PR #7379, opened 2021-12-10
docs: Make the k8s job log selector example more generic

terrytangyuan:
See https://stackoverflow.com/questions/70306051/argo-workflow-spark-operator-app-logs-not-generated


terrytangyuan:
cc @crenshaw-dev

docs/fields.md
```diff
@@ -180,9 +180,9 @@ Workflow is the definition of a workflow resource
 
 - [`input-artifact-s3.yaml`](https://github.com/argoproj/argo-workflows/blob/master/examples/input-artifact-s3.yaml)
 
-- [`k8s-jobs.yaml`](https://github.com/argoproj/argo-workflows/blob/master/examples/k8s-jobs.yaml)
+- [`k8s-jobs-log-selector.yaml`](https://github.com/argoproj/argo-workflows/blob/master/examples/k8s-jobs-log-selector.yaml)
```
crenshaw-dev, line 183: Maybe `k8s-resource-log-selector.yaml`? To me this is more about a feature of the "resource template" than about running jobs.

examples/k8s-jobs-log-selector.yaml
```diff
@@ -0,0 +1,47 @@
...
+        spec:
+          tfReplicaSpecs:
+             Worker:
+               replicas: 2
+               restartPolicy: Never
+               template:
+                 metadata:
+                   # We add this label to the pods created by TFJob custom resource to inform Argo Workflows
+                   # that we want to include the logs from the created pods. Once the pods are created with this
+                   # label, you can then use `argo logs -c tensorflow -l <logs-label-selector>` to the logs from
```
crenshaw-dev, line 35: Is the label selector necessary, or do you get the logs for free just because they're labeled as this workflow?

crenshaw-dev (commented):
I notice that some docs about success/failure conditions and about output parameters are left out of the new file. Is that intentional?

docs/fields.md
terrytangyuan: Sounds good.

examples/k8s-jobs-log-selector.yaml
terrytangyuan: Good catch. This is optional. I moved this to the second paragraph.

terrytangyuan:
> I notice that some docs about success/failure conditions and about output parameters are left out of the new file. Is that intentional?

Yes intentional as they are not necessary to illustrate log selector. Those are already part of other existing examples.

crenshaw-dev (approved):

crenshaw-dev:
Do you want to add a note about the earliest version in which this is (will be?) available?

terrytangyuan:
> Do you want to add a note about the earliest version in which this is (will be?) available?

Good idea. Added.

crenshaw-dev (approved):

terrytangyuan:
@sarabala1979 Could you take a look at this when you get a chance so that I can start pointing users to this example going forward?

sarabala1979 (approved):

Issue #7054, opened 2021-10-25
UI: make somewhat skinnable / themable


alexec:
# Summary

Allows users to skin the UI with new colours, their own loge, but keep Argo bits.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

krrrr38:
In my cases, it's enoght to change banner colour by environments.
It may be possible by argocd this feature https://argo-cd.readthedocs.io/en/stable/operator-manual/custom-styles/.

agilgur5:
See also #10825 for custom themes

PR #6282, opened 2021-07-05
feat(ui): Inject Custom CSS

dtaniwaki:
Checklist:

* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md).

Screen Shot 2021-07-05 at 12 41 58 PM

Fixes #1451

You can test this feature by executing the following commands.

```bash
$ kubectl -n argo create configmap custom-css --from-literal "custom.css=.nav-bar { background-color: #2196f3; }"
$ kubectl -n argo patch configmap workflow-controller-configmap -p '{"data":{"ui.cssurl":"./extra/custom.css"}}'
$ kubectl -n argo patch deployment argo-server -p '{"spec":{"template":{"spec":{"containers":[{"name":"argo-server","args":["server", "--secure=false", "--auth-mode=hybrid", "--extra-files-dir=/shared/app/custom"], "volumeMounts": [{"name": "custom-css", "mountPath": "/shared/app/custom"}]}], "volumes": [{"name":"custom-css", "configMap": {"name": "custom-css"}}]}}}}'

```

PR #7387, opened 2021-12-12
feat: customize nav bar background color

### LINKED ISSUE #1451: UI: Add ability to visually distinguish between deployment environments
**Is this a BUG REPORT or FEATURE REQUEST?**: **FEATURE REQUEST**

We run a dev environment and a production environment, which is fairly typical. We currently access the `argo-ui` through a proxy cmd, and I think it's easy enough to not specific the environment (dev/prod) properly and get confused as both look visually the same. 

What do you think about this problem? If/When we place domains in front of the UIs,  eg. `dev-argo-ui.company-ds.com` and `prod-argo-ui.company-ds.com` I think this issue will be alleviated, but not totally fixed as no element exists in the UI to display the deployment environment.

krrrr38:
Signed-off-by: krrrr38 <k.kaizu38@gmail.com>

## Motivation 

I want to distinguish multiple argo-workflows easily not to operate invalid operation on different environments.

Fixes #1451

## How to solve

- Make argo-ui NavBar customizable to change nav bar style
  - https://github.com/argoproj/argo-ui/pull/167
- Add navigation bar background color configuration into argo-workflow config map
- Pass background color info from `GetInfo` api

## Discussions

There are some discussions about this feature.
- [argo-cd way customize css impl on argo-workflows](https://github.com/argoproj/argo-workflows/pull/6282)
- argo-workflows should provide argo-workflow way.
  - [customize css sometimes breaks the ui, so provide argo-workflow `theme` and css maintained by argo](https://github.com/argoproj/argo-workflows/pull/6282#pullrequestreview-739570927)
- [More pluggable styles](https://github.com/argoproj/argo-workflows/issues/7054)

To archive my goal, I add nav bar color config for following reasons.

- [argo-ui v2 has `theme` feature which support only `dark` or `style`](https://github.com/argoproj/argo-ui/blob/49f257794525a5d6051bff109147b6133c7ee550/v2/shared/context/theme.tsx)
  - this theme feature doesn't enough to distinguish multiple environments
- When using `theme`, to distinguish multiple environments, we need to provides 3?, 4?,... themes. (dev, stg, prd, e2e, x division? x teams?,...)
  - I have no skills to create new design argo themes..., but color settings are available too many colors and easy to distinguish.
- pluggable ui and nav bar background color are something different.
  - pluggable ui provides new feature which provides **new component** in specific domains.
    - (if this is not, sorry 🙇 )
  - nav bar background color is **global settings** which is useful to distinguish multiple environments
- Not to customize too many css, just add a color setting.

So, I add nav bar background color as **global argo-workflow ui config**. I hope this `ui` config might be support icon, title, and so on.

## Check

https://github.com/argoproj/argo-workflows/pull/7387#issuecomment-1018985618


krrrr38:
CI / UI failed because this feature requires https://github.com/argoproj/argo-ui/pull/167.

pkg/apis/workflow/v1alpha1/info.go
```diff
@@ -11,3 +11,9 @@ type Link struct {
 	// The URL. Can contain "${metadata.namespace}", "${metadata.name}", "${status.startedAt}", "${status.finishedAt}" or any other element in workflow yaml, e.g. "${workflow.metadata.annotations.userDefinedKey}"
 	URL string `json:"url" protobuf:"bytes,3,opt,name=url"`
 }
+
+// Ui defines a config for UI
+type Ui struct {
+	// NavColor is an ui navigation bar theme which switch the design
+	NavColor string `json:"navColor,omitempty" protobuf:"bytes,1,opt,name=navColor"`
+}
```
simster7, line 19: If anything, wouldn't it be more powerful to allow a way to inject CSS overrides?

alexec (commented):
The use case for this are pretty simple, the ability to set colour differently for different environments. I think this is probably good enough for that. Previous solutions have been, these have been complex, I like the simplicity

pkg/apis/workflow/v1alpha1/info.go
```diff
@@ -11,3 +11,9 @@ type Link struct {
 	// The URL. Can contain "${metadata.namespace}", "${metadata.name}", "${status.startedAt}", "${status.finishedAt}" or any other element in workflow yaml, e.g. "${workflow.metadata.annotations.userDefinedKey}"
 	URL string `json:"url" protobuf:"bytes,3,opt,name=url"`
 }
+
+// Ui defines a config for UI
+type Ui struct {
```
alexec, line 16: should be `UI` ,i.e. upper-case

alexec (changes requested):

alexec (approved):

alexec (changes requested):
I don't think we actually need the "UI" wrapper, as we already have `links` top-level. Can just be `info.navColor`

krrrr38:
> I don't think we actually need the "UI" wrapper, as we already have `links` top-level. Can just be `info.navColor`

I also worry about this. I'll remove ui wrapper

pkg/apis/workflow/v1alpha1/info.go
krrrr38: Based on the description`Discussions`, I think this way is better.

krrrr38:
re-confirmed that nav bar color was changed by following config map

```
> kubectl get cm -n argo workflow-controller-configmap -ojsonpath='{.data.navColor}'
red
```

alexec (approved):

alexec (changes requested):
Please fix build.

krrrr38:
@alexec thank you for your review.
To build this PR, it requires argo-ui changes.
Can anyone review https://github.com/argoproj/argo-ui/pull/167?

alexec:
Up-stream PR now merged, you should. be able to use argo-ui version v2.5.0.

krrrr38:
I tried `yarn upgrade argo-ui@"https://github.com/argoproj/argo-ui.git"`, but there are some errors which depends on other library updates.
I will create another PR to upgrade argo-ui.

krrrr38:
now works fine again.

image

server/info/info_server_test.go
```diff
@@ -24,3 +25,31 @@ func Test_infoServer_GetUserInfo(t *testing.T) {
 		assert.Equal(t, "my-sa", info.ServiceAccountName)
 	}
 }
+
+func Test_infoServer_GetInfo(t *testing.T) {
```
alexec, line 29: nice

alexec (approved):
LGTM

krrrr38:
@alexec sorry, fixed golang-ci 😇 . CI succeeded.

Issue #7406, opened 2021-12-14
Using podSpecPatch and emissary together -> Pods stuck in running


Bobgy:
## Summary

What happened/what you expected to happen?

* I created a workflow that uses podSpecPatch to patch Pod command.
* I ran the workflow
* I noticed that this task with podSpecPatch is stuck in running forever.

I expect that workflows can run properly to the end.

What version of Argo Workflows are you running?

3.1.x

## Diagnostics

What executor are you running? Emissary

After debugging, I found that the pod's main container's command does not have the emissary executor prepended like in https://github.com/argoproj/argo-workflows/blob/3d47a5d29dee66775e6fa871dee1b6ca1ae6acda/workflow/controller/workflowpod.go#L331.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

Bobgy:
I think the fix is to move [the code that prepends /var/run/argo/argoexec emissary](https://github.com/argoproj/argo-workflows/blob/3d47a5d29dee66775e6fa871dee1b6ca1ae6acda/workflow/controller/workflowpod.go#L318-L335) to the bottom, so that it runs after pod spec patch has already been patched.

Bobgy:
Workaround: by adding "/var/run/argo/argoexec", "emissary", "--" in the podSpecPatch, I can get the workflow running. However, that makes the workflow coupled to emissary executor, so that's not ideal.

PR #7407, opened 2021-12-14
fix(controller): fix pod stuck in running when using podSpecPatch and emissary

Bobgy:
Fixes https://github.com/argoproj/argo-workflows/issues/7406
I tested via unit test.

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -315,10 +315,6 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 		c.VolumeMounts = append(c.VolumeMounts, volumeMountVarArgo)
 		pod.Spec.InitContainers[i] = c
 	}
-	for i, c := range pod.Spec.Containers {
-		c.VolumeMounts = append(c.VolumeMounts, volumeMountVarArgo)
-		pod.Spec.Containers[i] = c
-	}
 
 	// Add standard environment variables, making pod spec larger
 	envVars := []apiv1.EnvVar{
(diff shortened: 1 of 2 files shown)


Bobgy:
I might need some help understanding the test failure: https://github.com/argoproj/argo-workflows/runs/4517399752?check_suite_focus=true

The error message is not very intuitive:

```
    agent_test.go:83: 
        	Error Trace:	agent_test.go:83
        	            				then.go:66
        	            				then.go:41
        	            				agent_test.go:64
        	Error:      	Should be true
        	Test:       	TestAgentSuite/TestParallel
=== FAIL: AgentSuite/TestParallel
FAIL	github.com/argoproj/argo-workflows/v3/test/e2e	13.146s
FAIL
make: *** [Makefile:490: test-functional] Error 1
Error: Process completed with exit code 2.
```

Bobgy:
OK, after a new push, all tests are passing!

alexec (changes requested):
failing test due to new bug, please fix

Bobgy:
Hmm, I guess I fixed the failure.
However, I believe my change makes the test https://github.com/argoproj/argo-workflows/blob/master/test/e2e/testdata/sidecar-injected-kill-annotation-workflow.yaml useless, because now podSpecPatch is no longer a good way to simulate injected sidecars.

Any suggestions?

alexec (dismissed):

alexec:
> However, I believe my change makes the test https://github.com/argoproj/argo-workflows/blob/master/test/e2e/testdata/sidecar-injected-kill-annotation-workflow.yaml useless, because now podSpecPatch is no longer a good way to simulate injected sidecars.

This is as important test, do you have any other suggestions?

Bobgy:
I understand it's very complex to set up a webhook for e2e tests, so the best solution I came up with is adding a special annotation for test purposes only, which will have equivalent behavior of pod spec patch, but applied at the very end.

What do you think? Any easier ways to do this?

alexec:
Just refreshing, we need to apply the patch before we modify the command to insert the emissary. Otherwise, the patch mangles the command - removing emissary binary from it and preventing it from running. 

But, we use patch in the test. 

So clearly your code change is correct, and the test needs fixing. But question is - how?

alexec:
Coming back to this. Can you skip the tests in the PR? Create a new tech debt issue to fix the tests.

Bobgy:
Thank you for the flexibility!
I created https://github.com/argoproj/argo-workflows/issues/7422 to track the tech debt.

alexec:
Please sync with master.

Bobgy:
Everything ready : )

Bobgy:
Hi @alexec, is there anything else you want for this PR?

alexec (approved):

PR #7419, opened 2021-12-15
fix: Disable SDK release from master branch

terrytangyuan:
Currently this is also running on master branch, which would fail the CI. Example: https://github.com/argoproj/argo-workflows/runs/4534564425?check_suite_focus=true

### DIFF
--- a/.github/workflows/sdks.yaml
+++ b/.github/workflows/sdks.yaml
@@ -3,8 +3,6 @@ on:
   push:
     tags:
       - v*
-    branches:
-      - master
 jobs:
   sdk:
     if: github.repository == 'argoproj/argo-workflows'


flaviuvadan (approved):

alexec:
What if I want to test a tip version of SDK?

terrytangyuan:
> What if I want to test a tip version of SDK?

Not sure if you can do something similar for Java but for Python you can use `pip install git+https://github.com/argoproj/argo-workflows@master#subdirectory=sdks/python/client` which installs from the master branch.

alexec:
I think you could just build the Java version locally if you wanted tip of master. I think that would be fine solution.

alexec (approved):

PR #7420, opened 2021-12-15
fix: Increase padding between elements in workflow template creator. …

### LINKED ISSUE #7309: UI: Not intuitive parameter helper text when submitting workflows
# Summary

The helper text of text inputs in "submit workflow" UI are not intuitive. The user does not know which parameter is being edited. See example for a better understanding:

obraz

# Use Cases

Everytime the user is submitting the workflow, and making necessary changes in the parameters, he should be sure what parameter is edited exactly. The user should not be worried about editing incorrect parameter due to poor UI.

---

lstolcman:
Fixes #7309

Before:
before

After:
after

@simster7


sarabala1979:
@lstolcman I didn't see any difference in the above picture. Can you highlight it? Please fix the lint  error

lstolcman:
A wider gap to easily distinguish which text imput belongs to what item (explained well in https://github.com/argoproj/argo-workflows/issues/7309).
obraz

Also, there's no linting errors, jsut some weird item:
obraz

sarabala1979:
@lstolcman thanks for screenshot. Please fix the lint failure.

PR #7426, opened 2021-12-16
fix: Page workflow event bindings basehref

### LINKED ISSUE #7416: workflow-event-bindings page support basehref
# Summary

image

If the `basehref` is set, the placeholder address of the `workflow-event-bindings` page should also be added synchronously

Similarly, the Swagger-API docs needs to be handled

---

chf007:
Fixes #7416 

- Page workflow event bindings curl example support basehref.
- Enhance the robustness of the code.
  - export shared/base baseUrl 
  - base.getAttribute('href') may return null (`<base target="_blank">`)
  -  userinfo/components/cli-help.tsx argoBaseHref may report undefined error

### DIFF
--- a/ui/src/app/shared/base.ts
+++ b/ui/src/app/shared/base.ts
@@ -1,6 +1,6 @@
-function baseUrl(): string {
+export function baseUrl(): string {
     const base = document.querySelector('base');
-    return base ? base.getAttribute('href') : '/';
+    return base && base.getAttribute('href') ? base.getAttribute('href') : '/';
 }
 
 export function uiUrl(uiPath: string): string {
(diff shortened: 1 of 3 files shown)


sarabala1979:
have you tested? please. provide the screen shot and examples

chf007:
> have you tested? please. provide the screen shot and examples

I started it in the local environment and tested it through basepath=/.

Since argo-server can set basepath=xxx when it is started locally, it does not actually take effect, so the case of basepath=xxx has not been measured.

The following is a description of my local environment:

minikube

`export BASE_HREF=/my-path/ && make start API=true UI=true`

image

image

sarabala1979:
@simster7 Can you take a quick look for this PR?

sarabala1979:
@chf007 Please  fix  the lint

sarabala1979:
@chf007 Do you want to merge this  PR for v3.3?

sarabala1979 (changes requested):
Changes  requested

agilgur5:
This was replaced by #10759

Issue #7423, opened 2021-12-16
Fix Python SDK versioning to work with existing PyPI releases


terrytangyuan:
## Summary

What happened/what you expected to happen?

Currently, we are using the tag, e.g. 3.2.5, as the Python SDK versions. However, since we will be publishing to the existing package https://pypi.org/project/argo-workflows/ where `argo-workflows==5.0.0` exists, we have to update the versioning scheme for Python SDK. 

One solution is to use tag's major version + 3, e.g. 
GitHub tag -> PyPI package version
3.2.5 -> 6.2.5
3.3.5 -> 6.3.5
4.1 -> 7.1

Since we missed the boat for 3.2.5, let's target this for the next release.

cc @flaviuvadan 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

PR #7429, opened 2021-12-16
feat: Add Python SDK versioning script

flaviuvadan:
When the Python SDK was migrated to the Argo Workflows repository, we kept the `argo-workflows` name for the Python
package as we wanted to publish it to the same PyPi project on the public index. However, the existing `argo-workflows`
package was on version 5.0.0 already, while Argo Workflows was on 3.x.x. We wanted to publish the SDK as version 6.0.0
to indicate backwards incompatibility. So, this script takes the Argo Workflows tag, when a new release is created,
takes the major version, adds 3 to it, and prints to stdout the new version, which is:
- ARGO_MAJOR+3.ARGO_MINOR.ARGO_PATCH

Fixes: #7423 

Functional testing:
```
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ ./sdk_version.py                                                                                                                                                             
untagged
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ git tag -a v7.0.0 -m "tag 7 test"                                                                                                                                           
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ ./sdk_version.py                                                                                                                                                              
10.0.0
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ git tag -a v7.1.0 -m "tag 7 test"                                                                                                                                             
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ ./sdk_version.py                                                                                                                                                              
10.1.0
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ git tag -a v7.1.3 -m "tag 7 test"                                                                                                                                             
((python) ) ➜  python git:(fv/py-sdk-versioning) ✗ ./sdk_version.py                                                                                                                                                              
10.1.3
```

CC: @terrytangyuan

### DIFF
--- a/sdks/python/Makefile
+++ b/sdks/python/Makefile
@@ -1,5 +1,7 @@
-# SDK version
 VERSION := $(shell ./sdk_version.py)
+ifndef VERSION
+  $(error "Failed to obtain a valid version for the SDK")
+endif
 
 # work dir
 WD := $(shell echo "`pwd`/client")
@@ -7,10 +9,7 @@ WD := $(shell echo "`pwd`/client")
 DOCKER = docker run --rm -v $(WD):/wd --workdir /wd
 CHOWN = chown -R $(shell id -u):$(shell id -g)
 
-verify_version:
-	@if [ ${VERSION} == "FAILED" ]; then echo "Failed to obtain a valid version from the SDK versioning script! Version: ${VERSION}"; exit 1; fi
-
-publish: verify_version generate
+publish: generate
 	pip install setuptools twine build
 	python -m build --sdist --wheel --outdir client/dist/ client
 	twine check client/dist/*
--- a/sdks/python/Makefile
+++ b/sdks/python/Makefile
@@ -1,13 +1,8 @@
-GIT_TAG               := $(shell git describe --exact-match --tags --abbrev=0 2> /dev/null || echo untagged)
-ifeq ($(GIT_TAG),untagged)
-VERSION               := 0.0.0-latest
-else
-# remove the "v" prefix, not allowed
-VERSION               := $(subst v,,$(GIT_TAG))
-endif
+# SDK version
+VERSION := $(shell ./sdk_version.py)
 
 # work dir
-WD                    := $(shell echo "`pwd`/client")
+WD := $(shell echo "`pwd`/client")
 
 DOCKER = docker run --rm -v $(WD):/wd --workdir /wd
 CHOWN = chown -R $(shell id -u):$(shell id -g)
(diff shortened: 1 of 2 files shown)


sdks/python/sdk_version.py
```diff
@@ -0,0 +1,31 @@
...
+git_tag = os.popen(git_tag_cmd).read().strip()
+if git_tag == UNTAGGED:
+    print(UNTAGGED_VERSION)  # this goes to sys.stdout, so it's captured by the Makefile
+    sys.exit(0)
+
+version_digits = [int(i) for i in git_tag.replace(VERSION_PREFIX, '').split('.')]
+version_digits[MAJOR_VERSION_INDEX] += VERSION_INCREMENT
+
+version = '.'.join([str(i) for i in version_digits])
+print(version)
```
terrytangyuan, line 31: Would this be unreliable if any of these lines fail and an exception is printed out that becomes `VERSION`?

flaviuvadan: Added some exception handling and a `FAILED` indicator so that publication fails in CICD if the version is incorrect.

sdks/python/sdk_version.py
```diff
@@ -0,0 +1,50 @@
...
+UNTAGGED = 'untagged'
+FAILED = 'FAILED'  # indicator captured by the makefile to know when something failed
+UNTAGGED_VERSION = '0.0.0-latest'
+
+git_tag_cmd = 'git describe --exact-match --tags --abbrev=0 2> /dev/null || echo untagged'
+
+git_tag = None
+try:
+    git_tag = os.popen(git_tag_cmd).read().strip()
+except ValueError:
```
terrytangyuan, line 26: What about other types of exceptions?

sdks/python/Makefile
```diff
@@ -1,18 +1,16 @@
...
 # work dir
-WD                    := $(shell echo "`pwd`/client")
+WD := $(shell echo "`pwd`/client")
 
 DOCKER = docker run --rm -v $(WD):/wd --workdir /wd
 CHOWN = chown -R $(shell id -u):$(shell id -g)
 
-publish: generate
+verify_version:
+	@if [ ${VERSION} == "FAILED" ]; then echo "Failed to obtain a valid version from the SDK versioning script! Version: ${VERSION}"; exit 1; fi
```
terrytangyuan, line 11: This "FAILED" might not be needed. We can use `$?` to obtain the exit status of the python script (any exceptions will be non-zero exit). We can also make a big try-catch block and `os.exit(1)` expliclty during exceptions.

sdks/python/sdk_version.py
flaviuvadan: I think we can go with what you suggested: a big try/catch and capture `Exception` rather than independent ones.

sdks/python/Makefile
flaviuvadan: Ended up using a `ifndef` block.

terrytangyuan (approved):

Issue #7436, opened 2021-12-17
I would like a mentor


softwarebloat:
I would like to start contributing to argo workflows but i'm new to Golang and have no experience with this codebase.

I was looking at `good first issue`, in particular the #7239 but wasn't able to reproduce it locally. But probably, even if i reproduced it, i wouldn't know what to do to solve it

> Any `good first issue` would be cool if helps me grasping the code base though

alexec:
@sarabala1979 @terrytangyuan @simster7 I'm about to go OOV for 1 month. Could I please ask if you would be interested?

terrytangyuan:
Yes, I can help to mentor this.

terrytangyuan:
@softwarebloat Could you tell me a bit about what you have tried to reproduce? When I ran `argo submit https://raw.githubusercontent.com/argoproj/argo-workflows/master/examples/http-hello-world.yaml --watch`, I observed the `0/0` progress. Could you try that? 

If you'd like to work on #7239, one potential cause of this issue is that we currently do not support reading progress from HTTP type node. Some relevant code: https://github.com/argoproj/argo-workflows/blob/master/workflow/progress/updater.go. I haven't looked into details yet but you could start understanding the code relevant to progress updating and understand  its relation to the implementation of HTTP template https://github.com/argoproj/argo-workflows/pull/5750.

softwarebloat:
@terrytangyuan you are right, i can reproduce it (i was using a wrong workflow example).

The **updater.go** file was really helpful thanks! I also check the HTTP template PR (#5750) but was really big, more difficult to understand everything :sweat_smile: 
but the problem seems to be in the `updater.go`

I wrote a test and also think i fixed the bug but, while the test passed, if i submit the workflow i still see `progress 0/0`.
Is `make start API=true` enough? i also tried to run `make clean` and `make build` before but nothing changed.

And also, if i try to run `make pre-commit -B` i get the following
```
make pre-commit -B
GIT_COMMIT=0f567ffe2ab2d81fe308901d13acdac954323405 GIT_BRANCH=fix/http-templates-progress GIT_TAG=untagged GIT_TREE_STATE=dirty RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=k3d-argo-workflow DOCKER_DESKTOP=false K3D=true DOCKER_PUSH=false
[...]

PR #7450, opened 2021-12-21
fix: http template doesn't update progress. Fixes #7239

### LINKED ISSUE #7239: HTTP template always reports progress as zero
It should be 0/1 when pending/running and 1/1 when complete.

image

softwarebloat:
To me the PR is ready but i'm not able to try it locally as stated in #7436 
That's why i left it as Draft

### DIFF
--- a/workflow/progress/updater.go
+++ b/workflow/progress/updater.go
@@ -31,7 +31,7 @@ func PodProgress(pod *apiv1.Pod, node *wfv1.NodeStatus) wfv1.Progress {
 func UpdateProgress(wf *wfv1.Workflow) {
 	wf.Status.Progress = "0/0"
 	for _, node := range wf.Status.Nodes {
-		if node.Type != wfv1.NodeTypePod {
+		if node.Type != wfv1.NodeTypePod && node.Type != wfv1.NodeTypeHTTP {
 			continue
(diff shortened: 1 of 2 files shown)


sarabala1979:
@softwarebloat is it ready to review?

softwarebloat:
@sarabala1979 yes it is. only thing is that i wasn't able to try it locally but, if the test i wrote is correct, it should work :crossed_fingers:

sarabala1979:
> @sarabala1979 yes it is. only thing is that I wasn't able to try it locally but, if the test I wrote is correct, it should work 🤞

Sure, have you tried `make start`? I will try it on my env

softwarebloat:
> > @sarabala1979 yes it is. only thing is that I wasn't able to try it locally but, if the test I wrote is correct, it should work 🤞
> 
> Sure, have you tried `make start`? I will try it on my env

Yes I tried `make start` but seems like my fix has no effect. So I don't understand if it's not loaded or not working.
Another strange thing is that I don't have the controller pod in my cluster 🤷‍♂️

simster7 (approved):

softwarebloat:
@sarabala1979 were you able to test it locally?

PR #7474, opened 2021-12-30
chore: fix typo in docs

pquadri:
There was a reference in `variables.md` to the `when` key in retries, but the key is actually `expression`.


sarabala1979 (approved):

sarabala1979:
@pquadri Please fix the  lint error

pquadri:
I tried merging changes from master since I actually didn't make any change that should impact that linting (if I'm understanding correctly what it's supposed to check :D )

PR #7501, opened 2022-01-04
fix: update old buildkit version in buildkit-template.yaml

cod-r:
I spent a few hours trying to figure out why the example in buildkit-template.yaml was not working with a private insecure docker registry.
The problem was a bug in the moby/buildkit version used in this example. I updated the version so other people waste time like I did.


terrytangyuan (changes requested):
Please fix the DCO check. It would also be great to link to the issue in the older version of buildkit.

cod-r:
> Please fix the checks. It would also be great to link to the issue in the older version of buildkit.

This is the pr that solved the insecure registry issue: https://github.com/moby/moby/pull/40814
I can't figure out in what version of moby/buildkit was this introduced though, but it was definitely not working in 0.7.2 and working in 0.9.3

Issue #6807, opened 2021-09-27
`generateName` with customized suffix length


treeguard:
## Summary

Currently the workflows generated by workflowTemplate where the `generateName` is defined will have a suffix which is a random string of length 5, see https://github.com/argoproj/argo-workflows/blob/e5b131a333afac0ed3444b70e2fe846b86dc63e1/workflow/util/util.go#L613 this has brought up workflow name conflicts in our use case. It would be great if the length of the random string can be customized. 

What change needs making?
Add one more parameter in the workflow definition, for example, `generateNameSuffixLength` so that it can be customized. 

## Use Cases

When would you use this?

ASAP, we are now trying to override the `name` field when submitting the workflow as a workaround now. 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

tooptoop4:
@treeguard how do u handle this for cronworkflows?

LinuxSuRen:
I'm wondering if it's ok to let Kubernetes generate the name. I mean just provide the `generateName` instead of the `name` field.

tooptoop4:
@LinuxSuRen generateName only provides 5 random characters which is not unique, we facing collisions

LinuxSuRen:
> @LinuxSuRen generateName only provides 5 random characters which is not unique, we facing collisions

That is the limit of Argo workflow. I'm talking about letting Kubernetes to generate it.

agilgur5:
https://github.com/argoproj/argo-workflows/issues/7900 seems like a more generic option that would supersede this.

PR #7554, opened 2022-01-13
fix: longer random string suffix on workflow name to avoid conflict

tooptoop4:
fixes  https://github.com/argoproj/argo-workflows/issues/6806 / https://github.com/argoproj/argo-workflows/issues/6807

Create your PR as a draft.

  does not need to pass.
* Say how how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* You can ask for help!

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -600,7 +600,7 @@ func randString(n int) string {
 
 // RandSuffix generates a random suffix suitable for suffixing resource name.
 func RandSuffix() string {
-	return randString(5)
+	return randString(10)
 }
 
 // FormulateResubmitWorkflow formulate a new workflow from a previous workflow, optionally re-using successful nodes


workflow/util/util.go
```diff
@@ -600,7 +600,7 @@ func randString(n int) string {
 
 // RandSuffix generates a random suffix suitable for suffixing resource name.
 func RandSuffix() string {
-	return randString(5)
+	return randString(10)
```
terrytangyuan, line 603: This will break some of the existing workflows where some names (workflow pods) might exceed the maximum length.

tooptoop4: would u merge if I put a if check on the max length?

terrytangyuan: There is already a check on the length. The problem with this change is that people's workflows will break unexpectedly if the length is already close to the limit.

tooptoop4:
maybe can pad length here https://github.com/argoproj/argo-workflows/blob/09987a6dd03c1119fa286ed55cc97a2f4e588e09/workflow/creator/creator.go#L31-L33 with random suffix

terrytangyuan (changes requested):
The linked issues have not had enough interest so this seems to be a rare case. I'd recommend closing this PR until we observe that more people are having issues. One workaround is to provide your own name to the workflow instead of using generateName.

Issue #7557, opened 2022-01-13
Include the timezone in argo cron list output


bobh66:
# Summary

It would be helpful if the output of the argo cron list command included the timezone along with the schedule, as it is impossible to correctly interpret the schedule without knowing what timezone it is using.

# Use Cases

This is useful any time the user requests a list of existing cronworkflow objects

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
Looks like OP contributed and fixed this themselves in #7559

PR #7559, opened 2022-01-13
feat: Adds timezone to argo cron list output (#7557)

bobh66:
Signed-off-by: Bob Haddleton <bob.haddleton@nokia.com>

* [X] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

Added the timezone to the output of argo cron list):

```
NAME                         AGE   LAST RUN   NEXT RUN   SCHEDULE                           TIMEZONE     SUSPENDED
my-cron-workflow-c90cadee    20h   2h         14h        45 12,13,14,15,16,17,18,19 * * *                false
my-cron-workflow2-f4f27c90   20h   20h        26s        * * * * *                          US/Central   true
```

Note that the timezone field is empty when there is no timezone specified in the CronWorkflow object.  Since the cron server appears to default to UTC when no timezone is specified (at least as I understand that time.Now() returns the local time in UTC), would it make sense to print UTC when the timezone is null?

There do not appear to be any tests for this code.

### DIFF
--- a/cmd/argo/commands/cron/list.go
+++ b/cmd/argo/commands/cron/list.go
@@ -66,7 +66,7 @@ func printTable(wfList []wfv1.CronWorkflow, listArgs *listFlags) {
 	if listArgs.allNamespaces {
 		_, _ = fmt.Fprint(w, "NAMESPACE\t")
 	}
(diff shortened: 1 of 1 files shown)


sarabala1979 (approved):
LGTM

PR #7582, opened 2022-01-19
feat: lifecycle hook

sarabala1979:
hook

* [ ] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

  does not need to pass.
* Say how how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* You can ask for help!


examples/life-cycle-hooks.yaml
```diff
@@ -0,0 +1,96 @@
+    apiVersion: argoproj.io/v1alpha1
```
alexec, line 1: this is a long example, and I think it makes it hard to understand how this works

examples/life-cycle-hooks.yaml
```diff
@@ -0,0 +1,96 @@
...
+    kind: Workflow
+    metadata:
+      generateName: lifecycle-hook-
+    spec:
+      entrypoint: diamond
+      hooks:
+        exit:
+          template: http
+        running:
+          expression: "{{workflow.status}} == Running"
```
alexec, line 11: I strongly believe we should be avoiding `{{` if we can. Instead, could this be `workflow.status == 'Running'` instead.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -373,6 +373,24 @@ type WorkflowSpec struct {
 
 	// ArchiveLogs indicates if the container logs should be archived
 	ArchiveLogs *bool `json:"archiveLogs,omitempty" protobuf:"varint,40,opt,name=archiveLogs"`
+
+	// Hooks holds the lifecycle hook which is invoked at lifecycle of
+	// step, irrespective of the success, failure, or error status of the primary step
+	Hooks LifecycleHooks `json:"hooks,omitempty" protobuf:"bytes,41,opt,name=hooks"`
+}
+
+func (wfs *WorkflowSpec) GetExitHook(args Arguments) *LifecycleHook {
```
alexec, line 382: smart

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -373,6 +373,24 @@ type WorkflowSpec struct {
...
+		return nil
+	}
+	if wfs.OnExit != "" {
+		return &LifecycleHook{Template: wfs.OnExit, Arguments: args}
+	}
+	return wfs.Hooks.GetExitHook().WithArgs(args)
+}
+
+func (wfs *WorkflowSpec) HasExitHook() bool {
+	return (wfs.Hooks != nil && wfs.Hooks.GetExitHook() != nil) || wfs.OnExit != ""
```
alexec, line 393: minor- as Hooks is a map, and therefore a pointer, you could make Hooks.GetExitHook do a nil check

workflow/common/util.go
```diff
@@ -360,3 +360,7 @@ func IsDone(un *unstructured.Unstructured) bool {
 		un.GetLabels()[LabelKeyCompleted] == "true" &&
 		un.GetLabels()[LabelKeyWorkflowArchivingStatus] != "Pending"
 }
+
+func GenerateLifeHookNodeName(parentNodeName string, hookName string) string {
```
alexec, line 364: please avoid adding to this file, and move to own file, e.g. hooks.go

workflow/controller/dag.go
```diff
@@ -247,6 +247,19 @@ func (woc *wfOperationCtx) executeDAG(ctx context.Context, nodeName string, tmpl
...
 		taskNode := dagCtx.getTaskNode(taskName)
+
+		if taskNode != nil {
+			task := dagCtx.GetTask(taskName)
+			scope, err := woc.buildLocalScopeFromTask(dagCtx, task)
+			if err != nil {
+				woc.markNodeError(node.Name, err)
+			}
+			scope.addParamToScope(fmt.Sprintf("tasks.%s.status", task.Name), string(node.Phase))
+			err = woc.executeLifeCycleHook(ctx, scope, dagCtx.GetTask(taskName).Hooks, taskNode, dagCtx.boundaryID, dagCtx.tmplCtx, "tasks."+taskName)
```
alexec, line 258: scope will be nil because error fall through, should we exit on line 255?

workflow/controller/dag.go
```diff
@@ -358,6 +371,17 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
 			woc.controller.syncManager.Release(woc.wf, node.ID, tmpl.Synchronization)
 		}
 
+		task := dagCtx.GetTask(taskName)
+		scope, err := woc.buildLocalScopeFromTask(dagCtx, task)
+		if err != nil {
+			woc.markNodeError(node.Name, err)
```
alexec, line 377: should we return here?

workflow/controller/dag.go
```diff
@@ -247,6 +247,19 @@ func (woc *wfOperationCtx) executeDAG(ctx context.Context, nodeName string, tmpl
 		// complete (since the DAG itself will have succeeded). To ensure that their exit handlers are run we also run them here. Note that
 		// calls to runOnExitNode are idempotent: it is fine if they are called more than once for the same task.
 		taskNode := dagCtx.getTaskNode(taskName)
+
+		if taskNode != nil {
+			task := dagCtx.GetTask(taskName)
+			scope, err := woc.buildLocalScopeFromTask(dagCtx, task)
+			if err != nil {
+				woc.markNodeError(node.Name, err)
```
alexec, line 255: should we return here?

workflow/controller/hooks.go
```diff
@@ -0,0 +1,109 @@
+package controller
```
alexec, line 1: own file - nice

workflow/controller/hooks.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
+	"github.com/argoproj/argo-workflows/v3/workflow/templateresolution"
+)
+
+func IsExitHook(hookName wfv1.LifecycleEvent) bool {
+	return hookName == wfv1.ExitLifecycleEvent
+}
+
+func (woc *wfOperationCtx) executeWfLifeCycleHook(ctx context.Context, tmplCtx *templateresolution.Context) error {
+	if woc.wf.Spec.Hooks != nil && len(woc.wf.Spec.Hooks) > 0 {
```
alexec, line 19: minor - len(nil) == 0, you don't need nil check on this line,

moreover, as you're looping on 20, you don't need this guard at all `range nil` does not panic

workflow/controller/hooks.go
```diff
@@ -0,0 +1,109 @@
...
+		for hookName, hook := range woc.wf.Spec.Hooks {
+			//exit hook will be execute in runOnExitNode
+			if IsExitHook(hookName) {
+				continue
+			}
+			// Replace hook's parameters
+			hookBytes, err := json.Marshal(hook)
+			if err != nil {
+				return errors.InternalWrapError(err)
```
alexec, line 28: is there a strong reason to wrap like this? would it be good to use `fmt.Errorf` and add some diagnostics?

workflow/controller/hooks.go
```diff
@@ -0,0 +1,109 @@
...
+				return err
+			}
+			if execute {
+				hookNodeName := common.GenerateLifeHookNodeName(woc.wf.ObjectMeta.Name, string(hookName))
+				woc.log.WithField("lifeCycleHook", hookName).WithField("node", hookNodeName).Infof("Running workflow level hooks")
+				_, err := woc.executeTemplate(ctx, hookNodeName, &wfv1.WorkflowStep{Template: hook.Template}, tmplCtx, hook.Arguments, &executeTemplateOpts{})
+				if err != nil {
+					return err
+				}
+				woc.addChildNode(woc.wf.ObjectMeta.Name, hookNodeName)
```
alexec, line 50: minor - `ObjectMeta.` should not be needed

workflow/controller/hooks.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func (woc *wfOperationCtx) executeLifeCycleHook(ctx context.Context, scope *wfScope, lifeCycleHooks wfv1.LifecycleHooks, parentNode *wfv1.NodeStatus, boundaryID string, tmplCtx *templateresolution.Context, prefix string) error {
+
+	if lifeCycleHooks == nil {
```
alexec, line 59: guard not needed, you'll range on a nil, which is fine

workflow/controller/hooks.go
```diff
@@ -0,0 +1,109 @@
...
+					return err
+				}
+				woc.addChildNode(woc.wf.ObjectMeta.Name, hookNodeName)
+			}
+		}
+	}
+	return nil
+}
+
+func (woc *wfOperationCtx) executeLifeCycleHook(ctx context.Context, scope *wfScope, lifeCycleHooks wfv1.LifecycleHooks, parentNode *wfv1.NodeStatus, boundaryID string, tmplCtx *templateresolution.Context, prefix string) error {
```
alexec, line 57: Do you need two similar funds? Could one of the funds be enough?

workflow/controller/operator.go
```diff
@@ -1634,6 +1648,9 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 
 	// Inputs has been processed with arguments already, so pass empty arguments.
 	processedTmpl, err := common.ProcessArgs(resolvedTmpl, &args, woc.globalParams, localParams, false, woc.wf.Namespace, woc.controller.configMapInformer)
+	if errorsutil.IsTransientErr(err) {
```
alexec, line 1651: this _could_ come in a separate PR, but I don't feel strongly

alexec (changes requested):
Mostly minor comments, e.g. some code that can be simpler and one question about refactoring. 

Do not use `{{}}` syntax. It confuses users. We have bugs in that code that are hard to fix. We're trying to move away from it. Instead, use the "expr" style.

workflow/controller/hooks.go
sarabala1979: yes, one is workflow level and another one is template level. The template level has more logic to find the parent node.
Want to keep simple and small functions

pkg/apis/workflow/v1alpha1/workflow_types.go
alexec, line 382: do you have unit tests for this?

workflow/controller/dag.go
```diff
@@ -358,6 +373,17 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
...
 
+		task := dagCtx.GetTask(taskName)
+		scope, err := woc.buildLocalScopeFromTask(dagCtx, task)
+		if err != nil {
+			woc.markNodeError(node.Name, err)
+		}
+		scope.addParamToScope(fmt.Sprintf("tasks.%s.status", task.Name), string(node.Phase))
+		err = woc.executeTmplLifeCycleHook(ctx, scope, dagCtx.GetTask(taskName).Hooks, node, dagCtx.boundaryID, dagCtx.tmplCtx, "tasks."+taskName)
+		if err != nil {
+			woc.markNodeError(node.Name, err)
```
alexec, line 384: do you need to return here?

workflow/controller/hooks.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (woc *wfOperationCtx) executeTmplLifeCycleHook(ctx context.Context, scope *wfScope, lifeCycleHooks wfv1.LifecycleHooks, parentNode *wfv1.NodeStatus, boundaryID string, tmplCtx *templateresolution.Context, prefix string) error {
+
+	for hookName, hook := range lifeCycleHooks {
+		//exit hook will be execute in runOnExitNode
+		if IsExitHook(hookName) {
+			continue
+		}
+		execute, err := shouldExecuteHook(hook.Expression, woc.globalParams.Merge(scope.getParameters()))
```
alexec, line 50: I'm not a fan of passing parameters and then doing some sed on the names. I think we should pass the expr env explicitly as a map, e.g.

```golang
execute, err := shouldExecuteHook(hook.Expression, map[string]interface{} {
"step": map[string]interface{} {
  "status": ...
}
})
```

alexec (commented):
Continuing on the expressions, we should not lean on any parameters code at all.

Instead, explicitly create an expr env with the variables you need.

alexec (commented):
I think we could probably do with a few more tests too.

sarabala1979:
> I think we could probably do with a few more tests too.

added test

workflow/controller/dag.go
sarabala1979: No Exit handle need to run

pkg/apis/workflow/v1alpha1/workflow_types.go
sarabala1979: added

sarabala1979:
> I think we could probably do with a few more tests too.

added test

workflow/controller/hooks.go
```diff
@@ -0,0 +1,94 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func getEnv(parameters common.Parameters) map[string]interface{} {
+	params := make(map[string]interface{})
+	for key, value := range parameters {
+		params[strings.Replace(key, "-", "_", -1)] = value
```
alexec, line 83: this is the line of code I'd like to see gone

alexec (commented):
I'm going to push you harder to get rid of parameters in this. It's still translating from the old paradigm (parameters + templating) to new paradigm (expr + env). This hides the translation and could allow new bugs to appear. 

Instead, create a new env explicitly.

We could pair program this if you like.

workflow/controller/hooks.go
```diff
@@ -0,0 +1,99 @@
...
+
+	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+	"github.com/argoproj/argo-workflows/v3/util/expr/argoexpr"
+	"github.com/argoproj/argo-workflows/v3/util/expr/env"
+	"github.com/argoproj/argo-workflows/v3/util/template"
+	"github.com/argoproj/argo-workflows/v3/workflow/templateresolution"
+)
+
+func IsExitHook(hookName wfv1.LifecycleEvent) bool {
+	return hookName == wfv1.ExitLifecycleEvent
```
alexec, line 15: minor - I think you should just inline this func

workflow/controller/hooks.go
```diff
@@ -0,0 +1,99 @@
...
+		if !hookNode.Fulfilled() {
+			completed = false
+			break
+		}
+	}
+	return completed, nil
+}
+
+func shouldExecuteHook(expression string, env map[string]interface{}) (bool, error) {
+	return argoexpr.EvalBool(expression, env)
```
alexec, line 94: minor - I think you should inline

alexec (commented):
I think we need to have more tests, including tests that use `WorkflowTemplate` and `workflowTemplateRef` and `templateRef`

sarabala1979:
> I think we need to have more tests, including tests that use `WorkflowTemplate` and `workflowTemplateRef` and `templateRef`

Thanks  for identifying. I tested  all `templateRef` and `workflowTemplateRef` scenario and added unit  test

image

sarabala1979:
@alexec can you take a quick look ? I  addressed all your comments and added testcases for WFT sceanrios. I tested manually also.

alexec (approved):
I'm going to approve this and open two new tickets for more tests and for docs.

PR #7594, opened 2022-01-20
fix: Fix inconsistent ordering of workflows with the list command. 

### LINKED ISSUE #7581: argo CLI list command results not predictable with -o name flag
## Summary

Using argo CLI I want keep the latest 20 workflows and delete the rest. This runs in a cron job (not argo cron job, normal linux cron)

When I do this
```
argo list -n argo --no-headers  | tail -n 6
patch-generator-basic-c2pc5      Succeeded   15d   3m   0
patch-generator-basic-vxj92      Succeeded   15d   3m   0
patch-generator-basic-6db47      Succeeded   15d   2m   0
patch-generator-basic-wcwkb      Succeeded   19d   2m   0
patch-generator-basic-tgpqf      Succeeded   20d   2m   0
patch-generator-basic-ldsrb      Succeeded   20d   2m   0
```
and when I do it with -o name, I get different result
```
argo list -n argo --no-headers -o name | tail -n 6
patch-generator-basic-6db47
patch-generator-basic-ccklj
patch-generator-basic-tgpqf
patch-generator-basic-mc2bp
patch-generator-advanced-rm76g
patch-generator-basic-vtknt
```
As you can see both return a different list.

Is there a way to sort the output of `argo list` in order of creation time? Why ordering is different when `-o name` is used.

Version: v3.2.4
Kubernetes: v1.21.2
Docker: 20.10.7

dinever:
Fixes #7581

Hi, this PR fixes an issue where `argo list -o name` has inconsistent display order than `argo list`. The root cause is that we sort workflows based on `CreationTimestamp` and `FinishedAt`:

https://github.com/argoproj/argo-workflows/blob/f1fe3bee498ac7eb895af6f89a0eba5095410467/pkg/apis/workflow/v1alpha1/workflow_types.go#L130-L145

So we need to make sure these two fields are included in the name-only field selectors as well, in order to have a consistent sort.

### Before
Workflows have inconsistent display order with / without `-o name`
Screenshot from 2022-01-19 23-39-28

### After
Workflows have consistent display order now
Screenshot from 2022-01-19 23-40-00

Thanks!

Signed-off-by: Peixuan Ding <dingpeixuan911@gmail.com>

### DIFF
--- a/cmd/argo/commands/list.go
+++ b/cmd/argo/commands/list.go
@@ -38,7 +38,8 @@ type listFlags struct {
 }
 
 var (
-	nameFields    = "metadata,items.metadata.name"
+	// finishedAt and creationTimestamp must be included to have a consistent display order of workflows
+	nameFields    = "metadata,items.metadata.name,items.metadata.creationTimestamp,items.status.finishedAt"
 	defaultFields = "metadata,items.metadata,items.spec,items.status.phase,items.status.message,items.status.finishedAt,items.status.startedAt,items.status.estimatedDuration,items.status.progress"
 )
 
--- a/cmd/argo/commands/list_test.go
+++ b/cmd/argo/commands/list_test.go
@@ -76,11 +76,21 @@ func Test_listWorkflows(t *testing.T) {
 			assert.Len(t, workflows, 1)
 		}
 	})
+	t.Run("Names", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


alexec (approved):

alexec:
Good work!

PR #7607, opened 2022-01-21
fix: SSO to handle multiple authorization cookies such as from wildca…

### LINKED ISSUE #7592: SSO Bearer Token Formatting
## Summary

I have configured SSO on an upgraded version of Argo Workflow as per the documentation. Everything on the SSO side seems to be working. But after SSO makes the callback, I'm brought back to the login page with error "Failed to load version/info Error: Unauthorized". 

The developer console shows that there are escaped quotes in the bearer token from Argo's cookie when the login page tries to `GET /api/v1/info`

```
authorization: "\"Bearer   v2:ey...g\""
```

If I remove the extra quotes and use that token via client authentication, it works fine. And when I look at the /userinfo page after using that token, it shows the issuer as argo-server as well as the expected username under "subject". So it would appear that Argo Server is communicating with SSO just fine, and issuing a valid key, just in an invalid format. 

There was a previous bug opened for this #5916. The comments indicate that this was fixed in version 3.1, but I'm seeing this in the latest version (v3.2.6) even after clearing cache / running this in incognito (where I go through SSO again) and both Firefox and Chrome.

v3.2.6, but also tried on a prior version (3.1.15)

## Diagnostics

SSO configuration seems to be good:

```
time="2022-01-19T17:46:35.516Z" level=info authModes="[sso client]" baseHRef=/ managedNamespace=argo namespace=argo secure=true
time="2022-01-19T17:46:35.516Z" level=info msg="Generating Self Signed TLS Certificates for Secure Mode"
time="2022-01-19T17:46:35.525Z" level=info msg="config map" name=workflow-controller-configmap
time="2022-01-19T17:46:35.725Z" level=info msg="SSO configuration" clientId="{{oauth-client-secret} client-id-key <nil>}" insecureSkipVerify=false issuer="https://<my_sso>.com" issuerAlias=DISABLED redirectUrl="https://<argo_server>/oauth2/callback" scopes="[email openid]"
time="2022-01-19T17:46:35.725Z" level=info msg="SSO enabled"
```

Accessing the UI gives me these entries in the logs:

```
time="2022-01-19T18:15:38.392Z" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = token not valid for running mode" grpc.code=Unauthenticated grpc.method=GetInfo grpc.service=info.InfoService grpc.start_time="2022-01-19T18:15:38Z" grpc.time_ms=0.08 span.kind=server system=grpc
```
As mentioned in the summary, the token that argo-server issues is valid. It's just incorrectly formatted in the cookie. 

What Kubernetes provider are you using? Bare Metal deployed via Kubespray

What executor are you running? Docker

---

khyer:
…rd domains. Fixes #7592

Signed-off-by: Ken Hyer <khyer123@gmail.com>

This PR addresses a situation where Argo Workflow is hosted on a subdomain, and our organization has an org-wide cookie issued by our SSO provider with name 'authorization'. Meanwhile, Argo Server's UI authentication issues its own authorization cookie for both SSO and client authorization. Modern browsers will send both cookies, and Argo Server was only picking one to attempt to authenticate on. As a result, Argo Server would throw a "token not valid for running mode" error. This isn't necessarily Argo's fault, browsers don't send many details with cookies. But it's a real-world situation that we've run into in our organization. 

I replicated issue in my dev environment, showing an authorization cookie set on parent example.com domain, and Argo's bearer token set on the subdomain:
before_cookies

Resulting error:
before_auth_attempt

The PR modifies the gateway so that it will now fetch any available authorization cookies, and attempt to validate them to find one that is valid for its running mode(s) instead of just dying on the first one it finds. 

After the modification is applied, we still have the two cookies:

after_cookies

But now calls to the server are successfully authenticated: 

after_api_result

A unit test replicating this situation (authentication with multiple cookies with the same name) is also written.

### DIFF
--- a/server/auth/gatekeeper.go
+++ b/server/auth/gatekeeper.go
@@ -146,28 +146,43 @@ func GetClaims(ctx context.Context) *types.Claims {
 	return config
 }
 
-func getAuthHeader(md metadata.MD) string {
+func getAuthHeaders(md metadata.MD) []string {
 	// looks for the HTTP header `Authorization: Bearer ...`
 	for _, t := range md.Get("authorization") {
-		return t
+		return []string{t}
 	}
 	// check the HTTP cookie
+	// In cases such as wildcard domain cookies, there could be multiple authorization headers
+	var authorizations []string
 	for _, t := range md.Get("cookie") {
 		header := http.Header{}
 		header.Add("Cookie", t)
 		request := http.Request{Header: header}
-		token, err := request.Cookie("authorization")
-		if err == nil {
-			return token.Value
+		cookies := request.Cookies()
+		for _, c := range cookies {
+			if c.Name == "authorization" {
+				authorizations = append(authorizations, c.Value)
+			}
 		}
 	}
-	return ""
+	return authorizations
 }
 
 func (s gatekeeper) getClients(ctx context.Context, req interface{}) (*servertypes.Clients, *types.Claims, error) {
 	md, _ := metadata.FromIncomingContext(ctx)
-	authorization := getAuthHeader(md)
-	mode, valid := s.Modes.GetMode(authorization)
+	authorizations := getAuthHeaders(md)
+	valid := false
+	var mode Mode
+	var authorization string
+
+	for _, token := range authorizations {
+		mode, valid = s.Modes.GetMode(token)
+		// Stop checking after the first valid token
+		if valid {
+			authorization = token
+			break
+		}
+	}
 	if !valid {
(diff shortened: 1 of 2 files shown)


alexec (approved):

nicolai92:
Any ideas how to debug this issue? I guess we at Lufthansa are affected too, but I already upgraded to v3.3.0-rc3 ... still getting "token not valid for running mode". However, I am able to login and submit workflows using the UI, but the details view (steps, logs, etc.) is not loading, because of "token not valid for running mode".

PR #7619, opened 2022-01-24
examples: fixing missing namespace

serhatYilmazz:
If we don't specify namespace while creating a ConfigMap to read keys from it, we can get errors like:
Issue-1

- So, we can define what should be done and where to as a side note in example for user experience as in commit history.


terrytangyuan (changes requested):
This is intentional to be consistent with all examples https://github.com/argoproj/argo-workflows/tree/master/examples

The namespace shouldn't be hard-coded.

serhatYilmazz:
Thank you for your reply and explanation 🙋🏻‍♂️
Do you think that it is clearer than before if there is a `Note` comment for all examples ? Because I review all of them to discover Argo Workflows fully, I can do it for all.

PR #7621, opened 2022-01-24
feat: adding support for getting tls certificates from kubernetes secret (e.g.

ChaosInTheCRD:
Note: Continuation of #6542 (commit signing issues)

This PR allows a user to specify a secret name as a command-line parameter (--tls-certificate-secret-name) on the argo-server. Said secret would contain keys tls.key and tls.crt in order to provide a mechanism by which they can bring their own certificates (e.g. from cert-manager, Bitnami Sealed Secrets, generic Kubernetes secret).

This PR tries to follow a similar implementation to what is seen in Argo CD (argoproj/argo-cd#6071), for maintainability.

PR #7651, opened 2022-01-26
fix: TLS problem in UI. Fixes #7632

### LINKED ISSUE #7632: Quickstart: transport: authentication handshake failed
## Summary

When connecting to the UI of Argo Workflows (after port forwarding), I receive this message and virtually no components load:
```
Service Unavailable: connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority"
```

latest (ie 3.2.6)

## Diagnostics
Quickstart Guide - follow steps in guide to reproduce

What Kubernetes provider are you using?
EKS/Minikube

Configuration:
```yaml
    spec:
      containers:
      - args:
        - server
        - --namespaced
        - --auth-mode
        - server
        - --auth-mode
        - client
        - --insecure-skip-verify
        - "true"
```

kustomization.yaml
```yaml
resources:
  - namespace.yml
  - https://raw.githubusercontent.com/argoproj/argo-workflows/v3.2.6/manifests/quick-start-minimal.yaml

namespace: argo-workflows
```

image

> Note: This is reproducible w/ the minimal quickstart via minikube as well
---

ChaosInTheCRD:
Fixes #7632

**PLEASE NOTE**:
- [InsecureSkipVerify](https://pkg.go.dev/crypto/tls#:~:text=InsecureSkipVerify%20bool) is not relevant in the context of a server, as it is for client config so "crypto/tls accepts any certificate presented by the server and any host name in that certificate". In short, this line shouldn't be here
- As far as I can see, it seemed to be needed because it (as a part of `tls.Config` is passed into `apiserver.ArgoServerOpts` (seen [here](https://github.com/argoproj/argo-workflows/blob/master/cmd/argo/commands/server.go#L161))
- In turn, I think this `InsecureSkipVerify: true` is used in other areas ([here](https://github.com/argoproj/argo-workflows/blob/1df69afb6fa57598862edc083c6c0b968e728035/server/auth/sso/sso.go#L131) for instance)
- The correct way, would be that a flag was passed in here instead that simply allows the user to decide whether the clients elsewhere in the application should be able to skip verification. This would always need to be `true` in the case of self signed certs, unless some way could be determined that would allow the self signed certs to be trusted.

I believe it likely that this is the case, but cannot guarantee it without further testing. However, merging this pull request does put it in a better situation than before #7621 was merged.

### DIFF
--- a/util/tls/tls.go
+++ b/util/tls/tls.go
@@ -122,8 +122,9 @@ func GenerateX509KeyPairTLSConfig(tlsMinVersion uint16) (*tls.Config, error) {
 	}
 
 	return &tls.Config{
-		Certificates: []tls.Certificate{*cer},
-		MinVersion:   uint16(tlsMinVersion),
+		Certificates:       []tls.Certificate{*cer},
+		MinVersion:         uint16(tlsMinVersion),
+		InsecureSkipVerify: true,
 	}, nil
 }
 


alexec (approved):

alexec:
LGTM!

PR #7653, opened 2022-01-26
fix: Match cli display pod names with k8s. Fixes #7646

### LINKED ISSUE #7646: PODNAMES as shown from argocli does not match with actual generated pod names
## Summary

The Argo CLI's watch command should display the Argo generated pod names correctly, which in the new versions that use ```POD_NAMES=v2``` have the format of ```[defined prefix]-[task template name]-[random number string]```

However currently the Argo CLI show pod names as ```[defined prefix]-[random number string]```, missing the task template name component.

Example:
I have a workflow with ```generateName: test-addition``` and its first step is ```load-input```.
When using ```argo submit```, a workflow ID ```test-addition-zkzwx``` was generated.
The step ```load-input``` results in an actual pod named ```test-addition-zkzwx-load-input-2972163892``` which shows up from ```kubectl get pods```
But from the ```argo watch``` screen, its pod name is shown as ```test-addition-zkzwx-2972163892```

##Environment
v3.2.6 on argoexec, workflow-controller, and argocli
v1.23.1-2+9e8d1bf4219080 on Kubernetes
Kubernetes initialized using microk8s and using the containerd runtime
Environmental variable ```POD_NAMES=v2``` is set for both the argocli deployment and workflow-controller deployment

## Diagnostics

To reproduce this bug, use any workflow .yaml file and submit with ```argo submit workflow.yaml --watch```. The pod names shown in the watch screen would be incorrect.

Using ```kubectl``` to check confirms that the actual pods do have the V2 name format ```[defined prefix]-[task template name]-[random number string]```. So the pod name generation is working fine. It's just the Argo CLI watch screen not "retrieving" or showing these names right. Perhaps the Argo CLI was implemented in a way that doesn't actually retrieve the actual pod names, but instead using a pre-defined template to display the pod names, and that template wasn't updated together with the V2 POD NAME update.

Aside from the pod names, the workflows run perfectly fine with no errors.

Assigning to @JPZ13   Thank you.
---

JPZ13:
This PR:
- Invokes the workflow utils PodNames function in the cmd printNode function to add the updated pod name to the printout
- Modifies the PodNames function to take the version as a parameter, allowing it to be invoked in the cli when the `POD_NAMES` environment variable isn't present
- Fixes #7646 

```
Name:                dag-diamond-xtpd7
Namespace:           argo
Conditions:
 PodRunning          False
 Completed           True
Created:             Wed Jan 26 16:32:00 -0500 (13 seconds ago)
Started:             Wed Jan 26 16:32:00 -0500 (13 seconds ago)
Finished:            Wed Jan 26 16:32:13 -0500 (now)
Duration:            13 seconds
Progress:            4/4
ResourcesDuration:   8s*(1 cpu),0s*(100Mi memory)

STEP                  TEMPLATE  PODNAME                            DURATION  MESSAGE
 ✔ dag-diamond-xtpd7  diamond
 ├─✔ A                echo      dag-diamond-xtpd7-echo-3871903043  3s
 ├─✔ B                echo      dag-diamond-xtpd7-echo-3888680662  3s
 ├─✔ C                echo      dag-diamond-xtpd7-echo-3905458281  3s
 └─✔ D                echo      dag-diamond-xtpd7-echo-3922235900  3s
```

Signed-off-by: J.P. Zivalich <jp@pipekit.io>

### DIFF
--- a/test/e2e/fixtures/then.go
+++ b/test/e2e/fixtures/then.go
@@ -17,6 +17,7 @@ import (
 
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v3/pkg/client/clientset/versioned/typed/workflow/v1alpha1"
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
 	"github.com/argoproj/argo-workflows/v3/workflow/hydrator"
 	"github.com/argoproj/argo-workflows/v3/workflow/util"
 )
@@ -86,9 +87,15 @@ func (t *Then) ExpectWorkflowNode(selector func(status wfv1.NodeStatus) bool, f 
 		if n != nil {
 			_, _ = fmt.Println("Found node", "id="+n.ID, "type="+n.Type)
 			if n.Type == wfv1.NodeTypePod {
+				version := util.PodNameV1
+				annotations := metadata.GetAnnotations()
+				if annotations[common.AnnotationKeyPodNameVersion] == util.PodNameV2.String() {
+					version = util.PodNameV2
+				}
+				podName := util.PodName(t.wf.Name, n.Name, n.TemplateName, n.ID, version)
+
 				var err error
 				ctx := context.Background()
-				podName := util.PodName(t.wf.Name, n.Name, n.TemplateName, n.ID)
 				p, err = t.kubeClient.CoreV1().Pods(t.wf.Namespace).Get(ctx, podName, metav1.GetOptions{})
 				if err != nil {
 					if !apierr.IsNotFound(err) {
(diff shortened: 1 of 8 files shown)


test/e2e/fixtures/then.go
```diff
@@ -86,9 +87,15 @@ func (t *Then) ExpectWorkflowNode(selector func(status wfv1.NodeStatus) bool, f
 		if n != nil {
 			_, _ = fmt.Println("Found node", "id="+n.ID, "type="+n.Type)
 			if n.Type == wfv1.NodeTypePod {
+				version := util.PodNameV1
+				annotations := metadata.GetAnnotations()
+				if annotations[common.AnnotationKeyPodNameVersion] == util.PodNameV2.String() {
```
terrytangyuan, line 92: Reuse GetWorkflowPodNameVersion?

JPZ13: I was thinking about it but didn't have a workflow object. Would you recommend building one using the metadata variable, or just leaving as is?

terrytangyuan: I think creating a workflow object would be better to reduce code redundancy and make it easier later when we start removing v1 pod name.

JPZ13: Good call. I'll make the switch

PR #7711, opened 2022-01-31
feat: Support `workflow.parameters` in workflow meta-data. Fixes #3434

### LINKED ISSUE #3434: Cannot use `workflow.parameters` to set Workflow metadata?
## To reproduce

1. Write a Workflow template that attempts to set `metadata.labels` with some key/value pair, where the value is set from Workflow parameters (`job_id` being the workflow param in this example):

```
kind: Workflow
metadata:
  generateName: render-
  labels:
    jobId: "{{workflow.parameters.job_id}}"
```

2. Submit the workflow, passing in `job_id` as a workflow parameter.

3. Observe that the workflow fails to submit to Argo.

```
Kubeclient::HttpError (Workflow.argoproj.io "render-w85k4" is invalid: metadata.labels: Invalid value: "{{workflow.parameters.job_id}}": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')):

app/lib/argo/argo.rb:47:in `workflow_submit'
```

alexec:
Fixes #3434

---

This PR refactors `spec.workflowMetadata` to workflows, this results in about 1k lines of code being removed.

The PR then extends it with `labelsFrom` so that the workflows' labels are updated on first reconciliation to whatever expression is in there.

```yaml
 kind: Workflow
 metadata:
   generateName: label-value-from-
   annotations:
       This examples show you how to add labels based on an expression.
       You can then query workflows based on the parameters they were invoked with.
       In this specific case, the value of foo will set as a label on the workflow.
 spec:
   arguments:
     parameters:
       - name: foo
         value: bar
   workflowMetadata:
     labelsFrom:
       foo:
         expression: workflow.parameters.foo
   entrypoint: main
   templates:
     - name: main
       container:
         image: argoproj/argosay:v2 
```

---

Create your PR as a draft.

  does not need to pass.
* Say how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* You can ask for help!


sarabala1979:
I think we can easily substitute the global param into workflow metadata. Instead for workflow.spec, we should pass workflow object itself will substitute the value in label and annatations
https://github.com/argoproj/argo-workflows/blob/93c9c613d4759865eeacb5aedabe3015bf31b8bd/workflow/controller/operator.go#L3546

alexec:
> I think we can easily substitute the global param into workflow metadata. Instead for workflow. spec, we should pass workflow object itself or workflow metadata object will substitute the value in label and annotations
> 
> https://github.com/argoproj/argo-workflows/blob/93c9c613d4759865eeacb5aedabe3015bf31b8bd/workflow/controller/operator.go#L3546

I think you're right. My impl is restrictive.

alexec:
@sarabala1979 ready for review again.

examples/label-value-from-workflow.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: label-value-from-
+  annotations:
+    workflows.argoproj.io/description: |
```
alexec, line 6: add version

sarabala1979:
> > I think we can easily substitute the global param into workflow metadata. Instead for workflow. spec, we should pass workflow object itself or workflow metadata object will substitute the value in label and annotations
> > https://github.com/argoproj/argo-workflows/blob/93c9c613d4759865eeacb5aedabe3015bf31b8bd/workflow/controller/operator.go#L3546
> 
> I think you're right. My impl is restrictive.

@alexec This PR should have only changes in `substituteGlobalVariables`. WorkflowSpec should not have a `metadata` field which is confusing.  `WorkflowTemplate` has metadata to pass the metadata to workflow.

alexec:
> @alexec This PR should have only changes in substituteGlobalVariables. WorkflowSpec should not have a metadata field which is confusing. WorkflowTemplate has metadata to pass the metadata to workflow.

It is a bit confusing, but it is impossible to use `metadata.labels`, because Kubernetes only allows you to create workflows with valid labels. The cannot contain `{{` for example. Even if they did, to template them would be a breaking change.

Therefore, we need to have a way to do this in a field we control. This has to be `spec.X`. I used `spec.workflowMetadata` so that it was consistent with `CronWorkflow/ClusteWorkflowTemplate/WorkflowTemplate` .

Can you think of an alternative

alexec:
@sarabala1979 bump!

sarabala1979 (approved):

api/jsonschema/schema.json
```diff
@@ -4852,7 +4852,7 @@
           "$ref": "#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta"
         },
         "spec": {
-          "$ref": "#/definitions/io.argoproj.workflow.v1alpha1.WorkflowTemplateSpec"
+          "$ref": "#/definitions/io.argoproj.workflow.v1alpha1.WorkflowSpec"
```
terrytangyuan, line 4855: This is a breaking change?

alexec: no

PR #7746, opened 2022-02-04
fix: Missed workflow should not trigger if Forbidden Concurreny with no StartingDeadlineSeconds

### LINKED ISSUE #7682: Cron workflow is schedule workflow after configured  time
## Summary
We are experiencing some weird behaviour when using forbid concurrency policy, where new workflows are scheduled before the last workflow finishes, and then the new workflow is started immediately after the previous workflow finishes.

We have this CWF:

concurrencyPolicy: Forbid
  schedule: 0-59/5 0-10 * * *

and for example we have this WF, with start/creation time after the scheduled-time

annotations:
    workflows.argoproj.io/scheduled-time: "2022-01-24T10:35:00Z"
  creationTimestamp: "2022-01-24T10:38:19Z"
finishedAt: "2022-01-24T11:23:05Z" 

Now notice the next Workflow, which is scheduled way before the previous run finishes:

annotations:
    workflows.argoproj.io/scheduled-time: "2022-01-24T10:55:00Z"
  creationTimestamp: "2022-01-24T11:23:12Z"
but we shouldn't have any workflows that start past 11AM. Also WF1 (and previous ones) are not running when minute is at multiple of 5 minutes

## Diagnostics

Either a workflow that reproduces the bug, or paste you whole workflow YAML, including status, something like:

```yaml
kubectl get wf -o yaml ${workflow}
```

What Kubernetes provider are you using?

What executor are you running? Docker/K8SAPI/Kubelet/PNS/Emissary

```bash

```

---

sarabala1979:
…
closes #7682 

* [ ] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

  does not need to pass.
* Say how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* You can ask for help!

### DIFF
--- a/workflow/cron/operator.go
+++ b/workflow/cron/operator.go
@@ -258,8 +258,8 @@ func (woc *cronWfOperationCtx) shouldOutstandingWorkflowsBeRun() (time.Time, err
 
 		// We missed the latest execution time
 		if !missedExecutionTime.IsZero() {
-			// If StartingDeadlineSeconds is not set, or we are still within the deadline window, run the Workflow
-			if woc.cronWf.Spec.StartingDeadlineSeconds == nil || *woc.cronWf.Spec.StartingDeadlineSeconds == 0 || now.Before(missedExecutionTime.Add(time.Duration(*woc.cronWf.Spec.StartingDeadlineSeconds)*time.Second)) {
+			// if missedExecutionTime is within StartDeadlineSeconds, We are still within the deadline window, run the Workflow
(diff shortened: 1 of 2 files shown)


workflow/cron/operator.go
```diff
@@ -258,8 +258,8 @@ func (woc *cronWfOperationCtx) shouldOutstandingWorkflowsBeRun() (time.Time, err
 
 		// We missed the latest execution time
 		if !missedExecutionTime.IsZero() {
```
sarabala1979, line 260: missed execution should run only if StartdeadlineSeconds is configured, otherwise it should be forbidden.

alexec (approved):

PR #7805, opened 2022-02-08
fix: submitting Workflow from WorkflowTemplate will set correct serviceAccount and securityContext. Fixes #7726

### LINKED ISSUE #7726: Submitting from a WorkflowTemplate shows incorrect serviceAccount and securityContext nag
## Summary

Starting a Workflow from a WorkflowTemplate brings up the security nag "This workflow does not have security context set. ..." - even if the WorkflowTemplate defines a securityContext.  I've also (just) noticed that the serviceAccount is also shown as "default" even though I'm running with one (currently having to as using PNS).

Running with v3.2.7 client & v3.2.6 server

## Diagnostics

Example WorkflowTemplate

```yaml
kind: WorkflowTemplate
metadata:
  name: my-workflow-template
  namespace: my-namespace
spec:
  entrypoint: main
  securityContext:
    runAsNonRoot: false
  serviceAccountName: my-service-account
  templates:
    - name: main
      metadata:
      container:
        command:
          - bash
          - -c
          - echo 'Hello from WT'; env; sleep 15
        image: ubuntu
        name: main
```

Example command & output:

```sh
$ argo submit --from workflowtemplate/my-workflow-template
Name:                my-workflow-template-76qt4
Namespace:           my-namespace
ServiceAccount:      default
Status:              Pending
Created:             Tue Feb 01 21:38:02 +0000 (now)
Progress:            

This workflow does not have security context set. You can run your workflow pods more securely by setting it.
Learn more at https://argoproj.github.io/argo-workflows/workflow-pod-security-context/

```

What Kubernetes provider are you using? EKS

What executor are you running? PNS

---

dpadhiar:
fixes #7726 

Previously if a user started a Workflow from a WorkflowTemplate using the CLI, the serviceAccount and securityContext was not set based upon the StoredWorkflowSpec like is expected. Now, when serviceAccount and securityContext are checked for in `get.go`, `GetExecSpec()` is run to check if `wf.Status.StoredWorkflowSpec` is set or not. If so, it will reference this spec instead as expected.

Example WorkflowTemplate:

```
kind: WorkflowTemplate
metadata:
  name: my-workflow-template-test
  namespace: argo
spec:
  entrypoint: main
  securityContext:
    runAsNonRoot: false
  serviceAccountName: argo
  templates:
    - name: main
      metadata:
      container:
        command:
          - bash
          - -c
          - echo 'Hello from WT'; env; sleep 15
        image: ubuntu
        name: main
```

Submitting through CLI:
```
dist/argo submit --from workflowtemplate/my-workflow-template-test --watch
```

Output:
```
Name:                my-workflow-template-test-5ppq9
Namespace:           argo
ServiceAccount:      argo
Conditions:          
 PodRunning          False
 Completed           True
Created:             Mon Feb 07 15:18:53 -0800 (20 seconds ago)
Started:             Mon Feb 07 15:18:53 -0800 (20 seconds ago)
Finished:            Mon Feb 07 15:19:13 -0800 (now)
Duration:            20 seconds
Progress:            1/1
ResourcesDuration:   32s*(1 cpu),25s*(100Mi memory)

STEP                                TEMPLATE  PODNAME                          DURATION  MESSAGE
 ✔ my-workflow-template-test-5ppq9  main      my-workflow-template-test-5ppq9  19s 
```

### DIFF
--- a/cmd/argo/commands/get.go
+++ b/cmd/argo/commands/get.go
@@ -125,7 +125,7 @@ func printWorkflowHelper(wf *wfv1.Workflow, getArgs getFlags) string {
 		// also isn't set, we will use the 'default' ServiceAccount in the namespace
 		// the workflow will run in.
 		if wf.Spec.WorkflowTemplateRef != nil {
-			serviceAccount = "unset (will run with referred workflow/cluster template ServiceAccount)"
+			serviceAccount = "unset"
 		} else {
 			serviceAccount = "unset (will run with the default ServiceAccount)"
 		}
@@ -157,9 +157,9 @@ func printWorkflowHelper(wf *wfv1.Workflow, getArgs getFlags) string {
 	if !wf.Status.ResourcesDuration.IsZero() {
 		out += fmt.Sprintf(fmtStr, "ResourcesDuration:", wf.Status.ResourcesDuration)
 	}
-	if len(wf.Spec.Arguments.Parameters) > 0 {
+	if len(wf.GetExecSpec().Arguments.Parameters) > 0 {
 		out += fmt.Sprintf(fmtStr, "Parameters:", "")
-		for _, param := range wf.Spec.Arguments.Parameters {
+		for _, param := range wf.GetExecSpec().Arguments.Parameters {
 			if param.Value == nil {
 				continue
 			}
--- a/cmd/argo/commands/get.go
+++ b/cmd/argo/commands/get.go
@@ -118,13 +118,17 @@ func printWorkflowHelper(wf *wfv1.Workflow, getArgs getFlags) string {
 	out := ""
 	out += fmt.Sprintf(fmtStr, "Name:", wf.ObjectMeta.Name)
 	out += fmt.Sprintf(fmtStr, "Namespace:", wf.ObjectMeta.Namespace)
-	serviceAccount := wf.Spec.ServiceAccountName
+	serviceAccount := wf.GetExecSpec().ServiceAccountName
 	if serviceAccount == "" {
 		// if serviceAccountName was not specified in a submitted Workflow, we will
 		// use the serviceAccountName provided in Workflow Defaults (if any). If that
 		// also isn't set, we will use the 'default' ServiceAccount in the namespace
 		// the workflow will run in.
-		serviceAccount = "unset (will run with the default ServiceAccount)"
+		if wf.Spec.WorkflowTemplateRef != nil {
+			serviceAccount = "unset (will run with referred workflow/cluster template ServiceAccount)"
+		} else {
+			serviceAccount = "unset (will run with the default ServiceAccount)"
+		}
 	}
 	out += fmt.Sprintf(fmtStr, "ServiceAccount:", serviceAccount)
 	out += fmt.Sprintf(fmtStr, "Status:", printer.WorkflowStatus(wf))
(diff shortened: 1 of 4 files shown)


cmd/argo/commands/get.go
```diff
@@ -118,13 +118,17 @@ func printWorkflowHelper(wf *wfv1.Workflow, getArgs getFlags) string {
...
 	out += fmt.Sprintf(fmtStr, "Namespace:", wf.ObjectMeta.Namespace)
-	serviceAccount := wf.Spec.ServiceAccountName
+	serviceAccount := wf.GetExecSpec().ServiceAccountName
 	if serviceAccount == "" {
 		// if serviceAccountName was not specified in a submitted Workflow, we will
 		// use the serviceAccountName provided in Workflow Defaults (if any). If that
 		// also isn't set, we will use the 'default' ServiceAccount in the namespace
 		// the workflow will run in.
-		serviceAccount = "unset (will run with the default ServiceAccount)"
+		if wf.Spec.WorkflowTemplateRef != nil {
```
alexec, line 127: I think this can be simpler, just say "unset"

pkg/apis/workflow/v1alpha1/workflow_types_test.go
```diff
@@ -1096,3 +1096,40 @@ func TestStepSpecGetExitHook(t *testing.T) {
 	hooks = step.GetExitHook(step.Arguments)
 	assert.Equal(t, "hook", hooks.Template)
 }
+
+func TestGetExecSpec(t *testing.T) {
+	wf := Workflow{
```
alexec, line 1101: add a second test when stored spec is nil

alexec (changes requested):
when I did triage on this, I thought there was a third place in the cli where `wf.Spec` needs to be replaced by `wf.GetExecSpec()`, can you check please?

dpadhiar:
> when I did triage on this, I thought there was a third place in the cli where `wf.Spec` needs to be replaced by `wf.GetExecSpec()`, can you check please?

In `get.go` on lines 160 - 168, the spec is used here to access parameters. I believe this may be the third place for a CLI change, could you confirm?

```
if len(wf.Spec.Arguments.Parameters) > 0 {
		out += fmt.Sprintf(fmtStr, "Parameters:", "")
		for _, param := range wf.Spec.Arguments.Parameters {
			if param.Value == nil {
				continue
			}
			out += fmt.Sprintf(fmtStr, "  "+param.Name+":", *param.Value)
		}
	}
	
```

alexec:
> > when I did triage on this, I thought there was a third place in the cli where `wf.Spec` needs to be replaced by `wf.GetExecSpec()`, can you check please?
> 
> In `get.go` on lines 160 - 168, the spec is used here to access parameters. I believe this may be the third place for a CLI change, could you confirm?
> 
> ```
> if len(wf.Spec.Arguments.Parameters) > 0 {
> 		out += fmt.Sprintf(fmtStr, "Parameters:", "")
> 		for _, param := range wf.Spec.Arguments.Parameters {
> 			if param.Value == nil {
> 				continue
> 			}
> 			out += fmt.Sprintf(fmtStr, "  "+param.Name+":", *param.Value)
> 		}
> 	}
> 	
> ```

Yes. Fix that too.

alexec (approved):
🚀 LGTM

PR #7809, opened 2022-02-08
feat: Always Show Workflow Parameters

### LINKED ISSUE #7311: Allow to always pass workflow scoped variable via UI
# Summary

Previously you could submit a workflowtemplate in the UI by passing a parameter manually, it would not matter which entrypoint of the workflow you chose. Now you can only submit a parameter to the workflow if you choose the default entrypoint. If the parameter is workflow scoped, then it should be always possible to pass it regardless of which entrypoint you choose.

Default entrypoint -> able to pass the workflow parameter
image

Picked different entrypoint -> not able to pass the workflow parameter any longer
image

First discussed in https://github.com/argoproj/argo-workflows/issues/7294.

# Use Cases

Within our company we use argo workflows to trigger evaluations of our ML models manually from the UI. We have a workflow with different entrypoints based on the input you want to supply to the ML model. The workflow parameter is always the same and it would be redundant to add this parameter to each entrypoint individually.

---

kennytrytek:
Signed-off-by: Kenny Trytek <kenneth.g.trytek@gmail.com>

Fixes #7311

### Testing
The following GIF shows the submission of the example workflow from https://github.com/argoproj/argo-workflows/issues/7311#issuecomment-1026411700.
* Workflow can be submitted with workflow parameters given, even if a non-default entrypoint is selected.

submission

### DIFF
--- a/ui/src/app/workflows/components/submit-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/submit-workflow-panel.tsx
@@ -44,7 +44,7 @@ export class SubmitWorkflowPanel extends React.Component<Props, State> {
             entrypoints: this.props.templates.map(t => t.name),
             selectedTemplate: defaultTemplate,
             parameters: [] as Parameter[],
-            workflowParameters: this.props.workflowParameters,
+            workflowParameters: JSON.parse(JSON.stringify(this.props.workflowParameters)),
             templates: [defaultTemplate].concat(this.props.templates),
             labels: ['submit-from-ui=true'],
             isSubmitting: false
--- a/ui/src/app/workflows/components/submit-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/submit-workflow-panel.tsx
@@ -1,11 +1,10 @@
+import {Select, Tooltip} from 'argo-ui';
 import * as React from 'react';
 import {Parameter, Template, Workflow} from '../../../models';
 import {uiUrl} from '../../shared/base';
 import {ErrorNotice} from '../../shared/components/error-notice';
-import {services} from '../../shared/services';
-
-import {Select, Tooltip} from 'argo-ui';
 import {TagsInput} from '../../shared/components/tags-input/tags-input';
+import {services} from '../../shared/services';
 
 interface Props {
     kind: string;
@@ -20,6 +19,7 @@ interface State {
     entrypoint: string;
     entrypoints: string[];
     parameters: Parameter[];
+    workflowParameters: Parameter[];
     selectedTemplate: Template;
     templates: Template[];
     labels: string[];
@@ -28,6 +28,7 @@ interface State {
 }
 
 const workflowEntrypoint = '<default>';
+type ParamSelector = 'parameters' | 'workflowParameters';
 
 export class SubmitWorkflowPanel extends React.Component<Props, State> {
     constructor(props: any) {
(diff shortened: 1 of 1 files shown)


kennytrytek:
Hi Stale bot, I'm planning on updating this today.

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -133,36 +125,55 @@ export class SubmitWorkflowPanel extends React.Component<Props, State> {
...
-                            enum: p.enum
-                        }))
-                    });
+                    const update = {} as State;
+                    update[parameterStateName] = this.state[parameterStateName].map(p => ({
+                        name: p.name,
+                        value: p.name === parameter.name ? event.value : this.getValue(p),
+                        enum: p.enum
+                    }));
+                    this.setState(update);
```
alexec, line 134: do you want to use `this.setState(oldState => newState)` version here?

kennytrytek: There is only one example of the pattern you suggested in the code base:
https://github.com/argoproj/argo-workflows/blob/v3.3.1/ui/src/app/shared/components/input-filter.tsx#L63

What would this code gain by using that pattern? All other `setState`s in the UI look like this.

alexec: it is useful it you only want to modify part of the state, which is what I think you are trying to do here

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -28,21 +28,23 @@ interface State {
...
             }
         };
-        const state = {
+        const state: State = {
             entrypoint: workflowEntrypoint,
             entrypoints: this.props.templates.map(t => t.name),
             selectedTemplate: defaultTemplate,
-            parameters: this.props.workflowParameters || [],
+            parameters: [] as Parameter[],
+            workflowParameters: this.props.workflowParameters,
```
alexec, line 47: do we need to make a copy of this object rather than reference if we're going to change it? Use `Object.assign` to copy an object

ui/src/app/workflows/components/submit-workflow-panel.tsx
alexec: I think I'd do this differently, I'd add a new parameters to `displaySelectFieldForEnumValues` named `onChange` and pass in the correct func to it. The decouples the two functions.

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -133,36 +125,55 @@ export class SubmitWorkflowPanel extends React.Component<Props, State> {
...
                 onChange={event => {
-                    this.setState({
-                        parameters: this.state.parameters.map(p => ({
-                            name: p.name,
-                            value: p.name === parameter.name ? event.target.value : this.getValue(p),
-                            enum: p.enum
-                        }))
-                    });
+                    const update = {} as State;
+                    update[parameterStateName] = this.state[parameterStateName].map(p => ({
```
alexec, line 147: As above.

alexec (changes requested):

ui/src/app/workflows/components/submit-workflow-panel.tsx
kennytrytek: >it is useful it you only want to modify part of the state

I'm not convinced that `setState(update)` does _not_ do a partial update. At least, I haven't been able to find any documentation that reads otherwise. I also tried using the arrow function update anyway, but was unable to make it successfully update the state.

>I'd add a new parameters to displaySelectFieldForEnumValues named onChange and pass in the correct func to it

I understand the sentiment here, but I gave up after a few hours of trying to make it work this way. I ran into two problems:
* The update function would lose reference to `this`.
* The `setState` update would appear to work, but didn't actually update any values.

If you have any more specific direction, or just want someone else to make the changes, that would be fine with me.

ui/src/app/workflows/components/submit-workflow-panel.tsx
kennytrytek: Changed to a copy in https://github.com/argoproj/argo-workflows/pull/7809/commits/44da063ac2cb9965d9321d81a0f36276f7db733b.

PR #7820, opened 2022-02-08
docs: IDE Setup IntelliJ + K8s Plugin

sernst:
For users with the JetBrains Kubernetes plugin installed, there's
an alternate IDE configuration path that can be both beneficial
and avoid a conflicting validation state when both methods are
enabled on the same file.


docs/ide-setup.md
```diff
@@ -20,6 +20,20 @@ That's it. Open an Argo YAML file and you should see smarter behaviour, includin
 
 ![IDEA Example Functionality](assets/intellij-ide-step-1-example-functionality.png)
 
+### IntelliJ IDEA (Community & Utimate Editions) + Kubernetes Plugin
```
terrytangyuan, line 23: This should work for GoLand as well.

sernst: Yep, agreed. And every other JetBrains IDE as well. 

I'm certainly happy to change the language on that to be broader in terms of JetBrains IDE support. However, I'd advocate for changing the previous header to be consistent as well in that case:

https://github.com/argoproj/argo-workflows/blob/master/docs/ide-setup.md?plain=1#L7

Thoughts?

terrytangyuan: Sounds good to me

sernst: Great. I've made the change.

terrytangyuan (approved):
Thank you!

PR #7837, opened 2022-02-09
feat: remove pod workers. Fixes #4090

### LINKED ISSUE #4090: Deprecate and remove `--pod-workers`
# Summary

In #4064 when a pod is delete we skip the pod workflows and directly queue the workflow. The pod workflows don't do very much and create many additional Goroutines. Lets remove them completely. 

# Use Cases

Tech debt.

---

isubasinghe:
This removes the pod workers and the pod queue and directly writes into the workflow queue instead. 
Fixes #4090

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -99,7 +99,6 @@ type WorkflowController struct {
 	podInformer           cache.SharedIndexInformer
 	configMapInformer     cache.SharedIndexInformer
 	wfQueue               workqueue.RateLimitingInterface
-	podQueue              workqueue.RateLimitingInterface
 	podCleanupQueue       workqueue.RateLimitingInterface // pods to be deleted or labelled depend on GC strategy
 	throttler             sync.Throttler
 	workflowKeyLock       syncpkg.KeyLock // used to lock workflows for exclusive modification or access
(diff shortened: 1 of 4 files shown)


JPZ13:
@isubasinghe - I don't have access to push a commit to this pull request, but here's how to fix the linting error: https://github.com/pipekit/argo-workflows/blob/remove-pod-workers/workflow/controller/controller.go#L1033

workflow/controller/controller.go
```diff
@@ -1057,7 +1016,11 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) cache.SharedI
 					diff.LogChanges(oldPod, newPod)
 					return
 				}
-				wfc.podQueue.Add(key)
+				err = wfc.enqueueWfFromPodLabel(newVal)
+				if err != nil {
+					log.WithError(err).Warnf("could not enqueue workflow from pod label on add: %s", err)
```
alexec, line 1021: do not use `Warnf`, it is not structured, please use `Warn`

alexec (changes requested):

alexec:
I think we need one more change. In docs/upgrading.md, we should mention that this flag has been removed. So users may have custom configured it.

alexec (approved):

Issue #7683, opened 2022-01-28
Improve plugin templates docs


alexec:
* Do plugin templates only support Python? It's not clear in the docs:
  * https://github.com/argoproj/argo-workflows/blob/master/docs/executor_plugins.md
  * https://github.com/argoproj/argo-workflows/blob/master/docs/plugins.md
* Does a plugin failure cause a workflow failure? If so does it support the [continueOn](https://github.com/argoproj/argo-workflows/blob/4e450e250168e6b4d51a126b784e90b11a0162bc/examples/continue-on-fail.yaml) feature?
* How do users securely manage secrets needed for a plugin? Such as a Slack token in your example.
* Should we have an example in https://github.com/argoproj/argo-workflows/tree/master/examples for this feature? 
* Is the community going to maintain any common plugins such as a Slack plugin? 
* How do plugins interact with features such as resource requests/limits , pod disruption budgets, retry on the workflow side and activeDeadlineSeconds? It's a template so all template/step features should work right?
* What are the use cases for the aysnc / requeue feature? Is the 2 minutes configurable?

Shark:
@alexec The Executor Plugin [example](https://github.com/argoproj/argo-workflows/blob/master/docs/executor_plugins.md#a-simple-python-plugin) is currently broken because it's missing a step. There is a Python HTTP server and the plugin is configured to run `python -c`, but the script is never referenced or copied into a Docker image.

Do you have a neat idea for baking the script into the plugin? Custom Docker image is painful just for testing. I found a hacky [Gist](https://gist.github.com/ropnop/d477ed1cf7f78c1296e0a2f20b5d925f) that could help.

Shark:
💡 The file must have the name `server.* `.

This isn't mentioned in the docs, I'll open a PR.

PR #7256, opened 2021-11-20
feat: Template executor plugins. Fixes #5201

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>
Fixes #5201

Don't bother creating a PR until you've done this:

* [x] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it
  does not need to pass.
* Once required tests have passed, you can make it "Ready for review".
* Say how how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* If changes were requested, and you've made them, then dismiss the review to get it looked at again.
* Add you organization to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md) if you like.
* You can ask for help!

PR #7855, opened 2022-02-11
docs: Improve consistency of plugin template docs

Shark:
Refs #7683 

I improved the "Executor Plugins" docs page:

* Explain that `argo executor-plugin build` assumes the script is saved in `server.* `
* Pretty print the args + disable stdout buffering so that it's easy to understand the RPC
* Add the UID for the `nobody` user to make the given plugin runnable (before that it would complain that `runAsNonRoot` is true but the image uses UID 0 per default)

### DIFF
--- a/docs/executor_plugins.md
+++ b/docs/executor_plugins.md
@@ -21,6 +21,15 @@ spec:
               value: "true"
 ```
 
+When using the [Helm chart](https://github.com/argoproj/argo-helm/tree/master/charts/argo-workflows), add this to your `values.yaml`:
+
+```yaml
+controller:
+  extraEnv:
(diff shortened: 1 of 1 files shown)


alexec (approved):

agilgur5 (commented):
GH wouldn't let me comment in-line as the fork no longer exists (it throws an error), so leaving a comment outside:

This `runAsUser` added here appears to differ from the tests for the feature from #7256 :

https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/test/e2e/manifests/plugins/hello-executor-plugin-configmap.yaml#L71

PR #7856, opened 2022-02-11
fix: trim spaces while parse realtime metrics value. Fixes #7819

### LINKED ISSUE #7819: Inconsistency of parsing the realtime metric value.
## Summary
During the definition of real time metric in a workflow, the value of metric is not parsed correctly when defined 
using valid go template syntax
if value in real time metric is defined like this ` {{ workflow.duration }}`. The next error shown in the log

 However it works correctly like this `{{workflow.duration}}`  

I'd expect that it works correctly if either valid go template syntax is applied. 
It's minor issue. The custom parsing of the value doesn't trim suffix and leading white spaces. See [here](https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/operator.go#L3209).
If you think the  fix is good idea, i can submit a PR.

What executor are you using? PNS/Emissary
N/A

## Diagnostics

```yaml
# For complete documentation please see:
#
#   docs/metrics.md
kind: Workflow
metadata:
  generateName: hello-world-
spec:
  entrypoint: steps
  metrics:
    prometheus:
      - name: duration_gauge
        labels:
          - key: name
            value: workflow
        help: "Duration gauge by name"
        gauge:
          realtime: true                  # This metric will be emitted in real time. For more info see: docs/metrics.md
          value: "{{ workflow.duration }}"  # Use {{workflow.duration}} in workflow-level and {{duration}} in template-level
```

What executor are you running? PNS/Emissary

```bash

```
 ' workflow.duration ' is not available as a real time metric
---

olegmayko:
Signed-off-by: Oleg Mayko <oleg.mayko@swisscom.com>

Fixes #7819

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -3206,7 +3206,7 @@ func (woc *wfOperationCtx) computeMetrics(metricList []*wfv1.Prometheus, localSc
 				woc.reportMetricEmissionError("real time metrics can only be used with metric variables")
 				continue
 			}
-			value = strings.TrimSuffix(strings.TrimPrefix(value, "{{"), "}}")
+			value = strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(value, "{{"), "}}"))
 			valueFunc, ok := realTimeScope[value]
 			if !ok {
 				woc.reportMetricEmissionError(fmt.Sprintf("'%s' is not available as a real time metric", value))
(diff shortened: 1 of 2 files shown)


workflow/controller/operator.go
```diff
@@ -3206,7 +3206,7 @@ func (woc *wfOperationCtx) computeMetrics(metricList []*wfv1.Prometheus, localSc
 				woc.reportMetricEmissionError("real time metrics can only be used with metric variables")
 				continue
 			}
-			value = strings.TrimSuffix(strings.TrimPrefix(value, "{{"), "}}")
```
sarabala1979, line 3209: Can you add test for this?

sarabala1979 (approved):

PR #7859, opened 2022-02-11
fix: added priorityclass to workflow-controller. Fixes #7733

### LINKED ISSUE #7733: Add pod priority class to controller
We currently schedule the controller at the same priority as workflow pods. This mistake for almost all users, as it can result workflow pods kicking out the controller.

The controller, should, by default, be scheduled with higher priority.

rohankmr414:

Fixes #7733


alexec (changes requested):
Can you please remove `-priorityclass` suffix? I don't think we need to add kinds as suffixes to resources, it is redundant.

alexec (approved):

terrytangyuan (approved):

PR #7883, opened 2022-02-14
fix: Add license to python sdk. Fixes #7881

### LINKED ISSUE #7881: pypi package for rc missing Licence
## Summary

The tar/package of argo-workflows on pypi, is missing the LICENSE file, whereas in previous versions this was included. We internally use conda, and wanted to bump the conda-forge recipe, however, it has a dependency on the LICENSE file and therefore is not publishing. See https://github.com/conda-forge/hera-workflows-feedstock/pull/10 

What version is it broken in?
v6.0.0rc and higher

---

bchalk101:
Signed-off-by: Baz Chalk <boruch11@gmail.com>

Fixes #7881

### DIFF
--- a/sdks/python/Makefile
+++ b/sdks/python/Makefile
@@ -22,6 +22,7 @@ generate:
 		sed 's/io.k8s.api.core.v1.//' | \
 		sed 's/io.k8s.apimachinery.pkg.apis.meta.v1.//' \
 		> $(WD)/swagger.json
+	cp ../../LICENSE $(WD)/LICENSE
 	$(DOCKER) openapitools/openapi-generator-cli:v5.2.1 \
 		generate \
 		--input-spec /wd/swagger.json \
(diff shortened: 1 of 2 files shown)


sdks/python/Makefile
```diff
@@ -22,6 +22,7 @@ generate:
 		sed 's/io.k8s.api.core.v1.//' | \
 		sed 's/io.k8s.apimachinery.pkg.apis.meta.v1.//' \
 		> $(WD)/swagger.json
+	cp ../../LICENSE $(WD)/LICENSE
```
terrytangyuan, line 25: Could you check if this is being copied to be right path and that https://github.com/argoproj/argo-workflows/issues/7881 is resolved after this?

bchalk101: I have run `make --directory sdks/python publish -B` to produce the `argo-workflows.tar.gz` and it does now contain the license file at the root directory, as it did in versions pre 6.0.0. This will resolve #7881.

terrytangyuan: Thank you!

PR #7898, opened 2022-02-16
docs: provide more details in quick start during workflow submit

danielhelfand:
Adding slightly more detail during submission portion of workflow quick start doc.

A nice follow up may be to include some UI instructions for the workflow below.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -1,6 +1,6 @@
 # Quick Start
 
-To see how Argo works, you can install it and run examples of simple workflows and workflows that use artifacts.
+To see how Argo Workflows work, you can install it and run examples of simple workflows and workflows that use artifacts.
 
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):
Nice. Even though the commands are pretty intuitive to people who are already familiar with `kubectl`, it might still be nice to have.

danielhelfand:
> Nice. Even though the commands are pretty intuitive to people who are already familiar with `kubectl`, it might still be nice to have.

Thanks! Yeah, if the descriptions for each command is too much, happy to remove some of it. I thought though calling out some details of the unique name and at least mentioning what to expect would be helpful.

terrytangyuan:
I think it might be too much as we specified `generateName: hello-world-` which is a pretty basic concept. I'll let others chime in.

PR #7903, opened 2022-02-17
fix: Save containerSet logs in logs archive fixes #7897

### LINKED ISSUE #7897: Save containerSet logs in logs archive
# Summary

Currently, the logs archive only saves the "main" container logs. When using a ContainerSet, a lot of information is being lost once the pod is removed from the cluster.
I propose saving all the container logs in the archive (with the exception of the `init` and `wait` containers).

# Use Cases

In many cases, immediately or shortly after a wf has completed its run, its pod are getting garbage collected. In order to later view the Workflow's logs, it is very handy to set up a log archive in S3 (or similar storage solutions). It seems natural to expect the ContainerSet logs to also end up in the same log archive.

---

noam-codefresh:
Signed-off-by: Noam Gal <noam.gal@codefresh.io>

Fixes #7897

I have added code in the executor and workflow-controller.
In the Executor, `SaveLogs` now iterates over all the `mainContainerNames`, and saves each of their logs. Later, the `AnnotationOutputs` func is saving a different annotation for each container (using the container name to make each unique).
In the Operator, a new func `updateNodeExitData` is copying the container exitCode + matching annotation to the node. If the node has Children (= a ContainerSet), does the same recursively for them as well. Is there any way there could be a circular reference between the nodes like that? Should I protect against infinite recursion? It should be an easy fix to add.
I have also added some simple tests to the `executor_test`, but I didn't see any simple way to mock an ArtifactRepository, so they basically just test the `SaveLogs` function, and how it iterates over the containers.

### DIFF
--- a/cmd/argoexec/commands/wait.go
+++ b/cmd/argoexec/commands/wait.go
@@ -46,11 +46,7 @@ func waitContainer(ctx context.Context) error {
 	if err != nil {
 		wfExecutor.AddError(err)
 	}
-	// Saving logs
-	logArt, err := wfExecutor.SaveLogs(ctx)
-	if err != nil {
-		wfExecutor.AddError(err)
-	}
+
 	// Saving output parameters
 	err = wfExecutor.SaveParameters(ctx)
 	if err != nil {
@@ -61,11 +57,7 @@ func waitContainer(ctx context.Context) error {
 	if err != nil {
 		wfExecutor.AddError(err)
 	}
-	// Annotating pod with output
-	err = wfExecutor.ReportOutputs(ctx, logArt)
-	if err != nil {
-		wfExecutor.AddError(err)
-	}
 
+	wfExecutor.SaveLogs(ctx)
 	return wfExecutor.HasError()
 }
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -2745,14 +2745,14 @@ func (out *Outputs) HasParameters() bool {
 	return out != nil && len(out.Parameters) > 0
 }
 
-const MainLogsArtifactName = "main-logs"
+const LogsSuffix = "-logs"
 
 func (out *Outputs) HasLogs() bool {
 	if out == nil {
 		return false
 	}
 	for _, a := range out.Artifacts {
-		if a.Name == MainLogsArtifactName {
+		if strings.HasSuffix(a.Name, LogsSuffix) {
 			return true
 		}
 	}
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -2014,6 +2014,15 @@ func (n NodeStatus) HasChild(childID string) bool {
 	return false
 }
 
+func (node NodeStatus) GetContainerName() string {
+	// containerSet nodes are names after their container
+	if node.Type == NodeTypeContainer {
+		return node.DisplayName
+	}
+
+	return "main"
+}
+
 // S3Bucket contains the access information required for interfacing with an S3 bucket
 type S3Bucket struct {
 	// Endpoint is the hostname of the bucket endpoint
@@ -2549,7 +2558,7 @@ func (t *Template) IsDaemon() bool {
 
 // if logs should be saved as an artifact
 func (tmpl *Template) SaveLogsAsArtifact() bool {
-	return tmpl != nil && tmpl.ArchiveLocation.IsArchiveLogs() && (tmpl.ContainerSet == nil || tmpl.ContainerSet.HasContainerNamed("main"))
+	return tmpl != nil && tmpl.ArchiveLocation.IsArchiveLogs()
 }
 
 func (t *Template) GetRetryStrategy() (wait.Backoff, error) {
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -2014,15 +2014,6 @@ func (n NodeStatus) HasChild(childID string) bool {
 	return false
 }
 
-func (node NodeStatus) GetContainerName() string {
-	// containerSet nodes are names after their container
-	if node.Type == NodeTypeContainer {
-		return node.DisplayName
-	}
-
-	return "main"
-}
-
 // S3Bucket contains the access information required for interfacing with an S3 bucket
 type S3Bucket struct {
 	// Endpoint is the hostname of the bucket endpoint
--- a/workflow/controller/controller_test.go
+++ b/workflow/controller/controller_test.go
@@ -312,12 +312,12 @@ func listPods(woc *wfOperationCtx) (*apiv1.PodList, error) {
 
 type with func(pod *apiv1.Pod)
 
-func withOutputs(v interface{}) with {
+func withOutputs(cm string, v interface{}) with {
 	switch x := v.(type) {
 	case string:
-		return withAnnotation(common.AnnotationKeyOutputs, x)
+		return withAnnotation(common.AnnotationKeyOutputs+"-"+cm, x)
 	default:
-		return withOutputs(wfv1.MustMarshallJSON(x))
+		return withOutputs(cm, wfv1.MustMarshallJSON(x))
 	}
 }
 func withProgress(v string) with { return withAnnotation(common.AnnotationKeyProgress, v) }
@@ -338,7 +338,9 @@ func withExitCode(v int32) with {
 }
 
 func withAnnotation(key, val string) with {
-	return func(pod *apiv1.Pod) { pod.Annotations[key] = val }
+	return func(pod *apiv1.Pod) {
+		pod.Annotations[key] = val
+	}
 }
 
 // createRunningPods creates the pods that are marked as running in a given test so that they can be accessed by the
--- a/workflow/controller/controller_test.go
+++ b/workflow/controller/controller_test.go
@@ -312,12 +312,12 @@ func listPods(woc *wfOperationCtx) (*apiv1.PodList, error) {
 
 type with func(pod *apiv1.Pod)
 
-func withOutputs(cm string, v interface{}) with {
+func withOutputs(v interface{}) with {
 	switch x := v.(type) {
 	case string:
-		return withAnnotation(common.AnnotationKeyOutputs+"-"+cm, x)
+		return withAnnotation(common.AnnotationKeyOutputs, x)
 	default:
-		return withOutputs(cm, wfv1.MustMarshallJSON(x))
+		return withOutputs(wfv1.MustMarshallJSON(x))
 	}
 }
 func withProgress(v string) with { return withAnnotation(common.AnnotationKeyProgress, v) }
@@ -338,9 +338,7 @@ func withExitCode(v int32) with {
 }
 
 func withAnnotation(key, val string) with {
-	return func(pod *apiv1.Pod) {
-		pod.Annotations[key] = val
-	}
+	return func(pod *apiv1.Pod) { pod.Annotations[key] = val }
 }
 
 // createRunningPods creates the pods that are marked as running in a given test so that they can be accessed by the
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1232,16 +1232,7 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		new.PodIP = pod.Status.PodIP
 	}
 
-	if x, ok := pod.Annotations[common.AnnotationKeyOutputs]; ok {
-		woc.log.Warn("workflow uses legacy/insecure pod patch, see https://argoproj.github.io/argo-workflows/workflow-rbac/")
-		if new.Outputs == nil {
-			new.Outputs = &wfv1.Outputs{}
-		}
-		if err := json.Unmarshal([]byte(x), new.Outputs); err != nil {
-			new.Phase = wfv1.NodeError
-			new.Message = err.Error()
-		}
-	}
+	woc.updateNodeOuputs(pod, new)
 
 	new.HostNodeName = pod.Spec.NodeName
 
@@ -1259,12 +1250,13 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 	// We capture the exit-code after we look for the task-result.
 	// All other outputs are set by the executor, only the exit-code is set by the controller.
 	// By waiting, we avoid breaking the race-condition check.
-	if exitCode := getExitCode(pod); exitCode != nil {
-		if new.Outputs == nil {
-			new.Outputs = &wfv1.Outputs{}
-		}
-		new.Outputs.ExitCode = pointer.StringPtr(fmt.Sprint(*exitCode))
-	}
+	// TODO: noamg
+	// if exitCode := getExitCode(pod); exitCode != nil {
+	// 	if new.Outputs == nil {
+	// 		new.Outputs = &wfv1.Outputs{}
+	// 	}
+	// 	new.Outputs.ExitCode = pointer.StringPtr(fmt.Sprint(*exitCode))
+	// }
 
 	// if we are transitioning from Pending to a different state, clear out unchanged message
 	if old.Phase == wfv1.NodePending && new.Phase != wfv1.NodePending && old.Message == new.Message {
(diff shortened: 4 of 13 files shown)


workflow/controller/controller_test.go
```diff
@@ -279,7 +279,7 @@ func listPods(woc *wfOperationCtx) (*apiv1.PodList, error) {
 
 type with func(pod *apiv1.Pod)
 
-func withOutputs(v string) with  { return withAnnotation(common.AnnotationKeyOutputs, v) }
+func withOutputs(v string) with  { return withAnnotation(common.AnnotationKeyOutputs+"-main", v) }
```
alexec, line 282: this looks wrong to me

workflow/controller/operator.go
```diff
@@ -1265,9 +1255,47 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, node *wfv1.NodeStatu
 	return nil
 }
 
-func getExitCode(pod *apiv1.Pod) *int32 {
+func (woc *wfOperationCtx) updateNodeExitData(pod *apiv1.Pod, node *wfv1.NodeStatus) *wfv1.NodeStatus {
+	updated := false
+	containerName := common.MainContainerName
+	if node.Type == wfv1.NodeTypeContainer {
+		containerName = node.DisplayName
```
alexec, line 1262: interesting - I think this is a work-around - we should use some shared func to determine the container name - at least so the func acts are docs

workflow/executor/executor.go
```diff
@@ -531,24 +531,44 @@ func (we *WorkflowExecutor) SaveParameters(ctx context.Context) error {
...
 	tempLogsDir := "/tmp/argo/outputs/logs"
 	err := os.MkdirAll(tempLogsDir, os.ModePerm)
 	if err != nil {
-		return nil, argoerrs.InternalWrapError(err)
+		errs = append(errs, argoerrs.InternalWrapError(err))
+		return artifacts, errs
+	}
+
+	containerNames := we.Template.GetMainContainerNames()
+	for _, contianerName := range containerNames {
```
alexec, line 550: spelling

workflow/executor/executor.go
```diff
@@ -531,24 +531,44 @@ func (we *WorkflowExecutor) SaveParameters(ctx context.Context) error {
 }
 
 // SaveLogs saves logs
-func (we *WorkflowExecutor) SaveLogs(ctx context.Context) (*wfv1.Artifact, error) {
+func (we *WorkflowExecutor) SaveLogs(ctx context.Context) ([]wfv1.Artifact, []error) {
```
alexec, line 534: I don't think this should return mulitple errors - it not canonical Go - instead - could you add `containerName` as an argument and just call many times?

alexec (changes requested):
looks good so far - I think this should have an e2e test too

workflow/executor/executor.go
noam-codefresh: i moved the iteration over the `MainContainer` "outside" to the `waitContainer` func. it makes the function less concise.
another option i just tried out (haven't pushed) - 
```go
func (we *WorkflowExecutor) SaveAllLogs(ctx context.Context) ([]wfv1.Artifact, error) {
	logArts := make([]wfv1.Artifact, 0)
	if we.Template.SaveLogsAsArtifact() {
		return logArts, nil
	}

	// Saving logs
	log.Infof("Saving logs")
	tempLogsDir := "/tmp/argo/outputs/logs"
	err := os.MkdirAll(tempLogsDir, os.ModePerm)
	if err != nil {
		return logArts, err
	}

	containerNames := we.Template.GetMainContainerNames()
	for _, containerName := range containerNames {
		art, err := we.SaveLogs(ctx, tempLogsDir, containerName)
		if err != nil {
			we.AddError(err)
		} else if art != nil {
			logArts = append(logArts, *art)
		}
	}

	return logArts, nil
}
```

this will keep the iteration and error handling inside the WorkflowExecutor code, which is a bit different from how it is handled in the other calls from `waitContainer`. do you think this is preferable?

workflow/controller/operator.go
```diff
@@ -1265,9 +1255,43 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, node *wfv1.NodeStatu
 	return nil
 }
 
-func getExitCode(pod *apiv1.Pod) *int32 {
+func (woc *wfOperationCtx) updateNodeExitData(pod *apiv1.Pod, node *wfv1.NodeStatus) *wfv1.NodeStatus {
+	updated := false
+	containerName := node.GetContainerName()
+	if exitCode := getExitCode(pod, containerName); exitCode != nil {
```
alexec, line 1261: interesting --- this was  a bug

alexec: the existing code is inconsistent - some of it update variables like `newPhase` and some directly updates `node.Phase` - this will cause bugs (e.g. when updating node.Phase, we don't set updated=true)

I think we need to refactor it - but not in this PR

workflow/executor/executor.go
```diff
@@ -724,19 +715,25 @@ func (we *WorkflowExecutor) CaptureScriptResult(ctx context.Context) error {
...
+
+	key := fmt.Sprintf("%s-%s", common.AnnotationKeyOutputs, containerName)
 	outputBytes, err := json.Marshal(outputs)
 	if err != nil {
+		log.Infof("in error: %s", err.Error())
 		return argoerrs.InternalWrapError(err)
 	}
-	return we.AddAnnotation(ctx, common.AnnotationKeyOutputs, string(outputBytes))
+
+	log.Infof("returning")
```
alexec, line 735: always use `Info` not `Infof`

workflow/executor/executor.go
```diff
@@ -724,19 +715,25 @@ func (we *WorkflowExecutor) CaptureScriptResult(ctx context.Context) error {
...
 
 	if !outputs.HasOutputs() {
 		return nil
 	}
-	log.Infof("Annotating pod with output")
+
+	key := fmt.Sprintf("%s-%s", common.AnnotationKeyOutputs, containerName)
 	outputBytes, err := json.Marshal(outputs)
 	if err != nil {
+		log.Infof("in error: %s", err.Error())
```
alexec, line 731: always do structured logging - so use log.WithError, never use Infof

workflow/executor/executor.go
```diff
@@ -724,19 +715,25 @@ func (we *WorkflowExecutor) CaptureScriptResult(ctx context.Context) error {
 // AnnotateOutputs annotation to the pod indicating all the outputs.
 func (we *WorkflowExecutor) AnnotateOutputs(ctx context.Context, logArt *wfv1.Artifact) error {
 	outputs := we.Template.Outputs.DeepCopy()
+	containerName := common.MainContainerName
 	if logArt != nil {
+		containerName = strings.TrimSuffix(logArt.Name, "-logs")
```
alexec, line 720: this is too hacky - can we find another way?

alexec (changes requested):

alexec:
See this refactoring:

https://github.com/argoproj/argo-workflows/pull/7998

workflow/executor/executor.go
noam-codefresh: i moved the call to `AnnotateOutputs` into the loop on the containerNames, right after the call for `SaveLogs`, so now i send the name as a parameter.
but i can also just change `SaveLogs` to both save the logs in the artifact-repo *and* the annotation on the pod. and then this loop will only call a single func.
or wrap `SaveLogs` and `AnnotateOutputs` in a new func. whichever you prefer.

workflow/executor/executor.go
noam-codefresh: removed this debug log completely

workflow/executor/executor.go
noam-codefresh: nvm. i realized `SaveLogs` shouldn't contain `AnnotateOutputs`. the 2nd might happen without the 1st.
i think i got it now in the correct place - at the end of the `waitContainer` call, saving the logs if needed, and afterwards (in any case) save the output to the pod's annotation (using the containerName)

cmd/argoexec/commands/wait.go
```diff
@@ -61,10 +59,24 @@ func waitContainer(ctx context.Context) error {
...
+			// Saving logs
+			log.Info("Saving logs with new loop")
+			art, err = wfExecutor.SaveLogs(ctx, containerName)
+			if err != nil {
+				wfExecutor.AddError(err)
+			}
+		}
+
+		// Annotating pod with output
+		err = wfExecutor.AnnotateOutputs(ctx, art, containerName)
```
alexec, line 76: this will result in N patches, but we only want to do this once, can you move out of the loop please?

cmd/argoexec/commands/wait.go
```diff
@@ -61,10 +59,24 @@ func waitContainer(ctx context.Context) error {
 	if err != nil {
 		wfExecutor.AddError(err)
 	}
-	// Annotating pod with output
-	err = wfExecutor.AnnotateOutputs(ctx, logArt)
-	if err != nil {
-		wfExecutor.AddError(err)
+
+	containerNames := wfExecutor.Template.GetMainContainerNames()
```
alexec, line 63: could this all go inside SaveLogs?

workflow/controller/operator.go
```diff
@@ -1284,9 +1269,38 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 	return nil
 }
 
-func getExitCode(pod *apiv1.Pod) *int32 {
+func (woc *wfOperationCtx) updateNodeExitData(pod *apiv1.Pod, node *wfv1.NodeStatus) {
```
alexec, line 1272: i think this could be called `updateNodeOuputs`

workflow/executor/executor.go
```diff
@@ -530,25 +530,22 @@ func (we *WorkflowExecutor) SaveParameters(ctx context.Context) error {
...
 }
 
-// SaveLogs saves logs
-func (we *WorkflowExecutor) SaveLogs(ctx context.Context) (*wfv1.Artifact, error) {
-	if !we.Template.SaveLogsAsArtifact() {
-		return nil, nil
-	}
-	log.Infof("Saving logs")
+// SaveLogs saves a single container logs
+func (we *WorkflowExecutor) SaveLogs(ctx context.Context, containerName string) (*wfv1.Artifact, error) {
```
alexec, line 534: maybe call this `SaveContainerLogs`

workflow/executor/executor.go
```diff
@@ -530,25 +530,22 @@ func (we *WorkflowExecutor) SaveParameters(ctx context.Context) error {
 	return nil
 }
 
-// SaveLogs saves logs
-func (we *WorkflowExecutor) SaveLogs(ctx context.Context) (*wfv1.Artifact, error) {
```
alexec, line 534: maybe keep this func

alexec (changes requested):

cmd/argoexec/commands/wait.go
noam-codefresh: i moved the iteration over the containerNames into `SaveLogs`, and from there (inside the `WorkflowExecutor`) i made the calls to `we.AddError(err)` when needed.

cmd/argoexec/commands/wait.go
noam-codefresh: to get it outside of the loop i'd need to change `AnnotateOutputs` to accept an array of `Artifact/ContainerName` (maybe just add the `ContainerName` into the artifact struct?), update `AddAnnotation` to accept an array of `key/value` pairs, as well as `AddPodAnnotation` and `addPodMetadata`.
do you think there is a simpler way to do it?

maybe i should put all of the pod's outputs in a single annotation as a JSON object with the containerNames as keys? that might be simpler in that regard.

alexec:
Sorry - I'm afraid I think I did a u-turn on `SaveLogs`.

noam-codefresh:
i fixed the last e2e test fail i had. i hope the PR is ready now

tczhao (commented):
perhaps an e2e test for containerSet archiveLogs as well?

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2731,14 +2740,14 @@ func (out *Outputs) HasParameters() bool {
...
 }
 
-const MainLogsArtifactName = "main-logs"
+const LogsSuffix = "-logs"
 
 func (out *Outputs) HasLogs() bool {
 	if out == nil {
 		return false
 	}
 	for _, a := range out.Artifacts {
```
alexec, line 2749: one minor thought, this might break stuff, as users might have named their artifacts `mythink-log`. Is there any other field we can use?

workflow/controller/operator.go
```diff
@@ -1243,17 +1243,6 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
...
-		}
-		if err := json.Unmarshal([]byte(x), new.Outputs); err != nil {
-			new.Phase = wfv1.NodeError
-			new.Message = err.Error()
-		}
-	}
-
 	new.HostNodeName = pod.Spec.NodeName
 
 	if !new.Progress.IsValid() {
```
alexec, line 1248: Should this be moved into `updateNodeOutputs` ?

cmd/argoexec/commands/wait.go
```diff
@@ -61,11 +57,7 @@ func waitContainer(ctx context.Context) error {
 	if err != nil {
 		wfExecutor.AddError(err)
 	}
-	// Annotating pod with output
```
alexec, line 64: This does not seem correct to me. We should not have one annotation per container. Instead, we should have a single output for all containers.

alexec (changes requested):

alexec:
@noam-codefresh we're doing a lot of circuts on this PR review. If you want, we could sit down for 1h on Zoom and bash it out.

pkg/apis/workflow/v1alpha1/workflow_types.go
noam-codefresh: i thought maybe adding some `IsLogs bool` field to the Artifact, or maybe some `Type ArtifactType` enum, just to differentiate between log artifact and other types.

noam-codefresh:
@alexec - i would be happy to schedule a meeting to go over the suggested changes.
what days and hours will be good for you? i think we have a 10h time difference, so i guess your morning/my evening will work.

alexec:
I'm in PST, you're in?

noam-codefresh:
i am in Israel time - GMT+2.

alexec:
Ok. That is 10h difference, so I don't think we will be Zooming. Lets just continue with the way we were doing it.

noam-codefresh:
ok.
i was thinking about maybe changing the `NodeResult struct` to contain several `Outputs`. so either
```golang
Outputs  map[string]Outputs // cotainerName -> output map
```
or 

```golang
Outputs []Output
```
(and have the `Output struct` also contain some `ContainerName string` field).
either way, it will also effect the `WorkflowTaskSet struct`, and i am not %100 sure i understand where it is being used.

another possible option - create a separate `WorkflowTaskResult` for each container. if i understand correctly, it is currently being created once per `Pod`, so now the containerSet containers overwrite each others' result. if i add some `containerName string` parameter to the `upsertTaskResult` func, i will be able to separate them, and then re-struct it in the controller to put the information back on the `Workflow`.

which approach do you prefer? or maybe there is a different way?

alexec:
I don’t think either approach would work because they would break any running workflows at time of upgrade. But, I think you could add a `container` field to `Artifact`, but you always need to name the artifact something like `main-logs` so I think that has to be the way to name them.

noam-codefresh:
@alexec - i am not sure i understand. do you think i should add the `container` field? and some `isLogs` field? or just get the container name from the artifact name by removing the `-logs` suffix, and keep this naming convention?
also - should the containerSet logs archiving work also when using the legacy pod annotation way? or should i just concentrate on the new `WorkflowTaskResult`?

if you have time some morning, i'd be happy for a zoom meeting. whenever you are comfortable, i can do my evening-night (my 8pm/your 11am?)

alexec:
Hi @noam-codefresh I think maybe I'm overthinking this. Lets assume that `*-logs` are always logs.

noam-codefresh:
i hope i got it now. i collect all Artifacts, and then put them in the `WorkflowTaskResult` (or fallback to the annotation) together.
i'll add another e2e test to make sure the annotation fallback works as required, and i'll mark it ready for review in the coming days.

noam-codefresh:
i have rebased by branch, and i checked that the legacy support for passing output information on the Pod works with containerset logs as well.
is an e2e test to check the legacy support automatically required?

noam-codefresh:
other than that, i think the pr can be re-opened, and is ready for review.
@alexec

alexec:
@noam-codefresh can I please ask you to open a new PR? I can reopn this one at all, see grayed out button:

image

agilgur5:
Replaced by #8491

PR #7928, opened 2022-02-18
feat: Add Python plugin. Fixes #7913

### LINKED ISSUE #7913: Python plugin
A Python plugin that allows you to execute basic Python scripts.

---

alexec:

Fixes #7913

### DIFF
--- a/plugins/executor/python/python-executor-plugin-configmap.yaml
+++ b/plugins/executor/python/python-executor-plugin-configmap.yaml
@@ -1,0 +1,128 @@
+# This is an auto-generated file. DO NOT EDIT
+apiVersion: v1
+data:
+  sidecar.container: |
+    args:
+    - |
+      import json
+      from http.server import BaseHTTPRequestHandler, HTTPServer
+
+
+      class Plugin(BaseHTTPRequestHandler):
+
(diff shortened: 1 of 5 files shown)


plugins/executor/python/python-executor-plugin-configmap.yaml
```diff
@@ -0,0 +1,117 @@
+# This is an auto-generated file. DO NOT EDIT
```
terrytangyuan, line 1: Perhaps I missed this somewhere but have we documented how to auto-generate this?

plugins/executor/python/server.py
```diff
@@ -0,0 +1,74 @@
+import json
```
terrytangyuan, line 1: Should we expect plugin authors to write some tests?

alexec: good question - we should discus this - I think we should expect quality - band that may need tests - what kind of tests?

plugins/executor/python/python-executor-plugin-configmap.yaml
alexec: `argo executor-plugin build .`

PR #7929, opened 2022-02-18
fix: add TLS config option to HTTP template. Fixes #7390

### LINKED ISSUE #7390: TLS config for the HTTP workflow trigger
# Summary

TLS config options for HTTP workflow trigger

# Use Cases

Endpoints with self signed certificate

```
    - name: check-http
      http:
        successCondition: "response.statusCode == 200"
        url: "https://xxx.example"
```

```
Get "https://xxx.example": x509: certificate signed by unknown authority
```

---

rohankmr414:

Fixes #7390


workflow/controller/agent.go
```diff
@@ -128,6 +128,13 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 					Image:           woc.controller.executorImage(),
 					ImagePullPolicy: woc.controller.executorImagePullPolicy(),
 					Env:             envVars,
+
+					VolumeMounts: []apiv1.VolumeMount{{
+						Name:      "cert",
+						MountPath: "/etc/ssl/certs/ca-certificates",
```
alexec, line 134: This file will already exist. So that it why the tests are failing, you're replacing a working file, with an non-working directory.

alexec (commented):
Looking at this PR, I think we should make some changes. It is a bit more complex that first looks. 

The agent is created by the controller as you've found. We should mount a volume as you've tried . However, I think we a way to identify which secret should be mounted and where.

I suggest we use the existing secret informer in the controller for this.

We can look up secrets based on same namespace as workflow and has special name: `argo-workflows-agent-ca-certificates`. If that it exists, we mount it as a secret volume.

```yaml
kind: Secret
metadata: 
  name: argo-workflows-agent-ca-certificates
stringData:
  ca-certificates: abc123...
```

workflow/controller/agent.go
```diff
@@ -128,6 +164,9 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 					Image:           woc.controller.executorImage(),
 					ImagePullPolicy: woc.controller.executorImagePullPolicy(),
 					Env:             envVars,
+
+					VolumeMounts: woc.getCertVolumeMount(ctx, "argo-workflows-agent-ca-certificates"),
```
alexec, line 168: move cont to `common.CACertificatesVolumeMountName`

rohankmr414: done

workflow/controller/agent.go
```diff
@@ -70,6 +70,52 @@ func assessAgentPodStatus(pod *apiv1.Pod) (wfv1.WorkflowPhase, string) {
...
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (woc *wfOperationCtx) getCertVolumeMount(ctx context.Context, name string) []apiv1.VolumeMount {
+	exists, err := woc.secretExists(ctx, name)
+	if err != nil {
```
alexec, line 86: You should surface this error

workflow/controller/agent.go
```diff
@@ -70,6 +70,52 @@ func assessAgentPodStatus(pod *apiv1.Pod) (wfv1.WorkflowPhase, string) {
...
+			ReadOnly:  true,
+		}}
+	}
+	return nil
+}
+
+func (woc *wfOperationCtx) getCertVolume(ctx context.Context, name string) []apiv1.Volume {
+	exists, err := woc.secretExists(ctx, name)
+	if err != nil {
+		woc.log.WithError(err).Errorf("Failed to check if secret %s exists", name)
```
alexec, line 103: Please bubble/surface error

workflow/controller/agent.go
```diff
@@ -70,6 +70,50 @@ func assessAgentPodStatus(pod *apiv1.Pod) (wfv1.WorkflowPhase, string) {
...
+	if err != nil {
+		if apierr.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (woc *wfOperationCtx) getCertVolumeMount(ctx context.Context, name string) ([]apiv1.VolumeMount, error) {
```
alexec, line 84: what about combining the two funcs to return `(apiv1.VolumeMount, apiv1.Volume, error)`

alexec (changes requested):
needs an e2e test, just adding the `argo-workflow-agent-ca-certificates` as a resource should be enough

tczhao:
would be good to amend the examples as well, to let people know how to add TLS

workflow/controller/agent.go
```diff
@@ -70,6 +70,43 @@ func assessAgentPodStatus(pod *apiv1.Pod) (wfv1.WorkflowPhase, string) {
...
+
+		certVolumeMount := []apiv1.VolumeMount{{
+			Name:      name,
+			MountPath: "/etc/ssl/certs/ca-certificates",
+			ReadOnly:  true,
+		}}
+
+		return certVolume, certVolumeMount, nil
+	}
+	return nil, nil, nil
```
simster7, line 107: If `!exists` we return `nil` here... have we made sure that wherever `certVolume, certVolumeMount` is used `nil` is an acceptable value? Otherwise we could have panics

rohankmr414:
> needs an e2e test, just adding the `argo-workflow-agent-ca-certificates` as a resource should be enough

@alexec where do we add the manifests for the resources for the e2e tests? Is it in `test/e2e/manifests/mixins`?

alexec:
> > needs an e2e test, just adding the `argo-workflow-agent-ca-certificates` as a resource should be enough
> 
> @alexec where do we add the manifests for the resources for the e2e tests? Is it in `test/e2e/manifests/mixins`?

Yes.

workflow/controller/agent.go
rohankmr414: yes you are right. `nil` is not an acceptable value inside the `[]apiv1.Volume` and `[]apiv1.VolumeMount` so I've added an if else logic to prevent errors.

test/e2e/manifests/mixins/argo-workflows-agent-ca-certificates.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: argo-workflows-agent-ca-certificates
+stringData:
+  ca-certificates: >
+    -----BEGIN CERTIFICATE-----
```
alexec, line 7: can you add a comment for how to generate this?

rohankmr414: done @alexec

alexec: cool - did you run it with invalid cert? to make sure at least one test fails?

rohankmr414: @alexec On mounting at `/etc/ssl/cert.pem` the pod creation fails with the following error: 
`Error: failed to create containerd task: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:75: mounting "/var/lib/kubelet/pods/4cf0518f-35a4-4c01-b442-6f03bd8d3bb6/volumes/kubernetes.io~secret/argo-workflows-agent-ca-certificates" to rootfs at "/etc/ssl/cert.pem" caused: mount through procfd: not a directory: unknown`

Same for `/etc/ssl/ca-certificates.crt`

Is there a limitation on mounts?

rohankmr414: Okay so kubernetes does not allow us to mount secrets as a file. We have to mount the secret as a directory and I guess the key will be the filename and the value will be the content of the file in that directory.

rohankmr414: @alexec the workflow does give an error `x509: certificate signed by unknown authority` on providing an invalid certificate. I guess we can merge now.

alexec (approved):

PR #7930, opened 2022-02-18
feat!: Remove non-Emissary executors

alexec:

Proof-of-concept to explore the impact of these changes.


terrytangyuan:
Any findings?

alexec:
Two

- It was not as hard as I expected it to be. Which is good news. 
- I think we can remove the `pod patch` permission, by using the Taskset to send data back to the controller.

PR #7952, opened 2022-02-22
Fixes# specify the cmd of influxdb

runninghai:
Fixes #example/daemon-step.yaml init problem
>   Warning  WorkflowNodeError    9m51s  workflow-controller  Error node daemon-step-zn9tk[0].influx(0): container "main" in template "influxdb", does not have the command specified: when using the emissary executor you must either explicitly specify the command, or list the image's command in the index: https://argoproj.github.io/argo-workflows/workflow-executors/#emissary-emissary

to fix this problem, I add the init cmd in template

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1300,6 +1300,7 @@ spec:
       limit: 10                         # retry container if it fails
     container:
       image: influxdb:1.2
+      command: ["influxd"]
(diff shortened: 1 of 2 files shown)


terrytangyuan (changes requested):
Thanks. Could you fix the checks?

agilgur5:
This is no longer necessary after #8344

PR #7960, opened 2022-02-22
fix: Report container, plugin and HTTP  progress. Fixes #7918

### LINKED ISSUE #7918: Plugins don't count toward progress

### LINKED ISSUE #7925: Individual containers within container set do not count towards progress
This may be a minor issue, creating for reference.

The progress of a workflow could a whole container set a 1 toward progress. This is because we count pods, not containers, towards progress.

alexec:

Fixes #7918
Fixes #7925 

This PR also include what I think is a correction to behaviour. Previously failed/error nodes were considered complete. That seems incorrect to me. I think only succeded nodes should count.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ git-ask-pass.sh
 /go-diagrams/
 /.run/
 sdks/python/client/dist/*
+/cmd/argoexec/commands/test.txt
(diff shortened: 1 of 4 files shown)


terrytangyuan (approved):
Great. The logic is much more intuitive now.

PR #7987, opened 2022-02-23
fix: Reduce agent permissions. Fixes #7986

### LINKED ISSUE #7986: Reduce agent permissions
# Summary

The agent should not be able to update the task set, only patch its status. This means the agent cannot add work to the workflow. 

This got overlooking in the implementation and code review.

---

alexec:

Fixes #7986

Removing `pod patch` is not as useful, if the attacker can still create work using the task-set. This fixes that.


alexec:
@jessesuen can I get your rubber-stamp approval of the proto change? It is only adding status sub-resource

workflow/executor/agent.go
```diff
@@ -176,7 +176,7 @@ func (ae *AgentExecutor) patchWorker(ctx context.Context, taskSetInterface v1alp
 
 			ae.log.Info("Processing Patch")
 
-			_, err = taskSetInterface.Patch(ctx, ae.WorkflowName, types.MergePatchType, patch, metav1.PatchOptions{})
+			_, err = taskSetInterface.Patch(ctx, ae.WorkflowName, types.MergePatchType, patch, metav1.PatchOptions{}, "status")
```
alexec, line 179: this can be made a non-breaking change by trying to patch "status" first, if 403 error, try and patch resource instead (print warning about permissions)

sarabala1979 (approved):
LGTM

Issue #5405, opened 2021-03-15
[Proposal] Support resubmit/retry on archived workflows


terrytangyuan:
# Summary

What change needs making?

Support resubmit/retry on archived workflows.

# Use Cases

When would you use this?

Currently `argo retry/resubmit` only supports workflows resources that currently exist in the cluster. For clusters where workflows are GC-ed immediately after they are completed, it's not possible to retry/resubmit a workflow. There are workaound for this, e.g. saving the workflow resource manually either as a plain YAML or into archived workflow before we GC it, and then do a `kubectl create` before we use `argo retry/resubmit`. It would be better and less hacky if we have a CLI for this.

# API/Implementation Discussion

We can implement this in one of the following ways:
1. `argo retry/resubmit wf-name --from-archive`
2. `argo archive retry/resubmit`
3. `argo retry/resubmit --source=location-to-saved-wf.yaml`

Tasks
- [x] #7906
- [x] #7907
- [x] #7908
- [x] #7909
- [x] #7910
- [ ] #7911

Any preference?

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

alexec:
We should also provide this feature in the UI.

dpadhiar:
I will take this issue

terrytangyuan:
Wow this got popular. 👍🏻 @dpadhiar assigned to you.

PR #7988, opened 2022-02-23
feat: add retry API for archived workflows. Fixes #7906

### LINKED ISSUE #7906: Add Retry  API  on Archive workflow

dpadhiar:
Signed-off-by: Dillen Padhiar <dpadhiar99@gmail.com>

Fixes #7906

Previously workflows were only able to be retried if they existed in the cluster. If a workflow is GC-ed immediately after completion, you must save the workflow manually as a plain YAML or into workflow archive, then `kubectl create` before using `argo retry`. This API allows for archived workflows to be created and retried removing the need for this otherwise hacky method.


dpadhiar:
@sarabala1979

pkg/apiclient/workflowarchive/workflow-archive.proto
```diff
@@ -24,6 +24,13 @@ message ListArchivedWorkflowLabelKeysRequest {
 message ListArchivedWorkflowLabelValuesRequest {
     k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 1;
 }
+message RetryArchivedWorkflowRequest {
```
alexec, line 27: this should be same parameters as normal retry - can you please check

persist/sqldb/workflow_archive.go
```diff
@@ -72,6 +73,10 @@ func (r *workflowArchive) IsEnabled() bool {
 	return true
 }
 
+func (r *workflowArchive) ValidateWorkflow(wf *wfv1.Workflow) error {
```
alexec, line 76: this responsibily does not fit within the archive - what does validation to do with archiving?

server/workflowarchive/archived_workflow_server.go
```diff
@@ -188,3 +191,24 @@ func (w *archivedWorkflowServer) ListArchivedWorkflowLabelValues(ctx context.Con
...
+func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req *workflowarchivepkg.RetryArchivedWorkflowRequest) (*wfv1.Workflow, error) {
+	wfClient := auth.GetWfClient(ctx)
+	kubeClient := auth.GetKubeClient(ctx)
+
+	wf, err := w.wfArchive.GetWorkflow(req.Uid)
+	if err != nil {
+		return nil, err
+	}
+
+	err = w.wfArchive.ValidateWorkflow(wf)
```
alexec, line 204: this is correct - we should not retry workflows that belong to another instance - but is this check not redundant? GetWorkflow should never return a workflow which does not have correct instance ID

workflow/util/util.go
```diff
@@ -748,17 +752,29 @@ func RetryWorkflow(ctx context.Context, kubeClient kubernetes.Interface, hydrato
...
 	if err != nil {
 		return nil, err
 	}
+	return updated, err
+}
+
+// retryWorkflow takes a wf in method signature instead and has boolean to determine if this is from archive or not - archive means we must create the workflow before update
+func retryWorkflow(ctx context.Context, kubeClient kubernetes.Interface, hydrator hydrator.Interface, wfClient v1alpha1.WorkflowInterface, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string, retryArchive bool) (*wfv1.Workflow, error) {
+
 	switch wf.Status.Phase {
```
alexec, line 772: I think we should have a new func `prepareWorkflowForRettry`, specifically it should not take any `kubeClient`, `hydrator`, or `wfClient` - all it should do is take a workflow and make it ready for retrying

the calling function can do any hydration or whatever

workflow/util/util.go
```diff
@@ -872,6 +888,17 @@ func retryWorkflow(ctx context.Context, kubeClient kubernetes.Interface, hydrato
 		newWF.Status.StoredTemplates[id] = tmpl
 	}
 
+	if retryArchive {
```
alexec, line 891: behaviour flags like this can be a code smell

alexec (commented):
Great work identifying that you cannot just resubmit an archived workflow.

server/workflowarchive/archived_workflow_server.go
dpadhiar: This check comes from following the `RetryWorkflow` method and Bala and I discussed that it is redundant with the Get. In this case, we could remove `ValidateWorkflow` altogether for wfArchive also addressing the other issue that it doesn't belong with workflow archive.

pkg/apiclient/workflowarchive/workflow-archive.proto
dpadhiar: This is the same parameters as a normal retry however it also includes the `uid` to be passed for `GetArchivedWorkflow`.

workflow/util/util.go
dpadhiar: To address this, we can remove the flag and call `wfClient.create()` in `RetryArchiveWorkflow` after refactoring `retryWorkflow` into `prepareWorkflowForRetry`. This way, we do not need to rely on a flag to determine whether we are retrying an archived workflow or not.

alexec (changes requested):
changes requested

workflow/util/util.go
dpadhiar:
```
func prepareWorkflowForRetry(wf *wfv1.Workflow) (*wfv1.Workflow, error) {

	switch wf.Status.Phase {
	default:
		return nil, errors.Errorf(errors.CodeBadRequest, "workflow must be Failed/Error to retry")
	}

	newWF := wf.DeepCopy()

	// Delete/reset fields which indicate workflow completed
	delete(newWF.Labels, common.LabelKeyCompleted)
	delete(newWF.Labels, common.LabelKeyWorkflowArchivingStatus)
	newWF.Status.Conditions.UpsertCondition(wfv1.Condition{Status: metav1.ConditionFalse, Type: wfv1.ConditionTypeCompleted})
	newWF.ObjectMeta.Labels[common.LabelKeyPhase] = string(wfv1.NodeRunning)
	newWF.Status.Phase = wfv1.WorkflowRunning
	newWF.Status.Nodes = make(wfv1.Nodes)
	newWF.Status.Message = ""
	newWF.Status.StartedAt = metav1.Time{Time: time.Now().UTC()}
	newWF.Status.FinishedAt = metav1.Time{}
	newWF.Spec.Shutdown = ""
	if newWF.Spec.ActiveDeadlineSeconds != nil && *newWF.Spec.ActiveDeadlineSeconds == 0 {
		// if it was terminated, unset the deadline
		newWF.Spec.ActiveDeadlineSeconds = nil
	}

	newWF.Status.StoredTemplates = make(map[string]wfv1.Template)
	for id, tmpl := range wf.Status.StoredTemplates {
		newWF.Status.StoredTemplates[id] = tmpl
	}

	return newWF, nil
}
```

dpadhiar: Should this function replace `retryWorkflow`? In that case, we would move all the node logic into the calling functions `RetryWorkflow` and `RetryArchiveWorkflow`

server/workflowarchive/archived_workflow_server.go
```diff
@@ -188,3 +191,25 @@ func (w *archivedWorkflowServer) ListArchivedWorkflowLabelValues(ctx context.Con
...
+func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req *workflowarchivepkg.RetryArchivedWorkflowRequest) (*wfv1.Workflow, error) {
+	wfClient := auth.GetWfClient(ctx)
+	kubeClient := auth.GetKubeClient(ctx)
+
+	wf, err := w.wfArchive.GetWorkflow(req.Uid)
+	if err != nil {
+		return nil, err
+	}
+
+	err = w.hydrator.Hydrate(wf)
```
alexec, line 204: this is a no-op - workflows in the archive will always be hydrated

workflow/util/util.go
```diff
@@ -735,36 +735,60 @@ func convertNodeID(newWf *wfv1.Workflow, regex *regexp.Regexp, oldNodeID string,
 }
 
 // RetryWorkflow updates a workflow, deleting all failed steps as well as the onExit node (and children)
-func RetryWorkflow(ctx context.Context, kubeClient kubernetes.Interface, hydrator hydrator.Interface, wfClient v1alpha1.WorkflowInterface, name string, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, error) {
+func RetryWorkflow(ctx context.Context, podIf v1.PodInterface, wfClient v1alpha1.WorkflowInterface, name string, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, error) {
+	var updatedWf *wfv1.Workflow
+	err := waitutil.Backoff(retry.DefaultRetry, func() (bool, error) {
```
alexec, line 740: I don't think we should do backoff deep in the stack. In fact, not at all in this case. The client will always need to implement some kind of retry themselves, so 2n retries will happen.

workflow/util/util.go
```diff
@@ -735,36 +735,60 @@ func convertNodeID(newWf *wfv1.Workflow, regex *regexp.Regexp, oldNodeID string,
...
+		var err error
+		wf, err := wfClient.Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		updatedWf, err = prepareWorkflowForRetry(ctx, wf, podIf, restartSuccessful, nodeFieldSelector)
+		if err != nil {
+			return false, err
+		}
+		updatedWf, err = wfClient.Update(ctx, updatedWf, metav1.UpdateOptions{})
```
alexec, line 750: could you make this an upsert? then it would work for both archived and normal workflows, and would also be idempotent and therefore robust

workflow/util/util.go
alexec: `prepareWorkflowForRetry` is great

workflow/util/util_test.go
```diff
@@ -811,7 +811,7 @@ func TestDeepDeleteNodes(t *testing.T) {
 	ctx := context.Background()
 	wf, err := wfIf.Create(ctx, origWf, metav1.CreateOptions{})
 	if assert.NoError(t, err) {
-		newWf, err := RetryWorkflow(ctx, kubeClient, hydratorfake.Noop, wfIf, wf.Name, false, "")
+		newWf, err := RetryWorkflow(ctx, kubeClient.CoreV1().Pods(wf.ObjectMeta.Namespace), wfIf, wf.Name, false, "")
```
alexec, line 814: minor - simpler as `wf.Namespace`, no need for `ObjectMeta`

alexec (commented):
Some quick comments.

What happens if you move the Update/Create to the service level?

server/workflowarchive/archived_workflow_server.go
dpadhiar: In this case, I think we can remove the hydrator altogether from the workflow archive since it was only added for this call.

dpadhiar:
> What happens if you move the Update/Create to the service level?

Functionality is the same if we move the update/create calls to the service level. We can remove the `RetryArchiveWorkflow` function altogether since `RetryWorkflow` will only just prepare a workflow for retry and the server will either update or create.

alexec:
I think that's the right thing to do. The current util finc combines two responsibilies:

* Preparing the workflow for retry.
* Persisting the workflow.

It if only does the first, the second responsibly can be moved to the service layer.

https://en.wikipedia.org/wiki/Single-responsibility_principle

server/workflowarchive/archived_workflow_server.go
```diff
@@ -188,3 +191,28 @@ func (w *archivedWorkflowServer) ListArchivedWorkflowLabelValues(ctx context.Con
...
+
+	wf, err = util.RetryWorkflow(ctx, kubeClient.CoreV1().Pods(wf.Namespace), wfClient.ArgoprojV1alpha1().Workflows(req.Namespace), wf, req.RestartSuccessful, req.NodeFieldSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	wf, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Create(ctx, wf, metav1.CreateOptions{})
+	if err != nil {
+		if apierr.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("workflow already exists : %s", err)
```
alexec, line 212: either:

* We should we ignore this error for idempotence? @sarabala1979 
* Do not wrap it .

server/workflowarchive/archived_workflow_server.go
```diff
@@ -19,10 +20,12 @@ import (
 	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow"
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v3/server/auth"
+	"github.com/argoproj/argo-workflows/v3/workflow/util"
 )
 
 type archivedWorkflowServer struct {
 	wfArchive sqldb.WorkflowArchive
+	// hydrator  hydrator.Interface
```
alexec, line 28: delete

server/workflow/workflow_server.go
```diff
@@ -319,10 +319,21 @@ func (s *workflowServer) RetryWorkflow(ctx context.Context, req *workflowpkg.Wor
...
 	if err != nil {
 		return nil, err
 	}
+
+	wf, err = util.RetryWorkflow(ctx, kubeClient.CoreV1().Pods(wf.Namespace), wfClient.ArgoprojV1alpha1().Workflows(req.Namespace), wf, req.RestartSuccessful, req.NodeFieldSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	wf, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Update(ctx, wf, metav1.UpdateOptions{})
```
alexec, line 332: any dehydrate too.

dpadhiar: Are we dehydrating the wf returned from `RetryWorkflow` or should we differentiate the two?

dpadhiar: Added.

alexec (commented):
can you please add an e2e (API) test?

terrytangyuan:
Great work! I’ll review this soon as well since I originally proposed/requested this feature in https://github.com/argoproj/argo-workflows/issues/5405.

terrytangyuan (commented):
LGTM. Do you also plan to add UI support for this?

dpadhiar:
> LGTM. Do you also plan to add UI support for this?

Yes the plan is to next implement the Resubmit API then CLI and UI for both.

pkg/apiclient/workflowarchive/workflow-archive.proto
```diff
@@ -41,4 +48,10 @@ service ArchivedWorkflowService {
     rpc ListArchivedWorkflowLabelValues (ListArchivedWorkflowLabelValuesRequest) returns (github.com.argoproj.argo_workflows.v3.pkg.apis.workflow.v1alpha1.LabelValues) {
         option (google.api.http).get = "/api/v1/archived-workflows-label-values";
     }
+    rpc RetryArchivedWorkflow (RetryArchivedWorkflowRequest) returns (github.com.argoproj.argo_workflows.v3.pkg.apis.workflow.v1alpha1.Workflow) {
+        option (google.api.http) = {
+			put: "/api/v1/archived-workflows/{namespace}/{uid}/retry"
```
dpadhiar, line 53: Should this request be `/api/v1/archived-workflows/{uid}/retry` instead since no other workflow archive request uses the namespace?

@terrytangyuan @sarabala1979

terrytangyuan (approved):
LGTM.

workflow/util/util.go
```diff
@@ -735,36 +735,35 @@ func convertNodeID(newWf *wfv1.Workflow, regex *regexp.Regexp, oldNodeID string,
...
 
 // RetryWorkflow updates a workflow, deleting all failed steps as well as the onExit node (and children)
-func RetryWorkflow(ctx context.Context, kubeClient kubernetes.Interface, hydrator hydrator.Interface, wfClient v1alpha1.WorkflowInterface, name string, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, error) {
-	var updated *wfv1.Workflow
-	err := waitutil.Backoff(retry.DefaultRetry, func() (bool, error) {
-		var err error
-		updated, err = retryWorkflow(ctx, kubeClient, hydrator, wfClient, name, restartSuccessful, nodeFieldSelector)
-		return !errorsutil.IsTransientErr(err), err
-	})
+func RetryWorkflow(ctx context.Context, podIf v1.PodInterface, wfClient v1alpha1.WorkflowInterface, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, error) {
```
alexec, line 738: can we follow the pattern we found in `Resubmit`, i.e. create a `FormulateResubmitWorkflow` and move all the client-go calls into the service

alexec (changes requested):

test/e2e/argo_server_test.go
```diff
@@ -1302,6 +1326,24 @@ spec:
...
 
+	// we have to delete wf from namespace before retrying
+	s.Run("DeleteForRetry", func() {
+		s.e().DELETE("/api/v1/workflows/argo/" + name).
+			Expect().
+			Status(200)
+	})
+
+	s.Run("Retry", func() {
+		s.Need(fixtures.BaseLayerArtifacts)
```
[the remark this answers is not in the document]
dpadhiar: Currently, retry archived workflow will fail and return a 500 internal server error if the original workflow isn't deleted so for testing purposes, this is required.

workflow/util/util.go
dpadhiar: We can remove wfClient as a whole from the RetryWorkflow and RetryArchivedWorkflow methods since they no longer are used at all in the helper `prepareWorkflowForRetry`

alexec: yes please

workflow/util/util.go
```diff
@@ -735,36 +735,35 @@ func convertNodeID(newWf *wfv1.Workflow, regex *regexp.Regexp, oldNodeID string,
...
 	}
-	return updated, err
+
+	return updatedWf, err
 }
 
-func retryWorkflow(ctx context.Context, kubeClient kubernetes.Interface, hydrator hydrator.Interface, wfClient v1alpha1.WorkflowInterface, name string, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, error) {
-	wf, err := wfClient.Get(ctx, name, metav1.GetOptions{})
+// RetryWorkflow creates a workflow from the workflow archive
+func RetryArchivedWorkflow(ctx context.Context, podIf v1.PodInterface, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, error) {
```
alexec, line 748: This is a weird one. We want to retry an archived workflow.  But the workflow itself may or may not have been deleted. So I think logic can be:

1. Hydrate workflow
2. Formulate the workflow for retry.
3. Delete any pods for any failed node.
4. De-hydrate workflow.
5. Update the workflow. If 404 error, zero fields like UID and resourceVersion, then create it.

I want to see these changes:

* Rename `prepareWorkflowForRetry` `FormulateRetryWorkflow`.
* Rather than pass in `podIf`. Instead return a list of pod names that need to be deleted.
* Wrap that in `RetryWorkflow`. I think you can use that is both cases.

dpadhiar: Should we move deleting of the pods all to the server side then?

```
// Iterate the previous nodes. If it was successful Pod carry it forward
	deletedNodes := make(map[string]bool)
	for _, node := range wf.Status.Nodes {
		doForceResetNode := false
		if _, present := nodeIDsToReset[node.ID]; present {
			// if we are resetting this node then don't carry it across regardless of its phase
			doForceResetNode = true
		}
		switch node.Phase {
		case wfv1.NodeSucceeded, wfv1.NodeSkipped:
			if !strings.HasPrefix(node.Name, onExitNodeName) && !doForceResetNode {
				newWF.Status.Nodes[node.ID] = node
				continue
			}
		case wfv1.NodeError, wfv1.NodeFailed, wfv1.NodeOmitted:
			if !strings.HasPrefix(node.Name, onExitNodeName) && (node.Type == wfv1.NodeTypeDAG || node.Type == wfv1.NodeTypeTaskGroup || node.Type == wfv1.NodeTypeStepGroup) {
				newNode := node.DeepCopy()
				newNode.Phase = wfv1.NodeRunning
				newNode.Message = ""
				newNode.StartedAt = metav1.Time{Time: time.Now().UTC()}
				newNode.FinishedAt = metav1.Time{}
				newWF.Status.Nodes[newNode.ID] = *newNode
				continue
			} else {
				deletedNodes[node.ID] = true
			}
			// do not add this status to the node. pretend as if this node never existed.
		default:
			// Do not allow retry of workflows with pods in Running/Pending phase
			return nil, errors.InternalErrorf("Workflow cannot be retried with node %s in %s phase", node.Name, node.Phase)
		}
		if node.Type == wfv1.NodeTypePod {
			templateName := getTemplateFromNode(node)
			version := GetWorkflowPodNameVersion(wf)
			podName := PodName(wf.Name, node.Name, templateName, node.ID, version)
			log.Infof("Deleting pod: %s", podName)
			err := podIf.Delete(ctx, podName, metav1.DeleteOptions{})
			if err != nil && !apierr.IsNotFound(err) {
				return nil, errors.InternalWrapError(err)
			}
		} else if node.Name == wf.ObjectMeta.Name {
			newNode := node.DeepCopy()
			newNode.Phase = wfv1.NodeRunning
			newNode.Message = ""
			newNode.StartedAt = metav1.Time{Time: time.Now().UTC()}
			newNode.FinishedAt = metav1.Time{}
			newWF.Status.Nodes[newNode.ID] = *newNode
			continue
		}
	}

	if len(deletedNodes) > 0 {
		for _, node := range newWF.Status.Nodes {
			var newChildren []string
			for _, child := range node.Children {
				if !deletedNodes[child] {
					newChildren = append(newChildren, child)
				}
			}
			node.Children = newChildren

			var outboundNodes []string
			for _, outboundNode := range node.OutboundNodes {
				if !deletedNodes[outboundNode] {
					outboundNodes = append(outboundNodes, outboundNode)
				}
			}
			node.OutboundNodes = outboundNodes

			newWF.Status.Nodes[node.ID] = node
		}
	}
```

> Update the workflow. If 404 error, zero fields like UID and resourceVersion, then create it.

This can also occur on the server side so we can remove `RetryArchivedWorkflow` and use `RetryWorkflow` for both.

alexec: correct - to the service layer

dpadhiar: Moved deletion of pods to server side. `RetryWorkflow` now also returns a list of podNames to delete which allows us to remove podIf needing to be passed at all.

```
func RetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, []string, error) {
	updatedWf, podsToDelete, err := FormulateRetryWorkflow(ctx, wf, restartSuccessful, nodeFieldSelector)
	if err != nil {
		return nil, nil, err
	}

	return updatedWf, podsToDelete, err
}
```

```
func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string) (*wfv1.Workflow, []string, error) {
...
		if node.Type == wfv1.NodeTypePod {
			templateName := getTemplateFromNode(node)
			version := GetWorkflowPodNameVersion(wf)
			podName := PodName(wf.Name, node.Name, templateName, node.ID, version)
			podsToDelete = append(podsToDelete, podName)
```

```
func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req *workflowarchivepkg.RetryArchivedWorkflowRequest) (*wfv1.Workflow, error) {
	wfClient := auth.GetWfClient(ctx)
	kubeClient := auth.GetKubeClient(ctx)

	wf, err := w.GetArchivedWorkflow(ctx, &workflowarchivepkg.GetArchivedWorkflowRequest{Uid: req.Uid})
	if err != nil {
		return nil, err
	}

	wf, podsToDelete, err := util.RetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector)
	if err != nil {
		return nil, err
	}

		log.Infof("Deleting pod: %s", podName)
		err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
		if err != nil && !apierr.IsNotFound(err) {
			return nil, errors.InternalWrapError(err)
		}
	}

	wf, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Update(ctx, wf, metav1.UpdateOptions{})
	if apierr.IsBadRequest(err) {
		wf.ObjectMeta.ResourceVersion = ""
		wf, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Create(ctx, wf, metav1.CreateOptions{})
		if err != nil {
			return nil, err
		}
	}

	return wf, nil
}
```

test/e2e/argo_server_test.go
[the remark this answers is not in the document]
alexec: `s.Need(fixtures.BaseLayerArtifacts)` should not be needed?

workflow/util/util_test.go
```diff
@@ -892,56 +891,56 @@ func TestRetryWorkflow(t *testing.T) {
...
-					assert.Equal(t, "", node.Message)
-					assert.Equal(t, wfv1.NodeRunning, node.Phase)
-					assert.Equal(t, metav1.Time{}, node.FinishedAt)
-					assert.True(t, node.StartedAt.After(createdTime.Time))
-				}
-			}
-		}
-	})
-}
+// func TestRetryArchivedWorkflow(t *testing.T) {
```
alexec, line 894: clean-up?

dpadhiar: Missed on a quick scan, removed

server/workflow/workflow_server.go
```diff
@@ -319,10 +319,34 @@ func (s *workflowServer) RetryWorkflow(ctx context.Context, req *workflowpkg.Wor
...
+	wf, podsToDelete, err := util.RetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, podName := range podsToDelete {
+		log.Infof("Deleting pod: %s", podName)
+		err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
+		if err != nil && !apierr.IsNotFound(err) {
+			return nil, errors.InternalWrapError(err)
```
alexec, line 336: remove InternalWrapError, not needed

server/workflowarchive/archived_workflow_server.go
```diff
@@ -209,3 +212,37 @@ func (w *archivedWorkflowServer) ResubmitArchivedWorkflow(ctx context.Context, r
...
+		return nil, err
+	}
+
+	wf, podsToDelete, err := util.RetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, podName := range podsToDelete {
+		log.Infof("Deleting pod: %s", podName)
```
alexec, line 231: structured logging please

server/workflowarchive/archived_workflow_server.go
```diff
@@ -209,3 +212,37 @@ func (w *archivedWorkflowServer) ResubmitArchivedWorkflow(ctx context.Context, r
...
+		err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
+		if err != nil && !apierr.IsNotFound(err) {
+			return nil, errors.InternalWrapError(err)
```
alexec, line 234: remove InternalWrapError

server/workflowarchive/archived_workflow_server.go
```diff
@@ -209,3 +212,37 @@ func (w *archivedWorkflowServer) ResubmitArchivedWorkflow(ctx context.Context, r
...
+		}
+	}
+
+	wf, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Update(ctx, wf, metav1.UpdateOptions{})
+	if apierr.IsAlreadyExists(err) {
```
alexec, line 239: is this condition correct?

alexec: IsNotFound? how did you test this?

dpadhiar: used wrong error check, this is meant for going from Create to Update however we're doing Update to Create, will fix

alexec (approved):

PR #7998, opened 2022-02-24
fix: Re-factor `assessNodeStatus`. Fixes #7996

### LINKED ISSUE #7996: Re-factor `assessNodeStatus`
This func needs work:

* It can determine that a state change is needed, but that never gets persisted :- remove `update`.
* It can miss exit code and output changes :- remove guard condition on this.
* The daemon code is hard to understand, undocumented, spread out.

> Argo workflows can start containers that run in the background (also known as `daemon containers`) while the workflow itself continues execution. Note that the daemons will be *automatically destroyed* when the workflow exits the template scope in which the daemon was invoked. Daemon containers are useful for starting up services to be tested or to be used in testing (e.g., fixtures). We also find it very useful when running large simulations to spin up a database as a daemon for collecting and organizing the results. The big advantage of daemons compared with sidecars is that their existence can persist across multiple steps or even the entire workflow.

status.nodes.daemoned:

> Daemoned tracks whether or not this node was daemoned and need to be terminated

Completed daemon node are ALWAYS succeeded.

Related changes:

ca87f2906 fix: Daemon step in running state, but dependents don't start (#7107)
6875479db fix: Daemond status stuck with Running (#6742)
a147f178d fix(controller): Set finishedAt for workflow with Daemon steps (#6650)
72ee1cce9 fix: Set daemoned nodes to Succeeded when boudary ends (#5421)
d1abcb055 fix: Unmark daemoned nodes after stopping them (#5005)
2ff11cc98 fix(controller): Fix node status when daemon pod deleted but its children nodes are still running (#4683)
f07b5afea Issue #1113 - Wait for daemon pods completion to handle annotations (#1177)
174eb20a6 Issue #1040 - Kill daemoned step if workflow consist of single daemoned step (#1144)
311169f7e Fix issue where sidecars and daemons were not reliably killed (resolves #879)
e34728c66 Fix issue where daemoned steps were not terminated properly in DAG templates (resolves #832)
6a54b31f3 Support for automatic termination for daemoned workflow steps
227c19616 Initial support for daemon workflow steps (no termination yet)

alexec:

Fixes #7996

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -953,41 +953,28 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 		defer wfNodesLock.Unlock()
 		if node, ok := woc.wf.Status.Nodes[nodeID]; ok {
 			if newState := woc.assessNodeStatus(pod, &node); newState != nil {
-				woc.wf.Status.Nodes[nodeID] = *newState
-				woc.addOutputsToGlobalScope(node.Outputs)
-				if node.MemoizationStatus != nil {
-					if node.Succeeded() {
-						c := woc.controller.cacheFactory.GetCache(controllercache.ConfigMapCache, node.MemoizationStatus.CacheName)
-						err := c.Save(ctx, node.MemoizationStatus.Key, node.ID, node.Outputs)
+				woc.addOutputsToGlobalScope(newState.Outputs)
+				if newState.MemoizationStatus != nil {
+					if newState.Succeeded() {
+						c := woc.controller.cacheFactory.GetCache(controllercache.ConfigMapCache, newState.MemoizationStatus.CacheName)
+						err := c.Save(ctx, newState.MemoizationStatus.Key, newState.ID, newState.Outputs)
 						if err != nil {
-							woc.log.WithFields(log.Fields{"nodeID": node.ID}).WithError(err).Error("Failed to save node outputs to cache")
-							node.Phase = wfv1.NodeError
+							woc.log.WithFields(log.Fields{"nodeID": newState.ID}).WithError(err).Error("Failed to save node outputs to cache")
+							newState.Phase = wfv1.NodeError
+							newState.Message = err.Error()
 						}
 					}
 				}
-				if node.Phase == wfv1.NodeRunning {
+				if newState.Phase == wfv1.NodeRunning {
 					podRunningCondition.Status = metav1.ConditionTrue
 				}
+				woc.wf.Status.Nodes[nodeID] = *newState
 				woc.updated = true
-			}
-			node := woc.wf.Status.Nodes[nodeID]
-			if node.Type == wfv1.NodeTypePod {
-				if node.HostNodeName != pod.Spec.NodeName {
-					node.HostNodeName = pod.Spec.NodeName
-					woc.wf.Status.Nodes[nodeID] = node
-					woc.updated = true
-				}
-				podProgress := progress.PodProgress(pod, &node)
-				if podProgress.IsValid() && node.Progress != podProgress {
-					woc.log.WithField("progress", podProgress).Info("pod progress")
-					node.Progress = podProgress
-					woc.wf.Status.Nodes[nodeID] = node
-					woc.updated = true
-				}
-			}
-			if node.Fulfilled() && !node.IsDaemoned() {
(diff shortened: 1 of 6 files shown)


workflow/controller/operator.go
```diff
@@ -953,41 +953,28 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
...
 		if node, ok := woc.wf.Status.Nodes[nodeID]; ok {
 			if newState := woc.assessNodeStatus(pod, &node); newState != nil {
-				woc.wf.Status.Nodes[nodeID] = *newState
-				woc.addOutputsToGlobalScope(node.Outputs)
-				if node.MemoizationStatus != nil {
-					if node.Succeeded() {
-						c := woc.controller.cacheFactory.GetCache(controllercache.ConfigMapCache, node.MemoizationStatus.CacheName)
-						err := c.Save(ctx, node.MemoizationStatus.Key, node.ID, node.Outputs)
+				woc.addOutputsToGlobalScope(newState.Outputs)
+				if newState.MemoizationStatus != nil {
```
alexec, line 957: I think these changes may fix some memoization bugs.

workflow/controller/operator.go
```diff
@@ -953,41 +953,28 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
...
 					}
 				}
-				if node.Phase == wfv1.NodeRunning {
+				if newState.Phase == wfv1.NodeRunning {
 					podRunningCondition.Status = metav1.ConditionTrue
 				}
+				woc.wf.Status.Nodes[nodeID] = *newState
 				woc.updated = true
-			}
-			node := woc.wf.Status.Nodes[nodeID]
```
alexec, line 973: I've moved this to `assessNodeStatus` as I think it is good idea to be in one place.

workflow/controller/operator_test.go
```diff
@@ -1317,58 +1339,32 @@ func TestAssessNodeStatus(t *testing.T) {
...
+			assert.Equal(t, tt.daemon, wf.GetTemplateByName(tt.node.TemplateName).IsDaemon(), "check the test case is valid")
 			cancel, controller := newController()
 			defer cancel()
 			woc := newWorkflowOperationCtx(wf, controller)
 			got := woc.assessNodeStatus(tt.pod, tt.node)
 			assert.Equal(t, tt.want, got.Phase)
 		})
 	}
-
-	t.Run("Daemon Step finished - Pod running", func(t *testing.T) {
```
alexec, line 1344: moved these two tests up to the main block

workflow/progress/updater.go
```diff
@@ -1,29 +1,9 @@
...
 import (
-	apiv1 "k8s.io/api/core/v1"
-
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
-	"github.com/argoproj/argo-workflows/v3/workflow/common"
 )
 
-// PodProgress reads the progress annotation of a pod and ensures it's valid and synced
-// with the node status.
-func PodProgress(pod *apiv1.Pod, node *wfv1.NodeStatus) wfv1.Progress {
```
alexec, line 12: moved to `assessNodeStatus`

workflow/controller/operator.go
```diff
@@ -1204,64 +1192,65 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, node *wfv1.NodeStatu
...
-					woc.log.Infof("Setting node %v outputs: %s", node.ID, outputStr)
-					if err := json.Unmarshal([]byte(outputStr), node.Outputs); err != nil { // I don't expect an error to ever happen in production
-						node.Phase = wfv1.NodeError
-						node.Message = err.Error()
-					}
-				}
-			}
+	// both exit code and outputs maybe know before the pod terminates, we risk loosing them to pod being deleted
+	// if we do not capture them
+	if exitCode := getExitCode(pod); exitCode != nil {
```
alexec, line 1202: I've removed the guard condition on outputs, so we could capture them sooner, reducing impact of "pod deleted"

workflow/controller/operator.go
terrytangyuan: Good catch 👍

terrytangyuan (approved):
LGTM! Glad you fixed a couple of potential bugs along the way.

PR #8045, opened 2022-03-01
fix: argo plugins docs refer to plugin.* but should be server.*

NikeNano:
Signed-off-by: NikeNano <niklas.sven.hansson@gmail.com>

https://github.com/argoproj/argo-workflows/blob/b7a525be4014e3bdd28124c8736c25a007049ae7/cmd/argo/commands/executorplugin/io.go#L25

Fixes #TODO


alexec (approved):

alexec:
ops - should be `docs:` not `feat:` - but merged now

PR #8062, opened 2022-03-03
feat: Expand `mainContainer` config to support all fields. Fixes #7962

### LINKED ISSUE #7962: Ability to set default container security context
# Summary

Currently you can only set the default pod security context as part of `workflowDefaults`. It would be good to be able to set a default container security context as well. At the moment there doesn't seem to be a good way to do this, as the container securityContext can only be set in the container template definition in a workflow/workflowTemplate.

Moving this to defaults would potentially DRY up a lot of code, and make it easier to tighten up security across all workflows

Something like this:

```yaml
  workflowDefaults:
    spec:
      securityContext:
        runAsNonRoot: true
        runAsUser: 8737 #any non-root user
      containerSecurityContext:
        runAsNonRoot: true
        readOnlyRootFilesystem: true
        capabilities:
          drop:
            - all
```

# Use Cases

The main use for us is to avoid workflow/template authors needing to specify securityContexts to ensure their pods meet basic security requirements. It would also significantly reduce the amount of boilerplate yaml code in workflows.

---

alexec:

Fixes #7962

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -154,21 +154,22 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 			c.Name = common.MainContainerName
 		}
 		// Allow customization of main container resources.
-		if isResourcesSpecified(woc.controller.Config.MainContainer) {
-			c.Resources = *woc.controller.Config.MainContainer.Resources.DeepCopy()
-		}
-		// Template resources inv workflow spec takes precedence over the main container's configuration in controller
-		switch tmpl.GetType() {
-		case wfv1.TemplateTypeContainer:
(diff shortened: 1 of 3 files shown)


workflow/controller/workflowpod.go
```diff
@@ -154,21 +154,22 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
...
+				return nil, err
 			}
-		case wfv1.TemplateTypeScript:
-			if isResourcesSpecified(&tmpl.Script.Container) {
-				c.Resources = *tmpl.Script.Resources.DeepCopy()
+			b, err := json.Marshal(c)
+			if err != nil {
+				return nil, err
+			}
+			if err := json.Unmarshal(a, &c); err != nil {
```
sarabala1979, line 167: I like it. it is very simple to understand. Will it cover all usecase?

sarabala1979 (approved):

PR #8079, opened 2022-03-04
feat: add resubmit API for archived workflows. Fixes #7909

### LINKED ISSUE #7909: Add Resubmit API  on Archive workflow

dpadhiar:
Fixes #7909

Workflows currently can only be resubmitted if they exist on the cluster. If a workflow is GC-ed immediately after completion, you must save the workflow manually as a plain YAML or into workflow archive, then `kubectl create` before using `argo resubmit`. This API allows for archived workflows to be resubmitted removing the need for this otherwise hacky method.


alexec (commented):
PUT or POST?

dpadhiar:
> PUT or POST?

Resubmit for workflow server is a PUT request so I assumed resubmitting from the archive should be the same.

alexec:
> > PUT or POST?
> 
> Resubmit for workflow server is a PUT request so I assumed resubmitting from the archive should be the same. Since the workflow we are resubmitting has been deleted however, should we use a POST request instead?

Should be consistent. Whatever that is.

tczhao:
Should we upgrade the UI part as well?
when using resubmit button from archive-workflow, use the new endpoint instead of the old method

alexec (approved):

PR #8087, opened 2022-03-06
feat: added support for binary HTTP template bodies. Fixes #6888

### LINKED ISSUE #6888: support binary HTTP template bodies
Today, v3.2, HTTP template allows you to specify a body as a string. This is useful for `{{..}}` templating, but does not support all potential uses cases. It maybe necessary to support `[]byte` too. This could probably be done like this:

```yaml
bodyFrom:
   bytes: "b2341234==
````

This would allows us to extend this, e.g. to get the body from a config map or secret:

```yaml
bodyFrom:
  configMap:
    name: my-map
    key: my-key
````

### PR DESCRIPTION
**Pull request by rohankmr414**

Fixes #6888

### CODE REVIEW

File: workflow/executor/agent.go [hunk 1]
```diff
@@ -260,18 +260,35 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
...
+					"headers": tmpl.HTTP.Headers.ToHeader(),
+				},
+				"response": map[string]interface{}{
+					"statusCode": response.StatusCode,
+					"body":       string(bodyBytes),
+					"headers":    response.Header,
+				},
+			}
+		} else {
+			evalScope = map[string]interface{}{
```
sarabala1979, line 279: Don't you need to have nil check for  `BodyFrom`?

pkg/apis/workflow/v1alpha1/http_types.go
```diff
@@ -43,6 +47,8 @@ type HTTP struct {
 	SuccessCondition string `json:"successCondition,omitempty" protobuf:"bytes,6,opt,name=successCondition"`
 	// Body is content of the HTTP Request
 	Body string `json:"body,omitempty" protobuf:"bytes,5,opt,name=body"`
-	// insecureSkipVerify is a bool when if set to true will skip TLS verification for the HTTP client
-	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty" protobuf:"bytes,7,opt,name=insecureSkipVerify"`
+	// BodyFrom is  content of the HTTP Request as Bytes
+	BodyFrom *HTTPBodySource `json:"bodyFrom,omitempty" protobuf:"bytes,7,opt,name=bodyFrom"`
```
sarabala1979, line 51: Can you add it as last element 8?

sarabala1979 (changes requested):

pkg/apis/workflow/v1alpha1/http_types.go
rohankmr414: Sure

workflow/executor/agent.go
```diff
@@ -260,18 +260,35 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
...
+					"body":    string(tmpl.HTTP.BodyFrom.Bytes),
+					"headers": tmpl.HTTP.Headers.ToHeader(),
+				},
+				"response": map[string]interface{}{
+					"statusCode": response.StatusCode,
+					"body":       string(bodyBytes),
+					"headers":    response.Header,
+				},
+			}
+		} else {
```
simster7, line 278: Why are we duplicating this block twice if the only difference is the contents of `"body"`? Why not just have a statement setting `"body"` appropriately based on `tmpl.HTTP.BodyFrom`?

pkg/apis/workflow/v1alpha1/http_types.go
```diff
@@ -12,6 +12,10 @@ type HTTPHeaderSource struct {
 
 type HTTPHeaders []HTTPHeader
 
+type HTTPBodySource struct {
+	Bytes []byte `json:"bytes,omitempty" protobuf:"bytes,1,opt,name=bytes"`
```
simster7, line 16: Why are we using type `[]byte` here?

pkg/apis/workflow/v1alpha1/http_types.go
```diff
@@ -12,6 +12,10 @@ type HTTPHeaderSource struct {
 
 type HTTPHeaders []HTTPHeader
 
+type HTTPBodySource struct {
```
simster7, line 15: Can you add a doc string?

simster7 (changes requested):

workflow/executor/agent.go
```diff
@@ -260,18 +260,35 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
...
-				"body":    tmpl.HTTP.Body,
-				"headers": tmpl.HTTP.Headers.ToHeader(),
-			},
-			"response": map[string]interface{}{
-				"statusCode": response.StatusCode,
-				"body":       string(bodyBytes),
-				"headers":    response.Header,
-			},
+		var evalScope map[string]interface{}
+		if tmpl.HTTP.BodyFrom != nil {
```
sarabala1979, line 264: @simster7 has good point. Can you add `GetBody` function in `HTTP` which will return body for request?

http_type.go
```go
func (h *HTTP) GetBody() (string, error){
	if h.Body != "" {
		return h.Body, nil
	}
	if h.BodyFrom != nil && h.BodyFrom.Bytes != nil {
		return string (h.BodyFrom.Bytes), nil
	}
	return "", fmt.Errorf("HTTP body is not found")
}
```

pkg/apis/workflow/v1alpha1/http_types.go
alexec: I think this is probably correct - `string` would not work because it cannot encode certain characters, I think the JSON encoding is the body in base 64

pkg/apis/workflow/v1alpha1/http_types.go
rohankmr414: sure

blkperl (commented):
It doesn't look like we have an `http_types_test.go` file but it seems like we should have some test coverage for this PR.

pkg/apis/workflow/v1alpha1/http_types.go
```diff
@@ -43,6 +48,19 @@ type HTTP struct {
...
 	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty" protobuf:"bytes,7,opt,name=insecureSkipVerify"`
 }
+
+// GetBody returns the body of the HTTP request.
+func (h *HTTP) GetBody() string {
+	if h.Body != "" {
+		return h.Body
+	}
+	if h.BodyFrom != nil && h.BodyFrom.Bytes != nil {
+		return string(h.BodyFrom.Bytes)
```
simster7, line 63: @alexec mentioned:

> I think this is probably correct - string would not work because it cannot encode certain characters, I think the JSON encoding is the body in base 64

Does the `string(...)` cast here not have the same problem?

alexec: I think type-casting will introduce weird issues with character encoding.

Instead, lets have a separate `bodyBytes` and `GetBodyBytes`, that way the users cannot be impacted by it.

workflow/executor/agent.go
```diff
@@ -269,7 +269,7 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
 			"request": map[string]interface{}{
 				"method":  tmpl.HTTP.Method,
 				"url":     tmpl.HTTP.URL,
-				"body":    tmpl.HTTP.Body,
+				"body":    tmpl.HTTP.GetBody(),
```
alexec, line 272: I think we should have a separate `bodyBytes` field.

alexec (changes requested):
Changes requested.

workflow/executor/agent.go
rohankmr414: I've added a separate `bodyFrom` field so that we can add support for configmaps, artifacts, secrets in the future.

sarabala1979:
@rohankmr414 are you still like to address the change request?

rohankmr414:
> @rohankmr414 are you still like to address the change request?

@sarabala1979 Instead of `bodyBytes` I've added a separate `bodyFrom` field so that we can add support for configmaps, artifacts, secrets in the future. will that be ok?

alexec:
My request was that ‘bodyBytes’ be in the expression. Not the spec.

rohankmr414:
@alexec for some reason the unit tests are failing and I can't find the reason behind it. Can you please help me with it?

alexec:
I think the failing build might be fixed by re-basing to master.

workflow/executor/agent.go
```diff
@@ -267,10 +267,11 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
...
 		evalScope := map[string]interface{}{
 			"request": map[string]interface{}{
-				"method":  tmpl.HTTP.Method,
-				"url":     tmpl.HTTP.URL,
-				"body":    tmpl.HTTP.Body,
-				"headers": tmpl.HTTP.Headers.ToHeader(),
+				"method":    tmpl.HTTP.Method,
+				"url":       tmpl.HTTP.URL,
+				"body":      tmpl.HTTP.Body,
+				"bodyBytes": tmpl.HTTP.BodyFrom.Bytes,
```
alexec, line 273: would this `nil` panic?

alexec (changes requested):
Looks good, yet needs:

* You've a nil panic, I think this is why the test is failing.
* Test needs fixing.
* Needs sync with master.

workflow/executor/agent.go
```diff
@@ -265,14 +265,25 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
...
+			request = map[string]interface{}{
 				"method":    tmpl.HTTP.Method,
 				"url":       tmpl.HTTP.URL,
 				"body":      tmpl.HTTP.Body,
 				"bodyBytes": tmpl.HTTP.BodyFrom.Bytes,
 				"headers":   tmpl.HTTP.Headers.ToHeader(),
-			},
+			}
+		} else {
+			request = map[string]interface{}{
```
alexec, line 278: This is code duplication. I think you add `tmpl.HTTP.GetBodyBytes()` which can return `nil` when `tmpl.BodyFrom` is nil

alexec (approved):

Issue #7860, opened 2022-02-11
Resources plugin


alexec:
# Summary

Today - the resources template has two downsides:

* It is expensive as it requires one pod per resource.
* It only supports a single resource.
* Resource manifests must come from the workflow.
* Resources are strings, not structured YAML

It would also be an opportunity to complete this:

* #7725
* #7179 
* #7143
* #6862

I propose a new plugin:

```yaml
plugins:
  resources:
    -  manifest:
          value: 
          valueFrom: 
            url: ...
       wait:
          for: condition=Ready
          status:
            phase: Running
```

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

berlinsaint:
very useful feature.

alexec:
Correct. Plug-ins are coming in v3.3 and it would be easy for anyone to write a turbo-charged version of the resources template themselves.

berlinsaint:
> Correct. Plug-ins are coming in v3.3 and it would be easy for anyone to write a turbo-charged version of the resources template themselves.

Good. I have read the doc , and i'll  do a job for adapting the Volcano Job, when the v3.3 will be released

berlinsaint:
> Correct. Plug-ins are coming in v3.3 and it would be easy for anyone to write a turbo-charged version of the resources template themselves.

Good. I have read the doc , and i'll  do a job for adapting the Volcano Job

berlinsaint:
@alexec  hi , i just implement a demo plugin , and i found that  , every workflow will start an agent with a plugin sidecar container,  i have some question here, do we really need sidecar? why not we make the plugin run forever and all the same extended type workflow share the plugin?  i think   per workflow with an sidecar is expensive too.

NikeNano:
>  i think per workflow with an sidecar is expensive too.

I think it would begert if it is possible to have plugins that lived outside the scope of a workflow in ordet to allow plugins to persist over multiple workflows.

alexec:
#8142

berlinsaint:
> @berlinsaint that is currently the architecture. There is an open enhancement to run one agent per namespace. However, there are benefits of running one per workflow, e.g. it can use a different service account, it is not running when not needed.
> 
> If your plugin is heavy weight, then you could split it into two parts, one part is a normal Kubernetes, and another part is normal plugin which proxies requests.
> 
[...]

PR #8104, opened 2022-03-08
fix: executor plugin sidecar can be injected with sa secrets.

berlinsaint:
Signed-off-by: maybaby <berlinsaint@126.com>

Fixes #TODO

### DIFF
--- a/workflow/controller/agent.go
+++ b/workflow/controller/agent.go
@@ -104,6 +104,10 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 	}
 
 	serviceAccountName := woc.execWf.Spec.ServiceAccountName
+	enableAutoMountSA := woc.execWf.Spec.AutomountServiceAccountToken
+	if enableAutoMountSA == nil {
+		enableAutoMountSA = pointer.BoolPtr(false)
+	}
 	secretName, err := woc.getServiceAccountTokenName(ctx, serviceAccountName)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get token volumes: %w", err)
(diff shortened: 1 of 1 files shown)


berlinsaint:
By my testing, i found that it create sidecar with `automountServiceAccountToken` set to false, this means no easy way to communitate with k8s api if want do sth in plugin.

workflow/controller/agent.go
```diff
@@ -135,7 +135,7 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 				RunAsUser:    pointer.Int64Ptr(8737),
 			},
 			ServiceAccountName:           serviceAccountName,
-			AutomountServiceAccountToken: pointer.BoolPtr(false),
+			AutomountServiceAccountToken: pointer.BoolPtr(*woc.execWf.Spec.AutomountServiceAccountToken),
```
simster7, line 138: Why de-reference and re-reference? Simpler:

```suggestion
			AutomountServiceAccountToken: woc.execWf.Spec.AutomountServiceAccountToken,
```

simster7 (commented):
Minor change otherwise LGTM

workflow/controller/agent.go
alexec: I think this could change the default behaviour to mount the token. We don’t want that.

berlinsaint: do you mean that the sidecar(plugin)is designed to that cant mount sa by default?  why ？

alexec: Correct. Security.

berlinsaint: one more questuon,how can i init k8s api in sidecar?  Or as I mentioned in [here](https://github.com/argoproj/argo-workflows/issues/7860)
is there a design to make the plugin run forever such as using deployment?

alexec:
Ok, so this shows you how to do secrets:

https://github.com/argoproj-labs/argo-workflows-slack-executor-plugin

Do we need to add docs?

berlinsaint:
> Ok, so this shows you how to do secrets:
> 
> https://github.com/argoproj-labs/argo-workflows-slack-executor-plugin
> 
> Do we need to add docs?

I think what i need is  plugin authentication as [8144](https://github.com/argoproj/argo-workflows/issues/8144) says?

PR #8120, opened 2022-03-10
feat: Multi-cluster. Fixes #3523 

### LINKED ISSUE #2063: Managing multiple namespaces with the new version
# Summary
How do you suggest handling multiple namespaces that run Argo workflows in the new version?
The ways I’m thinking about right now are either:
a) workflow controller for each namespace and an Argo server for each namespace.
b) workflow controller for each namespace and a “managing” namespace which has a Argo server for each namespace with the “managed-namespace” patameter.

Is there a better way? Do you think it will be a good idea for you to implement an option to have one Argo server with multiple managed-namespaces for it?

# Motivation
We are the MLOps team and we want every DS team to have their own namespace with their own resource limitations and only show in the Argo UI their own workflows.

### LINKED ISSUE #3523: Multi-cluster, multi-namespace workflows
# Summary

Run workflows across multiple clusters.

# Motivation

So you only need to run one Argo Workflows installation.
So you can run a workflow that has nodes in different clusters.

# Proposal

Like Argo CD.

#3516

---

### LINKED ISSUE #4684: Multi-cluster UI: aka "control plane"
# Summary

Users would like the UI to be a single place to manage all workflows.

# Use Cases

Reduces the operational burden of installing the UI in many places

---

### LINKED ISSUE #6490: Casbin RBAC for Argo Server
# Summary

The current model of RBAC leans heavily on Kubernetes RBAC, so is easy to make secure, but may not scale well. 

Consider the situation where you have 1000s teams and 1000s namespace. At the very least you're like to need a OIDC group for each namespace, and then a service account, role and role binding.

It maybe better to use Casbin (or similar) to provide a more flexible way to configure this.

# Use Cases

When would you use this?

---

alexec:
Fixes #3523 
Fixes #6490 
Fixes #2063 
Fixes #4684


sarabala1979:
I think it should be in a feature branch. are you planning to merge master  branch?

.github/workflows/ci-build.yaml
```diff
@@ -79,33 +79,33 @@ jobs:
     name: E2E Tests
     runs-on: ubuntu-latest
     timeout-minutes: 20
-    needs: [ tests, argoexec-image ]
```
alexec, line 82: revert

cmd/workflow-controller/main.go
```diff
@@ -156,6 +169,58 @@ func NewRootCommand() *cobra.Command {
...
+	map[string]*restclient.Config,
+	map[string]kubernetes.Interface,
+	map[string]wfclientset.Interface,
+	map[string]metadata.Interface,
+	error,
+) {
+	configs := map[string]*restclient.Config{common.LocalCluster: config}
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, nil, nil, nil, err
```
alexec, line 182: wrap errors in this func

alexec:
> I think it should be in a feature branch. are you planning to merge master branch?

What is difference between "feature" branch and other branch?

cmd/argo/commands/cluster/get_profile.go
```diff
@@ -0,0 +1,135 @@
...
+				},
+				Data: map[string][]byte{"kubeconfig": data},
+			}
+
+			data, err = yaml.Marshal(profile)
+			if err != nil {
+				return err
+			}
+
+			println("ALEX", remoteServer)
```
alexec, line 119: delete

alexec:
TestDefaultParamValueWhenNotFound is failing.

PR #8122, opened 2022-03-10
fix(plugins): UX improvements

Shark:
Changes:

* `Content-Type` header is added to `template.execute` RPC calls → apart from technical correctness this makes it easier to use because many frameworks rely on this header for JSON parsing

   ✅ Has an integration test in [test/e2e/executor_plugins_test.go](https://github.com/argoproj/argo-workflows/blob/9aa04a1493c01782ed51b01c733ca6993608ea5b/test/e2e/executor_plugins_test.go)

* `argo executor-plugin build .` copies the `metadata/namespace` from the plugin manifest to the ConfigMap

   ✅ Tested on my local machine


alexec:
test?

sarabala1979 (approved):

PR #8128, opened 2022-03-10
fix: read/write mount mode for init / wait containers. Fixes #7755

### LINKED ISSUE #7755: Incompatibility with PSPs using read-only allowed host paths
## Summary

Argo Workflows seems to be incompatible with Pod Security Policies (PSPs) specifying read-only allowed host paths. [As mentioned in this thread](https://github.com/argoproj/argo-workflows/pull/1214#discussion_r789865942), the `init` / `wait` containers surrounding a workflow container mount the same volumes as the main container, but without respecting their read/write mode. For environments using PSPs with read-only allowed host paths, it becomes impossible to run workflows that use volume mounts (as sidecar containers will violate the PSP).

[Looking at the code mentioned above](https://github.com/argoproj/argo-workflows/blob/v3.3.0-rc2/workflow/controller/workflowpod.go#L1027-L1028), the author claims that mounts need to be read/write to allow overlapping mount paths. However, the `main` container will already need to mount paths in read/write mode if they overlap, so I don't see why the sidecar containers needs to modify this setting. Am I missing something or could we simply remove this override? 

**Argo Workflows install**: version v3.2.6, on a vanilla k8s cluster with the K8SAPI executor.

## Diagnostics

To replicate the issue, here's a sample PSP with a read-only allowed host path:
```yaml
$ kubectl apply -f - <<EOF
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: readonly-hostpaths
spec:
  allowedHostPaths:
  - pathPrefix: /etc/ssl/certs/
    readOnly: true
  volumes:
  - '*'

  runAsUser:
    rule: RunAsAny
  runAsGroup:
    rule: RunAsAny
  fsGroup:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
EOF
```

And a workflow that will violate it:
```shell
$ argo submit - <<EOF
kind: Workflow
metadata:
  generateName: hello-world-
spec:
  entrypoint: entrypoint
  templates:
  - name: entrypoint
    script:
      image: alpine
      source: |
        ls -lha /etc/ssl/certs/
      volumeMounts:
      - mountPath: /etc/ssl/certs/
        name: ssl-certs
        readOnly: true
    volumes:
    - hostPath:
        path: /etc/ssl/certs/
      name: ssl-certs
EOF
```

After inspecting the workflow in the UI, you get the following message, which indicates that the sidecar container is trying to mount the `ssl-certs` volume in write mode.
```
pods "hello-world-bq8pl" is forbidden: unable to validate against any pod security policy: [spec.containers[0].volumeMounts[0].readOnly: Invalid value: false: must be read-only]
```

This can be confirmed by disabling the PSP and inspecting the pod spec that is created.

---

louisblin:
As discussed in #7755, the `init` / `wait` containers surrounding a
workflow container mount the same volumes as the `main` container, but
without respecting their read/write mode. For environments using PSPs
with read-only allowed host paths, it becomes impossible to run workflows
that use volume mounts (as sidecar containers will violate the PSP).

The original code author (@jessesuen) claims that mounts need to be
read/write to allow overlapping mount paths. However, the `main`
container will already need to mount paths in read/write mode if they
overlap, so there does not seem to be a good reason for keeping this.

Fixes #7755


alexec:
I think it needs a test with overlapping volume mount.

louisblin:
Still planning to work on this in the next couple weeks fyi

tooptoop4:
where @louisblin

Issue #8124, opened 2022-03-10
Ensure all links in docs point to the latest stable release rather than master


ChaosInTheCRD:
# Summary
As a consequence of one of my PRs, a bug was introduced into the master branch of the Argo Workflows codebase. This had instant repercussions (see Github Issue [here](https://github.com/argoproj/argo-workflows/issues/7632)) for new users that were consuming the tool. Whenever the docs provide a link to manifests, they point to the master branch (see [here](https://github.com/argoproj/argo-workflows/blob/master/docs/quick-start.md?plain=1#L16)), rather than the latest stable release branch.

The impacts of this are that new users get stalled while trying out the tool due to an unknown bug or change that is not reflected in the documentation, but also could lead to instances where (while it is certainly not best practice) users are running Argo Workflows manifests from `master` branch, which would mean they are using the `:latest` tag on their images (see [here](https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/quick-start-postgres.yaml#:~:text=image%3A%20quay.io/argoproj/argocli%3Alatest) for an example.

What change needs making?
Change the docs to point exclusively to manifests at `https://raw.githubusercontent.com/argoproj/argo-workflows/stable/manifests/` rather than `https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/`

alexec:
`latest` is a anti-pattern:

* https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375

It does not mean anything. If we do support it as stable, it means operating known bad practice. 

Using master/latest is fine for development environments and I've just updated the installation guide because production set-up has several important steps.

ChaosInTheCRD:
> Using master/latest is fine for development environments and I've just updated the installation guide because production set-up has several important steps.

Personally I wouldn't agree. I think it has a negative impact on those that are trying to use the tool for the first time if they are hit by pre-release bugs and other things? Equally I think the deviation in the fact that Argo CD uses Stable release / semver tags for quick start envs whilst workflows uses master/`:latest`... I don't really see the upside? Maybe I am missing something. I would be interested in hearing the opinion of the ArgoCD community 😄

alexec:
To your first point - it is a good hypothesis - but we’ve never had a user report a pre-release bug from using quick-start. Only edge case bugs tend to get through testing. 

The only problem we seem to have is if users think they have pinned a version, but are in fact using latest. 

Here are some thoughts:

[...]

PR #8165, opened 2022-03-16
feat: Use pinned executor version.

alexec:

This should mitigate the problem for some users.

See #8124

### DIFF
--- a/workflow/controller/config.go
+++ b/workflow/controller/config.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/time/rate"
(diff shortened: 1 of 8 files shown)


workflow/controller/config.go
```diff
@@ -92,7 +98,10 @@ func (wfc *WorkflowController) executorImage() string {
 	if wfc.cliExecutorImage != "" {
 		return wfc.cliExecutorImage
 	}
-	return wfc.Config.GetExecutor().Image
+	if v := wfc.Config.GetExecutor().Image; v != "" {
+		return v
+	}
+	return fmt.Sprintf("quay.io/argoproj/argoexec:" + argo.GetVersion().Version)
```
alexec, line 104: use the same version as the controller

terrytangyuan (approved):
LGTM.

Issue #7586, opened 2022-01-19
Workflows with large/long scripts cause argument list too long error on init container


aachenmax:
## Summary
What happened/what you expected to happen?

When executing a workflow which contains a step configured as a script (e.g. Python, bash) with source code exceeding about 100000 characters, the pod cannot be started, with the following error message shown for the Argo init container:
 
`standard_init_linux.go:228: exec user process caused: argument list too long` 

This is due to the fact that the ARG_MAX / MAX_ARG_STRLEN limit for a single command, hardcoded at 131072 bytes in most kernels, is being exceeded. From my understanding/analysis, the actual script size is not a problem (until you approach the 1MB etcd limit I guess), since this is mounted into the container by Argo. The problem is that an environment variable named ARGO_TEMPLATE, containing the entire pod spec, is being set for the init, wait, and actual workload container. I believe in Argo versions < 3.2, this has been handled differently using pod annotations and volumes and therefore was not a problem and has been changed with this commit: https://github.com/argoproj/argo-workflows/commit/cecc379ce23e708479e4253bbbf14f7907272c9c

Of course one could ask: "Why do you need such a long script?" and argue that this should be better split up and run in consecutive or parallel containers. However, this might not be easily possible for certain applications. It does put an absolute limit on the spec/configuration size of a single pod that can be launched by Argo.

What version of Argo Workflows are you running?
This affects versions >= 3.2 and was tested on Argo v3.2.3
Not an issue in versions 3.1 for example

## Diagnostics

Example workflow YAML is attached as .txt
[large_script.txt](https://github.com/argoproj/argo-workflows/files/7898646/large_script.txt)

What Kubernetes provider are you using?

Tested it on 
- Kubernetes v1.21.5 running on Docker Desktop 4.2.0 (70708), Engine: 20.10.10
- Kubernetes v1.20.10 running on RKE, Engine 20.10.8
with same error

What executor are you running? Docker/K8SAPI/Kubelet/PNS/Emissary

Tested it with executors
- Docker
- Emissary
on above Docker Desktop Kubernetes with same error

# Logs from the workflow controller:
```
time="2022-01-19T15:49:50.320Z" level=info msg="Processing workflow" namespace=default workflow=scripts-python-tls7z
time="2022-01-19T15:49:50.357Z" level=info msg="Updated phase  -> Running" namespace=default workflow=scripts-python-tls7z
[...]

Issue #913, opened 2018-07-21
Fail workflows when hitting k8s resource size limit (status code 413)


jessesuen:
**Is this a BUG REPORT or FEATURE REQUEST?**:

It's possible for a workflow to get into a situation where it's unable to update a workflow because the object itself is simply too large. The symptom is a message in the controller:

```
Error updating workflow: the server responded with the status code 413 but did not return more information (put workflows.argoproj.io data-prep-ld52n)
```

K8s limits resource sizes to 1MB. In this event, we want to fail the workflow with a meaningful message. What is currently happening is that it stays in a running state until the workflow is deleted. It will never complete and simply causes the controller to continually process this for all of eternity.

**What happened**:
```
time="2018-07-20T20:00:01Z" level=info msg="All of node data-prep-ld52n.grader-1E547080-DB87-40D2-994E-EA21E401FC57 dependencies [] completed" namespace=default workflow=data-prep-ld52n
time="2018-07-20T20:00:01Z" level=warning msg="Deadline exceeded" namespace=default workflow=data-prep-ld52n
time="2018-07-20T20:00:01Z" level=warning msg="Error updating workflow: the server responded with the status code 413 but did not return more information (put workflows.argoproj.io data-prep-ld52n)" namespace=default workflow=data-prep-ld52n
time="2018-07-20T20:00:01Z" level=info msg="Processing workflow" namespace=default workflow=data-prep-ld52n
```

**What you expected to happen**:

If we cannot store the workflow because it's payload is too large, then at least fail the workflow by updating the phase to become Error with a message. NOTE: in order to do this, we would need to drop the current payload and simply update the status.phase.

**How to reproduce it (as minimally and precisely as possible)**:

Run a large workflow (workflow who's size will reach 1+ MB). 

**Anything else we need to know?**:

**Environment**:
- Argo version: v2.1.1

PR #8169, opened 2022-03-17
fix: Limit workflows to 128KB and return a friendly error message

blkperl:
Signed-off-by: William Van Hevelingen <william.vanhevelingen@acquia.com>

Part of: #7586


test/e2e/expectedfailures/large-workflow.yaml
```diff
@@ -3,15 +3,1572 @@ kind: Workflow
 metadata:
   generateName: large-workflow-
 spec:
-  entrypoint: large-workflow
```
blkperl, line 6: @jessesuen added this file in #913 but it doesn't look like the e2e tests use it.

workflow/controller/workflowpod.go
```diff
@@ -317,9 +317,16 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 		pod.Spec.InitContainers[i] = c
 	}
 
+	envVarTemplateValue := wfv1.MustMarshallJSON(tmpl)
+	templateSize := len(envVarTemplateValue)
+	if templateSize > 128000 {
```
blkperl, line 322: ARG_MAX is 131KB so we could allow a slightly higher number but there are other environment variables being added. 

```
docker run --entrypoint ""  -it quay.io/argoproj/argoexec:latest sh                       
/ # getconf ARG_MAX
131072
```

blkperl: We could loop over all the `envVars` and add them in a separate error message.

alexec (approved):

workflow/controller/workflowpod.go
blkperl: @alexec is there any special way to note this as a breaking change and where would be the best place to document the new limit?

alexec: I thought we were going to fix this by splitting the single ARG (max 130k) into two args (make 260k) which would fix the issue.

alexec: Docs are still good mind you, so I approve of this.

PR #8176, opened 2022-03-17
feat: Add support to auto-mount service account tokens for plugins.

alexec:

Users have requested a couple of times the ability to access the Kuberenetes API. This uses the same pattern as PodSpec.AutomountServiceAccountToken, but the name of the service account is fixed.


crenshaw-dev (commented):
Looks good. I'd recommend adding some docs: 1) why/how to use this and 2) "please please please only do this if necessary and then limit permissions carefully".

berlinsaint:
will this be merged at v3.3.2?

terrytangyuan (approved):

PR #8186, opened 2022-03-18
fix: task worker requeue wrong task. Fixes #8139

### LINKED ISSUE #8139: AgentExecutor: Requeue wrong tasks.
## Summary
When I test agent executor, I find if task worker execute wrong if task happens requeue.
So I dig into the code.
https://github.com/argoproj/argo-workflows/blob/master/workflow/executor/agent.go#L118
Here is the story.
There is two tasks in the taskQueue. Let's annotate _task1_ and _task2_.
the taskWorker first treat _task1_ and _task1_ replies to requeue in 10 sec, then worker treat _task2_.
After 10 secs, worker should treat _task1_ again. However this time worker treat _task2_ instead.

What version are you running?
latest

## Diagnostics
Here is my simplified code to illustrate the bug.
```go
package main

import (
	"fmt"
	"time"
)

type task struct {
	NodeId string
}

func main() {
	taskQueue := make(chan task)
	go worker(taskQueue)
	for i := 1; i < 10; i++ {
		taskQueue <- task{
			NodeId: fmt.Sprintf("node-%d", i),
		}
	}

	time.Sleep(3 * time.Second)
}

func worker(tq chan task) {
	for t := range tq {
		fmt.Println("worker", t.NodeId)
		time.AfterFunc(2*time.Second, func() {
			tq <- t
		})
	}
}
```
This demo code is expected that 10 tasks will be requeue and reexecuted in 2 seconds.
But If you execute the code, below is the result.
```bash
worker node-1
worker node-2
worker node-3
worker node-4
worker node-5
worker node-6
worker node-7
worker node-8
worker node-9
worker node-9
worker node-9
worker node-9
worker node-9
worker node-9
worker node-9
worker node-9
worker node-9
worker node-9
```
all tasks are polluted to the last task.
you can test with go playground. https://go.dev/play/p/XZyezyelrnj
This looks like that golang does not behave right in range channels with time after.
here is the fix code
```go
func worker(tq chan task) {
	for {
		t, ok := <-tq
		if !ok {
			break
		}
		fmt.Println("worker", t.NodeId)
		time.AfterFunc(2*time.Second, func() {
			tq <- t
		})
	}
}
```
Fix demo:https://go.dev/play/p/AiOO9dt11k4

---

wujayway:
Signed-off-by: wujayway <wujayway@gmail.com>

Fixes #8139
Use for loops instead of for range. Avoid go problems on `time.After` with closure.

### DIFF
--- a/workflow/executor/agent.go
+++ b/workflow/executor/agent.go
@@ -116,7 +116,11 @@ func (ae *AgentExecutor) Agent(ctx context.Context) error {
 }
 
 func (ae *AgentExecutor) taskWorker(ctx context.Context, taskQueue chan task, responseQueue chan response) {
-	for task := range taskQueue {
+	for {
+		task, ok := <- taskQueue
+		if !ok {
+			break
+		}
 		nodeID, tmpl := task.NodeId, task.Template
 		log := log.WithField("nodeID", nodeID)
 
--- a/workflow/executor/agent.go
+++ b/workflow/executor/agent.go
@@ -117,7 +117,7 @@ func (ae *AgentExecutor) Agent(ctx context.Context) error {
 
 func (ae *AgentExecutor) taskWorker(ctx context.Context, taskQueue chan task, responseQueue chan response) {
 	for {
-		task, ok := <- taskQueue
+		task, ok := <-taskQueue
 		if !ok {
 			break
 		}


alexec (approved):

alexec:
I've just noted, I'm not sure this will make a difference, as the semantics are not changed.

PR #8191, opened 2022-03-18
fix: Use `latest` image tag when version is `untagged`. Fixes #8188

### LINKED ISSUE #8188: Broken manifests
## Checklist

 

* [X] Double-checked my configuration.
* [X] Tested using the latest version.
* [X] Used the Emissary executor.

I want to run hello-world and expect it to successfully complete

Instead when describing the pod I get "failed to resolve reference "quay.io/argoproj/argoexec:untagged"

### What version are you running?

I use the latest cluster-install manifests rolled out by argo-cd

This bug does not occur if I pin version to before the following commit [2b5e4a1d2df7877d9b7b7fbedd7136a125a39c8d](https://github.com/argoproj/argo-workflows/commit/2b5e4a1d2df7877d9b7b7fbedd7136a125a39c8d)

It would seem that the above commit broke the manifests.

## Diagnostics

```yaml
kind: Workflow
metadata:
  generateName: hello-world-
  labels:
  annotations:
      This is a simple hello world example.
      You can also run it in Python: https://couler-proj.github.io/couler/examples/#hello-world
spec:
  entrypoint: whalesay
  templates:
  - name: whalesay
    container:
      image: docker/whalesay:latest
      command: [cowsay]
      args: ["hello world"]
```

---

alexec:

Fixes #TODO

### DIFF
--- a/workflow/controller/config.go
+++ b/workflow/controller/config.go
@@ -102,7 +102,14 @@ func (wfc *WorkflowController) executorImage() string {
 	if v := wfc.Config.GetExecutor().Image; v != "" {
 		return v
 	}
-	return fmt.Sprintf("quay.io/argoproj/argoexec:" + argo.GetVersion().Version)
+	return fmt.Sprintf("quay.io/argoproj/argoexec:" + getImageTag())
+}
+
+func getImageTag() string {
+	if v := argo.GetVersion().Version; v != "untagged" {
+		return v
+	}
+	return "latest"
 }
 
 // executorImagePullPolicy returns the imagePullPolicy to use for the workflow executor


version.go
```diff
@@ -17,6 +17,15 @@ var (
 	gitTreeState = ""                     // determined from `git status --porcelain`. either 'clean' or 'dirty'
 )
 
+// ImageTag return the image tag.
+// GetVersion().Version adulterates the version making it useless as the image tag.
+func ImageTag() string {
+	if version != "untagged" {
```
terrytangyuan, line 23: This should be using `GetVersion()` otherwise it's always `"v0.0.0"`?

alexec: Nope. The value of `version` and other vars is set by build flags in the `Makefile`:

```
GIT_TAG               := $(shell git describe --exact-match --tags --abbrev=0  2> /dev/null || echo untagged)
...
VERSION               := latest
...
ifeq ($(RELEASE_TAG),true)
VERSION               := $(GIT_TAG)
endif
...
override LDFLAGS += \
  -X github.com/argoproj/argo-workflows/v3.version=$(VERSION) \
  -X github.com/argoproj/argo-workflows/v3.buildDate=${BUILD_DATE} \
  -X github.com/argoproj/argo-workflows/v3.gitCommit=${GIT_COMMIT} \
  -X github.com/argoproj/argo-workflows/v3.gitTreeState=${GIT_TREE_STATE}
...
go build -v -ldflags '${LDFLAGS} -extldflags -static' -o $@.exe ./cmd/argo
```

alexec: Learn more: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

terrytangyuan (approved):

Issue #8185, opened 2022-03-18
 Template not executed when requeue time arrived.


berlinsaint:
I have implement a plugin to create volcano Job,  while the job has been created success,but

executor' failed:
```
time="2022-03-18T05:25:09.195Z" level=info msg="Starting Workflow Executor" version=v3.3.0
time="2022-03-18T05:25:09.196Z" level=info msg="Starting Agent" requeueTime=10s taskWorkers=16 workflow=hello-7kvk6
time="2022-03-18T05:25:09.296Z" level=info msg="Watch workflowtasksets 200"
time="2022-03-18T05:25:09.300Z" level=info msg="TaskSet Event" event_type=ADDED workflow=hello-7kvk6
time="2022-03-18T05:25:09.300Z" level=info msg="Processing task" nodeID=hello-7kvk6
time="2022-03-18T05:25:09.528Z" level=info msg="Sending result" message="no plugin executed the template" nodeID=hello-7kvk6 phase=Failed requeue=0s
time="2022-03-18T05:25:19.196Z" level=info msg="Processing Patch" workflow=hello-7kvk6
time="2022-03-18T05:25:19.201Z" level=info msg="Patch workflowtasksets/status 200"
time="2022-03-18T05:25:19.202Z" level=info msg="Patched TaskSet"
time="2022-03-18T05:25:19.202Z" level=info msg="TaskSet Event" event_type=MODIFIED workflow=hello-7kvk6
time="2022-03-18T05:25:19.202Z" level=info msg="Task is already considered" nodeID=hello-7kvk6
time="2022-03-18T05:25:29.264Z" level=info msg="TaskSet Event" event_type=MODIFIED workflow=hello-7kvk6
```

And argo get return: 
![image](https://user-images.githubusercontent.com/10629930/158943677-cbcc7b3f-54e3-4049-b85e-b273fbe08e82.png)

My plugin demo code is 

```
func (ct *Controller) ExecuteVolcanoJob(ctx *gin.Context) {
	c := &executorplugins.ExecuteTemplateArgs{}
	err := ctx.BindJSON(&c)
	if err != nil {
		klog.Error(err)
		return
	}
	inputBody := &VolcanoPluginBody{
		JobBody: &JobBody{
			JobSpec: &v1alpha1.JobSpec{},
		},
	}
	fmt.Println(string(c.Template.Plugin.Value))
	dd, _ := c.Template.Plugin.MarshalJSON()
	fmt.Println(string(dd))
[...]

PR #8194, opened 2022-03-19
fix:  requeue not delete the considererd Task flag

berlinsaint:
Signed-off-by: maybaby <berlinsaint@126.com>

Fixes [8185](https://github.com/argoproj/argo-workflows/issues/8185)


workflow/executor/agent.go
```diff
@@ -155,6 +155,9 @@ func (ae *AgentExecutor) taskWorker(ctx context.Context, taskQueue chan task, re
 		}
 		if requeue > 0 {
 			time.AfterFunc(requeue, func() {
+				if ok {
```
alexec, line 158: this guard is not needed

alexec (approved):

PR #8214, opened 2022-03-22
docs: Add deprecation notice on K8s API executor

terrytangyuan:


docs/workflow-executors.md
```diff
@@ -80,23 +80,6 @@ The emissary will exit with code 64 if it fails. This may indicate a bug in the
 * Configuration:
     * Additional Kubelet configuration maybe needed
 
-## Kubernetes API (k8sapi)
```
alexec, line 83: I don't think we should remove these, otherwise how will users running <=v3.3 know about them? Instead, we should mark them as depreated and unsupported.

alexec (approved):

PR #8226, opened 2022-03-23
feat: add archive retry command to argo CLI. Fixes #7907

### LINKED ISSUE #7907: Add Archive Retry Command on CLI

dpadhiar:
Fixes #7907 

It is currently only possible for an archived workflow to be retried through the CLI by using kubectl create then argo retry. This feature removes the need to do this set of operations by resubmitting workflows directly from the archive with the argo CLI: argo archive retry {uid}. This cli command supports the same wait, watch or log flags as argo resubmit.


agilgur5:
Replaced by #8229

Issue #7580, opened 2022-01-19
Better error handling for workflows which already exist


danbopes:
# Summary
Running the following command in my cluster (Using the default example template lovely-octopus), I get a `code = Unknown desc = The POST operation against Workflow.argoproj.io could not be completed at this time, please try again.`

This is because the workflow already exists, and I'm forcing the name using `--name specificname`. Instead of an unknown error, it would be prudent to return a better error code indicating this resource already exists:

```
argo submit --from WorkflowTemplate/lovely-octopus --name lovely-octopus-jobexists
time="2022-01-18T22:36:02.947Z" level=fatal msg="Failed to submit workflow: rpc error: code = Unknown desc = The POST operation against Workflow.argoproj.io could not be completed at this time, please try again."
```

# Use Cases
Through the API, I could make a request, and check if it already exists, and silently fail. When the error response is unknown, it could have been caused by some other error, and I'd only be guessing that it was because the workflow already exists.

When would you use this?
See above Use Case

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

PR #8229, opened 2022-03-23
feat: add archive retry command to argo CLI. Fixes #7907

### LINKED ISSUE #7907: Add Archive Retry Command on CLI

dpadhiar:
Fixes #7907 

It is currently only possible for an archived workflow to be retried through the CLI by using `kubectl create` then `argo retry`. This feature removes the need to do this set of operations by retrying workflows directly from the archive with the argo CLI: `argo archive retry {uid}`. This cli command supports the wait, watch or log flags as `argo retry`.

### DIFF
--- a/cmd/argo/commands/archive/resubmit.go
+++ b/cmd/argo/commands/archive/resubmit.go
@@ -45,27 +45,27 @@ func NewResubmitCommand() *cobra.Command {
 
 # Resubmit multiple workflows:
 
-  argo resubmit uid another-uid
+  argo archive resubmit uid another-uid
 
 # Resubmit multiple workflows by label selector:
 
-  argo resubmit -l workflows.argoproj.io/test=true
+  argo archive resubmit -l workflows.argoproj.io/test=true
 
 # Resubmit multiple workflows by field selector:
 
-  argo resubmit --field-selector metadata.namespace=argo
+  argo archive resubmit --field-selector metadata.namespace=argo
 
 # Resubmit and wait for completion:
 
-  argo resubmit --wait uid
+  argo archive resubmit --wait uid
 
 # Resubmit and watch until completion:
 
-  argo resubmit --watch uid
+  argo archive resubmit --watch uid
 
 # Resubmit and tail logs until completion:
 
-  argo resubmit --log uid
+  argo archive resubmit --log uid
 `,
 		Run: func(cmd *cobra.Command, args []string) {
 			if cmd.Flag("priority").Changed {
--- a/cmd/argo/commands/archive/retry.go
+++ b/cmd/argo/commands/archive/retry.go
@@ -1,0 +1,150 @@
+package archive
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/argoproj/pkg/errors"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
+
+	client "github.com/argoproj/argo-workflows/v3/cmd/argo/commands/client"
+	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands/common"
+	workflowpkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflow"
+	workflowarchivepkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflowarchive"
+	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+)
+
+type retryOps struct {
+	nodeFieldSelector string // --node-field-selector
+	restartSuccessful bool   // --restart-successful
+	namespace         string // --namespace
+	labelSelector     string // --selector
+	fieldSelector     string // --field-selector
+}
+
+// hasSelector returns true if the CLI arguments selects multiple workflows
+func (o *retryOps) hasSelector() bool {
+	if o.labelSelector != "" || o.fieldSelector != "" {
+		return true
+	}
+	return false
+}
+
+func NewRetryCommand() *cobra.Command {
+	var (
+		cliSubmitOpts common.CliSubmitOpts
+		retryOpts     retryOps
+	)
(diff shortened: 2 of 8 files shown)


dpadhiar:
```
result, err := wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Update(ctx, wf, metav1.UpdateOptions{})
	if err != nil {
		wf.ObjectMeta.ResourceVersion = ""
		wf.ObjectMeta.UID = ""
		result, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Create(ctx, wf, metav1.CreateOptions{})
		if err != nil {
			return nil, err
		}
	}
```
When updating a workflow from the archive, Update will return this error: 
`"Operation cannot be fulfilled on workflows.argoproj.io \"retry-fail-cqsb7\": the object has been modified; please apply your changes to the latest version and try again"` 
since the resource version differs between the workflow stored in the archive and the workflow on the cluster. Creating the workflow instead also currently fails with this error: 
`code = Unknown desc = The POST operation against [Workflow.argoproj.io](http://workflow.argoproj.io/) could not be completed at this time, please try again.` 
This is because we are attempting to create a workflow with the same name while it still exists on the cluster (this error is also brought up in https://github.com/argoproj/argo-workflows/issues/7580). This isn't caught with `apierr.IsAlreadyExists()`.

Since this error occurs while the workflow still exists on the cluster, we could instead instruct the user to retry the workflow using `argo retry`.

alexec (approved):

dpadhiar:
> LGTM. Need to fix failing test.

If we delete the workflow before retrying, the test will pass. Updating the workflow on the cluster fails since the resource versions differ.

alexec:
Builds go to pass, so the test must pass.

sarabala1979:
> ```
> result, err := wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Update(ctx, wf, metav1.UpdateOptions{})
> if err != nil {
> 	wf.ObjectMeta.ResourceVersion = ""
> 	wf.ObjectMeta.UID = ""
> 	result, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Create(ctx, wf, metav1.CreateOptions{})
> 	if err != nil {
> 		return nil, err
> 	}
> }
> ```
> 
> When updating a workflow from the archive, Update will return this error: `"Operation cannot be fulfilled on workflows.argoproj.io \"retry-fail-cqsb7\": the object has been modified; please apply your changes to the latest version and try again"` since the resource version differs between the workflow stored in the archive and the workflow on the cluster. Creating the workflow instead also currently fails with this error: `code = Unknown desc = The POST operation against [Workflow.argoproj.io](http://workflow.argoproj.io/) could not be completed at this time, please try again.` This is because we are attempting to create a workflow with the same name while it still exists on the cluster (this error is also brought up in #7580). This isn't caught with `apierr.IsAlreadyExists()`.
> 
> Since this error occurs while the workflow still exists on the cluster, we could instead instruct the user to retry the workflow using `argo retry`.
We check the workflow exist in cluster. If it is there then just return message say wf exist on cluster use `argo retry`

dpadhiar:
``` go
_, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Get(ctx, wf.Name, metav1.GetOptions{})
	if apierr.IsNotFound(err) {

		wf, podsToDelete, err := util.FormulateRetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector)
		if err != nil {
			return nil, err
		}

			log.WithFields(log.Fields{"podDeleted": podName}).Info("Deleting pod")
			err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
			if err != nil && !apierr.IsNotFound(err) {
				return nil, err
			}
		}

		wf.ObjectMeta.ResourceVersion = ""
		wf.ObjectMeta.UID = ""
		result, err := wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Create(ctx, wf, metav1.CreateOptions{})
		if err != nil {
			return nil, err
		}

		return result, nil
	}

	if err == nil {
		return nil, status.Error(codes.AlreadyExists, "Workflow already exists on cluster, use argo retry {name} instead")
	}

	return nil, err
```
To resolve issues with Update and Create, we can instead try to Get the workflow from the cluster first to check if it still exists.  If the workflow is not found, then we can retry the workflow from the archive with Create. If it is found and doesn't return an error, we instruct the user to use `argo retry {name}` instead. Otherwise, we return error if any comes with the Get.

sarabala1979:
> ```go
> _, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Get(ctx, wf.Name, metav1.GetOptions{})
> 	if apierr.IsNotFound(err) {
> 
> 		wf, podsToDelete, err := util.FormulateRetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector)
> 		if err != nil {
> 			return nil, err
> 		}
> 
> 		for _, podName := range podsToDelete {
> 			log.WithFields(log.Fields{"podDeleted": podName}).Info("Deleting pod")
> 			err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
> 			if err != nil && !apierr.IsNotFound(err) {
> 				return nil, err
> 			}
> 		}
> 
> 		wf.ObjectMeta.ResourceVersion = ""
> 		wf.ObjectMeta.UID = ""
> 		result, err := wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Create(ctx, wf, metav1.CreateOptions{})
> 		if err != nil {
> 			return nil, err
> 		}
> 
> 		return result, nil
> 	}
> 
> 	if err == nil {
> 		return nil, status.Error(codes.AlreadyExists, "Workflow already exists on cluster, use argo retry {name} instead")
> 	}
> 
> 	return nil, err
> ```
> 
> To resolve issues with Update and Create, we can instead try to Get the workflow from the cluster first to check if it still exists. If the workflow is not found, then we can retry the workflow from the archive with Create. If it is found and doesn't return an error, we instruct the user to use `argo retry {name}` instead. Otherwise, we return error if any comes with the Get.

LGTM

PR #8235, opened 2022-03-24
feat!: Refactor/simplify configuration code

alexec:

* Remove the ability to hot-reload config. Users should never rely on this. Some config changes required restart.
* Remove the risk of having incorrect config loaded briefly before start-up.


config/controller.go
```diff
@@ -71,63 +51,14 @@ func (cc *controller) parseConfigMap(cm *apiv1.ConfigMap) (interface{}, error) {
...
-					}
-				}
-			},
-		})
-	controller.Run(stopCh)
-	log.Info("Watching config map updates")
-}
-
-func (cc *controller) Get(ctx context.Context) (interface{}, error) {
+func (cc *controller) Get(ctx context.Context, config interface{}) error {
```
terrytangyuan, line 57: Should we re-use the configmap informer?

terrytangyuan (approved):

server/apiserver/argoserver.go
```diff
@@ -376,15 +376,6 @@ func mustRegisterGWHandler(register registerFunc, ctx context.Context, mux *runt
 	}
 }
 
-// Unlike the controller, the server creates object based on the config map at init time, and will not pick-up on
-// changes unless we restart.
-// Instead of opting to re-write the server, instead we'll just listen for any old change and restart.
-func (as *argoServer) restartOnConfigChange(interface{}) error {
```
sarabala1979, line 382: Will this change the existing behavior of argo server?  Argo server will not restart if there is any config change. is it right?

alexec: Correct. This PR should be marked as a breaking change.

sarabala1979 (approved):

alexec:
TestFunctionalSuite/TestEventOnNodeFailSentAsPod fails

PR #8265, opened 2022-03-28
fix: patch workflow status to workflow. Fixes #8245

### LINKED ISSUE #8245: `pod deleted` + `re-apply` error with `plugin` task
## Summary
This is a bug related to #4798.
we can get a pod deleted error under high load. It not only happens when workflow finished, also in the middle of the dag.
#4798 is fixed well, even pod deleted error happened in this situation.
However, if a workflow consists of some plugin tasks it stuck.
Here is the example:
DAG: A --> B --> C
`A` and `C` are plugin tasks. And `B` is a normal task.
When B pod got deleted error, workflowtaskset will be labeled `workflows.argoproj.io/completed=true`.
The fix #4798 will prevent from writing wrong status to workflow crd.
Workflow will be reconciled to correct status next time and continue to process task `C`. But workflowtaskset now is already labeled `workflows.argoproj.io/completed=true`, and agent will not process this workflowtaskset.
In conclusion, this workflow will be stuck in process C, controller is waiting result from agent, and agent does not treat workflowtaskset.

What version are you running?
v3.3.0

## Solution

I think we should always patch workflow status to workflowset. So next time when controller reconcile the workflowset will correct the status.

---

wujayway:
Signed-off-by: wujayway <wujayway@gmail.com>

Fixes #8245 
Every time controller reconciles the workflowset patch workflow status on workflowset.

### DIFF
--- a/workflow/controller/taskset.go
+++ b/workflow/controller/taskset.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


alexec (approved):

alexec:
This makes sense. Always overwrite the value.

PR #8271, opened 2022-03-28
fix: Default value is ignored when loading params from configmap. Fixes #8262

### LINKED ISSUE #8262: ConfigmapRef does not respect optional and default keyword
## Checklist

 

## Summary

`Optional` and `default` keywords are not respected when using  `configMapKeyRef`.
Workflow errors while trying to reference a key that does not exist.

What version are you running?
v3.2.9
Emissary Executor

## Diagnostics

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: retry-sample
  labels:
    workflows.argoproj.io/configmap-type: Parameter
data:
  dummy: "{}"
  dummy1: ""
```

```yaml
kind: WorkflowTemplate
metadata:
  name: retry-sample
spec:
  entrypoint: main
  templates:
    - name: main
      inputs:
        parameters:
          - name: message
            valueFrom:
              default: "hello"
              configMapKeyRef:
                name: retry-sample
                key: dummy-missing
                optional: true
      container:
        image: docker/whalesay
        command: [cowsay]
```

```bash

```

---

terrytangyuan:
Fixes https://github.com/argoproj/argo-workflows/issues/8262.


workflow/common/util.go
```diff
@@ -151,9 +151,14 @@ func ProcessArgs(tmpl *wfv1.Template, args wfv1.ArgumentsProvider, globalParams,
 
 				cmValue, err := GetConfigMapValue(configMapInformer, namespace, cmName, cmKey)
 				if err != nil {
-					return nil, errors.Errorf(errors.CodeBadRequest, "unable to retrieve inputs.parameters.%s from ConfigMap: %s", inParam.Name, err)
+					if inParam.ValueFrom.Default != nil {
```
alexec, line 154: should we make sure `err` is IsNotFound?
what about test?

terrytangyuan: > Default specifies a value to be used if retrieving the value from the specified source fails
Based on the original purpose from the doc-string, I think we want to have default no matter what specific error we are getting. 

I'll add tests.

alexec: I'm not sure about this, should a transient error result in my workflow running with unexpected parameters?

terrytangyuan: That makes sense. Added a check for IsNotFound as well as a test.

alexec (approved):

PR #8266, opened 2022-03-28
Remove hardcoded namespace in kustomize file #8250

tiwarisanjay:
Fixes #TODO
Please refer to https://github.com/argoproj/argo-workflows/issues/8250 for more information.

Issue #8250, opened 2022-03-25
Hardcoded namespaces in crb for kustomize cluster-install


tiwarisanjay:
## Checklist

 

* [x] Double-checked my configuration.
* [x] Tested using the latest version.
* [x] Used the Emissary executor.

## Summary
Following file has hardcoded namespaces in clusterrolebinding which is holding us to change namespace when we call the module with new namespace : 
[manifestesworkflow-controller-clusterrolebinding.yaml ](https://github.com/argoproj/argo-workflows/blob/v3.3.1/manifests/cluster-install/workflow-controller-rbac/workflow-controller-clusterrolebinding.yaml)

[argo-server-clusterolebinding.yaml ](https://github.com/argoproj/argo-workflows/blob/v3.3.1/manifests/cluster-install/argo-server-rbac/argo-server-clusterolebinding.yaml)

What happened/what you expected to happen?
Issue : 
Following Code should change all namsepaces to argoworkflow namespace : 
```yaml  
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- https://github.com/argoproj/argo-workflows.git/manifests/cluster-install/?ref=v3.2.4

namespace: argoworkflow
```
But I still see clusterrolebinding has namespaces as argo : 
```yaml 
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: argo-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: argo-cluster-role
subjects:
- kind: ServiceAccount
  name: argo
  namespace: argo
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: argo-server-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: argo-server-cluster-role
subjects:
- kind: ServiceAccount
  name: argo-server
  namespace: argo
``` 
What version are you running?
3.3.1
## Diagnostics

Paste the smallest workflow that reproduces the bug. We must be able to run the workflow.

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- https://github.com/argoproj/argo-workflows.git/manifests/cluster-install/?ref=v3.2.4

namespace: argoworkflow
```

---

**Message from the maintainers**:

[...]

PR #8280, opened 2022-03-31
fix: install.yaml missing crb subject ns

gdsoumya:
This PR fixes the issue with missing ns in crb subjects inside the `install.yaml` file, introduced in PR #8266 

ref: https://github.com/argoproj/argo-workflows/issues/8250#issuecomment-1084046919


jessesuen (approved):

tiwarisanjay:
@gdsoumya  Why exactly we are hardcoding the namespace for cluster install ? whoever refers the cluster install can mention the namespace in their own kustomization.yaml 
@alexec @jessesuen

tiwarisanjay:
@gdsoumya Yes, Apologies for install.yaml bug I introduced, as it was showing diff when I was fixing kustomize. But isn't kustomize supposed to be for generic use instead of hard-coding a namespace.
Also with kustomize you can not override namespace if its hardcoded. There is a open ticket for the same. :

https://github.com/kubernetes-sigs/kustomize/issues/880

PR #8284, opened 2022-03-31
fix: `taskresults` owned by pod rather than workflow.

alexec:

I think making task results owned by the workflow is incorrect. If they are owned by the pod, they are cleaned-up with the pod. This will save memory and reduce etcd usage while a workflow is running. When a workflow completes, they were already being deleted.


terrytangyuan (approved):
+1. GC on pods makes more sense.

PR #8296, opened 2022-04-01
feat: Remove the PNS executor. Fixes #7804

### LINKED ISSUE #7804: Remove `pns` executor
👀 

Should be able to use Emissary.

Also, remove the `libcap` and `procps` libraries one complete.

alexec:

Fixes #7804


alexec:
This PR will always wait for those checks, because the name changed when we removed the executor.

cmd/argoexec/commands/root.go
```diff
@@ -76,11 +75,10 @@ func NewRootCommand() *cobra.Command {
 
 func initExecutor() *executor.WorkflowExecutor {
 	version := argo.GetVersion()
-	executorType := os.Getenv(common.EnvVarContainerRuntimeExecutor)
-	log.WithFields(log.Fields{"version": version.Version, "executorType": executorType}).Info("Starting Workflow Executor")
+	log.WithFields(log.Fields{"version": version.Version}).Info("Starting Workflow Executor")
 	config, err := clientConfig.ClientConfig()
 	checkErr(err)
-	config = restclient.AddUserAgent(config, fmt.Sprintf("argo-workflows/%s argo-executor/%s", version.Version, executorType))
+	config = restclient.AddUserAgent(config, fmt.Sprintf("argo-workflows/%s", version.Version))
```
terrytangyuan, line 81: I think we can just remove the `executorType` but keep `argo-executor` to be consistent with UAs from controller, server, etc.

terrytangyuan (commented):
Just one minor comment. Otherwise LGTM!

cmd/argoexec/commands/root.go
alexec: hard code to emissary?

terrytangyuan: Either `fmt.Sprintf("argo-workflows/%s emissary", version.Version)` or `fmt.Sprintf("argo-workflows/%s argo-executor", version.Version)` would work. The benefit of including a more general term like "executor" would give the operator a more meaningful signal when monitoring the traffic.

alexec: `argo-executor=not-applicable`?

alexec: So just `argo-executor` then?

terrytangyuan: Sounds good

terrytangyuan (approved):

PR #8303, opened 2022-04-01
fix: reduce number of workflows displayed in UI by default. Fixes #8297

### LINKED ISSUE #8297: Reduce the number of workflows display in the UI by default
The UI is so slow to load, that users go to CLI instead. We should display 50 rather 500 workflows by default.

dpadhiar:
Signed-off-by: Dillen Padhiar <dpadhiar99@gmail.com>

Fixes #8297 

Previously, the default number of workflows displayed by the UI was 500. This can cause the UI to load for a long time if there are too many workflows being shown. Instead, we can list only by default to decrease this load.


alexec:
LGTM

alexec (approved):

PR #8325, opened 2022-04-06
feat: Show artifacts in UI DAG. Fixes #8324

### LINKED ISSUE #8324: Show artifacts in the UI DAG
Artifacts are key to each workflow, yet we treat them as an implementation detail.

We should surface them in the UI as part of the workflow itself.

More generally, we should focus more on data, less on workflows.

alexec:

Fixes #8324

This PR:

* Introduces 3 new endpoints to help work with artifacts.
  * `GetArtifact` which superceeds the existing API. Supports name/uid and input/output.
  * `GetArtifactDescription` which return information such as filename, content-type, and list of items any tgz contains. `.tgz` is VERY common (the default in fact) so this is more than just a utility.
  * `GetArtifactItem` which returns a single file from within an artifact.
* Supports new UI elements which use the new APIs to display artifacts.


terrytangyuan (commented):
Comments I mentioned during the weekly meeting:
1. The DAG looks a bit clunky and it's hard to distinguish nodes vs. artifacts
2. How to support rendering additional formats other then plain text files?
3. Should we not enable this by default?

alexec:
> * The DAG looks a bit clunky and it's hard to distinguish nodes vs. artifacts

I think we can work on that. I've some idea.

> * How to support rendering additional formats other then plain text files?

I'm polling users to find out what they want support for. 

> * Should we not enable this by default?

* If we enable by default, should be able to disable it (persistently, e.g.  using cookie).
* If we disable by default, we need to figure out how to make the feature discoverable.

PR #8345, opened 2022-04-08
feat!: Add entrypoint lookup. Fixes #8344

### LINKED ISSUE #8344: Allow images map for emissary to be overridden at a namespace level
# Summary

When using `emissary`, we have to either specify the `command` on every `container` or we have to specify a map in the `workflow-controller-configmap`. The problem is that the default deployment has the `workflow-controller` as a cluster level resource and the individual workflows are namespace specific. In a shared cluster, that makes it impractical to use the image map. Having to specify command is bad enough (as the containers are all designed to have that embedded), but having to specify it in _many_ places is far worse as the maintenance effort goes way up.

# Use Cases

This would be used for ALL images run in argo workflows in individual namespaces.  When we create new images to be run or have to change an existing image, this would provide a single place that needs updating rather than every deployed WorkflowTemplate/CronWorkflow that happens to use that image.

I would propose something like `workflow-emissary-images-configmap` that if defined in a namespace, would be used to augment or override the configuration in the `workflow-controller-configmap`. Another nice-to-have would be a _default_ as a large number of images provide a conventional `/docker-entrypoint.sh`.

_Side note, is there any documentation as to the reason why it is necessary or a good idea to force `command` to be specified to start with?  It is certainly something that can easily lead to hard to reason bugs as you might change the image but not see your changes reflected because the workflow still contains an old `command`...._

---

alexec:

Fixes #8344

This is similar to how Tekton does this. Only support v2 manifests.

This could be a breaking change for set of users. I thought I would be better to correct the behaviour for this that try and work-around it.


workflow/controller/entrypoint/container_registery_index.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, err
+	}
+	cmd := &Image{}
+	if len(f.Config.Entrypoint) > 0 {
+		cmd.Command = f.Config.Entrypoint
+	} else {
+		cmd.Command = f.Config.Cmd
+	}
+	return cmd, nil
```
lucastheisen, line 44: I left this comment [_elsewhere_](https://github.com/alexec/argo-workflows/commit/4bff0bc88d15b05d0c31147f3eb901c7dc8724e8) but seems that is not this MR...

This could be surprising.  The `template.container` may already have `args` specified which would replace the `f.Config.Cmd`, but this approach would effectively append `template.container.args` to the end of `f.Config.Cmd`.  Should probably return a

```
type EntrypointAndCmd struct {
    Entrypoint []string
    Cmd []string
}
```
So that downstream (where you prepend `argoexec emissary`) you could decide if the `CMD` should be used, or is replaced by `template.container.args`.

Right?

marwatk: Also, this ignores `CMD` if there's an `ENTRYPOINT`, correct? (I could be misreading). 

For example, if a `Dockerfile` had:
```Dockerfile
ENTRYPOINT [initialization_script.sh]
CMD [do_the_thing.sh]
```

This code would effectively drop the `do_the_thing.sh`, when instead it needs to concatenate `CMD` onto `ENTRYPOINT`, assuming the `template` doesn't override either as @lucastheisen suggests.

alexec: I struggle to know what is correct here. The fact that cmd is an array seems very odd to me. I think we may have been lumbered with some very old Docker design decision. 

This logic, I copied from Tekton.

https://github.com/tektoncd/pipeline/blob/62c9b0d861bd2ae01263dfa67ba41fb0fa2bc9a9/pkg/pod/entrypoint_lookup_impl.go#L163

lucastheisen: I would agree with the statement:

> behaviour must be exactly the same as a Dockerfile

but that does require deviation from what Tekton is doing.  They do indeed appear to be using `ENTRYPOINT || CMD` style logic which is definitely not the _same as a Dockerfile_.  It should be relatively simple to propagate the tuple of entrypoint and cmd to the consumer (though it may have impact on your cache code that i didn't think through).  But at least then you can faithfully reproduce the ENTRYPOINT / CMD behavior.

alexec: Case 1 - nothing user specified:

```
(image_entrypoint, image_cmd, image_args, nil, nil) -> (concat(image_entrypoint, image_cmd), image_args)
```

Case 2 - fully user specified:

```
(image_entrypoint, image_cmd, image_args, user_cmd, user_args) -> (user_cmd, user_args), image_args)
```

Case 3 - no user cmd

```
(image_entrypoint, image_cmd, image_args, nil, user_args) -> (image_entrypoint, image_args)
```

alexec: https://stackoverflow.com/questions/44316361/difference-between-docker-entrypoint-and-kubernetes-container-spec-command
https://sahansera.dev/closer-look-at-kubernetes-pod-commands-args/

alexec:
```
If you do not supply command or args for a Container, the defaults defined in the Docker image are used.
If you supply only args for a Container, the default Entrypoint defined in the Docker image is run with the args that you supplied.
If you supply a command for a Container, only the supplied command is used. The default EntryPoint and the default Cmd defined in the Docker image are ignored. Your command is run with the args supplied (or no args if none supplied).
```

alexec:
I've updated this based on the docs:

```golang
	// Entrypoint array. Not executed within a shell.
	// The docker image's ENTRYPOINT is used if this is not provided.
	// Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
	// cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
	// to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
	// produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
	// of whether the variable exists or not. Cannot be updated.
	// More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
	// +optional
	Command []string `json:"command,omitempty" protobuf:"bytes,3,rep,name=command"`
	// Arguments to the entrypoint.
	// The docker image's CMD is used if this is not provided.
	// Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
	// cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
	// to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
	// produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
	// of whether the variable exists or not. Cannot be updated.
	// More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
	// +optional
	Args []string `json:"args,omitempty" protobuf:"bytes,4,rep,name=args"`
```

lucastheisen:
Looks good to me!  Thanks, this will be great once it gets released.

workflow/controller/entrypoint/container_registery_index.go
```diff
@@ -0,0 +1,53 @@
...
+	f, err := img.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+	cmd := &Image{}
+	if len(f.Config.Entrypoint) > 0 {
+		cmd.Cmd = f.Config.Entrypoint
+	} else {
+		cmd.Cmd = f.Config.Cmd
+	}
```
lucastheisen, line 43: Just noticed this, this still seems wrong.  Shouldn't it be:

```
	cmd := &Image{}
	if len(f.Config.Entrypoint) > 0 {
		cmd.Entrypoint = f.Config.Entrypoint
	}
	if len(f.Config.Cmd) > 0 {
		cmd.Cmd = f.Config.Cmd
	}
```

though i guess both if's are unnecessary

alexec:
@terrytangyuan @sarabala1979 this has been reviewed by some interested people and given 👍 . Could you please review?

docs/upgrading.md
```diff
@@ -4,6 +4,49 @@
...
+## Upgrading to v3.4
+
+### feat!: Add entrypoint lookup. Fixes #8344
+
+Affected if:
+
+* Using the Emissary executor.
+* Used the `args` field for any entry in `images`.
+
+This PR automatically looks up the command and entrypoint. This implementation for config look-up was incorrect (it
```
alexec, line 16: This --> The

terrytangyuan:
> @terrytangyuan @sarabala1979 this has been reviewed by some interested people and given 👍 . Could you please review?

On my TODO list. Since this will be a breaking change, I'll spend dedicated time reviewing this soon.

docs/upgrading.md
```diff
@@ -4,6 +4,49 @@
...
+allowed you to specify `args` but not `entrypoint`). `args` has been removed to correct the behaviour.
+
+If you are incorrectly configured, the workflow controller will error on start-up.
```
terrytangyuan, line 19: We should probably add this to FAQs.

config/image.go
```diff
@@ -1,6 +1,6 @@
 package config
 
 type Image struct {
-	Command []string `json:"command"`
-	Args    []string `json:"args,omitempty"`
+	Entrypoint []string `json:"entrypoint,omitempty"`
+	Cmd        []string `json:"cmd,omitempty"`
```
terrytangyuan, line 5: Would it be better to stick with Cmd?

terrytangyuan (commented):
Just some nitpicks. Otherwise LGTM

docs/upgrading.md
alexec: Maybe include an example that will come up in Google results.

config/image.go
alexec: The behaviour has changed, so I wanted to change it.

terrytangyuan (approved):

vitalyrychkov:
Hello, 
We have recently upgraded to Argo Workflows 3.3.8 and changed the executor to Emissary. We are using Argo since about two years and built a framework around it. Until now we have used pns executor. After switching to Emissary, all our workflows fail with:   
 "task <task name> errored: container "main" in template <template>, does not have the command specified: when using the emissary executor you must either explicitly specify the command, or list the image's command in the index: https://argoproj.github.io/argo-workflows/workflow-executors/#emissary-emissary"


We always provide ENRTYPOINT and CMD in the Dockerfile and override the CMD with args in the contaner spec in the workflow template. 

I have seen in this thread that you expect it to work this way:

> - If you do not supply command or args for a Container, the defaults defined in the Docker image are used.
> - If you supply only args for a Container, the default Entrypoint defined in the Docker image is run with the args that you supplied.
> - If you supply a command for a Container, only the supplied command is used. The default EntryPoint and the default Cmd defined in the Docker image are ignored. Your command is run with the args supplied (or no args if none supplied).

However the second point (ENTRYPOINT in the Dockerfile and no command in the Container) does not work. Probably you did not have a test case for this one, I am attaching my simple test case for you then.

Could you please confirm this is the bug which must be fixed?

vitalyrychkov:
TEST CASE
Create a Dockerfile:
```
FROM debian
ENTRYPOINT ["echo"]
CMD ["DOCKERFILE"]
```

Build the image, e.g. cmdtest:latest and push to your repo

Create a test workflow to use the built image:

```
kind: Workflow
metadata:
  generateName: cmdtest-
  labels:
  annotations:
      This is a test for image command and entrypoint 
spec:
  entrypoint: cmdtest
  templates:
  - name: cmdtest
    container:
      image: <repo>/cmdtest:latest
      #command: ["echo"]
      #args: ["ARGS"]
```
In our Argo installation with Emissary executor the workflow fails with the mentioned error. If I uncomment the command, then the workflow succeeds.

vitalyrychkov:
I assume this issue has been closed long ago, therefore I have created a new one: https://github.com/argoproj/argo-workflows/issues/9139

PR #8375, opened 2022-04-12
fix: Updating completed node status

### LINKED ISSUE #8279: v3.3: stop results in `Failed` workflow, rather than `Succeeded`
## Checklist

 

* [ ] Double-checked my configuration.
* [ ] Used the Emissary executor.

## Summary

What version are you running? v3.3, worked in v3.2

> stopping a pipeline cause it to have a status of Failed instead of Succeeded as in the previous version.

## Diagnostics

```yaml

```

```bash

```

---

sarabala1979:

Fixes #8279


workflow/controller/exec_control.go
```diff
@@ -21,7 +21,11 @@ func (woc *wfOperationCtx) applyExecutionControl(ctx context.Context, pod *apiv1
 	}
 
 	nodeID := woc.nodeID(pod)
-
+	node := woc.wf.Status.Nodes[nodeID]
```
sarabala1979, line 24: No able to reproduce on my local env. Looks like some edge case pod state and node status is out of sync and it try to update the completed node

sarabala1979: Added extra check to make sure node status is not fulfilled for update the status

PR #8405, opened 2022-04-15
feat: added support for http as an option for artifact upload. Fixes #785

### LINKED ISSUE #785: Allow 'http' as an option for artifact upload
From Slack discussion with @jessesuen: Currently HTTP artifacts are 'input only'. It would be cool if we were able to upload them as well.

There are 2 main ways that I see this could be implemented:
* There is typical ‘form upload’ type HTTP stuff, that any upload field on a website would use
* The other way I would do it is more a ‘http api’ style, where it is a POST (potentially structured in some way) of the file contents, maybe as json/similar.

The second option is probably getting more nuanced and less generic. That sort of method might make more sense in somehow supporting ‘artifact plugins’, where these specific things could be ‘plugged in’ to the current architecture. That way I could say “upload this file as DevaliasJsonPluginStyle” and it would ‘just work’.

My main use case/thought for this is to support large file outputs (eg. too big to use `globalOutput`) back to some wrapping 'golang worker code', without the 'overhead' of having to upload then re-download from minio/etc.

Ideally in this case, i'd like to create a tiny API endpoint that accepts some kind of POST, and then uses the uploaded file directly 'in code', without it ever having to touch disk.

Currently we use a similar method with the http 'download' to populate a 'file' into the argo workflow from an in-memory store.

rohankmr414:

Fixes #785


rohankmr414:
@alexec how do we test this

alexec:
It would be acceptable to manually test this.

rohankmr414:
> It would be acceptable to manually test this, rather than set-up any system to test it.

@alexec I've used the same implementation which is present in `workflow/artifacts/artifactory/artifactory.go`

alexec:
There is a lot of code duplication between the artifatory and HTTP artifact repositories. Would this be a good time to merge them together and re-use the code?

rohankmr414:
> There is a lot of code duplication between the artifatory and HTTP artifact repositories. Would this be a good time to merge them together and re-use the code?

@alexec They are a little different as we are not including basic auth credentials in HTTP artifact repository requests.

alexec:
> @alexec They are a little different as we are not including basic auth credentials in HTTP artifact repository requests.

So we would need to add basic auth to HTTP?

rohankmr414:
> > @alexec They are a little different as we are not including basic auth credentials in HTTP artifact repository requests.
> 
> So we would need to add basic auth to HTTP?

@alexec users can add token in header. I don't think there is a need to add auth to HTTP

alexec:
I meant, that if we want to re-use the HTTP code for the artifact driver, we'd need a way for that driver to support username/password.

rohankmr414:
> I meant, that if we want to re-use the HTTP code for the artifact driver, we'd need a way for that driver to support username/password.

@alexec can we add a BasicAuth field in the spec just like the artifactory?
```go
type BasicAuth struct {
	// UsernameSecret is the secret selector to the repository username
	UsernameSecret *apiv1.SecretKeySelector `json:"usernameSecret,omitempty" protobuf:"bytes,1,opt,name=usernameSecret"`

	// PasswordSecret is the secret selector to the repository password
	PasswordSecret *apiv1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,2,opt,name=passwordSecret"`
}
.
.
.
 type HTTPArtifact struct { 
 	// URL of the artifact 
 	URL string `json:"url" protobuf:"bytes,1,opt,name=url"` 
  
 	// Headers are an optional list of headers to send with HTTP requests for artifacts 
 	Headers []Header `json:"headers,omitempty" protobuf:"bytes,2,opt,name=headers"` 
 	
 	BasicAuth `json:",inline" protobuf:"bytes,2,opt,name=basicAuth"`
 } 
```

https://github.com/argoproj/argo-workflows/blob/4471b59a52873ca66d6834a06519407c858f5906/pkg/apis/workflow/v1alpha1/workflow_types.go#L2274-L2280

rohankmr414:
or we can base64 decode the basic authorization header entered by the user to get the username password?

alexec:
Seems smart to me.

rohankmr414:
@alexec 
I have made the required changes, can u please review it?

alexec (approved):

PR #8414, opened 2022-04-19
feat: added support for http as option for artifact upload. Fixes #785

### LINKED ISSUE #785: Allow 'http' as an option for artifact upload

rohankmr414:

Fixes #785


workflow/artifacts/artifacts.go
```diff
@@ -79,7 +78,23 @@ func NewDriver(ctx context.Context, art *wfv1.Artifact, ri resource.Interface) (
 		return &driver, nil
 	}
 	if art.HTTP != nil {
-		return &http.ArtifactDriver{}, nil
+		if art.HTTP.UsernameSecret != nil && art.HTTP.PasswordSecret != nil {
```
alexec, line 81: I think this needs to be the same as below, you can have a password without a users (e.g. when you use a Personal Access Token)

alexec (approved):

rohankmr414:
@alexec for some reason the unit tests are failing on `master`

PR #8460, opened 2022-04-23
feat: Track UI event #8402

### LINKED ISSUE #8402: Track/log UI usage
# Summary

We could add a beacon endpoint that could by the UI to track usage of the UI

```xml
<TrackEvent event='viewNodeSummary'/>
```

This just makes a HTTP request to `/api/v1/collect?t=trackEvent&event=viewNodeSummary`, similar to:

https://developers.google.com/analytics/devguides/collection/protocol/v1/reference

This event could be logged, with the SSO subject/email so Splunk queries could be created and we can speak to users about their experience.

---

momom-i:
Signed-off-by: momom <51011095+momom-i@users.noreply.github.com>

Fixes #8402

I've confirmed the logs  using this `TrackEvent` component, but as I wrote in the issue I'm not sure where to insert it and I've not implemented that part.


alexec:
This looks great! Can I ask for some changes?

I think all API operation IDs must be REST conventional verb+noun So what is our verb+noun? I think it is Collect+Event. Can you change it to `CollectEvent`? That means the request payload should be `CollectEventRequest` rather than `TrackUserInterfaceUsageRequest`, and it should be POST rather than GET. The HTTP URL should match the payload type, so it can we make it `/api/v1/tracking/event`?

Regarding the `Event`, I think we should have type. Google Analytics allows up to 25 arbitrary parameters. I think GRPC can do this with a `map<>`. Could you change that?

I think this should also include a test in argo_server_test.go, just to make sure the endpoint returns 200 OK.

Finally, for the UI, we could track lots of different things, but lets not do that! Instead, how about embedding the tracker in each page in the left hand nagging, workflow list, workflow template list etc. The type could be `openedWorkflowList` etc.

We should test that it works correctly, sometimes React just renders everything at once, so would send tracking events all at once. You can see what happens in your Brower’s the Web Inspector Network view.

If you don’t know this already, you can start the UI by running `make start UI=true`. 

I don’t think we need any feature flag to disable this. It is very low risk, and the data never leaves your company network.

alexec:
Great stuff by the way!

momom-i:
Thank you for your kind instructions!
I'm still working on it, and I think it'll be done tomorrow.
By the way, while trying to run argo-workflows locally, I met the error same as [this issue](https://github.com/mattn/goreman/issues/121) at [this](https://github.com/argoproj/argo-workflows/blob/e368f504307ecbbfa2f228bccfab1f943efaca9d/Makefile#L465).
I bumped up goreman version and it seemed to be fixed. 
If this situation seems to occur in someone's environment, I'd like to create its issue. What do you think? 🤔

alexec:
Looks like we need a new version of goreman, do you want to update the Makefile?

https://github.com/argoproj/argo-workflows/blob/e368f504307ecbbfa2f228bccfab1f943efaca9d/Makefile#L465

ui/src/app/shared/components/track-user-interface-event.tsx
```diff
@@ -0,0 +1,10 @@
+import {services} from '../../shared/services';
+import { EventParam } from '../services/info-service';
+
+export const TrackEvent = ({name}: {name: string}): null => {
```
alexec, line 4: minor - no need to specify null,

alexec (changes requested):

alexec:
Run `yarn --cwd ui lint` to fix UI problems. Run `make pre-commit -B` to fix codegen issues.

.github/workflows/gh-pages.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       - run: pip install mkdocs==1.2.4 mkdocs_material==8.1.9
       - run: mkdocs build
       - run: make parse-examples
-      - uses: peaceiris/actions-gh-pages@v2.9.0
+      - uses: peaceiris/actions-gh-pages@v3.8.0
```
alexec, line 27: ??

server/info/info_server.go
```diff
@@ -50,6 +53,27 @@ func (i *infoServer) GetVersion(context.Context, *infopkg.GetVersionRequest) (*w
 	return &version, nil
 }
 
+func (i *infoServer) CollectEvent(ctx context.Context, req *infopkg.CollectEventRequest) (*infopkg.CollectEventResponse, error) {
+	if len(req.Param) > 5 {
```
alexec, line 57: 👍

alexec:
Looks good. You should revert the change to gh-pages file. 

Can I check - when you run the UI, do the events get recorded, only when the page is displayed?

ui/src/app/shared/components/track-user-interface-event.tsx
momom-i: https://github.com/argoproj/argo-workflows/runs/6196069776?check_suite_focus=true#step:5:1333
if I remove this, error occurs like this.

pkg/apiclient/info/info.proto
```diff
@@ -42,4 +49,10 @@ service InfoService {
   rpc GetUserInfo(GetUserInfoRequest) returns (GetUserInfoResponse) {
     option (google.api.http).get = "/api/v1/userinfo";
   }
+  rpc CollectEvent(CollectEventRequest) returns (CollectEventResponse) {
+    option (google.api.http) = {
+      post : "/api/v1/tracking/event"
+      body : "*"
```
alexec, line 55: try changing ithis to `body: "`param`"`

.github/workflows/gh-pages.yaml
momom-i: I didn't touch this yaml 🤔  It's a bit weird but I'm reverting it.

pkg/apiclient/info/info.proto
momom-i: I tried it, but it didn't work.
After changing it and generating code from this proto file, I met the error like 
```sh
--grpc-gateway_out: repeated field not allowed in field path: param in param
```
It seems to be related to https://github.com/grpc-ecosystem/grpc-gateway/issues/765 and unfortunately it's still not supported.

alexec: Hmm, my smart idea of using `map<>` is look less smart by the moment.

Lets pivot, go back to your original idea. Lets just have a singe `name` field. We can add other in the future if we want them.

I hope you don't mind my about face.

momom-i:
> Can I check - when you run the UI, do the events get recorded, only when the page is displayed?

Yes it is in my local. I saw sometimes requests were sending 2 or 3 times per page, but they send it only when the page is displayed.

pkg/apiclient/info/info.proto
momom-i: I understand! Thank you very much for your advice :)

alexec:
I think it is important we only send one request per page. I think the page is probably getting re-rendered due to a React key change. I'll investigate.

momom-i:
> I think it is important we only send one request per page. I think the page is probably getting re-rendered due to a React key change. I'll investigate.

Thank you, I'm also trying to find a way to do so.

alexec:
Another, perhaps easier, solution would be to rate limit the track on the client side. I.e. the React component keeps track of the last time it sent an event, and does not send another one, even if asked, for (say) 5s.

alexec:
Those should both work too.

momom-i:
I removed TrackEvent component because it just passes to service function and resends requests.
Now tracking log looks like as follows and is sent only when each page is displayed at the beginning:
```sh
argo-server | time="2022-05-01T03:09:51.503Z" level=info msg="tracking UI usage️️" name=openedWorkflowList
```

ui/src/app/apidocs/components/apiDocs.tsx
```diff
@@ -1,13 +1,20 @@
...
 
-export const ApiDocs = () => (
-    <Page title='Swagger'>
-        <div className='argo-container'>
-            <SwaggerUI url={uiUrl('assets/openapi-spec/swagger.json')} defaultModelExpandDepth={0} />
-        </div>
-    </Page>
-);
+export const ApiDocs = () => {
+    useEffect(() => {
```
alexec, line 10: could we wrap this. func up in, say `useCollectEvent('openedApiDocs');`, if that does not work, then `useEffect(() => collectEvent('openedApiDocs'))`

momom-i: I think we can wrap that where `useEffect` is used, but what should we do where it's used in class like [here](https://github.com/argoproj/argo-workflows/blob/bd1c2baf57083a0c2b35be8c111e0620bec3ad43/ui/src/app/workflows/components/workflows-list/workflows-list.tsx#L126)?
Should we leave them as it is and just wrap it for `useEffect` use case?

alexec: Use both?

momom-i: Ah OK, I'm sorry for the confusion.

alexec (changes requested):
I have one last change request. Can you add tracking to `artifact-panel.tsx`? This is a new feature I'd like to track usage of. When that is done, I will approve and merge.

alexec (approved):

alexec:
LGTM! 🚀

PR #8468, opened 2022-04-25
feat: Add WebHDFS support for HTTP artifacts. Fixes #7540

### LINKED ISSUE #7540: support webhdfs artifacts
# Summary

There is already support for hfds in argo-workflows (https://github.com/argoproj/argo-workflows/blob/master/examples/hdfs-artifact.yaml#L41) but not for the webhdfs protocol. I had a quick look in this repository and I could not find another question about webhdfs.  My team could help in with this implementation (i.e. implement the solution and open a PR in this repository). My question is if this kind of addition would we merged (given that it will follow the guidelines and address any comments that might arise in the PR)?

# Use Cases

Some datasets can be accessed using the webhdfs protocol. We would like to read and write them through argo workflows.

---

alexdittmann:
Signed-off-by: Alexander Dittmann <alexander.dittmann@sap.com>

Fixes #7540

This PR adds support for webHDFS for HTTP input/output artifacts. This is accomplished by:
- explicitly following temporary redirects (status code 307, see [here](https://github.com/argoproj/argo-workflows/blob/87108995155c88197ede933feb2ec90757859867/workflow/artifacts/http/http.go#L109-L112)) 
- adding supporting for two ways of authentication:
  - via certificates: the client authenticates via a tls certificate pair
  - via OAuth2: the client authenticates via an OAuth2 token

An example yaml can be found here: https://github.com/argoproj/argo-workflows/compare/master...alexdittmann:feat-webhdfs-support-for-http-art?expand=1#diff-731816132908abcbbf4286af6ac8c6d9241ef1d7de7f6f00d92a6416e8106e8f


workflow/controller/workflowpod.go
```diff
@@ -1150,6 +1158,14 @@ func createSecretVolume(volMap map[string]apiv1.Volume, art wfv1.Artifact, keyMa
...
 	} else if art.GCS != nil {
 		createSecretVal(volMap, art.GCS.ServiceAccountKeySecret, keyMap)
+	} else if art.HTTP != nil {
+		createSecretVal(volMap, art.HTTP.BasicAuth.UsernameSecret, keyMap)
+		createSecretVal(volMap, art.HTTP.BasicAuth.PasswordSecret, keyMap)
+		createSecretVal(volMap, art.HTTP.ClientCert.ClientCertSecret, keyMap)
+		createSecretVal(volMap, art.HTTP.ClientCert.ClientKeySecret, keyMap)
+		createSecretVal(volMap, art.HTTP.OAuth2.ClientIDSecret, keyMap)
+		createSecretVal(volMap, art.HTTP.OAuth2.ClientSecretSecret, keyMap)
+		createSecretVal(volMap, art.HTTP.OAuth2.TokenURLSecret, keyMap)
```
alexdittmann, line 1168: these lines (and the ones in the `createArchiveLocationSecret`) were not added in the recent update of the http artifact. But for my understanding they are needed, so i addded them. Correct me if im wrong.

workflow/artifacts/http/http.go
```diff
@@ -77,39 +81,64 @@ func (h *ArtifactDriver) Save(path string, outputArtifact *wfv1.Artifact) error
...
+		url = outputArtifact.HTTP.URL
+	}
+
+	res, err := h.doRequest(http.MethodPut, url, reader, outputArtifact)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == 307 && outputArtifact.HTTP.FollowTemporaryRedirects {
```
alexec, line 100: I think we should follow redirects always, I don't know when you'd not want to.

alexdittmann: i was not completely sure about that. Normally, the go client should follow redirects automatically, however, for these 307 it seems it doesnt follow (maybe there are some security concerns?). Thats why introduced the flag, to keep the "normal" behavior.

alexdittmann:
What is the procedure with the required checks before making the PR ready for review, is there some action required? Three of them passed, but the others are in the "Expected — Waiting for status to be reported". 

And another thing regarding updates of the PR: Do I always have to make sure that there is only one single commit, or is this only required for the PR creation? 
Thanks so far!

workflow/artifacts/http/http.go
alexec: https://github.com/golang/go/issues/7912

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2279,16 +2279,49 @@ type BasicAuth struct {
...
+type OAuth2Auth struct {
+	ClientIDSecret     *apiv1.SecretKeySelector `json:"clientIDSecret,omitempty" protobuf:"bytes,1,opt,name=clientIDSecret"`
+	ClientSecretSecret *apiv1.SecretKeySelector `json:"clientSecretSecret,omitempty" protobuf:"bytes,2,opt,name=clientSecretSecret"`
+	TokenURLSecret     *apiv1.SecretKeySelector `json:"tokenURLSecret,omitempty" protobuf:"bytes,3,opt,name=tokenURLSecret"`
+	Scopes             []string                 `json:"scopes,omitempty" protobuf:"bytes,5,rep,name=scopes"`
+	EndpointParams     []EndpointParam          `json:"endpointParams,omitempty" protobuf:"bytes,4,rep,name=endpointParams"`
+}
+
+// EndpointParam is for requesting optional fields that should be sent in the oauth request
+type EndpointParam struct {
```
alexec, line 2298: maybe call this `OAuth2EndpointParam`

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2279,16 +2279,49 @@ type BasicAuth struct {
 	PasswordSecret *apiv1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,2,opt,name=passwordSecret"`
 }
 
+// ClientCertAuth holds necessary information for client authentication via certificates
+type ClientCertAuth struct {
+	ClientCertSecret *apiv1.SecretKeySelector `json:"clientCertSecret,omitempty" protobuf:"bytes,1,opt,name=clientCertSecret"`
```
alexec, line 2284: I'm not sure we suffix `Secret` in other places, can you check please?

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2279,16 +2279,49 @@ type BasicAuth struct {
...
-	Headers []Header `json:"headers,omitempty" protobuf:"bytes,2,opt,name=headers"`
+	Headers []Header `json:"headers,omitempty" protobuf:"bytes,2,rep,name=headers"`
+
+	// HTTPAuth contains information for client authentication
+	HTTPAuth `json:",inline" protobuf:"bytes,3,opt,name=hTTPAuth"`
 
-	// BasicAuth is the secret selector for basic authentication
-	BasicAuth `json:",inline" protobuf:"bytes,3,opt,name=basicAuth"`
+	// whether to follow temporary redirects, needed for webHDFS
+	FollowTemporaryRedirects bool `json:"followTemporaryRedirects,omitempty" protobuf:"varint,4,opt,name=followTemporaryRedirects"`
```
alexec, line 2324: This defaults to false. You can change it to `*bool` if you want it to default to `nil` (which can imply true).

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2279,16 +2279,49 @@ type BasicAuth struct {
...
 type HTTPArtifact struct {
 	// URL of the artifact
 	URL string `json:"url" protobuf:"bytes,1,opt,name=url"`
 
 	// Headers are an optional list of headers to send with HTTP requests for artifacts
-	Headers []Header `json:"headers,omitempty" protobuf:"bytes,2,opt,name=headers"`
+	Headers []Header `json:"headers,omitempty" protobuf:"bytes,2,rep,name=headers"`
+
+	// HTTPAuth contains information for client authentication
+	HTTPAuth `json:",inline" protobuf:"bytes,3,opt,name=hTTPAuth"`
```
alexec, line 2321: This feature is new and unreleased. I think we should not use `,inline`. Instead, please have a `json:auth,omitempty` field. HTTPAuth will need to change to `*HTTPAuth` .

workflow/artifacts/artifacts.go
```diff
@@ -87,21 +88,52 @@ func newDriver(ctx context.Context, art *wfv1.Artifact, ri resource.Interface) (
 		return &driver, nil
 	}
 	if art.HTTP != nil {
+		client := gohttp.DefaultClient
```
alexec, line 91: I think you need to `.Clone()` this

workflow/artifacts/artifacts.go
```diff
@@ -87,21 +88,52 @@ func newDriver(ctx context.Context, art *wfv1.Artifact, ri resource.Interface) (
...
+			clientKey, err := ri.GetSecret(ctx, art.HTTP.ClientCert.ClientKeySecret.Name, art.HTTP.ClientCert.ClientKeySecret.Key)
+			if err != nil {
+				return nil, err
+			}
+			client, err = http.CreateClientWithCertificate([]byte(clientCert), []byte(clientKey))
+			if err != nil {
+				return nil, err
+			}
+		}
+		driver.Client = client
```
alexec, line 136: should you be setting `client.CheckRedirect` here? 

https://stackoverflow.com/questions/23297520/how-can-i-make-the-go-http-client-not-follow-redirects-automatically

workflow/artifacts/http/util.go
```diff
@@ -0,0 +1,41 @@
+package http
```
alexec, line 1: maybe rename this file `client_cert.go`

workflow/artifacts/http/util_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+)
+
+func TestCreateOauth2Client(t *testing.T) {
```
alexec, line 11: great tests!

alexec (changes requested):
some minor change, I think we probably want to have an E2E test for HTTP, can you check? if not, can you add a `http_test.go`, you might be able use the `minio` to test against.

workflow/artifacts/http/util.go
alexdittmann: i would go for just `clients.go`, since the oauth2 client is also in there

workflow/artifacts/http/http.go
alexdittmann: thanks for pointing me to this. I finally found out that i have to manually define the `GetBody func() (io.ReadCloser, error)` for the redirect to work. I think, we can then also get rid of the `followRedirectFlag`.

pkg/apis/workflow/v1alpha1/workflow_types.go
alexdittmann: as suggested above i would now remove this and add the `GetBody func() (io.ReadCloser, error)`per default

workflow/artifacts/artifacts.go
alexdittmann: with the `GetBody func() (io.ReadCloser, error)`works "natively", so i dont think we need this

pkg/apis/workflow/v1alpha1/workflow_types.go
alexdittmann: yes, its done e.g. here: https://github.com/argoproj/argo-workflows/blob/1a7d50fd32a30070e78bdf544dd709c20647378d/pkg/apis/workflow/v1alpha1/workflow_types.go#L2032-L2036

alexdittmann:
> some minor change, I think we probably want to have an E2E test for HTTP, can you check? if not, can you add a `http_test.go`, you might be able use the `minio` to test against.
First of all, thanks for your feedback!
I agree, an e2e test would make sense, but I am not sure how to start. I was looking for an example and i only found something for s3 in the `e2e/functional_test.go` (e.g. [here](https://github.com/argoproj/argo-workflows/blob/master/test/e2e/functional_test.go#L859)). Should  I create something similar for http in a new `http_test.go` file or am i on the wrong track?

alexec:
Yes. I would base it on `ArtifactsSuite`.

alexec:
To write through tests, you'd need to start (I'm guessing) Hadoop. We do something similar for S3, by starting a single MinIO pod in the test system. MinIO is lightweight, so is Hadoop lightweight?

test/e2e/http_artifacts_test.go
```diff
@@ -0,0 +1,63 @@
+//go:build executor
+// +build executor
```
alexec, line 2: I think this should be in `functional` set of tests.

alexdittmann:
Can we make this ready for review and enable the tests? Or is there still something missing?

test/e2e/functional_test.go
```diff
@@ -956,6 +956,46 @@ func (s *FunctionalSuite) TestPauseAfterAndBefore() {
 		WaitForWorkflow(fixtures.ToBeSucceeded)
 }
 
+func (s *FunctionalSuite) TestInputArtifactHttp() {
```
alexec, line 959: So we can just run HTTP tests, please we move these to their own file? It should still have the `functional` build tag.

test/e2e/functional_test.go
```diff
@@ -956,6 +956,46 @@ func (s *FunctionalSuite) TestPauseAfterAndBefore() {
...
+	s.Given().
+		Workflow("@testdata/input-artifact-http.yaml").
```
alexec, line 961: Minor - but you could put them in `testdata/http` so they're in one place.

test/e2e/functional_test.go
```diff
@@ -956,6 +956,46 @@ func (s *FunctionalSuite) TestPauseAfterAndBefore() {
...
+
+func (s *FunctionalSuite) TestOAuthArtifactHttp() {
+	s.Given().
+		Workflow("@testdata/oauth-artifact-http.yaml").
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeSucceeded)
+}
+
+func (s *FunctionalSuite) TestClientCertAuthArtifactHttp() {
```
alexec, line 991: This is an 1st-class set of tests!

alexec (commented):
Just more minor comments. Because the PR is draft, I've only provided comments so far. One you mark it ready for review I'll check it out and run it myself and give a thorough review.

test/e2e/functional_test.go
alexdittmann: yes i see! i misunderstood another comment before

alexec:
CI builds did not run. You also need to sync with master. Syncing should cause the builds to run.

alexec:
We should have a doc (in ./docs) to show how to use this. E..g with Hadoop. This could be in this PR, or in another PR/issue.

 #8635

alexec (approved):

PR #8486, opened 2022-04-26
feat: Artifact streaming: enable artifacts to be streamed to users rather than loading the full file to disk first. Fixes #8396

### LINKED ISSUE #8396: `argo cp`/UI: Artifact streaming
Today, when we work on artifacts, we download the entires object and copy it around.

This is extremely memory and disk intensive for large artifacts (e.g. 1Gb). Instead, we should use the streaming APIs when available to serve artifacts.

Tasks

- [x] Upgrade argoproj/pkg to v0.12.0 to get the new S3 OpenFile func.
- [ ] Modify `ArtifactDriver` to add `Open` func. Implement for drivers that can return files (e.g. S3).
- [ ] Modify `artifact_server.go` to use `Open` rather than `Load`.

juliev0:
Fixes #8396


server/artifacts/artifact_server.go
```diff
@@ -186,41 +182,31 @@ func (a *ArtifactServer) returnArtifact(ctx context.Context, w http.ResponseWrit
...
-
-	contentLength := strconv.FormatInt(stats.Size(), 10)
-	log.WithFields(log.Fields{"size": contentLength}).Debug("Artifact file size")
-
 	key, _ := art.GetKey()
 	w.Header().Add("Content-Disposition", fmt.Sprintf(`filename="%s"`, path.Base(key)))
 
-	http.ServeContent(w, r, "", time.Time{}, file)
+	// Send the initial headers saying we're going to stream the response.
+	w.Header().Set("Transfer-Encoding", "chunked") // todo: do we need this?
```
juliev0, line 201: @alexec  were you saying you didn't think we need this?

alexec: not needed

server/artifacts/artifact_server.go
```diff
@@ -186,41 +182,28 @@ func (a *ArtifactServer) returnArtifact(ctx context.Context, w http.ResponseWrit
...
+	stream, err := driver.OpenStream(art)
 	if err != nil {
 		return err
 	}
 
 	defer func() {
-		if err := file.Close(); err != nil {
-			log.Fatalf("Error closing file[%s]: %v", tmpPath, err)
+		if err := stream.Close(); err != nil {
+			log.Fatalf("Error closing stream[%s]: %v", stream, err)
```
alexec, line 193: this should have been fixed ages ago - Fatal will panic and crash! We don't want that at all - instead, just ignore the error. We will never take any action.

server/artifacts/artifact_server.go
```diff
@@ -186,41 +182,28 @@ func (a *ArtifactServer) returnArtifact(ctx context.Context, w http.ResponseWrit
...
-	contentLength := strconv.FormatInt(stats.Size(), 10)
-	log.WithFields(log.Fields{"size": contentLength}).Debug("Artifact file size")
-
 	key, _ := art.GetKey()
 	w.Header().Add("Content-Disposition", fmt.Sprintf(`filename="%s"`, path.Base(key)))
 
-	http.ServeContent(w, r, "", time.Time{}, file)
+	_, err = io.Copy(w, stream)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
alexec, line 202: I don't think this line is needed? the calling function should capture the error and do it

workflow/artifacts/common/common.go
```diff
@@ -1,12 +1,18 @@
...
+
+	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+)
 
 // ArtifactDriver is the interface for loading and saving of artifacts
 type ArtifactDriver interface {
 	// Load accepts an artifact source URL and places it at specified path
 	Load(inputArtifact *v1alpha1.Artifact, path string) error
 
+	OpenStream(inputArtifact *v1alpha1.Artifact) (io.ReadCloser, error)
```
alexec, line 14: The qualifier "input" is not needed,  this can be any kind of artifact, I'd just change to "a" in fact. You should make this change in each driver.

I think we should add a comment saying the following:

```golang
// OpenStream opens an artifact for reading. If the artifact is a file, then the file should be opened.
// If the artifact is a directory, the driver may return that as a tarball. OpenStream is intended to be efficient, 
// so implementations should minimise usage of disk, CPU and memory. 
// Implementations must not implement retry mechanisms. This will be handled by the client, so would result in O(nm) cost.
```

workflow/artifacts/s3/s3.go
```diff
@@ -103,6 +104,43 @@ func loadS3Artifact(s3cli argos3.S3Client, inputArtifact *wfv1.Artifact, path st
...
+		return nil, fmt.Errorf("failed to test if %s is a directory: %v", inputArtifact.S3.Key, err)
+	}
+	if !isDir {
+		// It's neither a file, nor a directory. Return the original NoSuchKey error
+		return nil, errors.New(errors.CodeNotFound, origErr.Error())
+	}
+	// directory case:
+	// todo: make a .tgz file which can be streamed to user
+	return nil, fmt.Errorf("Directory Stream capability currently unimplemented for S3")
+}
```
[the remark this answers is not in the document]
alexec: I think this is fine.

workflow/artifacts/s3/s3_test.go
```diff
@@ -57,7 +57,11 @@ func (s *mockS3Client) GetFile(bucket, key, path string) error {
 }
 
 func (s *mockS3Client) OpenFile(bucket, key string) (io.ReadCloser, error) {
-	return io.NopCloser(&bytes.Buffer{}), s.getMockedErr("OpenFile")
+	err := s.getMockedErr("OpenFile")
+	if err == nil {
+		return io.NopCloser(&bytes.Buffer{}), nil
```
alexec, line 62: clever!

workflow/artifacts/s3/s3_test.go
```diff
@@ -110,6 +114,121 @@ func (s *mockS3Client) MakeBucket(bucketName string, opts minio.MakeBucketOption
 	return s.getMockedErr("MakeBucket")
 }
 
+func TestOpenStreamS3Artifact(t *testing.T) {
```
alexec, line 117: love this detailed test

workflow/artifacts/s3/s3.go
```diff
@@ -103,6 +104,43 @@ func loadS3Artifact(s3cli argos3.S3Client, inputArtifact *wfv1.Artifact, path st
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new S3 client: %v", err)
+	}
+
+	return streamS3Artifact(s3cli, inputArtifact)
+
+}
+
+func streamS3Artifact(s3cli argos3.S3Client, inputArtifact *wfv1.Artifact) (io.ReadCloser, error) {
+	stream, origErr := s3cli.OpenFile(inputArtifact.S3.Bucket, inputArtifact.S3.Key)
```
alexec, line 123: minor - but `origErr` could just be `err` I think

workflow/artifacts/s3/s3.go
```diff
@@ -103,6 +104,43 @@ func loadS3Artifact(s3cli argos3.S3Client, inputArtifact *wfv1.Artifact, path st
...
+		return nil, fmt.Errorf("failed to get file: %v", origErr)
+	}
+	// If we get here, the error was a NoSuchKey. The key might be an s3 "directory"
+	isDir, err := s3cli.IsDirectory(inputArtifact.S3.Bucket, inputArtifact.S3.Key)
+	if err != nil {
+		return nil, fmt.Errorf("failed to test if %s is a directory: %v", inputArtifact.S3.Key, err)
+	}
+	if !isDir {
+		// It's neither a file, nor a directory. Return the original NoSuchKey error
+		return nil, errors.New(errors.CodeNotFound, origErr.Error())
```
alexec, line 137: correct, good stuff

workflow/artifacts/s3/s3.go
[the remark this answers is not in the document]
alexec: the error should be errors.CodeNotImplemented

workflow/artifacts/common/load_to_stream.go
```diff
@@ -0,0 +1,36 @@
...
+func (w selfDestructingFile) Close() error {
+	err := w.File.Close()
+	_ = os.Remove(w.Name())
+	return err
+}
+
+// Use ArtifactDriver.Load() to get a stream, which we can use for all implementations of ArtifactDriver.OpenStream()
+// that aren't yet implemented the "right way" and/or for those that don't have a natural way of streaming
+func LoadToStream(a *wfv1.Artifact, g ArtifactDriver) (io.ReadCloser, error) {
+	filename := "/tmp/" + rand.String(32)
```
alexec, line 26: Can we add a log line here? I'd like it to say:

"Efficient artifact streaming is not supported for %s: see INSERT GITHUB ISSUE HERE", g.GetType()

Maybe users will see it and implement for us?

alexec (changes requested):

server/artifacts/artifact_server.go
juliev0: Curious how the calling function knows which type of status based on error (i.e. 400 level vs 500 level), but I can take a look

workflow/artifacts/common/common.go
juliev0: is that an error in the Load method above too then?

server/artifacts/artifact_server.go
juliev0: how about maybe just a warning? otherwise, for the drivers that are using the common code, they'll leave files around and we may be more likely to notice if there's a warning ?

workflow/artifacts/common/common.go
alexec: no need - just for this

server/artifacts/artifact_server.go
juliev0: I see the caller is assuming it's a 500 error. That's fine in this case at least so I'll use it.

juliev0: I presume the call to writeHeader(http.StatusOK) below is probably also unnecessary...

workflow/artifacts/common/load_to_stream.go
juliev0: Info or Warn level?

server/artifacts/artifact_server.go
alexec: check the call-stack?

workflow/artifacts/common/load_to_stream.go
juliev0: Are you referring to the current "Artifact Streaming" github issue or some new issue we leave outstanding related to implementing the rest of these?

workflow/artifacts/s3/s3.go
juliev0: It's used on line 137 after err is modified

workflow/artifacts/common/load_to_stream.go
alexec: Warn level I think is probably correct. Yes, link to the artifact streaming issue.

workflow/artifacts/logging/driver.go
```diff
@@ -30,6 +31,10 @@ func (d driver) Load(a *wfv1.Artifact, path string) error {
 	return err
 }
 
+func (d driver) OpenStream(inputArtifact *wfv1.Artifact) (io.ReadCloser, error) {
+	return common.LoadToStream(inputArtifact, d)
```
alexec, line 35: This is wrong, the driver is a decorator, and should delegate all calls to `d.delegate`. It should also logs the same kind of info as `Load`.

alexec (changes requested):

workflow/artifacts/common/load_to_stream.go
juliev0: added a new issue and using Info level

alexec (approved):

workflow/artifacts/gcs/gcs.go
```diff
@@ -200,6 +200,11 @@ func listByPrefix(client *storage.Client, bucket, prefix, delim string) ([]strin
 	return results, nil
 }
 
+func (g *ArtifactDriver) OpenStream(a *wfv1.Artifact) (io.ReadCloser, error) {
+	// todo: this is a temporary implementation which loads file to disk first
+	return common.LoadToStream(a, g)
+}
+
```
moonyoungCHAE, line 207: @juliev0 
Hi. I found a todo comment you left.
I'm wondering about your intention with the LoadToStream function.
I'd like to improve it. :)

PR #8490, opened 2022-04-26
feat: Added a delete function to the artifacts storage. Fixes #8470

### LINKED ISSUE #8470: Add `ArtifactDriver.Delete` (internal API)
Create a `Delete` func to support #3102 .

https://github.com/argoproj/argo-workflows/blob/e368f504307ecbbfa2f228bccfab1f943efaca9d/workflow/artifacts/common/common.go#L11

Only implement for S3.

https://github.com/argoproj/argo-workflows/blob/e368f504307ecbbfa2f228bccfab1f943efaca9d/workflow/artifacts/s3/s3.go#L22

If we have not yet implemented we should return an error, and that error should ask the user to vote for support by linking to a Github issue.

Implementations should be idempotent.

mboussarov:
Fixes #8470

### DIFF
--- a/workflow/artifacts/s3/s3.go
+++ b/workflow/artifacts/s3/s3.go
@@ -3,13 +3,14 @@ package s3
 import (
 	"context"
 	"fmt"
-	"io"
-	"os"
-
+	argoerr "github.com/argoproj/argo-workflows/v3/util/errors"
 	"github.com/argoproj/pkg/file"
 	argos3 "github.com/argoproj/pkg/s3"
 	"github.com/minio/minio-go/v7"
 	log "github.com/sirupsen/logrus"
+	"io"
+	"k8s.io/client-go/util/retry"
+	"os"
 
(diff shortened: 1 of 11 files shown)


mboussarov:
Fixes #8470

workflow/artifacts/s3/s3.go
```diff
@@ -155,6 +156,23 @@ func (s3Driver *ArtifactDriver) Save(path string, outputArtifact *wfv1.Artifact)
 	return err
 }
 
+// Delete deletes an artifact from an S3 compliant storage
+func (s3Driver *ArtifactDriver) Delete(artifact *wfv1.Artifact) error {
+	ctx, cancel := context.WithCancel(context.Background())
```
alexec, line 161: minor - but @juliev0 and I found out this context is wrong, it should not have a cancel, it should just be `context.TODO()` - minor - so only

alexec (approved):

PR #8491, opened 2022-04-26
fix: Save containerSet logs in logs archive fixes #7897

### LINKED ISSUE #7897: Save containerSet logs in logs archive

noam-codefresh:
Signed-off-by: Noam Gal [noam.gal@codefresh.io](mailto:noam.gal@codefresh.io)

Fixes https://github.com/argoproj/argo-workflows/issues/7897

Re-opens https://github.com/argoproj/argo-workflows/pull/7903


alexec (approved):

alexec:
@noam-codefresh 

image

But it is not displaying them. Github has been less robust of late. Maybe transient issue.

noam-codefresh:
the conflict was a single log line that was removed in master `executor.go`

alexec:
🚀

PR #8517, opened 2022-04-27
Artifact GC: Implement Workflow.SearchArtifacts(). Fixes #8473

### LINKED ISSUE #8473: Artifact GC: Implement `Workflow.SearchArtifacts()` and tests
We need a func to search artifacts by various query. For MVP, this only need to be the GC strategy, but we will later on add options like `ArtifactName` or `TemplateName`:

```golang
type ArtifactSearchQuery struct {
   ArtiactGCStrategies [] ArtiactGCStrategy
}

func (w *Workflow) SearchArtifacts(ArtifactSearchQuery) Artifacts { 
   // ...
}
```

It should return all output artifacts that have that strategy. The logic is a bit more complex, because not all fields from template are copied into the `NodeStatus`, but you can query 

```golang
	for _, n := range wf.Status.Nodes {
 		// NodeStatus has the ArtifactGC field, but it is never set
 		t := wf.GetTemplateByName(n.TemplateName)
 		for _, a := range n.GetOutputs().GetArtifacts() {
 			artifactByName := t.GetOutputs().GetArtifactByName(a.Name)
 			trueStrategy := artifactByName.GetArtifactGC().GetStrategy()
````

### PR DESCRIPTION
**Pull request by dpadhiar**
Fixes #8473 

Implements `workflow.SearchArtifacts()` to query artifacts based on specified artifactGC strategy.

### CODE REVIEW

File: pkg/apis/workflow/v1alpha1/workflow_types.go [hunk 1]
```diff
@@ -447,6 +447,15 @@ func (wfs WorkflowSpec) GetVolumeClaimGC() *VolumeClaimGC {
 	return wfs.VolumeClaimGC
 }
 
+// ArtifactGC returns the ArtifactGC that was defined in the workflow spec.  If none was provided, a default value is returned.
+func (wfs WorkflowSpec) GetArtifactGC() *ArtifactGC {
+	if wfs.ArtifactGC == nil {
```
alexec, line 452: smart

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -752,6 +761,14 @@ func (tmpl *Template) setTemplateObjs(steps []ParallelSteps, dag *DAGTemplate, c
 	tmpl.Suspend = suspend
 }
 
+func (tmpl *Template) GetOutputs() *Outputs {
+	return &tmpl.Outputs
```
alexec, line 765: minor - we can add a nil check in the future to make this nil-safe

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1233,6 +1250,39 @@ func (r *ArtifactRepositoryRefStatus) String() string {
...
 }
 
+type ArtifactSearchQuery struct {
+	ArtifactGCStrategies map[ArtifactGCStrategy]bool `protobuf:"bytes,1,rep,name=artifactGCStrategies,castkey=ArtifactGCStrategy"`
+}
+
+func NewArtifactSearchQuery() *ArtifactSearchQuery {
+	var q ArtifactSearchQuery
+	q.ArtifactGCStrategies = make(map[ArtifactGCStrategy]bool)
+	return &q
```
alexec, line 1260: good, very minor - but i would do `&ArtifactSearchQuery{ArtifactGCStrategies: ...}`, but that's me

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1252,6 +1302,10 @@ type Outputs struct {
 	ExitCode *string `json:"exitCode,omitempty" protobuf:"bytes,4,opt,name=exitCode"`
 }
 
+func (o *Outputs) GetArtifacts() Artifacts {
+	return o.Artifacts
```
alexec, line 1306: minor - we can make this nil-safe in the future

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2025,6 +2079,10 @@ func (n *NodeStatus) GetTemplateRef() *TemplateRef {
 	return n.TemplateRef
 }
 
+func (n *NodeStatus) GetOutputs() *Outputs {
+	return n.Outputs
```
alexec, line 2083: minor - we can make this nil-safe in the future

pkg/apis/workflow/v1alpha1/workflow_types_test.go
```diff
@@ -753,6 +768,55 @@ func TestPrometheus_GetDescIsStable(t *testing.T) {
...
+		Spec: WorkflowSpec{
+			ArtifactGC: &ArtifactGC{
+				Strategy: ArtifactGCOnWorkflowCompletion,
+			},
+			Templates: []Template{
+				{
+					Name: "test-template",
+					Outputs: Outputs{
+						Artifacts: Artifacts{
+							Artifact{Name: "test-artifact", Path: ""},
```
alexec, line 786: very minor - default value for string is "", so you could omit `Path: ""`

alexec (approved):

pkg/apis/workflow/v1alpha1/workflow_types.go
dpadhiar: That's a quick fix, in that case do we want to return an empty `Outputs` or `nil` still if `tmpl.Outputs == nil`? Can do the same fix for the others.

alexec: I think you can rleae for the moment - I'd return nil

PR #8530, opened 2022-04-28
feat: Artifact GC - POC

alexec:
Hi @jessesuen ,

This PoC is the last part of Artifact GC.

## Why do we want Artifact GC?

Deleting artifacts at on workflow completion or deletion saves users cost. It also makes sure that defunct data is deleted. 

## Why Can't You Use The Artifact Repository Built-in Mechanisms?

It is impossible for S3 to know when a workflow is deleted.

## Solution

We must delete workflows by running a processes (i.e. pod) in the user namespace:

* Each workflow might have a different IAM role annotation. The controller cannot be given IAM roles, because it cannot be give multiple values.
* The agent's design is orientated around task-set which is not amenable to the job.

We need to store:

* Was the artifact deleted?

We need a way to communicate between the user pod and the controller:

* Please delete an artifact.
* Was delete successful?

Currently a pod can update the following:

* `WorkflowTaskResult` - this is for results, not jobs.
* `WorkflowTaskSet` - this is for running templates, we do not have templates
* `Pod` - interesting...

So...

* Please delete an artifact = create a "GC pod" with the info needed to delete the artifact
* Was the delete successful? = `pod.status.phase == "Succeeded"`

This means we need to run one pod to delete each artifact. This will always be at least equal to the number of nodes that produced artifacts. If every node produces 2 artifacts, then this will be 2x.

Won't that be expensive? No. We can control the resource requests. A pod deleting an artifact only needs enough CPU and memory to run the process and make the API call. 

Won't this result too many pods runinng, overwhelming the Kuberenetes API Server? We can limit the number of concurrent GC pods.


alexec:
@jessesuen as per Slack message, can I please ask you to review the design proposed here?

PR #8548, opened 2022-04-29
feat: New endpoint capable of serving directory listing or raw file, from non-archived or archived workflow

### LINKED ISSUE #8481: `argo cp`/UI: Argo Server `ArtifactServer.ServeFile` endpoint
Depends on #8396 

We need a more flexible endpoint to get artifact files from. 

The essential goal of this is to allow the Argo Server to allow directory like access to artifacts.

The format of the new URL is:

```
http://localhost:2746/artifact-files/{namespace}/[archived-workflows|workflows]/{id}/{nodeId}/outputs/{artifactName}
```

* `{namespace}` = The workflow namespace.
* `[archived-workflows|workflows]` the string literal. When this is `workflows`, it is a normal workflow. When `archived-workflow` then it is an archived workflow.
* `{id}` either the workflow name or the workflow UID.
* `{nodeId}` the node ID.
* `outputs` - the string literal "outputs". So that in the future we may support `inputs`. 
* `{artifactName}` - the name of the input or output artifact.

What should we return? 

We can use `ListObjects` to determine if an object is a directory (it'll error if it is a file) and return the file listing if not. This is only implemented for S3, but that is fine for MVP.

* If the artifact is a file, return that. 
* If the artifact is a directory, return a directory listing.

/artifact-files/{namespace}/[uid|name]/{id}/{nodeId}/outputs/{artifactName}/{filename}

## Access Control

The rule is, you can only see an artifact if you can see the workflow that created it. There is existing examples of this in the code-base.

juliev0:
Fixes #8481


server/artifacts/artifact_server.go
```diff
@@ -81,6 +83,174 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
 	a.getArtifact(w, r, true)
 }
 
+// single endpoint to be able to handle serving directories as well as files, both those that have been archived and those that haven't
```
alexec, line 86: good docs

server/artifacts/artifact_server.go
```diff
@@ -81,6 +83,174 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+
+	var fileName *string
+	requestPath := strings.Split(r.URL.Path, "/")
+	if len(requestPath) >= FILE_NAME_FIRST_INDEX+1 { // they included a file path in the URL (not just artifact name)
+		joined := strings.Join(requestPath[FILE_NAME_FIRST_INDEX:], "/")
+		// sanitize file name
+		cleanedPath := filepath.Clean(joined)
+		fileName = &cleanedPath
+	} else if len(requestPath) < ARTIFACT_NAME_INDEX+1 {
+		a.serverInternalError(fmt.Errorf("request path is not valid, expected at least %d fields, got %d", ARTIFACT_NAME_INDEX+1, len(requestPath)), w)
```
alexec, line 112: this should be changed to the brand new func `badRequest`

server/artifacts/artifact_server.go
```diff
@@ -81,6 +83,174 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+		ARCHIVE_DISCRIM_INDEX = 3
+		ID_INDEX              = 4
+		NODE_ID_INDEX         = 5
+		DIRECTION_INDEX       = 6
+		ARTIFACT_NAME_INDEX   = 7
+		FILE_NAME_FIRST_INDEX = 8
+	)
+
+	var fileName *string
+	requestPath := strings.Split(r.URL.Path, "/")
```
alexec, line 105: minor you could use `SplitN`, this will put all trailing part, saving the join

server/artifacts/artifact_server.go
```diff
@@ -81,6 +83,174 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+
+	namespace := requestPath[NAMESPACE_INDEX]
+	archiveDiscriminator := requestPath[ARCHIVE_DISCRIM_INDEX]
+	id := requestPath[ID_INDEX] // if archiveDiscriminator == "archived-workflows", this represents workflow UID; if archiveDiscriminator == "workflows", this represents workflow name
+	nodeId := requestPath[NODE_ID_INDEX]
+	direction := requestPath[DIRECTION_INDEX]
+	artifactName := requestPath[ARTIFACT_NAME_INDEX]
+
+	if direction != "outputs" { // for now we just handle output artifacts
+		a.serverInternalError(fmt.Errorf("request path is not valid, expected field at index %d to be 'outputs', got %s", DIRECTION_INDEX, direction), w)
```
alexec, line 124: `badRequest`

server/artifacts/artifact_server.go
```diff
@@ -81,6 +83,174 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+
+		wf, err = a.wfArchive.GetWorkflow(uid)
+		if err != nil {
+			a.serverInternalError(err, w)
+			return
+		}
+
+		// check that the namespace passed in matches this workflow's namespace
+		if wf.GetNamespace() != namespace {
+			a.serverInternalError(fmt.Errorf("request namespace '%s' doesn't match Workflow namespace: '%s'", namespace, wf.GetNamespace()), w)
```
alexec, line 160: badRequest I think

test/e2e/argo_server_test.go
```diff
@@ -1072,6 +1072,80 @@ func (s *ArgoServerSuite) TestArtifactServer() {
...
+	// In this case, the artifact name is a file
+	s.Run("GetArtifactFile", func() {
+		resp := s.e().GET("/artifact-files/argo/workflows/" + name + "/" + name + "/outputs/main-file").
+			Expect().
+			Status(200)
+
+		resp.Body().
+			Contains(":) Hello Argo!")
+
+		resp.Header("Content-Security-Policy").
```
alexec, line 1085: good

workflow/artifacts/s3/s3.go
```diff
@@ -221,8 +221,20 @@ func (s3Driver *ArtifactDriver) ListObjects(artifact *wfv1.Artifact) ([]string,
...
 			}
+			log.Debugf("successfully listing S3 directory associated with bucket: %s and key %s: %v", artifact.S3.Bucket, artifact.S3.Key, files)
 			return true, nil
 		})
 
 	return files, err
 }
+
+func (s3Driver *ArtifactDriver) IsDirectory(artifact *wfv1.Artifact) (bool, error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
alexec, line 232: minor -  use `context.TODO()` , the code you're copying is broken

server/artifacts/artifact_server.go
juliev0: Not sure if I understand what you mean by the advantage, but the reason I didn't use SplitN was because the number is undefined: could be "../artifactName", "../artifactName/filename", "../artifactName/filedir/filename", "../artifactName/filedir1/filedir2/filename", etc

server/artifacts/artifact_server.go
```diff
@@ -84,6 +86,174 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+// single endpoint to be able to handle serving directories as well as files, both those that have been archived and those that haven't
+// Valid requests:
+//  /artifact-files/{namespace}/[archived-workflows|workflows]/{id}/{nodeId}/outputs/{artifactName}
+//  /artifact-files/{namespace}/[archived-workflows|workflows]/{id}/{nodeId}/outputs/{artifactName}/{fileName}
+//  /artifact-files/{namespace}/[archived-workflows|workflows]/{id}/{nodeId}/outputs/{artifactName}/{fileDir}/.../{fileName}
+// 'id' field represents 'uid' for archived workflows and 'name' for non-archived
+func (a *ArtifactServer) GetArtifactFile(w http.ResponseWriter, r *http.Request) {
+
+	const (
+		NAMESPACE_INDEX       = 2
```
alexec, line 98: Minor - Golang convention for const is different to Java. They should be camel-case.`namespaceIndex`

server/artifacts/artifact_server.go
```diff
@@ -37,9 +39,9 @@ const (
 	EnvArgoArtifactXFrameOptions = "ARGO_ARTIFACT_X_FRAME_OPTIONS"
 	// DefaultContentSecurityPolicy is the default policy added to the Content-Security-Policy HTTP header
 	//   if no environment override has been added
-	DefaultContentSecurityPolicy = "sandbox; base-uri 'none'; default-src 'none'; image-src: 'self'; style-src: 'self'"
```
alexec, line 40: these were incorrect, @mboussarov will fix, but no matter if we fix here too

server/artifacts/artifact_server.go
alexec: if you SplitN() then you can:

```golang
fileName = requestPath[fileNameFirstIndex]
```

no need to join

juliev0: I see. And it looks like if you request n, it's "at most" n, so okay if you only have n - 1. Good to know.

alexec (approved):

PR #8552, opened 2022-04-30
fix: Polish artifact visualisation. Fixes #7743

### LINKED ISSUE #7743: Visualizations in the Argo UI
# Summary
Would like to have functionality, that can show generated visualizations (images, videos) inside a workflow in the Argo UI (like Metaflow cards)
https://docs.metaflow.org/metaflow/visualizing-results
https://towardsdatascience.com/integrating-pythonic-visual-reports-into-ml-pipelines-a163d150ed04

It could be integrated in some way with a step, or be a separate step, executed after the necessary artifacts are generated and prepare a HTML/javascript template for visualization. 
The visualizations would be available either directly on the "visualization node" in workflow details, or shown under details in the workflows overview page, where one could compare visualizations of multiple workflow runs

# Use Cases
- Comparing ML pipeline runs from generated charts  
- Visualizing end results of ML pipeline runs 
- Debugging workflows where visual artifacts are the most helpful

---

### LINKED ISSUE #8477: Artifact GC: docs
Depends on #8476. 

We should create

* A doc (e.g `docs/artifact_gc.md`) to show users how to use this feature.
* An example in `examples`.
* It's a cool feature, lets add it to the "new release modal": https://github.com/argoproj/argo-workflows/blob/e368f504307ecbbfa2f228bccfab1f943efaca9d/ui/src/app/modals/new-version/new-version-modal.tsx#L9

### LINKED ISSUE #8551: Add a link the previous runs of a workflow
I don't know why we do not have this. 

Add a link to the workflow details page saying "Previous Run". This should go to the workflow list, but only display previous runs of this workflow, i.e. same workflow-template, cron-workflow, etc.

### LINKED ISSUE #8558: Polish artifact server error response
Users would could benefit with some polish. 

The main issue is that we seem to use `http.WriteHeader` for errors, when we should use `http.Error`

```golang
w.WriteHeader(500) ;// no HTTP body to help user
```

vs

```golang
http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```

alexec:

Fixes #7743
Fixes #8477 
Fixes #8551
Fixes #8558


ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
```diff
@@ -71,7 +74,7 @@ export const ArtifactPanel = ({
                                         }}
                                     />
                                 ) : (
-                                    <iframe sandbox='' src={downloadUrl} style={{width: '100%', height: '500px', border: 'none'}} />
```
alexec, line 74: @mboussarov this is what was breaking the CSP.

sarabala1979 (approved):

PR #8555, opened 2022-04-30
fix: remove list and watch on secrets. Fixes #8534

### LINKED ISSUE #8534: Argo Server with SSO enabled should not require List on all Secrets
# Summary

When Argo Server starts up with SSO, it will call `cache.NewResourceCache()`:

```golang
func NewArgoServer(ctx context.Context, opts ArgoServerOpts) (*argoServer, error) {
	configController := config.NewController(opts.Namespace, opts.ConfigName, opts.Clients.Kubernetes)
	var resourceCache *cache.ResourceCache = nil
	ssoIf := sso.NullSSO
	if opts.AuthModes[auth.SSO] {
		c, err := configController.Get(ctx)
		if err != nil {
			return nil, err
		}
		ssoIf, err = sso.New(c.SSO, opts.Clients.Kubernetes.CoreV1().Secrets(opts.Namespace), opts.BaseHRef, opts.TLSConfig != nil)
		if err != nil {
			return nil, err
		}
		resourceCache = cache.NewResourceCache(opts.Clients.Kubernetes, ctx, getResourceCacheNamespace(opts))
		log.Info("SSO enabled")
	} else {
		log.Info("SSO disabled")
	}
```

But NewResourceCache wants the ability to list all service accounts and all secrets of the entire cluster.

```golang
func NewResourceCache(client kubernetes.Interface, ctx context.Context, namespace string) *ResourceCache {
	informerFactory := informers.NewSharedInformerFactoryWithOptions(client, time.Minute*20, informers.WithNamespace(namespace))
	cache := &ResourceCache{
		ServiceAccountLister: informerFactory.Core().V1().ServiceAccounts().Lister(),
		SecretLister:         informerFactory.Core().V1().Secrets().Lister(),
	}
	informerFactory.Start(ctx.Done())
	informerFactory.WaitForCacheSync(ctx.Done())
	return cache
}
```

There are two problems with this:

1) While list ServiceAccounts might be acceptable, listing all secrets of the cluster is generally unacceptable.
2) This does not work with an Argo Server running with `--namespaced` mode

Instead of listing all secrets, I think we could look up the secret of the ServiceAccount on-demand, to find the bearer token. This will allow administrators to only give the controller limited permissions to Secrets. If the use of a Lister was because of performance, we could cache this in-memory rather than rely on a lister.

# Use Cases

When would you use this?

In a secure, multi-tenant environment.

---

basanthjenuhb:
Signed-off-by: bjenuhb <Basanth_JenuHB@intuit.com>

Fixes https://github.com/argoproj/argo-workflows/issues/8534


server/cache/cache.go
```diff
@@ -10,17 +12,24 @@ import (
...
+		client:               client,
 		ServiceAccountLister: informerFactory.Core().V1().ServiceAccounts().Lister(),
-		SecretLister:         informerFactory.Core().V1().Secrets().Lister(),
 	}
 	informerFactory.Start(ctx.Done())
 	informerFactory.WaitForCacheSync(ctx.Done())
 	return cache
 }
+
+func (c *ResourceCache) GetSecret(namespace string, secretName string) (*v12.Secret, error) {
```
basanthjenuhb, line 32: Reason to introduce this method is if we want to introduce caching in future, we can easily change its implementation and keep external contract same

terrytangyuan (approved):

alexec (changes requested):
We added to avoid repeated API calls. Prior to this there was a rought 1-to-1 mapping between Argo Server API calls and Kubernetes API calls O(1). This is now O(2). 

Please use a cache (e.g `lru`) so that this is mitigated.

basanthjenuhb:
The problem with a cache is cache invalidation,
But got to know one info,
If the secret for that service account is deleted,
a new secret with a different name is recreated
so, since the name would be different, cache would be automatically invalidated.
Will add an LRU cache with 20 mins of refresh time

alexec:
It can be less than 20m. We only want to avoid doing 1 per call. A cache of 1m would work.

basanthjenuhb:
Just wanted to keep the timeout same as the informer
Since now cache invalidation is not an issue. 20m seems fine

jessesuen (changes requested):
There was one bug that is not addressed in this PR: `getResourceCacheNamespace()` needs to consider `--namespaced` as well as `--managed-namespace`.


As you can see, the code only considers `--namespaced` option:

```golang
func getResourceCacheNamespace(opts ArgoServerOpts) string {
	if opts.Namespaced {
		return opts.SSONameSpace
	}
	return v1.NamespaceAll
}
```

basanthjenuhb:
> There was one bug that is not addressed in this PR: `getResourceCacheNamespace()` needs to consider `--namespaced` as well as `--managed-namespace`.
> 
> As you can see, the code only considers `--namespaced` option:
> 
> ```go
> func getResourceCacheNamespace(opts ArgoServerOpts) string {
> 	if opts.Namespaced {
> 		return opts.SSONameSpace
> 	}
> 	return v1.NamespaceAll
> }
> ```

Actually that piece of code is handled here
You can specify `ssoNamespace`. To determine which namespace to lookup service accounts from
https://github.com/argoproj/argo-workflows/blob/master/cmd/argo/commands/server.go#L146-L155
```go
if namespaced {
    // Case 1: If ssoNamespace is not specified, default it to installation namespace
    if ssoNamespace == "" {
        ssoNamespace = namespace
    }
    // Case 2: If ssoNamespace is not equal to installation or managed namespace, default it to installation namespace
    if ssoNamespace != namespace && ssoNamespace != managedNamespace {
        log.Warn("--sso-namespace should be equal to --managed-namespace or the installation namespace")
        ssoNamespace = namespace
    }
}
```

basanthjenuhb:
@alexec I feel 20 min should be fine since we don't have cache invalidation issue.
Let me know if you want it to be 1 min.

alexec:
Please make it 1m. That way it will often be up-to-date, but also protects us from bursts. 

Regarding `SSONamespace`, I found this confused me a lot. Do users really want is effectively another system namespace? Could we get rid of it?

jessesuen:
> Actually that piece of code is handled here
> You can specify --sso-namespace.

I'm confused then about the usage. The use case is this:

* End user team A wants to run workflows in namespace: `foo`
* Operator team B run the workflow API server / controller for team A in a **_different namespace_**: 'foo-argo`
* Operator team B does not want to run a cluster-wide list/watch

What is the right combination of flags to achieve the above use case? I thought it would just be:

```
argo server --managed-namespace=foo
```

But I do not believe it works like that.

> Regarding SSONamespace, I found this confused me a lot. Do users really want is effectively another system namespace? 

This also confused me. Is there a scenario where it's desired that this is different than the namespace the server is running in?

basanthjenuhb:
Probably not.
But something we couldn't decide as to which namespace to be chosen when managed namespace was used in the initial design.
Will default it to the managed namespace when managed-namespace option is chosen

basanthjenuhb:
But yeah,
The right arguments would have been
```
argo server --managed-namespace=foo --sso-namespace=foo
```

But yeah, removing this for since you feel this is confusing
We will be be using this to decide sso-namespace
```
func getSSONamespace(opts ArgoServerOpts) string {
	if opts.Namespaced {
		if opts.ManagedNamespace != "" {
			return opts.ManagedNamespace
		}
		return opts.Namespace
	}
	return v1.NamespaceAll
}
```

jessesuen:
Shouldn't this be:

```golang
func getSSONamespace(opts ArgoServerOpts) string {
	if opts.ManagedNamespace != "" {
		return opts.ManagedNamespace
	} else if opts.Namespaced {
		return opts.Namespace
	}
	return v1.NamespaceAll
}
```

basanthjenuhb:
Both should return the same output
As for anyone to use `ManagedNamespace`, `Namespaced` has to be set to true based on 
https://github.com/argoproj/argo-workflows/blob/master/cmd/argo/commands/server.go#L93-L96

basanthjenuhb:
Segregated methods

```go
func getSSONamespace(opts ArgoServerOpts) string {
	if opts.ManagedNamespace != "" {
		return opts.ManagedNamespace
	}
	return opts.Namespace
}

func getResourceCacheNamespace(opts ArgoServerOpts) string {
	if opts.ManagedNamespace != "" {
		return opts.ManagedNamespace
	}
	return v1.NamespaceAll
}
```

This should be more clear now

basanthjenuhb:
Segregated methods

```go
func getSSONamespace(opts ArgoServerOpts) string {
	if opts.ManagedNamespace != "" {
		return opts.ManagedNamespace
	}
	return opts.Namespace
}

func getResourceCacheNamespace(opts ArgoServerOpts) string {
	if opts.ManagedNamespace != "" {
		return opts.ManagedNamespace
	}
	return v1.NamespaceAll
}
```

This should be more clear now

server/apiserver/argoserver.go
```diff
@@ -116,9 +115,16 @@ func init() {
 	}
 }
 
+func getSSONamespace(opts ArgoServerOpts) string {
+	if opts.ManagedNamespace != "" {
```
alexec, line 119: replace with `opts.ManagedNamespace` and then inline

server/cache/cache.go
```diff
@@ -1,26 +1,6 @@
...
-func NewResourceCache(client kubernetes.Interface, ctx context.Context, namespace string) *ResourceCache {
-	informerFactory := informers.NewSharedInformerFactoryWithOptions(client, time.Minute*20, informers.WithNamespace(namespace))
-	cache := &ResourceCache{
-		ServiceAccountLister: informerFactory.Core().V1().ServiceAccounts().Lister(),
-		SecretLister:         informerFactory.Core().V1().Secrets().Lister(),
-	}
-	informerFactory.Start(ctx.Done())
-	informerFactory.WaitForCacheSync(ctx.Done())
-	return cache
+type Cache interface {
```
alexec, line 3: If a package has a single key interface, I think it should be named `Interface`.

server/cache/cache.go
```diff
@@ -1,26 +1,6 @@
...
+	Get(key string) (any, bool)
```
alexec, line 4: `any` 👀

server/cache/lru_ttl_cache.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type lruTtlCache struct {
+	timeout time.Duration
+	cache   *lru.Cache
+}
+
+type timeValueHolder struct {
+	createTime time.Time
+	value      any
```
alexec, line 16: minor - go full generics

```golang
type item[T] struct  {
creationTime time.Time
item T
}
```

server/cache/lru_ttl_cache.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func NewLruTtlCache(timeout time.Duration, size int) *lruTtlCache {
```
alexec, line 19: `Lru` should be upper-case `LRU`

server/cache/lru_ttl_cache.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (c *lruTtlCache) Add(key string, value any) {
+	c.cache.Add(key, &timeValueHolder{
+		createTime: getCurrentTime(),
+		value:      value,
+	})
+}
+
+func getCurrentTime() time.Time {
```
alexec, line 45: minor- I think you could turn `getCurrentTime` into a `var` and then you could set it for testing. I think there might be other examples in deps.

server/cache/resource_cache.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/listers/core/v1"
+)
+
+type ResourceCache struct {
+	ctx    context.Context
+	cache  Cache
```
alexec, line 17: There are other examples in the code where we use the decorator pattern to add caching to an interface. Maybe use that?

server/cache/resource_cache.go
```diff
@@ -0,0 +1,62 @@
...
+
+func NewResourceCacheWithTimeout(client kubernetes.Interface, ctx context.Context, namespace string, timeout time.Duration) *ResourceCache {
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(client, time.Minute*20, informers.WithNamespace(namespace))
+	cache := &ResourceCache{
+		ctx:                  ctx,
+		cache:                NewLruTtlCache(timeout, 2000),
+		client:               client,
+		ServiceAccountLister: informerFactory.Core().V1().ServiceAccounts().Lister(),
+	}
+	informerFactory.Start(ctx.Done())
```
alexec, line 30: I don't think `Start` or `Run` funcs should ever appear within a `New` func. In Java, starting "threads" in a constructor is bad-practice.

server/cache/lru_ttl_cache.go
basanthjenuhb: I had done that,
The problem is there is a bug while doing go generate for mockery
https://github.com/golang/go/issues/51629
Had to revert the generics code

server/cache/cache.go
basanthjenuhb: yeah, could be anything.
could be used to store other than secrets also

server/cache/lru_ttl_cache.go
alexec: ah

server/cache/resource_cache.go
basanthjenuhb: so thing is , ResourceCache is not actually a generic cache,
It is specific for K8s objects, currently Service accounts and secrets

And `Cache` object is for generic `get` and `put`

I feel in this case simple composition makes more sense since contracts are different for `ResourceCache` and `Cache`

server/cache/resource_cache.go
basanthjenuhb: should I add a `init` method in `resource_cache` and call if after `New` ?

server/apiserver/argoserver.go
basanthjenuhb: so `opts.ManagedNamespace` will be set to `""` if its not a namespaced installation
In that case `SSONamespace` will be `opts.Namespace`

server/cache/lru_ttl_cache.go
basanthjenuhb: Done

server/cache/lru_ttl_cache.go
basanthjenuhb: hmm, not sure I understand
Can you share an example

server/cache/cache.go
basanthjenuhb: you mean `type Interface interface {`
in a file called `Interface.go` ?

server/cache/cache.go
basanthjenuhb: its just an alias for `interface{}`

server/apiserver/argoserver.go
alexec: Ah - I need

alexec: Minor - but I would have had `opts.GetSSONamespace()`

server/cache/resource_cache.go
alexec: Yes. Or Run.

jessesuen:
> so if Namespaced is set to false, there won't be a namespaced installation anyway
> Basically, I just wanted to honour ManagedNamespace only if Namespaced is true

I see. I think something should be done to improve the UX since this was something the user we worked with overlooked (and even myself). I understand now that currently `--managed-namespace` is ignored if `--namespaced` is not set. 

I would suggest one of two things to improve:
1) Presence of `--managed-namespace=foo` automatically implying that server is running in a namespace mode, going so far as setting o.Namspaced=true in the code.
2) Adding validation/usage error that `--managed-namespace=foo` and `--namespaced=false` is not a valid combination of flags.

@alexec WDYT?

server/cache/resource_cache.go
basanthjenuhb: done

server/cache/lru_ttl_cache.go
basanthjenuhb: done

server/cache/cache.go
basanthjenuhb: done

server/apiserver/argoserver.go
basanthjenuhb: I see, 
then I'll just resolve SSONamespace and pass it in opts.
and get as `opts.ssoNamespace`

server/cache/resource_cache.go
```diff
@@ -17,6 +17,7 @@ type ResourceCache struct {
 	cache  Interface
```
alexec, line 17: remove `ctx` - structs should not store `ctx`

server/cache/resource_cache.go
```diff
@@ -26,16 +27,20 @@ func NewResourceCacheWithTimeout(client kubernetes.Interface, ctx context.Contex
...
-	informerFactory.Start(ctx.Done())
-	informerFactory.WaitForCacheSync(ctx.Done())
 	return cache
 }
 
 func NewResourceCache(client kubernetes.Interface, ctx context.Context, namespace string) *ResourceCache {
 	return NewResourceCacheWithTimeout(client, ctx, namespace, time.Minute*1)
 }
 
+func (c *ResourceCache) Run() {
```
alexec, line 39: pass `stopCh` (which wil be `ctx.Done()`)

server/apiserver/argoserver.go
basanthjenuhb: done

cmd/argo/commands/server.go
```diff
@@ -152,6 +158,7 @@ See %s`, help.ArgoServer),
 				RestConfig:               config,
 				AuthModes:                modes,
 				ManagedNamespace:         managedNamespace,
+				SSONamespace:             ssoNamespace,
```
alexec, line 161: sorry - don't re-introduce this, instead add a func that determines it's value

server/cache/resource_cache.go
basanthjenuhb: done

server/cache/resource_cache.go
basanthjenuhb: done

cmd/argo/commands/server.go
basanthjenuhb: isn't it a better way to resolve `SSONamespace` like other parameters like `namespace` and `managedNamespace` in `server.go`
I am not introducing it as command line argument though.

alexec (approved):

basanthjenuhb:
@jessesuen 
Can you have another look at the PR
I think the change regarding `--managed-namespac` and `--namespaced` could be done in a separate PR and needs discussion.

basanthjenuhb:
@jessesuen Anything else needed in this PR. want to take the PR to closure

jessesuen (approved):

cmd/argo/commands/server.go
```diff
@@ -224,7 +213,6 @@ See %s`, help.ArgoServer),
 	command.Flags().StringVar(&configMap, "configmap", common.ConfigMapName, "Name of K8s configmap to retrieve workflow controller configuration")
 	command.Flags().BoolVar(&namespaced, "namespaced", false, "run as namespaced mode")
 	command.Flags().StringVar(&managedNamespace, "managed-namespace", "", "namespace that watches, default to the installation namespace")
-	command.Flags().StringVar(&ssoNamespace, "sso-namespace", "", "namespace that will be used for SSO RBAC. Defaults to installation namespace. Used only in namespaced mode")
```
agilgur5, line 227: This is a breaking change. I see it was discussed above, but it wasn't mentioned in the changelog or release notes. The PR title is also missing an exclamation point indicating breakage 😕

cmd/argo/commands/server.go
```diff
@@ -98,10 +97,16 @@ See %s`, help.ArgoServer),
 				managedNamespace = namespace
 			}
 
+			ssoNamespace := namespace
+			if managedNamespace != "" {
+				ssoNamespace = managedNamespace
+			}
```
agilgur5, line 103: this actually is different from the logic on line 151 and line 160 below. Those set the SSO namespace to the installation namespace when using a managed namespace. That makes sense, since your Argo configuration and Argo Server are in the installation namespace. Only your Workflows are in the managed namespace.

This change is actually quite confusing as a result. And it was an undocumented breaking change that caused a regression too: https://github.com/argoproj/argo-workflows/issues/9989#issuecomment-2077892941.

agilgur5 (commented):
There appear to have been some undocumented breaking changes in this PR 😕 

Those changes were unrelated to the goals and title of this PR too, which themselves were fully backward-compatible 😕

PR #8557, opened 2022-04-30
fix(Artifact GC): modified `SearchArtifact` to return `ArtifactSearchResults`. Fixes #8543

### LINKED ISSUE #8543: Artifact GC: `SearchArtifact` to return `ArtifactSearchResults`
Currently, this func returns `Artifacts`, but we need to know which node ID the artifact was from, we can't add that to `Artifact`, but we can introduce some new types:

```golang
type ArtifactSearchResults []ArtifactSearchResult

type ArtifactSearchResult struct {
Artifact
NodeID string
}
```

rohankmr414:

Fixes #8543


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1301,11 +1308,11 @@ func (w *Workflow) SearchArtifacts(q *ArtifactSearchQuery) Artifacts {
 				match = false
 			}
 			if match == true {
-				artifacts = append(artifacts, a)
+				results = append(results, ArtifactSearchResult{Artifact: a, NodeID: n.ID})
```
alexec, line 1311: test?

rohankmr414: done @alexec

alexec (approved):

PR #8575, opened 2022-05-02
feat: Send more workflow info to plugins. Fixes #8573

### LINKED ISSUE #8573: [plugin] add more workflow data in executor plugin
# Summary

[executor](https://github.com/argoproj/argo-workflows/blob/master/pkg/plugins/executor/objectmeta.go) api  only have the workflow name data and it is not enough in many usecases. so we should add more workflow meta data in executor.

# Use Cases

writing argo volocano  plugin, and when we setup the volcanojob, we want to set it's ownnerreference in the plugin code, we need workflow id, and namespace.

---

whybeyoung:
Fixes   https://github.com/argoproj/argo-workflows/issues/8573


workflow/executor/agent.go
```diff
@@ -341,7 +343,11 @@ func (ae *AgentExecutor) executeHTTPTemplateRequest(ctx context.Context, httpTem
 func (ae *AgentExecutor) executePluginTemplate(ctx context.Context, tmpl wfv1.Template, result *wfv1.NodeResult) (time.Duration, error) {
 	args := executorplugins.ExecuteTemplateArgs{
 		Workflow: &executorplugins.Workflow{
-			ObjectMeta: executorplugins.ObjectMeta{Name: ae.WorkflowName},
+			ObjectMeta: executorplugins.ObjectMeta{
+				Name:      ae.WorkflowName,
```
alexec, line 347: Test?

cmd/argoexec/commands/agent.go
```diff
@@ -103,6 +103,10 @@ func initAgentExecutor() *executor.AgentExecutor {
 	if !ok {
 		log.Fatalf("Unable to determine workflow name from environment variable %s", common.EnvVarWorkflowName)
 	}
+	workflowUid, ok := os.LookupEnv(common.EnvVarPodUID)
```
alexec, line 106: minor - but I don't think you need to error here, just ignore

alexec (changes requested):
This looks great, but does need a test.

berlinsaint:
> This looks great, but does need a test.

OK，i will add them

cmd/argoexec/commands/agent.go
alexec: this looks wrong to me, why `EnvVarPodUID` not `EnvVarWorkflowUID`, maybe shows why test is important

test/e2e/agent_test.go
```diff
@@ -61,7 +61,12 @@ spec:
 		SubmitWorkflow().
 		WaitForWorkflow(time.Minute).
 		Then().
-		ExpectWorkflow(func(t *testing.T, _ *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+		ExpectWorkflow(func(t *testing.T, meta *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
```
alexec, line 64: this seems to change the wrong file, I think you need to look at plugin_test.go

berlinsaint: > this seems to change the wrong file, I think you need to look at plugin_test.go

gitlab ci always showme error here ....  it misunderstood me

alexec: You're not testing the plugin in this test. You're testing the HTTP template. Having re-reviewed the changes, if you revert this file, I think that we probably don't need a test, because we can add an assertion into the code.

cmd/argoexec/commands/agent.go
```diff
@@ -103,6 +103,10 @@ func initAgentExecutor() *executor.AgentExecutor {
...
+	if !ok {
+		log.Warn("Unable to determine workflow Uid from environment variable %s", common.EnvVarPodUID)
+	}
```
alexec, line 109:
```suggestion
	workflowUid:= os.GetEnv(common.EnvVarPodUID)
	if workflowUid == "" {
		log.Warn("Unable to determine workflow Uid from environment variable %s", common.EnvVarPodUID)
	}
```

alexec: this would also make sure the UID is not empty

alexec:
@berlinsaint  are you still planning to complete this PR please?

PR #8582, opened 2022-05-02
feat: `argo cp` to download artifacts. Fixes #695

### LINKED ISSUE #695: Proposal: argo cp command to download artifacts
This proposal is for an argo CLI command to download artifacts from a workflow step. Given a workflow result like:

```
$ argo get artifact-passing-jg2k8 -o wide
Name:             artifact-passing-jg2k8
Namespace:        default
ServiceAccount:   default
Status:           Succeeded
Created:          Thu Jan 18 15:02:42 -0800 (9 hours ago)
Started:          Thu Jan 18 15:02:42 -0800 (9 hours ago)
Finished:         Thu Jan 18 15:03:31 -0800 (9 hours ago)
Duration:         49 seconds

STEP                       PODNAME                            DURATION  ARTIFACTS  MESSAGE
 ✔ artifact-passing-jg2k8
 ├---✔ generate-artifact   artifact-passing-jg2k8-2089725168  25s       hello-art
 └---✔ consume-artifact    artifact-passing-jg2k8-2707584000  23s
```

One could do something like:
```
$ argo cp artifact-passing-jg2k8-2089725168:hello-art /tmp/
```

The workflow YAML already has most of the information required to download the artifact (e.g. bucket name and key) as seen in the node status:
```
    artifact-passing-jg2k8-2089725168:
      finishedAt: 2018-01-18T23:03:07Z
      id: artifact-passing-jg2k8-2089725168
      name: artifact-passing-jg2k8[0].generate-artifact
      outputs:
        artifacts:
        - name: hello-art
          path: /tmp/hello_world.txt
          s3:
            accessKeySecret:
              key: accesskey
              name: argo-artifacts-minio-user
            bucket: my-bucket
            endpoint: argo-artifacts-minio-svc.default:9000
            insecure: true
            key: artifact-passing-jg2k8/artifact-passing-jg2k8-2089725168/hello-art.tgz
            secretKeySecret:
              key: secretkey
              name: argo-artifacts-minio-user
```

The issues here are that:
1. the artifact repository may not directly accessible from the client (e.g. internal minio server)
2. the secrets may not be accessible

However, we could at least support the use case for GCS and AWS, where the repository *will* be directly accessible. Regarding bucket secrets, we can either lookup the secrets using the k8s go-client, or rely on the .aws config/credentials in the user's home directory for the download.

mihirpandya-greenops:
Signed-off-by: mihirpandya-greenops <mihir@greenops.io>

Fixes [#695](https://github.com/argoproj/argo-workflows/issues/695)

Tested feature locally by running workflows with artifacts and copying artifacts to a directory using different flag combinations.

### DIFF
--- a/cmd/argo/commands/cp.go
+++ b/cmd/argo/commands/cp.go
@@ -1,0 +1,151 @@
+package commands
+
+import (
+	"crypto/tls"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands/client"
+	"github.com/argoproj/argo-workflows/v3/pkg/apiclient"
+	workflowpkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflow"
+	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+)
+
+type copyArtifactOps struct {
+	namespace    string // --namespace
+	workflowName string // --workflow-name
+	nodeId       string // --node-id
+	templateName string // --template-name
+	artifactName string // --artifact-name
+}
+
+func NewCpCommand() *cobra.Command {
+	var copyArtifactArgs copyArtifactOps
+	var argoServerArgs apiclient.ArgoServerOpts
+
+	command := &cobra.Command{
+		Use:   "cp outputDir ...",
+		Short: "copy artifacts from workflow",
+		Example: `# Copy a workflow's artifacts to a local output directory:
+
+  argo cp output-directory --workflow-name=my-wf
+
+# Copy artifacts from a specific node in a workflow to a local output directory:
+
+  argo cp output-directory --workflow-name=my-wf --node-id=my-wf-node-id-123
+`,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, apiClient := client.NewAPIClient(cmd.Context())
+			serviceClient := apiClient.NewWorkflowServiceClient()
+			namespace := client.Namespace()
+			if len(copyArtifactArgs.namespace) > 0 {
+				namespace = copyArtifactArgs.namespace
+			}
+			workflow, err := serviceClient.GetWorkflow(ctx, &workflowpkg.WorkflowGetRequest{
+				Name:      copyArtifactArgs.workflowName,
+				Namespace: namespace,
+			})
+			if err != nil {
+				return fmt.Errorf("failed to get workflow: %w", err)
+			}
+
+			copyArtifactArgs.workflowName = workflow.Name
+			artifactSearchQuery := v1alpha1.ArtifactSearchQuery{
+				ArtifactName: copyArtifactArgs.artifactName,
+				TemplateName: copyArtifactArgs.templateName,
+				NodeId:       copyArtifactArgs.nodeId,
+			}
+			artifactSearchResults := workflow.SearchArtifacts(&artifactSearchQuery)
+
+			outputDir := args[0]
+			basePath := filepath.Join(outputDir, namespace, copyArtifactArgs.workflowName)
+			err = os.MkdirAll(basePath, os.ModePerm)
+			if err != nil {
+				return fmt.Errorf("failed to create folder: %w", err)
+			}
+
+			c := makeClient(argoServerArgs)
+
+			for _, artifact := range artifactSearchResults {
+				nodeIdFolder := filepath.Join(basePath, artifact.NodeID, "outputs")
+				err = os.MkdirAll(nodeIdFolder, os.ModePerm)
+				if err != nil {
+					return fmt.Errorf("failed to create nodeId folder: %w", err)
+				}
+				key, err := artifact.GetKey()
+				if err != nil {
+					return fmt.Errorf("error getting key for artifact: %w", err)
+				}
+				err = getAndStoreArtifactData(namespace, copyArtifactArgs.workflowName, artifact.NodeID, artifact.Name, path.Base(key), nodeIdFolder, c, argoServerArgs)
+				if err != nil {
+					return fmt.Errorf("failed to get and store artifact data: %w", err)
+				}
+			}
+			return nil
+		},
+	}
+	command.Flags().StringVarP(&copyArtifactArgs.namespace, "namespace", "n", "", "namespace of workflow")
+	command.Flags().StringVar(&copyArtifactArgs.workflowName, "workflow-name", "", "name of workflow")
+	command.Flags().StringVar(&copyArtifactArgs.nodeId, "node-id", "", "id of node in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.templateName, "template-name", "", "name of template in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.artifactName, "artifact-name", "", "name of output artifact in workflow")
+
+	command.PersistentFlags().StringVarP(&argoServerArgs.URL, "argo-server", "s", os.Getenv("ARGO_SERVER"), "API server `host:port`. e.g. localhost:2746. Defaults to the ARGO_SERVER environment variable.")
+	command.PersistentFlags().StringVar(&argoServerArgs.Path, "argo-base-href", os.Getenv("ARGO_BASE_HREF"), "An path to use with HTTP client (e.g. due to BASE_HREF). Defaults to the ARGO_BASE_HREF environment variable.")
+	command.PersistentFlags().BoolVarP(&argoServerArgs.InsecureSkipVerify, "insecure-skip-verify", "k", os.Getenv("ARGO_INSECURE_SKIP_VERIFY") == "true", "If true, the Argo Server's certificate will not be checked for validity. This will make your HTTPS connections insecure. Defaults to the ARGO_INSECURE_SKIP_VERIFY environment variable.")
+
+	return command
+}
+
+func makeClient(argoServerArgs apiclient.ArgoServerOpts) *http.Client {
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: argoServerArgs.InsecureSkipVerify,
+			},
+		},
+	}
+}
+
+func getAndStoreArtifactData(namespace string, workflowName string, nodeId string, artifactName string, fileName string, nodeIdFolder string, c *http.Client, argoServerArgs apiclient.ArgoServerOpts) error {
+	authString := ""
+	if len(argoServerArgs.URL) == 0 {
+		argoServerArgs.URL = "localhost:2746"
+	} else {
+		authString = client.GetAuthString()
+	}
+	request, err := http.NewRequest("GET", argoServerArgs.GetURL()+fmt.Sprintf("/artifacts/%s/%s/%s/%s", namespace, workflowName, nodeId, artifactName), nil)
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+	if len(authString) != 0 {
+		request.Header.Set("Authorization", client.GetAuthString())
+	}
+	resp, err := c.Do(request)
+	if err != nil {
+		return fmt.Errorf("request failed with: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("request failed %s", resp.Status)
+	}
+	artifactFilePath := filepath.Join(nodeIdFolder, fileName)
+	fileWriter, err := os.Create(artifactFilePath)
+	if err != nil {
+		return fmt.Errorf("creating file failed: %w", err)
+	}
+	_, err = io.Copy(fileWriter, resp.Body)
+	if err != nil {
+		return fmt.Errorf("copying file contents failed: %w", err)
+	}
+	log.Printf("Stored artifact %s from node %s", fileName, nodeId)
+	return nil
+}
--- a/cmd/argo/commands/cp.go
+++ b/cmd/argo/commands/cp.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -23,10 +24,11 @@ type copyArtifactOpts struct {
 	nodeId       string // --node-id
 	templateName string // --template-name
 	artifactName string // --artifact-name
+	path         string // --path
 }
 
 func NewCpCommand() *cobra.Command {
-	var copyArtifactArgs copyArtifactOpts
+	var cpArtifactOpts copyArtifactOpts
 
 	command := &cobra.Command{
 		Use:   "cp my-wf output-directory ...",
@@ -48,23 +50,11 @@ func NewCpCommand() *cobra.Command {
 			workflowName := args[0]
 			outputDir := args[1]
 
-			argoServerUrl, err := cmd.Parent().Flags().GetString("argo-server")
-			argoBasePath, err := cmd.Parent().Flags().GetString("argo-base-href")
-			insecureSkipVerify, err := cmd.Parent().Flags().GetBool("insecure-skip-verify")
-			if err != nil {
-				return fmt.Errorf("not able to read flags correctly %w", err)
-			}
-			argoServerOpts := apiclient.ArgoServerOpts{
-				URL:                argoServerUrl,
-				Path:               argoBasePath,
-				InsecureSkipVerify: insecureSkipVerify,
-			}
-
 			ctx, apiClient := client.NewAPIClient(cmd.Context())
 			serviceClient := apiClient.NewWorkflowServiceClient()
 			namespace := client.Namespace()
-			if len(copyArtifactArgs.namespace) > 0 {
-				namespace = copyArtifactArgs.namespace
+			if len(cpArtifactOpts.namespace) > 0 {
+				namespace = cpArtifactOpts.namespace
 			}
 			workflow, err := serviceClient.GetWorkflow(ctx, &workflowpkg.WorkflowGetRequest{
 				Name:      workflowName,
@@ -76,32 +66,39 @@ func NewCpCommand() *cobra.Command {
 
 			workflowName = workflow.Name
 			artifactSearchQuery := v1alpha1.ArtifactSearchQuery{
-				ArtifactName: copyArtifactArgs.artifactName,
-				TemplateName: copyArtifactArgs.templateName,
-				NodeId:       copyArtifactArgs.nodeId,
+				ArtifactName: cpArtifactOpts.artifactName,
+				TemplateName: cpArtifactOpts.templateName,
+				NodeId:       cpArtifactOpts.nodeId,
 			}
 			artifactSearchResults := workflow.SearchArtifacts(&artifactSearchQuery)
 
 			c := &http.Client{
 				Transport: &http.Transport{
 					TLSClientConfig: &tls.Config{
-						InsecureSkipVerify: argoServerOpts.InsecureSkipVerify,
+						InsecureSkipVerify: client.ArgoServerOpts.InsecureSkipVerify,
 					},
 				},
 			}
 
-			basePath := filepath.Join(outputDir, namespace, workflowName)
 			for _, artifact := range artifactSearchResults {
-				nodeIdFolder := filepath.Join(basePath, artifact.NodeID, "outputs")
-				err = os.MkdirAll(nodeIdFolder, os.ModePerm)
+				customPath := filepath.Join(outputDir, cpArtifactOpts.path)
+				nodeInfo := workflow.Status.Nodes.Find(func(n v1alpha1.NodeStatus) bool { return n.ID == artifact.NodeID })
+				if nodeInfo != nil {
+					customPath = strings.Replace(customPath, "{templateName}", nodeInfo.TemplateName, 1)
+				}
+				customPath = strings.Replace(customPath, "{namespace}", namespace, 1)
+				customPath = strings.Replace(customPath, "{workflowName}", workflowName, 1)
+				customPath = strings.Replace(customPath, "{nodeId}", artifact.NodeID, 1)
+				customPath = strings.Replace(customPath, "{artifactName}", artifact.Name, 1)
+				err = os.MkdirAll(customPath, os.ModePerm)
 				if err != nil {
-					return fmt.Errorf("failed to create nodeId folder: %w", err)
+					return fmt.Errorf("failed to create folder path: %w", err)
 				}
 				key, err := artifact.GetKey()
 				if err != nil {
 					return fmt.Errorf("error getting key for artifact: %w", err)
 				}
-				err = getAndStoreArtifactData(namespace, workflowName, artifact.NodeID, artifact.Name, path.Base(key), nodeIdFolder, c, argoServerOpts)
+				err = getAndStoreArtifactData(namespace, workflowName, artifact.NodeID, artifact.Name, path.Base(key), customPath, c, client.ArgoServerOpts)
 				if err != nil {
 					return fmt.Errorf("failed to get and store artifact data: %w", err)
 				}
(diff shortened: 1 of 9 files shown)


cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+	resp, err := c.Do(request)
+	if err != nil {
+		log.Fatalf("Request failed with %s", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		log.Fatalf("Request failed %s", resp.Status)
+	}
+	contentDispositionInfo := resp.Header.Get("Content-Disposition")
+	re := regexp.MustCompile("filename=\"(.*?)\"")
```
alexec, line 168: interesting, I had not considered this, but smart! could do you `artifact.GetKey()` to achive the same?

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+	if len(matches) == 0 {
+		log.Fatal("Couldn't find filename")
+	}
+	fileName := matches[1]
+	artifactFilePath := filepath.Join(nodeIdFolder, fileName)
+	fileWriter, err := os.Create(artifactFilePath)
+	if err != nil {
+		log.Fatal(err)
+	}
+	_, err = io.Copy(fileWriter, resp.Body)
```
alexec, line 179: good stuff

test/e2e/cli_test.go
```diff
@@ -1332,6 +1332,47 @@ func (s *CLISuite) TestWorkflowTemplateRefSubmit() {
 	})
 }
 
+func (s *CLISuite) TestWorkflowCopyArtifact() {
+	s.Given().
+		Workflow("@testdata/artifact-repository-ref.yaml").
```
alexec, line 1337: I don't think you should re-use this workflow, use `basic-workflow.yaml` instead, that is less likely to change. Or create your own for the test.

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+						keysToDelete = append(keysToDelete, key)
+					}
+				}
+				for _, key := range keysToDelete {
+					delete(artifactOutput, key)
+				}
+			}
+
+			outputDir := args[0]
+			if len(outputDir) == 0 {
```
alexec, line 98: meh - I would not check this - you have to write another test

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+	}
+	command.Flags().StringVarP(&copyArtifactArgs.namespace, "namespace", "n", "", "namespace of workflow")
+	command.Flags().StringVar(&copyArtifactArgs.workflowName, "workflow-name", "", "name of workflow")
+	command.Flags().StringVar(&copyArtifactArgs.nodeId, "node-id", "", "id of node in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.templateName, "template-name", "", "name of template in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.artifactName, "artifact-name", "", "name of output artifact in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.artifactType, "artifact-type", "output", "type of artifact to copy (only supports output as of now)")
+
+	command.PersistentFlags().StringVarP(&argoServerArgs.URL, "argo-server", "s", os.Getenv("ARGO_SERVER"), "API server `host:port`. e.g. localhost:2746. Defaults to the ARGO_SERVER environment variable.")
```
alexec, line 128: is this already accessible in another variable? I think this is dupe

cmd/argo/commands/copy_artifacts.go
alexec, line 98: meh - I would not check this - you have to write another test

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+	"regexp"
+)
+
+type copyArtifactOps struct {
+	namespace    string // --namespace
+	workflowName string // --workflow-name
+	nodeId       string // --node-id
+	templateName string // --template-name
+	artifactName string // --artifact-name
+	artifactType string // --artifact-type
```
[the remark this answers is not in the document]
mihirpandya-greenops: Removed artifactType

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+		Short: "copy artifacts from workflow",
+		Example: `# Copy a workflow's artifacts to a local output directory:
+
+  argo cp output-directory --workflow-name=my-wf
+
+# Copy artifacts from a specific node in a workflow to a local output directory:
+
+  argo cp output-directory --workflow-name=my-wf --node-id=my-wf-node-id-123
+`,
+		Run: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
mihirpandya-greenops: Switched to using `RunE` & added in `SilenceUsage: true`

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+				log.Fatal(err)
+			}
+
+			copyArtifactArgs.workflowName = workflow.Name
+
+			//{<node name1>: [artifact list], <node name2>: [artifact list],...}
+			artifactOutput := make(map[string]v1alpha1.Artifacts)
+			//var artifactOutput v1alpha1.Artifacts
+			//Getting a list of Artifacts based on user input
+			if len(copyArtifactArgs.nodeId) > 0 && len(copyArtifactArgs.templateName) > 0 {
```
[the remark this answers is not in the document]
mihirpandya-greenops: Removed the "one or the other" check

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+				log.Fatal("Only one of --template-name and --node-id can be filled out")
+			}
+			var nodes v1alpha1.Nodes
```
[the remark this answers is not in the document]
mihirpandya-greenops: Switched to using SearchArtifacts instead of the original method

cmd/argo/commands/copy_artifacts.go
mihirpandya-greenops: They are duplicates from `conn.go`, the main roadblock here was that the `argoServerOpts` method that the values are added to is private, and cannot be accessed. I could make a method in `conn.go` to make the client/setup the base request if that would be smoother?

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+func getAndStoreArtifactData(namespace string, workflowName string, nodeId string, artifactName string, nodeIdFolder string, c *http.Client, argoServerArgs apiclient.ArgoServerOpts) {
+	authString := ""
+	if len(argoServerArgs.URL) == 0 {
+		argoServerArgs.URL = "localhost:2746"
+	} else {
+		authString = client.GetAuthString()
+	}
+	request, err := http.NewRequest("GET", argoServerArgs.GetURL()+fmt.Sprintf("/artifacts/%s/%s/%s/%s", namespace, workflowName, nodeId, artifactName), nil)
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
mihirpandya-greenops: Replaced all `log.Fatalf` instances with `fmt.Errorf`

cmd/argo/commands/copy_artifacts.go
mihirpandya-greenops: With `artifact.GetKey()` there is a path included (not just filename) that doesn't match the format you had mentioned in the issue. I can use this instead if it is preferred though.

cmd/argo/commands/copy_artifacts.go
```diff
@@ -0,0 +1,184 @@
...
+type copyArtifactOps struct {
+	namespace    string // --namespace
+	workflowName string // --workflow-name
+	nodeId       string // --node-id
+	templateName string // --template-name
+	artifactName string // --artifact-name
+	artifactType string // --artifact-type
+}
+
+func NewCopyArtifactCommand() *cobra.Command {
```
alexec, line 28: I'd probably call this func `NewCpCommand` and the file `cp.go`

cmd/argo/commands/copy_artifacts.go
alexec: filename is just `path.Base(key)`

alexec: but you're correct, that isn't what I said -- hmmm -- maybe we need a way to specify the file name using templating? Give the user options like kubcetl

```
-f `%nodeId/outputs/%artifactName
```

```
-f "my-file.js"
```

```
-f "%filename"

File: cmd/argo/commands/copy_artifacts.go [hunk 12]
```diff
@@ -0,0 +1,184 @@
...
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: argoServerArgs.InsecureSkipVerify,
+			},
+		},
+	}
+}
+
+func getAndStoreArtifactData(namespace string, workflowName string, nodeId string, artifactName string, nodeIdFolder string, c *http.Client, argoServerArgs apiclient.ArgoServerOpts) {
+	authString := ""
+	if len(argoServerArgs.URL) == 0 {
```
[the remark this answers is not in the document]
mihirpandya-greenops: Without these checks/defaults, I don't think a user would be able to use the cp method out of the box. Should I still remove the checks?

cmd/argo/commands/copy_artifacts.go
mihirpandya-greenops: Updated to use artifact.GetKey()

test/e2e/cli_test.go
mihirpandya-greenops: Added in a workflow (`basic-artifact-workflow.yaml`) specifically for the cp e2e test.

cmd/argo/commands/copy_artifacts.go
mihirpandya-greenops: I removed the check, what should I verify for the test?

cmd/argo/commands/copy_artifacts.go
mihirpandya-greenops: Updated naming

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+	resp, err := c.Do(request)
+	if err != nil {
+		return fmt.Errorf("request failed with: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("request failed %s", resp.Status)
+	}
+	artifactFilePath := filepath.Join(nodeIdFolder, fileName)
+	fileWriter, err := os.Create(artifactFilePath)
```
terrytangyuan, line 141: fileWriter needs to be closed

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+		Example: `# Copy a workflow's artifacts to a local output directory:
+
+  argo cp output-directory --workflow-name=my-wf
+
+# Copy artifacts from a specific node in a workflow to a local output directory:
+
+  argo cp output-directory --workflow-name=my-wf --node-id=my-wf-node-id-123
+`,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
mihirpandya-greenops: Checked len(args) & verified same form as other commands

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1285,6 +1285,9 @@ func (w *Workflow) SearchArtifacts(q *ArtifactSearchQuery) ArtifactSearchResults
 
 	for _, n := range w.Status.Nodes {
 		t := w.GetTemplateByName(n.TemplateName)
+		if n.GetOutputs() == nil {
```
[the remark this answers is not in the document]
mihirpandya-greenops: Done

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+	artifactFilePath := filepath.Join(nodeIdFolder, fileName)
+	fileWriter, err := os.Create(artifactFilePath)
+	if err != nil {
+		return fmt.Errorf("creating file failed: %w", err)
+	}
+	_, err = io.Copy(fileWriter, resp.Body)
+	if err != nil {
+		return fmt.Errorf("copying file contents failed: %w", err)
+	}
+	log.Printf("Stored artifact %s from node %s", fileName, nodeId)
```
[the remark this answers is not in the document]
mihirpandya-greenops: Done

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands/client"
+	"github.com/argoproj/argo-workflows/v3/pkg/apiclient"
+	workflowpkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflow"
+	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+)
+
```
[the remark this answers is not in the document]
mihirpandya-greenops: Updated name from Ops (type) to Opts.

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+type copyArtifactOps struct {
+	namespace    string // --namespace
+	workflowName string // --workflow-name
```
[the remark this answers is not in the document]
mihirpandya-greenops: Updated command syntax

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+			artifactSearchQuery := v1alpha1.ArtifactSearchQuery{
+				ArtifactName: copyArtifactArgs.artifactName,
+				TemplateName: copyArtifactArgs.templateName,
+				NodeId:       copyArtifactArgs.nodeId,
+			}
+			artifactSearchResults := workflow.SearchArtifacts(&artifactSearchQuery)
+
+			outputDir := args[0]
+			basePath := filepath.Join(outputDir, namespace, copyArtifactArgs.workflowName)
+			err = os.MkdirAll(basePath, os.ModePerm)
```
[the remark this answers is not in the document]
mihirpandya-greenops: Removed the extra Mkdir

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+	command.Flags().StringVar(&copyArtifactArgs.artifactName, "artifact-name", "", "name of output artifact in workflow")
+
+	command.PersistentFlags().StringVarP(&argoServerArgs.URL, "argo-server", "s", os.Getenv("ARGO_SERVER"), "API server `host:port`. e.g. localhost:2746. Defaults to the ARGO_SERVER environment variable.")
+	command.PersistentFlags().StringVar(&argoServerArgs.Path, "argo-base-href", os.Getenv("ARGO_BASE_HREF"), "An path to use with HTTP client (e.g. due to BASE_HREF). Defaults to the ARGO_BASE_HREF environment variable.")
+	command.PersistentFlags().BoolVarP(&argoServerArgs.InsecureSkipVerify, "insecure-skip-verify", "k", os.Getenv("ARGO_INSECURE_SKIP_VERIFY") == "true", "If true, the Argo Server's certificate will not be checked for validity. This will make your HTTPS connections insecure. Defaults to the ARGO_INSECURE_SKIP_VERIFY environment variable.")
+
+	return command
+}
+
+func makeClient(argoServerArgs apiclient.ArgoServerOpts) *http.Client {
```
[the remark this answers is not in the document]
mihirpandya-greenops: Inlined the client creation

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+			return nil
+		},
+	}
+	command.Flags().StringVarP(&copyArtifactArgs.namespace, "namespace", "n", "", "namespace of workflow")
+	command.Flags().StringVar(&copyArtifactArgs.workflowName, "workflow-name", "", "name of workflow")
+	command.Flags().StringVar(&copyArtifactArgs.nodeId, "node-id", "", "id of node in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.templateName, "template-name", "", "name of template in workflow")
+	command.Flags().StringVar(&copyArtifactArgs.artifactName, "artifact-name", "", "name of output artifact in workflow")
+
+	command.PersistentFlags().StringVarP(&argoServerArgs.URL, "argo-server", "s", os.Getenv("ARGO_SERVER"), "API server `host:port`. e.g. localhost:2746. Defaults to the ARGO_SERVER environment variable.")
```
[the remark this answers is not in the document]
mihirpandya-greenops: Switched to inheriting the flags from the parent command

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+				InsecureSkipVerify: argoServerArgs.InsecureSkipVerify,
+			},
+		},
+	}
+}
+
+func getAndStoreArtifactData(namespace string, workflowName string, nodeId string, artifactName string, fileName string, nodeIdFolder string, c *http.Client, argoServerArgs apiclient.ArgoServerOpts) error {
+	authString := ""
+	if len(argoServerArgs.URL) == 0 {
+		argoServerArgs.URL = "localhost:2746"
```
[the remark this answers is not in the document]
mihirpandya-greenops: Removed default setting

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,151 @@
...
+	} else {
+		authString = client.GetAuthString()
+	}
+	request, err := http.NewRequest("GET", argoServerArgs.GetURL()+fmt.Sprintf("/artifacts/%s/%s/%s/%s", namespace, workflowName, nodeId, artifactName), nil)
```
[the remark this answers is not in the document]
mihirpandya-greenops: Moved URL creation completely into `fmt.Sprintf`

cmd/argo/commands/cp.go
mihirpandya-greenops: Closed filewriter before exiting method

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1335,6 +1335,9 @@ type Outputs struct {
 }
 
 func (o *Outputs) GetArtifacts() Artifacts {
+	if o == nil {
+		return []Artifact{}
```
alexec, line 1339: minor - can be nil

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,145 @@
...
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 2 {
+				cmd.HelpFunc()(cmd, args)
+				return fmt.Errorf("incorrect number of arguments")
+			}
+			workflowName := args[0]
+			outputDir := args[1]
+
+			argoServerUrl, err := cmd.Parent().Flags().GetString("argo-server")
```
alexec, line 51: change `argoServerOpts = apiclient.ArgoServerOpts{}` to `ArgoServerOpts = apiclient.ArgoServerOpts{}` then you can access that

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,145 @@
...
+				Name:      workflowName,
+				Namespace: namespace,
+			})
+			if err != nil {
+				return fmt.Errorf("failed to get workflow: %w", err)
+			}
+
+			workflowName = workflow.Name
+			artifactSearchQuery := v1alpha1.ArtifactSearchQuery{
+				ArtifactName: copyArtifactArgs.artifactName,
```
alexec, line 79: rename `copyArtifactArgs` to `opts`

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,145 @@
...
+	request.Header.Set("Authorization", client.GetAuthString())
+	resp, err := c.Do(request)
+	if err != nil {
+		return fmt.Errorf("request failed with: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("request failed %s", resp.Status)
+	}
+	artifactFilePath := filepath.Join(nodeIdFolder, fileName)
```
alexec, line 133: The user will want control over the path. What did you think of my idea of a temptable path?

mihirpandya-greenops: By temptable path do you mean allowing a user to flatten the directory?

alexec: I'm not sure. I think as a user, I've a few ways I want to use this.

Get all artifacts from several massive workflow. I'll invoke the CLI once per workflow. 

```bash
argo cp wf-1 ./out-1
argo cp wf-2 ./out-2
```

In that case, all the files should be put into the directory in such a way they are all uniquely named. E.g. `{nodeId}/outputs/{artifactName}`

Get a single artifacts from a workflow. 

```bash
argo cp wf-1 . -1 --template=output-tmpl --filename=output.json
```

Get a few artifacts from a workflow, where each template is only used once:

```bash
argo cp wf-1 . -1 --template=output-tmpl --filename='{templateName}/{artifactName}'
```

Get a artifacts from a workflow, but each has different key (filename=basename(key):

```bash
argo cp wf-1 . -1 --template=output-tmpl --filename='{filename}'
```

Any I missed?

mihirpandya-greenops: I see what you're saying. I think given the direction you're talking about, it might make sense to fully commit to exhaustive customizability for path construction to keep things as simple as possible for users (would probably be easier for a user to understand than case-by-case).

We could have 4 main variables - `workflowName`, `nodeId`, `templateName`, `artifactName`. A user could use the --filename (although if we're using it as a path, --path would make more sense) to define a path structure. Eg:

```
--path=`{templateName}/{artifactName}`
--path=`{workflowName}/{nodeId}/{artifactName}`
--path=`{workflowName}`
etc...
```
This would allow a user to create a fully customized path based on whatever their requirements are. The default would be `{workflowName}/{nodeId}/outputs/{artifactName}`.

All the other filter flags would still be honored as the path would only be made for artifacts that are being copied. The artifact file would be stored in whatever the constructed folder is.

Do you think this might be a good approach for customizable paths?

alexec: You’re missing one variable `filename`, which is `basename(key)`. But yes, and `—path` does seem more correct.

alexec (changes requested):
This is almost ready:

* You need to remove the copy-and-paste of the `ArgoServerOpts`. Left unfixed will create bugs in the future.
* Add the `--path` option.

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,143 @@
...
+type copyArtifactOpts struct {
+	namespace    string // --namespace
+	nodeId       string // --node-id
+	templateName string // --template-name
+	artifactName string // --artifact-name
+	path         string // --path
+}
+
+func NewCpCommand() *cobra.Command {
+	var cpArtifactOpts copyArtifactOpts
```
alexec, line 31: minor - but I think you could just inline all you options rather than using a struct - code would be simpler

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,143 @@
...
+					TLSClientConfig: &tls.Config{
+						InsecureSkipVerify: client.ArgoServerOpts.InsecureSkipVerify,
+					},
+				},
+			}
+
+			for _, artifact := range artifactSearchResults {
+				customPath := filepath.Join(outputDir, cpArtifactOpts.path)
+				nodeInfo := workflow.Status.Nodes.Find(func(n v1alpha1.NodeStatus) bool { return n.ID == artifact.NodeID })
+				if nodeInfo != nil {
```
alexec, line 86: if `nodeInfo == nil` that's an error surely?

cmd/argo/commands/cp.go
```diff
@@ -0,0 +1,143 @@
...
+	fileWriter, err := os.Create(artifactFilePath)
+	if err != nil {
+		return fmt.Errorf("creating file failed: %w", err)
+	}
+	defer fileWriter.Close()
+	_, err = io.Copy(fileWriter, resp.Body)
+	if err != nil {
+		return fmt.Errorf("copying file contents failed: %w", err)
+	}
+	log.Printf("Stored artifact %s", fileName)
```
alexec, line 141: Can you change this to `"Created %q`, filename"

cmd/argo/commands/cp.go
mihirpandya-greenops: That's true - that would mean the user entered the wrong template name. I'll update this.

cmd/argo/commands/cp.go
mihirpandya-greenops: Updated

cmd/argo/commands/cp.go
mihirpandya-greenops: Updated to throw an error if nodeInfo is `nil`

cmd/argo/commands/cp.go
mihirpandya-greenops: replaced cpArtifactOpts with direct variables

alexec (approved):
🚀

PR #8596, opened 2022-05-03
feat: Date range filter for workflow list. Fixes #8329

### LINKED ISSUE #8329: Date range filter for workflow list in UI
# Summary

Users want to be able to find workflows by time. Add a filter to the UI for date range (same as archived workflows).

Note that the API does not support time filters, and cannot, because it essentially proxies to the Kubernetes API, which does not support it.

That means we need to warn users that maybe need to increase the number of records per page..

This feature will never be perfect, but would be useful.

momom-i:
Fixes #8329

Add a date range filter for workflow

Now if a user has more than 20 workflows, show them a pop-up window that warns it may takes time.
But this number is a kind of temporary value so I'd appreciate if you could let me know more appropriate value.


ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -75,6 +77,8 @@ export class WorkflowsList extends BasePage<RouteComponentProps<any>, State> {
                 params.append('label', label);
             });
         }
+        params.append('minStartedAt', this.state.minStartedAt.toISOString());
```
alexec, line 80: I don't see how this can work. This is not supported my `minStartedAt`.

momom-i: I forgot to remove them.

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -194,12 +209,42 @@ export class WorkflowsList extends BasePage<RouteComponentProps<any>, State> {
...
+    private fetchWorkflows(namespace: string, selectedPhases: WorkflowPhase[], selectedLabels: string[], minStartedAt: Date, maxStartedAt: Date, pagination: Pagination): void {
         if (this.listWatch) {
             this.listWatch.stop();
         }
         this.listWatch = new ListWatch(
-            () => services.workflows.list(namespace, selectedPhases, selectedLabels, pagination),
+            () =>
+                services.workflows.list(namespace, selectedPhases, selectedLabels, pagination).then(x => {
+                    if (x.items && x.items.length > 20) {
+                        this.appContext.apis.notifications.show({
```
alexec, line 240: I don’t think you need to do this. The filter looks pretty fast to me.

momom-i: I understand.

alexec (changes requested):
I've run this locally and it seems to work. You've changed the defaults. By default, all workflows are show. Now only the last month are shown. I like this new behaviour. But, we need to make sure the user knows.

That can be archived using the new `<FirstTimeUserPanel/>`. This panel is inspired by the same UI element that iPhone shows after upgrades:

image

This element replaces the element it explains, and once dismissed, is never shown again.

image

This panel needs some small improvements:

* Center the text.
* Allow the user to add a margin.

E.g.

```typescript
export const FirstTimeUserPanel = ({id, explanation, children ,style}: {id: string; explanation: string; children: ReactChildren, style: CSSProperties}) => {
    const localStorageKey = 'FirstTimeUserPanel/' + id;
    const [dismissed, setDismissed] = useState(localStorage.getItem(localStorageKey) === 'true');

    useEffect(() => {
        localStorage.setItem(localStorageKey, String(dismissed));
    }, [dismissed]);

    if (!dismissed) {
        return (
            <div className='white-box' style={{textAlign: 'center', ...style }}>
                <p>{explanation}</p>
                <p>
                    <a onClick={() => setDismissed(true)}>Continue</a>
                </p>
            </div>
        );
    }
    return <>{children}</>;
};
```

The user may wish to quickly clear the dates, so can we add an "X" to the date-picker that clears the date?

momom-i:
I'm sorry for the delay. Now it looks like as follows.

- Add clear button
スクリーンショット 2022-05-08 22 14 30

- Add `FirstTimeUserPanel`
スクリーンショット 2022-05-08 22 15 08

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -153,53 +165,92 @@ export class WorkflowsList extends BasePage<RouteComponentProps<any>, State> {
...
-                        <SlidingPanel isShown={!!this.sidePanel} onClose={() => ctx.navigation.goto('.', {sidePanel: null})}>
-                            {this.sidePanel === 'submit-new-workflow' && (
-                                <WorkflowCreator
-                                    namespace={Utils.getNamespaceWithDefault(this.state.namespace)}
-                                    onCreate={wf => ctx.navigation.goto(uiUrl(`workflows/${wf.metadata.namespace}/${wf.metadata.name}`))}
+                        <FirstTimeUserPanel
+                            id='WorkflowListPanel'
+                            explanation={
+                                'This page shows a list of your workflows. You can filter by namespace, label and now date and time. ' +
+                                'By default, only the last months workflows are shown.'
```
alexec, line 172: minor - should be "month's" not "months".

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -153,53 +165,92 @@ export class WorkflowsList extends BasePage<RouteComponentProps<any>, State> {
...
-                        </div>
-                        <SlidingPanel isShown={!!this.sidePanel} onClose={() => ctx.navigation.goto('.', {sidePanel: null})}>
-                            {this.sidePanel === 'submit-new-workflow' && (
-                                <WorkflowCreator
-                                    namespace={Utils.getNamespaceWithDefault(this.state.namespace)}
-                                    onCreate={wf => ctx.navigation.goto(uiUrl(`workflows/${wf.metadata.namespace}/${wf.metadata.name}`))}
+                        <FirstTimeUserPanel
+                            id='WorkflowListPanel'
+                            explanation={
+                                'This page shows a list of your workflows. You can filter by namespace, label and now date and time. ' +
```
alexec, line 171: minor - delete "now"

alexec: to make this sentance scan, "and date/time"

alexec (approved):

PR #8614, opened 2022-05-04
test: e2e test support for artifacts updated. Fixes #8521.

### LINKED ISSUE #8521: Artifact GC: e2e test support
Currently, we don't have a way in our e2e tests to understand if an artifact has been deleted.

https://github.com/argoproj/argo-workflows/blob/d7bcaa7569ac15d85eb293a72a1a98779275bd6e/test/e2e/fixtures/then.go#L187

This checks the workflow, but does not check the storage, which is actually MinIO.

We need to make changes to this func:

* Allow it to return an error if the artifact is not found, e.g. by changing to ` ExpectArtifact(nodeName string, artifactName string, f func(t *testing.T, object *minio.Object, err error))`.
* Use `minio.New` to connect to Minio. Username is `admin`, password is `password.

dpadhiar:

Fixes #8521

### DIFF
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -11,6 +11,8 @@ import (
 	"github.com/stretchr/testify/suite"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/minio/minio-go/v7"
+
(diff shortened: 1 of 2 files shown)


alexec (approved):

alexec:
LGTM 🚀

PR #8620, opened 2022-05-05
fix: Terminate, rather than delete, deadlined pods. Fixes #8545

### LINKED ISSUE #8545: containers in containersets not appropriately reporting status if terminated
## Checklist
* [X] Double-checked my configuration.
* [X] Tested using the latest version.
* [X] Used the Emissary executor.

## Summary
When terminating a workflow due to a deadline, I expect that the workflow is terminated, yet we are still waiting for containerset containers to be terminated, even though they are finished with error in k8s.

An image of the workflow after timeout termination:
Screen Shot 2022-04-29 at 7 05 21 PM

An image of the containers for shard-13 in k8s:
Screen Shot 2022-04-29 at 7 07 00 PM

What version are you running?
3.3.2

# Reproducible Workflow
[reproducible-workflow.txt](https://github.com/argoproj/argo-workflows/files/8592860/reproducible-workflow.txt)

[controller-logs.txt](https://github.com/argoproj/argo-workflows/files/8592827/controller-logs.txt)

```apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

# Logs from in your workflow's wait container:
[wait-logs.txt](https://github.com/argoproj/argo-workflows/files/8592829/wait-logs.txt)

---

alexec:

Fixes #8545

Problem: When a step hits it deadline, we delete the pod. Because we delete the pod, we loose all information about the success or failure of the pod. Instead of deleting, we must SIGTERM the pod's containers.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -468,24 +468,6 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	err := func() error {
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
-		case shutdownPod:
-			// to shutdown a pod, we signal the wait container to terminate, the wait container in turn will
-			// kill the main container (using whatever mechanism the executor uses), and will then exit itself
-			// once the main container exited
-			pod, err := wfc.getPod(namespace, podName)
-			if pod == nil || err != nil {
-				return err
-			}
-			for _, c := range pod.Spec.Containers {
-				if c.Name == common.WaitContainerName {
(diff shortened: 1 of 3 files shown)


workflow/controller/controller.go
```diff
@@ -468,24 +468,6 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	err := func() error {
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
-		case shutdownPod:
```
alexec, line 471: this isn't needed anymore

sarabala1979 (approved):
LGTM, Will existing Test cover this case?

PR #8630, opened 2022-05-05
docs: Add link to Argo LinkedIn page

terrytangyuan:

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -239,3 +239,4 @@ nav:
   - Blog: https://blog.argoproj.io/
   - Slack: https://argoproj.github.io/community/join-slack
   - Twitter: https://twitter.com/argoproj
+  - LinkedIn: https://www.linkedin.com/company/argoproj/


terrytangyuan:
The LinkedIn page is something we are using officially for marketing. I believe it's managed by Argo SIG Marketing team.

hblixt (approved):
The LinkedIn ArgoProj page/account is managed by the Argo SIG marketing committee and posting there follows the same guidelines as for our other social accounts (e.g. twitter) with the goal of spreading information by and for the project. 
Vendor pitches/marketing are explicitly not allowed, so adding the link should be fine.

alexec:
OK. That was my 1st objection addressed. My second concern is that it clutters docs the toolbar with a link most users won't want.

alexec:
image

hblixt:
> image

Can you group Blog/Slack/Twitter/LinkedIn under a "Social" drop-down?

alexec:
Today, if we want it, I'd leave it where it is. Putting it in the nav will drive traffic there. It'll do so at the cost of other options. I'm not sure what the benefit is.

terrytangyuan:
The nav bar indeed looks pretty long. Perhaps we can merge user/operator/developer guide into one dropdown for guide. Then another dropdown for socials like Slack/Twitter/LinkedIn.

alexec (approved):

Issue #8683, opened 2022-05-09
Workflows "Started" / "Finished" should be showed in detailed date and time


tgruenert:
# Summary

STARTED and FINISHED information should contain date and time. This should be done at 
* Workflows view
* Archive view
* detail view for Workflows

# Use Cases

We use CronWorkflows a lot and with a schedule trigger more than 10 times a day. After one day in lists you can´t see which one runs fine and which one has failed. Every workflow has the same information like "Start time = 1d ago" (or one 1h ago and so one). 
The label "Start time" point to the fact a user might expect - a concrete time.

![image](https://user-images.githubusercontent.com/6465773/167359933-32587aaa-b8b0-4f71-8268-983021fa1536.png)

The list is used by our developers to check out the success of our recurring processing.

Also the detail view could not be helpful at this point.

![image](https://user-images.githubusercontent.com/6465773/167360220-605b3189-5d14-4e25-b525-ff6c2384ec1f.png)

In my opinion this is relevant for all other processes done more than one time a day with the same (generated) name.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

Acarlso:
> When you hover over "2d ago", you should see a detailed timestamp.

This is an unnecessary and frustrating step when the detailed timestamp could be displayed instead of the diff.  How long ago something occurred obfuscates and inherently adds less value than when something occurred.

tooptoop4:
close since https://github.com/argoproj/argo-workflows/pull/13284 is merged?

PR #8686, opened 2022-05-09
feat: Increased default significant figures in formatDuration. Fixes #8650

### LINKED ISSUE #8650: show higher resolution of time for duration > 1 hour
# Summary

Show higher resolution of time when duration is > 1 hour (show minutes or hours and minutes). This applies everywhere duration is shown e.g. the "Workflows" view, "Workflow Details" view, and stage "Summary" view.

# Use Cases

I frequently use Argo Workflows to understand how long jobs take and how that changes over time as data and code changes. Because duration may only show resolution of one hour, there are ambiguous cases. For example, suppose workflow A shows a duration of 1 hour and workflow B shows a duration of 2 hours. I cannot tell if the real durations are
A = 1.000
B = 2.000
in which case B took twice as long, or

A = 1.499
B = 1.511
in which case they took nearly the same time.

ibuder @ Intuit

---

ibuder:
Fixes #8650

This increases the default number of significant figures when formatting durations.

## Testing

Before:
Screen Shot 2022-05-09 at 5 55 15 PM

After: 
Screen Shot 2022-05-09 at 5 31 23 PM

Signed-off-by: Immanuel Buder <immanuel_buder@intuit.com>

### DIFF
--- a/ui/src/app/shared/duration.ts
+++ b/ui/src/app/shared/duration.ts
@@ -3,10 +3,12 @@ import * as models from '../../models';
 /**
  * Format the given number number of seconds in the form _d_h_m_s.
  * @param seconds Number of seconds to format. Will be rounded to the nearest whole number.
- * @param sigfigs Level of significant figures to show
+ * @param sigfigs Level of significant figures to show.
+ *  Here, a sigfig is one of days, hours, minutes, or seconds.
+ *  Examples: 13h has one sigfig and 13h22s has two sigfigs.
  */
 
-export function formatDuration(seconds: number, sigfigs = 1) {
+export function formatDuration(seconds: number, sigfigs = 2) {
     let remainingSeconds = Math.abs(Math.round(seconds));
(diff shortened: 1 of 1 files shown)


ibuder:
@alexec The lint check failed, but on .go files I didn't change. What should I do in this case?

ibuder:
@alexec @tgruenert I'm not convinced this fixes #8683 . This does not add a "concrete time" as requested. I think the increased accuracy added in this PR will help with the use cases mentioned in #8683, but I think we should leave that issue open to add the absolute start and end times in another PR. What do you think?

tgruenert:
I agree to @ibuder. Detailed runtimes are nice but not ment in #8683. Before i wrote #8683 i found #8650 and decide that theese issues describe two complete different requirements.
So finally this PR don´t resolve #8683. A separate PR would be really nice!

alexec (approved):

PR #8690, opened 2022-05-09
fix: Executor signal handling

### LINKED ISSUE #8680: OOM error not caught by the `emissary` executor forcing the workflow to hang in "Running" state
_I am opening a new issue, but you can check https://github.com/argoproj/argo-workflows/issues/8456#issuecomment-1120206141 for context._

---

The below error has been reproduced on `master` (07/05/2022), `3.3.5` and `3.2.11`.

When a workflow get OOM killed by K8s, the `emissary` executor sometime **can't detect it**. In consequence, the workflow **hangs in "Running" state forever**.

The error **does not happen** when using the `pns` or `docker` executor. This is a major regression for us since the previous executors were working just fine. For now, we are falling back to `docker`.

I have been able to make Argo detect the killed process by manually sshing to the pod and killing the `/var/run/argo/argoexec emissary -- bash --login /argo/staging/script` process (sending a `SIGTERM` signal). When doing that the `main` container get immediately killed, as well as the workflow. The workflow is correctly marked as failed with the correct `OOMKilled (exit code 137)` error (the same error when using the `pns` and `docker` executors).

**Unfortunately**, so far all my attempts to reproduce it using openly available code, images and packages has been **unsuccessful** (I'll keep trying). I can only reproduce it using our private internal stack and images.

The workload is a deeply nested machine learning code that rely a lot on the python and pytorch `multiprocessing` and `distributed` module. My guess is that some zombie child processes prevent the argo executor or workflow controller to detect the `main` container as completed.

I will be happy to provide you more information, logs or config if it can help you to make sense of this (while on my side I'll keep trying to make a workflow reproducing the bug that I can share).

While this bug affects us, I am quite confident other people running ML workload with Python on Argo will get that bug at some point.

### LINKED ISSUE #8687: Wait container is terminated with wrong command
In v3.4 we still try to terminate using `/bin/sh` but we need to introduce `argoexec kill` instead.

https://github.com/argoproj-labs/argo-dataflow/blob/8fbb255c01e744b6331faa269c96d8558b71e15a/kill/kill.go#L21

This will work for all cases now.

`signalContainers` will exit on the first error running `SignalContainer`. It should instead continue or error, otherwise `killContainers` will never be invoked.

### LINKED ISSUE #8699: Wait container does not handle SIGTERM correctly
The wait container should trap SIGTERM and gracefully shutdown, capturing artifacts.

However, if it recieves SIGTERM while capture artifacts, it will just exit.

alexec:
This PR fixes problems with signal handling.

* Previously, if you SIGTERM the wait container during output capture, outputs would be lost and container would be failed.
* The controller was using a `sh -c kill`, but `sh` has been removed in v3.4.

See #8680
Fixes #8687 
Fixes #8699


cmd/argoexec/commands/emissary.go
```diff
@@ -127,21 +127,13 @@ func NewEmissaryCommand() *cobra.Command {
...
-				if combined != nil {
-					combined.Close()
-				}
-				command, stdout, combined, err = createCommand(name, args, template)
+				command, stdout, combined, err := createCommand(name, args, template)
 				if err != nil {
 					return fmt.Errorf("failed to create command: %w", err)
 				}
-
+				defer stdout.Close()
```
alexec, line 135: this was a bug, should close output

cmd/argoexec/commands/kill.go
```diff
@@ -0,0 +1,40 @@
...
+			if err != nil {
+				return err
+			}
+			pid, err := strconv.Atoi(args[1])
+			if err != nil {
+				return err
+			}
+			if err != nil {
+				return err
+			}
```
tczhao, line 27: duplicated with line 22

alexec:
@sarabala1979 bump! I think this will fix #8869 too.

.github/workflows/ci-build.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
   tests:
     name: Unit Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 8
+    # timeout-minutes: 8
```
alexec, line 20: revert

alexec:
I'm going to break this up into smaller PRs.

agilgur5:
Replaced by #8908

PR #8750, opened 2022-05-13
docs: user add h2o

tczhao:
sort user
add h2o


terrytangyuan (approved):
Congrats on the new gig!

PR #8779, opened 2022-05-15
feat: sanitize config links

### LINKED ISSUE #8644: Fix issues found in CNCF security engagement
The report has not been finalized. This looks like the following need to be fixed. 

- [ ] #8645

### LINKED ISSUE #8645: Validate config-map links do not contain XSS
We need to validate all links on Argo Server start-up to ensure they do not contain XSS:

- [ ] ADA-ARGO-SA-03: XSS in Argo Workflows chat url
- [ ] ADA-ARGO-SA-04: XSS in Workflows workflow link
- [ ] ADA-ARGO-SA-05: XSS in Pod Logs link
- [ ] ADA-ARGO-SA-06: XSS in sensor logs link
- [ ] ADA-ARGO-SA-07: XSS in Event Source logs url
- [ ] ADA-ARGO-SA-08: XSS in Pod link url

I think all links are set here:

https://github.com/argoproj/argo-workflows/blob/e6e0c9bb3b923a6d977875cbbd2744b8bacfce15/config/config.go#L71

Add `Config.Validate()` that checks all links are in-fact valid URLs (e.g. using `url.Parse)` so they cannot contain `javascript:` . Call `Validate()` when Argo Server starts-up.

tczhao:
Fixes #8644 
Fixes #8645


config/config.go
```diff
@@ -130,6 +133,35 @@ func (c Config) GetPodGCDeleteDelayDuration() time.Duration {
 	return c.PodGCDeleteDelayDuration.Duration
 }
 
+func (c Config) Validate(input_url string) error {
+	u, err := url.Parse(input_url)
+	if err != nil {
+		return err
+	}
+	if u.Scheme == "javascript" {
```
alexec, line 141: Could you use an allow list? Ie only allow http and https protocol

config/config.go
```diff
@@ -130,6 +133,35 @@ func (c Config) GetPodGCDeleteDelayDuration() time.Duration {
...
+		return fmt.Errorf("detect javascript link: %s", input_url)
+	}
+	return nil
+}
+
+func (c Config) HTMLEscape(unescaped string) string {
+	return html.EscapeString(unescaped)
+}
+
+func (c *Config) Sanitize() {
```
alexec, line 151: I think we should report an error on invalid links

config/config.go
```diff
@@ -130,6 +133,35 @@ func (c Config) GetPodGCDeleteDelayDuration() time.Duration {
 	return c.PodGCDeleteDelayDuration.Duration
 }
 
+func (c Config) Validate(input_url string) error {
```
alexec, line 136: This signature should not have an parameters

config/config.go
alexec: This list should be configurable. Some user might want to use JavaScript.

alexec (commented):
Looks good so far!

config/config.go
tczhao: do you mean the argo server should log and exit 
instead of just log
when encountering link validation error?

alexec: correct, error and exit

config/config.go
alexec: I'd use an env var as simpler

alexec (approved):

PR #8796, opened 2022-05-18
feat: remove size limit of 128kb for workflow templates. Fixes #8789

### LINKED ISSUE #8789: Add max template size environment variable
Currently, the max workflow template size is hardcoded and not configurable. We should introduce the `MAX_TEMPLATE_SIZE` environment variable so this may be changed since it is possible for a template to be below 128KB and be well over after unmarshaling the entire workflow spec. If the variable is not set, the max size should default to 128KB.

https://github.com/argoproj/argo-workflows/blob/39fbdb2a551482c5ae2860fd266695c0113cb7b7/workflow/controller/workflowpod.go#L268-L273

dpadhiar:

Fixes #8789

### DIFF
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -339,29 +339,6 @@ func (s *FunctionalSuite) TestEventOnWorkflowSuccess() {
 		)
 }
 
-func (s *FunctionalSuite) TestLargeWorkflowFailure() {
-	var uid types.UID
-	s.Given().
-		Workflow("@expectedfailures/large-workflow.yaml").
-		When().
-		SubmitWorkflow().
-		WaitForWorkflow(120*time.Second).
-		Then().
-		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):

jszendre:
@blkperl FYI

blkperl:
@jszendre There is still a bash/kernel limit on the size. Debian had a higher limit than Alpine so I added the helpful error message in the first place so the workflow would fail earlier.

jszendre:
I'm downstream of argo by two teams but this change was breaking several of our pipeilnes. For whatever reason the images we were using could tolerate workflows > 128kb

dpadhiar:
@jszendre The workflows that are over 128kb are now breaking the pipeline but previously they didn't?

cc @sarabala1979

dpadhiar:
@jszendre Hi, what version of Argo Workflows are these pipelines using? cc @sarabala1979

lkad:
@alexec  hello  you can see this . this  https://github.com/argoproj/argo-workflows/pull/8169  's limit workflows to 128KB and return a friendly error message  code and test  unit  has been delete .

PR #8808, opened 2022-05-19
feat: Add jobs template. Fixes #8786

### LINKED ISSUE #8786: Job template
# Summary

Workflows is used by some as a CI solution. Yet it is really needs a new template type to be comparable to GitHub Actions. 

The template type could be called “job”. A job runs a series of steps. Each step runs a command (including a shell script) and the job fails if the command fails. Steps execute in sequence. The job fails if any step fails. Steps execute conditionally, usually only if the job is passing. 

How’s this different to a steps template? Each step of a job runs in the same container. This makes storing and restoring build and caches faster and simpler. 

It also allows users to visualise steps within a container. 

Code changes will be similar to adding the container-set template. 

# Use Cases

When would you use this?

---

alexec:

Fixes #8786


agilgur5:
I think this was replaced by #10517?

PR #8811, opened 2022-05-19
feat: Add priority field to workflow steps and DAG tasks. Fixes #1092.

### LINKED ISSUE #1092: Task level priorities.
**Is this a BUG REPORT or FEATURE REQUEST?**: FEATURE REQUEST

**What happened**:
Implement priorities (a.k.a. weights) when dealing with tasks collectively consumes more resource than the underlying system can handle. A typical use case would be to set a (global| dag) parallelism and prioritise tasks that need to be scheduled first. When we have long running tasks mixed with short tasks, this can improve overall execution time.

Apache Airflow captured the concept by `priority_weight` and `weight_rule` variables where each task can accumulate or discard weights from its downstream dependencies. [link](https://airflow.apache.org/code.html#baseoperator)

A key difference from k8s pod priority and preemption is that k8s does not have the dependency information so that weight_rule calculation is not possible.

dpadhiar:
Fixes #1092


tczhao:
I think it'd be good to add an example to the `examples` folder

dpadhiar:
> I think it'd be good to add an example to the `examples` folder

Will do that tomorrow as we as add an example to this PR.

workflow/controller/steps.go
```diff
@@ -222,6 +223,22 @@ func (woc *wfOperationCtx) executeStepGroup(ctx context.Context, stepGroup []wfv
 	// The template scope of this step group.
 	stepTemplateScope := stepsCtx.tmplCtx.GetTemplateScope()
 
+	// Sort steps by priority
+	sort.SliceStable(stepGroup, func(i, j int) bool {
```
tczhao, line 227: Perhaps we can wrap this as a function to `util/sorting`

workflow/controller/steps_test.go
```diff
@@ -314,3 +314,47 @@ func TestOptionalArgumentAndParameter(t *testing.T) {
...
+      image: alpine:3.7
+      command: [sh, -c, "exit 0"]
+     
+  - name: fail
+    container:
+      image: alpine:3.7
+      command: [sh, -c, "exit 1"]
+`
+
+func TestStepsWithPriority(t *testing.T) {
```
tczhao, line 347: I think we also need a test for the dag counterpart

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1384,6 +1384,9 @@ type WorkflowStep struct {
 	// Hooks holds the lifecycle hook which is invoked at lifecycle of
 	// step, irrespective of the success, failure, or error status of the primary step
 	Hooks LifecycleHooks `json:"hooks,omitempty" protobuf:"bytes,12,opt,name=hooks"`
+
+	// Priority is used if controller is configured to process limited number of workflows in parallel. Steps with higher priority are processed first.
```
tczhao, line 1388: I think the description here is not quite right,
the priority tag at `step/task` is different to `priority` at the `workflowspec`

alexec: @tczhao is correct. This should be consistent. Conventionally, N is higher priority than N+1 (i.e. 1 is higher than 2). Which is different to precedence. "Steps with higher priority are processed first." confused me, and made me wonder if it was the reverse. If you follow the convention, delete "Steps with higher priority are processed first.".

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1454,6 +1457,10 @@ func (step *WorkflowStep) GetTemplateRef() *TemplateRef {
 	return step.TemplateRef
 }
 
+func (step *WorkflowStep) GetPriority() *int32 {
+	return step.Priority
```
alexec, line 1461: why not return "0" when nil? i.e. return `int` not `*int32`

pkg/apis/workflow/v1alpha1/workflow_types.go
alexec: I think this field should be `int32` not `*int32` as it looks like 0 should be the default.

workflow/controller/dag.go
```diff
@@ -651,22 +651,47 @@ func (woc *wfOperationCtx) resolveDependencyReferences(dagCtx *dagContext, task
...
-		if isLeaf {
-			leafTaskNames = append(leafTaskNames, taskName)
+	for _, task := range tasks {
+		if taskIsLeaf[task.Name] {
+			leafTaskNames = append(leafTaskNames, task.Name)
 		}
 	}
-	sort.Strings(leafTaskNames) // execute tasks in a predictable order
+
+	if !prioritySort {
```
alexec, line 692: You should to sort by priority first and then by name:

```golang
var taskSorter = func(i, j int) bool {
 		a := tasks[i].GetPriority()
                b := tasks[j].GetPriority()

               if a < b {return true}

              return strings.Compare(tasks[i].Name, tasks[j]Name) > 0
 	}


sort.Slice(tasks, taskSorter)
```

If you put the sorter into a global, you can write a unit test.

workflow/controller/steps.go
alexec: Put into a shared func and re-use.

alexec (changes requested):

pkg/apis/workflow/v1alpha1/workflow_types.go
tczhao: I think the higher priority part is correct. It aligns with how kubernetes prioirtyClass is defined. (The pods with higher priority get scheduled first. e.g. system-cluster-critical class has 2000000000 priority value).

dpadhiar: To keep in line with k8s priorityClass, I will keep the convention that higher priority = higher value. I will update the comments to accurately reflect that.

alexec:
Still needs changes.

workflow/controller/dag.go
```diff
@@ -651,22 +651,42 @@ func (woc *wfOperationCtx) resolveDependencyReferences(dagCtx *dagContext, task
 // findLeafTaskNames finds the names of all tasks whom no other nodes depend on.
 // This list of tasks is used as the the default list of targets when dag.targets is omitted.
 func (d *dagContext) findLeafTaskNames(tasks []wfv1.DAGTask) []string {
+
+	prioritySort := false
+
+	// Sort tasks by priority
+	sort.Slice(tasks, func(i, j int) bool {
```
tczhao, line 658: With Alex's comment, maybe we should put these into global util?
https://go.dev/doc/tutorial/generics

workflow/controller/steps_test.go
```diff
@@ -314,3 +314,47 @@ func TestOptionalArgumentAndParameter(t *testing.T) {
...
+	wfcset := controller.wfclientset.ArgoprojV1alpha1().Workflows("")
+
+	ctx := context.Background()
+	wf := wfv1.MustUnmarshalWorkflow(stepsWithPriority)
+	wf, err := wfcset.Create(ctx, wf, metav1.CreateOptions{})
+	assert.NoError(t, err)
+	woc := newWorkflowOperationCtx(wf, controller)
+
+	woc.operate(ctx)
+	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
```
tczhao, line 359: I think you need additional assert here to check `StartedAt` for the `fail` step is earlier than the `succeed` step

tczhao (changes requested):

workflow/controller/dag.go
dpadhiar: Will put the sort function into a generic so that it works with both tasks and steps.

dpadhiar:
```
func SortByPriority[T []wfv1.WorkflowStep | []wfv1.DAGTask](m T) {
	sort.Slice(m, func(i, j int) bool {

		if m[i].GetPriority() > m[j].GetPriority() {
			return true
		}

		return strings.Compare(m[i].Name, m[j].Name) > 0
	})
}
```

@tczhao This is the generic function I am trying to use to sort both steps and tasks however I keep getting an invalid operation syntax error when trying to index m. 

`cannot index m (variable of type T constrained by []"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1".WorkflowStep|[]"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1".DAGTask)`

workflow/controller/dag_test.go
```diff
@@ -3448,3 +3448,178 @@ func TestDagHttpChildrenAssigned(t *testing.T) {
...
+      startedAt: "2022-07-18T22:45:27Z"
+      templateName: intentional-fail
+      templateScope: local/dag-contiue-on-fail-987nb
+      type: Pod
+  phase: Failed
+  progress: 1/2
+  resourcesDuration:
+    cpu: 8
+    memory: 4
+  startedAt: "2022-07-18T22:45:27Z"
```
tczhao, line 3610: I think we can remove all these static `status` field
and run `woc.operate(ctx)` twice to achieve the same purpose

workflow/controller/steps_test.go
```diff
@@ -314,3 +314,186 @@ func TestOptionalArgumentAndParameter(t *testing.T) {
...
+      startedAt: "2022-07-18T21:30:51Z"
+      templateName: fail
+      templateScope: local/steps-with-priority-r46kf
+      type: Pod
+  phase: Failed
+  progress: 1/2
+  resourcesDuration:
+    cpu: 11
+    memory: 5
+  startedAt: "2022-07-18T21:30:51Z"
```
tczhao, line 481: same here

tczhao (changes requested):
overall looks good with some minor changes

workflow/controller/dag_test.go
tczhao: Hi @dpadhiar , it won't mark the workflow as complete

---

here's how it works in my head:
- 1st operate, executes pod B due priority
- 2nd operate, executes pod A due priority

Then we can compare the `startedAt`

At this stage, the workflow is at phase.status running, and I think this is all we need.

---

Giving it the status with phase Failed won't cause operate() to perform any further action, we will end up evaluating the test data instead of the controller functionality

---

Let me know what you think :)

dpadhiar: @tczhao This implementation worked exactly as you said! Thank you for the help, ready to add.

tczhao (approved):
Lgtm 👍

alexec:
This PR is 2 months old. Is it abandoned?

dpadhiar:
> This PR is 2 months old. Is it abandoned?

No, should be ready to merge.

pkg/apis/workflow/v1alpha1/common.go
```diff
@@ -36,6 +38,17 @@ type TemplateReferenceHolder interface {
...
 	// GetTemplateName returns the template name. This maybe empty. This is last precedence.
 	GetTemplateName() string
+	// GetPriority returns the priority of the step or task.
+	GetPriority() int32
+	// GetName returns the name of the step or task.
+	GetName() string
+}
+
+func CompareByPriority(a, b TemplateReferenceHolder) bool {
+	if a.GetPriority() > b.GetPriority() {
```
alexec, line 48: I think you need a test.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1408,6 +1408,9 @@ type WorkflowStep struct {
 	// Hooks holds the lifecycle hook which is invoked at lifecycle of
 	// step, irrespective of the success, failure, or error status of the primary step
 	Hooks LifecycleHooks `json:"hooks,omitempty" protobuf:"bytes,12,opt,name=hooks"`
+
+	// Priority is used if controller is configured to process limited number of steps in parallel. Steps with higher priority are processed first.
```
alexec, line 1412: I think this could be a simpler comment. E.g. "Priority controlles which step is scheduled first. Lower values are processed first."

workflow/controller/dag.go
alexec: nice

workflow/controller/dag.go
```diff
@@ -651,22 +651,37 @@ func (woc *wfOperationCtx) resolveDependencyReferences(dagCtx *dagContext, task
 // findLeafTaskNames finds the names of all tasks whom no other nodes depend on.
 // This list of tasks is used as the the default list of targets when dag.targets is omitted.
 func (d *dagContext) findLeafTaskNames(tasks []wfv1.DAGTask) []string {
+
+	prioritySort := false
```
alexec, line 655: I think we should get rid of this boolean and always process in order.

alexec (changes requested):

tczhao (approved):

pkg/apis/workflow/v1alpha1/common_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCompareByPriority(t *testing.T) {
+
+	stepOne := WorkflowStep{Name: "A", Priority: 100}
```
alexec, line 11: I'd have two sub-tests: one for sorting by priority and another by name

dpadhiar: Added.

pkg/apis/workflow/v1alpha1/common_test.go
```diff
@@ -1,22 +1,37 @@
...
+		stepGroup := []WorkflowStep{{Name: "Banana"}, {Name: "Durian"}, {Name: "Apple"}, {Name: "Cherry"}}
 
+		sort.Slice(stepGroup, func(i, j int) bool {
+			return CompareByPriority(&stepGroup[i], &stepGroup[j])
+		})
+
+		assert.Equal(t, "Durian", stepGroup[0].Name)
+		assert.Equal(t, "Cherry", stepGroup[1].Name)
+		assert.Equal(t, "Banana", stepGroup[2].Name)
+		assert.Equal(t, "Apple", stepGroup[3].Name)
```
tczhao, line 35: in the original version, the sort is in alphabetical order, I think we shouldn't change that behavior

tczhao (changes requested):

PR #8856, opened 2022-05-25
feat: Artifact GC Finalizer needs to be added if any Output Artifacts have a strategy

### LINKED ISSUE #8612: The code that adds the artifact GC finalizer ignores the strategy of artifact
This code only takes into account the strategy set at workflow level. I needs to take into account the value set on the artifact.

juliev0:
Fixes #8612 

My understanding is that we should set the Finalizer if:
1. WorkflowSpec.ArtifactStrategy set
OR:
2. any Output Artifact sets ArtifactStrategy

Thus, in the case that WorkflowStrategy.ArtifactStrategy is "none" but the artifact's strategy is set, the artifact value overrides the workflow value.

### DIFF
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -404,7 +404,8 @@ type WorkflowSpec struct {
 	// WorkflowMetadata contains some metadata of the workflow to be refer
 	WorkflowMetadata *WorkflowMetadata `json:"workflowMetadata,omitempty" protobuf:"bytes,42,opt,name=workflowMetadata"`
 
-	// ArtifactGC describes the strategy to use when to deleting artifacts from completed or deleted workflows
+	// ArtifactGC describes the strategy to use when to deleting artifacts from completed or deleted workflows (applies to all output Artifacts
+	// unless Artifact.ArtifactGC is specified, which overrides this)
 	ArtifactGC *ArtifactGC `json:"artifactGC,omitempty" protobuf:"bytes,43,opt,name=artifactGC"`
 }
 
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -3517,22 +3517,20 @@ func (woc *wfOperationCtx) addFinalizers() {
 }
 
 func (woc *wfOperationCtx) addArtifactGCFinalizer() {
-	if woc.execWf.Spec.ArtifactGC != nil && woc.execWf.Spec.ArtifactGC.Strategy != wfv1.ArtifactGCNever {
-		// verify that the output Artifacts aren't all set to ArtifactGCNever (undefined is okay)
-		garbageCollect := false
-		for _, template := range woc.execWf.Spec.Templates {
-			for _, artifact := range template.Outputs.Artifacts {
-				if artifact.ArtifactGC == nil || artifact.ArtifactGC.Strategy != wfv1.ArtifactGCNever {
-					garbageCollect = true
-					break
-				}
+
+	artifactDefinedGC := false
+	for _, template := range woc.execWf.Spec.Templates {
+		for _, artifact := range template.Outputs.Artifacts {
+			if artifact.ArtifactGC != nil && artifact.ArtifactGC.Strategy != wfv1.ArtifactGCNever {
+				artifactDefinedGC = true
+				break
 			}
 		}
+	}
 
-		if garbageCollect {
-			finalizers := append(woc.wf.GetFinalizers(), common.FinalizerArtifactGC)
-			woc.wf.SetFinalizers(finalizers)
-		}
+	if (woc.execWf.Spec.ArtifactGC != nil && woc.execWf.Spec.ArtifactGC.Strategy != wfv1.ArtifactGCNever) || artifactDefinedGC {
+		finalizers := append(woc.wf.GetFinalizers(), common.FinalizerArtifactGC)
+		woc.wf.SetFinalizers(finalizers)
 	}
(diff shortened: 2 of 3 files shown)


workflow/controller/operator.go
```diff
@@ -3517,7 +3517,18 @@ func (woc *wfOperationCtx) addFinalizers() {
 }
 
 func (woc *wfOperationCtx) addArtifactGCFinalizer() {
-	if woc.execWf.Spec.ArtifactGC != nil && woc.execWf.Spec.ArtifactGC.Strategy != wfv1.ArtifactGCNever {
+
+	artifactDefinedGC := false
+	for _, template := range woc.execWf.Spec.Templates {
```
alexec, line 3522: what about encapsulating into `Workflow.HasArtifactGC`?

workflow/controller/operator_test.go
```diff
@@ -3809,6 +3807,107 @@ func TestArtifactGCOnWorkflowCompletion(t *testing.T) {
...
+// verify Finalizer gets added when the artifact strategy indicates GC for Artifacts
+func TestArtifactGC_ArtifactStrategy(t *testing.T) {
+	cancel, controller := newController()
+	defer cancel()
+	ctx := context.Background()
+
+	wf := wfv1.MustUnmarshalWorkflow(`
+  apiVersion: argoproj.io/v1alpha1
+  kind: Workflow
+  metadata:
```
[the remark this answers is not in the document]
juliev0: sounds good

workflow/controller/operator.go
juliev0: Yeah, thought about doing that myself. Will do.

juliev0: Done. Let me know if you think I should also create a method on the Template level or not.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -196,6 +196,20 @@ func (w *Workflow) GetExecSpec() *WorkflowSpec {
 	return &w.Spec
 }
 
+func (w *Workflow) HasArtifactGC() bool {
+	artifactDefinedGC := false
+	for _, template := range w.Spec.Templates {
+		for _, artifact := range template.Outputs.Artifacts {
```
alexec, line 202: GetOutputs()

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -196,6 +196,20 @@ func (w *Workflow) GetExecSpec() *WorkflowSpec {
...
+			if artifact.GetArtifactGC().Strategy != ArtifactGCNever {
+				artifactDefinedGC = true
+				break
```
alexec, line 205: `return true` here?

pkg/apis/workflow/v1alpha1/workflow_types.go
juliev0: @alexec GetOutputs() is a method of NodeStatus, not Template. Did you want me to implement Template.GetOutputs()?

alexec: Only if outputs can be nil

workflow/controller/operator_test.go
```diff
@@ -3778,37 +3778,6 @@ func TestNestedOptionalOutputArtifacts(t *testing.T) {
...
-	wf := wfv1.MustUnmarshalWorkflow(artifactGCOnWorkflowCompletion)
-	woc := newWorkflowOperationCtx(wf, controller)
-	woc.operate(ctx)
-
-	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
-	assert.Equal(t, 1, len(woc.wf.GetFinalizers()))
-	assert.Equal(t, "workflows.argoproj.io/artifact-gc", woc.wf.GetFinalizers()[0])
-}
-
 //  TestPodSpecLogForFailedPods tests PodSpec logging configuration
```
juliev0, line 3781: The logic got moved into workflow_types.go, so the unit test moved to workflow_types_test.go. The remaining logic in operator.go seemed too simple to be worthy of a unit test, so I removed it.

pkg/apis/workflow/v1alpha1/workflow_types_test.go
```diff
@@ -116,6 +117,102 @@ func TestWorkflowHappenedBetween(t *testing.T) {
...
+	tests := []struct {
+		name                      string
+		workflowArtGCStrategySpec string
+		artifactGCStrategySpec    string
+		expectedResult            bool
+	}{
+		{
+			name: "WorkflowSpecGC_Completion",
+			workflowArtGCStrategySpec: `
+      artifactGC:
```
juliev0, line 130: The Go formatter is doing some weird alignment of text...

pkg/apis/workflow/v1alpha1/workflow_types_test.go
```diff
@@ -116,6 +117,97 @@ func TestWorkflowHappenedBetween(t *testing.T) {
...
+			name: "WorkflowSpecGC_Completion",
+			workflowArtGCStrategySpec: `
+              artifactGC:
+                strategy: OnWorkflowCompletion`,
+			artifactGCStrategySpec: "",
+			expectedResult:         true,
+		},
+		{
+			name:                      "ArtifactSpecGC_Completion",
+			workflowArtGCStrategySpec: "",
```
juliev0, line 137: The Go formatter is doing some weird alignment of text...

alexec (approved):

Issue #4286, opened 2020-10-15
Allow whitespace in templates


idsvandermolen:
# Summary

Currently Argo does not allow white space in templating, rendering of `{{workflow.parameters.variable}}` does correctly work, however `{{ workflow.parameters.variable }}` doesn't render the variable. Also `argo lint` did not issue any warning or error.

# Use Cases

I often find it more readable when you can add whitespace to separate markers like `{{` and `}}` from variable names.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

alexec:
Good idea. Would you like to investigate how to fix this?

idsvandermolen:
I'm not that proficient in Go, but looks like in function `Replace` this https://github.com/argoproj/argo/blob/8facaceeb3515d804c3fd276b1802dbd6cf773e8/workflow/common/util.go#L375 
could probably be replaced with:
```golang
  replacement, ok := replaceMap[strings.TrimSpace(tag)]
```

alexec:
Want to submit a PR?

idsvandermolen:
👍 let me have a look

idsvandermolen:
There you go, PR #4310

idsvandermolen:
btw I had to make a few `Makefile` adjustments to be able to build (see PR #4311)

PR #8867, opened 2022-05-26
Allow optional space for param detection

RyanCopley:
Really, I intend for this to be more of a discussion than anything. Now that I think I understand what's going on, i can proceed with my work, but want to make sure this is known.

I work for a company that has a large argo workflow, and we've been struggling to reconcile strange behaviors around the space needed for the interpolated variables. I believe this is because this specific regex is not meeting the contract that was documented, and does not allow for those additional white space between the bracket and variable.

https://github.com/argoproj/argo-workflows/blob/master/docs/variables.md#simple

> Simple tags may have white-space between the brackets and variable.

https://github.com/argoproj/argo-workflows/blob/master/util/template/simple_template.go#L13 provides that contract during the workflows actual execution (e.g. `argo submit`), however, there is a code path during `argo lint` that tries to use the file i'm changing to determine if it is or isn't a parameter, in order to detect if it should allow that as a "global" param, but it doesn't follow that same contract?

https://github.com/argoproj/argo-workflows/blob/master/workflow/validate/validate.go#L872

I posted about this in the slack:

https://cloud-native.slack.com/archives/C01QW9QSSSK/p1653497167145119

https://cloud-native.slack.com/archives/C01QW9QSSSK/p1653600495901169

### DIFF
--- a/workflow/validate/validate.go
+++ b/workflow/validate/validate.go
@@ -1373,7 +1373,7 @@ func sortDAGTasks(tmpl *wfv1.Template, ctx *dagValidationContext) error {
 
 var (
 	// paramRegex matches a parameter. e.g. {{inputs.parameters.blah}}
-	paramRegex               = regexp.MustCompile(`{{[-a-zA-Z0-9]+(\.[-a-zA-Z0-9_]+)*}}`)
+	paramRegex               = regexp.MustCompile(`{{\s?[-a-zA-Z0-9]+(\.[-a-zA-Z0-9_]+)*\s?}}`)
 	paramOrArtifactNameRegex = regexp.MustCompile(`^[-a-zA-Z0-9_]+[-a-zA-Z0-9_]*$`)
 	workflowFieldNameRegex   = regexp.MustCompile("^" + workflowFieldNameFmt + "$")
 )


tczhao:
related https://github.com/argoproj/argo-workflows/issues/4286

tczhao (changes requested):
A few things:

We'd better move this to an issue if its intended for discussion
For the PR. Can you fix the DCO, see [link](https://github.com/argoproj/argo-workflows/pull/8867/checks?check_run_id=6616955074) or `git commit -s --amend`

There seems more cases where this white space thing is not respected, e.g. https://github.com/argoproj/argo-workflows/blob/master/workflow/validate/validate.go#L1270. I'd probably start with searching all the `{{` codes...

We need to add more test to validate these edge cases

workflow/validate/validate.go
```diff
@@ -1373,7 +1373,7 @@ func sortDAGTasks(tmpl *wfv1.Template, ctx *dagValidationContext) error {
 
 var (
 	// paramRegex matches a parameter. e.g. {{inputs.parameters.blah}}
-	paramRegex               = regexp.MustCompile(`{{[-a-zA-Z0-9]+(\.[-a-zA-Z0-9_]+)*}}`)
```
sarabala1979, line 1376: Can you add  test?

RyanCopley:
I don't intend to come across as rude, but unfortunately my schedule doesn't allow me to put more effort into this when, now that I did this analysis, we have a way forward avoiding this problem. That means I have no interest (read: time, motivation) in any further commits to this PR to bring it up to your standards. I simply wish to let you all know that this contract is being broken, and a very thin example of what the issue is so you can understand and prioritize it on your own.

Good luck though, this issue was incredibly hard to wrap my head around and has led to toil on my team :(  For the record, we'll be standardizing on wether or not to allow spaces in these tags, and update our existing workflows accordingly.

alexec:
No more work expected on this PR. Closing.

agilgur5:
I believe this was superseded by #11781 and perhaps some PRs before it

PR #8893, opened 2022-05-31
feat: add Artifact.Deleted

### LINKED ISSUE #8665: Add `Artifact.Deleted` flag
We need to know if an artifact was deleted. Add this field.

juliev0:

Fixes #8665


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -946,6 +946,9 @@ type Artifact struct {
 
 	// ArtifactGC describes the strategy to use when to deleting an artifact from completed or deleted workflows
 	ArtifactGC *ArtifactGC `json:"artifactGC,omitempty" protobuf:"bytes,12,opt,name=artifactGC"`
+
+	// Has this been deleted?
+	Deleted bool `json:"deleted,omitempty" protobuf:"varint,13,opt,name=deleted"`
 }
```
juliev0, line 952: I know we don't want somebody to specify 'Deleted' in their spec, but I assume this is the right way to add this based on how other status type fields are specified....

alexec (approved):

PR #8925, opened 2022-06-06
feat: refactoring e2e test timeouts to support multiple environments.

dpadhiar:
Currently, running `make test-%` to run e2e tests is hardcoded to have a 15 minute timeout. With this, it is possible that tests that normally would pass in some environments will fail in others based on timeout. This PR changes these hardcoded values to reflect the env variables `E2E_TIMEOUT`, `E2E_PARALLEL`, `E2E_SUITE_TIMEOUT`. This way, a user can adjust timeout until the test cases pass as planned. This PR also introduces `E2E_ENV_FACTOR`, a variable that can adjust the time allowed for certain test cases that wait for a workflow to reconcile. Making this value variable allows for a user to ensure all tests will pass on their environment.


.github/workflows/ci-build.yaml
```diff
@@ -151,7 +151,7 @@ jobs:
       - run: make start PROFILE=${{matrix.profile}} AUTH_MODE=client STATIC_FILES=false LOG_LEVEL=info API=${{matrix.test == 'test-api' || matrix.test == 'test-cli' || matrix.test == 'test-java-sdk' || matrix.test == 'test-python-sdk'}} UI=false > /tmp/argo.log 2>&1 &
       - run: make wait
         timeout-minutes: 4
-      - run: make ${{matrix.test}} E2E_TIMEOUT=1m STATIC_FILES=false
+      - run: make ${{matrix.test}} E2E_TIMEOUT=15m STATIC_FILES=false
```
tczhao, line 154: I think the E2E_TIMEOUT here is for each workflow or each WaitForWorkflow,
We shouldn't have a workflow that takes up to 15min to run

tczhao (changes requested):

.github/workflows/ci-build.yaml
sarabala1979: @tczhao  This configuration was never applied into make file. This PR is fixing it. Makefile had 15m time out for entire tests to complete.

tczhao: @sarabala1979 I see what you meant,
I think the original intention of this `E2E_TIMEOUT` is timeout for individual workflows, it is getting loaded here
https://github.com/argoproj/argo-workflows/blob/master/test/e2e/fixtures/e2e_suite.go#L39
https://github.com/argoproj/argo-workflows/blob/master/test/e2e/fixtures/when.go#L245

maybe we should separate these parameters since we have parameters for overall timeout and individual workflow timeout from this PR

dpadhiar: In this case, `timeout` for waiting on workflows is defaulting to E2E_TIMEOUT correct? And we want to make it so E2E_TIMEOUT only corresponds to overall timeout, not a single workflow?

sarabala1979: @tczhao you are right. @dpadhiar we can add new variable E2E_SUIT_TIMEOUT This is mainly for entire suite timeout

Makefile
```diff
@@ -29,6 +29,10 @@ K3D_CLUSTER_NAME      ?= k3s-default
 KUBE_NAMESPACE        ?= argo
 MANAGED_NAMESPACE     ?= $(KUBE_NAMESPACE)
 
+E2E_TIMEOUT           ?= 1m
+E2E_PARALLEL          ?= 10
```
terrytangyuan, line 33: Parallelism?

Makefile
```diff
@@ -29,6 +29,10 @@ K3D_CLUSTER_NAME      ?= k3s-default
 KUBE_NAMESPACE        ?= argo
 MANAGED_NAMESPACE     ?= $(KUBE_NAMESPACE)
 
+E2E_TIMEOUT           ?= 1m
```
terrytangyuan, line 32: Individual timeout? A bit confusing with suite timeout.

Makefile
dpadhiar: This is for parallel tests.

`go test -failfast -v -timeout $(E2E_SUITE_TIMEOUT) -count 1 --tags api,cli,cron,executor,examples,functional,plugins -parallel $(E2E_PARALLEL) ./test/e2e  -run='.*/$*'`

Makefile
dpadhiar: This variable represents the timeouts for waiting on workflows, pods etc. We could change this to a different name ie. `E2E_WAIT` or `E2E_WAIT_TIMEOUT`.

terrytangyuan: Thanks. I don't have a preference on names as long as we document them. Could you add a comment above the declaration or document in one of the developer guides?

dpadhiar: Yes! Will add in the Makefile.

terrytangyuan (approved):

PR #8944, opened 2022-06-10
docs: Add another CI example. Fixes #8813

tim-sendible:
Signed-off-by: tim-sendible <tim@sendible.com>

Fixes #8813


terrytangyuan (approved):
IMO it might be better to keep all example content in this repo.

Issue #8917, opened 2022-06-05
Make side panels resizable


dakotalillie:
# Summary

It would be nice if side panels (such as the ones for viewing workflow node details and logs) were resizable.

# Use Cases

The primary use case I have in mind would be to make it easier to view content which exceeds the standard width of the side panel. It might also be useful with the logs side panel, to allow the primary page content to be visible alongside the logs.

I did a rough mockup (sorry for the poor performance, screen recording slows my computer down significantly):

https://user-images.githubusercontent.com/20373585/172034555-5ada41e7-3e61-443a-8680-676bae3a50a3.mov

I'd be happy to submit a PR if this gets the go-ahead.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

sarabala1979:
@dakotalillie Nice enhancement. Do you like to submit the PR for this feature?

dakotalillie:
Sure thing! I'll submit a PR shortly.

PR #8947, opened 2022-06-10
fix: long code blocks overflow in ui. Fixes #8916

### LINKED ISSUE #8916: Long container args and/or script source overflows in UI
## Checklist

 

## Summary

Providing a lengthy argument to a container template, or a lengthy source to a script template, causes the provided argument/source to visually overflow in the UI.

I have observed this issue in multiple versions, including:

- v3.3.6
- "latest" (master branch)

## Diagnostics

Here are two YAML examples which can be used to reproduce this issue, along with screenshots of the UI for each:

```yaml
kind: Workflow
metadata:
  generateName: example-
spec:
  entrypoint: example
  templates:
  - name: example
    container:
      image: ubuntu:latest
      command: [bash]
      args:
        - -c
        - |-
          # The exact contents of this script don't matter.
          # All that matters for this example is that the line is sufficiently long.
          if [ "$EXAMPLE_ENV_VAR" == "hello darkness my old friend" ]; then
            echo "Hello world!"
          fi
```

Screen Shot 2022-06-04 at 6 12 40 PM

```yaml
kind: Workflow
metadata:
  generateName: example-
spec:
  entrypoint: example
  templates:
  - name: example
    script:
      image: ubuntu:latest
      command: [bash]
      source: |-
        # The exact contents of this script don't matter.
        # All that matters for this example is that the line is sufficiently long.
        if [ "$EXAMPLE_ENV_VAR" == "hello darkness my old friend" ]; then
          echo "Hello world!"
        fi
```

Screen Shot 2022-06-04 at 6 14 00 PM

It looks like a similar issue was addressed in #7876. Adopting the same fix for this should be straightforward. However, I wanted to check first to see whether an alternative solution might be preferable, such as using a scrollbar. Here's what two potential solutions might look like:

Screen Shot 2022-06-04 at 6 21 23 PM

https://user-images.githubusercontent.com/20373585/172031016-3c0ba8b4-2d61-4a5f-b4e5-9a61abe1d76b.mov

Either way, I'm happy to submit a PR.

---

dakotalillie:
Signed-off-by: Dakota Lillie <dakota.lillie@icloud.com>

Fixes #8916

This PR addresses an issue where code blocks would overflow their container in the UI.

**Before:**

Screen Shot 2022-06-10 at 9 30 25 AM

**After:**

Screen Shot 2022-06-10 at 9 31 24 AM

I have tested this in Chrome, Firefox, and Safari. Safari does have an issue where the scrollbar does not increase the height of the element (notice the lack of padding below the code block):

Screen Shot 2022-06-10 at 10 12 15 AM

Without the `hidden` overflow-y setting, this would cause an additional vertical scrollbar to appear. I'm inclined to believe this is a bug in Safari, as manually updating any height-related style in the browser, even a random one like "line-height", causes Safari to fix itself:

https://user-images.githubusercontent.com/20373585/173118493-043429b7-796f-4331-b2fc-df49fdee27e3.mov

Needless to say, changing the "line-height" in the code does not fix the issue.

I chose to add a scrollbar, rather than wrapping the text, because I find line-wrapped code to be more difficult to read. I can switch approaches if folks would prefer differently, but hopefully this should become a largely moot issue once #8917 is implemented.

I verified to ensure this change does not cause a regression for the bug fixed in #7876. It appears it does not:

Screen Shot 2022-06-10 at 10 34 22 AM

Even with the removal of `white-space: pre-wrap`, the text still wraps due to `overflow-wrap: break-word` [set here](https://github.com/argoproj/argo-ui/blob/a5f7fa7197d297a0aec9a7ee7c6d919763611060/src/styles/elements/containers.scss#L97).

I had to disable the pre-commit checks to make this commit, as `make test` runs a test which [attempts to clone a private GitHub repository](https://github.com/argoproj/argo-workflows/blob/750c4e1f699b770a309843f2189b4e703305e44f/workflow/artifacts/git/git_test.go#L35) I don't have access to. I was successfully able to run `make pre-commit -B`.

### DIFF
--- a/ui/src/app/workflows/components/workflow-node-info/workflow-node-info.tsx
+++ b/ui/src/app/workflows/components/workflow-node-info/workflow-node-info.tsx
@@ -68,7 +68,7 @@ interface Props {
 const AttributeRow = (attr: {title: string; value: any}) => (
     <div className='row white-box__details-row' key={attr.title}>
         <div className='columns small-4'>{attr.title}</div>
-        <div className='columns columns--narrower-height small-8' style={{whiteSpace: 'pre-wrap'}}>
+        <div className='columns columns--narrower-height small-8' style={{overflow: 'auto hidden'}}>
             {attr.value}
         </div>
     </div>


tczhao (approved):
LGTM
The PR description is very neat

PR #8963, opened 2022-06-14
feat: make node info side panel resizable. Fixes #8917

### LINKED ISSUE #8917: Make side panels resizable

dakotalillie:
Signed-off-by: Dakota Lillie <dakota.lillie@icloud.com>

Fixes #8917 

## Change Summary

- Allow the user to resize the node info side panel

https://user-images.githubusercontent.com/20373585/173668411-553a2141-884e-4e03-92e6-c4f5dc596e97.mov

## Notes

I have tested this feature in Chrome, Firefox, and Safari.

As a result of this change, when opening the sidebar to inspect the details of an artifact node, the sidebar will open to the same width it opens to for other nodes (570px) rather than the 50% width to which it previously opened.

I had to update the grid system used by the attributes in the various tabs, because with the original grid system, both columns would increase in size when the side panel was resized, which looked kind of weird. The new grid system uses CSS grid, and ensures that only the right-most column increases in size. This change also addresses some other superficial issues pertaining to line height.

The original issue mentioned also potentially applying this feature to the logs sidebar. I elected not to do that here, to reduce the scope of the PR.

I would love to add tests for this feature, but it doesn't seem like the UI has testing infrastructure in place. Please let me know if I missed something — I will gladly add them.


alexec (approved):
This works fine tested locally.

alexec:
Please resolve conflicts.

PR #8966, opened 2022-06-15
feat: If artifact has been deleted, show a message to that effect in the iFrame in the UI

### LINKED ISSUE #8666: Don't try and show deleted artifacts in the UI
The UI wil try and display deleted artifacts, returning an error. Instead, just don't try and show them. Show "Artifact has been deleted".

juliev0:
Fixes #8666


server/artifacts/artifact_server.go
```diff
@@ -214,6 +214,10 @@ func (a *ArtifactServer) GetArtifactFile(w http.ResponseWriter, r *http.Request)
 
 	} else { // stream the file itself
 		log.Debugf("not a directory, artifact: %+v", artifact)
+		if artifact.Deleted {
+			a.notFoundError(w)
+			return
+		}
 		err = a.returnArtifact(w, artifact, driver)
```
juliev0, line 221: Return 404 if we garbage collected the file. If for some reason, it got deleted manually outside of Argo Workflows, we could theoretically return a 404 as well, but that means needing to get the variety of error back from ArtifactDriver for each type of driver, which is more work.

server/artifacts/artifact_server.go
```diff
@@ -343,6 +357,10 @@ func (a *ArtifactServer) httpBadRequestError(w http.ResponseWriter) {
 	http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
 }
 
+func (a *ArtifactServer) notFoundError(w http.ResponseWriter) {
```
alexec, line 360: looks fine

ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
```diff
@@ -43,12 +45,25 @@ export const ArtifactPanel = ({
...
         setObject(null);
         setError(null);
+        setShowDownloadLink(true);
         if (ext === 'json') {
             requests
                 .get(services.workflows.artifactPath(workflow, artifact.nodeId, artifact.name, archived, input))
                 .then(r => r.text)
                 .then(setObject)
                 .catch(setError);
+        } else {
```
alexec, line 55: I don't think you need this. The iframe will show "Not found". I think that is enough.

alexec: In fact, I don't think you need any UI changes.

server/artifacts/artifact_server.go
juliev0: oh, I didn't realize about the wrapped errors...okay, I'll use that to get the real status I suppose

ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
juliev0: Oh, I thought you wanted a friendly message here? The story says to show "Artifact has been deleted". But yes, I can see how showing the default "File not found" message would be sufficient.

server/artifacts/artifact_server.go
```diff
@@ -344,14 +344,42 @@ func (a *ArtifactServer) httpBadRequestError(w http.ResponseWriter) {
...
+				statusCode = http.StatusNotFound
+			case argoerrors.CodeBadRequest:
+				statusCode = http.StatusBadRequest
+			case argoerrors.CodeNotImplemented:
+				statusCode = http.StatusNotImplemented
+			case argoerrors.CodeTimeout, argoerrors.CodeInternal:
+				statusCode = http.StatusInternalServerError
+			default:
+				statusCode = http.StatusInternalServerError
+			}
```
juliev0, line 373: while I could have excluded argoerrors.CodeTimeout and argoerrors.CodeInternal and let them fall through to default, I left them in to show they were addressed

ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
alexec: I think that should be fine, because they will get an error. What you don't want to do is download a large file twice.

juliev0: I guess it won't cache it, will it?

juliev0: Another possibility is to set the iFrame content equal to the result of the earlier request. But perhaps that can be for another day.

server/artifacts/artifact_server.go
```diff
@@ -344,14 +346,38 @@ func (a *ArtifactServer) httpBadRequestError(w http.ResponseWriter) {
...
-		http.Error(w, http.StatusText(statusCode), statusCode)
+		statusCode = int(e.Status().Code)
 	} else {
-		// Unknown error - return internal error
-		a.serverInternalError(err, w)
+		// check if it's an internal ArgoError
+		switch resolvedError := err.(type) {
+		case argoerrors.ArgoError:
+			switch resolvedError.Code() {
+			case argoerrors.CodeUnauthorized:
```
sarabala1979, line 362: Are ArgoErrors codes and HTTP codes different? can we have same code so it can be send it?

juliev0: So, the argoError code is a string, not an int. I am guessing that Jesse intended that the error be generic and not specific to http. But it may be reasonable to include HTTPCode() into ArgoError, or otherwise to have some generic function which translates ArgoError into an http code, or possibly to move this whole function somewhere more central, since it's not really specific to ArtifactServer.  @jessesuen do you have a preference?

juliev0: @sarabala1979 Maybe no response from Jesse. In that case, we can just decide. It looks like [ArgoError](https://github.com/argoproj/argo-workflows/blob/master/errors/errors.go#L22) is being used in places that are not http related. Do you think it makes sense to add HTTPCode() as a method of the ArgoError interface anyway? Or I can add a function to errors/errors.go which translates an ArgoError into an http code.

sarabala1979: I would prefer to add HTTPCode in the ArgoError interface. Is ArgoError interface implemented by any other object other than argoerr?

juliev0: don't think so. I added HTTPCode() to ArgoError and argoerr

server/artifacts/artifact_server.go
```diff
@@ -356,22 +356,9 @@ func (a *ArtifactServer) httpFromError(err error, w http.ResponseWriter) {
...
-			case argoerrors.CodeBadRequest:
-				statusCode = http.StatusBadRequest
-			case argoerrors.CodeNotImplemented:
-				statusCode = http.StatusNotImplemented
-			case argoerrors.CodeTimeout, argoerrors.CodeInternal:
-				statusCode = http.StatusInternalServerError
-			}
+		argoerr, typeOkay := err.(argoerrors.ArgoError)
+		if typeOkay {
+			statusCode = argoerr.HTTPCode()
```
sarabala1979, line 361: like it

sarabala1979 (approved):

PR #8988, opened 2022-06-16
feat: log workflow size before hydrating/dehydrating. Fixes #8976

### LINKED ISSUE #8976: Logging  workflow size before Dehydrate the workflow
# Summary

What change needs making?

# Use Cases

When would you use this?

---

dpadhiar:

Fixes #8976 

If a workflow fails to be hydrated / dehydrated, it can be hard to tell what the issue may be. Many times, it may be because the workflow is too large and thus this change will log the size of the workflow so they may investigate for this case.

### DIFF
--- a/workflow/hydrator/hydrator.go
+++ b/workflow/hydrator/hydrator.go
@@ -85,6 +85,7 @@ func (h hydrator) Hydrate(wf *wfv1.Workflow) error {
 		}
 		h.HydrateWithNodes(wf, offloadedNodes)
 	}
+	log.WithField("Workflow Size", wf.Size()).Info("Workflow hydrated")
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


workflow/hydrator/hydrator.go
```diff
@@ -85,6 +85,7 @@ func (h hydrator) Hydrate(wf *wfv1.Workflow) error {
 		}
 		h.HydrateWithNodes(wf, offloadedNodes)
 	}
+	log.WithField("Workflow Size", wf.Size()).Info("Workflow hydrated")
```
sarabala1979, line 88: can you move to in  86 after h.HydrateWithNodes?

dpadhiar: yes

sarabala1979 (approved):

PR #8991, opened 2022-06-16
fix: add nil check for retryStrategy.Limit in deadline check. Fixes #8990

### LINKED ISSUE #8990: retryStrategy limit does not have a nil check
`retryStrategy.Limit` is a pointer value which maybe be nil. This will because a panic when it is not checked before attempting to use `IntVal`.

https://github.com/argoproj/argo-workflows/blob/19eae92db339fa79eb91fb71a50d330a18b09bcf/workflow/controller/operator.go#L907-L912

dpadhiar:

Fixes #8990


tczhao (approved):
lgtm

terrytangyuan (approved):

PR #8993, opened 2022-06-17
feat: api ratelimiter for argoserver

### LINKED ISSUE #7982: Rate-limit Argo Server API requests
# Summary

I'm not sure if we have a rate-limiter. If not, we should add one.

# Use Cases

When would you use this?

---

tczhao:
Fixes #7982

### DIFF
--- a/server/apiserver/argoserver.go
+++ b/server/apiserver/argoserver.go
@@ -151,7 +151,7 @@ func NewArgoServer(ctx context.Context, opts ArgoServerOpts) (*argoServer, error
 		return nil, err
 	}
 	apiRateLimiter := apiratelimiter.NewApiRateLimiter(opts.ApiRateLimit, opts.ApiRateBurst)
-	go apiRateLimiter.CleanupVisitors()
+	go apiRateLimiter.CleanupVisitors(1*time.Minute, 5*time.Minute)
 	return &argoServer{
 		baseHRef:                 opts.BaseHRef,
 		tlsConfig:                opts.TLSConfig,
--- a/server/apiserver/argoserver.go
+++ b/server/apiserver/argoserver.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/gorilla/handlers"
@@ -86,7 +87,10 @@ type argoServer struct {
 	eventAsyncDispatch       bool
 	xframeOptions            string
 	accessControlAllowOrigin string
-	rateLimiter              *rate.Limiter
+	apiRateLimit             int
+	apiRateBurst             int
+	visitors                 map[string]*rate.Limiter
+	mu                       *sync.Mutex
 	cache                    *cache.ResourceCache
 }
 
@@ -150,7 +154,10 @@ func NewArgoServer(ctx context.Context, opts ArgoServerOpts) (*argoServer, error
 	if err != nil {
 		return nil, err
 	}
-	rateLimiter := rate.NewLimiter(rate.Limit(opts.ApiRateLimit), opts.ApiRateBurst)
+
+	// Create a map to hold the rate limiters for each visitor and a mutex.
+	var visitors = make(map[string]*rate.Limiter)
+	var mu sync.Mutex
 	return &argoServer{
 		baseHRef:                 opts.BaseHRef,
 		tlsConfig:                opts.TLSConfig,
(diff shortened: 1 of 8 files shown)


alexec (commented):
This is a non trivial PR. It is also hard to test. My question would be - can we use some off the shelf component?

util/apiratelimiter/apiratelimiter.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Retrieve and return the rate limiter for the current visitor if it
+// already exists. Otherwise create a new rate limiter and add it to
+// the visitors map, using the IP address as the key.
+func (r *apiRateLimiter) GetVisitor(ip string) *rate.Limiter {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	v, exists := r.visitors[ip]
+	if !exists {
```
juliev0, line 46: Could consider using an RWMutex to prevent all of the readers on line 45 from locking each other out...or maybe these calls are infrequent enough that it doesn't matter?

tczhao: good point, let me change that

juliev0: great!

juliev0 (approved):

.github/workflows/ci-build.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
   e2e-tests:
     name: E2E Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 20
```
alexec, line 81: Please revert the changes to build files. Those should go in another PR if needed.

Makefile
```diff
@@ -511,7 +511,7 @@ mysql-cli:
 test-cli: ./dist/argo
 
 test-%:
-	go test -failfast -v -timeout 15m -count 1 --tags $* -parallel 10 ./test/e2e
+	go test -failfast -v -timeout 20m -count 1 --tags $* -parallel 10 ./test/e2e
```
alexec, line 514: Please revert.

server/apiserver/argoserver.go
```diff
@@ -106,6 +108,8 @@ type ArgoServerOpts struct {
 	EventAsyncDispatch       bool
 	XFrameOptions            string
 	AccessControlAllowOrigin string
+	ApiRateLimit             int
```
alexec, line 111: minor - should be `APIRateLimit` and `APIRateBurst`

util/grpc/interceptor.go
```diff
@@ -45,3 +50,38 @@ var (
...
+		}
+		return handler(srv, stream)
+	}
+}
+
+// GetClientIP inspects the context to retrieve the ip address of the client
+func getClientIP(ctx context.Context) string {
+	p, ok := peer.FromContext(ctx)
+	if !ok {
+		log.Errorf("couldn't parse client IP address")
```
alexec, line 81: warn rather than error

alexec (changes requested):
I don't think you need to write a lot of this code. Instead you can use this library:

https://github.com/sethvargo/go-limiter

You don't need to write so many tests then.

util/apiratelimiter/apiratelimiter.go
juliev0: @alexec Would the sync.Map cause readers to lock each other out, though?

juliev0: Never mind. I see the documentation for sync.Map says "The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex."

util/grpc/interceptor.go
```diff
@@ -45,3 +50,38 @@ var (
 		return TranslateError(handler(srv, ss))
 	}
 )
+
+// RatelimitUnaryServerInterceptor returns a new unary server interceptor that performs request rate limiting.
+func RatelimitUnaryServerInterceptor(ratelimiter apiratelimiter.ApiRateLimiter) grpc.UnaryServerInterceptor {
```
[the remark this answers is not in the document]
tczhao: The current implementation is a modified version of the example ratelimit.go
the example ratelimit.go is ratelimit at a global level, the current implementation is ratelimit per IP

alexec (approved):

PR #9010, opened 2022-06-20
fix: add logging to expose root cause of panic in Issue 8968

juliev0:

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1943,10 +1943,24 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 	}
 
 	node = woc.wf.GetNodeByName(node.Name)
+	if node == nil {
+		err := fmt.Errorf("no Node found by the name of %s;  wf.Status.Nodes=%+v", node.Name, woc.wf.Status.Nodes)
+		woc.log.Error(err)
+		woc.markWorkflowError(ctx, err)
+		return node, err
+	}
 
 	// Swap the node back to retry node
 	if retryNodeName != "" {
 		retryNode := woc.wf.GetNodeByName(retryNodeName)
+
+		if retryNode == nil {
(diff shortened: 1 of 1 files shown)


workflow/controller/operator.go
```diff
@@ -1943,10 +1943,24 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 	}
 
 	node = woc.wf.GetNodeByName(node.Name)
+	if node == nil {
+		err := fmt.Errorf("no Node found by the name of %s;  wf.Status.Nodes=%+v", node.Name, woc.wf.Status.Nodes)
+		woc.log.Error(err)
+		woc.markWorkflowError(ctx, err)
+		return node, err
+	}
```
juliev0, line 1951: I'm no expert on this code, but I assume it should error if node is nil. ? If not, I will move the error to where we try to reference node below.

tczhao: I think here is fine

looking at the original commit
https://github.com/argoproj/argo-workflows/pull/1552/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1224

The purpose of getting the node here is to retrieve node info after executing node base on template (step/dag..)

One way to look at this is Retry should be based on some kind of node template, so in theory, if there's no template there's no retry.

tczhao: I think here is fine

looking at the original commit
https://github.com/argoproj/argo-workflows/pull/1552/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1224

The purpose of getting the node here is to retrieve node info after executing node base on template (step/dag..)

One way to look at this is Retry should be based on some kind of node template, so in theory, if there's no template there's no retry.

sarabala1979 (approved):

PR #9056, opened 2022-06-24
feat: inform users on UI if an artifact will be deleted. Fixes #8667

### LINKED ISSUE #8667: Inform the user in the UI if an artifact will be delete and when
There is no way to know when at artifact will be delete in the UI. Add to `<ArtifactPanel/>` some text to tell users if this will happen. E.g.  "Artifact will be automatically deleted shortly after the workflow completes".

https://github.com/alexec/argo-workflows/blob/e771dc381a95d144d0ad09838c184fc0aaaf849e/ui/src/app/workflows/components/workflow-details/artifact-panel.tsx#L101

dpadhiar:
Fixes #8667 

If `artifactGC` is set on either an output artifact or in a workflow spec, the user will see a message on the artifact(s) describing when they will be deleted.

Screen Shot 2022-06-24 at 9 51 25 AM


ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
```diff
@@ -63,12 +70,16 @@ export const ArtifactPanel = ({
...
+                        </h3>
                         <p>
                             <small>{urn}</small>
                         </p>
                         {error && <ErrorNotice error={error} />}
-                        {show ? (
+                        {artifact.deleted ? (
+                            <p>Artifact has been deleted.</p>
+                        ) : show ? (
                             <ViewBox>
```
juliev0, line 83: This is interesting. It's actually doing the work that is from this issue, which I worked on and have a PR for: https://github.com/argoproj/argo-workflows/issues/8666

Alex's point of view was that we shouldn't just go by the 'deleted' flag but instead we should go to S3 to determine whether it's actually been deleted or not. So, in that PR basically if the backend returns a 404 (FileNotFound) error, then it shows "File not found", and he thought that was sufficient, not requiring UI changes.

Still, it seems reasonable to me that if GC deleted the Artifact and set the deleted flag, then the only thing that would cause the Artifact to re-exist would be a new run of that Workflow, and in that case it seems reasonable to call the original Artifact from the original Workflow 'deleted'. So, maybe this is fine.

ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
```diff
@@ -53,6 +53,13 @@ export const ArtifactPanel = ({
     }, [downloadUrl]);
     useCollectEvent('openedArtifactPanel');
 
+    const spec = execSpec(workflow);
+    const artifacts = spec.templates.find(t => t.name === workflow.status.nodes[artifact.nodeId].templateName)?.outputs?.artifacts;
+    let artifactGCStrategy = '';
+    if (artifacts !== undefined) {
+        artifactGCStrategy = artifacts.find(a => a.name === artifact.name).artifactGC?.strategy || spec.artifactGC?.strategy;
+    }
+
```
juliev0, line 62: Wow, you already seem like a natural in Typescript.

dpadhiar: I can't take all the credit for this haha - all thanks to @alexec extensive PoC.

ui/src/app/workflows/components/workflow-details/artifact-panel.tsx
```diff
@@ -53,6 +53,13 @@ export const ArtifactPanel = ({
     }, [downloadUrl]);
     useCollectEvent('openedArtifactPanel');
 
+    const spec = execSpec(workflow);
+    const artifacts = spec.templates.find(t => t.name === workflow.status.nodes[artifact.nodeId].templateName)?.outputs?.artifacts;
+    let artifactGCStrategy = '';
```
juliev0, line 58: I wonder if we need to do error handling for artifact.nodeId not being found in workflow.status.nodes (which I recognize probably shouldn't happen unless there's a bug somewhere). @alexec Do you have an opinion? Maybe the result is an error in the console in either case?

juliev0 (commented):
Looks good overall. Since this PR addresses deleted artifacts, just like [this](https://github.com/argoproj/argo-workflows/pull/8966/files) one does, once one of them is merged the other one will need to re-test to make sure everything's working nicely together. (The other PR deals with artifacts that are deleted in S3 (either by GC or manually) while this one deals with those that were deleted by GC). What should happen is that your 'Artifact has been deleted' text will show if GC happened, but if they manually deleted the artifact themselves then the 404 response will show. The file download link will be hidden in the former case but not the latter. (It should ideally be hidden in the latter case but Alex didn't think it was a big deal - we could potentially address it later.)

juliev0:
@dpadhiar Sorry mine got merged before yours. Mine basically returns an HTTP status code of 404 ("File not found") [here](https://github.com/argoproj/argo-workflows/blob/master/server/artifacts/artifact_server.go#L222) when it's called from [here](https://github.com/argoproj/argo-workflows/blob/master/ui/src/app/workflows/components/workflow-details/artifact-panel.tsx#L48) or [here](https://github.com/argoproj/argo-workflows/blob/master/ui/src/app/workflows/components/workflow-details/artifact-panel.tsx#L85), if it doesn't find the file in S3. 
If you don't mind testing it, I think we can consider the following cases:
1. Artifact has been deleted by GC, and doesn't exist in S3 (your error should appear)
2. Artifact has been deleted by GC, and does exist in S3 - say somebody ran a new Workflow which created it (your error should appear)
3. Artifact was not deleted by GC, but doesn't exist in S3 (somebody deleted it manually) - should see "File not found" error

To manually change a Workflow to add the Artifact.Deleted flag, use "kubectl patch"
To delete files from AWS S3, locally you will use minio to stand in for S3. In your browser open localhost:9001 and login as username: "admin", password: "password"

sarabala1979 (approved):

PR #9064, opened 2022-06-27
docs: BackUp Command Update

reddymh:
Signed-off-by: Rajshekar Reddy [reddymh@gmail.com](mailto:reddymh@gmail.com)

Current document shows command takes back up only on default namespace not from all the namespace


sarabala1979:
@reddymh Please do DCO to fix the build failure

reddymh (commented):
Signed-off-by: Rajshekar Reddy (reddymh@gmail.com)

tczhao:
> Thanks for this. Just to make sure, what happens if you pass in "-A" but don't have access to other namespaces? Does it just report the namespaces you do have read access to?

The namespaced will use Role/ClusterRole rbac, -A will use ClusterRole rbac
If the account doesn't have cluster scope access, it will get blocked, so -A either returns all or none

tczhao:
@reddymh to fix DCO you need to 
```
git commit -s --amend
git push -f
```

juliev0:
> > Thanks for this. Just to make sure, what happens if you pass in "-A" but don't have access to other namespaces? Does it just report the namespaces you do have read access to?
> 
> The namespaced request will use Role/ClusterRole rbac, -A will use ClusterRole rbac If the account doesn't have cluster scope access, it will get blocked, so -A either returns all or none

Thanks. In that case, should we include both options depending on whether all namespaces are required @reddymh ?

reddymh:
@juliev0  current command will only work for default namespace or I can add cluster wide with -A option as well as namespace back up

reddymh:
> @reddymh to fix DCO you need to
> 
> ```
> git commit -s --amend
> git push -f
> ```

I used the same command now for pushing the changes but still same issue.

alexec (approved):

PR #9069, opened 2022-06-28
docs: Artifact GC proposal

juliev0:


agilgur5:
This was replaced by #9070 and then #9145

Issue #8998, opened 2022-06-17
Remove argo-dataflow pipelines component from the UI


terrytangyuan:
# Summary

What change needs making?

People ask a lot about what this is and why it always errors out on this page. Given that argo-dataflow is still in argoproj-labs and under major refactoring/updates (possibly no longer part of Argo?), we should consider removing it from the UI.

WDYT? @alexec 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

alexec:
Yes. I think this is a dupe?

terrytangyuan:
https://github.com/argoproj/argo-workflows/issues/7088 is a more general request. I can take this specific one though.

alexec:
Not a dupe. I think most of the original work would have been done in a57b3ad9e. Try reverting that. You'll find a lot of conflicts, prefer what is on `master`. The search for `dataflow`, `Dataflow`, `pipeline` to see what was missed.

louisnow:
@alexec @terrytangyuan
We currently use argo dataflow pipelines and rely on it being present and usable within argo.
Does this change mean we can't use argo dataflow pipelines with Argo 3.4+ ?

terrytangyuan:
Yes you'll have to use it separately.

PR #9071, opened 2022-06-28
chore!: Remove dataflow pipelines from codebase

terrytangyuan:
Closes https://github.com/argoproj/argo-workflows/issues/8998.


terrytangyuan:
> (a) Fix lint. (b) I'm not sure if this is is breaking change or not. Dataflow never went to v1.0. Either way, it should probably be mentioned in the upgrading guide.

Good point. Updated the docs.

alexec (approved):

louisnow:
@alexec @terrytangyuan 
We currently use argo dataflow pipelines and rely on it being present on the UI.
Does this change mean we can't use argo dataflow pipelines with Argo 3.4+ ?

PR #9086, opened 2022-06-30
feat: add token auth mode. Fixes #9090

### LINKED ISSUE #9090: Static token authentication mode
# Summary
This would implement a new authentication method using tokens from a specific secret.

# Use Cases
This could be useful for automatically managing workflows from a different cluster or namespace.
Right now,
+ `--auth-mode server` will allow to bypass any other authentication method by design.
+ `--auth-mode client` requires access to a ServiceAccount-s secret token, which can be complicated from another cluster or even namespace.
+ `--auth-mode sso` will require a managed application and a flow to generate a token in order to authenticate.

---

LoricAndre:
Signed-off-by: Loric ANDRE <loric.a@padoa-group.com>
Fixes #9090

This implements a new server authentication mode, using a kube secret for static Bearer token authentication.
We require this change for cross-cluster workflow submit from other workflows without server auth mode. See #9090 for details
This is linked to [this chart update PR](https://github.com/argoproj/argo-helm/pull/1352)


alexec:
Any new auth method impact security and therefore has a higher bar to acceptance. A quick visual inspection of the code tells me that I don't think it will work when your run `argo server --auth-mode=sso --auth-mode=token`.

Before we review this PR, I did not see any enhancement proposal. Can I please ask you to create an issue and explain the use base and get a few 👍 to show that people want this?

LoricAndre:
You're right about the SSO+Token, the previous code made them incompatible. It is fixed now, and I added a test to verify it.

config/controller.go
```diff
@@ -47,7 +47,8 @@ func parseConfigMap(cm *apiv1.ConfigMap, config *Config) error {
 			}
 		}
 	}
-	err := yaml.UnmarshalStrict([]byte(rawConfig), config)
+	// err := yaml.UnmarshalStrict([]byte(rawConfig), config)
```
alexec, line 50: revert

config/config.go
```diff
@@ -105,6 +105,9 @@ type Config struct {
 
 	// SSO in settings for single-sign on
 	SSO SSOConfig `json:"sso,omitempty"`
+
+	// Secret name for token auth mode
+	TokenSecretName string `json:"tokenSecretName,omitempty"`
```
alexec, line 110: I think this should be

```yaml
token:
  secretName: abc
```

The secret name should have a default value, e.g. `argo-auth-tokens`

config/controller_test.go
```diff
@@ -30,9 +30,9 @@ func Test_parseConfigMap(t *testing.T) {
 			assert.NotEmpty(t, c.ArtifactRepository)
 		}
 	})
-	t.Run("Garbage", func(t *testing.T) {
-		c := &Config{}
-		err := parseConfigMap(&apiv1.ConfigMap{Data: map[string]string{"garbage": "garbage"}}, c)
-		assert.Error(t, err)
-	})
+	// t.Run("Garbage", func(t *testing.T) {
```
alexec, line 33: revert

config/controller.go
LoricAndre: Yes, this is on the Todo before merging, this was set for our testing.

alexec:
I think this is a good idea. I do think that we need the following:

1. This should be self-serve, I.e. look in the namespace of the request for the secret. This will allow teams in multi-tentant mode to self-serve.
2. Docs.

LoricAndre:
@alexec about the self-serve part, implementing this would then require secrets in every workflow's namespace, instead of just the server's. This seems pretty heavy to me, I can implement a toggle for this but I'm not sure this is what you mean.

alexec:
What we do for similar self-serve:

* Is there secret in user namespace? If so, use that.
* Otherwise, is there secret in install namespace.

Makes sense?

LoricAndre:
I'm not sure I get what you mean by 'user namespace', since there is no user with this auth method. I can use the request's namespace instead, but this doesn't seem very useful in itself.

alexec:
You're correct. I mean "request namespace", not "user namespace".

alexec:
This PR is still a great idea as it makes automation easier and more secure.

PR #9093, opened 2022-07-02
fix: need to update artifact according to revised location

### LINKED ISSUE #9092: wait container doesn't update Artifact location in Workflow Status if Template.ArchiveLocation is specified
If I set Template.ArchiveLocation, the wait container doesn't seem to reflect that location in the NodeStatus. I believe it is a change that needs to be made in executor.go on this [line](https://github.com/argoproj/argo-workflows/blob/master/workflow/executor/executor.go#L342).

```yaml
kind: Workflow
metadata:
  generateName: artifact-gc-
spec:
  entrypoint: main
  artifactGC:
    strategy: OnWorkflowDeletion
  podGC:
    strategy: ""
  templates:
    - name: main
      archiveLocation:
        s3:
          key: on-deletion-key
          bucket: my-bucket-2
          endpoint: minio:9000
          insecure: true
          accessKeySecret:
            name: my-minio-cred
            key: accesskey
          secretKeySecret:
            name: my-minio-cred
            key: secretkey
      container:
        image: argoproj/argosay:v2
        command:
          - sh
          - -c
        args:
          - |
            echo "hello world" > /tmp/message
      outputs:
        artifacts:
          - name: on-deletion
            path: /tmp/message
```

juliev0:

Fixes #9092

### DIFF
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -180,28 +180,14 @@ spec:
 			Then().
 			ExpectWorkflow(func(t *testing.T, m *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
 				n := status.Nodes[m.Name]
-				expectedOutputs := &wfv1.Outputs{
-					Artifacts: wfv1.Artifacts{
-						{
-							Name: "a-logs",
-							ArtifactLocation: wfv1.ArtifactLocation{
-								S3: &wfv1.S3Artifact{
-									Key: fmt.Sprintf("%s/%s/a.log", m.Name, m.Name),
-								},
-							},
-						},
-						{
-							Name: "b-logs",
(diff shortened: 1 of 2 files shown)


juliev0:
@alexec Can you confirm that I am correct that executor.go should in fact update the artifact's S3 Location information to reflect the revised location? 

My confusion is [this](https://github.com/argoproj/argo-workflows/blob/master/test/e2e/functional_test.go#L383) test line which fails if I make this change. In this case, there's an artifactRepositoryRef which indicates the bucket. Should the Artifact in the NodeStatus not be updated to reflect this information?

juliev0:
I talked to Bala, and there's an intention not to expand the NodeStatus with this Artifact location information in each artifact. So, instead, the Artifact GC code will need to look for the artifacts in the correct locations. Closing this PR.

PR #9115, opened 2022-07-05
feat: report Artifact GC failures in user interface. Fixes #8518

### LINKED ISSUE #8518: Artifact GC: report GC failures
One thing missing in the PoC is a way to feedback to the user that artifact GC has failed. It only logs it. Most users, that do not have access to those logs. Instead, use `/status/conditons` to report errors:

https://github.com/argoproj/argo-workflows/blob/e368f504307ecbbfa2f228bccfab1f943efaca9d/pkg/apis/workflow/v1alpha1/workflow_types.go#L1698

```
    Status:      True
    Type:         ArtifactGCError
    Message: "such and such error occurred"
```

Make the error in the user interface. This is an important error, so display in the same way as "Workflow deleted".

dpadhiar:
Fixes #8518


ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -255,6 +255,8 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
             e => {
                 if (e.type === 'DELETED') {
                     setError(new Error('Workflow deleted'));
+                } else if (hasArtifactGCError(e.object.status.conditions)) {
+                    setError(new Error('Artifact garbage collection failed'));
```
qhuai, line 259: `conditions` may also include MetricsError, SubmissionError, SpecError, etc. Is the artifact GC error more special than other errors so that we only need to report the former?

dpadhiar: There may be a different place to surface this error, this is the implementation I went with to start since Alex specified this error should show the same way we show when an artifact is deleted.

ui/src/app/shared/conditions-panel.tsx
```diff
@@ -26,6 +26,20 @@ export function hasWarningConditionBadge(conditions: Condition[]): boolean {
     return false;
 }
 
+export function hasArtifactGCError(conditions: Condition[]): boolean {
+    if (conditions.length === 0) {
+        return false;
+    }
+
```
juliev0, line 33: this "if" statement could probably be removed, right? (since it would be taken care of below)

juliev0: hmm, I'd be surprised but could be a peculiarity of the language I suppose

qhuai: I think the for-loop will cover the case when conditions is nil or if its length is 0. Check out [Go spec on RangeClause](https://go.dev/ref/spec#RangeClause). In addition, a simple "hello world" test can confirm the case.

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
qhuai: If it's not the right place, where is that different place to surface this error?

ui/src/app/shared/conditions-panel.tsx
dpadhiar: According to this, a nil slice would cause 0 iterations, therefore we should be able to remove this check in both functions. Testing now.

juliev0: @qhuai You may be right, but this is Typescript actually, not Go. :)

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
dpadhiar: I will have to surface this error in a different block of code. Currently, this implementation will keep the workflow diagram from appearing when the error appears which is expected for a deleted workflow but not failed artifact garbage collection.

juliev0 (approved):
Looks good. Only thing I'm seeing is that we could theoretically have a constant to represent the 'ArtifactGCError' string, and then we could test against that instead of the literal string, which is basically what we do in the Go code (reduces likelihood of misspelling in the check), but I'm not sure if Alex does that anywhere else or not. I've given approval.

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -255,6 +255,9 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
...
+                    setWorkflow(e.object);
```
qhuai, line 260: Nitpick. An alternative way to avoid duplicating the setWorkflow step and thus make the overall logic slightly clearer:
```
else {
    if (hasArtifactGCError(e.object.status.conditions)) {
        setError(new Error('Artifact garbage collection failed'));
    }
    setWorkflow(e.object);
}
```

ui/src/app/shared/conditions-panel.tsx
```diff
@@ -26,6 +22,16 @@ export function hasWarningConditionBadge(conditions: Condition[]): boolean {
     return false;
 }
 
+export function hasArtifactGCError(conditions: Condition[]): boolean {
+    for (const condition of conditions) {
+        if (condition.type === 'ArtifactGCError') {
```
qhuai, line 27: Nitpick. If we want to report all errors (not just ArtifactGCError), this function could be
```
function getErrorMessages(conditions: Condition[]): Error[] {
    let errs = Error[]
    for (const condition of conditions) {
        if (ErrorConditions.includes(condition.type) {
            errs.push(ErrorMessages[condition.type]);
        }
    }
    return errs;
}
```

qhuai (approved):
Overall, LGTM

juliev0 (approved):

PR #9118, opened 2022-07-06
fix: support RemainingItemCount in archivedWrokflow

hadesy:
Signed-off-by: hadesy <6346047+hadesy@users.noreply.github.com>

Fixes support RemainingItemCount in archivedWrokflow


juliev0 (approved):
Looks good. 

I guess the number of workflows can change between the calls to ListWorkflows() and CountWorkflows() but since metav1.ListMeta.RemainingItemCount is only supposed to be treated as an estimate, that's totally fine (as you probably rationalized).

tczhao (changes requested):
@hadesy Could you fix the generated code
```
make -B codegen
```

PR #9119, opened 2022-07-06
feat(ui): Add workflow author information to workflow summary and drawer

terrytangyuan:


alexec (approved):
Seems ok. Could use a functional component rather than class, but otherwise fine

Issue #9024, opened 2022-06-22
build timeout


alexec:
The `functional` test suite is hitting is 20m limit. 20m is a long time for a test suite to run because it means that the engineer does not have fast enough feedback. IMHO less than 10m is a good target time.

 This usually happens when:

1. A slow test has been introduced. 
2. A build step has become slow due to problems with cache.
3. Incremental new tests push the time limit.

If it is (2), then all builds become slower. The fastest build takes about 4m20. I think that's pretty fast so it is probably not (2).

Therefore, I think it is probably (1). 

Actions:

* Look at the build and see if anything is slow (look for "slow" in the logs). Anything systemic?
* Are there just too many tests and maybe we should break some out into a new suite?

juliev0:
Added this one to address intermittent failures as well: https://github.com/argoproj/argo-workflows/issues/9027

PR #9150, opened 2022-07-14
test: increase e2e timeout for ci

dpadhiar:
Currently, it is possible for multiple e2e test to fail when making pull requests when the code is not changed at all. we assume this is because the tests timeout with the current 15 minute threshold. We should increase this threshold to mitigate these intermittent failures to remove the need to rerun tests constantly until they all pass. We can also increase the Github action timeout in order to stop tests from simply cancelling when exceeding 20 minutes.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -151,7 +151,7 @@ jobs:
       - run: make start PROFILE=${{matrix.profile}} AUTH_MODE=client STATIC_FILES=false LOG_LEVEL=info API=${{matrix.test == 'test-api' || matrix.test == 'test-cli' || matrix.test == 'test-java-sdk' || matrix.test == 'test-python-sdk'}} UI=false > /tmp/argo.log 2>&1 &
       - run: make wait
         timeout-minutes: 4
-      - run: make ${{matrix.test}} E2E_TIMEOUT=1m STATIC_FILES=false
+      - run: make ${{matrix.test}} E2E_SUITE_TIMEOUT=15m STATIC_FILES=false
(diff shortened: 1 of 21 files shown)


terrytangyuan (commented):
Has the build been passing stably though?

dpadhiar:
> Has the build been passing stably though?

This is the first run I have done on Github since @juliev0 showed me we could adjust the Github action timeout. We could continue to manually rerun this PR to see if it has any random failures, we would also know if this solution works if we see how it affects PRs over the next day or so if it were to be merged.

terrytangyuan:
I'd suggest we trigger empty commits with `--allow-empty` to see if this solves the problem consistently. Otherwise, we will need to bump timeout again.

juliev0:
> I'd suggest pushing empty commits with `--allow-empty` and trigger the build to see if this solves the problem consistently. Otherwise, we will need to bump timeout again.

Hey Terry. @sarabala1979 and I talked about bumping it up to 25m for now. Ultimately, we need to address [this](https://github.com/argoproj/argo-workflows/issues/9024) issue that the tests are taking too long. But the current state of things seems to be pretty aggravating. I encountered this timeout issue twice in a row and then Dillen did as well.

juliev0 (approved):

terrytangyuan:
Sure happy to merge once it passes again.

terrytangyuan (approved):

dpadhiar:
@terrytangyuan Had 1 fail so far but this looks like it wasn't related to any of the other intermittent failures at all.

juliev0:
> Sure happy to merge once it passes again.

thank you!

PR #7605, opened 2022-01-20
fix: Use v1 pod name if no template name or ref. Fixes #7595 and #7749

JPZ13:
Signed-off-by: J.P. Zivalich <jp@pipekit.io>

This PR:
- Falls back to the v1 pod names in the UI if no templateName or templateRef is set on a given node - my hypothesis is that certain virtual nodes might not have a a templateName or templateRef set as seen here: https://github.com/argoproj/argo-workflows/issues/7595#issuecomment-1017776672
- Fixes #7595 

@liuzqt Could you run one of your workflows locally using this PR branch to test if it is working? I want to validate that I don't have to make an additional change to the workflow controller

PR #9168, opened 2022-07-18
fix: Workflow details accessing undefined templateRef. Fixes #9167

### LINKED ISSUE #9167: UI: `e.templateRef is undefined` when using inline template
## Checklist

 

* [ ] Double-checked my configuration.
* [ ] Used the Emissary executor.

## Summary

Creates a simple workflow with inline template. It should succeed.

What version are you running?
Argo v3.2.11

## Diagnostics

From https://raw.githubusercontent.com/argoproj/argo-workflows/master/examples/steps-inline-workflow.yaml:
```yaml
kind: Workflow
metadata:
  generateName: steps-inline-
  labels:
  annotations:
      This workflow demonstrates running a steps with inline templates.
    workflows.argoproj.io/version: ">= 3.2.0"
spec:
  entrypoint: main
  templates:
    - name: main
      steps:
        - - name: a
            inline:
              container:
                image: argoproj/argosay:v2
```

---

terrytangyuan:
Fixes #9167. This is a regression and needs to be backported.

### DIFF
--- a/ui/src/app/shared/pod-name.ts
+++ b/ui/src/app/shared/pod-name.ts
@@ -55,7 +55,7 @@ export const getTemplateNameFromNode = (node: NodeStatus): string => {
     }
 
     // fall back to v1 pod names if no templateName or templateRef defined
-    if (!node.templateRef) {
+    if (node?.templateRef === undefined || node?.templateRef.template === '') {
         return '';
     }
 


alexec:
We do not back-port fixes to unsupported versions. Please upgrade to v3.3

ui/src/app/shared/pod-name.ts
```diff
@@ -55,7 +55,7 @@ export const getTemplateNameFromNode = (node: NodeStatus): string => {
     }
 
     // fall back to v1 pod names if no templateName or templateRef defined
-    if (!node.templateRef) {
+    if (node?.templateRef === undefined || node?.templateRef.template === '') {
```
agilgur5, line 58: the `node?.` part is not necessary; the `node.templateName` line above this would error out before reaching this line if it didn't exist.

#7605 did fix this, it just never got backported to 3.2.x. So #9167 was a pure duplicate.

This whole function can actually be simplified into a one-liner, which I'm working on.

agilgur5: Simplified in #12964

PR #9180, opened 2022-07-19
docs: remove false line from retry docs

nielstenboom:
This small PR removes a line from the docs that threw me off for a couple of hours today. What the docs state here is false unfortunately, therefore this PR removes it from the docs.

Can be tested with the following examples:

```yaml
retryStrategy:
  retryPolicy: "Always"
  expression: 'lastRetry.status == "Failed" && asInt(lastRetry.exitCode) not in [-1,0,1] or lastRetry.status == "Error"'
  limit: 3
```
Submit a simple workflow with above retrypolicy and delete the pod while it's running, the workflow will do a retry.

```yaml
retryStrategy:
  expression: 'lastRetry.status == "Failed" && asInt(lastRetry.exitCode) not in [-1,0,1] or lastRetry.status == "Error"'
  limit: 3
```
Submit a simple workflow again with above retrypolicy and delete the pod while it's running, the workflow will fail immediately.


juliev0 (commented):
@jmcarp Sorry to bother you, but do you have bandwidth to review this small change? It looks like you added this file originally and probably know the most about it. Thanks.

alexec (approved):

PR #9184, opened 2022-07-19
feat: adding new CRD type "ArtifactGCTask" 

juliev0:
Fixes: #9200 

New CRD type is for Controller to provide information about artifacts that need deleting for Artifact GC Pod to read and write Status to.


pkg/apis/workflow/v1alpha1/artifact_gc.go
```diff
@@ -0,0 +1,52 @@
+package v1alpha1
```
sarabala1979, line 1: follow naming convention.
artifact_gc_types.go

pkg/apis/workflow/v1alpha1/artifact_gc.go
```diff
@@ -0,0 +1,52 @@
...
+type ArtifactGCSpec struct {
+	ArtifactsByNode map[string]ArtifactNodeSpec `json:"artifactsByNode,omitempty" protobuf:"bytes,1,rep,name=artifactsByNode"`
+}
+
+type ArtifactNodeSpec struct {
+	ArchiveLocation *ArtifactLocation       `json:"archiveLocation,omitempty" protobuf:"bytes,1,opt,name=archiveLocation"`
+	Artifacts       map[string]ArtifactList `json:"artifacts,omitempty" protobuf:"bytes,2,rep,name=artifacts"`
+}
+
+type ArtifactList struct {
```
sarabala1979, line 27: Do you really need Artifactlist?

pkg/apis/workflow/v1alpha1/artifact_gc.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type ArtifactResultNodeStatus struct {
+	ArtifactResults []ArtifactResult `json:"artifactResults,omitempty" protobuf:"bytes,1,rep,name=artifactResults"`
+}
+
+type ArtifactResult struct {
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+
+	Deleted bool `json:"deleted,omitempty" protobuf:"varint,2,opt,name=deleted"`
```
sarabala1979, line 42: will status make sense instead `deleted` ?
status=failed/deleted

sarabala1979 (changes requested):

pkg/apis/workflow/v1alpha1/artifact_gc.go
juliev0: I initially tried not using it and just using "Artifacts" there but I guess because "Artifacts" doesn't have a json definition associated with it, the .proto generator turned it into a map<string, Artifact>.

juliev0: I did try making json "inline" but that seemed to fail compilation as well so I went back to this version.

pkg/apis/workflow/v1alpha1/artifact_gc.go
juliev0: I guess it would be nice if we could still have a boolean type rather than a new type. Maybe "Failed" as a boolean? What do you think? We also have the "Error" there which will be set if it fails.

pkg/apis/workflow/v1alpha1/artifact_gc.go
juliev0: will do

pkg/apis/workflow/v1alpha1/artifact_gc.go
juliev0: Actually, I realized I had a bug there and that instead of having a "map[string]ArtifactList" I actually do want a "map[string]Artifact" anyway :) so I can remove that extra struct.

pkg/apis/workflow/v1alpha1/artifact_gc.go
juliev0: I changed it to a "Success" boolean

pkg/apis/workflow/v1alpha1/artifact_gc_task_types.go
```diff
@@ -0,0 +1,48 @@
...
+type ArtifactGCSStatus struct {
+	ArtifactResultsByNode map[string]ArtifactResultNodeStatus `json:"artifactResultsByNode,omitempty" protobuf:"bytes,1,rep,name=artifactResultsByNode"`
+}
+
+type ArtifactResultNodeStatus struct {
+	ArtifactResults map[string]ArtifactResult `json:"artifactResults,omitempty" protobuf:"bytes,1,rep,name=artifactResults"`
+}
+
+type ArtifactResult struct {
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
```
sarabala1979, line 36: @juliev0 Can you add comments for these fields?

manifests/quick-start-minimal.yaml
```diff
@@ -908,6 +908,7 @@ rules:
   - workflows/finalizers
   - workflowtasksets
   - workflowtasksets/finalizers
+  - artifactgctasks
```
alexec, line 911: For consistency`workflowartifactgctaskset`

juliev0: do you like "workflowartifactgctaskset" better than "workflowartifactgctask"? I was thinking this was more of a task than a task set.? (WorkflowTaskSet, on the other hand, actually defines a list of Tasks)

juliev0: changed to "workflowartifactgctaskset"

juliev0: changed to _workflowartifactgctaskset_

juliev0: back to _workflowartifactgctask_ since you said you were open to that in a separate conversation (thank you)

sarabala1979 (approved):

PR #9192, opened 2022-07-20
ci: Pin golangci-lint to v1.47.1

terrytangyuan:
See context in https://github.com/argoproj/argo-workflows/pull/9186. 


juliev0 (approved):
Seems reasonable, to keep this from happening again. I suppose we won't benefit from any future checks that are added, but with all of the false failures in the CI, it seems like a reasonable trade off to me.

sarabala1979 (approved):

PR #9213, opened 2022-07-22
fix: retryStrategy.Limit is now read properly for backoff strategy. Fixes #9170.

### LINKED ISSUE #9170: `backoff` under `retryStrategy` does not work as expected for version 3.3.8
The `backoff` feature for `retryStrategy` does not behave as expected under version `3.3.8`. 
```
      retryStrategy:
        limit: "10"
        backoff:
          duration: "30m"   
```
The workflow was retried 10 times, but the wait time between each retry call was about 21 seconds. 

Tested under version `3.2.4`, and it worked as expected.

dpadhiar:

Fixes #9170 

Previously, this if statement always evaluated to false if you set a `Limit` for `RetryStrategy` as the `retryStrategy.Limit.IntVal` always equaled 0. The appropriate way to get the integer value from a `intstr.IntOrString` is to use `IntValue` which returns an int we cast to `int32`. 

```
if time.Now().Before(waitingDeadline) && retryStrategy.Limit != nil && int32(len(node.Children)) <= 
    int32(retryStrategy.Limit.IntValue()) {
    woc.requeueAfter(timeToWait)
    retryMessage := fmt.Sprintf("Backoff for %s", humanize.Duration(timeToWait))
    return woc.markNodePhase(node.Name, node.Phase, retryMessage), false, nil
}
```


test/e2e/retry_test.go
```diff
@@ -48,6 +49,35 @@ spec:
...
+          image: 'argoproj/argosay:v2'
+          args: [ exit, "1" ]
+`).
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(time.Minute).
+		Then().
+		ExpectWorkflow(func(t *testing.T, _ *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			assert.Equal(t, wfv1.WorkflowPhase("Failed"), status.Phase)
+			assert.Equal(t, "Max duration limit exceeded", status.Message)
```
tczhao, line 77: I think you need to validate the number of retries since the issue is "string" `limit` does not get interpret correctly

workflow/controller/operator.go
```diff
@@ -898,7 +898,7 @@ func (woc *wfOperationCtx) processNodeRetries(node *wfv1.NodeStatus, retryStrate
 		}
 
 		// See if we have waited past the deadline
-		if time.Now().Before(waitingDeadline) && retryStrategy.Limit != nil && int32(len(node.Children)) <= retryStrategy.Limit.IntVal {
+		if time.Now().Before(waitingDeadline) && retryStrategy.Limit != nil && int32(len(node.Children)) <= int32(retryStrategy.Limit.IntValue()) {
```
tczhao, line 901: 👍

tczhao (changes requested):

test/e2e/retry_test.go
dpadhiar: In this case, we could check the length of child nodes at the end of the run which should be 5 based on the backoff time.

test/e2e/retry_test.go
```diff
@@ -48,6 +49,36 @@ spec:
...
+	s.Given().
+		Workflow(`
+metadata:
+  generateName: test-backoff-strategy-
+spec:
+  entrypoint: main
+  templates:
+    - name: main
+      retryStrategy:
+        limit: 10
```
tczhao, line 62: I should've pointed this out last time, since the error is with the intstring conversion maybe we should test both `'10'` and `10`

dpadhiar: Will add a separate workflow to test this, good idea.

tczhao (approved):

sarabala1979 (approved):

tanken:
am on v3.3.9 but backoff doesnt seem to apply any delays between retries

PR #9241, opened 2022-07-27
fix: main oom but wait stuck in running state

### LINKED ISSUE #9083: Wait container cannot exit if main container is OOMKilled

wait container cannot exit if main container is OOMKilled

Use Case: Start a workflow, then let the main container oom, and find that the wait container does not exit.

截屏2022-06-30 上午10 47 17

截屏2022-06-30 上午10 48 03

Reason: I read the source code and found that the wait container decides whether to exit by checking if the /var/run/argo/ctr/main/exitcode file exists, and the main container calls the defer function in the NewEmissaryCommand function to create the file. But when the main container is OOMKilled, the main container crashes directly, the defer function will not be called, and the wait container will not exit.

What version are you running?
v3.3.5

## Diagnostics

```yaml
kind: Workflow
metadata:
  name: dag-test2
  namespace: icdc-test
spec:
  arguments: {}
  entrypoint: scheduler-alpha-entrypoint
  podSpecPatch: '{"terminationGracePeriodSeconds": 0, "enableServiceLinks": false}'
  serviceAccountName: argo-workflow
  templates:
    - name: A
      container:
        image: ubuntu
        resources:
          requests:
            cpu: 1
            memory: 10Ki
          limits:
            cpu: 1
            memory: 50Mi
        command: ["/bin/bash"]
        args: ["-c","for x in {1..200}; do echo 'Round $x'; bash -c 'for b in {0..99999999}; do a=$b$a; done'; done"]
    - name: scheduler-alpha-entrypoint
      dag:
        tasks:
          - arguments: { }
            name: C
            template: A
      inputs: {}
      metadata: {}
      outputs: {}

```

```bash

```

---

firetaker:
Fixes #9083
Test on v3.3.5 and v3.3.8
PTAL.
Thanks

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -510,6 +510,26 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 			if err != nil && !apierr.IsNotFound(err) {
 				return err
 			}
+		case remarkExitCode:
+			innerPod, err := wfc.getPod(namespace, podName)
+			if innerPod == nil || err != nil {
+				return err
+			}
+			for _, c := range innerPod.Spec.Containers {
+				if c.Name == common.WaitContainerName {
+					//exec cmd  /bin/sh -c "echo 137 > /var/run/argo/ctr/main/exitcode"
+					toPath := common.VarRunArgoPath + "/ctr/" + common.MainContainerName + "/exitcode"
+					command := []string{"/bin/sh", "-c", fmt.Sprintf("echo 137 > %s", toPath)}
+					if err := signal.ExecPodContainerAndGetOutput(wfc.restConfig,
+						innerPod.Namespace, innerPod.Name,
+						common.WaitContainerName, command...); errorsutil.IgnoreContainerNotFoundErr(err) != nil {
+						logCtx.WithError(err).Warn("failed to force terminate wait pod")
+						return err
+					}
+					log.WithFields(log.Fields{"key": "pod", "name": innerPod.Name, "cmd": command}).Info("force terminate wait pod")
+					return nil // done
+				}
+			}
 		}
 		return nil
 	}()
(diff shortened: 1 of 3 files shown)


workflow/controller/controller.go
```diff
@@ -510,6 +510,26 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
+		case remarkExitCode:
+			innerPod, err := wfc.getPod(namespace, podName)
+			if innerPod == nil || err != nil {
+				return err
+			}
+			for _, c := range innerPod.Spec.Containers {
+				if c.Name == common.WaitContainerName {
+					//exec cmd  /bin/sh -c "echo 137 > /var/run/argo/ctr/main/exitcode"
+					toPath := common.VarRunArgoPath + "/ctr/" + common.MainContainerName + "/exitcode"
+					command := []string{"/bin/sh", "-c", fmt.Sprintf("echo 137 > %s", toPath)}
```
alexec, line 522: this cannot work as `sh` is not installed on the wait container anymore

firetaker: when main container been killed by OOM-killer(cgroup or host os), no 'exitcode' file wirte to ctr/main dir, but wait container stuck in running state,  up to this condition, just use sh command write a FAKE exitcode file, this FAKE exitode will trigger wait-container exit。

firetaker: this line can work ? 
[command := []string{"/bin/sh", "-c", "kill -%d 1"}](https://github.com/argoproj/argo-workflows/blob/2d1758fe90fd60b37d0dfccb55c3f79d8a897289/workflow/signal/signal.go#L18)

alexec: No. That is not on master anymore

firetaker: thanks for kind reply.

sarabala1979: @firetaker Are you still able to reproduce on Master? if not, can you close this PR?

tooptoop4:
@firetaker dis fixed?

PR #9243, opened 2022-07-27
feat: Implement resource allowance feature (fixes #8982)

### LINKED ISSUE #8982: Resource sharing amongst users at the template level
## Summary

I would like to create workflows on behalf of 3rd party users, and I would like those users to be able to share resources equally at the template level via "user-level" parallelism.

## Potential interface

One instantiation of such a feature could look like this (note that I continue to update this as I experiment more with Argo codebase):

The workflow spec has a `ResourceSharingID` that identifies a single participant who will receive the same number of resources as all other participants (basically, a "user").

A workflow could have this ID passed along like this:

```yaml
kind: Workflow
metadata:
  generateName: pipeline-
spec:
  entrypoint: sg-wrapper
  resourceSharingID: abc-1233333
  templates:
```

but more realistically, when the workflow is created, it'd be passed in dynamically:

```sh
argo submit -l resourceSharingID=abc123333 workflow.yaml
```

Resource limits could be set in the following way. In `workflow-controller-configmap`:

```yaml
  resourcesAvailable: |
    template-name-1: "4"
    template-name-1: "20"

```

Any templates that are not defined here (or are not used in steps / group) will bypass resource sharing checks, (but will still be subject to all other parallelism checks).

## Details / edge cases:

- If parallelism for a particular step is defined, AND a resource limit is set for it's template, the behavior is undefined but the Workflow will submit.

- Workflows that do not contain resourceSharingID will have no limits set on pods with regards to available resources.

## Notes:

- I do not think such behavior exists (trivially) at the moment, but if I have missed it I'd love to be pointed in the right direction! (I have written a cron that adjusts template-level parallelism on an interval, but this approach seems bad for a variety of reasons)

- I have already set up the codebase in a dev container and have been reading/experimenting with it. I'm beginning to get a feel for what would need to be done - I would be more than happy to work with a mentor to bring this feature back upstream (as I need this feature for a project, I'll probably be implementing either way).

## Use Cases

I am currently building software for a render farm that processes long-running workflows (several hours to days). Access to our resources is bottlenecked by a finite number of provisioned software licenses for our farm.

---

robertkotcher:
NOTE: This is a draft and does not yet completely implement the proposed feature. I have implemented in a separate, local, branch, and will move code over as discussions with other maintainers progress. I believe it'll be easier to digest the intention here by starting with a small amount of code and documentation. Planning to move out of draft when this PR represents a complete feature.

See issue https://github.com/argoproj/argo-workflows/issues/8982

So far, the code in this pull request does the following:

* add function that computes resource allowance for arbitrary users/resource

Signed-off-by: Robert Kotcher <robert@synthesis.ai>

### DIFF
--- a/docs/resource-sharing.md
+++ b/docs/resource-sharing.md
@@ -1,0 +1,40 @@
+# Resource Sharing
+
+It could be the case that a particular template execution uses 1 of a finite number of resources. Examples of this could be software licenses or compute nodes, just to name two examples.
+
+## Examples
+
+### Example 1
+
+At a particular moment, we have two users, Alice and Bob, who are trying to render images via Workflows that we have submitted on their behalf. The software running inside the render step’s Template requires access to a license, and licenses are scarce resources. Therefore, at this particular moment, we give Alice and Bob each exactly half of the available licenses. (Executing a template will cause exactly one license to be claimed.)
+
+### Example 2
+
+Somebody is running an in-house cluster with only a few nodes of a particular type. A particular Template’s Pod has a toleration towards these nodes, but there are enough concurrent Pods that just a small number of them will end up blocking other users’ workflows from progressing.
+
+### Example 3
+
+??
+
+## Proposal
+
+Please see related issue [here](https://github.com/argoproj/argo-workflows/issues/8982).
+
+## Discussion
+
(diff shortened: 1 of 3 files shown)


agilgur5:
I believe this was replaced by #9992

PR #9246, opened 2022-07-28
docs: Update RetryStrategy.limit documentation

jobimrobinsantos-drizly:
`limit` is currently ambiguous in documentation. It represents the maximum number of retries that will be attempted in addition to the first attempt. For example, `limit = 1` means two attempts total, `limit = 99` means one hundred attempts total, etc. This change attempts to clarify the ambiguity.

Fixes #TODO


terrytangyuan (commented):
This file is auto-generated. You may want to update `workflow_types.go` and then run `make codegen`

terrytangyuan (approved):

PR #9247, opened 2022-07-28
feat: Add custom event aggregator function with annotations

terrytangyuan:


jessesuen (commented):
Just so I understand the PR.

When retrying workflows, we have an issue such that Kubernetes Events are aggregated (incorrectly) from a previous submission. And so this PR provides a way to prevent that aggregation?

And is the EVENT_AGGREGATION_WITH_ANNOTATIONS just a feature flag?

terrytangyuan:
My apologies. I thought I filled out the PR description with use cases but it somehow wasn't saved. 

> Just so I understand the PR.
> 
> When retrying workflows, we have an issue such that Kubernetes Events are aggregated (incorrectly) from a previous submission. And so this PR provides a way to prevent that aggregation?

When retrying, the events for the same nodes will be aggregated since they have exactly the same information (ID, name, type) that is used by the default event key aggregator `EventAggregatorByReasonFunc`. To avoid that, we introduce a timestamp in the annotations to differentiate the retrying nodes from the previously finished nodes. The custom aggregator function takes the annotations into account so that those events won't be combined. Our users who consume K8s events are able to use the fine-grained events to tell whether the nodes are being re-run and the exact timestamps for those unique events.
> 
> And is the EVENT_AGGREGATION_WITH_ANNOTATIONS just a feature flag?

Yes, we don't want to change the default behavior but also want to support the additional use cases so an environment variable is a way for us to toggle experimental features in Workflows.

workflow/events/event_recorder_manager.go
```diff
@@ -24,14 +27,43 @@ type eventRecorderManager struct {
...
+// including annotation values as part of the event aggregation key.
+func customEventAggregatorFuncWithAnnotations(event *apiv1.Event) (string, string) {
+	var joinedAnnotationsStr string
+	includeAnnotations := env.LookupEnvStringOr("EVENT_AGGREGATION_WITH_ANNOTATIONS", "false")
+	if annotations := event.GetAnnotations(); includeAnnotations == "true" && annotations != nil {
+		var joinedAnnotations = make([]string, 0, len(annotations))
+		for _, v := range annotations {
+			joinedAnnotations = append(joinedAnnotations, v)
+		}
+		joinedAnnotationsStr = strings.Join(joinedAnnotations, "")
```
jessesuen, line 40: I think there is a bug here. Annotations is a map and when iterating maps in golang the order is random. Therefore your string will be different every time.

terrytangyuan: Good catch! Fixed.

jessesuen (approved):
LGTM

alexec:
Could you use the workflow UID instead of timestamp?

terrytangyuan:
> Could you use the workflow UID instead of timestamp?

That only works for resubmitted workflows but for retried workflows the UID does not change.

PR #9267, opened 2022-08-01
Fix: user namespace override (Fixes #9266)

### LINKED ISSUE #9266: Fix: differentiate between no currentNamespace and empty currentNamespace
## Checklist

 

## Summary

After logging in with SSO on the server (using the UI) in the latest master, the namespace is always reset to the user's namespace when the current namespace is empty, but it should stay at cluster level. This is due to https://github.com/argoproj/argo-workflows/pull/9008

What version are you running?
Latest master

## Diagnostics

This is observed from the UI.
---

LoricAndre:
Signed-off-by: Loric ANDRE <loric.a@padoa-group.com>

Fixes https://github.com/argoproj/argo-workflows/issues/9266

### DIFF
--- a/ui/src/app/shared/utils.ts
+++ b/ui/src/app/shared/utils.ts
@@ -103,8 +103,12 @@ export const Utils = {
     },
 
     get currentNamespace() {
-        // we always prefer the user namespace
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):
Thanks!

PR #9275, opened 2022-08-02
docs: Fix inconsistent key-value in memoization example

terrytangyuan:


examples/memoize-simple.yaml
```diff
@@ -37,7 +37,7 @@ spec:
     container:
       image: docker/whalesay:latest
       command: [sh, -c]
-      args: ["cowsay {{inputs.parameters.message}} > /tmp/hello_world.txt"]
+      args: ["cowsay world > /tmp/hello_world.txt"]
     outputs:
```
juliev0, line 41: We don't need to parameterize this anymore? (just trying to follow along since this memoization capability is new to me)

terrytangyuan: Here we only want to parameterize the key in the cache so that we can clearly see that the `memoize.key` is the key in the configmap.

juliev0: So, basically does the template not get run in this case since the output parameter is named "hello" and the "hello" key is memoized?

Unfortunately, I had a hard time comprehending this - I wonder if we should add more [details](https://argoproj.github.io/argo-workflows/memoization/) to the page on memoization?

terrytangyuan: Yes it will be skipped if the cache is already created and not outdated (based on `maxAge`).

juliev0: got it, thanks!

examples/memoize-simple.yaml
```diff
@@ -22,7 +22,7 @@ spec:
   arguments:
     parameters:
```
alexec, line 23: can you delete this?

terrytangyuan: Done. This could simplify the example a bit.

terrytangyuan:
@sarabala1979 This should be ready for another look

alexec (approved):

PR #9280, opened 2022-08-03
fix(gcs): Wrap errors using `%w` to make retrying work

iainlane:
We have a function `isTransientGCSErr()` here which attempts to
`Unwrap()` errors which we have wrapped using `fmt.Errorf()`. This will
only work if the `%w` format string has been used, otherwise no
`Unwrap()` method is created.

This should enable retrying on transient GCS upload errors, and avoid
failures like

```
executor error: upload /tmp/argo/outputs/logs/main.log: writer close:
googleapi: Error 503: We encountered an internal error. Please try
again., backendError
```

Which we've been seeing.


terrytangyuan (commented):
Is this ready for review?

iainlane:
> Is this ready for review?

Just need to get CI to pass

iainlane:
> > Is this ready for review?
> 
> Just need to get CI to pass

And there we go!

terrytangyuan (approved):

PR #9291, opened 2022-08-04
feat: modify "argoexec artifact delete" to handle multiple artifacts. Fixes #9143

### LINKED ISSUE #9143: modify "argoexec artifact delete" to handle multiple artifacts
Implement the logic of the GC Pod according to Option 2 in the slides [here](https://docs.google.com/presentation/d/1LhtWuVSk4bxM_a1-lt5ys2yd-AcdT9F8NhczJ2rVfVU/edit#slide=id.g13af713bf61_0_113).

For this MVP:
- don't worry about retries in the case of failure
- optionally don't worry about having multiple goroutines - it's okay if it runs serially

New CRD type `ArtifactGCWork` is shown [here](https://github.com/juliev0/argo-workflows/blob/art-gc-crd/pkg/apis/workflow/v1alpha1/artifact_gc.go).

1. Read the environment variable "ARGO_ARTIFACT_POD_NAME" (need to add this env var const to workflow/common/common.go). This is how you will know which ArtifactGCWork objects you need to read, as they will be labeled with this.
2. Locate all ArtifactGCWork objects that have label "workflows.argoproj.io/artifact-gc-pod" set to value of "ARGO_ARTIFACT_POD_NAME" (need to add new Label to workflow/common/common.go): should be able to do something like this: 
         ```
        artifactGCTaskInterface := workflowInterface.ArgoprojV1alpha1().WorkflowArtifactGCTasks(namespace)
	labelSelector := fmt.Sprintf("%s = %s", common.LabelKeyArtifactGCPodName, podName)
	
	taskList, err := artifactGCTaskInterface.List(context.Background(), metav1.ListOptions{LabelSelector: labelSelector})
         ```
3. For each ArtifactNodeSpec object in ArtifactsByNode:
          - Create a new ArtifactResultNodeStatus (which you'll add to your ArtifactResultsByNode, keyed by the node name (i.e. same key as ArtifactsByNode))
          - Delete each Artifact (use the ArtifactLocation that's specified in the ArtifactNodeSpec except overridden by any fields specified in the Artifact itself)
          - add each Artifact as an ArtifactResult to the ArtifactResultNodeStatus. If deletion succeeds (it's okay if the key doesn't exist if for some reason it was already deleted), _Success_ should be set to true. Otherwise, set _Success_ to false and _Error_ to the error message.
          - add your ArtifactResultNodeStatus to the ArtifactGCStatus
     - write your ArtifactGCWorkStatus to the ArtifactGCWork CRD object
 4. Return exit code 0 if everything succeeded, else return a non-zero exit code (this should probably happen automatically if you return an error)

---

dpadhiar:
Fixes #9143


cmd/argoexec/commands/artifact/delete.go
```diff
@@ -9,40 +9,112 @@ import (
...
 
+	path := filepath.Join(common.SecretVolMountPath, name, key)
+	if file, ok := r.Files[path]; ok {
+		return string(file), nil
+	}
+
+	file, err := os.ReadFile(path)
+	r.Files[path] = file
+	return string(file), err
 }
```
juliev0, line 118: This is good except I'm thinking about the case where there's an error reading the file on line 115, but then the error doesn't get returned on line 112 later.
Two options I can think of:
1. don't just map to the []byte but instead to a new struct which includes both the []byte and the err
2. in the case of error, don't bother storing it in the map

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -9,40 +9,112 @@ import (
...
+					return err
+				}
 			}
 			return nil
 		},
 	}
 }
 
-type resources struct{}
+func deleteArtifacts(taskList *v1alpha1.WorkflowArtifactGCTaskList, cmd *cobra.Command, artifactGCTaskInterface wfv1alpha1.WorkflowArtifactGCTaskInterface) error {
```
juliev0, line 57: just a minor suggestion, so feel free to take it or leave it: since you're only using the cmd to get the Context from it, you may just want to pass in the Context as an argument. Reasons: 1. Context is often passed in as an argument; 2. it shows you've separated out the concern of what the actual cmd was from this function

juliev0 (changes requested):
made some comments, in-lined

juliev0 (approved):
LGTM!

juliev0:
FYI, for anyone reviewing: for the MVP we kept it simple and didn't do:
- retries in the case of failure
- parallel deletions for performance
These have now been added as Issues to the project.

sarabala1979 (changes requested):
@dpadhiar @juliev0 Why are Workflowartifactgctasks codegen changes in this PR too?

dpadhiar:
> @dpadhiar @juliev0 Why are Workflowartifactgctasks codegen changes in this PR too?

The PR wasn't passing codegen tests unless I included the manifest changes.

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -19,30 +25,104 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+
+				artifactGCTaskInterface := workflowInterface.ArgoprojV1alpha1().WorkflowArtifactGCTasks(namespace)
+				labelSelector := fmt.Sprintf("%s = %s", common.LabelKeyArtifactGCPodName, podName)
 
-			if err := drv.Delete(a); err != nil {
-				return fmt.Errorf("failed to delete artifact: %w", err)
+				taskList, err := artifactGCTaskInterface.List(context.Background(), metav1.ListOptions{LabelSelector: labelSelector})
```
sarabala1979, line 43: minor, you can move this line into deleteArtifacts.

sarabala1979 (approved):

PR #9321, opened 2022-08-08
feat: add argo delete --force. Fixes #9315.

### LINKED ISSUE #9315: Add "argo delete --force" to force deletion of a workflow which has a finalizer
Artifact GC creates a finalizer to prevent Workflows from being deleted until their artifacts have been deleted. But if the artifacts fail to get deleted, the finalizer will remain.

This CLI command would force delete a workflow which has a finalizer, removing the finalizer first. (And thus the ArtifactGCTask, which is owned by it, and the ArtifactGC pod owned by the ArtifactGCTask would all get deleted.)

---

dpadhiar:

Fixes #9315


cmd/argo/commands/delete.go
```diff
@@ -40,6 +43,14 @@ func NewDeleteCommand() *cobra.Command {
 				os.Exit(1)
 			}
 
+			var workflowClientset *workflow.Clientset
+			if force {
+				clientConfig := client.GetConfig()
+				config, err := clientConfig.ClientConfig()
+				errors.CheckError(err)
+				workflowClientset = workflow.NewForConfigOrDie(config)
```
juliev0, line 51: I wonder if we can use the same "serviceAccount" variable that's being used for the deletion instead of needing to add this.

cmd/argo/commands/delete.go
```diff
@@ -22,6 +24,7 @@ func NewDeleteCommand() *cobra.Command {
 		all           bool
 		allNamespaces bool
 		dryRun        bool
+		force         bool
```
juliev0, line 27: make sure this new argument is added on lines 30-31

cmd/argo/commands/delete.go
```diff
@@ -87,5 +103,6 @@ func NewDeleteCommand() *cobra.Command {
 	command.Flags().StringVarP(&flags.labels, "selector", "l", "", "Selector (label query) to filter on, not including uninitialized ones, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2)")
 	command.Flags().StringVar(&flags.fields, "field-selector", "", "Selector (field query) to filter on, supports '=', '==', and '!='.(e.g. --field-selectorkey1=value1,key2=value2). The server only supports a limited number of field queries per type.")
 	command.Flags().BoolVar(&dryRun, "dry-run", false, "Do not delete the workflow, only print what would happen")
+	command.Flags().BoolVar(&force, "force", false, "Force delete workflows by removing finalizers")
```
juliev0, line 106: Do you happen to know why we have some flags which are in the listFlags structure and others that aren't? Maybe the listFlags are the ones that apply to all commands? Just want to make sure this shouldn't be part of that...

juliev0 (commented):
Great job figuring out how to implement this

server/workflow/workflow_server.go
```diff
@@ -298,6 +299,12 @@ func (s *workflowServer) DeleteWorkflow(ctx context.Context, req *workflowpkg.Wo
 	if err != nil {
 		return nil, err
 	}
+	if req.Force {
+		_, err := auth.GetWfClient(ctx).ArgoprojV1alpha1().Workflows(wf.Namespace).Patch(ctx, wf.Name, types.MergePatchType, []byte("{\"metadata\":{\"finalizers\":null}}"), metav1.PatchOptions{})
+		if err != nil {
+			return nil, err
+		}
+	}
 	err = auth.GetWfClient(ctx).ArgoprojV1alpha1().Workflows(wf.Namespace).Delete(ctx, wf.Name, metav1.DeleteOptions{PropagationPolicy: argoutil.GetDeletePropagation()})
```
alexec, line 308: I think DeleteOptions might have force option?

dpadhiar: It does not, Bala suggested we do this method instead.

```
type DeleteOptions struct {
	TypeMeta `json:",inline"`

	
	GracePeriodSeconds *int64 `json:"gracePeriodSeconds,omitempty" protobuf:"varint,1,opt,name=gracePeriodSeconds"`

	Preconditions *Preconditions `json:"preconditions,omitempty" protobuf:"bytes,2,opt,name=preconditions"`

	OrphanDependents *bool `json:"orphanDependents,omitempty" protobuf:"varint,3,opt,name=orphanDependents"`

	PropagationPolicy *DeletionPropagation `json:"propagationPolicy,omitempty" protobuf:"varint,4,opt,name=propagationPolicy"`


	DryRun []string `json:"dryRun,omitempty" protobuf:"bytes,5,rep,name=dryRun"`
}
```

alexec: OK. What does `kubectl delete --force` do? Can you run that with debug? I'd expect it be the same

dpadhiar: I patched this wf to add the artifact-gc finalizer and tried running force delete. The console prints that the workflow is deleted but the process doesn't actually finish and the workflow isn't deleted. Same behavior as just running `kubectl delete` or `argo delete` without removing the finalizer.

```
warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.
workflow.argoproj.io "hello-world-scgnt" force deleted
```

sarabala1979: kubectl delete --force basically make GracePeriod=0 delete the resource ASAP. kubectl delete doesn't support the `finalizer` removable. They don't want to bypass the other controllers which are holding the finalizer.

server/workflow/workflow_server.go
```diff
@@ -298,6 +299,15 @@ func (s *workflowServer) DeleteWorkflow(ctx context.Context, req *workflowpkg.Wo
 	if err != nil {
 		return nil, err
 	}
+	if wf.Finalizers != nil && !req.Force {
+		return nil, fmt.Errorf("%s has finalizer. Use argo delete --force to delete this workflow", wf.Name)
+	}
```
juliev0, line 304: great idea to add the helpful hint!

alexec (approved):

PR #9352, opened 2022-08-12
documentation changes

Yh010:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=260wWuS)
Fixes #TODO

### DIFF
--- a/docs/cli/argo_terminate.md
+++ b/docs/cli/argo_terminate.md
@@ -1,4 +1,4 @@
-## argo terminate
+## argo terminate:
 
 terminate zero or more workflows immediately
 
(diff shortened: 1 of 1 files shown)


terrytangyuan (changes requested):
The changes look unnecesary. What was your reasoning behind these changes? This file is also autogenerated and you'll need to modify the Go files in commands instead.

Yh010:
> The changes look unnecesary. What was your reasoning behind these changes? This file is also autogenerated and you'll need to modify the Go files in commands instead.

Felt like there were grammatical errors and the code seemed quite congested so modified that✌.

alexec:
These files are generated from the source code.  You can edit the correct file (e.g. terminate.go) and then run `make codegen`

Yh010:
Ok sir will do the necessary

PR #9354, opened 2022-08-12
feat: Add Artifact GC Strategies for "OnWorkflowSuccessOrDeletion" and "OnWorkflowFailureOrDeletion"

### LINKED ISSUE #9318: ArtifactGC: Add strategies for "OnWorkflowSuccessOrDeletion" and "OnWorkflowFailureOrDeletion"
Consider giving users the ability to configure their Workflows so artifacts be automatically deleted only in the case of success or failure.

Strategies like "OnWorkflowSuccessOrDeletion" and "OnWorkflowFailureOrDeletion" would make it so that artifacts could also be deleted when the Workflow gets deleted if they weren't already deleted.

Need to determine if we also want to have "OnWorkflowSuccess" and "OnWorkflowFailure" or no need...

---

juliev0:
Fixes #9318 

Note the reason to add strategies for "OnWorkflowSuccessOrDeletion" and "OnWorkflowFailureOrDeletion" rather than "OnWorkflowSuccess" and "OnWorkflowFailure" is to ensure artifacts get deleted at some point.

We can add "OnWorkflowSuccess" and "OnWorkflowFailure" at some point if there's demand for those.

I also considered trying to consolidate multiple strategies into a single GC Pod but I decided it wasn't really worth the complexity - and I think having each in a separate Pod named by its strategy will make the inspection of ArtifactGCTasks and Pods more intuitive.

### DIFF
--- a/test/e2e/fixtures/when.go
+++ b/test/e2e/fixtures/when.go
@@ -233,12 +233,16 @@ var ToBeWaitingOnAMutex Condition = func(wf *wfv1.Workflow) (bool, string) {
 	return wf.Status.Synchronization != nil && wf.Status.Synchronization.Mutex != nil, "to be waiting on a mutex"
 }
 
+type WorkflowCompletionOkay bool
+
 // Wait for a workflow to meet a condition:
 // Options:
 // * `time.Duration` - change the timeout - 30s by default
 // * `string` - either:
 //    * the workflow's name (not spaces)
 //    * or a new message (if it contain spaces) - default "to finish"
+// * `WorkflowCompletionOkay`` (bool alias): if this is true, we won't stop checking for the other options
+//    * just because the Workflow completed
 // * `Condition` - a condition - `ToFinish` by default
 func (w *When) WaitForWorkflow(options ...interface{}) *When {
 	w.t.Helper()
@@ -248,6 +252,7 @@ func (w *When) WaitForWorkflow(options ...interface{}) *When {
 		workflowName = w.wf.Name
 	}
 	condition := ToBeDone
+	var workflowCompletionOkay WorkflowCompletionOkay
 	for _, opt := range options {
 		switch v := opt.(type) {
 		case time.Duration:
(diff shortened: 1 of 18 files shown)


test/e2e/fixtures/when.go
```diff
@@ -294,11 +301,13 @@ func (w *When) WaitForWorkflow(options ...interface{}) *When {
...
-				if ok, _ = ToBeDone(wf); ok {
-					w.t.Errorf("condition never and cannot be met because the workflow is done")
-					return w
+				if !workflowCompletionOkay {
+					// once done the workflow is done, the condition can never be met
+					// rather than wait maybe 30s for something that can never happen
+					if ok, _ = ToBeDone(wf); ok {
+						w.t.Errorf("condition never and cannot be met because the workflow is done")
+						return w
+					}
```
juliev0, line 310: In this case, it's okay that the Workflow is completed so I needed to be able to allow for that here without affecting the other tests. Hence, I added a new Option type.

I gave it a whole new type "WorkflowCompletionOkay" rather than just bool, in case we want to add other bool types in the future. (I tend to think we should do the same for the "string" option.)

alexec:
I think this should be abandoned.

* I can't see a use case where you'd want to delete on failure.
* Zero 👍 or comments on the linked issue. Where is the customer problem?

I think we/I might have made a mistake with having a single strategy field, when you might want options.

```
strategies:
- OnFailure
- OnDeletion
- OnSuccess
- OnError
- OnAliensDiscovered
```


Etc.

juliev0:
> I think this should be abandoned.
> 
> * I can't see a use case where you'd want to delete on failure.
> * Zero 👍 or comments on the linked issue. Where is the customer problem?
> 
> I think we/I might have made a mistake with having a single strategy field, when you might want options.
> 
> ```
> strategies:
> - OnFailure
> - OnDeletion
> - OnSuccess
> - OnError
> - OnAliensDiscovered
> ```
> 
> Etc.

Yeah, was thinking about the combination of strategies as well but went with this since it was a more trivial change and because some of the combinations don’t really make sense (but of course that could be okay in terms of either just letting people do what they want or prohibiting those combinations.)

juliev0:
If others want to wait and see if we start getting interest in this that’s fine and I’ll hold off. I have some e2e test improvements in here that I could incorporate into a separate PR.

PR #9383, opened 2022-08-16
fix: removed error check which prevented deleting successful artGC wfs. 

dpadhiar:

Quick fix for artifactGC. Error message check is preventing successful artifact GC workflows from being deleted. Will add better check before 3.4 release.

### DIFF
--- a/server/workflow/workflow_server.go
+++ b/server/workflow/workflow_server.go
@@ -299,9 +299,6 @@ func (s *workflowServer) DeleteWorkflow(ctx context.Context, req *workflowpkg.Wo
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


sarabala1979 (approved):

server/workflow/workflow_server.go
```diff
@@ -299,9 +299,6 @@ func (s *workflowServer) DeleteWorkflow(ctx context.Context, req *workflowpkg.Wo
 	if err != nil {
 		return nil, err
 	}
-	if wf.Finalizers != nil && !req.Force {
```
qhuai, line 302: Will this change have unintended side effects? For example, if the workflow has a non-GC related finalizer, should we still require --force for the delete command?

PR #9387, opened 2022-08-17
feat: added support for DAG task name as variables in workflow templates

### LINKED ISSUE #7698: Variable for DAG task name (or step name)
# Summary

It would be useful to have a way of retrieving the task name of a DAG task. For all of the DAG template variables in the documentation, the TASKNAME is required before they can be accessed. But there is no way of referencing the TASKNAME; it's presumed it is known beforehand.  

# Use Cases

When would you use this?

For example, let's say I've got several different volumes and I want to mount each in a different task in my DAG. But everything else in the template used for each task is the same. IN the template, it would be useful to add a placeholder in volumeMounts referencing the task.name, like this:

```
  templates:
  - name: myscript
    inputs:
      parameters:
        - name: task
        - name: script
    container:
      image: alpine:v3.14
      imagePullPolicy: Never
      command: [sh, -c]
      args: ["echo 'test' > /mnt/myvolume/file.txt"]
      volumeMounts:
      - name: "{tasks.name}"
        mountPath: /mnt/myvolume

  - name: processing-tasks
    dag:
      tasks:
        - name: task1
          template: myscript
        - name: task2
          template: myscript

```

This way I only need to have one template which allows me to dynamically reference different volumes based on the name of the task. Currently, (as far as I can tell), I have to create a different template referencing a different volume for each dag task. This seems very inefficient. 

---

rohankmr414:
Signed-off-by: Rohan Kumar <rohan@pipekit.io>

Fixes #7698


sarabala1979:
@rohankmr414 can we add step name and task name as local param by default?

sarabala1979:
@rohankmr414 Can you add a test for this?

docs/variables.md
```diff
@@ -150,6 +150,7 @@ returns `0`. Please review the Sprig documentation to understand which functions
 
 | Variable | Description|
 |----------|------------|
+| `tasks.name` | Name of the task |
```
sarabala1979, line 153: add variable for step

workflow/controller/operator.go
```diff
@@ -1672,7 +1672,9 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 	if resolvedTmpl.IsPodType() && woc.retryStrategy(resolvedTmpl) == nil {
 		localParams[common.LocalVarPodName] = woc.getPodName(nodeName, resolvedTmpl.Name)
 	}
-
+	if orgTmpl.IsDAGTask() {
```
sarabala1979, line 1675: Can you refactor this code to support Step and Task?

Rachel-Rabinowitz:
Can you finish with that?
We are waiting for it.
Thanks

sarabala1979 (approved):

PR #9396, opened 2022-08-19
fix(executor): Resource template gets incorrect plural for certain types

### LINKED ISSUE #9393: Some resources cannot use success/failure conditions at all due to bug in UnsafeGuessKindToResource
## Checklist

 

## Summary

Due to this bug https://github.com/kubernetes/client-go/issues/1082
And the usage of this function here: https://github.com/argoproj/argo-workflows/blob/74cdf5d870cc4d0b5576e6d78da7a6fde6a1be99/workflow/executor/resource.go#L80
certain custom resources will get the incorrect plural and fail when using a success/failure condition.  

gateway is one such word.  

The relevant log from the executor is here
```
time="2022-08-18T19:55:50.232Z" level=info msg="Resource: kaas-egress-ip/eipgateway.egressips.acquia.io/eipgateway-post-sync-wf-1660852548-7f461cfc5b5e. SelfLink: apis/egressips.acquia.io/v1alpha1/namespaces/kaas-egress-ip/eipgatewaies/eipgateway-post-sync-wf-1660852548-7f461cfc5b5e"
time="2022-08-18T19:55:50.232Z" level=info msg="Waiting for conditions: status.eipAssignments.1.state==ASSIGNED,status.eipAssignments.2.state==ASSIGNED"
time="2022-08-18T19:55:50.240Z" level=info msg="Get eipgatewaies 403"
time="2022-08-18T19:55:50.240Z" level=warning msg="Non-transient error: eipgatewaies.egressips.acquia.io \"eipgateway-post-sync-wf-1660852548-7f461cfc5b5e\" is forbidden: User \"system:serviceaccount:kaas-egress-ip:eip-post-sync-tests\" cannot get resource \"eipgatewaies\" in API group \"egressips.acquia.io\" in the namespace \"kaas-egress-ip\""
```

## Diagnostics

You'll need a CRD to test this, but anything ending in `gateway` will trigger this bug. 

---

wreed4:
Fixes #9393


terrytangyuan (commented):
Thank you! Since we only rely on a single function from gengo, it might be better to fork the implementation (with a link and credit to original source code) instead of adding a less popular dependency.

workflow/executor/resource.go
```diff
@@ -77,7 +78,11 @@ func inferObjectSelfLink(obj unstructured.Unstructured) string {
 	gvk := obj.GroupVersionKind()
 	// This is the best guess we can do here and is what `kubectl` uses under the hood. Hopefully future versions of the
 	// REST client would remove the need to infer the plural name.
-	pluralGVR, _ := meta.UnsafeGuessKindToResource(gvk)
```
terrytangyuan, line 80: Can you check if there’s any improvement to this function in the new versions of meta?

wreed4:
re: forking the implementation it's not _terribly_ much to fork and with a comment maybe it's okay, but using this function directly was suggested in the client-go bug because this is what is used by the tools that generate things like DeepCopy in controller-gen and a lot of other boiler plate code.  Personally I wouldn't want to split away from that logic, but I do understand it's a small bit of code to add a whole dependency for.  So whichever you guys feel happier with I'm happy to do.

This is the function implementation just to make things simple.  There are a handful of calls out, but it's not massive.  It is more than I would want to copy and paste personally, but again up to you guys. 
```golang
// Name returns the plural form of the type's name. If the type's name is found
// in the exceptions map, the map value is returned.
func (r *pluralNamer) Name(t *types.Type) string {
	singular := t.Name.Name
	var plural string
	var ok bool
	if plural, ok = r.exceptions[singular]; ok {
		return r.finalize(plural)
	}
	if len(singular) < 2 {
		return r.finalize(singular)
	}

	switch rune(singular[len(singular)-1]) {
	case 's', 'x', 'z':
		plural = esPlural(singular)
	case 'y':
		sl := rune(singular[len(singular)-2])
		if isConsonant(sl) {
			plural = iesPlural(singular)
		} else {
			plural = sPlural(singular)
		}
	case 'h':
		sl := rune(singular[len(singular)-2])
		if sl == 'c' || sl == 's' {
			plural = esPlural(singular)
		} else {
			plural = sPlural(singular)
		}
	case 'e':
		sl := rune(singular[len(singular)-2])
		if sl == 'f' {
			plural = vesPlural(singular[:len(singular)-1])
		} else {
			plural = sPlural(singular)
		}
	case 'f':
		plural = vesPlural(singular)
	default:
		plural = sPlural(singular)
	}
	return r.finalize(plural)
}
```

workflow/executor/resource.go
wreed4: This is the PR that is open to fix the issue, but it's not merged and it's unclear to me if people are happy with this solution or if they want to use the solution I did in this PR. 

https://github.com/kubernetes/kubernetes/pull/110053

In master it's still in silly mode https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L143

terrytangyuan:
> this is what is used by the tools that generate things like DeepCopy in controller-gen and a lot of other boiler plate code
Could you link me to where this is being used?

Another concern is that there isn't any stable releases yet: https://github.com/kubernetes/gengo/releases

wreed4:
https://github.com/kubernetes/gengo
image

I picked a random stable version of the kubernetes project to look into
https://github.com/kubernetes/kubernetes/blob/v1.21.10/staging/src/k8s.io/code-generator/cmd/informer-gen/generators/packages.go
https://github.com/kubernetes/kubernetes/blob/a55bd631728590045b51a4f65bba31aed1415571/staging/src/k8s.io/code-generator/cmd/lister-gen/generators/lister.go#L43

lots of places
https://github.com/kubernetes/kubernetes/search?p=3&q=%22k8s.io%2Fgengo%2Fnamer%22

terrytangyuan (approved):
Looks good. Thanks!

PR #9412, opened 2022-08-22
docs: add 'how to install' to argo-cli

asfaltboy:
## Context
I was missing this bit of information when I got to this page. Also includes a reference of using homebrew.

## Testing

Checked the markdown renders in github.

Going to try and rely on CI to make docs, if that doesn't happen will update with testing strategy.

Update: need to address the failing "spelling check"

### DIFF
--- a/docs/walk-through/argo-cli.md
+++ b/docs/walk-through/argo-cli.md
@@ -2,6 +2,18 @@
 
 In case you want to follow along with this walk-through, here's a quick overview of the most useful argo command line interface (CLI) commands.
 
+## Install the Argo CLI tool
+
+You may obtain [the latest release from the releases page](https://github.com/argoproj/argo-workflows/releases/latest).
+
+On MacOS you may also use homebrew to install the tool:
+
+```sh
+brew install argo
+```
+
+## Quick start with Argo CLI
+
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
I am not sure if the changes are necessary. The installation instruction is here: https://github.com/argoproj/argo-workflows/blob/83fa63794a958881a2cd24182a4f101889d68f15/docs/installation.md

It refers to more specific links to install from releases.

asfaltboy:
> The installation instruction is here

I know this is pretty basic. My goal is to improve the discoverability. I wanted to install the CLI tool, and searching through the docs brought me to this page, not the release page. Note that the release page doesn't mention Argo CLI, and I was less likely to follow to that page (I didn't need to install argo server or controller)

Would it better if the page should instead instruct users to refer to the `releases.md` page for up to date instructions (to avoid maintaining two methods of downloading)? Happy to change it if this makes sense, could also add a note there about what's available on the releases page.

On a related note, do you think the release notes should also mention homebrew, or you feel it's better to instruct users to download the pre-compiled binary as we do today?

agilgur5:
Recreated in #11751

PR #9421, opened 2022-08-24
docs: Improve quickstart flow and installation docs. Fixes #9398

### LINKED ISSUE #9398: Diminish the use of quickstart manifests to improve first-time user experience
# Summary

The Argo Workflows quickstart manifests are referenced heavily within the Argo Workflows documentation. It is highly likely that the first experience of Argo Workflows to newcomers will be via these quickstart manifests. (Note, for the purpose of this document, “Quick Start” refers to anything in the argo/manifests directory on github.)

These manifests rely on referencing the `latest` image for various Workflows components. Examples:

https://github.com/workflow/argo/blob/master/manifests/base/workflow-controller/workflow-controller-deployment.yaml#L17

https://github.com/workflow/argo/blob/master/manifests/base/argo-server/argo-server-deployment.yaml#L17

The `latest` images are naturally bleeding-edge versions of Argo Workflows. At best, this produces unexpected results for new users, which gives a poor first impression. At worst, Argo Workflows simply doesn’t install properly, users get frustrated and move to an alternative tool.

A high number of slack queries have a root cause of the user using an unstable quickstart manifest and receiving an unexpected outcome.

I am proposing that the Getting Started documentation stops referencing the quick start manifests and instead references the official release manifests. This then smooths the upgrade flow from quick start > production.

I am NOT proposing to remove the quickstart manifests themselves, certainly as part of this enhancement. There may be people who historically rely on them, so we shouldn't just make them disappear.

I am happy to make this change myself.

---

tico24:
Signed-off-by: Tim Collins <tim@thecollins.team>

Fixes #9398


docs/installation.md
```diff
@@ -1,28 +1,40 @@
 # Installation
 
-## Argo on Desktop
+## Non-production installation
```
caelan-io, line 3: I like how you clarified the non-production install use case here 👍

docs/quick-start.md
```diff
@@ -1,31 +1,52 @@
 # Quick Start
 
-To see how Argo Workflows work, you can install it and run examples of simple workflows and workflows that use artifacts.
+To see how Argo Workflows work, you can install it and run examples of simple workflows.
 
-Before you start you need a Kubernetes cluster and `kubectl` set-up
+Before you start you need a Kubernetes cluster and `kubectl` set-up to be able to access that cluster.
```
caelan-io, line 5: Nit: I'd change "set-up" => "set up" while we're in  here

docs/quick-start.md
```diff
@@ -1,31 +1,52 @@
...
+
+⚠️ These instructions are intended to help you get started quickly. They are not suitable in production. For production installs, please refer to [the installation documentation](installation.md) ⚠️
```
caelan-io, line 7: thank you for adding this

docs/quick-start.md
caelan-io: Also, I think this is a good place to link out to the common local K8s deployment tools: minikube/kind/k3d/etc. 

I'm thinking of users coming here without a lot of K8s experience. It could help to ensure they are directly linked to a tool for this, or at least know that all of these work fine for Quickstart.

Local K8s cluster options:
- [MiniKube](https://minikube.sigs.k8s.io/docs/)
- [kind](https://kind.sigs.k8s.io/)
- [K3s](https://rancher.com/docs/k3s/latest/en/quick-start/) or [K3d](https://k3d.io/)
- [Docker Desktop](https://docs.docker.com/desktop/kubernetes/)

docs/quick-start.md
```diff
@@ -1,31 +1,52 @@
...
 
 ## Install Argo Workflows
 
-To get started quickly, you can use the quick start manifest which will install Argo Workflows as well as some commonly used components:
+To install Argo Workflows, navigate to the [releases page](https://github.com/argoproj/argo-workflows/releases/latest) and find the release you wish to use (the latest full release is preferred).
+
+Scroll down to the `Controller and Server` section and execute the `kubectl` commands.
```
caelan-io, line 13: I think we should include the commands here with a placeholder for the version number so the user doesn't have to nav away from this doc. What do you think about something like this?

```
kubectl create namespace argo
kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.X.Y/install.yaml
```

tico24: I did initially write it like that. Then came to the conclusion that some people will skim the page, then copy/paste the code and then cry a bit. What do you think?

caelan-io: ha yeah, I see what you mean. Maybe something like the below and we leave it at 3.3.8 for now since that's stable and we let the community update it via PR in the future:

Below is an example of the install commands. Ensure you update the commend to install the correct version number.
```
kubectl create namespace argo
kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.3.8/install.yaml

Comment (tico24) [reply docs/quick-start.md:13]: I decided to go with your initial idea. I think it's better that an install completely fails (due to not reading) than the user ending up with a potentially old, unsupported version and not realising.

File: docs/quick-start.md [hunk 5]
```diff
@@ -59,5 +90,21 @@ You can also observe the logs of the Workflow run by running the following:
...
 
+### Submit an example workflow (GUI)
+
+* Open a port-forward so you can access the UI:
+
+```bash
+kubectl -n argo port-forward deployment/argo-server 2746:2746
+```
+
+* Navigate your browser to <https://localhost:2746>.
```
terrytangyuan, line 101: Remove `<>` so readers can just click the link?

docs/quick-start.md
```diff
@@ -1,31 +1,62 @@
...
+### Port-forward the UI
+
 Open a port-forward so you can access the UI:
 
 ```bash
 kubectl -n argo port-forward deployment/argo-server 2746:2746
 ```
 
-This will serve the UI on <https://localhost:2746>
+This will serve the UI on <https://localhost:2746>. Due to the self-signed certificate, you will receive a TLS error which you will need to manually approve.
```
terrytangyuan, line 51: Same here

terrytangyuan (commented):
Thank you! This looks much better.

docs/quick-start.md
tico24: This breaks the CI. (This url was also written like this previously)

docs/quick-start.md
tico24: As above.

docs/quick-start.md
tico24: The `>` chars are removed at rendering. Look at https://argoproj.github.io/argo-workflows/quick-start/

terrytangyuan (approved):

PR #9423, opened 2022-08-24
Add --tls-certificate-secret-name parameter to server command. Fixes #5582

### LINKED ISSUE #5582: TLS configuration and/or docs enhancement
# Summary

It looks like after 3.0 release argo server listens on https port with self-signed certificate (#5581 )
I think it would be great to have a dedicated ConfigMap with tls.cert and tls.key parameters which would be mounted inside pod (Like it is implemented in ArgoCD) so it would be very easy to generate a valid certificate with cert-manager or ue any pre-isued one.
At least it would be great to have a parameter or doc which would tell where the cert is read from. It looks like it is harcoded now and it's impossible to mount CM inside pod without even knowing where the self-signed cert is located...

# Use Cases

When deploying a new instance with Ingress it would be great to use ACME from cert-manager

---

chtcvl:
Signed-off-by: vladimir ivanov <vladimir.ivanov@grasshopperasia.com>

Fixes #5582

Add --tls-certificate-secret-name parameter to server command.
Since server runs on localhost:2746, get an error in the UI "certificate is valid for ... not for localhost". Therefore the merge request set `InsecureSkipVerify` on TLSconfig for grpc client on the gateway.

Tested to to work with secret provided by cert-manager (added to manifests).

### DIFF
--- a/manifests/base/argo-server/argo-server-certificate.yaml
+++ b/manifests/base/argo-server/argo-server-certificate.yaml
@@ -1,0 +1,20 @@
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: argo-workflows-issuer
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: argo-server-cert
+spec:
+  dnsNames:
+  - argo-server.argo.svc.cluster.local
+  - argo-server.argo.svc
+  - argo-server
+  issuerRef:
+    kind: Issuer
+    name: argo-workflows-issuer
+  secretName: argo-server-tls
--- a/server/apiserver/argoserver.go
+++ b/server/apiserver/argoserver.go
@@ -334,7 +334,10 @@ func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServe
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxGRPCMessageSize)),
 	}
 	if as.tlsConfig != nil {
-		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(as.tlsConfig)))
+		tlsConfig := as.tlsConfig
+		tlsConfig.InsecureSkipVerify = true
+		dCreds := credentials.NewTLS(tlsConfig)
+		dialOpts = append(dialOpts, grpc.WithTransportCredentials(dCreds))
 	} else {
 		dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	}
--- a/Makefile
+++ b/Makefile
@@ -421,6 +421,9 @@ test: server/static/files.go
 .PHONY: install
 install: githooks
 	kubectl get ns $(KUBE_NAMESPACE) || kubectl create ns $(KUBE_NAMESPACE)
+	# install cert-manager if Certificate CRD is not available
+	kubectl get ns cert-manager || kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.9.1/cert-manager.yaml
+	kubectl wait --for=condition=Ready pods --all --namespace cert-manager
 	kubectl config set-context --current --namespace=$(KUBE_NAMESPACE)
 	@echo "installing PROFILE=$(PROFILE)"
 	kubectl kustomize --load-restrictor=LoadRestrictionsNone test/e2e/manifests/$(PROFILE) | sed 's|quay.io/argoproj/|$(IMAGE_NAMESPACE)/|' | sed 's/namespace: argo/namespace: $(KUBE_NAMESPACE)/' | kubectl -n $(KUBE_NAMESPACE) apply --prune -l app.kubernetes.io/part-of=argo -f -
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ The following commands install Argo Workflows as well as some commmonly used com
 
 ```bash
 kubectl create ns argo
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.9.1/cert-manager.yaml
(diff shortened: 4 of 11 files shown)


juliev0 (commented):
Interesting. Thanks for the addition. Just wondering if we should support the case in which somebody doesn't want the InsecureSkipVerify? (Don't want to add more unnecessary work but not sure what others think)

manifests/base/argo-server/argo-server-certificate.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: argo-server-cert
+spec:
+  dnsNames:
+  - argo-server.argo.svc.cluster.local
+  - argo-server.argo.svc
+  - argo-server
+  issuerRef:
+    kind: Issuer
+    name: argo-workflows-issuer
+  secretName: argo-server-tls
```
juliev0, line 20: Did you intend for this to be deployed as part of installing our manifests? I see manifests/base/argo-server/kustomization.yaml calls out the yaml files in this directory. Should we add it to that?

chtcvl:
> Interesting. Thanks for the addition. Just wondering if we should support the case in which somebody doesn't want the InsecureSkipVerify? (Don't want to add more unnecessary work but not sure what others think)

This `InsecureSkipVerify` is for communication within the same instance of argo-server application - between grpc client and server goroutines, so don't expect this requirement.

manifests/base/argo-server/argo-server-certificate.yaml
chtcvl: Thanks, added to `manifests/base/argo-server/kustomization.yaml`.

Yes, intend to install as part manifests, since certificate from `Certificate` resource is persistent compared to one argo-server generates itself. This way user will only need to set up trust once.

juliev0:
> > Interesting. Thanks for the addition. Just wondering if we should support the case in which somebody doesn't want the InsecureSkipVerify? (Don't want to add more unnecessary work but not sure what others think)
> 
> This `InsecureSkipVerify` is for communication within the same instance of argo-server application - between grpc client and server goroutines, so don't expect this requirement.

Oh, I see…

juliev0:
> Details

Great. As far as I can tell, that will propagate into all of the various manifests. You'll need to run "make codegen -B STATIC_FILES=false" and "git diff" and check in whatever changed (which you can see [here](https://github.com/argoproj/argo-workflows/runs/8126613955?check_suite_focus=true)).

juliev0 (commented):
Thanks for updating the manifests. As far as I can tell, the change looks good to me, but unfortunately I'm not too knowledgeable about this area of the code and hoping someone more knowledgeable can review.

juliev0:
@sarabala1979 @terrytangyuan Are either of you able to review and merge this PR?

server/apiserver/argoserver.go
```diff
@@ -334,7 +334,10 @@ func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServe
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxGRPCMessageSize)),
 	}
 	if as.tlsConfig != nil {
-		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(as.tlsConfig)))
+		tlsConfig := as.tlsConfig
+		tlsConfig.InsecureSkipVerify = true
```
sarabala1979, line 338: @chtcvl Do you need to make InsecureSkipVerify=true for all cases? Can you make as command line arg and configure in `tsConfig`?

chtcvl: Right now `gRPC server` explicitly runs on `localhost` ( https://github.com/argoproj/argo-workflows/blob/769896eb5bf0a7d8db1a94b423e5bc16cf09d5aa/server/apiserver/argoserver.go#L260 ) meaning `https://localhost`. For `grpc-gateway` to access `gRPC server` it has some options:
1 . set `InsecureSkipVerify=true` (not really a concern as this is communication between two goroutines in the same app) - let grpc-gateway trust gRPC server running on localhost. 
2.  have certificate Subject Alternate Name include `localhost` - this is not supported by publicly trusted certificates. 
3. serve on https://realdomain... for which certificate is issued instead of localhost.

This being said, adding new option for `InsecureSkipVerify=false` can lead to confusion as option will only be usable in pt.3 which is not supported by current code.

sarabala1979: Thanks for your response. LGTM @alexec Can you quickly take a look? if I am missing anything.

sarabala1979 (approved):

kpiroddi-yieldstreet:
Hi, This PR enforces users of the manifests to now have cert-manger CRD's installed. Was this the intention? 

To use the new manifests (while happy with a self signed cert), had to add the following to kustomize:

````
patchesStrategicMerge:
  - |-
    apiVersion: cert-manager.io/v1
    kind: Issuer
    metadata:
      name: argo-workflows-issuer
    $patch: delete
  - |-
    apiVersion: cert-manager.io/v1
    kind: Certificate
    metadata:
      name: argo-server-cert
    $patch: delete
   ````

Piroddi:
Hi, This PR enforces users of the manifests to have the cert-manger CRD's installed. Is this the intention? 

To use the new manifests (while happy with a self signed cert). Had to add the following to kustomize:

````
patchesStrategicMerge:
  - |-
    apiVersion: cert-manager.io/v1
    kind: Issuer
    metadata:
      name: argo-workflows-issuer
    $patch: delete
  - |-
    apiVersion: cert-manager.io/v1
    kind: Certificate
    metadata:
      name: argo-server-cert
    $patch: delete
````

allemp:
> patchesStrategicMerge:
>   - |-
>     apiVersion: cert-manager.io/v1
>     kind: Issuer
>     metadata:
>       name: argo-workflows-issuer
>     $patch: delete
>   - |-
>     apiVersion: cert-manager.io/v1
>     kind: Certificate
>     metadata:
>       name: argo-server-cert
>     $patch: delete

This was a huge issue for me as well.

PR #9438, opened 2022-08-25
fix: default to 'main' container in Sensor logs. Fixes #9459

### LINKED ISSUE #9459: UI hangs when attempting to view Sensor logs if pod has more than one container
## Checklist

 

## Summary

We're using a Kubernetes mutating webhook to inject a fluent-bit container into certain running pods for the purposes of collecting logs and sending them to Splunk. When attempting to view Sensor logs in the UI, the Argo Workflows API hangs and eventually returns a 504 - Gateway Timeout.

This seems to happen because the API encounters this message when attempting to query container logs:

```
time="2022-08-25T14:30:21.960Z" level=error msg="a container name must be specified for pod adobe-platform--[snip], choose one of: [main fluent-bit]" namespace=ns-team-adobe-platform--[snip] podName=adobe-platform--[snip]
```

The API does support parsing `podLogOptions.container`, and so by appending `podLogOptions.container=main` to the API call, I confirmed the call returns instantly with the logs we expect, so this is a matter of updating this at the UI level.

The container name of `main` can be seen here: https://github.com/argoproj/argo-events/blob/56196143ecbf8b451d5ecb02a0ca13835e20954c/controllers/sensor/resource.go#L267

A potential fix to this is here: https://github.com/argoproj/argo-workflows/pull/9438

What version are you running?

Argo Workflows v3.3.8

## Diagnostics

This happens with any workflow so long as the Sensor pod has more than 1 container.

```yaml

```

```bash

```

---

jsvk:
Fixes https://github.com/argoproj/argo-workflows/issues/9459

### DIFF
--- a/ui/src/app/shared/services/sensor-service.ts
+++ b/ui/src/app/shared/services/sensor-service.ts
@@ -33,8 +33,8 @@ export class SensorService {
         return requests.loadEventSource(`api/v1/stream/sensors/${namespace}`).pipe(map(line => line && (JSON.parse(line).result as SensorWatchEvent)));
     }
 
-    public sensorsLogs(namespace: string, name = '', triggerName = '', grep = '', tailLines = -1) {
-        const params = ['podLogOptions.follow=true'];
+    public sensorsLogs(namespace: string, name = '', triggerName = '', grep = '', tailLines = -1, container = 'main') {
+        const params = ['podLogOptions.follow=true', `podLogOptions.container=${container}`];
         if (name) {
             params.push('name=' + name);
         }


terrytangyuan (commented):
What problem is this trying to solve?

jsvk:
sorry still working on the write-up (in the form of Github Issue, hope that's ok), but the short summary is we're injecting a sidecar container to the Sensor pod (fluent-bit in our case, for forwarding logs to Splunk). 

Without specifying the container, the UI fails to load the logs and instead returns with a 504 - Gateway Timeout.

Error logs in the Server pod say it encountered this error - 
```
time="2022-08-25T14:30:21.960Z" level=error msg="a container name must be specified for pod adobe-platform--[snip], choose one of: [main fluent-bit]" namespace=ns-team-adobe-platform--[snip] podName=adobe-platform--[snip]
```

I saw the API does support parsing `podLogOptions.container`, and so by appending `?podLogOptions.container=main` to the API call, I confirmed the call returns instantly with the logs we expect, so this is a matter of updating this at the UI level.

Is there a better approach I'm not thinking of?

alexec (approved):

PR #9448, opened 2022-08-26
feat: improve e2e test for ArtifactGC

juliev0:

These were some changes I made as part of [this](https://github.com/argoproj/argo-workflows/pull/9354/files#) PR, which may not get merged in, so I'm putting them here. This test checks not just that the Artifact GC happened but that the right strategy happened.


test/e2e/fixtures/when.go
```diff
@@ -294,11 +301,13 @@ func (w *When) WaitForWorkflow(options ...interface{}) *When {
...
-					w.t.Errorf("condition never and cannot be met because the workflow is done")
-					return w
+				if !workflowCompletionOkay {
+					// once done the workflow is done, the condition can never be met
+					// rather than wait maybe 30s for something that can never happen
+					if ok, _ = ToBeDone(wf); ok {
+						w.t.Errorf("condition never and cannot be met because the workflow is done")
+						return w
+					}
 				}
```
juliev0, line 311: In this case, it's okay that the Workflow is completed so I needed to be able to allow for that here without affecting the other tests. Hence, I added a new Option type.

I gave it a whole new type "WorkflowCompletionOkay" rather than just bool, in case we want to add other bool types in the future. (I tend to think we should do the same for the "string" option.)

sarabala1979 (approved):

PR #9464, opened 2022-08-29
fix: link to "get artifacts from logs" was assuming Node ID was equal to Pod Name

### LINKED ISSUE #9434: 3.4-rc2 - Workflows UI can no longer get logs (s3)
## Checklist

 

## Summary
This occurs following an upgrade from workflows 3.3.9 to 3.4-rc2.

Logs are still correctly sent to s3 by argo workflows, I can see main.log in s3 and the contents of the log file are correct.

However, once the workflow has finished the pod has been archived, the logs field is now empty in the UI.
Clicking `Try getting logs from the artifacts` results in `Internal Server Error`

The argo-server logs show this error, only when trying to clikc the 'try getting logs from artifacts' link.: `level=error msg="Artifact Server returned internal error" error="artifact not found: main-logs"`

No errors at all when just trying to view in the UI. None on the controller either.

Note, as I'm using IRSA, I have the following patch on my argo-server:
```
    spec:
      securityContext:
        fsGroup: 65534
```

What version are you running? 3.4-rc2

Config summary:
controller-configmap:
```yaml
  artifactRepository: |
    # archiveLogs will archive the main container logs as an artifact
    archiveLogs: true

    s3:
      endpoint: s3.amazonaws.com
      bucket: my-bucket-name
      region: us-east-1
      insecure: false
      keyFormat: "my-artifacts\
        /{{workflow.creationTimestamp.Y}}\
        /{{workflow.creationTimestamp.m}}\
        /{{workflow.creationTimestamp.d}}\
        /{{workflow.name}}\
        /{{pod.name}}"
      useSDKCreds: true
```

The argo-server ServiceAccount has the `eks.amazonaws.com/role-arn` annotation as always.

---

juliev0:
Hoping this Fixes #9434 but we'll see
At least it does fix the issue seen when running examples/dag-coinflip.yaml

Root cause was that this UI code was never updated to reflect the change to POD_NAMES v2, so it was still assuming that the Pod Name and Node ID would be equal.

### DIFF
--- a/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
+++ b/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
@@ -148,7 +148,7 @@ export const WorkflowLogsViewer = ({workflow, nodeId, initialPodName, container,
                         <InfoIcon /> Wait containers logs are useful when debugging output artifact problems.
                     </>
                 )}
-                {podName && (
+                {podName && podNamesToNodeIDs.get(podName) && (
                     <>
                         Still waiting for data or an error? Try getting{' '}
                         <a href={services.workflows.getArtifactLogsPath(workflow, podNamesToNodeIDs.get(podName), selectedContainer, archived)}>logs from the artifacts</a>.
--- a/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
+++ b/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
@@ -71,7 +71,8 @@ export const WorkflowLogsViewer = ({workflow, nodeId, initialPodName, container,
     }
     const podNameVersion = annotations[ANNOTATION_KEY_POD_NAME_VERSION];
 
-    const podNamesToNodeNames = new Map<string,string>();
+    // map pod names to corresponding node IDs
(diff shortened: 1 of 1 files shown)


ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
```diff
@@ -106,8 +110,14 @@ export const WorkflowLogsViewer = ({workflow, nodeId, initialPodName, container,
...
-                <Autocomplete items={containers} value={selectedContainer} onSelect={setContainer} />
+                <Autocomplete
+                    items={podNames}
+                    value={(podNames.find(x => x.value[0] === podName) || {label: ''}).label}
+                    onSelect={(_, item) => {
+                        setPodName(item.value);
+                    }}
+                />{' '}
+                / <Autocomplete items={containers} value={selectedContainer} onSelect={setContainer} />
                 <span className='fa-pull-right'>
```
juliev0, line 121: (this part was the linter's doing, not mine)

sarabala1979 (approved):

ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
agilgur5: noting that there was a bug here (#9491) that was fixed in #9515

PR #9499, opened 2022-09-02
feat: Support retrying complex workflows with nested group nodes

terrytangyuan:
This includes:
* Major refactoring with clearer logic.
* Connected parent group nodes are reset correctly.
* Only the correct branch that matches the selected node path in the DAG is retried.
* Handle pod deletion properly and fix some cases where nodes fail with "pod deleted" error.
* Added extensive test cases to cover retrying on any selected node in the workflow.


workflow/util/util.go
```diff
@@ -813,6 +847,7 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 	deletedNodes := make(map[string]bool)
 	deletedPods := make(map[string]bool)
 	var podsToDelete []string
+	var resettedParentGroupNodes []string
```
juliev0, line 850: sorry to be annoying but can this be changed to "resetParentGroupNodes" since the past tense verb is "reset"?

juliev0: (same for all uses of this word)

terrytangyuan: Definitely! I’ll update this. Thanks!

workflow/util/util.go
```diff
@@ -765,6 +765,40 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+		if !containsNode(resettedParentGroupNodes, currentNodeID) {
+			newWF.Status.Nodes[currentNodeID] = resetNode(*currentNode.DeepCopy())
+			resettedParentGroupNodes = append(resettedParentGroupNodes, currentNodeID)
+			log.Debugf("Resetting connected group node %s", currentNode.Name)
+		}
+		if currentNode.BoundaryID != "" && currentNode.BoundaryID != oldWF.ObjectMeta.Name {
+			parentNode := oldWF.Status.Nodes[currentNode.BoundaryID]
+			if isGroupNode(parentNode) {
+				currentNodeID = parentNode.ID
+			}
```
juliev0, line 794: What happens in the implicit else case here? If I'm reading it correctly, it seems currentNodeID doesn't get changed and we end up in an infinite loop, no? Or am I mistaken?

workflow/util/util.go
```diff
@@ -821,37 +856,62 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
...
 						}
+						if nodeGroupNeedsReset {
+							newWF, resettedParentGroupNodes = resetConnectedParentGroupNodes(wf, newWF, node, resettedParentGroupNodes)
+						}
 					}
 				} else {
+					if node.Type == wfv1.NodeTypePod || node.Type == wfv1.NodeTypeSuspend {
+						newWF, resettedParentGroupNodes = resetConnectedParentGroupNodes(wf, newWF, node, resettedParentGroupNodes)
+						// Only remove the descendants of a suspended node but not the suspended node itself. The descendants
+						// of a suspended node need to be removed since because the conditions should be re-evaluated based on
```
juliev0, line 883: typo in comment: change "since because" to either "since" or "because"

juliev0 (changes requested):
Requested changes (inline)

terrytangyuan:
@juliev0 Thanks for your feedback! All comments have been addressed.

juliev0 (approved):

terrytangyuan:
Test failure is unrelated: https://github.com/argoproj/argo-workflows/issues/9591

rbreeze (approved):
LGTM!

PR #9552, opened 2022-09-09
ci: add minimum GitHub token permissions for workflows

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using https://github.com/step-security/secure-workflows.

GitHub Actions workflows have a GITHUB_TOKEN with write access to multiple scopes.
Here is an example of the permissions in one of the workflows:
https://github.com/argoproj/argo-workflows/runs/8125260622?check_suite_focus=true#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflows:

- changelog.yaml
- ci-build.yaml
- gh-pages.yaml
- release.yaml
- sdks.yaml
- snyk.yml

The following workflow file already has the least privileged token permission set:

- dependabot-reviewer.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- GitHub recommends defining minimum GITHUB_TOKEN permissions.
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -5,8 +5,14 @@ on:
     tags:
       - v*
       - "!v0.0.0"
+permissions:
+  contents: read
+
 jobs:
   generate_changelog:
+    permissions:
+      contents: write  # for peter-evans/create-pull-request to create branch
+      pull-requests: write  # for peter-evans/create-pull-request to create a PR
     if: github.repository == 'argoproj/argo-workflows'
     runs-on: ubuntu-latest
     name: Generate changelog
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -13,6 +13,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   tests:
     name: Unit Tests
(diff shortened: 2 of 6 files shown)


alexec:
These work fine today. Why change them?

boahc077:
> https://github.com/argoproj/argo-workflows/runs/8125260622?check_suite_focus=true#step:1:19

Thanks for reviewing my PR @alexec. If you visit https://github.com/argoproj/argo-workflows/runs/8125260622?check_suite_focus=true#step:1:19, you would see that all GitHub action workflows have elevated GitHub token permissions which increases the attack surface drastically. I calculated minimum token permissions based on the actions included in these workflows, you would see that most of them only need limited read access. This change will update token permissions for all workflows so that they operate at the least privileged level as opposed to using an elevated GitHub token by default.

alexec (approved):

terrytangyuan (approved):

PR #9557, opened 2022-09-09
docs: new S3 access requirements as of v3.4

juliev0:

### DIFF
--- a/docs/configure-artifact-repository.md
+++ b/docs/configure-artifact-repository.md
@@ -75,7 +75,8 @@ $ cat > policy.json <<EOF
          "Effect":"Allow",
          "Action":[
             "s3:PutObject",
-            "s3:GetObject"
+            "s3:GetObject",
+            "s3:ListBucket"
          ],
          "Resource":"arn:aws:s3:::$mybucket/*"
       }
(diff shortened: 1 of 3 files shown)


docs/configure-artifact-repository.md
```diff
@@ -75,7 +75,8 @@ $ cat > policy.json <<EOF
          "Effect":"Allow",
          "Action":[
             "s3:PutObject",
-            "s3:GetObject"
+            "s3:GetObject",
+            "s3:ListBucket"
          ],
```
juliev0, line 80: "ListBucket" is required as of 3.4 (written for the Artifact Visualization feature) because we do a check to determine if the file is a directory, which requires this

docs/configure-artifact-repository.md
```diff
@@ -75,7 +75,8 @@ $ cat > policy.json <<EOF
          "Effect":"Allow",
          "Action":[
             "s3:PutObject",
-            "s3:GetObject"
+            "s3:GetObject",
+            "s3:ListBucket"
```
terrytangyuan, line 79: Would existing workflows break if this permission is not there when s3 is used? If so, we also need to document this in upgrading guide.

juliev0: wil do!

juliev0: (and yes)

terrytangyuan (approved):

PR #9575, opened 2022-09-12
fix: Log early abort. Fixes #9573

### LINKED ISSUE #9573: logs incomplete comparing `kubectl logs xxx`
## Summary

Logs incomplete comparing `kubectl logs xxx`

What version are you running?

argo v3.2.9

## Diagnostics

```yaml
kind: Workflow
metadata:
  generateName: hello-world-
  labels:
    test: "true"
spec:
  entrypoint: whalesay
  arguments:
    parameters:
    - name: hello
      value: world
  templates:
  - name: whalesay
    script:
      image: docker/whalesay:latest
      command: [bash]
      source: |
        echo "this will be shown"
        for((j=0;j<=10;j+=1))
        do
          b=""
          for((i=0;i<=100;i+=1))
          do
              printf "progress %3d: [%-100s] %3d%%\r" $j $b $i
              sleep 0.002
              b+='#'
          done
        done
        echo "this won't be shown in argo logs"
```
# Logs from the argo cli:
argo315 logs hello-world-ktfw4 -n argo
```
hello-world-ktfw4: time="2022-09-12T08:54:35.491Z" level=info msg="capturing logs" argo=true
hello-world-ktfw4: this will be shown
```
# Logs from in your kubectl logs:
kubectl logs hello-world-ktfw4 -n argo -c main
```
time="2022-09-12T08:54:35.491Z" level=info msg="capturing logs" argo=true
this will be shown
progress   1: [#####################2022-09-12T16:54:36.690118203+08:00 ############                                              progress   2: [#########################################################2022-09-12T16:54:37.181210511+08:00 #########             progress   3: [#############################################################################################2022-09-12T16:54:37.78progress   5: [#######2022-09-12T16:54:38.252760981+08:00 #########################                                               progress   6: [###########################################2022-09-12T16:54:38.732911680+08:00 ######################              progress   7: [###############################################################################2022-09-12T16:54:39.176877542+08:00 progress2022-09-12T16:54:39.627235787+08:00    9: [###############################                                                progress  10: [#############################2022-09-12T16:54:39.640302442+08:00 ###################################               this won't be shown in argo logs################################################################################################] 100%
```
image

---

jibuji:
Fixes #9573
argo logs abort early caused by bufio.ErrTooLong

### DIFF
--- a/util/logs/workflow-logger.go
+++ b/util/logs/workflow-logger.go
@@ -41,6 +41,27 @@ type sender interface {
 	Send(entry *workflowpkg.LogEntry) error
 }
 
+const maxTokenLength = 1024 * 1024
+const startBufSize = 16 * 1024
+
+func scanLinesOrGiveLong(data []byte, atEOF bool) (advance int, token []byte, err error) {
+	advance, token, err = bufio.ScanLines(data, atEOF)
+	if advance > 0 || token != nil || err != nil {
+		// bufio.ScanLines found something, use it
+		return
+	}
+
+	// bufio.ScanLines found nothing
+	// if our buffer is still a reasonable size, continue scanning for regular lines
+	if len(data) < maxTokenLength {
+		return
+	}
+
+	// our buffer is getting massive, stop waiting for line breaks and return data now
+	// this avoids bufio.ErrTooLong
+	return maxTokenLength, data[0:maxTokenLength], nil
+}
+
 func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient kubernetes.Interface, req request, sender sender) error {
 	wfInterface := wfClient.ArgoprojV1alpha1().Workflows(req.GetNamespace())
 	_, err := wfInterface.Get(ctx, req.GetName(), metav1.GetOptions{})
@@ -111,7 +132,12 @@ func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient 
 					logCtx.Error(err)
 					return
 				}
+
 				scanner := bufio.NewScanner(stream)
+				//give it more space for long line
+				scanner.Buffer(make([]byte, startBufSize), maxTokenLength)
+				//avoid bufio.ErrTooLong error when encounters a very very long line
+				scanner.Split(scanLinesOrGiveLong)
 				for scanner.Scan() {
 					select {
 					case <-ctx.Done():
@@ -119,7 +145,11 @@ func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient 
 					default:
 						line := scanner.Text()
 						parts := strings.SplitN(line, " ", 2)
-						content := parts[1]
+						//on old version k8s, the line may contains no space, hence len(parts) would equal to 1
+						content := ""
+						if len(parts) > 1 {
+							content = parts[1]
+						}
 						timestamp, err := time.Parse(time.RFC3339, parts[0])
 						if err != nil {
 							logCtx.Errorf("unable to decode or infer timestamp from log line: %s", err)


util/logs/workflow-logger.go
```diff
@@ -111,15 +132,24 @@ func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient
...
 				for scanner.Scan() {
 					select {
 					case <-ctx.Done():
 						return
 					default:
 						line := scanner.Text()
 						parts := strings.SplitN(line, " ", 2)
-						content := parts[1]
+						//on old version k8s, the line may contains no space, hence len(parts) would equal to 1
+						content := ""
```
terrytangyuan, line 149: How old is your k8s cluster?

jibuji: 1.18

jibuji: Ok, I see.
But, in my opinion, it may still be necessary to ensure that the length of parts is greater than 1 here.
If there is a very very long token, which doesn't contain any space in it, the length of parts would equal to 1, thus it
 would cause a crash.

terrytangyuan: This sounds like additional scan that could be time-consuming when there are a lot of logs

jibuji: Actually, there isn't additional scan here. `scanLinesOrGiveLong` is just a custom scan function replacing the default one. Comparing the default scan which is `bufio.ScanLines`, `scanLinesOrGiveLong` just a simple function wrapping `bufio.ScanLines` with some extra checking to ensure the token returned wouldn't be too large to cause `bufio.ErrTooLong` error.

terrytangyuan: @alexec Any objections?

util/logs/workflow-logger.go
```diff
@@ -111,15 +132,24 @@ func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient
 					logCtx.Error(err)
 					return
 				}
+
 				scanner := bufio.NewScanner(stream)
+				//give it more space for long line
+				scanner.Buffer(make([]byte, startBufSize), maxTokenLength)
```
alexec, line 138: I don’t really understand this PR. Is is based on any docs or blog post you could share please?

jibuji: `scanLinesOrGiveLong` is to let scanner avoid `bufio.ErrTooLong` error, which will cause `scanner.Scan()` return false, thus break the loop early, leaving logs in the stream not scanned.  This method is borrowed from [here](https://gist.github.com/AlbinoDrought/c292adc185ea6593338fb5b56ac89aa7). 

Actually, the method above alone is sufficient to solve the issue #9675, but it will be nicer with the complementary of the following.

`scanner.Buffer(make([]byte, startBufSize), maxTokenLength)` is to enlarge the scanner's inner buffer(the default size is 4096), and replace the MaxScanTokenSize(the default value is 64 * 1024) with a larger one.  According to our observation, it is common that users print visual progress bar to stdout. Most of the visual progress bar use the '\r' character to return to the start point of the same line, as to give user-friendly visual effect on terminal. Thus, it forms a very long line for scanner, as scanner treats '\n' as the end of each line by default.  if we don't set more larger MaxScanTokenSize, the progress bar will be chopped in the middle very often. it is not a big deal, but larger one will make the logs looks nicer in this case.

It is worthy to mention that this issue can be solved by another method:
_loop over stream manually without the help of scanner, and get the new line when seeing a new '\n'._  
This method have a drawback, user may print a long line in a very long time, such as the progress-bar-like logs, then the user will wait very long time and then see a very long line log pop up abruptly.

alexec (approved):

PR #9586, opened 2022-09-14
fix: avoid panic when not passing AuthSupplier

### LINKED ISSUE #9585: panic when not passing  AuthSupplier option
## Checklist

 
* [x] Tested using `:latest` images.
* [x] Attached the smallest workflow that reproduces the issue.

## Summary

when I try to use [argo client library ](https://argoproj.github.io/argo-workflows/client-libraries/#client-libraries) I got  panic `panic: runtime error: invalid memory address or nil pointer dereference`                                                                     
,when I try  connecting to a server without auth. 

no error or panic expected. SHOULD NOT panic in any way.

What version are you running?
v3.3.9
## Diagnostics

[this code](https://gist.githubusercontent.com/wujunwei/0672ecf83f3cd72f8c8633e677851c8f/raw/0e6c7e99b78d248bcee62da9c0af3bf4ad731691/argo.go) can reproduce it .
```go
ctx, client, err := apiclient.NewClientFromOpts(
		apiclient.Opts{
			ArgoServerOpts: apiclient.ArgoServerOpts{
				URL:                "localhost:31092",  //nodeport
				Secure:             true,
				InsecureSkipVerify: true,
			},
			ClientConfigSupplier: func() clientcmd.ClientConfig { return GetConfig() },
			Context:              context.Background(),
		})
```

to fix this ,I have to add `AuthSupplier: func() string {return ""},` to `apiclient.Opts`

wujunwei:
Signed-off-by: wujunwei <wjw3323@live.com>

Fixes #9585

Fix panic: runtime error: invalid memory address or nil pointer dereference

### DIFF
--- a/pkg/apiclient/apiclient.go
+++ b/pkg/apiclient/apiclient.go
@@ -69,8 +69,14 @@ func NewClientFromOpts(opts Opts) (context.Context, Client, error) {
 		return nil, nil, fmt.Errorf("cannot use instance ID with Argo Server")
 	}
 	if opts.ArgoServerOpts.HTTP1 {
+		if opts.AuthSupplier == nil {
+			return nil, nil, fmt.Errorf("AuthSupplier cannot be empty when connecting to Argo Server")
+		}
 		return newHTTP1Client(opts.ArgoServerOpts.GetURL(), opts.AuthSupplier(), opts.ArgoServerOpts.InsecureSkipVerify, opts.ArgoServerOpts.Headers)
 	} else if opts.ArgoServerOpts.URL != "" {
+		if opts.AuthSupplier == nil {
+			return nil, nil, fmt.Errorf("AuthSupplier cannot be empty when connecting to Argo Server")
+		}
 		return newArgoServerClient(opts.ArgoServerOpts, opts.AuthSupplier())
 	} else {
 		if opts.ClientConfigSupplier != nil {


pkg/apiclient/apiclient.go
```diff
@@ -68,10 +68,19 @@ func NewClientFromOpts(opts Opts) (context.Context, Client, error) {
 	if opts.ArgoServerOpts.URL != "" && opts.InstanceID != "" {
 		return nil, nil, fmt.Errorf("cannot use instance ID with Argo Server")
 	}
+	var auth string
 	if opts.ArgoServerOpts.HTTP1 {
-		return newHTTP1Client(opts.ArgoServerOpts.GetURL(), opts.AuthSupplier(), opts.ArgoServerOpts.InsecureSkipVerify, opts.ArgoServerOpts.Headers)
+		if opts.AuthSupplier != nil {
```
terrytangyuan, line 73: Should we check auth supplier and return error instead? Otherwise it's `auth=""`

wujunwei: > Should we check auth supplier and return error instead? Otherwise it's `auth=""`

Of course, but we can only do something like below when there is no any Authentication.
```go
apiclient.Opts{
			ArgoServerOpts: ArgoServerOpts,
			AuthSupplier: func() string {
				return ""
			},
		})
```

alexec (approved):

terrytangyuan (approved):

PR #9617, opened 2022-09-19
ci: Auto-retry e2e tests

alexec:
Fixes #TODO


examples/data-transformations.yaml
```diff
@@ -3,8 +3,6 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: data-transformations-
-  labels:
```
alexec, line 6: cannot be retried

terrytangyuan: Do you know why this cannot be retried?

sarabala1979 (approved):

PR #9620, opened 2022-09-19
fix: Support Kubernetes v1.24. Fixes #8320

### LINKED ISSUE #8320: Kubernetes 1.24: Token-handling code assumes auto-created service account token secrets
Sweeping token-scraping of auto-generated Kubernetes token secrets in preparation for Kubernetes 1.24 showed the following code locations assume auto-generated tokens will exist:

https://github.com/argoproj/argo-workflows/blob/62e0a8ce4e74d2e19f3a9c0fb5e52bd58a6b944b/workflow/controller/operator.go#L3605

https://github.com/argoproj/argo-workflows/blob/a3c326fdf0d2133d5e78ef71854499f576e7e530/server/auth/webhook/interceptor.go#L90

https://github.com/argoproj/argo-workflows/blob/a3c326fdf0d2133d5e78ef71854499f576e7e530/server/auth/gatekeeper.go#L321

That assumption is not universally correct.

In 1.21+, secret-based tokens are no longer used for mounting into pods (ephemeral time-limited tokens are), and the token controller can be turned off.

In 1.24+, secret-based tokens are no longer auto-created by default for new service accounts.

Using ephemeral time-bound tokens is preferred in 1.21+ (see the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API) if possible.

If a secret-based token is still desired, one can be [created manually](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token), but will not be referenced from the service account's .secrets list.

---

alexec:

Fixes #8320

Changes:

* Update Github Actions to use a good looking action for K3S, running latest K3S.
* Reduced build time by 3m.
* Added diagnostics to builds.


terrytangyuan:
Will continue the work in https://github.com/argoproj/argo-workflows/pull/9680

docs/manually-create-secrets.md
```diff
@@ -0,0 +1,40 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: default.service-account-token
+  annotations:
+    kubernetes.io/service-account.name: default
+type: kubernetes.io/service-account-token
+```
+
+This option is simpler that option 2, as you can combine creating the secret with making is discoverable by name.
```
alexec, line 25:
```suggestion
This option is simpler than option 2, as you can combine creating the secret with making it discoverable by name.
```

alexec:
We're getting close!

docs/manually-create-secrets.md
```diff
@@ -0,0 +1,40 @@
...
+
+Annotate the service account with the secret name:
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: default
+  annotations:
+    kubernetes.io/service-account-token.name: my-token
```
liggitt, line 37: Please don't use non-standard kubernetes.io-namespaced annotations. If this is the annotation argo will look for to find the secret, please use an argo-specific annotation name to hold this

.github/workflows/ci-build.yaml
```diff
@@ -136,20 +140,49 @@ jobs:
...
-        timeout-minutes: 4
+        timeout-minutes: 6
         name: Wait for MinIO/MySQL etc to be ready
       - name: Run tests ${{matrix.test}}
         # https://github.com/marketplace/actions/retry-step
         uses: nick-fields/retry@v2.8.1
         with:
           timeout_minutes: 20
-          max_attempts: 2
+          max_attempts: 1
```
alexec, line 150: revert

manifests/quick-start/mysql/mysql-deployment.yaml
```diff
@@ -28,10 +29,10 @@ spec:
               value: "yes"
           ports:
             - containerPort: 3306
-          readinessProbe:
-            exec:
-              command: ["mysql", "-u", "mysql", "-ppassword", "argo", "-e", "SELECT 1"]
-            initialDelaySeconds: 15
-            timeoutSeconds: 2
+          #readinessProbe:
```
alexec, line 32: reinstate

docs/manually-create-secrets.md
alexec: This is a standard one:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token

.github/workflows/ci-build.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
     name: E2E Tests
     runs-on: ubuntu-latest
     timeout-minutes: 25
-    needs: [ tests, argoexec-image ]
+    needs: [ argoexec-image ]
```
alexec, line 62: 3m faster

docs/manually-create-secrets.md
liggitt: that is `kubernetes.io/service-account.name`, not `kubernetes.io/service-account-token.name`, and is [defined for use on Secret objects](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-service-account-name) pointing back at the service account, not the other way around

alexec: Oh. You’re right. It should be `workflows.argoproj.io/service-account-token.name`. Which makes me wonder how any tests are passing.

manifests/quick-start/mysql/mysql-deployment.yaml
```diff
@@ -28,10 +29,5 @@ spec:
               value: "yes"
           ports:
             - containerPort: 3306
-          readinessProbe:
```
alexec, line 31: I could not get this to work with K3s v1.24, so I gave up and removed it.

docs/manually-create-secrets.md
alexec: Fixed. It’s only wrong in the docs.

alexec:
Phew. That was epic.

docs/manually-create-secrets.md
```diff
@@ -0,0 +1,40 @@
+# Kubernetes Secrets
+
+As of Kubernetes v1.24, secrets are no longer automatically created for service accounts.
+
+You must create the
```
terrytangyuan, line 5: missing the rest of the sentences?

.github/workflows/ci-build.yaml
```diff
@@ -103,15 +103,19 @@ jobs:
         with:
           python-version: '3.x'
           cache: pip
-      - name: Install and start K3S
+      - uses: AbsaOSS/k3d-action@v2
```
terrytangyuan, line 106: Is there anywhere to explicitly specify the k3s version?

terrytangyuan (commented):
Thanks! LGTM overall. Just a couple of small comments. We should also test this to make sure it still works in older versions of k8s.

.github/workflows/ci-build.yaml
alexec: Good spot. I switched from manual install. 

https://github.com/AbsaOSS/k3d-action

I'm now not clear what K3s we're running. I need to re-visit this.

alexec: Hmmm. `docker.io/rancher/k3s:v1.22.7-k3s1`

.github/workflows/ci-build.yaml
alexec, line 62: `argoexec-image` is 3m faster than `tests`, so this speeds up the build by 3m

hack/free-port.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/sh
-set -eu
+set -eu -o pipefail
 
 port=$1
 
-lsof -s TCP:LISTEN -i ":$port" | grep -v PID | awk '{print $2}' | xargs -L 1 kill || true
+pids=$(lsof -t -s TCP:LISTEN -i ":$port" || true)
```
alexec, line 6: this wrote a lot of errors on linux

alexec:
@terrytangyuan green!

terrytangyuan:
Actually just one last failing test!

.github/workflows/ci-build.yaml
```diff
@@ -86,7 +86,19 @@ jobs:
             profile: minimal
           - test: test-python-sdk
             profile: minimal
+          - test: test-executor
+            install_k3s_version: v1.21.2+k3s1
```
terrytangyuan, line 90: Hmm this is not right?

alexec: This is the version were running

terrytangyuan: We should be testing 1.24

alexec: It defaults to latest (v1.25 ATM). These are for testing backwards compatibility.

docs/manually-create-secrets.md
```diff
@@ -0,0 +1,40 @@
...
+metadata:
+  name: default.service-account-token
+  annotations:
+    kubernetes.io/service-account.name: default
+type: kubernetes.io/service-account-token
+```
+
+This option is simpler than option 2, as you can combine creating the secret with making it discoverable by name.
+
+## Option 2 - Discovery By Annotation
```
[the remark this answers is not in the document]
alexec: This is only if there are long service account names.

.github/workflows/ci-build.yaml
terrytangyuan: I see. Can we rename these tests to differentiate them from the existing ones? It might be better to explicitly set the version so that we know what change breaks the tests?

alexec: Yes/no. I wast trying to avoid forcing all the required checks to be changed. That impacts all PRs, which would need to be synced with master.

Using latest reduces maintenance, we don’t need to update it.

terrytangyuan (approved):
LGTM. @sarabala1979 Take another look?

PR #9666, opened 2022-09-23
fix: improve semaphore concurrency performance

jessesuen:
Our semaphore implementation, can heavily underutilize the max semaphore counter in the event that multiple workflows are **_simultaneously finishing_** at the same time. 

We use a priority semaphore which will only allow the next workflow in line to acquire the semaphore. But when a workflow tries to acquire the semaphore and fails because it is not in the front of the line, it will go back to sleep and rely on someone else to enqueue it. There exists a problem when a batch of waiting workflows get enqueued at the same time, but only one of them acquires it, and at the current moment theres nothing to requeue the others if there is still room.

Secondly, our `nextWorkflow` function enqueues the next workflow immediately. This means that the next workflow will reconcile at a point in time when the semaphore metadata may not yet reflect any increments back to the semaphore counter. By adding a short delay of 1s, we would have completed our metadata update and the next workflow will properly know it can run. 1 second is enough to finish up the current set of function calls, update counters, etc...

This change improves semaphore capacity performance in two ways:
1. We now enqueue the next N workflows after the current workflow acquires the semaphore and there is still more room for others
2. When enqueuing the next workflow, we should not enqueue it while we have yet to finish the semaphore book keeping. This is because the next workflow will immediately discover it is blocked (because current workflow has not yet updated semaphore metadata). By delaying the enqueuing of the next workflow by using AddAfter(), we will enqueue it at a point in time when the semaphore metadata has been updated.

To reproduce:
1. submit 20 workflows referencing a semaphore configmap of size 10. the workflow can just be a single pod `sleep 20` 
2. The first batch of 10 workflows will run. But after the workflow completes after the `sleep 20`, you will notice that only 1-2 workflows (of the next 10) will start running

Signed-off-by: Jesse Suen <jesse@akuity.io>


workflow/controller/controller.go
```diff
@@ -333,7 +333,12 @@ func (wfc *WorkflowController) createSynchronizationManager(ctx context.Context)
 	}
 
 	nextWorkflow := func(key string) {
-		wfc.wfQueue.AddRateLimited(key)
+		// We now use AddAfter(1s) instead of AddRateLimited() because if we add immediately,
+		// the enqueued workflow will likely be reconciled before we have finished reconciling
+		// *this* workflow. If we reconcile the next workflow too soon (before we have finished
+		// incremented the semaphore counters), the next workflow will believe it cannot run.
+		wfc.wfQueue.AddAfter(key, 1*time.Second)
+		//wfc.wfQueue.AddRateLimited(key)
```
terrytangyuan, line 341: Remove this?

workflow/sync/semaphore.go
```diff
@@ -188,6 +188,11 @@ func (s *PrioritySemaphore) tryAcquire(holderKey string) (bool, string) {
 	if s.acquire(holderKey) {
 		s.pending.pop()
 		s.log.Infof("%s acquired by %s ", s.name, nextKey)
+		if s.pending.Len() > 0 && len(s.lockHolder) < s.limit {
```
terrytangyuan, line 191: This may not be as effective if the number of pending workflows is large, maybe we should do a for loop here whenever there’s room to enqueue the next workflows.

workflow/controller/controller.go
```diff
@@ -333,7 +333,13 @@ func (wfc *WorkflowController) createSynchronizationManager(ctx context.Context)
...
 	nextWorkflow := func(key string) {
-		wfc.wfQueue.AddRateLimited(key)
+		// We now use AddAfter(1s) instead of AddRateLimited() because if we add immediately,
+		// the next workflow will likely be reconciled before we have finished reconciling
+		// *this* workflow. If we reconcile the next workflow too soon, we will do so before
+		// we have finished incrementing the semaphore counters as part of this workflow, and
+		// so the next workflow will believe it cannot run. By delaying for 1s, we will reconcile
+		// the next workflow at a point in time when semaphore counters have been updated.
+		wfc.wfQueue.AddAfter(key, 1*time.Second)
 	}
```
terrytangyuan, line 343: We probably want to expose this as an environment variable.

workflow/sync/semaphore.go
```diff
@@ -175,11 +186,10 @@ func (s *PrioritySemaphore) tryAcquire(holderKey string) (bool, string) {
 	// If it is not a front key, it needs to wait for its turn.
 	if s.pending.Len() > 0 {
 		item := s.pending.peek()
-		nextKey = fmt.Sprintf("%v", item.key)
-		if holderKey != nextKey && !isSameWorkflowNodeKeys(holderKey, nextKey) {
+		if holderKey != nextKey && !isSameWorkflowNodeKeys(holderKey, item.key) {
 			// Enqueue the front workflow if lock is available
 			if len(s.lockHolder) < s.limit {
-				s.nextWorkflow(nextKey)
```
jessesuen, line 182: this appears to be a bug with notify implementation. nextKey might be in the form of `namespace/workflowname/nodeid`, but we must enqueue it as `namespace/workflowname`

workflow/sync/semaphore.go
jessesuen: I didn't notice your concern during testing, but it does not hurt to enqueue other waiters, so I implemented the suggestion.

workflow/sync/semaphore.go
```diff
@@ -175,11 +186,10 @@ func (s *PrioritySemaphore) tryAcquire(holderKey string) (bool, string) {
 	// If it is not a front key, it needs to wait for its turn.
 	if s.pending.Len() > 0 {
 		item := s.pending.peek()
-		nextKey = fmt.Sprintf("%v", item.key)
```
jessesuen, line 178: item.key is already a string, so I didn't understand the purpose of `fmt.Sprintf("%v", item.key)`

terrytangyuan (approved):

jessesuen:
@terrytangyuan will you handle cherry-picking this to v3.4?

terrytangyuan:
Yeah we can cherry pick this. Usually @alexec or @sarabala1979 will create an issue to discuss which commits to cherry-pick since the last release.

sarabala1979:
sure I will release 3.4.1 this friday

PR #9680, opened 2022-09-25
fix: Support Kubernetes v1.24. Fixes #8320

### LINKED ISSUE #8320: Kubernetes 1.24: Token-handling code assumes auto-created service account token secrets

terrytangyuan:
Fixes #8320. On top of existing work in https://github.com/argoproj/argo-workflows/pull/9620.

Co-authored-by: Alex Collins <alex_collins@intuit.com>


jessesuen (commented):
Will this change  bump the workflow minimum k8s version to something like v1.22 (when the TokenRequestAPI was marked stable)?

workflow/controller/operator.go
```diff
@@ -3686,10 +3688,7 @@ func (woc *wfOperationCtx) getServiceAccountTokenName(ctx context.Context, name
 	if err != nil {
 		return "", err
 	}
-	if len(account.Secrets) == 0 {
-		return "", fmt.Errorf("service account %s/%s does not have any secrets", account.Namespace, account.Name)
-	}
-	return account.Secrets[0].Name, nil
+	return secrets.SecretName(account), nil
```
jessesuen, line 3691: I think we are breaking backwards compatibility here.

Users of v1.23 and below, are not conditioned to generate ServiceAccount secrets yet and so having the assumption that there exists the secret by naming convention is going to break a lot of.

I think the implementation needs to have both implementations and preserve the older v1.23 assumed behavior for backward compatibility

docs/manually-create-secrets.md
```diff
@@ -0,0 +1,5 @@
+# Kubernetes Secrets
+
+As of Kubernetes v1.24, secrets are not automatically created for service accounts by default. [Find out how to create these yourself manually](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token).
+
+Argo discovers your token by name, not annotation. They must be named `${serviceAccountName}.service-account-token`.
```
jessesuen, line 5: I think we could do better job explaining when users need to do this. When using Agent http template? Auth delegation in api server? Does this need to be mentioned in the SSO docs?

jessesuen (commented):
I am wondering if we have considered alternative techniques that doesn't require user to create a ServiceAccount Secret as an added step for v1.24. For example, using the token request API?

terrytangyuan:
@alexec @sarabala1979 Any thoughts on Jesse's questions above? Is the token request API something you thought about when agent was initially designed/implemented?

terrytangyuan:
I'll investigate the approach that uses the TokenRequest API. It seems like a better option instead of using a token that never expires.

sarabala1979:
@terrytangyuan Mainly we need to think about the API calls increase. Will Controller call once to create a secret for SA or  Will it call every reconcile time?

terrytangyuan:
> @terrytangyuan Mainly we need to think about the API calls increase. Will Controller call once to create a secret for SA or Will it call every reconcile time?

We'll need to manage the lifecycle of tokens so that they can be rotated and re-requested only when necessary (e.g. based a configured timeout/expiration for bearer tokens) instead of making requests frequently.

terrytangyuan:
Remaining issues:
* `ExecutorPluginsSuite/TestTemplateExecutor` is failing: https://github.com/argoproj/argo-workflows/actions/runs/3160377225/jobs/5160709121
* `ArtifactsSuite/TestArtifactGC` is failing: https://github.com/argoproj/argo-workflows/actions/runs/3160377225/jobs/5160708864
* `make wait` fails randomly. As an example, `test-corefunctional` failed but succeeded after retry:
    * https://github.com/argoproj/argo-workflows/actions/runs/3124105856/jobs/5067786142
    * https://github.com/argoproj/argo-workflows/actions/runs/3129444039/jobs/5078620426

agilgur5:
For reference, #9620 was re-opened and completed after this was closed, so marking this as superseded

PR #9686, opened 2022-09-26
fix: Label on Artifact GC Task no longer exceeds max characters

### LINKED ISSUE #9628: artifactgc not working because of labels
### Pre-requisites

expected artifact to be gc'd

but see "Logs from the workflow controller" below, nothing on the UI indicates issue except that the workflow stays in 'workflows 
 UI' instead of moving to 'archived workflows UI', also the artifact still remains on s3

### Version

3.4.0

```YAML
n/a
```

`time="2022-09-20T04:34:37.029Z" level=error msg="failed to GC artifacts" error="failed to Create WorkflowArtifactGCTask '' for Garbage Collection: WorkflowArtifactGCTask.argoproj.io \"bitbucket-workflow-20220920t1414421000pbjt2-artgc-wfcomp-2166136261-0\" is invalid: metadata.labels: Invalid value: \"bitbucket-workflow-20220920t1414421000pbjt2-artgc-wfcomp-2166136261\": must be no more than 63 characters" namespace=auth workflow=bitbucket-workflow-20220920t1414421000pbjt2
`

juliev0:
Fixes #9628 

This is the label that the Controller puts on the Artifact GC Tasks and then passes as an Environment Variable to the Artifact GC Pod so that it can identify the Tasks. The only requirement really is that it's unique for each GC Pod, so by hashing the pod name it's at least as unique as any 32 bit hash.


tooptoop4 (approved):

juliev0:
> @juliev0 Can you add test for this?

Do you just mean a unit test which confirms that the label is < 63 characters?
(We do already have end to end tests to confirm that Artifact GC is working, which are passing)

test/e2e/testdata/artifactgc/artgc-multi-strategy-multi-anno.yaml
```diff
@@ -1,9 +1,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
-  generateName: two-artgc-
-  #finalizers:
-  #  - "blah"
+  generateName: testing-on-completion-and-on-deletion-strategies-
```
juliev0, line 4: made this workflow name really long as a way of testing this as part of the e2e test. I confirmed it fails on master and passes here.

juliev0:
> @juliev0 Can you add test for this?

Done. I modified the e2e test workflow to make the workflow name really long. I confirmed it fails on master and passes here.

sarabala1979 (approved):

tooptoop4:
@sarabala1979 can u pls merge this?

PR #9693, opened 2022-09-27
feat: Added workflow summary to workflow-list page.

### LINKED ISSUE #6535: Add summary to workflows list page
Like Apache Fink:

image

anil10032:
Signed-off-by: Anil Kumar <anil225sri@gmail.com>

Adds workflows summary with the status of jobs and their count in the view. The counts get updated in real-time as and when the jobs get added. Fixes #6535

Check the screenshot below to see how it gets rendered in the UI.

Screenshot 2022-10-07 at 10 39 09 PM

**NOTE:** The summary is for the workflows present in the view. It gets updated as the view gets updated on filters, pagination, etc

Examples:

Screenshot 2022-10-07 at 10 55 58 PM

Screenshot 2022-10-07 at 10 57 15 PM

Screenshot 2022-10-07 at 10 57 36 PM


anil10032:
@alexec @terrytangyuan Can one of you review this PR? Or assign it to someone who can review the UI changes.

Thanks!

terrytangyuan (commented):
Looks fine to me. @alexec Could you take another look?

ui/src/app/workflows/components/workflows-summary-container/workflows-summary-container.tsx
```diff
@@ -0,0 +1,51 @@
...
+            setWfSummary(summary);
+        }
+    }, [props.workflows]);
+
+    return (
+        <div className='wf-summary-container'>
+            <p className='wf-summary-container__title'>Workflows Summary</p>
+            <div className='row'>
+                <div className='columns small-12 xlarge-12'>
+                    <span className='wf-summary-container__text'>Running Jobs &nbsp;</span>
```
terrytangyuan, line 25: Running workflows

anil10032:
@alexec Can you take a look at this please?

ui/src/app/workflows/components/workflows-summary-container/workflows-summary-container.scss
```diff
@@ -0,0 +1,45 @@
+@import 'node_modules/argo-ui/src/styles/config';
+
+.wf-summary-container {
```
alexec, line 3: can we avoid using new styles?

anil10032: Are you suggesting using inline styles?

alexec: I’m suggesting no new styles.

anil10032: Removing the new styles and adding the available `white-box` class to the outer div looks like this.

Screenshot 2022-09-30 at 7 20 50 AM

I think some inline styling is needed if no new styles are allowed.

alexec:
What about putting it above the list?

alexec:
Will this work? It does not know about all workflows. So is the summary right?

anil10032:
> Will this work? It does not know about all workflows. So is the summary right?

This works on the workflows that are available with the state. The numbers reflect the workflows in the current view.

If we need to show the summary for the entire list irrespective of the filters applied...then the implementation needs to be changed.

anil10032:
@alexec Can you share the expectations of issue #6535? Please close this PR if you feel it is not addressing the issue.

alexec:
Ok. This is what I think:

* Remove the term “Job”. They’re workflows. 
* Just show summary for workflows in view (i.e. no change).
* Add an <InfoIcon/> to example only workflows in view are shown. That was the user are not confused by what they see.

anil10032:
@alexec I have made changes as per your request. Also updated the PR description and included some examples.

Please take a look.

alexec:
Ops. I added a code sample, but it was hidden because of my use of `<>`:

* Add an <InfoIcon/> to the summary. When hovered, this should say "Only workflows in view are summarized". This will help user not be confused by what they see, if they expected all workflows.

anil10032:
It looks like this now.

image

alexec (approved):

anil10032:
@alexec Can you merge this, please?

Issue #8790, opened 2022-05-18
duplicative env var `ARGO_TEMPLATE` result in etcd size 1MB limit


MagicFrogSJTU:
## Summary
I am running a containerSet with lots of steps. And I have large parameter input for the step.
Because there is an env variable called ARGO_TEMPLATE each container, the overall pod description is too large to fit in etcd, exceeding the 1MB limit.
(10 + init + wait) = 12 containers
12 * 100k = 1200k > 1MB
What is the usage of ARGO_TEMPLATE? Can we delete it?

I am using the v3.1.10

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

alexec:
> What is the usage of ARGO_TEMPLATE? Can we delete it?

No.

However, we could replace it with just the information needed by the emissary. The emissary only needs to know which containers it must wait on. It does not need the whole template.

MagicFrogSJTU:
That would be nice. Look forward to the coming improve!

terrytangyuan:
> The emissary only needs to know which containers it must wait on.

@alexec Looks like tests are breaking after I removed the parameters. https://github.com/argoproj/argo-workflows/pull/9698 Any hints on what I may have missed?

juliev0:
At Intuit we are actually relying on the $ARGO_TEMPLATE environment variable `Inputs` because we had an issue in which we were originally trying to do this in our Step:
```
image: "docker.intuit.com/dev/patterns/kubernetes/dev/kubectl-awscli:v1.22.15"
command:
  - sh
  - -c
args:
  - |
    echo "{{`{{inputs.parameters.manifest}}`}}" | ....
```

but we found that we exceeded the maximum argument size limit allowed by Linux. So, we changed the Step to use the ARGO_TEMPLATE environment variable:
```
image: "docker.intuit.com/dev/patterns/kubernetes/dev/kubectl-awscli:v1.22.15"
command:
  - sh
  - -c
args: ["
  getconf ARG_MAX;
  echo $ARGO_TEMPLATE | yq '.inputs.parameters[0].value' --unwrapScalar > /tmp/manifest-64;
  ....
"]
```

juliev0:
(We had also tried an alternative solution of mounting the Input parameter as a RawArtifact but that faced an issue and didn't work either)

juliev0:
now perhaps there could be an alternative solution in which the value of the environment variable is compressed or something...

[...]

PR #9698, opened 2022-09-27
fix: Avoid hitting etcd limit for workflow pod with large parameters

terrytangyuan:
Fixes https://github.com/argoproj/argo-workflows/issues/8790.


tooptoop4:
🙏

juliev0:
added a [comment](https://github.com/argoproj/argo-workflows/issues/8790#issuecomment-1743614552) to the original issue on how we're relying on the ARGO_TEMPLATE inputs

tooptoop4:
@terrytangyuan could we have an env flag to enable omitting this?

tooptoop4:
would line 325 break as it expects inputs? 
```pod, err = substitutePodParams(pod, woc.globalParams, tmpl)```

terrytangyuan:
There are certainly things breaking because of this change. The tests were not passing. If someone wants to pick this up, feel free to do so.

PR #9729, opened 2022-10-01
feat (executor, controller): reduce poll time

tooptoop4:
Improve e2e workflow performance by reducing sleep durations


alexec:
I’m sorry. The risk of negative impact is high with this PR and is fundamentally changes how the controller behaves. I don’t think the benefit of faster tests makes that worthwhile.

Making the tests faster and more robust is a really good idea. Thank you for trying to help here.

tooptoop4:
just want to make my own workflows more snappy

PR #9740, opened 2022-10-03
fix: node links on UI use podName instead of workflow name

dpadhiar:

Currently on the UI, links on the UI from the node use `workflow.name` instead of the `podName`. This causes issue with Splunk links for example that need the podName (which includes step name as well) to filter for data.


terrytangyuan (approved):

vladlosev:
Is is possible to cherry-pick this into the 3.4 branch? Thanks!

PR #9753, opened 2022-10-06
fix: `value` is required when parameter is of type `enum`

missingcharacter:
Without this change we get error:
`Recovered from panic: runtime error: invalid memory address or nil pointer dereference`

Signed-off-by: Ricardo Rosales <728243+missingcharacter@users.noreply.github.com>


terrytangyuan (approved):

missingcharacter:
@terrytangyuan yes, I will, 1 moment

terrytangyuan (approved):

Issue #9742, opened 2022-10-03
Expr expressions can't make use of "workflow.parameters" 


juliev0:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

In [this](https://github.com/argoproj/argo-workflows/blob/master/util/expand/expand.go#L18) function we remove "workfow.parameters" from our map if it contains "workflow.parameters.x". This means we can't evaluate an _expr_ expression if it contains "workflow.parameters". 

In the Workflow example I've included, the result of "argo logs" is:
```
 % argo logs global-parameters-64mnf 
global-parameters-64mnf: time="2022-10-03T22:45:08.830Z" level=info msg="capturing logs" argo=true
global-parameters-64mnf:  __________________________________ 
global-parameters-64mnf: / {{=jsonpath(workflow.parameters, \
global-parameters-64mnf: \ '$.value')}}                   /
global-parameters-64mnf:  ---------------------------------- 
global-parameters-64mnf:     \
global-parameters-64mnf:      \
global-parameters-64mnf:       \     
global-parameters-64mnf:                     ##        .            
global-parameters-64mnf:               ## ## ##       ==            
global-parameters-64mnf:            ## ## ## ##      ===            
global-parameters-64mnf:        /""""""""""""""""___/ ===        
global-parameters-64mnf:   ~~~ {~~ ~~~~ ~~~ ~~~~ ~~ ~ /  ===- ~~~   
global-parameters-64mnf:        \______ o          __/            
global-parameters-64mnf:         \    \        __/             
global-parameters-64mnf:           \____\______/   
global-parameters-64mnf: time="2022-10-03T22:45:09.847Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: global-parameters-
spec:
  entrypoint: whalesay1
  arguments:
    parameters:
    - name: message
      value: 99

  templates:
  - name: whalesay1
    container:
      image: docker/whalesay:latest
      command: [cowsay]
[...]

Issue #8911, opened 2022-06-03
Error with expr from parameter output another step


DiegoFlop:
## Checklist

 

* [x] Double-checked my configuration.
* [x] Tested using the latest version.
* [x] Used the Emissary executor.

## Summary

What happened/what you expected to happen?

When on a when field add an expr with output task I get the next error:

>  Failed: invalid spec: templates.main when doesn't support 'expr' expression '{{='. 'When' expression is only support govaluate format {{

the expr I'm using is:

> when: "{{=jsonpath(tasks.checkout.outputs.parameters.custom_config, '$.steps.test_unit.skip')}}"

However when I use and input parameter with expr everything works correctly, but I don't have use an input parameter because, I need use the output parameter to another task and I get a json file.

> when: "{{=inputs.parameters.BRANCH == 'false'}}"

Which this expr works correctly

What version are you running?

Argo-workflows: 3.3.6
Kubernetes: 1.22.9

## Diagnostics

Paste the smallest workflow that reproduces the bug. We must be able to run the workflow.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: example
  entrypoint: main
  templates:
    - name: main
      inputs:
        parameters:
          - name: URL
          - name: BRANCH 
      dag:
        tasks:
        - name: checkout
          templateRef:
            name: checkout
            template: main
          arguments:
            parameters:
              - name: git_ssh_url
                value: "git@{{= sprig.replace('net/', 'net:', sprig.substr( 8, -1, inputs.parameters.URL)) }}"
              - name: BRANCH
                value: "{{inputs.parameters.BRANCH}}"
...
...
        - name: test-unit
          templateRef:
            name: unit-test
            template: node-web
          arguments:
            parameters:
              - name: PROJECT_NAME
                value: "{{workflow.parameters.PROJECT_NAME}}"
# Get value for testing
              - name: SKIP
                value: "{{=jsonpath(tasks.checkout.outputs.parameters.custom_config, '$.steps.test_unit.skip')}}"
          when: "{{=jsonpath(tasks.checkout.outputs.parameters.custom_config, '$.steps.test_unit.skip') == 'false'}}"

...
```

```bash
kubectl logs -n argo-workflows deploy/workflow-controller | grep example-db6tm

# If the workflow's pods have not been created, you can skip the rest of the diagnostics.

# The workflow's pods that are problematic:
[...]

PR #6410, opened 2021-07-23
fix(controller): Randomly expr expression fail to resolve

sarabala1979:
Signed-off-by: Saravanan Balasubramanian <sarabala1979@gmail.com>
closes #6257 
Checklist:

* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md).

Tips:

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it does not need to pass
* Sign-off your commits to pass the DCO check: `git commit --signoff`.
* Run `make pre-commit -B` to fix codegen or lint problems. 
* Say how how you tested your changes. If you changed the UI, attach screenshots.

PR #9757, opened 2022-10-06
fix: enable when expressions to use expr; add new json variables to avoid expr conflicts

juliev0:
Fixes https://github.com/argoproj/argo-workflows/issues/8911
Fixes https://github.com/argoproj/argo-workflows/issues/9742

This removes the validation check which prevents expr from being used in a "when" condition, which had gotten added [here](https://github.com/argoproj/argo-workflows/pull/6410/files).

Apparently, the actual problem that was being addressed there was the fact that "workflow.parameters" couldn't be used in an expr expression. The issue was that there were conflicts since both "workflow.parameters" and "workflow.parameters.x" were variables that need replacing, and the call to evaluate expr can't handle cases in which one variable is a prefix of another. Therefore, this adds a new variable "workflow.parameters.json" which is equivalent to "workflow.parameters". "workflow.parameters" should eventually be phased out in favor of this. (For now, we can keep it: it's just that it can't be handled in an expr expression.)


agilgur5:
Replaced by #9761

PR #9760, opened 2022-10-06
fix: SSO integration not considering HTTP_PROXY when making requests. Fixes #9259

### LINKED ISSUE #9259: The SSO integration does not consider HTTP_PROXY env vars when making requests
## Checklist

## Summary

The `github.com/coreos/go-oidc/v3/oidc` library by default uses an http client that is not reading the env vars for `HTTP_PROXY` variables. This results in SSO failing to initialize at startup.

## Diagnostics

I added several `log.Print` statements in my fork of the `release-3.3` branch and I discovered that the root cause is occuring within the `oidc.NewProvider` call:

```go
func providerFactoryOIDC(ctx context.Context, issuer string) (providerInterface, error) {
	log2.Print("Attemping to create new OIDC provider")
	provider, err := oidc.NewProvider(ctx, issuer)
	log2.Print("Finished oidc.NewProvider(ctx, issuer)")
	return provider, err
}
```

By setting the context myself with the native http client of go, it works correctly:

```go
func providerFactoryOIDC(ctx context.Context, issuer string) (providerInterface, error) {
	log2.Print("Attemping to create new OIDC provider")
	// Set the client context explicitly in order to use proxy configuration from environment(if any)
	// See https://github.com/coreos/go-oidc/blob/8d771559cf6e5111c9b9159810d0e4538e7cdc82/oidc.go#L43-L53
	oidc_ctx := oidc.ClientContext(ctx, &http.Client{})
	provider, err := oidc.NewProvider(oidc_ctx, issuer)
	log2.Print("Finished oidc.NewProvider(ctx, issuer)")
	return provider, err
}
```

---

rohankmr414:
Signed-off-by: Rohan Kumar <rohan@pipekit.io>

Fixes #9259

### DIFF
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -79,7 +79,8 @@ type providerInterface interface {
 type providerFactory func(ctx context.Context, issuer string) (providerInterface, error)
 
 func providerFactoryOIDC(ctx context.Context, issuer string) (providerInterface, error) {
-	return oidc.NewProvider(ctx, issuer)
+	oidcCtx := oidc.ClientContext(ctx, &http.Client{})
+	return oidc.NewProvider(oidcCtx, issuer)
 }
 
 func New(c Config, secretsIf corev1.SecretInterface, baseHRef string, secure bool) (Interface, error) {


alexec (changes requested):
How tested please?

alexec (approved):

PR #9763, opened 2022-10-07
docs: add Gepardec to users

fhochleitner:
As we are heavily using the argo ecosystem for the management of our OpenShift Clusters and applications as well as implementing our CICD solutions we would like to reflect that on the users list :)


terrytangyuan (approved):
Thanks!

PR #9789, opened 2022-10-11
fix: Add --tls-certificate-secret-name parameter to server command. Fixes #5582 

### LINKED ISSUE #5582: TLS configuration and/or docs enhancement

chtcvl:
Signed-off-by: vladimir.ivanov <vladimir.ivanov@grasshopperasia.com>

Fixes #5582 

Add --tls-certificate-secret-name parameter to server command.
Since server runs on localhost:2746, get an error in the UI "certificate is valid for ... not for localhost". Therefore the merge request set InsecureSkipVerify on TLSconfig for grpc client on the gateway.

Tested to to work with secret generated by cert-manager.

### DIFF
--- a/server/apiserver/argoserver.go
+++ b/server/apiserver/argoserver.go
@@ -334,7 +334,10 @@ func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServe
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxGRPCMessageSize)),
 	}
 	if as.tlsConfig != nil {
-		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(as.tlsConfig)))
+		tlsConfig := as.tlsConfig.Clone()
+		tlsConfig.InsecureSkipVerify = true
+		dCreds := credentials.NewTLS(tlsConfig)
+		dialOpts = append(dialOpts, grpc.WithTransportCredentials(dCreds))
 	} else {
 		dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	}
--- a/cmd/argo/commands/server.go
+++ b/cmd/argo/commands/server.go
@@ -46,7 +46,7 @@ func NewServerCommand() *cobra.Command {
 		baseHRef                 string
 		secure                   bool
 		tlsCertificateSecretName string
-		htst                     bool
+		hsts                     bool
 		namespaced               bool   // --namespaced
 		managedNamespace         string // --managed-namespace
 		enableOpenBrowser        bool
@@ -153,7 +153,7 @@ See %s`, help.ArgoServer),
 			opts := apiserver.ArgoServerOpts{
 				BaseHRef:                 baseHRef,
 				TLSConfig:                tlsConfig,
-				HSTS:                     htst,
+				HSTS:                     hsts,
 				Namespaced:               namespaced,
 				Namespace:                namespace,
 				Clients:                  clients,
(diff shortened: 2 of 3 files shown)


server/apiserver/argoserver.go
```diff
@@ -334,7 +334,10 @@ func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServe
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxGRPCMessageSize)),
 	}
 	if as.tlsConfig != nil {
-		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(as.tlsConfig)))
+		tlsConfig := as.tlsConfig
+		tlsConfig.InsecureSkipVerify = true
```
alexec, line 338: should you delete this line?

chtcvl: This `InsecureSkipVerify` is for communication within the same instance of argo-server process - between grpc-gateway and grpc-server goroutines (since server is serving on `localhost` and we can't have it in certificate SAN for publically trusted certificate). Setting it as `false` will break the gateway.
I provided more reasoning and alternative implementations at https://github.com/argoproj/argo-workflows/pull/9423#discussion_r970628566

tczhao: I think the correct place to do this is here instead
https://github.com/argoproj/argo-workflows/blob/9a32232a85ce0f24719a6aaa46eef86f86a500b4/util/tls/tls.go#L147-L149

tczhao (changes requested):

server/apiserver/argoserver.go
chtcvl: `func GetServerTLSConfigFromSecret` returns a **server** TLS config with proper certificates. Here we are patching **client** TLS config (dialOpts) for aforementioned reasons. 

We set `InsecureSkipVerify` on server TLS config as well, but this is unnecessary and can become a security risk in the long run.

server/apiserver/argoserver.go
```diff
@@ -334,7 +334,10 @@ func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServe
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxGRPCMessageSize)),
 	}
 	if as.tlsConfig != nil {
-		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(as.tlsConfig)))
+		tlsConfig := as.tlsConfig
```
alexec, line 337: Will `as.tlConfig` be modified? Do you need to do `copy()` or similar?

chtcvl: Fixed with `Clone()`

alexec (approved):
LGTM. I'd like a second pair of eyes on this. @juliev0 ?

juliev0 (approved):

chtcvl:
I don't seem to be able to merge this PR as it requires write access. Would appreciate help merging this.

PR #9807, opened 2022-10-13
fix: exit code always be '0' in windows container. Fixes #9797

### LINKED ISSUE #9797: [bug] The command exit code always be '0' in windows container
The command exit code always be '0' in windows container. Because the wait function on windows only check the err but not with the exit code of the stat (The error might be redirect).
[https://github.com/argoproj/argo-workflows/blob/aa59b43748f78e599709add871af7ec14e1fd3c1/workflow/executor/os-specific/signal_windows.go#L29](url)

I tried to fix this by the code below:
``` go
func Wait(process *os.Process) error {
	stat, err := process.Wait()
	if stat.ExitCode() != 0 {
		var errStr string
		if err != nil {
			errStr = err.Error()
		} else {
			errStr = "<nil>"
		}

		return errors.New(fmt.Sprintf("Exited failed. Error:%s", errStr))
	}
	return nil
}

```
This is just a temporary fix, so I just show my code here. Hope helpful.

The yaml below will reproduce the issue.

``` yaml
kind: WorkflowTemplate
metadata:
  namespace: default
  name: mytemplate

spec:
  templates:
    - name: test
      template: test-main
          

    - name: test-main
      nodeSelector:
        kubernetes.io/os: windows
      container:
        - name: main
          image: mcr.microsoft.com/windows/nanoserver:ltsc2022 # I used the nanoserver-ltsc2022 but the issue should also exist on ltsc2019
          command: ["cmd", "/s", "/c"]
          args: ["call exit /b 1"]
```

wsj0000:
Fixes https://github.com/argoproj/argo-workflows/issues/9797

### DIFF
--- a/workflow/executor/os-specific/signal_windows.go
+++ b/workflow/executor/os-specific/signal_windows.go
@@ -1,6 +1,7 @@
 package os_specific
 
 import (
+	"fmt"
 	"os"
 	"syscall"
 )
@@ -25,6 +26,16 @@ func Setpgid(a *syscall.SysProcAttr) {
 }
 
 func Wait(process *os.Process) error {
-	_, err := process.Wait()
+	stat, err := process.Wait()
+	if stat.ExitCode() != 0 {
+		var errStr string
+		if err != nil {
+			errStr = err.Error()
+		} else {
+			errStr = "<nil>"
+		}
+
+		return fmt.Errorf("exit with non-zero code. exit-code: %d, error:%s", stat.ExitCode(), errStr)
+	}
 	return err
 }


tczhao:
Because we squash merge, the title of the PR will become the commit message.
Perhaps changing the title to
`fix: exit code always be '0' in windows container. Fixes #9797`
to align with the PR template guide

tczhao (changes requested):
Please follow the PR template

- PR title
Because we squash merge, the title of the PR will become the commit message.
Perhaps changing the title to
`fix: exit code always be '0' in windows container. Fixes #9797`
to align with the PR template guide

- PR description
In the PR description please use
`Fixes #9797` instead of `associated with #9797`
The "Fixes" is a special keyword in github, that will link the PR to issue
image

alexec (approved):

Issue #9769, opened 2022-10-07
Bring CLOMonitor Score to 100%


eddie-knight:
Per discussion in slack, this issue is to track the efforts necessary to bring argo-workflow [CLOMonitor](https://clomonitor.io/projects/cncf/argo#argo-workflows) score to 100% as part of the CNCF security slam.

Below is a checklist of action items. I will add comments if there is anything that can't be addressed or if I am unsure of how to address it. It is possible to [exclude checks](https://github.com/cncf/clomonitor/blob/0942e124e116c2d46b3eeb8204579722ba1ff329/docs/metadata/.clomonitor.yml) with a written justification if that becomes necssary.
### Summary

**Repository**: argo-workflows
**URL**: https://github.com/argoproj/argo-workflows
**Checks sets**:  `CODE`
**Score**: 88

### Checks passed per category

| Category  |  Score  |
| :----------------- | --------: |
|  Documentation  |  100%  |
|  License  |  75%  |
|  Best Practices  |  100%  |
|  Security  |  80%  |
|  Legal  |  n/a  |

### Documentation [100%]

  - [x] [Changelog](https://github.com/argoproj/argo-workflows/blob/master/CHANGELOG.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#changelog))
  - [x] [Contributing](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#contributing))
  - [x] [Maintainers](https://github.com/argoproj/argo-workflows/blob/master/OWNERS) ([_docs_](https://clomonitor.io/docs/topics/checks/#maintainers))
  - [x] [Readme](https://github.com/argoproj/argo-workflows/blob/master/README.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#readme))
  
### License [75%]

  - [x] Apache-2.0 ([_docs_](https://clomonitor.io/docs/topics/checks/#license_spdx_id))
  - [x] Approved license ([_docs_](https://clomonitor.io/docs/topics/checks/#license_approved))
  - [ ] License scanning ([_docs_](https://clomonitor.io/docs/topics/checks/#license_scanning))
  
### Best Practices [100%]

  - [x] [Artifact Hub badge](https://artifacthub.io/packages/helm/argo/argo-workflows) ([_docs_](https://clomonitor.io/docs/topics/checks/#artifacthub_badge))
[...]

PR #9837, opened 2022-10-16
chore: sign container images on quay and signed releases

34fathombelow:
Signed releases #9769 
This PR implements container images to be signed using ```sigstore/cosign```  It also consolidates the checksums of the cli-binaries into a single file then signs it.   

#### Three GitHub secrets will need to be created before this PR is merged. The process to do this is listed below.

TLDR: https://docs.sigstore.dev/cosign/git_support

1. [Install Cosign](https://docs.sigstore.dev/cosign/installation) on your workstation. Linux, Homebrew, and container options are available.  Execute the command ```cosign --version``` to verified it has been installed correctly.
2. Create or have a valid GitHub PAT token available.
3. Export your PAT as the environment variable "GITHUB_TOKEN"
4. Execute ```cosign generate-key-pair github://argoproj/argo-workflows``` This will start the process of creating the GitHub Secrets automatically for you, and prompt you to enter a password.  This Password will be stored as the Github secret ```COSIGN_PASSWORD``` I would recommend to use well respected password generator such as KeePassXC or Bitwarden and use a paranoid level of characters of at least 32. 
5. Three GitHub secretes should have been created, please verify.  ```COSIGN_PASSWORD``` ```COSIGN_PRIVATE_KEY``` & ```COSIGN_PUBLIC_KEY```
6. Please comment on this PR with the newly created public key (cosign.pub)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -429,7 +429,7 @@ jobs:
         run: |
           cosign sign-blob --key env://COSIGN_PRIVATE_KEY ./dist/argocd-workflows-checksums.txt > ./dist/argocd-workflows-checksums.sig
           # Retrieves the public key to release as an asset
-          cosign public-key --key env://COSIGN_PRIVATE_KEY > ./dist/argo-rollouts-cosign.pub
+          cosign public-key --key env://COSIGN_PRIVATE_KEY > ./dist/argo-workflows-cosign.pub
 
       # https://github.com/softprops/action-gh-release
       # This will publish the release and upload assets.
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -388,8 +443,9 @@ jobs:
           body_path: release-notes
           files: |
             dist/argo-*.gz
```
alexec, line 445: Don’t we want to keep this file?

34fathombelow:
Updated to only sign multi-image tags

alexec (approved):
LGTM. Github Actions will need to be monitored when this is merged and settings updated.

34fathombelow:
Not stale

terrytangyuan (approved):
Thank you!

terrytangyuan:
We need some additional secrets required by this PR. @alexec @sarabala1979 Could you help add those?

PR #9846, opened 2022-10-17
fix(9656): stores all states except workflows, fixes #9656

### LINKED ISSUE #9656: Archived Workflows UI error: `Failed to execute 'setItem' on 'Storage'`
### Pre-requisites

- [ ] I can confirm the issues exists when I tested with `:latest`

When opening the tab "Archived Workflows" in the web UI I got the following error:

`Failed to execute 'setItem' on 'Storage': Setting the value of 'ArchiveListOptions/options' exceeded the quota.`

Screenshot 2022-09-21 at 16 59 32

In particular, I am using a Postgres storage backend to archive the workflows and I already tried to increase resources for both workflow-controller and argo-server.

Any idea?

### Version

v3.3.5

```YAML
Cannot view any workflow in the UI.
```

athityakumar:
Fixes #9656 

The `workflows` key/value is no longer stored in browser's local storage for archived workflows page, after this change:

image


ltartarini (approved):

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
```diff
@@ -181,8 +181,54 @@ export class ArchivedWorkflowList extends BasePage<RouteComponentProps<any>, Sta
         return params;
     }
 
+    private filterStateObject(state: State, whitelistedStates: string[], blacklistedStates: string[]) {
+        const filteredState: State = {} as State;
```
juliev0, line 185: Just curious: why do we need both a blacklist and a whitelist? Is there a use case for that?

It seems if you pass `{whitelistedStates:['pagination'], blacklistedStates:[]}`, that's equivalent to  `{whitelistedStates:[], blacklistedStates:[]}`, so I'm wondering if `whitelistedStates` is really needed? 

I see that if you pass `{whitelistedStates:['pagination'], blacklistedStates:[`pagination`]}` then that would not be equivalent to `{whitelistedStates:[], blacklistedStates:[`pagination`]}` but I'm not sure if anyone would ever need to call it that way ?

Do let me know if I'm missing something, though. Is there some industry standard that uses both whitelists and blacklists together like this?

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
```diff
@@ -181,8 +181,54 @@ export class ArchivedWorkflowList extends BasePage<RouteComponentProps<any>, Sta
...
+            filteredState.deep = state.deep;
+        }
+
+        return filteredState;
+    }
+
     private saveHistory() {
-        this.storage.setItem('options', this.state, {} as State);
+        this.storage.setItem('options', this.filterStateObject(this.state, [], ['workflows']), {} as State);
         const newNamespace = Utils.managedNamespace ? '' : this.state.namespace;
```
juliev0, line 232: I do like the way you abstracted this into a function

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
```diff
@@ -181,8 +181,54 @@ export class ArchivedWorkflowList extends BasePage<RouteComponentProps<any>, Sta
...
+        }
+
+        if (whitelistedStates.includes('error') || !blacklistedStates.includes('error')) {
+            filteredState.error = state.error;
+        }
+
+        if (whitelistedStates.includes('deep') || !blacklistedStates.includes('deep')) {
+            filteredState.deep = state.deep;
+        }
+
```
juliev0, line 226: I'm a little confused. Shouldn't `workflows` be in here? Otherwise, passing it in here seems to have no effect

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
juliev0: Although, I'm concerned about the way the function lists out each member individually. There seems to be some maintenance risk there if somebody adds a new field to the struct, they need to update this function, right? Is there any way to do this in a way that doesn't require the maintenance?

juliev0: Maybe we should consider simply clearing the value for `this.state.workflows` and perhaps this is a little over-designed? (I'm open, though)

athityakumar: Agree regarding the maintenance risk if the state expands - I've added looping to blacklist the keys instead of statically going through each key via multiple if conditions

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
athityakumar: Makes sense, I've removed whitelistedStates as we don't currently need it rn - I was initially thinking of a more generic function that filters the states (with either whitelist to select few states, or blacklist to reject few states)

sarabala1979 (approved):

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
```diff
@@ -181,8 +181,25 @@ export class ArchivedWorkflowList extends BasePage<RouteComponentProps<any>, Sta
...
+        return filteredState;
+    }
+
+    private fetchBrowserStorageStateObject(state: State) {
+        return this.filterOutStateObject(state, ['workflows']);
+    }
+
     private saveHistory() {
-        this.storage.setItem('options', this.state, {} as State);
+        this.storage.setItem('options', this.fetchBrowserStorageStateObject(this.state), {} as State);
```
alexec, line 202: I don't think this is the correct approach. A new objects should be constructed with only allowed fields.

alexec:
```typescript
interface Options {
        namespace: string
        name: string
        namePrefix: string
        selectedPhases: string[]
        selectedLabels: string[]
        minStartedAt: Date
        maxStartedAt: Date
        pagination: Pagination
}

alexec (changes requested):

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
athityakumar: @alexec - But that'd mean we have to set the attributes manually for the new `options` variable, which is an `Options` interface object right? @juliev0 had raised a concern earlier that manually specifying attributes may have a maintenance overhead (if/when we add a new attribute to `State` interface, we'd also have to add it to the `Options` interface and in the function that converts the `State` object to `Options` object) - due to which I tried to automate the logic by taking `blacklistedStates` as an input.

Please lmk if `Options` is the preferred approach - I can update my PR with the same changes 😄

athityakumar: Ping @alexec @sarabala1979 ^

athityakumar:
@sarabala1979 @alexec - Any next steps on this PR? Please lmk if any action is required from my side 😄

ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx
alexec: I think it preferable to construct new object to write lots of code to black list fields. Less code to maintain I think.

alexec (approved):

PR #9859, opened 2022-10-19
feat: Hide secrets in logs. Fixes #8685

### LINKED ISSUE #8685: Hide secrets in logs
# Summary

Currently, the secrets are printed in the logs without any filtering. It would be nice if the secrets could be redacted in the logs while being printed

# Use Cases

This would prevent unintentional leaks of a secret in the log by running something like `env` in the workflow

NOTE: This does not prevent intentionally leaks of the secrets. Someone can still run something like `env | base64` to extract the secrets from the logs, as the filtering will not be able to filter out the secret output.

---

anilkumar-pcs:
Fixes #8685 

FIlter and mask secrets in workflow logs.

Logs when running `examples/secrets.yaml` workflow with no redaction

```
secrets-tdwm7: time="2022-10-19T12:47:54.251Z" level=info msg="capturing logs" argo=true
secrets-tdwm7: secret from env: S00perS3cretPa55word
secrets-tdwm7: secret from file: S00perS3cretPa55word
secrets-tdwm7: time="2022-10-19T12:47:55.263Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

Logs when running `examples/secrets.yaml` workflow with redaction enabled

```
secrets-tdwm7: time="2022-10-19T12:47:54.251Z" level=info msg="capturing logs" [*********]=true
secrets-tdwm7: secret from env: [*********]
secrets-tdwm7: secret from file: [*********]
secrets-tdwm7: time="2022-10-19T12:47:55.263Z" level=info msg="sub-process exited" [*********]=true error="<nil>"
```


alexec:
This is cool. I wonder if it can be even smarter? E.g. auto-detect secrets using a regex?

anil10032:
@alexec while the idea to auto-detect the secrets is nice, secrets are something user configurable and they can be literally anything. I believe it would be difficult to have a regex to identify all possible secrets.

Do you have any references in mind? Please share your thoughts.

util/logs/workflow-logger.go
```diff
@@ -165,6 +174,17 @@ func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient
 						}
 						if rx.MatchString(content) { // this means we filter the lines in the server, but will still incur the cost of retrieving them from Kubernetes
 							logCtx.WithFields(log.Fields{"timestamp": timestamp, "content": content}).Debug("Log line")
+
+							// log redaction for secrets
+							if secrets != nil && enablePodLogRedaction == "true" {
+								for _, s := range secrets.Items {
+									for _, v := range s.Data {
+										if strings.Contains(content, string(v)) {
+											content = strings.Replace(content, string(v), "[ redacted ]", -1)
```
tczhao, line 183: I wonder if using *** is better than redacted

tczhao (approved):
I tested how Github Action hides the secret
from the result, I think this is exaclty how it's done.

string replace rather than any regex.
(e.g. if I echo "text same as secret", it will hide it, if I remove 1 char from it, it will show)

docs/environment-variables.md
```diff
@@ -22,6 +22,7 @@ most users. Environment variables may be removed at any time.
...
 | `ARGO_AGENT_MEMORY_LIMIT`              | `resource.Quantity` | `256m`                                                                                      | Memory resource limit for the agent.                                                                                                                                                                                                                                     |
+| `ARGO_REDACT_POD_LOGS`                 | `bool`              | `false`                                                                                      | Whether to redact pod logs to hide/mask secrets.                                                                                                                                                                                                                        |
```
terrytangyuan, line 25: I think this should be turned on by default.

util/logs/workflow-logger.go
```diff
@@ -80,6 +81,14 @@ func WorkflowLogs(ctx context.Context, wfClient versioned.Interface, kubeClient
 
 	var podListOptions metav1.ListOptions
 
+	// get env variable for pod logs redaction
+	enablePodLogRedaction := os.Getenv("ARGO_REDACT_POD_LOGS")
+	// get secrets for redaction
+	secrets, err := kubeClient.CoreV1().Secrets(req.GetNamespace()).List(ctx, metav1.ListOptions{})
```
terrytangyuan, line 87: I don't think we can list all the secrets. See https://github.com/argoproj/argo-workflows/issues/8534

docs/environment-variables.md
terrytangyuan: Although I am not sure if we should turn on new features by default. WDYT? @sarabala1979 @alexec

util/logs/workflow-logger.go
anil10032: @terrytangyuan I can see the [GetSecret](https://github.com/argoproj/argo-workflows/blob/master/server/cache/resource_cache.go#L42) method is available but it requires the service account name and we can only fetch the secrets associated with a service account using this method.

One approach is using [Service account Lister](https://github.com/argoproj/argo-workflows/blob/master/server/cache/resource_cache.go#L27) to get all the service accounts and fetch the secrets of each SA using above mentioned GetSecret method. Although we might not have all the secrets with this approach.

Can you suggest any other alternate approaches?

alexec (commented):
I don't think this is the right approach. It moves responsibly for secret redaction from the user to the operator, yet it is the user's fault secrets are being logged. 

I'd be amenable to a redaction that was zero-configuration. Github Actions automatically redacts secrets for example.

anil10032:
@alexec I have introduced configuration as it was suggested in [here](https://github.com/argoproj/argo-workflows/issues/8685#issuecomment-1121328515)

We can remove the env variable configuration and mask the secrets by default. WDYT?

anilkumar-pcs:
@alexec made changes to remove the env configuration and masking secrets by default.

util/logs/workflow-logger.go
da-head0: @terrytangyuan @anilkumar-pcs 
I'd like to resolve #8685 but I don't have any better idea then 
1. list secrets and keep then in cache or queue
2. when command like "echo" or "export", check command if they include one of those secrets
3. hide secrets

but I think those feature make large burden on argo-workflows.... 
any suggestion on this??

tooptoop4:
where @anilkumar-pcs

agilgur5:
> where @anilkumar-pcs

Per [above](https://github.com/argoproj/argo-workflows/pull/9859#pullrequestreview-1170671482), this approach has security ramifications. See https://github.com/argoproj/argo-workflows/issues/8685#issuecomment-1401477067 for an alternative without those ramifications

PR #9876, opened 2022-10-21
feat: Added tar archives can be used as input. Fixes #9818

### LINKED ISSUE #9818: Artifact with compressionLevel:0 failed on v3.4.1
### Pre-requisites

With the setting `artifact.tar.compressionLevel: 0` and an artifact `test.tar`. I have this error `Error (exit code 1): gzip: invalid header`. 

It seems like argo-workflows is expecting a gzip file even the compressionLevel is 0.

The same code works fine with v3.3.9

### Version

v3.4.1

```YAML
kind: Workflow
metadata:
  generateName: test-
spec:
  entrypoint: echo
  arguments:
    artifacts:
      - name: test
        archive:
          tar:
            compressionLevel: 0
        s3:
          bucket: test-bucket
          key: test.tar
          useSDKCreds: true
          endpoint: s3.amazonaws.com
          region: ca-central-1
  templates:
  - name: echo
    inputs:
      parameters:
      - name: message
        default: Bye
      artifacts:
        - name: test
          path: /tmp/test
    container:
      image: alpine:3.7
      command: [sh, '-c']
      args:
        - |
          echo "{{inputs.parameters.message}}-modified"
```

$ kubectl logs -n argo-workflows deploy/argo-workflows-workflow-controller |  grep test-pdrd5
Found 2 pods, using pod/argo-workflows-workflow-controller-74d976d56c-bfdt2

Error from server (BadRequest): container "wait" in pod "test-pdrd5" is terminated

dhaven:
Signed-off-by: dhaven <dev_ddh@pm.me>

Fixes #9818 

This features allows to use uncompressed tar archives as input. Instead of always assuming that a tar archive is gzipped we check first if it is just a tar and apply the appropriate unpacking thereafter. An alternative implementation would be to extend the workflow template to allow a tar (non gzip) archive strategy which would allow to support both output and input. Happy to receive some feedback at this point!

### DIFF
--- a/workflow/executor/executor.go
+++ b/workflow/executor/executor.go
@@ -224,8 +224,8 @@ func (we *WorkflowExecutor) LoadArtifacts(ctx context.Context) error {
 				isGzip, err = isgzip(tempArtPath)
 				if err != nil {
 					return err
-                }
-            }
+				}
+			}
 		} else if art.GetArchive().Zip != nil {
 			// explicitly a zip
 			isZip = true
--- a/workflow/executor/executor.go
+++ b/workflow/executor/executor.go
@@ -211,28 +211,41 @@ func (we *WorkflowExecutor) LoadArtifacts(ctx context.Context) error {
 		}
 
 		isTar := false
+		isGzip := false
 		isZip := false
 		if art.GetArchive().None != nil {
 			// explicitly not a tar
 			isTar = false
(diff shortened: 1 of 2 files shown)


terrytangyuan (commented):
The direction looks good to me. Please ping me when ready to review.

PR #9912, opened 2022-10-26
fix(ui): Apply url encode and decode to a `ProcessURL`. Fixes #9791

### LINKED ISSUE #9791: Link templates do not work when template is in URL path
### Pre-requisites

I have workflow controller links configuration
```
links: |
   - name: Spark UI
      scope: workflow
      url: https://example.com/workflow/${metadata.name}/jobs/
```
when I press `Spark UI` button it redirects me to `https://example.com/workflow/$%7Bmetadata.name%7D/jobs/`.
Info API returns already encoded links. They become encoded when URL object reassembles string https://github.com/argoproj/argo-workflows/blob/master/config/config.go#L155

### Version

latest

```YAML
Any workflow can be used here
```

No errors in logs expected

No errors in logs expected

AndrewChubatiuk:
Fixes #9791


alexec (approved):
LGTM

terrytangyuan (approved):

PR #9915, opened 2022-10-27
docs: explain why maxCharsInObjectName is 63 characters

Mzhiing:
When I encountered this problem, I spent some time to find the reason, so I want people to know the reason faster.


alexec (approved):
LGTM

PR #9916, opened 2022-10-27
fix: Return correct http error codes. Fixes #9237

### LINKED ISSUE #9237: Workflow API response failure due to RBAC restrictions
## Summary

If we are not authorized for a particular operation according to RBAC, the Workflow API will return an HTTP status 500 error.
Not authorized, we expect the API to return a HTTP status 4xx error such as 403.

Server administrators often monitor the API for HTTP status errors at 5xx.
However, since an unauthorized request is normal for the server, it is a false positive and turns into "The Boy Who Cried Wolf".

version: **v3.3.8**

## Diagnostics

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: ui-readonly
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - pods/log
  - events
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - argoproj.io
  resources:
  - workflows
  - workfloweventbindings
  - workflowtemplates
  - cronworkflows
  - cronworkflows/finalizers
  - clusterworkflowtemplates
  verbs:
  - get
  - list
  - watch
```

For example, if the above privileges are assigned to a service account, the user using that service account will not be able to **retry** the workflow.

When the user performs a **retry** operation from UI, the following API was called:

`PUT /api/v1/workflows/namespace/workflow-name/retry`

The response to this would be a HTTP status of 500.

In this case, the error log for argocli is as follows.

```
time="2022-07-27T00:04:43.918Z" level=error msg="finished unary call with code Unknown" error="pods \"sample-workflow-pod-name\" is forbidden: User \"system:serviceaccount:argo:ui-readonly\" cannot delete resource \"pods\" in API group \"\" in the namespace \"sample-namespace\"" grpc.code=Unknown grpc.method=RetryWorkflow grpc.service=workflow.WorkflowService grpc.start_time="2022-07-27T00:04:43Z" grpc.time_ms=49.319 span.kind=server system=grpc
```

---

isubasinghe:

Fixes #9237

### DIFF
--- a/server/utils/errors.go
+++ b/server/utils/errors.go
@@ -12,6 +12,21 @@ import (
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 )
 
+var errorMaps = map[int]codes.Code{
+	http.StatusOK:                  codes.OK,
+	http.StatusRequestTimeout:      codes.Canceled,
+	http.StatusGatewayTimeout:      codes.DeadlineExceeded,
+	http.StatusNotFound:            codes.NotFound,
+	http.StatusConflict:            codes.AlreadyExists,
+	http.StatusForbidden:           codes.PermissionDenied,
+	http.StatusUnauthorized:        codes.Unauthenticated,
+	http.StatusTooManyRequests:     codes.ResourceExhausted,
+	http.StatusBadRequest:          codes.InvalidArgument,
+	http.StatusNotImplemented:      codes.Unimplemented,
+	http.StatusInternalServerError: codes.Internal,
+	http.StatusServiceUnavailable:  codes.Unavailable,
+}
+
 // Return a new status error and true else nil and false
 // is meant to be as close to the opposite of grpc gateways status -> http error code converter
 // exceptions made where one to one mappings were not available
@@ -19,21 +34,6 @@ func httpToStatusError(code int, msg string) (error, bool) {
 	// handle success & information  codes in one go
 	if code < 300 {
 		return status.Error(codes.OK, msg), true
-	}
-
-	errorMaps := map[int]codes.Code{
-		http.StatusOK:                  codes.OK,
-		http.StatusRequestTimeout:      codes.Canceled,
-		http.StatusGatewayTimeout:      codes.DeadlineExceeded,
-		http.StatusNotFound:            codes.NotFound,
-		http.StatusConflict:            codes.AlreadyExists,
-		http.StatusForbidden:           codes.PermissionDenied,
-		http.StatusUnauthorized:        codes.Unauthenticated,
-		http.StatusTooManyRequests:     codes.ResourceExhausted,
-		http.StatusBadRequest:          codes.InvalidArgument,
-		http.StatusNotImplemented:      codes.Unimplemented,
-		http.StatusInternalServerError: codes.Internal,
-		http.StatusServiceUnavailable:  codes.Unavailable,
 	}
 
 	statusCode, ok := errorMaps[code]
--- a/server/utils/errors.go
+++ b/server/utils/errors.go
@@ -12,62 +12,35 @@ import (
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 )
 
-func ArgoToStatusError(err argoerrors.ArgoError) error {
-	switch err.Code() {
-	case argoerrors.CodeUnauthorized:
-		return status.Error(codes.Unauthenticated, err.Error())
-	case argoerrors.CodeForbidden:
-		return status.Error(codes.PermissionDenied, err.Error())
-	case argoerrors.CodeNotFound:
-		return status.Error(codes.NotFound, err.Error())
-	case argoerrors.CodeBadRequest:
-		return status.Error(codes.InvalidArgument, err.Error())
-	case argoerrors.CodeNotImplemented:
-		return status.Error(codes.Unimplemented, err.Error())
-	case argoerrors.CodeTimeout:
-		return status.Error(codes.DeadlineExceeded, err.Error())
-	case argoerrors.CodeInternal:
-		return status.Error(codes.Internal, err.Error())
-	default:
-		return status.Error(codes.Unknown, err.Error())
-	}
-}
-
 // Return a new status error and true else nil and false
 // is meant to be as close to the opposite of grpc gateways status -> http error code converter
 // exceptions made where one to one mappings were not available
-func HTTPToStatusError(code int, msg string) (error, bool) {
+func httpToStatusError(code int, msg string) (error, bool) {
 	// handle success & information  codes in one go
 	if code < 300 {
 		return status.Error(codes.OK, msg), true
 	}
 
-	switch code {
-	case http.StatusOK:
-		return nil, true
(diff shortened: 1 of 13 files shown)


server/apiserver/argoserver.go
```diff
@@ -314,6 +316,15 @@ func (as *argoServer) newGRPCServer(instanceIDService instanceid.Service, offloa
 	return grpcServer
 }
 
+func customHandler(ctx context.Context, mux *runtime.ServeMux, marshaller runtime.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
+	_, ok := status.FromError(err)
+	if !ok {
+		// InvalidArgument is translated to a BadRequest
```
sarabala1979, line 322: @isubasinghe this will convert all errors into badrequest. we need to send this error if any spec validation error.

isubasinghe: @sarabala1979 Do you know anyway to determine if an error is a spec validation error? I wasn't able to find how, hence I felt forced to go with this approach. The limitation is from the fact that some errors are returned by fmt.Errorf, which makes it impossible to check against since they are just strings. 

I think personally the fix is to create a new `ServerError` struct and have it implement error. This way I think we can use reflection to check the error. This would be a pretty big change (I'd argue a breaking change) and require me to change every server type (info, workflow ...)

isubasinghe: Do you think the `ServerError` path is the way to go ?

alexec (commented):
This does not look like the correct approach to me. There is an error translator somewhere you could take a look at.

sarabala1979:
@alexec Can you provide more information to @isubasinghe, So, he can refactor?

isubasinghe:
@sarabala1979 this is a refactor from the old one which Alex commented on. Every error goes through ToStatusError. 
Here I try to extract an ArgoError and map it into a grpc error, failing that I try to extract a k8s error and map that into a grpc error. 

In a general sense, the most relevant file here is just: https://github.com/argoproj/argo-workflows/pull/9916/files#diff-6ec3cdfd853ed9734ca0ba5e09bfc02c4a2ff21478a1d264ce3e4277e40f4c69

The others are just calling this.

caelan-io:
What are your thoughts @alexec ? Can we merge to close #9237?

server/utils/errors.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	argoerrors "github.com/argoproj/argo-workflows/v3/errors"
+
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+)
+
+func ArgoToStatusError(err argoerrors.ArgoError) error {
```
JPZ13, line 15: nit: Can you add a comment here given it's an exported function?

JPZ13 (approved):
Code looks great. Running well when tested locally. One quick nit, can you change the PR title to be something like `fix: Return correct http error codes. Fixes #9237` or something similar?

alexec (changes requested):
This looks great. I think maybe a couple more tests?

server/utils/errors.go
```diff
@@ -0,0 +1,123 @@
...
+
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// REFER TO https://github.com/grpc-ecosystem/grpc-gateway/blob/8952e38d5addd28308e29c272c696a578aa8ace8/runtime/errors.go#L36 in order to understand
+// how the mapping behaviours for both HTTPToStatusError and ArgoToStatusError were chosen.
+
+// Converts an ArgoError into a grpc status error.
+// It defaults to a codes.Unknown if an appropriate error was not able to be found
+func ArgoToStatusError(err argoerrors.ArgoError) error {
```
alexec, line 20: Test?

server/utils/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	case argoerrors.CodeInternal:
+		return status.Error(codes.Internal, err.Error())
+	default:
+		return status.Error(codes.Unknown, err.Error())
+	}
+}
+
+// Return a new status error and true else nil and false
+// is meant to be as close to the opposite of grpc gateways status -> http error code converter
+// exceptions made where one to one mappings were not available
```
alexec, line 43: Test? Can be private?

server/utils/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil, false
+}
+
+// Try to see if we can obtain a http
+// error code from the k8s layer or the ArgoError layer
+// if not we resort to a default value of `code`
+// NOTE: errors passed of the type from grpc's status are not converted
+// and returned as is. This is to keep user code as dumb as possible.
+// The assumption here is that the error in the lowest layer of the error stack is the most relevant error.
+func ToStatusError(err error, code codes.Code) error {
```
alexec, line 99: More tests?

server/utils/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	if err == nil {
+		return nil
+	}
+	argoerr, typeOkay := err.(argoerrors.ArgoError)
```
alexec, line 103: Could use `errors.As()`?

server/utils/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	if typeOkay {
+		return ArgoToStatusError(argoerr)
+	}
+	// allow callers to call ToStatusError on the same processed error
+	_, alreadyConverted := status.FromError(err)
```
alexec, line 108: Could probably move this up?

server/utils/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// REFER TO https://github.com/grpc-ecosystem/grpc-gateway/blob/8952e38d5addd28308e29c272c696a578aa8ace8/runtime/errors.go#L36 in order to understand
+// how the mapping behaviours for both HTTPToStatusError and ArgoToStatusError were chosen.
+
+// Converts an ArgoError into a grpc status error.
+// It defaults to a codes.Unknown if an appropriate error was not able to be found
+func ArgoToStatusError(err argoerrors.ArgoError) error {
+	switch err.Code() {
```
alexec, line 21: Minor, but you could use a `map` to do a look up here.

server/utils/errors.go
```diff
@@ -0,0 +1,94 @@
...
+// Return a new status error and true else nil and false
+// is meant to be as close to the opposite of grpc gateways status -> http error code converter
+// exceptions made where one to one mappings were not available
+func httpToStatusError(code int, msg string) (error, bool) {
+	// handle success & information  codes in one go
+	if code < 300 {
+		return status.Error(codes.OK, msg), true
+	}
+
+	errorMaps := map[int]codes.Code{
```
sarabala1979, line 24: @isubasinghe Can you move to constants instead of initializing every call?

isubasinghe: done

sarabala1979 (approved):

PR #9944, opened 2022-11-02
docs: add 3.4 links encode guide. Fixes #9935

### LINKED ISSUE #9935: HTML encoding still breaking some log links
### Pre-requisites

Reference #9435 where this was originally reported.

Given I have a custom link in my configmap:

`http://grafana.home.local/explore?orgId=1&left=%7B%22datasource%22:%22Loki%22,%22queries%22:%5B%7B%22expr%22:%22%7Bcontainer%3D%5C%22main%5C%22,pod%3D%5C%22${metadata.name}%5C%22%7D%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-3h%22,%22to%22:%22now%22%7D%7D`

${metadata.name} is correctly resolved to a pod BUT %3D is not resolved to a = so loki gets this query:

`{container%3D"main",pod%3D"backfill-worker-go62jrw-962151937"}`

instead of this:

`{container="main",pod="backfill-worker-go62jrw-962151937"}`

It all worked well in 3.3.x

### Version

3.4.3

```YAML
Any workflow.
```

n/a

n/a

tczhao:
Fixes #9935

### DIFF
--- a/docs/links.md
+++ b/docs/links.md
@@ -35,3 +35,7 @@ For example, one may find it useful to define a custom label in the workflow and
 
 We can also access workflow fields in a pod link. For example, `${workflow.metadata.name}` returns
 the name of the workflow instead of the name of the pod.
+
+> v3.4 and after
+
+The links now undergo a encode -> decode -> encode process, make sure the links you provide do not include the encoded version of the reseved character `$&+,/:;=?@` to avoid bugs.


tczhao:
From https://www.ietf.org/rfc/rfc2396.txt

2.4.2. When to Escape and Unescape

   A URI is always in an "escaped" form, since escaping or unescaping a
   completed URI might change its semantics.  **Normally, the only time
   escape encodings can safely be made is when the URI is being created
   from its component parts**; each component may have its own set of
   characters that are reserved, so only the mechanism responsible for
   generating or interpreting that component can determine whether or

   not escaping a character will change its semantics. Likewise, a URI
   must be separated into its components before the escaped characters
   within those components can be safely decoded.

   In some cases, data that could be represented by an unreserved
   character may appear escaped; for example, some of the unreserved
   "mark" characters are automatically escaped by some systems.  If the
   given URI scheme defines a canonicalization algorithm, then
   unreserved characters may be unescaped according to that algorithm.
   For example, "%7e" is sometimes used instead of "~" in an http URL
   path, but the two are equivalent for an http URL.

   Because the percent "%" character always has the reserved purpose of
   being the escape indicator, it must be escaped as "%25" in order to
   be used as data within a URI.  Implementers should be careful not to
   escape or unescape the same string more than once, since unescaping
   an already unescaped string might lead to misinterpreting a percent
   data character as another escaped character, or vice versa in the
   case of escaping an already escaped string.

---

Given the above, maybe we shouldn't perform any encoding/decoding for the user-provided URL.
We only keep the validation (if valid url, if not js)

sarabala1979 (approved):

tico24:
Poking stalebot. Could really do with having this merged in.

PR #9992, opened 2022-11-07
feat: Implement rebalance queue and config option (Fixes #8982)

### LINKED ISSUE #8982: Resource sharing amongst users at the template level

robertkotcher:
Signed-off-by: Robert Kotcher <rkotcher+1@gmail.com>

Fixes #8982


config/config.go
```diff
@@ -99,6 +99,10 @@ type Config struct {
 
 	RetentionPolicy *RetentionPolicy `json:"retentionPolicy,omitempty"`
 
+	// SemaphoreStrategies maps a lock name to the type of strategy that should be used to assign locks to
+	// pending lock holders.
+	SemaphoreStrategies SemaphoreStrategyTypes `json:"semaphoreStrategies,omitempty"`
```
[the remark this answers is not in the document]
sarabala1979: Instead of having it controller config, Can we have it in the semaphore configmap itself.
The current approach will not support multi-tenant.

sarabala1979 (changes requested):

config/config.go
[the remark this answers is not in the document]
robertkotcher: Sorry, pinged you on slack but figured it might be best to comment here for better visibility

workflow/controller/controller.go
```diff
@@ -341,6 +341,30 @@ func (wfc *WorkflowController) createSynchronizationManager(ctx context.Context)
...
+		lockName, err := sync.DecodeLockName(lockKey)
+		if err != nil {
+			return config.SemaphoreStrategyDefault, err
+		}
+		configMap, err := wfc.kubeclientset.CoreV1().ConfigMaps(lockName.Namespace).Get(ctx, lockName.ResourceName, metav1.GetOptions{})
+		if err != nil {
+			return config.SemaphoreStrategyDefault, err
+		}
+
+		value, found := configMap.Data[fmt.Sprintf("%s-strategy", lockName.Key)]
```
isubasinghe, line 354: Why do you need this "%s-strategy" string formatting? Where does this get populated?

In general I do think it would be nice to move out key generation into functions, it just helps to trace usage better when you can just use the LSP to find references.

workflow/sync/rebalanced.go
```diff
@@ -0,0 +1,144 @@
...
+	overflow := make(map[string]*item, 0)
+	cant := make([]*item, 0)
+
+	maxLocksPerUserFloor := math.Floor(float64(r.semaphore.getLimit()) / float64(len(allRebalanceKeys)))
+	// iterate through pending queue to determine what has changed
+	for _, p := range r.items {
+		if p.key != key {
+			rebalanceKey := r.rebalanceKeyCache[p.key]
+			// we want to ignore "key" since removal from queue happens after release in semaphore
+			if float64(holderRebalanceKeys[rebalanceKey]) < maxLocksPerUserFloor {
```
isubasinghe, line 61: Minor thing but why convert this to a float? Is it not better to cast maxLocksPerUserFloor into an int instead? 
I know we will probably never hit numbers large enough where comparison being imprecise is a worry, but probs nice to use ints regardless.

workflow/controller/controller.go
robertkotcher: The user configures this in config map.

I'm not 100% sure I follow the suggestion. How would the user configure semaphore strategy with a function?

Thx!

workflow/sync/rebalanced.go
robertkotcher: @isubasinghe it's because `maxLocksPerUserFloor` is a float, since Floor returns float64. See line 55.

workflow/controller/controller.go
isubasinghe: Thanks for your clarification, I meant refactor the string formatting out to a function, but if this is the only place it is going to be used, it is probably fine to remain as it is. A comment about it would be great though. 

Thanks

workflow/sync/rebalanced.go
isubasinghe: Yes, I was more suggesting to go with approach on line 67 instead of casting to floats. It honestly is a very minor nit though, probs fine as it is, this is more me just being paranoid

workflow/controller/controller.go
robertkotcher: Oh, thanks! I understand! I will add a comment. I appreciate the review @isubasinghe

isubasinghe: Thanks for doing that! No worries! I've just approved this

isubasinghe (approved):
Awesome work! Thanks for your contribution.

config/config.go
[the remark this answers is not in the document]
robertkotcher: @sarabala1979 I have implemented this change - do you think I could get your approval on this PR? Or, are there any further changes you'd like me to make? Thanks!

caelan-io (approved):
Looks like all the requested changes have been made. Thanks, @robertkotcher ! Excited to build on top of this feature and gather more user feedback once it's merged

sarabala1979:
@terrytangyuan Can you review this PR?

terrytangyuan:
> @terrytangyuan Can you review this PR?

Will do

docs/proposals/rebalanced-semaphore.md
```diff
@@ -0,0 +1,118 @@
...
+The final config file should look something like:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+ name: my-config
+data:
+  template: "5"
+  template-strategy: "rebalanced"
```
terrytangyuan, line 50: Where is this used?

docs/proposals/rebalanced-semaphore.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+Note that the strategy should be added to the same config file that the limit is defined in.
+
+The workflow controller should be restarted if a change in strategy takes place. This is because the strategy is fixed when the controller in initialized.
```
terrytangyuan, line 55: This does not seem ideal. Why isn't this config dynamically picked up?

terrytangyuan:
Great work on this, but I wonder why https://github.com/argoproj/argo-workflows/issues/8982 is pretty quiet and not popular.

robertkotcher:
> Great work on this, but I wonder why #8982 is pretty quiet and not popular.

It's a fair point. @caelan-io has mentioned that this is something that some of Pipekit customers would benefit from, but I haven't heard any others come out to say it'd be useful. Maybe we can talk more about this in next maintainer meeting. Maybe there's a way to poll or figure out if this would be a useful addition?

PR #10059, opened 2022-11-18
Enable full offline lint of all resources

julienduchesne:
Currently, the offline linting only works for Workflow because other resources (`WorkflowTemplate`, `ClusterWorkflowTemplate` and `CronWorkflow`) can depend on references to other resources. However, this behavior is very limiting in a CI context where a user may modify both a "top-level" file and its dependencies at the same time. The dependencies are fetched from the server and the validation fails.

This PR extends the offline linter so that it uses the whole list of files passed as arguments in order to validate. This is useful in a GitOps context where a user can pass the whole list of Argo files to the CI check.

Signed-off-by: Julien Duchesne <julien.duchesne@grafana.com>


pkg/apiclient/apiclient.go
```diff
@@ -32,7 +32,7 @@ type Opts struct {
 	// DEPRECATED: use `ClientConfigSupplier`
 	ClientConfig         clientcmd.ClientConfig
 	ClientConfigSupplier func() clientcmd.ClientConfig
-	Offline              bool
```
alexec, line 35: this is a breaking change, can we keep the old variable please

alexec (changes requested):

pkg/apiclient/apiclient.go
julienduchesne: I put it back. It should still work as it did before for workflows. But now, having an empty `OfflineFiles` will fail when trying to lint a workflowtemplate that has a ref

cmd/argo/commands/lint.go
```diff
@@ -35,6 +35,9 @@ func NewLintCommand() *cobra.Command {
   cat manifests.yaml | argo lint --kinds=workflows,cronworkflows -`,
 		Run: func(cmd *cobra.Command, args []string) {
 			client.Offline = offline
+			if offline {
+				client.OfflineFiles = args
```
alexec, line 39: I like the idea that all the files form a "linting context", do you even `OfflineFiles`?

julienduchesne: What do you mean?

alexec: I think your contextual lint should be the default. No need to enable any flag. I think we can always do what line 39 does. This is always better and more correct.

julienduchesne: I removed the condition. It doesn't make sense indeed. The `OfflineFiles` are only used when the "offline client" is used

julienduchesne:
Hey @alexec. Any chance I could get another review of this? This would greatly improve performance, reliability and usefulness of the linter for us

juliev0:
> I approve of these changes. @juliev0 do you want to review? If not, we can merge.

My feeling is that in the interest of moving more PRs along, if you find a PR you approve of, go ahead and merge it? (but if it's in an area of the code I've worked in, I'm definitely happy to review it first)

PR #10120, opened 2022-11-27
feat: allow switching timezone for date rendering. Fixes #3474

### LINKED ISSUE #3474: Correct timezone in logs
The logs provided by the UI are timestamped. The dates and times provided are always the European ones.

I tried to set `TZ=America/Toronto` to both the `workflow-controller` and `argo-server` deployment without success.

Is that possible to configure the timezone?

isubasinghe:

Fixes #3474

* [] Run `make pre-commit -B` to fix codegen and lint problems.

### DIFF
--- a/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
+++ b/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {useEffect, useState} from 'react';
 
 import {Autocomplete} from 'argo-ui';
+import moment = require('moment-timezone');
 import {Observable} from 'rxjs';
 import {map, publishReplay, refCount} from 'rxjs/operators';
 import * as models from '../../../../models';
@@ -10,9 +11,14 @@ import {ANNOTATION_KEY_POD_NAME_VERSION} from '../../../shared/annotations';
 import {ErrorNotice} from '../../../shared/components/error-notice';
 import {InfoIcon, WarningIcon} from '../../../shared/components/fa-icons';
 import {Links} from '../../../shared/components/links';
+import {useLocalStorage} from '../../../shared/hooks/uselocalstorage';
 import {getPodName, getTemplateNameFromNode} from '../../../shared/pod-name';
 import {services} from '../../../shared/services';
 import {FullHeightLogsViewer} from './full-height-logs-viewer';
+
+const TZ_LOCALSTORAGE_KEY = 'DEFAULT_TZ';
+
+const DEFAULT_TZ = process.env.DEFAULT_TZ || 'UTC';
 
 interface WorkflowLogsViewerProps {
     workflow: models.Workflow;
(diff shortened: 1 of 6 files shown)


alexec:
How does this work once webpack builds the app? I don’t think it has access to time zone anymore.

isubasinghe:
@alexec it is basically a C macro is my understanding according to this: https://webpack.js.org/plugins/define-plugin/. 
All it does is text replacement: process.env.DEFAULT_TZ becomes "UTC". 

I can also remove this, I just thought it being configurable is nice. 
Alternatively we can use dotenv files.

JPZ13 (commented):
Hey @isubasinghe - the way the build process works for the UI is that webpack bakes these environment variables in a buildtime rather than getting them at runtime, which is annoying but common with client-side single-page-apps. To set up a way to get configurable items, there are a couple solutions:
1. Build an API endpoint that the UI can hit - this will have access to environment variables in the argo-server
2. Implement server-side rendering
3. Other stuff that I haven't seen yet

Happy to continue the thread here or chat about it on our next call

isubasinghe:
@JPZ13 oh yeah I am aware of that aspect, correct me if I am wrong but I believe @alexec concern was that the process.env.DEFAULT_TZ was not available anymore after webpack has built the JS bundle, which I don't believe is correct. DefinePlugin should strip out  process.env.DEFAULT_TZ and replace it with just 'UTC', like this static build time thing was specifically what I was aiming for with this. 

I wanted to just have a configurable default timezone in the webapp that someone can configure as a part of the **build process** itself. 

This baking step is done here through webpack itself here https://github.com/argoproj/argo-workflows/pull/10120/files#diff-17798992c0235eb019e197c72e2967fd83ba5e2703ae7672ae54c56141e90aa8R63. 

I made a minimal web page here to demonstrate that process.env.DEFAULT_TZ is in fact not undefined here: https://github.com/isubasinghe/argo-workflows/blob/test-webpack/ui/src/app/testapp.tsx. 

It is 'UTC' and this value is from the previous DefinePlugin call to webpack. 


I hope I made my intentions more clearer now. 
Honestly perhaps reaching out to the server and querying its configured timezone is better, but for now imo this baking is good enough. Happy to change it if you disagree though.

JPZ13:
Ah understood. It does seem like an odd developer experience to me. The user would have to create a separate docker image for the new timezone that they want to add. It seems strange to make the user do that rather than have the environment variable for `DEFAULT_TZ` be configurable on the argo server deployment

isubasinghe:
@JPZ13 the intention is that this is more something we change when we(Argo Project members) would like to change the default timezone we ship with. 

The timezone is still configurable from the UI. Here is the image where one may select the timezone they would like to view logs in.
image

I just pushed a new change which saves the selected timezone into local storage as well. A fresh workflow run results in the following output, because I detected that I had previously chosen AEDT as my timezone: 
image


Sorry to be a pain on this haha, I am going to retract what I said before regarding querying the server. I understand the default timezone being the same on the server and the client is a nice to have, but not convinced it is worth the effort.  Especially given that the timezone the user selects is now persisted on localstorage.

JPZ13:
> Sorry to be a pain on this haha

You're not being a pain in the slightest @isubasinghe! Thanks for answering all my clarifying questions

I understand what you're getting at now. My apologies for the confusion on my end. Storing the timezone in local storage makes sense. Let's not worry about the API endpoint for now. If someone really wants it, we can always add it in later. Always better to ship small, incremental progress. I'll clone down the latest changes later today and give it a review

ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
```diff
@@ -26,13 +32,60 @@ function identity<T>(value: T) {
...
     const [grep, setGrep] = useState('');
     const [error, setError] = useState<Error>();
     const [loaded, setLoaded] = useState(false);
     const [logsObservable, setLogsObservable] = useState<Observable<string>>();
+    // timezone used for ui rendering only
+    const [uiTimezone, setUITimezone] = useState<string>(DEFAULT_TZ);
+    // timezone used for timezone formatting
+    const [timezone, setTimezone] = useLocalStorage<string>(TZ_LOCALSTORAGE_KEY, DEFAULT_TZ);
+    // list of timezones moment tz supports
+    const [timezones, setTimezones] = useState<string[]>([]);
```
juliev0, line 84: is there a typo on line 83? can you clarify this comment?

ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
```diff
@@ -116,7 +186,20 @@ export const WorkflowLogsViewer = ({workflow, nodeId, initialPodName, container,
...
+                        <input type='search' defaultValue={logFilter} onChange={v => setLogFilter(v.target.value)} placeholder='Filter (regexp)...' />
+                        <i className='fa fa-globe' />{' '}
+                        <Autocomplete
+                            items={filteredTimezones}
+                            value={uiTimezone}
+                            onChange={v => setUITimezone(v.target.value)}
+                            onSelect={tz => {
+                                setUITimezone(tz);
+                                setTimezone(tz);
+                            }}
```
juliev0, line 200: If I'm understanding correctly, it looks like setUITimezone() would be called automatically when timezone changes (lines 86-88), so maybe setUITimezone() wouldn't need to be explicitly called on line 198?

ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
isubasinghe: No, I just meant that it stores the list of timezones the library supports. I abbreviated timezone with tz. 
I'll make it clearer now.

ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
isubasinghe: Ah good pickup, this isn't needed anymore. I forgot to update this line after introducing the useEffect.

juliev0 (approved):

isubasinghe:
not stale

isubasinghe:
not stale

JPZ13 (approved):

sarabala1979 (approved):

PR #10135, opened 2022-11-29
build: Use alpine, buildkit, and refactor pkg caching. Fixes #10134

### LINKED ISSUE #10134: Speed up docker image builds
# Summary
Docker image builds for Argo Workflows take a long time (10 minutes per image) on my machine (Macbook Pro M1). This makes reviewing PRs and developing a hassle. Fortunately, we can speed up the build times by using Buildkit and better caching in the Dockerfile

# Use Cases
This speeds up iteration time while:
1. Developing
2. Reviewing PRs

---

JPZ13:
This PR:
- Refactors the Dockerfile to cache packages in the builder stage
- Caches go modules
- Caches node modules
- Uses alpine images for the builder and ui
- Refactors the makefile to use buildkit

Note: On this PR, I can get the build of all three images needed to run the codebase down to about 30 seconds total (with caching), where a build of all three images on the master branch takes about 30 minutes total on my machine

Fixes #10134

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ WORKDIR /tmp
 WORKDIR /go/src/github.com/argoproj/argo-workflows
 COPY go.mod .
 COPY go.sum .
-RUN go mod download
+RUN --mount=type=cache,target=/go/pkg/mod go mod download
 
 COPY . .
 
@@ -35,12 +35,17 @@ FROM node:16 as argo-ui
 
 COPY ui/package.json ui/yarn.lock ui/
 
-RUN JOBS=max yarn --cwd ui install --network-timeout 1000000
+RUN --mount=type=cache,target=/root/.yarn \
+  YARN_CACHE_FOLDER=/root/.yarn JOBS=max \
+  yarn --cwd ui install --network-timeout 1000000
 
 COPY ui ui
 COPY api api
 
-RUN NODE_OPTIONS="--max-old-space-size=2048" JOBS=max yarn --cwd ui build
+RUN --mount=type=cache,target=/root/.yarn \
+  --mount=type=cache,target=/assets \
+  YARN_CACHE_FOLDER=/root/.yarn JOBS=max \
+  NODE_OPTIONS="--max-old-space-size=2048" JOBS=max yarn --cwd ui build
 
 ####################################################################################################
 
@@ -60,7 +65,7 @@ RUN curl -o /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/j
 RUN cat .dockerignore >> .gitignore
 RUN git status --porcelain | cut -c4- | xargs git update-index --skip-worktree
 
-RUN --mount=type=cache,target=/root/.cache/go-build make dist/argoexec
+RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build make dist/argoexec
 
 ####################################################################################################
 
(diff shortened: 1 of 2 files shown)


terrytangyuan (commented):
It's fast on my laptop. Maybe it's because Buildkit is enabled by default on Docker Desktop. 

I think we are missing `DOCKER_BUILDKIT=1` here?

JPZ13:
I'm using Buildkit in this PR with the buildx command.
Docs: https://docs.docker.com/engine/reference/commandline/buildx/
Makefile line: https://github.com/argoproj/argo-workflows/pull/10135/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R234

This way it'll use Buildkit even without the environment variable set, but regardless, the Dockerfile refactor will speed up the image builds to take just seconds after the first time building the image locally. Previously, this was taking a long time on my machine, even with Buildkit enabled because the go packages aren't cached in the builder stage and the kubectl binary gets re-downloaded each time someone changes a file

terrytangyuan:
Sounds good. We are actually already using `buildx` for release builds but not for CI build yet. https://github.com/argoproj/argo-workflows/blob/master/.github/workflows/release.yaml

JPZ13:
Nice! I wonder if this will make release builds build faster. I'd also be down to open up a ticket to add buildkit for the CI if that's desired

Dockerfile
```diff
@@ -71,31 +59,33 @@ FROM builder as workflow-controller-build
...
 RUN cat .dockerignore >> .gitignore
 RUN git status --porcelain | cut -c4- | xargs git update-index --skip-worktree
 
-RUN --mount=type=cache,target=/root/.cache/go-build make dist/argo
+RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build make dist/argo
+
+####################################################################################################
+
+# NOTE: kubectl version should be one minor version less than https://storage.googleapis.com/kubernetes-release/release/stable.txt
+FROM bitnami/kubectl:1.24.8 as kubectl
```
rohankmr414, line 81: do we need this step, can't we curl the binary in the `builder` step and copy it later?
https://github.com/argoproj/argo-workflows/blob/1225d8b546c1e5093047d8e6e8a46d053f051d97/Dockerfile#L52-L53

https://github.com/argoproj/argo-workflows/blob/1225d8b546c1e5093047d8e6e8a46d053f051d97/Dockerfile#L97

JPZ13: Using the bitnami base image is less verbose and keeps the builder step smaller. Additionally, if we want to upgrade the kubectl version, it'll only affect the argoexec image rather than all three images, which would be the case if we move the kubectl import to the builder step

JPZ13: Also, using the bitnami import will prevent us from having to worry about the OS/arch when importing kubectl, which is another nice to have

isubasinghe (approved):

Dockerfile
```diff
@@ -1,66 +1,54 @@
...
-        /usr/share/doc-base
-
-WORKDIR /tmp
-
-# https://blog.container-solutions.com/faster-builds-in-docker-with-go-1-11
+    curl \
+    gcc \
+    bash \
+    jq \
+    mailcap
```
terrytangyuan, line 14: We might need to double-check whether all these changes are okay to avoid causing any missing dependencies/binaries.

terrytangyuan:
This works great for me! Left one comment.

Dockerfile
JPZ13: We do use the distroless static base image for the final stages in the Dockerfile: https://github.com/argoproj/argo-workflows/pull/10135/files/9987392605dff9421a1beb8c8d8e562979bd3724#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R85

That means none of the binaries from the builder stage get included unless explicitly copied from the base layer. Also, the final steps will complain if there's something missing from the builder step. I had to add `mailcap` to get the `/etc/mime.types` to show up

JPZ13:
> This works great for me! Left one more comment.

Thanks, @terrytangyuan!

terrytangyuan (approved):
LGTM. @alexec Want to take another look?

rohankmr414 (approved):

alexec:
LGTM

PR #10140, opened 2022-11-30
fix (argo wait): use functions to constrain ctx instead of blocks

sxllwx:
Signed-off-by: scott <scottwangsxll@gmail.com>

Fixes: as the title says

### DIFF
--- a/cmd/argoexec/commands/wait.go
+++ b/cmd/argoexec/commands/wait.go
@@ -32,8 +32,8 @@ func waitContainer(ctx context.Context) error {
 	defer stats.LogStats()
 	stats.StartStatsTicker(5 * time.Minute)
 
-	// use a block to constrain the scope of ctx
-	{
+	// use a function to constrain the scope of ctx
+	func() {
 		// this allows us to gracefully shutdown, capturing artifacts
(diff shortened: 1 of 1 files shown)


alexec:
What problem does this fix?

sxllwx:
Please correct me if I understand wrong, thx.


```golang
     // use a function to constrain the scope of ctx
      {
		// this allows us to gracefully shutdown, capturing artifacts
		ctx, cancel := signal.NotifyContext(ctx, syscall.SIGTERM)
		defer cancel()

		// Wait for main container to complete
		err := wfExecutor.Wait(ctx)
		if err != nil {
			wfExecutor.AddError(err)
		}
	}
```

As mentioned in the comment, the block here is to limit the scope of ctx.  expected execution result is to call `cancel` after the scope of ctx ends (that is, when the code segment of the block is executed). However, due to the use of the `defer` keyword, defer will only be executed after the end of the function, not after the end of the code segment in the block.

sample-code: https://go.dev/play/p/Bvwy9tswNm8

ps: The current code runs without any problems. This patch will make our code more grammatically correct. 😄

alexec (approved):
Better grammar is good!

PR #10188, opened 2022-12-08
docs: add makefile improvement proposal

isubasinghe:


rohankmr414:
@isubasinghe I'd suggest we also try experimenting with [Vagrant](https://www.vagrantup.com/). It is used to spin up isolated development environments inside VMs or docker, the config for that will be stored in the repo in a file named `Vagrantfile`.

docs/proposals/makefile-improvement-proposal.md
```diff
@@ -0,0 +1,68 @@
...
+
+### What will Nix get us?
+
+As mentioned previously Nix gets us closer to reproducible build environment's. It should ease significantly the on-boarding process of developers onto the project.
+There have been at least one confirmed developer who wanted to work on Argo Workflows but found the Makefile to be a barrier, it is likely that there are more developers on this boat. With a reproducible build environment, we hope that
```
caelan-io, line 50: We've had several mentees provide feedback that getting started with the Argo Workflows makefile is challenging and time consuming, so I would say 5+ developers here instead of 1.

docs/proposals/makefile-improvement-proposal.md
```diff
@@ -0,0 +1,68 @@
...
+everyone who would like to contribute to the project is able to do so easily. It should also save time for engineers on-boarding onto the project, especially if they are using a more niche
```
caelan-io, line 51: I don't think Linux systems like Fedora are niche in the realm of OSS.

docs/proposals/makefile-improvement-proposal.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Proposal
+
+In order to maximize the benefit we receive from using something like Nix, I suggest that we initially start off with a modest change to the Makefile.
+The first proposal would be to remove out all dependency management code and replace this functionality with Nix, where it is trivially possible. This may not be possible for some go lang related binaries we use, we will retain the Makefile functionality in those cases, at least for a while. Eventually we will migrate more and more of this responsibility away from the Makefile. Following Nix being responsible for all dependency management, we could start to consider moving more of our build system itself into Nix, perhaps it is easiest to start off with UI build as it is relatively painless. However, do note that this is not a requirement, I do not see a problem with the Makefile and the Nix file co-existing, it is more about finding a good balance between the reproducibility we desire and the effort we put into obtaining said reproducibility
```
caelan-io, line 45: Can you please point to the specific lines in the makefile that you would begin replacing with Nix so we can see an example or two?

docs/proposals/makefile-improvement-proposal.md
```diff
@@ -0,0 +1,68 @@
...
+## Introduction
+
+The motivation for this proposal is to enable developers working on Argo Workflows to use build tools in a more reproducible way.
+Currently the Makefile is unfortunately too opinionated and as a result is often a blocker when first setting up Argo Workflows locally.
+I believe we should shrink the responsibilities of the Makefile and where possible outsource areas of responsibility to more specialized technology, such
+as Devenv/Nix in the case of dependency management.
+
+## Proposal Specifics
+
+In order to better address reproducibility, it is better to split up the duties the Makefile currently performs into various sub components, that can be assembled in more appropriate technology. One important aspect here is to completely shift the responsibility of dependency management away from the Makefile and into technology such as Nix or Devenv.
```
caelan-io, line 12: I would also add a sentence here that this will improve access to the Argo Workflows open source code for non-Mac devs and make it easier to run it locally.

caelan-io (approved):

isubasinghe:
Thanks for the review @caelan-io , addressed the feedback.

alexec:
Getting started is a real issue. I think think the right solution, because we're on Github, it Codespace. Have you experimented with that?

isubasinghe:
@alexec I think that works as well but I think having Nix could be in addition to that to provide a better build experience on local machines. 

I think I can get a Nix file up such that the dependencies we use will be the exact same for everyone, I don't think I even have to modify the Makefile for this. This alone should solve most of the issues I've had in the past with getting Argo Workflows running on Fedora.

alexec (commented):
1. I agree that the `Makefile` today is too complex. It has many options that are not used and could be simplified removed, e.g. KUBERNETES_NAMESPACE. 
2. Let's use DevContainers for the tooling. This is much more mature that it was 1 years ago and is now usable.
3. Let's avoid any tools that are programatic in nature rather than declarative. They're always harder to understand, which results in a weaker eco-system, and less popularity and adoption. I've not heard of Nix or Devenv, but we already have a toolchain.

@isubasinghe if you're interested, I can list out the improvement we could make to the `Makefile` to reduce it complexity and rely more on the underlying tools.

caelan-io:
Thanks @alexec - yes if you're able to list out the improvements you have in mind, we can make the changes:
> I can list out the improvement we could make to the Makefile to reduce it complexity and rely more on the underlying tools.

> I don't think we need PR for proposals to build tool chains. Let's open an issue instead.
@isubasinghe opened this proposal at @sarabala1979 's request. Can we keep the discussion thread going here until we have another PR opened?

docs/proposals/makefile-improvement-proposal.md
```diff
@@ -0,0 +1,67 @@
+# Proposal for Makefile improvements
+
+## Introduction
+
+The motivation for this proposal is to enable developers working on Argo Workflows to use build tools in a more reproducible way.
+Currently the Makefile is unfortunately too opinionated and as a result is often a blocker when first setting up Argo Workflows locally.
```
juliev0, line 6: Sorry, can you clarify what you mean by "opinionated"?

docs/proposals/makefile-improvement-proposal.md
```diff
@@ -0,0 +1,67 @@
...
+
+### What will Nix get us?
+
+As mentioned previously Nix gets us closer to reproducible build environments. It should ease significantly the on-boarding process of developers onto the project.
+There have been several developers who wanted to work on Argo Workflows but found the Makefile to be a barrier, it is likely that there are more developers on this boat. With a reproducible build environment, we hope that
```
juliev0, line 50: Can you clarify some of the problems faced? And were they Go modules or React/JS/TS?

docs/proposals/makefile-improvement-proposal.md
isubasinghe: I've found that the Makefile makes assumptions about the system it's running on, for example that the system is either running macOS or Ubuntu. 

It will attempt to use apt to install packages for example. 

I think a build system should just not make any assumptions without proving them to be true. Nix is a hermetic build system/package manger which ensures all assumptions made can be proven.

docs/proposals/makefile-improvement-proposal.md
isubasinghe: Well I run a Fedora machine, the Makefile just isn't suited for this. 
When I initially on boarded to this project,  I had to spend a week or so getting all the dependencies exactly right. 
I went on leave and came back only to have to spend another two days to get Argo Workflows working again. 

In fact, the Makefile is broken on my laptop as we speak. My development process when working from my laptop is to push changes, pull them to my desktop and check everything on there. 

`make pre-commit` pretty much never works for me.

I think providing a Hermetic build system is a massive improvement over this. 
With my latest PR, I was able to get Argo running in < 15 minutes. This was a huge improvement.

isubasinghe: I do not believe having global installs of packages like [this](https://github.com/argoproj/argo-workflows/blob/master/Makefile#L640) are great for reproducibility. 
For example it might be entirely feasible for some pip packages to not work on certain versions of Python. 

Argo relies on a lot of external tooling, I think it would be a good idea to pin these down to exact versions. 
As an example [these](https://github.com/argoproj/argo-workflows/pull/10999/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R37-R141) download exactly a known working combination of python interpreter and packages.

isubasinghe:
@juliev0 As a general comment, we can keep the Makefile, I suggest Nix as an additional option to make life easier for people who are not on OSX/Ubuntu or just want a reproducible build environment. I strongly believe the two can co-exist.

isubasinghe:
> @isubasinghe I'd suggest we also try experimenting with [Vagrant](https://www.vagrantup.com/). It is used to spin up isolated development environments inside VMs or docker, the config for that will be stored in the repo in a file named `Vagrantfile`.

Sorry, I forgot to reply. But yeah I think we can have a look at this as well. I also think it might be trivial to create a VM image from this Nix file itself. See this https://github.com/nix-community/awesome-nix#virtualisation

docs/proposals/makefile-improvement-proposal.md
juliev0: yes, that makes a lot of sense. I guess macos developers probably weren't thinking about users of other OS's.

juliev0:
> I don't think we need PR for proposals to build tool chains. Let's open an issue instead.

@alexec Isitha opened up a PR for this proposal because it's more likely to get seen by more people and by folks like you than opening an Issue. Also, I had mentioned that this was the way to go. I was originally told to put the ArtifactGC proposal in this same directory. As I recall, this is what Argo CD is doing.

Also, Isitha opened up a [PR](https://github.com/argoproj/argo-workflows/pull/10999) for Nix. It sounds like it would address the big hurdles of devs getting started on non OSX/Ubuntu. And it sounds like it would not affect the current processes for those who want to use it. Any reason in that case not to merge it? I know you mentioned an alternative strategy, so if you feel that that should be put in place instead please provide the details. Thanks.

> 1. I agree that the `Makefile` today is too complex. It has many options that are not used and could be simplified removed, e.g. KUBERNETES_NAMESPACE.
> 2. Let's use DevContainers for the tooling. This is much more mature that it was 1 years ago and is now usable.
> 3. Let's avoid any tools that are programatic in nature rather than declarative. They're always harder to understand, which results in a weaker eco-system, and less popularity and adoption. I've not heard of Nix or Devenv, but we already have a toolchain.
> 
> @isubasinghe if you're interested, I can list out the improvement we could make to the `Makefile` to reduce it complexity and rely more on the underlying tools.
>

terrytangyuan (commented):
One concern is that a lot of the commands seem to duplicate what's in Makefile. If we can somehow refactor that it would be great since this will double maintenance effort. An example: https://github.com/argoproj/argo-workflows/pull/10999#discussion_r1186396202

JPZ13:
Not stale

juliev0:
since we already merged the actual work, I can enable auto-merge for this

terrytangyuan (approved):

Issue #8684, opened 2022-05-09
Workflow deadlocks on mutex in steps template if controller is restarted


Gibstick:
## Checklist

 

* [x] Double-checked my configuration
* [x] Tested using the latest version.
* [x] Used the Emissary executor.

## Summary

**What happened/what you expected to happen?**

I expect my workflow to complete successfully even if the workflows controller is restarted. The same issue does not occur if the mutex is on the container template or at the workflow level.

**What version are you running?**
v3.3.4 is what I've been using for this reproduction, but I've also been able to reproduce it in v3.3.3, v3.3.2, and v3.2.7 with a more complicated workflow.

## Diagnostics

This is a workflow that reliably reproduces the issue. The steps I use are:
```
argo submit deadlock.yaml
sleep 5
kubectl delete pod -l app.kubernetes.io/component=workflow-controller
```

```yaml
# deadlock.yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: deadlock-test-
spec:
  entrypoint: main-steps

  templates:
    - name: main-steps
      synchronization:
        mutex:
          name: "mutex-{{workflow.name}}"
      steps:
        - - name: main
            template: main-container

    - name: main-container
      container:
        image: "busybox:latest"
        command: ["sh"]
        args: ["-c", "sleep 10; exit 0"]
```

**Pod logs**
No pod logs from the main container since it just sleeps and exits.

**Wait container logs**
<details>

```
time="2022-05-09T12:57:42.216Z" level=info msg="Starting Workflow Executor" executorType=docker version=v3.3.2
time="2022-05-09T12:57:42.217Z" level=info msg="Creating a docker executor"
time="2022-05-09T12:57:42.217Z" level=info msg="Using executor retry strategy" Duration=1s Factor=1.6 Jitter=0.5 Steps=5
[...]

PR #10009, opened 2022-11-10
[WIP] Fix deadlock when controller is restarted. Fixes #8684

isubasinghe:
Fixes #8684

Please do not open a pull request until you have checked ALL of these:

* [ ] Create the PR as draft .
* [ ] Run `make pre-commit -B` to fix codegen and lint problems.
* [ ] Sign-off your commits (otherwise the DCO check will fail).
* [ ] Use [a conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) (otherwise the commit message check will fail).
* [ ] "Fixes #" is in both the PR title (for release notes) and this description (to automatically link and close the issue).
* [ ] Add unit or e2e tests. Say how you tested your changes. If you changed the UI, attach screenshots.
* [ ] Github checks are green.
* [ ] Once required tests have passed, mark your PR "Ready for review".

If changes were requested, and you've made them, dismiss the review to get it reviewed again.

PR #10191, opened 2022-12-08
docs: add mutex improvement proposal

isubasinghe:
Related to #8684


sarabala1979:
@isubasinghe Can you present in the contributor meeting? We can discuss

caelan-io (approved):

isubasinghe:
update: Discussed with @sarabala1979 what the next steps are for this. Bala mentioned to try replace the getResourceKey with getHolderKey. I thought during the meeting this would work as well, looking into that again this will not work without changing some data in one of the structs. 

@sarabala1979 could you also please clarify here why we can't go with the config map approach? I believe it was about push back regarding introducing another resource? 

The next steps for this would be to go with modifying the format of the key in the MutexHolding struct. @sarabala1979 with your okay, I will proceed with this fix.

isubasinghe:
not stale and not resolved.

isubasinghe:
Hey @juliev0 or @terrytangyuan  could you please give me some feedback on this if you have any time available? Mutexes have been broken since they were implemented. I have a partial solution here: https://github.com/argoproj/argo-workflows/pull/10009.
 
But I need to know if I can go ahead with this approach before continuing to invest more time/money in it. 

TL;DR -> store a couple more mutex related information in config maps

terrytangyuan:
I haven't got a chance to look into details yet but would this be a breaking change for current users who rely on mutexes?

isubasinghe:
@terrytangyuan it shouldn't change anything for users, it will introduce an additional config map where mutex related data is stored.

docs/proposals/mutex-improvements.md
```diff
@@ -0,0 +1,96 @@
...
+* Relatively simple change.
+* Single source of truth
+
+##### Disadvantages of solution #1
+
+* Changing a field could be major change and might break projects in the ecosystem that rely on this behavior. This has been the behavior of mutexes ever since they have been implemented, it is high risk to change this in my opinion.
+
+#### Store holding/pending information in a config map
+
+We can store the holder keys inside a config map, on release, we refer to the holder keys inside this config map. There is an existing WIP PR for this, [here](https://github.com/argoproj/argo-workflows/pull/10009).
```
sarabala1979, line 71: Do you think there is a security concern in multi-tenant env?  If Controller updates the holder on Configmap, The controller should have cluster-level RBAC to update the config map across the namespace.

docs/proposals/mutex-improvements.md
```diff
@@ -0,0 +1,96 @@
...
+
+### How should they work?
+
+We should not be generating information after it has already been generated. It is extremely important to only maintain a single source of truth,
+generating this information (the holder key) produces sources of truth at every generation point. Ideally we should be storing this information somewhere
+as we will be needing it eventually when a release call is made.
+
+### Solutions
+
+#### Store the holder key directly in the MutexStatus/SemaphoreStatus structs
```
sarabala1979, line 53: How will these solutions form the holder keys?  Will it use string format?
Can you provide the sample yaml how it will look?

docs/proposals/mutex-improvements.md
```diff
@@ -0,0 +1,96 @@
...
+
+* More sensible approach than including a new dependency (ConfigMap).
+* Relatively simple change.
+* Single source of truth
+
+##### Disadvantages of solution #1
+
+* Changing a field could be major change and might break projects in the ecosystem that rely on this behavior. This has been the behavior of mutexes ever since they have been implemented, it is high risk to change this in my opinion.
+
+#### Store holding/pending information in a config map
```
sarabala1979, line 69: Can you add sample configmap content?

docs/proposals/mutex-improvements.md
```diff
@@ -0,0 +1,96 @@
...
+We can store the holder keys inside a config map, on release, we refer to the holder keys inside this config map. There is an existing WIP PR for this, [here](https://github.com/argoproj/argo-workflows/pull/10009).
+It requires handling pending workflows. If we are going with this solution, a small amendment to deal with pending items will have to be made, we may have to introduce two config maps. One will be used for storing information regarding acquired locks, the other for storing information regarding pending lock acquisitions.
+There is a possibility that it might be possible to use a single config map here, but that solution needs to be explored in order to confirm this.
+
+##### Advantages of solution #2
+
+* The change in behavior is transparent to users.
+* Existing PR should allow for us to be quite quick in pushing a fix.
+
+##### Disadvantages of solution #2
```
sarabala1979, line 80: how can we handle if holder information crosses the 1MB limit on configmap?

JPZ13:
not stale, and not resolved

Bwvolleyball:
This issue still feels relevant, right?

docs/proposals/mutex-improvements.md
```diff
@@ -0,0 +1,96 @@
...
+`
+getHolderKey({"Namespace": "argo", Name: "example"}, "node") = "argo/example/node"
+
+-- MutexStatus After LockAcquired Call --
+items = ["argo", "example", "node"]
+holdingName = "node"
+ms.Holding = [MutexHolding{Mutex: lockKey, Holder: holdingName}]
+
+getResourceKey("argo", "example", "node") = "argo/example/node"
+`
```
agilgur5, line 23:
```suggestion
\```go
getHolderKey({"Namespace": "argo", Name: "example"}, "node") = "argo/example/node"

// -- MutexStatus After LockAcquired Call --
items = ["argo", "example", "node"]
holdingName = "node"
ms.Holding = [MutexHolding{Mutex: lockKey, Holder: holdingName}]

getResourceKey("argo", "example", "node") = "argo/example/node"
\```
```

agilgur5 (commented):
some formatting changes below that make the [rendered version](https://github.com/argoproj/argo-workflows/blob/49865b1783b481ba7600441999559821d1a03a18/docs/proposals/mutex-improvements.md) a good bit easier to read

docs/proposals/mutex-improvements.md
```diff
@@ -0,0 +1,96 @@
...
+
+##### Advantages of solution #1
+
+* More sensible approach than including a new dependency (ConfigMap).
+* Relatively simple change.
+* Single source of truth
+
+##### Disadvantages of solution #1
+
+* Changing a field could be major change and might break projects in the ecosystem that rely on this behavior. This has been the behavior of mutexes ever since they have been implemented, it is high risk to change this in my opinion.
```
agilgur5, line 67: Does this change a field on the CR itself? Or just the internal structs?

If it is just internal structs, a restart of the controller should get everything into the new format.

If it is on the CR, perhaps there could be a in-between option -- a struct that is encoded into a string?

isubasinghe: I don't think it is just internal structs unfortunately, it gets serialised into json and protobufs hence why I assumed there might be some usage here from the outside world. 

Look, honestly seems very unlikely though, think we can probably just change the struct. Chucking everything in a ConfigMap would keep things consistent between different versions. 

I am open to just changing the struct. This is basically an RFC to get some idea of what to do.

agilgur5: So I believe this impacts the `status.synchronization` field on the CR. We technically can't change the CRD without a breaking version change (e.g. `v1alpha1` -> `v1alpha2` or similar). 

We can _add_ fields though, as we do quite often. So perhaps we deprecate the existing ones and only use the new ones moving forward. In either manner, we would need to ensure backward-compat with older Workflows that used the old format.

I think that would mitigate the breaking change part and not have the drawbacks of additional ConfigMaps

isubasinghe:
closed thanks to https://github.com/argoproj/argo-workflows/pull/13553

PR #10314, opened 2023-01-05
fix: fix minio image at older working version

### LINKED ISSUE #10310: new release of minio Docker image gets broken pipe error when port forwarded on localhost
### Pre-requisites

If you look at manifests/quick-start/base/minio/minio-service.yaml it deploys latest minio image, and the service for that gets port forwarded to localhost. If I try to log into localhost:9001 with username="admin" and password="password", I can't login and instead see:
```
E0104 18:03:24.034605   81979 portforward.go:378] error copying from remote stream to local connection: readfrom tcp6 [::1]:9001->[::1]:54749: write tcp6 [::1]:9001->[::1]:54749: write: broken pipe
E0104 18:03:54.035287   81979 portforward.go:346] error creating error stream for port 9001 -> 9001: Timeout occurred
```

If I use an older version of minio it works. It seems to break at `minio/minio:RELEASE.2022-12-12T19-27-27Z` I think. 

We can either fix the image at the previous version or investigate what changed and/or log a bug on their side.

### Version

latest

```YAML
n/a
```

```text
n/a
```

```text
n/a
```

juliev0:
Signed-off-by: Julie Vogelmani <julie_vogelman@intuit.com>

Fixes #10310

### DIFF
--- a/manifests/quick-start-minimal.yaml
+++ b/manifests/quick-start-minimal.yaml
@@ -1749,7 +1749,7 @@ spec:
           value: admin
         - name: MINIO_SECRET_KEY
           value: password
-        image: minio/minio
+        image: minio/minio:RELEASE.2022-11-17T23-20-09Z
         lifecycle:
           postStart:
             exec:
--- a/manifests/quick-start-mysql.yaml
+++ b/manifests/quick-start-mysql.yaml
@@ -1793,7 +1793,7 @@ spec:
           value: admin
         - name: MINIO_SECRET_KEY
           value: password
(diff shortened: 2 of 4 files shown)


sarabala1979:
@juliev0 you should change it in https://github.com/argoproj/argo-workflows/tree/master/manifests/quick-start/base/minio. you change on generated files. if you do `make manifest` it will override it.

sarabala1979 (changes requested):
wrong files

sarabala1979 (approved):

PR #10343, opened 2023-01-11
feat: ensure immediate deletion of pods. Fixes #10219

### LINKED ISSUE #10219: Handle rate limiting for large operations
# Summary
We're working with a customer who is running GKE and Argo Workflows for data processing workloads. Occasionally, the CPU and memory on the workflow controller grows exponentially causing the workflow controller to crash. We think this is being caused by the pod cleanup being rate limited by the GKE API and the internal queue backing up, but we're not sure because there isn't clear logging telling us that it's happening

Note: We spoke with @sarabala1979 during the last contributors' meeting about a similar issue. I'm creating this issue to track out work around getting better logging around rate limiting

# Use Cases
This would be useful for users whose K8s distro has some sort of rate limiting to have better insights on what's happening during the CPU/memory buildup

---

isubasinghe:

Fixes ##10219

### DIFF
--- a/workflow/controller/pod_cleanup.go
+++ b/workflow/controller/pod_cleanup.go
@@ -8,8 +8,9 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 
-	"github.com/argoproj/argo-workflows/v3/workflow/common"
 	"github.com/sirupsen/logrus"
+
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
 
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v3/workflow/controller/indexes"
@@ -66,7 +67,7 @@ func (woc *wfOperationCtx) runImmediateCleanup(pod *apiv1.Pod, podGC *wfv1.PodGC
 
 	var lselector metav1.LabelSelector
 	if woc.wf.Status.Fulfilled() {
-		log.Info("workflow is fullfilled, deleting everything which may be deleted as per PodGC.Strategy.LabelSelector")
+		log.Info("workflow is fulfilled, deleting everything which may be deleted as per PodGC.Strategy.LabelSelector")
 		lselector = getLabelSelector(wfName, podGC, false)
 	} else {
 		log.Info("deleting existing pods marked as completed")
@@ -112,7 +113,7 @@ func (woc *wfOperationCtx) cleanupPods() {
 		log := log.WithField("podName", pod.Name)
 
 		// specific handling for OnWorkflowCompletion
-		// we will be creating unecessary DeleteCollection requests after the first one
+		// we will be creating unnecessary DeleteCollection requests after the first one
 		// to fix this, we would need a variable in wfOperationCtx. I think it is possibly okay to avoid this optimisation for now.
 		if woc.isImmediateDeletionStrategy() {
 			err := woc.runImmediateCleanup(pod, podGC, workflowPhase)
--- a/workflow/controller/pod_cleanup.go
+++ b/workflow/controller/pod_cleanup.go
@@ -1,21 +1,102 @@
 package controller
 
 import (
+	"context"
+
 	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


workflow/controller/pod_cleanup.go
```diff
@@ -1,21 +1,104 @@
...
 	"github.com/argoproj/argo-workflows/v3/workflow/common"
+	"github.com/sirupsen/logrus"
 
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v3/workflow/controller/indexes"
 )
 
-func (woc *wfOperationCtx) queuePodsForCleanup() {
+func (woc *wfOperationCtx) isImmediateDeletionStrategy() bool {
+	return woc.orig.Spec.PodGC != nil && woc.orig.Spec.PodGC.Strategy == wfv1.PodGCOnWorkflowCompletion
```
isubasinghe, line 19: This should change to something else, what should it be?

isubasinghe: I think we should add `PodGCImmediate` as a GC strategy.

workflow/controller/pod_cleanup.go
```diff
@@ -1,21 +1,103 @@
...
+			metav1.PatchOptions{},
+		)
+		if err != nil {
+			log.Errorf("was not able to patch pods with completed as true for pods %s in workflow %s due to %s hence abandoning deletion", pod.Name, woc.execWf.ObjectMeta.Name, err)
+			return err
+		}
+	}
+
+	var lselector metav1.LabelSelector
+	if woc.wf.Status.Fulfilled() {
```
isubasinghe, line 69: No need to run on this in a loop, write to a variable in wfOperationCtx to indicate everything has been deleted and then perform an early exit

isubasinghe:
We may hit the rate limit with this solution unfortunately since there is no mechanism to stop that here. 

I have some ideas that may help here: 
1. We rate limit on the function itself. This is actually fairly doable with a constant amount of memory, so perhaps this should be the solution. Additionally we actually don't need to run the cleanup function multiple times, we just need it to run. 
This might be our solution, **but if we imagine a situation where a huge chunk of pods were immediately updated, then we might once again run into rate limitation issues**.  

2. We do the batch deletion as originally talked about. This might come at increased cloud costs


So yeah this approach is not without it's tradeoffs unfortunately, but I hope my proposals are good enough to mitigate the cons @JPZ13

JPZ13:
Hey @isubasinghe - to me, option 3 makes the most sense for scaling to high levels while keeping cloud costs down. What would you recommend for a persistent queue following this route? I would guess that we would need a queue Kubernetes deployment and then we could either have a separate goroutine running on the workflow controller to handle the messages or we could spin up an entirely separate deployment for the handler

isubasinghe:
@JPZ13 Honestly I think even just using an embedded queue would be good enough given we are running only a single controller. Perhaps we could embed nsq like this example: https://gist.github.com/joshrotenberg/ad49d39dbee8b48789d9 

But honestly I also don't see why we don't just wait, its not like the most technical solution hahaha, but I suspect it should be a good enough solution in mean time at least. 

I do think this is something we should properly discuss about sometime though. I've also opened https://github.com/argoproj/argo-workflows/issues/10360, I think it is feasible that other memory growth issues contribute to the issue as well.

isubasinghe:
Also we could just add the rate limiting from https://github.com/argoproj/argo-workflows/pull/10359 to this PR too.

workflow/controller/pod_cleanup.go
JPZ13: `OnPodCompletion` is already a strategy that I'd consider to be immediate

JPZ13: Current options are here: https://argoproj.github.io/argo-workflows/fields/#podgc

JPZ13:
@isubasinghe I think the rate limiter makes sense as a minimum viable solution. Let's:
- Consolidate your work from this PR and #10359
- Allow the rate limiter interval to be passed in as an environment variable, something like `RATE_LIMIT_INTERVAL_MILLISECONDS` or whatever unit you think makes sense

We can separate out the memory leak stuff in another PR and then tackle queueing separately

isubasinghe:
@JPZ13 yeah that sounds good to me, on that now.

JPZ13:
Closing in favor of #10379

PR #10379, opened 2023-01-21
feat: allow deletion of pods via direct API call. Fixes #10219

### LINKED ISSUE #10219: Handle rate limiting for large operations

isubasinghe:
Fixes #10219


workflow/controller/operator.go
```diff
@@ -140,6 +143,20 @@ type failedNodeStatus struct {
 
 // newWorkflowOperationCtx creates and initializes a new wfOperationCtx object.
 func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOperationCtx {
+	var rateLimiter *waitutil.RateLimiter
+	rateLimiter = nil
```
NikeNano, line 147: Do we need to set this to nil?

workflow/controller/pod_cleanup.go
```diff
@@ -26,7 +146,24 @@ func (woc *wfOperationCtx) queuePodsForCleanup() {
...
-		switch determinePodCleanupAction(selector, pod.Labels, strategy, workflowPhase, pod.Status.Phase) {
+
+		// specific handling for OnWorkflowCompletion
+		// we will be creating unnecessary DeleteCollection requests after the first one
+		// to fix this, we would need a variable in wfOperationCtx. I think it is possibly okay to avoid this optimisation for now.
+		if woc.isImmediateSkipQueues() {
+			err := woc.runImmediateCleanup(pod, podGC, workflowPhase)
+			if err != nil {
+				log.Errorf("was unable to run cleanup due to %s", err)
+			}
```
NikeNano, line 157: nit: 
```suggestion
			if err := woc.runImmediateCleanup(pod, podGC, workflowPhase); err != nil {
				log.Errorf("was unable to run cleanup due to %s", err)
			}
```

workflow/controller/pod_cleanup.go
```diff
@@ -1,22 +1,142 @@
...
+	// lets patch everything up
+	if woc.wf.Status.Fulfilled() {
+		markDone := true
+		podList, err := pods.List(context.Background(), metav1.ListOptions{LabelSelector: selectorString})
+		if err != nil {
+			log.Errorf("unable to obtain list of pods due to %s", err)
+		} else {
+			for _, nonCompletedPod := range podList.Items {
+				woc.rateLimiter.Wait()
+				_, err := pods.Patch(
```
NikeNano, line 100: I am probably missing something (that I would like to learn) but if we patch each pod here and then delete them as a batch, what do we gain? Why not just delete them instead? Will we reduce the API calls? Thanks!

workflow/controller/pod_cleanup.go
```diff
@@ -1,22 +1,142 @@
...
+	}
+
+	var lselector metav1.LabelSelector
+	if woc.wf.Status.Fulfilled() {
+		log.Info("workflow is fulfilled, deleting everything which may be deleted as per PodGC.Strategy.LabelSelector")
+		lselector = getLabelSelector(wfName, podGC, false)
+	} else {
+		log.Info("deleting existing pods marked as completed")
+		lselector = getLabelSelector(wfName, podGC, true)
+	}
```
NikeNano, line 81: Would it make sens to not have the else statement but instead have the else as default:
```suggestion
	lselector := getLabelSelector(wfName, podGC, true)
	if woc.wf.Status.Fulfilled() {
		log.Info("workflow is fulfilled, deleting everything which may be deleted as per PodGC.Strategy.LabelSelector")
		lselector = getLabelSelector(wfName, podGC, false)
	}
```

workflow/controller/pod_cleanup.go
```diff
@@ -1,22 +1,142 @@
...
+			woc.finishedCleanup = true
+		}
+	}
+
+	woc.rateLimiter.Wait()
+	err = pods.DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selectorString})
+
+	if err != nil {
+		log.Errorf("was not able to delete collection to due %s", err)
+	}
```
NikeNano, line 123:
```suggestion
	if err = pods.DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selectorString}); err != nil {
		log.Errorf("was not able to delete collection to due %s", err)
	}
```

workflow/controller/pod_cleanup.go
```diff
@@ -1,22 +1,142 @@
...
+
+	selectorString := selector.String()
+
+	// optimisation when function is called with a fulfilled workflow
+	// lets patch everything up
+	if woc.wf.Status.Fulfilled() {
+		markDone := true
+		podList, err := pods.List(context.Background(), metav1.ListOptions{LabelSelector: selectorString})
+		if err != nil {
```
NikeNano, line 95: Is it safe to continue execution after this error or should we retry the whole delete cycle?

NikeNano (commented):
Looks really cool, left some comments :)

workflow/controller/pod_cleanup.go
JPZ13: Thanks for the review @NikeNano! The goal is to reduce the API calls from a series of deletes in rapid succession to batch deletes every so often. We're noticing in some deployments on the major managed Kubernetes providers that workflows that spin up lots of pods that then get deleted in quick succession are being rate limited

There's a bit more context on the issue #10219 but the discussion then shifted to some of the draft PRs. We originally proposed one single batch deletion but @isubasinghe came up with this approach after community members voiced their feedback that they needed to delete pods quicker in order to avoid increased cloud spend

workflow/controller/operator.go
isubasinghe: it should be nil already but I think being more explicit is better ?

workflow/controller/pod_cleanup.go
isubasinghe: @NikeNano we actually do need to patch them, just to state that they have been completed. Argo expects this to be the case for completed workflows(all their pods are marked as completed), so I don't think we should skip this step.

workflow/controller/pod_cleanup.go
isubasinghe: addressed

workflow/controller/pod_cleanup.go
isubasinghe: Yeah we can do that, easy.

workflow/controller/pod_cleanup.go
isubasinghe: We do actually retry, but I did forget one thing, markDone = false should be stated in that branch. 
This enables the retry to happen when the next clean up call is made. 

Addressed.

workflow/controller/pod_cleanup.go
isubasinghe: addressed

workflow/controller/pod_cleanup.go
isubasinghe: Perhaps though in the future to avoid this patching even, we could have a flag that says to the controller to not expect a completed field, this way we can batch delete by just the workflow name.

workflow/controller/operator.go
NikeNano: I think this makes it harder to read since the default value is nil. But if we really want to set it explicit I suggest: 

```
var rateLimiter *waitutil.RateLimiter = nil 
```

workflow/controller/pod_cleanup.go
NikeNano: > Argo expects this to be the case for completed workflows(all their pods are marked as completed), so I don't think we should skip this step.

Good point!

workflow/controller/pod_cleanup.go
NikeNano: A nit on the use of `else` from [Effective Go
](https://go.dev/doc/effective_go):
```
In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.
```

I am quite new to Go but i seldom see the use of else(to the level that it is avoided) in other code bases, especially in the roam of err checking. For me the user here dont feel very go @isubasinghe  but it might be because I am new to it.

isubasinghe: I think else is unavoidable here unless you introduce a goto statement or `if err == nil` or repeat some code. To me, else still is the option that makes the most sense here. 

But I think that the text there just states when the if statement doesn't flow into the next statement that is the recommended approach, in this case here it does flow into the next statement in both branches, so that rule doesn't apply here.

workflow/controller/operator.go
isubasinghe: addressed

util/wait/rate_limiter_test.go
```diff
@@ -0,0 +1 @@
+package wait
```
rohankmr414, line 1: can get rid of this file if we're not adding any tests

rohankmr414 (changes requested):

rohankmr414:
qq: how will this change behave in a highly available deployment of argo-workflows (multiple replicas of controller and server)?

workflow/controller/pod_cleanup.go
JPZ13: You could factor out into a helper function if you really want to avoid the else. Something like `patchSelectedPods(ctx context.Context, labelSelector string) error`

JPZ13:
> qq: how will this change behave in a highly available deployment of argo-workflows (multiple replicas of controller and server)?

Hey @rohankmr414 - I don't think multiple replicas of the controller is suppoorted, so we shouldn't have to worry about it for now. It'll be a good thing to watch out for in the future though. Here's a docs link: https://argoproj.github.io/argo-workflows/scaling/#horizontally-scaling

workflow/controller/pod_cleanup.go
isubasinghe: Ah yes that works, addressed.

dpadhiar (approved):

workflow/controller/pod_cleanup.go
```diff
@@ -1,22 +1,147 @@
...
 	"github.com/argoproj/argo-workflows/v3/workflow/common"
 
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v3/workflow/controller/indexes"
 )
 
-func (woc *wfOperationCtx) queuePodsForCleanup() {
+func (woc *wfOperationCtx) isImmediateSkipQueues() bool {
+	return woc.rateLimiter != nil && woc.orig.Spec.PodGC != nil && woc.orig.Spec.PodGC.Strategy == wfv1.PodGCOnPodCompletion
+}
```
juliev0, line 23: I'm not sure if it's just me, but I'm a little confused at the name of the method - can you add a comment ahead of it to explain what it is?

workflow/controller/operator.go
```diff
@@ -140,6 +143,19 @@ type failedNodeStatus struct {
...
+	intervalMsStr, present := os.LookupEnv(common.EnvCleanupRateLimitIntervalMilliSeconds)
+	if present {
+		intervalMs, err := strconv.Atoi(intervalMsStr)
+		if err != nil {
+			log.Errorf("was not able to convert %s to int, defaulting to queue based cleanup, err: %s", intervalMsStr, err)
+		} else {
+			rl := waitutil.NewRateLimiter(time.Duration(intervalMs) * time.Millisecond)
+			rateLimiter = &rl
+		}
+	}
```
juliev0, line 157: the rate limiter is just being used for batch deletion, right? Can we add a comment to that effect here?

util/wait/rate_limiter.go
```diff
@@ -0,0 +1,28 @@
...
+	d.mutex.Lock()
+	canRunAt := d.last.Add(d.interval)
+	now := time.Now()
+	diff := canRunAt.Sub(now)
+	if diff > 0 {
+		time.Sleep(diff)
+	}
+	d.last = now
+	d.mutex.Unlock()
+}
```
juliev0, line 28: so, I guess the effect of this is that our pod deletion goroutines will block? does this mean there's a risk there will be no workers available some of the time?

sarabala1979:
@isubasinghe I am hard to understand the implementations. Basically, this feature is reducing K8s API for Pod deletion. Am I right?
The solution could be:
1. Ratelimit for Single pod deletion if the workflow has `OnPodCompletion`
2. Delete all workflows using the label selector if the workflow has `OnWorkflowCompletion`

RateLimit should support # of calls with duration (1s).

Can we have a quick sync call to go over the PR?

JPZ13:
Thanks for taking a look at this @sarabala1979. Based on what you said in option 1, I think quickly modifying this PR to only do the batch deletion if the workflow has `OnPodCompletion` is doable. Would that allay your concerns? To activate the batch deletions with rate limiting, the user would have to set that as the PodGC strategy along with the environment variable. Users without both set would continue unaffected

We already spoke about option 2 with our affected customer, and unfortunately solution 2 doesn't work for them because of the increased cloud cost of having lots of pods be live, especially in heavy, long-running workflows with a large number of pods

tico24:
> We already spoke about option 2 with our affected customer, and unfortunately solution 2 doesn't work for them because of the increased cloud cost of having lots of pods be live, especially in heavy, long-running workflows with a large number of pods

To add to this. 'completed' pods can indectly cost money because if the pod still exists, it's using both ip address(es) and a 'pod allocation' (my words, not the k8s term) on the node. Nodes are limited to a certain number of pods (the number depends on the node size and the cloud provider up to a kubernetes max of 256).

GCP's doc on the topic: https://cloud.google.com/kubernetes-engine/docs/how-to/flexible-pod-cidr 

So orgs with large workflows may well be being hit by having to scale in more nodes (or have larger nodes) to compensate.

util/wait/rate_limiter.go
```diff
@@ -0,0 +1,28 @@
+package wait
+
+import (
+	"sync"
+	"time"
+)
+
+type RateLimiter struct {
```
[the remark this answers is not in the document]
JPZ13: We synced with @isubasinghe and he mentioned that he didn't see where in the code base that there is a rate limiter that behaves in a similar fashion. Were you referring to [this](https://github.com/argoproj/argo-workflows/blob/master/docs/workflow-controller-configmap.yaml#L29) @alexec or something different?

workflow/controller/pod_cleanup.go
isubasinghe: Sorry about being AWOL.  I just added a comment now.

workflow/controller/operator.go
isubasinghe: Sure will do!

util/wait/rate_limiter.go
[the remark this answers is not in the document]
alexec: There is a library in go.mod for rate limiting. Don’t reinvent the wheel. Maybe this

curl -q https://raw.githubusercontent.com/kitproj/kit/main/install.sh | sh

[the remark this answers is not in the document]
JPZ13: I looked in go.mod and found two libraries that do rate limiting ([go-flowrate](https://github.com/mxk/go-flowrate) and [go-limiter](https://github.com/sethvargo/go-limiter)) but neither is actively maintained. I also found that go has a [built-in rate limiter package](https://pkg.go.dev/golang.org/x/time/rate#Limiter). Could we use that one instead? If anything, I'd suggest we sunset the other two packages and stick with the built-in throughout the codebase

PR #10418, opened 2023-01-28
fix: Only look at kubeclientset pods if needed

tooptoop4:


LinuxSuRen (approved):
LGTM

LinuxSuRen:
hi @tooptoop4 I guess you might need to keep the PR template instead of deleting it.

alexec (changes requested):
Looks good, but can you:

* Sign-off your commits to the DCO check passes.
* Add comments saying why this is a good idea.

tooptoop4:
crunchety

workflow/controller/controller.go
```diff
@@ -502,7 +502,6 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	logCtx := log.WithFields(log.Fields{"key": key, "action": action})
 	logCtx.Info("cleaning up pod")
 	err := func() error {
-		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
```
terrytangyuan, line 505: This doesn't actually look up pods, it only invokes the API. The patch/delete calls are the actual calls to K8s. I don't see any benefits this brings.

PR #10423, opened 2023-01-30
HasArtifactGC() can't access Status through execWf

juliev0:
HasArtifactGC() was running on `execWf`, which doesn't necessarily have access to the Workflow's Status in order to access `Status.StoredTemplates`, so I moved the function to where we can have access to both `execWf` for Spec and `wf` for Status. 


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -202,23 +202,6 @@ func (w *Workflow) GetExecSpec() *WorkflowSpec {
...
-	for _, template := range w.GetTemplates() {
-		for _, artifact := range template.Outputs.Artifacts {
-			if artifact.GetArtifactGC().Strategy != ArtifactGCNever && artifact.GetArtifactGC().Strategy != ArtifactGCStrategyUndefined {
-				return true
-			}
-		}
-	}
-	return false
-}
-
```
juliev0, line 221: function moved to where we can access both `execWf` for Spec and `wf` for Status

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3193,13 +3176,6 @@ func (wf *Workflow) GetTemplateByName(name string) *Template {
 	return nil
 }
 
-func (w *Workflow) GetTemplates() []Template {
-	return append(
-		w.GetExecSpec().Templates,
-		w.Status.GetStoredTemplates()...,
-	)
-}
-
```
juliev0, line 3202: this function was only ever called by `HasArtifactGC()`. I also noticed that `w.GetExecSpec().Templates` and `w.Status.GetStoredTemplates()` were not mutually exclusive so appending the two together might be misleading if somebody tries to use this function in the future

pkg/apis/workflow/v1alpha1/workflow_types_test.go
```diff
@@ -117,96 +117,6 @@ func TestWorkflowHappenedBetween(t *testing.T) {
...
-			wf := MustUnmarshalWorkflow(workflowSpec)
-
-			hasArtifact := wf.HasArtifactGC()
-
-			assert.Equal(t, hasArtifact, tt.expectedResult)
-		})
-	}
-
-}
-
```
juliev0, line 209: unit test moved since the underlying function was moved

test/e2e/artifacts_test.go
```diff
@@ -90,46 +93,64 @@ func (s *ArtifactsSuite) TestArtifactGC() {
...
 		},
+		// entire Workflow based on a WorkflowTemplate which has a Step that references another WorkflowTemplate's template
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-from-ref-template.yaml",
-			expectedGCPodsOnWFCompletion: 0,
+			hasGC:                        true,
+			expectedGCPodsOnWFCompletion: 1,
 			expectedArtifacts: []artifactState{
-				artifactState{"on-completion", "my-bucket-2", false, true},
+				artifactState{"on-completion", "my-bucket-2", true, false},
```
juliev0, line 142: the expectation of this test was wrong previously - the test assumed that in the case of a Workflow defining strategy A and the WorkflowTemplate it's based on defining an artifact with Strategy B, that the net result should be A, but it makes more sense for it to be B (artifact level always overrides workflow level).

workflow/controller/artifact_gc.go
```diff
@@ -71,6 +71,34 @@ func (woc *wfOperationCtx) garbageCollectArtifacts(ctx context.Context) error {
...
+			strategy := woc.execWf.GetArtifactGCStrategy(&artifact)
+			if strategy != wfv1.ArtifactGCStrategyUndefined && strategy != wfv1.ArtifactGCNever {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
```
juliev0, line 101: function moved from workflow_types.go to here where we can access both execWf for Spec and wf for Status

workflow/controller/artifact_gc_test.go
```diff
@@ -609,3 +609,106 @@ func TestProcessCompletedWorkflowArtifactGCTask(t *testing.T) {
...
+			defer cancel()
+			woc := newWorkflowOperationCtx(wf, controller)
+
+			hasArtifact := woc.HasArtifactGC()
+
+			assert.Equal(t, hasArtifact, tt.expectedResult)
+		})
+	}
+
+}
```
juliev0, line 714: unit test moved here

test/e2e/artifacts_test.go
juliev0: (the only reason the test was operating the other way before was due to our not successfully accessing the `Status.StoredTemplates`)

sarabala1979 (approved):
LGTM

PR #10424, opened 2023-01-30
fix: delete PVCs upon onExit error when OnWorkflowCompletion is enabled. Fixes #10408

### LINKED ISSUE #10408: volumeClaimGC OnWorkflowCompletion does not work when onExit/entrypoint template errors
### Pre-requisites

Using `OnWorkflowCompletion` pvc cleanup strategy:

```yaml
  volumeClaimGC:
```

The PVC will not be deleted if there is a template error for the `onExit` handler (and possibly other failure scenarios with onExit). 

```
$ k get pvc
NAME                STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
wf-gz95t-builddir   Pending                                      local-path     10s

$ argo list
NAME       STATUS   AGE   DURATION   PRIORITY   MESSAGE
wf-gz95t   Error    19s   10s        0          error in exit template execution : template 'notify' type is unknown
```

The key parts about this issue seems to be a combination of:
* using WorkflowTemplates
* an invalid onExit template 

When I convert the WorkflowTemplate to a Workflow, the PVC clean up happens expectedly.

### Version

latest as of 1/27

```YAML
kind: WorkflowTemplate
metadata:
  name: wft-pvc
spec:
  volumeClaimTemplates:
    - metadata:
        name: builddir
      spec:
        accessModes: [ "ReadWriteMany" ]
        resources:
          requests:
            storage: 1Mi

  volumeClaimGC:

  entrypoint: whalesay
  onExit: notify

  templates:
  - name: whalesay
    container:
      image: docker/whalesay:latest
      command: [cowsay]
      args: ["hello world"]

  - name: notify
    # note this is a bad template

---
kind: Workflow
metadata:
  generateName: wf-
spec:
  workflowTemplateRef:
    name: wft-pvc
```

Note that this seems to be a combination of WorkflowTemplates and onExit

```text
$ kubectl logs -n argo deploy/workflow-controller | grep wf-gz95t
time="2023-01-28T01:49:06.711Z" level=info msg="Processing workflow" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.737Z" level=info msg="Updated phase  -> Running" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.737Z" level=info msg="Creating pvc wf-gz95t-builddir" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.745Z" level=info msg="Pod node wf-gz95t initialized Pending" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.758Z" level=info msg="Created pod: wf-gz95t (wf-gz95t)" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.758Z" level=info msg="TaskSet Reconciliation" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.758Z" level=info msg=reconcileAgentPod namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:06.764Z" level=info msg="Workflow update successful" namespace=argo phase=Running resourceVersion=406075 workflow=wf-gz95t
time="2023-01-28T01:49:16.712Z" level=info msg="Processing workflow" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Task-result reconciliation" namespace=argo numObjs=1 workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="task-result changed" namespace=argo nodeID=wf-gz95t workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="node changed" namespace=argo new.message= new.phase=Succeeded new.progress=0/1 nodeID=wf-gz95t old.message= old.phase=Pending old.progress=0/1 workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="TaskSet Reconciliation" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg=reconcileAgentPod namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Running OnExit handler: notify" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Skipped node wf-gz95t-309339311 initialized Error (message: template 'notify' type is unknown)" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Updated phase Running -> Error" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Updated message  -> error in exit template execution : template 'notify' type is unknown" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Marking workflow completed" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Marking workflow as pending archiving" namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.713Z" level=info msg="Checking daemoned children of " namespace=argo workflow=wf-gz95t
time="2023-01-28T01:49:16.722Z" level=info msg="cleaning up pod" action=deletePod key=argo/wf-gz95t-1340600742-agent/deletePod
time="2023-01-28T01:49:16.724Z" level=info msg="Workflow update successful" namespace=argo phase=Error resourceVersion=406101 workflow=wf-gz95t
time="2023-01-28T01:49:16.729Z" level=info msg="archiving workflow" namespace=argo uid=74611389-a1da-4b8a-9c96-947241c5afe8 workflow=wf-gz95t
time="2023-01-28T01:49:16.734Z" level=info msg="cleaning up pod" action=labelPodCompleted key=argo/wf-gz95t/labelPodCompleted
```

```text
N/A
```

jiachengxu:
Signed-off-by: Jiacheng Xu <xjcmaxwellcjx@gmail.com>

Fixes #10408
This PR enables PVC removal upon onExit error when `OnWorkflowCompletion` is enabled so that PVCs provisioned by the workflow will be GCed.
An e2e helper function and test are added as well.


workflow/controller/operator.go
```diff
@@ -421,9 +421,12 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 			default:
 				if !errorsutil.IsTransientErr(err) && !woc.wf.Status.Phase.Completed() && os.Getenv("BUBBLE_ENTRY_TEMPLATE_ERR") != "false" {
 					woc.markWorkflowError(ctx, x)
-					return
 				}
 			}
+			err = woc.deletePVCs(ctx)
+			if err != nil {
+				woc.log.WithError(err).Warn("failed to delete PVCs")
+			}
```
jessesuen, line 429: Do you know why this failed when using `workflowTemplateRef` but not when the whole workflow was inlined? I'm wondering if there is a deeper problem to be fixed rather than adding another deletePVC call

jiachengxu: Yeah, so I think the root issue is that before the controller creates PVCs, when it does validation for `Workflow`s, there are 2 cases:
- If the `Workflow` doesn't refer to any `WorkflowTemplate` and it is inlined, controller will validate the workflow itself (inlined workflow), this is fine since the bad onExit templates will not pass the validation, and PVC will not even be created.
- If the `Workflow` is referred to a `WorkflowTemplate`, the controller will also only validate the workflow, but not the templates from `WorkflowTemplate` spec. (https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/operator.go#L3589). In this case, PVC will be created, and never be deleted. That's why we have different behavior.

I will open a PR to fix the above issue. However, this  deletePVC call is still needed since it could be other errors when onExit executes(not only because onExit bad templates).

jessesuen: > I will open a PR to fix the above issue since I feel like it is an enhancement and bugfix for workflow validation

Yes, I agree, seems with using `workflowTemplateRef`, we do less validation than when they are inlined.

> Also, this deletePVC call is still needed since it could be other errors when onExit executes(not only because onExit bad templates).

I originally had the same thought as you and felt that this was a bug with cleaning PVCs when onExit has errors. However, when I tried the following workflow, it creates and then properly deletes the PVC when onExit fails.

```yaml
kind: Workflow
metadata:
  generateName: wft-pvc-
spec:
  volumeClaimTemplates:
    - metadata:
        name: builddir
      spec:
        accessModes: [ "ReadWriteMany" ]
        resources:
          requests:
            storage: 1Mi

  volumeClaimGC:

  entrypoint: whalesay
  onExit: notify

  templates:
  - name: whalesay
    container:
      image: docker/whalesay:latest
      command: [cowsay]
      args: ["hello world"]

  - name: notify
    container:
      image: docker/whalesay:latest
      command: [sh, -c]
      args: ["exit 1"]
```

This tells me that probably onExit PVC cleanup is working properly, and we only need the extra validation that is missing when using `workflowTemplateRef`. In other words, I think this PR may not be necessary if we fix the validation.

jiachengxu: Yeah, the above workflow works since the `Status` of the workflow is `Failed` not `Error`, I am trying to construct a workflow that will cause an error here and return here(this is the same place that the workflow that uses `workflowTemplateRef` will return): https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/operator.go#L427, and I don't know how I can do that. Maybe you have an idea? If we can construct a workflow return here, then I guess the PVCs will not be deleted?

jiachengxu: I have added the workflow validating logic in this commit: https://github.com/argoproj/argo-workflows/pull/10424/commits/d18db259ff617909ca1ff24bfb72e994024befc0, and I have tested without the `deletePVC` call added by this PR, and it works, and now the workflow will fail to pass the validation, and PVCs will not be created.
If you think what I mentioned in the last comment is not an issue, I will just remove the `deletePVC` call and change the PR title and description.

test/e2e/fixtures/then.go
```diff
@@ -185,6 +186,38 @@ func (t *Then) ExpectAuditEvents(filter func(event apiv1.Event) bool, num int, b
 	return t
 }
 
+func (t *Then) ExpectPVCDeleted() *Then {
+	t.t.Helper()
+	timeout := defaultTimeout
+	_, _ = fmt.Println("Checking", timeout.String(), "for expecting PVCs deletion")
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
NikeNano, line 193: would it make sense to use [`time.After`](https://pkg.go.dev/time#After) instead since the context is not really used here?

workflow/validate/validate.go
```diff
@@ -241,10 +259,21 @@ func ValidateWorkflow(wftmplGetter templateresolution.WorkflowTemplateNamespaced
...
+			}
+		}
+	} else {
+		for _, template := range wf.Spec.Templates {
+			_, err := ctx.validateTemplateHolder(&wfv1.WorkflowStep{Template: template.Name}, tmplCtx, &FakeArguments{}, opts.WorkflowTemplateValidation)
+			if err != nil {
+				return errors.Errorf(errors.CodeBadRequest, "templates.%s %s", template.Name, err.Error())
+			}
 		}
 	}
```
NikeNano, line 278: Would it make sens to remove the `else` and do an early return instead?

```suggestion
	if hasWorkflowTemplateRef {
		for _, template := range wfSpecHolder.GetWorkflowSpec().Templates {
			_, err := ctx.validateTemplateHolder(&wfv1.WorkflowStep{TemplateRef: wf.Spec.WorkflowTemplateRef.ToTemplateRef(template.Name)}, tmplCtx, &FakeArguments{}, opts.WorkflowTemplateValidation)
			if err != nil {
				return errors.Errorf(errors.CodeBadRequest, "templates.%s %s", template.Name, err.Error())
			}
		}
	return nil 
	} 
	for _, template := range wf.Spec.Templates {
			_, err := ctx.validateTemplateHolder(&wfv1.WorkflowStep{Template: template.Name}, tmplCtx, &FakeArguments{}, opts.WorkflowTemplateValidation)
			if err != nil {
				return errors.Errorf(errors.CodeBadRequest, "templates.%s %s", template.Name, err.Error())
			}
		}
```

workflow/validate/validate.go
```diff
@@ -225,9 +225,27 @@ func ValidateWorkflow(wftmplGetter templateresolution.WorkflowTemplateNamespaced
...
+	// If the OnExit is empty in Workflow:
+	//   - If OnExit is specified in referred WorkflowTemplate, the OnExit of the referred WorkflowTemplate will be validated.
+	//   - If OnExit is empty in referred WorkflowTemplate, nothing will be validated.
+	var tmplHolder *wfv1.WorkflowStep
 	if wf.Spec.OnExit != "" {
+		if hasWorkflowTemplateRef {
+			tmplHolder = &wfv1.WorkflowStep{TemplateRef: wf.Spec.WorkflowTemplateRef.ToTemplateRef(wf.Spec.OnExit)}
+		} else {
+			tmplHolder = &wfv1.WorkflowStep{Template: wf.Spec.OnExit}
+		}
```
NikeNano, line 242: Would it make sens to set a default and skip the else?
```suggestion
		tmplHolder = &wfv1.WorkflowStep{Template: wf.Spec.OnExit}
		if hasWorkflowTemplateRef {
			tmplHolder = &wfv1.WorkflowStep{TemplateRef: wf.Spec.WorkflowTemplateRef.ToTemplateRef(wf.Spec.OnExit)}
		}
```

workflow/controller/operator.go
```diff
@@ -357,6 +357,11 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 				woc.markWorkflowError(ctx, x)
 			}
 		}
+		// Garbage collect PVCs if Entrypoint template execution returns error
+		err = woc.deletePVCs(ctx)
+		if err != nil {
+			woc.log.WithError(err).Warn("failed to delete PVCs")
+		}
```
NikeNano, line 364:
```suggestion
		if err := woc.deletePVCs(ctx); err != nil {
			woc.log.WithError(err).Warn("failed to delete PVCs")
		}
```

workflow/validate/validate.go
NikeNano: hmm formatting got screwed but I hope you get the idea :)

test/e2e/fixtures/then.go
jiachengxu: Sorry I don't get it. The context is used in the following for loop for pvcClient to get PVCs: https://github.com/argoproj/argo-workflows/pull/10424/files#diff-75dd759badaa0fcc320bb8cac2c095d3585bb87b257fe4fe967279d6b21cf830R204

workflow/validate/validate.go
jiachengxu: Done

workflow/validate/validate.go
jiachengxu: Done. Thank you!

workflow/controller/operator.go
jiachengxu: Done. Thank you!

test/e2e/fixtures/then.go
```diff
@@ -185,6 +186,38 @@ func (t *Then) ExpectAuditEvents(filter func(event apiv1.Event) bool, num int, b
...
+	defer cancel()
+	for {
+		select {
+		case <-ctx.Done():
+			t.t.Errorf("timeout after %v waiting for condition", timeout)
+			return t
```
NikeNano, line 199:
```suggestion
	_, _ = fmt.Println("Checking", timeout.String(), "for expecting PVCs deletion")
	for {
		select {
		case <-time.After(defaultTimeout)::
			t.t.Errorf("timeout after %v waiting for condition", timeout)
			return t
```

test/e2e/fixtures/then.go
NikeNano: My bad I missed the use of `ctx` and you only used the `context.WithTimeout` to have a timeout.

I though that we could replace: 

```
case <-ctx.Done():
```

with 

```
case <-time.After(defaultTimeout):
```

and remove the creation but missed the use of `ctx`.

jessesuen (approved):

PR #10457, opened 2023-02-03
fix: use env when pod version annotation is missing. Fixes #10237

### LINKED ISSUE #10237: Inconsistent pod name with parallelization enabled
### Pre-requisites

I am noticing some inconsistent behavior for the pod names calculation when using parallelization with v1 POD_NAME spec. I have parallelization enabled to run 5 max workflows at the same time. I submit 20 workflows at the same time, the first 5 calculate pod names for all tasks without the template name(according to v1 spec), after that every workflow that was stuck in the Pending state appends template name to the pod name as well(v2 spec). Argo UI doesn't show this, but when you monitor the pods using kubectl, you can see the name of the pods being created after the initial 5 workflows will have template names appended to it(v2 spec)

### Version

3.4.0

```YAML
kind: Workflow
metadata:
  generateName: steps-
spec:
  entrypoint: hello-hello-hello

  templates:
  - name: hello-hello-hello
    steps:
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello1"
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello2a"
      - name: hello2b           # single dash => run in parallel with previous step
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello2b"

  - name: whalesay
    inputs:
      parameters:
      - name: message
    container:
      image: docker/whalesay
      command: [cowsay]
```

```text
```

```text
```

isubasinghe:

Fixes #10237

* [] Add unit or e2e tests. Say how you tested your changes. If you changed the UI, attach screenshots.


workflow/util/pod_name.go
```diff
@@ -79,8 +79,10 @@ func ensurePodNamePrefixLength(prefix string) string {
 // given workflow
 func GetWorkflowPodNameVersion(wf *v1alpha1.Workflow) PodNameVersion {
 	annotations := wf.GetAnnotations()
-	version := annotations[common.AnnotationKeyPodNameVersion]
-
+	version, ok := annotations[common.AnnotationKeyPodNameVersion]
```
NikeNano, line 82: Is the `default` in the switch case below still valid?  Otherwise this could potentially be rewritten so that `GetPodNameVersion()` is the `default` case?

isubasinghe: I was tempted to implement this fix that way but that default switch case will be hit if the annotation is invalid. That is a different behaviour to the GetPodNameVersion call. 

But yeah I am on the fence on this about what the "right behaviour" is.  This is why I went with the current approach, at least it will always be consistent when the user makes a wrong annotation (default to v2).

NikeNano: Ok, sounds good!

NikeNano (approved):
LGTM

alexec (approved):

PR #10474, opened 2023-02-06
chore: 'Enforcing history limit' log to Debug level

tooptoop4:
Currently this is being logged every 10 seconds per different cron workflow. Can be costly to log aggregators if having many cronWorkflows


alexec (approved):

alexec (changes requested):
Please sign-off your commits.

tooptoop4:
crispy, these logs costing $70 a day!

terrytangyuan (commented):
Please fix DCO check

PR #10493, opened 2023-02-09
fix: evaluated debug env vars value

GoshaDo:
Useful when passing debug flag to a pipeline as a parameter. 

Fixes #evulated debug env vars value

* [X] Github checks are green.
* [X] Once required tests have passed, mark your PR "Ready for review".


alexec (changes requested):
Thank you for your PR.

* I'm not clear what issue you're trying to fix. Could you please like to it from the PR title?
* I've suggested an simplification.

GoshaDo:
> Thank you for your PR.
> 
> * I'm not clear what issue you're trying to fix. Could you please like to it from the PR title?
> * I've suggested an simplification.

I haven't open an issue for that one, but it enables to set the env var and check the value of it.
The problem with you suggested fix is that the value "True" or "TRUE" not evaluated. also if anything that non boolean value injected, it's false.. that not bad, but don't covers all the edge cases

alexec:
I see. I don't think that we should support variations of "true" here when we do not do so elsewhere. That would result in inconsistent and unexpected behaviour.

GoshaDo:
@alexec ok, so lets leave it like that.

alexec (approved):

Issue #10490, opened 2023-02-08
Wait container failing: 'level=info msg="stopping progress monitor (context done)" error="context canceled"'


lvijnck:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I am trying to implement a mechanism to clone a repository and pass it downstream. This has been failing due to the "unable to resolve delta" that was happening in 3.4.4. I just updated to argo v3.4.5.

I am now getting errors on Argo's "wait" container. I performed several tests and concluded the following:

- The repo is succesfully cloned, as the status is correctly displayed
- Error does not happen when I omit the output artifact
- Error does _not_ happen for public repositories (even with output artifact defined)

### Version

v.3.4.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: test-clone-repo
spec:
  entrypoint: run
  templates:
    - name: run
      inputs:
        parameters:
          artifacts:
          - name: repo-artifact
            path: /src
            git:
              repo: "https://gitlab.com/url"
              revision: "main"
              usernameSecret:
                name: argo-git-creds
                key: username
              passwordSecret:
                name: argo-git-creds
                key: token
      container:
        image: "python:3.9-buster"
        command: [ sh, -c ]
        args: [ "git status && ls && pwd" ]
        workingDir: /src
      outputs:
        artifacts:
          - name: repo-artifact
            path: /src
```

### Logs from the workflow controller

```text
time="2023-02-08T19:14:09.342Z" level=info msg="Processing workflow" namespace=fix-arg-workflows workflow=test-clone-repo-drqpb                                                                                                  │
│ time="2023-02-08T19:14:09.342Z" level=info msg="Task-result reconciliation" namespace=fix-arg-workflows numObjs=0 workflow=test-clone-repo-drqpb                                                                                 │
[...]

PR #10530, opened 2023-02-14
fix: added logs related to executing commands in the container

sxllwx:
Signed-off-by: scott <scottwangsxll@gmail.com>

Fixes：

In the process of troubleshooting #10490, we found that our `wait` received more than one sigterm.

However, when wf-controller sends sigterm for the second time (actually through argoexec kill), it may cause container-crash (signal processing is asynchronous), and we need this part of the log to support our troubleshooting.

### DIFF
--- a/workflow/common/util.go
+++ b/workflow/common/util.go
@@ -60,7 +60,16 @@ func (d *WebsocketRoundTripper) RoundTrip(r *http.Request) (*http.Response, erro
 }
 
 // ExecPodContainer runs a command in a container in a pod and returns the remotecommand.Executor
-func ExecPodContainer(restConfig *rest.Config, namespace string, pod string, container string, stdout bool, stderr bool, command ...string) (remotecommand.Executor, error) {
+func ExecPodContainer(restConfig *rest.Config, namespace string, pod string, container string, stdout bool, stderr bool, command ...string) (exec remotecommand.Executor, err error) {
+	defer func() {
+		log.WithField("namespace", namespace).
+			WithField("pod", pod).
+			WithField("container", container).
+			WithField("command", command).
+			WithError(err).
+			Debug("exec container command")
+	}()
+
 	clientset, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
 		return nil, errors.InternalWrapError(err)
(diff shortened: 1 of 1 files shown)


tooptoop4 (changes requested):
Use Debug level instead of Info level (already the amount of logs are expensive). Could the command contain sensitive text (ie result of secret/pw)?

workflow/signal/signal.go
```diff
@@ -45,6 +45,13 @@ func SignalContainer(restConfig *rest.Config, pod *corev1.Pod, container string,
 
 func ExecPodContainerAndGetOutput(restConfig *rest.Config, namespace string, pod string, container string, command ...string) error {
 	x, err := common.ExecPodContainer(restConfig, namespace, pod, container, true, true, command...)
+	log.
+		WithField("namespace", namespace).
+		WithField("pod", pod).
+		WithField("container", container).
+		WithField("command", command).
+		WithError(err).
+		Info("exec command")
```
alexec, line 54:
```suggestion
		Debug("exec command")
```

alexec (changes requested):
Do we need this if #10523 works?

workflow/signal/signal.go
sxllwx: > Do we need this if #10523 works?

I think it is needed, it can provide a complete error context.

sxllwx: fixed, PTAL. 😄

workflow/signal/signal.go
```diff
@@ -45,6 +45,13 @@ func SignalContainer(restConfig *rest.Config, pod *corev1.Pod, container string,
 
 func ExecPodContainerAndGetOutput(restConfig *rest.Config, namespace string, pod string, container string, command ...string) error {
 	x, err := common.ExecPodContainer(restConfig, namespace, pod, container, true, true, command...)
+	log.
```
alexec, line 48: do you want to move logging into `ExecPodContainer`?

sxllwx: ok~, PTAL, thx

alexec (approved):

PR #10532, opened 2023-02-14
feat: Enable Codespaces with `kit`

alexec:
To try out:

image

Codespaces will start.
When it is ready, run `kit up` in the terminal.
Wait.
Open "ports" tab.
Click on the globe symbol on the webpack row.


tasks.yaml
```diff
@@ -0,0 +1,89 @@
+metadata:
+  annotations:
+    help: |
+      Install `kit` by following https://github.com/alexec/kit#install.
+      
+      Run `kit up` to start argo.
```
NikeNano, line 6: Should we include something in the docs about how to run this as well? Looks really cool.

alexec: You just run `kit up` :)

NikeNano: But how do new users of Argo find out about this? Do we expect them to do read the `task.yaml` file?

alexec: I think we should trial. It is intended to work with Codespaces.

NikeNano (approved):
LGTM

isubasinghe (approved):

sarabala1979 (approved):

PR #10559, opened 2023-02-21
fix: Resolve issues with offline linter + add tests

julienduchesne:
https://github.com/argoproj/argo-workflows/pull/10059 was tested in very specific cases it seems (sorry about that 😬), because a couple things are wrong:
- Linting directories does not work anymore
- Resolving references in a namespace which doesn't show up in the given args, panics

In this PR, I fix those issues but I also add functional tests that checks that all of these cases work from the cmd level 

I did not add to the `argo/lint/lint_test.go` tests and instead put the tests earlier down the line because the current tests do not cover the creation and configuration of the offline client (which is the part that is failing here, the actual linter was not modified in the PR above)

### DIFF
--- a/pkg/apiclient/offline-client.go
+++ b/pkg/apiclient/offline-client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/argoproj/argo-workflows/v3/pkg/apiclient/clusterworkflowtemplate"
 	"github.com/argoproj/argo-workflows/v3/pkg/apiclient/cronworkflow"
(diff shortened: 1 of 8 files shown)


pkg/apiclient/offline-client.go
```diff
@@ -4,76 +4,106 @@ import (
...
+		err := filepath.Walk(basePath, func(path string, info os.FileInfo, err error) error {
+			if info.IsDir() {
+				return nil
 			}
-
-			if _, ok := clusterWorkflowTemplateGetter.clusterWorkflowTemplates[cwftmpl.Name]; ok {
-				return nil, nil, fmt.Errorf("duplicate ClusterWorkflowTemplate found: %q", cwftmpl.Name)
+			if err != nil {
+				return err
 			}
```
julienduchesne, line 59: Reviewing without whitespace yields a nicer result here. The only real change is nestic the logic in a `filepath.Walk`

alexec (approved):

julienduchesne:
Thanks for merging @alexec. 

FYI, I have now been able to test this internally (we use a combination of CronWorkflows, WorkflowTemplates and ClusterWorkflowTemplates, with refs between each other) and have not seen any errors. We will start using it in our CI pipeline very soon.

PR #10569, opened 2023-02-22
chore: Upgrade Go to v1.19

terrytangyuan:
v1.18 has reached end of life: https://endoflife.date/go


alexec (changes requested):
Please update the Dockerfiles.

Please dismis this review when done.

alexec (approved):

PR #10576, opened 2023-02-24
fix: panic in offline linter + handling stdin

julienduchesne:
Follow-up to https://github.com/argoproj/argo-workflows/pull/10059 and https://github.com/argoproj/argo-workflows/pull/10559

The offline linter is working well in ideal scenarios, I am linting a set of files/directory in 5 seconds where it used to take ~5 minutes

However, I did find issues after testing some more:
Given an non-existent file, it panics (I inverted the `info.IsDir()` statement and the one checking for errors) 
There's a disconnect between the `filepath.Walk` in the linter and the one building the offline client which means that there were other issues (such as stdin not working)

To fix those issues once and for all, I extracted the walk func to a common path (`file.WalkManifests`); that way, we are sure that changes are always reflected in both locations

I also added some tests for non-existent files as well as offline linting from stdin

### DIFF
--- a/cmd/argo/commands/client/conn_test.go
+++ b/cmd/argo/commands/client/conn_test.go
@@ -1,9 +1,11 @@
 package client
 
 import (
+	"context"
 	"os"
 	"testing"
 
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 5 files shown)


alexec (approved):

tooptoop4:
@julienduchesne is there an argo binary anywhere with this fix? or do I need to compile the binary?

PR #10599, opened 2023-02-27
build: Enhance Codespaces

alexec:
This PR:

- Makes `make argoexec-image` run in just a few seconds when rebuilding from cache. The `.git` folder gets modified every time you run git. This mean that the Docker caches being invalidated. `GIT_COMMIT` etc must be passed as Docker build .args
- Codespaces is not very polished. E.g. the "universal" image is not "universal" because it does not work on Apple Silicon.
- I've pinned Node to 16. While both 16 and 18 are LTS, 18 has the legacy SSL issue, to which there does not appear to be any simple fix.

If pre-built, you can start a Codespace instantly. When not pre-built, it takes several minutes.


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,18 @@
...
+      "version": "16"
+    },
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {}
+  },
+  "hostRequirements": {
+    "cpus": 4
+  },
+  "onCreateCommand": ".devcontainer/pre-build.sh",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/home/vscode/go/src/github.com/argoproj/argo-workflows,type=bind",
+  "workspaceFolder": "/home/vscode/go/src/github.com/argoproj/argo-workflows"
```
tczhao, line 17: Could we keep using the `${localEnv:HOME}` instead of `/home/vscode` so it still works on local devcontainer

tczhao (changes requested):
Perhaps we need to add some info about the `kit up` in 
https://github.com/tczhao/argo/blob/master/docs/running-locally.md

terrytangyuan (approved):
LGTM. Please take a look at the comment above.

.devcontainer/devcontainer.json
alexec: Let me test. I was fixing something by doing this.

sarabala1979 (approved):

alexec:
I'll fix vscode problem if needed later.

PR #10639, opened 2023-03-04
fix: ensure children containers are killed for container sets. Fixes #10491

### LINKED ISSUE #10491: Workflow hangs after clicking stop and terminate
### Pre-requisites

When clicking stop or terminate within the UI, we are noticing that workflows are left hanging in a running state. The only way to get them to actually stop is to delete the workflow. This causes a problem for users who want to stop the workflow but keep the record in their archives

Here's a list of some other issues that I think might be related. I will keep this list up to date as more arise:
- https://github.com/argoproj/argo-workflows/issues/10063
- https://github.com/argoproj/argo-workflows/issues/8680
- https://github.com/argoproj/argo-workflows/issues/9934

### Version

latest

```YAML
Unfortunately, this happens intermittently. I will update this issue with some videos and see if I can find some workflows that consistently trigger it using only public images
```

```text
see above
```

```text
See above
```

isubasinghe:
Fixes #10491

### DIFF
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -1735,6 +1735,30 @@ func (s Nodes) Children(parentNodeId string) Nodes {
 	return childNodes
 }
 
+func (s Nodes) NestedChildrenStatus(parentNodeId string) ([]NodeStatus, error) {
+	parentNode, ok := s[parentNodeId]
+	if !ok {
+		return nil, fmt.Errorf("could not find %s in nodes when searching for nested children", parentNodeId)
+	}
+
+	children := []NodeStatus{}
+	q := []NodeStatus{parentNode}
+
+	for {
+		if len(q) <= 0 {
+			break
+		}
+		childNode := q[0]
+		q = q[1:]
+		for _, nodeID := range childNode.Children {
+			q = append(q, s[nodeID])
+		}
+		children = append(children, childNode)
+	}
+
+	return children, nil
+}
+
 // Filter returns the subset of the nodes that match the predicate, e.g. only failed nodes
 func (s Nodes) Filter(predicate func(NodeStatus) bool) Nodes {
 	filteredNodes := make(Nodes)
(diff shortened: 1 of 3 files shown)


workflow/controller/exec_control.go
```diff
@@ -82,10 +82,23 @@ func (woc *wfOperationCtx) handleExecutionControlError(nodeID string, wfNodesLoc
 	node := woc.wf.Status.Nodes[nodeID]
 	woc.markNodePhase(node.Name, wfv1.NodeFailed, errorMsg)
 
+	children := []wfv1.NodeStatus{}
```
alexec, line 85: This looks like a useful utility function,:

- Is there already a similar function?
- If not, extract it, e.g. `woc.wf.Status.Nodes.GetTransientChildren(nodeID)`

Extracting it to a function will:

- Allow reuse.
- Write test.

alexec (changes requested):
Please dismis this review to request re-review.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1735,6 +1735,30 @@ func (s Nodes) Children(parentNodeId string) Nodes {
 	return childNodes
 }
 
+func (s Nodes) NestedChildrenStatus(parentNodeId string) ([]NodeStatus, error) {
```
alexec, line 1738: now, please add tests!

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1735,6 +1735,30 @@ func (s Nodes) Children(parentNodeId string) Nodes {
...
+	q := []NodeStatus{parentNode}
+
+	for {
+		if len(q) <= 0 {
+			break
+		}
+		childNode := q[0]
+		q = q[1:]
+		for _, nodeID := range childNode.Children {
+			q = append(q, s[nodeID])
```
alexec, line 1754: do we need to go down the tree?

alexec (approved):

sarabala1979 (approved):

Issue #10447, opened 2023-02-01
Customizable Argo UI (navigation links, custom columns)


jessesuen:
# Summary

Many users use Argo Workflows for their CI or other application purposes (AI/ML). However, the UI does not lend itself to any specific application use case. It is a generic UI that allows you to do very generic things: filter on labels, namespace, etc... However, end users who might be visiting the Argo Workflows UI may not know where to start or how to use the UI to get to the filtered or curated view they are looking for.

This enhancement proposal is to allow **_workflow administrators_** to customize the UI in a couple of ways:
* providing tabs/links at the top of the interface which makes it obvious for their own users to navigate to curated/filtered views
* allowing the workflow administrator to customize the **_columns_** shown on the list view so that users can discern different workflow categories, e.g. by labels

# Use Cases

When would you use this?

For someone using workflows as CI, they could create various navigation links with predefined filters for their users, e.g.
* `Guestbook Builds`
* `Sockshop Builds`

These links would be predefined with filters based on labels that they choose in their environment. 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

jessesuen:
Spoke with @alexec yesterday and he was supportive of the feature. @rbreeze and @jiachengxu will find time to work on this in the coming weeks.

jessesuen:
@jiachengxu - can you comment on this issue so I can add you as an assignee?

jiachengxu:
@jessesuen thank you! Looking forward to working on it!

rbreeze:
Proposed CM settings structure:

```yaml
# These will be exposed as buttons in the top bar of the UI
customLinks:
  - url: http://localhost:8080/workflows?label=my.custom.label%foo
    name: My Custom Label

# These will be exposed as columns in the Workflow List View
exposeColumns:
  - key: my.custom.label
    name: Custom Label
```

alexec:
There's already a feature for links you could extend. I think you need to write something new for the columns.

I wonder if we should bundle predefined/default views?

jiachengxu:
> There's already a feature for links you could extend. I think you need to write something new for the columns.
> 
> I wonder if we should bundle predefined/default views?

I saw we have links support: https://argoproj.github.io/argo-workflows/links/, I will try to reuse or extend this. Thank you!

jessesuen:
Would this be possible (so that domain does not need to be configured in each link)?

```
customLinks:
[...]

PR #10677, opened 2023-03-13
feat: extend links feature for custom workflow views

jiachengxu:
This PR extends the current links feature(https://argoproj.github.io/argo-workflows/links/), with new scope `workflow-list`, which will be used as buttons on the top of the workflow view, and users can use them to navigate to filtered/curated view of workflows. 
This PR only contains the necessary on the backend and @rbreeze will help with the UI support.
Related #10447


pkg/apis/workflow/v1alpha1/info.go
```diff
@@ -6,7 +6,7 @@ package v1alpha1
 type Link struct {
 	// The name of the link, E.g. "Workflow Logs" or "Pod Logs"
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
-	// "workflow", "pod", "pod-logs", "event-source-logs", "sensor-logs" or "chat"
+	// "workflow", "pod", "pod-logs", "event-source-logs", "sensor-logs", "custom-buttons" or "chat"
```
terrytangyuan, line 9: Is this used anywhere?

jiachengxu: Thank you for the review @terrytangyuan!
This `scope`  will only be used by the UI code to create buttons on the workflow view, and it is not used by anywhere for now.
@rbreeze will create UI based on the scope `custom-buttons` in the next UI PRs, and this PR only make the necessary backend change

alexec:
”custom buttons” is rather generic. What about “workflow list”?

jiachengxu:
> ”custom buttons” is rather generic. What about “workflow list”?

I think that makes more sense, and I just updated the PR, thanks!

pkg/apis/workflow/v1alpha1/info.go
terrytangyuan: But the docs and manifests are also updated, which might mislead users. Does it make more sense to work on a single PR? @alexec Any thoughts?

jiachengxu: Personally, I feel like the `main` branch can always contain un-released or incompleted features, and most users use releases and this shouldn't affect them much. But if you think that is an issue, I can also remove the user-facing change(e.g., docs, manifests) in this PR and add them back once the feature is completed.

rbreeze: I agree with @jiachengxu - I think it's okay to get this merged first given that it's such a small change to the docs

rbreeze (approved):
LGTM!

terrytangyuan (approved):

pkg/apis/workflow/v1alpha1/info.go
terrytangyuan: SGTM

PR #10681, opened 2023-03-14
fix(agent): no more requeue when the node succeeded

sxllwx:
Fixes: As the title

### DIFF
--- a/workflow/executor/agent.go
+++ b/workflow/executor/agent.go
@@ -368,6 +368,9 @@ func (ae *AgentExecutor) executePluginTemplate(ctx context.Context, tmpl wfv1.Te
 			return 0, err
 		} else if reply.Node != nil {
 			*result = *reply.Node
+			if reply.Node.Phase == wfv1.NodeSucceeded {
+				return 0, nil
+			}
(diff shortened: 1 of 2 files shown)


alexec (commented):
Thank you for this fix. How would you go about testing it? Unit tests?

sxllwx:
> Thank you for this fix. How would you go about testing it? Unit tests?

ok~

sxllwx:
done~ PTAL, thx @alexec

workflow/executor/agent.go
```diff
@@ -368,6 +368,9 @@ func (ae *AgentExecutor) executePluginTemplate(ctx context.Context, tmpl wfv1.Te
 			return 0, err
 		} else if reply.Node != nil {
 			*result = *reply.Node
+			if reply.Node.Phase == wfv1.NodeSucceeded {
```
alexec, line 371: Should this include failed and errored nodes?

sxllwx: I don't think so. Because in failure, the user needs to `explicitly give` the delay of the requeue.

alexec (approved):

PR #10685, opened 2023-03-15
fix: PVC in wf.status should be reset when retrying workflow

terrytangyuan:
This fixes a bug where a retried workflow would fail with the error `persistentvolumeclaim xxx not found` if the PVC is deleted previously (e.g. volumeClaimGC strategy is set to `OnWorkflowCompletion`). We can fix this by resetting the PVC in workflow status so that the operator could recreate the PVC.


jessesuen (commented):
Should this instead be reset when the pvc is deleted? In the controller code?

alexec (approved):

PR #10689, opened 2023-03-16
fix(ui): reword Workflow `DELETED` error

### LINKED ISSUE #10688: reword `Workflow deleted`
### Pre-requisites

1. I'm looking at a workflow run, and something deletes the workflow (not me!)
2. I see:
    image
3. The code is from:
    https://github.com/argoproj/argo-workflows/blob/94f66a20eb5fb3aca63556ecf67a77a9900b9a99/ui/src/app/workflows/components/workflow-details/workflow-details.tsx#L296

Note that from my perspective `Workflow deleted` is a reasonable message to show in response to me actively clicking a button in the ui labeled `Delete` or `Delete workflow` which results in a workflow.

As I didn't click a `Delete (Workflow)` button, this message is surprising / inappropriate.

## Expected results

Something that doesn't imply I just deleted the workflow from this screen.

`Workflow gone` is the simplest change that would do that, I'm not entirely committed to that phrase, but I believe it's an improvement over `Workflow deleted`.

### Version

latest

```YAML
any workflow -- so long as you look at it and then have something else delete it
```

```text
...
```

```text
...
```

jsoref:
Fixes #10688

### DIFF
--- a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
+++ b/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
@@ -293,7 +293,7 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
             () => setError(null),
             e => {
                 if (e.type === 'DELETED') {
-                    setError(new Error('Workflow deleted'));
+                    setError(new Error('Workflow gone'));
                 } else {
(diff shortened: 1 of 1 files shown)


alexec (approved):

alexec:
Approved!

PR #10693, opened 2023-03-16
feat: add custom columns support for workflow list views

jiachengxu:
This PR adds custom columns support on the backend, which will be used as custom columns on the workflow list views to display extra columns based on workflow labels and annotations.
The column information is exposed via `/info` endpoint like what we are doing for the links feature(https://argoproj.github.io/argo-workflows/links/):
Pasted Graphic 1
This PR only contains the necessary on the backend and @rbreeze will help with the UI support.
Related to https://github.com/argoproj/argo-workflows/issues/10447


manifests/quick-start-minimal.yaml
```diff
@@ -1396,6 +1396,10 @@ data:
       secretKeySecret:
         name: my-minio-cred
         key: secretkey
+  columns: |
+    - name: Workflow Completed
+      type: label
+      key: workflows.argoproj.io/completed
```
terrytangyuan, line 1402: I wonder if users would also want this column to include workflows that match multiple label keys. Would it be better for this to be a label selector?

rbreeze (approved):
LGTM!

terrytangyuan (approved):

PR #10735, opened 2023-03-22
fix: Release locks by terminating pending workflows immediately. Fixes #10733

### LINKED ISSUE #10733: Terminating Pending workflows waiting on a mutex/semaphore are not immediate
### Pre-requisites

After submitting a workflow that becomes Pending because it is waiting on a mutex or sempahore, when that workflow is terminated, it will not be terminated immediately, even though it has yet to begun. It will only be terminated once that workflow can acquire the semaphore/mutex, which may be a long time because it needs to wait for it to be released by another workflow.

We should allow Pending workflows to terminate immediately if they never had aquired the semaphore, because there is no consequence.

To reproduce:
1. Submit two workflows in example below
2. Run `argo terminate` on second workflow that is in Pending state:
3. Notice that the Pending workflow is still waiting on lock:

```
$ argo terminate synchronization-wf-level-cl7rp
workflow synchronization-wf-level-cl7rp terminated

$ argo list
NAME                             STATUS    AGE   DURATION   PRIORITY   MESSAGE
synchronization-wf-level-cl7rp   Pending   11s   0s         0          Waiting for argo/Mutex/test lock. Lock status: 0/1
synchronization-wf-level-rcnqz   Running   12s   12s        0
```

To fix this, I think the controller need to consider examining the following when reconciling Pending workflows waiting on mutexes/semaphores:

```yaml
spec:
  shutdown: Terminate
```

And proceed with the termination even if the workflow is not at the top of the priority queue. 

### Version

v3.4.5

```YAML
kind: Workflow
metadata:
  generateName: synchronization-wf-level-
spec:
  entrypoint: whalesay
  synchronization:
    mutex:
      name:  test
  templates:
    - name: whalesay
      container:
        image: docker/whalesay:latest
        command: [sh, -c]
        args: ["sleep 99999"]
```

```text
time="2023-03-22T17:22:19.611Z" level=info msg="Processing workflow" namespace=argo workflow=synchronization-wf-level-cl7rp
time="2023-03-22T17:22:19.620Z" level=warning msg="Workflow processing has been postponed due to concurrency limit" namespace=argo workflow=synchronization-wf-level-cl7rp
time="2023-03-22T17:22:19.620Z" level=info msg="Updated phase  -> Pending" namespace=argo workflow=synchronization-wf-level-cl7rp
time="2023-03-22T17:22:19.620Z" level=info msg="Updated message  -> Waiting for argo/Mutex/test lock. Lock status: 0/1" namespace=argo workflow=synchronization-wf-level-cl7rp
time="2023-03-22T17:22:19.626Z" level=info msg="Workflow update successful" namespace=argo phase=Pending resourceVersion=366960480 workflow=synchronization-wf-level-cl7rp
time="2023-03-22T17:22:38.098Z" level=info msg="Processing workflow" namespace=argo workflow=synchronization-wf-level-cl7rp
time="2023-03-22T17:22:38.098Z" level=warning msg="Workflow processing has been postponed due to concurrency limit" namespace=argo workflow=synchronization-wf-level-cl7rp
```

```text
N/A
```

terrytangyuan:
Fixes #10733.

TODO: Add tests

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -246,6 +246,12 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 		}
 		woc.updated = wfUpdate
 		if !acquired {
+			if sdStrategy := woc.GetShutdownStrategy(); sdStrategy == wfv1.ShutdownStrategyTerminate && woc.execWf.Status.Phase == wfv1.WorkflowPending {
+				if woc.controller.syncManager.ReleaseAll(woc.execWf) {
+					woc.log.WithFields(log.Fields{"key": woc.execWf.Name}).Info("Released all locks since this pending workflow is terminating")
+					woc.markWorkflowSuccess(ctx)
+				}
+			}
 			woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
 			phase := woc.wf.Status.Phase
 			if phase == wfv1.WorkflowUnknown {
@@ -1867,6 +1873,12 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 		if !lockAcquired {
 			if node == nil {
 				node = woc.initializeExecutableNode(nodeName, wfutil.GetNodeType(processedTmpl), templateScope, processedTmpl, orgTmpl, opts.boundaryID, wfv1.NodePending, msg)
+			}
+			if sdStrategy := woc.GetShutdownStrategy(); sdStrategy == wfv1.ShutdownStrategyTerminate && woc.execWf.Status.Phase == wfv1.WorkflowPending {
+				if woc.controller.syncManager.ReleaseAll(woc.execWf) {
+					woc.log.WithFields(log.Fields{"key": woc.execWf.Name}).Info("Released all locks since this pending workflow is terminating")
+					woc.markWorkflowSuccess(ctx)
+				}
 			}
 			lockName, err := argosync.GetLockName(processedTmpl.Synchronization, woc.wf.Namespace)
 			if err != nil {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -246,19 +246,15 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 		}
 		woc.updated = wfUpdate
 		if !acquired {
-			if sdStrategy := woc.GetShutdownStrategy(); sdStrategy == wfv1.ShutdownStrategyTerminate && woc.execWf.Status.Phase == wfv1.WorkflowPending {
-				if woc.controller.syncManager.ReleaseAll(woc.execWf) {
-					woc.log.WithFields(log.Fields{"key": woc.execWf.Name}).Info("Released all locks since this pending workflow is terminating")
-					woc.markWorkflowSuccess(ctx)
+			if !woc.releaseLocksForPendingTerminatingWfs(ctx) {
+				woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
+				phase := woc.wf.Status.Phase
+				if phase == wfv1.WorkflowUnknown {
+					phase = wfv1.WorkflowPending
 				}
+				woc.markWorkflowPhase(ctx, phase, msg)
+				return
 			}
-			woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
-			phase := woc.wf.Status.Phase
-			if phase == wfv1.WorkflowUnknown {
-				phase = wfv1.WorkflowPending
-			}
-			woc.markWorkflowPhase(ctx, phase, msg)
-			return
 		}
 	}
 
@@ -494,6 +490,17 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 	if err := woc.deletePVCs(ctx); err != nil {
 		woc.log.WithError(err).Warn("failed to delete PVCs")
 	}
+}
+
+func (woc *wfOperationCtx) releaseLocksForPendingTerminatingWfs(ctx context.Context) bool {
+	if sdStrategy := woc.GetShutdownStrategy(); sdStrategy == wfv1.ShutdownStrategyTerminate && woc.execWf.Status.Phase == wfv1.WorkflowPending {
+		if woc.controller.syncManager.ReleaseAll(woc.execWf) {
+			woc.log.WithFields(log.Fields{"key": woc.execWf.Name}).Info("Released all locks since this pending workflow is terminating")
+			woc.markWorkflowSuccess(ctx)
+			return true
+		}
+	}
+	return false
 }
 
 // set Labels and Annotations for the Workflow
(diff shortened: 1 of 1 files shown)


workflow/controller/operator.go
```diff
@@ -490,6 +492,17 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
+func (woc *wfOperationCtx) releaseLocksForPendingTerminatingWfs(ctx context.Context) bool {
+	if sdStrategy := woc.GetShutdownStrategy(); sdStrategy == wfv1.ShutdownStrategyTerminate && woc.execWf.Status.Phase == wfv1.WorkflowPending {
+		if woc.controller.syncManager.ReleaseAll(woc.execWf) {
+			woc.log.WithFields(log.Fields{"key": woc.execWf.Name}).Info("Released all locks since this pending workflow is terminating")
+			woc.markWorkflowSuccess(ctx)
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
terrytangyuan: I think so. I'll update to include `ShutdownStrategyStop`.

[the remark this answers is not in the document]
terrytangyuan: @jessesuen Actually, `argo stop` should be treated differently since all exit handlers can still run.

alexec (approved):

workflow/controller/operator.go
```diff
@@ -490,6 +492,23 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
+			// 1. If a workflow is being stopped, all exit handlers should still run.
+			// 2. If a workflow is being terminated, workflow should be marked as succeeded.
+			if woc.execWf.Spec.HasExitHook() && woc.GetShutdownStrategy().ShouldExecute(true) {
+				woc.markWorkflowRunning(ctx)
+				return true
+			}
+			woc.markWorkflowSuccess(ctx)
+			return true
+		}
+	}
```
jessesuen, line 508: Trying to understand something: 
if we stop a pending workflow that wants to acquire a mutex/semaphore, are we allowing it to move to Running without acquiring the lock?

jessesuen: Because if so, the user's exit handler might also do things that need exclusive access and we would be violating the mutex/semaphore

jessesuen: I think it is incorrect behavior for stop. We can't assume the intent of exit handler/hook logic. It could be considered part of a critical code section by a user. The guarantee we must always make with workflow-level semaphores/mutexes is that at most N workflows will run at a given time. Argo stop should not change that.

We do have to decide: `argo stop` of a Pending workflow should either:
1. allow a workflow to acquire the mutex/semaphore normally, and then essentially just run exit handlers as soon as it becomes Running
2. or behave identically to `argo terminate` and never execute any part of the workflow

Given that `argo terminate` already handles cancelling a workflow without running any logic, I think 2 may be more desirable?

jessesuen: If the user wants their exit handlers to always run but have their main logic protected by the semaphore/mutex, then the user can do the following:
1. use a **_template_** level mutex/semaphore for their entrypoint template (not workflow level)
2. do not protect exit handlers/hooks with any mutex/semaphore
3. Run `argo stop` against their Running workflows instead of terminate

terrytangyuan: Conclusion after discussion with Jesse:
1. `argo terminate` of a pending workflow waiting on a lock, immediately transitions it to `Succeeded` without waiting for lock aquisition
2. `argo stop` of a pending workflow waiting on a lock, will acquire the lock normally, and reconcile normally

terrytangyuan:
@jessesuen Updated the PR based on our discussions.

workflow/controller/operator.go
```diff
@@ -490,6 +492,22 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
 }
 
+func (woc *wfOperationCtx) releaseLocksForPendingShuttingdownWfs(ctx context.Context) bool {
+	if woc.GetShutdownStrategy().Enabled() && woc.wf.Status.Phase == wfv1.WorkflowPending {
+		// 1. If a workflow is being stopped, all exit handlers should still run.
+		// 2. If a workflow is being terminated, workflow should be marked as succeeded.
+		if woc.GetShutdownStrategy() == wfv1.ShutdownStrategyStop {
+			return false
+		}
+		if woc.controller.syncManager.ReleaseAll(woc.execWf) {
```
jessesuen, line 502: It seems we treat ShutdownStrategyStop special, whereas the logic would be more correct to treat the Terminate case specially. Because we actually don't care about  ShutdownStrategyStop at all, and should just let it fall down to the `return false` at the end of the function.

```golang
if woc.GetShutdownStrategy().Enabled() && woc.wf.Status.Phase == wfv1.WorkflowPending && woc.GetShutdownStrategy() == wfv1.ShutdownStrategyTerminate {
    if woc.controller.syncManager.ReleaseAll(woc.execWf) {
			woc.log.WithFields(log.Fields{"key": woc.execWf.Name}).Info("Released all locks since this pending workflow is being shutdown")
			woc.markWorkflowSuccess(ctx)
			return true
    }
}
```

terrytangyuan: Oh yes you are right. I should’ve just reverted to my original commit :-) 

Updated.

workflow/controller/operator.go
```diff
@@ -1871,13 +1884,15 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
...
-				// unexpected behavior and is a bug.
-				panic("bug: GetLockName should not return an error after a call to TryAcquire")
+			if !woc.releaseLocksForPendingShuttingdownWfs(ctx) {
+				lockName, err := argosync.GetLockName(processedTmpl.Synchronization, woc.wf.Namespace)
+				if err != nil {
+					// If an error were to be returned here, it would have been caught by TryAcquire. If it didn't, then it is
+					// unexpected behavior and is a bug.
+					panic("bug: GetLockName should not return an error after a call to TryAcquire")
+				}
+				return woc.markNodeWaitingForLock(node.Name, lockName.EncodeName()), nil
```
jessesuen, line 1894: Sorry, I just noticed this.

I don't think it is correct to be releasing locks as part of `executeTemplate()` when terminating a Running workflow. Because at this point, the workflow is already Running and has acquired the lock.  By releasing the locks here, I believe we will violate mutex and semaphore. Why not let it release naturally?

terrytangyuan: Good catch indeed. I just removed this part of the logic.

jessesuen (approved):

workflow/controller/operator.go
```diff
@@ -246,13 +246,15 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
+				woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
+				phase := woc.wf.Status.Phase
+				if phase == wfv1.WorkflowUnknown {
+					phase = wfv1.WorkflowPending
+				}
+				woc.markWorkflowPhase(ctx, phase, msg)
+				return
 			}
-			woc.markWorkflowPhase(ctx, phase, msg)
-			return
```
agilgur5, line 255: this could be simplified / doesn't need another level of nesting

```suggestion
		if !acquired && !woc.releaseLocksForPendingShuttingdownWfs(ctx) {
			woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
			phase := woc.wf.Status.Phase
			if phase == wfv1.WorkflowUnknown {
				phase = wfv1.WorkflowPending
			}
			woc.markWorkflowPhase(ctx, phase, msg)
			return
```

Simplified this in #12965

PR #10745, opened 2023-03-24
feat: Add workflow var for originName

### LINKED ISSUE #6772: Use template name in custom metric label definition
# Summary

We want to collect duration steps for every part of the workflow but for now template name is not available in variables so we can not refer to it while defining labels.

# Use Cases

Collect the same metrics for all steps/tasks in workflow.

```yaml
templateDefaults:
  metrics:
    prometheus:
      - name: result_counter
        help: "Count of step execution by result status"
        labels:
          - key: template_name
            value: "{{templateName}}"
           - key: status
             value: "{{status}}"    # Argo variable in `labels`
        counter:
          value: "1"
```
Also it would be nice to have some common metrics defined on the controller level so later it's possible to reference this definitions in specs. i.e.:
```yaml
  metrics:
    prometheus:
      - name: result_counter
        metricsRef: result_counter
```
---

Joibel:
Add a global workflow variables to allow generic template metrics to identify which workflow they originate from.

Issue #6772 asks for the template name, but we cannot always know what this is (simplest unknowable is kubectl apply of a workflow with a generateName). So provide that by deduction as originName. See the tests added in this commit for what we're aiming for there.

I have added both variables and tested them by invoking very simple workflows. The regex used to generate originName from the Name has a test suite, and has been checked from simple workflows also.

Fixes #6772

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -246,6 +246,7 @@ For `Template`-level metrics:
 | Variable | Description|
 |----------|------------|
 | `workflow.name` | Workflow name |
+| `workflow.originName` | Workflow template name. This works only for template names, it may give wrong answers for non-generated names |
 | `workflow.namespace` | Workflow namespace |
 | `workflow.mainEntrypoint` | Workflow's initial entrypoint |
 | `workflow.serviceAccountName` | Workflow service account name |
--- a/workflow/controller/operator_test.go
+++ b/workflow/controller/operator_test.go
@@ -366,6 +366,7 @@ func TestGlobalParams(t *testing.T) {
 
 	assert.Contains(t, woc.globalParams, "workflow.duration")
 	assert.Contains(t, woc.globalParams, "workflow.name")
+	assert.Contains(t, woc.globalParams, "workflow.originName")
 	assert.Contains(t, woc.globalParams, "workflow.namespace")
 	assert.Contains(t, woc.globalParams, "workflow.mainEntrypoint")
 	assert.Contains(t, woc.globalParams, "workflow.parameters")
@@ -377,6 +378,22 @@ func TestGlobalParams(t *testing.T) {
 	// Ensure that the phase label is included after the first operation
 	woc.operate(ctx)
 	assert.Contains(t, woc.globalParams, "workflow.labels.workflows.argoproj.io/phase")
+}
+
+func TestOriginName(t *testing.T) {
+	for input, expected := range map[string]string{
+		`template-name-1234567890`:       `template-name`, // Cronworkflows run on a cron
+		`template-name1234567890`:        `template-name`, // Cronworkflow without trailing -
+		`foobar-fs4hy`:                   `foobar`,        // Standard template submission
+		`foobar-foobar-ashda`:            `foobar-foobar`,
+		`foobarfs4hy`:                    `foobar`, // Template without trailing -
+		`foobar`:                         `foobar`, // We don't replace the end of this
+		`foo-bar`:                        `foo-bar`,
+		`foobar-foobar-0987654321-4ab1a`: `foobar-foobar`, // Cronworkflow, resubmitted
+		`foobar-foobar`:                  `foobar-f`,      // Can't really prevent this kind of thing, not desired though
+	} {
+		assert.Equal(t, expected, getOriginNameFromName(input))
+	}
 }
 
 // TestSidecarWithVolume verifies ia sidecar can have a volumeMount reference to both existing or volumeClaimTemplate volumes
--- a/workflow/common/common.go
+++ b/workflow/common/common.go
@@ -165,6 +165,8 @@ const (
 
 	// GlobalVarWorkflowName is a global workflow variable referencing the workflow's metadata.name field
 	GlobalVarWorkflowName = "workflow.name"
+	// GlobalVarWorkflowOriginName is a global workflow variable referencing the workflow's name reverse engineered to the template name from a generateName
+	GlobalVarWorkflowOriginName = "workflow.originName"
 	// GlobalVarWorkflowNamespace is a global workflow variable referencing the workflow's metadata.namespace field
 	GlobalVarWorkflowNamespace = "workflow.namespace"
 	// GlobalVarWorkflowMainEntrypoint is a global workflow variable referencing the workflow's top level entrypoint name
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -577,9 +577,21 @@ func (woc *wfOperationCtx) getWorkflowDeadline() *time.Time {
 	return &deadline
 }
 
+// originNameFromName tries to strip the suffix for a cronworkflow [\d]{10} or from a
+// generateName [a-z0-9]{5} from the name.
+var originNameFromName = regexp.MustCompile(`^(.{3,}?)(\-?[\d]{10})?(\-?[a-z0-9]{5})?$`)
+
+func getOriginNameFromName(name string) string {
+	if originNameFromName.MatchString(name) {
+		return originNameFromName.ReplaceAllString(name, "$1")
+	}
+	return name
+}
+
 // setGlobalParameters sets the globalParam map with global parameters
 func (woc *wfOperationCtx) setGlobalParameters(executionParameters wfv1.Arguments) error {
 	woc.globalParams[common.GlobalVarWorkflowName] = woc.wf.ObjectMeta.Name
+	woc.globalParams[common.GlobalVarWorkflowOriginName] = getOriginNameFromName(woc.wf.ObjectMeta.Name)
 	woc.globalParams[common.GlobalVarWorkflowNamespace] = woc.wf.ObjectMeta.Namespace
 	woc.globalParams[common.GlobalVarWorkflowMainEntrypoint] = woc.execWf.Spec.Entrypoint
 	woc.globalParams[common.GlobalVarWorkflowServiceAccountName] = woc.execWf.Spec.ServiceAccountName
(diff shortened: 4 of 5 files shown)


docs/variables.md
```diff
@@ -243,7 +243,9 @@ For `Template`-level metrics:
 | Variable | Description|
 |----------|------------|
 | `workflow.name` | Workflow name |
+| `workflow.templateName` | Workflow template name. This works only for template names, it may give wrong answers for non-generated names |
 | `workflow.namespace` | Workflow namespace |
+| `workflow.main entrypoint` | Workflow's initial entrypoint |
```
rohankmr414, line 248:
```suggestion
| `workflow.mainEntrypoint` | Workflow's initial entrypoint |
```

rohankmr414 (changes requested):
LGTM, requested a minor typo fix change

docs/variables.md
Joibel: Oops, thanks

rohankmr414 (approved):
LGTM!

docs/variables.md
```diff
@@ -243,7 +243,9 @@ For `Template`-level metrics:
 | Variable | Description|
 |----------|------------|
 | `workflow.name` | Workflow name |
+| `workflow.templateName` | Workflow template name. This works only for template names, it may give wrong answers for non-generated names |
 | `workflow.namespace` | Workflow namespace |
+| `workflow.mainEntrypoint` | Workflow's initial entrypoint |
```
terrytangyuan, line 248: This is static. What's the use case to have this available?

Joibel: The original problem was distinguishing which high level workflow has invoked us. This is not a static for a WorkflowTemplate called from a another place.
Given this use in a workflowTemplate:
```
kind: WorkflowTemplate
metadata:
  name: say-main-entrypoint
spec:
  entrypoint: echo
  templates:
  - name: echo
    container:
      image: alpine
      command: [echo]
      args: ["{{workflow.mainEntrypoint}}"]
```
I can distinguish my caller:
```
kind: Workflow
metadata:
  generateName: foo-
spec:
  entrypoint: foo
  templates:
    - name: foo
      steps:
      - - name: step
          templateRef:
            name: say-main-entrypoint
            template: echo
```
results in a log of `foo`


```
kind: Workflow
metadata:
  generateName: bar-
spec:
  entrypoint: bar
  templates:
    - name: bar
      steps:
      - - name: step
          templateRef:
            name: say-main-entrypoint
            template: echo
```
results in a log of `bar`

It fulfills at least some of the original need in #6772 in a different way, and in a less naughty fashion than the templateName option. @terrytangyuan - if you're happy that it's actually useful, I'll add some more documentation to this PR as it's obvious I haven't explained what use this is.

isubasinghe (approved):

docs/variables.md
```diff
@@ -243,7 +243,9 @@ For `Template`-level metrics:
 | Variable | Description|
 |----------|------------|
 | `workflow.name` | Workflow name |
+| `workflow.templateName` | Workflow template name. This works only for template names, it may give wrong answers for non-generated names |
```
juliev0, line 246: At the risk of being verbose, I wonder if we should call this `workflowTemplateName`? When I see `templateName`, I think of the individual step, not the WorkflowTemplate

juliev0: From [this](https://argoproj.github.io/argo-workflows/workflow-templates/) link: "The terms WorkflowTemplate and template have created an unfortunate naming collision and have created some confusion in the past."

Joibel: I know what you mean. To fit the two part naming of all the rest of the variables how about workflowTemplate.name?

juliev0: For my understanding, does this also represent the CronWorkflow name for a Workflow generated by a CronWorkflow?

Joibel: @juliev0, yes, it should do. Ideally it would always be able to derive the name from the parent, but as the names are generated and this attempts to reverse that, I'm basing it on what others in #6772 have found works.

juliev0: So, in that case using `workflowTemplate` in the name would also be inaccurate. I'm thinking something like `origin`, or otherwise dividing this into multiple variables. What do you think?

Joibel: So `workflow.origin` is your proposal? Or `workflow.originName` to demonstrate it's the name of something perhaps.

juliev0: `workflow.originName` sounds good to me. If you'd like you can wait a few days to see if anyone chimes in first, for risk of your changing it once and then needing to change it again.

workflow/controller/operator_test.go
```diff
@@ -375,6 +376,22 @@ func TestGlobalParams(t *testing.T) {
...
+		`foobar-foobar-ashda`:            `foobar-foobar`,
+		`foobarfs4hy`:                    `foobar`, // Template without trailing -
+		`foobar`:                         `foobar`, // We don't replace the end of this
+		`foo-bar`:                        `foo-bar`,
+		`foobar-foobar-0987654321-4ab1a`: `foobar-foobar`, // Cronworkflow, resubmitted
+		`foobar-foobar`:                  `foobar-f`,      // Can't really prevent this kind of thing, not desired though
+	} {
+		assert.Equal(t, expected, getTemplateNameFromName(input))
+	}
+}
```
juliev0, line 393: So, it looks like if there's a Workflow whose actual name is for example `template-name-1234567890`, we will assume that the origin is `template-name`, right?

Joibel: Yes, that's the point of these tests. Show how originName will behave given a specific name being fed to it as the workflow name.

juliev0: Oh, sorry, let me clarify my question. What I meant was, say there's a Workflow that wasn't created from a WorkflowTemplate or a CronWorkflow, and it just happens to be named in a way in which it has the "-" followed by a certain number of letters or numbers. In that case, the actual origin should be the entire Workflow name, right?

Joibel: It should, but it won't be. We're reverse engineering the name. The same might happen for a workflow submitted with a `generateName:` or even just a name that matches. I'd prefer to fix this with some "buyer beware" stickers in the documentation - if you want to fool this you're welcome to but you won't get what you want. We could get the controller generated workflows to add annotations of their original name so that we can then re-injest that as the correct name, but that is a fair amount of work to avoid users deliberately being silly when they can just read the instructions instead and not fool the regex. And it still won't work in all cases where workflows are being injected by other tooling including kubectl and argo-events.

workflow/controller/operator_test.go
```diff
@@ -6634,6 +6651,33 @@ func TestWorkflowScheduledTimeVariable(t *testing.T) {
...
+	wf := wfv1.MustUnmarshalWorkflow(wfMainEntrypointVariable)
+	cancel, controller := newController(wf)
+	defer cancel()
+
+	ctx := context.Background()
+	woc := newWorkflowOperationCtx(wf, controller)
+	woc.operate(ctx)
+	assert.Equal(t, "whalesay", woc.globalParams[common.GlobalVarWorkflowMainEntrypoint])
+}
+
```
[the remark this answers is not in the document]
Joibel: Yes, I'm pretty sure that's doable, I will do once we've settled on functionality.

workflow/controller/operator_test.go
juliev0: The result of a Workflow submitted with a `generateName` might actually be pretty good - if your `generateName` is `foobar-` I guess the `origin` would be `foobar`, right?

juliev0: @terrytangyuan what do you think? Is there any precedence for template variables that come with a "buyer beware" sticker (i.e. don't use in certain cases or you might get the wrong result)?

juliev0: Just thinking about the alternative - I guess somebody would need to include their own sprig function to perform the same transformation you did, right? I recognize that would make the Workflow more complex.

Joibel: > The result of a Workflow submitted with a `generateName` might actually be pretty good - if your `generateName` is `foobar-` I guess the `origin` would be `foobar`, right?

Yes. But it's impossible to distinguish between `name: foobar-123467890` and `generateName: foobar-`.

Joibel: > Just thinking about the alternative - I guess somebody would need to include their own sprig function to perform the same transformation you did, right? I recognize that would make the Workflow more complex.

They can already do this, and are... see the issue #6772 this is trying to assist with. We don't need to support overriding the regex in this facility as users can and are already implementing that.

The reliable answer is to use the alternative I've provided in this PR, the entrypoint name instead.

juliev0: I like the entrypoint functionality. Are you thinking of maybe just including that in this PR then?

Joibel: I'd prefer to include both.
`mainEntrypoint` is the 'right answer', but it requires work and requires that users write all their workflows with a differing entrypoint, which is not initially likely, as probably they all have them called `main` or at least haven't given them good, unique names. Adding controls to ensure `mainEntrypoint` is unique for all your workflows ever is problematic.
Because of this, the alternative `originName` is a quick fix for people who want a nice distinguishing name with minimal effort and has been shown to work in the field effectively. We can then determine whether we should put more effort into that solution when issues arise, but also point at `mainEntrypoint` in the docs. Perhaps we can implement workflow automatically generated annotations from common sources if it proved a popular but flawed feature.

Let me do the documentation that's needed from Terry's first comment, and that might help.

juliev0: Okay, thanks. I'll leave it to @terrytangyuan to decide if we should include a variable that should only be used in certain circumstances and won't work if used in the wrong circumstance, if it provides value to people. (I'm not sure if we've come across that type of thing before)

terrytangyuan:
We discussed it today at the contributors meeting. My concern is that this is one-off solution and introduces confusion to users when it doesn't work for other cases. Any thoughts? @sarabala1979

Joibel:
I'm finding more use cases for this, so as a basic proposal I'm planning to add a label to the sources of workflows that we control that can supply this originName. The label will be used where present to provide this variable.

juliev0:
Any plans to continue this?

Joibel:
Not as is. I'll write up a proposal as an issue instead for discussion.

PR #10759, opened 2023-03-25
fix: updates the curl example to use the BASE_HREF. Fixes #7416

### LINKED ISSUE #7416: workflow-event-bindings page support basehref

mikeringrose:
Fixes #7416
* Update the Workflow Event Bindings page such that the curl command respects the `BASE_HREF`
* The update creates a util method named `absoluteUrl` that takes a path returns a URL rooted at `document.baseURI`.
* The changes were tested locally both with a `BASE_HREF` specified and with out.

Screen Shot 2023-03-25 at 4 07 36 PM
Screen Shot 2023-03-25 at 4 08 02 PM


ui/src/app/workflow-event-bindings/components/workflow-event-bindings/workflow-event-bindings.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import {useContext, useEffect, useState} from 'react';
 import {RouteComponentProps} from 'react-router-dom';
 import {WorkflowEventBinding} from '../../../../models';
-import {uiUrl} from '../../../shared/base';
+import {absoluteUrl, uiUrl} from '../../../shared/base';
```
tczhao, line 6: would `uiUrl` achieve the same purpose?

alexec (approved):

PR #10770, opened 2023-03-28
docs: Add Salesforce to USERS.md

vnzongzna:
As we are heavily using the argo ecosystem for the management of our workflows, we would like to reflect that on the users list :)


terrytangyuan (approved):
🎉

PR #10778, opened 2023-03-29
docs: Clarify DinD usage on newer k8s versions

tico24:
Docker in Docker is (to all intents and purposes) not possible in k8s 1.24+

We still support older versions of k8s so I didn't want to just delete this page, but we should make it clear that DinD isn't a cool thing to be trying on modern k8s.


rohankmr414 (approved):

terrytangyuan (approved):

agilgur5:
Apparently DinD is actually still possible without `dockershim` and the `docker:dind` sidecar in this example _already_ includes Docker Engine to make that possible. C.f. https://github.com/argoproj/argo-workflows/discussions/13010#discussioncomment-9372559, https://aws.amazon.com/blogs/containers/all-you-need-to-know-about-moving-to-containerd-on-amazon-eks/.
Buildkit, Kaniko, etc would still be preferred, so just the DinD part needs correction

PR #10786, opened 2023-03-30
fix: Fix inlined templates in templates

### LINKED ISSUE #10737: inline templates not executed correctly
### Pre-requisites

When there are two or more inline templates in a workflowtemplate that is called from a workflow, the content of the first executed inline templates is replaced by the content of the subsequent inline script and then executed.

### Version

v3.4.5 

```YAML
kind: Workflow
metadata:
  generateName: workflow-test-
spec:
  entrypoint: hello-hello
  templates:
    - name: hello-hello
      steps:
        - - name: start
            templateRef:
              name: template-hello
              template: main

---
kind: WorkflowTemplate
metadata:
  name: template-hello
spec:
  templates:
    - name: main
      steps:
        - - name: hello1
            inline:
              script:
                image: alpine:3.7
                command: [ sh ]
                source: |
                  echo "hello1"
        - - name: hello2
            inline:
              script:
                image: alpine:3.7
                command: [ sh ]
                source: |
                  echo "hello2"
```

```text
workflow-hello-xd4px--1205290903: time="2023-03-19T06:35:30.018Z" level=info msg="capturing logs" argo=true
workflow-hello-xd4px--1205290903: hello1
workflow-hello-xd4px--1205290903: time="2023-03-19T06:35:31.025Z" level=info msg="sub-process exited" argo=true error="<nil>"
workflow-hello-xd4px--3965209339: time="2023-03-19T06:35:41.045Z" level=info msg="capturing logs" argo=true
workflow-hello-xd4px--3965209339: hello1
workflow-hello-xd4px--3965209339: time="2023-03-19T06:35:42.069Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

```text
```

Joibel:
When a template is inline inside a WorkflowTemplate or ClusterWorkflowTemplate we must still create a stored version of it. This fix changes the 'name' given to those from `/scope/templateName/` (where after the last / should be a name but there isn't) to /scope/templateName/inline/name' where name can be step name or task name.

I have added inline as it's possible to name an inline step the same as a template in the same WorkflowTemplate.

Tested against the test cases from #10737.

Fixes #10737

### DIFF
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -3280,6 +3280,10 @@ func resolveTemplateReference(callerScope ResourceScope, resourceName string, ca
 		return fmt.Sprintf("%s/%s/%s", referenceScope, tmplRef.Name, tmplRef.Template), true
 	} else if callerScope != ResourceScopeLocal {
 		// Either a WorkflowTemplate or a ClusterWorkflowTemplate is calling a template inside itself. Template storage is needed
+		if caller.GetTemplate() != nil {
+			// If we have an inlined template here, use the inlined name
+			return fmt.Sprintf("%s/%s/inline/%s", callerScope, resourceName, caller.GetName()), true
+		}
 		return fmt.Sprintf("%s/%s/%s", callerScope, resourceName, caller.GetTemplateName()), true
 	} else {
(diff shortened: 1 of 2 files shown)


alexec:
Could you please add a unit test?

terrytangyuan (approved):
Thanks!

PR #10473, opened 2023-02-06
feat: validate the name of workflow template.

sxllwx:
Signed-off-by: scott <scottwangsxll@gmail.com>

Fixes 

Please do not open a pull request until you have checked ALL of these:

* [x] Create the PR as draft .
* [ ] Run `make pre-commit -B` to fix codegen and lint problems.
* [x] Sign-off your commits (otherwise the DCO check will fail).
* [x] Use [a conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) (otherwise the commit message check will fail).
* [x] "Fixes #" is in both the PR title (for release notes) and this description (to automatically link and close the issue).
* [x] Add unit or e2e tests. Say how you tested your changes. If you changed the UI, attach screenshots.
* [x] Github checks are green.
* [x] Once required tests have passed, mark your PR "Ready for review".

If changes were requested, and you've made them, dismiss the review to get it reviewed again.

---

A note for reviewers:

Please correct me if I'm wrong thx:

> The name of the template (not workflowTemplate) affects the name of the Workflow Pod.

In order to prevent the pod names created by `workflow-controller` from not conforming to [kubernetes named constraints](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names). If we can verify when submitting the workflow, there will be a better user experience. Otherwise, the argo-workflow CR resource has been generated but cannot be run because the Pod name is invalid, which brings a complicated troubleshooting process.

PR #10793, opened 2023-03-31
fix: workflow-of-workflows example

tico24:
todo


terrytangyuan (commented):
https://github.com/argoproj/argo-workflows/pull/10473

PR #10795, opened 2023-03-31
chore: Print out transient S3 errors

terrytangyuan:
These errors are in the backoff retry loop and users are not clear what/whether errors are retried.

### DIFF
--- a/workflow/artifacts/s3/errors.go
+++ b/workflow/artifacts/s3/errors.go
@@ -2,6 +2,7 @@ package s3
 
 import (
 	argos3 "github.com/argoproj/pkg/s3"
+	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


rbreeze (approved):

workflow/artifacts/s3/errors.go
```diff
@@ -27,6 +28,7 @@ func isTransientS3Err(err error) bool {
 	}
 	for _, transientErrCode := range s3TransientErrorCodes {
 		if argos3.IsS3ErrCode(err, transientErrCode) {
+			log.Infof("Transient S3 error: %v", err)
```
jessesuen, line 31: Wouldn't you want this to be log.Errorf?

jessesuen (approved):

PR #10806, opened 2023-04-01
docs: fix for async-pattern.md

vasi-grigo:
`async-pattern` docs contained errors and were not clear enough
PR fixes and expands on the docs for the async pattern using both argo CLI and curl
Changes were tested by running both setups in GKE.


terrytangyuan (commented):
Please fix docs build. Thanks.

vasi-grigo:
can you point me to what is failing?
i couldn't really understand

terrytangyuan:
See details in https://github.com/argoproj/argo-workflows/actions/runs/4586747576/jobs/8099799966?pr=10806

You also deleted the template in your PR description. It has instructions: https://raw.githubusercontent.com/argoproj/argo-workflows/master/.github/pull_request_template.md

docs/async-pattern.md
```diff
@@ -56,37 +58,89 @@ spec:
     suspend: {}
 ```
 
-In this case the ```job-cmd``` parameter can be a command that makes an HTTP call via curl to an endpoint that returns a job UUID. More sophisticated submission and parsing of submission output could be done with something like a Python script step.
+In this case the `job-cmd` parameter can be a command that makes an HTTP call via curl to an endpoint that returns a job UUID. More sophisticated submission and parsing of submission output could be done with something like a Python script step.
```
alexec, line 61:
```suggestion
In this case the `cmd` parameter can be a command that makes an HTTP call via curl to an endpoint that returns a job UUID. More sophisticated submission and parsing of submission output could be done with something like a Python script step.
```

vasi-grigo:
- cleaned up the loose ends (retested stuff)
- no idea how `git diff --exit-code` is supposed to produce a 0 for a change that DID happen for docs
- something else is failing within the build checks
so what do we do guys?

agilgur5:
> * no idea how `git diff --exit-code` is supposed to produce a 0 for a change that DID happen for docs

The diff is run _on top_ of your commit. It runs on every PR and does work correctly. This is also a fairly common CI pattern.

Per the logs, it runs `make docs` and _then_ `git diff --exit-code`. If it fails, that means that `make docs` produced a diff, meaning it hadn't been run locally. In this case, it looks like lint errors from `markdownlint` were not fixed

PR #10810, opened 2023-04-04
feat: enable configuration for finalizer removal if artifact GC fails

### LINKED ISSUE #10788: Enable the option of Workflows being deleted when Artifact GC fails
# Summary

I'm trying to gauge the interest for this. I've heard some people mention in other threads they may want it. 

Currently, if Artifact GC fails (even after exponential backoff), the Artifact GC Pods and the Workflow (which has an ownership reference to it) remain around (there is a finalizer on the Workflow that's not being removed in that case). The idea behind that was:
- if you want your Artifacts deleted first, maybe your Workflow shouldn't be deleted yet
- there is evidence of what went wrong there that you may be able to use, both in a "Condition" in the Workflow Status and the log of the Artifact GC Pod

Is there interest in enabling the finalizer to be removed if Artifact GC fails? (I'm thinking it would make sense to define it on the Workflow level if so, although potentially defined for the Workflow Controller as a whole ?.)

---

juliev0:
Fixes #10788 

Why: some users seem to want their Workflows to be deletable when Artifact GC fails (through other means besides `argo delete --force`). The idea is to make it something that's configurable.

What: 

I considered:
- making it configurable in the ArtifactGC struct (which lives on the Workflow and artifact level)
- making it configurable only on the Workflow level
- making it configurable in the configmap somewhere

I tried both the 1st and 2nd options. The 1st was a possibility but it seemed to probably be unnecessary in that users won't need different configurability for different artifacts. It also meant needing to make the field something other than a boolean, since otherwise you can't really distinguish between:
- `wfLevel: forceFinalizerRemoval=true, artifactLevel: forceFinalizerRemoval=false`, and
- `wfLevel: forceFinalizerRemoval=true, artifactLevel: forceFinalizerRemoval not set`

I went with the 2nd option. 

Testing: Expanded the e2e test to test whether the finalizer was removed when ArtifactGC failed both when `forceFinalizerRemoval` was set to true and when it was set to false.


alexec (approved):
LGTM.

terrytangyuan (approved):

PR #10835, opened 2023-04-05
fix: trivialVersions is no longer a supported argument for controller-gen

juliev0:
I upgraded from controller-gen version 0.4.1 to 0.8 recently and noticed that the `trivialVersions` argument is no longer supported. When running `make codegen` I'd get:
```
Error: unable to parse option "crd:trivialVersions=true,maxDescLen=0": [unknown argument "trivialVersions" (at <input>:1:16) extra arguments provided: "true,maxDescLen=0" (at <input>:1:17)]
Usage:
  controller-gen [flags]
```

There seems to be no change to the generated files by removing it.

### DIFF
--- a/hack/crdgen.sh
+++ b/hack/crdgen.sh
@@ -8,7 +8,7 @@ add_header() {
   mv tmp "$1"
 }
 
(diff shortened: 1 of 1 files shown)


juliev0:
I see the makefile is specifying v0.4.1. Maybe I won't mess with it.

PR #10852, opened 2023-04-07
docs: forceFinalizerRemoval

juliev0:
Explain the new forceFinalizerRemoval flag in the Artifact GC docs

### DIFF
--- a/docs/walk-through/artifacts.md
+++ b/docs/walk-through/artifacts.md
@@ -258,7 +258,7 @@ kubectl patch workflow my-wf \
     --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'
 ```
 
-### Release Versions >= v3.5:
+### Release Versions >= 3.5:
 A flag has been added to the Workflow Spec called `forceFinalizerRemoval` (see [here](../fields.md#workflowlevelartifactgc)) to force the finalizer's removal even if Artifact GC fails:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/walk-through/artifacts.md
```diff
@@ -257,3 +257,12 @@ kubectl patch workflow my-wf \
...
 ```
+
+A flag has been added to the Workflow Spec called `forceFinalizerRemoval` (see [here](../fields.md#workflowlevelartifactgc)) to force the finalizer's removal even if Artifact GC fails:
+```yaml
+spec:
+  artifactGC:
+    strategy: OnWorkflowDeletion 
+    forceFinalizerRemoval: true
+
+```
```
juliev0, line 268: Should I indicate a release version for this? (unfortunately, it hasn't been released so I could only guess) Otherwise, users should assume documentation is essentially guaranteed to be correct for "latest".

juliev0:
(hoping we can merge this without waiting for e2e to pass)

terrytangyuan:
Failed checks are unrelated. See https://github.com/k3s-io/k3s/issues/7247

terrytangyuan:
CI will be fixed in https://github.com/argoproj/argo-workflows/pull/10870

alexec (dismissed):
You should state which version this is available in.

juliev0:
> You should state which version this is available in.

Yeah, I was wondering about that. Would that be both the minor release as well as the patch release or just the minor release? I guess I could wait until this does get released.

docs/walk-through/artifacts.md
caelan-io: this will be confusing until the feature is released. maybe we could hold on merging or say "coming in v3.5"?

terrytangyuan (approved):

PR #10864, opened 2023-04-09
Update workflow-of-workflows.md

jyje:
Names of template contain capital characters such as 'triggerWorkflowUsingResourceWithoutArgument'. It does not agrees with the rule: 'a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters'. So the workflow did not work well. So I changed two names from camelCase to dash case. Please check my changes.

### DIFF
--- a/docs/workflow-of-workflows.md
+++ b/docs/workflow-of-workflows.md
@@ -50,13 +50,13 @@ spec:
     - name: main
       steps:
         - - name: workflow1
-            template: triggerWorkflowUsingResourceWithoutArgument
+            template: trigger-workflow-using-resource-without-argument
             arguments:
               parameters:
               - name: workflowtemplate
                 value: "workflow-template-submittable"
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

caelan-io (approved):

jyje:
I'm ashamed to ask this question, but this is my first pull request. Is there something I need to do to finish this well?

tico24 (approved):

tico24 (changes requested):
Would it not make sense to match the naming convention with the example? https://github.com/argoproj/argo-workflows/blob/master/examples/workflow-of-workflows.yaml

jyje:
I didn't know there was an example associated with doc. Thank you for your information. I changed it according to the example. Could you check again?

tico24 (approved):
Lgtm!

Issue #9530, opened 2022-09-06
Support Large Bearer token for SSO authentication


sarabala1979:
If the user has part of a lot of groups. The bearer token is failed to be set as a cookie because it is crossing its size limit.

![image](https://user-images.githubusercontent.com/33908564/188675287-894fc9ab-a193-4f76-b892-19de71bb5e6e.png)

## Checklist

 

* [ ] Double-checked my configuration.
* [ ] Tested using `:latest` images.
* [ ] Used the Emissary executor.

## Summary

What happened/what you expected to happen?

What version are you running?

## Diagnostics

Paste the smallest workflow that reproduces the bug. We must be able to run the workflow.

```yaml

```

```bash
# Logs from the workflow controller:
kubectl logs -n argo deploy/workflow-controller | grep ${workflow} 

# If the workflow's pods have not been created, you can skip the rest of the diagnostics.

# The workflow's pods that are problematic:
kubectl get pod -o yaml -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded

# Logs from in your workflow's wait container, something like:
kubectl logs -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

sarabala1979:
@alexec @basanthjenuhb  how can we address this issue? Can we split two cookies if the token size is more than 4096?

alexec:
Cookie is compressed already. Browsers only allow 4k of cookie in total. I.e. you cannot split the cookie.

Work-around - user should remove some groups.

The only solution is to store the JWT in a cache (e.g. Redis key-value store) and the cookie is just the key within that store.

sarabala1979:
I will change to enhancement.

basanthjenuhb:
@alexec can we store in K8s secret object.
it will be just another key in the secret

alexec:
Bala and I saw today that a commit to sso.go copied the entire claimns into the argo issued JWE. This will >=2x the size of the JWE:

https://github.com/argoproj/argo-workflows/blob/9d66b69f0bca92d7ef0c9aa67e87a2e334797530/server/auth/sso/sso.go#L280

This was added in #6444 by @liafizan .

@liafizan I don't think we need this copy?

liafizan:
[...]

PR #10908, opened 2023-04-14
feat: compress bearer token (fixes #10153 #9530)

### LINKED ISSUE #10153: SSO does not work for users with a lot of groups due to exceeding max browser cookie size limit
### Pre-requisites

After setting up Argo Workflow with an external SSO we noticed certain users can login with correct groups showing up while others cannot. Argo server logs show "token not valid for running mode".

To debug, we compiled Argo Workflow ourselves and added tons of debug logs. 

Eventually we were able to pinpoint everything is normal until the part that [sets authorization Cookie in HandleCallback](https://github.com/argoproj/argo-workflows/blob/55e96972ec67d0fb5b871d1b1acc649837904fd3/server/auth/sso/sso.go). In non-working case the final Cookie size is 5000+ bytes, which exceeds the [4096 bytes limit](https://chromestatus.com/feature/4946713618939904). This causes the `authorization` Cookie not to be written at all, which makes subsequent requests do not contain the "authorization" header and leads to "token not valid for running mode" since it's empty.

I think we probably need to consider reducing the size of this Cookie or use a different approach?

Thanks.

### Version

V3.4.3

```text
N/A, just set up an fresh Argo Workflow with SSO then login with a user that has a ton of groups so cookie will exceed 4096 bytes
```

```text
N/A
```

```text
N/A
```

basanthjenuhb:
# Description
- Implementing the change after discussion with @sarabala1979 
- Currently users logging in via SSO with lot of groups, exceed the 4KB Cookie limit
- This happens because we send the full set of claims as part of cookie
- Inorder to fix this, instead of the full encrypted claims, we will just generate a UUID and send it as part of the cookie
- And we will use this UUID for authorization
- Users can choose to opt in by setting `compressBearerToken` to `true` (defaults to `false`)
- Users can also control max number of sessions to keep in memory by `maxConcurrentUsers` (defaults to 1000)
- Any change in config will require a restart

# Knows issues
- Whenever the argo server restarts, users will have to re-login
- Might have to increase argo server memory if the users have extremely large claims

Fixes #10153 #9530 

# Before
Screenshot 2023-04-15 at 11 17 50

# After
Screenshot 2023-04-15 at 10 52 31

### DIFF
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -8,6 +8,8 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"github.com/google/uuid"
+	"k8s.io/utils/lru"
 	"net/http"
 	"strings"
 	"time"
@@ -49,18 +51,20 @@ var _ Interface = &sso{}
 type Config = config.SSOConfig
 
 type sso struct {
-	config          *oauth2.Config
-	issuer          string
-	idTokenVerifier *oidc.IDTokenVerifier
-	httpClient      *http.Client
-	baseHRef        string
-	secure          bool
-	privateKey      crypto.PrivateKey
-	encrypter       jose.Encrypter
-	rbacConfig      *config.RBACConfig
-	expiry          time.Duration
-	customClaimName string
-	userInfoPath    string
+	config              *oauth2.Config
+	issuer              string
+	idTokenVerifier     *oidc.IDTokenVerifier
+	httpClient          *http.Client
+	baseHRef            string
+	secure              bool
+	privateKey          crypto.PrivateKey
+	encrypter           jose.Encrypter
+	rbacConfig          *config.RBACConfig
+	expiry              time.Duration
+	customClaimName     string
+	userInfoPath        string
+	bearerTokenMap      *lru.Cache
+	compressBearerToken bool
 }
 
 func (s *sso) IsRBACEnabled() bool {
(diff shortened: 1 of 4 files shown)


terrytangyuan (commented):
Can you describe the issue or use cases you are trying to address?

server/auth/sso/sso.go
```diff
@@ -298,6 +305,11 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	value := Prefix + raw
+	if s.compressBearerToken {
+		compressedToken := Prefix + uuid.New().String()
+		s.bearerTokenMap.Add(compressedToken, value)
+		value = compressedToken
+	}
```
juliev0, line 312: I was a little confused by the name `compressedToken` and `compressBearerToken` since in my mind that's an actual compression algorithm but I'm okay if others are good with it. Otherwise, I would think something like `tokenRefererence` or maybe `reducedToken`.

server/auth/sso/sso.go
```diff
@@ -323,16 +335,23 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
 	c := &types.Claims{}
 	if err := tok.Claims(s.privateKey, c); err != nil {
+		s.bearerTokenMap.Remove(authorization)
 		return nil, fmt.Errorf("failed to parse claims: %v", err)
 	}
 
 	if err := c.Validate(jwt.Expected{Issuer: issuer}); err != nil {
+		s.bearerTokenMap.Remove(authorization)
 		return nil, fmt.Errorf("failed to validate claims: %v", err)
 	}
```
juliev0, line 356: I'm curious - what's the user experience if the user's uuid got purged from the LRU cache and they are passing in their UUID? Will they see the error: "failed to parse encrypted token" or will they simply be asked to log in again?

alexec (commented):
How work with 2 replicas?

basanthjenuhb:
> How work with 2 replicas?

Yeah, won't work with 2 replicas.
Sorry, my understanding was there would always be 1 replica.

server/auth/sso/sso.go
basanthjenuhb: Yeah, it would say failed to parse.
And they get brought back to login page

sarabala1979:
yes, it is a good point. it won't work for multiple replicas. I just had a call with @basanthjenuhb, Going back to the original solution like splitting payload into multiple cookies and merging it back, @basanthjenuhb is looking into it

juliev0:
@basanthjenuhb can you close this then?

PR #10923, opened 2023-04-15
docs: Add Swagger UI to the API reference page

changhc:
### Description
Currently the link in the [API reference page](https://argoproj.github.io/argo-workflows/swagger/) points to the API spec in JSON, which is not readable. This PR adds a simple HTML code snippet that renders Swagger UI with the spec to `swagger.md` so that it displays the API docs directly.

We can also put the HTML code in a separate file and simply make the link in the API reference page point to that file. I think it's more straightforward to embed Swagger UI like this.

One issue with Swagger UI is that it's slow when it loads information about a certain path or schema after a user clicks on it. This is probably because the API spec is too large. I tested generating Swagger UI in static HTML, but it also took some time for my browser to load a certain path.

### Test
`make docs`

Though I also updated Makefile to exempt `swagger.md` from checks because it has HTML code, which is not allowed by the linter.

### Screenshot

Peek 2023-04-15 10-31

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -618,7 +618,7 @@ docs/cli/argo.md: $(CLI_PKGS) go.sum server/static/files.go hack/cli/main.go
 .PHONY: docs-spellcheck
 docs-spellcheck: /usr/local/bin/mdspell
 	# check docs for spelling mistakes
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):
Thanks!

PR #10926, opened 2023-04-16
fix: Replace expressions with placeholders in resource manifest template. Fixes #10924

### LINKED ISSUE #10924: Unquoted expressions in WorkflowTemplate resource manifest fail validation
### Pre-requisites

When submitting a workflow template that has a 'complex'=(not directly resolvable) unquoted expression (see attached workflow's containerPort in the resource manifest) we get a validation error which is a false positive. It occurs because the expression remains unresolved which breaks the yaml syntax during the unmarshalling.

It is a valid concern because as we know all input parameters are strings  (or at least should be according to the schema), so there are cases when we must cast the string to a type which is expected by the applied resource's manifest (e.g: containerPort).
The current 'workaround' is to have the input parameter's value supplied as an int (unquoted):
```
        parameters:
          - name: intParam
            value: 20
 ```
So that in the resource manifest we can reference the input parameter directly and unquoted (i.e: `containerPort: {{inputs.parameters.intParam}}`), which works because it is not an expression and is replaced with a placeholder before yaml validation, but there are several issues with that:
- this is not valid according to the schema (https://github.com/argoproj/argo-workflows/blob/master/api/jsonschema/schema.json#L5803)
- sometimes we need more complex logic which can't be supplied as input parameter

The solution is quite simple - basically we should replace the unresolved expressions with placeholders as well, so that the yaml unmarshalling succeeds - just for validation (because the actual expression evaluation will happen later). Will have a PR submitted later today.

### Version

latest

```YAML
kind: WorkflowTemplate
metadata:
  generateName: valid-resource-
spec:
  entrypoint: whalesay
  templates:
    - name: whalesay
      inputs:
        parameters:
          - name: intParam
            value: '20'
          - name: strParam
            value: 'foobarbaz'
      outputs: {}
      metadata: {}
      resource:
        action: create
        manifest: |
          apiVersion: v1
          kind: Pod
          metadata:
            generateName: {{workflow.name}}-
          spec:
            restartPolicy: Never
            containers:
            - name: 'foo'
              image: docker/whalesay
              command: [cowsay]
              args: ["{{=sprig.replace("bar", "baz", inputs.parameters.strParam)}}"]
              ports:
              - containerPort: {{=asInt(inputs.parameters.intParam)}}
```

```text
# workflow controller doesn't print any errors (just returns 500: {"code":2,"message":"templates.whalesay.resource.manifest must be a valid yaml"})

# logs from workflow-server
time="2023-04-16T09:19:43.867Z" level=error msg="finished unary call with code Unknown" error="templates.whalesay.resource.manifest must be a valid yaml" grpc.code=Unknown grpc.method=CreateWorkflowTemplate grpc.service=workflowtemplate.WorkflowTemplateService grpc.start_time="2023-04-16T09:19:43Z" grpc.time_ms=0.626 span.kind=server system=grpc
```

```text
N/A
```

crowcrow:
Fixes #10924

### DIFF
--- a/test/e2e/testdata/workflow-template-with-resource-expr.yaml
+++ b/test/e2e/testdata/workflow-template-with-resource-expr.yaml
@@ -1,0 +1,33 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: workflow-template-with-resource-expr
+spec:
+  entrypoint: whalesay
+  templates:
+    - name: whalesay
+      inputs:
+        parameters:
+          - name: intParam
+            value: '20'
+          - name: strParam
+            value: 'foobarbaz'
+      outputs: {}
+      metadata: {}
+      resource:
+        action: create
+        setOwnerReference: true
+        manifest: |
+          apiVersion: v1
+          kind: Pod
+          metadata:
+            name: foo
+          spec:
+            restartPolicy: Never
+            containers:
+            - name: 'foo'
+              image: docker/whalesay
+              command: [cowsay]
+              args: ["{{=sprig.replace("bar", "baz", inputs.parameters.strParam)}}"]
+              ports:
+              - containerPort: {{=asInt(inputs.parameters.intParam)}}
--- a/test/e2e/workflow_template_test.go
+++ b/test/e2e/workflow_template_test.go
@@ -77,6 +77,19 @@ func (s *WorkflowTemplateSuite) TestSubmitWorkflowTemplateWorkflowMetadataSubsti
 		})
 }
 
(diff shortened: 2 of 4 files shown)


crowcrow:
[Test failure](https://github.com/argoproj/argo-workflows/actions/runs/4731763885/jobs/8397150006?pr=10926) doesn't look related to the changes in this PR

maxsxu (commented):
@crowcrow Thanks for this fix! I'm sure we've run into this problem many times.

Just give more details for reproduction: this issue only occurs when the WorkflowTemplate is submitted via the argo CLI or argo UI (to argo-server), not the kubectl (to kube-apiserver).

alexec (approved):

terrytangyuan (approved):

PR #4644, opened 2020-12-03
feat(server): Support email for SSO+RBAC. Closes #4612

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>
Closes #4612

Checklist:

* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).

Screenshot 2020-12-03 at 12 46 08

Issue #4612, opened 2020-11-29
Use Email for SSO RBAC with Google OIDC


avifried1:
# Summary

When using SSO auth mode with Google's OIDC, the issuer identifier (subject) is a unique user ID that isn't very useful and is hard to use for identifying users and managing RBAC permissions in a scalable way.

It will be more useful to use the user's email as the subject when using google as the OIDC provider.
per Google's [OIDC documentation](https://developers.google.com/identity/protocols/oauth2/openid-connect):

```json
{
  "iss": "https://accounts.google.com",
  "azp": "1234987819200.apps.googleusercontent.com",
  "aud": "1234987819200.apps.googleusercontent.com",
  "sub": "10769150350006150715113082367",
  "at_hash": "HK6E_P6Dh8Y93mRNtsDB1Q",
  "hd": "example.com",
  "email": "jsmith@example.com",
  "email_verified": "true",
  "iat": 1353601026,
  "exp": 1353604926,
  "nonce": "0394852-3190485-2490358"
}
```

# Use Cases

When google is set as the OIDC provider (issuer `https://accounts.google.com`). This is very useful for teams using G suite.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

alexec:
This should be easy for you to create a PR

```golang
type Claims struct {
	jwt.Claims
	Groups []string `json:"groups,omitempty"`
}
```

That needs an `Email` field, e.g.:


```golang
type Claims struct {
	jwt.Claims
        Email string `json:"email,omitempty"`
	Groups []string `json:"groups,omitempty"`
}
```

Also need to update the docs and your config to add `email` to `sso.groups`

avifried1:
> This should be easy for you to create a PR
> 
> ```go
> type Claims struct {
> 	jwt.Claims
> 	Groups []string `json:"groups,omitempty"`
> }
> ```
> 
> That needs an `Email` field, e.g.:
> 
> ```go
> type Claims struct {
> 	jwt.Claims
>         Email string `json:"email,omitempty"`
> 	Groups []string `json:"groups,omitempty"`
> }
> ```
> 
> Also need to update the docs and your config to add `email` to `sso.groups`

Will also be better to use Email, if available, in the `workflows.argoproj.io/creator` label

alexec:
I've created a dev build for you to test to see that it does what you need:

```
docker pull argoproj/argocli:fix-4612
```

[...]

PR #10927, opened 2023-04-17
feat(server): support name claim for RBAC SSO

agilgur5:
## Summary

Add `name` claim for SSO RBAC
- Closely follows #4644

## Details

- my organization uses a unique identifier in the `name` claim for SSO
  - without this change, we are unable to use the SSO identifier in order to authorize users to Argo
  - we can still use `groups` and other supported claims, but as `name` is the SSO identifier, it is commonly used for authorization

- it took quite a lot of debugging and trial-and-error to figure out why our `rbac-rule` expression wasn't working with `name`
  - eventually I found some helpful logs from the server:
    ```
    time="2023-04-13T21:34:30.243Z" level=info msg="Error while SSO Delegation" error="failed to evaluate rule: unable to evaluate expression 'name.name == 'agilgur5'': cannot fetch name from <nil> (1:6)\n | name.name == 'agilgur5'\n | .....^"
    ```
    - these logs showed that `name` was `nil`, which finally made me think that `name` might not be supported, even though we did add [`profile` to our `scopes`](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims)
    - from there I dove into the codebase and eventually found that `Claims` did not support / marshal `name`
      - and `RawClaim` is set to private, so I wasn't able to make use of that either unfortunately
    - from there I eventually found #4644 and #4612 that added support for `email`
      - this PR closely follows that one to support `name`

- also fix the versioning of `protobuf` in the `protoc` function
  - use 1.3.2 which matches what's `go install`ed by the `Makefile`
      
## Review Notes

**EDIT**: and then I got through all the errors and my employer's auto review, so nothing here anymore! yay!

## Follow-ups

There's a lot more `claims` in [the `profile` `scope`](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims) than just `name` that I would think various organizations might want depending on their SSO set-up. Exposing more fields or exposing `RawClaims` could help with that.
* I intentionally left this PR pretty small for this change that my organization really needs, but if you'd _also_ like to implement the above, I could contribute that too in a separate PR(s). Let me know what approach you might prefer.


agilgur5:
Not sure why one of the CLI's log tests is failing, don't think I changed anything there?

alexec (approved):
Great PR!

terrytangyuan (approved):

PR #10928, opened 2023-04-17
chore: Adjust gc_controller log to be more accurate. Fixes #10929

### LINKED ISSUE #10929: Adjust gc_controller deletion log to be more accurate
# Summary
Argo gc_controller [Delete function](https://github.com/argoproj/argo-workflows/blob/master/workflow/gccontroller/gc_controller.go#L218) has log `"Successfully deleted"` without double checking. It will be confused us when we see it but the workflow still exists.

What change needs making?
Adjust log to `Successfully request '%s' to be deleted`.

# Use Cases
When we are debugging and searching in logs, this log will confuse us to make clear what is going on.

When would you use this?
Debug
---

peterandluc:
- Say why you made your changes:
This log can not reflect the workflow which is to be deleted correctly. The call of Delete function returns nil does not mean the workflow has been successfully. We have seen it in production environment sometimes which is confused us.

- Say what changes you made:
Adjust the log msg.

- Say how you tested your changes:
Add finalizer and wait gc_controller to delete

Fixes #10929

### DIFF
--- a/workflow/gccontroller/gc_controller.go
+++ b/workflow/gccontroller/gc_controller.go
@@ -215,7 +215,7 @@ func (c *Controller) deleteWorkflow(ctx context.Context, key string) error {
 			return err
 		}
 	} else {
-		log.Infof("Successfully deleted '%s'", key)
+		log.Infof("Successfully request '%s' to be deleted", key)
 	}
 	return nil
 }


workflow/gccontroller/gc_controller.go
```diff
@@ -215,7 +215,7 @@ func (c *Controller) deleteWorkflow(ctx context.Context, key string) error {
 			return err
 		}
 	} else {
-		log.Infof("Successfully deleted '%s'", key)
+		log.Infof("Successfully request '%s' to be deleted", key)
```
terrytangyuan, line 218: This would depend on `WF_DEL_PROPAGATION_POLICY`

peterandluc: @terrytangyuan Hi, thank you for comments. I think this issue is not related to k8s DeletionPropagation. The log `Successfully deleted` should depend on the real status of workflow, which is whether existed or deleted. In our case, our workflow has many different finalizers attached by different services, any existing finalizer will prevent the deletion, which will be confused by this log msg.

terrytangyuan (approved):

PR #10955, opened 2023-04-21
feat: argo sso group field change to configurable param.

Shanjian2033:
Code Change For: feat: argo sso group field change to configurable param.


agilgur5:
Replaced by #10957

PR #10967, opened 2023-04-23
feat(exit_handler): Make current step outputs available to exit handler expressions

iainlane:
The [documentation][docs] currently states

> You can use outputs from previous steps, just not the one you're hooking into. If you'd like to use outputs create an exit handler instead - all the status variable are available there so you can still conditionally decide what to do.

This implies that expressions will have outputs from the current step available too. They don't though - the scope for an exit hook does not have the outputs in it.

It should be possible to do this for exit hooks. They only run when their step is finished. When that is the case, their outputs will be available. Add current step outputs to the exit hook's context. This allows much more powerful exit hook expressions to be written, such as sending a message conditionally if a certain output is set. For example, imagine a step which opens a pull request which may be a draft. That would be an output parameter, and we could conditionally run an exit hook to post a status update in that case.

Beef up the tests to verify that previous step outputs are there also.

(All of this applies to DAG tasks too.)

[docs]: https://argoproj.github.io/argo-workflows/lifecyclehook/#unsupported-conditions

### DIFF
--- a/workflow/controller/dag.go
+++ b/workflow/controller/dag.go
@@ -265,6 +265,8 @@ func (woc *wfOperationCtx) executeDAG(ctx context.Context, nodeName string, tmpl
 				if taskNode.Completed() {
 					// Run the node's onExit node, if any. Since this is a target task, we don't need to consider the status
 					// of the onExit node before continuing. That will be done in assesDAGPhase
+					prefix := fmt.Sprintf("tasks.%s", task.Name)
+					woc.buildLocalScope(scope, prefix, taskNode)
 					_, _, err := woc.runOnExitNode(ctx, dagCtx.GetTask(taskName).GetExitHook(woc.execWf.Spec.Arguments), taskNode, dagCtx.boundaryID, dagCtx.tmplCtx, "tasks."+taskName, scope)
 					if err != nil {
 						return node, err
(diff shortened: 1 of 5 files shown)


iainlane:
I'm sure this isn't ready yet, but wanted to open up for review comments early on the idea itself - in case we're way off base - and any specifics which I might need to take care of.

Have tried to motivate it a little in the description, but essentially we've found conditional exit hooks to be a really powerful tool in the workflows we're developing - and there are a few ways that they could potentially be better still. Being able to run them conditionally based on their step's output would be a good first step.

In terms of this PR, one thing I had wanted to add is tests to make sure artifacts can still be fetched given the refactoring I did in the code which sets those up for exit handlers. But I couldn't figure out how to make the test workflows actually fetch from a mocked artifact repository, unfortunately.

iainlane:
unstale

workflow/controller/dag.go
```diff
@@ -407,6 +412,8 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
 
 		if node.Completed() {
 			// Run the node's onExit node, if any.
+			prefix := fmt.Sprintf("tasks.%s", task.Name)
+			woc.buildLocalScope(scope, prefix, node)
 			hasOnExitNode, onExitNode, err := woc.runOnExitNode(ctx, task.GetExitHook(woc.execWf.Spec.Arguments), node, dagCtx.boundaryID, dagCtx.tmplCtx, "tasks."+taskName, scope)
 			if hasOnExitNode && (onExitNode == nil || !onExitNode.Fulfilled() || err != nil) {
```
juliev0, line 418: just want to make sure - I noticed `woc.runOnExitNode` is also called on line 562. Is that part good to go as far as the scope?

juliev0:
Thanks for doing this!

workflow/controller/exit_handler.go
```diff
@@ -24,43 +32,75 @@ func (woc *wfOperationCtx) runOnExitNode(ctx context.Context, exitHook *wfv1.Lif
 		execute := true
 		var err error
 		if exitHook.Expression != "" {
-			execute, err = argoexpr.EvalBool(exitHook.Expression, env.GetFuncMap(template.EnvMap(woc.globalParams.Merge(scope.getParameters()))))
+			scopeParams := scope.getParameters()
+			mergedParams := woc.globalParams.Merge(scopeParams)
```
sarabala1979, line 36: Does Scope not contains the global param? if yes, we should add it caller function. Scope should have all the paramers

PR #10971, opened 2023-04-23
docs: mention defaults for PodGC `strategy` and `volumeClaimGC`

agilgur5:
## Summary

Mention defaults for PodGC `strategy` and `volumeClaimGC` in the Fields reference page
- Given that these are for garbage collection (GC), they're kind of important to know for operators 😅 

## Details

- the defaults were previously not mentioned in the docs, so I had to look through the source code to find these:
  - `volumeClaimGC`: https://github.com/argoproj/argo-workflows/blob/1ba8d317113509c48d9a7959f2bee6ee2e812802/pkg/apis/workflow/v1alpha1/workflow_types.go#L484
  - PodGC `strategy`: https://github.com/argoproj/argo-workflows/blob/1ba8d317113509c48d9a7959f2bee6ee2e812802/pkg/apis/workflow/v1alpha1/workflow_types.go#L1050

I found this out the hard way when I realized our clusters had a bunch of unused PVCs laying around 😬 

## Follow-ups

Should there be a separate column in the tables for all the defaults? I feel like that would be useful and would catch missing ones like this. Similar to how Helm charts often do this in their `Values` reference

### DIFF
--- a/api/jsonschema/schema.json
+++ b/api/jsonschema/schema.json
@@ -5831,7 +5831,7 @@
           "description": "LabelSelector is the label selector to check if the pods match the labels before being added to the pod GC queue."
         },
         "strategy": {
-          "description": "Strategy is the strategy to use. One of \"OnPodCompletion\", \"OnPodSuccess\", \"OnWorkflowCompletion\", \"OnWorkflowSuccess\"",
+          "description": "Strategy is the strategy to use. One of \"OnPodCompletion\", \"OnPodSuccess\", \"OnWorkflowCompletion\", \"OnWorkflowSuccess\". If unset, does not delete Pods",
           "type": "string"
         }
       },
@@ -6956,7 +6956,7 @@
       "description": "VolumeClaimGC describes how to delete volumes from completed Workflows",
       "properties": {
         "strategy": {
-          "description": "Strategy is the strategy to use. One of \"OnWorkflowCompletion\", \"OnWorkflowSuccess\"",
+          "description": "Strategy is the strategy to use. One of \"OnWorkflowCompletion\", \"OnWorkflowSuccess\". Defaults to \"OnWorkflowSuccess\"",
           "type": "string"
         }
       },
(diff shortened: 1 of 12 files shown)


terrytangyuan (commented):
Please fix the codegen. Thanks

agilgur5:
I'm out on vacation for the next week, so I'll get back to this late next week.

I also did not realize this part of the docs had codegen but that makes sense given it's a reference!

caelan-io (approved):
Thanks for adding clarity here. I'm sure this will help other users as well!

Suggestions that we can address in a separate PR if not this one:

Since the Examples are highly referenced, we might also consider adding a similar comment to these examples, perhaps in the realm of lines 14-16 where comments already exist:
- https://github.com/argoproj/argo-workflows/blob/master/examples/pod-gc-strategy.yaml
- https://github.com/argoproj/argo-workflows/blob/master/examples/pod-gc-strategy-with-label-selector.yaml

We might also consider mentioning the PodGC default here under the Cost Optimization doc (i.e. "By default, pods are not deleted" or similar): https://argoproj.github.io/argo-workflows/cost-optimisation/#limit-the-total-number-of-workflows-and-pods

agilgur5:
@terrytangyuan fixed codegen! (took a bit for me to get that running locally again 😅 )

@caelan-io I added a second PR for the examples #11154 .
re: the Cost Optimization doc, I'd also like to mention `volumeClaimGC` there (as I address that in this PR as well), but the section/list is about "total number of workflows" and `podGC` already feels a little out of place with that naming. I can add `volumeClaimGC` to the list as well, but perhaps there's some better naming/titling to be used there. It feels like maybe `podGC` and `volumeClaimGC` belong in a separate section entirely? Any thoughts on that?

terrytangyuan (approved):
Thanks!

agilgur5:
Added a third PR for the Cost Optimization doc: #11293

PR #10975, opened 2023-04-25
fix: Match retryConditional example with its summary

tico24:
The retry-conditional example summary says it'll try a maximum of 10 times, but the limit was set to 3.

This matches the limit to the description.


maxsxu (approved):
LGTM. Very meticulous observation!

@tico24 But I'm thinking would it be better to update the summary to *...limit of 3...*, since it'll hardly reach *10 limit* in this example.

terrytangyuan (approved):

PR #10981, opened 2023-04-25
ci: add PR title conventional commits checker

maxsxu:
### Motivation
To provide the same contributing settings for Argo projects. since other projects like the argo-rollouts and argo-cd have already adopted the PR title checker to conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,0 +1,17 @@
+name: PR
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+
+jobs:
+  title-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/pr.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - opened
+      - edited
+      - reopened
+      - synchronize
+
+jobs:
+  title-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
```
terrytangyuan, line 15: Where is the configuration? Also need explicitly set permissions similar to ArgoCD https://github.com/argoproj/argo-cd/blob/master/.github/workflows/pr-title-check.yml

maxsxu: @terrytangyuan No additional config file is required for this action `amannn/action-semantic-pull-request`.

Here are more details about comparison of the PR title checker:

- the argo-cd is using `thehanimo/pr-title-checker`, which requires additional conf files. See https://github.com/argoproj/argo-cd/blob/master/.github/workflows/pr-title-check.yml
- the argo-rollouts is using `amannn/action-semantic-pull-request` , which doesn't require additional conf files and it's more popular and active maintained. See https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml

Let's adopt the argo-rollouts one, but make it simpler inspired by https://github.com/microsoft/SynapseML/blob/master/.github/workflows/check-semantic-prs.yaml

terrytangyuan (approved):

PR #10987, opened 2023-04-26
feat: Add lastRetry.message

### LINKED ISSUE #9937: Add message field to lastRetry for conditional retries
# Summary

In https://argoproj.github.io/argo-workflows/retries/#conditional-retries, `lastRetry` cannot access workflow error/failure messages such as `Pod was terminated due to imminent node shutdown`. This request is for adding a `message` on `lastRetry` containing this error message.

# Use Cases

In https://github.com/argoproj/argo-workflows/issues/9880, we captured preemptions by setting `OnTransientError` and `TRANSIENT_ERROR_PATTERN`. However, we would also like to retry `OnError`. Using conditional retries, we cannot do both since `lastRetry` does not include a `message` field - e.g.

```
retryStrategy:
        limit: "3"
        expression: "lastRetry.status == 'Error' || lastRetry.message.contains('imminent node shutdown')"
```

---

Joibel:
As requested in #9937 add a message field to lastRetry to allow fine grained control of retries.

Tested with unit test and a manually with a simple workflow which conditionally retried on `lastRetry.message=="some words"`

Fixes #9937

### DIFF
--- a/docs/retries.md
+++ b/docs/retries.md
@@ -63,7 +63,7 @@ access to the following variables:
 - `lastRetry.exitCode`: The exit code of the last retry, or "-1" if not available
 - `lastRetry.status`: The phase of the last retry: Error, Failed
 - `lastRetry.duration`: The duration of the last retry, in seconds
-- `lastRetry.message`: The message output from the last retry
+- `lastRetry.message`: The message output from the last retry (available from version 3.5)
 
 If `expression` evaluates to false, the step will not be retried.
 
--- a/docs/retries.md
+++ b/docs/retries.md
@@ -63,6 +63,7 @@ access to the following variables:
 - `lastRetry.exitCode`: The exit code of the last retry, or "-1" if not available
 - `lastRetry.status`: The phase of the last retry: Error, Failed
 - `lastRetry.duration`: The duration of the last retry, in seconds
+- `lastRetry.message`: The message output from the last retry
 
 If `expression` evaluates to false, the step will not be retried.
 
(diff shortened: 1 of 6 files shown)


docs/retries.md
```diff
@@ -63,6 +63,7 @@ access to the following variables:
 - `lastRetry.exitCode`: The exit code of the last retry, or "-1" if not available
 - `lastRetry.status`: The phase of the last retry: Error, Failed
 - `lastRetry.duration`: The duration of the last retry, in seconds
+- `lastRetry.message`: The message output from the last retry
```
tico24, line 66: Ah the fun part of the docs is they aren't versioned. So we'll have to put some sort of "3.5+" message onto this.

docs/variables.md
```diff
@@ -189,8 +189,9 @@ When using the `expression` field within `retryStrategy`, special variables are
 | Variable | Description|
 |----------|------------|
 | `lastRetry.exitCode` | Exit code of the last retry |
-| `lastRetry.Status` | Status of the last retry |
-| `lastRetry.Duration` | Duration in seconds of the last retry |
+| `lastRetry.status` | Status of the last retry |
+| `lastRetry.duration` | Duration in seconds of the last retry |
+| `lastRetry.message` | Message output from the last retry |
```
tico24, line 194: same docs-versioning issue here.

docs/retries.md
Joibel: I've added `(available from version 3.5)` to this

docs/variables.md
Joibel: As above

terrytangyuan (approved):

Issue #10872, opened 2023-04-11
Don't Crash Controller when an Invalid Workflow Object is present


hbrewster-splunk:
# Summary

When running an argo as a platform, there is always the potential for users to create invalid Workflow objects. 

Usually, invalid CRDs are detected by the controller, however in some cases they're not. For example, the following Workflow is invalid whenever an invalid parameter "execution_label" is provided since when used in workflowMetadata, it must not contain special chars etc. 

`
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: name
  namespace: default
spec:
  activeDeadlineSeconds: 65000
  arguments:
    - name: execution_label
      value: some/special/char
.....
  workflowMetadata:
    labelsFrom:
      execution_label:
        expression: workflow.parameters.execution_label
`

When this Workflow is submitted, the controller spits out a (valid) error: 
`
time="X" level=warning msg="Error updating workflow: Workflow.argoproj.io \"X\" is invalid: metadata.labels: Invalid value: \"some/special/char\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?') Invalid" namespace=default workflow=X
`

and subsequently enters a crashLoopBackOff with 'Workflow never reconciled" error. 

I would prefer if the controller would not enter a crashLookBackOff, or have the option not to, since this disrupts other users of the controller. 

# Use Cases

When would you use this?

Any platform level usage of Argo Workflows 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

terrytangyuan:
Could you share a reproducible example? I got this error instead.

```
cannot unmarshall spec: cannot restore struct from: slice
```

jessesuen:
Hmmm, I thought we could deal with this. We use an unstructured informer that ignores/fails malformed workflows. See:

https://github.com/argoproj/argo-workflows/blob/2edf2cf17f0ddfafff12da869e9524d34403714e/workflow/util/util.go#L55-L60

Before this, a bad workflow would cause a reconciliation hot loop:
https://github.com/kubernetes/kubernetes/issues/57705
https://github.com/argoproj/argo-workflows/issues/632

[...]

PR #10995, opened 2023-04-27
fix: Validate label values from workflowMetadata.labels to avoid controller crash

terrytangyuan:
Fixes https://github.com/argoproj/argo-workflows/issues/10872.


rbreeze (approved):
LGTM

PR #10999, opened 2023-04-28
feat: Add functionality for nix. Fixes #10998

### LINKED ISSUE #10998: [Maintainer quality of life improvement] - Add rudimentary support for Nix.
# Summary

The dependencies in Argo Workflows are complex and it is hard to get a reproducible build and in my past experience it was hard to get a build at all. 
The Makefile is heavily opinionated and takes on the role of a task runner, build tool and a package manager. Migrating some of it's functionality away to a more reproducible setup will make developing easier for developers. 

# Use Cases
I would use this as my daily development environment. 

I actually have an existing flake now in our [fork](https://github.com/pipekit/argo-workflows/pull/21), I spent the day porting it over to the upstream repo, so not only will this work help us, it should help even more developers to get started on Argo. 

# Why Nix? 
https://serokell.io/blog/what-is-nix
TL;DR 
The claims (which I also agree with are) are
  1. That the Filesystem Hierarchy Standard is fundamentally incompatible with reproducibility 
  2. Docker is a band-aid over this.

isubasinghe:
Please see #10998 for a detailed comment regarding why this is needed. 

This is a change from our [fork](https://github.com/pipekit/argo-workflows/pull/21) that I ported to the upstream repo because it works quite well. 

The steps to run Argo now are quite simple, and guaranteed to work exactly the same on every machine. 
Not only this, the entire Argo build system is isolated from the rest of the user's filesystem. 

 1. modify the hosts file according to [this](https://argoproj.github.io/argo-workflows/running-locally/), don't worry about the other instructions. 
  2. setup a k8s cluster, k3d is the recommended solution here. 
  3. Install [Nix](https://nixos.org/download.html) 
  4. Run "nix develop" (you will have to use the experimental feature flags on a fresh install, the features you need are "nix-command" and "flakes").
  5. Run "devenv up".


hack/swagger/secondaryswaggergen.go
isubasinghe, line 1: This file changed due to gofmt.

server/static/files.go
isubasinghe, line 1: We need a basic stub committed so that it will compile, previously the Makefile copied over the files.go.stub to files.go

flake.nix
```diff
@@ -0,0 +1,379 @@
...
+                    pythonEnv
+                    clang-tools
+                    protobuf
+                    myyarn
+                  ];
+                  enterShell = ''
+                    ./hack/port-forward.sh;
+                    ./hack/free-port 9090;
+                    ./hack/free-port 2746;
+                    ./hack/free-port 8080;
```
isubasinghe, line 351: you forgot the .sh extension here.

juliev0 (approved):

README.md
```diff
@@ -45,6 +45,19 @@ Argo is a [Cloud Native Computing Foundation (CNCF)](https://cncf.io/) hosted pr
...
+Argo Workflows has some basic support for Nix which is enough to get Argo Workflows up and running with minimal effort. 
+Here are the steps to follow: 
+  1. Modify the hosts file according to [this](https://argoproj.github.io/argo-workflows/running-locally/), don't worry about the other instructions. 
+  2. Setup a k8s cluster, k3d is the recommended solution here. 
+  3. Install [Nix](https://nixos.org/download.html).
+  4. Make a directory in the argo workflows folder called ".devenv".
+  5. Run "nix develop --extra-experimental-features nix-command --extra-experimental-features flakes" (you can add this as a default in your nix.conf file).
+  6. Run "devenv up". 
+
+As I said, this is still fairly barebones at the moment, any feature in the Makefile not mentioned here is excluded at the moment.
```
terrytangyuan, line 59: Remove "I"s and any personal language since this is a publicly facing README

README.md
```diff
@@ -45,6 +45,19 @@ Argo is a [Cloud Native Computing Foundation (CNCF)](https://cncf.io/) hosted pr
...
+## Try Argo using Nix
+Nix is a package manager / build tool which focusses on reproducible build environments.
+Argo Workflows has some basic support for Nix which is enough to get Argo Workflows up and running with minimal effort. 
+Here are the steps to follow: 
+  1. Modify the hosts file according to [this](https://argoproj.github.io/argo-workflows/running-locally/), don't worry about the other instructions. 
+  2. Setup a k8s cluster, k3d is the recommended solution here. 
+  3. Install [Nix](https://nixos.org/download.html).
+  4. Make a directory in the argo workflows folder called ".devenv".
+  5. Run "nix develop --extra-experimental-features nix-command --extra-experimental-features flakes" (you can add this as a default in your nix.conf file).
+  6. Run "devenv up".
```
terrytangyuan, line 57: Maybe move to a separate doc so this root-level README can be concise

flake.nix
```diff
@@ -0,0 +1,382 @@
...
+              };
+              vendorHash = "sha256-+B3cAuV+HllmB1NaPeZitNpX9udWuCKfDFv+mOVHw2Y=";
+              doCheck = false;
+            };
+
+            staticfiles = pkgs.buildGoModule {
+              pname = "staticfiles";
+              version = "0.0.1"; # no official version
+              src = pkgs.fetchFromGitHub {
+                owner = "isubasinghe";
```
terrytangyuan, line 280: Remove reference to personal repos

flake.nix
```diff
@@ -0,0 +1,382 @@
...
+          packages = {
+            ${package.name} = pkgs.buildGoModule {
+              pname = package.name;
+              inherit (package) version;
+              inherit src;
+              vendorSha256 = "sha256-1ivPt7m65B89jOTlYA21AcrW6OZeMlgpoPZd9Lw2Ohc=";
+              doCheck = false;
+            };
+
+            mockery = pkgs.buildGoModule rec {
```
terrytangyuan, line 167: Can these be automatically updated since we have versions in other places too or somehow reuse existing Makefile?

terrytangyuan (commented):
There are a lot of new files at top-level. Is that necessary?

isubasinghe:
> There are a lot of new files at top-level. Is that necessary?

I believe I can probably refactor away the node-env.nix and node-package.nix by bringing some of that into flake.nix. 
The others, I do need them.

flake.nix
isubasinghe: There is `nix flake update` but nothing to update them to exactly the version in the Makefile. 
We can't really reuse the Makefile, Nix is designed to be completely sealed off, this is really it's core advantage. Because it manages dependencies itself, it can guarantee that the build will be reproducible on any machine whether it be a Ubuntu, Fedora or even 20 years in the future.

flake.nix
isubasinghe: I definitely agree that this is generally a bad idea, unfortunately the repo is dead. https://github.com/bouk/staticfiles. 

The current method to install this dependency is actually quite dangerous: https://github.com/argoproj/argo-workflows/blob/master/Makefile#L164. 

It uses an sha abbreviation instead of the full hash, on top of this it downloads off a random URL. 
We should get rid of this ASAP as it opens up the codebase to a code injection attack. 

I needed to fork his code because it was missing a [go.mod](https://github.com/isubasinghe/staticfiles/blob/master/go.mod). 

downloading from my repo is not great, but still safer than what the Makefile does. 

I will create an issue to get rid of this dep all together.

README.md
```diff
@@ -45,6 +45,19 @@ Argo is a [Cloud Native Computing Foundation (CNCF)](https://cncf.io/) hosted pr
 
 ![Screenshot](docs/assets/screenshot.png)
 
+## Try Argo using Nix
+Nix is a package manager / build tool which focusses on reproducible build environments.
+Argo Workflows has some basic support for Nix which is enough to get Argo Workflows up and running with minimal effort. 
+Here are the steps to follow: 
+  1. Modify the hosts file according to [this](https://argoproj.github.io/argo-workflows/running-locally/), don't worry about the other instructions. 
+  2. Setup a k8s cluster, k3d is the recommended solution here. 
+  3. Install [Nix](https://nixos.org/download.html).
```
juliev0, line 54: "Setup" -> "Set up"

conf.nix
```diff
@@ -0,0 +1,56 @@
+# most of these configuration options might not do anything yet. 
+# the plan is to gradually introduced more and more functionality into Nix
```
juliev0, line 2: "introduced" -> "introduce"

flake.nix
juliev0: Maybe we can add clear, hard-to-miss comments in the Makefile regarding how to go about modifying Nix, and possibly the other way around as well?

Does it make any sense to have CI compare the two?

isubasinghe:
Sorry @juliev0 for my late engagement on this issue, was quite busy with another bug (https://github.com/argoproj/argo-workflows/issues/11102).

flake.nix
isubasinghe: Yeah I think just documentation is really the best we can do unfortunately, I think this is fine anyway if we explicitly state in the docs that this feature is experimental and not supported (and should therefore not be relied on in production etc). This is very much a solution for developers only, to get them developing as quickly as possible. 

I can't really think of how we can compare them in the CI process unfortunately.

isubasinghe: I'll write some script to notify me on a Makefile change, so I can get to it ASAP. I have the capacity (and get paid) to do this kind of ad-hoc work anyway, so shouldn't be a problem.

server/static/files.go
```diff
@@ -0,0 +1,8 @@
+// File built without static files
+package static
+
+import "net/http"
+
+func ServeHTTP(http.ResponseWriter, *http.Request) {}
+
+func Hash(string) string { return "" }
```
terrytangyuan, line 8: Is this file used?

terrytangyuan (commented):
LGTM just left a minor comment. Thanks!

terrytangyuan:
Should we move this to /hack folder instead of creating a dedicated one in root directory.

server/static/files.go
isubasinghe: Yeah it is, the Makefile does this at runtime by copying files.stub.go, but since I don't rely on the Makefile this needs to be explicit. 

This doesn't change any of the existing behaviour though, just means the Makefile doesn't have to do the runtime copy.

isubasinghe:
@terrytangyuan I am happy to move it over if you think that is better, but I personally don't think it belongs there. 

The hack folder is specifically reserved for hacks that the Makefile uses, this is entirely separate from that so we shouldn't put it there imo. 

Is that alright? I'll chuck it in the hack folder otherwise. 
Thanks!

Makefile
```diff
@@ -613,31 +626,35 @@ docs/cli/argo.md: $(CLI_PKGS) go.sum server/static/files.go hack/cli/main.go
...
 
 .PHONY: docs-lint
 docs-lint: /usr/local/bin/markdownlint
 	# lint docs
 	markdownlint docs --fix --ignore docs/fields.md --ignore docs/executor_swagger.md --ignore docs/swagger.md --ignore docs/cli --ignore docs/walk-through/the-structure-of-workflow-specs.md
 
 /usr/local/bin/mkdocs:
-	python -m pip install mkdocs==1.2.4 mkdocs_material==8.1.9  mkdocs-spellcheck==0.2.1
+	python -m pip install mkdocs==1.2.4 mkdocs_material==8.1.9  mkdocs-spellcheck==0.2.1 # update this in Nix when upgrading it here
+
```
juliev0, line 657: thanks for doing this
do you think it makes sense to have any comments in the opposite direction?

terrytangyuan:
> The hack folder is specifically reserved for hacks that the Makefile uses, this is entirely separate from that so we shouldn't put it there IMO.

I feel like a `nix-files` folder might be overkill. If not in `hack`, I've seen folders like `dev` or `build` for this purpose which might be more suitable. We should move `.devcontainer` to the same folder as `nix-files` if dev container allows it to be in a different folder.

isubasinghe:
@terrytangyuan 
Yeah I think a dev folder works in that case, I just wanted to avoid overloading the meaning of the hack folder.  

I don't use devcontainers but I can ask someone who does (@Joibel ) to help me out as it does look like devcontainers does support the config in subdirectories: https://github.com/microsoft/vscode-remote-release/issues/2413

Makefile
isubasinghe: I made that change as well, thanks for the feedback!

Joibel:
> > The hack folder is specifically reserved for hacks that the Makefile uses, this is entirely separate from that so we shouldn't put it there IMO.
> 
> I feel like a `nix-files` folder might be overkill. If not in `hack`, I've seen folders like `dev` or `build` for this purpose which might be more suitable. We should move `.devcontainer` to the same folder as `nix-files` if dev container allows it to be in a different folder.

For autodiscovery of "this project has a devcontainer config" I think `.devcontainer` needs to stay where it is.

juliev0 (approved):

terrytangyuan (approved):

terrytangyuan:
@isubasinghe Can you help look into why the build is failing after merge? https://github.com/argoproj/argo-workflows/actions/runs/5145625809/jobs/9263514577

isubasinghe:
@terrytangyuan 

If it is really this PR, the only source of error I can think of is this -> https://github.com/argoproj/argo-workflows/pull/10999/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L1-R1

The only reason I can see is the Makefile changing, but that seems somewhat unlikely given the CI tests ran fine. Might just be a coincidence, looking into it now.

isubasinghe:
It could also be the files.go actually, will fix ASAP.

isubasinghe:
@terrytangyuan  here you go, it was missing a space :

https://github.com/argoproj/argo-workflows/pull/11160

PR #11016, opened 2023-04-30
fix: ui getPodName should use v2 format by default (fixes #11015)

### LINKED ISSUE #11015: ui: getPodName is using v1 by default
### Pre-requisites

I noticed that for some workflows the podName is wrongfully set as the nodeID (which is the v1 format).

This resulted in bad log streaming for running templates. (I thought it was related to https://github.com/argoproj/argo-workflows/issues/10124 but I think it's a pure UI issue).

Looking at the ts code I saw this condition:
https://github.com/argoproj/argo-workflows/blob/0c24ca1ba8a5c38c846d595770e16398f6bd84a5/ui/src/app/shared/pod-name.ts#L14

The `version` here is taken straight from the direct annotation of the workflow: https://github.com/argoproj/argo-workflows/blob/133b4a384364079e5c82580c4969ded636cbadf5/ui/src/app/archived-workflows/components/archived-workflow-details/archived-workflow-details.tsx#L56

I think that after https://github.com/argoproj/argo-workflows/commit/cc9d14cf0d60812e177ebb447181df933199b722 we should change the condition so that v2 would be used by default also for UI.

### Version

v3.4.5

```YAML
{
  "metadata": {
    "name": "fantastic-rhino",
    "namespace": "argo-workflows",
    "labels": {
      "example": "true"
    }
  },
  "spec": {
    "arguments": {
      "parameters": [
        {
          "name": "message",
          "value": "hello argo"
        }
      ]
    },
    "entrypoint": "argosay",
    "templates": [
      {
        "name": "argosay",
        "inputs": {
          "parameters": [
            {
              "name": "message",
              "value": "{{workflow.parameters.message}}"
            }
          ]
        },
        "container": {
          "name": "main",
          "image": "argoproj/argosay:v2",
          "command": [
            "/argosay"
          ],
          "args": [
            "echo",
            "{{inputs.parameters.message}}"
          ]
        }
      }
    ],
    "ttlStrategy": {
      "secondsAfterCompletion": 300
    },
    "podGC": {
      "strategy": "OnPodCompletion"
    }
  }
}
```

```text
Irrelevant
```

```text
Irrelevant
```

or-shachar:
I noticed that sometimes the UI shows the wrong pod name. Specifically for running pods.

I noticed that this change https://github.com/argoproj/argo-workflows/commit/cc9d14cf0d60812e177ebb447181df933199b722 made pod name format v2 to be default even when the annotation is missing in the backend, but this behavior was never coded into the UI code.

This tries to fix it.

Fixes #11015

### DIFF
--- a/ui/src/app/shared/pod-name.test.ts
+++ b/ui/src/app/shared/pod-name.test.ts
@@ -28,9 +28,12 @@ describe('pod names', () => {
     });
 
     test('getPodName', () => {
-        expect(getPodName(shortWfName, nodeName, shortTemplateName, nodeID, POD_NAME_V2)).toEqual('wfname-templatename-1454367246');
-        expect(getPodName(shortWfName, nodeName, shortTemplateName, nodeID, POD_NAME_V1)).toEqual(nodeID);
-        expect(getPodName(shortWfName, nodeName, shortTemplateName, nodeID, '')).toEqual(nodeID);
+        const v1podName = nodeID;
+        const v2podName = `${shortWfName}-${shortTemplateName}-${createFNVHash(nodeName)}`;
+        expect(getPodName(shortWfName, nodeName, shortTemplateName, nodeID, POD_NAME_V2)).toEqual(v2podName);
+        expect(getPodName(shortWfName, nodeName, shortTemplateName, nodeID, POD_NAME_V1)).toEqual(v1podName);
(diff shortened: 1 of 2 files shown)


terrytangyuan (commented):
cc @JPZ13

isubasinghe:
I am surprised that the UI has to reconstruct the podName, that is somewhat alarming. Anyway this looks fine.

isubasinghe (approved):

terrytangyuan (approved):

PR #11033, opened 2023-05-04
fix: Remove 401 Unauthorized when customClaimGroup retrieval fails, Fixes #11032

### LINKED ISSUE #11032: Usage of customGroupClaimName can lead to unexpected 401 unauthorized if custom claim is missing
### Pre-requisites

## Bug Description
Setting `customGroupClaimName` in the `sso` configuration leads to a 401 unauthorized exception if the specified claim is missing from the `idToken`

### Why this should not happen
This leads to an inconsistent behaviour compared to the usage of the `groups` claim that is consumed by default. When the latest is being used these 2 scenarios can happen:
- `groups` claim is present -> the various ServiceAccounts are evaluated in order of priority
- `groups` claim is absent -> same as above, but "rbac-rule" that checks the presence of elements in `groups` will always evaluate to false

When a custom claim is used instead this is what we're getting:
- `custom` claim is present -> the various ServiceAccounts are evaluated in order of priority
- `custom` claim is absent -> throws 401 unauthorized 

### Suggested solution
Do not throw an exception if the retrieval of the customClaimGroup fails but, instead, log the error/warning the evaluate the rbac-rules as false (as it is currently happening in for the default behaviour)

It looks like converting this snippet in `server/auth/sso/sso.go`
```go
	groups := c.Groups
	if s.customClaimName != "" {
		groups, err = c.GetCustomGroup(s.customClaimName)
		if err != nil {
			w.WriteHeader(401)
			return
		}
	}
```
to 
```go
	groups := c.Groups
	if s.customClaimName != "" {
		groups, err = c.GetCustomGroup(s.customClaimName)
		if err != nil {
			log.Warn(err)
		}
	}
```
could be a viable solution (I'll create a PR with these changes later on)

### Version

latest

```YAML
Unrelated to workflows, it's about authentication
```

```text
No relevant logs are returned
```

```text
No relevant logs are returned
```

vanny96:
Changed way customClaimGroups are handled in `server/auth/sso/sso.go` `HandleCallback`. 

Before: if failed to retrieve, throws 401 Unauthorized 
After: keeps `groups` nil, logs the error encountered

Fixes #11032

### DIFF
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -263,8 +263,7 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 	if s.customClaimName != "" {
 		groups, err = c.GetCustomGroup(s.customClaimName)
 		if err != nil {
-			w.WriteHeader(401)
-			return
+			log.Warn(err)
 		}
 	}
 


server/auth/sso/sso.go
```diff
@@ -263,8 +263,7 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 	if s.customClaimName != "" {
 		groups, err = c.GetCustomGroup(s.customClaimName)
 		if err != nil {
-			w.WriteHeader(401)
-			return
+			log.Warn(err)
```
terrytangyuan, line 266: This error handling seems inconsistent with other similar places above

vanny96: Hey terry, thank you for taking a look at this PR! 
I gave a more descriptive explanation in the linked issue, so if possible take a look there as well :) 
In general, the idea here is that by throwing a 401 exception when failing to retrieve a `customGroup` we are being inconsistent with the default behaviour (argo workflows doesn't throw 401 Unauthorized if the groups claim is missing)

This is causing a bunch of issues with my integration with Azure, which uses "roles" as a way to authorise users to an App Registration. If the roles claim is empty it shouldn't throw 401 Unauthorized, but evaluate it as empty (as it's currently the case for groups)

terrytangyuan: cc @liafizan @alexec to take a look as well

vanny96: Hey @terrytangyuan , any update on this? As we're about deploying argo workflows to production we'd really need this feature to work properly in order to grant granular access to users

juliev0: I see that there was a problem of inconsistency and this is addressing it. I'm gathering that in the case where customGroupClaimName is not defined, c.Groups may in fact be set to empty and nothing really happens if that's the case. So, I think this seems okay to me.

juliev0 (approved):

terrytangyuan (approved):

PR #11040, opened 2023-05-05
fix: handle panic from type assertion

juliev0:
Noticing occasional panics when the Workflow is no longer in the Informer cache and we get a deletion event due to a type assertion. The actual type is `cache.DeletedFinalStateUnknown`.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -827,7 +827,12 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context) 
 	wfc.wfInformer.AddEventHandler(
 		cache.FilteringResourceEventHandler{
 			FilterFunc: func(obj interface{}) bool {
-				return reconciliationNeeded(obj.(*unstructured.Unstructured))
+				un, ok := obj.(*unstructured.Unstructured)
+				if !ok {
+					log.Warnf("Workflow FilterFunc: '%v' is not an unstructured", obj)
+					return false
+				}
+				return reconciliationNeeded(un)
 			},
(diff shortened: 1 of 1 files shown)


workflow/controller/controller.go
```diff
@@ -827,7 +830,12 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
 	wfc.wfInformer.AddEventHandler(
 		cache.FilteringResourceEventHandler{
 			FilterFunc: func(obj interface{}) bool {
-				return reconciliationNeeded(obj.(*unstructured.Unstructured))
+				un, ok := obj.(*unstructured.Unstructured)
+				if !ok {
+					log.Warnf("Workflow FilterFunc: '%v' is not an unstructured", obj)
+					return false
+				}
+				return reconciliationNeeded(un)
```
maxsxu, line 838:
```suggestion
				return ok && reconciliationNeeded(un)
```

Would it be better to keep the same style as here? https://github.com/argoproj/argo-workflows/blob/dd08e95d5b7e77fa7fe4b2230d590c54ec8144f7/workflow/controller/controller.go#L870-L872

maxsxu (approved):

workflow/controller/controller.go
juliev0: I guess it depends on whether we think there's a benefit in having the Warn line - do you think there isn't?

maxsxu: yep, of course more logs are usually more helpful, I've no objection to this. 
maybe we can think about why other similar lines within informer don't print such logs? 🤔

Issue #10781, opened 2023-03-29
Unify the workflow + workflow archive UI/API


jessesuen:
# Summary

We have two categories/tabs of workflows when presenting them in our UI:
* in-cluster (only in k8s cluster)
* archived to a database

Depending on the archive configuration, the ttl strategy, and whether or not workflows are completed or not, a workflow may appear in the in-cluster list, the archive list, or even both.

A complaint among users is that end users somehow need to understand where to find their workflow in either of these two places. But they don't necessarily understand or care where to look.

Contrast this to just about any other batch processing system, which makes no distinction between something that is live vs. archived. You just arrive at a list of your recent runs, and there is a single page where you can find all your jobs.

The feature that would benefit the UX, is providing a single unified table where both in-cluster workflows and workflows from the archive can be seen in a single place.

The technical challenge is that archived workflows must be retrieved by UID (but can be searched by name), whereas live workflows must be retrieved by name.

One way to potentially mitigate this is: give an option such that a workflow name can be **_a primary/unique value in the database_**. It would be the responsibility of the end user, to name their workflows with a name that would not collide with a previous run. And if there was a collision, the latest one would override the previous.

# Use Cases

Any users that making use of the persistence feature would likely prefer this mode of operation.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

jessesuen:
One key thing about this feature is that I consider this more of an Argo API Server feature rather than a UI feature. It's not just about unifying the UI view. It's also about unifying the API itself, so things like RBAC, API requests, etc... are consistent on whether you are getting a workflow in-cluster or from the archive.

Another way to put it, it should be possible to write the front-end or CLI in a way where it does not know if workflows are archived or not.

jessesuen:
Here is another reason why needs to be addressed in the API layer. Below is a valid complaint about what an application needs to deal with the argo API:

> * We need to check the status of workflows that may be running or may have been archived.
> * We call the k8s api and look for a workflow. If it exists we return the current running state.
> * If it does not exist we call the /api/v1/archived-workflows?namePrefix=WORKFLOW_NAME
> * If we find a workflow in that list we grab the UID and then make another api call to /api/v1/archived-workflows/UID and grab the finished status.
> 
> All that results in up to 3 api calls to return the status

[...]

PR #11055, opened 2023-05-08
feat: Support GetWorkflow regardless of its archival status

terrytangyuan:
This includes the backend changes for https://github.com/argoproj/argo-workflows/issues/10781.


server/workflow/workflow_server.go
```diff
@@ -590,8 +591,14 @@ func (s *workflowServer) getWorkflow(ctx context.Context, wfClient versioned.Int
 		return latest, nil
 	}
 	wf, err := wfClient.ArgoprojV1alpha1().Workflows(namespace).Get(ctx, name, options)
-	if err != nil {
-		return nil, sutils.ToStatusError(err, codes.Internal)
+	if wf == nil || err != nil {
```
sarabala1979, line 594: Do you think instead of merging two different functionality in one API, UI can make two API calls? So the API call will be very clear 1-on-1  functionality.

pkg/apiclient/workflowarchive/workflow-archive.proto
```diff
@@ -15,6 +15,7 @@ message ListArchivedWorkflowsRequest {
 message GetArchivedWorkflowRequest {
   string uid = 1;
   string namespace = 2;
+  string name = 3;
```
sarabala1979, line 18: I would recommend creating a new API and deprecating the existing one for supporting backward capability.

sarabala1979 (changes requested):

server/workflow/workflow_server.go
terrytangyuan: The original purpose is to address use case mentioned in https://github.com/argoproj/argo-workflows/issues/10781 so that users only have to make one API call.

pkg/apiclient/workflowarchive/workflow-archive.proto
sarabala1979: Is it going to be a breaking change? Is there a way we can support this feature with backward compatibility?

persist/sqldb/null_workflow_archive.go
```diff
@@ -29,7 +29,7 @@ func (r *nullWorkflowArchive) CountWorkflows(string, string, string, time.Time,
 	return 0, nil
 }
 
-func (r *nullWorkflowArchive) GetWorkflow(string) (*wfv1.Workflow, error) {
```
sarabala1979, line 32: @terrytangyuan Here is the API changes

pkg/apiclient/workflowarchive/workflow-archive.proto
jessesuen: The change would not be considered breaking since introducing fields would not break any existing behavior. We're introducing a new `name` field as part of the existing request to allow users to retrieve workflows by name/namespace, but existing users who are retrieving workflow by UID would not be affected.

persist/sqldb/workflow_archive.go
```diff
@@ -257,14 +257,43 @@ func namePrefixClause(namePrefix string) db.Cond {
...
+				From(archiveTableName).
+				Where(r.clusterManagedNamespaceAndInstanceID()).
+				And(namespaceEqual(namespace)).
+				And(nameEqual(name)).
+				One(total)
+			if err != nil {
+				return nil, err
+			}
+			if int64(total.Total) > 1 {
+				return nil, fmt.Errorf("there are more than 1 workflows with the same name and namespace")
```
jessesuen, line 284: Suggest providing more detail:

`found %d archived workflows found with namespace/name: %s/%s`

persist/sqldb/workflow_archive.go
```diff
@@ -257,14 +257,43 @@ func namePrefixClause(namePrefix string) db.Cond {
...
+			}
+			err = r.session.
+				Select("workflow").
+				From(archiveTableName).
+				Where(r.clusterManagedNamespaceAndInstanceID()).
+				And(namespaceEqual(namespace)).
+				And(nameEqual(name)).
+				One(archivedWf)
+		} else {
+			return nil, fmt.Errorf("both name and namespace are required if uid is not specified")
```
jessesuen, line 294: Is there a way to return an error that would cause us to return 400 Bad Request as an http response?

jessesuen (commented):
@terrytangyuan how do we address the need for a unified list API? It's one thing to GET a workflow by namespace/name from either etcd or archive, but don't we want to improve the UX such that archived workflows could also be presented in same list view?

pkg/apiclient/workflowarchive/workflow-archive.proto
sarabala1979: Thanks, @jessesuen Just had a discussion with @terrytangyuan. I request to test the using current SDK with the new Argo server(signature changes) whether it passes the "" for new arguments or user needs to update the new SDK.

terrytangyuan: I tested this with existing SDK against the new Argo Server and existing behavior preserves. Users can still use only UID to retrieve the archived workflow.

persist/sqldb/workflow_archive.go
terrytangyuan: Yes, see my latest commit.

persist/sqldb/workflow_archive.go
terrytangyuan: Great suggestion.

persist/sqldb/workflow_archive.go
terrytangyuan: The mapping can be found in https://github.com/argoproj/argo-workflows/blob/master/server/utils/errors.go

sarabala1979 (approved):

PR #11070, opened 2023-05-11
fix: allow azure blobs to not exist when deleting

roelarents:
Fixes n/a

### Motivation

The docs [say](https://github.com/argoproj/argo-workflows/blob/master/docs/walk-through/artifacts.md?plain=1#L244) about artifact garbage collection:

> Artifact already having been deleted which is not considered a failure

However it is considered a failure at the moment. (At least for azure blobs.)

### Modifications

Changed the azure implementation of the `ArtifactDriver` to swallow `StorageErrorCodeBlobNotFound` errors from the azure client when `Delete`ing an artifact.

### Verification

I didn't test it yet.


terrytangyuan (commented):
Looks good. Could you test it and verify the changes?

Dirc:
Hi @terrytangyuan, I have tested the change and added a small fix. This fixes the bug for us.

terrytangyuan (approved):
Thank you!

PR #11074, opened 2023-05-11
fix: Support inputs for inline steps templates

iainlane:
Inline templates don't support inputs in steps templates, but they work for DAGs. This brings the feature up to parity.

Previously we validated the whole `stepGroup` at once, but I don't see how to do that with inputs on inline templates: given the scope there we can't simply add those params to the context. So I've switched to validating each step individually.


workflow/validate/validate.go
```diff
@@ -932,20 +932,30 @@ func (ctx *templateValidationCtx) validateSteps(scope map[string]interface{}, tm
...
+					stepScope["inputs.parameters."+p.Name] = placeholderGenerator.NextPlaceholder()
+				}
+			}
+
+			err = resolveAllVariables(stepScope, ctx.globalParams, string(stepBytes), workflowTemplateValidation)
+			if err != nil {
+				return errors.Errorf(errors.CodeBadRequest, "templates.%s.steps %s", tmpl.Name, err.Error())
+			}
 
-		for _, step := range stepGroup.Steps {
```
sarabala1979, line 946: Why do we need to remove this loop? will Code iterate all steps in stepGroup?

iainlane: It's not this loop that _needed_ to be removed, it was the serialization of the `stepGroup` that I removed a few lines above - the input parameters for the inline template aren't in the scope and so validation fails. They can't just be added to the overall scope or they would be visible to all steps in the group (AFAIK).

And then once that has been done you have two loops right next to each other looping over the same slice (`stepGroup.Steps`) - it seems to make sense that you should merge them then.

sarabala1979 (approved):

terrytangyuan (approved):

PR #11075, opened 2023-05-12
feat(workflow-controller): add timeout of WaitForCacheSync

sxllwx:
Fixes #TODO

### Motivation

add a timeout for Workflow-Controller `WaitForCacheSync`. if it cannot be completed within the specified time(1m) controller exit and output logs to help user locate the bug.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -137,6 +137,7 @@ const (
 	clusterWorkflowTemplateResyncPeriod = 20 * time.Minute
 	workflowExistenceCheckPeriod        = 1 * time.Minute
 	workflowTaskSetResyncPeriod         = 20 * time.Minute
+	waitForCacheSyncTimeout             = time.Minute
 )
 
 var (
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
What issue is this PR trying to address?

sxllwx:
ahha, thank you for your quick reply.

Sorry, I didn't provide the full relevant background.

This is what I encountered during the upgrade from v3.1.1 to v3.4.6. At that time, after the workflow was submitted, the workflow-controller did not drive the workflow for a long time, and there was no output in the log that could help locate the fault [our log level was not adjusted to a high level]. Finally, it was found through pprof because it was blocked in With `WaitForCacheSync`, although we control it through ctx, the timeout of waitForCacheSync. But unfortunately this ctx never ends. [The more fundamental reason is that we added a new CRD `workflowartifactgctasks`, but it is not installed in our cluster]

terrytangyuan:
>The more fundamental reason is that we added a new CRD workflowartifactgctasks, but it is not installed in our cluster

That should be included in the upgrade.

sxllwx:
Do we still need this PR to fix this problem?

terrytangyuan:
If missing `workflowartifactgctasks` CRD is the root cause, then we probably don't need it.

sxllwx:
Thanks for your review. I close this PR.

PR #11076, opened 2023-05-12
feat: Improve schema validation. Fixes #11021

### LINKED ISSUE #11021: Improve schema by utilizing "additionalProperties": false
# Summary

Many definitions are missing `additionalProperties`, and by default JSON Schema assumes additional properties are allowed if `additionalProperties` is missing. Consequently, mistyped properties are easily missed as validators considers those valid additional properties even if an object actually doesn't allow additional properties.

Basically only need to add `"additionalProperties": false` to each definition that doesn't permit additional properties.

Reference:
https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties
https://github.com/argoproj/argo-workflows/blob/master/api/jsonschema/schema.json

# Use Cases

Better validation, particularly catching mistyped properties early on.

---

Kitanotori:
See #11021 for details

`"additionalProperties": false` applied in following way:

```
yq -i -o=json '(.. | select(has("type") and .type == "object" and has("properties") and (has("additionalProperties") | not))).additionalProperties |= false' swagger.json

make
```

i.e., for each object that has "properties" but no existing "additionalProperties" set. In case there are no "properties" (i.e. no defined structure) the change would prevent any properties being passed, so wanted to avoid that.

### DIFF
--- a/api/jsonschema/schema.json
+++ b/api/jsonschema/schema.json
@@ -3,6 +3,7 @@
   "$schema": "http://json-schema.org/schema#",
   "definitions": {
     "eventsource.CreateEventSourceRequest": {
+      "additionalProperties": false,
       "properties": {
         "eventSource": {
           "$ref": "#/definitions/io.argoproj.events.v1alpha1.EventSource"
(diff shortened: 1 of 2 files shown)


terrytangyuan (commented):
These are generated files so you may want to modify the original source and run codegen.

Kitanotori:
@terrytangyuan Do you know which are all the files I should modify, or if there's some documentation regarding how to modify the definitions?

terrytangyuan:
I think the ones you are modifying belong in Argo Events repo.

tooptoop4:
dis 🔥 @nakamorichi

PR #11091, opened 2023-05-16
fix: Update Bitbucket SSH host key

toredash:
https://bitbucket.org/blog/ssh-host-key-changes


toredash:
related;

Had an issue where cloning resulted in:
```
level=fatal msg="artifact bitbucket-repository failed to load: failed to clone \"git@bitbucket.org:X/X.git\": ssh: handshake failed: knownhosts: key mismatch"
```

This started to occur after bitbucket _added_ their keys as of 15th of May (blogpost). This would not happen if go-git was upgraded to >=5.5.0, as it includes a fix described here (https://github.com/go-git/go-git/issues/411)

I'm not sure why the go-git module is forked in argo-workflow (https://github.com/argoproj/argo-workflows/blob/c1c3a5a1fe47508da7bc2942dbee4763dccadcd6/go.mod#L260), but argo-cd uses the upstream version of go-git (https://github.com/argoproj/argo-cd/pull/11893). They have had similar problems (https://github.com/argoproj/argo-cd/issues/7723)[](url)

terrytangyuan (approved):

terrytangyuan:
Thank you. For anyone else having the issue, the workaround is to mount the updated host keys in `/etc/ssh/`

PR #11097, opened 2023-05-17
fix: Fixed path separator in .tgz output artifacts created on windows. Fixes #10562

### LINKED ISSUE #10562: Incorrectly compressed output artifact on Windows
### Pre-requisites

Output archive artifact compress folder incorrectly in case of windows image. Instead of keeping the folder structure, it puts all files in the root folder and appends the folder names to the file names, e.g structure: 
argo_test/
|----- test.txt

In the compressed artifact, there will be the empty argo_test folder and the file called argo_test\test.txt
The problem does not persist on Linux pods.

I assume this should be a 7zip misconfiguration.

### Version

v3.4.4

```YAML
kind: Workflow
metadata:
  generateName: artifact-passing-
spec:
  entrypoint: artifact-example
  templates:
  - name: artifact-example
    steps:
    - - name: hello-win
        template: hello-win

  - name: hello-win
    nodeSelector:
          kubernetes.io/os: windows
    container:
      image: mcr.microsoft.com/windows/nanoserver:1809
      command: ["cmd", "/c"]
      args: ["mkdir argo-test && echo Welcome from windows container! > argo-test/test.txt"]
    outputs:
      artifacts:
      - name: hello-win
        path: /argo-test
        archive:
          tar:
            compressionLevel: 0
```

```text
-
```

```text
-
```

PeterKoegel:
### Motivation

Output artifacts in .tgz format that are created on Windows result in wrong directory and file structure when unpacked by 7-zip or on Linux.

Example:
1. Folder structure Windows is
`A\B\C.txt `

2. An output artifact in .tgz format is generated from this (on Windows)

3. The resulting .tgz archive is opened in 7zip
```
(folder) "A\"
(folder) "A\B\"
(file)   "A\B\C.txt"
```

### Modifications

When generating the tar archive the pathes that are written to the tar-file's file header are converted to use forward slashes instead of backslashes independent of the current operation systems path format.

### Verification

When the above example is repeaded after this fix then 7zip shows the correct folder structure:
```
(folder) A
(folder)   B
(file)       C.txt
```

### DIFF
--- a/util/archive/archive.go
+++ b/util/archive/archive.go
@@ -88,7 +88,7 @@ func tarDir(sourcePath string, tw *tar.Writer) error {
 		if err != nil {
 			return errors.InternalWrapError(err)
 		}
-		nameInArchive = filepath.Join(baseName, nameInArchive)
+		nameInArchive = filepath.ToSlash(filepath.Join(baseName, nameInArchive))
 		log.Debugf("writing %s", nameInArchive)
 		count++
 


terrytangyuan (commented):
@lippertmarkus Would you mind helping us review this change?

lippertmarkus (approved):

lippertmarkus:
@terrytangyuan names in tar headers need to use forward slashes for the tar archive to extract correctly on both Linux and Windows, so the fix is fine

terrytangyuan (approved):
Thanks!

PR #11100, opened 2023-05-18
fix: Emoticon

luyang93:
fix: Emoticon


agilgur5:
Looks like this re-introduced #2788

PR #11104, opened 2023-05-19
fix: Setup /etc/hosts for running inside devcontainer

Joibel:
If you attempt `make start` from inside a devcontainer you will be told you need to setup various entries in hosts as per docs/running-locally.md. This fix does that step for you.

### Motivation

I develop inside the supplied devcontainer with just the devcontainer CLI, and this commit would mean I can skip a step in getting started.

### Modifications

Only changes to the devcontainer startup. There are few other options as to how to do this change, as /etc/hosts is initially managed by devcontainer itself.

### Verification

After the change /etc/hosts in a container brought up with `devcontainer up` contain the entries listed in running-locally.md that are necessary for make start.

### DIFF
--- a/.devcontainer/pre-build.sh
+++ b/.devcontainer/pre-build.sh
@@ -1,6 +1,13 @@
 #!/usr/bin/env sh
 set -eux
 
+# Add hosts
+sudo bash -c 'echo "127.0.0.1 dex" >> /etc/hosts'
+sudo bash -c 'echo "127.0.0.1 minio" >> /etc/hosts'
+sudo bash -c 'echo "127.0.0.1 postgres" >> /etc/hosts'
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

agilgur5:
Was looking through some devcontainer PRs and noticed that this actually already existed and was (accidentally?) removed in #10532 ([this line](https://github.com/argoproj/argo-workflows/pull/10532/files#diff-49205b36371a0a673e7a6d64ab5b7ae5f1bdb471c54299bf63ae060d7abfac1aL8)). Wonder if that was related to switching from `postStartCommand` to `onCreateCommand` in that PR 🤔

See also #11439 and the fix in #11440 that moves it to `postStart`

Issue #10807, opened 2023-04-03
Flakey tests


terrytangyuan:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

Unit tests failed: https://github.com/argoproj/argo-workflows/actions/runs/4592799925/jobs/8110154390

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
See recent CI builds
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

tico24:
This is caused by #10768  - Until kit is fixed or reverted out, this will continue to happen.

terrytangyuan:
Here's the log: https://pipelines.actions.githubusercontent.com/serviceHosts/49efa180-38ba-4f73-8389-f407aa841894/_apis/pipelines/1/runs/34951/signedlogcontent/2?urlExpires=2023-04-04T12%3A52%3A27.4837908Z&urlSigningMethod=HMACV1&urlSignature=7YLqPxOPpolP4cnkttw6DW97ZgvYY6ui%2Fs438NRwSMY%3D

https://github.com/argoproj/argo-workflows/actions/runs/4592799925/jobs/8110154390

It's not related to kit. Are you looking at somewhere else?

tico24:
I was looking at the failed e2e test, apologies.

terrytangyuan:
Also flaky e2e test https://github.com/argoproj/argo-workflows/actions/runs/4608012677/jobs/8143253843:
```
 Condition "to have running pod" met after 6s
Waiting 1m40s for workflow metadata.name=stop-terminate-dktw8
 ● stop-terminate-dktw8   Workflow 0s      
 └ ● stop-terminate-dktw8 DAG      0s      
 └ ● A                    Pod      0s      

    signals_test.go:78: timeout after 1m40s waiting for condition
Checking expectation stop-terminate-dktw8
stop-terminate-dktw8 : Running 
[...]

PR #11105, opened 2023-05-19
fix: Disable unreliable test

Joibel:
### Motivation

The test TestSynchronizationForPendingShuttingdownWfs/PendingShuttingdownStoppingWf regularly fails in github actions unit test stage.

This should probably be looked at, but for now stop needing quite so many retries to push through simple changes.

### Modifications

Commented out test with a note on reason why.

### DIFF
--- a/workflow/controller/operator_concurrency_test.go
+++ b/workflow/controller/operator_concurrency_test.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"strconv"
 	"strings"
 	"testing"
@@ -903,6 +904,9 @@ func TestSynchronizationForPendingShuttingdownWfs(t *testing.T) {
 	})
 
 	t.Run("PendingShuttingdownStoppingWf", func(t *testing.T) {
+		if githubActions, ok := os.LookupEnv(`GITHUB_ACTIONS`); ok && githubActions == "true" {
+			t.Skip("This test regularly fails in Github Actions CI")
+		}
 		// Create and acquire the lock for the first workflow
 		wf := wfv1.MustUnmarshalWorkflow(pendingWfWithShutdownStrategy)
 		wf.Name = "one-stopping"


terrytangyuan (commented):
There are other ones too. We might want to fix each one of them. https://github.com/argoproj/argo-workflows/issues/10807

terrytangyuan:
Can you do a `go test -v -run TestSynchronizationForPendingShuttingdownWfs ./workflow/controller -count=5` to see if you can reproduce it?

Joibel:
> There are other ones too. We might want to fix each one of them. #10807

Are you only interested in fixed tests, or is disabling an option with an issue to investigate and re-enable?

workflow/controller/operator_concurrency_test.go
```diff
@@ -902,54 +902,55 @@ func TestSynchronizationForPendingShuttingdownWfs(t *testing.T) {
...
-		woc.wf.Status.Phase = wfv1.WorkflowSucceeded
-		woc.operate(ctx)
-		assert.Nil(t, woc.wf.Status.Synchronization)
-		// The pending workflow should now be running normally
-		wocTwo.operate(ctx)
-		assert.Equal(t, wfv1.WorkflowRunning, wocTwo.execWf.Status.Phase)
-	})
+	// This test regularly fails in Github Actions CI
+	// t.Run("PendingShuttingdownStoppingWf", func(t *testing.T) {
+	// 	// Create and acquire the lock for the first workflow
```
JPZ13, line 907: Instead of commenting out, it's probably better to do:
```
t.Run("PendingShuttingdownStoppingWf", func(t *testing.T) {
  t.Skip("This test regularly fails in Github Actions CI")
  ...
}
```

I don't know if Github Actions also has any environment variables that it sets, but if so, you could make it a conditional based on that environment variable being there

Joibel: This is just skipped in CI now based on [GITHUB_ACTIONS=true](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

Joibel:
> > Are you only interested in fixed tests, or is disabling an option with an issue to investigate and re-enable?
> 
> A bit concerned if it's related to core functionalities. Perhaps we can skip it only in GitHub Actions but enable it locally.

I've done that now.

terrytangyuan (approved):

PR #11121, opened 2023-05-23
feat: Unified workflows list UI and API

terrytangyuan:
Fixes https://github.com/argoproj/argo-workflows/issues/10781.


terrytangyuan:
For workflow deletion, we need to delete workflow in K8s cluster and in the DB. Here's what I propose we should do:
1. When deleting a workflow, we always delete the workflow in the K8s cluster.
2. In addition, there's a popup in the UI (when clicking the delete button) where the user can select whether to delete the workflow in the DB.

For the backend, we can choose either of the following:
1. Keep the existing behavior of `workflowServer.DeleteWorkflow()` and UI should call both `workflowServer.DeleteWorkflow()` and `archivedWorkflowServer.DeleteWorkflow()`.
1. Add an argument `deleteArchived bool` to `workflowServer.DeleteWorkflow()`. If `deleteArchived=true`, `workflowServer.DeleteWorkflow()` will also call `archivedWorkflowServer.DeleteWorkflow()` under the hood.

Any thoughts? @jessesuen @sarabala1979

terrytangyuan:
Discussed with Bala, for the backend we'll use this approach:
> Keep the existing behavior of workflowServer.DeleteWorkflow() and UI should call both workflowServer.DeleteWorkflow() and archivedWorkflowServer.DeleteWorkflow().

> Are you going to show the icon in the workflow list archived?

Right now it's text "true/false" in the column.  The icon doesn't look good in the column and the meaning of the icon is not intuitive.

rbreeze (approved):
LGTM! Agree with you that global var UI solution isn't ideal, but for now works and shouldn't have side effects.

server/workflow/workflow_server.go
```diff
@@ -140,6 +150,23 @@ func (s *workflowServer) ListWorkflows(ctx context.Context, req *workflowpkg.Wor
...
+		Namespace:   req.Namespace,
+	})
+	if err != nil {
+		log.Warnf("unable to list archived workflows:%v", err)
+	} else {
+		if archivedWfList != nil {
+			for _, item := range archivedWfList.Items {
+				if !containsWorkflow(*wfList, item) {
+					wfList.Items = append(wfList.Items, item)
+				}
```
jessesuen, line 165: I see a O^2 problem here. We iterate `wfList` every time for `archivedWfList`. I think we need to improve this by building up a map of UIDs that we seen in the live.

terrytangyuan: Thanks for the feedback! I've improved this in the changes.

persist/sqldb/workflow_archive.go
```diff
@@ -174,20 +172,13 @@ func (r *workflowArchive) ListWorkflows(namespace string, name string, namePrefi
...
-			Status: wfv1.WorkflowStatus{
-				Phase:      md.Phase,
-				StartedAt:  v1.Time{Time: md.StartedAt},
-				FinishedAt: v1.Time{Time: md.FinishedAt},
-			},
+	for i, archivedWf := range archivedWfs {
+		wf := wfs[i]
+		err = json.Unmarshal([]byte(archivedWf.Workflow), &wf)
+		if err != nil {
+			return nil, err
```
jessesuen, line 179: Seems like one bad workflow in database could cause whole request to fail. It may not even be a "bad" workflow per se. It could be an older workflow that doesnt marshal to current workflow API definition (maybe because of spec change).

I wonder if it would be better to log an error and continue to the next workflow, instead of return error immediately.

sarabala1979: I too agree with it. UI should move on if any one element is bad

terrytangyuan: Updated the PR to print a warning instead.

server/workflow/workflow_server.go
```diff
@@ -128,6 +129,28 @@ func (s *workflowServer) GetWorkflow(ctx context.Context, req *workflowpkg.Workf
...
+	for _, item := range archivedWfs.Items {
+		uidToWfs[item.UID] = append(uidToWfs[item.UID], item)
+	}
+	numWfs := 0
+	for _, v := range uidToWfs {
+		if numWfsToKeep == 0 || numWfs < numWfsToKeep {
+			finalWfs = append(finalWfs, v[0])
+			numWfs += 1
+		}
+	}
```
terrytangyuan, line 147: @jessesuen Added logic to correct the limit

server/workflow/workflow_server.go
```diff
@@ -128,6 +129,28 @@ func (s *workflowServer) GetWorkflow(ctx context.Context, req *workflowpkg.Workf
...
+	finalWfsList := v1alpha1.WorkflowList{Items: finalWfs, ListMeta: liveWfs.ListMeta}
+	sort.Sort(finalWfsList.Items)
+
+	return &finalWfsList
```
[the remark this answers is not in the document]
terrytangyuan: > iterating maps is unpredictable

This is a great catch! I have fixed it and added test to make sure the result is always consistent.

terrytangyuan:
Merging this since all comments have been addressed.

ui/src/models/workflows.ts
```diff
@@ -539,6 +539,15 @@ export interface Workflow {
 
 export const execSpec = (w: Workflow) => Object.assign({}, w.status.storedWorkflowTemplateSpec, w.spec);
 
+// The label may not have been updated on time but usually this indicates that they are already archived.
+export function isArchivedWorkflow(wf: Workflow): boolean {
+    return (
+        wf.metadata.labels &&
+        (wf.metadata.labels['workflows.argoproj.io/workflow-archiving-status'] === 'Archived' ||
```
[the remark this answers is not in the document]
terrytangyuan: Yes, however, the result of a `workflowServer.List` call will only contain either the live wf or the archived wf (to avoid duplicating the entry in the list view). If the live wf is used, this label will be "Archived". If the one in the database is used, then this label will be "Pending".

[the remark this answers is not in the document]
terrytangyuan: Improved and corrected this in https://github.com/argoproj/argo-workflows/pull/11336

persist/sqldb/workflow_archive.go
```diff
@@ -158,7 +156,7 @@ func (r *workflowArchive) ListWorkflows(namespace string, name string, namePrefi
 	}
 
 	err = r.session.
-		Select("name", "namespace", "uid", "phase", "startedat", "finishedat").
```
agilgur5, line 161: hmm this might be part of the performance regression per #12025, or rather #12030 more specifically. In particular, if this pulls the `status`, it could be huge -- even 20 archived workflows could total more than a 20MB response from the DB to the Server

agilgur5: This performance regression seems to have been confirmed by a user [on Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1706628618551249)

ui/src/app/shared/workflow-operations-map.tsx
agilgur5, line 1: This seems to have no JSX in it? Not sure why it was renamed

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -147,20 +184,45 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
...
+                                                .catch(setError);
                                         }
-                                    })
-                                    .catch(setError);
-                            }
-                        });
+                                        navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
+                                        // TODO: This is a temporary workaround so that the list of workflows
+                                        //  is correctly displayed. Workflow list page needs to be more responsive.
+                                        window.location.reload();
```
agilgur5, line 211: I couldn't reproduce this issue, so I have a fix going out to remove this. It's possible this was happening due to the bug that #11711 fixed, as that would prevent Workflows from being deleted

agilgur5 (commented):
Added some back-links to fixes for issues here for posterity

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -474,7 +549,13 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
             {workflow && (
                 <SlidingPanel isShown={!!sidePanel} onClose={() => setSidePanel(null)}>
                     {parsedSidePanel.type === 'logs' && (
-                        <WorkflowLogsViewer workflow={workflow} initialPodName={podName} nodeId={parsedSidePanel.nodeId} container={parsedSidePanel.container} archived={false} />
+                        <WorkflowLogsViewer
+                            workflow={workflow}
+                            initialPodName={podName}
+                            nodeId={parsedSidePanel.nodeId}
+                            container={parsedSidePanel.container}
+                            archived={isArchivedWorkflow(workflow)}
```
agilgur5, line 557: this appears to cause a subtle regression for logs for Workflows that are _both_ live and archived (e.g. ones that have not TTL'd yet): https://github.com/argoproj/argo-workflows/issues/12948#issuecomment-2061378878

ui/src/app/shared/workflow-operations-map.tsx
agilgur5: Fixed in #13815

PR #11140, opened 2023-05-26
chore: adding test coverage for ListSensors func

zhaque44:
### Motivation

Adding test coverage for the `ListSensors` function: Without these tests there would be no validation of the ListSensors behavior. This functionality is invoked in the `_SensorService_ListSensors_Handler` func & the `request_SensorService_ListSensors_0` func

### Modifications

I introduced two tests, `TestListSensors` and T`estListSensors_SensorClientNotSet`, to cover different scenarios in the ListSensors function. The first test validates the behavior when the sensor client is not set, while the second test specifically tests the case where the sensor client is set to nil.

### Verification

`TestListSensors` verifies that the ListSensors function correctly handles the scenario where the sensor client is not set. It expects an error indicating the sensor client is missing, then checks that the list is nil.

`TestListSensors_SensorClientNotSet` explicitly sets the sensor client to nil to simulate the absence of the client. It verifies that calling the ListSensors function in this state results in an error with the expected error message, and the sensor list returns nil.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -148,6 +148,7 @@ require (
 	github.com/go-openapi/swag v0.22.3 // indirect
 	github.com/golang-jwt/jwt/v4 v4.5.0 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
+	github.com/golang/mock v1.6.0
 	github.com/google/btree v1.0.1 // indirect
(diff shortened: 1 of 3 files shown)


terrytangyuan (approved):

terrytangyuan:
Thanks!

Issue #11131, opened 2023-05-25
workflow.parameters.json does not escape double quotes


captaincoordinates:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

The JSON provided by `workflow.parameters.json`, as described [here](https://argoproj.github.io/argo-workflows/variables/#global), does not escape quotes in parameter values and can produce invalid JSON.

I expected double quotes in workflow input parameters to be escaped so that `workflow.parameters.json` always provides a valid JSON string.

The provided workflow reproduces this error. If the `"` is removed from `argument-1`'s value the error goes away.

### Version

3.4.7

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
kind: Workflow
metadata:
  generateName: json-test-
spec:
  entrypoint: params-dumper
  arguments:
    parameters:
      - name: argument-1
        value: with " quote
  templates:
    - name: params-dumper
      container:
        image: python
        command:
          - python
          - -c
        args:
          - |
            from json import loads
            from json.decoder import JSONDecodeError
            try:
              loads("""{{workflow.parameters.json}}""")
            except JSONDecodeError as e:
              print(f"not ok: {e}")
              raise e
            else:
              print("ok")
```

### Logs from the workflow controller

```text
time="2023-05-25T19:00:05.217Z" level=info msg="Processing workflow" namespace=argo workflow=json-test-v265d
time="2023-05-25T19:00:05.221Z" level=info msg="Updated phase  -> Running" namespace=argo workflow=json-test-v265d
time="2023-05-25T19:00:05.221Z" level=info msg="Pod node json-test-v265d initialized Pending" namespace=argo workflow=json-test-v265d
time="2023-05-25T19:00:05.227Z" level=info msg="Created pod: json-test-v265d (json-test-v265d)" namespace=argo workflow=json-test-v265d
time="2023-05-25T19:00:05.227Z" level=info msg="TaskSet Reconciliation" namespace=argo workflow=json-test-v265d
time="2023-05-25T19:00:05.227Z" level=info msg=reconcileAgentPod namespace=argo workflow=json-test-v265d
[...]

PR #11186, opened 2023-06-07
test: add and refactor test for parameters with double and single quote for template replace

Transmitt0r:
### Motivation
This PR adds some tests that were created during the investigation of #11131

### Modifications
Two new tests were added to `template_test.go`, which previously only tested templates with expressions (e.g. `{{= true == false}}`.
The new tests test the behavior of replacing variables with values from the `replaceMap`.

Also, `template_test.go` has been refactored to a table driven design, because there was a lot of duplicated code betwen the test cases.

### Verification

Ran `go test ./util/template/...`

### DIFF
--- a/util/template/template_test.go
+++ b/util/template/template_test.go
@@ -11,10 +11,10 @@ type SimpleValue struct {
 	Value string `json:"value,omitempty"`
 }
 
-func processTemplate(t *testing.T, tmpl SimpleValue) SimpleValue {
+func processTemplate(t *testing.T, tmpl SimpleValue, replaceMap map[string]string) SimpleValue {
 	tmplBytes, err := json.Marshal(tmpl)
 	assert.NoError(t, err)
-	r, err := Replace(string(tmplBytes), map[string]string{}, true)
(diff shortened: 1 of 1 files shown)


util/template/template_test.go
```diff
@@ -24,60 +24,45 @@ func processTemplate(t *testing.T, tmpl SimpleValue) SimpleValue {
...
+		}
+
+		for name, tc := range testCases {
+			t.Run(name, func(t *testing.T) {
+				tmpl := SimpleValue{Value: tc.input}
+				newTmpl := processTemplate(t, tmpl, map[string]string{})
+				assert.Equal(t, tc.want, newTmpl.Value)
+			})
+		}
+	})
```
juliev0, line 50: Just curious that we don't test expressions with a non-empty replaceMap? (not that I'm asking you to do it) But do you know why that would be?

juliev0 (approved):

terrytangyuan (approved):

util/template/template_test.go
Transmitt0r: Idk either, but I think we should add some just in case

PR #11202, opened 2023-06-11
feat: Added label selectors to argo cron list. Fixes #11158

### LINKED ISSUE #11158: Add label selectors to `argo cron list` command
# Summary

My team has many different cronworkflows, and we often use the `argo cron list` command to summarize them. Sometimes we wish to only view different groups of related workflows, which we cluster together using kubernetes labels. We really like the functionality of `argo cron list` for the information it provides regarding last runs, next runs, etc, but sometimes we only want to select a subset of cronworkflows based on specific labels. Currently our solution is to pipe the output of `argo cron list` to something like `grep`, but this only allows us to filter the result based on the name of the cronworkflow. 

I would like for there to be a label selector option when using the `argo cron list` command. I imagine it would look something like this:

```bash
argo cron list -l environment=production,client=importantclient -n argo
```

Where these labels are stored in the metadata of the cronworkflow itself:

```yaml
kind: CronWorkflow
metadata:
   name: my-cronwf
   labels:
      environment: production
      client: importantclient
```

What change needs making?

It appears that there is a `labelSelector` being used in 

https://github.com/argoproj/argo-workflows/blob/58793a8ca54486c0a929ba7197d30b9f3cb3ce17/cmd/argo/commands/cron/list.go#L41

Perhaps this can be extended to appropriate the labels passed by the user before querying the kubernetes cluster for existing cronworkflows?

# Use Cases

When would you use this?

As noted above, it would be quite useful to utilize label selectors when querying for cronworkflows, especially when teams like ours have many, in order to give a glimpse at how certain "groups" of cronworkflows are running in our cluster. As another example, if a team has multiple cron workflows related to a specific use case, such as all cronworkflows configured for a specific business client, it would be useful use label selectors to grab all cronworkflows related to that client, in order to get an easily readable summary, rather than the entire list of cronworkflows in your cluster/namespace. 

Thank you argo team for your time and for an amazing product! 

---

dbragdon1:
Fixes #11158 by adding label selectors to `argo cron list` command.

### Motivation

As stated in #11158 : 

> My team has many different cronworkflows, and we often use the argo cron list command to summarize them. Sometimes we wish to only view different groups of related workflows, which we cluster together using kubernetes labels. We really like the functionality of argo cron list for the information it provides regarding last runs, next runs, etc, but sometimes we only want to select a subset of cronworkflows based on specific labels. Currently our solution is to pipe the output of argo cron list to something like grep, but this only allows us to filter the result based on the name of the cronworkflow.

### Modifications

Added `--selector` flag to `argo cron list` command to support selectively querying `CronWorkflow` objects based on labels defined under `metadata` section. Functionality is identical to the `kubectl` `--selector` flag. 

Optional `-l` flag now appears in usage message:
```
Usage:
  argo cron list [flags]

Flags:
  -A, --all-namespaces    Show workflows from all namespaces
  -h, --help              help for list
  -o, --output string     Output format. One of: wide|name
  -l, --selector string   Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2). Matching objects must satisfy all of the specified label constraints.

```

Example: 

If a workflow is created with the following metadata:

```yaml
kind: CronWorkflow
metadata:
  name: test-cwf-with-labels
  labels:
    client: importantclient
```
You can query for this label by running

`argo cron list -l client=importantclient`

or

`argo cron list --selector client=importantclient`

and the result will be the standard `argo cron list` output:

```
NAME                   AGE   LAST RUN   NEXT RUN   SCHEDULE    TIMEZONE              SUSPENDED
test-cwf-with-labels   1h    40s        19s        * * * * *   America/Los_Angeles   false
``` 

### Verification

Added `List with Labels` unit test to `test/e2e/cli_test.go` which creates two cron workflows, one with label `client: importantclient` and one without, calls `argo cron list -l client=importantclient`, and makes sure that only cron workflow with appropriate labels is selected.


terrytangyuan (approved):

terrytangyuan (commented):
You need to `make codegen -B STATIC_FILES=false`

PR #11215, opened 2023-06-14
feat: Allow non flat addressing for withItems/withParams. Fixes #11192

### LINKED ISSUE #11192: Allow for more complex data structures used with `withParam`
# Summary

When indexing the supplied data using `{{item.x}}` in `withParam`, only one level of depth is allowed when passing an array of objects. This means that only flat objects are allowed in the supplied array. It would be beneficial to enable the supplying of more complex data structures for indexing while using `withParam`, and I'd assume, by extension `withItems`.

# Use Cases

This feature would be useful when attempting to create an arbitrary number of steps with a defined object. Argo expecting the list of objects being a flat object forces the developer to contort incoming data to fit what workflows is expecting. That or, create a json object with the architecture of argo workflows in mind, if that makes sense. This opposed to, if more complex indexing were implemented, the incoming data could just be accepted as is. 

In my specific case, I'm creating k8s resources using the resource template with data passed in from a pubsub event. The object being consumed is a complex data structure (It's not really complex, just some key value pairs and a key or two that is an object), and there is an unknown amount of these objects being sent in, all within a single object. I would like to template and pass the values into the resource being created by the resource template. However, I have to find a way to transform the object before passing it in because I can't index into the nested object.

---

awilcots:
Fixes #11192 

### Motivation

I wanted to fix the issue.

### Modifications

In workflow/controller/operator.go
  Created function `fillReplaceMap` to go through an item's value and
   recursively create an index/address for each nested value type. In
   effect implementing https://github.com/tidwall/sjson#path-syntax, but
   with as little impact to the structure as I could muster.
  Flattened a conditional.

In workflow/controller/operator_test.go
  Updated processItem test to test more scenarios.

### Verification
Added tests, verified functionality locally.


awilcots:
As this change updates the keys that get added to the replaceMap, it also alters the name of the resulting node. The idea was that it was to be deterministic, which my change doesn't change, but it does add potentially significant length to the resulting name. 

Where if the input json is:
```json
[
  {
    "json": {
      "list": [
        0,
        "1"
      ],
      "number": 2,
      "string": "foo"
    },
    "list": [
      0,
      "1"
    ],
    "number": 2,
    "string": "foo"
  }
]
```
The node name would go from:
> task-name(0:json:{"list":[0,"1"],"number":2,"string":"foo"},list:[0,"1"],number:2,string:foo)

to

> task-name(0:json.list.0:0,json.list.1:1,json.list:[0,"1"],json.number:2,json.string:foo,json:{"list":[0,"1"],"number":2,"string":"foo"},list.0:0,list.1:1,list:[0,"1"],number:2,string:foo)


I'm unsure if this would be considered an interface change.

jaen:
@awilcots two questions as someone who got just bitten by it (luckily it was us who generated the JSON, but it certainly sunk some time to discover this):
  * I see the test cases added only test node name, not that the `{{ interpolation.syntax.works }}` with nested data - since this is supposed to resolve that issue, maybe it would be good to add a test for that? I guess it is a fix, because the interpolation happens against the node name, but maybe there's some place that it can be verified?
  * I see that the test data does not have a `{ "nested": { "object": "example" } }` only one with a nested list. Looking at the code it quacks as if nested objects should also work, but maybe it would be good to add a test case for this as well?

Thanks for fixing this!

juliev0:
Closing this PR as it's been stale for awhile

weafscast:
@juliev0 can I take this up?

juliev0:
> @juliev0 can I take this up?

for sure! do you want to continue on this branch or start a new one? (if you are able to reopen it, feel free to do so)

weafscast:
@juliev0 adding  few more test cases and [this](https://github.com/argoproj/argo-workflows/pull/13574) should be gtg.

PR #7439, opened 2021-12-17
fix: Support inputs for inline DAG template. Fixes #7432

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>
Fixes #7432

Don't bother creating a PR until you've done this:

* [ ] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it
  does not need to pass.
* Once required tests have passed, you can make it "Ready for review".
* Say how how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* If changes were requested, and you've made them, then dismiss the review to get it looked at again.
* Add you organization to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md) if you like.
* You can ask for help!

PR #11242, opened 2023-06-20
fix: Support inputs for inline Steps

### LINKED ISSUE #7432: can not reference variable in inline template
## Summary

```
  templates:
    - name: main
      dag:
        tasks:
          - name: dummy
            inline:
              container:
                image: docker/whalesay:latest
                command: [sh, -c]
                args: ["echo {{inputs.parameters.dummy-input}}"]
              inputs:
                parameters:
                  - name: dummy-input
                    value: 0
```
I had an inline template in dag tasks, but failed to reference the input parameters.

got this error: `failed to resolve {{inputs.parameters.series}}`

## Diagnostics

What Kubernetes provider are you using? aws eks

---

Joibel:
Make inline step templates work with inputs. Basically #7439 but for steps. Does not address any other concerns over inlines.

Fixes #7432 (which is closed, but left this unfixed)

### Verification

Test added and logs checked for correct behaviour

### DIFF
--- a/workflow/controller/inline_test.go
+++ b/workflow/controller/inline_test.go
@@ -39,3 +39,33 @@ spec:
 	woc.operate(context.Background())
 	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
 }
+
+func TestInlineSteps(t *testing.T) {
+	wf := wfv1.MustUnmarshalWorkflow(`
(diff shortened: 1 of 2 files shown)


Joibel:
Actually, this is fixed properly in #11074, but it needs rerunning to make it pass

PR #11246, opened 2023-06-21
fix: Upgrade windows container to ltsc2022

### LINKED ISSUE #10738: build argoexec for windows 2022
# Summary

Would you accept a pull request which adds a windows 2022 build for argoexec?

# Use Cases

We're in the process of switching to 2022 and therefore will build an own version of it but of course it would be great to have it in the standard container image.

ping @lippertmarkus 

---

tico24:
Fixes #10738
and the fact that Windows builds were not releasing.

Older builds were on the 1809 container, which isn't supported any more. Chocolatey recently dropped support and this is what killed the build.

The container we are replacing went out of Microsoft support in November 2020, so this shouldn't be a huge surprise for Windows users: https://learn.microsoft.com/en-us/lifecycle/announcements/windows-10-1809-end-of-servicing

### Motivation

Updates to a supported version of Windows and allows the release mechanism to work again.

### Modifications

- Upgraded dockerfile to ltsc2022
- CI runs on server 2022

I'm not super familiar with how windows containers work (if they do at all :p) but I think this now means that workflows will only run on server 2022 - it looks like containers are pinned to their host in the least containery-way possible. So this probably requires a minor release of Workflows, not a patch.

### Verification
Built on the Pipekit fork and released as mentioned below in the comments.

image

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -190,7 +190,7 @@ jobs:
   build-windows:
     name: Build & push windows
     if: github.repository == 'argoproj/argo-workflows'
-    runs-on: windows-2019
+    runs-on: windows-2022
     steps:
       - uses: actions/checkout@v3
       - name: Docker Login
@@ -336,7 +336,7 @@ jobs:
   test-images-windows:
     name: Try pulling windows
     if: github.repository == 'argoproj/argo-workflows'
-    runs-on: windows-2019
+    runs-on: windows-2022
     needs: [ push-images ]
     steps:
       - name: Docker Login
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -4,7 +4,7 @@
 # Also used as the image in CI jobs so needs all dependencies
 ####################################################################################################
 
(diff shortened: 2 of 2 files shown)


terrytangyuan (commented):
Can you try trigger a release build but only builds the image? https://github.com/argoproj/argo-workflows/blob/master/.github/workflows/release.yaml

terrytangyuan:
You can comment out most of it and only runs a docker build on Windows. You can run modify it to run it in your fork too

tico24:
This PR remains in draft for now, but I have something working

@terrytangyuan if you wish to test:
```
docker pull quay.io/pipekitdev/workflow-controller:dev-windows-linux-amd64
docker pull quay.io/pipekitdev/argoexec:dev-windows-linux-amd64
```

terrytangyuan:
If it's working, we can merge it. Let me know when ready.

JPZ13 (approved):
Validated on our fork

terrytangyuan (approved):
Passing now. Thanks!

tico24:
Err. It's not. But... ok?

PR #5951, opened 2021-05-19
fix(executor): resource patch for non-json patches regression

wreed4:

PR #11249, opened 2023-06-21
fix: Remove -f flag from 'kubectl patch' commands in resource executor (#11248)

### LINKED ISSUE #11248: Resource patch fails if `mergeStrategy: merge` or `strategic` and patch file with `flags` provided
### Pre-requisites

Whenever a "resource patch" is performed with a `mergeStrategy` different from `json` the patching fails.
This error seems to have been introduced with the [this PR](https://github.com/argoproj/argo-workflows/pull/5951)

My recommended solution would be to revert this snippet of code in `workflow/executor/resource.go`
```go
	appendFileFlag := true
	if action == "patch" {
		mergeStrategy := "strategic"
		if we.Template.Resource.MergeStrategy != "" {
			mergeStrategy = we.Template.Resource.MergeStrategy
			if mergeStrategy == "json" {
				// Action "patch" require flag "-p" with resource arguments.
				// But kubectl disallow specify both "-f" flag and resource arguments.
				// Flag "-f" should be excluded for action "patch" here if it's a json patch.
				appendFileFlag = false
			}
		}

		args = append(args, "--type")
		args = append(args, mergeStrategy)
		args = append(args, "-p")
		args = append(args, string(buff))
	}
	if len(flags) != 0 {
		args = append(args, flags...)
	}

	if len(buff) != 0 && appendFileFlag {
		args = append(args, "-f")
		args = append(args, manifestPath)
	}
```

to
```go
	if action == "patch" {
		mergeStrategy := "strategic"
		if we.Template.Resource.MergeStrategy != "" {
			mergeStrategy = we.Template.Resource.MergeStrategy
		}

		args = append(args, "--type")
		args = append(args, mergeStrategy)
		args = append(args, "-p")
		args = append(args, string(buff))
	}
	if len(flags) != 0 {
		args = append(args, flags...)
	}

	// Action "patch" require flag "-p" with resource arguments.
	// But kubectl disallow specify both "-f" flag and resource arguments.
	// Flag "-f" should be excluded for action "patch" here.
	if len(buff) != 0 && action != "patch" {
		args = append(args, "-f")
		args = append(args, manifestPath)
	}
```

### Version

latest

```YAML
metadata:
  name: test-workflow
spec:
  entrypoint: main
  templates:
    - name: main
      steps:
        - - name: create-cm
            template: create-cm

        - - name: patch-cm
            template: patch-cm
            arguments:
              parameters:
                - name: config-map-name
                  value: {{ steps.create-cm.outputs.config-map-name }}

    - name: create-cm
      resource:
        action: create
        manifest: |
          apiVersion: v1
          kind: ConfigMap
          metadata:
            generateName: config-map-test-
          data:
            test: "Hello World"
      outputs:
        parameters:
          - name: config-map-name
            valueFrom:
              jsonPath: '{.metadata.name}'

    - name: patch-cm
      inputs:
        parameters:
          - name: config-map-name
      resource:
        action: patch
        mergeStrategy: merge
        flags:
          - cm
          - "{{ inputs.parameters.config-map-name }}"
        manifest: |
          {
            "data": {
              "test": "Goodbye"
            }
          }
```

```text
time="2023-06-21T09:29:13 UTC" level=info msg="kubectl patch --type merge -p {\n  \"data\": {\n\"test\": \"Goodbye\"\n  }\n}\n cm -f /tmp/manifest.yaml -o json"
time="2023-06-21T09:29:13 UTC" level=warning msg="Non-transient error: exit status 1"
time="2023-06-21T09:29:13 UTC" level=error msg="executor error: no more retries error: when paths, URLs, or stdin is provided as input, you may not specify resource arguments as well"
time="2023-06-21T09:29:13 UTC" level=fatal msg="no more retries error: when paths, URLs, or stdin is provided as input, you may not specify resource arguments as well"
```

```text
Nothing of relevance
```

vanny96:
Fixes #11248

As described by the already existing comment, the `kubectl patch` command is used with the `-p` flag in the Resource executor. 
Since the `-p` flag and the `-f` flag can't be used together, we need to ensure that the second is not added for _any_ patch operation, not just the json ones (restoring the behaviour present before [PR 5991](https://github.com/argoproj/argo-workflows/pull/5951))

### DIFF
--- a/workflow/executor/resource.go
+++ b/workflow/executor/resource.go
@@ -125,31 +125,25 @@ func (we *WorkflowExecutor) getKubectlArguments(action string, manifestPath stri
 		output = "name"
 	}
 
-	appendFileFlag := true
 	if action == "patch" {
 		mergeStrategy := "strategic"
 		if we.Template.Resource.MergeStrategy != "" {
 			mergeStrategy = we.Template.Resource.MergeStrategy
-			if mergeStrategy == "json" {
-				// Action "patch" require flag "-p" with resource arguments.
-				// But kubectl disallow specify both "-f" flag and resource arguments.
-				// Flag "-f" should be excluded for action "patch" here if it's a json patch.
-				appendFileFlag = false
-			}
 		}
 
 		args = append(args, "--type")
 		args = append(args, mergeStrategy)
-
 		args = append(args, "-p")
 		args = append(args, string(buff))
 	}
-
 	if len(flags) != 0 {
 		args = append(args, flags...)
 	}
 
-	if len(buff) != 0 && appendFileFlag {
+	// Action "patch" require flag "-p" with resource arguments.
+	// But kubectl disallow specify both "-f" flag and resource arguments.
+	// Flag "-f" should be excluded for action "patch" here.
+	if len(buff) != 0 && action != "patch" {
 		args = append(args, "-f")
 		args = append(args, manifestPath)
 	}


terrytangyuan (commented):
Could you add some tests to cover the cases?

vanny96:
> Could you add some tests to cover the cases?

Hey @terrytangyuan, I was mistaken about the issue and the PR (and as such I've already closed the issue, about to close the PR as well)

The main issue I was facing is that I wanted to apply a `kubectl patch` with `--type=merge` without referencing the `name`/`namespace` in the manifest. 

So I would do something like this:
```yaml
      resource:
        action: patch
        mergeStrategy: merge
        flags: 
          - cm
          - name-of-configmap
        manifest: |
          data:
            test: "Hello World"
```

Which doesn't work of course, as argo workflows will pass both `-f /tmp/manifest.yaml/` and `cm name-of-configmap`, and `kubectl` will complain

The current way to do it is to write instead
```yaml
      resource:
        action: patch
        mergeStrategy: merge
        manifest: |
          apiVersion: v1
          kind: ConfigMap
          metadata:
            name: name-of-configmap
          data:
            test: "Goodbye"
```

I still wish that the documentation had this explained more explicitly, and that the behaviour of `mergeStrategy: json` wouldn't be so different from the others, but that's a matter for another PR/Issue 😁

agilgur5:
> The main issue I was facing is that I wanted to apply a `kubectl patch` with `--type=merge` without referencing the `name`/`namespace` in the manifest.

It turns out this somewhat of a workaround, per discussion in #13279. The `kind` and `name` are needed, provided by either `kubectl patch <kind> <name> -p` or `kubectl patch -f <header-file> -p`

Providing a patch file with a header might work, but it is often used in the first form, which we basically don't support as it will cause an error.

The solution is a bit more nuanced than this PR; reverting #5991 can also cause issues for JSON patches. I wrote up a PR with tests that fixes the error in #13317

PR #11270, opened 2023-06-27
fix: SSO Callback redirect fix for unavailable state cookie. Fixes #11268

### LINKED ISSUE #11268: SSO Login in argo-workflow failing with HTTP 400 Error with authentication through Microsoft AD (OIDC)
### Pre-requisites

SSO Login in argo-workflow failing with HTTP 400 Error with authentication through Microsoft AD (OIDC)

### Summary

We have configured SSO authentication mode with Microsoft AD as OIDC provider. Authentication through Microsoft AD is happening successfully, but while handling a callback we can see error page with HTTP 400 error. 

On further investigation into code, issue found to be in below code 

https://github.com/argoproj/argo-workflows/blob/aa2b66a5b7d8c3ab2af900a5fbda948b13d14085/server/auth/sso/sso.go#L230C2-L235C3

Here code is looking for a cookie with value of variable `state` and then setting a Cookie Header with value of `state`. Post that it is return HTTP 400 Error because cookie named with value of var `state` wasn't available. That doesn't make sense. If Cookie not found, then while error handling it should set the cookie with name `state`

### Configuration
```
sso:
      issuer: <issuer-url>
      clientId:
        name: argo-workflow-oidc-secret
        key: client-id
      clientSecret:
        name: argo-workflow-oidc-secret
        key: client-secret
      redirectUrl: http://arog-workflows.my.org.com/oauth2/callback
      rbac:
        enabled: true
      scopes: 
        - groups
        - emails
```

I already fixed and tested the fix, and would like to contribute the fix as a PR along with another enhancement that I raised with issue #11252 

### Version

v3.4.7

```YAML
Not Applicable
```

```text
Not Applicable
```

```text
Not Applicable
```

praveenjindal62:
Fixes #11268

### Motivation

SSO Callback from Microsoft Azure AD returns with HTTP 400 error because of missing `state` cookie header. Code is trying to check if Cookie header with name `state` exists -> Set the cookie header with name `state` -> Return with HTTP 400 error. Reload of page after HTTP 400 error redirect user to argo-workflow UI with successful authentication (Because during reload Cookie header with name `state` is present)

### Modifications

Modified function `HandleCallback` of file `server/auth/sso/sso.go`. Logic is changed to look for cookie header with name `state`. If cookie header not found then set the cookie header. 

Bottom of the function is changed as well to verify if there was an error during initial cookie header fetch and set `redirect` only if there was no error along with pre-existing logic.

### Verification

I have tested by setting up argo-workflows in AKS cluster and performed SSO Login using Microsoft Azrue AD. I no longer get HTTP 400 error and whole login and redirect journey is working fine. In the end user is redirected to the argo-workflow UI with correct authentication.

### DIFF
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -227,11 +227,9 @@ func (s *sso) HandleRedirect(w http.ResponseWriter, r *http.Request) {
 func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	state := r.URL.Query().Get("state")
-	cookie, err := r.Cookie(state)
-	http.SetCookie(w, &http.Cookie{Name: state, MaxAge: 0})
-	if err != nil {
-		w.WriteHeader(400)
-		return
+	cookie, cookieErr := r.Cookie(state)
+	if cookieErr != nil {
+		http.SetCookie(w, &http.Cookie{Name: state, MaxAge: 0})
 	}
 	redirectOption := oauth2.SetAuthURLParam("redirect_uri", s.getRedirectUrl(r))
 	// Use sso.httpClient in order to respect TLSOptions
@@ -315,7 +313,7 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 	}
 	prefix := fmt.Sprintf("%s://%s%s", proto, r.Host, s.baseHRef)
 
-	if strings.HasPrefix(cookie.Value, prefix) {
+	if cookieErr == nil && strings.HasPrefix(cookie.Value, prefix) {
 		redirect = cookie.Value
 	}
 	http.Redirect(w, r, redirect, 302)


server/auth/sso/sso.go
```diff
@@ -227,11 +227,9 @@ func (s *sso) HandleRedirect(w http.ResponseWriter, r *http.Request) {
...
 	ctx := r.Context()
 	state := r.URL.Query().Get("state")
-	cookie, err := r.Cookie(state)
-	http.SetCookie(w, &http.Cookie{Name: state, MaxAge: 0})
-	if err != nil {
-		w.WriteHeader(400)
-		return
+	cookie, cookieErr := r.Cookie(state)
+	if cookieErr != nil {
+		http.SetCookie(w, &http.Cookie{Name: state, MaxAge: 0})
```
juliev0, line 232: It looks like this will change the default path, won't it? I assume the typical case is that `cookieErr == nil`. Before we were setting `http.SetCookie()` in that case, and now we are not. Right?

juliev0:
Trying to understand the existing code, I believe we first call `HandleRedirect()` to redirect the user to the SSO endpoint, and we pass a query parameter called `redirect_uri` which looks like this (and has `state` as a query parameter in it):
Screenshot 2023-06-28 at 11 04 50 AM

And then there's an assumption that the SSO endpoint is going to call back to us using that `redirect_uri`. And then in `HandleCallback()` we are expecting to receive a query parameter called "state". And for some reason you are not receiving it?

juliev0:
From the code, it does look like this "state" cookie is merely being used to elegantly redirect the user back to the original page requested, as opposed to being used for any actual authorization, so I think you're right that it shouldn't return a 400 error just because the cookie isn't present.

praveenjindal62:
So, do you think if changes looks ok ?.

I am sure that it is not the problem with browser as there are other
components who are doIng SSO authentication with OIDC provider including
ArgoCD. And whole flow os working fine.

juliev0:
> So, do you think if changes looks ok ?. 

Please see, I did add a comment in-line in the code.

praveenjindal62:
After a close look I found that, I am testing argo workflow behind reverse proxy and doing URL rewrite, because default callback url is intercepted by oauth2-proxy at reverse-proxy. Because of URL Rewrite I am loosing the `state` cookie that was created while initiating the SSO login and hence during callback handling, cookie isn't available and seeing 400 error. 

We may not need this PR. Also, fix I introduced isn't correct. I will see if I can reject this PR and close it.

Issue #9894, opened 2022-10-23
Move global delayed delete to `podGC`/workflow-level (`PodGCDeleteDelayDuration`)


papandadj:
# Summary
The pod controller(in kubernetes) deletes redundant pods where there are too many pod.  delete is sortd by name. If a large number pods are not processed, "pod delete"  will occur in argo. Add delayed duration  to podgc field is used to delay pod deletion time for each workflow.
 
# Use Cases
workflow now supports PodGCDeleteDelayDuration as a global parameter to delay deletion, but in some special cases, it is necessary to delay deletion for individual workflow.  For example, we set  PodGCDeleteDelayDuration to 2 days by default, but soma argo task are too many, we want to delete them after 2 hours.

# When would you use this?
Control echo argo podgc delayed delete time instead of use default value.

#### I can add this feature if needed. I will use fellowing scheme.
1. Add `DelayDuration` fieled to podgc config at `/pkg/apis/workflow/v1alpha1/workflow_types.go`.
<img width="981" alt="image" src="https://user-images.githubusercontent.com/25424898/197383962-4e108f93-3d37-40ea-af5f-b884aad31cfd.png">
2. When clearing the pod,  determine whether the `DelayDuration` field is defined and  if It's defined replace PodGCDeleteDelayDuration. So add code to `workflow/controller/pod_cleanup.go` as fellow
<img width="991" alt="image" src="https://user-images.githubusercontent.com/25424898/197384202-d0892ee1-057b-4e8b-8878-91f4b32db995.png">
3. Add function to get `DelayDuration` as fellow
<img width="898" alt="image" src="https://user-images.githubusercontent.com/25424898/197384262-90e8f235-f63e-446e-bc68-bb68d86ae1bc.png">
4. When deploy workflow yaml, add DelayDuration in podgc as fellow
<img width="817" alt="image" src="https://user-images.githubusercontent.com/25424898/197384400-ff58af77-f2ab-434f-b9d7-26c55b5fa8a1.png">

Now when workflow completion, pod will be deleted after 2 minutes.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

Issue #8539, opened 2022-04-29
PodGC Timeout


SebastianGoeb:
# Summary

Would it be possible to introduce a `podGC.timeout: [15m|1h|2d]` property to delay deletion of pods?

# Use Cases

We have an issue with our fluentbit not collecting logs from pods that vanish quickly (or so it seems at least). The pod in question also provisions a fat EC2 instance that we want to get rid of ASAP when it's done, hence we use `podGC: OnPodCompletion`. Unfortunately, sometimes our logs are cut off, and sometimes logs from short-lived pods are cut off too.

I don't think this is fixable in fluentbit because if the pod is gone, then it's gone. Fluentbit wouldn't be able to do anything about it (or maybe it would, by attaching some kind of finalizer that doesn't release the pod until all logs have been collected???).

Anyway, I thought podGC might benefit from a timeout property, that would let us add a grace period of some seconds/minutes/whatever to pod termination. Do you think this would be a useful feature to have?

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

alexec:
This already exists. See docs.

SebastianGoeb:
@alexec I'm sorry, either I'm blind or we misunderstood each other. The only properties I can find in the docs relating to timeout/ttl/deadline/etc. are
- podGC.strategy (don't support a timeout)
- activeDeadlineSeconds (kills a workflow if it does over X time)
- startingDeadlineSeconds (analogous to CronJob startingDeadline)
- ttlStrategy (removes the Workflow X time after completion)

Those are all not what I was describing. I was after something like this:

config: {podGC.ttl: 60s}
0s -> workflow start, pod 1 start
30m -> pod1 completed (stays in Completed status for now), pod2 start
31m -> pod1 deleted
60m -> pod2 completed, workflow complete
61m -> pod2 deleted

I tried out [TTLStrategy](https://argoproj.github.io/argo-workflows/fields/#ttlstrategy), which I believe is what you're referring to, but that results in this behavior:

config: {workflow.ttlStrategy: 60s}
0s -> workflow start, pod 1 start
30m -> pod1 completed (deleted immediately), pod2 start
60m -> pod2 completed (deleted immediately), workflow complete
61m -> workflow deleted

Did I miss something? I was after podGC, not workflowGC.

SebastianGoeb:
Another two I found are
[...]

PR #6168, opened 2021-06-17
feat(controller): Delay pod GC deletion by a configurable amount (default 5s)

stefansedich:
The PR https://github.com/argoproj/argo-workflows/pull/5033 was meant to fix the issue of delaying pod deletion for some period so things like logging scrapers can do their thing before the pod goes away.

Based on my comments https://github.com/argoproj/argo-workflows/pull/5033#discussion_r581354568 this won't actually work as intended and another solution is needed.

My fix in this PR aims to delay the actual deletion by some configurable value and the deletion logic will then either delete immediately or queue the deletion for after this period.

@terrytangyuan tagging you as you did the original PR and tagging @alexec as you were the reviewer, as you might have more insight here. I have written an initial test here but am happy to expand coverage if desired.

One area I am not sure about is what other docs need to be updated for an API change and if that is a manual process or if there is some automation around that, it was not clear to me as I could not find any documentation explaining how to make contributions that add to the spec and this is my first time so just let me know what else I need to do.

Checklist:

* [X] My organization is added to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md).

Tips:

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it does not need to pass
* Sign-off your commits to pass the DCO check: `git commit --signoff`.
* Run `make pre-commit -B` to fix codegen or lint problems. 
* Say how how you tested your changes. If you changed the UI, attach screenshots.

PR #11293, opened 2023-07-05
docs: add GC details to Cost Optimization

agilgur5:
## Summary

Add details for `podGC` and `volumeClaimGC` in the Cost Optimization doc, including both of their default settings
- Follow-up to https://github.com/argoproj/argo-workflows/pull/10971#issuecomment-1566556322 . I didn't get a response there unfortunately, so used my best judgment here. Feel free to modify!

### Motivation

Same as #10971

### Modifications

- per feedback on #10971, mention the `podGC` default here
  - `podGC` also causes Pods to be deleted _immediately_ upon the configured condition - as in, not "after a set time"
    - for instance, when `OnWorkflowCompletion` is set, all Pods iare deleted as soon as the Workflow completes, with no wait time
      - my teams actually had problems due to the lack of wait time, as `fluentd` would sometimes fail to capture Pod details if they were deleted too fast - I had to put a `suspend` at the end of Workflows to workaround that
  - also consistently use the phrase "set time" (instead of just "time"), as the time is specifically configured

- mention `volumeClaimGC` and its default under the Volume section

### Verification

N/A

### Misc

It would be great if `podGC` and `volumeClaimGC` had a TTL to them! That would permanently fix the quick deletion issue that I had to workaround above.
- Potentially, I could set the Workflow `ttlStrategy` instead, which should hopefully delete all Pods and PVCs that are descendant from it, but unfortunately my teams had yet to implement the [Workflow Archive](https://argoproj.github.io/argo-workflows/workflow-archive/), so we could not delete any Workflows yet (especially for compliance purposes)

### DIFF
--- a/docs/cost-optimisation.md
+++ b/docs/cost-optimisation.md
@@ -35,6 +35,8 @@ Consider:
 * Data storage costs (object storage vs. volume)
 * Requirement for parallel access to data (NFS vs. block storage vs. artifact)
 
+When using volume claims, consider configuring [Volume Claim GC](fields.md#volumeclaimgc). By default, claims are only deleted when a workflow is successful.
+
 ### Limit The Total Number Of Workflows And Pods
 
 > Suitable for all.
@@ -48,8 +50,8 @@ You should delete workflows once they are no longer needed, or enable a [Workflo
 Limit the total number of workflows using:
 
 * Active Deadline Seconds - terminate running workflows that do not complete in a set time. This will make sure workflows do not run forever.
-* [Workflow TTL Strategy](fields.md#ttlstrategy) - delete completed workflows after a time
-* [Pod GC](fields.md#podgc) - delete completed pods after a time
+* [Workflow TTL Strategy](fields.md#ttlstrategy) - delete completed workflows after a set time.
+* [Pod GC](fields.md#podgc) - delete completed pods. By default, Pods are not deleted.
 
 Example
 


agilgur5:
> It would be great if `podGC` and `volumeClaimGC` had a TTL to them! That would permanently fix the quick deletion issue that I had to workaround above.

Responding to myself, but I did find an undocumented configuration for this last night, `PodGCDeleteDelayDuration`: https://github.com/argoproj/argo-workflows/blob/8ec5623093181cfb5ef7c34a5df8a8c917282b7b/config/config.go#L91

As this is entirely undocumented, I found this in a super roundabout way:
1. Stumbled upon [a value in the Helm chart](https://github.com/argoproj/argo-helm/blob/5f55ef2c4cc16d4c91faaf1bdb7e0c1d7385435a/charts/argo-workflows/values.yaml#L353)
1. `git blame`d that to see https://github.com/argoproj/argo-helm/pull/1728
1. Which redirects to https://github.com/argoproj/argo-workflows/issues/8539, which is full of comments describing how this is undocumented
    - The issue described there with `fluentbit` is literally _identical_ to the issue I described above with `fluentd`
1. Where someone found the original PR: https://github.com/argoproj/argo-workflows/pull/6168
1. and then found the mostly implemented feature request: https://github.com/argoproj/argo-workflows/issues/9894 after searching for `PodGCDeleteDelayDuration`

I'll follow-up with a PR to document this in the configmap example in the docs. Will probably also implement #9894 too as that is useful and OP basically wrote everything too (will credit them of course as co-author), but never got a response unfortunately.

agilgur5:
Implemented #9894 in #11325, albeit with a slightly different approach to simplify the code

PR #10808, opened 2023-04-03
feat: Expose customized UI links and columns

rbreeze:
Fixes #10447

The following in `workflow-controller-configmap.yaml`:

```yaml
data: 
  links:
    - name: Completed Workflows
      scope: workflow-list
      url: http://workflows?label=workflows.argoproj.io/completed=true
  columns: |
    - name: Workflow Completed
      type: label
      key: workflows.argoproj.io/completed
```

yields:

Screenshot 2023-04-03 at 10 08 20

PR #11297, opened 2023-07-05
docs: add all remaining controller configurations

agilgur5:
### Motivation

Several configurations were missing, and, as a result, entirely undocumented
  - several of these can be very useful for different use-cases, lack of docs makes these settings hard to discover
  - addresses https://github.com/argoproj/argo-workflows/pull/11293#issuecomment-1622162284 and https://github.com/argoproj/argo-workflows/issues/8539

#### Primary Changes
- add namespace, columns, podSpecLogStrategy, podGCGracePeriodSeconds, podGCDeleteDelayDuration, initialDelay, and retentionPolicy
  - add comments for all of the above, primarily copying from `config.go`: https://github.com/argoproj/argo-workflows/blob/30a916880ff21027ad1004c4ae176037f1a956da/config/config.go#L22
    - a few comments from other places, like `info.go` for Column: https://github.com/argoproj/argo-workflows/blob/8ec5623093181cfb5ef7c34a5df8a8c917282b7b/pkg/apis/workflow/v1alpha1/info.go#L18
    - `argo-helm` for retentionPolicy: https://github.com/argoproj/argo-helm/blob/5f55ef2c4cc16d4c91faaf1bdb7e0c1d7385435a/charts/argo-workflows/values.yaml#L322
      - big kudos to the `argo-helm` maintainers and contributors for keeping all the config up-to-date despite lack of docs!

#### Minor Secondary Changes
- fix typing issues -- configmap values must be strings
- minor copy-edits to docs and comments
  - add comments for RetentionPolicy as it was the odd one out with no comments
  - fix comment that said DeleteDelayDuration was integer seconds when it is actually a Duration type

### Verification

n/a

### DIFF
--- a/docs/workflow-controller-configmap.yaml
+++ b/docs/workflow-controller-configmap.yaml
@@ -13,6 +13,9 @@ data:
   # controller watches workflows and pods that *are not* labeled with an instance id.
   instanceID: my-ci-controller
 
+  # Namespace is a label selector filter to limit the controller's watch to a specific namespace
+  namespace: my-namespace
+
   # Parallelism limits the max total parallel workflows that can execute at the same time
   # (available since Argo v2.3). Controller must be restarted to take effect.
   parallelism: 10
@@ -77,6 +80,16 @@ data:
     - name: Completed Workflows
       scope: workflow-list
       url: http://workflows?label=workflows.argoproj.io/completed=true
+
+  # Columns are custom columns that will be exposed in the Workflow List View.
+  columns: |
+    # Adds a column to the Workflow List View
+    - # The name of this column, e.g., "Workflow Completed".
+      name: Workflow Completed
+      # The type of this column, "label" or "annotation".
+      type: label
+      # The key of the label or annotation, e.g., "workflows.argoproj.io/completed".
+      key: workflows.argoproj.io/completed
 
   # uncomment following lines if you want to change navigation bar background color
   # navColor: red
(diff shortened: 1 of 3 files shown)


agilgur5:
Might need to run codegen on this. On a different computer, so gotta set up my dev environment again 😅

terrytangyuan (commented):
@jmeridth Would you like to help review this PR? This borrows some docs from argo-helm :-)

agilgur5:
added a tiny fix to existing docs. realized the comment for DeleteDelayDuration incorrectly stated the type as `int` when it is a `Duration`

agilgur5:
@jmeridth just a friendly follow-up that this is awaiting your review 🙂

docs/workflow-controller-configmap.yaml
```diff
@@ -78,6 +81,16 @@ data:
       scope: workflow-list
       url: http://workflows?label=workflows.argoproj.io/completed=true
 
+  # Columns are custom columns that will be exposed in the Workflow List View.
+  columns: |
```
jmeridth, line 85: this is follow-up to [this PR](https://github.com/argoproj/argo-workflows/pull/10808), correct?

jmeridth (approved):
based on my understanding this looks good

docs/workflow-controller-configmap.yaml
agilgur5: Yep! Docs/example for that feature

terrytangyuan (approved):

docs/workflow-controller-configmap.yaml
agilgur5: added a comment that this apparently requires Argo v3.5 (I had not realized that earlier) in #11474

Issue #4991, opened 2021-02-02
"token not valid for running mode" after SSO authentication


Deepa050:
## Summary 

Argo UI comes back to login page with error `{"code":16,"message":"token not valid for running mode”}` after completing SSO user authentication.

## Diagnostics

argo version:
argo: v2.12.3
  BuildDate: 2021-01-05T00:49:48Z
  GitCommit: 93ee530126cc1fc154ada84d5656ca82d491dc7f
  GitTreeState: clean
  GitTag: v2.12.3
  GoVersion: go1.13
  Compiler: gc
  Platform: darwin/amd64

I have added sso configuration in  the config map "workflow-controller-configmap"
```yaml
sso: |
      issuer: https://cloudsso-test.xxx.com
      clientId:
        name: argo-server-sso
        key: clientID
      clientSecret:
        name: argo-server-sso
        key: clientSecret
      redirectUrl: http://argo-code.xxx.com/workflows/argo
      scopes:
       - groups
       - email
      rbac:
        enabled: false
```
I have added the required secrets for SSO and also starting argo-server  in "--auth-mode sso"

argo-server logs :
```
time="2021-02-02T07:34:01.198Z" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = token not valid for running mode" grpc.code=Unauthenticated grpc.method=GetVersion grpc.service=info.InfoService grpc.start_time="2021-02-02T07:34:01Z" grpc.time_ms=0.34 span.kind=server system=grpc
time="2021-02-02T07:34:01.218Z" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = token not valid for running mode" grpc.code=Unauthenticated grpc.method=ListWorkflowTemplates grpc.service=workflowtemplate.WorkflowTemplateService grpc.start_time="2021-02-02T07:34:01Z" grpc.time_ms=0.065 span.kind=server system=grpc
```

**Please note we use JWT token for SSO and not JWE.** 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

alexec:
Did you also add —autho-mode=sso?

Deepa050:
Yes I did. Below is the yaml for argo-server.  Application goes through the entire SSO authentication successfully but fails to run after that. Can you also confirm if using JWT token doesnt impact in anyway since argo document says it uses JWE. 
```yaml
[...]

PR #11300, opened 2023-07-06
docs(auth): add link to FAQ on invalid token error

### LINKED ISSUE #5832: link to SSO docs when login fails: "token not valid for running mode"
Many users are confused by this error. This means they have not configured `--auth-mode` correctly.

We could enhance this error message with more help.

agilgur5:
### Motivation

- link was requested by Alex, fixes #5832

- also shorten the error message to "token not valid" as, in my experience at least, "for running mode" actually significantly increased confusion (c.f. https://github.com/argoproj/argo-workflows/issues/4991#issuecomment-1466465686)
  - everyone on my teams thought that it meant we misconfigured our auth or that only one mode could be used at a time - but actually, we were just missing the `"Bearer "` prefix for our token

- my comment about the `"Bearer "` prefix (https://github.com/argoproj/argo-workflows/issues/4991#issuecomment-1466465686) got a few upvotes, so thought this would be worth documenting as well

### Modifications

- add a link to the auth error message in `gatekeeper.go` when the token is invalid
  -  shorten the error message as well, per above motivation

- add the `"Bearer "` prefix as another common reason for this error in the FAQ

### Verification

Ran `make docs`

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,8 +1,11 @@
 # FAQ
 
-## "token not valid for running mode", "any bearer token is able to login in the UI or use the API"
+## "token not valid", "any bearer token is able to login in the UI or use the API"
 
-You've not configured Argo Server authentication correctly. If you want SSO, try running with `--auth-mode=sso`.
+You may not have configured Argo Server authentication correctly.
+
+If you want SSO, try running with `--auth-mode=sso`.
+If you're using `--auth-mode=client`, make sure you have `Bearer ` in front of the token, as mentioned in [Access Token](access-token.md#token-creation).
 
 [Learn more about the Argo Server set-up](argo-server.md)
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -5,7 +5,7 @@
 You may not have configured Argo Server authentication correctly.
 
 If you want SSO, try running with `--auth-mode=sso`.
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):

terrytangyuan:
docs build need to be fixed

agilgur5:
hmm let me figure out which part broke -- looks like it may be detecting the `Bearer ` reference as a false positive token match. from [the logs](https://github.com/argoproj/argo-workflows/actions/runs/5479062739/jobs/9980381761?pr=11300):

```diff
-If you're using `--auth-mode=client`, make sure you have `*** in front of the token, as mentioned in [Access Token](access-token.md#token-creation).
+If you're using `--auth-mode=client`, make sure you have `Bearer` in front of the token, as mentioned in [Access Token](access-token.md#token-creation).
```
it's also missing the intentional space after `Bearer` in the docs as well 🤔

agilgur5:
ah okay, looks like removing the intentional space in front of `Bearer ` was the change, and then GitHub Actions's log is redacting it as a false positive.

agilgur5:
@terrytangyuan error should be fixed now

Seems like `markdownlint` has no line ignore functionality unfortunately (https://github.com/markdownlint/markdownlint/issues/16), so just removed the intentional space

PR #11325, opened 2023-07-10
feat(podGC): add Workflow-level `DeleteDelayDuration`

### LINKED ISSUE #9894: Move global delayed delete to `podGC`/workflow-level (`PodGCDeleteDelayDuration`)

agilgur5:
### Motivation

- previously, this could only be set globally for the entire controller
- now, users can configure this per Workflow as needed / wanted
  - and can override the global setting as desired
  - in particular, this matches how `strategy` etc can already be set per Workflow
    - so users have a single place to configure all options of `podGC`

- in the future (I assume after some deprecation window), the global setting could be removed
  - `workflowDefaults` can achieve the same thing now
  
Resolves #9894
- I took a bit of a different approach from there; with more explicit and consistent naming and typing
  - Using `metav1.Duration` was more consistent with the global and simplified the logic as well
  - Added more comments and more consistent comments as well 

### Modifications

- add `DeleteDelayDuration` to Workflow-level `podGC`
  - add to `workflow_types.go`, then add logic to use this in `pod_cleanup.go`
  - add to PodGC examples as well
  - +codegen

### Verification

So `pod_cleanup_test.go` actually has no unit test for the whole `queuePodsForCleanup()` function, so I couldn't just add a unit there. That function could require some mocking to properly test, so I didn't introduce a whole new test suite for that at this point.

### Review Notes

Main issue I have is with the codegen for `fields.md`, where it explains `metav1.Duration` a bit misleadingly 😕 
- Everywhere else correctly shows that it boils down to just a string as well, but not this generated section 😕 
- See in-line comment about this

Related, durations seem a bit inconsistently typed in the codebase and spec. Some are ints, some are strings, and others are Durations.
- Not sure if there's a reason for that? I would guess there is some history?


docs/fields.md
```diff
@@ -5122,6 +5123,24 @@ ObjectReference contains enough information to let you inspect or modify the ref
...
+- [`pod-gc-strategy-with-label-selector.yaml`](https://github.com/argoproj/argo-workflows/blob/master/examples/pod-gc-strategy-with-label-selector.yaml)
+
+- [`pod-gc-strategy.yaml`](https://github.com/argoproj/argo-workflows/blob/master/examples/pod-gc-strategy.yaml)
+</details>
+
+### Fields
+| Field Name | Field Type | Description   |
+|:----------:|:----------:|---------------|
+|`duration`|`string`|_No description available_|
+
```
agilgur5, line 5143: this is the part I find issues with from the codegen. It's missing spaces (between multi-line comments in the Go), for one thing, but also it confusingly lists `duration` as a sub-field. That is correct for the Go type itself, but it marshals to a string, so the configuration is just a plain string.

This doc makes me think I would do:

```yaml
  podGC:
    deleteDelayDuration:
      duration: 30s
```

instead of just `deleteDelayDuration: 30s`, which is how it actually is.

Is there a way to override this or something?

agilgur5: Well, I was able to fix the spacing by modifying [this line](https://github.com/argoproj/argo-workflows/blob/82310dd459aa080754169d6a1667d30d9b7c75bf/hack/docgen.go#L60) in `docgen.go`, but while fixing this it ends up adding extraneous spaces in several places 😕
The resulting diff for `fields.md` ends up quite large as well, so I didn't even commit that change

terrytangyuan (commented):
@juliev0 Could you help review this?

juliev0:
> Details

I wonder if it could be running a different version of Go from what you are? looks like this is using 1.20: https://github.com/argoproj/argo-workflows/blob/master/.github/workflows/ci-build.yaml#L205

juliev0:
@agilgur5 Looks good. Once you get the CI all passing, let me know and I can approve.

agilgur5:
> I wonder if it could be running a different version of Go from what you are? looks like this is using 1.20: https://github.com/argoproj/argo-workflows/blob/master/.github/workflows/ci-build.yaml#L205

So I'm using the dev container on my current machine, which is also using Go [`1.20`](https://github.com/argoproj/argo-workflows/blob/a0b3c136c0d139a724aaf5095a827436e7cbea66/.devcontainer/devcontainer.json#L6) 😅 
Confirmed that too:

```sh
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (feat-workflow-podgc-delay) $ go version
go version go1.20.5 linux/amd64
```

Strange. Might try rebuilding the dev container from scratch 🤷

agilgur5:
> Might try rebuilding the dev container from scratch 🤷

fresh build solved it 🤷 

> @agilgur5 Looks good. Once you get the CI all passing, let me know and I can approve.

@juliev0 should be good to go now!

juliev0 (approved):

docs/fields.md
agilgur5: > This doc makes me think I would do:

This also affects the OpenAPI schema, not just `fields.md`, per #12631. The CRDs were correct though per https://github.com/argoproj/argo-workflows/issues/12631#issuecomment-1931128000 as the k8s tooling knows [how to handle this specific type](https://github.com/kubernetes/apimachinery/blob/2465dc5239ab8827a637148a78b380c278b4a5f4/pkg/apis/meta/v1/duration.go#L65).

I wrote up a workaround for this in #13543 

> Is there a way to override this or something? Casting somewhere? Not sure

I tried casting to a string in the `proto` struct tags of the type with `casttype`, but that didn't work and caused errors. It's possible I was doing it wrong though 🤷

PR #11335, opened 2023-07-11
chore: Reduce release yaml

tico24:
### Motivation

I noticed some duplicate code in the release yaml that could be compacted down.

### Verification

Ran the release a number of times on our fork.

(If you're reviewing all my commits you'll see that I experimented with improving the caching. Due to the way the docker container is constructed, more agressive caching results in failed builds, so I have reverted those changes.)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -86,6 +86,9 @@ jobs:
             --build-arg GIT_TAG=$GIT_TAG \
             --build-arg GIT_TREE_STATE=$GIT_TREE_STATE \
             --output "type=image,push=true" \
+            --build-arg GIT_COMMIT=$GIT_COMMIT \
+            --build-arg GIT_TAG=$GIT_TAG \
+            --build-arg GIT_TREE_STATE=$GIT_TREE_STATE \
             --platform="${PLATFORM}" \
             --target $TARGET \
             --provenance=false \
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -80,6 +80,9 @@ jobs:
           docker buildx build \
             --cache-from "type=local,src=/tmp/.buildx-cache" \
             --cache-to "type=local,dest=/tmp/.buildx-cache" \
+            --build-arg GIT_COMMIT=$GIT_COMMIT \
+            --build-arg GIT_TAG=$GIT_TAG \
+            --build-arg GIT_TREE_STATE=$GIT_TREE_STATE \
             --output "type=image,push=true" \
             --platform="${PLATFORM}" \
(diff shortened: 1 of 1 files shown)


caelan-io (approved):

.github/workflows/release.yaml
```diff
@@ -81,110 +81,10 @@ jobs:
...
-          tag_suffix=$(echo $PLATFORM | sed -r "s/\//-/g")
-          image_name="${DOCKERIO_ORG}/${TARGET}:${tag}-${tag_suffix}"
-
-          docker buildx build \
-            --cache-from "type=local,src=/tmp/.buildx-cache" \
-            --cache-to "type=local,dest=/tmp/.buildx-cache" \
-            --output "type=image,push=true" \
-            --build-arg GIT_COMMIT=$GIT_COMMIT \
-            --build-arg GIT_TAG=$GIT_TAG \
-            --build-arg GIT_TREE_STATE=$GIT_TREE_STATE \
```
terrytangyuan, line 172: Why did we remove these?

tico24: over-enthustic delete caused by looking at the 3.4.8 branch. They are back now.

terrytangyuan (approved):
Could you run another test release before I merge this?

tico24:
> Could you run another test release before I merge this?

my test release had those lines in. I just failed to cherry pick the right lines across from one repo to another.

terrytangyuan:
Commits seem messed up. Mind cleaning up a bit?

tico24:
I have no idea what happened there, but I'm enjoying the fact that this PR has nearly the same number of commits as LOC changed.

terrytangyuan (approved):

.github/workflows/release.yaml
```diff
@@ -20,102 +20,19 @@ permissions:
...
-        platform: [ linux/arm64 ]
+        platform: [ linux/amd64, linux/arm64 ]
         target: [ workflow-controller, argocli, argoexec ]
     steps:
       - uses: actions/checkout@v3
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
-        with:
-          platforms: arm64
```
terrytangyuan, line 118: Just saw this change. Could you confirm that the built images work as expected for multi-arch?

tico24: yes, if you don't set it, qemu and buildkit just do the right thing.

tico24: and here's the screenshot from quay showing the multi-arch image
image

terrytangyuan: Great thanks!

PR #11336, opened 2023-07-11
fix: Live workflow takes precedence during merge to correctly display in the UI

terrytangyuan:
See discussions in https://github.com/argoproj/argo-workflows/pull/11121#discussion_r1259744758

Live workflow is more up-to-date.


rbreeze (approved):
LGTM

PR #11342, opened 2023-07-12
docs: link Core Components to Architecture

RafaelClovr:
### Motivation
The documentation does not talk about Workflows Server and Workflow controller

### Modifications
I added few basic descriptions about that components.

### Verification

Just take a look at the rest of the documentation.

### DIFF
--- a/docs/workflow-concepts.md
+++ b/docs/workflow-concepts.md
@@ -1,10 +1,6 @@
 # Core Concepts
 
 This page serves as an introduction to the core concepts of Argo.
-## The `Server`
-Argo Workflows Server handles the authentication, the UI, and the authentication mechanism (SSO, tokens, or none [server mode]).
-
-
 
 ## The `Workflow`
 
--- a/docs/workflow-concepts.md
+++ b/docs/workflow-concepts.md
@@ -154,5 +154,6 @@ In this example `A` runs first. Once it is completed, `B` and `C` will run in pa
         template: echo
 ```
 
-## The `Workflow Controller`
-It allows to handle where could we run the Workflow, and let the user restricts which namespace is able to run any Workflow. 
+## Architecture
+
+If you are interested in Argo's underlying architecture, see [Architecture](architecture.md).
--- a/docs/workflow-concepts.md
+++ b/docs/workflow-concepts.md
@@ -1,6 +1,10 @@
 # Core Concepts
 
-This page serves as an introduction into the core concepts of Argo.
+This page serves as an introduction to the core concepts of Argo.
+## The `Server`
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
@agilgur5 Would you like to help review this?

agilgur5 (changes requested):
So documentation on the internals is typically placed in an "Architecture" page in most k8s projects. Argo has an ["Architecture" page](https://argoproj.github.io/argo-workflows/architecture/) in the Developer Guide.

We could link to the Architecture page from here, and that page could certainly use more details, but I don't think the architectural details belong on this page, which is part of the User Guide.
A user does not need to know the underlying architecture to get started with Argo, but they certainly may want to. I think a link to the Architecture page would serve better for that purpose

RafaelClovr:
@agilgur5 I agree, could be interesting to just add the link.

docs/workflow-concepts.md
```diff
@@ -149,3 +153,6 @@ In this example `A` runs first. Once it is completed, `B` and `C` will run in pa
         dependencies: [B, C]
         template: echo
 ```
+
+## The `Workflow Controller`
+It allows to handle where could we run the Workflow, and let the user restricts which namespace is able to run any Workflow.
```
agilgur5, line 158:
```suggestion
## Architecture

If you are interested in Argo's underlying architecture, see [Architecture](architecture.md)
```

agilgur5 (commented):
@RafaelClovr would you like to update your PR to add a link instead?

I provided a suggestion below as to what that may look like. The Architecture page itself could use a bit more details if you would like to improve that.

Will have to update the title of your PR to be more specific as well, for instance, `docs: link Core Components to Architecture`. CI is also currently failing `markdownlint`

docs/workflow-concepts.md
```diff
@@ -1,6 +1,10 @@
 # Core Concepts
 
-This page serves as an introduction into the core concepts of Argo.
+This page serves as an introduction to the core concepts of Argo.
+## The `Server`
+Argo Workflows Server handles the authentication, the UI, and the authentication mechanism (SSO, tokens, or none [server mode]).
+
+
```
agilgur5, line 7:
```suggestion
This page serves as an introduction to the core concepts of Argo.
```

We should revert this section addition now that the link at the bottom of the page exists. 
Reverting this should also fix the remaining `markdownlint` CI issues

terrytangyuan (approved):

agilgur5:
Thanks for your contribution @RafaelClovr !
If you'd like to contribute more details to the Architecture page or any other docs, feel free to make additional PRs 🙂

PR #11355, opened 2023-07-14
fix: Enable the workflow created by a wftmpl to retry after manually stopped

TongKangheng:
…ally after manually stopped

No, I just met a problem in my work, but I made no issue but fix it myself because of time constraints

### Motivation

Most of the users I am responsible for use workflow-templates to create workflows. However, there is a problem that has been bothering us: after the workflows created through the workflow templates are manually stopped, when retrying the failed node (because of the Stop operation) and all subsequent nodes will be skipped. Here is an example:

After manually stopped:
image

Ater manually retried:
image

### Modifications

After carefully reading the relevant code, we finally realized that in the `FormulaRetryWorkflow` function, `newWF.Status.StoredWorkflowSpec.Shutdown` should be set to an empty string instead of keeping the shutdown information.

### Verification

The following is a sample workflow template file：
```yaml
metadata:
  name: clf-test-retry
  namespace: argo
spec:
  templates:
    - name: echo
      inputs:
        parameters:
          - name: message
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: alpine:3.7
        command:
          - sleep
          - 1000s
        resources: {}
    - name: diamond
      inputs: {}
      outputs: {}
      metadata: {}
      dag:
        tasks:
          - name: A
            template: echo
            arguments:
              parameters:
                - name: message
                  value: A
          - name: B
            template: echo
            arguments:
              parameters:
                - name: message
                  value: B
            dependencies:
              - A
          - name: C
            template: echo
            arguments:
              parameters:
                - name: message
                  value: C
            dependencies:
              - A
          - name: D
            template: echo
            arguments:
              parameters:
                - name: message
                  value: D
            dependencies:
              - B
              - C
  entrypoint: diamond
  arguments: {}
```
You can just stop and retry the workflow when node A is running and the images in the ***Motivation*** can be reproduced.
After our fix, the workflow after retrying is as follows:
image

Signed-off-by: tongkangheng.tkh <tongkangheng.tkh@bytedance.com>

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -820,6 +820,9 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 	newWF.Status.Message = ""
 	newWF.Status.StartedAt = metav1.Time{Time: time.Now().UTC()}
 	newWF.Status.FinishedAt = metav1.Time{}
+	if newWF.Status.StoredWorkflowSpec != nil {
+		newWF.Status.StoredWorkflowSpec.Shutdown = ""
+	}
 	newWF.Spec.Shutdown = ""
 	newWF.Status.PersistentVolumeClaims = []apiv1.Volume{}
 	if newWF.Spec.ActiveDeadlineSeconds != nil && *newWF.Spec.ActiveDeadlineSeconds == 0 {


terrytangyuan (commented):
Thanks. Could you sign your commits?

terrytangyuan:
Tests are failing

TongKangheng:
> Tests are failing

TongKangheng:
> Thanks. Could you sign your commits?

I have done this

terrytangyuan:
Thanks. Please fix test failures.

terrytangyuan (approved):

Issue #11374, opened 2023-07-17
Periodly healthz check may cause serious problems of apiserver/etcd


cwdsuzhou:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

lightly do healthz check

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
When we deploy the controller by namespace and we have more one hundreds namespaces. We found apiserver/etcd may meet serious problems due to the health check.
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

terrytangyuan:
>When we deploy the controller by namespace and we have more one hundreds namespaces. We found apiserver/etcd may meet serious problems due to the health check.

What is the specific issue you are trying to address? What do you mean by "serious problems"?

cwdsuzhou:
> > When we deploy the controller by namespace and we have more one hundreds namespaces. We found apiserver/etcd may meet serious problems due to the health check.
> 
> What is the specific issue you are trying to address? What do you mean by "serious problems"? Using the lister might not surface all potential issues.

workflow is a `CRD`, kueb-apiserer do not have caches, so if we list from apiserver, we should first list from etcd every time. Issue in my environment is frequent healthz check can cause apiserver cpu/menory increase when deployed in many namespace and we have a large number of `CR` exist in the cluster.
And, actually we have used informer where caches exist, why not get from it directly?
IMO, Health check directly relays on apiserver/etcd is too heavy.

terrytangyuan:
The default only lists 200 wfs and you can reduce `HEALTHZ_LIST_LIMIT` if you want. Would that be sufficient?

cwdsuzhou:
> The default only lists 200 wfs and you can reduce `HEALTHZ_LIST_LIMIT` if you want. Would that work for you?

[...]

PR #11375, opened 2023-07-17
fix: Health check from lister not apiserver

cwdsuzhou:
Fixes [#11374](https://github.com/argoproj/argo-workflows/issues/11374)

### Motivation

Reduce the `List` check from controller. IMO, it is not reasonable to list from apiserver for health check. Informer is not so reliable and if informer is not healthy, the controller may exit or resync from apiserver.

### Modifications

Change the health check logic from apiserver to the local cache


terrytangyuan (changes requested):
See discussions in https://github.com/argoproj/argo-workflows/issues/11374

terrytangyuan:
Bump

cwdsuzhou:
Since we have disscussed https://github.com/argoproj/argo-workflows/issues/11374, can we work on this PR now?

@terrytangyuan

terrytangyuan (approved):
I am ok with this change. Let's watch for any issues once merged/released

astraw99:
> I am ok with this change. Let's watch for any issues once merged/released

Raised a PR https://github.com/argoproj/argo-workflows/pull/11770 to fix an issue https://github.com/argoproj/argo-workflows/issues/11769 introduced from this.
PTAL thanks.

juliev0:
@cwdsuzhou When you tested this, were you able to successfully execute line 37 while catching a Workflow that hadn't been reconciled yet? We noticed a panic occurring on [this](https://github.com/argoproj/argo-workflows/blob/main/pkg/client/listers/workflow/v1alpha1/workflow.go#L68) line, when the `*Unstructured` type in the Informer gets type asserted to a `*Workflow` type instead. It seems like the Informer holds `*Unstructured` types, so I'm not sure if this type conversion ever works. If you have any thoughts, let me know. Thanks!

agilgur5:
> We noticed a panic occurring on [this](https://github.com/argoproj/argo-workflows/blob/main/pkg/client/listers/workflow/v1alpha1/workflow.go#L68) line, when the `*Unstructured` type in the Informer gets type asserted to a `*Workflow` type instead. It seems like the Informer holds `*Unstructured` types, so I'm not sure if this type conversion ever works.

For posterity, this was followed up in #12326 and #12353

PR #11393, opened 2023-07-19
fix: log level for client-side throttling

### LINKED ISSUE #11387: Better exposure for client-side throttling
# Summary

When the `--qps` and `--burst` are set too low on the Workflow Controller, this can create client-side throttling by the underlying client-go library. (It appears that this serves to slow down the messages going out as opposed to reject requests based on my testing.) 

The only way currently to know this is happening is that the library spits out messages like:
```
I0525 09:00:21.037886       1 request.go:601] Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t
```
Ideally, maybe we'd have a warning-level message of our own or perhaps a metric to create greater exposure for the issue and help it to be more quickly identified.

Note that there is debug-level logging right now for messages in general here: https://github.com/argoproj/argo-workflows/blob/master/util/logs/log-k8s-requests.go#L20

# Use Cases

When would you use this?

---

Jack-R-lantern:
Expresses when throttling occurs on the
client side by setting the log message level to Warn.

Fixes #11387

### DIFF
--- a/util/logs/log-k8s-requests.go
+++ b/util/logs/log-k8s-requests.go
@@ -17,7 +17,7 @@ func (m k8sLogRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	x, err := m.roundTripper.RoundTrip(r)
 	if x != nil {
 		verb, kind := k8s.ParseRequest(r)
-		log.Debugf("%s %s %d", verb, kind, x.StatusCode)
+		log.Warnf("%s %s %d", verb, kind, x.StatusCode)
 	}
 	return x, err
 }


util/logs/log-k8s-requests.go
```diff
@@ -17,7 +17,7 @@ func (m k8sLogRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	x, err := m.roundTripper.RoundTrip(r)
 	if x != nil {
 		verb, kind := k8s.ParseRequest(r)
-		log.Debugf("%s %s %d", verb, kind, x.StatusCode)
+		log.Warnf("%s %s %d", verb, kind, x.StatusCode)
```
terrytangyuan, line 20: What's the motivation?

tooptoop4: it should be kept at debug, this log is very expensive!

tooptoop4 (changes requested):

Jack-R-lantern:
Sorry about that. I'll think about it a bit more and submit a PR

PR #11409, opened 2023-07-20
feat: cli allow retry successful workflow if nodeFieldSelector is set. Fixes #11020)

### LINKED ISSUE #11020: feat: Retry (rerun) successful workflows from certain steps
# Summary

- On #2431 we added the option to rerun successful steps in the Retry operation.
- This Allowed the user to select which steps to reset as part of the Retry command.
- I think we can use the `restartSuccesful` property that was added to allow rerunning successful workflow from given steps.

# Use Cases

- I created a mechanism for rerun with debug pause (in short: all templates have the env variable `ARGO_DEBUG_PAUSE_{{workflow.parameters.pause}}`, the default value for pause is `NEVER`, when I use the Retry I override the value to `AFTER`).
- I want to be able to use it also for successful workflows (for instance: I need to do manual operation on terraform state, I need to test out something manually etc etc).

# Implmentation

The Retry operation will fail if workflow is not in Failed or Error state:
https://github.com/argoproj/argo-workflows/blob/ec856835a3a4ec78164aa737f98d4b1653809781/workflow/util/util.go#L806

IIUC, it should be fairly simple to change this condition to allow other states, in case the `restartSuccesful` and node selector were given.

WDYT?

cc: @markterm

---

or-shachar:
Fixes #11020 

### Motivation

As described in #11020 - I want to be able to retry successful workflow, given I specify explicitly `retrySuccessful` and node selector.

### Modifications

I added a condition to only fail the `FormulateRetryWorkflow` method if the original workflow is not Error/Failed **AND either `restartSuccessful==False` or `nodeFieldSelector` is empty**.

### Verification

I verified it with local deployment + added tests for both cases (retry should fail for a successful run without those extra flags, retry should work for a successful workflow with those flags).

### Notes
- ~I'm not completely sure about the error message (_"To retry either: (1) workflow must be Failed/Error (2) Provide also restartSuccessful and nodeFieldSelector value"_). Too long? understandable?~ changed it according to @terrytangyuan 's advise 🙏 
- I think that to complement this feature - we should also re-enable the "Retry" button from the UI and require the user to select which node to retry. I don't think this is a stopper for this PR as allowing to do this from CLI is already super helpful.

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -810,8 +810,12 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow, 
 func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string, parameters []string) (*wfv1.Workflow, []string, error) {
 	switch wf.Status.Phase {
 	case wfv1.WorkflowFailed, wfv1.WorkflowError:
+	case wfv1.WorkflowSucceeded:
+		if !(restartSuccessful && len(nodeFieldSelector) > 0) {
+			return nil, nil, errors.Errorf(errors.CodeBadRequest, "To retry a succeeded workflow, set the options restartSuccessful and nodeFieldSelector")
+		}
 	default:
-		return nil, nil, errors.Errorf(errors.CodeBadRequest, "workflow must be Failed/Error to retry")
+		return nil, nil, errors.Errorf(errors.CodeBadRequest, "Cannot retry a workflow in phase %s", wf.Status.Phase)
 	}
 
 	newWF := wf.DeepCopy()
--- a/cmd/argo/commands/retry.go
+++ b/cmd/argo/commands/retry.go
@@ -71,6 +71,9 @@ func NewRetryCommand() *cobra.Command {
 # Retry the latest workflow:
 
   argo retry @latest
+
+# Restart node with id 5 on successful workflow, using node-field-selector
+  argo retry my-wf --restart-successful --node-field-selector id=5
 `,
 		Run: func(cmd *cobra.Command, args []string) {
 			if len(args) == 0 && !retryOpts.hasSelector() {
--- a/docs/cli/argo_retry.md
+++ b/docs/cli/argo_retry.md
@@ -41,6 +41,9 @@ argo retry [WORKFLOW...] [flags]
 
   argo retry @latest
 
+# Restart node with id 5 on successful workflow, using node-field-selector
+  argo retry my-wf --restart-successful --node-field-selector id=5
+
 ```
 
 ### Options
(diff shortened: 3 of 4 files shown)


workflow/util/util.go
```diff
@@ -805,7 +805,9 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 	switch wf.Status.Phase {
 	case wfv1.WorkflowFailed, wfv1.WorkflowError:
 	default:
-		return nil, nil, errors.Errorf(errors.CodeBadRequest, "workflow must be Failed/Error to retry")
+		if !(restartSuccessful && len(nodeFieldSelector) > 0) {
+			return nil, nil, errors.Errorf(errors.CodeBadRequest, "To retry either: (1) workflow must be Failed/Error (2) Provide also restartSuccessful and nodeFieldSelector value")
```
terrytangyuan, line 809: This might be a bit hard to understand.

or-shachar: I know. Do you have an idea on how to simplify it? Maybe a link to docs?

terrytangyuan: Looks good

or-shachar:
```suggestion
			return nil, nil, errors.Errorf(errors.CodeBadRequest, "To retry a successful workflow, set the options restartSuccessful and nodeFieldSelector")
```

or-shachar: done

workflow/util/util.go
```diff
@@ -811,7 +811,9 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 	switch wf.Status.Phase {
 	case wfv1.WorkflowFailed, wfv1.WorkflowError:
 	default:
```
Joibel, line 813: Would it make sense to pop this new code under a `case wfv1.WorkflowSucceeded:` as that's the intended and tested use case.

workflow/util/util.go
```diff
@@ -811,7 +811,9 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
...
-		return nil, nil, errors.Errorf(errors.CodeBadRequest, "workflow must be Failed/Error to retry")
```
Joibel, line 814: And then perhaps a different error if you attempt to restart a running workflow. (If something catches that before here, I can't see it).

or-shachar: @Joibel  - what about pending / suspended workflows?

Joibel: I believe so. Previously if you got here with a pending/suspended, attempting a retry, you'd get a failure telling you it should have failed/errored. You're not intentionally changing that behaviour, so best tell people the conditions where retry is allowed.

or-shachar: Fixed... it now explicitly mentions `wfv1.WorkflowSucceeded` case, and then adds default behavior.
Thanks, it makes more sense now.

workflow/util/util.go
or-shachar: done!

Joibel:
Looks good to me. Thanks @or-shachar

terrytangyuan (approved):

isubasinghe (approved):
LGTM

PR #11416, opened 2023-07-22
docs: fix typo exit handler status to Succeeded

PanagiotisS:
### Motivation

There is a typo in the documentation https://argoproj.github.io/argo-workflows/lifecyclehook/ :

>  Put differently, an exit handler is like a workflow-level LifecycleHook with an expression of workflow.status == "Completed" or workflow.status == "Failed" or workflow.status == "Error".

It should be workflow.status == "Succeeded"  and not "Completed"

### Verification

- We can see here what the `workflow.status` values may be for exit handler: https://argoproj.github.io/argo-workflows/variables/#exit-handler

```md
### Exit Handler

| Variable | Description|
|----------|------------|
| `workflow.status` | Workflow status. One of: `Succeeded`, `Failed`, `Error` |
```

- Additionally, tested on my environment with argo-workflows v3.4.7

### Modifications

Changed word from "Completed" -> "Succeeded"

### DIFF
--- a/docs/lifecyclehook.md
+++ b/docs/lifecyclehook.md
@@ -61,4 +61,4 @@ spec:
        url: http://dummy.restapiexample.com/api/v1/employees
 ```
 
-> Put differently, an exit handler is like a workflow-level `LifecycleHook` with an expression of `workflow.status == "Completed"` or `workflow.status == "Failed"` or `workflow.status == "Error"`.
+> Put differently, an exit handler is like a workflow-level `LifecycleHook` with an expression of `workflow.status == "Succeeded"` or `workflow.status == "Failed"` or `workflow.status == "Error"`.


terrytangyuan (approved):

agilgur5:
> ### Verification

Double-checked this against the actual code (in case the docs were incorrect): [`workflow.status`](https://github.com/argoproj/argo-workflows/blob/6e209cdaa6d01180239551d6c13bf320d380b887/workflow/controller/operator.go#L371) refers to [`WorkflowPhase`](https://github.com/argoproj/argo-workflows/blob/6e209cdaa6d01180239551d6c13bf320d380b887/pkg/apis/workflow/v1alpha1/workflow_phase.go#L4).

`WorkflowPhase` would indeed only have one of those 3 upon exit, and `Completed` is indeed _not_ one of them.

Thanks for the fix!

PR #11440, opened 2023-07-25
fix: Devcontainer resets /etc/hosts (#11439)

### LINKED ISSUE #11439: The container created by Devcontainer keeps resetting /etc/hosts
### Pre-requisites

# What happened
Whenever I reboot my computer, /etc/hosts entries added by Devcontainer are gone

```bash
devcontainer up --workspace-folder .
devcontainer exec --workspace-folder . /bin/bash
cat /etc/hosts

#127.0.0.1	localhost
#::1	localhost ip6-localhost ip6-loopback
#fe00::0	ip6-localnet
#ff00::0	ip6-mcastprefix
#ff02::1	ip6-allnodes
#ff02::2	ip6-allrouters
#172.17.0.1	host.docker.internal
#172.17.0.2	be7abe611b1c
#127.0.0.1 dex
#127.0.0.1 minio
#127.0.0.1 postgres
#127.0.0.1 mysql
#127.0.0.1 azurite

reboot
devcontainer up --workspace-folder .
devcontainer exec --workspace-folder . /bin/bash
cat /etc/hosts

#127.0.0.1	localhost
#::1	localhost ip6-localhost ip6-loopback
#fe00::0	ip6-localnet
#ff00::0	ip6-mcastprefix
#ff02::1	ip6-allnodes
#ff02::2	ip6-allrouters
#172.17.0.1	host.docker.internal
#172.17.0.2	be7abe611b1c
```
# What I expect
/etc/hosts should be preserved

# Why it happens
That's Because Devcontainer's container is running with bridge networking 
https://stackoverflow.com/questions/27521678/etc-hosts-file-of-a-docker-container-gets-overwritten

# How to fix It
~~https://code.visualstudio.com/remote/advancedcontainers/start-processes~~
~~Add postStartCommand field that adds /etc/hosts entries to devcontainer.json~~
Add --add-host options as runArgs in devcontainer.json

### Version

latest

```YAML
none
```

```text
none
```

```text
none
```

binkoni:
Fixes #11439 

### Motivation

A container created by Devcontainer keeps resetting /etc/hosts whenever I reboot or restart the container.

### Modifications

.devcontainer/devcontainer.json
.devcontainer/pre-build.sh
~~.devcontainer/post-start.sh~~

### Verification

To ensure this issue is fixed, I manually tested by running devcontainer commands that after I made this change.
I also tested with VsCode devcontainer extension.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,14 @@
   "hostRequirements": {
     "cpus": 4
   },
-  "runArgs": ["--add-host=host.docker.internal:host-gateway"],
+  "runArgs": [
+    "--add-host=host.docker.internal:host-gateway",
+    "--add-host=dex:127.0.0.1",
+    "--add-host=minio:127.0.0.1",
+    "--add-host=postgres:127.0.0.1",
+    "--add-host=mysql:127.0.0.1",
+    "--add-host=azurite:127.0.0.1"
+  ],
   "onCreateCommand": ".devcontainer/pre-build.sh",
   "workspaceMount": "source=${localWorkspaceFolder},target=/home/vscode/go/src/github.com/argoproj/argo-workflows,type=bind",
   "workspaceFolder": "/home/vscode/go/src/github.com/argoproj/argo-workflows",
--- a/.devcontainer/pre-build.sh
+++ b/.devcontainer/pre-build.sh
@@ -1,13 +1,6 @@
 #!/usr/bin/env sh
 set -eux
 
-# Add hosts
-sudo bash -c 'echo "127.0.0.1 dex" >> /etc/hosts'
-sudo bash -c 'echo "127.0.0.1 minio" >> /etc/hosts'
-sudo bash -c 'echo "127.0.0.1 postgres" >> /etc/hosts'
-sudo bash -c 'echo "127.0.0.1 mysql" >> /etc/hosts'
-sudo bash -c 'echo "127.0.0.1 azurite" >> /etc/hosts'
-
 # install kubernetes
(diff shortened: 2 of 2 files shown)


.devcontainer/post-start.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+set -eux
+
+# Add hosts
+sudo bash -c 'echo "127.0.0.1 dex" >> /etc/hosts'
+sudo bash -c 'echo "127.0.0.1 minio" >> /etc/hosts'
+sudo bash -c 'echo "127.0.0.1 postgres" >> /etc/hosts'
+sudo bash -c 'echo "127.0.0.1 mysql" >> /etc/hosts'
+sudo bash -c 'echo "127.0.0.1 azurite" >> /etc/hosts'
```
agilgur5, line 9: what if these entries already exist in `/etc/hosts`? these may end up getting listed twice as a result. that won't break anything, but may want to check the file first

binkoni: Thanks for the review! I think it's impossible for the entries to already exist. Because when the container restarts it wipes the entries added by the script. But when we suppose that the entries already exist for some reason, that will add duplicate entries.
So I changed the code to make it more reliable by adding --add-host options as runArgs in devcontainer.json instead of using postStartCommand

agilgur5: Oh good idea on `--add-host`!

If someone configured different networking or used a different client, there could be some overlap. But in general, just better to be safer if we can make the code purely additive without much effort

agilgur5 (approved):
Good change! Thanks for the contribution!

terrytangyuan (approved):

PR #11444, opened 2023-07-25
fix: UI toolbar sticky

### LINKED ISSUE #11442: UI: The position sticky on the workflow toolbar is not working
### Pre-requisites

Jul-25-2023 02-41-53
- The sticky property applied to the toolbar when the workflow list checkbox is selected is not applied, so the toolbar is not visible when scrolling down.

### How to fix the problem?
The Page component provided by argo-ui has `overflow:hidden` as a default property in the page__content-wrapper, so sticky cannot be applied.

Alternatively, you can use `position:fixed` and classname to make it behave like sticky.
If you have a better way, I'd appreciate any suggestions :)

### Version

latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

yunwoo-yu:
Fixes #11442 

### Motivation

The `position:sticky` in the workflows-toolbar is not working.

### Modifications

For a better UX, we changed the workflows-toolbar component from not being position sticky to being able to see the toolbar after scrolling.

- before

Jul-25-2023 02-41-53

- after

Jul-25-2023 20-01-13

### DIFF
--- a/ui/src/app/workflows/components/workflows-list/workflows-list.scss
+++ b/ui/src/app/workflows/components/workflows-list/workflows-list.scss
@@ -1,226 +1,231 @@
 @import 'node_modules/argo-ui/src/styles/config';
 
 .workflows-list {
-  padding: 1em;
+    padding: 1em;
 
-  &__row-container {
-      a {
+    &__row-container {
+        a {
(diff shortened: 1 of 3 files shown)


terrytangyuan (approved):
Thanks!

PR #622, opened 2017-12-20
Implementation for resource templates and resource wait/fail conditions

jessesuen:
This implements issue #606.

It introduces a new template type, resource, which has the following fields:
```
// ResourceTemplate is a template subtype to manipulate kubernetes resources
type ResourceTemplate struct {
	// Action is the action to perform to the resource.
	// Must be one of: create, apply, delete
	Action string `json:"action"`

	// Manifest contains the kubernetes manifest
	Manifest string `json:"manifest"`

	// WaitCondition is a label selector expression which describes the conditions
	// of the k8s resource in which it is acceptable to proceed to the following step
	WaitCondition string `json:"waitCondition,omitempty"`

	// FailCondition is a label selector expression which describes the conditions
	// of the k8s resource in which the step was considered failed
	FailCondition string `json:"failCondition,omitempty"`
}
```
It has support for "wait conditions" which block the step from proceeding until the expression evaluates true, as well as "fail conditions", which fail the step immediately if evaluated true.

PR #11450, opened 2023-07-25
docs: add Resource Template page

agilgur5:
### Motivation

- previously, the Template Types section of the User Guide never mentioned Resource Templates
  - the Walk Through section of Getting Started did have a page on this though
    - so just link the two together as that one already documents it in detail
    
- I felt like something was missing in this section, finally realized what 😅
  - Was trying to find this doc while answering a question [on Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1690286713376469?thread_ts=1690276970.326629&cid=C01QW9QSSSK) and went to Template Types only to not see it 💭 

### Modifications

- create a new Resource Template page under Template Types of the User Guide
  - just link to the existing docs in the Walk Through 
  - also add a version as all the Template Type pages have one
    - this was released in a 2.0 alpha as part of #622
- add `v2.0` to the `.spelling` ignore file

### Verification

Used `make docs-serve` and confirmed it works on local. Screenshot:

Screenshot 2023-07-27 at 1 42 42 PM

### Misc Notes

Not the first time I've found myself thinking that some of the docs could be consolidated and go to one page only to realize that the relevant docs are on another page (e.g. #11296). I am perhaps a more unique user as both an Argo Operator and User (and Developer nowadays), so some of those are perhaps a given. But the Walk Through seems like it addresses the same audience as the User Guide, no?

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -198,6 +198,7 @@ v1.2
 v1.3
 v1.24
 v2
+v2.0
 v2.10
 v2.11
 v2.12
@@ -236,6 +237,6 @@ devenv
 vendored
 nix.conf
 LDFlags
-dev 
+dev
 vendorSha256
 dependabot
--- a/docs/resource-template.md
+++ b/docs/resource-template.md
@@ -1,0 +1,5 @@
+# Resource Template
+
+> v2.0
+
+See [Kubernetes Resources](../walk-through/kubernetes-resources.md).
--- a/docs/resource-template.md
+++ b/docs/resource-template.md
@@ -2,4 +2,4 @@
 
 > v2.0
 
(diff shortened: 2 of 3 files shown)


agilgur5:
Strange, CI is failing on spellcheck:

>     docs/resource-template.md
>        3 | > v2.0 

Yet this is the same syntax used everywhere else 🤔

agilgur5:
> Strange, CI is failing on spellcheck:

Oh, figured it out. Each version actually has a manual ignore in [`.spelling`](https://github.com/argoproj/argo-workflows/blob/90930ab88b18b7fba3074cdc06059eb6460b50d9/.spelling#L200) and this one is not in there already

terrytangyuan (approved):

Issue #10285, opened 2022-12-28
argo3.3.10 termante workflow then retry meet error: Workflow operation error


yeicandoit:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

https://user-images.githubusercontent.com/22470950/209812826-6c417bff-ad4d-41a0-88af-f30a9d512834.mp4

when the workflow is running, stop/terminate it, then retry it. the workflow will meet issue: Workflow operation error.
I except the workflow to re-run correctly

### Version

v3.3.10

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: dag-diamond-1
spec:
  entrypoint: diamond
  templates:
  - name: echo
    inputs:
      parameters:
      - name: message
    container:
      image: alpine:3.7
      command: [sleep, "600s"]
  - name: diamond
    dag:
      tasks:
      - name: A
        template: echo
        arguments:
          parameters: [{name: message, value: A}]
      - name: B
        dependencies: [A]
        template: echo
        arguments:
          parameters: [{name: message, value: B}]
      - name: C
        dependencies: [A]
        template: echo
        arguments:
          parameters: [{name: message, value: C}]
      - name: D
        dependencies: [B, C]
        template: echo
        arguments:
          parameters: [{name: message, value: D}]
```

### Logs from the workflow controller

```text
time="2022-12-28T10:37:43.765Z" level=info msg="Create pods 201"
time="2022-12-28T10:37:43.766Z" level=info msg="Created pod: dag-diamond-1-lkhtv.A (dag-diamond-1-lkhtv-1695834569)" namespace=enos workflow=dag-diamond-1-lkhtv
[...]

PR #11451, opened 2023-07-26
fix: add guard against NodeStatus. Fixes #11102

### LINKED ISSUE #11102: Default initialised NodeStatus causes random false failures.
### Pre-requisites

When a nodeID is missing from the NodeStatus hashmap, this should cause an error and that condition should be handled. 
Unfortunately thanks to Go's design of maps, a default initialised value of a NodeStatus is returned instead of panicking (which is arguably the sane thing to do). Instead of assuming a value for the NodeStatus, either mark that node as an error or wait till the field has been populated. 

### Version

:latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #11102 and also issues in the same class (such as #10285)

### Motivation

Adds a guard to all NodeStatus accesses, this enforces users to check the HashMap before using a value in it. 
This is needed because the HashMap will return a default initialised NodeStatus. Default initialised NodeStatus have triggered edge case bugs for our customers. This fix existed internally in Pipekit for a while now and was able to solve the customers issues with container sets. 

### Modifications

Fairly simple changes were added. 
  * panic when we are absolutely certain a NodeStatus should exist but we see an error
  * bubble up errors instead of relying on default initalised NodeStatus

### Verification

  * Manual testing


isubasinghe:
Simple changes, just makes error handling mandatory.

We encountered too many random failing container sets without these fixes. 

Now invalid access errors are just bubbled upstream

isubasinghe:
Conflicts with #11493

Joibel (approved):
Looks good to me.

terrytangyuan (approved):

agilgur5 (commented):
LGTM. Simple change, but so many places needed changing; nice work getting through them!

For my own edification, did the edge case bugs primarily happen on retried workflows? Reading through this code, it seems that retried workflows are the only place where Nodes get deleted?
Please correct me if I'm wrong, just trying to understand when/how this class of bugs would happen

isubasinghe:
@agilgur5 some areas of the code weren't checking if the NodeStatus actually existed, thanks to Go's (in)sane way of dealing with missing entries (zero initialised), we were getting valid structs. 

```go
x := make(map[string]NodeStatus)
assert x["blah"] == DefaultInitialisedNodeStatus() // function doesn't actually exist, just making a point 
```

This didn't just happen for retried workflows, for some reason workflows with container sets pretty much randomly failed all the time. I believe this was because some code path was falsely relying on the default initalised values. 

This change now forces you to explicitly handle the case where a NodeStatus is missing.

isubasinghe:
> This is the part I was curious about. As I would think the only time an uninitialized value would be accessed would be if the key previously existed, i.e. it was deleted. Otherwise, I would think that the Nodes map would only grow (until it is GC'd), and so a non-existent key should never occur.
Or to put it differently, trying to understand the root cause of when this would happen. Not just that the node could not exist, but why it would not exist

This is a good question, unfortunately one I do not have a definitive answer to. I believe that the codebase is somewhat loosely consistent, code paths are allowed to fail as long as eventually they succeed. That is my educated guess anyway. 
I dealt with this issue for months before realising that this was the issue. This simple change was likely the most difficult PR I opened in the codebase.

agilgur5:
> This is a good question, unfortunately one I do not have a definitive answer to.

Gotcha. Was hoping you might know. Guess we still have to watch out for the root cause somewhere (which may give this new error now at least!). If it's not just impacting retries, there might be a thread safety issue somewhere 🤔 

> This simple change was likely the most difficult PR I opened in the codebase.

I know that feeling all too well 😅 Like to use these types of fixes as great examples that amount of code is often not the most impactful contribution!

agilgur5:
Jotting down some notes here as I attempted to find the root cause:
1. Searched the codebase for places where `NodeStatus` could have been accessed or removed (keywords: "[delete](https://github.com/search?q=repo%3Aargoproj%2Fargo-workflows+delete++language%3AGo&type=code&p=1&l=Go)" and "[Status.Nodes](https://github.com/search?q=repo%3Aargoproj%2Fargo-workflows+status.nodes++language%3AGo&type=code)").
1. `delete` indeed seems only used during retries
1. The only other place I could find where nodes are removed is in [this line](https://github.com/argoproj/argo-workflows/blob/f5e31f8f36b32883087f783cb1227490bbe36bbd/workflow/packer/packer.go#L80) in `compressWorkflow`, which I believe is _only_ used for [node status offload](https://argoproj.github.io/argo-workflows/offloading-large-workflows/).
    - Which does lead to a question if this has been reproduced _without_ node status offload?
1. The only other potential anomaly is that some places seem to lock when modifying the nodes and others don't. I did not inspect carefully why that is though.
    - I also thought that the goroutines did not share a workflow between each other (roughly [this codepath](https://github.com/argoproj/argo-workflows/blob/f5e31f8f36b32883087f783cb1227490bbe36bbd/workflow/controller/controller.go#L314)), so thought locking wouldn't be necessary. But I may have misinterpreted the code; the controller is the most complex part of the codebase after all.
    - It would be nice if the Go compiler had more advanced race condition detection (like Rust, Pony, etc). I also don't think(?) we use Go's [race detector](https://go.dev/doc/articles/race_detector)
    - Even if there were some race conditions with modifying, that should never result in a Node _not existing_, so those cases don't really explain it either 😕

Issue #11454, opened 2023-07-26
Expose cronWorkflowWorkers as config item.


sarabala1979:
# Summary
Currently `cronWorkflowWorkers' was hardcoded. In some scenarios, if the user is using cron workflow heavily, they need more workers to process the cronworkflow.

What change needs making?

# Use Cases

When would you use this?

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

tooptoop4:
does it mean if I have 9 different cronworkflows set to run hourly that one will be missed?

agilgur5:
Added in #11457

agilgur5:
> does it mean if I have 9 different cronworkflows set to run hourly that one will be missed?

I believe it's more that there are 8 threads running by default. So if you have 9 trigger simultaneously, they all get added to the queue and up to 8 can be processed in parallel (also depends on number of cores and other things running etc).
A CronWorkflow primarily [creates Workflows](https://argoproj.github.io/argo-workflows/cron-workflows/#workflowspec-and-workflowmetadata), so I'd imagine all 8 could be processed in less than a second before the 9th then gets processed.

@sarabala1979 feel free to correct me as you obviously know this code better than me!

agilgur5:
(^that is also given a single controller instance without any [sharding](https://argoproj.github.io/argo-workflows/scaling/#sharding))

PR #11457, opened 2023-07-27
feat: Expose the Cron workflow workers as argument

sarabala1979:
Fixes #TODO


juliev0 (approved):

terrytangyuan (approved):

agilgur5:
Fixes #11454 (though as arg, not configmap option)

agilgur5:
We should add some docs for this new feature as well.
Also, I might be wrong, but I think some of the other flags are not documented? Flags could be a use case for codegen

agilgur5:
> We should add some docs for this new feature as well.

Docs were added for this feature in #11731

PR #11470, opened 2023-07-28
feat: Use WorkflowTemplate/ClusterWorkflowTemplate Informers when validating CronWorkflows

### LINKED ISSUE #11372: Cron Workflow Controller should use WorkflowTemplateInformer and ClusterWorkflowTemplateInformer
# Summary

It seems to me that the CronWorkflow validation code is unnecessarily calling the K8S API to "GET WorkflowTemplate" [here](https://github.com/argoproj/argo-workflows/blob/master/workflow/cron/operator.go#L114) (for CronWorkflows that use WorkflowTemplates) when it could be using the Informer.

---

juliev0:
Fixes #11372 

### Motivation

When a CronWorkflow gets run it first gets validated. If it contains a WorkflowTemplate or CronWorkflowTemplate, it goes to the K8S API to get the spec to do the validation, rather than using what's already been stored locally in the Informer caches.

### Modifications

Uses the WorkflowTemplate Informer and ClusterWorkflowTemplate Informer instead.

### Verification

- Ran locally with a CronWorkflow which uses a WorkflowTemplate, as well as one which uses a CronWorkflowTemplate.


examples/cluster-workflow-template/clustertemplates.yaml
```diff
@@ -135,7 +135,7 @@ kind: ClusterWorkflowTemplate
 metadata:
   name: cluster-workflow-template-submittable
 spec:
-  entryPoint: whalesay-template
+  entrypoint: whalesay-template
   arguments:
     parameters:
```
juliev0, line 140: not related but this is a typo

workflow/controller/informer/cluster_workflow_template_convert.go
```diff
@@ -25,10 +39,25 @@ func objectToClusterWorkflowTemplate(object runtime.Object) (*wfv1.ClusterWorkfl
...
+	cwfTmpl, err := interfaceToClusterWorkflowTemplate(obj)
+	if err != nil {
+		return nil, err
+	}
+	return cwfTmpl, nil
+}
+
+func NewClusterWorkflowTemplateFromInformerGetter(cwftmplInformer wfextvv1alpha1.ClusterWorkflowTemplateInformer) templateresolution.ClusterWorkflowTemplateGetter {
+	return &ClusterWorkflowTemplateFromInformerGetter{cwftmplInformer: cwftmplInformer}
 }
```
juliev0, line 63: let me know if there's a better place to put this struct...

terrytangyuan (approved):
LGTM. Thank you!

PR #11480, opened 2023-07-29
fix: Datepicker Style Malfunction Issue. Fixes #11476

### LINKED ISSUE #11476: UI: Datepicker Style Malfunction Issue
# Summary

I am creating this issue to address the broken style of the date filter in the workflow. The current state of the date filter's appearance is causing issues with its functionality, making it challenging to use the date filtering feature effectively.

I have already checked issue #11421, which seems related to the problem. However, before resolving that specific issue, I believe it is crucial to prioritize the improvement of the basic date picker style.

스크린샷 2023-07-29 오후 5 20 43

Is it okay if I solve this issue myself?

zel0rd:
Fixes #11476 

### Motivation

The date filter is a crucial element when viewing the list of workflows. However, when clicking on the date-picker to change the date, the styles get messed up, resulting in a UI state where the filter functionality cannot be used properly. Moreover, the close button overlaps with the selected date, making it difficult to select the desired date.

### Modifications

This issue is more related to layout and styles rather than a major functional problem. Therefore, we plan to make the following two modifications:

1. Move the close button to the right so that it doesn't overlap with the date-picker.
2. Adjust the styles of the date-picker using the calendar to ensure proper functionality.

스크린샷 2023-07-29 오후 7 22 32
KakaoTalk_20230729_153542342

### DIFF
--- a/ui/src/app/workflows/components/workflow-filters/workflow-filters.scss
+++ b/ui/src/app/workflows/components/workflow-filters/workflow-filters.scss
@@ -26,3 +26,10 @@
     background-color: transparent;
     border: 0;
 }
+
+.wf-filters-container__content {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    justify-content: space-between;
+}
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):
Thank you!

PR #11493, opened 2023-08-01
fix: always fail dag when shutdown is enabled. Fixes #11452

### LINKED ISSUE #11452: Failed workflows can never be diagnosed as failed sometimes, meaning it is impossble to stop
### Pre-requisites

Expected: 
When the user issues a stop/terminate request, the workflow should be marked as Failed. 

Happens:
When the user issues a stop/terminate (depending the progress to the targetTasks) ,the workflow will be never marked as Failed. 

Conditional make this easy to replicate, but I think this is a general issue. 

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: dag-diamond-coinflip-
  annotations:
      This is an example of coin flip defined as a DAG.
      You can also run it in Python: https://couler-proj.github.io/couler/examples/#dag
spec:
  entrypoint: diamond
  templates:
  - name: diamond
    retryStrategy:
      limit: "10"
    dag:
      tasks:
      - name: A
        template: coinflip
      - name: B
        depends: "A"
        template: main
      - name: C
        depends: "A"
        template: main
      - name: D
        depends: "B && C"
        template: coinflip

  - name: coinflip
    retryStrategy:
      limit: "10"
    steps:
    - - name: flip-coin
        template: flip-coin
    - - name: heads
        template: heads
        when: "{{steps.flip-coin.outputs.result}} == heads"
      - name: tails
        template: coinflip
        when: "{{steps.flip-coin.outputs.result}} == tails"

  - name: flip-coin
    retryStrategy:
      limit: "10"
    script:
      image: python:alpine3.6
      command: [python]
      source: |
        import random
        result = "heads" if random.randint(0,1) == 0 else "tails"
        print(result)

  - name: heads
    retryStrategy:
      limit: "10"
    container:
      image: alpine:3.6
      command: [sh, -c]
      args: ["echo \"it was heads\""]

  - name: main
    retryStrategy:
      limit: "10"
    containerSet:
      containers:
        - name: a
          image: ubuntu:20.04
          command:
            - "sleep"
            - "20"
        - name: b
          image: ubuntu:20.04
          command:
            - "sleep"
            - "20"
          dependencies: ["a"]
        - name: c
          image: ubuntu:20.04
          command:
            - "sleep"
            - "20"
          dependencies: ["a"]
        - name: d
          image: ubuntu:20.04
          command:
            - "sleep"
            - "20"
          dependencies: ["b", "c"]
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #11452

### Motivation

Ensure that stopped workflows are always marked as Failed (as expected). 

### Modifications

When shutdown is enabled, we don't do the DAG exploration. 

### Verification

Manually tested


isubasinghe:
Conflicts with: #11451

workflow/controller/operator_test.go
```diff
@@ -6843,7 +6843,7 @@ func TestNoPodsWhenShutdown(t *testing.T) {
 
 	node := woc.wf.Status.Nodes.FindByDisplayName("hello-world")
 	if assert.NotNil(t, node) {
-		assert.Equal(t, wfv1.NodeSkipped, node.Phase)
+		assert.Equal(t, wfv1.NodeFailed, node.Phase)
```
isubasinghe, line 6846: Double check this. 

Not sure why it was NodeSkipped before

agilgur5: If the Pod never ran at all, NodeSkipped could make sense. The source (this is the test) is marking it prior to it running.

isubasinghe: Okay reverting this

Joibel (approved):
Makes sense to me

workflow/controller/dag.go
```diff
@@ -140,15 +146,24 @@ func (d *dagContext) assessDAGPhase(targetTasks []string, nodes wfv1.Nodes) wfv1
 		targetTaskPhases[d.taskNodeID(task)] = ""
 	}
 
+	boundaryNode, err := nodes.Get(d.boundaryID)
```
agilgur5, line 149: for context for other readers, the `nodes.Get` modifications here are just error guards a la #11451. no real semantic changes

agilgur5 (approved):
LGTM. just adds an early return and error guards

workflow/controller/dag.go
```diff
@@ -128,7 +128,13 @@ func (d *dagContext) getTaskNode(taskName string) *wfv1.NodeStatus {
 }
 
 // assessDAGPhase assesses the overall DAG status
-func (d *dagContext) assessDAGPhase(targetTasks []string, nodes wfv1.Nodes) wfv1.NodePhase {
+func (d *dagContext) assessDAGPhase(targetTasks []string, nodes wfv1.Nodes, isShutdown bool) (wfv1.NodePhase, error) {
+	// we cannot rely upon the DAG traversal alone
+	// I believe conditionals, self references and ContinuesOn (every one of those features in unison) makes this an undecidable problem.
```
terrytangyuan, line 133: Remove "I"

workflow/controller/dag.go
```diff
@@ -128,7 +128,13 @@ func (d *dagContext) getTaskNode(taskName string) *wfv1.NodeStatus {
...
+	// However, we can just use the isShutdown to automatically fail the DAG.
```
terrytangyuan, line 134:
```suggestion
	// We cannot only rely on the DAG traversal. Conditionals, self-references, 
	// and ContinuesOn (every one of those features in unison) make this an undecidable problem.
	// However, we can just use isShutdown to automatically fail the DAG.
```

terrytangyuan (approved):

workflow/controller/operator_test.go
agilgur5: #13214 is changing this to `NodeFailed` for a good reason: `NodeSkipped` breaks retries. Otherwise I'm still not sure why it was `NodeSkipped` initially.

See also #11849 for the "Cancelled" status that would more accurately describe this

PR #11504, opened 2023-08-02
fix: update nix hash

isubasinghe:
Update nix hash to match dependencies in use.

### DIFF
--- a/dev/nix/flake.nix
+++ b/dev/nix/flake.nix
@@ -180,7 +180,7 @@
               pname = package.name;
               inherit (package) version;
               inherit src;
-              vendorSha256 = "sha256-OKUiHkVZ/rfjPFs7Md5WDa5K1SNJQvLMxlYClUe7Umk=";
+              vendorSha256 = "sha256-beZOlw9QaHf+wwwIMAj3OzvXUK+d539+zdbCdGA54v0=";
               doCheck = false;
             };
 


dev/nix/flake.nix
```diff
@@ -180,7 +180,7 @@
               pname = package.name;
               inherit (package) version;
               inherit src;
-              vendorSha256 = "sha256-OKUiHkVZ/rfjPFs7Md5WDa5K1SNJQvLMxlYClUe7Umk=";
+              vendorSha256 = "sha256-beZOlw9QaHf+wwwIMAj3OzvXUK+d539+zdbCdGA54v0=";
```
terrytangyuan, line 183: How is this generated (any docs?) and can this be automated?

isubasinghe: Yes the Nix docs show how this can be updated. 

It is a relatively trivial process to do so. Run `nix build`, nix will complain about the hash mismatch, use the correct hash from this output. 

I could write a tool to update this hash automatically. I can make the hash checking a part of the CI process (via `make lint`).

isubasinghe: https://argoproj.github.io/argo-workflows/running-nix/#why-am-i-getting-a-vendorsha256-mismatch

terrytangyuan: Sounds good.

terrytangyuan (changes requested):
We'll wait for the automation for this instead of updating manually every time

PR #11505, opened 2023-08-02
feat: update nix version and build with ldflags

isubasinghe:
Updates to mockery v2.26.0 (what is used in Makefile). 
Adds ldflags to ensure stdout is the same as official package when the command `mockery --version` is run.

### DIFF
--- a/dev/nix/flake.nix
+++ b/dev/nix/flake.nix
@@ -196,19 +196,20 @@
 
             mockery = pkgs.buildGoModule rec { 
               pname = "mockery";
-              version = "2.10.0"; # upgrade this in the Makefile if upgraded here
+              version = "2.26.0"; # upgrade this in the Makefile if upgraded here
 
               src = pkgs.fetchFromGitHub {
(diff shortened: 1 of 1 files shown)


dev/nix/flake.nix
```diff
@@ -186,16 +186,20 @@
 
             mockery = pkgs.buildGoModule rec { 
               pname = "mockery";
-              version = "2.10.0"; # upgrade this in the Makefile if upgraded here
+              version = "2.26.0"; # upgrade this in the Makefile if upgraded here
```
terrytangyuan, line 189: How was this version not consistent with Makefile in the first place?

isubasinghe: Frankly not sure, I think I started working on this a while ago so the version changed while I was developing. 

Just so you are aware the versions being different is not too worrying to me because the GitHub CI tests will differ in output and fail any PR that uses the Nix dependencies incorrectly.

terrytangyuan (approved):

PR #11515, opened 2023-08-03
docs: "Other" use-case should always be last

agilgur5:
### Motivation

- normal convention to leave "the rest" for last instead of in the middle

### Modifications

- move "Other" to be last under "Use Cases"
  - add a comment to hopefully make sure it stays last

- slightly modify the page description as well
  - previously it had a self-explanatory description, this way it specifies that the listed use-cases are not exhaustive
  - also `awesome-argo` is not listed in the specific use-case docs, so this is actually a difference
  - this also leaves the door open to list more unconventional use-cases on this page

### Verification

n/a


terrytangyuan (approved):

agilgur5:
Woops I accidentally hit the update button so guess we'll see if it passes this time 😅

PR #11557, opened 2023-08-11
feat(cli): option to delete pending workflows. fixes #11555

### LINKED ISSUE #11555: CLI: option to delete/list only workflows in 'pending' state
ie  want cli like:
`argo delete --pending`

to delete all workflows that are pending (ie before moved to running state)

Paritosh-Anand:
### Motivation

This is my first contribution to Argo Workflows. I am trying to fix the issue #11555
IMO it is good to have `argo delete` with `--pending` for cleaning up stuck or pending tasks.

### Modifications

I added support for `--pending` flag in `argo delete` command. This flag will delete all the workflows which are in pending state.

### Verification

WIP


terrytangyuan (commented):
I am not 100% whether this would be necessary since you can do this via `argo list --field-selector status.phase=Pending`

tooptoop4:
if I'm not mistaken '--field-selector status.phase=Pending' would only filter **pods** that are pending (which would be from an already **running workflow**), this PR seems like it would be the only way to filter on **pending workflows** (ie where they have not started any pod)

agilgur5:
> I am not 100% whether this would be necessary since you can do this via `argo list --field-selector status.phase=Pending`

@terrytangyuan So I thought this might be an acceptable change given that `--completed` and `--running` are already available flags. Otherwise I would prefer not to add more flags as well.
If you want to deprecate those flags or make a decision that new status flags would not be accepted (as other flags can be used equivalently), I'd be fine with that too. Just wanna make sure we're consistent one way or the other

cmd/argo/commands/delete.go
```diff
@@ -36,7 +36,7 @@ func NewDeleteCommand() *cobra.Command {
   argo delete @latest
 `,
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) == 0 && !(all || allNamespaces || flags.completed || flags.resubmitted || flags.prefix != "" || flags.labels != "" || flags.fields != "" || flags.finishedAfter != "") {
+			if len(args) == 0 && !(all || allNamespaces || flags.completed || flags.resubmitted || flags.pending || flags.prefix != "" || flags.labels != "" || flags.fields != "" || flags.finishedAfter != "") {
 				cmd.HelpFunc()(cmd, args)
 				os.Exit(1)
 			}
```
agilgur5, line 42: [line 55](https://github.com/argoproj/argo-workflows/blob/28ffd1c442260dc8ef222062c41d2c33e880bd65/cmd/argo/commands/delete.go#L55) below would also need modifications

cmd/argo/commands/list.go
```diff
@@ -87,6 +88,7 @@ func NewListCommand() *cobra.Command {
 	command.Flags().BoolVar(&listArgs.completed, "completed", false, "Show completed workflows. Mutually exclusive with --running.")
 	command.Flags().BoolVar(&listArgs.running, "running", false, "Show running workflows. Mutually exclusive with --completed.")
 	command.Flags().BoolVar(&listArgs.resubmitted, "resubmitted", false, "Show resubmitted workflows")
+	command.Flags().BoolVar(&listArgs.pending, "pending", false, "Show pending workflows")
```
agilgur5, line 91: Technically speaking, I believe this is mutually exclusive with the other two statuses as well, as a workflow can only have [one phase](https://github.com/argoproj/argo-workflows/blob/6a1c7ef8abb3e9f6b310f9b35de94115115cf62b/pkg/apis/workflow/v1alpha1/workflow_phase.go#L6).

So this would need some more modifications

cmd/argo/commands/list.go
```diff
@@ -27,6 +27,7 @@ type listFlags struct {
 	completed     bool
 	running       bool
 	resubmitted   bool
+	pending       bool
```
agilgur5, line 30: Should add [a test](https://github.com/argoproj/argo-workflows/blob/28ffd1c442260dc8ef222062c41d2c33e880bd65/cmd/argo/commands/list_test.go#L43) for this as well

agilgur5 (changes requested):
If we choose to go forward with this PR, there would still be some changes necessary. See below in-line comments

cmd/argo/commands/delete.go
Paritosh-Anand: added. thanks for pointing it out.

terrytangyuan:
Is this superseded by another PR?

agilgur5:
#11577 implemented a variant of this, but technically _both_ can be done. Per my [above comment](https://github.com/argoproj/argo-workflows/pull/11557#issuecomment-1678434961), I'd defer to you if we want to implement both or just no longer accept features for other statuses and point to the `--status` flag instead.

Paritosh-Anand:
I believe the #11577 is more meaningful and having multiple approaches will confuse the users. 
so closing this PR.

agilgur5:
Thanks for your efforts @Paritosh-Anand!

Issue #10620, opened 2023-03-01
Once-off scheduled Workflow execution


byrneo:
# Summary

There are times where you may wish to schedule just a *single* Workflow execution at a specific, future point in time. E.g. I may wish to execute Workflow-A on `Wed  1 Mar 2023 14:05:30 GMT`, Workflow-B on `Fri  3 Mar 2023 20:10:05 GMT` etc

[Cron Workflows](https://argoproj.github.io/argo-workflows/cron-workflows/) are a great fit for *recurring* Workflow executions e.g. once every hour, week etc etc but, by design, don't enable a single invocation for a **specific** future date/time.

What is being suggested here is a Workflow type that would mimic the feature of linux's [at](https://linux.die.net/man/1/at) command where you can schedule a job for execution at a future point in time.

# Use Cases

Execute Workflows, individually, on a scheduled basis with **no** recurrence.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

Kitanotori:
How about using some notification engine (e.g. [Cronicle](https://github.com/jhuckaby/Cronicle/blob/master/docs/WebUI.md#schedule-tab)) for sending scheduled notifications to Argo Events and then trigger Workflows from there?

Kitanotori:
Not sure how difficult it would be to implement such functionality into Workflow. Having scheduled execution for Workflow, would be nice to have feature, though.

caelan-io:
Another option until this is a native feature in Argo WF: create a plugin step that uses a script or notification engine integration to schedule the workflow:
https://argoproj.github.io/argo-workflows/plugins/
> a plugin just responds to RPC HTTP requests
https://argoproj.github.io/argo-workflows/executor_plugins/

eduardodbr:
Hi! 
I would also need to run workflows on a schedule so did a bit of research on how this could be implemented using cron workflows:

- Define the year on the cron expression:  Currently argo workflows uses https://github.com/robfig/cron to schedule  cron jobs but this lib does not support setting the year. There is a [PR](https://github.com/robfig/cron/pull/340) opened for it,  but there is no response from the maintainer. The project seems stale for the last 2.5 years.
- Create a new config field to set the max number of runs for a CronWorkflow: the controller would stop scheduling the cronjob once it reached the configured max runs. This seems fairly simple to implement. 

Other than this I can only see a new controller being created.

[...]

PR #11559, opened 2023-08-11
docs: add cron wf improvements proposal

eduardodbr:
### Motivation

Cron workflow improvements proposal as prompted by #10620

### DIFF
--- a/docs/proposals/cron-wf-improvement-proposal.md
+++ b/docs/proposals/cron-wf-improvement-proposal.md
@@ -1,0 +1,103 @@
+# Proposal for Cron Workflows improvements
+
+## Introduction
+
+Currently, cronworkflows are a great resource if we want to run recurring tasks to infinity. However, it is missing the ability to customize it, for example define how many times a workflow should run or how to handle multiple failures. I believe argo workflows would benefit of having more configuration options for cron workflows, to allow to change its behavior based on the result of its child’s success or failures. Below I present my thoughts on how we could improve them, but also some questions and concerns on how to properly do it. 
+
+## Proposal 
+
+This proposal discusses the viability of adding 2 more fields into the cron workflow configuration:
+
+```
+RunStrategy:
+	maxSuccess:
+	maxFailures:
+```
+
+`maxSuccess` - defines how many child workflows must have success before suspending the workflow schedule
+
+`maxFailures` - defines how many child workflows must fail before suspending the workflow scheduling. This may englobe `Failed` workflows, `Errored` workflows or spec errors.
+
+For example, if we want to run a workflow just once, we could just set:
+
+```
+RunStrategy:
+	maxSuccess: 1
+```
+
+This configuration will make sure the controller will keep scheduling workflows until one of them finishes with success.
+
+As another example, if we want to stop scheduling workflows when they keep failing, we could configure the CronWorkflow with:
+
+```
+RunStrategy:
+	maxFailures: 2
+```
+
+This config will stop scheduling workflows if fails twice.
+
+
+## Total vs consecutive
+
+One aspect that needs to be discussed is whether these configurations apply to the entire life of a cron Workflow or just in consecutive schedulings. For example, if we configure a workflow to stop scheduling after 2 failures, I think it makes sense to have this applied when it fails twice consecutively. Otherwise, we can have 2 outages in different periods which will suspend the workflow. On the other hand, when configuring a workflow to run twice with success, it would make more sense to have it execute with success regardless of whether it is a consecutive success or not. If we have an outage after the first workflow succeeds, which translates into failed workflows, it should need to execute with success only once. So I think it would make sense to have:
+
+- maxFailures - maximum number of **consecutive failures** before stopping the scheduling of a workflow
+
+- maxSuccess - maximum number of workflows with sucess. 
+
+## How to store state
+
+Since we need to control how many child workflows had success/failure we must store state. With this some questions arise:
+
+- Should we just store it through the lifetime of the controller or should we store it to a database? 
+
+	- Probably only makes sense if we can backup the state somewhere (like a BD). However, I dont have enough knowledge about workflow's architecture to tell how good of an idea this is. 
+- If a cronworkflow gets re-applied, does it maintain or reset the number of success/failures?
+
+	- I guess it should reset since a configuration change should be seen as a new start.
+
+
+## How to stop the workflow
+
+Once the configured number of failures or successes is reached, it is necessary to stop the workflow scheduling. 
+I believe we have 3 options:
+
+- Delete the workflow: In my opinion, this is the worst option and goes against gitops principles. 
+- Suspend it (set suspend=true): the workflow spec is changed to have the workflow suspended. I may be wrong but this conflicts with gitops as well.
+- Stop scheduling it: The workflow spec is the same. The controller needs to check if the max number of runs was already attained and skip scheduling if it did.
+
+Option 3 seems to be the only possibility. After reaching the max configured executions, the cron workflow would exist forever but never scheduled. Maybe we could add a new status field, like `Inactive` and have something the UI to show it? 
+
+## How to handle suspended workflows
+
+One possible case that comes to mind is a long outage where all workflows are failing. For example, imagine a workflow that needs to download a file from some storage and for some reason that storage is down. Workflows will keep getting scheduled but they are going to fail. If they fail the number of configured `maxFailures`, the workflows gets stoped forever. Once the storage is back up, how can the user enable the workflow again?
+
+- Manually re-create the workflow: could be an issue if the user has multiple cron workflows
+- Instead of stoping the workflow scheduling, introduce a back-off period as suggested by [#7291](https://github.com/argoproj/argo-workflows/issues/7291). Or maybe allow both configurations. 
+
+I believe option 2 would allow the user to select if they want to stop scheduling or not. If they do, when cron workflows are wrongfully halted, they will need to manually start them again. If they don't, Argo will only introduce a back-off period between schedules to avoid rescheduling workflows that are just going to fail. Spec could look something like:
+
+```
+RunStrategy:
+	maxSuccess:
+	maxFailures:
+		value: # this would be optional
+		back-off:
+			enabled: true
+			factor: 2
+```
+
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
cc @sarabala1979 @caelan-io @Kitanotori @wesleyscholl @tico24

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+ maxFailures: 2
+```
+
+This config will stop scheduling workflows if fails twice.
+
+## Total vs consecutive
+
+One aspect that needs to be discussed is whether these configurations apply to the entire life of a cron Workflow or just in consecutive schedules. For example, if we configure a workflow to stop scheduling after 2 failures, I think it makes sense to have this applied when it fails twice consecutively. Otherwise, we can have 2 outages in different periods which will suspend the workflow. On the other hand, when configuring a workflow to run twice with success, it would make more sense to have it execute with success regardless of whether it is a consecutive success or not. If we have an outage after the first workflow succeeds, which translates into failed workflows, it should need to execute with success only once. So I think it would make sense to have:
+
+- maxFailures - maximum number of **consecutive failures** before stopping the scheduling of a workflow
```
caelan-io, line 43: I would propose adding another field to toggle between consecutive or total failures. It could be a boolean with default as `false`. 

Maybe: `consecFailures` or `consecFail`

This would make the default logic align for both `maxSuccess` and `maxFailures` which is simpler logic for a new user to understand. If they want consecutive failures counted, then they can set `consecFailures` to `true`.

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+Currently, CronWorkflows are a great resource if we want to run recurring tasks to infinity. However, it is missing the ability to customize it, for example define how many times a workflow should run or how to handle multiple failures. I believe argo workflows would benefit of having more configuration options for cron workflows, to allow to change its behavior based on the result of its child’s success or failures. Below I present my thoughts on how we could improve them, but also some questions and concerns on how to properly do it.
+
+## Proposal
+
+This proposal discusses the viability of adding 2 more fields into the cron workflow configuration:
+
+```
+RunStrategy:
+ maxSuccess:
+ maxFailures:
```
caelan-io, line 14: maybe `maxFail` for simplicity and align with  "success vs. fail"

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+`maxSuccess` - defines how many child workflows must have success before suspending the workflow schedule
+
+`maxFailures` - defines how many child workflows must fail before suspending the workflow scheduling. This may contain `Failed` workflows, `Errored` workflows or spec errors.
```
caelan-io, line 19: very important to document this clearly in the docs when the feature is released.

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+## How to store state
+
+Since we need to control how many child workflows had success/failure we must store state. With this some questions arise:
+
+- Should we just store it through the lifetime of the controller or should we store it to a database?
+
+    - Probably only makes sense if we can backup the state somewhere (like a BD). However, I don't have enough knowledge about workflow's architecture to tell how good of an idea this is.
+- If a CronWorkflow gets re-applied, does it maintain or reset the number of success/failures?
+
+    - I guess it should reset since a configuration change should be seen as a new start.
```
caelan-io, line 56: I agree with resetting state upon re-application of the CronWorkflow

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+
+## How to handle suspended workflows
+
+One possible case that comes to mind is a long outage where all workflows are failing. For example, imagine a workflow that needs to download a file from some storage and for some reason that storage is down. Workflows will keep getting scheduled but they are going to fail. If they fail the number of configured `maxFailures`, the workflows gets stopped forever. Once the storage is back up, how can the user enable the workflow again?
+
+- Manually re-create the workflow: could be an issue if the user has multiple cron workflows
+- Instead of stopping the workflow scheduling, introduce a back-off period as suggested by [#7291](https://github.com/argoproj/argo-workflows/issues/7291). Or maybe allow both configurations.
+
+I believe option 2 would allow the user to select if they want to stop scheduling or not. If they do, when cron workflows are wrongfully halted, they will need to manually start them again. If they don't, Argo will only introduce a back-off period between schedules to avoid rescheduling workflows that are just going to fail. Spec could look something like:
```
caelan-io, line 76: I agree with this, but I would say that we should probably ship this feature in a separate PR than the above requirements you outlined. It feels like this would add additional complexity.

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
+# Proposal for Cron Workflows improvements
+
+## Introduction
+
+Currently, CronWorkflows are a great resource if we want to run recurring tasks to infinity. However, it is missing the ability to customize it, for example define how many times a workflow should run or how to handle multiple failures. I believe argo workflows would benefit of having more configuration options for cron workflows, to allow to change its behavior based on the result of its child’s success or failures. Below I present my thoughts on how we could improve them, but also some questions and concerns on how to properly do it.
```
caelan-io, line 5: I would also note here how a comparable tool, Airflow, includes this functionality so it is a strong precedent to enable it for Argo Workflows as well: https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/scheduler.html#triggering-dag-with-future-date

caelan-io (changes requested):
Thanks @eduardodbr ! Looks real great. I left some suggestions.

docs/proposals/cron-wf-improvement-proposal.md
eduardodbr: Maybe we could have the run strategy to be configurable by success or failure. For example:

```yaml
RunStrategy:
  onSuccess:
    maxSuccess: 3 
  onFail:
    maxFail:  3
    consecFail: false 
```

this would allow us to further upgrade the configurable behavior in the future without breaking the schema. For example,  if we decide to ship the back-off in another PR, you could just upgrade it to:

```yaml
RunStrategy:
  onSuccess:
    maxSucess: 3 
  onFail:
    maxFail:  3
    consecFail: true
    back-off:
      enabled: true
      factor: 2
```

However, I do agree with you, this may add a lot of complexity, and its best to keep it simple for now.

docs/proposals/cron-wf-improvement-proposal.md
eduardodbr: I didn't mention that because this proposal was aiming to be something wider than just allowing to run a workflow once (in the future or not). However, I agree that it is one of the most important features that Argo workflows is lacking. I believe a lot of people would need it more frequently than configuring the behavior of recurrent success/fail. This proposal aimed to allow both use cases. However, if we come to the conclusion that it may be to complex to implement all this behavior, I would suggest at least implementing the trigger once feature.

docs/proposals/cron-wf-improvement-proposal.md
caelan-io: I like this suggestion of yours:

> Maybe we could have the run strategy to be configurable by success or failure. this would allow us to further upgrade the configurable behavior in the future without breaking the schema. For example, if we decide to ship the back-off in another PR, you could just upgrade it

I'm up for this if you think it'll be simple enough to implement on the first attempt.

caelan-io (approved):

caelan-io:
@terrytangyuan - what are your thoughts on moving forward with this?

terrytangyuan:
This is on my list of items to review.

docs/proposals/cron-wf-improvement-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+I believe option 2 would allow the user to select if they want to stop scheduling or not. If they do, when cron workflows are wrongfully halted, they will need to manually start them again. If they don't, Argo will only introduce a back-off period between schedules to avoid rescheduling workflows that are just going to fail. Spec could look something like:
+
+```yaml
+RunStrategy:
+ maxSuccess:
+ maxFailures:
+  value: # this would be optional
+  back-off:
+   enabled: true
+   factor: 2
```
terrytangyuan, line 85: What about something like the following so it's more extensible to other phases?
```
runStrategy:
  condition: "failure >= 3"
  backOff:
    factor: ...
```

eduardodbr:
@caelan-io  @terrytangyuan  I have some time to come back to this and after your suggestions I suggest the following first approach:

 - New Spec field `stopStrategy` with a `condition` that is used to stop scheduling workflows. Controller stops scheduling workflows if the stopping condition is achieved:
 ```
 stopStrategy:
    condition: "succeeded >= 1"
 ```
- New status fields:
     -  Succeeded - counts the number of workflows that completed in phase `Succeeded`
     - Failed -  counts the number of workflows that failed to submit or completed in phase  `Error` or `Failed`  
     - Completed - whether the stopping condition was achieved or not (is it worth it to have it as an annotation?) 
- Use the `completed` status field instead of changing the cronworkflow with “suspend=true” to avoid changing the cronworkflow spec, which conflicts with gitops users
- Keywords allowed for stopping conditions are `failed` and `succeeded`:
      - Failed - uses the status field `failed` 
      - Succeeded - uses the status field `succeeded`
 ```
stopStrategy:
  condition: "failed >= 3"
```
 ```
stopStrategy:
  condition: "succeeded >= 1"
```
- Updating the stopping condition after the workflow is completed, will continue the normal cron workflow execution.


Just wondering if instead of naming it `stopStrategy` we could use something more generic. However, I feel that @terrytangyuan's suggestion `runStrategy` can be easily missed if the user just wants to schedule the workflow for a X amount of times. 

If it is ok with you I can open a PR with the above changes.

terrytangyuan:
I don't have a strong preference regarding the naming.

terrytangyuan (approved):

terrytangyuan:
Sure

caelan-io:
Thank you, Terry!

PR #11368, opened 2023-07-15
fix: prevent stdout from disappearing in script templates. Fixes #11330

boiledfroginthewell:
Fixes #11330

### Motivation

Outputs of the executed command are asynchronously read and copied to `cmd.Stdout`. `cmd.Stdout` must be closed after copies are done otherwise command outputs are gone. 

As described below, however, copies can never be end because sub-processes can keep running and can write to stdout indefinitely.

https://github.com/argoproj/argo-workflows/blob/6166464aa6961bcb375705753ed0b58707d68222/workflow/executor/os-specific/signal_darwin.go#L27-L38

### Modifications

`Wait()` is added to wait the copy.
`WaitDelay` is also added. It will close the pipes (stdout/stderr) of the executed command to stop sub-processes writing any more texts. 

### Verification

Tests were added.
I also ran the following command for more than an hour.

```sh
$ docker run \
    --rm -it \
    --cpus 0.5 \
    -e ARGO_CONTAINER_NAME=main \
    -e ARGO_INCLUDE_SCRIPT_OUTPUT=true \
    -v $PWD/dist/argoexec:/var/run/argo/argoexec \
    busybox:1.35 \
    sh
$ echo '{}' > /var/run/argo/template
$ while :; do
    rm -f /var/run/argo/ctr/main/*
    < /dev/null /var/run/argo/argoexec emissary --loglevel info --log-format text -- echo hello
    ls -l /var/run/argo/ctr/main
    [ ! -s /var/run/argo/ctr/main/stdout ] && break
    [ ! -s /var/run/argo/ctr/main/combined ] && break
done
```

PR #11576, opened 2023-08-15
fix: fail test on pr #11368

sonbui00:
Fixes #11368

### Motivation

Run test fail

### Modifications

Replace echo -n with echo ..\c"

### Verification

make test

### DIFF
--- a/cmd/argoexec/commands/emissary_test.go
+++ b/cmd/argoexec/commands/emissary_test.go
@@ -49,7 +49,7 @@ func TestEmissary(t *testing.T) {
 	})
 	t.Run("Sub-process", func(t *testing.T) {
 		_ = os.Remove(varRunArgo + "/ctr/main/stdout")
-		err := run("(sleep 60; echo 'should not wait for sub-process')& echo -n hello")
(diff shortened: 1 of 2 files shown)


cmd/argoexec/commands/emissary_test.go
```diff
@@ -49,7 +49,7 @@ func TestEmissary(t *testing.T) {
 	})
 	t.Run("Sub-process", func(t *testing.T) {
 		_ = os.Remove(varRunArgo + "/ctr/main/stdout")
-		err := run("(sleep 60; echo 'should not wait for sub-process')& echo -n hello")
+		err := run(`(sleep 60; echo 'should not wait for sub-process')& echo "hello\c"`)
```
agilgur5, line 52: For other readers, note that `\c` is equivalent to the `-n` flag. They [both mean](https://www.gnu.org/software/bash/manual/bash.html#index-echo) to not output a trailing newline

agilgur5 (approved):
Nice [catch](https://github.com/argoproj/argo-workflows/pull/11368#issuecomment-1678283651)!
LGTM

terrytangyuan (approved):

PR #11577, opened 2023-08-15
feat(cli): Add a flag status to delete cmd like list cmd of argo cli

spencercjh:
### Motivation

For https://github.com/argoproj/argo-workflows/issues/11555. @tooptoop4 wants to filter workflows more accurately and flexibly by status.

There are two supported flags associated with it: 

- `--field-selector`: `metadata.name=foo`, `metadata.namespace=bar`
- `--selector`: `workflows.argoproj.io/phase=Succeeded`

It is not feasible to use the `--field-selector` to filter workflows by status because of the [well-known reasons](https://github.com/kubernetes/kubernetes/issues/53459).

Using `-l(--selector)` would satisfy the need, but the labels are too long with a common prefix `workflows.argoproj.io/`. It is not convenient to use.

Luckily, `argo list` has a directly related flag `--status` to filter workflows by status like `argo list --status Running`. It is more convenient to use. I think it could be taken a step further by exempting users from using the shell combination like `argo delete $(argo list --status Succeeded | awk '{print $1}')`

### Modifications

1. Add a flag `--status` to `argo delete` to filter workflows by status. It's an existing field of struct `listFlags` and can be used directly.

### Verification

There was no test code for `argo delete` before. I run it locally and it works as expected.

image

### DIFF
--- a/cmd/argo/commands/delete.go
+++ b/cmd/argo/commands/delete.go
@@ -25,7 +25,7 @@ func NewDeleteCommand() *cobra.Command {
 		force         bool
 	)
 	command := &cobra.Command{
-		Use:   "delete [--dry-run] [WORKFLOW...|[--all] [--older] [--completed] [--resubmitted] [--prefix PREFIX] [--selector SELECTOR] [--force] ]",
+		Use:   "delete [--dry-run] [WORKFLOW...|[--all] [--older] [--completed] [--resubmitted] [--prefix PREFIX] [--selector SELECTOR] [--force] [--status STATUS] ]",
 		Short: "delete workflows",
 		Example: `# Delete a workflow:
 
@@ -36,7 +36,10 @@ func NewDeleteCommand() *cobra.Command {
   argo delete @latest
 `,
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) == 0 && !(all || allNamespaces || flags.completed || flags.resubmitted || flags.prefix != "" || flags.labels != "" || flags.fields != "" || flags.finishedAfter != "") {
+			hasFilterFlag := all || allNamespaces || flags.completed || flags.resubmitted || flags.prefix != "" ||
+				flags.labels != "" || flags.fields != "" || flags.finishedAfter != "" || len(flags.status) > 0
+
+			if len(args) == 0 && !hasFilterFlag {
 				cmd.HelpFunc()(cmd, args)
 				os.Exit(1)
 			}
@@ -52,7 +55,7 @@ func NewDeleteCommand() *cobra.Command {
 					ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: flags.namespace},
 				})
 			}
-			if all || flags.completed || flags.resubmitted || flags.prefix != "" || flags.labels != "" || flags.fields != "" || flags.finishedAfter != "" {
+			if hasFilterFlag {
 				listed, err := listWorkflows(ctx, serviceClient, flags)
 				errors.CheckError(err)
 				workflows = append(workflows, listed...)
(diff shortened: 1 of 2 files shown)


spencercjh:
https://github.com/argoproj/argo-workflows/pull/11557 does something similar, but I don't think flag pending makes sense, there shouldn't be such a tiny, fine-grained flag. 

Also thanks for his/her PR. 🙏

cmd/argo/commands/delete.go
```diff
@@ -52,7 +55,7 @@ func NewDeleteCommand() *cobra.Command {
 					ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: flags.namespace},
 				})
 			}
-			if all || flags.completed || flags.resubmitted || flags.prefix != "" || flags.labels != "" || flags.fields != "" || flags.finishedAfter != "" {
+			if hasFilterFlag {
```
agilgur5, line 58: `--all-namespaces` was previously not in this conditional.
from a quick glance, not sure if that actually makes a difference though

cmd/argo/commands/delete.go
```diff
@@ -52,7 +55,7 @@ func NewDeleteCommand() *cobra.Command {
...
 				listed, err := listWorkflows(ctx, serviceClient, flags)
```
agilgur5, line 59: nice that this works without needing any changes 🙂

agilgur5:
> It is not feasible to use the `--field-selector` to filter workflows by status because of the [well-known reasons](https://github.com/kubernetes/kubernetes/issues/53459).
> 
> Using `-l(--selector)` would satisfy the need, but the labels are too long with a common prefix `workflows.argoproj.io/`. It is not convenient to use.

Agreed with this rationale. After you brought up the inadequacy of `--field-selector` in https://github.com/argoproj/argo-workflows/issues/11555#issuecomment-1676990143, I thought label selectors could be a workaround as well, but I would agree that it is quite long to type out.

Moreover, `argo list` already supports `--status`, so this is just exposing that existing flag to `argo delete`.

agilgur5:
> #11557 does something similar, but I don't think flag pending makes sense, there shouldn't be such a tiny, fine-grained flag.

I would normally agree with this, but given that there are already `--running` and `--completed` flags, it's not altogether new either. If those didn't exist already, I would definitely prefer to not increase the surface area.

In any case, I think that discussion should be had in that PR. This PR does not technically preclude that PR

spencercjh:
> > It is not feasible to use the `--field-selector` to filter workflows by status because of the [well-known reasons](https://github.com/kubernetes/kubernetes/issues/53459).
> > Using `-l(--selector)` would satisfy the need, but the labels are too long with a common prefix `workflows.argoproj.io/`. It is not convenient to use.
> 
> Agreed with this rationale. After you brought up the inadequacy of `--field-selector` in [#11555 (comment)](https://github.com/argoproj/argo-workflows/issues/11555#issuecomment-1676990143), I thought label selectors could be a workaround as well, but I would agree that it is quite long to type out.
> 
> Moreover, `argo list` already supports `--status`, so this is just exposing that existing flag to `argo delete`.

Yes. My implementation is reuse listFlags.

agilgur5 (commented):
one tiny comment I need to look at more closely, otherwise LGTM.
Thanks for your contribution!

cmd/argo/commands/delete.go
agilgur5: When `--all-namespaces` is _not_ set, the [client's namespace is set as a flag](https://github.com/argoproj/argo-workflows/blob/2229fbac5ed1c61beda4e313f4b8a15655d919a3/cmd/argo/commands/delete.go#L51) (a few lines above).

However, if it is set, I _believe_ a `list` would be necessary anyway, so I _think_, functionally, these are equivalent

agilgur5 (approved):

spencercjh:
Thank's for your review, your analysis is very detailed and careful. 👍 💯

terrytangyuan (approved):

PR #11475, opened 2023-07-29
[Snyk] Upgrade monaco-editor from 0.20.0 to 0.40.0

sarabala1979:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade monaco-editor from 0.20.0 to 0.40.0.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **493 versions** ahead of your current version.
- The recommended version was released **21 days ago**, on 2023-07-07.

<hr/>

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open upgrade PRs.*

For more information:  

🧐 [View latest project report](https://app.snyk.io/org/sarabala1979/project/05009f53-cf59-41dc-9a3c-4db00251e16a?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr)

🛠 [Adjust upgrade PR settings](https://app.snyk.io/org/sarabala1979/project/05009f53-cf59-41dc-9a3c-4db00251e16a/settings/integration?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr)

🔕 [Ignore this dependency or unsubscribe from future upgrade PRs](https://app.snyk.io/org/sarabala1979/project/05009f53-cf59-41dc-9a3c-4db00251e16a/settings/integration?pkg&#x3D;monaco-editor&amp;utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr#auto-dep-upgrades)

PR #11593, opened 2023-08-16
chore(deps): upgrade `monaco-editor` to 0.30

agilgur5:
Partially replaces #11475, but there is still more work to be done as written below

### Motivation

Was [asked](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.qogh82i9yij0) to help upgrade deps, particularly ones that may have vulns since Snyk opened #11475
  - Though it doesn't actually mention any CVEs, but good to upgrade anyway (and I got a good part of the way in just investigating the issue)

### Modifications

- upgrade `monaco-editor` to 0.30
  - 0.30 requires `monaco-editor-webpack-plugin` 6, per the [version matrix](https://github.com/microsoft/monaco-editor/tree/v0.41.0/webpack-plugin#version-matrix)
  - [0.29.0](https://github.com/microsoft/monaco-editor/blob/v0.41.0/CHANGELOG.md#0290-08102021) had a breaking change to `renderIndentGuides`; converted to the new option
    - `true` might not be necessary as it's the default, but `guides` is empty by default, so not 100% sure -- left it in for now to ensure the behavior remains the same after the upgrade

- 0.30 is the latest version I could upgrade to without requiring significant build changes
  - 0.31+ uses newer JS syntax per the [breaking changes](https://github.com/microsoft/monaco-editor/blob/v0.41.0/CHANGELOG.md#0310-10122021)
    - Per https://stackoverflow.com/a/70622999/3431180, this requires an updated parser, otherwise the Webpack build was getting similar errors as in the SO question
      - tried to remove `monaco-editor` from the `babel-loader` `exclude` to remedy this, but then the build would crash due to out-of-memory (OOM) heap allocation errors
        - as a result, only upgraded to 0.30 for now, can upgrade to latest once the build system itself is upgraded

### Verification

Build passes, install succeeds


ui/package.json
```diff
@@ -27,7 +27,7 @@
         "json-merge-patch": "^0.2.3",
         "moment": "^2.29.4",
         "moment-timezone": "^0.5.43",
-        "monaco-editor": "0.20.0",
+        "monaco-editor": "^0.30.0",
```
terrytangyuan, line 30: What about `react-monaco-editor`? https://github.com/argoproj/argo-workflows/pull/11598?

agilgur5: Oh I hadn't seen that one, you had only mentioned #11475 in the meeting. Upgrading `monaco-editor` didn't seem to require updating `react-monaco-editor` in tandem, but I can take a look at the error there too now that I know of it

agilgur5: Oh that Snyk PR was literally opened a few hours ago just before you commented here 😅 
Was that a manual run?

agilgur5: Hmm build seems unrelated for that one. And even the version of `react-monaco-editor` that we're on, 0.50.1, actually requires a newer version of `monaco-editor`, [^0.34.0](https://github.com/react-monaco-editor/react-monaco-editor/blob/0f119c8e805d489ed35616b90af595dad32a7709/package.json#L53). 
That Snyk PR upgrades to 0.54.0, which requires `monaco-editor` [^0.39.0](https://github.com/react-monaco-editor/react-monaco-editor/blob/6ab247bc9e17271b43e27c729902def77ef1009c/package.json#L53).

So er, the current set-up already uses incompatible peerDeps, but this one brings it at least closer 😅

agilgur5: We might want to make CI UI install fail on invalid peerDeps. Though right now all the React versioning is incompatible too, but still working (as an unsupported peerDep doesn't _necessarily_ mean it's incompatible, but means it probably hasn't been tested at least)

agilgur5: `react-monaco-editor` seems to be mostly pass through, as the `options` prop was directly impacted by the `monaco-editor` version

terrytangyuan (approved):

terrytangyuan:
Did you verify that the editor still works?

agilgur5:
Oof that is a good question, I did not check that specifically. Most of my UI PRs just fix the entire build, so just checking if the UI works is a decent verification, but this is actually a very specific component of the UI.

My bad, let me double check that now.

agilgur5:
Took a bit, but got it all working. Did need a small fix, see #11655

ui/package.json
agilgur5: Investigated `react-monaco-editor` in https://github.com/argoproj/argo-workflows/pull/11598#issuecomment-1696429512

agilgur5:
Updated Webpack in #11628 which allowed this dep to be further updated.
Updated `monaco-editor` to latest 0.41.0 in #11710

PR #11599, opened 2023-08-17
build: run openapi-generator-cli in docker with current user.

luyang93:
### Motivation
Due to docker running under root, it is necessary to modify the owner of the generated file.

### Modifications
Add --user with docker run
I've read [link](https://vsupalov.com/docker-shared-permissions/#set-the-docker-user-when-running-your-container)
bash and openapi-generator-cli works normally


luyang93:
Can we remove chown if run docker with --user?

terrytangyuan (approved):

Issue #11156, opened 2023-05-31
`Pod was active on the node longer than the specified deadline` issue with v3.4.2+


sakshimalani:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

We have upgraded argo-workflow to v3.4.8 and workflow controller pod is running fine. However, workflow pods are giving below message-
**level=info msg="Pod failed: Pod was active on the node longer than the specified deadline"**
This issue was observed from v3.4.2 onwards.
This issue is not allowing us to upgrade to latest version and older versions are having security violations.

There is already an issue created regarding this https://github.com/argoproj/argo-workflows/issues/9934 but there is no recent updates.

### Version

v3.4.8

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
We cannot share those information.
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep kaas-external-secrets-canary-1685518200

time="2023-05-31T07:33:27.121Z" level=info msg="node changed" namespace=external-secrets new.message= new.phase=Running new.progress=0/1 nodeID=kaas-external-secrets-canary-1685518200-1847453086 old.message= old.phase=Running old.progress=0/1 workflow=kaas-external-secrets-canary-1685518200
time="2023-05-31T07:33:27.122Z" level=info msg="SG Outbound nodes of kaas-external-secrets-canary-1685518200-3820792552 are [kaas-external-secrets-canary-1685518200-1729983040]" namespace=external-secrets workflow=kaas-external-secrets-canary-1685518200
time="2023-05-31T07:33:27.123Z" level=info msg="Workflow step group node kaas-external-secrets-canary-1685518200-3290429874 not yet completed" namespace=external-secrets workflow=kaas-external-secrets-canary-1685518200
time="2023-05-31T07:33:27.123Z" level=info msg="Workflow step group node kaas-external-secrets-canary-1685518200-413055525 not yet completed" namespace=external-secrets workflow=kaas-external-secrets-canary-1685518200
[...]

Issue #9934, opened 2022-10-31
Pod failed with error: Pod was active on the node longer than the specified deadline -remain in status Running 


shiraOvadia:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I set a timeout of 10 seconds in a template.
activeDeadlineSeconds: 10
After 10 seconds the pod received an error: the pod was active on the node for more than the specified deadline
After a few minutes the pod is deleted in kubernetes, but its remain in status Pending or Running.
The flow is blocked and no continue to the next template.
I expected that the pod get failed status

In previous version the same workflow work ok.

### Version

V3.4.2

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: hello-world-  # Name of this Workflow
spec:
  entrypoint: engine
  podGC:
    strategy: OnPodSuccess
  templates:
  - name: whalesay            # Defining the "whalesay" template
    activeDeadlineSeconds: 10
    container:
      image: docker/whalesay
      command: ['sh','-c']
      args: ["cowsay hello world && sleep 600"]   # This template runs "cowsay" in the "whalesay" image with arguments "hello world"
      resources:
        requests:
          memory: "3Gi"
          cpu: "2000m"
        limits:
          memory: "3Gi"
          cpu: "2000m"
  - name: engine
    parallelism: 7000
    steps:
      - - name: whalesay
          template: whalesay
          withSequence:
            count:  1
```

### Logs from the workflow controller

kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
[workflow.log](https://github.com/argoproj/argo-workflows/files/9904813/workflow.log)

### Logs from in your workflow's wait container

kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
[wait.log](https://github.com/argoproj/argo-workflows/files/9904815/wait.log)

shiraOvadia:
This is a fatal error. It prevents us from upgrading argo version to latest.
Can someone please help here?

sarabala1979:
[...]

PR #11603, opened 2023-08-17
fix(workflow): match discovery burst and qps for `kubectl` with upstream kubectl binary

vadasambar:
May fix or alleviate issues like https://github.com/argoproj/argo-workflows/issues/11156 and https://github.com/argoproj/argo-workflows/issues/9934

### Motivation

We migrated to using `kubectl` as a package starting v3.4.7:
* v3.4.7: https://github.com/argoproj/argo-workflows/blob/v3.4.7/go.mod#L71
* v3.4.8: https://github.com/argoproj/argo-workflows/blob/v3.4.8/go.mod#L71
* master: https://github.com/argoproj/argo-workflows/blob/master/go.mod#L71

Before v3.4.7 we used `kubectl` binary ([ref1](https://github.com/argoproj/argo-workflows/blob/v3.4.6/Dockerfile#L82-L83), [ref2](https://github.com/argoproj/argo-workflows/blob/988706dd131cf98808f09fb7cc03780e2af94c73/workflow/executor/resource.go#L39))

`kubectl` binary uses uses higher values for discovery burst and QPS [since v1.24](https://github.com/kubernetes/kubernetes/pull/107131 ). `kubectl` passes the new QPS and burst through [`NewDefaultKubectlCommand`](https://github.com/kubernetes/kubernetes/blob/a5d2d6fec3a1a9538255d796eaf58e0777940135/cmd/kubectl/kubectl.go#L29) while in argo workflows we use [`NewKubectlCommand`](https://github.com/argoproj/argo-workflows/blob/1adc6c208717d0f2a460cefe452f651baa1dd6c8/workflow/executor/resource.go#L374) instead [which uses default burst of `100`](https://github.com/kubernetes/cli-runtime/blob/4fdf49ae46a0caa7fafdfe97825c6129d5153f06/pkg/genericclioptions/config_flags.go#L442) ([we are using `v.0.24.3` version of `kubectl` pkg](https://github.com/argoproj/argo-workflows/blob/7b80ce19e8afe6690aed5c2f3d6c123c812e468b/go.mod#L74) at the time of writing this) 

If there are 100 or more CRDs in the cluster, discovery can take time. This leaves less time for pod to run because of which pod can exceed `activeDeadlineSeconds` (check https://github.com/argoproj/argo-workflows/issues/11156 and https://github.com/argoproj/argo-workflows/issues/9934). Related: https://github.com/kubernetes/kubectl/issues/1126

### Modifications

This PR matches the the new discovery burst and QPS set in the upstream. 
Related:
* https://github.com/kubernetes/kubernetes/pull/120024 is merged

Note that in 1.26 version (v0.26.x for kubectl package), [default burst across all client-go clients was bumped up to 300](https://github.com/kubernetes/kubernetes/pull/109141) (no increase in the default QPS though) but we can't use this since we are on v0.24.3. Upgrading the `kubectl` package is an option but I am not sure about the consequences of doing that. We might want to re-visit this PR again in the future and remove lines which are already set in the upstream.

### DIFF
--- a/workflow/executor/resource.go
+++ b/workflow/executor/resource.go
@@ -372,9 +372,14 @@ func runKubectl(args ...string) ([]byte, error) {
 	}()
 	var buf bytes.Buffer
 	if err := kubectlcmd.NewKubectlCommand(kubectlcmd.KubectlOptions{
-		Arguments:   args,
-		ConfigFlags: genericclioptions.NewConfigFlags(true),
-		IOStreams:   genericclioptions.IOStreams{Out: &buf, ErrOut: os.Stderr},
+		Arguments: args,
+		// TODO(vadasambar): use `DefaultConfigFlags` variable from upstream
+		// as value for `ConfigFlags` once https://github.com/kubernetes/kubernetes/pull/120024 is merged
+		ConfigFlags: genericclioptions.NewConfigFlags(true).
+			WithDeprecatedPasswordFlag().
+			WithDiscoveryBurst(300).
+			WithDiscoveryQPS(50.0),
+		IOStreams: genericclioptions.IOStreams{Out: &buf, ErrOut: os.Stderr},
 	}).Execute(); err != nil {
 		return nil, err
 	}


tooptoop4:
300/50 seems too high, I raised https://github.com/argoproj/argo-helm/issues/2195 but 100/50 is only recommended from k8s 1.27, under k8s 1.27 its 20/10

vadasambar:
> 300/50 seems too high, I raised [argoproj/argo-helm#2195](https://github.com/argoproj/argo-helm/issues/2195) but 100/50 is only recommended from k8s 1.27, under k8s 1.27 its 20/10

I think you are talking about rest client burst/qps. This PR is about discovery client burst/qps. 300/50 is the default upstream value since v1.24 version of Kubernetes. All of the details are in the description of the PR.

vadasambar:
`CI / E2E Tests (test-executor, minimal)` seems flaky. Similar test is failing in other open PRs as well:
* https://github.com/argoproj/argo-workflows/pull/11628
* https://github.com/argoproj/argo-workflows/pull/11641
* https://github.com/argoproj/argo-workflows/pull/11638
* https://github.com/argoproj/argo-workflows/pull/11644

Makes me think the test failure is not related to this PR.

vadasambar:
I think the PR is ready for review. I don't think the test failure is related to the changes in this PR. Happy to take another look if the reviewers think the failure is related to changes in this PR.

workflow/executor/resource.go
```diff
@@ -372,9 +372,13 @@ func runKubectl(args ...string) ([]byte, error) {
 	}()
 	var buf bytes.Buffer
 	if err := kubectlcmd.NewKubectlCommand(kubectlcmd.KubectlOptions{
-		Arguments:   args,
-		ConfigFlags: genericclioptions.NewConfigFlags(true),
-		IOStreams:   genericclioptions.IOStreams{Out: &buf, ErrOut: os.Stderr},
+		Arguments: args,
+		// TODO(vadasambar): re-visit this once https://github.com/kubernetes/kubernetes/pull/120024 is merged
+		ConfigFlags: genericclioptions.NewConfigFlags(true).
+			WithDeprecatedPasswordFlag().
```
terrytangyuan, line 378: This is not needed?

terrytangyuan (commented):
Should this be chore instead of fix?

workflow/executor/resource.go
vadasambar: Thank you for the comment. 

I have tried matching the upstream default config: https://github.com/kubernetes/kubernetes/pull/120024/files#diff-7e82ada63a2e6664b803ac3ccdb84759ac97338ce807df63be2f81e3c9e0ba9cR95 

Once https://github.com/kubernetes/kubernetes/pull/120024 is merged we are going to end up using it anyway. I don't have a strong opinion on this. Happy to remove it for now if you want. Let me know.

workflow/executor/resource.go
```diff
@@ -372,9 +372,13 @@ func runKubectl(args ...string) ([]byte, error) {
 	}()
 	var buf bytes.Buffer
 	if err := kubectlcmd.NewKubectlCommand(kubectlcmd.KubectlOptions{
-		Arguments:   args,
-		ConfigFlags: genericclioptions.NewConfigFlags(true),
-		IOStreams:   genericclioptions.IOStreams{Out: &buf, ErrOut: os.Stderr},
+		Arguments: args,
+		// TODO(vadasambar): re-visit this once https://github.com/kubernetes/kubernetes/pull/120024 is merged
```
terrytangyuan, line 376: Can you add add instructions on how to update this part once that PR is merged? (in case in you are away we can also remove the TODO in the future)

vadasambar: Thank you for the comment. I have re-worded the TODO explaining what needs to be done. Let me know if you think we should reword it (or add more).

terrytangyuan (approved):

agilgur5:
> May fix or alleviate issues like #11156 and #9934

> If there are 100 or more CRDs in the cluster, discovery can take time. This leaves less time for pod to run because of which pod can exceed `activeDeadlineSeconds` (check #11156 and #9934)

This specific change would only help those issues if they were using many `resource` templates with many CRDs. They don't give reproducible Workflows unfortunately, but those are typically a small subset of Workflows.

PR #11605, opened 2023-08-18
feat: upgrade expr to v1.14 for richer language definition

jessesuen:
### Motivation

Updates antonmedv/expr from v1.13 to v1.14 which has a [much richer language definition](https://expr.medv.io/docs/Language-Definition) and includes useful helpers such as `hasPrefix`, `max`

I think this will really help with simplifying workflows. I also think it will make the WorkflowEventBinding much more useful and reduce the need for argo-events webhook eventsource.

The other problem that this will solve is that our docs constantly refer to the expr project to see how to write expressions. But if we continue to use the old version, users will hit unexpected errors. e.g.:

```
Non-transient error: failed to evaluate workflow template expression: unable to evaluate expression 'hasPrefix(payload.Key, \"foo\") == true': reflect: call of reflect.Value.Call on zero Value (1:1)\n | hasPrefix(payload.Key, \"foo\") == true\n | ^
```

In fact, this is how I discovered the new version in the first place (I was confused on why certain functions were unavailable).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/alibabacloud-go/tea v1.2.1
 	github.com/aliyun/aliyun-oss-go-sdk v2.2.8+incompatible
 	github.com/aliyun/credentials-go v1.3.1
-	github.com/antonmedv/expr v1.13.0
+	github.com/antonmedv/expr v1.14.0
 	github.com/argoproj/argo-events v1.7.3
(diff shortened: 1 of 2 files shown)


jessesuen:
v1.14:
image

vs v1.13:
image

terrytangyuan (approved):

agilgur5:
Shortly after this PR it looks like the creator/maintainer filed #11608 about these new functions

PR #11613, opened 2023-08-18
fix: upgrade base image for security and build support arm64 #10435

### LINKED ISSUE #10435: `argosay` uses `amd64` architecture instead of `arm64` on `arm64` machine
# Summary

[argosay](https://hub.docker.com/r/argoproj/argosay/tags) should also be available for arm64, not only for amd64.

# Use Cases

Many people use arm64 machines.

---

sonbui00:
Fixes #10435

### Motivation

- Fix security
- Run examples on MacOS

### Modifications

- Upgrade base image
- Add/update build command for multiple arch

### Verification

https://hub.docker.com/repository/docker/sonbui00/argosay

### Note
 Run `DOCKER_PUSH=true make argosayv1` to build and push v1
Then  run `DOCKER_PUSH=true make argosay` to build and push v2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ E2E_PARALLEL          ?= 20
 E2E_SUITE_TIMEOUT     ?= 15m
 
 VERSION               := latest
-DOCKER_PUSH           := false
+DOCKER_PUSH           ?= false
 
 # VERSION is the version to be used for files in manifests and should always be latest unless we are releasing
 # we assume HEAD means you are on a tag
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):

sonbui00:
@terrytangyuan @isubasinghe @agilgur5 Can you help me build and push the docker image follow my note on the description?

agilgur5:
See user report in https://github.com/argoproj/argo-workflows/issues/10435#issuecomment-1872522148

Apparently this image is not pushed as part of CI?
Terry, Bala, and Alex are the only ones who might have access to the Argo DockerHub account; it looks like Alex was the last to manually push this image.

Could we instead add a CI process to push this on `main` and release branches, like the other images?

agilgur5:
I'm wondering if we should just remove and replace `argosay` entirely. It seems like it was primarily used to add a [small `sleep` for the PNS executor](https://github.com/argoproj/argo-workflows/blob/a68ea0feabc87c09d5e13d12e6f0d1a61adc5b16/test/e2e/images/argosay/v2/argosay#L21), which is no longer supported (`argosay` v1 was added in #2917 and v2 in #3001, for reference)

PR #4653, opened 2020-12-05
docs: Add more information about terminate and stop process

knabben:
Checklist:

* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).

Adding more information on Cobra short helper and cli docs.

Ref https://github.com/argoproj/argo/issues/4454

Issue #4454, opened 2020-11-03
Docs doesn't  tell user differences between  workflow Stop and Terminate behaviors


bklaubos:
## Summary 
I expect  CLI docs to tell end user differences between `argo terminate` and `argo stop`  intentions and its expected behavior. I found none.

## Diagnostics

What Kubernetes provider are you using? 
AWS
What version of Argo Workflows are you running? 
2.16

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

simster7:
Feel free to submit a PR

bklaubos:
I logged this bug because as end user, I don't know the difference between stop vs terminate workflow.  Are they the same?. If not, what are the expected behaviours.

alexec:
They seem to be related to "shutdown strategy". 

* `Terminate` do not run any more node nodes in the workflow.
* `Stop` do not run any more nodes -  except exit handlers.

I'm not clear what happens to running nodes.

simster7:
>  I'm not clear what happens to running nodes.

In both cases running nodes are stopped.

`Terminate` stops the workflow immediately, while `Stop` stops the workflow _but allows all exit handlers to run_.

knabben:
What about adding these notes to the Cobra helper? I can help with this ticket.

https://github.com/argoproj/argo/blob/master/cmd/argo/commands/terminate.go#L16
https://github.com/argoproj/argo/blob/master/cmd/argo/commands/stop.go#L27

simster7:
Sure, that would be great. I would also find a place in the docs to add the same information

antoinetran:
> `Terminate` stops the workflow immediately, while `Stop` stops the workflow _but allows all exit handlers to run_.

@simster7 I was confused and thought terminate was stop and delete the pods. I think "Force stop" instead of "Terminate" would be better.

PR #11626, opened 2023-08-19
docs(cli): clarify `stop` v. `terminate` with `Long` descriptions

agilgur5:
### Motivation

Similar to #11625, this is another one of the most frequently asked questions by users and so could definitely use some clarification

### Modifications

- Following the same pattern as #11625, add a `Long` description to the CLI commands that details how `stop` and `terminate` work in more detail
  - via `make codegen`, this resulted in a `### Synopsis` section in the generated docs as well, which is also helpful
  
Descriptions loosely follow the official issue response in https://github.com/argoproj/argo-workflows/issues/4454#issuecomment-721821469.
This is effectively a follow-up to #4653

### Verification

- `make codegen` passes

### Notes for Reviewers

- I wrote this as a "fix" commit so that it can be added as a patch to the CLI (and make it into any cherry-picks), but could make it a "docs" commit instead.

- Feel free to add more modifications to the long descriptions!

### Future Work

I'd like to add these descriptions to the API docs as well, but that seems to be a larger effort as the API docs have _no_ descriptions right now 😕 I'm also not sure where codegen tries to pull the descriptions from, as there are GoDoc descriptions, but perhaps not in the right places?

### DIFF
--- a/cmd/argo/commands/stop.go
+++ b/cmd/argo/commands/stop.go
@@ -38,6 +38,7 @@ func NewStopCommand() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "stop WORKFLOW WORKFLOW2...",
 		Short: "stop zero or more workflows allowing all exit handlers to run",
+		Long:  "Stop a workflow but still run exit handlers.",
 		Example: `# Stop a workflow:
 
   argo stop my-wf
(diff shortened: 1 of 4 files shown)


agilgur5:
Huh, lint [timed out](https://github.com/argoproj/argo-workflows/actions/runs/5913955535/job/16038915129?pr=11626). Haven't seen that happen before. Will need to retry that but may need to bump the timeout

isubasinghe (approved):

terrytangyuan (approved):

terrytangyuan:
Please fix lint

agilgur5:
> Please fix lint

Per my comment [above](https://github.com/argoproj/argo-workflows/pull/11626#issuecomment-1685131400), it timed out and needs a retry

agilgur5:
Looks like this is good to merge now

PR #11630, opened 2023-08-20
chore(deps): remove unused JS deps

agilgur5:
### Motivation

None of these deps are used -- note how the `yarn.lock` had no changes
- Less to install, less to update. No real security benefits as none of these made it into the bundle (just potential install security benefits)

### Modifications

- Remove `json-merge-patch` and `react-moment` from deps
- Remove `copyfiles`, `glob`, and `webfonts-generator` from devDeps

### Verification

- Confirmed that `yarn.lock` had no changes after `yarn install`
- Searched the codebase and confirmed that none of these are used in the current revision
- Install still succeeds, build still passes


terrytangyuan (commented):
Is there anything similar to go mod tidy but for typescript?

agilgur5:
Not built-in like that, unfortunately. `npm` isn't even part of the language, the way deps can be implicitly used (e.g. loaders are referenced but not imported in `webpack.config.js`) complicates things a lot, and all the compile-to-JS languages (e.g. TS, Vue, Svelte, etc) make it even harder (not to mention ambient types in TS).

The closest thing are a few external packages that try their best but can absolutely have false positives:
- [depcheck](https://github.com/depcheck/depcheck) is the most popular one, but it hasn't had a release in [a few years](https://github.com/depcheck/depcheck/discussions/800). It also has a good amount of false positives in the issues (some due to outdated nature)
- [knip](https://github.com/webpro/knip) is a recent one that is currently well maintained.


That being said, I have never used either and nor have I ever really needed something like that. Usually once you stop using a dep, you remove it from the `package.json` as well.
It happens that someone forgets, but pretty infrequently in my experience.

agilgur5:
Related, I am adding [`yarn-deduplicate`](https://github.com/scinos/yarn-deduplicate) in a separate PR. It has some vague similarities to `go mod tidy` but for the lockfile specifically (`go.sum` / `yarn.lock`), not for the main dependency manifest (`go.mod` / `package.json`)

terrytangyuan (approved):

PR #11635, opened 2023-08-20
build: Fix nix build

sudosubin:
### Motivation

Current `./dev/nix/flake.nix` is using go 1.19 for `pkgs.buildGoModule`, and the build fails.

Previously it worked well anyway, but after [this change](https://github.com/argoproj/argo-workflows/pull/11368/files#diff-b655785d3d88b4f926fa9bb629e20bee6b91d8f361832ba7942dfe266d0834beR21), the added `cmd.WaitDelay` makes the build completely fail on versions below go 1.20.

```sh
$ nix develop --extra-experimental-features nix-command --extra-experimental-features flakes ./dev/nix/ --impure
...
# github.com/argoproj/argo-workflows/v3/workflow/executor/os-specific
workflow/executor/os-specific/command.go:21:7: cmd.WaitDelay undefined (type *exec.Cmd has no field or method WaitDelay)
note: module requires Go 1.20
...

```

### Modifications

- Update `_module.args`, and add overlays for `go`, `buildGoModule` to use go 1.20.
- Update controller's `vendorSha256` hash

### Verification

```sh
$ nix develop --extra-experimental-features nix-command --extra-experimental-features flakes ./dev/nix/ --impure
(devenv) bash-5.1$
```

### DIFF
--- a/dev/nix/flake.nix
+++ b/dev/nix/flake.nix
@@ -13,7 +13,7 @@
     inputs.flake-parts.lib.mkFlake { inherit inputs; } {
       systems = [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ];
       imports = [ inputs.treefmt-nix.flakeModule ];
-      perSystem = { pkgs, lib, config, ... }:
+      perSystem = { pkgs, lib, config, system, ... }:
         let
           argoConfig = import ./conf.nix;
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
@isubasinghe Could you take a look?

isubasinghe (approved):
LGTM, we really need to do something about the dependency hashes needing constant updating.

terrytangyuan (approved):

PR #11628, opened 2023-08-20
build(ui): upgrade to Webpack v5 + upgrade loaders + plugins

agilgur5:
### Motivation

- Upgrade the UI build to use latest versions of many dev deps
  - Webpack v4 hasn't been updated in [2.5 years](https://github.com/webpack/webpack/releases/tag/v4.46.0).
  - Webpack v5 stable was released [3 years ago](https://github.com/webpack/webpack/releases/tag/v5.0.0), and RC / beta / alpha are even older
  - This unblocks other dep upgrades as well

#### Webpack Core

- upgrade `webpack` and `webpack-cli` to latest major v5
  - (`webpack-dev-server` does not have a v5, is already at [latest major and minor](https://github.com/webpack/webpack-dev-server/blob/v4.15.1/CHANGELOG.md))
  - fix config errors and deprecation warnings:
    - `[hash]` -> `[contenthash]` (https://webpack.js.org/migrate/5/#clean-up-configuration)
      - `contenthash` is just a better hash, as it is calculated per entrypoint instead for the entire bundle
        - that being said, we only have one entrypoint, so not super relevant. but `hash` was also deprecated
      - also same as [Argo CD config](https://github.com/argoproj/argo-cd/blob/6ca2e90b5094eba3878746fee5f2c357bfbc6cb8/ui/src/app/webpack.config.js#L20)
    - `node.fs: "empty"` -> `resolve.fallback.fs: false` (https://webpack.js.org/migrate/5/#clean-up-configuration)
      - also same as [Argo CD config](https://github.com/argoproj/argo-cd/blob/6ca2e90b5094eba3878746fee5f2c357bfbc6cb8/ui/src/app/webpack.config.js#L28)
    - `loaders` -> `use` (https://webpack.js.org/migrate/5/#update-outdated-options)
      - `!` syntax for chaining loaders has been fully removed as well. replaced by using arrays

#### Plugins

- upgrade `HtmlWebpackPlugin` and `CopyWebpackPlugin` to latest major:
  - `html-webpack-plugin` to v5's main [breaking changes](https://github.com/jantimon/html-webpack-plugin/blob/v5.5.3/CHANGELOG.md#500-2021-02-03) are to drop support for Webpack v4 and Node <= 10
  - `copy-webpack-plugin` to v11 has some options changes as well as Webpack and Node support drops:
[...]

PR #11637, opened 2023-08-20
ci(deps): dedupe `yarn.lock`, add check for dupes

agilgur5:
### Motivation

Use as few duplicates of dependencies as possible, reducing install size, memory usage, and bundle size
- See also https://github.com/atlassian/yarn-deduplicate#why-is-this-necessary and a PR from a repo I used to maintain: https://github.com/jaredpalmer/tsdx/pull/683
- Removes 700+ lines from the `yarn.lock` by condensing deps

### Modifications

- add a `deduplicate` script that uses `yarn-deduplicate`
- run `deduplicate` during `make lint` and UI CI
  - I added this as another one-liner in `make lint` b/c that was honestly _much_ more readable than trying to do multi-line shell conditionals inside of a `Makefile` (which require `\` and `;` as they get interpreted [as a single line](https://stackoverflow.com/a/58602879/3431180))
- run `deduplicate` once and commit the results
  - double-checked 50%+ of these automated changes and they all look good to me

similar to a commit I made in a library I maintained a few years ago: https://github.com/jaredpalmer/tsdx/commit/22133ce025f5f3a80cc797b15cdec3960b2ba3ae (part of https://github.com/jaredpalmer/tsdx/pull/683)

Also, upgrade `ts-loader` to v8.4.0
- To support TS v4.7+ per [the changelog](https://github.com/TypeStrong/ts-loader/blob/v9.4.4/CHANGELOG.md#v840)
  - TS v4.6 was [deduped](https://github.com/argoproj/argo-workflows/pull/11637/files#diff-3a968206d6de2fecfc5dacd7d94bab7744c9f5d5c999a816164d95cbc135c316L10092). Previously there were two versions of TS being used by the build chain -- which itself could lead to some inconsistencies 
- `ts-loader` [v8](https://github.com/TypeStrong/ts-loader/blob/v9.4.4/CHANGELOG.md#v800)'s only breaking change was to no longer support TS <3.6
  - `ts-loader` [v9](https://github.com/TypeStrong/ts-loader/blob/v9.4.4/CHANGELOG.md#v900) requires Webpack v5, so cannot bump to that yet (see also #11628)

### Verification

Ran the new `make lint`: confirmed it worked and did not make an additional diff.
Spot check on UI render was ok.
CI checks pass

### Notes to Reviewers

This will likely merge conflict _hard_ with a few my other open UI deps PRs, so it probably makes more sense to merge those first and then update this PR after those are all merged

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -257,6 +257,7 @@ jobs:
       - run: yarn --cwd ui build
       - run: yarn --cwd ui test
       - run: yarn --cwd ui lint
+      - run: yarn --cwd ui deduplicate
       - run: git diff --exit-code
       # check to see if it'll start (but not if it'll render)
       - run: yarn --cwd ui start &
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ ui/dist/app/index.html: $(shell find ui/src -type f && find ui -maxdepth 1 -type
 $(GOPATH)/bin/staticfiles:
 # update this in Nix when updating it here
 ifneq ($(USE_NIX), true)
-	go install bou.ke/staticfiles@dd04075 
+	go install bou.ke/staticfiles@dd04075
 endif
 
 ifeq ($(STATIC_FILES),true)
@@ -456,6 +456,8 @@ lint: server/static/files.go $(GOPATH)/bin/golangci-lint
 	$(GOPATH)/bin/golangci-lint run --fix --verbose
 	# Lint the UI
 	if [ -e ui/node_modules ]; then yarn --cwd ui lint ; fi
+	# Deduplicate Node modules
+	if [ -e ui/node_modules ]; then yarn --cwd ui deduplicate ; fi
 
 # for local we have a faster target that prints to stdout, does not use json, and can cache because it has no coverage
 .PHONY: test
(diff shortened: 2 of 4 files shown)


agilgur5:
Huh Prod UI build is [failing](https://github.com/argoproj/argo-workflows/actions/runs/5919743411/job/16050307225?pr=11637) on a `ts-loader` error. Was able to reproduce that locally, but it doesn't seem to occur in dev mode probably because [type-checking is off](https://github.com/argoproj/argo-workflows/blob/8a52da5e8ee6eeabffb6c7e5858702129b37b525/ui/src/app/webpack.config.js#L37C79-L37C103).

That error suggests that one of the deps was relying on an outdated version of TS

agilgur5:
Ah yep, [TS 4.6 was deduped](https://github.com/argoproj/argo-workflows/pull/11637/files#diff-3a968206d6de2fecfc5dacd7d94bab7744c9f5d5c999a816164d95cbc135c316L10092) and [`ts-loader` needs an update to support TS 4.7+](https://github.com/TypeStrong/ts-loader/blob/v9.4.4/CHANGELOG.md#v840).
(For reference, TS [does not follow SemVer](https://github.com/microsoft/TypeScript/issues/14116), so there can be breaking changes in minors like this one)

terrytangyuan (commented):
Can we combine this with other similar PRs?

agilgur5:
So all 3 (#11641, #11630, and this one) are technically independent of each other.
In particular, this is the only one that massively merge conflicts when there's a dep change, the other two can be merged pretty cleanly. Pretty sure both of those can be merged right now.

terrytangyuan (approved):

agilgur5:
Geh, some other PR ended up merge conflicting as well 😭 Possibly #11678.

Rebased on top

terrytangyuan (approved):

PR #11638, opened 2023-08-20
fix: ensure labels is defined before key access. Fixes #11602

### LINKED ISSUE #11602: UI - label undefined for some workflows
### Pre-requisites

UI accesses "label" field of non existing object causing it to crash. 

Will fix this myself. 

### Version

latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Verified manually, this is a simple change, just checks if a key existed before accessing it.


ui/src/app/workflows/components/workflows-row/workflows-row.tsx
```diff
@@ -94,7 +94,8 @@ export class WorkflowsRow extends React.Component<WorkflowsRowProps, WorkflowRow
                         </div>
                         <div className='columns small-1'>{isArchivedWorkflow(wf) ? 'true' : 'false'}</div>
                         {(this.props.columns || []).map(column => {
-                            const value = wf.metadata?.labels[column.key];
+                            // best not to make any assumptions and wait until this data is filled
+                            const value = wf?.metadata?.labels ? wf.metadata.labels[column.key] : 'unknown';
```
agilgur5, line 98:
```suggestion
                            const value = wf?.metadata?.labels?.[column.key] ?? 'unknown';
```

I believe this is the more conventional syntax nowadays with some slightly nicer properties

Though I think with `?.[column.key]`, `'unknown'` would technically no longer be required, so could also also do:

```suggestion
                            const value = wf?.metadata?.labels?.[column.key];
```

isubasinghe:
Thanks @agilgur5, I went with the first option.

terrytangyuan (approved):

PR #11585, opened 2023-08-15
fix: upgrade `argo-ui` components to latest

agilgur5:
Replaces #11510, fixes all issues I commented about
Fixes #11448 by incorporating https://github.com/argoproj/argo-ui/pull/418

### Motivation

- Upgrade to latest version of `argo-ui` components
  - Upgrade and force resolve other deps to make sure build passes with latest version
- Fix intermediate parameters issue with the overflow fix
- Fix deprecation warnings in build

### Modifications

- Pin `argo-ui` to latest SHA: https://github.com/argoproj/argo-ui/commit/87d27fb1cb4f6e3ac4a49f85747e471b2efa7512
  - Add `@types/uuid` to fix lack of typing (see https://github.com/argoproj/argo-ui/pull/367 and https://github.com/argoproj/argo-cd/pull/13836)
  - Force resolve `react-toastify` to 9.0.3
    - This was pinned to 9.0.8 in https://github.com/argoproj/argo-ui/pull/379 but due to https://github.com/fkhadra/react-toastify/issues/775, that version is incompatible with Webpack/CRA v4 and was causing errors. Followed [this comment](https://github.com/fkhadra/react-toastify/issues/775#issuecomment-1150715526) to pin to 9.0.3, which fixed the errors
      - Argo CD [uses](https://github.com/argoproj/argo-cd/blob/bfaac2b5ac4b96c8af8158902c1b086f7ced7389/ui/package.json#L119) Webpack v5, so I assume that is why they didn't see the same issue
    - Enable `esModuleInterop` as this was added [in Argo UI](https://github.com/argoproj/argo-ui/pull/367/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R9), and since Argo UI is not compiled (we import source TS), we have to match its `tsconfig.json` to get the same results
       - This fixes a few import errors (as the errors say)
       - Use default imports for `classnames` and `moment` instead of namespace imports to fix the remaining import errors (as the errors suggest)

- Update `foundation-sites` to fix SCSS deprecation warnings
  - see https://github.com/foundation/foundation-sites/releases/tag/v6.7.0 which had various upgrades and compat fixes 

### Verification

Build passes, install succeeds

PR #3009, opened 2020-05-12
fix: Improve robustness of releases. Fixes #3004

alexec:
Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed the CLA.
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not. 
* [x] My organization is added to USERS.md.

Fixes #3004

v2.8.0 had no less than three problems:

* Manifests pointed to the incorrect version.
* Images were not pushed correctly.
* Homebrew was wrong.

I believe this all stems from the v2.8.0 tag having been applied to an old commit. Either way, we (I) should have a checklist to prevent these problems.

Additionally, what should be clean builds show `dirty`. In the past this has been things like file permissions. I'd like to understand why this is.

PR #11641, opened 2023-08-21
chore(deps): remove unneeded Yarn resolutions

agilgur5:
### Motivation

- removing these has no impact on resolution, as can be seen in the `yarn.lock`
  - any bumps in these deps will produce either the same minor or same major version (most are same minor version) -- i.e. no breaking changes
  - basically, these are extraneous as of this time
  
Related to #11630, which removes unused deps; this removes unneeded `resolutions`

### Modifications

- remove resolutions for `@types/react`, `autolinker`, `fast-json-patch`, `lodash`, and `prismjs`

- did _not_ touch `semver` or `react-toastify`
  - the `semver` resolution currently forces _all versions_ to the same as the resolution (including different majors, which can be seen [in the `yarn.lock`](https://github.com/argoproj/argo-workflows/blob/8a52da5e8ee6eeabffb6c7e5858702129b37b525/ui/yarn.lock#L8959))
  - ~the `react-toastify` resolution was needed for #11585 and it is removed as part of build upgrades in #11628~
    - **EDIT**: rebased on top so this is no longer present

### Verification

Install succeeds, build passes.
Spot check on render.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -85,11 +85,6 @@
         "webpack-dev-server": "^4.15.0"
     },
     "resolutions": {
-        "@types/react": "16.8.5",
-        "autolinker": "3.16.1",
-        "fast-json-patch": "3.1.1",
-        "lodash": "4.17.21",
-        "prismjs": "1.27.0",
         "semver": "^7.5.2"
     }
 }
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -1879,7 +1879,7 @@
     "@types/history" "*"
     "@types/react" "*"
 
-"@types/react@*", "@types/react@16.8.5", "@types/react@^16.8.5":
+"@types/react@*", "@types/react@^16.8.5":
   version "16.8.5"
   resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.5.tgz#03b9a6597bc20f6eaaed43f377a160f7e41c2b90"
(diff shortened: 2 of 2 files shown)


terrytangyuan (commented):
I remember these resolutions were added to resolve security issues.

agilgur5:
Per the first line of the PR, removing these has no impact on resolution. I imagine that those resolutions are either outdated or were used incorrectly to update a version in `yarn.lock`. Looking at #10842, that was indeed used to update a version `yarn.lock`.

`yarn.lock` can just be updated independently if the bump is within the SemVer range; it does not need a resolution.
You will need to comment out the `--frozen-lockfile` flag temporarily in [the `.yarnrc` though](https://github.com/argoproj/argo-workflows/blob/d74929a69130fdefed0608f708767025b2de90a7/ui/.yarnrc#L3). Normally that flag is only set in CI and not during development (it was made as a default [here](https://github.com/argoproj/argo-workflows/pull/3009/files#r423909775), seemingly so that it doesn't have to be individually set in multiple CI builds?)

terrytangyuan (approved):
We can merge and see if Snyk check passes

agilgur5:
It [did](https://github.com/argoproj/argo-workflows/actions/runs/5980065079/job/16225432786) 😉 .
This is basically a no-op as all the versions in the `yarn.lock` don't change.

PR #2134, opened 2020-02-03
fix: route all unknown web content requests to index.html

tcolgate:
Untested.

Fixes: #2103

Issue #3080, opened 2020-05-22
BASE_HREF  is not working


yuvraj9:
Cant Access argo UI if I change the BASE_HREF to something other than /

I want it to be accessible at xxx.com/argo 

To reproduce it you can just change the env variable BASE_HREF to something other than /. I have tried these - argo, /argo , /argo/ . None of these work.

Getting this on the console of my browser -
```
 Uncaught SyntaxError: Unexpected token '<'
```
THis is the image we are using: argoproj/argocli:v2.7.4

- Kubernetes version : 1.16
---

**Message from the maintainers**:

If you are impacted by this bug please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

ramanNarasimhan77:
Hello, 

### BASE_HREF setup
```
- name: BASE_HREF
          value: /argo/
```

### Ingress setup

```
rules:
  - host: {{domain-name}}
    http:
      paths:
      - backend:
          serviceName: argo-server
          servicePort: http
        path: /argo(/|$)(.*)

```

I can see workflows in https://{{domain-name}}/argo/workflows

But when I click on a workflow, I see a blank screen

<img width="1605" alt="Screenshot 2020-06-23 at 8 43 46 PM" src="https://user-images.githubusercontent.com/52691361/85424151-2e507200-b595-11ea-8a21-71c6e04ce299.png">


Under network I can see that {{namespace}}?listOptions.fieldSelector=metadata.name=artifact-passing-wj8b4 takes considerable time and it is successful after ~2mins and response does not show on the UI

<img width="1680" alt="Screenshot 2020-06-23 at 9 00 15 PM" src="https://user-images.githubusercontent.com/52691361/85424159-327c8f80-b595-11ea-9549-4d5eac7dad66.png">

alexec:
> I can see workflows in https://{{domain-name}}/argo/workflows

this screen can be blank if the workflow in invalid

ramanNarasimhan77:
> This screen can be blank if the workflow in invalid. Check under the summary tab (one of the the small buttons top right)

@alexec 
When i submit a workflow from the ui, i can see the steps
When i go back to main workflows/ page and then click on the workflow, I get a blank screen

On clicking resubmit from the blank screen, I can see all the steps
![image](https://user-images.githubusercontent.com/52691361/85428612-15e35600-b59b-11ea-9167-c01a7c715df9.png)

going back to main page
[...]

PR #11644, opened 2023-08-21
fix: base href does not work #11401 #10194 #2134

### LINKED ISSUE #7767: Implement rootpath similar to argo-cd
# Summary

Currently, running argo-workflows on a subpath requires rewrites on the reverse proxy. In argo-cd you can avoid that by setting extra arguments to the entrypoint command.
```
- --rootpath
- /argocd
- --basehref
- /argocd
```
Would it be possible to replicate this behavior on argo-workflows as well?

# Use Cases

AWS Application Load Balancers do not support rewrites. Our setup doesn't use any additional controller like nginx.

---

### LINKED ISSUE #10194: baseUrl doesnt appear to work
### Pre-requisites

When running the argo-server, I have added the environment variable `BASE_HREF=/argo/`. However, when I port-forward the argo-server to localhost, I notice that `localhost:8081` still works and redirects to `localhost:8081/workflows`, and I can navigate the UI. However, if I try to navigate to `localhost:8081/argo/` I get `Uncaught SyntaxError: Unexpected token '<' (at main.2430295409b8b54e52ad.js:1:1)`. I confirmed in the logs that argo is picking up the BASE_HREF env var (see logs below), but it just seems to ignore it. 

I expected to be able to see the UI on `localhost:8081/argo`. I am trying to use the BASE_HREF in order to point an nginx instance at the UI. 

Here are the logs from the server:
```
time="2022-12-08T21:27:31.747Z" level=info msg="not enabling pprof debug endpoints"
time="2022-12-08T21:27:31.747Z" level=info authModes="[server client]" baseHRef=/argo/ managedNamespace=quilter-dev-sergiy namespace=quilter-dev-sergiy secure=false ssoNamespace=quilter-dev-sergiy
time="2022-12-08T21:27:31.747Z" level=warning msg="You are running in insecure mode. Learn how to enable transport layer security: https://argoproj.github.io/argo-workflows/tls/"
time="2022-12-08T21:27:31.747Z" level=info msg="SSO disabled"
time="2022-12-08T21:27:31.758Z" level=info msg="Starting Argo Server" instanceID= version=v3.4.3
time="2022-12-08T21:27:31.767Z" level=info msg="Creating event controller" asyncDispatch=false operationQueueSize=16 workerCount=4
time="2022-12-08T21:27:31.769Z" level=info msg="GRPC Server Max Message Size, MaxGRPCMessageSize, is set" GRPC_MESSAGE_SIZE=104857600
time="2022-12-08T21:27:31.770Z" level=info msg="Argo Server started successfully on http://localhost:2746" url="http://localhost:2746"
time="2022-12-08T21:27:51.135Z" level=info duration="80.64µs" method=GET path=index.html size=478 status=0
```

### Version

3.4.3 or latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

### LINKED ISSUE #11401: v3.4.8 only works in / endpoint
### Pre-requisites

if i configure 
env:
          - name: BASE_HREF
            value: /argo/

i have a blank page

### Version

v3.4.8

```YAML
blank page
```

```text
theres are no error messages
```

```text
theres are no error messages
```

### LINKED ISSUE #11643: Setting base-href to "/argo/" and then accessing via portforward does not work.
### Pre-requisites

I installed https://github.com/argoproj/argo-workflows/blob/master/manifests/quick-start-minimal.yaml 

Made the following changes.
1. Added extra server argument `--insecure=true` for testing without tls
2. Added env BASE_HREF as `/argo/`
3. Changed HTTPS to HTTP in container probe

image

I port-forwarded Argo-server:2746 to localhost:2747 and got blank page

image

I expected to be able to see the Argo home page at "/argo/"

### Version

latest

```YAML
NA
```

```text

NA
```

```text

NA
```

sonbui00:
Fix #11401 
Fix #10194 
Fix #11643
Fix #7767
Related #2134

### Motivation

Support BASE_HREF on the document https://argoproj.github.io/argo-workflows/argo-server/

### Modifications

Add handle base href for static file and test

### Verification
make test

### More
As I check this issue, I see the lib to embed static files into Go binary has been deprecated (https://github.com/bouk/staticfiles). We should replace it.

### DIFF
--- a/server/static/static.go
+++ b/server/static/static.go
@@ -11,15 +11,22 @@ type FilesServer struct {
 	hsts            bool
 	xframeOpts      string
 	corsAllowOrigin string
+	Hash            func(string) string
 }
 
 func NewFilesServer(baseHRef string, hsts bool, xframeOpts string, corsAllowOrigin string) *FilesServer {
-	return &FilesServer{baseHRef, hsts, xframeOpts, corsAllowOrigin}
+	return &FilesServer{baseHRef, hsts, xframeOpts, corsAllowOrigin, Hash}
 }
 
 func (s *FilesServer) ServerFiles(w http.ResponseWriter, r *http.Request) {
+	if s.baseHRef != "/" {
+		p := strings.TrimPrefix(r.URL.Path, s.baseHRef)
+		if len(p) < len(r.URL.Path) {
+			r.URL.Path = p
+		}
+	}
 	// If there is no stored static file, we'll redirect to the js app
-	if Hash(strings.TrimLeft(r.URL.Path, "/")) == "" {
+	if s.Hash(strings.TrimLeft(r.URL.Path, "/")) == "" {
 		r.URL.Path = "index.html"
 	}
 
--- a/server/static/static_test.go
+++ b/server/static/static_test.go
@@ -1,0 +1,28 @@
+package static
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestServeFile(t *testing.T) {
+	t.Run("should handle base href for static file", func(t *testing.T) {
+		filesServer := NewFilesServer("/argotest/", false, "", "")
+		filesServer.Hash = func(s string) string {
+			resources := map[string]string{
+				"main.js": "main.js",
+			}
+			return resources[s]
+		}
+
+		req := httptest.NewRequest(http.MethodGet, "/argotest/main.js", nil)
+		w := httptest.NewRecorder()
+
+		filesServer.ServerFiles(w, req)
+
+		if req.URL.Path != "main.js" {
+			t.Errorf("should handle base href, expect %v, got %v", "main.js", req.URL.Path)
+		}
+	})
+}


agilgur5 (commented):
Hmm I need to take a look at this more closely, but this looks more like partial rootpath support rather than base href from a glance. 

With just base href, the proxy in front should be rewriting these subpaths. The Argo Server still sees requests from `/`. Or, effectively, base href primarily impacts links, such as those on [the front-end](https://github.com/argoproj/argo-workflows/blob/cbf0ae9171cf3cb33e815ffc252d0e8f14ae22bd/ui/src/app/shared/base.ts#L1)

sonbui00:
@agilgur5 what "partial rootpath support" mean? 

> With just base href, the proxy in front should be rewriting these subpaths
=> Can you give me an example?

> The Argo Server still sees all requests from /
=> I tried to keep this compatible with other functions, ex: readiness check

agilgur5:
> @agilgur5 what "partial rootpath support" mean?

As in a partial implementation of #7767.

> => Can you give me an example?

#3080 has many examples of ingress config with rewrites. [This one](https://github.com/argoproj/argo-workflows/issues/3080#issuecomment-657956161) from the top is pretty concise and uses the common NGINX IGC

sonbui00:
@agilgur5 I'm sorry, but I do not understand. Let's me know if you want any info here

agilgur5:
> @agilgur5 I'm sorry, but I do not understand.

Let me try two other explanations:
1. [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) is an HTML element. Per the MDN docs, the `href` attribute only affects HTML links. It does _not_ change anything on the back-end.
    - With a Single Page App (SPA) on the UI, JS links also need to be changed. Hence [the JS](https://github.com/argoproj/argo-workflows/blob/cbf0ae9171cf3cb33e815ffc252d0e8f14ae22bd/ui/src/app/shared/base.ts#L1) I linked before.
  
2. `/argotest/main.js` should never occur. The proxy in front of the server is supposed to rewrite that to `/main.js`. The server only sees `/main.js` and never sees `/argotest/main.js` because there is a proxy in front.
    - Without rootpath support (i.e. #3080), the back-end basically has no knowledge of the rewrite. Rootpath support is similar to implementing the rewrite within Argo.

agilgur5 (changes requested):
I don't think this is actionable as a "fix" to base href. base href is working correctly.

This could be used as part of a feature to implement rootpath (#7767), but that should probably be done as one larger PR

sonbui00:
> I don't think this is actionable as a "fix" to base href. base href is working correctly.

Can you make a test? I don't think it work like the document and I also made some test before. That's why many people report this in some issues. Also you can read the fix code to know the reason

sonbui00:
@agilgur5 with ingress rewrite, I think this is not a issue. I tried without ingress rewrite for this

agilgur5:
> Can you make a test?

The back-end is working correctly. Per https://github.com/argoproj/argo-workflows/issues/11401#issuecomment-1685877020, the Server logs are correct, correctly detecting the environment variable. Per https://github.com/argoproj/argo-workflows/issues/11643#issuecomment-1685871836, the Server correctly sets `<base href="/argo/">`.

That is the only impact to the back-end that base href has. All other impacts are front-end facing.

> That's why many people report this in some issues

Per above, both users' own issues show that it is working correctly.
Per https://github.com/argoproj/argo-workflows/issues/10194#issuecomment-1685866941, I think many users are _misinterpreting_ the base href feature for rootpath, but those are two _separate_ features.
Base href is for the front-end, rootpath is for the back-end.

agilgur5:
> @agilgur5 with ingress rewrite, I think this is not a issue. I tried without ingress rewrite for this

Yes exactly! Base href is intended to be used _in conjunction to_ an ingress rewrite

sonbui00:
> Yes exactly! Base href is intended to be used in conjunction to an ingress rewrite

@agilgur5 Maybe ingress rewrite added to the document after have this issue. This fix can run without ingress rewrite. The document updated after this issue https://github.com/argoproj/argo-workflows/issues/3080

agilgur5:
> This fix can run without ingress rewrite.

That's just a separate feature entirely. That is part of rootpath support #7767. And totally feel free to work on #7767!

Base href and rootpath are just two separate (but related) features. Note that the opening comment of #7767 uses _both_ `--basehref` _and_ `--rootpath` together.

"Base href" _specifically_ is a front-end only feature designed to be used with ingress rewrites. It is making the front-end _aware_ of a back-end configuration. The back-end configuration can be done either with an ingress rewrite _or_ with `--rootpath`.

sonbui00:
@agilgur5 
I think it this project basehref and rootpath is the same. You will see that they are set to same value. 
Also please note on https://github.com/argoproj/argo-workflows/issues/7767, purpose of the the issue is to fix `running argo-workflows on a subpath requires rewrites on the reverse proxy `. The suggest about `rootpath` is a way to archive the purpose. I think this PR also fix #7767

agilgur5:
> I think in this project basehref and rootpath is the same.

No they are separate, it's just that rootpath was never implemented.
I also think it is good for Argo projects to be consistent where possible (and where it makes sense), so if Argo CD has those as two separate flags, Argo Workflows can implement that consistently.

Moreover, "base href" refers to a Web standard, literally the HTML `base` element and its `href` attribute. I don't think we should redefine a Web standard. Things like that tend to cause more confusion rather than less. We should try to teach users about the existing standards and how they work -- which is why those were good docs changes. We might be able to further improve those docs.

`--rootpath` also would require more changes. Ingress rewrites change the external API route as well (e.g. `my-domain.com/api` -> `my-domain.com/argo/api`). So we'd need to do the same thing for `--rootpath`. https://github.com/argoproj/argo-workflows/issues/7767#issuecomment-1031207499 shows several of the changes in CD required (like the GRPC gateway).

sonbui00:
@agilgur5 Could you give me a real life example for rootpath and basehref with different values for running Argo Workflow?

sonbui00:
> `my-domain.com/argo/api`with prefix /argo/ -> should work without rewrite on additional controller for #7767

I think my PR do not handle this

sonbui00:
@agilgur5 Maybe we also need rootpath to handle all cases. let me try to figure it out

agilgur5:
> @agilgur5 Could you give me a real life example for rootpath and basehref with different values for running Argo Workflow?

The main one is an ingress rewrite. For instance, if I have `my-cluster.com` as my main cluster domain and all services as subpaths, then I could have `my-cluster.com/argo-cd/`, `my-cluster.com/argo-workflows/`, `my-cluster.com/argo-events/`, `my-cluster.com/prometheus/`, etc.

If all those services did not support rootpath, it would be easier to configure ingress rewrites for all of them in the same way -- only the subpath would be different between them. If they did all support rootpath, they may all also require different configuration (e.g. a configmap vs a flag vs an env var, etc), so rewrites may be easier/more consistent as well.

I can think of some other wacky use-cases, but that's the only one I can think of off the top of my head that is a "real life example" that I and others have definitely used before.

agilgur5:
> I think my PR do not handle this

> @agilgur5 Maybe we also need rootpath to handle all cases. let me try to figure it out

Yea that's what I meant when I said that full rootpath support to handle all cases would require a larger PR.
Feel free to take that on if you want!

PR #11645, opened 2023-08-22
fix: argo logs completion

zhangzhiqiangcs:
Fix arog logs command completion

### DIFF
--- a/cmd/argo/commands/completion.go
+++ b/cmd/argo/commands/completion.go
@@ -42,22 +42,13 @@ __argo_get_cron_workflow() {
 
 __argo_get_logs() {
 	# Determine if were completing a workflow or not.
-	local workflow=0
-	for comp_word in "${COMP_WORDS[@]}"; do
-		if [[ $comp_word =~ ^(-w|--workflow)$ ]]; then
-			workflow=1
-			break
-		fi
-	done
-
-	# If completing a workflow, call normal function.
-	if [[ $workflow -eq 1 ]]; then
+	if [[ $prev == "logs" ]]; then
 		__argo_get_workflow && return $?
-	fi
-
+ 	fi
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
Can you paste the output of the new completions?

zhangzhiqiangcs:
> Can you paste the output of the new completions?

First tab, will output workflow suggestions. Second tab, will output pod list of the workflow.

image

Total completion content has too many lines. Hers is the changed part
```bash
__argo_get_logs() {
    # Determine if were completing a workflow or not.
    if [[ $prev == "logs" ]]; then     
          __argo_get_workflow && return $?
    fi
    local workflow=$prev  
    # Otherwise, complete the list of pods
    local -a kubectl_out
    if kubectl_out=($(kubectl get pods --no-headers --selector=workflows.argoproj.io/workflow="${workflow}" 2>/dev/null | awk '{print $1}' 2>/dev/null)); then
        COMPREPLY+=( $( compgen -W "${kubectl_out[*]}" -- "$cur" ) )
    fi
}
```

cmd/argo/commands/completion.go
```diff
@@ -42,22 +42,13 @@ __argo_get_cron_workflow() {
...
-
-	# If completing a workflow, call normal function.
-	if [[ $workflow -eq 1 ]]; then
-		__argo_get_workflow && return $?
-	fi
-
+	if [[ $prev == "logs" ]]; then
+          __argo_get_workflow && return $?
+ 	fi
+    local workflow=$prev
```
terrytangyuan, line 48: Wrong indentation here

zhangzhiqiangcs: fixed

terrytangyuan (approved):

Issue #11499, opened 2023-08-01
Indirect infinite recursion doesn't get prevented


Joibel:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

With a workflow which recurses the controller will attempt to recurse forever. Take the recursion example from [the documentaton](../blob/master/docs/walk-through/recursion.md) and make it always coinflip for an example.
It feels like this should be prevented by having a configurable maximum recursion depth.

### Version

Current master

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: recursive2-
spec:
  entrypoint: coinflip
  templates:
  - name: coinflip
    steps:
    # flip a coin
    - - name: flip-coin
        template: flip-coin
    # evaluate the result in parallel
    - - name: heads
        template: coinflip              # call heads template if "heads"
        when: "{{steps.flip-coin.outputs.result}} == heads"
      - name: tails                     # keep flipping coins if "tails"
        template: coinflip
        when: "{{steps.flip-coin.outputs.result}} == tails"

  - name: coinflip
    script:
      image: python:alpine3.6
      command: [python]
      source: |
        import random
        result = "heads" if random.randint(0,1) == 0 else "tails"
        print(result)

  - name: heads
    container:
      image: alpine:3.6
      command: [sh, -c]
      args: ["echo \"it was heads\""]
```

### Logs from the workflow controller

```text
Not interesting
```

### Logs from in your workflow's wait container

```text
Not relevant
```

terrytangyuan:
Perhaps some additional checks can be specified in the `when` condition?

Joibel:
> Perhaps some additional checks can be specified in the `when` condition?

The actual workflow here isn't the problem, it's an obvious case which will recurse indefinitely. I'm really looking at trying to stop much more complex workflows where the recursion isn't obvious from inspection, but the outcome of chewing up resources is undesirable.

#4180 has thought about this problem before and a fix wasn't agreed on, but the basic problem does remain and does happen in the real world.

jessesuen:
[...]

PR #4193, opened 2020-10-02
fix(controller): Set a configurable max recursion depth limit. Closes: #4180

simster7:
Closes: #4180 

Needs tests

PR #11646, opened 2023-08-22
feat: Set a max recursion depth limit

Joibel:
This is a resurrection of #4193 to address #11499

# Motivation
Intentional or unintentionally it is possible to write templates that indefinitely recurse. This change prevents that by monitoring the stack depth of templates, and stopping when it reaches 150 which seems big enough for anyone.

It can be disabled globally with a controller environment variable in case anyone comes across problems and we can address those problems with new fixes if they are reported.


docs/scaling.md
```diff
@@ -52,6 +52,12 @@ argo --instanceid i1 submit my-wf.yaml
 
 You do not need to have one instance ID per namespace, you could have many or few.
 
+### Maximum Recursion Depth
+
+In order to protect users against infinite recursion, the controller has a default maximum recursion depth of 150 calls to templates.
+
+This protection can be disabled with the environment variable `DISABLE_MAX_RECURSION=true`
```
agilgur5, line 59:
```suggestion
This protection can be disabled with the [environment variable](environment-variables.md#controller) `DISABLE_MAX_RECURSION=true`
```

isubasinghe (approved):
Generally looks good, I have a nit with the `os.Getenv`, but this is how the rest of the codebase does it so not a big deal. 

The nit: os.Getenv is a syscall, this value may differ throughout the execution of the operator function. I can see both sides to this, perhaps this is desired behaviour. The functional programmer in me does not like it one bit though. 

Regardless I've approved, grepping shows that argo commonly does this, so changing it just here won't change a thing.

Joibel:
@isubasinghe if these env vars were used for anything 'non experimental' I'd suggest we read them at startup and provide them via a service, but that's a bit overkill for the current intended use.

isubasinghe:
@Joibel yeah that is fair enough, I think my previous comment was largely motivated by my subjective sense of "niceness" rather than practicality.
I've approved it anyway.

workflow/controller/controller.go
```diff
@@ -67,6 +67,8 @@ import (
 	plugin "github.com/argoproj/argo-workflows/v3/workflow/util/plugins"
 )
 
+const defaultMaxStackDepth = 150
```
terrytangyuan, line 70: Would this be a reasonable default?

Joibel: This number comes from #4193, where it started off as 500, and was amended to 150. #10785 removed 10 as the limit, but that feels unreasonably low and likely to cause issues with existing workflows.
Note, this is not actually a default because it's uncontrollable (again #4193 started with it configurable via the environment, but switched to a disable instead). I will change it's name.
My opinion is that 150 is too big, if I was plucking a number out of the air I'd set it lower at around 50. What's your guess at a good number @terrytangyuan?

terrytangyuan: Is there a number we can use where the controller crashes once the limit is reached?

deHaar (approved):
I really like this kind of recursion prevention, we produced infinite recursion due to equal names of a step and an entrypoint. Spent a lot of time to stop/terminate/delete the workflow that called its own entrypoint before we spent even more time finding out why this occurred.
This prevention would have saved us much time, so I think it should be merged to save time for others in future.

workflow/controller/controller.go
Joibel: I'm think by the time the controller is having trouble we've gone too far.
Without status offloading you'll probably have wasted a lot of resources but hit failure to store status before you get the controller in trouble. I don't think we should just be aiming to stop the controller getting into trouble.
I think this number should aim for a value which is reasonable for workflows already in existence so as few users as possible consider this a regression, but is otherwise low enough to prevent wasting time doing work which will be wasted because we're actually in an infinite recursion loop.
I'm going to propose reducing the 150 to 100 and see what happens in the real world, what do you think?

Joibel: My preference would be to leave it enabled as a saner default. Switching it 'later' doesn't really help with not breaking stuff, and I'm expecting this to go in a major not minor release (hence flagged as `feat`).

docs/environment-variables.md
```diff
@@ -27,6 +27,7 @@ most users. Environment variables may be removed at any time.
...
 | `DEFAULT_REQUEUE_TIME`                 | `time.Duration`     | `10s`                                                                                       | The re-queue time for the rate limiter of the workflow queue.                                                                                                                                                                                                            |
+| `DISABLE_MAX_RECURSION`                | `bool`              | `false`                                                                                     | Set to true to disable the recursion preventer, which will stop a workflow running which has called into a child template 150 times                                                                                                                                      |
```
terrytangyuan, line 30:
```suggestion
| `DISABLE_MAX_RECURSION`                | `bool`              | `false`                                                                                     | Set to true to disable the recursion preventer, which will stop a workflow running which has called into a child template 100 times                                                                                                                                      |
```

docs/scaling.md
```diff
@@ -83,6 +83,12 @@ argo --instanceid i1 submit my-wf.yaml
 
 You do not need to have one instance ID per namespace, you could have many or few.
 
+### Maximum Recursion Depth
+
+In order to protect users against infinite recursion, the controller has a default maximum recursion depth of 150 calls to templates.
```
terrytangyuan, line 88:
```suggestion
In order to protect users against infinite recursion, the controller has a default maximum recursion depth of 100 calls to templates.
```

terrytangyuan (approved):

PR #11665, opened 2023-08-24
fix: add guard against NodeStatus. Fixes #11102 

### LINKED ISSUE #11102: Default initialised NodeStatus causes random false failures.

isubasinghe:


terrytangyuan (approved):

agilgur5:
For posterity, as it may not be immediately obvious to a reader, this is a backport of #11451 for the 3.4.11+ patch series

PR #11670, opened 2023-08-25
ci: Improved readability of log of release actions

Yaminyam:
### Motivation

image

When I checked the log of release actions, I was initially puzzled to see that it was running actions/cache three times.
However, I found out that this is to use multiple caches, and I want to improve it so that it can be recognized right away in the log.


terrytangyuan (approved):

agilgur5 (commented):
Nice addition! Thanks for the contribution 🙂

Issue #11661, opened 2023-08-23
Deleting a workflow on the workflow details page doesn't do anything


tico24:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

If you go to the workflows details page and hit delete (and then confirm the delete), you are taken back to the workflows list page... where your workflow is still sat there staring at you :)

We do not have workflows archiving enabled.

### Version

v3.5.0-rc1

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
any workflow
```

### Logs from the workflow controller

```text
I couldn't find anything useful. I'll update if I find anything.
```

### Logs from in your workflow's wait container

```text
n/a
```

terrytangyuan:
cc @toyamagu-2021 This might be related to your fix

terrytangyuan:
If you refresh, would it be gone?

toyamagu-2021:
I cannot reproduce this both on master branch and on 3.5.0-rc1 on k3d cluster... 🤔 

<img width="1034" alt="image" src="https://github.com/argoproj/argo-workflows/assets/83329336/e3ff491f-9946-414e-a90f-e180fdf5d077">
<img width="612" alt="image" src="https://github.com/argoproj/argo-workflows/assets/83329336/c44da420-a6f2-4608-bccc-788060d52bf7">

toyamagu-2021:
Could you give me your workflow metadata.labels?
Might be following line returns false.

https://github.com/argoproj/argo-workflows/blob/7cb22a2aa33b7f5c92d5d87bfa69faed35b3d06a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx#L194-L195

Or please give me your workflow (might be meaningless...).

tico24:
Happens with a basic hello world though:

```
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: hello-world-
  labels:
    workflows.argoproj.io/archive-strategy: "false"
  namespace: ci
  annotations:
    workflows.argoproj.io/description: |
      This is a simple hello world example.
spec:
  serviceAccountName: ci
  entrypoint: whalesay
  templates:
  - name: whalesay
[...]

PR #11676, opened 2023-08-25
fix(ui): Only redirect/reload to wf list page when wf deletion succeeds

terrytangyuan:
Potentially fix and surface the underlying issue in https://github.com/argoproj/argo-workflows/issues/11661.

### DIFF
--- a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
+++ b/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
@@ -197,10 +197,12 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
                                         if (isArchivedWorkflow(workflow) && (globalDeleteArchived || !isWorkflowInCluster(workflow))) {
                                             services.workflows.deleteArchived(workflow.metadata.uid, workflow.metadata.namespace).catch(setError);
                                         }
-                                        navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
-                                        // TODO: This is a temporary workaround so that the list of workflows
-                                        //  is correctly displayed. Workflow list page needs to be more responsive.
-                                        window.location.reload();
+                                        if (error === null) {
+                                            navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
+                                            // TODO: This is a temporary workaround so that the list of workflows
+                                            //  is correctly displayed. Workflow list page needs to be more responsive.
+                                            window.location.reload();
+                                        }
                                     }
                                 });
                         } else if (workflowOperation.title === 'RESUBMIT') {


evgeny-goldin (approved):

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -197,10 +197,12 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
                                         if (isArchivedWorkflow(workflow) && (globalDeleteArchived || !isWorkflowInCluster(workflow))) {
                                             services.workflows.deleteArchived(workflow.metadata.uid, workflow.metadata.namespace).catch(setError);
```
agilgur5, line 198: I'm pretty sure that this line and the equivalent above are asynchronous promises. If they're not waited on to finish, the redirect and reload may happen _before_ the request completes.

It is possible the requests get cancelled as a result of that, would need to see some traces to confirm that though (as it depends).

That _could_ be the root cause here, but not really sure without looking in more detail

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -197,10 +197,12 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
...
                                         }
-                                        navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
-                                        // TODO: This is a temporary workaround so that the list of workflows
-                                        //  is correctly displayed. Workflow list page needs to be more responsive.
-                                        window.location.reload();
+                                        if (error === null) {
+                                            navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
+                                            // TODO: This is a temporary workaround so that the list of workflows
+                                            //  is correctly displayed. Workflow list page needs to be more responsive.
+                                            window.location.reload();
```
agilgur5, line 204: a better workaround would be to reload _only_ the list of workflows, i.e. have it refetch instead of reloading the entire window

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -197,10 +197,12 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
...
+                                        }
```
agilgur5, line 205: this `if` statement could be inverted to remove nesting. same with the `if (yes)` one above too

```suggestion
                                        if (error !== null) {
                                            return
                                        }
                                        navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
                                         // TODO: This is a temporary workaround so that the list of workflows
                                         //  is correctly displayed. Workflow list page needs to be more responsive.
                                         window.location.reload();
```

agilgur5 (commented):
Only 1 in-line comment related to the contents of this change. 

I also see some potential issues in the surrounding code that _could_ be root cause related

rbreeze (approved):

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
agilgur5: Fixed in #11711

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
agilgur5: Refactored in #11711

PR #11684, opened 2023-08-27
feat: Search by name for WorkflowTemplates in UI

### LINKED ISSUE #11004: Search by name for WorkflowTemplates in UI
# Summary

Add search bar in UI to search for WorkflowTemplates, ClusterWorkflowTemplates, CronWorkflows by name.

# Use Cases

Searching by name makes sense if there is a large number of templates in a cluster. Some devs in our team submit WorkflowTemplates manually regularly. It's always a pain to scroll through all the templates.

---

sunyeongchoi:
Fixes #11004

### Motivation

It was implemented because the function to filter the result of the workflow template by name was considered necessary.

If the corresponding PR is merged, the same will be applied to other resource pages.

### Modifications

* It was implemented by adding a namePattern field to search based on name contains.
* Implementation of cursor pagination logic based on resourceVersion to newly paginate the results of filtering on the kubernetes list results.

image

### Verification

* Pass workflow_template_server_test.go
* UI workflow template first page normal
* Normal when the workflow template page limit is 5, 10, or 50
* Test the above with name pattern filtering search
* After searching without a name pattern, search again by adding a name pattern
* After searching including the name pattern, delete the name pattern and search again


server/workflowtemplate/workflow_template_server.go
```diff
@@ -69,15 +71,74 @@ func (wts *WorkflowTemplateServer) getTemplateAndValidate(ctx context.Context, n
...
+		}
+		wfList.Items = items[0:endIndex]
+	} else {
+		wfList.Items = items
+	}
+
+	// Calculate new offset for next batch
+	if limit != 0 && len(wfList.Items) == int(limit) {
+		wfList.ListMeta.Continue = wfList.Items[len(wfList.Items)-1].ResourceVersion
+	}
```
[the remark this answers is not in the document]
sunyeongchoi: @terrytangyuan 
Of course!
We have implemented namePattern filtering.
Since Kubernetes doesn't provide filtering like namePattern, we implemented logic to retrieve results from Kubernetes, then separately filter the results.

In the previous logic, we were utilizing the "continue" option for Kubernetes' built-in pagination.
However, when customizing Kubernetes' results, features like namePattern disrupt the proper usage of "continue" pagination.

Hence, we implemented our own pagination by utilizing Kubernetes' resourceVersion for cursor pagination.

1. Sort in descending order based on resourceVersion.
2. Filter only smaller values compared to the resourceVersion held from the previous page.
3. Index the filtered results based on the limit.
4. Store the resourceVersion for use in the next page.

The pagination logic is implemented in the sequence described above. Thank you.

Also I added a comment about the logic and committed it. https://github.com/argoproj/argo-workflows/pull/11684/commits/e939bb8e81404635d49e76ba3b079692dfe9b282

server/workflowtemplate/workflow_template_server.go
```diff
@@ -69,15 +71,82 @@ func (wts *WorkflowTemplateServer) getTemplateAndValidate(ctx context.Context, n
...
+
+	// Save the original Continue and Limit for custom filtering.
+	resourceVersion := k8sOptions.Continue
+	limit := k8sOptions.Limit
+
+	// Search whole with Limit 0.
+	// Reset the Continue "" to prevent Kubernetes native pagination.
+	// Kubernetes api will search for all results without limit and pagination.
+	k8sOptions.Continue = ""
+	k8sOptions.Limit = 0
```
terrytangyuan, line 88: Would it be possible to stick with the native pagination when search functionality is not used?

sunyeongchoi: I think maybe it is possible. I will try to implement it like that.

sunyeongchoi: I uploaded commit about sticking with native pagination when search functionality is not used. 392222bd097eaa2374de08eb59f9e5feb78c2c61

If this PR is reflected, we plan to implement the same namePattern function on other pages as well. I will consider having `cursorPaginatonByResourceVersion` common in one place and one function.

terrytangyuan (approved):
Thanks! Great work 👍

server/workflowtemplate/workflow_template_server.go
```diff
@@ -67,17 +69,87 @@ func (wts *WorkflowTemplateServer) getTemplateAndValidate(ctx context.Context, n
...
+	// Save the original Continue and Limit for custom filtering.
+	resourceVersion := k8sOptions.Continue
+	limit := k8sOptions.Limit
+
+	if req.NamePattern != "" {
+		// Search whole with Limit 0.
+		// Reset the Continue "" to prevent Kubernetes native pagination.
+		// Kubernetes api will search for all results without limit and pagination.
+		k8sOptions.Continue = ""
+		k8sOptions.Limit = 0
```
agilgur5, line 132: correct me if I'm wrong, but won't this have the same problem as #11761 / i.e. #12025 when there are enough `WorkflowTemplates`? It looks like the exact problem was mentioned in https://github.com/argoproj/argo-workflows/issues/11004#issuecomment-1690298225

agilgur5: The workaround here may be to only retrieve the `limit` at a time from the k8s API. If there aren't enough after the filter, then retrieve more.

That might make the request take longer (potentially significantly, but maybe less than retrieving all at once, which also takes time, and, crucially, memory), but would make sure we never retrieve too many (i.e. it throttles it, effectively). There are some other side-effects to that too (the k8s API rate limit, for instance).

The other way may be to do this fully client-side or partially client-side.
Fully would be the UI (and CLI? a flag wasn't implemented here though) fetches and does a filter, and then if it doesn't have enough, requests more.
Partially would be the API does a filter but can underfetch, in which case the UI has to request more.

agilgur5: I'm also thinking we may want to explicitly point out in the UI which search fields are not k8s native and can therefore be non-performant when used; basically a "hey we implemented this, but you might want to be careful using this".

and at that point, potentially let admins disable them with flags too, if they don't want users to overload the Argo Server or k8s API accidentally.

pkg/apiclient/workflowtemplate/workflow-template.proto
```diff
@@ -24,7 +24,8 @@ message WorkflowTemplateGetRequest {
 
 message WorkflowTemplateListRequest {
   string namespace = 1;
-  k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 2;
+  string namePattern = 2;
+  k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 3;
```
agilgur5, line 28: changing the tag numbers is apparently a breaking change in protobufs, see https://github.com/argoproj/argo-workflows/pull/13160#discussion_r1634092491 for more details.

ui/src/app/workflow-templates/components/workflow-template-filters/workflow-template-filters.tsx
```diff
@@ -51,7 +53,17 @@ export function WorkflowTemplateFilters({templates, namespace, labels, onChange}
                         autocomplete={labelSuggestion}
                         tags={labels}
                         onChange={tags => {
-                            onChange(namespace, tags);
+                            onChange(namespace, namePattern, tags);
+                        }}
+                    />
+                </div>
+                <div className='columns small-2 xlarge-12'>
+                    <p className='wf-filters-container__title'>Name Pattern</p>
```
agilgur5, line 61: "Pattern" is a bit ambiguous as it can mean several things. "Contains" is a more explicit naming. See https://github.com/argoproj/argo-workflows/pull/13160#discussion_r1633477891 for more details

server/workflowtemplate/workflow_template_server.go
agilgur5: For future reference, the current pattern to mitigate this would be to follow the SQLite pattern here a la #12736, see also https://github.com/argoproj/argo-workflows/pull/13633#discussion_r1776448455 et al

PR #11686, opened 2023-08-27
fix: cron workflow initial filter value. Fixes #11685

### LINKED ISSUE #11685: ui: Cron workflow initial filter value
# Summary

What change needs making?

I'm encountering some issues while using the Cron workflow feature.

The first issue is that when I enter the cron workflow, the lists are displayed, but all the filters are removed and set to default values. This can confuse users compared to having all the phases filters selected from the beginning. Therefore, the state of the cron workflow should start with all selections made.
[before]
스크린샷 2023-08-27 오전 2 14 06
[after]
스크린샷 2023-08-27 오전 2 14 12

The second issue is related to modifying the cron in the cron workflow. The style of the Concurrency Policy is broken. The position and height of the Select box and Select arrow need to be adjusted.
[before]
스크린샷 2023-08-27 오전 2 06 32
[after]
스크린샷 2023-08-27 오전 2 06 07

# Use Cases

When would you use this?

---

zel0rd:
Fixes #11685

### Modifications

Encountering some issues while using the Cron workflow feature.

### Modifications

1. When I enter the cron workflow, the lists are displayed, but all the filters are removed and set to default values. This can confuse users compared to having all the phases filters selected from the beginning. Therefore, the state of the cron workflow should start with all selections made.스크린샷 2023-08-27 오전 2 14 12

2. The second issue is related to modifying the cron in the cron workflow. The style of the Concurrency Policy is broken. The position and height of the Select box and Select arrow need to be adjusted.스크린샷 2023-08-27 오전 2 06 07

### DIFF
--- a/ui/src/app/shared/components/checkbox-filter/checkbox-filter.tsx
+++ b/ui/src/app/shared/components/checkbox-filter/checkbox-filter.tsx
@@ -13,6 +13,7 @@ interface Props {
 export class CheckboxFilter extends React.Component<Props> {
     constructor(props: any) {
         super(props);
+        this.props.items.forEach(item => this.props.selected.push(item.name));
     }
 
     public render() {
--- a/ui/src/app/cron-workflows/components/cron-workflow-spec-editior.tsx
+++ b/ui/src/app/cron-workflows/components/cron-workflow-spec-editior.tsx
@@ -24,8 +24,9 @@ export const CronWorkflowSpecEditor = ({onChange, spec}: {spec: CronWorkflowSpec
                 </div>
                 <div className='row white-box__details-row'>
                     <div className='columns small-3'>Concurrency Policy</div>
-                    <div className='columns small-9'>
+                    <div className='columns small-9' style={{lineHeight: '30px'}}>
                         <Select
+                            placeholder='Select concurrency policy'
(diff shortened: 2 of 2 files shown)


terrytangyuan (approved):

ui/src/app/shared/components/checkbox-filter/checkbox-filter.tsx
```diff
@@ -13,6 +13,7 @@ interface Props {
 export class CheckboxFilter extends React.Component<Props> {
     constructor(props: any) {
         super(props);
+        this.props.items.forEach(item => this.props.selected.push(item.name));
```
agilgur5, line 16: mmm this is a bit of an anti-pattern, especially for a reusable component like this one (it's used in more than one place).

It would be better to just set the initial `selected` props from the parent component. That ends up going two parents up with the initial state [here](https://github.com/argoproj/argo-workflows/blob/9cb378342283c9ef9f2f3b999bec7cf10c8aab91/ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx#L38)

agilgur5: Fixed in #11739

PR #11690, opened 2023-08-28
docs: Add notes on using pprof for development

terrytangyuan:
This is asked very often. Adding these notes to help new contributors.

### DIFF
--- a/docs/running-locally.md
+++ b/docs/running-locally.md
@@ -227,8 +227,8 @@ git commit --signoff -m 'feat: Added a new feature. Fixes #1234'
   have checked out your code into `$GOPATH/src/github.com/argoproj/argo-workflows`.
 * If you encounter "out of heap" issues when building UI through Docker, please validate resources allocated to Docker.
   Compilation may fail if allocated RAM is less than 4Gi.
-* To start profiling with [pprof](https://go.dev/blog/pprof), pass `ARGO_PPROF=true` when starting the controller locally,
-  and run the following:
+* To start profiling with [`pprof`](https://go.dev/blog/pprof), pass `ARGO_PPROF=true` when starting the controller locally.
+  Then run the following:
 
 ```bash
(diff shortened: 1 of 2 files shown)


docs/running-locally.md
```diff
@@ -227,6 +227,14 @@ git commit --signoff -m 'feat: Added a new feature. Fixes #1234'
   have checked out your code into `$GOPATH/src/github.com/argoproj/argo-workflows`.
 * If you encounter "out of heap" issues when building UI through Docker, please validate resources allocated to Docker.
   Compilation may fail if allocated RAM is less than 4Gi.
+* To start profiling with [pprof](https://go.dev/blog/pprof), pass `ARGO_PPROF=true` when starting the controller locally,
+  and run the following:
```
agilgur5, line 231:
```suggestion
* To start profiling with [`pprof`](https://go.dev/blog/pprof), pass `ARGO_PPROF=true` when starting the controller locally.
  Then run the following:
```

I believe you won't need the `.spelling` change if it is in backticks. Also fix run-on sentence

juliev0 (commented):
Thanks for adding documentation. I didn't know about this.

I see [this](https://github.com/argoproj/argo-workflows/blob/master/util/pprof/pprof.go#L14) code for where we're serving the http endpoint. Do you know what causes port 6060 to be used?

terrytangyuan:
> Thanks for adding documentation. I didn't know about this.
> 
> I see [this](https://github.com/argoproj/argo-workflows/blob/master/util/pprof/pprof.go#L14) code for where we're serving the http endpoint. Do you know what causes port 6060 to be used?

See the bottom of this page https://go.dev/blog/pprof. Basically that import statement will install the handlers.

image

And then we invoke the `Init()` method in controller https://github.com/argoproj/argo-workflows/blob/master/cmd/workflow-controller/main.go#L79 and executor https://github.com/argoproj/argo-workflows/blob/master/workflow/executor/executor.go#L1177

juliev0 (approved):

PR #11693, opened 2023-08-28
fix: Make defaultWorkflow hooks work more than once

Joibel:
As workflowDefaults are passed by pointer we're modifying them directly here before modifying them such that StrategicMerge can work.

Undo the modification we make directly so that the next time we patch it's still there.

Fixes a bug introduced in the fix in #11214 

### Verification

Run with workflowDefaults having hooks more than once.


isubasinghe (approved):
LGTM

terrytangyuan (approved):

PR #11703, opened 2023-08-29
fix: offset reset when pagination limit onchange

### LINKED ISSUE #11702: A problem that is not initialized when changing the number of page limits
### Pre-requisites

There appeared to be an error with the pagination limit. The issue was observed when performing the following steps:

1. Set the limit to 5 on the workflows page or workflow templates page from the frontend. and Click "Next Page."
image

2. While on the next page, change the limit to 10. Although there should be enough items to display 10, they do not appear.
image
image

### Version

master

```YAML
This is a front matter.
```

```text
```

```text
```

sunyeongchoi:
Fixes #11702

### Motivation

Upon investigating the cause, it was found that when changing the pagination limit in the onChange function, the offset used as the pagination reference was not reset. Instead, the offset value received from the server was directly used.

As a solution, I have modified the onChange function to reset the offset during each change. This change ensures that when the pagination limit is modified, the search restarts from the beginning with the updated limit.

### Modifications

I just set offset '' in page limit onChange function.

image

1. Set the limit to 5 on the workflows page or workflow templates page from the frontend. and Click "Next Page."
image

2. While on the next page, change the limit to 10. they  appear.
image
image

### DIFF
--- a/ui/src/app/shared/components/pagination-panel.tsx
+++ b/ui/src/app/shared/components/pagination-panel.tsx
@@ -40,8 +40,9 @@ export class PaginationPanel extends React.Component<{pagination: Pagination; on
                             // Only return the offset if we're actually going to be limiting
                             // the results we're requesting.  If we're requesting all records,
                             // we should not skip any by setting an offset.
+                            // The offset must be initialized whenever the pagination limit is changed.
                             if (limit) {
-                                newValue.offset = this.props.pagination.offset;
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
Can you check if the pagination works correctly for other pages (e.g. workflows list) as well?

terrytangyuan (approved):

Issue #11608, opened 2023-08-18
Use `expr` builtin function and pipes


antonmedv:
> IMO, `sprig.date()` should not silently fall back to `now()` when you pass an invalid type but return an error instead, but that's another story.

_Originally posted by @Gerrit-K in https://github.com/argoproj/argo-workflows/issues/11488#issuecomment-1666485905_

agilgur5:
Thanks for notifying us!

It looks like we got a PR to upgrade `expr` to 1.14 in #11605, just a couple hours before this issue 😅 So I think this is complete now.

The [expression docs](https://argoproj.github.io/argo-workflows/variables/#expression) link directly to https://github.com/antonmedv/expr/blob/master/docs/Language-Definition.md.

We could potentially update [the examples](https://argoproj.github.io/argo-workflows/variables/#examples) a bit, but it's just a short list.

> I recently added a bunch of functions in `expr` lib which should cover most of the `sprig` libs.

This is great to know! We do have a pretty big warning in that same part of the docs about Sprig errors. We could potentially add to that warning that newer versions of `expr` can replace Sprig and forego those warnings.

We would need to note that these require Argo v3.5+ (https://github.com/argoproj/argo-workflows/issues/11381) to use

Do you have any specific suggestions for the docs? Would you like to submit a PR to update them?

antonmedv:
> Do you have any specific suggestions for the docs? Would you like to submit a PR to update them?

Yes. I can update the docs. Will send a PR.  

What do you think about trying a pipe operator? It is experimental right now, but still? 

Another thing: lets add some env types and Argo examples to expr tests. To catch errors early on expr side as well? 
For example, here is expressions for Crowdsec https://github.com/antonmedv/expr/blob/master/testdata/crowdsec.json

caelan-io:
+1 we would greatly appreciate adding more documentation about this. Expressions and Sprig are common Q&As in our Slack community.

PR #11704, opened 2023-08-29
chore: update pkg dep

b-erdem:
### Motivation

Argo was using an outdated JSONPath library and it was only supporting simple expressions. This was causing issues or workarounds when you need a bit more complex JSONPath expressions.

### Modifications

The current JSONPath library replaced with something better that supports almost all JSONPath spec. See here: https://github.com/argoproj/pkg/pull/486

### Verification

I've added additional test cases in the first PR. If you think we should add some additional tests here as well, please let me know. I'll update accordingly.

### Caveats

When working on this, I noticed just updating the dependency is not enough to be able to work with more complex JSONPath expressions. In Argo, it's assumed that simple values(like string, number, etc.) will be extracted from JSON data using JSONPath expressions. When you use filtering expressions, it'll return list or object. Current implementation does not support that. For instance, if your expression returns `["bar", "baz", "foo"]`, in YAML file it'll be `"["bar", "baz", "foo"]"` which is in correct(quotes should be escaped). I fixed this issue and existing tests are passing but I thought it'll take a bit more time to discuss and to not blocking updating `pkg` dependency, I'm just opening this PR with only dependency updates. After this, I'll open a PR to fix `expressionReplace` function to work properly when the value is list or object.


b-erdem:
@agilgur5 Hi! Do you want me to open the next PR now or should I wait to get this one merged?

agilgur5:
Sorry I hadn't had a chance to take a look in depth yet. Been a bit under the weather this week 😕

b-erdem:
Got it. No problem. Just wanted to make sure if there’s anything else i need to do. Thanks!

rohankmr414:
@b-erdem can you update the https://github.com/argoproj/pkg package to the latest commit, It should fix the issue  https://github.com/argoproj/argo-workflows/issues/11723 with the minio changes

agilgur5:
> I've added additional test cases in the first PR. I tested it with some example YAML files. If you think we should add some additional tests here as well, please let me know. I'll update accordingly.

Since you added tests in the upstream `argoproj/pkg` PR, I'm fine either way. If there's a specific Workflows scope use-case test you want to add, that may be good to have.
We may also add tests upstream to `expr` itself -- see https://github.com/argoproj/argo-workflows/issues/11608#issuecomment-1685250948.

> For instance, if your expression returns `["bar", "baz", "foo"]`, in YAML file it'll be `"["bar", "baz", "foo"]"` which is in correct(quotes should be escaped. note: just to clarify this is existing functionality and it's not something introduced with this change).

Good note! Yea the previous library also supported slices and filters, so it would still have that issue.
It is pretty rare that you'd want to return a list or object from an expression though, so maybe that's why this hasn't been reported as a bug before 🤔 

> After this, I'll open a PR to fix `expressionReplace` function to work properly when the value is list or object.

Sounds good to me! That does indeed sound like a separate bugfix that was just incidentally found as part of this.

agilgur5 (approved):
LGTM. Per above, may be good to further upgrade as well. I'd approve either.

b-erdem:
@rohankmr414 I updated `pkg` to the latest commit. @agilgur5 Thanks for your thought. I'll open another PR soon to fix `expressionReplace`.

agilgur5 (approved):

rohankmr414 (approved):

sarabala1979 (approved):

b-erdem:
@sarabala1979 Hi. Thanks for approving the PR! It seems like a flaky test is failing. Can you please re-run the job so I can merge it to master? Thanks!

rohankmr414:
> @sarabala1979 Hi. Thanks for approving the PR! It seems like a flaky test is failing. Can you please re-run the job so I can merge it to master? Thanks!

@b-erdem you can also push an empty commit to retrigger the tests

agilgur5:
> @b-erdem you can also push an empty commit to retrigger the tests

This retriggers all of CI though (which can result in more flakes), whereas an admin can retry a single failing job (which has a lower chance of flakes) 😕

b-erdem:
Yeah, it takes too long as well. :D So it's better if someone who can rerun it.(I just pushed an empty commit for now).

b-erdem:
@agilgur5 @rohankmr414 @sarabala1979 @terrytangyuan Hi, can someone merge this to master? It seems I can't do it.

PR #11711, opened 2023-08-29
fix(ui): don't reload the page until _after_ deletion

### LINKED ISSUE #11661: Deleting a workflow on the workflow details page doesn't do anything

agilgur5:
Hopefully fixes #11661 

### Motivation

- Per my comments in https://github.com/argoproj/argo-workflows/pull/11676#discussion_r1306119462, workflow deletions happen asynchronously and were not waited on
  - this meant that the error check could occur too early
  - and that the page reload might occur too early, before the requests finished
    - in particular, this scenario could happen on a higher latency connection (e.g. on a remote server, not running locally)
      - I and others were not able to reproduce locally, but it was reliably reproduced in the issue (e.g. https://github.com/argoproj/argo-workflows/issues/11661#issuecomment-1691839976)
    - if this were to happen, the UI would be dropping its connection and reconnecting during the reload
      - if the connection were dropped while the request were in-flight, the server could drop the request entirely
        - based on Server logs in https://github.com/argoproj/argo-workflows/issues/11661#issuecomment-1697752783, a 408 status code suggested that this might in fact be the case
        - if the request were dropped before the response, it is possible that the server may not have completed deletion
          - e.g. may have not yet sent a request to the k8s API server, or drops the request to the k8s API server, etc

### Modifications

- modify the code to actually wait on the requests
  - use `Promise.all` to run the delete and archive delete simultaneously
- partially rewrite in newer async/await syntax as it is considerably cleaner that way
  - otherwise, the scoping/conditionals get pretty confusing and require some redundant code
  - ideally, we should more generally rewrite Promise-based code to async/await style syntax, which is generally cleaner & preferred by the community over Promises or callback hell

- refactor the conditionals to have less nesting and early return instead
  - per review comments I made in https://github.com/argoproj/argo-workflows/pull/11676#discussion_r1306123032
  - there's several places where we can rewrite to have less nesting. async/await will also help with that on the UI

### Verification

Build passes. UI still works.
I was unable to reproduce the original problem, so as a result I can't verify that this fixes the issue completely, but the evidence made my hypothesis seem likely.

### Future Work

Per above, refactor to have less nesting and use async/await

### DIFF
--- a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
+++ b/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
@@ -189,21 +189,27 @@ export const WorkflowDetails = ({history, location, match}: RouteComponentProps<
                         if (workflowOperation.title === 'DELETE') {
                             popup
                                 .confirm('Confirm', () => <DeleteCheck isWfInDB={isArchivedWorkflow(workflow)} isWfInCluster={isWorkflowInCluster(workflow)} />)
-                                .then(yes => {
-                                    if (yes) {
-                                        if (isWorkflowInCluster(workflow)) {
-                                            services.workflows.delete(workflow.metadata.name, workflow.metadata.namespace).catch(setError);
-                                        }
-                                        if (isArchivedWorkflow(workflow) && (globalDeleteArchived || !isWorkflowInCluster(workflow))) {
-                                            services.workflows.deleteArchived(workflow.metadata.uid, workflow.metadata.namespace).catch(setError);
-                                        }
-                                        if (error === null) {
-                                            navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
-                                            // TODO: This is a temporary workaround so that the list of workflows
-                                            //  is correctly displayed. Workflow list page needs to be more responsive.
-                                            window.location.reload();
-                                        }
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
@tico24 Could you test the changes to see if that fixes your issue?

terrytangyuan:
cc @rbreeze to help review

rbreeze (approved):
LGTM pending testing

agilgur5:
This is a fix / good change even if it does not fully resolve #11661, so I would say it is safe to merge in regardless.

terrytangyuan (approved):

tico24:
This is on my todo list to test, I just haven't had the time today.

tico24:
This does fix issue #11661, thank you.

agilgur5:
> ### Future Work
> Per above, refactor to have less nesting and use async/await

Created #11740 to track refactor to async/await

PR #11713, opened 2023-08-30
docs: link to Installation instead of `manifests/README.md`

agilgur5:
### Motivation

- `manifests/README.md` just has a one-line link to the Installation docs
  - instead of redirecting a contributor twice, just redirect once

- also follows existing pattern of the docs using relative links to other pages

Stumbled upon this for the first time recently during PR reviews for the Developer tab of the docs

### Modifications

Modify `public-api.md` to link to `installation.md` directly
- instead of `manifests/README.md` which [itself then links](https://github.com/argoproj/argo-workflows/blob/f00b6ff631bae954f749f52d2d94c5dbc6a0df53/manifests/README.md?plain=1#L3) to `installation.md`

### Verification

`make docs` passes

### DIFF
--- a/docs/public-api.md
+++ b/docs/public-api.md
@@ -4,4 +4,4 @@ Argo Workflows public API is defined by the following:
 
 * The file `api/openapi-spec/swagger.json`
 * The schema of the table `argo_archived_workflows`.
-* The installation options listed in `manifests/README.md`.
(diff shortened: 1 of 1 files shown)


caelan-io (approved):
good catch!

terrytangyuan (approved):

PR #11714, opened 2023-08-30
docs(auth): clarify "Bearer token" terminology

agilgur5:
Follow-up to #11300, where my wording could have been clearer

### Motivation

Per https://github.com/argoproj/argo-workflows/discussions/11680 and some others, this is still a point of confusion with users

### Modifications

- be specific when referring to `$ARGO_TOKEN`, which includes "Bearer ", and the SA Secret, which is the part after "Bearer "
- also fix the one doc that has `Bearer $ARGO_TOKEN` whereas the rest of the codebase has `$ARGO_TOKEN` with "Bearer " part of it

### Verification

`make docs` passes

### Related Notes

- **Note**: the _internal_ terminology is actually a bit off here. In the [IETF standards spec](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1), the "token" is referred to as the part after "Bearer "
  - The spec does not quite specify this in its [terminology section](https://datatracker.ietf.org/doc/html/rfc6750#section-1.2), however, which means this is not as clear as it could be in the industry
  - See also another common industry reference's use of the terminology: https://swagger.io/docs/specification/authentication/bearer-authentication/
  - Internally though (which is more than just docs, it includes source code and tests as well), `$ARGO_TOKEN` _always_ includes "Bearer "
    - so this change makes everything _internally_ consistent. changing all internals would be very complex at this time


agilgur5:
`Bearer ` was removed in #12014, which was reviewed and merged before this PR, despite being made ~1.5 months after it... 😕
That PR did not have any description, but this one does describe in detail that `Bearer ` is currently included in the `$ARGO_TOKEN` variable in the codebase, but probably shouldn't be according to industry terminology.

I merged `master` as a result, but this description and remaining changes are still relevant

terrytangyuan (approved):

jmeridth (approved):

PR #11716, opened 2023-08-30
chore: remove ui source map for production build

sonbui00:
### Motivation

- Reduce release package size, container size (~20MB)

### Modifications

- Disable source map for production build

### Opinion
- The source map is unnecessary to add to the release package. If we want to have the source map to debug, we can run a dev ui and connect to API

### DIFF
--- a/ui/src/app/webpack.config.js
+++ b/ui/src/app/webpack.config.js
@@ -24,7 +24,7 @@ const config = {
     path: __dirname + "/../../dist/app"
   },
 
-  devtool: "source-map",
+  devtool: isProd ? false : "source-map",
 
   resolve: {
     extensions: [".ts", ".tsx", ".js", ".json", ".ttf"],
--- a/ui/src/app/webpack.config.js
+++ b/ui/src/app/webpack.config.js
@@ -24,7 +24,7 @@ const config = {
     path: __dirname + "/../../dist/app"
   },
 
-  devtool: isProd ? false : "source-map",
+  devtool: isProd ? false : "eval",
 
   resolve: {
     extensions: [".ts", ".tsx", ".js", ".json", ".ttf"],


agilgur5:
> * The source map is unnecessary to add to the release package

Mmm I don't agree with this. There is a reason production source maps exist. Source maps can be very useful to consumers, especially of an open-source tool. Minified code is not really readable without them (not to mention that the source is TS, not ES5 JS).

You can certainly argue the pros & cons of production sourcemaps, and one can certainly decide not to provide them, but they are not useless

ui/src/app/webpack.config.js
```diff
@@ -24,7 +24,7 @@ const config = {
     path: __dirname + "/../../dist/app"
   },
 
-  devtool: "source-map",
+  devtool: isProd ? false : "source-map",
```
agilgur5, line 27:
```suggestion
  devtool: isProd ? false : "eval",
```

If we only have dev source maps, there are much faster options. `eval` is [recommended](https://webpack.js.org/configuration/devtool/) for dev

agilgur5:
Also, FWIW, the UI build _is_ fairly large and there are likely ways to decrease its size. I haven't examined it in detail yet though (e.g. via bundle analysis and other methods)

ui/src/app/webpack.config.js
sonbui00: Great. I also think eval is better for development

sonbui00:
@agilgur5 Should I do anything to go forward for this?

agilgur5:
Similar process as I mentioned in your other PR. But here we'd want some more folks to weigh in since I don't necessarily favor this change. The change itself is otherwise simple, so not much to review.

cc @isubasinghe do you have any opinion on this change?

isubasinghe:
@agilgur5 I don't really have an opinion on this, I suspect users might not particularly care about the source map being absent in the release build. 

Then again, 20Mb is relatively cheap, so might as well leave it in?

terrytangyuan (commented):
What do other projects do?

sonbui00:
> What do other projects do?

You can see here: https://github.com/search?q=%22devtool%3A+isProd%22&type=code

agilgur5:
> You can see here: https://github.com/search?q=%22devtool%3A+isProd%22&type=code

A more generic search: https://github.com/search?q=%22devtool%3A%22+%28language%3ATypeScript+OR+language%3AJavaScript+%29&type=code (as `isProd` is not always used)

> What do other projects do?

For open-source _libraries_, it is generally best practice to have source maps (here's a relevant issue: https://github.com/jaredpalmer/tsdx/issues/490. I have also maintained TS library tooling for years now).
The Argo Server's UI is not a library, however, but it is a "developer tool" of sorts. More specifically, if users have an issue, they may be capable of debugging it themselves, and as an OSS tool, why not let them? It makes self-fixing a bug report easier too.

The "official" suggestion from [Webpack](https://webpack.js.org/configuration/devtool/) et al (see also this popular [SO answer](https://stackoverflow.com/a/46570345/3431180) and this popular [blog post](https://css-tricks.com/should-i-use-source-maps-in-production/)) has always been nuanced -- it depends on what your goal is.
Better debuggability in Prod? High quality source maps. Fast Prod build speed? Lower quality source maps or none. Obscurity of source code? Hidden source maps or none. Smaller build size? None.

So there is no single convention as such. Personally, I tend to lean more pro-source map unless there's a convincing reason to not include them. It's a general debugging best practice (that I recommend to library authors) and, as someone who root causes a lot of issues, can be _very_ helpful personally. As I mentioned above, we haven't gotten complaints about it, so I don't see much reason to change it.

To be fair though, I didn't expect the source maps to be 20MB. It's almost all in the deps I imagine though (which there are almost certainly optimizations for, like [DLLs](https://webpack.js.org/plugins/dll-plugin/)). It isn't a lot in today's day and age, but it isn't a little either.

Issue #10566, opened 2023-02-22
Add documentation about `--qps` and `--burst` settings of controller


sgutwein:
# Summary

The [vertically-scaling](https://argoproj.github.io/argo-workflows/scaling/#vertically-scaling) section in the argo-workflows documentation currently only recommends to adapt the `--qps` and `--burst` setting of workflows controller. I couldn't find any (other) documentation about these settings, neither what the default values are.

# Related issues

https://github.com/argoproj/argo-workflows/issues/4634
https://github.com/argoproj/argo-workflows/issues/8576

# Use Cases

Setup the scaling of argo-workflows.

---

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

caelan-io:
@juliev0 - this become a top voted issue and it's a docs issue. Do you have any notes from your recent work related to this internally that we could use as a first draft for a docs PR to close this? Thanks!

juliev0:
Oh wow, I didn't know this issue existed. I'd be happy to do this in the near future.

Issue #8576, opened 2022-05-02
Looks like the `burst` configuration for the Workflow Controller doesn't actually do anything


Radolumbo:
Was hunting down what `burst` actually does for https://cloud-native.slack.com/archives/C01QW9QSSSK/p1651410938829379

Looking for `burst` in https://github.com/argoproj/argo-workflows takes us to https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/config.go#L85

Looks like it's using a built in golang Library, time/rate .. so we can take a look at https://pkg.go.dev/golang.org/x/time/rate#Limiter.Burst

This tells us
>Burst returns the maximum burst size. Burst is the maximum number of tokens that can be consumed in a single call to Allow, Reserve, or Wait, so higher Burst values allow more events to happen at once. A zero Burst allows no events, unless limit == Inf.

Interesting. Where are we using any of those calls in the Workflow controller on this `rateLimiter` object from the first link? Looks like .. here

https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/workflowpod.go#L429

So, we're using `Allow`. Based on

https://pkg.go.dev/golang.org/x/time/rate#Limiter.Allow

`Allow` is shorthand for `AllowN(time.Now(), 1)`.

So wait -- `burst` tells us how many tokens for rate limiting can be requested in a single call to `Allow`. However, the controller only ever requests a single token at time.

Thus -- I actually think that the configured `burst` for the Workflow Controller doesn't do anything at all.

Let me know if it seems like I missed anything here, but I'm pretty sure `burst` does nothing for the Workflow controller.

alexec:
This bug is correct. Yet, I don't think we can change the code to use `AllowN`, `createWorkflowPod` does not know how many pods will be created.

Radolumbo:
Sure, I totally agree -- not suggesting we try to add that functionality, but rather we should probably clear out code/guidance that is confusing to users. Maybe I can take a crack at some point soon.

thor:
Coming across this by chance, I suggest re-opening this issue and marking it as frozen.

CiprianAnton:
Seems like --qps and --burst are used in controller main: https://github.com/argoproj/argo-workflows/blob/master/cmd/workflow-controller/main.go#L86.

Can it have an impact on calls to k8s api?

PR #11731, opened 2023-09-03
docs: Add some more info to the Scaling Doc

juliev0:
Fixes: https://github.com/argoproj/argo-workflows/issues/10566

### Motivation

Want to provide more information to those who are trying to scale Argo Workflows.

### DIFF
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -44,7 +44,7 @@ or for >= v3.5: a warning like this (could be any CR, not just `WorkflowTemplate
 
 Then, if your K8S API Server can handle more requests:
 
-- Increase both `--qps` and `--burst`. The `qps` value indicates the average number of queries per second allowed by the K8S Client. The `--burst` value is the number of queries/sec the Client receives before it starts enforcing `qps`, so typically `--burst` > `qps`.
+- Increase both `--qps` and `--burst`. The `qps` value indicates the average number of queries per second allowed by the K8S Client. The `burst` value is the number of queries/sec the Client receives before it starts enforcing `qps`, so typically `burst` > `qps`.
 
 ## Sharding
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -32,7 +32,7 @@ If you have sufficient CPU cores, you can take advantage of them with more gorou
 
 ### K8S API Client Side Rate Limiting
 
-The K8S client library rate limits the messages that can go out. The default values are fairly low.
+The K8S client library rate limits the messages that can go out. 
 
 If you frequently see messages similar to this in the Controller log (issued by the library):
 
@@ -48,7 +48,7 @@ Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1a
 
 Then, if your K8S API Server can handle more requests:
 
-- Increase both `--qps` and `--burst`. The `qps` value indicates the average number of queries per second allowed by the K8S Client. The `burst` value is the number of queries/sec the Client receives before it starts enforcing `qps`, so typically `burst` > `qps`.
+- Increase both `--qps` and `--burst` arguments for the Controller. The `qps` value indicates the average number of queries per second allowed by the K8S Client. The `burst` value is the number of queries/sec the Client receives before it starts enforcing `qps`, so typically `burst` > `qps`.  If not set, the default values are `qps=20` and `burst=30` (as of v3.5 (refer to `cmd/workflow-controller/main.go` in case the values change)).
 
 ## Sharding
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -19,6 +19,8 @@ If you observe the Controller using its total allocated CPU or memory, you shoul
 
 ### Adding Goroutines to Increase Concurrency
 
+If you have sufficient CPU you can take advantage of it with more goroutines:
+
 - If you have many Workflows and you notice they're not being reconciled fast enough, increase `--workflow-workers`.
 - If you're using TTLStrategy in your Workflows and you notice they're not being deleted fast enough, increase `--workflow-ttl-workers`.
 - If you're using PodGC in your Workflows and you notice the Pods aren't being deleted fast enough, increase `--pod-cleanup-workers`.
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -16,7 +16,7 @@ You can scale the controller vertically in these ways:
 
 ### Container Resource Requests
 
-If you observe the Controller using its total request CPU or memory, you should increase those.
+If you observe the Controller using its total CPU or memory requests, you should increase those.
 
 ### Adding Goroutines to Increase Concurrency
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -42,7 +42,7 @@ Waited for 7.090296384s due to client-side throttling, not priority and fairness
 
 or for >= v3.5: a warning like this (could be any CR, not just `WorkflowTemplate`):
 
-```
+```txt
 Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t
 ```
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -12,7 +12,7 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
 
 ## Vertically Scaling
 
-You can scale the controller vertically:
+You can scale the controller vertically in these ways:
 
 ### Adding Goroutines to Increase Concurrency
 
@@ -24,11 +24,11 @@ You can scale the controller vertically:
 
 ### K8S API Client Side Rate Limiting
 
-The K8S client library rate limits the messages that can go out. The default values are fairly low. If you frequently see a message similar to this (issued by the library):
+The K8S client library rate limits the messages that can go out. The default values are fairly low. If you frequently see a message similar to this in the Controller log (issued by the library):
 
 `Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
 
-in the Controller log, or for >= v3.5: a warning like this (could be any CR, not just WorkflowTemplate):
+or for >= v3.5: a warning like this (could be any CR, not just WorkflowTemplate):
 
 `Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -13,6 +13,7 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
 ## Vertically Scaling
 
 You can scale the controller vertically in these ways:
+
 ### Container Resource Requests
 
 If you observe the Controller using its total allocated CPU or memory, you should increase those.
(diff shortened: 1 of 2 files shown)


docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+`Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
+
+or for >= v3.5: a warning like this (could be any CR, not just `WorkflowTemplate`):
+
+`Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
+
+then assuming your K8S API Server can handle it:
+
+- Increase both `--qps` and `--burst`. The `qps` value indicates the average number of queries per second allowed by the K8S Client. The `--burst` value is the number of queries/sec the Client receives before it starts enforcing `qps`, so typically `--burst` > `qps`.
```
terrytangyuan, line 45: Burst may not work as expected. See https://github.com/argoproj/argo-workflows/issues/8576

juliev0: So, that's actually a different "Burst". That's for the [RateLimiter](https://github.com/argoproj/argo-workflows/blob/master/config/config.go#L66) configured in the workflow-controller-configmap controlling the Pod creation rate.

This setting just gets passed directly into the Kubernetes Client to control the rate of outgoing K8S API requests.

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+### Adding Goroutines to Increase Concurrency
+
+If you have sufficient CPU you can take advantage of it with more goroutines:
+
+- If you have many Workflows and you notice they're not being reconciled fast enough, increase `--workflow-workers`.
+- If you're using `TTLStrategy` in your Workflows and you notice they're not being deleted fast enough, increase `--workflow-ttl-workers`.
+- If you're using `PodGC` in your Workflows and you notice the Pods aren't being deleted fast enough, increase `--pod-cleanup-workers`.
+
+>= v3.5
```
agilgur5, line 29:
```suggestion
> v3.5 and after
```

Blockquote is the stylistic convention used in most of the docs ([random example](https://github.com/argoproj/argo-workflows/blob/8e01696a9c63f25b3660ed479c6f7a57bb61d82d/docs/rest-api.md?plain=1#L5))

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
 
-You can scale the controller vertically:
+You can scale the controller vertically in these ways:
 
-- If you have many workflows, increase `--workflow-workers` and `--workflow-ttl-workers`.
-- Increase both `--qps` and `--burst`.
+### Container Resource Requests
 
-You will need to increase the controller's memory and CPU.
+If you observe the Controller using its total request CPU or memory, you should increase those.
```
agilgur5, line 19:
```suggestion
If you observe the Controller using its total CPU or memory requests, you should increase those.
```

"resource requests" is the [usual](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container) order, so move "requests" to after CPU/mem

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+### Adding Goroutines to Increase Concurrency
+
+If you have sufficient CPU you can take advantage of it with more goroutines:
```
agilgur5, line 23:
```suggestion
If you have sufficient CPU cores, you can take advantage of them with more goroutines:
```
- fix comma splice
- perhaps be a bit more specific by specifying "cores" as in "one thread per core". could be convinced otherwise on that though since k8s language more or less refers to "CPU" as number of cores

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+>= v3.5
+
+- If you're using a lot of `CronWorkflows` and they don't seem to be firing on time, increase `--cron-workflow-workers`.
+
+### K8S API Client Side Rate Limiting
+
+The K8S client library rate limits the messages that can go out. The default values are fairly low. If you frequently see a message similar to this in the Controller log (issued by the library):
+
+`Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
```
agilgur5, line 37:
```suggestion
\```
Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t
\```
```

code block instead of plain backticks.

GH can't properly escape code blocks in code blocks, so the suggestion is a bit wonky 😕

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+or for >= v3.5: a warning like this (could be any CR, not just `WorkflowTemplate`):
+
+`Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
```
agilgur5, line 41:
```suggestion
\```
Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t
\```
```

same as above, code block instead of plain backticks

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+- If you're using `TTLStrategy` in your Workflows and you notice they're not being deleted fast enough, increase `--workflow-ttl-workers`.
+- If you're using `PodGC` in your Workflows and you notice the Pods aren't being deleted fast enough, increase `--pod-cleanup-workers`.
+
+>= v3.5
+
+- If you're using a lot of `CronWorkflows` and they don't seem to be firing on time, increase `--cron-workflow-workers`.
+
+### K8S API Client Side Rate Limiting
+
+The K8S client library rate limits the messages that can go out. The default values are fairly low. If you frequently see a message similar to this in the Controller log (issued by the library):
```
agilgur5, line 35:
```suggestion
The K8S client library rate limits the messages that can go out. The default values are fairly low.

If you frequently see messages similar to this in the Controller log (issued by the library):
```

- split the exposition with the action
- "frequently see messages"

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+`Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
+
+or for >= v3.5: a warning like this (could be any CR, not just `WorkflowTemplate`):
```
agilgur5, line 39:
```suggestion
Or, in >= v3.5, a message similar to this (could be any CR, not just `WorkflowTemplate`):
```

- comma splices
- use same style & wording as the above for parallelism (as in grammatical parallelism, not the CPU kind 😅)

docs/scaling.md
```diff
@@ -12,12 +12,37 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+`Waited for 7.090296384s, request:GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t`
+
+then assuming your K8S API Server can handle it:
```
agilgur5, line 43:
```suggestion
Then, if your K8S API Server can handle more requests:
```

docs/scaling.md
agilgur5, line 45:
```suggestion
- Increase both `--qps` and `--burst`. The `qps` value indicates the average number of queries per second allowed by the K8S Client. The `burst` value is the number of queries/sec the Client receives before it starts enforcing `qps`, so typically `burst` > `qps`.
```

- Consistently use `--` or not when describing the values.

agilgur5: Is there a good piece of external documentation you might recommend here? Could add a sentence at the end as "See [x]() for more details".

the k8s docs themselves [don't seem to](https://kubernetes.io/search/?q=qps) document this very explicitly. the [godoc for the k8s client](https://pkg.go.dev/k8s.io/client-go/rest#Config) doesn't either 😕

agilgur5 (commented):
Great additions! Tuning has a good amount of nuance to it, so I think guidance here could be very helpful!

Mostly made copy-edits below

agilgur5:
We may want to list the defaults for these flags here as well so that there is a comparison point for when they are not explicitly set

docs/scaling.md
juliev0: I'd like to keep the word "warning" in there. I think the fact that it's a message at "warning" level will indicate to people that they can filter the logs to just show warning and error, and they will see this.

docs/scaling.md
juliev0: unfortunately, I don't really have a good link either

docs/scaling.md
agilgur5: Is the pre-3.5 version also a warning? Can change them to both be "warning" then to be consistent

juliev0: No, it wasn't a warning, as it wasn't our code printing the log line - it was the library printing. The motivation for adding the warning log line was so it would be more visible.

agilgur5: Ohhhh, I didn't know that was an internal print -- thought it was just due to a dep upgrade that the message was changed. Thanks for the clarification!
So does that mean that _both_ are printed in 3.5 then?

agilgur5:
```suggestion
Or, in >= v3.5, if you see warnings similar to this (could be any CR, not just `WorkflowTemplate`):
```

I think that would be most consistent while still having the warning piece then 🤔

docs/scaling.md
```diff
@@ -12,12 +12,43 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+
+- If you're using a lot of `CronWorkflows` and they don't seem to be firing on time, increase `--cron-workflow-workers`.
+
+### K8S API Client Side Rate Limiting
+
+The K8S client library rate limits the messages that can go out. The default values are fairly low.
+
+If you frequently see messages similar to this in the Controller log (issued by the library):
+
+```
```
agilgur5, line 39:
```suggestion
```txt
```

ah I forgot `markdownlint` requires code blocks specify a language, my bad

docs/scaling.md
```diff
@@ -12,12 +12,43 @@ As of v3.0, the controller supports having a hot-standby for [High Availability]
...
+Waited for 7.090296384s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/argoproj.io/v1alpha1/namespaces/argo/workflowtemplates/s2t
+```
+
+or for >= v3.5: a warning like this (could be any CR, not just `WorkflowTemplate`):
+
+```
```
agilgur5, line 45:
```suggestion
```txt
```

same as above for `markdownlint`

agilgur5 (approved):
LGTM, thanks for the copy edits!

[Listing defaults](https://github.com/argoproj/argo-workflows/pull/11731#issuecomment-1704502523) would be good, but that can be done as a follow-up, this is already a solid improvement as is

juliev0:
> We may want to list the defaults for these flags here as well so that there is a comparison point for when they are not explicitly set

I suppose I was reluctant to do that since the values may change and we may forget to update the documentation, although admittedly I did say that the defaults were low, which could potentially change as well.

agilgur5:
> I suppose I was reluctant to do that since the values may change and we may forget to update the documentation

We don't change defaults that often though. Right now I think it's more likely that users have no idea what the defaults are 😬

In https://github.com/argoproj/argo-workflows/pull/11457#issuecomment-1656549995 I also mentioned that flag options would probably be a good use-case for docgen. So maybe we consider docgen as future state for these but may still want something in the interim as docgen is non-trivial.

terrytangyuan (approved):

PR #11733, opened 2023-09-03
feat: Update to use markdownlint compatible changelog

sudosubin:
Resolved markdownlint errors for `CHANGELOG.md`. Related with 

### Motivation

Argo Workflows project has markdownlint configuration, but many markdown documents does not respect the lint rules. (I'm currently working on markdownlint.)

### Modifications

I also updated `./hack/changelog.sh` script.

- `MD024`: https://github.com/DavidAnson/markdownlint/blob/main/doc/md024.md
  - Changed `Multiple headings with the same content` allows only sibling headings
- `MD034`: https://github.com/DavidAnson/markdownlint/blob/main/doc/md034.md
  - Changed `Bare URL used` allowed

### Verification

```sh
$ markdownlint ./CHANGELOG.md
(clear)
```


.markdownlint.yaml
```diff
@@ -3,3 +3,6 @@ MD013: false
 # mkdocs uses 4 spaces indent
 MD007:
   indent: 4
+MD024:
+  siblings_only: true
+MD034: false
```
terrytangyuan, line 8: Would this break the links to the existing headers, e.g. https://github.com/argoproj/argo-workflows/blob/master/CHANGELOG.md#v350-rc1-2023-08-15?

sudosubin: I'm not sure about the concerns, but the anchor to the header seems to work well: https://github.com/argoproj/argo-workflows/blob/428ddb118e826d322527a7a5c59e80418401ae1f/CHANGELOG.md#v350-rc1-2023-08-15

[`MD034`](https://github.com/DavidAnson/markdownlint/blob/main/doc/md034.md) is a lint rule about using url in markdown file. (ex. `https://example.com` in `*.md` will produce warning)

CHANGELOG.md
terrytangyuan, line 1: Can you remove changes in this file? This will be auto-generated.

sudosubin: Of course! I force-pushed the latest commit. Thansk for the comment :D

terrytangyuan (approved):

.markdownlint.yaml
agilgur5: I think it would be better to configure these inline for the whole file rather than globally.

As in, using:
```html
<!-- markdownlint-disable MD034 -->
<!-- markdownlint-configure-file { "MD024": { "siblings_only": true } } -->
```

Also some comments would be nice as to why these were disabled, as is convention

PR #11736, opened 2023-09-03
fix: Permit enums w/o values. Fixes #11471.

### LINKED ISSUE #11471: Argo CLI: Lint: Enums can't have default in WorkflowTemplate
### Pre-requisites

This appears to be a variation of #6044.

When creating a WorkflowTemplate I would expect that I can define an `enum` parameter that also provides a `default` matching one of the enum values if no value was provided when the `WorkflowTemplate` becomes an actual `Workflow`. This does not appear to be the case. When I attempt to `argo lint` this I get the following error:

```
in "value-bug-example" (WorkflowTemplate): rpc error: code = InvalidArgument desc = spec.arguments.test-parameter.value is required
```

### Version

v3.4.9

```YAML
kind: WorkflowTemplate
metadata:
  name: value-bug-example

spec:
  arguments:
    parameters:
      - name: test-parameter
        enum: ['hello', 'goodbye']
        default: 'hello'

  templates:
    - name: test
      script:
        image: ubuntu
        source: echo {{workflow.parameters.test-parameter}}
```

```text
N/A
```

```text
N/A
```

farmdawgnation:
Fixes #11471 

### Motivation

As described in the above issue, it's currently pretty impossible to use `enum` fields and have `lint` pass. This makes them effectively useless in workflow templates.

### Modifications

Updated the enum checks in the validate code in order to respect `allowEmptyValues`.

### Verification

Updated unit tests and tweaked code until they passed as expected.

### DIFF
--- a/workflow/validate/validate.go
+++ b/workflow/validate/validate.go
@@ -865,10 +865,11 @@ func validateArgumentsValues(prefix string, arguments wfv1.Arguments, allowEmpty
 			if len(param.Enum) == 0 {
 				return errors.Errorf(errors.CodeBadRequest, "%s%s.enum should contain at least one value", prefix, param.Name)
 			}
-			if param.Value == nil && !allowEmptyValues {
+			if param.Value == nil {
+				if allowEmptyValues {
+					return nil
+				}
 				return errors.Errorf(errors.CodeBadRequest, "%s%s.value is required", prefix, param.Name)
-			} else if param.Value == nil {
-				return nil
 			}
 			valueSpecifiedInEnumList := false
 			for _, enum := range param.Enum {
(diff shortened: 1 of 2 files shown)


workflow/validate/validate.go
```diff
@@ -865,8 +865,10 @@ func validateArgumentsValues(prefix string, arguments wfv1.Arguments, allowEmpty
 			if len(param.Enum) == 0 {
 				return errors.Errorf(errors.CodeBadRequest, "%s%s.enum should contain at least one value", prefix, param.Name)
 			}
-			if param.Value == nil {
+			if param.Value == nil && !allowEmptyValues {
```
Joibel, line 868: I think this would read easier as:
```go
if param.Value == nil {
  if allowEmptyValues {
    return nil
  }
  return errors.Errorf(errors.CodeBadRequest, "%s%s.value is required", prefix, param.Name)
}
```

farmdawgnation: Sounds good, I'll make that change.

Joibel (approved):
Looks good to me, thank you.

farmdawgnation:
Someone may need to rerun the E2E tests... I don't seem to have permission to do so!

Joibel:
> Someone may need to rerun the E2E tests... I don't seem to have permission to do so!

@Just push an empty commit or something similar to get the CI to re-run

terrytangyuan (approved):

PR #11738, opened 2023-09-03
fix: add workflow outputs to scope for arguments

### LINKED ISSUE #11610: Cannot access output global artifacts in exit handler after upgrading to v3.4.6+
### Pre-requisites

Whenever we are calling the global output artifact in the exithandler Its failing with the error:

Bad Request: templates.exit-handler.steps failed to resolve ```{{workflow.outputs.artifacts.output-result-car}}```

This error was introduced after upgrading to the v3.4.6,. Also have tested with v3.4.7, v3.4.8, v3.4.9, v3.4.10 same issue persist. 

After downgrading to v3.4.5  same workflow is working without any issue

### Version

v3.4.6

```YAML
kind: Workflow
metadata:
  name: exit-handler-with-artifacts
spec:
  onExit: exit-handler
  entrypoint: main
  arguments:
    parameters:
    - name: params
      value: |
        [
          { "variable": "car"},
          { "variable": "bike"}
        ]
  templates:
    - name: main
      steps:
        - - name: step-1
            template: output
			arguments:
			  parameters:
			  - name: variable
			    value: "{{item.variable}}"
            withParam: "{{workflow.parameters.params}}" 
    
    - name: output
	  inputs:
        parameters:
        - name: variable
      script:
        image: python:alpine3.6
        command: [sh, -c]
        args: ["echo $variable > /result.txt "]
      outputs:
        artifacts:
          - name: result-{{inputs.parameters.variable}}
            path: /result.txt
            globalName: output-result-{{inputs.parameters.variable}}

    - name: printer
      inputs:
        artifacts:
          - name: message-bike
            path: /tmp/message-bike
          - name: message-car
            path: /tmp/message-car
      container:
        image: python:alpine3.6
        command: [sh, -c]
        args: ["cat /tmp/message.*"]
			
    - name: exit-handler
	  steps:
	    - - name: printer
			template: printer
			arguments:
			  artifacts:
			  - name: message-car
				from: "{{workflow.outputs.artifacts.output-result-car}}"          
			  - name: message-bike
				from: "{{workflow.outputs.artifacts.output-result-bike}}" 
```
```


```text
NA
```

```text
NA
```

rohankmr414:
Fixes #11610

### Modifications

Added workflow global outputs to scope for arguments

### Verification

Tested locally


isubasinghe (approved):
LGTM, is it possible to add a test for this ? Perhaps in e2e

workflow/validate/validate.go
```diff
@@ -1374,7 +1379,7 @@ func (ctx *templateValidationCtx) validateDAG(scope map[string]interface{}, tmpl
 		ancestry := common.GetTaskAncestry(dagValidationCtx, task.Name)
 		for _, ancestor := range ancestry {
 			ancestorTask := dagValidationCtx.GetTask(ancestor)
-			resolvedTmpl := resolvedTemplates[ancestor]
+			resolvedTmpl = resolvedTemplates[ancestor]
```
Joibel, line 1382: This change doesn't actually do anything but is probably a wrong change and at minimum would be best reverted.

`resolvedTmpl` is not used outside of this for loop in the outer scope **after** this loop. So the fact that this variable used to shadow the outer `resolvedTmpl` but doesn't any more doesn't actually change the behaviour. However - this value of resolvedTmpl is definitely _not_ useful outside of this loop, and the formerly shadowed version might be. The right fix is to use a different name for this variable inside this loop.

rohankmr414: reverting this

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3054,6 +3054,11 @@ func (tmpl *Template) HasOutput() bool {
 	return tmpl.Container != nil || tmpl.ContainerSet.HasContainerNamed("main") || tmpl.Script != nil || tmpl.Data != nil || tmpl.HTTP != nil || tmpl.Plugin != nil
 }
 
+// HasArguments returns true if the template can have arguments.
+func (tmpl *Template) HasArguments() bool {
```
Joibel, line 3058: I'd prefer if this was `CanHaveArguments` as it's slightly misleadingly named as is.

Joibel:
@rohankmr414 shall I'll take over this PR

tooptoop4:
☂️

Issue #10826, opened 2023-04-04
Better first-time user experience with popups


terrytangyuan:
Currently, there are various popups:
![image](https://user-images.githubusercontent.com/4269898/229902635-74ba0999-7b13-459e-87a6-721c40124abc.png)

These can be disabled by setting 3 different env vars in argo-server: 

![image](https://user-images.githubusercontent.com/4269898/229902725-0de5db37-e412-4614-b516-8b5f4a0635b3.png)

Perhaps we should consider changing the default or make them less intrusive.

caelan-io:
We also get a ton of negative feedback about these via the in-product survey.

alexec:
If you disable them, you mights as well remove them. 

We will not get anymore survey results if you do that. 

I think there is a bug in them, as they should disappear after being first shown. Sounds like they pop up every time someone opens the app.

What about adding instructions to the pop-ups on how to disable them?

agilgur5:
One of the other issues with this is that it'll show up whenever cookies are cleared/reset, which is not necessarily a first-time user. That being said, there really isn't a better way to detect FTUs. 

I have definitely gotten annoyed by these modals as a result of that, and I just automatically close these when they pop-up without even reading -- I've also seen many colleagues do that too. I've unintentionally missed new version specifics because of that habit as well, so this is perhaps encouraging the opposite behavior now 😕 

I do think that the first 3 modals at least could be made less obtrusive. Perhaps as much smaller, closeable toasts in the corner instead of huge modals.

Issue #9696, opened 2022-09-27
UI does not show all running workflows (it used to do so) / UI and argo list differ (missing running workflows)


scravy:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

When I am submitting a workflow in the UI it is shown in the list of running workflows. When refreshing this page all workflows should be listed.

We just upgraded to argo 3.4.0 and not all running workflows are shown in the UI. When you wait long enough eventually workflows pop up. In our day to day work we noticed that workflows seem to be added to the UI once a transition happens, i.e. when a new step is entered in the workflow.

In short: The output of `argo list` and the workflows shown in the UI are not the same. Some running flows are missing. They are being added once they transition from one node to another while the UI is shown.

### Version

3.4.0

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: zargo-debug-child
  namespace: argo
spec:
  entrypoint: main
  templates:
    - name: main
      steps:
        - - name: node1
            template: work
        - - name: node2
            template: work
    - name: work
      script:
        image: bash:5.2.0-alpine3.15
        command:
          - bash
        source: |
          echo "Doing some work for 60 seconds"
          sleep 60
```

### Logs from the workflow controller

```
time="2022-09-27T15:57:49.577Z" level=info msg="Processing workflow" namespace=argo workflow=zargo-debug-child-vcmgq
time="2022-09-27T15:57:49.591Z" level=info msg="Updated phase  -> Running" namespace=argo workflow=zargo-debug-child-vcmgq
time="2022-09-27T15:57:49.593Z" level=info msg="Steps node zargo-debug-child-vcmgq initialized Running" namespace=argo workflow=zargo-debug-child-vcmgq
time="2022-09-27T15:57:49.593Z" level=info msg="StepGroup node zargo-debug-child-vcmgq-3865638504 initialized Running" namespace=argo workflow=zargo-debug-child-vcmgq
[...]

PR #11742, opened 2023-09-04
fix(ui): merge WF List FTU Panel with New Version Modal

### LINKED ISSUE #10821: Workflows are not showing up for first-time users
Image

agilgur5:
Fixes #10821
Technically a follow-up to #8596

### Motivation

- Workflows being hidden behind a first-time user panel was confusing
  - reported in #10821 with a few upvotes
  - personally, I did not even realize that panel specified a new default, it just seemed like another unnecessary click when my cookies get cleared -- similar to the survey etc (i.e. #10826)

### Modifications

- Move the description of the new default instead to the New Version Modal
  - and specify in **bold** that there is a new default
  
- unrelated: simplify some syntax in this file using optional chaining (from [investigating](https://github.com/argoproj/argo-workflows/issues/9696#issuecomment-1704688457) #9696)

### Verification

1. Cleared cookies and loaded `/workflows`
   1. Confirmed that New Version Modal has the new bullet point and has the new default in bold. 
   1. Confirmed that Workflows List is shown immediately.
   
See below screenshot with the New Version Modal as well as the Workflow List in the background:
Screenshot 2023-09-04 at 2 20 39 AM
 
### Future Work

`WorkflowList` could be optimized a bit. It's got a few internal functions that are purely static and could be moved out of the component as such. Left the diff pretty minimal in this PR though

### DIFF
--- a/ui/src/app/modals/new-version/new-version-modal.tsx
+++ b/ui/src/app/modals/new-version/new-version-modal.tsx
@@ -47,6 +47,9 @@ export const NewVersionModal = ({version, dismiss}: {version: string; dismiss: (
                     </a>
                     in the middle of the workflow
                 </li>
+                <li>
+                    Filter by date and time in the UI. <b>New default</b>: only show the past month's Workflows.
+                </li>
             </ul>
             <p>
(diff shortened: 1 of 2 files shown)


isubasinghe (approved):
LGTM

terrytangyuan (approved):

Issue #11741, opened 2023-09-04
UI: `links` for label containing dot not working


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

using links: in wfcontroller configmap:

for a label key not containing slash it works, ie this works:
```label=s3-folder%3D${workflow.metadata.labels.s3-folder}```

for a label key containing dot slash it returns null in the generated link url:
```label=workflows.argoproj.io%2Fworkflow-template%3D${workflow.metadata.labels.workflows.argoproj.io/workflow-template}```
below also returns null
```label=workflows.argoproj.io%2Fworkflow-template%3D${workflow.metadata.labels.workflows.argoproj.io%2Fworkflow-template}```

### Version

3.4.8

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

agilgur5:
If this is a UI-only error, [this function](https://github.com/argoproj/argo-workflows/blob/8e01696a9c63f25b3660ed479c6f7a57bb61d82d/ui/src/app/shared/components/links.tsx#L24) may need updating to handle this.

Is it slashes specifically that fail, or the fact that the label has dots in it?
[This line](https://github.com/argoproj/argo-workflows/blob/8e01696a9c63f25b3660ed479c6f7a57bb61d82d/docs/workflow-controller-configmap.yaml#L82) in the example configmap works, but that's a static link, not a templated link

tooptoop4:
my guess is that it doesn't know how to reference label key with dot because it might be searching for a key of 'workflows'

agilgur5:
[...]

PR #11743, opened 2023-09-04
refactor(ui): optimize Link functionality

agilgur5:
### Motivation

- Optimize the `links` module and its usage a bit
- Simplify code a bit by reducing nesting
- Fix some JS conventions etc

Stumbled upon this while investigating #11741 and thought it could use some improvements

### Modifications

- move static functions outside of the component
  - this way they are not created every time the component is and hence use less memory etc
- use named `function`s instead of `const`s assigned to anonymous functions
  - this makes tracing a bit easier and source maps a bit clearer
  - (also technically a memory improvement as well)
- both of these also reduce nesting

- re-use `openLinkWithKey` in `WorkflowDetails` instead of having duplicate code
  - rename to `openLinkWithKey` so that it doesn't overlap with `openLink` and to be a bit more specific
  - (also a very tiny memory improvement)

- remove redundant `formatURL` function

- use lowercase "p" in `processURL`
  - JS convention that only classes are capitalized (or components in React, which used to always be classes)
    - I imagine this might have been capitalized b/c Go uses capitals for exports?

- be more specific with `urlExpression` as the argument since it is not necessarily a raw URL and that is why it needs "processing"

### Verification

Stylistic and performance changes, but no usage changes (minor semantic differences that improve tracing, memory, etc as mentioned above, but do not change usage).


isubasinghe (approved):
Could totally be my lack of frontend knowledge, but I didn't quite understand the function name for `openLinkWithKey` here, I assumed this was more my fault and approved.

agilgur5:
> I didn't quite understand the function name for `openLinkWithKey` here

I wrote this a few weeks ago, but it's quite literal, since it's when clicked with `ctrlKey` or `metaKey` (i.e. Ctrl+Click or Middle Click) to open in a new tab. "Key" as in with one of those two keys (or without)

terrytangyuan (commented):
This doesn't seem to fix https://github.com/argoproj/argo-workflows/issues/11741?

terrytangyuan (approved):

agilgur5:
poop, this had a one-line semi-colon conflict with #11882 😭 . mentioned [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1695705273908319?thread_ts=1694644337.627309&cid=C0510EUH90V) that I tried pretty hard to avoid a conflict on this old PR but missed a single character

PR #11748, opened 2023-09-05
docs: Modify CronWorkflowSpec, Status fields examples

newdeal123:
### Motivation

Docs examples that do not use cronworkflow are included in the CronWorkflowSpec, CronWorkflowStatus fields, causing confusion in finding the necessary data.

### Modifications

Examples in the CronWorkflowSpec, CronWorkflowStatus field allow only examples using cronworkflow to find the necessary materials.

### Verification

I submitted workflows by copy & paste.


terrytangyuan (commented):
This is an autogenerated file so you'd want to fix the docgen script.

PR #11752, opened 2023-09-05
ci: better naming for various steps/jobs

agilgur5:
### Motivation

- partly inspired by #11670
  - as well as seeing users be confused by some steps
    - `git diff --exit-code` had [some confusion](https://github.com/argoproj/argo-workflows/pull/10806#issuecomment-1537913872) in particular
  - and me just reading more closely and wanting better descriptions (as I am working on a few CI changes)
    - for example, I did not even realize that the steps after an E2E test failure were for debugging purposes until I read the CI code itself
      - GH runs some post action hooks etc, so thought that was just clean-up or something

### Modifications

- add names to more steps & jobs in GH Actions Workflows
  - specific names for all `git diff --exit-code` as per above
  - names for Docs steps
  - names PR Title check steps
  - names for Snyk dependency scans workflow
  - CI workflow names:
    - all E2E test failure debugging steps per above
    - add a "get" (as in `kubectl get`) in front of the ones that just log out a resource for clarity as well
    - Java and Python SDK set-up/install steps
    - a few other clarifications / improvements
    - additions where there were no names

- also add some new lines and comments in a few places for clarity & readability

- reordering of fields
  - follow a consistent order of `name` -> `if` -> `needs`
  - then `uses`, `steps`, or `run` depending on the context

### Verification

- Purely stylistic and DX changes, no semantic changes
- CI on this PR should pass with all the new names too 🙂

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -10,12 +10,12 @@ permissions:
 
 jobs:
   generate_changelog:
+    name: Generate changelog
+    if: github.repository == 'argoproj/argo-workflows'
     permissions:
       contents: write  # for peter-evans/create-pull-request to create branch
       pull-requests: write  # for peter-evans/create-pull-request to create a PR
-    if: github.repository == 'argoproj/argo-workflows'
     runs-on: ubuntu-latest
-    name: Generate changelog
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   tests:
-    name: Unit Tests
+    name: Go Unit Tests
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -29,11 +29,11 @@ jobs:
           cache: true
       - run: make test STATIC_FILES=false GOTEST='go test -p 20 -covermode=atomic -coverprofile=coverage.out'
       # engineers just ignore this in PRs, so lets not even run it
-      - run: bash <(curl -s https://codecov.io/bash)
-        if: github.ref == 'refs/heads/master'
+      - if: github.ref == 'refs/heads/master'
+        run: bash <(curl -s https://codecov.io/bash)
 
   argoexec-image:
-    name: Create argoexec image
+    name: argoexec image
     # needs: [ lint ]
     runs-on: ubuntu-latest
     timeout-minutes: 10
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   tests:
-    name: Go Unit Tests
+    name: Unit Tests
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -34,7 +34,7 @@ jobs:
         run: bash <(curl -s https://codecov.io/bash)
 
   argoexec-image:
-    name: argoexec image
+    name: argoexec-image
     # needs: [ lint ]
     runs-on: ubuntu-latest
     timeout-minutes: 10
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -33,10 +33,10 @@ jobs:
         if: github.ref == 'refs/heads/master'
 
   argoexec-image:
-    name: argoexec-image
+    name: Create argoexec image
+    # needs: [ lint ]
     runs-on: ubuntu-latest
     timeout-minutes: 10
-    # needs: [ lint ]
     steps:
       - uses: actions/checkout@v4
       - uses: docker/setup-buildx-action@v3
@@ -49,7 +49,8 @@ jobs:
           target: argoexec
           cache-from: type=gha
           cache-to: type=gha,mode=max
-      - uses: actions/upload-artifact@v3
+      - name: Upload
+        uses: actions/upload-artifact@v3
         with:
           name: argoexec
           path: /tmp/argoexec_image.tar
@@ -57,9 +58,9 @@ jobs:
 
   e2e-tests:
     name: E2E Tests
+    needs: [ argoexec-image ]
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    needs: [ argoexec-image ]
     env:
       KUBECONFIG: /home/runner/.kubeconfig
     strategy:
@@ -96,22 +97,23 @@ jobs:
             install_k3s_version: v1.25.11-k3s1
             profile: minimal
     steps:
-      - name: Install socat
-        # needed by Kubernetes v1.25
+      - name: Install socat (needed by Kubernetes v1.25)
         run: sudo apt-get -y install socat
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v4
         with:
           go-version: "1.21"
           cache: true
-      - uses: actions/setup-java@v3
+      - name: Install Java for the SDK
         if: ${{matrix.test == 'test-java-sdk'}}
+        uses: actions/setup-java@v3
         with:
           java-version: '8'
           distribution: adopt
           cache: maven
-      - uses: actions/setup-python@v4
+      - name: Install Python for the SDK
         if: ${{matrix.test == 'test-python-sdk'}}
+        uses: actions/setup-python@v4
         with:
           python-version: '3.x'
           cache: pip
@@ -129,13 +131,13 @@ jobs:
           echo "  user:" >> $KUBECONFIG
           echo "    token: xxxxxx" >> $KUBECONFIG
           until kubectl cluster-info ; do sleep 10s ; done
-      - uses: actions/download-artifact@v3
-        name: Download argoexec image
+      - name: Download argoexec image
+        uses: actions/download-artifact@v3
         with:
           name: argoexec
           path: /tmp
-      - run: docker load < /tmp/argoexec_image.tar
-        name: Load argoexec image
+      - name: Load argoexec image
+        run: docker load < /tmp/argoexec_image.tar
       - name: Set-up /etc/hosts
         run: |
           echo '127.0.0.1 dex'      | sudo tee -a /etc/hosts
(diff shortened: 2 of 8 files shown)


agilgur5:
Ok top-level workflow names and job names don't look great when they're long. Gonna make some modifications...

agilgur5:
Hmmm required checks evidently require the name to be exactly the same instead of the key 😕 

See screenshot:
Screenshot 2023-09-05 at 9 09 54 PM

I'm going to revert the "Unit Tests" and "argoexec-image" name changes then, especially as those did not add altogether much

Yaminyam:
image
Because the actions list is divided by the yaml file name, you should be aware that two actions with the same name may result in confusion.
I think it would be a good idea to delete the existing list when there are enough logs of newly changed docs actions and the old logs are no longer needed.

agilgur5:
> Because the actions list is divided by the yaml file name, you should be aware that two actions with the same name may result in confusion.

Well that's annoying and confusing 😕
To be fair though, I almost never use the [Actions tab](https://github.com/argoproj/argo-workflows/actions) of the repo and normally just look at checks on commits and PRs.

> I think it would be a good idea to delete the existing list when there are enough logs of newly changed docs actions and the old logs are no longer needed.

I don't know if this is actually possible 😐 [This action](https://github.com/argoproj/argo-workflows/actions/workflows/default-branch-check.yaml) hasn't been run in 10 months but still appears in the list.
If there is such a capability (idk), I don't have permissions to delete any of these unfortunately.

So not sure that we can actually do anything about this, it's a GitHub issue 😕

Yaminyam:
That's right, this is a GitHub issue and discussions are continuing, but there is no response from GitHub.
https://github.com/orgs/community/discussions/26256

So the only way to delete the meaningless list from the workflow list is to delete all logs.
https://stackoverflow.com/questions/57927115/delete-a-workflow-from-github-actions/67000032#67000032

image
As shown in the photo above, github actions logs cannot be checked after 90 days anyway, so it is okay to delete all logs older than 90 days.
I am using the script above to delete and manage logs that are no longer used due to changes in the yaml file.

Yaminyam:
I often look at the actions tab, and the problem when the meaningless list becomes large is that you have to go in to check which of the workflows with the same name is currently in use.
When the number of workflows increases, it may be difficult to find a valid workflow because the workflows are sorted alphabetically.

agilgur5:
> So the only way to delete the meaningless list from the workflow list is to delete all logs.
> https://stackoverflow.com/questions/57927115/delete-a-workflow-from-github-actions/67000032#67000032

I doubt I have permissions to do this, but maybe an approver does.
These [2](https://github.com/argoproj/argo-workflows/actions/workflows/default-branch-check.yaml) [actions](https://github.com/argoproj/argo-workflows/actions/workflows/security.yaml) can be pruned.

`docs.yaml` can't be pruned as it's the new file name here and contains logs from everything other than this PR (future old logs)

Yaminyam:
Even if a lot of logs are piled up, they cannot be accessed after 90 days or more, so I think it is okay to delete the existing logs because they become meaningless after 90 days after the 90-day pr is applied.
But ultimately, this is a matter for those in power to decide. :D

agilgur5:
> Even if a lot of logs are piled up, they cannot be accessed after 90 days or more

If this PR is merged, I would say it would be ok to delete `gh-pages.yaml` after 90 days (although PRs that are based on an old branch will still have it). But I don't have permission to do this regardless

.github/workflows/changelog.yaml
```diff
@@ -10,12 +10,12 @@ permissions:
 
 jobs:
   generate_changelog:
+    name: Generate changelog
+    if: github.repository == 'argoproj/argo-workflows'
```
terrytangyuan, line 14: What's the purpose of this change?

.github/workflows/changelog.yaml
```diff
@@ -30,4 +30,4 @@ jobs:
           title: 'docs: updated CHANGELOG.md'
           commit-message: 'docs: updated CHANGELOG.md'
           branch: create-pull-request/changelog
-          signoff: true
```
terrytangyuan, line 33: Same here

terrytangyuan: And other places in this PR

.github/workflows/ci-build.yaml
```diff
@@ -96,22 +98,23 @@ jobs:
...
-      - uses: actions/setup-java@v3
+      - name: Install Java for the SDK
         if: ${{matrix.test == 'test-java-sdk'}}
+        uses: actions/setup-java@v3
         with:
           java-version: '8'
           distribution: adopt
           cache: maven
-      - uses: actions/setup-python@v4
+      - name: Install Python for the SDK
```
terrytangyuan, line 115: Isn't it already intuitive by using this action?

.github/workflows/ci-build.yaml
```diff
@@ -129,72 +132,75 @@ jobs:
...
+      # failure debugging below
+      - name: Failure debug - get MinIO/MySQL deployment
+        if: ${{ failure() }}
         run: |
           set -eux
           kubectl get deploy
           kubectl describe deploy
-      - if: ${{ failure() }}
-        name: MinIO/MySQL pods
+      - name: Failure debug - get MinIO/MySQL pods
```
terrytangyuan, line 171: This is too implementation specific and may change

.github/workflows/snyk.yml
```diff
@@ -14,18 +14,20 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - name: Run Snyk to check for vulnerabilities
+      - name: Run Snyk to check for Go vulnerabilities
         uses: snyk/actions/golang@master
         with:
           args: --severity-threshold=high
 
   node:
+    name: Scan Node deps
```
terrytangyuan, line 30: Not necessary. It's already under a `node` section

terrytangyuan (commented):
Please avoid unnecessary changes

.github/workflows/ci-build.yaml
agilgur5: I did not know why Python was being installed, given that it's not a language used inside of Argo. The SDK has a test here, and I mention the SDK in the name

.github/workflows/changelog.yaml
agilgur5: there's no change here, it's just a newline at EOF, which GitHub and many editors point out. Mine automatically adds it when missing in any file

.github/workflows/changelog.yaml
agilgur5: See my opening comment, which explicitly says:

> * reordering of fields
>     * follow a consistent order of `name` -> `if` -> `needs`
then `uses`, `steps`, or `run` depending on the context

.github/workflows/ci-build.yaml
agilgur5: Not sure what you mean, it already said "MinIO/MySQL pods", which I had no clue why it was doing that before, so I added "Failure debug" to be explicit.
I also didn't know what "MinIO/MySQL pods" meant. Versus, I run an SRE team and if you say "`[kubectl] get` a Pod", everyone knows what means. If you say "MinIO/MySQL pods", I would say "what about them?".
I can change it to `describe` which would be even more explicit.

There's two lines of code here, so yes, I would very much expect that the name would need changing if the two lines change. I don't think that's a bad thing, that means the name is explicit about what the two lines do and there is no guesswork to be had when looking at the CI logs, which normally auto-collapse these under the name.

.github/workflows/snyk.yml
agilgur5: It makes more sense in context; the below screenshot just says "Snyk / node":

Screenshot_2023-09-26-10-36-32-04_40deb401b9ffe8e1df2f1cc5ba480b12

That is not entirely clear what it does since Snyk has many functionalities (see also automated PRs #11844, scanning Docker images, etc). This makes it very explicit at first glance

agilgur5:
> Please avoid unnecessary changes
 
I would disagree that these are unnecessary. I had several iterations and trial-and-error testing on this, as can be seen in the commit log and first two comments.

It seemed like you mostly didn't like the reordering and may have misinterpreted those as removals?

Per my opening comment, that was a consistency change:

> * reordering of fields
>   * follow a consistent order of `name` -> `if` -> `needs`
>   * then `uses`, `steps`, or `run` depending on the context

would be great if we had an automated formatter to sort fields in the same order, but I'm not sure that exists, so did so manually for now.
It's quite a bit easier on the eyes when they are consistent as opposed to all over the place, which makes it harder to follow and/or find the fields that I'm looking for. `name` first acts similarly to a comment, then `if` and `needs` are conditionals that say whether the rest is relevant (in code, the rest would be nested under these) or can be skipped.

Same as [in-line comment](https://github.com/argoproj/argo-workflows/pull/11752#discussion_r1337275193)

agilgur5:
ah poop looks like this already ended conflicting with #11782. That was one of the reasons I asked for reviews [over Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1694726319948949?thread_ts=1694644337.627309&cid=C0510EUH90V) as renames and reorderings can be really easy to conflict with

terrytangyuan (approved):

.github/workflows/ci-build.yaml
agilgur5: I updated it to `describe`

agilgur5:
Rebased and fixed merge conflicts! fingers crossed those are the last ones before this get merged 🤞

terrytangyuan:
Build failed

agilgur5:
Yea [docs](https://github.com/argoproj/argo-workflows/actions/runs/6322548346/job/17168375993?pr=11752) just timed out on checkout. That's happened a few times recently in CI actually.
The other one is [a test failure](https://github.com/argoproj/argo-workflows/actions/runs/6322548348/job/17168505793?pr=11752) that I've actually never seen before (`TestJSONVariables`).

But my changes don't affect either of those, so those are flakes in context. I can do an empty commit but I was hoping you could retry those @terrytangyuan

agilgur5:
one of the flakey hooks tests is [failing](https://github.com/argoproj/argo-workflows/actions/runs/6341809982/job/17226399531?pr=11752) now 😕

PR #11759, opened 2023-09-06
feat: Add support to configure Plugin Agent Retry Backoff Policy.

sid8489:
Signed-off-by: sid8489 <agrawal8489@gmail.com>

### Motivation
Plugin agent Pod main container makes http call to plugin containers. That Http call is auto retried for transient errors using a Retry Backoff Policy. Currently the Retry Backoff Policy is hardcoded to default values. These default values might not be fit for certain use cases (eg: My Plugin Container is a python http server, but I am downloading the source code at run time, so this can take some seconds and default retry policy is too aggressive for this). So proposing to make it configurable .

### Modifications
- Modified Retry Backoff Policy used by Plugin agent to lookup for environment variables and configure correspondingly.
- Added support to supply Env Variable to Extended Agent Pod Init and Main  containers  through Executor Config of Workflow Controller

### Verification

 TODO: If these changes are acceptable, I will add the verification details.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -129,6 +129,10 @@ spec:
 | `REMOVE_LOCAL_ART_PATH`                | `bool`          | `false` | Whether to remove local artifacts.                                                                     |
 | `RESOURCE_STATE_CHECK_INTERVAL`        | `time.Duration` | `5s`    | The time interval between resource status checks against the specified success and failure conditions. |
 | `WAIT_CONTAINER_STATUS_CHECK_INTERVAL` | `time.Duration` | `5s`    | The time interval for wait container to check whether the containers have completed.                   |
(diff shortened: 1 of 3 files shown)


terrytangyuan (commented):
Can we reuse existing retry environment variables?

sid8489:
Initially

> Can we reuse existing retry environment variables?

Initially I thought of using existing variables only. But then it couples the retry policy of executor init container and plugin container. One might want to just configure  retry policy  of the plugin executor container or just of the executor init container. Let me know your thoughts, IMO plugin retry policy should be allowed to configured independently.

terrytangyuan:
Let's reuse existing env vars for now.

sid8489:
> Let's reuse existing env vars for now.

Have made the change.

terrytangyuan:
Could you fix the checks?

terrytangyuan:
Closing since this has been inactive for several months. Feel free to re-open if you pick it up again.

Issue #11381, opened 2023-07-18
Release v3.5 preparation


terrytangyuan:
We will release everything on the master branch. 

There might be outstanding high-priority issues we'd like to address before v3.5 release. Let's use this issue to track and discuss.

terrytangyuan:
Per discussions from today's meeting, CronWorkflow related issues are low priority and not blockers for this release.

advithVashist:
Hi!
I was wondering if we could get this PR included in a this release or a testing build: https://github.com/argoproj/argo-workflows/pull/10927

advithVashist:
Hi!
I was wondering if we could get this PR included in v3.5 and RC-1: https://github.com/argoproj/argo-workflows/pull/10927

terrytangyuan:
```
git log --format=' * [%h](https://github.com/argoproj/argo-workflows/commit/%H) %s' --grep "^feat(*.*)*:.*(#" v3.4.0...master | cat
```
Complete list of features:

 * [6b3620091](https://github.com/argoproj/argo-workflows/commit/6b362009138ac2ee16cb07f9206b56794b6de0c4) feat: Use WorkflowTemplate/ClusterWorkflowTemplate Informers when validating CronWorkflows (#11470)
 * [e53a26579](https://github.com/argoproj/argo-workflows/commit/e53a265799bd4ae10681a4c5d4dba8ae03c0a62f) feat: improve alibaba cloud credential providers in OSS artifacts (#11453)
 * [be0bdf9b0](https://github.com/argoproj/argo-workflows/commit/be0bdf9b0eab7d9d23fbb8df0426b4075af6830d) feat: Expose the Cron workflow workers as argument (#11457)
 * [90930ab88](https://github.com/argoproj/argo-workflows/commit/90930ab88b18b7fba3074cdc06059eb6460b50d9) feat: cli allow retry successful workflow if nodeFieldSelector is set. Fixes #11020 (#11409)
 * [336d8a41a](https://github.com/argoproj/argo-workflows/commit/336d8a41a455896d97ea751e46c8e2bcb712fa84) feat: logging for client-side throttling (#11437)
 * [a76674c82](https://github.com/argoproj/argo-workflows/commit/a76674c829461bfb252e904c3fef9c231cadbb56) feat: Allow memoization without outputs (#11379)
[...]

PR #11761, opened 2023-09-06
fix: apply custom cursor pagination where workflows and archived workflows are merged

### LINKED ISSUE #11715: Pagination may not work correctly for archived workflows
### Pre-requisites

When we perform the merge of live workflows and archived workflows (in order to only show one of them in the UI), the `listMeta`  in `WorkflowList` is obtained from live workflows. This may cause issues for pagination for archived workflows.

```go
mergedWfsList := v1alpha1.WorkflowList{Items: mergedWfs, ListMeta: liveWfs.ListMeta}
```

We need to test the pagination more thoroughly and revisit this part of the code. https://github.com/argoproj/argo-workflows/blob/master/server/workflow/workflow_server.go#L155C2-L155C86

### Version

latest

```YAML
TBA
```

```text
```

```text
```

sunyeongchoi:
Fixes #11715

### Motivation

Workflows and archive workflows use different pagination logics.
However, in the current workflow page, if both workflows and archive workflows coexist, we want them to be retrieved together.
The issue arises from passing the pagination options used for workflows pagination directly to archive workflows, causing a mismatch between the types used in archive workflows and workflows, resulting in archive workflows not being retrieved.

### Modifications

I have implemented the following logic to resolve this issue:

1. fetch the entire list of workflows and archived workflows separately.
2. merge the workflows and archived workflows.
3. perform cursor pagination on the merged result based on the resourceVersion.

After contemplating on how to address this issue, I concluded that the conventional approach of paginating workflows and archived workflows separately would not easily resolve this problem. Therefore, I implemented it in this manner.

### Verification

* Confirming the pagination is functioning correctly.
* Checking if Archive workflows are retrieved successfully.
* Verifying that Workflows are retrieved correctly.
* Validating the proper functioning of tests with the 'make test' command.


terrytangyuan (commented):
This is great! Could you also look into adding some tests around `cursorPaginationByResourceVersion`?

sunyeongchoi:
> This is great! Could you also look into adding some tests around `cursorPaginationByResourceVersion`?

Sure! I'll write a test code and upload a commit tomorrow :)

server/workflow/workflow_server.go
```diff
@@ -152,43 +195,49 @@ func mergeWithArchivedWorkflows(liveWfs v1alpha1.WorkflowList, archivedWfs v1alp
...
 	}
-	s.instanceIDService.With(listOption)
-	wfList, err := wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).List(ctx, *listOption)
+
+	// Save the original Continue and Limit.
+	resourceVersion := options.Continue
+	limit := options.Limit
+
+	// Search whole with Limit 0.
+	// Reset the Continue "" to prevent Kubernetes native pagination.
```
terrytangyuan, line 214: Would this approach require fetching the entire list every time user clicks another page?

sunyeongchoi: Yes, the entire page is fetched every time the page changed.

sunyeongchoi: I have written and uploaded test code for cursorPaginationByResourceVersion. 859c0b2ddd2b59b55d542354698fcacb1b198c5a

I am concerned about the potential performance impact of my implementation because it retrieves the entire dataset with each page change. However, I can't think of a better way to solve this problem on the server-side other than this method.

In my opinion, unless there's a better approach, the server should send the entire list to the front-end once, and then pagination can be handled on the front-end. However, one drawback of this approach is that the data won't be updated on the front-end without refreshing the page. To achieve real-time data updates, we might need to consider using websockets or polling.

Do you have any suggestions or better ideas on how to handle this?

server/workflow/workflow_server.go
```diff
@@ -129,7 +129,50 @@ func (s *workflowServer) GetWorkflow(ctx context.Context, req *workflowpkg.Workf
 	return wf, nil
 }
 
-func mergeWithArchivedWorkflows(liveWfs v1alpha1.WorkflowList, archivedWfs v1alpha1.WorkflowList, numWfsToKeep int) *v1alpha1.WorkflowList {
+func cursorPaginationByResourceVersion(items []v1alpha1.Workflow, resourceVersion string, limit int64, wfList *v1alpha1.WorkflowList) {
+	// Use Kubernetes resourceVersion for cursor pagination.
+	// Sort the Kubernetes results in descending order by resourceVersion.
+	// To implement cursor pagination with filtering based on resourceVersion, start by sorting in descending order according to the resourceVersion.
```
terrytangyuan, line 135:
```suggestion
	// Sort the workflow list in descending order by resourceVersion.
```

server/workflow/workflow_server.go
```diff
@@ -129,7 +129,50 @@ func (s *workflowServer) GetWorkflow(ctx context.Context, req *workflowpkg.Workf
...
+	sort.Slice(items, func(i, j int) bool {
+		itemIRV, _ := strconv.Atoi(items[i].ResourceVersion)
+		itemJRV, _ := strconv.Atoi(items[j].ResourceVersion)
+		return itemIRV > itemJRV
+	})
+
+	// Due to the descending sorting above, the items are filtered to have a resourceVersion smaller than the received value.
```
terrytangyuan, line 142: Clarify what "received value/resoureVersion" means

server/workflow/workflow_server.go
```diff
@@ -129,7 +129,50 @@ func (s *workflowServer) GetWorkflow(ctx context.Context, req *workflowpkg.Workf
...
+		endIndex := int(limit)
+		if endIndex > len(items) || limit == 0 {
+			endIndex = len(items)
+		}
+		wfList.Items = items[0:endIndex]
+	} else {
+		wfList.Items = items
+	}
+
+	// Calculate new offset for next batch
```
terrytangyuan, line 167: For next page?

server/workflow/workflow_server.go
```diff
@@ -152,43 +195,49 @@ func mergeWithArchivedWorkflows(liveWfs v1alpha1.WorkflowList, archivedWfs v1alp
...
-	numWfs := 0
-	var finalWfs []v1alpha1.Workflow
-	for _, item := range mergedWfsList.Items {
-		if numWfsToKeep == 0 || numWfs < numWfsToKeep {
-			finalWfs = append(finalWfs, item)
-			numWfs += 1
-		}
-	}
-	return &v1alpha1.WorkflowList{Items: finalWfs, ListMeta: liveWfs.ListMeta}
+	return &v1alpha1.WorkflowList{Items: mergedWfs, ListMeta: liveWfs.ListMeta}
```
terrytangyuan, line 198: Could you add a comment to explain why we are passing `liveWfs.ListMeta` here?

server/workflow/workflow_server.go
terrytangyuan: I don't have a better idea. I think correctness is important right now. Performance-wise, we should do something from the front-end like what you said to only refetch the whole list when refreshing. This can be something that's configurable.

server/workflow/workflow_server.go
sunyeongchoi: Thanks for your review.
I add some comment about `resourceVersion` and `recievedRV` to clarify what "received value/resoureVersion" means.

d32ed71a533c71526d4c50efc27706c0ea9d76fa

server/workflow/workflow_server.go
sunyeongchoi: Yes, for next page.
to clarify I modify the word `batch` to `page`

d433216c5e6739ccd5cc17aa1d006019af2b431a

terrytangyuan:
This looks good once the last comment and the conflicts are resolved. Great work! This should be included in the next RC https://github.com/argoproj/argo-workflows/issues/11381

server/workflow/workflow_server.go
sunyeongchoi: Then, after this PR, I will think about the front page pagination and give it a try.

Also, since I am curious about the performance of the currently implemented pagination, I will think about ways to test its performance.

I will share the test results later when they come out.

terrytangyuan: Great!

terrytangyuan (approved):

server/workflow/workflow_server.go
sunyeongchoi: I add comment why `liveWfs.ListMeta` here :)

77118763136270cfe406f57499af210abd198b15

terrytangyuan:
Need to sort the imports

terrytangyuan:
@sunyeongchoi Looks like the test is flaky. See https://github.com/argoproj/argo-workflows/actions/runs/6166062084/job/16734921912

sunyeongchoi:
> @sunyeongchoi Looks like the test is flaky. See https://github.com/argoproj/argo-workflows/actions/runs/6166062084/job/16734921912

Oh no, I'll quickly post a PR fixing that test. sorry.

sunyeongchoi:
> @sunyeongchoi Looks like the test is flaky. See https://github.com/argoproj/argo-workflows/actions/runs/6166062084/job/16734921912

I uploaded bugfix PR about this :) https://github.com/argoproj/argo-workflows/pull/11816
Thaks for your report.

PR #6712, opened 2021-09-12
feat: Add template node to pod name. Fixes #1319

JPZ13:
This PR is a continuation of Alex's PoC  #6647 that fixes #1319. It:
- Sets pod names to a concatenation of the wf name, template name, and hash
- Allows for an override of the pod name to the node id if the env `USE_LEGACY_POD_NAMES` is set to true
- Stores the node id as an annotation on the pod

Sample terminal output:
```
$ kubectl get pods -n argo
NAME                                   READY   STATUS      RESTARTS   AGE
minio-786d5cbc44-c9ppg                 1/1     Running     7          19h
retry-test-p7jzr-whalesay-2308805457   0/2     Completed   0          13h
retry-test-p7jzr-whalesay-2378053249   0/2     Completed   0          13h
```

PR #11766, opened 2023-09-06
fix(ui): don't use `Buffer` for FNV hash

### LINKED ISSUE #11764: [master branch only] UI crashes when clicking node with an error message: Buffer is not defined
### Pre-requisites

THIS DOES NOT RELATES TO RELEASED IMAGES either v3.5.0-rc nor v3.10.0.
Might be local development only?

* Summary
  * UI crashes when we click a node `hello1`.
  * Curiously, `step-mnss7` is ok.
  * image
  * image
* Procedure
  * checkout master branch
  * `make start UI=true`
* My Env
  * WSL
  * k3d

### Version

master

```YAML
kind: Workflow
metadata:
  generateName: steps-
spec:
  entrypoint: hello-hello-hello

  templates:
  - name: hello-hello-hello
    steps:
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello1"
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello2a"
      - name: hello2b           # single dash => run in parallel with previous step
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello2b"

  - name: whalesay
    inputs:
      parameters:
      - name: message
    container:
      image: docker/whalesay
      command: [cowsay]
```
```


```text
```

```text
```

agilgur5:
Fixes #11764

### Motivation

- `Buffer` requires a polyfill in client-side JS
  - #11628 upgraded to Webpack v5, which [no longer auto adds Node polyfills](https://github.com/webpack/changelog-v5#automatic-nodejs-polyfills-removed)
    - which ended up breaking 2 year old code from #6925
  - this seems to be the only one we use (as far as I can tell) -- and this code quite literally copies server-side Go code, so makes sense that it may use server-side libraries
    - add comments in the code about where this is copied from so that it is easier to track (I had to go through the `blame` + commit history)

### Modifications

- use `charCodeAt` to extract an int from a character of a string instead of converting the string to a Buffer
  - was inspired by looking at some JS FNV implementations and stumbled upon some lines like this: https://github.com/schwarzkopfb/fnv1a/blob/2c5ee2889e42a6bd38c64a29b227eaf9e048e40d/index.js#L22

- also re-enable the `tslint` rule at the end of the code

### Verification

- unit tests for FNV hashing still pass, so think the implementation is equivalent and we're good!

### Misc Notes

I decided not to replace the FNV hash implementation with a library as basically [all](https://www.npmjs.com/search?q=fnv1a) of them are unmaintained and this change only ended up being a few LoC. Some of the libraries are also [quite large](https://github.com/tjwebb/fnv-plus/blob/v1.3.0/index.js) (implementing many different algorithms and are not tree-shakeable), or have [build issues](https://github.com/schwarzkopfb/fnv1a/issues/2), or [use BigInt](https://github.com/sindresorhus/fnv1a/blob/ddbe4f242c6d9a86258400dcf6a1dabae156a3e1/index.js#L5) (which is not supported in older browsers... although those are all EoL and I wouldn't personally support such old browsers, but still something to consider).

### DIFF
--- a/ui/src/app/shared/pod-name.ts
+++ b/ui/src/app/shared/pod-name.ts
@@ -9,6 +9,7 @@ export const k8sNamingHashLength = 10;
 // getPodName returns a deterministic pod name
 // In case templateName is not defined or that version is explicitly set to  POD_NAME_V1, it will return the nodeID (v1)
 // In other cases it will return a combination of workflow name, template name, and a hash (v2)
+// note: this is inteded to be equivalent to the server-side Go code in workflow/util/pod_name.go
 export const getPodName = (workflowName: string, nodeName: string, templateName: string, nodeID: string, version: string): string => {
     if (version !== POD_NAME_V1 && templateName !== '') {
         if (workflowName === nodeName) {
@@ -35,17 +36,17 @@ export const ensurePodNamePrefixLength = (prefix: string): string => {
 };
 
 export const createFNVHash = (input: string): number => {
-    const data = Buffer.from(input);
-
     let hashint = 2166136261;
 
-    /* tslint:disable:no-bitwise */
-    for (const character of data) {
+    for (let i = 0; i < input.length; i++) {
+        const character = input.charCodeAt(i);
+        /* tslint:disable:no-bitwise */
         hashint = hashint ^ character;
         hashint += (hashint << 1) + (hashint << 4) + (hashint << 7) + (hashint << 8) + (hashint << 24);
     }
 
     return hashint >>> 0;
+    /* tslint:enable:no-bitwise */
 };
 
 export const getTemplateNameFromNode = (node: NodeStatus): string => {
--- a/ui/src/app/shared/pod-name.ts
+++ b/ui/src/app/shared/pod-name.ts
@@ -9,7 +9,7 @@ export const k8sNamingHashLength = 10;
 // getPodName returns a deterministic pod name
 // In case templateName is not defined or that version is explicitly set to  POD_NAME_V1, it will return the nodeID (v1)
 // In other cases it will return a combination of workflow name, template name, and a hash (v2)
-// note: this is inteded to be equivalent to the server-side Go code in workflow/util/pod_name.go
+// note: this is intended to be equivalent to the server-side Go code in workflow/util/pod_name.go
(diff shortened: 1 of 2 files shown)


terrytangyuan (commented):
@toyamagu-2021 Can you help test this change?

ui/src/app/shared/pod-name.ts
```diff
@@ -9,6 +9,7 @@ export const k8sNamingHashLength = 10;
 // getPodName returns a deterministic pod name
 // In case templateName is not defined or that version is explicitly set to  POD_NAME_V1, it will return the nodeID (v1)
 // In other cases it will return a combination of workflow name, template name, and a hash (v2)
+// note: this is inteded to be equivalent to the server-side Go code in workflow/util/pod_name.go
```
agilgur5, line 12:
```suggestion
// note: this is intended to be equivalent to the server-side Go code in workflow/util/pod_name.go
```

typo

toyamagu-2021 (approved):
Thanks! Works fine.

But I wonder why we should calculate podName by nix hash in frontend?
We may derive podName from nodeId...

`nodeId`=`steps-dflj9-1988631110`
`podName`=`steps-dflj9-whalesay-1988631110`

terrytangyuan (approved):

agilgur5:
> But I wonder why we should calculate podName by nix hash in frontend? We may derive podName from nodeId...
> 
> `nodeId`=`steps-dflj9-1988631110` `podName`=`steps-dflj9-whalesay-1988631110`

[the logic](https://github.com/agilgur5/argo-workflows/blob/38a09001bebcdece74b1a51f0feedef7ec5fc7ff/ui/src/app/shared/pod-name.ts#L18), as far as I can tell, is that when the name is too long it needs to be shortened, so a hash is used. The back-end PR is #6712 which also has a few back-links to other issues and PRs. Could also ask JP if he remembers

I've definitely had some long names before and it's common to do some truncation or hashing when names are too long for k8s's limit. So it makes sense to me.

toyamagu-2021:
Thanks for clarifying!
Now also makes sense to me.

Issue #9295, opened 2022-08-04
Artifact GC performance improvement - delete artifacts in parallel


juliev0:
The current "argoexec delete" command operates serially. We can improve performance by using workers that run in parallel.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

juliev0:
@dpadhiar You can model the code in agent.go [here](https://github.com/argoproj/argo-workflows/blob/master/workflow/executor/agent.go#L85) which starts up workers running in parallel and uses channels to communicate with them. (We can discuss more in detail if you like when you start this.)

juliev0:
I unassigned you, @dpadhiar, in case somebody else wants to pick this up (since we haven't been working on it)

PR #11768, opened 2023-09-06
feat: Parallel artifact GC

Joibel:
Parallelise deletion of artifacts when they are being garbage collected.

This is theoretically a transparent change and hence can't be integration tested.

Addresses #9295

Tested on minio and real S3

### DIFF
--- a/cmd/argoexec/commands/artifact/delete.go
+++ b/cmd/argoexec/commands/artifact/delete.go
@@ -71,20 +71,24 @@ func deleteArtifacts(labelSelector string, ctx context.Context, artifactGCTaskIn
 	if err != nil {
 		return err
 	}
-	taskWorkers := env.LookupEnvIntOr(common.EnvExecGCWorkers, 4)
+	gcWorkers := env.LookupEnvIntOr(common.EnvExecArtifactGCWorkers, 4)
 
 	totalTasks := 0
 	for _, task := range taskList.Items {
 		totalTasks += len(task.Spec.ArtifactsByNode)
 	}
+	// Channel for dispatching GC work to the workers
+	// large enough for every single task
 	taskQueue := make(chan *request, totalTasks)
+	// Channel for the replies when they complete
 	responseQueue := make(chan response, totalTasks)
-	for i := 0; i < taskWorkers; i++ {
+	for i := 0; i < gcWorkers; i++ {
 		go deleteWorker(ctx, taskQueue, responseQueue)
 	}
 
 	taskno := 0
 	nodesToGo := make(map[*v1alpha1.WorkflowArtifactGCTask]int)
+	// Dispatch all GC tasks to workers
 	for _, task := range taskList.Items {
 		nodesToGo[&task] = len(task.Spec.ArtifactsByNode)
 		task.Status.ArtifactResultsByNode = make(map[string]v1alpha1.ArtifactResultNodeStatus)
@@ -95,17 +99,27 @@ func deleteArtifacts(labelSelector string, ctx context.Context, artifactGCTaskIn
 	}
 	close(taskQueue)
 	completed := 0
+	// And now process all the responses as they are completed
 	for {
 		response := <-responseQueue
+		// A worker has had an error, we abort collecting any more information
+		// This is probably not ideal.
 		if response.Err != nil {
 			return response.Err
 		}
+		// If we get a nil response this means a worker has reached the end of the queued
+		// work, so keep a record and complete if all workers are done.
 		if response.Task == nil {
 			completed++
-			if completed >= taskWorkers {
+			if completed >= gcWorkers {
 				break
 			}
 		} else {
+			if response.Task == nil {
+				// Internal error, this shouldn't happen
+				return err
+			}
+			// Process the response
 			response.Task.Status.ArtifactResultsByNode[response.NodeName] = v1alpha1.ArtifactResultNodeStatus{ArtifactResults: response.Results}
 			// Check for completed tasks
 			nodesToGo[response.Task]--
@@ -129,7 +143,7 @@ func deleteWorker(ctx context.Context, taskQueue chan *request, responseQueue ch
 	for {
 		item, ok := <-taskQueue
 		if !ok {
-			// Done
+			// Report a done to the response queue
 			responseQueue <- response{Task: nil, NodeName: "", Err: nil}
 			return
 		}
@@ -151,6 +165,7 @@ func deleteWorker(ctx context.Context, taskQueue chan *request, responseQueue ch
 			}
 			drv, err := executor.NewDriver(ctx, &artifact, resources)
 			if err != nil {
+				// Report an error and process next item
 				responseQueue <- response{Task: item.Task, NodeName: item.NodeName, Results: results, Err: err}
 				continue
 			}
@@ -163,9 +178,15 @@ func deleteWorker(ctx context.Context, taskQueue chan *request, responseQueue ch
 					return false, err
 				}
 				results[artifact.Name] = v1alpha1.ArtifactResult{Name: artifact.Name, Success: true, Error: nil}
-				return true, err
+				return true, nil
 			})
+			if err != nil {
+				// Report an error and process next item
+				responseQueue <- response{Task: item.Task, NodeName: item.NodeName, Results: results, Err: err}
+				continue
+			}
 		}
+		// Report a success to the response queue
 		responseQueue <- response{Task: item.Task, NodeName: item.NodeName, Results: results, Err: nil}
 	}
 }
--- a/cmd/argoexec/commands/artifact/delete.go
+++ b/cmd/argoexec/commands/artifact/delete.go
@@ -16,6 +16,7 @@ import (
 	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	workflow "github.com/argoproj/argo-workflows/v3/pkg/client/clientset/versioned"
 	wfv1alpha1 "github.com/argoproj/argo-workflows/v3/pkg/client/clientset/versioned/typed/workflow/v1alpha1"
+	"github.com/argoproj/argo-workflows/v3/util/env"
 	"github.com/argoproj/argo-workflows/v3/util/retry"
 	waitutil "github.com/argoproj/argo-workflows/v3/util/wait"
 	executor "github.com/argoproj/argo-workflows/v3/workflow/artifacts"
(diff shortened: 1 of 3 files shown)


terrytangyuan (commented):
@juliev0 Could you help review this?

juliev0:
> @juliev0 Could you help review this?

Certainly. Will try to do today or tomorrow.

Joibel:
This is missing documentation as it is. I'm not sure how it undrafted itself. I'll try and get that done this evening.

agilgur5:
> This is missing documentation as it is.

Was gonna mention docs as there's a new number of workers. Potentially the [new scaling docs](https://argoproj.github.io/argo-workflows/scaling/#adding-goroutines-to-increase-concurrency) from #11731 could be a good place to add.

Also the CI [failure](https://github.com/argoproj/argo-workflows/actions/runs/6102068755/job/16559829554?pr=11768) is on `TestArtifactGC`, so I imagine it's related

Joibel:
The test failures aren't happening locally, but I have an idea.

The new variable isn't configurable - the pod that runs artifact GC is completely hardcoded at the moment. I suggest a separate PR to add configurability to it.

juliev0:
> The test failures aren't happening locally, but I have an idea.
> 
> The new variable isn't configurable - the pod that runs artifact GC is completely hardcoded at the moment. I suggest a separate PR to add configurability to it.

[This](https://github.com/argoproj/argo-workflows/pull/11586/files#diff-dbb0465647629d13c4dc6925a35205e042f03c61b5b6799d5fe9450dc14aa4fd) PR was put in recently to add configurability for resources in the Pod. Is this what you mean?

Joibel:
> > The test failures aren't happening locally, but I have an idea.
> > 
> > The new variable isn't configurable - the pod that runs artifact GC is completely hardcoded at the moment. I suggest a separate PR to add configurability to it.
> 
> [This](https://github.com/argoproj/argo-workflows/pull/11586/files#diff-dbb0465647629d13c4dc6925a35205e042f03c61b5b6799d5fe9450dc14aa4fd) PR was put in recently to add configurability for resources in the Pod. Is this what you mean?

Yep, I missed that PR. That's great, saves me a job.

juliev0:
Looking at the e2e test failure, I'm thinking it probably doesn't have anything to do with your code. This is the test that doesn't delete any artifacts. It looks like it waits for 1 minute to verify that the finalizer gets removed. I wonder if there is potential for this not to be long enough...if it just never did the Workflow reconciliation because maybe it was CPU starved in the CI or something?

If you do a series of empty commits, does it happen again?

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -52,63 +53,121 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+}
 
+func deleteArtifacts(labelSelector string, ctx context.Context, artifactGCTaskInterface wfv1alpha1.WorkflowArtifactGCTaskInterface) error {
 	taskList, err := artifactGCTaskInterface.List(context.Background(), metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
 		return err
 	}
+	taskWorkers := env.LookupEnvIntOr(common.EnvExecGCWorkers, 4)
+
+	totalTasks := 0
```
juliev0, line 76: I wonder if a different term than "task" could be used, since it's a little confusing with the existing use of `task` to mean something different in the code.

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -52,63 +53,121 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+				if err != nil {
+					errString := err.Error()
+					results[artifact.Name] = v1alpha1.ArtifactResult{Name: artifact.Name, Success: false, Error: &errString}
+					return false, err
+				}
+				results[artifact.Name] = v1alpha1.ArtifactResult{Name: artifact.Name, Success: true, Error: nil}
+				return true, err
+			})
 		}
+		responseQueue <- response{Task: item.Task, NodeName: item.NodeName, Results: results, Err: nil}
```
juliev0, line 169: I know the existing code doesn't have much commenting, but can we add more now?

juliev0 (commented):
I'm curious if you tried testing with multiple `WorkflowArtifactGCTasks`? I know we never actually enabled the creation of multiple but there was always the possibility.

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -52,63 +53,121 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+			err = waitutil.Backoff(retry.DefaultRetry, func() (bool, error) {
+				err = drv.Delete(&artifact)
+				if err != nil {
+					errString := err.Error()
+					results[artifact.Name] = v1alpha1.ArtifactResult{Name: artifact.Name, Success: false, Error: &errString}
+					return false, err
+				}
+				results[artifact.Name] = v1alpha1.ArtifactResult{Name: artifact.Name, Success: true, Error: nil}
+				return true, err
+			})
```
juliev0, line 167: It looks like this is an error with the original code: it's not doing anything if `err` returns true

workflow/common/common.go
```diff
@@ -159,6 +159,9 @@ const (
 	// EnvAgentPatchRate is the rate that the Argo Agent will patch the Workflow TaskSet
 	EnvAgentPatchRate = "ARGO_AGENT_PATCH_RATE"
 
+	// EnvExecGCWorkers is the number of artifact GC workers for argoexec
+	EnvExecGCWorkers = "ARGO_EXEC_GC_WORKERS"
+
```
juliev0, line 164: I don't think there's a code path that will apply this to the ArtifactGCPod. Somebody would have to enable environment variable specification as part of the `ArtifactGC` spec. Were you planning on adding that?

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -52,63 +53,121 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+	Results  map[string]v1alpha1.ArtifactResult
+	Err      error
+}
 
+func deleteArtifacts(labelSelector string, ctx context.Context, artifactGCTaskInterface wfv1alpha1.WorkflowArtifactGCTaskInterface) error {
 	taskList, err := artifactGCTaskInterface.List(context.Background(), metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
 		return err
 	}
+	taskWorkers := env.LookupEnvIntOr(common.EnvExecGCWorkers, 4)
```
isubasinghe, line 74: Should we look into how many cores are available and base the parallelism off that ? 
I can imagine some deployments being backed by some heavy servers. 


```python
if log2(cores) + 4 >= cores:
  return cores
else:
  return log2(cores) + 4
```

Something like the above growth rate could make sense because it doesn't scale with cores linearly. 

Not even sure if this is good idea, just a suggestion.

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -52,63 +53,121 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+		}
+		if response.Task == nil {
+			completed++
+			if completed >= taskWorkers {
+				break
 			}
+		} else {
+			response.Task.Status.ArtifactResultsByNode[response.NodeName] = v1alpha1.ArtifactResultNodeStatus{ArtifactResults: response.Results}
+			// Check for completed tasks
+			nodesToGo[response.Task]--
```
isubasinghe, line 111: Do you think we need to add an ok check here? I see that this should obviously exist, just tempted to be more defensive so that we future proof this a bit more. 

Alternatively add a comment that we expect the response.Task to be present here because it was populated in the previous loop, probably should be accompanied by a comment at line 89 where it gets populated. 

Sorry for being pedantic/annoying when it is obvious, I just don't want things to go in the way of NodeStatus, tempted to use a linter to force ok checks everywhere to be honest.

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -52,63 +53,121 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+					continue
+				}
+			}
+			drv, err := executor.NewDriver(ctx, &artifact, resources)
+			if err != nil {
+				responseQueue <- response{Task: item.Task, NodeName: item.NodeName, Results: results, Err: err}
+				continue
+			}
+
+			err = waitutil.Backoff(retry.DefaultRetry, func() (bool, error) {
```
isubasinghe, line 158: As @juliev0 states, the err is not handled

workflow/common/common.go
```diff
@@ -159,6 +159,9 @@ const (
 	// EnvAgentPatchRate is the rate that the Argo Agent will patch the Workflow TaskSet
 	EnvAgentPatchRate = "ARGO_AGENT_PATCH_RATE"
 
+	// EnvExecGCWorkers is the number of artifact GC workers for argoexec
+	EnvExecGCWorkers = "ARGO_EXEC_GC_WORKERS"
```
isubasinghe, line 163: imo if its only used for artifact gc perhaps the env name should also carry this information? Just to be more obvious to anyone who is looking at a .env file. 
`ARGO_EXEC_ARTI_GC_WORKERS` ?

isubasinghe (commented):
Generally looks great, the error isn't handled. I've added more comments, let me know what you think :)

cmd/argoexec/commands/artifact/delete.go
Joibel: Done

cmd/argoexec/commands/artifact/delete.go
Joibel: This is deployed as a separate pod, so users can tune resource limits and `ARGO_EXEC_GC_WORKERS` to fit. This is a heavily IO bound process as we're mostly just dispatching deletes to a remote blob store, and it's dependent on how that is configured how much it will benefit from parallelism, so attempting an autotune is probably not worthwhile.

isubasinghe: Makes sense, yeah I can imagine this being majorly bound on the network. Makes sense that an autotune isn't worthwhile.

cmd/argoexec/commands/artifact/delete.go
isubasinghe: Thanks!

workflow/common/common.go
isubasinghe: Nice

workflow/common/common.go
juliev0: nice!

cmd/argoexec/commands/artifact/delete.go
juliev0: sorry, is this the `if response.Task == nil` check? I'm actually kind of confused why that's in there. It seems redundant, no?

juliev0:
Sorry but would you mind adding a unit test?

Joibel:
I'll get back to this one day!

tooptoop4:
dis forgotten

Joibel:
> dis forgotten

Nope, but it's not a priority for me right now.

If someone wants to work on it they are welcome

Issue #10970, opened 2023-04-23
Reload Server configuration when `ConfigMap` is updated


agilgur5:
# Summary

Currently, when the `argo-workflows-workflow-controller-configmap` is updated, the configuration is not reloaded, and so I have to restart some `Deployment`s to get them to have the updated configuration.

# Use Cases

For instance, when I was working on #10927, I had to change the `scopes` field, and this did not automatically apply in the Argo Server. For it to apply, I had to restart the Server Deployment. While this works, it can lead to a gotcha for some operators.

Automatically restarting `Deployment`s when a `ConfigMap` changes is one of the primary purposes of a tool like [Reloader](https://github.com/stakater/Reloader), but it feels like maybe this should be built into Argo in some way instead of requiring a different third-party tool that is not mentioned in any docs.

Let me know what you think. This could be considered a bug instead of a feature as well, depending on how you look at it.

# Root cause analysis

The config is generally reloaded when [`Get`](https://github.com/argoproj/argo-workflows/blob/e2657ad996f77edd8917be1ee883c37dda52f73c/config/controller.go#L55) is called, but I see for SSO for the Server, it is only configured once [during instantiation](https://github.com/argoproj/argo-workflows/blob/e2657ad996f77edd8917be1ee883c37dda52f73c/server/apiserver/argoserver.go#L141). As a result it is never reloaded if there is an update.

I'd be open to contributing this as well but may need a bit more guidance as this requires a bit of refactoring. And there are potentially multiple ways of achieving this with different trade-offs (e.g. `goroutine` to poll on configmap changes vs. always reading the configmap vs. caching etc) which would be for a maintainer to decide on the approach.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

tico24:
What workflows version are you running? I'm sure this was fixed in 3.4.5.

agilgur5:
I'm on 3.3.8. #10660 looks like it's for the controller, not the server. The latest server code I [linked to](https://github.com/argoproj/argo-workflows/blob/e2657ad996f77edd8917be1ee883c37dda52f73c/server/apiserver/argoserver.go#L141) above does not seem to ever refresh the config (e.g. for SSO config).

tico24:
Apologies, I mis-read.

terrytangyuan:
@agilgur5 Would you like to submit a PR?

agilgur5:
[...]

PR #11774, opened 2023-09-08
feat: filter sso groups based on regex

### LINKED ISSUE #10153: SSO does not work for users with a lot of groups due to exceeding max browser cookie size limit

basanthjenuhb:
Fixes #10153 #9530

### Motivation
- Currently users logging in via SSO with lot of groups, exceed the 4KB Cookie limit
- This happens because we send the full set of claims as part of cookie
- Adding capability to filter sso groups matching a regex.
- This would eliminate all SSO groups which won't be used in SSO RBAC and reduce the number of groups in the authorization cookie

### Modifications
- Add a property `filterGroupsRegex ` in SSO config
- FIlter groups based on the property

### Verification
- Tested locally with `filterGroupsRegex `

### Example usage:
```
scopes:
- groups
- email
- profile
rbac:
  enabled: true
filterGroupsRegex:
- ".*argowf.*"
```

### DIFF
--- a/config/sso.go
+++ b/config/sso.go
@@ -18,10 +18,10 @@ type SSOConfig struct {
 	Scopes        []string        `json:"scopes,omitempty"`
 	SessionExpiry metav1.Duration `json:"sessionExpiry,omitempty"`
 	// customGroupClaimName will override the groups claim name
-	CustomGroupClaimName string `json:"customGroupClaimName,omitempty"`
-	UserInfoPath         string `json:"userInfoPath,omitempty"`
-	InsecureSkipVerify   bool   `json:"insecureSkipVerify,omitempty"`
-	FilterGroupsRegex    string `json:"filterGroupsRegex,omitempty"`
+	CustomGroupClaimName string   `json:"customGroupClaimName,omitempty"`
+	UserInfoPath         string   `json:"userInfoPath,omitempty"`
+	InsecureSkipVerify   bool     `json:"insecureSkipVerify,omitempty"`
+	FilterGroupsRegex    []string `json:"filterGroupsRegex,omitempty"`
 }
 
 func (c SSOConfig) GetSessionExpiry() time.Duration {
--- a/config/sso.go
+++ b/config/sso.go
@@ -21,6 +21,7 @@ type SSOConfig struct {
 	CustomGroupClaimName string `json:"customGroupClaimName,omitempty"`
 	UserInfoPath         string `json:"userInfoPath,omitempty"`
 	InsecureSkipVerify   bool   `json:"insecureSkipVerify,omitempty"`
+	FilterSSOGroupsRegex string `json:"filterSSOGroupsRegex,omitempty"`
 }
 
 func (c SSOConfig) GetSessionExpiry() time.Duration {
--- a/config/sso.go
+++ b/config/sso.go
@@ -21,7 +21,7 @@ type SSOConfig struct {
 	CustomGroupClaimName string `json:"customGroupClaimName,omitempty"`
 	UserInfoPath         string `json:"userInfoPath,omitempty"`
 	InsecureSkipVerify   bool   `json:"insecureSkipVerify,omitempty"`
-	FilterSSOGroupsRegex string `json:"filterSSOGroupsRegex,omitempty"`
+	FilterGroupsRegex    string `json:"filterGroupsRegex,omitempty"`
 }
 
 func (c SSOConfig) GetSessionExpiry() time.Duration {
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -62,7 +62,7 @@ type sso struct {
 	expiry            time.Duration
 	customClaimName   string
 	userInfoPath      string
-	filterGroupsRegex *regexp.Regexp
+	filterGroupsRegex []*regexp.Regexp
 }
 
 func (s *sso) IsRBACEnabled() bool {
@@ -187,11 +187,14 @@ func newSso(
 	if c.IssuerAlias != "" {
 		lf["issuerAlias"] = c.IssuerAlias
 	}
-	var filterGroupRegex *regexp.Regexp = nil
-	if c.FilterGroupsRegex != "" {
-		filterGroupRegex, err = regexp.Compile(c.FilterGroupsRegex)
-		if err != nil {
-			return nil, fmt.Errorf("failed to compile sso.filterGroupRegex: %s %w", c.FilterGroupsRegex, err)
+	var filterGroupRegex []*regexp.Regexp
+	if c.FilterGroupsRegex != nil && len(c.FilterGroupsRegex) > 0 {
+		for _, regex := range c.FilterGroupsRegex {
+			compiledRegex, err := regexp.Compile(regex)
+			if err != nil {
+				return nil, fmt.Errorf("failed to compile sso.filterGroupRegex: %s %w", regex, err)
+			}
+			filterGroupRegex = append(filterGroupRegex, compiledRegex)
 		}
 	}
 	log.WithFields(lf).Info("SSO configuration")
@@ -290,11 +293,14 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	if s.filterGroupsRegex != nil {
+	if s.filterGroupsRegex != nil && len(s.filterGroupsRegex) > 0 {
 		var filteredGroups []string
 		for _, group := range groups {
-			if s.filterGroupsRegex.Match([]byte(group)) {
-				filteredGroups = append(filteredGroups, group)
+			for _, regex := range s.filterGroupsRegex {
+				if regex.MatchString(group) {
+					filteredGroups = append(filteredGroups, group)
+					break
+				}
 			}
 		}
 		groups = filteredGroups
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -187,14 +187,14 @@ func newSso(
 	if c.IssuerAlias != "" {
 		lf["issuerAlias"] = c.IssuerAlias
 	}
-	var filterGroupRegex []*regexp.Regexp
+	var filterGroupsRegex []*regexp.Regexp
 	if c.FilterGroupsRegex != nil && len(c.FilterGroupsRegex) > 0 {
 		for _, regex := range c.FilterGroupsRegex {
 			compiledRegex, err := regexp.Compile(regex)
 			if err != nil {
 				return nil, fmt.Errorf("failed to compile sso.filterGroupRegex: %s %w", regex, err)
 			}
-			filterGroupRegex = append(filterGroupRegex, compiledRegex)
+			filterGroupsRegex = append(filterGroupsRegex, compiledRegex)
 		}
 	}
 	log.WithFields(lf).Info("SSO configuration")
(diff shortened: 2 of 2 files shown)


config/sso.go
```diff
@@ -21,6 +21,7 @@ type SSOConfig struct {
 	CustomGroupClaimName string `json:"customGroupClaimName,omitempty"`
 	UserInfoPath         string `json:"userInfoPath,omitempty"`
 	InsecureSkipVerify   bool   `json:"insecureSkipVerify,omitempty"`
+	FilterSSOGroupsRegex string `json:"filterSSOGroupsRegex,omitempty"`
```
agilgur5, line 24:
```suggestion
	FilterGroupsRegex string `json:"filterGroupsRegex,omitempty"`
```

"SSO" is redundant here as this is a field of `SSOConfig`

agilgur5 (commented):
Would an allowlist of groups suffice as well? Or is a regex necessary for the amount of relevant groups?

server/auth/sso/sso.go
```diff
@@ -280,6 +283,20 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
+	if s.filterSsoGroupsRegex != "" {
+		var filteredGroups []string
+		regex, err := regexp.Compile(s.filterSsoGroupsRegex)
+		if err != nil {
+			log.WithError(err).Errorf("failed to compile filterSsoGroupsRegex: %s", s.filterSsoGroupsRegex)
```
agilgur5, line 290: I think it would make more sense to compile this on start-up and immediately error out then as opposed to erroring out much later when a request is made.

basanthjenuhb:
> Would an allowlist of groups suffice as well? Or is a regex necessary for the amount of relevant groups?

Reason for regex is to be able to add groups dynamically. And be able to filter our groups which will be relevant for argo SSO rbac

with `allowlist` we would need to update config and restart everytime we want to add a new group

config/sso.go
basanthjenuhb: done

juliev0 (commented):
Good idea. So, for a given cluster there would be only one particular regex that can be matched, right? I wonder if we should make it an array of regex?

sarabala1979 (approved):
LGTM

sarabala1979:
@basanthjenuhb Can you add the document for this?
@terrytangyuan can we include this feature in v3.5?

basanthjenuhb:
> Good idea. So, for a given cluster there would be only one particular regex that can be matched, right? I wonder if we should make it an array of regex?

yeah
we could make it an array
but wanted to keep it simple and see how it works out for us
we can take it up as an enhancement after we see this one’s usage

basanthjenuhb:
> @basanthjenuhb Can you add the document for this? @terrytangyuan can we include this feature in v3.5?

https://github.com/argoproj/argo-workflows/pull/11778

agilgur5 (commented):
> So, for a given cluster there would be only one particular regex that can be matched, right? I wonder if we should make it an array of regex?

Julie brings up a good point here

> we could make it an array
> but wanted to keep it simple and see how it works out for us
> we can take it up as an enhancement after we see this one’s usage

Adding a feature afterward is a bit easier said than done -- there would then be two features that need to be maintained for backward-compatibility. 

If we can get ahead of a problem now, it would certainly aid maintainability

agilgur5:
> with `allowlist` we would need to update config

correct. that is not necessarily a bad thing though.

> restart everytime we want to add a new group

there is a separate issue to improve that: #10970

agilgur5:
> > So, for a given cluster there would be only one particular regex that can be matched, right? I wonder if we should make it an array of regex?
> 
> Julie brings up a good point here

Wait a minute, maybe I spoke too soon. This is a regex after all, you can do "or" conditions within a regex. Would that suffice? That would actually be simpler than having to tell users whether the list is "or" vs. "and"

Sorry I haven't been home all day, responded from my phone. Apologies if this results in rework 🙇

juliev0:
> > > So, for a given cluster there would be only one particular regex that can be matched, right? I wonder if we should make it an array of regex?
> > 
> > 
> > Julie brings up a good point here
> 
> Wait a minute, maybe I spoke too soon. This is a regex after all, you can do "or" conditions within a regex. Would that suffice? That would actually be simpler than having to tell users whether the array is "or" vs. "and"
> 
> Sorry I haven't been home all day, responded from my phone. Apologies if this results in rework 🙇 Will take a closer look later. Also want to check how Dex handles this case

You're right. Just looked that up. It's more doable than I thought it was.

basanthjenuhb:
> Wait a minute, maybe I spoke too soon. This is a regex after all, you can do "or" conditions within a regex. Would that suffice? That would actually be simpler than having to tell users whether the array is "or" vs. "and"

yes, I had looked it up
regex does support OR. it would look like `(.*patternA.*|.*patternB.*)`, although doable, its just not very clean compared to
```
filterGroupsRegex:
- ".*patternA.*"
- ".*patternB.*"
```
I would prefer list over single regex if the idea is to support multiple patterns.
anyways, let me know your preference. Will make the changes accordingly

juliev0:
> > Wait a minute, maybe I spoke too soon. This is a regex after all, you can do "or" conditions within a regex. Would that suffice? That would actually be simpler than having to tell users whether the array is "or" vs. "and"
> 
> yes, I had looked it up regex does support OR. it would look like `(.*patternA.*|.*patternB.*)`, although doable, its just not very clean compared to
> 
> ```
> filterGroupsRegex:
> - ".*patternA.*"
> - ".*patternB.*"
> ```
> 
> I would prefer list over single regex if the idea is to support multiple patterns. anyways, let me know your preference. Will make the changes accordingly

I agree if you’re open to changing it. Thanks

basanthjenuhb:
> > > Wait a minute, maybe I spoke too soon. This is a regex after all, you can do "or" conditions within a regex. Would that suffice? That would actually be simpler than having to tell users whether the array is "or" vs. "and"
> > 
> > 
> > yes, I had looked it up regex does support OR. it would look like `(.*patternA.*|.*patternB.*)`, although doable, its just not very clean compared to
> > ```
> > filterGroupsRegex:
> > - ".*patternA.*"
> > - ".*patternB.*"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I would prefer list over single regex if the idea is to support multiple patterns. anyways, let me know your preference. Will make the changes accordingly
> 
> I agree if you’re open to changing it. Thanks

already did

juliev0 (approved):

server/auth/sso/sso.go
```diff
@@ -185,21 +187,32 @@ func newSso(
...
+	if c.FilterGroupsRegex != nil && len(c.FilterGroupsRegex) > 0 {
+		for _, regex := range c.FilterGroupsRegex {
+			compiledRegex, err := regexp.Compile(regex)
+			if err != nil {
+				return nil, fmt.Errorf("failed to compile sso.filterGroupRegex: %s %w", regex, err)
+			}
+			filterGroupsRegex = append(filterGroupsRegex, compiledRegex)
+		}
+	}
 	log.WithFields(lf).Info("SSO configuration")
```
agilgur5, line 200: `filterGroupsRegex` should also be added to `lf` above.

or perhaps better, the whole code block should be moved above the `lf` assignment as those statements were intended to be together. this got spliced in between the `lf` code block

can put newlines around this code block and that code block so that it is more clear that they are blocks of related code

server/auth/sso/sso.go
```diff
@@ -280,6 +293,18 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
+	if s.filterGroupsRegex != nil && len(s.filterGroupsRegex) > 0 {
```
agilgur5, line 296:
```suggestion
        // only return groups that match at least one of the regexes
	if s.filterGroupsRegex != nil && len(s.filterGroupsRegex) > 0 {
```
a comment would be useful to describe the logic here. that would be consistent with the above code and help readers as the code here is very nested

agilgur5 (commented):
there's some stylistic issues in the code (it was spliced in between another block of related code), see below

agilgur5:
> I agree if you’re open to changing it. Thanks

I don't have a strong opinion on this so long as the documentation makes clear that the list is "OR"'d together. Otherwise it may be ambiguous to users. I mentioned this in my [docs review](https://github.com/argoproj/argo-workflows/pull/11778#discussion_r1320646434)

server/auth/sso/sso.go
```diff
@@ -280,6 +293,18 @@ func (s *sso) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+		var filteredGroups []string
```
agilgur5, line 297: it would be great if the provider could do this logic so that the response from it is not so large, but it does not seem there is a uniform provider interface for that.

Azure AD implements [group filtering per app](https://learn.microsoft.com/en-us/azure/active-directory/hybrid/connect/how-to-connect-fed-group-claims#group-filtering)
LDAP more generally has [group search](https://dexidp.io/docs/connectors/ldap/).

There is a related Dex issue on this (https://github.com/dexidp/dex/issues/1476) and generalizing it was basically closed out due to provider-specific nuances.

So I think there may be no way around filtering in the Server code

server/auth/sso/sso.go
basanthjenuhb: done

server/auth/sso/sso.go
basanthjenuhb: done

server/auth/sso/sso.go
```diff
@@ -181,25 +183,41 @@ func newSso(
...
+		}
+	}
+
 	lf := log.Fields{"redirectUrl": config.RedirectURL, "issuer": c.Issuer, "issuerAlias": "DISABLED", "clientId": c.ClientID, "scopes": config.Scopes, "insecureSkipVerify": c.InsecureSkipVerify}
 	if c.IssuerAlias != "" {
 		lf["issuerAlias"] = c.IssuerAlias
 	}
+	if c.FilterGroupsRegex != nil && len(c.FilterGroupsRegex) > 0 {
+		lf["filterGroupsRegex"] = c.FilterGroupsRegex
+	}
```
agilgur5, line 204: I believe this can just be added to `lf` directly. `issuerAlias` is the odd one out since it logs out "DISABLED", whereas the rest log out the field directly.

`scopes` above is similarly a `[]string` (i.e. same as `filterGroupsRegex`), with regard to typing

basanthjenuhb: done

agilgur5 (approved):
LGTM. Thanks for iterating on this!

terrytangyuan:
> @terrytangyuan can we include this feature in v3.5?

@sarabala1979 Sorry just saw it. Let's discuss this in the next contributors meeting on v3.5 release strategy. https://github.com/argoproj/argo-workflows/issues/11381

PR #11775, opened 2023-09-08
chore: pkg imported more than once

testwill:
Fixes #TODO


agilgur5 (approved):
Nice catch!

Surprised `gofmt` doesn't do this something like this 🤔

terrytangyuan (approved):

PR #11788, opened 2023-09-09
feat: View the history of completions in cron-workflows

yunwoo-yu:
Fixes #11706 

### Motivation

- You can view completed workflows on the cron-workflows detail page.
- You can explore workflows that you want users to see more easily. Better UX
- We've taken the component that was applied to the workflow view and reused it.

### Modifications

- before
image

- after
image

Issue #11740, opened 2023-09-04
UI: Refactor to use async/await syntax


agilgur5:
# Summary

Instead of using promises with many callbacks and chains, refactor to use the newer and cleaner async/await syntax where possible.

As mentioned in #11711

# Use Cases

All UI logic that uses asynchronous promise chains

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
A quick search on [all appearances of `.then` in the UI](https://github.com/search?q=repo%3Aargoproj%2Fargo-workflows+%22.then%22+path%3Aui%2F&type=code) showed mostly correct async code. There was one case that was parallelized and async with a render, but not sure if that was intentional or not (as it could potentially work fine that way). Some large chains though.

But the case of #11711 actually doesn't have a `.then` and a request could just end in a `.get` or `.list` etc, so that search above is a bit naive, but likely gets the majority of cases.

Issue #10846, opened 2023-04-06
Pagination for Cron workflows


terrytangyuan:
Reference implementation: https://github.com/argoproj/argo-workflows/pull/7163

Joibel:
Still valid on main - the cron workflow list page loads every CronWorkflow in one request, with no limit/continue.

This is a UI-only change: `ListCronWorkflows` already forwards `ListOptions` to the Kubernetes client (`server/cronworkflow/cron_workflow_server.go`), so `limit`/`continue` work server-side today. The work is to mirror what `ui/src/workflow-templates/workflow-template-list.tsx` does (see #7163): add a `Pagination` argument to `CronWorkflowService.list` in `ui/src/shared/services/cron-workflow-service.ts`, and add the pagination state, `<PaginationPanel>` and URL query params to `ui/src/cron-workflows/cron-workflow-list.tsx`.

One thing to be aware of: the state filter (Running/Suspended) and the label options in `cron-workflow-filters.tsx` are currently computed client-side from the whole list, so once pagination is in they will only see the current page - the same caveat that applies to the workflows list.

Krupesh-21:
Hey @Joibel, I'd like to work on this issue - could you assign it to me? I've looked at the reference implementation in #7163 and plan to follow the same pagination pattern for the CronWorkflows list.

Joibel:
Assigned: There is a similar #10845 if you'd like a follow up.

Krupesh-21:
> Assigned: There is a similar [#10845](https://github.com/argoproj/argo-workflows/issues/10845) if you'd like a follow up.

Happy to pick up #10845 too once this one's merged, if it's still open.

Bryandero98:
Hey @Krupesh-21, checking in — are you still working on this? It's been about 40 days since you were assigned with no update, and I'd be happy to pick it up (following the same pattern as #7163, per @Joibel's notes above) if you've moved on to something else. No rush either way, just want to avoid duplicate effort.

PR #11790, opened 2023-09-09
feat: show history about completed runs in each cron workflow

### LINKED ISSUE #11706: UI - show history about completed runs in each cron workflow
# Summary

We can view completed cron-workflows in Workflow view. 
However, I suggest an UI enhancement to view the history about completed runs in each cron workflow.

<br>

**Example**
image
If You click a cron-workflow, 

image
You can look history of completed runs about the cron-workflows.

<br>

# Use Cases

It is easy to view completed runs of each cron workflow.

---

yunwoo-yu:
Fixes #11706 

### Motivation

- You can view completed workflows on the cron-workflows detail page.
- You can explore workflows that you want users to see more easily. Better UX
- We've taken the component that was applied to the workflow view and reused it.

### Modifications

- before
image

- after
image


ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -62,6 +73,16 @@ export const CronWorkflowDetails = ({match, location, history}: RouteComponentPr
 
     useEffect(() => setEdited(true), [cronWorkflow]);
 
+    useEffect(() => {
+        services.workflows.list(namespace, selectedPhases, selectedLabels, pagination).then(res => {
```
agilgur5, line 77: Ideally we can filter in the request to not retrieve unnecessary workflows, although that might not be possible

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -25,6 +30,12 @@ export const CronWorkflowDetails = ({match, location, history}: RouteComponentPr
     const [name] = useState(match.params.name);
     const [sidePanel, setSidePanel] = useState(queryParams.get('sidePanel'));
     const [tab, setTab] = useState(queryParams.get('tab'));
+    const [workflows, setWorkflows] = useState<Workflow[]>([]);
+    const [{selectedPhases, selectedLabels, paginationLimit}, setSelectedData] = useState<WorkflowListRenderOptions>(JSON.parse(localStorage.getItem('ListOptions/options')));
```
agilgur5, line 34: there are no phases or labels here. those are from the sidebar filter and the Workflows list which is not present here.

pagination does not seem to be present either. there seems to be some logic here, but no UI elements for pagination.

also the `localStorage` call would not apply here as it is specific to the Workflows list page

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -183,6 +204,17 @@ export const CronWorkflowDetails = ({match, location, history}: RouteComponentPr
 
         return items;
     };
+    const updateCurrentlySelectedAndBatchActions = (newSelectedWorkflows: Map<string, Workflow>): void => {
```
agilgur5, line 207: So the checkbox is unused here. No batch actions were added to this page, so it's not useable currently

agilgur5 (changes requested):
Thanks for implementing this!

There are some issues with the implementation right now, as not everything is transferable from the Workflows List. See some comments below

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
agilgur5: this is possible -- you can pass in a label with the Cron Workflow name: `workflows.argoproj.io/cron-workflow=<name-of-cron-workflow>`

agilgur5: also should use async/await per #11740

yunwoo-yu: - Thanks for the feedback, would it be ok to change to async/await and get the data via label names like on the workflows List page?

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
yunwoo-yu: - Thanks I was wondering this, would it be a good idea to put a selection or pagination in the detail cron page?

yunwoo-yu (commented):
Thanks for the feedback, I've got a few questions and will try to revise it in the direction of your feedback 👍🙇‍♂️

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
yunwoo-yu: - The `workflows-row` component has fixed selection and checkboxes. What do you think about creating a new `cron-workflws-row` component?

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
agilgur5: You can replace `selectedLabels` (the 3rd argument to [`list`](https://github.com/argoproj/argo-workflows/blob/477b3caf415d1f65f71dd366d9ebc5c04c64c099/ui/src/app/shared/services/workflows-service.ts#L42)) with the `cron-workflow` label from above

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
agilgur5: I don't think filtering is needed right now. pagination would be nice to have, but for a first-pass MVP, I think we can leave out pagination as well.

You may have not noticed, but the Cron Workflows page itself actually doesn't implement pagination yet 😅 : #10846

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
agilgur5: So ideally we would factor out a `workflows-row` without checkboxes and then use that in both places.
In particular, this history view you've created could also be useful for Workflow Templates as well, so making it reusable would be useful.

That being said, that requires a bit of refactoring. As a first-pass, we could just make the checkboxes no-op for now and refactor later -- that was why I commented on the function itself and not the UI elements.
I'll leave that decision up to you for if you want to try to tackle a refactor or just leave the checkboxes non-functional for now.

yunwoo-yu:
@agilgur5

hello 😄

I'll hide the checkbox for now, thanks for the advice

The PR got stuck in the commit log, so I'll repromote it again

Thanks!

agilgur5:
> The PR got stuck in the commit log, so I'll repromote it again

You can rebase the branch and force push. No need to open a new PR

yunwoo-yu:
Uncontributed commits from collaborators got mixed up in the middle, and it was really hard to pick them out 😅

And I deleted the already forked repository. Maybe juijeong8324 will PR it again!

agilgur5:
@yunwoo-yu I noticed in this PR and in #11788 that you used your fork's `master` branch. You usually want to use [feature branches](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow). That makes it easier to separate your work and could make it easier to rebase and cherry-pick as well.

yunwoo-yu:
@agilgur5 

For the next PR, I'll create a separate feat branch and upload it. Thanks 😁

And like you said, I was able to get rid of the merged fact with reflog. I had already deleted the repository so I didn't get to try it, but I'll do it the way you suggested next time. 👍

PR #11793, opened 2023-09-10
refactor(ui): migrate `UserInfo` to functional component

agilgur5:
Partial fix for #9810 
Partial fix for #11740 

### Motivation

- Use newer, modern functional components compatible with hooks instead of legacy class-based components

- Use newer, modern async/await syntax instead of Promise chains

- also remove a deprecated usage of `BasePage`
  - `BasePage` is a deprecated usage of `react-router`'s private context before React made context an official API (it's quite old)
    - removing this is necessary to upgrade `react-router-dom` as the old, private context no longer exists in new versions
  - I'm also in process of migrating its remaining usages to functional components, this is one of a few PRs to do so (the others are _significantly_ more complex, so I have split them out. See #11794)

### Modifications

- mostly straightforward state refactor from `this.state` to `useState`
  - plus refactor `componentDidMount` to `useEffect`

- remove deprecated `BasePage` `extends` and typings
  - it wasn't even used in `UserInfo` (or at least not currently), so just removed references
    - no routing primitives are used in `UserInfo`

- plus remove unused `BasePage` methods
  - these are not used by anything at this time
    - searched the codebase and found 0 references

### Verification

- Pure refactor, no real semantic changes
- Manually tested locally that the UserInfo page works the same as it did before

### Notes for Reviewers

Turn on ["ignore whitespace"](https://github.com/argoproj/argo-workflows/pull/11793/files?w=1) to make the diff much simpler to read

Issue #9810, opened 2022-10-13
Refactor UI to use functional component


tczhao:
# Summary

Say for UI features, new feature added in workflow page would likely be useful in archived workflow as well,
but right now we have both class and functional components at various parts of the UI, porting features from one part to another can be a pain.

class component file list

- [x] 1. ui/src/app/app.tsx:
- [ ] 2. ui/src/app/archived-workflows/components/archived-workflow-details/archived-workflow-details.tsx:
- [ ] 3. ui/src/app/archived-workflows/components/archived-workflow-filters/archived-workflow-filters.tsx:
- [ ] 4. ui/src/app/archived-workflows/components/archived-workflow-list/archived-workflow-list.tsx:
- [x] 5. ui/src/app/cluster-workflow-templates/components/cluster-workflow-template-list/cluster-workflow-template-list.tsx:
- [ ] 6. ui/src/app/reports/components/reports.tsx:
- [ ] 7. ui/src/app/shared/components/base-page.tsx:
- [ ] 8. ui/src/app/shared/components/checkbox-filter/checkbox-filter.tsx:
- [ ] 9. ui/src/app/shared/components/dropdown/dropdown.tsx:
- [ ] 10. ui/src/app/shared/components/error-panel.tsx:
- [ ] 11. ui/src/app/shared/components/graph/coffman-graham-sorter.ts:
- [ ] 12. ui/src/app/shared/components/graph/dfs-sorter.ts:
- [ ] 13. ui/src/app/shared/components/graph/types.ts:
- [ ] 14. ui/src/app/shared/components/inline-table/inline-table.tsx:
- [ ] 15. ui/src/app/shared/components/input-filter.tsx:
- [ ] 16. ui/src/app/shared/components/nudge.tsx:
- [ ] 17. ui/src/app/shared/components/pagination-panel.tsx:
- [ ] 18. ui/src/app/shared/components/resource-editor/resource-editor.tsx:
- [ ] 19. ui/src/app/shared/components/tags-input/tags-input.tsx:
- [ ] 20. ui/src/app/shared/conditions-panel.tsx:
- [ ] 21. ui/src/app/shared/list-watch.ts:
- [ ] 22. ui/src/app/shared/resources-duration.tsx:
- [ ] 23. ui/src/app/shared/retry-observable.ts:
- [ ] 24. ui/src/app/shared/retry-watch.ts:
- [ ] 25. ui/src/app/shared/scoped-local-storage.ts:
[...]

PR #11794, opened 2023-09-10
refactor(ui): migrate `Reports` to functional component and split files

agilgur5:
Partial fix for #9810 
Partial fix for #11740 

### Motivation

- Use newer, modern async/await syntax instead of Promise chains

- also remove a deprecated, upgrade-blocking usage of `BasePage` similar to #11793 

### Modifications

Follow a very similar process to #7035
  - `state` -> `useState`
  - `this.props` -> arguments
  - `BasePage` + `this.queryParams` -> `location` argument + `URLSearchParams`
  - `componentDidMount` -> `useEffect`
  - `services.info.collectEvent` -> `useCollectEvent`
  - `saveHistory` -> `useEffect` + `historyUrl`
  - `Consumer` -> `useContext`
  - merge `renderReport` into the main component
  - split out filters into a separate `ReportsFilters` component similar to `CronWorkflowFilters`, `WorkflowFilters`, etc
    - be consistent with how all these work
    - similar to #7035, refactor logic to have an `onChange` function and retrieve Workflows in a top-level effect (after all state has been set)
    - and similarly add an SCSS file with the appropriate styles
      - note: this may be worth refactoring as nearly all the filters have the same styles. comments in #7035 suggest making a shared filters component if possible too.

- Promise chains -> async/await
  - refactor `ReportsFilters` logic to use async/await syntax
  - also refactor `CronWorkflowList` to use newer async/await syntax consistent with #11740 
- getters and setters to regular functions
- split out a pure function into a util file
  - to make clear that it is pure and unrelated to React state, context, etc

Also some tiny optimizations similar to #11743:
- use named functions instead of `const`s assigned to anonymous functions
  - this makes tracing a bit easier and source maps a bit clearer
  - (also technically a memory improvement as well)

### Verification

- Pure refactor, no real semantic changes
- Test manually, seems to work, still very interactive with fancy animations
  - It was my first time using this page though, I literally didn't know it existed 👀 
    - I believe it's not really documented either?
  - Screenshot below of it working: 
Screenshot 2023-09-10 at 5 19 54 PM


ui/src/app/reports/components/reports.tsx
```diff
@@ -2,351 +2,132 @@ import {Page} from 'argo-ui/src/index';
...
+    // internal state
+    const [charts, setCharts] = useState<Chart[]>();
+    const [error, setError] = useState<Error>();
+
+    // save history
+    useEffect(() => {
+        history.push(historyUrl('reports' + (Utils.managedNamespace ? '' : '/{namespace}'), {namespace, labels: labels.join(',')}));
+    }, [namespace, labels]);
+
+    async function onChange(newNamespace: string, newLabels: string[]) {
```
agilgur5, line 44: potential `useCallback` optimization

ui/src/app/reports/components/reports-filters.tsx
```diff
@@ -0,0 +1,95 @@
...
+        setLabel(labelKeyPhase, value);
+    }
+
+    function setWorkflowTemplate(value: string) {
+        setLabel(labelKeyWorkflowTemplate, value);
+    }
+
+    function setCronWorkflow(value: string) {
+        setLabel(labelKeyCronWorkflow, value);
+    }
```
agilgur5, line 43: potential `useCallback` optimizations

agilgur5 (commented):
It's possible to optimize this a bit with [`useCallback`]. And I think most functions are memoizable too.

But that can be done in a follow-up optimization PR if needed, leaving as is for now for least potential breakage. It also runs pretty fast locally anyway so nbd probably.

ui/src/app/reports/components/reports-filters.tsx
```diff
@@ -0,0 +1,95 @@
...
+                    <NamespaceFilter
+                        value={namespace}
+                        onChange={newNamespace => {
+                            onChange(newNamespace, labels);
+                        }}
+                    />
+                </div>
+                <div className=' columns small-12 xlarge-12'>
+                    <p className='wf-filters-container__title'>Labels</p>
+                    <TagsInput placeholder='Labels' tags={labels} onChange={newLabels => onChange(namespace, newLabels)} />
```
toyamagu-2021, line 59: I assume you've removed `autocomplete` props intentionally since that does nothing in original code.

toyamagu-2021 (approved):
I've double chacked functionality.

ui/src/app/reports/components/reports-filters.tsx
agilgur5: yup exactly, it was just totally unused and an empty string `''`

isubasinghe (approved):
Looks good, no need to worry about useCallback for now imo. 
Might even be slower for some functions, anyway I think the performance of the UI is not a worry for me regardless. 

Oh the only exception being the swagger docs page which is unusable but yes not related to this issue.

agilgur5:
> Looks good, no need to worry about useCallback for now imo. [...] anyway I think the performance of the UI is not a worry for me regardless.

Agreed, figure I'd notate it at least. I used to write fully memoized components, but I'd forgotten some of the optimizations like `useCallback` (since React hasn't been my day job in years). Some of those UIs I wrote in the past were impressively performant, so it _can_ make a difference at scale (scale with respect to number of rerenders and components).

> Oh the only exception being the swagger docs page which is unusable but yes not related to this issue.

Yea it freezes everytime I try to use it. I can't even run a profiler on it. It may very well be on the `swagger-ui` side though, or due to the size of the swagger. I have not seen this problem in other codebases though (but idr if we used different renderers or not).

terrytangyuan (approved):

ui/src/app/reports/components/reports.tsx
```diff
@@ -2,351 +2,132 @@ import {Page} from 'argo-ui/src/index';
...
+                    {title: 'Reports', path: uiUrl('reports')},
+                    {title: namespace, path: uiUrl('reports/' + namespace)}
+                ]
+            }}>
+            <div className='row'>
+                <div className='columns small-12 xlarge-2'>
+                    <ReportFilters namespace={namespace} labels={labels} onChange={onChange} />
+                </div>
+                <div className='columns small-12 xlarge-10'>
+                    <ErrorNotice error={error} />;
```
agilgur5, line 90: woops just realized I copy+pasted an extra semi-colon here (from [the previous version](https://github.com/argoproj/argo-workflows/pull/11794/files#diff-471e092498589fe7d5aab61b99b2608a8eb1cd0403b1e8022d661f7a844d55a9L308)).

this actually shows up in rendering as a random semi-colon on the page, woooops 😅 

will have that fixed shortly

agilgur5: Fixed in #12060

PR #11797, opened 2023-09-11
docs: Add links to Distributed ML Patterns book

terrytangyuan:
The book is related to these two use cases that leverage Argo Workflows.

### DIFF
--- a/docs/use-cases/data-processing.md
+++ b/docs/use-cases/data-processing.md
@@ -10,4 +10,4 @@
 
 ## Books
 
-* [Distributed Machine Learning Patterns](https://github.com/terrytangyuan/distributed-ml-patterns)
+* [Distributed Machine Learning Patterns](https://github.com/terrytangyuan/distributed-ml-patterns) (see Chapter 2 on data processing/ingestion patterns)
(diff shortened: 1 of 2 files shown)


docs/use-cases/data-processing.md
```diff
@@ -7,3 +7,7 @@
 ## Videos
 
 * [Running a Data Replication Pipeline on Kubernetes with Argo and Singer.io](https://towardsdatascience.com/running-a-data-replication-pipeline-on-kubernetes-with-argo-and-singer-io-2fab5b0bad?utm_source=argo-docs)
+
+## Books
+
+* [Distributed Machine Learning Patterns](https://github.com/terrytangyuan/distributed-ml-patterns)
```
agilgur5, line 13: since we have the ML page, should this only be on that page? Whereas this page would be more focused on more "pure" data work (e.g. ETLs)?

terrytangyuan: There's a dedicated chapter in the book for data processing.

agilgur5: Could you specify the chapter here? Something like "(see Chapter 12 on Data Processing)"

That makes it more clear why this one is in this list specifically (since we do have separate lists).

terrytangyuan: Done

agilgur5 (approved):

juliev0 (approved):

Issue #11657, opened 2023-08-23
Controller: `invalid config map object received in config watcher. Ignored processing` with CPU spike


juranir:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

We're migrating our Argo workflow from v3.0.2 hosted in an old Rancher cluster to the last version in an AWS EKS cluster.

Ramdomically the workflow controller increases de CPU usage and starts to show this error.
`level=error msg="invalid config map object received in config watcher. Ignored processing"`

If we remove the pod, the new one starts well, but after 1 or 2 days (random), the error reappears.

The last one started yesterday at the moment without workflow running nor scheduled tasks on the infra side.

We don't have an error in the k8s API log to indicate some EKS outage.

The Argo server component doesn't have errors.

We don't have this error in the old version.

**Env:**
Argo: v3.4.10
Installation method: Official Helm Chart (0.33.1)
K8s Cluster: AWS EKS 1.27

---

**Workflow-controller k8s config map:**
> Name:         argo-workflow-controller-configmap
> Namespace:    argo
> Labels:       app.kubernetes.io/component=workflow-controller
>               app.kubernetes.io/instance=argo
>               app.kubernetes.io/managed-by=Helm
>               app.kubernetes.io/name=argo-workflows-cm
>               app.kubernetes.io/part-of=argo-workflows
>               helm.sh/chart=argo-workflows-0.33.1
> Annotations:  meta.helm.sh/release-name: argo
>               meta.helm.sh/release-namespace: argo
> 
> Data
> 
> config:
> 
> metricsConfig:
>   enabled: true
>   path: /metrics
>   port: 9090
>   ignoreErrors: false
>   secure: false
> persistence:
>   archive: true
>   archiveTTL: 60d
>   postgresql:
>     database: argo
>     host: <REDACTED>
>     passwordSecret:
>       key: password
>       name: argo-postgres-config
>     port: 5432
>     ssl: false
>     sslMode: disabled
>     tableName: argo_workflows
>     userNameSecret:
>       key: username
>       name: argo-postgres-config
> workflowDefaults:
>   spec:
>     ttlStrategy:
>       secondsAfterFailure: 604800
> sso:
[...]

PR #11799, opened 2023-09-11
fix: controller watches global configMap in the namespace where it is running and not in managedNamespace - fixes #11463

### LINKED ISSUE #11463: Controller not picking up configmap updates when running with `--managed-namespace`
### Pre-requisites

# Description

When updating the workflow controller configmap, the controller does not automatically pick up those changes. Only after I delete the controller pod and it gets recreated does it pick them up. This happens only when the controller is running with `--namespaced` and with `--managed-namespace=<other>`

For added context, the docs suggest configurations like these, with both flags, when restricting the namespace where the workflows run to just one (or a few): https://github.com/argoproj/argo-workflows/blob/2d20532a814a654075764b5901625083b690c8db/docs/managed-namespace.md?plain=1#L8-L24

# How to reproduce
To reproduce the issue:
1. Have a k8s cluster with 2 namespaces, e.g. `argo` (where the controller will run) and `argo-runs` (where workflows would run),
2. Run the workflow controller in `argo` with a workflow-controller-configmap in `argo` as well. Run it with both `--namespaced` and with `--managed-namespace=argo-runs`.
3. Edit the configmap, for example adding `workflowDefaults.spec.podMetadata.annotations: {some: value}`
4. Check the controller logs. ➡️ You won't see any log of it picking up the updated configmap, which is wrong. You'll only see the initial configuration picked up on the controller startup.
5. _(Optional)_ Run any workflow in the `argo-runs` namespace. Then, check the resulting workflow resource for it. ➡️ If you added a podMetadata annotation in step 3, you won't see it appear on the new workflow, which is wrong
6. Restart the controller (e.g. by bouncing the pod)
7. Re-do step 4 - Check the logs again. ➡️ Now, the logs will show the new configuration, as it is correctly read on startup
8. _(Optional)_ Re-do step 5 - Deploy any workflow in the `argo-runs` namespace ➡️ Now the resulting workflow resource will show the `podMetadata.annotations.some: value` defined

# Analysis

The issue comes from the following line: 
https://github.com/argoproj/argo-workflows/blob/2d20532a814a654075764b5901625083b690c8db/workflow/controller/controller.go#L389-L391 

In it, the controller is watching the config map in the **managed namespace** instead of the namespace where the controller itself is running. When it starts, and what makes most sense IMHO, it reads it from the **namespace where the controller itself runs** (as the controller AFAIK does not accept per-namespace config maps): https://github.com/argoproj/argo-workflows/blob/2d20532a814a654075764b5901625083b690c8db/workflow/controller/controller.go#L178

I've tested changing locally the namespace where it watches for the configmap and that seems to fix the issue. I'd like to further test this and make an actual PR to fix the issue myself, but I'd really appreciate any feedback/input as would be my first time contributing to this project 🙏 

# Appendix / extra info 

For reference, here is my full workflow controller deployment. See the args, where both `--namespaced` (deployed in the `argo` namespace) and `--managed-namespace=argo-runs` are specified.
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "2"
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"annotations":{},"labels":{"app.kubernetes.io/part-of":"argo"},"name":"workflow-controller","namespace":"argo"},"spec":{"replicas":0,"selector":{"matchLabels":{"app":"workflow-controller","app.kubernetes.io/part-of":"argo"}},"template":{"metadata":{"labels":{"app":"workflow-controller","app.kubernetes.io/part-of":"argo"}},"spec":{"containers":[{"args":["--namespaced"],"command":["workflow-controller"],"env":[{"name":"LEADER_ELECTION_IDENTITY","valueFrom":{"fieldRef":{"apiVersion":"v1","fieldPath":"metadata.name"}}}],"image":"quay.io/argoproj/workflow-controller:latest","imagePullPolicy":"Never","livenessProbe":{"failureThreshold":3,"httpGet":{"path":"/healthz","port":6060},"initialDelaySeconds":90,"periodSeconds":60,"timeoutSeconds":30},"name":"workflow-controller","ports":[{"containerPort":9090,"name":"metrics"},{"containerPort":6060}],"securityContext":{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"readOnlyRootFilesystem":true,"runAsNonRoot":true}}],"nodeSelector":{"kubernetes.io/os":"linux"},"priorityClassName":"workflow-controller","securityContext":{"runAsNonRoot":true},"serviceAccountName":"argo"}}}}
  creationTimestamp: "2023-07-27T20:00:05Z"
  generation: 5
  labels:
    app.kubernetes.io/part-of: argo
  name: workflow-controller
  namespace: argo
  resourceVersion: "1619"
  uid: 613ec8e2-7e84-4ede-b0d8-29cf120fba08
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: workflow-controller
      app.kubernetes.io/part-of: argo
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: workflow-controller
        app.kubernetes.io/part-of: argo
    spec:
      containers:
      - args:
        - --namespaced
        - --managed-namespace=argo-runs
        command:
        - workflow-controller
        env:
        - name: LEADER_ELECTION_IDENTITY
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        image: quay.io/argoproj/workflow-controller:latest
        imagePullPolicy: Never
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /healthz
            port: 6060
            scheme: HTTP
          initialDelaySeconds: 90
          periodSeconds: 60
          successThreshold: 1
          timeoutSeconds: 30
        name: workflow-controller
        ports:
        - containerPort: 9090
          name: metrics
          protocol: TCP
        - containerPort: 6060
          protocol: TCP
        resources: {}
        securityContext:
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: true
          runAsNonRoot: true
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      nodeSelector:
        kubernetes.io/os: linux
      priorityClassName: workflow-controller
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext:
        runAsNonRoot: true
      serviceAccount: argo
      serviceAccountName: argo
      terminationGracePeriodSeconds: 30
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2023-07-27T20:00:05Z"
    lastUpdateTime: "2023-07-27T20:13:29Z"
    message: ReplicaSet "workflow-controller-5d78457cbf" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
  - lastTransitionTime: "2023-07-27T20:22:39Z"
    lastUpdateTime: "2023-07-27T20:22:39Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  observedGeneration: 5
  readyReplicas: 1
  replicas: 1
  updatedReplicas: 1

```

### Version

latest

```YAML
<N/A>: No workflow is needed to reproduce this issue
```

```text
```

```text
<N/A>: As no workflow is needed for this issue, no wait container logs are needed either
```

toblich:
Fixes #11463 

### Motivation

There was a bug by which the workflow controller did not pick up changes made to its configMap until its pod was deleted. This only manifestated when the controller is running in a particular namespace (`--namespaced`) but is managing workflows in a different namespace (`--managed-namespace=<some_other_ns>`)

### Modifications

Fixed the namespace that the controller watches for changes made to its own configMap, so that it watches the same namespace that it initially reads. See the analysis in #11463 for more details. 

### Verification

I did manual verification by:
1. Deploying into a cluster using both flags (`--namespaced` `--managed-namespace=<other_ns>`) 
2. Making changes to the config map (e.g. to the workflowDefaults)
3. Reading the controller logs
4. Starting a workflow to validate whether the changed made to the configmap had been picked up or not

I also ran all tests, but given that this bug is related to configmaps and namespaces in a k8s cluster, and I don't see a clear way of simulating that in `workflow/controller/controller_test.go`, I haven't added any new tests, but I'm obviously open to do it if there's any example of how to mock the namespaces and configmaps to actually test the watcher (which, as far as I can tell, relies on k8s control plane actually being there and is thus complicated to test without a real k8s cluster).

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -398,7 +398,7 @@ func (wfc *WorkflowController) runConfigMapWatcher(stopCh <-chan struct{}) {
 	ctx := context.Background()
 	retryWatcher, err := apiwatch.NewRetryWatcher("1", &cache.ListWatch{
 		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-			return wfc.kubeclientset.CoreV1().ConfigMaps(wfc.managedNamespace).Watch(ctx, metav1.ListOptions{})
+			return wfc.kubeclientset.CoreV1().ConfigMaps(wfc.namespace).Watch(ctx, metav1.ListOptions{})
 		},
 	})
 	if err != nil {


agilgur5 (approved):
Thanks for this fix! As stated in the issue, this may also potentially fix some related issues as well, if this was their root cause.

This is a one-line fix, so I'm ok approving as is with manual verification.

I did verify the codepaths myself that this is logically correct:
1. [`config` controller creation](https://github.com/argoproj/argo-workflows/blob/86ee23a9161b9a38f1985392e8536591ccb9231e/workflow/controller/controller.go#L185) uses `namespace` and _not_ `managedNamespace`
1. `UpdateConfig` [calls `Get`](https://github.com/argoproj/argo-workflows/blob/86ee23a9161b9a38f1985392e8536591ccb9231e/workflow/controller/controller.go#L472) on the `config` controller, which [uses `namespace`](https://github.com/argoproj/argo-workflows/blob/f9bb71da8504cbcda8c8f90463975e0b6a9f0302/config/controller.go#L57).
1. Both of those explain why it still works on initialization, but fails to `watch` correctly

So `WatchFunc` here is the odd one out. And ofc, as stated in PR and issue, the `workflow-controller-configmap` _should_ be in the same namespace as the Workflow Controller.

agilgur5 (commented):
Hmm I do see a potential problem that I [suspected](https://github.com/argoproj/argo-workflows/issues/11463#issuecomment-1714074509) in the issue...

On line 422 below in this same function, `wfc.notifySemaphoreConfigUpdate` is called for _every_ ConfigMap in the namespace. Semaphore configs are indeed in the `managedNamespace`.

This may require two separate watchers if there isn't a separate watcher on semaphores already

agilgur5:
So there is a separate watcher. The `configMapInformer` is [created on the `managedNamespace`](https://github.com/argoproj/argo-workflows/blob/86ee23a9161b9a38f1985392e8536591ccb9231e/workflow/controller/controller.go#L1056). But that seems to be used almost exclusively [for `parameters` from ConfigMaps](https://github.com/argoproj/argo-workflows/blob/f9bb71da8504cbcda8c8f90463975e0b6a9f0302/workflow/controller/operator.go#L620) (and [GC cache](https://github.com/argoproj/argo-workflows/blob/f9bb71da8504cbcda8c8f90463975e0b6a9f0302/workflow/controller/cache_gc.go#L28)??), if I'm understanding the code correctly.

That informer should probably be extended to semaphores then if I'm understanding correctly.
(Vaguely related, but the `retryWatcher` should probably be converted to an informer too per https://github.com/argoproj/argo-workflows/issues/11657#issuecomment-1705576977).

Would definitely be good to get someone more familiar with the ConfigMap codepaths. @sarabala1979 wrote some of this code originally in #4421, which is why I requested his review.

juliev0:
thanks @toblich for the fix! I just reviewed the [PR](https://github.com/argoproj/argo-workflows/pull/11855) referenced above , which looks like it could get merged soon and takes care of the issue you found as well, so in that case this whole method would go away.

agilgur5:
@juliev0 not quite, that PR doesn't actually resolve this issue at all. seems like the author there did not interpret the bug here correctly. See my comment there: https://github.com/argoproj/argo-workflows/pull/11855#issuecomment-1741676801

juliev0:
> @juliev0 not quite, that PR doesn't actually resolve this issue at all. seems like the author there did not interpret the bug here correctly. See my comment there: [#11855 (comment)](https://github.com/argoproj/argo-workflows/pull/11855#issuecomment-1741676801)

yes, you're totally right. Not sure how I missed that :|

juliev0 (approved):

juliev0 (changes requested):
actually, holding off on approval until maybe we can get some feedback on Anton's questions

agilgur5:
After reviewing #11855 in-depth, I am pretty sure that `wfc.notifySemaphoreConfigUpdate` is indeed problematic here and needs moving (it also needs moving in #11855's current state due to the `FilterFunc`).

As far as I understand, that function is responsible for requeueing if a semaphore's ConfigMap value increases in size (say you had a limit of 5 and now allow 10, for example). To retain that functionality, that function call must continue to be called on the managed namespace.

juliev0:
@agilgur5 

> for `parameters` from ConfigMaps

Yeah, I see what you're saying. `runConfigMapWatcher()` has been handling both the WorkflowControllerConfigmap and [these](https://argoproj.github.io/argo-workflows/synchronization/) ConfigMaps. It would make sense that the latter live on the managed namespace, same as the Workflows that they're used by.

juliev0:
> created on the `managedNamespace`

Are you thinking that instead of using a `RetryWatcher` for the managed namespace, we should take advantage of the informer's [UpdateFunc](https://github.com/argoproj/argo-workflows/blob/86ee23a9161b9a38f1985392e8536591ccb9231e/workflow/controller/controller.go#L1090) since otherwise we're creating an unnecessary Watch?

agilgur5:
> we should take advantage of the informer's [UpdateFunc](https://github.com/argoproj/argo-workflows/blob/86ee23a9161b9a38f1985392e8536591ccb9231e/workflow/controller/controller.go#L1090) since otherwise we're creating an unnecessary Watch?

ostensibly, it would be a good idea to re-use an existing Informer such as the `configMapInformer`. that specific line/event handler wouldn't fit though, as it's [filtered on executor plugins](https://github.com/argoproj/argo-workflows/blob/26481a2146107ad0937ef7698c27f3686f93c81e/workflow/controller/controller.go#L1069) (and only runs if there are [executor plugins defined](https://github.com/argoproj/argo-workflows/blob/26481a2146107ad0937ef7698c27f3686f93c81e/workflow/controller/controller.go#L1062)). A new event handler outside of that conditional can be added to work around that though.

I was also concerned in https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1341910749 that the `configMapInformer`'s [label selector](https://github.com/argoproj/argo-workflows/blob/26481a2146107ad0937ef7698c27f3686f93c81e/workflow/controller/controller.go#L1059) may not apply for semaphore configmaps. If that is accurate, we would either need a third ConfigMap Informer (the two in #11855 plus another one) or to modify that one to handle both scenarios (e.g. move the label selector into the `FilterFunc` -- but that could change the indices, so that usage would still need to be further checked)

juliev0:
> > we should take advantage of the informer's [UpdateFunc](https://github.com/argoproj/argo-workflows/blob/86ee23a9161b9a38f1985392e8536591ccb9231e/workflow/controller/controller.go#L1090) since otherwise we're creating an unnecessary Watch?
> 
> ostensibly, it would be a good idea to re-use an existing Informer such as the `configMapInformer`. that specific line/event handler wouldn't fit though, as it's [filtered on executor plugins](https://github.com/argoproj/argo-workflows/blob/26481a2146107ad0937ef7698c27f3686f93c81e/workflow/controller/controller.go#L1069) (and only runs if there are [executor plugins defined](https://github.com/argoproj/argo-workflows/blob/26481a2146107ad0937ef7698c27f3686f93c81e/workflow/controller/controller.go#L1062)). A new event handler outside of that conditional can be added to work around that though.
> 
> I was also concerned in [#11855 (comment)](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1341910749) that the `configMapInformer`'s [label selector](https://github.com/argoproj/argo-workflows/blob/26481a2146107ad0937ef7698c27f3686f93c81e/workflow/controller/controller.go#L1059) may not apply for semaphore configmaps. If that is accurate, we would either need a third ConfigMap Informer (the two in #11855 plus another one) or to modify that one to handle both scenarios (e.g. move the label selector into the `FilterFunc` -- but that could change the indices, so that usage would still need to be further checked).
> 
> I can tackle that if our two contributors here aren't able to pick it up.

Nice deep diving into all of this. Yes, you seem to have become an expert on this.

toblich:
Closing this PR in favor of #11855, which is doing a much more comprehensive fix and also taking care of #11463

PR #11791, opened 2023-09-09
refactor(ui): migrate several components from class to functional

agilgur5:
Partial fix for #9810

### Motivation

- Use newer, modern functional components compatible with hooks instead of legacy class-based components

- Several of the simpler components in `worflows/components/` were relatively quick to refactor, so did so
  - Stumbled upon them while reviewing some UI PRs like #11790, figured I might as well convert them

- Others I just added to this PR as I was asked to consolidate PRs due to lack of approvers right now
  - most of these are pure functional components

### Modifications

- migrate several components from class-based to functional:
  - `shared/components/error-panel.tsx`
  - `shared/components/inline-table/inline-table.tsx`
  - `shared/components/inline-table/nudge.tsx`
    - this one is technically the most "complex" (still very simple) on this list as it has a single variable of boolean state and uses `localStorage`
  - `shared/components/pagination-panel.tsx`
  - `shared/conditions-panel.tsx`
  - `shared/resources-duration.tsx`
  - `workflows/components/workflow-creator-info/workflow-creator-info.tsx`
  - `workflows/components/workflow-dag/workflow-dag-render-options-panel.tsx`
    - this one required changing a getter to a regular function as no more classes
  - `workflows/components/workflow-labels/workflow-labels.tsx`
  - `workflows/components/workflow-node-info/workflow-node-info.tsx`
    - this one has a single variable of string state
  - `workflows/components/workflow-panel/workflow-panel.tsx`
  - `workflows/components/workflow-yaml-viewer/workflow-yaml-viewer.tsx`
  - `workflows/components/workflows-row/workflows-row.tsx`
    - this one has a single variable of boolean state

Also some tiny optimizations similar to #11743:
- move static functions outside of the component
  - this way they are not created every time the component is and hence use less memory etc
- use named functions instead of consts assigned to anonymous functions
  - this makes tracing a bit easier and source maps a bit clearer
  - (also technically a memory improvement as well)

### Verification

- Pure refactor, no real semantic changes.
- Build / typecheck pass.
- Opened up the UI locally and had no issues

### Notes for Reviewers

Turning on [hide whitespace](https://github.com/argoproj/argo-workflows/pull/11791/files?w=1) shrinks the diff quite significantly, as most of it is indentation changes

PR #11800, opened 2023-09-11
refactor(ui): convert a few components to use hooks

agilgur5:
Partial fix for #9810 
Partial fix for #11740 

### Motivation

- Use newer, modern React hooks syntax instead of legacy React component lifecycle hooks

- Use newer, modern async/await syntax instead of Promise chains

Combined a few components together as previously requested. These are all components that make use of effects and other hooks (e.g. refs), but are not significantly complicated (unlike #11794, which is one complex component -- an entire page. and #11791 is all components _without_ effects).

### Modifications

Convert 3 components to use hooks: `ResourceEditor`, `WorkflowDrawer`, and `TagsInput`
- `this.props` -> `props`
- `this.state` -> `useState`
  - simplify some callbacks now too
- `ref` -> `useRef`
  - rename variables to clarify that these are refs which are accessed via `.current`
- `componentDidMount` -> `useEffect`
- methods -> inner functions
  - also convert to async/await syntax as well
  - hoist child render functions into the main component

- also shorten some logic by using modern optional chaining syntax

See commit messages for more details

### Verification

- Manually tested a solid bit for `TagsInput` (as the refs gave some errors for a bit due to an accidental incorrect renaming) and a bit for `WorkflowDrawer`. Screenshots of them working post-refactor:
Screenshot 2023-09-11 at 5 33 48 PM
Screenshot 2023-09-11 at 5 35 50 PM

### Notes for Reviewers

[Ignoring whitespace changes](https://github.com/argoproj/argo-workflows/pull/11800/files?w=1) makes it a bit easier to read. Though there are still a good amount of other refactorings

### DIFF
--- a/ui/src/app/shared/components/resource-editor/resource-editor.tsx
+++ b/ui/src/app/shared/components/resource-editor/resource-editor.tsx
@@ -1,5 +1,6 @@
 import * as kubernetes from 'argo-ui/src/models/kubernetes';
 import * as React from 'react';
+import {useState} from 'react';
 import {Button} from '../button';
 import {ErrorNotice} from '../error-notice';
 import {ObjectEditor} from '../object-editor/object-editor';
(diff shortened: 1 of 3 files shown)


toyamagu-2021 (approved):
I've double checked functionalities.

terrytangyuan (approved):

PR #11803, opened 2023-09-12
refactor(ui): workflow panel components from class to functional

tetora1053:
Partial fix for https://github.com/argoproj/argo-workflows/issues/9810
Partial fix for https://github.com/argoproj/argo-workflows/issues/11740

### Motivation

### Modifications

- `this.state` -> `useState`
- `this.props` -> `props`
- methods -> inner functions
- anonymous functions -> named functions
- chaining promise callbacks -> async/await & try/catch
- useMemo for optimization
- hoist up static variable out of component

### Verification

- manually tested in local environment

- submit panel
Screen Shot 2023-09-12 at 10 48 02
Screen Shot 2023-09-12 at 11 22 09
Screen Shot 2023-09-12 at 11 22 24
Screen Shot 2023-09-12 at 11 27 33

- resubmit panel
Screen Shot 2023-09-12 at 10 48 33

- retry panel
Screen Shot 2023-09-12 at 10 49 25

## References PR
- https://github.com/argoproj/argo-workflows/pull/11800

### DIFF
--- a/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
@@ -1,5 +1,5 @@
 import {Checkbox} from 'argo-ui';
-import * as React from 'react';
+import React, {useState} from 'react';
 import {Parameter, ResubmitOpts, Workflow} from '../../../models';
 import {uiUrl} from '../../shared/base';
 import {ErrorNotice} from '../../shared/components/error-notice';
@@ -12,103 +12,93 @@ interface Props {
     isArchived: boolean;
 }
 
-interface State {
-    overrideParameters: boolean;
-    workflowParameters: Parameter[];
-    memoized: boolean;
-    error?: Error;
-    isSubmitting: boolean;
-}
+export function ResubmitWorkflowPanel(props: Props) {
+    const [overrideParameters, setOverrideParameters] = useState(false);
+    const [workflowParameters, setWorkflowParameters] = useState<Parameter[]>(JSON.parse(JSON.stringify(props.workflow.spec.arguments.parameters || [])));
+    const [memoized, setMemoized] = useState(false);
+    const [error, setError] = useState<Error>();
+    const [isSubmitting, setIsSubmitting] = useState(false);
 
-export class ResubmitWorkflowPanel extends React.Component<Props, State> {
-    constructor(props: any) {
-        super(props);
-        const state: State = {
-            workflowParameters: JSON.parse(JSON.stringify(this.props.workflow.spec.arguments.parameters || [])),
-            memoized: false,
-            isSubmitting: false,
-            overrideParameters: false
-        };
-        this.state = state;
-    }
-
-    public render() {
-        return (
-            <>
-                <h4>Resubmit Workflow</h4>
-                <h5>
-                    {this.props.workflow.metadata.namespace}/{this.props.workflow.metadata.name}
-                </h5>
-                {this.state.error && <ErrorNotice error={this.state.error} />}
-                <div className='white-box'>
-                    <div key='override-parameters' style={{marginBottom: 25}}>
-                        <label>Override Parameters</label>
-                        <div className='columns small-9'>
-                            <Checkbox checked={this.state.overrideParameters} onChange={overrideParameters => this.setState({overrideParameters})} />
-                        </div>
-                    </div>
-
-                    {this.state.overrideParameters && (
-                        <div key='parameters' style={{marginBottom: 25}}>
-                            <label>Parameters</label>
-                            {this.state.workflowParameters.length > 0 && (
-                                <ParametersInput parameters={this.state.workflowParameters} onChange={workflowParameters => this.setState({workflowParameters})} />
-                            )}
-                            {this.state.workflowParameters.length === 0 ? (
-                                <>
-                                    <br />
-                                    <label>No parameters</label>
-                                </>
-                            ) : (
-                                <></>
-                            )}
-                        </div>
-                    )}
-
-                    <div key='memoized' style={{marginBottom: 25}}>
-                        <label>Memoized</label>
-                        <div className='columns small-9'>
-                            <Checkbox checked={this.state.memoized} onChange={memoized => this.setState({memoized})} />
-                        </div>
-                    </div>
-
-                    {this.state.overrideParameters && this.state.memoized && (
-                        <div key='warning-override-with-memoized'>
-                            <i className='fa fa-exclamation-triangle' style={{color: '#f4c030'}} />
-                            Overriding parameters on memoized submitted workflows may have unexpected results.
-                        </div>
-                    )}
-
-                    <div key='resubmit'>
-                        <button onClick={() => this.submit()} className='argo-button argo-button--base' disabled={this.state.isSubmitting}>
-                            <i className='fa fa-plus' /> {this.state.isSubmitting ? 'Loading...' : 'Resubmit'}
-                        </button>
-                    </div>
-                </div>
-            </>
-        );
-    }
-
-    private submit() {
-        this.setState({isSubmitting: true});
-        const parameters: ResubmitOpts['parameters'] = this.state.overrideParameters
-            ? [...this.state.workflowParameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p))]
+    const submit = () => {
+        setIsSubmitting(true);
+        const parameters: ResubmitOpts['parameters'] = overrideParameters
+            ? [...workflowParameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p))]
             : [];
         const opts: ResubmitOpts = {
             parameters,
-            memoized: this.state.memoized
+            memoized
         };
 
-        if (!this.props.isArchived) {
+        const handleSuccess = (submitted: Workflow) => {
+            document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
+        };
+
+        const handleError = (err: Error) => {
+            setError(err);
+            setIsSubmitting(false);
+        };
+
+        if (!props.isArchived) {
             services.workflows
-                .resubmit(this.props.workflow.metadata.name, this.props.workflow.metadata.namespace, opts)
-                .then((submitted: Workflow) => (document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`)))
-                .catch(error => this.setState({error, isSubmitting: false}));
+                .resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts)
+                .then(handleSuccess)
+                .catch(handleError);
         } else {
             services.workflows
-                .resubmitArchived(this.props.workflow.metadata.uid, this.props.workflow.metadata.namespace, opts)
-                .then((submitted: Workflow) => (document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`)))
-                .catch(error => this.setState({error, isSubmitting: false}));
+                .resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts)
+                .then(handleSuccess)
+                .catch(handleError);
         }
-    }
+    };
+
+    return (
+        <>
+            <h4>Resubmit Workflow</h4>
+            <h5>
+                {props.workflow.metadata.namespace}/{props.workflow.metadata.name}
+            </h5>
+            {error && <ErrorNotice error={error} />}
+            <div className='white-box'>
+                <div key='override-parameters' style={{marginBottom: 25}}>
+                    <label>Override Parameters</label>
+                    <div className='columns small-9'>
+                        <Checkbox checked={overrideParameters} onChange={setOverrideParameters} />
+                    </div>
+                </div>
+
+                {overrideParameters && (
+                    <div key='parameters' style={{marginBottom: 25}}>
+                        <label>Parameters</label>
+                        {workflowParameters.length > 0 && <ParametersInput parameters={workflowParameters} onChange={setWorkflowParameters} />}
+                        {workflowParameters.length === 0 && (
+                            <>
+                                <br />
+                                <label>No parameters</label>
+                            </>
+                        )}
+                    </div>
+                )}
+
+                <div key='memoized' style={{marginBottom: 25}}>
+                    <label>Memoized</label>
+                    <div className='columns small-9'>
+                        <Checkbox checked={memoized} onChange={setMemoized} />
+                    </div>
+                </div>
+
+                {overrideParameters && memoized && (
+                    <div key='warning-override-with-memoized'>
+                        <i className='fa fa-exclamation-triangle' style={{color: '#f4c030'}} />
+                        Overriding parameters on memoized submitted workflows may have unexpected results.
+                    </div>
+                )}
+
+                <div key='resubmit'>
+                    <button onClick={submit} className='argo-button argo-button--base' disabled={isSubmitting}>
+                        <i className='fa fa-plus' /> {isSubmitting ? 'Loading...' : 'Resubmit'}
+                    </button>
+                </div>
+            </div>
+        </>
+    );
 }
--- a/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
@@ -19,7 +19,7 @@ export function ResubmitWorkflowPanel(props: Props) {
     const [error, setError] = useState<Error>();
     const [isSubmitting, setIsSubmitting] = useState(false);
 
-    const submit = () => {
+    function submit() {
         setIsSubmitting(true);
         const parameters: ResubmitOpts['parameters'] = overrideParameters
             ? [...workflowParameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p))]
@@ -29,14 +29,14 @@ export function ResubmitWorkflowPanel(props: Props) {
             memoized
         };
 
-        const handleSuccess = (submitted: Workflow) => {
+        function handleSuccess(submitted: Workflow) {
             document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
-        };
+        }
 
-        const handleError = (err: Error) => {
+        function handleError(err: Error) {
             setError(err);
             setIsSubmitting(false);
-        };
+        }
 
         if (!props.isArchived) {
             services.workflows
@@ -49,7 +49,7 @@ export function ResubmitWorkflowPanel(props: Props) {
                 .then(handleSuccess)
                 .catch(handleError);
         }
-    };
+    }
 
     return (
         <>
--- a/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
@@ -19,7 +19,7 @@ export function ResubmitWorkflowPanel(props: Props) {
     const [error, setError] = useState<Error>();
     const [isSubmitting, setIsSubmitting] = useState(false);
 
-    function submit() {
+    async function submit() {
         setIsSubmitting(true);
         const parameters: ResubmitOpts['parameters'] = overrideParameters
             ? [...workflowParameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p))]
@@ -29,25 +29,14 @@ export function ResubmitWorkflowPanel(props: Props) {
             memoized
         };
 
-        function handleSuccess(submitted: Workflow) {
+        try {
+            const submitted = props.isArchived
+                ? await services.workflows.resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts)
+                : await services.workflows.resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts);
             document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
-        }
-
-        function handleError(err: Error) {
+        } catch (err) {
             setError(err);
             setIsSubmitting(false);
-        }
-
-        if (!props.isArchived) {
-            services.workflows
-                .resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts)
-                .then(handleSuccess)
-                .catch(handleError);
-        } else {
-            services.workflows
-                .resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts)
-                .then(handleSuccess)
-                .catch(handleError);
         }
     }
 
--- a/ui/src/app/workflows/components/retry-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/retry-workflow-panel.tsx
@@ -1,5 +1,5 @@
 import {Checkbox} from 'argo-ui';
-import * as React from 'react';
+import React, {useState} from 'react';
 import {Parameter, RetryOpts, Workflow} from '../../../models';
 import {uiUrl} from '../../shared/base';
 import {ErrorNotice} from '../../shared/components/error-notice';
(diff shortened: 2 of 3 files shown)


tetora1053:
Sorry, I added a commit.
Removed a state that is set but not used at `SubmitWorkflowPanel`.
I have confirmed that no degradation has occurred in the local environment.

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
-                        </button>
-                    </div>
-                </div>
-            </>
-        );
-    }
+    const [entrypoint, setEntrypoint] = useState<string>(workflowEntrypoint);
+    const [parameters, setParameters] = useState<Parameter[]>([]);
+    const [workflowParameters, setWorkflowParameters] = useState<Parameter[]>(JSON.parse(JSON.stringify(props.workflowParameters)));
+    const [templates] = useState<Template[]>([defaultTemplate].concat(props.templates));
```
toyamagu-2021, line 33: Originally state is used, but should we change here without useState?

```
 const templates = [defaultTemplate].concat(props.templates);
```

toyamagu-2021 (approved):
Added comment, but I confirmed no regressions.

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
```diff
@@ -12,103 +12,93 @@ interface Props {
...
 
-export class ResubmitWorkflowPanel extends React.Component<Props, State> {
-    constructor(props: any) {
-        super(props);
-        const state: State = {
-            workflowParameters: JSON.parse(JSON.stringify(this.props.workflow.spec.arguments.parameters || [])),
-            memoized: false,
-            isSubmitting: false,
-            overrideParameters: false
+    const submit = () => {
```
agilgur5, line 22:
```suggestion
    function submit() {
```

prefer named function over const assigned to anonymous function (it's more performant and easier to trace in debug logs etc since "anonymous" is not the name -- I also used this in all of my recent PRs)

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
```diff
@@ -12,103 +12,93 @@ interface Props {
...
+                .resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts)
+                .then(handleSuccess)
+                .catch(handleError);
+        } else {
+            services.workflows
+                .resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts)
+                .then(handleSuccess)
+                .catch(handleError);
+        }
+    };
```
agilgur5, line 52:
```suggestion
    }
```

once it's a named func, the semi-colon will be auto removed by lint

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
```diff
@@ -12,103 +12,93 @@ interface Props {
...
-                <div className='white-box'>
-                    <div key='override-parameters' style={{marginBottom: 25}}>
-                        <label>Override Parameters</label>
-                        <div className='columns small-9'>
-                            <Checkbox checked={this.state.overrideParameters} onChange={overrideParameters => this.setState({overrideParameters})} />
-                        </div>
-                    </div>
+        const handleSuccess = (submitted: Workflow) => {
+            document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
+        };
```
agilgur5, line 34:
```suggestion
        function handleSuccess(submitted: Workflow) {
            document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
        }
```

same as above.

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
```diff
@@ -12,103 +12,93 @@ interface Props {
...
-                {this.state.error && <ErrorNotice error={this.state.error} />}
-                <div className='white-box'>
-                    <div key='override-parameters' style={{marginBottom: 25}}>
-                        <label>Override Parameters</label>
-                        <div className='columns small-9'>
-                            <Checkbox checked={this.state.overrideParameters} onChange={overrideParameters => this.setState({overrideParameters})} />
-                        </div>
-                    </div>
+        const handleSuccess = (submitted: Workflow) => {
+            document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
```
agilgur5, line 33: this is the same as it was before, but shouldn't this use react-router calls instead?

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
```diff
@@ -12,103 +12,93 @@ interface Props {
...
-                                </>
-                            ) : (
-                                <></>
-                            )}
-                        </div>
-                    )}
+        const handleError = (err: Error) => {
+            setError(err);
+            setIsSubmitting(false);
+        };
```
agilgur5, line 39:
```suggestion
        function handleError(err: Error) {
            setError(err);
            setIsSubmitting(false);
        }
```

same as above

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
```diff
@@ -12,103 +12,93 @@ interface Props {
...
+        if (!props.isArchived) {
+            services.workflows
+                .resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts)
+                .then(handleSuccess)
+                .catch(handleError);
+        } else {
+            services.workflows
+                .resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts)
+                .then(handleSuccess)
+                .catch(handleError);
```
agilgur5, line 50:
```suggestion
        let submitted;
        try {
            if (!props.isArchived) {
                submitted = await services.workflows.resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts);
            } else {
                submitted = await services.workflows.resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts);
           }
           handleSuccess(submitted);
       } catch (newError) {
         handleError(newError);
       }
```

would be good to rewrite this with async/await while we're at it, which also makes it cleaner.

we can actually remove the `handleSuccess` and `handleError` functions and instead in-line them this way. `submit` just has to be an `async` function.

ui/src/app/workflows/components/retry-workflow-panel.tsx
```diff
@@ -12,112 +12,98 @@ interface Props {
...
+        process
+            .then((submitted: Workflow) => {
+                document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);
+                setIsSubmitting(false);
+            })
+            .catch(err => {
+                setError(err);
+                setIsSubmitting(false);
+            });
+    };
```
agilgur5, line 47: same as above regarding functions and async/await

ui/src/app/workflows/components/submit-workflow-panel.tsx
agilgur5: Yea this doesn't seem like it actually uses state, just derives.
If that was meant to be a performance optimization, `useMemo` should be used instead (I believe there's a few places in the code where it seemed like it might be more appropriate for `useMemo`).

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
-                    <div key='submit'>
-                        <button onClick={() => this.submit()} className='argo-button argo-button--base' disabled={this.state.isSubmitting}>
-                            <i className='fa fa-plus' /> {this.state.isSubmitting ? 'Loading...' : 'Submit'}
-                        </button>
-                    </div>
-                </div>
-            </>
-        );
-    }
+    const [entrypoint, setEntrypoint] = useState<string>(workflowEntrypoint);
```
agilgur5, line 30:
```suggestion
    const [entrypoint, setEntrypoint] = useState(workflowEntrypoint);
```

`string` straightforwardly inferred here, so type annotation is not needed.

you didn't use type annotations for primitives in `resubmit-workflow-panel.tsx` and I didn't use them in my PRs, so let's keep it consistent and only use them when it's not correctly inferred (which tends to happen with more complex types or different initial states -- `Parameter[]` does need the annotation, for instance).

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
+    const [parameters, setParameters] = useState<Parameter[]>([]);
+    const [workflowParameters, setWorkflowParameters] = useState<Parameter[]>(JSON.parse(JSON.stringify(props.workflowParameters)));
+    const [templates] = useState<Template[]>([defaultTemplate].concat(props.templates));
+    const [labels, setLabels] = useState<string[]>(['submit-from-ui=true']);
```
agilgur5, line 34:
```suggestion
    const [labels, setLabels] = useState(['submit-from-ui=true']);
```

same as above

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
+    const [error, setError] = useState<Error>();
+    const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
```
agilgur5, line 36:
```suggestion
    const [isSubmitting, setIsSubmitting] = useState(false);
```

same as above

ui/src/app/workflows/components/retry-workflow-panel.tsx
```diff
@@ -12,112 +12,98 @@ interface Props {
...
-    overrideParameters: boolean;
-    restartSuccessful: boolean;
-    workflowParameters: Parameter[];
-    nodeFieldSelector: string;
-    error?: Error;
-    isSubmitting: boolean;
-}
+export function RetryWorkflowPanel(props: Props) {
+    const [overrideParameters, setOverrideParameters] = useState<boolean>(false);
+    const [restartSuccessful, setRestartSuccessful] = useState<boolean>(false);
```
agilgur5, line 17:
```suggestion
    const [overrideParameters, setOverrideParameters] = useState(false);
    const [restartSuccessful, setRestartSuccessful] = useState(false);
```

same as above

ui/src/app/workflows/components/retry-workflow-panel.tsx
```diff
@@ -12,112 +12,98 @@ interface Props {
...
+    const [workflowParameters, setWorkflowParameters] = useState<Parameter[]>(JSON.parse(JSON.stringify(props.workflow.spec.arguments.parameters || [])));
+    const [nodeFieldSelector, setNodeFieldSelector] = useState<string>('');
```
agilgur5, line 19:
```suggestion
    const [nodeFieldSelector, setNodeFieldSelector] = useState('');
```

same as above

ui/src/app/workflows/components/retry-workflow-panel.tsx
```diff
@@ -12,112 +12,98 @@ interface Props {
...
+    const [error, setError] = useState<Error>();
+    const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
```
agilgur5, line 21:
```suggestion
    const [isSubmitting, setIsSubmitting] = useState(false);
```

same as above

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
-        };
-        this.state = state;
-    }
+export function SubmitWorkflowPanel(props: Props) {
+    const defaultTemplate: Template = {
+        name: workflowEntrypoint,
+        inputs: {
+            parameters: []
+        }
+    };
```
agilgur5, line 28: this didn't change, but this looks like it's static. we can hoist this to the module level instead of the component-level as an optimization then.

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
-        for (const t of this.state.templates) {
-            if (t.name === entrypoint) {
-                return t;
-            }
-        }
-        return null;
-    }
+    const getSelectedTemplate = (name: string): Template | null => {
+        return templates.find(t => t.name === name) || null;
+    };
```
agilgur5, line 40:
```suggestion
    function getSelectedTemplate(name: string): Template | null {
        return templates.find(t => t.name === name) || null;
    }
```

same as above re: named function over anonymous

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
+                    <Select
+                        value={entrypoint}
+                        options={templates.map(t => ({
+                            value: t.name,
+                            title: t.name
+                        }))}
+                        onChange={selected => {
+                            const selectedTemp = getSelectedTemplate(selected.value);
+                            setEntrypoint(selected.value);
+                            setParameters((selectedTemp && selectedTemp.inputs.parameters) || []);
```
agilgur5, line 79:
```suggestion
                            setParameters(selectedTemp?.inputs?.parameters || []);
```

can simplify this with optional chaining syntax

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
-                            value={this.state.entrypoint}
-                            options={this.state.templates.map(t => ({
-                                value: t.name,
-                                title: t.name
-                            }))}
-                            onChange={selected => {
-                                const selectedTemplate = this.getSelectedTemplate(selected.value);
-                                this.setState({
-                                    entrypoint: selected.value,
-                                    selectedTemplate,
```
agilgur5, line 77: huh this was just unused state. good removal!

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
+                labels: labels.join(',')
             })
             .then((submitted: Workflow) => (document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`)))
-            .catch(error => this.setState({error, isSubmitting: false}));
-    }
+            .catch(err => {
+                setError(err);
+                setIsSubmitting(false);
+            });
+    };
```
agilgur5, line 58: could we convert this to async/await and a named function as well?

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -17,119 +17,92 @@ interface Props {
...
                 parameters: [
-                    ...this.state.workflowParameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p)),
-                    ...this.state.parameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p))
+                    ...workflowParameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p)),
+                    ...parameters.filter(p => Utils.getValueFromParameter(p) !== undefined).map(p => p.name + '=' + Utils.getValueFromParameter(p))
                 ],
-                labels: this.state.labels.join(',')
+                labels: labels.join(',')
             })
             .then((submitted: Workflow) => (document.location.href = uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`)))
```
agilgur5, line 53: this hasn't changed, but also should perhaps use react-router navigation instead

agilgur5 (commented):
mostly looks good to me, though I've requested some further improvements in-line. basically to use named functions, async/await, optional chaining, type inference, and some small optimizations

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
agilgur5: or this gets removed entirely if `submit` is [rewritten](https://github.com/argoproj/argo-workflows/pull/11803#discussion_r1323481200) as it gets in-lined

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
agilgur5: or this gets removed entirely if `submit` is [rewritten](https://github.com/argoproj/argo-workflows/pull/11803#discussion_r1323481200) as it gets in-lined

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
agilgur5: could also set a variable for the function name as `retry` does instead of the `submitted` variable. that variant is maybe cleaner actually?

agilgur5:
```suggestion
        try {
            const submitted = !props.isArchived
                ? await services.workflows.resubmit(props.workflow.metadata.name, props.workflow.metadata.namespace, opts);
                : submitted = await services.workflows.resubmitArchived(props.workflow.metadata.uid, props.workflow.metadata.namespace, opts);
           handleSuccess(submitted);
       } catch (newError) {
           handleError(newError);
       }
```

yea the ternary is definitely cleaner

agilgur5:
Most of the PR description verbatim copies my previous PRs (e.g. #11800). That's fine, but please cite your sources, especially when it is verbatim copy+paste.

agilgur5:
Would either of you like to review the other PRs I have open for the refactor? Could help speed up getting them merged 🙂

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
tetora1053: I reverted handleSuccess and handleError funcs to inline because there was no need to extract functions by using await/async and try/catch as you pointed out. thanks!

ui/src/app/workflows/components/submit-workflow-panel.tsx
tetora1053: I changed `templates` from state to const and memoized the select options created based on 'templates'
Could you review if the changes are appropriate?
https://github.com/argoproj/argo-workflows/pull/11803/commits/9050132b89d453d9faaef5ce77af15590bee6a2d

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
agilgur5: Ah I think the current pattern is to use the History API with `history.push`. We are on an older version of react-router though (I'm working on upgrading it -- have to finish removing `BasePage` first), so I don't think the `useHistory` hook is available. Other components use `history` from props currently

tetora1053: thanks! I was just researching how to write in current react router version.
I'll try!

tetora1053: I tryed passing `history` in props and replaced 'document.location.href' with 'history.push', but it didn't work as I expected :(

```props.history.push(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);```

tetora1053: If the scope of the changes is likely to be larger than initially expected, it might be good to reconsider the approach after your react-router version up I think.

agilgur5: > `props.history.push(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`);`

you do still need `uiUrl`, as in `props.history.push(uiUrl("..."))`. though if `BASE_HREF` isn't set (and it's not when running locally), all it does is [add a `/`](https://github.com/argoproj/argo-workflows/blob/55bb51885d2a6690727f97dce25fffef1afb34f2/ui/src/app/shared/base.ts#L3) to the beginning.

I'm not sure if that fixes this?

> but it didn't work as I expected :(

how did it end up working? just so we know for future reference next time somebody tries tackling this.

> If the scope of the changes is likely to be larger than initially expected, it might be good to reconsider the approach after your react-router version up I think.

Yea I'm good approving as is, though if you can figure it out, would be a nice addition too 🙂

ui/src/app/workflows/components/submit-workflow-panel.tsx
agilgur5: I think the memoization might not work as you're comparing `templates`, which is an array. `useMemo` [uses `Object.is`](https://react.dev/reference/react/useMemo#skipping-expensive-recalculations) to check equality of the dependencies, which uses [reference equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#description) for objects (and arrays). So it will probably get recalculated every time as `templates` is recalculated it above it every time.

I'm fine with no `useMemo` and just having it be a normal variable, as the performance is not really significant. Otherwise can try using `props.templates` as the dependency instead (as `templates` is wholly derived from `props.templates`, they are semantically equivalent dependencies), which I _think_ would pass reference equality checks as it [does not change after the initial fetch of the `WorkflowTemplate`](https://github.com/argoproj/argo-workflows/blob/55bb51885d2a6690727f97dce25fffef1afb34f2/ui/src/app/workflow-templates/components/workflow-template-details/workflow-template-details.tsx#L133)

agilgur5:
Btw, the unrelated unit test failure was fixed in #11816. may want to rebase on top (or merge)

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
tetora1053: Thanks a lot!!
by `props.history.push(uiUrl("..."))`,  it works for resubmit and retry component!
but in the resubmit case, the icon of the node in the workflow graph at the destination of the location does not change.
before the modification, the icon of the node in the graph panel had changed to progress status.

Screen Shot 2023-09-14 at 22 04 05

I confirmed that the workflow pod outputs logs, so I think the workflow itself was working.
do I need change some state of the workflow graph node?

and in the retry case, this degradation doesnt occur. 🤒

tetora1053: > Does it eventually update itself to in progress?

no, the icon does not change at all from this image
Screen Shot 2023-09-14 at 22 33 50

before the modification, the icon had changed yellow -> blue -> green

agilgur5: Huh. Since this is on the same page (just different Workflow), I wonder if maybe the page is not quite unloading correctly and still has the old state. The effects seem to have [proper disposers](https://github.com/argoproj/argo-workflows/blob/7aedf973356c8b57510a554b6b759f2684f88839/ui/src/app/workflows/components/workflow-details/workflow-details.tsx#L386) and [dependencies](https://github.com/argoproj/argo-workflows/blob/7aedf973356c8b57510a554b6b759f2684f88839/ui/src/app/workflows/components/workflow-details/workflow-details.tsx#L427), so I don't see anything strange off the top of my head.

At this point, especially since it potentially involves a whole other page's logic, I think we should punt on the navigation improvements for now. Appreciate your efforts and investigation here! Hopefully that will help the next person to tackle this too (which could be yourself 😉 ).

agilgur5 (commented):
Since we're [punting](https://github.com/argoproj/argo-workflows/pull/11803#discussion_r1326410891) on navigation improvements for now, the one remaining thing is the [memoization fixes](https://github.com/argoproj/argo-workflows/pull/11803#discussion_r1325243496)

ui/src/app/workflows/components/submit-workflow-panel.tsx
tetora1053: thanks for elaborating on useMemo!
I now have a better understanding. (I didn't understand it before😃)
I fixed it and confirmed that variables templates and templatesOptions are not recalculated by using console.log.

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
tetora1053: In the retry component, the state appears to be changing as expected, so the difference in implementation between retry and resubmit may be a clue, though I haven't been able to found out.

tetora1053: If I figure it out, I'll make a PR🔥

agilgur5 (approved):
LGTM. Thanks for getting the refactors, iteration, and performance improvements done!

ui/src/app/workflows/components/submit-workflow-panel.tsx
agilgur5: Glad to help 🙂
New version looks correct to me!

ui/src/app/workflows/components/resubmit-workflow-panel.tsx
agilgur5: > so the difference in implementation between retry and resubmit may be a clue

So a retry does not create a new Workflow, it just reruns failed steps. A resubmit creates a wholly new Workflow, effectively a clone of the old one.

So for a retry, the URL does not change. For a resubmit it does.
For the retry, the existing `ListWatch` might be able to pick up the change automatically. For a resubmit, an entirely new request is needed for the new Workflow.

Maybe those details may help you debug?
I already knew that though (I wrote a bit of the documentation on retry vs resubmit in #11625) and couldn't figure it out, though I didn't stare too hard

terrytangyuan (approved):

tetora1053:
@agilgur5 
I want to express my gratitude.
this was my first contribution to OSS, and your constructive feedback and guidance made the process much smoother.
thank you for taking much time to review my pull request and for your invaluable advice.

having my pull request merged means a lot to me.
I'm looking forward to contributing more in the future!

agilgur5:
Thanks for your heartfelt words @tetora1053! ❤️  Experiences like these are one of the reasons I continue to work in OSS 🙂 

Glad I could help make it a great experience for you and hope to see more of your contributions! Big congrats on your first contribution 🎉 🎉 

Thanks @toyamagu-2021 for getting more folks involved as well!

PR #11805, opened 2023-09-12
docs: Remove mentoring related docs and issue template

terrytangyuan:
Closes #11802

PR #11753, opened 2023-09-06
ci: reduce test & CI flakes with retries + longer waits & timeouts

agilgur5:
Partial fix for #10807

### Motivation

- signal_test.go was flaking quite a bit in CI
   - examples (all same test, same error): [1](https://github.com/argoproj/argo-workflows/actions/runs/6090397163/job/16525212870?pr=11752), [2](https://github.com/argoproj/argo-workflows/actions/runs/6090397163/job/16525214144?pr=11752), [3](https://github.com/argoproj/argo-workflows/actions/runs/6057348108/job/16438451395?pr=11727), [4](https://github.com/argoproj/argo-workflows/actions/runs/6024154407/job/16342412510?pr=11716), [5](https://github.com/argoproj/argo-workflows/actions/runs/5910880044/job/16032931699?pr=11568), [6](https://github.com/argoproj/argo-workflows/actions/runs/5910880485/job/16032933770?pr=11567), [7](https://github.com/argoproj/argo-workflows/actions/runs/6066706542/job/16457612480), [8](https://github.com/argoproj/argo-workflows/actions/runs/6008824810/job/16297128151), [9](https://github.com/argoproj/argo-workflows/actions/runs/6076177904/job/16483849857), [10](https://github.com/argoproj/argo-workflows/actions/runs/5979752022/job/16224632564?pr=11637), [11](https://github.com/argoproj/argo-workflows/actions/runs/5922167534/job/16055845745), [12](https://github.com/argoproj/argo-workflows/actions/runs/6092222043/job/16530030316), [13](https://github.com/argoproj/argo-workflows/actions/runs/6088747445/job/16520226969), [14](https://github.com/argoproj/argo-workflows/actions/runs/6019750405/job/16330112382), [15](https://github.com/argoproj/argo-workflows/actions/runs/6005141116/job/16287356878), [16](https://github.com/argoproj/argo-workflows/actions/runs/5980829119/job/16227812975), [17](https://github.com/argoproj/argo-workflows/actions/runs/5959534969/job/16165466761), [18](https://github.com/argoproj/argo-workflows/actions/runs/5898116342/job/15998761453), [19](https://github.com/argoproj/argo-workflows/actions/runs/5886793569/job/15965217053)
     - the most frequent test flake I've seen by a good margin (~20x (?) more than any other one, anecdotally)

- the rest are incidental, but related, fixes for flakes that occurred in CI runs for this PR _itself_ (and that have been seen in CI runs elsewhere too)

<details>
[...]

PR #10307, opened 2023-01-04
fix: Fix not working Running state lifecycle hooks in dag task. Fixes #9897

GeunSam2:
Fixes #9897 

Please do not open a pull request until you have checked ALL of these:

* [x] Create the PR as draft .
* [x] Run `make pre-commit -B` to fix codegen and lint problems.
* [x] Sign-off your commits (otherwise the DCO check will fail).
* [x] Use [a conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) (otherwise the commit message check will fail).
* [x] "Fixes #" is in both the PR title (for release notes) and this description (to automatically link and close the issue).
* [x] ~Add unit or e2e tests. Say how you tested your changes. If you changed the UI, attach screenshots.~
* [x] Github checks are green.
* [x] Once required tests have passed, mark your PR "Ready for review".

If changes were requested, and you've made them, dismiss the review to get it reviewed again.

PR #11814, opened 2023-09-13
docs: remove no longer needed spelling exceptions

agilgur5:
### Motivation

Several of the `.spelling` exceptions are no longer needed

### Modifications

- in particular, since mentoring docs were removed in #11805, several of these are no longer needed
  - names and usernames, GSoC, etc

- also remove other exceptions that are no longer used
  - words that are in backticks do not need exceptions

### Verification

`make docs-lint` and `make docs-spellcheck` pass


agilgur5:
I bumped `TestStopBehavior` again here as it failed here. I bumped it in #11753, which did dramatically reduce flakes, but there were still a few here [and there](https://github.com/argoproj/argo-workflows/actions/runs/6159386743/job/16714258017).

Now there's a different CI [failure](https://github.com/argoproj/argo-workflows/actions/runs/6179936074/job/16775787988?pr=11814) that seems more like a bug and is not due to a timeout. Left a comment on https://github.com/argoproj/argo-workflows/pull/10307#issuecomment-1720028813 about it

terrytangyuan (commented):
Is it possible to automate? I don't think we can keep this file updated.

agilgur5:
Good question, though unfortunately it doesn't seem like [the library](https://github.com/lukeapage/node-markdown-spellcheck) has options for that. There are some related issues (didn't find quite the same) like https://github.com/lukeapage/node-markdown-spellcheck/issues/54 that have been open for several years.
There is a [manual workaround for sorting](https://github.com/lukeapage/node-markdown-spellcheck/issues/83#issuecomment-643215591), but there isn't an explicit way to output the "word list" for a manual workaround. As in, to automate, you'd need to check the entire corpus's set of words against the `.spelling` file, and if the `.spelling` list is not a subset, then can remove the extraneous ones. The main problem is that word boundaries can be defined differently, so ideally using the same library would be optimal.

I could potentially hack together a script that tries to use similar word boundaries to automate this, but it may not be perfect and likely not worth the effort at this time. Here's a simple implementation I wrote up quickly but that doesn't quite work (my GNU is a bit rusty):

```sh
all_words=$(tail -n +6 .spelling) # skip the first 5 lines that are comments
words=(${all_words}) # convert string to array
for word in ${words[*]}; do
  grep -rqv $word docs
done
```

It's pretty rare that we remove substantial portions of the docs, so usually there's few reasons for removal; normally it only grows. The removal of the mentoring page was a pretty big one (that also contained a lot of ignores for things like usernames etc) as I mentioned in my opening comment.

Ignore files in general (like `.gitignore`, `.dockerignore`, etc) don't often have automation around them 🤔

jmeridth (approved):

juliev0:
Won't some of them be useful in the future even if they're not being used now? Then we would need to re-add those?

agilgur5:
90% of the ones removed here are mentoring related (with 80% of those being usernames), another 5% are from incorrect usage of backticks, and then a few unrelated. So I don't think most would ever be needed again, but we can always add them back if we ever need them.

juliev0 (approved):

Issue #8489, opened 2022-04-26
Artifact Streaming: implement for drivers other than S3


juliev0:
Currently, [this](https://github.com/argoproj/argo-workflows/issues/8396) feature has only been implemented the "right way" for S3. The other drivers are downloading the file first and then streaming it.

# Use Cases

When would you use this?

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

terrytangyuan:
Artifactory streaming is also available for HTTP/Artifactory in v3.5. Added some docs https://github.com/argoproj/argo-workflows/pull/11863

PR #11823, opened 2023-09-15
feat: Support artifact streaming for HTTP/Artifactory artifact driver

terrytangyuan:
Part of https://github.com/argoproj/argo-workflows/issues/8489.

### DIFF
--- a/workflow/artifacts/http/http.go
+++ b/workflow/artifacts/http/http.go
@@ -75,16 +75,17 @@ func (h *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string) error {
 func (h *ArtifactDriver) OpenStream(inputArtifact *wfv1.Artifact) (io.ReadCloser, error) {
 	var req *http.Request
 	var url string
+	var err error
 	if inputArtifact.Artifactory != nil && inputArtifact.HTTP == nil {
 		url = inputArtifact.Artifactory.URL
-		req, err := http.NewRequest(http.MethodGet, url, nil)
+		req, err = http.NewRequest(http.MethodGet, url, nil)
 		if err != nil {
 			return nil, err
 		}
 		req.SetBasicAuth(h.Username, h.Password)
 	} else {
 		url = inputArtifact.HTTP.URL
-		req, err := http.NewRequest(http.MethodGet, url, nil)
+		req, err = http.NewRequest(http.MethodGet, url, nil)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


workflow/artifacts/http/http.go
```diff
@@ -72,9 +72,41 @@ func (h *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string) error {
...
+	var err error
+	if inputArtifact.Artifactory != nil && inputArtifact.HTTP == nil {
+		url = inputArtifact.Artifactory.URL
+		req, err = http.NewRequest(http.MethodGet, url, nil)
+		if err != nil {
+			return nil, err
+		}
+		req.SetBasicAuth(h.Username, h.Password)
+	} else {
+		url = inputArtifact.HTTP.URL
```
juliev0, line 87: just looking at this code it seems like HTTP can be nil and Artifactory can be nil on this line - should we add a case for that even if currently maybe it's not possible to enter this code for that case? (in case it is in the future)

workflow/artifacts/http/http.go
```diff
@@ -72,9 +72,41 @@ func (h *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string) error {
...
+		req, err = http.NewRequest(http.MethodGet, url, nil)
+		if err != nil {
+			return nil, err
+		}
+		for _, h := range inputArtifact.HTTP.Headers {
+			req.Header.Add(h.Name, h.Value)
+		}
+		if h.Username != "" && h.Password != "" {
+			req.SetBasicAuth(h.Username, h.Password)
+		}
```
juliev0, line 97: can we consolidate the logic that's copy/pasted from `Load()` into a single place?

juliev0: is the only actual difference the very last part in fact?

workflow/artifacts/http/http.go
terrytangyuan: Thanks. Added

workflow/artifacts/http/http.go
terrytangyuan: Done. Yes there's only slight difference.

terrytangyuan: Another difference is that the response body cannot be closed in `OpenStream()` since we will consume the stream later.

juliev0 (approved):

PR #11841, opened 2023-09-19
docs: add hint for limiting the number of wfs

mikutas:
Fixes #TODO

### Motivation

As platform owner we have enabled [Workflow Archive](https://argoproj.github.io/argo-workflows/workflow-archive/) in our company.
On the other hand, some developer set their CronWorkflow's history limits too large and caused a flood of (completed) pods and workflows, then as a result memory consumption of workflow-controller and other some controllers increased.

I took some times to identify the root cause is CronWorkflow's history limits (are too large, and large limit is not necessary ), so I'd like to suggest to refer that in cost-optimisation.md.

### DIFF
--- a/docs/cost-optimisation.md
+++ b/docs/cost-optimisation.md
@@ -52,6 +52,7 @@ Limit the total number of workflows using:
 * Active Deadline Seconds - terminate running workflows that do not complete in a set time. This will make sure workflows do not run forever.
 * [Workflow TTL Strategy](fields.md#ttlstrategy) - delete completed workflows after a set time.
 * [Pod GC](fields.md#podgc) - delete completed pods. By default, Pods are not deleted.
+* [`CronWorkflow`'s history limit](cron-workflows.md#cronworkflow-options) - delete completed/failed workflows which exceed the limit.
 
 Example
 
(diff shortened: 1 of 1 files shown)


docs/cost-optimisation.md
```diff
@@ -52,6 +52,7 @@ Limit the total number of workflows using:
 * Active Deadline Seconds - terminate running workflows that do not complete in a set time. This will make sure workflows do not run forever.
 * [Workflow TTL Strategy](fields.md#ttlstrategy) - delete completed workflows after a set time.
 * [Pod GC](fields.md#podgc) - delete completed pods. By default, Pods are not deleted.
+* [`CronWorkflow`'s history limit](cron-workflows.md#cronworkflow-options) - delete completed/failed workflows which exceed the limit.
```
agilgur5, line 55:
```suggestion
* [`CronWorkflow` history limits](cron-workflows.md#cronworkflow-options) - delete successful or failed workflows which exceed the limit.
```

- minor grammar changes
  - `CronWorkflow` with no possessive. Same parallelism as the other 3 bullets which refer to a specific field without a possessive 
  - multiple limits, so plural "limits"
  
- "completed" -> "successful" -- there is no "completedJobsHistoryLimit", only "successfulJobsHistoryLimit", which are not identical.

agilgur5 (commented):
I think this is a good recommendation to add. Tiny changes below

agilgur5 (approved):
Thanks for the add!

terrytangyuan (approved):

PR #11842, opened 2023-09-19
feat: added mongodb template for easy addition/updation/deletion of mongo document from work…

sajjanjyothi:
Add mongo operations from workflow itself as a template.Will be helpful for projects which are creating/updating mongo documents from workflows

Fixes #TODO

### Motivation

In my project I have to update mongodb database based on some worflow steps outputs. Created a mongo template for this activity so that it can be re used by any one who has the same use case.

### Modifications

Added mongodb template as part of workflow

### Verification

Tested with a sample workflow locally


agilgur5:
> Why not build some reusable plugins instead of adding a specific template type?

Agreed. This is tightly coupled to a very specific provider and introduces dependencies for that provider as well.
I think this is more suited to an executor plugin as well.

terrytangyuan:
I don't think we would introduce a new template type. I'd suggest implementing your plugin and add a link in https://github.com/argoproj/argo-workflows/blob/master/docs/plugin-directory.md

sajjanjyothi:
Sure,Let me move this to a plugin

Issue #11542, opened 2023-08-08
custom metrics gc no longer cleans old emitted metrics 


fhochleitner:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

After upgrading to v3.4.9 (and latest) custom metrics will no longer get cleaned up after "metricsTTL" time has been reached and will stay exposed at the metrics endpoint endlessly. 

e.g. example workflow emits the following metric `argo_workflows_result_task_counter{id="b68baed7-b2a5-4718-bac4-221d559c750f",status="Succeeded",task_name="example-metrics"} 1` which will stay until pod get recreated. 

downgrade to v3.4.8 without any configuration change leads to expected behaviour of metric no longer being exposed after TTL time has been reached. 

workflow-controller-configmap for reference: 

```parallelism: 10
namespaceParallelism: 5
mainContainer:
  imagePullPolicy: IfNotPresent
  resources:
    limits:
      cpu: 1
      memory: 4Gi
    requests:
      cpu: 250m
      memory: 512Mi
  securityContext:
    runAsGroup: 0
    runAsNonRoot: true
    runAsUser: 1000
executor:
  imagePullPolicy: IfNotPresent
  securityContext:
    runAsGroup: 0
    runAsNonRoot: true
    runAsUser: 1000
artifactRepository:
  archiveLogs: true
  s3:
    --- omitted ---
metricsConfig:
  enabled: true
  path: /metrics
  port: 9090
  metricsTTL: 5m
  ignoreErrors: false
  secure: false
persistence:
  archive: true
  archiveTTL: 120d
  connectionPool:
    maxIdleConns: 50
  nodeStatusOffLoad: true
  postgresql:
    --- omitted --- 
workflowDefaults:
  spec:
    securityContext:
      fsGroup: 0
    ttlStrategy:
      secondsAfterCompletion: 3600
sso:
 --- omitted ---
nodeEvents:
  enabled: true```


### Version

v3.4.9, latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: metrics-regressions-example-
[...]

Issue #11541, opened 2023-08-07
workflow-controller-configmap/parallelism setting not working in 3.4.9


andef:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I have pasted a small bash script that starts minikube and installs argo with 3.4.9 and configuring parallelism to 5.
Then submit 10 workflows and watches the result. In 3.4.8 this works as expected, but not in 3.4.9 or latest.

### Version

3.4.9

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```bash
#!/usr/bin/env bash

minikube delete
minikube start
kubectl create namespace argo

# Install argo.
#kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.4.8/install.yaml
kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.4.9/install.yaml
#kubectl apply -n argo -k https://github.com/argoproj/argo-workflows/manifests/cluster-install

echo "Wait for controller to start..."
kubectl wait --for=condition=available --timeout=60s deployment/workflow-controller -n argo

# Add parallelism: 5 to configmap.
kubectl patch configmap workflow-controller-configmap -n argo -p '{"data":{"parallelism": "5"}}'

# Restart controller.
kubectl -n argo get pod|grep controller|cut -d " " -f1|xargs kubectl -n argo delete pod

echo "Wait for server to start..."
kubectl wait --for=condition=available --timeout=60s deployment/argo-server -n argo

WORKFLOW=$(cat <<EOF
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: sleep-
  namespace: argo
spec:
  entrypoint: sleep
  templates:
  - name: sleep
    container:
      image: busybox:latest
      command: [sleep]
      args: ["30"]
EOF
)

for _ in {1..10}
do
   echo "${WORKFLOW}" | argo submit -n argo - 
done

watch argo -n argo list

```

### Logs from the workflow controller

```text
N/A
```

### Logs from in your workflow's wait container

```text
N/A
```

terrytangyuan:
[...]

Issue #11177, opened 2023-06-04
Argo DB init conflict when deploy workflow-controller with multiple replicas


astraw99:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

DB init encounter conflict, when deploy the workflow-controller with multiple replicas (leader election enabled by default).
```
kubectl get po -n argo

NAME                                   READY   STATUS             RESTARTS   AGE
argo-server-849b557f68-9q8wk           1/1     Running            0          10m
argo-server-849b557f68-cjkw2           1/1     Running            0          10m
argo-server-849b557f68-v5vmp           1/1     Running            0          10m
minio-588f94977f-9bd6r                 1/1     Running            0          10m
workflow-controller-7796ff4cbb-87kbj   0/1     CrashLoopBackOff   6          10m
workflow-controller-7796ff4cbb-dfjgb   0/1     CrashLoopBackOff   6          10m
workflow-controller-7796ff4cbb-hwbd4   0/1     CrashLoopBackOff   6          10m
```
The conflict log is pasted below.

I will raise a PR to fix this: move the DB init logic to after the leader election completed.

### Version

v3.4.6

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
workflow-controller crashed by `kubectl get po -n argo`:

NAME                                   READY   STATUS             RESTARTS   AGE
workflow-controller-7796ff4cbb-87kbj   0/1     CrashLoopBackOff   6          10m
workflow-controller-7796ff4cbb-dfjgb   0/1     CrashLoopBackOff   6          10m
workflow-controller-7796ff4cbb-hwbd4   0/1     CrashLoopBackOff   6          10m
```

### Logs from the workflow controller

```text
kubectl logs -n argo workflow-controller-7796ff4cbb-87kbj

time="2023-05-26T09:11:26Z" level=info msg="index config" indexWorkflowSemaphoreKeys=true
time="2023-05-26T09:11:26Z" level=info msg="cron config" cronSyncPeriod=10s
time="2023-05-26T09:11:26Z" level=info msg="Memoization caches will be garbage-collected if they have not been hit after" gcAfterNotHitDuration=30s
[...]

PR #11855, opened 2023-09-21
fix: refactor `runConfigMapWatcher` to use Informers. Fixes #11657

### LINKED ISSUE #11463: Controller not picking up configmap updates when running with `--managed-namespace`

### LINKED ISSUE #11657: Controller: `invalid config map object received in config watcher. Ignored processing` with CPU spike

juranir:
Fixes #11657 
Fixes #11463 

Supersedes / Closes #11799 

### Motivation

After some time running, when the connection to k8s API is broken (timeout for example), the watcher cannot handle the exception increases the CPU usage, and floods the log with the message `level=error msg="invalid config map object received in config watcher. Ignored processing"`. After reaching the pod limit, it'll be restarted. See #11657  for details.

### Modifications

I replaced the function with a new one using informers (many thanks @agilgur5 ).
I tried to use the code pattern used by other functions, but please let me know if something can be improved.
I also changed the namespace used to "managed-namespace" instead of "namespace" (This change will solve #11463 )

### Verification
Manual changes in the ConfigMap are being watched and applied;
An instance with this change is running for a while;
We also did some integration and performance tests;

shell


workflow/controller/controller.go
```diff
@@ -392,38 +395,34 @@ func (wfc *WorkflowController) initManagers(ctx context.Context) error {
...
-		case event := <-retryWatcher.ResultChan():
-			cm, ok := event.Object.(*apiv1.ConfigMap)
-			if !ok {
-				log.Errorf("invalid config map object received in config watcher. Ignored processing")
-				continue
+	informer.AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			cm, err := meta.Accessor(obj)
+			if err != nil {
+				return false
```
juliev0, line 411: should we log something here?

juliev0 (commented):
just one small comment, but otherwise looks good!

agilgur5:
> Fixes #11463

> I also changed the namespace used to "managed-namespace" instead of "namespace" (This change will solve #11463 )

I don't see how this fixes #11463? The RetryWatcher was already listening to `managedNamespace`.
#11463 also points out the opposite -- the Controller's ConfigMap is _not_ in the `managedNamespace`, it's in the Controller's own namespace. Using `managedNamespace` is the bug that #11463 points out.
Per my review comments in https://github.com/argoproj/argo-workflows/pull/11799#issuecomment-1714735706, there's also an issue with the semaphore notification that will have to be moved, as semaphores _are_ in the `managedNamespace`

workflow/controller/controller.go
```diff
@@ -392,38 +395,34 @@ func (wfc *WorkflowController) initManagers(ctx context.Context) error {
...
+func (wfc *WorkflowController) newConfigMapCtrlInformer() cache.SharedIndexInformer {
 	defer runtimeutil.HandleCrash(runtimeutil.PanicHandlers...)
 
 	ctx := context.Background()
-	retryWatcher, err := apiwatch.NewRetryWatcher("1", &cache.ListWatch{
-		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-			return wfc.kubeclientset.CoreV1().ConfigMaps(wfc.managedNamespace).Watch(ctx, metav1.ListOptions{})
-		},
+
+	informer := v1.NewConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
```
agilgur5, line 403: `wfc.GetManagedNamespace()` is better than `wfc.managedNamespace`, but still not quite correct, per my above comment. I'm fine if this PR doesn't fix the other bug, so either we can leave as is or we need to modify this to correctly fix the other bug

workflow/controller/controller.go
```diff
@@ -392,38 +395,34 @@ func (wfc *WorkflowController) initManagers(ctx context.Context) error {
...
+			cm, err := meta.Accessor(obj)
+			if err != nil {
+				return false
 			}
-			log.Debugf("received config map %s/%s update", cm.Namespace, cm.Name)
-			if cm.GetName() == wfc.configController.GetName() && wfc.namespace == cm.GetNamespace() {
+			return cm.GetName() == wfc.configController.GetName()
+		},
+		Handler: cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(_, obj interface{}) {
```
agilgur5, line 416: technically speaking, an `AddFunc` and `DeleteFunc` would be good too, though I don't think the previous RetryWatcher handled those scenarios either. They are also more complex, so I think this PR is probably good as is

agilgur5 (changes requested):
There's a bit of an issue here with the semaphore configmap notifications. See in-line [below](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1341910749)

workflow/controller/controller.go
juliev0: I'd vote for changing it to `wfc.namespace` (thanks for noticing that @agilgur5)

workflow/controller/controller.go
juranir: We can add a log here, even if it is as copy+past I also agree a log may be helpful.

workflow/controller/controller.go
juranir: Great, I'll update it.

workflow/controller/controller.go
```diff
@@ -109,6 +108,7 @@ type WorkflowController struct {
 	wftmplInformer        wfextvv1alpha1.WorkflowTemplateInformer
 	cwftmplInformer       wfextvv1alpha1.ClusterWorkflowTemplateInformer
 	podInformer           cache.SharedIndexInformer
+	configMapCtrlInformer cache.SharedIndexInformer
 	configMapInformer     cache.SharedIndexInformer
```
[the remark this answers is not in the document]
juranir: Great, I'll update it.

workflow/controller/controller.go
```diff
@@ -392,38 +395,34 @@ func (wfc *WorkflowController) initManagers(ctx context.Context) error {
...
+				cm := obj.(*apiv1.ConfigMap)
 				log.Infof("Received Workflow Controller config map %s/%s update", cm.Namespace, cm.Name)
 				wfc.UpdateConfig(ctx)
-			}
-			wfc.notifySemaphoreConfigUpdate(cm)
-		case <-stopCh:
-			return
-		}
-	}
+				wfc.notifySemaphoreConfigUpdate(cm)
```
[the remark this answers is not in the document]
juranir: @agilgur5, I'm so sorry, but I'm not sure if I understood well your comment. Is it only about keeping `namespace` instead `managedNamespace`?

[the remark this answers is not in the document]
agilgur5: Not entirely.

`notifySemaphoreConfigUpdate` must run on the `managedNamespace`, yes, as all semaphore ConfigMaps are in the managed namespace.

Further, your `FilterFunc` limits the events to only the Controller's own ConfigMap, while `notifySemaphoreConfigUpdate` needs to run on _all_ ConfigMaps in the managed namespace.

The previous code before your modifications was kind of strangely coupled together (and had a bug for managed namespaces) as those are unrelated features, so I imagine you may have missed that one-line nuance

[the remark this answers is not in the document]
juliev0: @juranir there's some complexity here since [this](https://github.com/argoproj/argo-workflows/pull/11799) PR is also out at the same time. Both of you seem to have covered part of the equation, but then @agilgur5 has also given a lot of thought on how to best optimize and consolidate code [here](https://github.com/argoproj/argo-workflows/pull/11799#issuecomment-1742309369). It sounds like @agilgur5 is willing to take both of these PRs forward unless you would like to go through his comments in that issue and try to take it on yourself? (@agilgur5 please let me know if I'm mistaken)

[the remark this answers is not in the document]
agilgur5: yep that's correct. @juranir seemed interesting, so I'm very willing to assist him as well 🙂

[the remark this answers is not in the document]
juranir: Thank you very much for all your help, @agilgur5  is very nice.

If I understand correctly, the idea is to refactor the current `newConfigMapInformer` function to handle both operations (plugins and the cm controller itself), so I think this [commit](https://github.com/argoproj/argo-workflows/pull/11855/commits/1a6110a7fae92ff3a6877f114f4c8564f4079164) might be on the way (I hope in the right one).

I also added a log as commented by @juliev0 

PS. One test failed, but I'm not sure about the reason or if it's related.

workflow/controller/controller.go
```diff
@@ -1053,42 +1017,62 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) cache.SharedI
 }
 
 func (wfc *WorkflowController) newConfigMapInformer() cache.SharedIndexInformer {
-	indexInformer := v1.NewFilteredConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
-		indexes.ConfigMapLabelsIndex: indexes.ConfigMapIndexFunc,
-	}, func(opts *metav1.ListOptions) {
-		opts.LabelSelector = common.LabelKeyConfigMapType
+
+	indexInformer := v1.NewConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
```
agilgur5, line 1021: So I mentioned [before](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1342787539) that there are _two_ namespaces to watch, and both are used for different purposes. This is on the managed namespace, which is not necessarily the same as Argo's own namespace. The `workflow-controller-configmap` will _always_ be in Argo's own namespace, but semaphores etc may be in a separate namespace. As in, if you deploy Argo in one namespace but run Workflows in a _different_ namespace(s).

So we may very well need two separate Informers to handle both namespace configurations, unless you can listen to both namespaces in one.

workflow/controller/controller.go
```diff
@@ -1053,42 +1017,62 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) cache.SharedI
...
 	})
-	log.WithField("executorPlugins", wfc.executorPlugins != nil).Info("Plugins")
-	if wfc.executorPlugins != nil {
-		indexInformer.AddEventHandler(cache.FilteringResourceEventHandler{
-			FilterFunc: func(obj interface{}) bool {
-				cm, err := meta.Accessor(obj)
-				if err != nil {
-					return false
+
+	isPluginCM := true
```
agilgur5, line 1025: the functions below can be called multiple times, so setting `isPluginCM` at the top-level here would be insufficient (and likely encounter race conditions as well). It would have to be set in each function individually.

workflow/controller/controller.go
```diff
@@ -1053,42 +1017,62 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) cache.SharedI
 }
 
 func (wfc *WorkflowController) newConfigMapInformer() cache.SharedIndexInformer {
-	indexInformer := v1.NewFilteredConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
-		indexes.ConfigMapLabelsIndex: indexes.ConfigMapIndexFunc,
-	}, func(opts *metav1.ListOptions) {
-		opts.LabelSelector = common.LabelKeyConfigMapType
+
+	indexInformer := v1.NewConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
+		cache.NamespaceIndex: cache.MetaNamespaceIndexFunc,
```
agilgur5, line 1022: I might be missing something here -- what was the purpose of changing the index?

workflow/controller/controller.go
```diff
@@ -1053,42 +1017,62 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) cache.SharedI
...
+			if cm.GetName() == wfc.configController.GetName() {
+				isPluginCM = false
+			} else {
+				isPluginCM = true
+				log.WithField("executorPlugins", wfc.executorPlugins != nil).
+					Info("Plugins")
+			}
+
+			return (cm.GetName() == wfc.configController.GetName()) ||
+				(cm.GetLabels()[common.LabelKeyConfigMapType] == common.LabelValueTypeConfigMapExecutorPlugin)
```
agilgur5, line 1047: so there is a _third_ type of ConfigMap that isn't handled here: the semaphore config update

agilgur5 (changes requested):
Sorry for the delay on my end, I was sick for a whole week and then playing catch up.

There's two main issues with the below code, per the in-line comments:
1. There are potentially two different namespaces to watch, not just one
2. There are three different types of ConfigMaps to watch, not just two

workflow/controller/controller.go
juranir: Do you think it would be better to convert this validation into a method and call it wherever we want?

workflow/controller/controller.go
juranir: As far as I know, we can only watch 1 or all namespaces, so I think we have 2 possibilities (based on my limited knowledge):
#1. Create another function and duplicate the code;
#2. Create an array with all necessary namespaces and move the current code into this loop;

So far I would prefer #2, but I'm not sure if it may cause some problem

workflow/controller/controller.go
agilgur5: That's one option, yea. `isPluginCM` as a function makes sense to me. Could do same for `isControllerCM` too.
They're both one-liners, but given the amount of times they're used here, helper functions would make it more readable and less error-prone.

workflow/controller/controller.go
juranir: Sorry for asking, but I would like to clarify, `notifySemaphoreConfigUpdate` should be called for all CM caught by this informer or there is some exception (in the [current code](https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/controller.go#L422) I think so)?

agilgur5: the current code is all CMs in the managed namespace. I don't think there's a specific label for them right now (but maybe there should be)

tooptoop4:
is this ok to be merged?

agilgur5:
> is this ok to be merged?

No, my [previous review](https://github.com/argoproj/argo-workflows/pull/11855#pullrequestreview-1698131756) has yet to be addressed

agilgur5:
Hey @juranir do you have any time to continue working on this next week? If not, I'd like to take over the work as #11657 has gotten to be one of the top issues (seemingly due to k8s 1.27 upgrades?).
I'll ofc credit you with the existing work and leave you as co-author etc 🙂

sstarcher:
@agilgur5 any updates on this it's becoming very problematic.

juranir:
The proposed suggestions by @agilgur5 and @juliev0  were implemented

agilgur5 (changes requested):
Thanks for picking this up again @juranir! I had to get a solid block of time to re-familiarize myself with this code, but I reviewed everything now.

For other reviewers, note that turning on "Hide Whitespace" in the GH diff is very helpful (tbh, I am thinking of using a different reviewing UI entirely given how many complex PRs I review now and how many insufficiencies GH has).

This does look much better now. Summary of my in-line comments:
- Biggest thing -- this is [missing](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1554360641) the `runConfigMapWatcher` removal. That might've gotten lost in the rebase?
- I think [unconditionally](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1554393709) creating both Informers would simplify the code. If we have two Informers on the same namespace, so be it. Informer resource usage tends to be more in the cache of data and filtering, rather than the overhead of the Informer itself.
- there are [some](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1554387769) [missing](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1554386025) namespace checks
- the label selector on the configmap informer [needs moving](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1554396079)

agilgur5:
While I have the context, and given the priority & impact of this, I'm just going to go ahead and make the changes requested above myself and push them to the PR

workflow/controller/controller.go
```diff
@@ -1270,43 +1285,62 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
 	return informer, nil
 }
 
-func (wfc *WorkflowController) newConfigMapInformer() (cache.SharedIndexInformer, error) {
-	indexInformer := v1.NewFilteredConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
+func (wfc *WorkflowController) newConfigMapInformer(ns string) (cache.SharedIndexInformer, error) {
+	indexInformer := v1.NewFilteredConfigMapInformer(wfc.kubeclientset, ns, 20*time.Minute, cache.Indexers{
 		indexes.ConfigMapLabelsIndex: indexes.ConfigMapIndexFunc,
 	}, func(opts *metav1.ListOptions) {
 		opts.LabelSelector = common.LabelKeyConfigMapType
```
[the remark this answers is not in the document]
agilgur5: oh this is the indexer... this might be necessary then. and you already have the check in `isPluginCM()`, my bad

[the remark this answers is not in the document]
agilgur5: wait no it _is_ the `TweakListOptionsFunc` (have to look a bit closely 😅). so this should be removed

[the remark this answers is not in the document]
agilgur5: Some E2Es [failed](https://github.com/argoproj/argo-workflows/actions/runs/8578095350/job/23511562662?pr=11855) without this. They seemed to miss the label without this... would've thought it would've picked up all configmaps without this? 🤔 

I might be misunderstanding something here, I'm certainly not an Informer/`client-go` expert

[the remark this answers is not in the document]
agilgur5: Oh, nvm, totally different bug; the wrong informer was being used -- fixed in [this commit](https://github.com/argoproj/argo-workflows/pull/11855/commits/191b441ec3730782ff5963e7feccfd5df9597bbc)

juranir:
> While I have the context, and given the priority & impact of this, I'm just going to go ahead and make the changes requested above myself and push them to the PR to make things faster (and reduce "telephone game" interpretations). In this review cycle I [now have permissions](https://github.com/argoproj/argoproj/issues/277) to do that.

No worries and many thanks for that.

@agilgur5  Please let me know if I can help with something.

workflow/controller/controller.go
```diff
@@ -1270,69 +1239,61 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
 	return informer, nil
 }
 
-func (wfc *WorkflowController) newConfigMapInformer() (cache.SharedIndexInformer, error) {
-	indexInformer := v1.NewFilteredConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
+var watchControllerSemaphoreConfigMaps = os.Getenv("WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS") != "false"
+
+func (wfc *WorkflowController) newConfigMapInformerManaged() (cache.SharedIndexInformer, error) {
+	indexInformer := v1.NewConfigMapInformer(wfc.kubeclientset, wfc.GetManagedNamespace(), 20*time.Minute, cache.Indexers{
 		indexes.ConfigMapLabelsIndex: indexes.ConfigMapIndexFunc,
```
agilgur5, line 1246: just noting here that this indexer [is used](https://github.com/argoproj/argo-workflows/blob/934bf179a703a365a6a5eddfa6edb788492eac79/workflow/controller/cache_gc.go#L28) for more than just plugins. that's why it previously only had a label selector on the label and did not specify the plugin type (and then why it filtered in the `FilterFunc` by plugin type instead -- it would populate the indexer but not the cache)

agilgur5 (commented):
I have substantially simplified this now (including refactoring out some common plugin handling functions) and fixed all the correctness issues. So this should be "good enough" now.

But:
1. There's a performance degradation, see in-line comment [below](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1556248023). **EDIT**: I have fixed that now with a third informer for semaphores
1. I'd like someone else to spot check this since I wrote a decent chunk of it. @juranir and @juliev0 can you both take a look?

workflow/controller/controller.go
```diff
@@ -1270,69 +1239,61 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
...
-	}, func(opts *metav1.ListOptions) {
-		opts.LabelSelector = common.LabelKeyConfigMapType
```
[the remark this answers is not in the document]
juranir: This option of getting a list of cm to process would be fantastic in terms of resource optimization. 

+1

workflow/controller/controller.go
```diff
@@ -1270,69 +1239,61 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
...
+
+	if !watchControllerSemaphoreConfigMaps {
+		return indexInformer, nil
+	}
+
+	_, err := indexInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(_, obj interface{}) {
+			cm := obj.(*apiv1.ConfigMap)
+			log.Infof("Received Workflow Controller config map %s/%s update", cm.GetNamespace(), cm.GetName())
+			wfc.UpdateConfig(ctx)
```
agilgur5, line 1291: there's an optimization that can be done here as well -- we can just pass the configmap directly. `UpdateConfig` does a `Get`, which is not necessary here.

that requires a bit of refactoring, including in the config controller, and is independent of this PR (the optimization probably could have been done before this PR as well), so I'm going to do that as a separate follow-up PR

workflow/controller/controller.go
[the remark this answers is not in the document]
agilgur5: Ok, I added the semaphore informer in [this commit](https://github.com/argoproj/argo-workflows/pull/11855/commits/f8e0872544602a96fae155f166d49709a32b28c2).

Now we're not changing the existing ConfigMap informer, which is nice, but tracking semaphores by name is a bit hacky 😕

workflow/controller/indexes/workflow_index.go
```diff
@@ -70,6 +71,16 @@ func WorkflowSemaphoreKeysIndexFunc() cache.IndexFunc {
 		if err != nil {
 			return nil, nil
 		}
-		return wf.GetSemaphoreKeys(), nil
+		keys := wf.GetSemaphoreKeys()
+		// store all keys while indexing as a side effect
+		for _, key := range keys {
+			semaphoreKeyMap[key] = struct{}{}
```
agilgur5, line 77: This is pretty hacky, I don't like this, but every way to track this seemed hacky 😕
This option had the least changes though, which is why I ended up going with it

workflow/controller/controller.go
```diff
@@ -1274,69 +1251,155 @@ func (wfc *WorkflowController) newConfigMapInformer() (cache.SharedIndexInformer
...
+		// only leave name and namespace, remove the rest as we don't use it
+		cm = cm.DeepCopy()
+		cm.ObjectMeta = metav1.ObjectMeta{
+			Name:      cm.Name,
+			Namespace: cm.Namespace,
 		}
+		cm.Data = map[string]string{}
+		cm.BinaryData = map[string][]byte{}
+		return cm, nil
+	})
```
agilgur5, line 1367: here's [an external example](https://squiggly.dev/2023/06/custom-cache-for-kubernetes-controllers/) of using a `cache.TransformFunc`, for reference

workflow/controller/controller.go
```diff
@@ -1270,71 +1238,150 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
...
-						WithField("name", cm.GetName()).
-						Info("Executor plugin updated")
-				},
-				DeleteFunc: func(obj interface{}) {
-					key, _ := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
-					namespace, name, _ := cache.SplitMetaNamespaceKey(key)
-					delete(wfc.executorPlugins[namespace], name)
-					log.WithField("namespace", namespace).WithField("name", name).Info("Executor plugin removed")
-				},
+	//nolint:errcheck // the error only happens if the informer was stopped, and it hasn't even started (https://github.com/kubernetes/client-go/blob/46588f2726fa3e25b1704d6418190f424f95a990/tools/cache/shared_informer.go#L580)
```
agilgur5, line 1253: see also https://github.com/argoproj/argo-workflows/pull/12847#discussion_r1556488443 -- will have a follow-up PR to remove the unnecessary error handling elsewhere.

a nice side-effect of this removal is that this will no longer merge conflict explicitly or implicitly during cherry-pick (as #12847 is breaking and may not be applied to 3.4.x/3.5.x)

agilgur5 (approved):
This LGTM to me now -- I've pretty much done every significant optimization that is specific and [isolated](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1556375760) to this change now
- `WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS` is not the most efficiently implemented (as I mentioned in the [initial commit](https://github.com/argoproj/argo-workflows/pull/11855/commits/43527e9c8073751cbd818b7bc48b7f14d90f5840)) as the Informers wouldn't need to be created at all in those cases (not just their event handlers), but given that's a workaround that could be removed (and env vars are explicitly documented as [unstable](https://argo-workflows.readthedocs.io/en/release-3.5/environment-variables/)), I didn't want to add too much code around it

There's also a comment I have [above](https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1556460508) on the semaphore keys hackery in the indexer, but I don't think there's a better way to do that without a breaking change (e.g. requiring labels on semaphore configmaps)

I'm also going to build a test image and push to a test registry for folks to try and test out

agilgur5:
> I'm also going to build a test image and push to a test registry for folks to try and test out

Pushed [`agilgur5/argo-workflow-controller:fix-configmap-informer`](https://hub.docker.com/layers/agilgur5/argo-workflow-controller/fix-configmap-informer/images/sha256-48c229a153ff18c9ab6fc94b4f62340e6155a6d68c25937c4ce733f73b444345?context=repo) for testing

workflow/controller/controller.go
```diff
@@ -1270,71 +1238,150 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
...
+		log.WithField("namespace", cm.GetNamespace()).
+			WithField("name", cm.GetName()).
+			WithError(err).
+			Error("failed to convert configmap to plugin")
+		return
+	}
+	if _, ok := wfc.executorPlugins[cm.GetNamespace()]; !ok {
+		wfc.executorPlugins[cm.GetNamespace()] = map[string]*spec.Plugin{}
+	}
+	wfc.executorPlugins[cm.GetNamespace()][cm.GetName()] = p
```
juliev0, line 1299: Don't we need thread safety between these functions?

juliev0: (if so, I see it wasn't there before)

agilgur5: Yea this part hasn't changed in this PR, it just got refactored into a common func. If you turn on "Hide Whitespace" on the GH diff, this line actually appears as unchanged (I mentioned the setting [earlier](https://github.com/argoproj/argo-workflows/pull/11855#pullrequestreview-1984237287), although it's not as useful after my refactor)

Technically speaking, they probably should be made thread-safe in case an `apply` and `delete` happen simultaneously, but realistically that will basically never happen. Also, this is ran on a single goroutine, so there aren't multiple threads modifying it. Safety is good to have just in case though.

juliev0: It's run on a single goroutine? Oh, I figured the `ResourceEventHandlerFuncs` spawned new goroutines - they don't?

agilgur5: Oh sorry I missed the notification for this comment; just manually checked and saw it.

I thought it would run on the same goroutine as the Informer's `.Run`, but no I think you're correct, it does seem to spawn new goroutines. It was a little non-trivial to trace back, but this is [the `client-go` line](https://github.com/kubernetes/client-go/blob/aa7909e7d7c0661792ba21b9e882f3cd6ad0ce53/tools/cache/shared_informer.go#L802) that spawns goroutines (`AddEventHandler` calls [`processor.addListener`](https://github.com/kubernetes/client-go/blob/aa7909e7d7c0661792ba21b9e882f3cd6ad0ce53/tools/cache/shared_informer.go#L612) and `wait` `Group`'s `Start` [spawns it](https://github.com/kubernetes/apimachinery/blob/d794766488ac2892197a7cc8d0b4b46b0edbda80/pkg/util/wait/wait.go#L70)).

In any case though, as that thread-safety bug had existed beforehand, I don't plan to fix it in this PR. Adding locks would also conventionally entail creating a new, thread-safe data structure for the executor plugins which becomes a sizeable change in and of itself that detracts a bit from the purpose of this PR.
There's a couple other follow-ups from this PR stemming from incidental findings as well like https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1556375760 and https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1556460508

Have you reviewed the other parts of this PR?

juliev0: sounds good. Let me go through the rest of the PR tomorrow. :)

workflow/controller/controller.go
```diff
@@ -1270,71 +1238,150 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
...
-		})
-		if err != nil {
-			return nil, err
+			UpdateFunc: func(_, obj interface{}) {
+				cm := obj.(*apiv1.ConfigMap)
+				wfc.applyPluginCM(cm, "updated")
+			},
+			DeleteFunc: func(obj interface{}) {
+				cm := obj.(*apiv1.ConfigMap)
+				wfc.deletePluginCM(cm)
```
juliev0, line 1276: This is a lot cleaner looking :)

workflow/controller/controller.go
```diff
@@ -1270,71 +1238,150 @@ func (wfc *WorkflowController) newPodInformer(ctx context.Context) (cache.Shared
...
+}
+
+var watchControllerSemaphoreConfigMaps = os.Getenv("WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS") != "false"
+
+func (wfc *WorkflowController) newConfigMapControllerInformer(ctx context.Context) cache.SharedIndexInformer {
+	indexInformer := v1.NewFilteredConfigMapInformer(wfc.kubeclientset, wfc.GetNamespace(), 20*time.Minute, nil, func(opts *metav1.ListOptions) {
+		opts.FieldSelector = fields.OneTermEqualSelector(metav1.ObjectNameField, wfc.configController.GetName()).String() // only the controller configmap
+	})
+
+	if !watchControllerSemaphoreConfigMaps {
```
juliev0, line 1319: I'm confused about this. Is it supposed to be here, or was it copy/pasted from below?

agilgur5: It is supposed to be there. If the flag is set, no event handlers are added to semaphore or controller informers.

They're together only because the flag was implemented prior to them being split (it was implemented as a workaround for the main bug this PR aims to fix)

juliev0: Oh jeez. Backward compatibility can be so annoying sometimes. :) I suppose some people do intend to not watch their Controller ConfigMap and are using that environment variable? In that case, can we make it more clear in the code through comments and/or naming that it represents both things?

juliev0:
Unfortunately, it looks like there are some file conflicts now.

Another question: which of the different configmaps have been effectively tested (either manually, unit testing, or e2e)?

workflow/controller/controller.go
agilgur5: > I suppose some people do intend to not watch their Controller ConfigMap and are using that environment variable?

Yes, https://github.com/argoproj/argo-helm/issues/2622 is one example, although I'm not committed to supporting that. I wanted to leave this flag in just-in-case there are remaining issues. The hot reload has also been buggy before, e.g. #11177, #11542, #11541, so it is potentially useful as a workaround if that reoccurs.

> In that case, can we make it more clear in the code through comments and possibly naming that it represents both things?

Sure I can add a comment. I made sure it was described in the docs in #12622 and #12400, but code comments are missing.

The naming follows the env variable exactly as the codebase conventionally does, so I wouldn't change that though.

juliev0: okay, sounds good

agilgur5: [added](https://github.com/argoproj/argo-workflows/pull/11855/commits/05ee735a985473e4f15b1fc642f6ff96571aa7c5) comment

Also fixed merge conflicts (it was with my own PR that stemmed from a finding in this PR: #12917. just had to choose this PR's version for each conflict)

juliev0:
I'm going through old PRs now, looking for stale ones to add "stale" label to. What's the status of this one @agilgur5 @juranir ?

agilgur5:
It's still very necessary, although the upstream issue may have been resolved in k8s 1.30 if our hypotheses are correct, but even if that is accurate it may break again as it was due to a breaking bugfix in etcd that is now feature flagged due to the unexpected breakage. This change is still good to have and fixes other issues either way as well.

But I never got around to writing the tests, and, as far as I could tell, there were no tests for this previously, which would explain why it wasn't caught during k8s upgrades.

Joibel:
Replaced by #16408

PR #11865, opened 2023-09-22
docs: Add K8s compatibility for v3.5 release

terrytangyuan:
This is based on https://endoflife.date/kubernetes. K8s v1.25's "Maintenance Support" will end next month but I think it's safe to keep it as fully supported since we have CI builds that test against it.


agilgur5 (approved):

agilgur5:
Good to have now that there are 2 RCs out. Agreed that we can keep v1.25 since it's tested

juliev0:
So, if our CI tests with v1.25, then how do we know we're compatible with >v1.25? testing outside of the CI?

terrytangyuan:
We are testing both v1.25 and latest. If we don’t specify any versions, it will use the latest: https://github.com/argoproj/argo-workflows/blob/799df790e14f20461f5a3831c7df52913425cff8/.github/workflows/ci-build.yaml#L73


If we explicitly set some builds to test the old version. https://github.com/argoproj/argo-workflows/blob/799df790e14f20461f5a3831c7df52913425cff8/.github/workflows/ci-build.yaml#L96

juliev0:
> We are testing both v1.25 and latest.
> 
> If we don’t specify any versions, it will use the latest:
> 
> https://github.com/argoproj/argo-workflows/blob/799df790e14f20461f5a3831c7df52913425cff8/.github/workflows/ci-build.yaml#L73
> 
> We also explicitly set some builds to test the old version.
> 
> https://github.com/argoproj/argo-workflows/blob/799df790e14f20461f5a3831c7df52913425cff8/.github/workflows/ci-build.yaml#L96

thanks for walking through that. And k3s version uses the same versioning as k8s?

terrytangyuan:
Yes but with suffix "-k3s1".

juliev0 (approved):

PR #11882, opened 2023-09-24
refactor(ui): use async/await in several components

agilgur5:
Partial fix for #11740 

### Motivation

Use newer, modern async/await syntax instead of Promise chains

### Modifications

- convert promise chains -> async/await syntax
  - this is largely cleaner, newer syntax, and easier to reason about
    - for instance, there were a few `.then(set)` in a row which should now be [batched](https://react.dev/learn/queueing-a-series-of-state-updates#react-batches-state-updates) together
      - they _may_ have been batched together before too, but not sure, as the chain could have resulted in callbacks being scheduled later, causing unnecessary re-renders

- also convert `services.info.collectEvent` -> `useCollectEvent` for consistency
  - and slightly optimize this by removing the unnecessary `.then()`
    - we don't do anything on `.then()` and the request is fired async with no await, so it's not needed
      - was basically just a scheduled void callback -- no need for schedule, no need for callback
   - all components now use `useCollectEvent`, except for `WorkflowsList`, which I am covering in a separate PR as the whole component needs a large refactor 

- also convert consts assigned to anonymous functions to named functions instead
  - which is better for tracing and debugging

### Verification

Largely semantically equivalent, outside of optimizations based above; mainly syntactic changes

### DIFF
--- a/ui/src/app/cluster-workflow-templates/components/cluster-workflow-template-details/cluster-workflow-template-details.tsx
+++ b/ui/src/app/cluster-workflow-templates/components/cluster-workflow-template-details/cluster-workflow-template-details.tsx
@@ -7,6 +7,7 @@ import {ClusterWorkflowTemplate} from '../../../../models';
 import {uiUrl} from '../../../shared/base';
 import {ErrorNotice} from '../../../shared/components/error-notice';
 import {Loading} from '../../../shared/components/loading';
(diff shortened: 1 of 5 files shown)


toyamagu-2021 (approved):
Thanks!

terrytangyuan (approved):

PR #11894, opened 2023-09-27
refactor(ui): convert ParametersInput to functional components

juijeong8324:
Partial fix for #9810 

### Motivation

Use newer, modern functional components compatible with hooks instead of legacy class-based components

### Modifications

- `this.props` -> arguments
- `this.state` -> useState
- methods -> inner functions

### Verification

- Pure refactor, no real semantic changes
- Tested in Local environment 
     - Screenshot below of `submit-workflows-panel`
     image

### DIFF
--- a/ui/src/app/shared/components/parameters-input/parameters-input.tsx
+++ b/ui/src/app/shared/components/parameters-input/parameters-input.tsx
@@ -51,9 +51,9 @@ export function ParametersInput(props: ParametersInputProps) {
         );
     }
 
-    function onParametersChange(parameters: Parameter[]) {
+    function onParametersChange(params: Parameter[]) {
         if (props.onChange) {
-            props.onChange(parameters);
+            props.onChange(params);
         }
     }
 
--- a/ui/src/app/shared/components/parameters-input/parameters-input.tsx
+++ b/ui/src/app/shared/components/parameters-input/parameters-input.tsx
@@ -1,15 +1,22 @@
 import {Select, Tooltip} from 'argo-ui';
-import React, {useState} from 'react';
+import React from 'react';
 import {Parameter} from '../../../../models';
 import {Utils} from '../../utils';
 
 interface ParametersInputProps {
     parameters: Parameter[];
-    onChange?: (parameters: Parameter[]) => void;
+    onChange: (parameters: Parameter[]) => void;
 }
 
 export function ParametersInput(props: ParametersInputProps) {
-    const [parameters, setParameters] = useState<Parameter[]>(props.parameters || []);
+    function onParameterChange(parameter: Parameter, value: string) {
+        const newParameters: Parameter[] = props.parameters.map(p => ({
+            name: p.name,
+            value: p.name === parameter.name ? value : Utils.getValueFromParameter(p),
+            enum: p.enum
+        }));
+        props.onChange(newParameters);
+    }
 
     function displaySelectFieldForEnumValues(parameter: Parameter) {
         return (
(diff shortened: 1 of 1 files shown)


ui/src/app/shared/components/parameters-input/parameters-input.tsx
```diff
@@ -42,39 +21,55 @@ export class ParametersInput extends React.Component<ParametersInputProps, {para
...
     }
 
-    private onParametersChange(parameters: Parameter[]) {
-        if (this.props.onChange) {
-            this.props.onChange(parameters);
+    function onParametersChange(params: Parameter[]) {
+        if (props.onChange) {
+            props.onChange(params);
         }
     }
```
agilgur5, line 58:
```suggestion
    function onParameterChange(parameter: Parameter, value: string) {
        const newParameters: Parameter[] = props.parameters.map(p => ({
            name: p.name,
            value: p.name === parameter.name ? value : Utils.getValueFromParameter(p),
            enum: p.enum
        }));
        props.onChange(newParameters);
    }
```

we can consolidate the change logic into this function. I also made it singular as this is a function that runs when a single parameter changes. (the previous name confused me a bit until I rewrote it)

agilgur5: Could you also move this function to the top of the file? Then it would have consistent ordering with other components in the codebase where they have `state`, then callbacks, then render functions

agilgur5 (commented):
The changes look good, but would you be willing to make some further optimizations to this component? These aren't in code you wrote, they were already there, but I figure if we're refactoring it, we might as well improve it too!

See below comments

juijeong8324:
> The changes look good, but would you be willing to make some further optimizations to this component? These aren't in code you wrote, they were already there in the previous code, but I figure if we're refactoring it, we might as well improve it too!
> 
> See below comments

Of course! It's my first time thinking about optimization. 🤔Thank you😎

ui/src/app/shared/components/parameters-input/parameters-input.tsx
juijeong8324: > we can consolidate the change logic into this function. I also changed the name to be singular as this is a function that runs when a _single_ parameter changes. (the previous name confused me a bit until I rewrote it)

I realize that it is important to change with good functions name!

juijeong8324:
I learn a lot about how to refactor through this feedback!!
 
Actually... I saw unnecessary code (e.g. distinguishing between props.parameter and state parameter..) when I read the code at first.  But I was wondering about whether it would be okay to change it, so I just changed it to functional component.. 

I will confidently refactor to make some further optimizations!! Thanks @agilgur5

ui/src/app/shared/components/parameters-input/parameters-input.tsx
agilgur5: > Ah! It means that `onParamterChange` functions is callbacks, and other display functions is render function right?

Yep! They're pretty common patterns in React components

agilgur5:
> Actually... I saw unnecessary code (e.g. distinguishing between props.parameter and state parameter..) when I read the code at first. But I was wondering about whether it would be okay to change it, so I just changed it to functional component..

Yea that definitely makes sense to make the safer change when you're not sure 👍 🙂 
Your thought process was on the mark though, there was some unnecessary shadowing due to the previously optional `onChange` prop. If `onChange` were _actually_ optional (it was [incorrectly marked](https://github.com/argoproj/argo-workflows/pull/11894#discussion_r1340776363)), then there would need to be some state. Since it isn't optional though, `parameter` is held in state by its parent component and this component itself doesn't have any state as a result -- it's a "pure" component

agilgur5 (approved):
Thanks for updating and optimizing this! It looks a lot cleaner now too! 😄

terrytangyuan (approved):
Thanks!

PR #11895, opened 2023-09-27
docs: Remove CI status badge from README.md

terrytangyuan:
It currently shows "failing" but is actually passing. I've observed the false positive many times so let's remove it.


agilgur5 (approved):
I believe CLOMonitor doesn't have [a check](https://clomonitor.io/docs/topics/checks/) for build status, so I think it's safe to remove without affecting that score.

Also I agree that it doesn't provide much value to readers (and is affected by flakes etc too).

PR #11899, opened 2023-09-28
refactor(ui): InputFilter and WorkflowTimeline components from class to functional

tetora1053:
Partial fix for https://github.com/argoproj/argo-workflows/issues/9810

> * [ ]  15. ui/src/app/shared/components/input-filter.tsx:
> * [ ]   44. ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx:

### Motivation

### Modifications

- `this.state` -> `useState`
- `this.props` -> `props`
- `useRef` to manage references to DOM elements 
- methods -> inner functions

### Verification

- manually tested in local environment

InputFilter
Screen Shot 2023-09-29 at 1 37 06
Screen Shot 2023-09-29 at 1 37 20

WorkflowTimeline
Screen Shot 2023-09-29 at 1 41 20

## Reference PRs
- https://github.com/argoproj/argo-workflows/pull/11800
- https://github.com/argoproj/argo-workflows/pull/11803

### DIFF
--- a/ui/src/app/shared/components/input-filter.tsx
+++ b/ui/src/app/shared/components/input-filter.tsx
@@ -13,8 +13,8 @@ export function InputFilter(props: InputProps) {
     const [localCache, setLocalCache] = useState((localStorage.getItem(props.name + '_inputs') || '').split(',').filter(item => item !== ''));
 
     function setValueAndCache(newValue: string) {
-        setLocalCache(state => {
-            const updatedCache = [...state];
+        setLocalCache(currentCache => {
+            const updatedCache = [...currentCache];
             if (!updatedCache.includes(newValue)) {
                 updatedCache.unshift(newValue);
             }
--- a/ui/src/app/shared/components/input-filter.tsx
+++ b/ui/src/app/shared/components/input-filter.tsx
@@ -1,5 +1,5 @@
 import {Autocomplete} from 'argo-ui';
-import * as React from 'react';
+import React, {useState} from 'react';
 
 interface InputProps {
     value: string;
@@ -8,68 +8,60 @@ interface InputProps {
     onChange: (input: string) => void;
 }
 
-interface InputState {
-    value: string;
-    localCache: string[];
-    error?: Error;
-}
+export function InputFilter(props: InputProps) {
+    const [value, setValue] = useState<string>(props.value);
+    const [localCache, setLocalCache] = useState<string[]>((localStorage.getItem(props.name + '_inputs') || '').split(',').filter(item => item !== ''));
 
-export class InputFilter extends React.Component<InputProps, InputState> {
-    constructor(props: Readonly<InputProps>) {
-        super(props);
-        this.state = {
-            value: props.value,
-            localCache: (localStorage.getItem(this.props.name + '_inputs') || '').split(',').filter(value => value !== '')
-        };
+    function setValueAndCache(newValue: string) {
+        setLocalCache(state => {
+            const updatedCache = [...state];
+            if (!updatedCache.includes(newValue)) {
+                updatedCache.unshift(newValue);
+            }
+            while (updatedCache.length > 5) {
+                updatedCache.pop();
+            }
+            localStorage.setItem(props.name + '_inputs', updatedCache.join(','));
+            return updatedCache;
+        });
     }
 
-    public render() {
+    function renderInput(inputProps: React.HTMLProps<HTMLInputElement>) {
         return (
-            <>
-                <Autocomplete
-                    items={this.state.localCache}
-                    value={this.state.value}
-                    onChange={(e, value) => this.setState({value})}
-                    onSelect={value => {
-                        this.setState({value});
-                        this.props.onChange(value);
-                    }}
-                    renderInput={inputProps => (
-                        <input
-                            {...inputProps}
-                            onKeyUp={event => {
-                                if (event.keyCode === 13) {
-                                    this.setValueAndCache(event.currentTarget.value);
-                                    this.props.onChange(this.state.value);
-                                }
-                            }}
-                            className='argo-field'
-                            placeholder={this.props.placeholder}
-                        />
-                    )}
-                />
-                <a
-                    onClick={() => {
-                        this.setState({value: ''});
-                        this.props.onChange('');
-                    }}>
-                    <i className='fa fa-times-circle' />
-                </a>
-            </>
+            <input
+                {...inputProps}
+                onKeyUp={event => {
+                    if (event.keyCode === 13) {
+                        setValue(event.currentTarget.value);
+                        setValueAndCache(event.currentTarget.value);
+                        props.onChange(value);
+                    }
+                }}
+                className='argo-field'
+                placeholder={props.placeholder}
+            />
         );
     }
 
-    private setValueAndCache(value: string) {
-        this.setState(state => {
-            const localCache = state.localCache;
-            if (!state.localCache.includes(value)) {
-                localCache.unshift(value);
-            }
-            while (localCache.length > 5) {
-                localCache.pop();
-            }
-            localStorage.setItem(this.props.name + '_inputs', localCache.join(','));
-            return {value, localCache};
-        });
-    }
+    return (
+        <>
+            <Autocomplete
+                items={localCache}
+                value={value}
+                onChange={(e, newValue) => setValue(newValue)}
+                onSelect={newValue => {
+                    setValue(newValue);
+                    props.onChange(newValue);
+                }}
+                renderInput={renderInput}
+            />
+            <a
+                onClick={() => {
+                    setValue('');
+                    props.onChange('');
+                }}>
+                <i className='fa fa-times-circle' />
+            </a>
+        </>
+    );
 }
--- a/ui/src/app/shared/components/input-filter.tsx
+++ b/ui/src/app/shared/components/input-filter.tsx
@@ -9,8 +9,8 @@ interface InputProps {
 }
 
 export function InputFilter(props: InputProps) {
-    const [value, setValue] = useState<string>(props.value);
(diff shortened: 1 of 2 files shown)


ui/src/app/shared/components/input-filter.tsx
```diff
@@ -8,68 +8,60 @@ interface InputProps {
     onChange: (input: string) => void;
 }
 
-interface InputState {
-    value: string;
-    localCache: string[];
-    error?: Error;
```
tetora1053, line 14: `error` state isn't used, so removed it.

ui/src/app/shared/components/input-filter.tsx
```diff
@@ -8,68 +8,60 @@ interface InputProps {
...
-}
+export function InputFilter(props: InputProps) {
+    const [value, setValue] = useState(props.value);
+    const [localCache, setLocalCache] = useState((localStorage.getItem(props.name + '_inputs') || '').split(',').filter(item => item !== ''));
```
tetora1053, line 13: ts infers the type of `localCache` correctly, so I omitted type specification. (`useState<string[]>`)
but at the same time, I think it also looks good to specify the type for readability.
which should I choose 😟

agilgur5: yea [same as your previous PR](https://github.com/argoproj/argo-workflows/pull/11803#discussion_r1323485883), I think when the inference is correct and straightforward (`string` in this case), there's no need to annotate it.

I also have experience working in ML-family languages like OCaml, which are even stronger typed and where the compiler infers the types of the entire program (so there are _no_ type annotations at all). On the other spectrum would be something like Java, which requires on type annotations on almost everything and feels very verbose as a result (despite having weaker guarantees than OCaml et al).

For TS, I prefer to have strict typing on, which we'll be able to do in this codebase once there is some progress on decoupling `argo-ui` (see https://github.com/argoproj/argo-ui/issues/453). With strict typing, if a type can't be correctly inferred, the compiler will tell you (right now I believe it only has a warning).

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
...
-    public componentWillUnmount() {
-        if (this.resizeSubscription) {
-            this.resizeSubscription.unsubscribe();
-            this.resizeSubscription = null;
-        }
-        if (this.refreshSubscription) {
-            this.refreshSubscription.unsubscribe();
-            this.refreshSubscription = null;
+    useEffect(() => {
+        ensureRunningWorkflowRefreshing(props.workflow);
```
agilgur5, line 45: we can in-line this function now as it's only used here now

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
     nodeClicked?: (node: models.NodeStatus) => any;
 }
 
-interface WorkflowTimelineState {
-    parentWidth: number;
-    now: moment.Moment;
-}
+export function WorkflowTimeline(props: WorkflowTimelineProps) {
+    let resizeSubscription: Subscription;
+    let refreshSubscription: Subscription;
```
agilgur5, line 23:
```suggestion
    const resizeSubscription = useRef<Subscription>(null);
    const refreshSubscription = useRef<Subscription>(null);
```

these actually should be refs as well so that they are persisted between renders.
Right now I think it only still works by coincidence, since all the functions have it in their scope, but the best practice (and least error prone) here would be to use refs

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
...
-    private refreshSubscription: Subscription;
+    const [parentWidth, setParentWidth] = useState(0);
+    const [now, setNow] = useState(moment());
 
-    constructor(props: WorkflowTimelineProps) {
-        super(props);
-        this.state = {parentWidth: 0, now: moment()};
-        this.ensureRunningWorkflowRefreshing(props.workflow);
-    }
+    const containerRef = useRef(null);
```
agilgur5, line 28:
```suggestion
    const containerRef = useRef<HTMLElement>(null);
```

this one _does_ need a type annotation 😉 
(since `null` can't be inferred to be an `HTMLElement`)

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
agilgur5: Could you also move these to be together with `containerRef` as they were before? so that all refs are together

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
...
 
-    public componentDidMount() {
-        this.resizeSubscription = fromEvent(window, 'resize').subscribe(() => this.updateWidth());
-        this.updateWidth();
-    }
+    useEffect(() => {
+        resizeSubscription = fromEvent(window, 'resize').subscribe(updateWidth);
```
agilgur5, line 31:
```suggestion
        resizeSubscription.current = fromEvent(window, 'resize').subscribe(updateWidth);
```

will need a `.current` when it's converted to a ref

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
...
-    public componentWillReceiveProps(nextProps: WorkflowTimelineProps) {
-        this.ensureRunningWorkflowRefreshing(nextProps.workflow);
-    }
+        return () => {
+            if (resizeSubscription) {
+                resizeSubscription.unsubscribe();
+            }
+            if (refreshSubscription) {
+                refreshSubscription.unsubscribe();
+            }
```
agilgur5, line 40:
```suggestion
        return () => {
            resizeSubscription.current?.unsubscribe();
            refreshSubscription.current?.unsubscribe();
```

can simplify these with optional chaining syntax

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
...
-            .sort((first, second) => {
-                const diff = moment(first.startedAt).diff(second.startedAt);
-                // If node started almost at the same time then sort by end time
-                if (diff <= 2) {
-                    return moment(first.finishedAt).diff(second.finishedAt);
-                }
-                return diff;
+    function ensureRunningWorkflowRefreshing(workflow: models.Workflow) {
+        const completedPhases = [models.NODE_PHASE.ERROR, models.NODE_PHASE.SUCCEEDED, models.NODE_PHASE.SKIPPED, models.NODE_PHASE.OMITTED, models.NODE_PHASE.FAILED];
+        const isCompleted = workflow && workflow.status && completedPhases.indexOf(workflow.status.phase) > -1;
```
agilgur5, line 56:
```suggestion
        const isCompleted = workflow && workflow.status && completedPhases.includes(workflow.status.phase);
```

can use the newer `includes` function now

ui/src/app/workflows/components/workflow-timeline/workflow-timeline.tsx
```diff
@@ -18,128 +18,124 @@ interface WorkflowTimelineProps {
...
+        if (!refreshSubscription && !isCompleted) {
+            refreshSubscription = interval(1000).subscribe(() => {
+                setNow(moment());
             });
-        if (nodes.length === 0) {
-            return <div />;
+        } else if (refreshSubscription && isCompleted) {
+            refreshSubscription.unsubscribe();
+            refreshSubscription = null;
```
agilgur5, line 63:
```suggestion
        if (!refreshSubscription.current && !isCompleted) {
            refreshSubscription.current = interval(1000).subscribe(() => {
                setNow(moment());
            });
        } else if (refreshSubscription.current && isCompleted) {
            refreshSubscription.current.unsubscribe();
            refreshSubscription.current = null;
```

will need to add `.current` here when it's a ref as well

ui/src/app/shared/components/input-filter.tsx
```diff
@@ -8,68 +8,60 @@ interface InputProps {
...
-export class InputFilter extends React.Component<InputProps, InputState> {
-    constructor(props: Readonly<InputProps>) {
-        super(props);
-        this.state = {
-            value: props.value,
-            localCache: (localStorage.getItem(this.props.name + '_inputs') || '').split(',').filter(value => value !== '')
-        };
+    function setValueAndCache(newValue: string) {
+        setLocalCache(currentCache => {
+            const updatedCache = [...currentCache];
```
agilgur5, line 17: nice catch to make this an immutable operation!

agilgur5 (commented):
just a few changes need to refs in the `WorkflowTimeline` component, otherwise the rest looks good 😄

ui/src/app/shared/components/input-filter.tsx
tetora1053: I see. unless it's a complex type, I'll try to omit the type annotations where TS correctly infers. thanks for detailed description.😄 

and I had no idea about argo-ui! Thank you for letting me know. I'm really hopeful for the progress in decoupling. It sounds it could bring great improvements.

agilgur5 (approved):
LGTM. Thanks for updating this and fixing some subtle bugs (like the [mutable updates](https://github.com/argoproj/argo-workflows/pull/11899#discussion_r1340796935)) in the process too!

terrytangyuan (approved):
Thanks!

PR #11901, opened 2023-09-29
refactor(ui): converting a drop-down component to a functional component

yunwoo-yu:
Partial fix for #9810 

### Motivation

Convert legacy class components to use the modern syntax of functional components and hooks

### Modifications

- Remove this statement and change to a React hook
- Change methods to regular functions
- Remove React lifecycle and change to useEffect

### Notes for Reviewers

- [Removing scroll events](https://github.com/argoproj/argo-workflows/pull/11901#pullrequestreview-1651714439)

### Verification

In my local environment, I checked the drop-down filters on the Event Flow page to see the change.

### DIFF
--- a/ui/src/app/shared/components/dropdown/dropdown.tsx
+++ b/ui/src/app/shared/components/dropdown/dropdown.tsx
@@ -4,15 +4,9 @@ import {createPortal} from 'react-dom';
 
 export interface DropDownProps {
     isMenu?: boolean;
-    anchor: () => JSX.Element;
+    anchor: JSX.Element;
     children: ReactNode;
     qeId?: string;
-}
-
-export interface DropDownState {
-    opened: boolean;
-    left: number;
-    top: number;
 }
 
 require('./dropdown.scss');
@@ -49,7 +43,7 @@ export function DropDown({isMenu, anchor, children, qeId}: DropDownProps) {
     }
 
     function open() {
-        if (!contentRef || !anchorRef) {
+        if (!contentEl || !anchorEl) {
             return;
         }
 
@@ -62,21 +56,37 @@ export function DropDown({isMenu, anchor, children, qeId}: DropDownProps) {
     }
 
     function close(event: MouseEvent) {
-        if (opened) {
-            // Doesn't close when clicked inside the portal area
-            if (contentEl.contains(event.target as Node) || anchorEl.contains(event.target as Node)) {
-                return;
-            }
+        // Doesn't close when clicked inside the portal area
+        if (contentEl.contains(event.target as Node) || anchorEl.contains(event.target as Node)) {
+            return;
+        }
 
-            setOpened(false);
+        setOpened(false);
+    }
+
+    function scroll() {
+        if (contentEl && anchorEl) {
+            const newState = refreshState();
+
+            setOpened(newState.opened);
+            setTop(newState.top);
+            setLeft(newState.left);
         }
     }
 
     useEffect(() => {
-        document.body.addEventListener('click', close);
+        if (!opened) {
+            return;
+        }
 
-        return () => document.body.removeEventListener('click', close);
-    });
+        document.addEventListener('click', close);
+        document.addEventListener('scroll', scroll, {capture: true});
+
+        return () => {
+            document.removeEventListener('click', close);
+            document.removeEventListener('scroll', scroll, {capture: true});
+        };
+    }, [opened]);
 
     return (
         <div className='argo-dropdown' ref={anchorRef}>
@@ -87,7 +97,7 @@ export function DropDown({isMenu, anchor, children, qeId}: DropDownProps) {
                     open();
                     event.stopPropagation();
                 }}>
-                {anchor()}
+                {anchor}
             </div>
             {createPortal(
                 <div className={classNames('argo-dropdown__content', {opened, 'is-menu': isMenu})} style={{top, left}} ref={contentRef}>
--- a/ui/src/app/shared/components/dropdown/dropdown.tsx
+++ b/ui/src/app/shared/components/dropdown/dropdown.tsx
@@ -1,13 +1,10 @@
 import classNames from 'classnames';
-import React, {ComponentType, ReactNode, useEffect, useRef} from 'react';
-import {useState} from 'react';
+import React, {ReactNode, useEffect, useRef, useState} from 'react';
 import {createPortal} from 'react-dom';
-import {BehaviorSubject, fromEvent, merge, Subscription} from 'rxjs';
-import {filter} from 'rxjs/operators';
 
 export interface DropDownProps {
     isMenu?: boolean;
-    anchor: ComponentType;
+    anchor: () => JSX.Element;
     children: ReactNode;
     qeId?: string;
 }
@@ -20,74 +17,39 @@ export interface DropDownState {
 
 require('./dropdown.scss');
 
-const dropDownOpened = new BehaviorSubject<typeof DropDown>(null);
-
-export function DropDown({isMenu, anchor: Anchnor, children, qeId}: DropDownProps) {
-    const anchorRef = useRef<null | HTMLDivElement>(null);
-    const contentRef = useRef<null | HTMLDivElement>(null);
-    const subscriptionsRef = useRef<null | Subscription[]>(null);
+export function DropDown({isMenu, anchor, children, qeId}: DropDownProps) {
     const [opened, setOpened] = useState(false);
     const [left, setLeft] = useState(0);
     const [top, setTop] = useState(0);
+    const anchorRef = useRef<HTMLDivElement>(null);
+    const contentRef = useRef<HTMLDivElement>(null);
+    const anchorEl = anchorRef.current;
+    const contentEl = contentRef.current;
 
-    useEffect(() => {
-        const content = contentRef.current;
-        const el = anchorRef.current;
-
-        subscriptionsRef.current = [
-            merge(
-                dropDownOpened.pipe(filter(dropdown => dropdown !== DropDown)),
-                fromEvent(document, 'click').pipe(
-                    filter((event: Event) => {
-                        return content && opened && !content.contains(event.target as Node) && !el.contains(event.target as Node);
-                    })
-                )
-            ).subscribe(() => {
-                close();
-            }),
-            fromEvent(document, 'scroll', {capture: true}).subscribe(() => {
-                if (opened && content && el) {
-                    const newState = refreshState();
-
-                    setOpened(newState.opened);
-                    setLeft(newState.left);
-                    setTop(newState.top);
-                }
-            })
-        ];
-
-        return () => {
-            (subscriptionsRef.current || []).forEach(s => s.unsubscribe());
-            subscriptionsRef.current = [];
-        };
-    }, []);
-
-    const refreshState = () => {
-        const content = contentRef.current;
-        const anchor = anchorRef.current;
-        const anchorHeight = anchor.offsetHeight + 2;
-        const {top: anchorTop, left: anchorLeft} = anchor.getBoundingClientRect();
-        const newState = {top: anchorTop, left: anchorLeft, opened};
+    function refreshState() {
+        const anchorHeight = anchorEl.offsetHeight + 2;
+        const {top: anchorTop, left: anchorLeft} = anchorEl.getBoundingClientRect();
+        const newState = {top, left, opened};
 
         // Set top position
-        if (content.offsetHeight + anchorTop + anchorHeight > window.innerHeight) {
-            newState.top = anchorTop - content.offsetHeight - 2;
+        if (contentEl.offsetHeight + anchorTop + anchorHeight > window.innerHeight) {
+            newState.top = anchorTop - contentEl.offsetHeight - 2;
         } else {
             newState.top = anchorTop + anchorHeight;
         }
 
         // Set left position
-        if (content.offsetWidth + anchorLeft > window.innerWidth) {
-            newState.left = anchorLeft - content.offsetWidth + anchor.offsetWidth;
+        if (contentEl.offsetWidth + anchorLeft > window.innerWidth) {
+            newState.left = anchorLeft - contentEl.offsetWidth + anchorEl.offsetWidth;
         } else {
             newState.left = anchorLeft;
         }
 
         return newState;
-    };
+    }
 
-    const open = () => {
-        if (!contentRef.current || !anchorRef.current) {
+    function open() {
+        if (!contentRef || !anchorRef) {
             return;
         }
 
@@ -97,12 +59,24 @@ export function DropDown({isMenu, anchor: Anchnor, children, qeId}: DropDownProp
         setOpened(newState.opened);
         setLeft(newState.left);
         setTop(newState.top);
-        dropDownOpened.next(DropDown);
-    };
+    }
 
-    const close = () => {
-        setOpened(false);
-    };
+    function close(event: MouseEvent) {
+        if (opened) {
+            // Doesn't close when clicked inside the portal area
+            if (contentEl.contains(event.target as Node) || anchorEl.contains(event.target as Node)) {
+                return;
+            }
+
+            setOpened(false);
+        }
+    }
+
+    useEffect(() => {
+        document.body.addEventListener('click', close);
+
+        return () => document.body.removeEventListener('click', close);
+    });
 
     return (
         <div className='argo-dropdown' ref={anchorRef}>
@@ -113,10 +87,10 @@ export function DropDown({isMenu, anchor: Anchnor, children, qeId}: DropDownProp
                     open();
                     event.stopPropagation();
                 }}>
-                <Anchnor />
+                {anchor()}
             </div>
             {createPortal(
-                <div className={classNames('argo-dropdown__content', {'opened': opened, 'is-menu': isMenu})} style={{top, left}} ref={contentRef}>
+                <div className={classNames('argo-dropdown__content', {opened, 'is-menu': isMenu})} style={{top, left}} ref={contentRef}>
                     {children}
                 </div>,
                 document.body
--- a/ui/src/app/shared/components/dropdown/dropdown.tsx
+++ b/ui/src/app/shared/components/dropdown/dropdown.tsx
@@ -64,27 +64,15 @@ export function DropDown({isMenu, anchor, children, qeId}: DropDownProps) {
         setOpened(false);
     }
 
-    function scroll() {
-        if (contentEl && anchorEl) {
-            const newState = refreshState();
-
-            setOpened(newState.opened);
-            setTop(newState.top);
-            setLeft(newState.left);
-        }
-    }
-
     useEffect(() => {
         if (!opened) {
             return;
         }
 
         document.addEventListener('click', close);
-        document.addEventListener('scroll', scroll, {capture: true});
 
         return () => {
             document.removeEventListener('click', close);
-            document.removeEventListener('scroll', scroll, {capture: true});
         };
     }, [opened]);
 
--- a/ui/src/app/shared/components/dropdown/dropdown.tsx
+++ b/ui/src/app/shared/components/dropdown/dropdown.tsx
@@ -1,13 +1,14 @@
 import classNames from 'classnames';
-import * as React from 'react';
-import * as ReactDOM from 'react-dom';
+import React, {ComponentType, ReactNode, useEffect, useRef} from 'react';
+import {useState} from 'react';
+import {createPortal} from 'react-dom';
 import {BehaviorSubject, fromEvent, merge, Subscription} from 'rxjs';
 import {filter} from 'rxjs/operators';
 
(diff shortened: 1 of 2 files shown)


yunwoo-yu:
Hi @agilgur5  😄

The changes are done, but there is actually one problem. I don't know rxjs 😅...

I tried to import the existing logic as it is, but it doesn't work properly when I close that dropdown filter on the `event-flow` page 🥲.

Can anyone help me with this? 🙇‍♂️

(Also, I found an existing bug where clicking the check all button works fine but gives an error - I'll put up a separate PR for this when I get a chance!)

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -1,13 +1,14 @@
 import classNames from 'classnames';
-import * as React from 'react';
-import * as ReactDOM from 'react-dom';
+import React, {ComponentType, ReactNode, useEffect, useRef} from 'react';
+import {useState} from 'react';
```
agilgur5, line 3:
```suggestion
import React, {ComponentType, ReactNode, useEffect, useRef, useState} from 'react';
```
consolidate the two import statements together

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
-                    </div>,
-                    document.body
-                )}
-            </div>
-        );
-    }
+export function DropDown({isMenu, anchor: Anchnor, children, qeId}: DropDownProps) {
+    const anchorRef = useRef<null | HTMLDivElement>(null);
+    const contentRef = useRef<null | HTMLDivElement>(null);
+    const subscriptionsRef = useRef<null | Subscription[]>(null);
```
agilgur5, line 28:
```suggestion
    const anchorRef = useRef<HTMLDivElement>(null);
    const contentRef = useRef<HTMLDivElement>(null);
    const subscriptionsRef = useRef<Subscription[]>(null);
```

you don't need to have an `| null` as [_all_ types are nullable](https://www.typescriptlang.org/docs/handbook/advanced-types.html#nullable-types) in JS.

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
                 fromEvent(document, 'click').pipe(
                     filter((event: Event) => {
-                        return this.content && this.state.opened && !this.content.contains(event.target as Node) && !this.el.contains(event.target as Node);
+                        return content && opened && !content.contains(event.target as Node) && !el.contains(event.target as Node);
                     })
                 )
             ).subscribe(() => {
-                this.close();
+                close();
             }),
```
agilgur5, line 47:
```suggestion
            ).subscribe(close),
```

can shorten this

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
             })
         ];
-    }
 
-    public componentWillUnmount() {
-        (this.subscriptions || []).forEach(s => s.unsubscribe());
-        this.subscriptions = [];
-    }
+        return () => {
+            (subscriptionsRef.current || []).forEach(s => s.unsubscribe());
```
agilgur5, line 60:
```suggestion
            subscriptionsRef.current?.forEach(s => s.unsubscribe());
```

optional chaining simplifies this a tiny bit

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
-    }
+        setOpened(newState.opened);
+        setLeft(newState.left);
+        setTop(newState.top);
+        dropDownOpened.next(DropDown);
+    };
+
+    const close = () => {
+        setOpened(false);
+    };
```
agilgur5, line 105:
```suggestion
    function close() {
        setOpened(false);
    }
```

can you use named functions for all these helper funcs? they're better for debugging than anonymous functions assigned to a const (and a tiny bit more optimized).

Also could you put these above the effect? somewhat conventionally in this codebase, it's state, refs, callbacks, then effects as the ordering inside each component

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
+    }, []);
 
-    public close() {
-        this.setState({opened: false});
-    }
-
-    private refreshState() {
-        const anchor = this.el.querySelector('.argo-dropdown__anchor') as HTMLElement;
-        const {top, left} = anchor.getBoundingClientRect();
+    const refreshState = () => {
```
agilgur5, line 65:
```suggestion
    function refreshState() {
```

named function same as above

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
+        if (content.offsetWidth + anchorLeft > window.innerWidth) {
+            newState.left = anchorLeft - content.offsetWidth + anchor.offsetWidth;
         } else {
-            newState.left = left;
+            newState.left = anchorLeft;
         }
+
         return newState;
-    }
+    };
```
agilgur5, line 87:
```suggestion
    }
```

the semi-colon will be removed by `prettier` for a named function

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
 
-    private open() {
-        if (!this.content || !this.el) {
+    const open = () => {
```
agilgur5, line 89:
```suggestion
    function open() {
```

named function same as above

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
+
         newState.opened = true;
-        this.setState(newState);
-        dropDownOpened.next(this);
-    }
+        setOpened(newState.opened);
+        setLeft(newState.left);
+        setTop(newState.top);
+        dropDownOpened.next(DropDown);
+    };
```
agilgur5, line 101:
```suggestion
    }
```

semi-colon will be removed by `prettier` for a named function, same as above

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
+    const subscriptionsRef = useRef<null | Subscription[]>(null);
+    const [opened, setOpened] = useState(false);
+    const [left, setLeft] = useState(0);
+    const [top, setTop] = useState(0);
 
-    public componentWillMount() {
-        this.subscriptions = [
+    useEffect(() => {
+        const content = contentRef.current;
+        const el = anchorRef.current;
```
agilgur5, line 35: you create temporary variables for these refs' current values in a few places -- may make sense to just move these variables to the component level and use them in all the functions

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -1,13 +1,14 @@
...
+import React, {ComponentType, ReactNode, useEffect, useRef} from 'react';
+import {useState} from 'react';
+import {createPortal} from 'react-dom';
 import {BehaviorSubject, fromEvent, merge, Subscription} from 'rxjs';
 import {filter} from 'rxjs/operators';
 
 export interface DropDownProps {
     isMenu?: boolean;
-    anchor: React.ComponentType;
-    children: React.ReactNode | (() => React.ReactNode);
```
agilgur5, line 10: ah I see that the function form is never actually this component's [sole consumer](https://github.com/argoproj/argo-workflows/blob/e91c2737f3dff1fee41ce97991e294a57c53fc93/ui/src/app/shared/components/filter-drop-down.tsx#L18) -- good simplification!

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
-                        style={{top: this.state.top, left: this.state.left}}
-                        ref={el => (this.content = el)}>
-                        {children}
-                    </div>,
-                    document.body
-                )}
-            </div>
-        );
-    }
+export function DropDown({isMenu, anchor: Anchnor, children, qeId}: DropDownProps) {
```
agilgur5, line 25:
```suggestion
export function DropDown({isMenu, anchor, children, qeId}: DropDownProps) {
```

the extra type is not needed since it has an interface already. `Anchnor` also has a typo

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
+    return (
+        <div className='argo-dropdown' ref={anchorRef}>
+            <div
+                qe-id={qeId}
+                className='argo-dropdown__anchor'
+                onClick={event => {
+                    open();
+                    event.stopPropagation();
+                }}>
+                <Anchnor />
```
agilgur5, line 116:
```suggestion
                <anchor />
```

if that doesn't work I think you can use the same syntax as `children`:

```suggestion
                {anchor}
```

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
+            </div>
+            {createPortal(
+                <div className={classNames('argo-dropdown__content', {'opened': opened, 'is-menu': isMenu})} style={{top, left}} ref={contentRef}>
```
agilgur5, line 119:
```suggestion
                <div className={classNames('argo-dropdown__content', {opened, 'is-menu': isMenu})} style={{top, left}} ref={contentRef}>
```
can use shorthand here since they have the same name now

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +20,107 @@ export interface DropDownState {
...
-    public componentWillMount() {
-        this.subscriptions = [
+    useEffect(() => {
+        const content = contentRef.current;
+        const el = anchorRef.current;
+
+        subscriptionsRef.current = [
             merge(
-                dropDownOpened.pipe(filter(dropdown => dropdown !== this)),
+                dropDownOpened.pipe(filter(dropdown => dropdown !== DropDown)),
```
agilgur5, line 39: hmm this function and `.next` might be the problem. did using `this` not work? functions still have `this`

I think it might make sense to just remove the `filter` entirely by moving `dropDownOpened` into the component itself as a ref. The `filter` is only needed because there could be multiple instantiations of this component and `dropDownOpened` is currently at module-level. If we move `dropDownOpened`  into component-level, then each instantiation will have their own ref -- i.e. `dropDownOpened` becomes component-local.

I think then the generic can be removed as well and it could just accept no arguments, i.e. `dropDownOpened.next()`.

Also Rx is not really needed for these simple event listeners, so we could even go a step further and remove Rx from this component entirely.

Making `dropDownOpened` a component-level ref is the easier change, so I'd probably suggest you do that unless you want to rewrite more 😅

agilgur5 (commented):
Thanks for tackling this! There are some simplifications and stylistic changes I requested below.

I also _think_ I found the Rx issue below, see comment in-line.
I don't know if I've ever used the Events Flow page, but if you found a bug, definitely feel free to fix it!

ui/src/app/shared/components/dropdown/dropdown.tsx
yunwoo-yu:
```ts
export interface DropDownProps {
    isMenu?: boolean;
    anchor: () => JSX.Element;
    children: ReactNode;
    qeId?: string;
}
```
```suggestion
                {anchor()}
```
When I tested, they all returned errors, but when I changed the 'anchor' type of 'DropDownProps' to () => JSX.Element and added {anchor()}, it worked 👍 Is it ok to proceed like this?

ui/src/app/shared/components/dropdown/dropdown.tsx
yunwoo-yu: Thanks for the great way to reduce repetitive code, I'll add it for reference at the component level!
```
const anchorEl = anchorRef.current;
const contentEl = contentRef.current;
```

ui/src/app/shared/components/dropdown/dropdown.tsx
yunwoo-yu: Yes this doesn't work I tested and found that the filter function this in the class component returns the value of the DropDown object
image
but in the functional component it returns a function and this returns undefined. 🤔
image
If I could pass this function as an object to `dropDownOpened`, I could save the existing code, but I can't think of a way to do it easily 😅 I moved `dropDownOpened` inside the component as you said, but `close()` still doesn't work 🥲 For now, I'm trying to figure out if there's a way to pass this function component as an object to `dropDownOpened.pipe`!

agilgur5: > but in the functional component it returns a function and this returns undefined. 🤔

Yea the former is what I expected. You were passing the function itself, whereas the class was passing an _instantiation_.

I thought the latter might work since all functions are also objects in JS, but I appear to have forgotten some of the [detailed nuances there](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this). That is a hacky method though, removing the filter entirely with a component-local ref would be better anyway.

Looks like you managed to figure out how to remove Rx entirely though in your recent commit (nice work!), so seems like this is moot now?

ui/src/app/shared/components/dropdown/dropdown.tsx
agilgur5: Yea as far as I can tell, the parent's usage of `anchor` is indeed a straight element.

Since you changed this already, could you go a step further and just make it a plain `JSX.Element` with no function? Would have to change the parent as well. It would be a good simplification for both to remove an unnecessary function!

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +17,84 @@ export interface DropDownState {
...
+            newState.left = anchorLeft;
         }
+
         return newState;
     }
 
-    private open() {
-        if (!this.content || !this.el) {
+    function open() {
+        if (!contentRef || !anchorRef) {
```
agilgur5, line 52:
```suggestion
        if (!contentEl || !anchorEl) {
```

you might've had a typo here since you had the elements in the previous version. the refs themselves will always return `true`

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +17,84 @@ export interface DropDownState {
...
+
+    function close(event: MouseEvent) {
+        if (opened) {
+            // Doesn't close when clicked inside the portal area
+            if (contentEl.contains(event.target as Node) || anchorEl.contains(event.target as Node)) {
+                return;
+            }
+
+            setOpened(false);
+        }
```
agilgur5, line 72:
```suggestion
        if (!opened) {
            return
        }
        
        // don't close when clicked inside the portal area
        if (contentEl.contains(event.target as Node) || anchorEl.contains(event.target as Node)) {
            return;
        }

        setOpened(false);
```

we can invert the conditional to simplify this a bit and reduce nesting

agilgur5 (commented):
Nice work removing Rx -- it substantially simplifies the code! 🙂 
Is this working now without Rx? Wasn't sure because your previous comment I think was before you removed it.

Two small comments below and one above as well

ui/src/app/shared/components/dropdown/dropdown.tsx
yunwoo-yu: - Thanks for the great idea! I'll remove the function and post it right away 👍

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +17,84 @@ export interface DropDownState {
...
     }
+
+    useEffect(() => {
+        document.body.addEventListener('click', close);
+
+        return () => document.body.removeEventListener('click', close);
+    });
```
agilgur5, line 79:
```suggestion
    useEffect(() => {
        if (!opened} {
          return
        }
        
        document.body.addEventListener('click', close);
        return () => document.body.removeEventListener('click', close);
    }, [opened]);
```


We can make some small optimizations here I think. So that the event listener is only attached when opened and the effect is only called when `opened` changes values

ui/src/app/shared/components/dropdown/dropdown.tsx
agilgur5: actually, with my further comments, I think we can remove the `!opened` check from here as it would be in the effect instead.

agilgur5 (commented):
there might some small misses with the effects

ui/src/app/shared/components/dropdown/dropdown.tsx
```diff
@@ -19,112 +17,84 @@ export interface DropDownState {
...
-                dropDownOpened.pipe(filter(dropdown => dropdown !== this)),
-                fromEvent(document, 'click').pipe(
-                    filter((event: Event) => {
-                        return this.content && this.state.opened && !this.content.contains(event.target as Node) && !this.el.contains(event.target as Node);
-                    })
-                )
-            ).subscribe(() => {
-                this.close();
-            }),
-            fromEvent(document, 'scroll', {capture: true}).subscribe(() => {
```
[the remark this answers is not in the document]
yunwoo-yu: image
- This scroll event simply moves this dropdown position to a new location when the scroll event occurs. However, this doesn't seem to be working correctly. 
- This image is a test of the scroll event on the master branch The filter moves up when the screen is small and works fine when the screen is large, so I'm not sure what the need for the event is 😅 Let me know what you think and I'll try to incorporate it, thanks 🙇‍♂️

ui/src/app/shared/components/dropdown/dropdown.tsx
yunwoo-yu: Yes when I tested it, both close and open work fine! Even when I clicked inside the filter list, there is no event propagation and everything looks fine 👍 I'll re-commit with the details you mentioned, thanks for your help as always 🙇‍♂️

ui/src/app/shared/components/dropdown/dropdown.tsx
yunwoo-yu: Oh I think I missed it, thank you 👍

yunwoo-yu (commented):
Oct-01-2023 04-22-21
I've incorporated all of your feedback into the code 👍
I'll leave a gif to show the dropdown working properly with Rxjs removed, thanks!

ui/src/app/shared/components/dropdown/dropdown.tsx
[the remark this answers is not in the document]
agilgur5: Ok, I didn't realize that scroll code was broken, thanks for bringing that up!
I tested myself, and yes it didn't seem very effective or useful. Maybe this logic predates the CSS being `fixed` or something? 🤷 

So I think your initial removal was the correct move forward, let's remove it again. Sorry for the re-work!

For future reference, when you change something that may be unexpected to a reviewer, a common practice is to either mention it in your PR description -- I use a "Notes for Reviewers" section occasionally -- or comment on your own PR and leave an in-line comment explaining why it was changed. You can kind of predict a question like this and answer it ahead of time then, making the review more efficient 🙂
It also saves that information for posterity in the comments instead of just in your brain or a reviewer's brain. So that, for example, if a future reader wonders why this was removed, they can just look at this comment and know -- no need to guess!

agilgur5 (commented):
I've been sick the past week, so just getting back to more complex reviews now.

I have just [one comment](https://github.com/argoproj/argo-workflows/pull/11901#discussion_r1357719273) to follow-up on the scroll event -- let's go with the removal you had earlier.

Otherwise everything LGTM, great work and iteration on this!

ui/src/app/shared/components/dropdown/dropdown.tsx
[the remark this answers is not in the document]
yunwoo-yu: - I wasn't aware of that... Am I correct in saying that I should either get a link to the comment with the explanation of why the scroll event was removed and create a separate comment, or create a section in the description part of the PR "Notes for Reviewers" and write it down?
- Thanks for always teaching me so much. Come to think of it, I don't think it would be easy to find this comment without a separate link 🤣.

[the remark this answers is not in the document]
agilgur5: I think adding a "Notes for Reviewers" with a link to this comment thread would be sufficiently helpful for posterity 🙂

agilgur5 (approved):
🚀

terrytangyuan (approved):

Issue #7701, opened 2022-01-31
argo lint panics on a certain invalid workflow


Ark-kun:
## Summary

What happened/what you expected to happen?
`argo lint workflow.yaml` panics.

This is not a blocker for me. Just stumbled upon it and decided to report.

What version of Argo Workflows are you running?
`v3.3.0-rc2`

## Diagnostics

Either a workflow that reproduces the bug, or paste you whole workflow YAML, including status, something like:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
spec:
  entrypoint: root
  templates:
    - name: template
      inputs:
        parameters:
          - name: data
      container:
        name: main
        image: alpine
    - name: root
      inputs:
        parameters:
          - name: anything_param
      dag:
        tasks:
          - name: task
            template: template
            arguments:
              parameters:
                - name: data
                  # There is a bug here. I still keep forgetting which syntax is right >_<
                  valueFrom:
                    parameter: "{{inputs.parameters.anything_param}}"
  arguments:
    parameters:
      - name: anything_param
        value: anything_param
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x19c6f3b]

goroutine 1 [running]:
github.com/argoproj/argo-workflows/v3/workflow/validate.validateDAGTaskArgumentDependency({{0xc000189080, 0x1, 0x4}, {0x0, 0x0, 0x0}}, {0x478a408, 0x0, 0x23ec660})
        /home/runner/work/argo-workflows/argo-workflows/workflow/validate/validate.go:1332 +0x3bb
github.com/argoproj/argo-workflows/v3/workflow/validate.(*templateValidationCtx).validateDAG(0xc00012bb40, 0xc00042dd40, 0x2526839, 0xc000180900)
        /home/runner/work/argo-workflows/argo-workflows/workflow/validate/validate.go:1311 +0xce5
github.com/argoproj/argo-workflows/v3/workflow/validate.(*templateValidationCtx).validateTemplate(0xc00012bb40, 0xc000180240, 0xc00012e9c0, {0x32dd130, 0xc00042dbf0})
        /home/runner/work/argo-workflows/argo-workflows/workflow/validate/validate.go:399 +0xf3c
[...]

PR #11902, opened 2023-09-29
fix: Allow valueFrom in dag arguments parameters. Fixes #11900

### LINKED ISSUE #11900: Combination valueFrom and withParam does not work for dag templates
### Pre-requisites

```
arguments:
  parameters:
  - name: param
    valueFrom:
      configMapKeyRef:
        name: cm-{{item.configMap}}
        key: param
withParam: "{{inputs.parameters.cm-list}}"      # parameter specifies the list to iterate over
```

throws the error in the sensor deployment:
```
time="2023-09-29T08:45:39.915Z" level=fatal msg="Failed to submit workflow: templates.main.tasks.print missing value for parameter 'param'"{"level":"error","ts":1695977139.9195628,"logger":"argo-events.sensor","caller":"sensors/listener.go:345","msg":"Failed to execute a trigger","sensorName":"test-resource","error":"failed to execute trigger, failed after retries: failed to execute submit command for workflow , exit status 1","triggerName":"workflow-trigger","triggeredBy":["webhook"],"triggeredByEvents":["62303161646637302d393732362d343865322d383730632d636636383962313466626533"],"stacktrace":"github.com/argoproj/argo-events/sensors.(*SensorContext).triggerWithRateLimit\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:345"}
```

when using a `dag` template but works with `steps` template.

Below are the entire sensors and workflows.

### Version

v3.4.11

```YAML
Throws an error:

kind: Sensor
metadata:
  name: test-resource
  namespace: argo
spec:
  dependencies:
    - name: webhook
      eventSourceName: webhook
      eventName: test
  revisionHistoryLimit: 1
  template:
    serviceAccountName: argo-events-sa
  triggers:
  - template:
      name: workflow-trigger
      argoWorkflow:
        group: argoproj.io
        version: v1alpha1
        resource: workflows
        operation: submit
        source:
          resource:
            kind: Workflow
            metadata:
              generateName: test-resource-
              namespace: argo
            spec:
              entrypoint: main
              serviceAccountName: workflow
              arguments:
                parameters:
                - name: cm-list                                     # a list of items
                  value: |
                    [
                      { "configMap": "param-test-ping" },
                      { "configMap": "param-test-pong" }
                    ]

              templates:
              - name: main
                inputs:
                  parameters:
                  - name: cm-list
                dag:
                  tasks:
                  - name: print
                    templateRef:
                      name: print-params
                      template: main
                    arguments:
                      parameters:
                      - name: param
                        valueFrom:
                          configMapKeyRef:
                            name: cm-{{item.configMap}}
                            key: param
                    withParam: "{{inputs.parameters.cm-list}}"      # parameter specifies the list to iterate over
```

Works as expected:
```
kind: Sensor
metadata:
  name: test-resource
  namespace: argo
spec:
  dependencies:
    - name: webhook
      eventSourceName: webhook
      eventName: test
  revisionHistoryLimit: 1
  template:
    serviceAccountName: argo-events-sa
  triggers:
  - template:
      name: workflow-trigger
      argoWorkflow:
        group: argoproj.io
        version: v1alpha1
        resource: workflows
        operation: submit
        source:
          resource:
            kind: Workflow
            metadata:
              generateName: test-resource-
              namespace: argo
            spec:
              entrypoint: main
              serviceAccountName: workflow
              arguments:
                parameters:
                - name: cm-list                                     # a list of items
                  value: |
                    [
                      { "configMap": "param-test-ping" },
                      { "configMap": "param-test-pong" }
                    ]

              templates:
              - name: main
                inputs:
                  parameters:
                  - name: cm-list
                steps:
                - - name: print
                    templateRef:
                      name: print-params
                      template: main
                    arguments:
                      parameters:
                      - name: param
                        valueFrom:
                          configMapKeyRef:
                            name: cm-{{item.configMap}}
                            key: param
                    withParam: "{{inputs.parameters.cm-list}}"      # parameter specifies the list to iterate over
```

Workflow template:
```
kind: WorkflowTemplate
metadata:
  name: print-params
spec:
  entrypoint: main
  templates:
  - name: main
    inputs:
      parameters:
      - name: param
    container:
      image: cgr.dev/chainguard/bash:latest
      command:
      - /bin/bash
      - -c
      - |
        set -eu
        echo "{{inputs.parameters.param}}"
```
```


```text
```

```text
```

nice-pink:
Fixes #11900 

### Motivation

Allow assigning values from config maps to dag.tasks.arguments.parameters.

### Modifications

The validation for dag task arguments was updated to allow `valueFrom`, not only `value`.


workflow/validate/validate_dag_test.go
```diff
@@ -1039,42 +1039,3 @@ func TestDAGOutputsResolveTaskAggregatedOutputs(t *testing.T) {
 	err := validate(dagOutputsResolveTaskAggregatedOutputs)
 	assert.NoError(t, err)
 }
-
-var dagMissingParamValueInTask = `
```
terrytangyuan, line 1043: Maybe update the test instead of removing?

nice-pink: I reverted my changes as I don't know what thoughts were behind forbidding ValueFrom.Parameter, so I added an additional case for ValueFrom.ConfigMapKeyRef. I also reverted the test and added a new one for the second case.

juliev0:
So sorry, I'm going through old PRs and I see that unfortunately this one has been forgotten. I will try to review it soon.

Argo Workflows needs more people approving PRs unfortunately. If your company is by chance interested, I am gauging interest in this idea: https://github.com/argoproj/argo-workflows/issues/12229

workflow/validate/validate.go
```diff
@@ -1424,9 +1424,11 @@ func validateDAGTaskArgumentDependency(arguments wfv1.Arguments, ancestry []stri
 
 	for _, param := range arguments.Parameters {
 		if param.Value == nil {
-			return errors.Errorf(errors.CodeBadRequest, "missing value for parameter '%s'", param.Name)
-		}
-		if strings.HasPrefix(param.Value.String(), "{{tasks.") {
+			// Disallow ValueFrom.Parameter, but allow ValueFrom.ConfigMapKeyRef
+			if param.ValueFrom == nil || param.ValueFrom.Parameter != "" {
+				return errors.Errorf(errors.CodeBadRequest, "missing value for parameter '%s'", param.Name)
+			}
```
juliev0, line 1430: Nice fix. I see that `ValueFrom` has a lot of sub-parameter options. How did you decide which to allow and which not to?

nice-pink: I just wanted to concentrate on the description in the ticket and allow as little as possible. Further, I wanted to align the behaviour of steps and dag for this very example.

Could be that other options might be valid here as well, though. That should be proposed and discussed in a follow up ticket.

nice-pink: I just wanted to concentrate on the description in the ticket and allow as little as possible. Further, I wanted to align the behaviour of steps and dag for this very example.

Could be that other options might be valid here as well, though. That should be proposed and discussed in a follow up ticket.

juliev0: allow as little as possible? But I think it will allow all of the other members to be set, right? (just `ValueFrom.Parameter` is prevented)

juliev0:
Looks good now. I saw in the original issue it says "throws the error when submitting in dag template but works for steps." Out of curiosity, is there any validation for ValueFrom members in steps?

workflow/validate/validate.go
nice-pink: You are completely right.

nice-pink:
No, there is no test for steps. Should I add one for consistency? The only test related is for Inputs.Parameters (`paramWithValueFromConfigMapRef`) in `validate_tetst.go`.

juliev0:
> No, there is no test for steps. Should I add one for consistency? The only test related is for Inputs.Parameters (`paramWithValueFromConfigMapRef`) in `validate_tetst.go`.

Thanks for offering that. In general, I'd say consistency is really important (plus consolidating the code into a single method). In this case, I'm concerned that if we apply your change, people will start reporting that certain ValueFrom members aren't working for steps, since before everything was allowed. One thought is: what if we took your previous iteration of code that only denied `ValueFrom.Parameter` (assuming you're fairly certain that that won't be evaluated correctly in either case - maybe it could be tested if not 100% sure?) and create a common function for that somewhere? Let me know what you think.

nice-pink:
I re-read the whole code related and checked why the `if` was added [Issue 7710](https://github.com/argoproj/argo-workflows/issues/7701). I think I found a better solution.

There is already the `validateArguments` method used for `steps` and `dag`. The method I updated was `validateDAGTaskArgumentDependency` which really should only validate if the dependencies exist before referencing them. I understand why the fix 7710 was added, but it makes all values defined by `valueFrom` invalid, even though only `valueFrom.Parameter` was intended to be considered invalid. Further, `valueFrom.Parameter` was only considered invalid for `dag`, even though it is invalid for `steps` also.

So I propse the current updates, so that `validateDAGTaskArgumentDependency` is used for dependency validation only. Further, I updated `validateArgumentsValues` to throw an error, if the values are not valid. 

To me it seems like only `ConfigMapKeyRef`, `Default` and likely `Expression` are valid options here.

Let me know what you think. If you also believe this makes sense, I'll add some more tests.

juliev0:
Oh darn, I'm not sure if you saw, but it seems that your last commit accidentally altered a bunch of other files as well. 

As for your suggestion, it sounds great. I appreciate the digging you did. 

You mentioned: "To me it seems like only ConfigMapKeyRef, Default and likely Expression are valid options here." Would you mind going through each of the other members and commenting here on why each does not make sense? We should make sure that your assumption applies to all 3 places I see `validateArguments()` is called from. In general, I tend to think that validation is a "nice extra", and if we're unsure at all, then maybe less validation is better than more.

Really appreciate the iterations. Thank you!

nice-pink:
Thanks for noticing me about the messy commit!

I agree with less validation could be an option as currently the abilities are cut to prevent one invalid case.

Here are all options of `valueFrom`:

```
// Path in the container to retrieve an output parameter value from in container templates
Path string `json:"path,omitempty" protobuf:"bytes,1,opt,name=path"`
```
Is used for output parameters only.

```
// JSONPath of a resource to retrieve an output parameter value from in resource templates
JSONPath string `json:"jsonPath,omitempty" protobuf:"bytes,2,opt,name=jsonPath"`
```
Is used for output parameters only.

```
// JQFilter expression against the resource object in resource templates
JQFilter string `json:"jqFilter,omitempty" protobuf:"bytes,3,opt,name=jqFilter"`
```
Seems to be used for output parameters only.

```
// Selector (https://github.com/antonmedv/expr) that is evaluated against the event to get the value of the parameter. E.g. `payload.message`
Event string `json:"event,omitempty" protobuf:"bytes,7,opt,name=event"`
```
Actually not sure.

```
// Parameter reference to a step or dag task in which to retrieve an output parameter value from
// (e.g. '{{steps.mystep.outputs.myparam}}')
Parameter string `json:"parameter,omitempty" protobuf:"bytes,4,opt,name=parameter"`
```
Maybe parameter never makes sense, as parameters are referenced using `value: {{...}}`?!

```
// Supplied value to be filled in directly, either through the CLI, API, etc.
Supplied *SuppliedValueFrom `json:"supplied,omitempty" protobuf:"bytes,6,opt,name=supplied"`
```
Should only be used for `inputs.Parameters`.

```
// ConfigMapKeyRef is configmap selector for input parameter configuration
ConfigMapKeyRef *apiv1.ConfigMapKeySelector `json:"configMapKeyRef,omitempty" protobuf:"bytes,9,opt,name=configMapKeyRef"`
```
Valid

```
// Default specifies a value to be used if retrieving the value from the specified source fails
Default *AnyString `json:"default,omitempty" protobuf:"bytes,5,opt,name=default"`
```
Valid, e.g. if config map can't be read.

```
// Expression, if defined, is evaluated to specify the value for the parameter
Expression string `json:"expression,omitempty" protobuf:"bytes,8,rep,name=expression"`
```
Would be nice, to e.g. set predefined strings based on boolean set in `inputs.Parameters`.

juliev0:
Thanks for walking through all of those. For a bunch you mentioned that they're only for Output Parameters. Is `validateArgumentsValues()` only being called for Input Parameters then? I'm not sure i see the code for that.

nice-pink:
Yes, these methods are only used for validating the arguments. The outputs are validated in `validateOutputs` and `validateOutputParameter`, the inputs are validated in `validateInputs`.

juliev0:
> Yes, these methods are only used for validating the arguments. The outputs are validated in `validateOutputs` and `validateOutputParameter`, the inputs are validated in `validateInputs`.

Yes, thanks!

juliev0:
Okay, so how about disallowing any that you're fairly certain should be and allowing the rest? (and maybe adding a "TODO" if not sure about some)

nice-pink:
Sounds good to me. I'll update the code and the test asap.

workflow/validate/validate.go
```diff
@@ -856,11 +856,23 @@ func validateArgumentsFieldNames(prefix string, arguments wfv1.Arguments) error
...
+		// check for valid valueFrom sub-parameters
+		// INFO: default needs to be accompanied by with ConfigMapKeyRef.
+		// TODO: check if ValueFrom.Expression is valid!
+		if param.ValueFrom != nil && param.ValueFrom.ConfigMapKeyRef == nil && param.ValueFrom.Expression == "" {
+			return errors.Errorf(errors.CodeBadRequest, "only default, configMapKeyRef and supplied allowed for valueFrom '%s'", param.Name)
+		}
+		// check for invalid valueFrom sub-parameters
+		if param.ValueFrom != nil && param.ValueFrom.Path != "" && param.ValueFrom.JSONPath != "" && param.ValueFrom.Event != "" && param.ValueFrom.Parameter != "" && param.ValueFrom.Supplied != nil {
+			return errors.Errorf(errors.CodeBadRequest, "only default, configMapKeyRef and supplied allowed for valueFrom '%s'", param.Name)
+		}
```
juliev0, line 874: should those `&&` be `II` by chance? I thought if you simply set `param.ValueFrom.JSONPath` that it would be invalid.?

nice-pink: You are completely right.

workflow/validate/validate.go
```diff
@@ -856,11 +856,23 @@ func validateArgumentsFieldNames(prefix string, arguments wfv1.Arguments) error
...
+		// check for valid valueFrom sub-parameters
+		// INFO: default needs to be accompanied by with ConfigMapKeyRef.
+		// TODO: check if ValueFrom.Expression is valid!
+		if param.ValueFrom != nil && param.ValueFrom.ConfigMapKeyRef == nil && param.ValueFrom.Expression == "" {
+			return errors.Errorf(errors.CodeBadRequest, "only default, configMapKeyRef and supplied allowed for valueFrom '%s'", param.Name)
+		}
+		// check for invalid valueFrom sub-parameters
+		if param.ValueFrom != nil && (param.ValueFrom.Path != "" || param.ValueFrom.JSONPath != "" || param.ValueFrom.Event != "" || param.ValueFrom.Parameter != "" || param.ValueFrom.Supplied != nil) {
+			return errors.Errorf(errors.CodeBadRequest, "only default, configMapKeyRef and supplied allowed for valueFrom '%s'", param.Name)
+		}
```
juliev0, line 874: you mentioned in your notes that you weren't sure about `Event` - are you now?

juliev0: I think if I search for `Event`, it is being used as an argument, no? (in `server/event/dispatch/operation.go`)

juliev0, line 874: This is a total nit, but can you put the whole thing inside one block so it's together for readability?:
```
if param.ValueFrom != nil {

}
```

nice-pink: I haven't found this one. Thanks for checking.

workflow/validate/validate.go
```diff
@@ -856,11 +856,25 @@ func validateArgumentsFieldNames(prefix string, arguments wfv1.Arguments) error
 // validateArgumentsValues ensures that all arguments have parameter values or artifact locations
 func validateArgumentsValues(prefix string, arguments wfv1.Arguments, allowEmptyValues bool) error {
 	for _, param := range arguments.Parameters {
+		// check if any value is defined
 		if param.ValueFrom == nil && param.Value == nil {
 			if !allowEmptyValues {
 				return errors.Errorf(errors.CodeBadRequest, "%s%s.value is required", prefix, param.Name)
 			}
 		}
+		// check for valid valueFrom sub-parameters
```
juliev0, line 865: I think this comment needs to be moved to the first block below?

workflow/validate/validate.go
```diff
@@ -856,11 +856,25 @@ func validateArgumentsFieldNames(prefix string, arguments wfv1.Arguments) error
...
+		if param.ValueFrom != nil {
+			// INFO: default needs to be accompanied by with ConfigMapKeyRef.
+			// TODO: check if ValueFrom.Expression is valid!
+			if param.ValueFrom.ConfigMapKeyRef == nil && param.ValueFrom.Expression == "" {
+				return errors.Errorf(errors.CodeBadRequest, "only default, configMapKeyRef and supplied allowed for valueFrom '%s'", param.Name)
+			}
```
juliev0, line 871: is it allowed to have `Event` and not `ConfigMapKeyRef` or `Expression`?

juliev0: I think from what I'm seeing, `Event` should be allowed but not `Expression`. 

Which IDE do you use? You should be able to do "Find all references" or the equivalent for each member of `ValueFrom` struct. That's where I can see who's calling it and how it's being used. It would be great if you might also do this to double check my findings.

shaharb-hs:
Any ETA on this? 🙏

nice-pink:
Sorry, I was pretty busy lately. I hope to complete the work on this PR this week.

workflow/validate/validate.go
nice-pink: Sorry for the very late response.

You are completely right, that is the current state. Even though expression is currently not used for input/args I already had cases where I would love to use it.

Imagine you have a workflow which has "isDev" bool as input. A sub workflow now needs to set an environment dev/prod/... based on the isDev bool. Currently we need to define the sub workflow twice, set the environment parameter of each explicitly and decide on which sub workflow to trigger based on a `when: "isDev == true"` condition. In these (rare) cases it would be nice to have expression as an option for inputs. It would then be used in arguments.

Anyways, this is not the case currently.

juliev0:
Is this ready for more review?

nice-pink:
Mabe thumbs up was not sufficient as reaction. Yes the PR is ready for review.

juliev0:
> Mabe thumbs up was not sufficient as reaction. Yes, the PR is ready for review.

Okay, will look when I have time

juliev0 (approved):

juliev0 (changes requested):
I approved before checking the CI. Can you check the unit test failure? It may be related.

nice-pink:
I don't see a relation between my changes and the failing e2e tests.

juliev0:
Okay, yes, those sometimes fail intermittently. Sorry for the hassle but can you submit [empty commits](https://www.freecodecamp.org/news/how-to-push-an-empty-commit-with-git/) until everything passes and then let me know?

PR #11907, opened 2023-09-30
fix(ui): use `popup.confirm` instead of browser `confirm`

### LINKED ISSUE #11905: Improve confirmation prompt message
When deleting a selected workflow in the workflow list view, we get a confirmation prompt to check whether to delete the workflow in the database. Currently, the checkbox says "don't allow...to prompt you again" but it should be a "yes". 

image
---

agilgur5:
Fixes #11905 
Partial fix for #11740 

### Motivation

- browser [`confirm`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) is a blocking sync method; replace with the conventional and consistently styled `popup.confirm`
  - `popup.confirm` creates an HTML modal instead
  - this was the last place in the codebase that still used browser `confirm` and not `popup.confirm`
  
- in particular, as #11905 shows, browser `confirm` was causing the browser to allow users to disable the prompt, which would make it _impossible_ to confirm
  - this is because, when there is an archived workflow, two `confirm`s would be used in a row -- one for the batch action and one for the archive delete
    - the browser detects two in a row as possible spam / abuse, and has a checkbox to disable prompts for the entire site

### Modifications

- sync `confirm` -> async `popup.confirm`

- also fix various typing issues in the `WorkflowsToolbar` component while at it
  - there were several unnecessary `any`s that could use an actual type
  - not quite all of them though

- convert some syntax to use async/await as well instead of promise chains
- invert some conditionals to reduce nesting as well

### Verification

Before, using browser API `confirm`:
Screenshot 2023-09-30 at 12 01 48 AM -- confirm browser api

After, using `popup.confirm` modal:
Screenshot 2023-09-29 at 11 45 16 PM -- confirm modal

### DIFF
--- a/ui/src/app/shared/workflow-operations-map.tsx
+++ b/ui/src/app/shared/workflow-operations-map.tsx
@@ -7,7 +7,7 @@ export type OperationDisabled = {
     [action in WorkflowOperationName]: boolean;
 };
 
-type WorkflowOperationName = 'RETRY' | 'RESUBMIT' | 'SUSPEND' | 'RESUME' | 'STOP' | 'TERMINATE' | 'DELETE';
+export type WorkflowOperationName = 'RETRY' | 'RESUBMIT' | 'SUSPEND' | 'RESUME' | 'STOP' | 'TERMINATE' | 'DELETE';
 
 export interface WorkflowOperation {
     title: WorkflowOperationName;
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):
LGTM

PR #11908, opened 2023-09-30
fix(ui): ensure `WorkflowsRow` message is not too long

agilgur5:
### Motivation

`WorkflowsRow` could have styling issues when the message was too long

### Modifications

- cut it off at 180 chars
  - tested with browser devtools on a 4k width screen and 180 was way more than it could display anyway inside 2 columns - so this is an overshoot
- note that per comment, this does _already_ have a CSS `text-overflow`, but when the message was _very_ long, it would still go beyond the width and cause styling issues
  - so doubly cut it off with less than a line of JS
  
- also simplify `onChange` callback
  - no need to wrap in an anonymous func

### Verification

Before, with styling issues:
Screenshot 2023-09-29 at 11 59 55 PM -- overflow issue

After, no styling issues and still properly overflows on a ~4k width screen:
Screenshot 2023-09-29 at 11 40 57 PM -- 4k still overflows fine


terrytangyuan (approved):
LGTM

PR #11909, opened 2023-09-30
fix: add Warning message on Version mismatch between CLI and Server. fixes #9212

### LINKED ISSUE #9212: CLI should warn when CLI version is outdated
# Summary

In  #9209, the user was using a version of the Argo CLI that was 2 minor versions behind the operator running on the cluster. It would be a nice feature if the Argo CLI could throw a warning if the CLI version is older than version running on the cluster and it would be nice if the Argo CLI could tell the user that a new version is available. 

# Use Cases

* Use case 1: As a cluster administrator, I want users of Argo to receive a warning when the version of Argo CLI is older than the version running in the cluster
* Use case 2: As an Argo maintainer, I want users aware they are using older versions of Argo CLI so that they can choose to update the CLI

---

fhochleitner:
add Warning message on Version mismatch between CLI and Server.
implemented a check whether argocli and arger-server version are based on the same git-tag and print a warning when versions are mismatched. 

Fixes #9212

### Motivation

Wanted to start contributing and found an issue that seemed to be a nice start for this journey

### Modifications

added function to compare cli and server version and prints a warning in case versions do not match. function fails with warning messages if it cannot contact server as it should not break everything in case of an error. 

### Verification

built cli and tested it against argo-workflows running on a cluster. happy to implement some e2e tests if required but would need some pointers.

### DIFF
--- a/cmd/argo/commands/root.go
+++ b/cmd/argo/commands/root.go
@@ -1,9 +1,13 @@
 package commands
 
 import (
+	"os"
+
 	"github.com/argoproj/pkg/cli"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
+	infopkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/info"
 
 	"github.com/argoproj/argo-workflows/v3"
 	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands/archive"
@@ -128,10 +132,35 @@ If your server is behind an ingress with a path (you'll be running "argo server 
 		cli.SetLogLevel(logLevel)
 		cmdutil.SetGLogLevel(glogLevel)
 		log.WithField("version", argo.GetVersion()).Debug("CLI version")
+		printVersionMismatchWarning(command)
 	}
 	command.PersistentFlags().StringVar(&logLevel, "loglevel", "info", "Set the logging level. One of: debug|info|warn|error")
 	command.PersistentFlags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enabled verbose logging, i.e. --loglevel debug")
 
 	return command
+}
+
+// printVersionMismatchWarning logs a warning if the CLI version does not match the server version
+func printVersionMismatchWarning(command *cobra.Command) {
+	// if ARGO_SERVER isn't set there's no need to compare server and cli version
+	if _, ok := os.LookupEnv("ARGO_SERVER"); !ok {
+		return
+	}
+	ctx, apiClient := client.NewAPIClient(command.Context())
+	serviceClient, err := apiClient.NewInfoServiceClient()
+	// fail with warning because check for version mismatch should not prevent other commands from running
+
+	if err != nil {
+		log.Warnf("Failed create service client: %v", err)
+		return
+	}
+	serverVersion, err := serviceClient.GetVersion(ctx, &infopkg.GetVersionRequest{})
+	if err != nil {
+		log.Warnf("Failed to connect to Argo Server: %v", err)
+		return
+	}
+	if serverVersion.GitTag != argo.GetVersion().GitTag {
+		log.Warnf("CLI version (%s) does not match server version (%s). This can lead to unexpected behavior.", argo.GetVersion().GitTag, serverVersion.GitTag)
+	}
 }


fhochleitner:
I've placed the version check into the root command so it get's executed and prints the warning no matter command the user actually uses and gets notified in case. but I'm not sure whether that's the right as it does execute a few network calls for all commands. happy to implement changes if you have a way to implement this in a smarter way

cmd/argo/commands/root.go
```diff
@@ -128,10 +132,33 @@ If your server is behind an ingress with a path (you'll be running "argo server
...
 	command.PersistentFlags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enabled verbose logging, i.e. --loglevel debug")
 
 	return command
 }
+
+// printVersionMismatchWarning logs a warning if the CLI version does not match the server version
+func printVersionMismatchWarning(command *cobra.Command) {
+	// if ARGO_SERVER isn't set there's no need to compare server and cli version
+	if _, ok := os.LookupEnv("ARGO_SERVER"); ok {
```
agilgur5, line 147: Let's invert this conditional so that there is less nesting:
```suggestion
	if _, ok := os.LookupEnv("ARGO_SERVER"); !ok {
	    return
	}
	// .. logic below
```

cmd/argo/commands/root.go
```diff
@@ -128,10 +132,33 @@ If your server is behind an ingress with a path (you'll be running "argo server
...
 	command.PersistentFlags().StringVar(&logLevel, "loglevel", "info", "Set the logging level. One of: debug|info|warn|error")
 	command.PersistentFlags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enabled verbose logging, i.e. --loglevel debug")
 
 	return command
 }
+
+// printVersionMismatchWarning logs a warning if the CLI version does not match the server version
+func printVersionMismatchWarning(command *cobra.Command) {
+	// if ARGO_SERVER isn't set there's no need to compare server and cli version
```
agilgur5, line 146:
```suggestion
	// if ARGO_SERVER isn't set there's no need to compare server and cli version
```

I'm not sure that this is necessarily accurate. Any usage of the CLI could have unexpected results if there is a version mismatch. For example, through the k8s API, the CLI could be using an outdated spec

fhochleitner: you're right. a check would definitely be useful in this case as well. do you know whether there's a way to get a version that's comparable with the cli version directly via api access? I haven't been able to find one but then again I am not very familiar with the codebase yet.

cmd/argo/commands/root.go
```diff
@@ -128,10 +132,35 @@ If your server is behind an ingress with a path (you'll be running "argo server
...
+		return
+	}
+	serverVersion, err := serviceClient.GetVersion(ctx, &infopkg.GetVersionRequest{})
+	if err != nil {
+		log.Warnf("Failed to connect to Argo Server: %v", err)
+		return
+	}
+	if serverVersion.GitTag != argo.GetVersion().GitTag {
+		log.Warnf("CLI version (%s) does not match server version (%s). This can lead to unexpected behavior.", argo.GetVersion().GitTag, serverVersion.GitTag)
+	}
```
isubasinghe, line 165: I think it might be a better idea if you separated out the logic that checks for a mismatch from the code that presents this information to the user. 

Only saying this because we could make this a bit more strict then, we could fail on a version mismatch and have a flag to force any action to be performed regardless.

fhochleitner: ah sorry for not having able to continue working on this PR. work and private life got quite busy as of late. 
I'm not sure this will change a lot in the coming months, but I do think that's a great idea @isubasinghe

isubasinghe: No worries, take your time!

driv:
Since this is marked as stale and relatively simple, I can take a look.

cmd/argo/commands/root.go
driv: While I agree with:
> Any usage of the CLI could have unexpected results if there is a version mismatch. 

Even the `version` command does not check the server version if `ARGO_SERVER` is not defined. I'm assuming is not that simple to determine the version in that case.

https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/version.go#L23

agilgur5: My comment was about the "there's no need to compare" comment. That is inaccurate.
You may not be able to retrieve it without `ARGO_SERVER` set, but is nonetheless inaccurate to say "there is no need compare". It would be more apt to say "you can't compare". But the comment is a bit redundant at that point

driv: Great, agree.

agilgur5:
Superseded by #13635

PR #11919, opened 2023-09-30
docs: Improve doc on how to do GitOps with WorkflowTemplates

Sebast1aan:
Adding some guidance on how to maintain WorkflowTemplates in a gitops manner.


docs/workflow-templates.md
```diff
@@ -353,7 +353,9 @@ Using `kubectl apply -f` and `kubectl get wftmpl`
 
 ### GitOps via Argo CD
 
-`WorkflowTemplate` resources can be managed with GitOps by using [Argo CD](https://github.com/argoproj/argo-cd)
+`WorkflowTemplate` resources can be managed with GitOps by using [Argo CD](https://github.com/argoproj/argo-cd).
+Store the `WorkflowTemplates` in a directory in a git repository, which you then monitor as a new application in Argo CD.
+When changes are made to the `WorkflowTemplates`, Argo CD will automatically update their definitions such that they become available in the Argo Workflows UI.
```
agilgur5, line 358:
```suggestion

For example, you can store `WorkflowTemplates` in a `git` repository that is tracked by an `Application` in Argo CD.
Argo CD can then sync any changes to the repository to your cluster.
```

- use [simpler and more direct language](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language) and [active voice](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice), per k8s style guide.
  - "directory" does not need to be specified
  - "new application" is not strictly necessary
  - "automatically update their definitions such that" is very verbose, simplified

- consistently [use UpperCamelCase](https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects) for API objects like `Application`, per k8s style guide
- also use terminology that is consistent with Argo CD's docs, such as "tracked" and "synced"

- the UI is separate and has its own section below, so should not be mentioned here. the UI merely reads from the cluster as well.
- "automatically" actually depends on if you have autosync turned on; we can just remove the word automatically

agilgur5 (changes requested):
I'm not sure that this description is necessary as that can be found in Argo CD's docs and it would be better to leave such an explanation to their docs.

I'll leave it to other contributors if they think this makes sense to add.
I also requested some specific changes in-line [below](https://github.com/argoproj/argo-workflows/pull/11919#discussion_r1342240337) if this is to be added.

Issue #11879, opened 2023-09-24
Artifacts not set in Status.Nodes | ArtifactGC prevents Workflow deletion


Garett-MacGowan:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

This issue stems from #10840. I expect to be able to delete a workflow and have the garbage collection remove the artifact from the repository.

In this specific case, the workflow finalizer isn't removed from the workflow due to a failure [here](https://github.com/argoproj/argo-workflows/blob/d0588ea092012c36ce06f44b008dc057bdf72154/workflow/controller/artifact_gc.go#L135), implying that the artifacts aren't being written to Status.Nodes. Upon inspecting the workflow manifest's `Status.Nodes`, I can confirm that no output artifacts are listed. Artifacts are successfully being written to GCS upon triggering "stop" for the workflow.

**Steps to Reproduce**
1) Have a workflow with a long-running process
2) Stop the workflow mid-execution
3) Artifacts are written to GCS
4) Attempt to delete the workflow
5) Notice that the workflow cannot be deleted since the finalizer can't be removed & the artifacts are not deleted.

**Partial Remedy**
While preparing this ticket, I notice that by adding at least one successful workflow and artifact to my DAG (in parallel), the workflow's finalizer is successfully deleted & the successful DAG step's artifacts are also successfully cleaned up, **however**, the failed (stopped) DAG step's artifacts are not cleaned up. This can lead to dangling resources in GCS.

**Suggested Fix**
Ensure that stopped/failed workflow step's output artifacts are written to `Status.Nodes`

### Version

3.5.0-rc1

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
# https://argoproj.github.io/argo-workflows/cron-workflows/
# https://argoproj.github.io/argo-workflows/fields/
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  name: TEMP
spec:
  # Start up 5 minutes to 6 AM to ensure system is online.
  schedule: "55 5 * * *"
  timezone: America/New_York
  concurrencyPolicy: "Replace"
  startingDeadlineSeconds: 0
[...]

Issue #10768, opened 2023-03-28
kit not downloading


tico24:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I submitted a Documentation enhancement PR #10767

The build repeatedly fails, not on the docs ci but on the main CI.  Specifically it loops over and over again with 

```
 + curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ jq -r .tag_name
+ tag=
+ [  =  ]
+ jq -r .tag_name
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ tag=
+ [  =  ]
+ jq -r .tag_name
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ tag=
+ [  =  ]
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ jq -r .tag_name
+ tag=
+ [  =  ]
+ jq -r .tag_name
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ tag=
+ [  =  ]
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ jq -r .tag_name
+ tag=
+ [  =  ]
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ jq -r .tag_name
+ tag=
+ [  =  ]
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ jq -r .tag_name
+ tag=
+ [  =  ]
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ jq -r .tag_name
+ tag=
+ [  =  ]
+ jq -r .tag_name
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ tag=
+ [  =  ]
+ jq -r .tag_name
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ tag=
+ [  =  ]
+ jq -r .tag_name
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
+ tag=
+ [  =  ]
+ curl -fsL https://api.github.com/repos/kitproj/kit/releases/latest
[...]

Issue #11423, opened 2023-07-23
Why can't use a devcontainer-based development environment using codespace?


Yaminyam:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

[running-locally](https://github.com/argoproj/argo-workflows/blob/master/docs/running-locally.md)
If you look at the documentation, it says that you can't work in codespace based on devcontainer, but it doesn't explain why.

As a result of running it myself, it seems that the installation fails in the middle due to insufficient capacity.
<img width="698" alt="image" src="https://github.com/argoproj/argo-workflows/assets/31057849/ad5cfe0a-bd2c-4fbf-996e-17144a9ed9c6">

So, as described in the document https://containers.dev/implementors/json_reference/#min-host-reqs, I think it can be solved by setting the storage of hostrequirements as needed.

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
.
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

agilgur5:
Feel free to send a PR for any changes if you can get Codespaces working!

As far as I know (I haven't tried myself), Codespaces have had various issues running Argo like this one. Had this pop up yesterday [on Slack](https://cloud-native.slack.com/archives/C020XM04CUW/p1690204870404329?thread_ts=1690204870.404329&cid=C020XM04CUW) too.

Joibel:
The `kit pre-up` line (https://github.com/argoproj/argo-workflows/blob/7310e9c41a03cb128bc644cb3b734d89a8b0436e/.devcontainer/pre-build.sh#L33) is the cause. Kit doesn't seem to work properly as part of pre-build. If you comment that line out on a fork and run up the codespace, and then run kit once it is working it seems to work, but that's as far as I got with codespaces.

agilgur5:
Got Codespaces working in #11981!

PR #11928, opened 2023-10-02
build: replace `kit` with `task`

agilgur5:
### Motivation

- [Task](https://github.com/go-task/task) is a very popular (~9k stars), well-maintained (has [releases](https://github.com/go-task/task/releases) at least once a month in the past 6 months), [well-documented](https://taskfile.dev/api/), Go-based task runner with a nearly identical spec to `kit`, as can be seen in the diff
  - basically can do virtually all of the same things and a good number of other things, with lots of docs and active maintenance

- Comparatively, [`kit`](https://github.com/kitproj/kit) was only created in the past year by Alex, is not well-known (~40 stars, and anecdotally, I'd never heard of it before), and seems to be sporadically maintained
  - It also has some unfixed bugs like https://github.com/argoproj/argo-workflows/issues/11879#issuecomment-1740263062 and has had other bugs like #10768 etc
    - Discussed some of these [over](https://cloud-native.slack.com/archives/C0510EUH90V/p1696023977780199) [Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1696264262161079) as well
    
To be explicit, I'm not big on Task or anything ([Earthly](https://github.com/earthly/earthly) is my current go-to, as mentioned [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1692215848943679) etc) -- it was just the most similar and most popular alternative I knew of

### Modifications

- intentionally made this PR have as few changes as possible to reduce the diff and make it look very familiar
  - used one-line arrays instead of multi-line to be similar to `kit`'s `make`-like ["stringy types"](https://github.com/kitproj/kit/blob/main/docs/STRINGY_TYPES.md)
  - used singular `cmd` syntax instead of the usual plural `cmds` syntax

- then followed Task's [styleguide](https://taskfile.dev/styleguide/):
  - name as `Taskfile.yaml`: https://taskfile.dev/styleguide/#use-taskfileyml-and-not-taskfileyml
  - new lines between tasks: https://taskfile.dev/styleguide/#add-spaces-between-tasks
  - consistent ordering of keywords: https://taskfile.dev/styleguide/#use-the-correct-order-of-keywords
    - strangely it doesn't mention keywords _within_ tasks, but I did `cmd` -> `dir` -> `deps` -> `sources` -> `env` consistently
      - similar to how I ordered the GH Actions keywords too in #11752 

- replace all installation and usage of `kit` with `task`
  - following the installation guide: https://taskfile.dev/installation/
    - use `npm` for the `Makefile` install as it's compatible on different OSes and matches some of our existing tools (like `mdspell`, `markdownlint`, etc)
      - NOTE: this does still need adding to the Nix dir
  - replace `kit` in `Makefile`, `pre-build.sh`, and `ci-build.yaml` with `task`      

### Verification

This is a draft, I actually haven't tested it yet


Taskfile.yml
```diff
@@ -0,0 +1,109 @@
...
+
+# The app will be up-and-running between 15s and 1m later (if hot compiled or cold).
+# When using `--watch`, any changes made to the source code will be automatically recompiled and the app restarted, typically within a few seconds.
+
+version: 3
+
+tasks:
+  go-deps:
+    cmd: go mod download
+    # mutex: download
```
agilgur5, line 18: I don't think any of these mutexes are actually strictly necessary, two downloads and two builds can run in parallel, they just may bottleneck, so it might be a concurrency optimization to run them serially instead.

Task does have the [`run`](https://taskfile.dev/usage/#limiting-when-tasks-run) key to ensure something runs only once. And it does have a [serial syntax](https://taskfile.dev/usage/#calling-another-task), but it's not as flexible as a string for a mutex lock.

Taskfile.yml
```diff
@@ -0,0 +1,109 @@
...
+  build-controller:
+    cmd: make ./dist/workflow-controller
+    deps: [go-deps]
+    sources: [cmd/workflow-controller, config, errors, persist, pkg, util, workflow]
+    # mutex: build
+
+  port-forward:
+    cmd: ./hack/port-forward.sh
+    deps: [install]
+    # ports: 9000
```
agilgur5, line 37: I believe the ports should already be opened, but I might be wrong

Taskfile.yml
```diff
@@ -0,0 +1,109 @@
+# yaml-language-server: $schema: https://taskfile.dev/schema.json
```
agilgur5, line 1: the schema is a pretty nice addition too 😉

Taskfile.yml
agilgur5: mm this is not forwarding all the way from my devcontainer to my Mac's localhost, so there may very well be something missing here

agilgur5:
So far `task` also seems much more efficient than `kit`. It's running pretty silently on my laptop in the background whereas previously `make start UI=true` would make all my fans turn on quite loudly.

Screenshot from Docker Desktop:
Screenshot 2023-10-02 at 4 21 21 PM

agilgur5:
All of the E2E tests seem to be [failing](https://github.com/argoproj/argo-workflows/actions/runs/6384463230/job/17328006179?pr=11928) on [port issues](https://github.com/argoproj/argo-workflows/pull/11928#discussion_r1343065238). CI doesn't run in a `devcontainer` though, so those should be a bit different

Makefile
```diff
@@ -517,18 +517,13 @@ dist/argosay:
...
-	brew tap kitproj/kit --custom-remote https://github.com/kitproj/kit
-	brew install kit
-else
-	curl -q https://raw.githubusercontent.com/kitproj/kit/main/install.sh | tag=v0.1.8 sh
-endif
+.PHONY: task
+task:
+# update this in Nix when upgrading it here
+ifneq ($(USE_NIX), true)
+	npm i -g @go-task/cli@3.30.1
```
isubasinghe, line 524: Needs updating in the Nix file, happy to do that myself if you want?
Technically the Nix setup doesn't need task though because it already has something similar.

isubasinghe:
Generally looks good to me, not approving or disapproving yet, will wait till tests pass.

terrytangyuan (changes requested):
Per our discussions at today's contributors meeting, one better option is to revert back to use goreman which is widely adopted and used by Argo CD.

isubasinghe:
> Per our discussions at today's contributors meeting, one better option is to revert back to use goreman which is widely adopted and used by Argo CD. 

@terrytangyuan 
This might be the easiest way to get rid of kit, a careful revert should do the trick.


There will almost certainly be conflicts.

Makefile
agilgur5: @isubasinghe could you elaborate on this? You mentioned `devenv` before, but I didn't know that had task runner support (again, my Nix knowledge is pretty dated).
We discussed it at the Contributor meeting too if it could be used as a replacement on all environments, but no one knew the answer so we were looking to you for more info

agilgur5:
> Per our discussions at today's contributors meeting, one better option is to revert back to use goreman which is widely adopted and used by Argo CD.

I don't think there was consensus on "better", just that Workflows, CD, and Akuity already use(d) it, so it would be more familiar within Argoproj.

We had consensus on removing `kit`, but there were a few options on the table, such as `task`, `goreman`, and Nix `devenv` (see [above](https://github.com/argoproj/argo-workflows/pull/11928#discussion_r1344759627)).

> @terrytangyuan This might be the easiest way to get rid of kit, a careful revert should do the trick.
> 
> There will almost certainly be conflicts.

If that's what we want to do, I can tackle that

.github/workflows/ci-build.yaml
```diff
@@ -149,10 +149,12 @@ jobs:
       - name: Install manifests
         run: make install PROFILE=${{matrix.profile}} STATIC_FILES=false
       - name: Build controller
-        run: make controller kit STATIC_FILES=false
+        run: make controller task STATIC_FILES=false
       - name: Build CLI
         run: make cli STATIC_FILES=false
         if: ${{matrix.test == 'test-api' || matrix.test == 'test-cli' || matrix.test == 'test-java-sdk' || matrix.test == 'test-python-sdk'}}
+      - name: Start port forward
+        run: ./hack/port-forward.sh
```
agilgur5, line 157: this was removed in the `goreman` -> `kit` migration in https://github.com/argoproj/argo-workflows/pull/10599/files#diff-b8790dc873e80c2f7f63bf198b4da0147787ad8954767b839967d29f65155e2cL148, added it back now since `task` does not handle port forwarding

Makefile
agilgur5: Checked it myself and it looks like [`devenv up`](https://devenv.sh/processes/) is a replacement for [`goreman`](https://github.com/mattn/goreman), but neither are full task runners, they just run processes concurrently (no `deps`, `sources`, `env`, etc).

`task` and `kit` are `make` replacements rather than `goreman` replacements.

isubasinghe: Yeah it does have a task runner, infact Nix can replace the Makefile and task/goreman/etc. 
https://github.com/argoproj/argo-workflows/blob/master/dev/nix/flake.nix#L393-L403 is how the processes to run are declared. 

There is a manual step of setting up the k8s cluster needed at the moment but even this can be fully integrated into Nix. 

That controller should have a version attached to it, I did not set this version. So production builds using Nix are not really possible yet. This is still very fixable though, more than happy to talk about it if this is something we want to do.

.github/workflows/ci-build.yaml
agilgur5: this fixed the Controller tests. some Server tests are failing because apparently the Server didn't start at all. apparently `task` was waiting for `deps` to finish (which totally makes sense), but starting a servers will _never_ finish. so the `cmd`s need to be backgrounded properly. then everything works ok.

I think `kit`'s `ports` keyword (mentioned [above](https://github.com/argoproj/argo-workflows/pull/11928#discussion_r1343065238)) was just watching for ports to be opened before proceeding to the next task. nifty feature for servers.

Taskfile.yml
agilgur5: ok, I misinterpreted the issue above actually. it was indeed forwarding due to the `./hack/port-forward.sh` script, but the UI server actually never started b/c `task` was waiting for the `server` task to complete. The `server` task never completes though as it just runs until it is closed. See [below comment](https://github.com/argoproj/argo-workflows/pull/11928#discussion_r1344883152) for more details

Makefile
agilgur5: > Yeah it does have a task runner, infact Nix can replace the Makefile and task/goreman/etc.
> https://github.com/argoproj/argo-workflows/blob/master/dev/nix/flake.nix#L393-L403 is how the processes to run are declared.

Yea I've seen that -- that's equivalent to `goreman` but that doesn't have `deps`, `sources`, `env` etc -- features of `task`, `kit`, and `make`. Unless that feature exists somewhere too and just isn't being used?

agilgur5:
I have good news and bad news.

Good news:
- I have figured out how to get the tests to pass

Bad news:
- `task` is _not_ a replacement for `goreman`. `goreman` is a port of [`foreman`](https://github.com/ddollar/foreman), which addresses the whole [foreground/background process issue](http://blog.daviddollar.org/2011/05/06/introducing-foreman.html).
  - `task` is a replacement for `make`
    - `make` was still used in tandem with `goreman` per #3074 and #2965, bc `make` does _not_ solve the foreground/background process issue
      - i.e. it is out-of-scope for both `make` and `task`
- `kit` is a replacement for _both_ `make` _and_ `goreman`

alexec:
Let’s correct for the record: 

- Kit is maintained, but it’s basically feature complete. A lack of activity is a good thing, look at ‘jq’. 
- That said, the last release was 3 weeks ago. 
- Kit has no open bugs. 
- Kit is probably not the cause of the fan issue you mention. I’ve heard no similar reports from other users. 

@juliev0 do you need to open a bug?

Why does it need to be changed? I don’t see any discussion of the problem being solved here. 

That’s said, simpler toolchains are cheaper to maintain. If Nix can do the job , can we try that?

juliev0:
> Let’s correct for the record:
> 
> * Kit is maintained, but it’s basically feature complete. A lack of activity is a good thing, look at ‘jq’.
> * That said, the last release was 3 weeks ago.
> * Kit has no open bugs.
> * Kit is probably not the cause of the fan issue you mention. I’ve heard no similar reports from other users.
> 
> @juliev0 do you need to open a bug?
> 
> Why does it need to be changed? I don’t see any discussion of the problem being solved here.
> 
> That’s said, simpler toolchains are cheaper to maintain. If Nix can do the job , can we try that?

Sure, I’ll open a bug. Thanks

juliev0:
@alexec I'm not sure the trade offs of the various tools. I think one issue if we end up keeping it, is that it needs more documentation on how users should use it in the "running locally" page. I think people are generally confused about where to find their log files, they don't know anything about the tasks.yaml file, etc. Also, need to make sure all of the core functionality works. Apologies for not writing the bug as an issue earlier.

alexec:
It looks like we've had some tool creep in workflows (CD may have had the same) and this has resulted in too many tools.

Kit prints logs to the console if you run `CI=1 kit up`, that might make it easier to debug by default.

agilgur5:
> * Kit is maintained, but it’s basically feature complete. A lack of activity is a good thing, look at ‘jq’.

To be fair, I didn't say it wasn't. I merely compared its activity to `task`. Discussed more in https://github.com/kitproj/kit/issues/36#issuecomment-1753486495.
Also I'm not sure that `jq` is a fair comparison, given that it's much smaller in scope. `kit`'s README also compares to `jq` as a single binary, but `jq` is ~270kb while `kit` is ~8mb. But let's keep that in the other issue.

> Kit is probably not the cause of the fan issue you mention. I’ve heard no similar reports from other users.

All I know is that when I swapped it out, my fans made a lot less noise and my memory usage was considerably lower 😅 .
I did not trace it carefully, but it seemed like there might be a memory leak in the auto-restart feature or something as I noticed the memory just kept increasing when I made changes to the code and not going back down. It could be a memory leak in one of the tasks themselves too though; again, I did not trace it, that was merely an observation and not one of the initial motivators.

> Why does it need to be changed? I don’t see any discussion of the problem being solved here.

I listed out related discussions in issues and Slack in my opening comment, but Julie has succinctly summarized them above; a handful of bugs and a lack of familiarity and documentation with `kit`.

> That’s said, simpler toolchains are cheaper to maintain. If Nix can do the job , can we try that?

There's a separate discussion on Nix [above](https://github.com/argoproj/argo-workflows/pull/11928#discussion_r1344890504). It can fulfill what `goreman` did, but it does do all that `kit` does.

@alexec while you are here, would you mind elaborating on the reasons for using `kit` inside of Workflows? It replaced `goreman`, but `kit` is more generally a `make` replacement (with process management as an additional feature), yet `make` was not fully replaced and still makes up the bulk of tooling. If `make` were wholly replaced, I'd understand that (I am certainly not the biggest `make` fan), but as only a partial replacement I was a bit confused. `goreman` is similarly a very tiny-scoped tool that does pretty much one thing

alexec:
> All I know is that when I swapped it out,

I've raised a bug. but more diagnostic info would be useful. Your screen shot indicates that a container is using a lot of memory, but does not say what that container is? Do you have a `ps` output that shows that it is `kit`?

> handful of bugs

I count 2 bugs?

>  lack of familiarity

This is a reasonable argument. This would make tools like Nix and Goreman more attractive.

>  kit is more generally a make replacement for Make

Kit is not a replacement for me Make. It is a replacement for foreman.

agilgur5:
> I've raised a bug. but more diagnostic info would be useful. Your screen shot indicates that a container is using a lot of memory, but does not say what that container is? Do you have a `ps` output that shows that it is `kit`?

It's the Workflows `devcontainer`. I can put more details in the relevant bug, but as I mentioned before, it was just an observation, so I hadn't looked into it in detail to check for a root cause.

> I count 2 bugs?

That is enough for a handful 😝.
There are some others like Codespaces running indefinitely on `kit pre-up`: #11423. I even got a notification from GH that it used up 24 hours of compute before I even had access to a terminal.
I just looked into that yesterday and think it can be solved by switching the `devcontainer`'s `onCreateCommand` to a `postStartCommand`, but something weird is going on with how Codespaces is interacting with `kit` specifically at build-time. I saw it restart a few processes repeatedly too, not sure why.
I wouldn't be surprised if there were other bugs too.

> This is a reasonable argument. This would make tools like Nix and Goreman more attractive.

That's the primary reason, to be fair 👍
It just pops up occasionally from a bug or contributor issue. I wrote this PR as I was literally just comparing `task`'s syntax and wanted to see how easy it would be to transition. It was pretty easy, though `task` doesn't do process management like `foreman` / `goreman` unfortunately 😕 

> Kit is not a replacement for me Make. It is a replacement for foreman.

🤔 It does a lot more than `foreman` though? As it's a general task runner (like `task` and `make`) rather than a simple process manager

agilgur5:
> 🤔 It does _a lot_ more than `foreman` though? As it's a general task runner (like `task` and `make`) rather than a simple process manager

talked a bit with Alex offline. we didn't get into the full meat of this yet but I think Alex specifically meant `make` as a "build tool" rather `make` as a "task runner".
but we mostly use it as a "task runner" in Workflows and that is the primary comparison to `task` as well (and `task` compares itself to `make`).
(Earthly is more of a "build tool", for comparison)
Nix is more of a "dependency management tool", so also a bit orthogonal 😅 

basically a lot of tools with _partial_ overlap

terrytangyuan:
Should we live with it if `kit` already has all the features we need and there are no known stressing bugs?

terrytangyuan:
Should we close this?

agilgur5:
Thanks for following up on this Terry.

> and there are no known stressing bugs?

I do still have [a few months-old open issues](https://github.com/kitproj/kit/issues/created_by/agilgur5) on Kit, so it is not quite all clear skies. In particular, https://github.com/kitproj/kit/issues/48 is one I hit into with some frequency (and can take a few minutes to realize). The workaround is simple (quit `kit` and run `yarn install` outside of `kit`, then restart `kit`), but not a great experience and could be especially confusing for new contributors.

Since `task` still [does not handle background processes](https://github.com/go-task/task/issues/762) and contributors preferred `kit`'s DX to `goreman`'s DX, I was thinking what we should do is reduce `kit` to a background process runner and move the rest of the tasks out to limit the scope and bugs like the above.

I had initially thought to do a partial move to `task` to complete that, but at that point I was wondering if we should just entirely replace `Make` with another tool (`task` or otherwise). I thought that some contributors may not want to move away from `Make`, especially since we've been using it for so long and have familiarity with the current `Makefile`, so I was looking to do the partial replace of `kit` by moving foreground tasks into `Make`. #12583 is one effort toward that -- I still need to fix some race conditions with that, but the other potential con is that there's no fancy log handling for parallel tasks in `Make`.

I think reducing the scope of `kit` to background processes would be good, but the question of what tool to move to is still up in the air. Does anyone have thoughts on using `Make` or replacing it?

> Should we close this?

Otherwise I think we can close this since `task` is not a full `kit` replacement and `goreman` is not desired to replace the other portion of `kit`. Especially if we don't want to replace `Make` with `task`, then I think there's no reason to keep this open, other than some discussion of the above.

alexec:
What about Tilt.dev, or DevSpace.sh?

agilgur5:
It's been a long time since I've played around with Tilt (and I don't think I ever gave Devspace a spin as it's quite similar) but it wasn't quite clear if it supports background tasks (i.e. servers) and host processes (i.e. running binaries or `yarn start` / `webpack`). Being discussed in https://github.com/kitproj/kit/issues/36
Technically, the current usage of `kit` in Argo is all background tasks and their dependents.

Taskfile.yml
agilgur5: Removed the `download` mutex in #13350

Joibel:
Kit was replaced by tilt in #16235

PR #11945, opened 2023-10-04
fix(ui): faulty `setInterval` -> `setTimeout` in clipboard

agilgur5:
### Motivation

`setInterval` is incorrect in this context, it should be `setTimeout`
- this is a tiny optimization
- I just stumbled upon this while looking at tooltip usage. Also _unrelated_ to the [recent Slack thread](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1696440502865469), I happened to already be looking at tooltips after a [different thread](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1695922722824439) from last week.

### Modifications

- [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/setInterval) is for calling a function on an interval, i.e. once every 2s
  - it also requires clean-up with [`clearInterval`](https://developer.mozilla.org/en-US/docs/Web/API/clearInterval), which this component does not do
  - whereas [`setTimeout`](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) calls a function once after a timeout, i.e. after 2s
    - that is the correct behavior here -- to close the tooltip 2s after it was copied, not to continuously close it every 2s
    
- also convert anonymous function to named function for better traceability etc

### Verification

simple semantic change

### DIFF
--- a/ui/src/app/shared/components/clipboard-text.tsx
+++ b/ui/src/app/shared/components/clipboard-text.tsx
@@ -2,8 +2,8 @@ import {Tooltip} from 'argo-ui';
 import * as React from 'react';
 import {useState} from 'react';
 
-export const ClipboardText = ({text}: {text: string}) => {
-    const [justClicked, setJustClicked] = useState<boolean>(false);
+export function ClipboardText({text}: {text: string}) {
+    const [justClicked, setJustClicked] = useState(false);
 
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):
Good catch

PR #11946, opened 2023-10-04
fix(ui): `ClipboardText` tooltip properly positioned

agilgur5:
### Motivation

- it seems that with the [new version of `tippy`](https://github.com/argoproj/argo-ui/pull/297) (which was pulled in as part of #11585) the `overflow` property made it show up directly above the text (in the z direction)
  - this meant that you could not click on the clipboard to copy because it would cover the icon 😕

This was found by a user and noted [on Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1696440502865469). I then reproduced this myself, see below screenshot with incorrect positioning:
Screenshot 2023-10-04 at 5 13 33 PM -- clipboard before

### Modifications

- the original PR #8947 that added `overflow` here was for horizontal scrolling for multi-line args in code blocks (`<pre>` elements)
  - so I moved the `overflow` into the `pre` element's `class` more specifically and made it specific for the `x` direction
    - note that these code block elements do not have a clipboard button to copy their text

### Verification

Tested locally with a few different variations, see screenshots below.

Single-line clipboard tooltip positioned correctly:
Screenshot 2023-10-04 at 5 34 25 PM -- clipboard one-line

Multi-line clipboard tooltip positioned correctly:
Screenshot 2023-10-04 at 5 06 11 PM -- clipboard multi-line

Horizontal overflow on code blocks still works:
Screenshot 2023-10-04 at 5 07 16 PM -- overflow horizontal still works

Horizontal overflow on multi-line code blocks still works:
Screenshot 2023-10-04 at 5 11 23 PM -- horizontal scroll multi-line

### DIFF
--- a/ui/src/app/workflows/components/workflow-node-info/workflow-node-info.scss
+++ b/ui/src/app/workflows/components/workflow-node-info/workflow-node-info.scss
@@ -16,6 +16,7 @@
         line-height: 24px;
         display: inline-block;
         vertical-align: middle;
+        overflow-x: auto !important; // override `.workflow-details__atribute-grid pre` selector
     }
 
     &__artifact-details span {
@@ -61,7 +62,7 @@
         span {
             color: $argo-color-teal-5;
         }
-        
+
         i {
             float: right;
             font-size: 1.5em;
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):
Thanks!

Issue #11898, opened 2023-09-28
Add examples to commands' --help output


tico24:
## Summary

The Argo CLI has a lot of commands for which we provide no example in the `--help` output.

## Motivation

Examples are awesome.

## Proposal

For each item in this checklist, add an `Example` field. *EDIT by agilgur5*: See already-checked items for inspiration, use similar style & examples when possible, such as `argo list` for `argo archive list`, etc.

In some cases, the code itself may provide sufficient information to write some examples. For other commands, you may have to read through the docs to find or create examples.

*--- list has been updated & modified by agilgur5 ---* 

### `argo` commands
- [ ] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/cp.go#L30
- [ ] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/delete.go#L27
- [ ] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/get.go#L22
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/list.go#L64 -- Completed in #12076
- [ ] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/lint.go#L26
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/logs.go#L27
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/node.go#L29
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/resubmit.go#L37
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/resume.go#L22
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/retry.go#L40
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/stop.go#L38
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/submit.go#L32
- [ ] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/suspend.go#L14
- [x] https://github.com/argoproj/argo-workflows/tree/1f66a399e/cmd/argo/commands/terminate.go#L46
[...]

PR #11951, opened 2023-10-06
chore: added an example field to root.go

rum1887:
Fixes #11898


agilgur5:
Looks like this was replaced by #11969.

For future reference, please use the same PR instead of opening a new PR.

PR #11969, opened 2023-10-09
 docs(cli): Add an example field to argo archive command

rum1887:
Fixes #11898  Added an example field to the archive command.

### Motivation
Changes to support more detailed information when someone runs the archive Cobra command CLI

### Verification
Run - argo archive


yarn.lock
agilgur5, line 1: This should not be added. This is probably from running `yarn install` in the root directory instead of in the `ui` directory

go.sum
```diff
@@ -53,8 +53,8 @@ cloud.google.com/go/storage v1.33.0/go.mod h1:Hhh/dogNRGca7IWv1RC2YqEn0c0G77ctA/
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
 github.com/Azure/azure-sdk-for-go v66.0.0+incompatible h1:bmmC38SlE8/E81nNADlgmVGurPWMHDX2YNXVQMrBpEE=
 github.com/Azure/azure-sdk-for-go v66.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
-github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.1 h1:/iHxaJhsFr0+xVFfbMr5vxz848jyiWuIEDhYq3y5odY=
-github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.1/go.mod h1:bjGvMhVMb+EEm3VRNQawDMUyMMjo+S5ewNjflkep/0Q=
+github.com/Azure/azure-sdk-for-go/sdk/azcore v1.8.0 h1:9kDVnTz3vbfweTqAUmk/a/pH5pWFCHtvRpHYC0G/dcA=
+github.com/Azure/azure-sdk-for-go/sdk/azcore v1.8.0/go.mod h1:3Ug6Qzto9anB6mGlEdgYMDF5zHQ+wwhEaYR4s17PHMw=
```
agilgur5, line 57: Can you rebase off (or cherry-pick onto) `master`? It seems like this is from a dependabot commit that you accidentally amended

yarn.lock
rum1887: yes, removed.  
The build fails now
FAIL: TestHooksSuite/TestTemplateLevelHooksDagSuccessVersion (26.83s)
Should I retry the test?

rum1887: @agilgur5 can you review this :) ?

cmd/argo/commands/archive/root.go
```diff
@@ -8,6 +8,22 @@ func NewArchiveCommand() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "archive",
 		Short: "manage the workflow archive",
+		Example: `
+   		# List workflow archives:
```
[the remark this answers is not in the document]
agilgur5: the comments should ideally be the same as the ones in the individual commands. for instance, `argo archive get` has ["get a workflow in the archive"](https://github.com/argoproj/argo-workflows/blob/f01dbb1df1584c6e5daa288fd6fe7e8416697bd8/cmd/argo/commands/archive/get.go#L23)

The Workflow Archive is also singular. There are "workflows in the archive" as the other commands have.

the commands should also be alphabetized to match the existing documentation

docs/cli/argo_archive.md
```diff
@@ -6,6 +6,27 @@ manage the workflow archive
...
+   		# Delete a specific workflow archive:
+   		argo archive delete uid
+   		# List workflow archives by label key:
+   		argo archive list-label-key
+   		# List workflow archives by label value:
+   		argo archive list-label-value
+   		# Resubmit a workflow archive:
+   		argo archive resubmit uid
+		# Retry a workflow archive:
+   		argo archive retry uid
```
agilgur5, line 26: so this whole section is actually already repeated in the "See Also" section below. and in the sidebar in the docs as well. and in the CLI help output already.

so I don't think we should duplicate documentation here or in any `root` command really. 

if we have anything here, I suppose we could have:

```sh
# Print this message
argo archive

# Otherwise, see subcommands below
```

agilgur5 (changes requested):
I think we'll want to rewrite this a bit or just not have any `Long` descriptions for `root.go`, as they don't have much value. See in-line comment below for one possible alternative. Otherwise, I would suggest to focus on other CLI commands whose examples would be more useful. I suggested some corrections to the list in https://github.com/argoproj/argo-workflows/issues/11898#issuecomment-1758685634

Also can you rename the title of this PR to something more clear? there are many files named `root.go`, so the current title is ambiguous. 

I would suggest "docs(cli): Add an example to `argo archive` command"

PR #11971, opened 2023-10-09
fix: Replace antonmedv/expr with expr-lang/expr

terrytangyuan:
The current link is down and the author has created a fork in expr-lang. We use a `replace` here since the original repo would likely recover once GitHub handles the issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -279,4 +279,7 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
 
-replace github.com/go-git/go-git/v5 => github.com/argoproj-labs/go-git/v5 v5.4.7
+replace (
+	github.com/antonmedv/expr => github.com/expr-lang/expr v0.0.0-20230912141041-709c5dd55aa7
+	github.com/go-git/go-git/v5 => github.com/argoproj-labs/go-git/v5 v5.4.7
+)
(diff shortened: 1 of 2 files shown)


rbreeze (approved):

go.mod
```diff
@@ -279,4 +279,7 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
 
-replace github.com/go-git/go-git/v5 => github.com/argoproj-labs/go-git/v5 v5.4.7
+replace (
+	github.com/antonmedv/expr => github.com/expr-lang/expr v0.0.0-20230912141041-709c5dd55aa7
```
agilgur5, line 283: Confirmed that this commit SHA matches the one I confirmed to be from the original repo in https://github.com/argoproj/argo-workflows/discussions/11964#discussioncomment-7232791. 

Per my comment there though, we do not yet have confirmation from the author that [`expr-lang/expr`](https://github.com/expr-lang/expr) is an "official" fork. I think we should wait for confirmation from the author or use a known, trusted fork in the interim (such as [my fork](https://github.com/agilgur5/expr))

terrytangyuan:
Let's hold off on merging until the original author confirms.

terrytangyuan:
The author [confirmed](https://x.com/antonmedv/status/1711448876677280051?s=20) so we should be good to proceed

terrytangyuan:
The author [confirmed](https://x.com/antonmedv/status/1711448876677280051?s=20) so we should be good to proceed

agilgur5 (approved):

agilgur5:
Reverted this in #12033 since the original/root fork `antonmedv/expr` is back now (see also https://github.com/argoproj/argo-workflows/discussions/11964#discussioncomment-7242678)

PR #11978, opened 2023-10-10
build: fix `ARGO_LOG_LEVEL` -> `ARGO_LOGLEVEL`

agilgur5:
Fixes https://github.com/kitproj/kit/issues/41 and fixes the issue in this comment: https://github.com/argoproj/argo-workflows/issues/11879#issuecomment-1740263062

### Motivation

- credit to Julie and Alex for debugging this, I merely confirmed the issue and fixed it

- the flag is [`loglevel`](https://github.com/argoproj/argo-workflows/blob/a45afc0c87b0ffa52a110c753b97d48f06cdf166/cmd/workflow-controller/main.go#L54) and with the [automatic env prefix](https://github.com/argoproj/argo-workflows/blob/a45afc0c87b0ffa52a110c753b97d48f06cdf166/cmd/workflow-controller/main.go#L195), that becomes `ARGO_LOGLEVEL` as an env var
  - `ARGO_LOG_LEVEL` was a misconfiguration / typo as such

### Modifications

- Find+Replace all instances of `ARGO_LOG_LEVEL` -> `ARGO_LOGLEVEL`
  - they were only in the `Makefile` and the `tasks.yaml`

- also add `watch` list for the `ui-deps` task after my dive into `task` ([this line](https://github.com/argoproj/argo-workflows/pull/11928/files#diff-cd2d359855d0301ce190f1ec3b4c572ea690c83747f6df61c9340720e3d2425eR84) from #11928)
  - it seemed like `kit` kept re-installing deps when this was missing
    - actually, `kit` might be missing this feature? 🤔 not sure as the docs are unclear here: https://github.com/kitproj/kit/issues/40#issuecomment-1755913391 . this change may end up just having it auto-restart when deps change, which is still a good change
      - NOTE that `ui/node_modules` was removed here, as that caused an infinite restart loop when there was a dep change

### Verification

After setting the env variable in `tasks.yaml`, the Controller is now showing debug logs during a `kit` run:
```sh
controller: time="2023-10-10T20:51:00.001Z" level=info msg="omniscient-python is suspended, skipping execution" namespace=argo workflow=omniscient-python
controller: time="2023-10-10T20:51:00.015Z" level=debug msg="Patch cronworkflows 200"
port-forward: Handling connection for 9000
port-forward: Handling connection for 9000
controller: time="2023-10-10T20:51:04.973Z" level=debug msg="Get leases 200"
controller: time="2023-10-10T20:51:04.986Z" level=debug msg="Update leases 200"
```

Notably, I realized that the Server [doesn't even take a loglevel](https://github.com/argoproj/argo-workflows/blob/b2e3676503999e699cd3aaca62ff7bfff64464af/cmd/argo/commands/server.go#L42). Its parent command, `argo`, does take a [loglevel](https://github.com/argoproj/argo-workflows/blob/b2e3676503999e699cd3aaca62ff7bfff64464af/cmd/argo/commands/root.go#L120), and so that should be inherited, but it seems like that flag doesn't pass through to Viper to be read as an env variable automatically.
That will require further fixes -- a CLI flag, like `--loglevel=debug`, is currently required for the Server as the env var is currently ignored.


terrytangyuan (approved):

juliev0:
thanks for doing this @agilgur5 !

juliev0 (approved):

PR #11979, opened 2023-10-10
fix(ui): don't show pagination warning on first page if all are displayed

### LINKED ISSUE #11968: BUG UI: pagination warning is not always shown on last page of workflows
### Pre-requisites

When limiting the number of records in the workflow page a warning is shown that 'Workflows cannot be globally sorted when paginated'. I get that warning on all pages, except for the last page:
image

Only when the number of workflows to show on that page equals the page limit, the warning is shown:
image

I suppose this piece of code in pagination-panel.tsx is causing the issue:
` {props.pagination.limit > 0 && props.pagination.limit <= props.numRecords ? (
                <>
                    <WarningIcon /> Workflows cannot be globally sorted when paginated
                </>
            )`
            
It seems sufficient to have following condition only:
`props.pagination.limit > 0`

If you agree I can open a PR to fix the code.

### Version

v3.4.10

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

agilgur5:
Fixes my review comment in https://github.com/argoproj/argo-workflows/pull/11973#discussion_r1353099743
Fixes #11968 without a regression

### Motivation

- example: if you're on the first page and your pagination limit is 50 while you only have 10 total workflows, then the warning shouldn't show, as you already see all 10 workflows

### Modifications

- this new logic should handle both the first page and the last page issues
  - added a comment to word it out as well, which was actually very helpful in debugging and properly writing the logic (as it's a bit convoluted)

### Verification

Tested manually. See https://github.com/argoproj/argo-workflows/pull/11973#discussion_r1353099743 for more details

Before:
Screenshot 2023-10-10 at 2 01 08 PM -- before

After:
Screenshot 2023-10-10 at 2 28 04 PM -- after

### DIFF
--- a/ui/src/app/shared/components/pagination-panel.tsx
+++ b/ui/src/app/shared/components/pagination-panel.tsx
@@ -19,7 +19,8 @@ export function PaginationPanel(props: {pagination: Pagination; onChange: (pagin
                 }>
                 Next page <i className='fa fa-chevron-right' />
             </button>
-            {props.pagination.limit > 0 ? (
+            {/* if pagination is used, and we're either not on the first page, or are on the first page and have more records than the page limit */}
+            {props.pagination.limit > 0 && (props.pagination.offset || (!props.pagination.offset && props.numRecords >= props.pagination.limit)) ? (
                 <>
                     <WarningIcon /> Workflows cannot be globally sorted when paginated
                 </>


terrytangyuan (approved):

terrytangyuan:
Will include in https://github.com/argoproj/argo-workflows/issues/11997

Issue #11390, opened 2023-07-18
Versioned documentation


jmeridth:
# Summary

Adding versioning to argo-workflows docs https://argoproj.github.io/argo-workflows/

# Use Cases

When searching docs for the specific version of argo workflows you have deployed

Currently argo-cd has versioning in their documentation. Was introduced with [this PR](https://github.com/argoproj/argo-cd/pull/5099)

Would require setup and then migration/rewriting of docs to have docs split by version. 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

tico24:
Happy to help with the re-write. I have tried to set up versioning in the past, and failed so I need an adult to help me get that bit set up.

caelan-io:
Tim and I just remembered why the Argo Workflows docs weren't an easy switch like it seemed it was for Argo CD's docs: Argo Workflows isn't using Readthedocs for its docs website. 

I believe this means we need to scope out a way to incorporate versioning into the mkdocs build of the Argo Workflows docs. 

Resources related to this:
- https://pypi.org/project/mkdocs-versioning/
- https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

agilgur5:
This would be useful for features that have yet to be released as well. Just had someone file an issue for a feature I wrote that hasn't been released yet (c.f. https://github.com/argoproj/argo-workflows/issues/11412 / https://github.com/argoproj/argo-workflows/pull/11445). Similar to how Argo CD has the `latest` version of the docs.

This gets into some complications about versioned docs that I've run into before too:
- A docs improvement or clarification can be back-ported to old docs versions
- New docs for previously undocumented features can be back-ported to old docs too (such as #11297)
- Docs about new features should be released with the feature
- Generated docs should follow the same principles

Or, in other words, SemVer applies to docs too 😅

agilgur5:
Just jotting down here that in the previous contributor meeting ([on Aug 15th](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.qogh82i9yij0)), we discussed that moving to Read the Docs like Argo CD may be easier and more consistent than adding versioning to `mkdocs`.

Briefly mentioned redirects as well. If that is the target state, I may want to hold off adding some `mkdocs` plugins, like one for redirects as I mentioned in #11516.

[...]

PR #11980, opened 2023-10-10
docs(cli): correct headings in `argo` command's `Long` description

agilgur5:
### Motivation

- these were H1s before, and there can only be 1 H1 per page
  - they also _should_ be H4s, because these are under the "Synopsis", which is an H3 
    - this fixes some very wacky formatting in the docs as well as HTML semantic errors

### Modifications

- h1 -> h4 in `argo` root command's `Long` description

### Verification

Before:
Screenshot 2023-10-10 at 2 39 08 PM --before

After:
Screenshot 2023-10-10 at 2 40 35 PM --after


terrytangyuan (commented):
This should be docs instead of fix?

terrytangyuan:
Generated docs are docs change.

agilgur5:
Sure, but it also changes the source code / metadata of the CLI. nbd either way to be fair, but we're not currently cherry-picking docs PRs, which we will need to do for [versioned docs](https://github.com/argoproj/argo-workflows/issues/11390). so we'll need some way to decipher docs PRs that are for new features vs docs PRs for existing features (most will fall into the latter, I imagine). semantically, it's pretty quick to tell from the PR, but there's no way to do that automatically right now unless we differentiate in some way

agilgur5:
fixed codegen

agilgur5:
Merged `master` to get new status checks / actions from #12006 to run

terrytangyuan (approved):

PR #12006, opened 2023-10-14
ci: only run jobs when relevant files have been changed

### LINKED ISSUE #10156: Suggestion: skip unnecessary workflows if only have documentary changes
# Summary

As I mentioned in the title. It's totally not necessary to run the whole workflow all the time.

# Use Cases

As you can see below, I only made some documentary changes in this PR

image

See https://github.com/argoproj/argo-workflows/pull/10155

We could skip the e2e workflow by some filters.
---

### LINKED ISSUE #10265: Better segmentation of tests
# Summary

Currently every single test is run on a PR, we should only run tests if they are relevant to the changed files. 

# Use Cases
This will increase the speed of iteration for developers, but more importantly it will lessen the probability of blocking PRs due to flakey tests. A documentation PR should not be blocked by a flakey test for example.

agilgur5:
Fixes #10156
Fixes #10265
- Note: these can be further segmented and improved as well, see "Future Work" below

### Motivation

Make CI run faster for certain kinds of PRs, use less CI minutes (and just reduce unnecessary compute in general), and limit the blast radius of flaky tests as well
- e.g. don't run E2E tests if only UI or only Docs have changed
  - or don't run UI CI when UI has not chaged
 

#### Changed Files checks

- use [`tj-actions/changed-files`](https://github.com/tj-actions/changed-files) action for this
  - the most popular and full featured one I could find

- run `changed-files` in its own job that must run before all other jobs
  - have it `output` booleans for specific changes -- limit all the `changed_files` nuances, naming, syntax, etc to that one job
    - job `outputs` are also needed for skipping other jobs, as step outputs can't be directly accessed
      - see https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs and https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context
- have other jobs specify it in their `needs` and then skip if not needed with `if`

- use multi-output variant of `changed-files` YAML directive
  - so can check e2e vs lint vs UI etc
  - note that `changed-files` supports a YAML variant [with list anchoring/aliasing](https://github.com/tj-actions/changed-files/blob/2a10bef1b42044172f2e64d40beeb8fbad792438/test/changed-files.yml#L8-L11)
- use `any_modified` as that includes added, copied, modified, renamed, and deleted (ACMRD)
  - `any_changed` does not include deletions

#### `needs` changes for `codegen` and `lint`

- both can run independently of each other and independently of tests
  - they should fail fast if the other one fails though, I suppose

- otherwise this breaks some of the checks, since if tests are skipped, then codegen and lint are skipped too

#### Changes to Status Checks

Note that due to lack of GH features such as https://github.com/actions/runner/issues/952 and https://github.com/orgs/community/discussions/9141, there is a workaround embedded here for the E2E test matrix
- skipping the overall `e2e-tests` job will cause status checks for individual E2E tests to just hang indefinitely
  - so instead we can change the required status checks to check the new composite result
    - that or we'd have to put an `if` for every single `step` of the E2E tests _and_ also check for changed files within each E2E test
      - as there is no early exit feature such as https://github.com/actions/runner/issues/662
      - as steps can't use outputs from other jobs
      - and even then, the E2E test job would report as "success" instead of "skipped" as there is similarly no way to manually mark a job as skipped, which is a bit misleading
- the new `e2e-tests-composite-result` job aggregates all of the results as a workaround
  - see https://github.com/orgs/community/discussions/9141#discussioncomment-2296809 and https://github.com/orgs/community/discussions/26822#discussioncomment-3305794
  - **once this PR is merged, the required status checks on branch protection need to be changed**
    - individual E2E test checks removed, composite check required now
    - note that this will also require that PRs opened earlier than this PR be rebased on top of this to get the new check so that they can pass CI

### Verification

I tested this quite extensively in my fork in https://github.com/agilgur5/argo-workflows/pull/1:
- [Correctly only ran UI job](https://github.com/agilgur5/argo-workflows/actions/runs/6519175589/job/17705386816?pr=1)
- [Correctly ran UI + Lint](https://github.com/agilgur5/argo-workflows/actions/runs/6519460406/job/17705986839?pr=1)
- [Correctly only ran Lint](https://github.com/agilgur5/argo-workflows/actions/runs/6519499844/job/17706063728?pr=1)
- [Correctly only ran Codegen](https://github.com/agilgur5/argo-workflows/actions/runs/6519534651/job/17706136684?pr=1) (this was cancelled early, but we only need to confirm that it ran)
- [Correctly only ran E2E Tests](https://github.com/agilgur5/argo-workflows/actions/runs/6519559416/job/17706194361?pr=1) (this was cancelled early, but we only need to confirm that it ran)
- [Correctly ran Unit Tests, Lint, + E2E Tests](https://github.com/agilgur5/argo-workflows/actions/runs/6519575693/job/17706228655?pr=1) (this was cancelled early, but we only need to confirm that it ran)
  - as Lint + E2E Tests will run whenever Unit Tests run too (since source files were changed)
- [Correctly ran all jobs](https://github.com/agilgur5/argo-workflows/actions/runs/6519606501/job/17706287740?pr=1)
- Confirmed that the new E2E composite check works correctly -- see all of the above

### Notes to Reviewers

- See the "Changes to Status Checks" section above that we will need to implement
- We will also need to be a bit careful in making sure that any new code is properly listed in the files lists

### Future Work

1. Further segment E2E tests -- not all tests necessarily need to run for all back-end changes 
1. Add to other GH Workflows, like the Docs Workflow?
    - would not be a huge efficiency increase, but would limit some unnecessary runs
1. Add a check for `argoexec-image` job as well
    - this one may require some more effort to get everything necessary, particularly with repetition from `.dockerignore` etc


.github/workflows/ci-build.yaml
```diff
@@ -17,8 +17,97 @@ permissions:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 50 # assume PRs are less than 50 commits
+      - name: Get relevant files changed per group
+        id: changed-files
+        uses: tj-actions/changed-files@v39
+        with:
+          files_yaml: |
+            common: &common
+              - .github/workflows/ci-build.yaml
```
agilgur5, line 40: we may want to add `Makefile` and `tasks.yaml` here, but I wanted to avoid excessive testing. It may be better to be more conservative in case of a breaking `Makefile` change, for example, but those are rarely done in isolation

.github/workflows/ci-build.yaml
```diff
@@ -17,8 +17,97 @@ permissions:
...
+            tests: &tests
+              - *common
```
agilgur5, line 42: as mentioned in the PR description, `tj-actions/changed-files` supports a YAML variant [that has array anchors & aliases](https://github.com/tj-actions/changed-files/blob/2a10bef1b42044172f2e64d40beeb8fbad792438/test/changed-files.yml#L8-L11)

.github/workflows/ci-build.yaml
```diff
@@ -198,9 +288,28 @@ jobs:
         if: ${{ failure() }}
         run: kubectl logs -c wait -l workflows.argoproj.io/workflow --prefix
 
+  # workaround for status checks -- check this one job instead of each individual E2E job in the matrix
+  # this allows us to skip the entire matrix when it doesn't need to run while still having accurate status checks
+  # see https://github.com/orgs/community/discussions/9141#discussioncomment-2296809 and https://github.com/orgs/community/discussions/26822#discussioncomment-3305794
+  e2e-tests-composite-result:
```
agilgur5, line 294: see the "Changes to Status Checks" section in the PR description

.github/workflows/ci-build.yaml
```diff
@@ -198,9 +288,28 @@ jobs:
...
+          if [[ $result == "success" || $result == "skipped" ]]; then
+            exit 0
+          else
+            exit 1
+          fi
+
   codegen:
     name: Codegen
-    needs: [ tests ]
+    needs: [ changed-files ]
```
agilgur5, line 311: See the "Changes to `needs` for `codegen` and `lint`" section in the PR description

agilgur5 (commented):
left some comments for some in-line clarity about some changes

.github/workflows/ci-build.yaml
isubasinghe: In my experience, the Makefile doesn't change that frequently and I think it is better to err on the side of safety and have this run test. 

I was actually quite reliant on the testing when I did my nix stuff.

isubasinghe:
> otherwise this breaks some of the checks, since if tests are skipped, then codegen and lint are skipped too

Can you elaborate on this? 
If tests are skipped the lint is also skipped? 
Would that be an issue for docs changes then? Those actually get linted. 
Would we be not linting them in that case?

agilgur5:
> Can you elaborate on this?

If one or more items of a `needs` array is skipped, the dependent job will also skip.
In the previous version of the code, if the `tests` job was skipped, the `lint` and `codegen` jobs would also be skipped. `tests` never had a conditional before though.

 So as a result, I had to remove the `needs` from the `lint` and `codegen` jobs so that they run regardless if `tests` is skipped or not. They are independent jobs anyway, so the `needs` block seemed to be more fast fail purposes; i.e. don't run `lint` or `codegen` if tests fail. Right now they will run either way

> Would that be an issue for docs changes then?

In `ci-build.yaml`, which is the only GH workflow I changed here, the docs lint runs as part of the `lint` job mentioned above 

There is also a separate `docs.yaml` GH workflow that builds the docs. That one I did not change in this PR; I left as "Future Work" if we want to tackle that. It's simple but duplicate and low value compared to `ci-build.yaml`. (it would have to be duplicative as GH workflows don't share any information with each other)

agilgur5:
> The 3 are independent jobs anyway, so the previous `needs` block seemed to be more for fast fail purposes; i.e. don't run `lint` or `codegen` if tests fail.

Mmm I think there is a way to leave the `needs` in if I re-work the conditional a bit, [per the `needs` docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds). Actions have some implicit conditionals that are very disorienting IMO. It would be kind of if Argo's DAG `dependencies` and enhanced `depends` could be used together -- some very wacky edge cases would ensue -- simpler for them to be mutually exclusive fields.

.github/workflows/ci-build.yaml
agilgur5: > I was actually quite reliant on the testing when I did my nix stuff.

Not really impactful on the decision, but CI doesn't use Nix? So I am a bit confused by this statement, how would CI be helpful for Nix changes?

isubasinghe: I had to make changes to the Makefile and was reliant on the CI to check if it retained its previous functionality

agilgur5: Oh for regression testing, gotcha

isubasinghe:
Ah thanks @agilgur5, makes sense now, I misunderstood previously, in that case you have fixed an issue withe the CI process really.

isubasinghe (commented):
I'd personally prefer if the Makefile was included, I think the tasks.yaml can probably be avoidede though

agilgur5:
> in that case you have fixed an issue withe [sic] the CI process really.

Sort of; it would never occur previously because there wasn't a conditional on `tests` previously. That was very subtle though; was previously very easy to add a conditional and not realize it would cause further skips

caelan-io (approved):

.github/workflows/ci-build.yaml
agilgur5: Added both `Makefile` and `tasks.yaml` for the safest option

.github/workflows/ci-build.yaml
```diff
@@ -17,8 +17,99 @@ permissions:
...
+      e2e-tests: ${{ steps.changed-files.outputs.e2e-tests_any_modified == 'true' }}
+      codegen: ${{ steps.changed-files.outputs.codegen_any_modified == 'true' }}
+      lint: ${{ steps.changed-files.outputs.lint_any_modified == 'true' }}
+      ui: ${{ steps.changed-files.outputs.ui_any_modified == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 50 # assume PRs are less than 50 commits
+      - name: Get relevant files changed per group
```
terrytangyuan, line 34: This list seems hard to maintain though

.github/workflows/ci-build.yaml
```diff
@@ -17,8 +17,99 @@ permissions:
...
+              # generation scripts
+              - hack/cli/**
+              - hack/jsonschema/**
+              - hack/swagger/**
+              - hack/auto-gen-msgh.sh
+              - hack/crdgen.sh
+              - hack/crds.go
+              - hack/docgen.go
+              - hack/parse_examples.go
+              - hack/swaggify.sh
```
terrytangyuan, line 84: Cannot we combine these using regex?

agilgur5: everything in `hack/` is not actually used for codegen -- there are various other scripts in there, it's kind of a catch-all directory. I picked out the individual scripts that are actually used in codegen.

The better question would be if we could better organize the `hack/` directory

.github/workflows/ci-build.yaml
agilgur5: There's no real way around that -- task runners pretty regularly ask that you keep a list of files that should be "memoized" / "cached" / "watched" as tooling does not usually track all the files that were touched.

Also we don't change tooling that frequently; these only need changing when we introduce something new. I think in the past 6+ months the only things that would have affected this list are the Kit replacement (#11928) and the `mkdocs` replacement (partly mentioned in #11390).

So I think it's actually pretty reasonable to maintain. We only really have to be careful with it when we change large parts of the build system or create a totally new directory. The vast majority of PRs do not do either of those.

terrytangyuan (approved):

.github/workflows/ci-build.yaml
agilgur5: > The better question would be if we could better organize the `hack/` directory (which I do think would be good to do), but that was not the goal of this PR

Took a first draft at this in #12774

PR #12015, opened 2023-10-16
fix: Fix the Maximum Recursion Depth prompt link in the CLI.

luyang93:


agilgur5 (approved):
Yea that's the correct link: https://argoproj.github.io/argo-workflows/scaling/#maximum-recursion-depth

`.md` is how we refer to it inside the docs themselves, which gets converted when rendered. `mkdocs` doesn't run in the CLI though, so this needs to be the rendered link.

Thanks for the fix!

terrytangyuan (approved):

PR #12020, opened 2023-10-17
fix: suppress error about unable to obtain node

mikutas:
Relates to #11451

### Motivation

After upgrading to v3.4.11, the error log is often reported from monitoring.

```
level=error msg="was unable to obtain node for " namespace=default workflow=hello-world-jq82w
```

### Modifications

When boundaryID is empty, we don't try to get node and also don't raise any error.


terrytangyuan (approved):

mikutas:
@terrytangyuan Thank you for reviewing! Please retry required `Docs/docs` job because it seems to be a temporary failure. 🙏 
image

PR #12024, opened 2023-10-17
feat: Fall back to retrieve logs from live pods

terrytangyuan:
Today, if the Argo server fails to pull archived logs, it won’t use pod logs even if the pod still exists on the cluster. This means that in cases where the final upload fails due to streaming errors (or if the user’s SA does not have RBAC access to the artifact creds), Argo server will not display any logs for the workflow step after failing to retrieve archived logs even if the pod and its logs still exist on the cluster.

This PR adds a fallback to retrieve logs from live pods in those cases.


rbreeze (approved):
LGTM!

Only question - when this fallback occurs, is it clear to the user that they are not viewing archived logs anymore?

terrytangyuan:
> Only question - when this fallback occurs, is it clear to the user that they are not viewing archived logs anymore? 

There's currently no differentiation. The type of logs they are seeing is not shown to users.

Issue #12031, opened 2023-10-18
Improve OpenSSF Scorecard checks


agilgur5:
# Summary

Closely related to #9769 and #11953

Can see our scores for OpenSSF Scorecard from the badge link here: https://api.securityscorecards.dev/projects/github.com/argoproj/argo-workflows

<details> <summary> Scorecard JSON output, pretty-printed </summary>

```json
{
  "date": "2023-10-16",
  "repo": {
    "name": "github.com/argoproj/argo-workflows",
    "commit": "165f57fd1b40256cdd41c41cf3fc2b9e4664c9fe"
  },
  "scorecard": {
    "version": "v4.13.0-21-g8eaf0d76",
    "commit": "8eaf0d7647a3f50d80615812c72a277fd568fb6d"
  },
  "score": 6.5,
  "checks": [
    {
      "name": "Maintained",
      "score": 10,
      "reason": "30 commit(s) out of 30 and 18 issue activity out of 30 found in the last 90 days -- score normalized to 10",
      "details": null,
      "documentation": {
        "short": "Determines if the project is \"actively maintained\".",
        "url": "https://github.com/ossf/scorecard/blob/8eaf0d7647a3f50d80615812c72a277fd568fb6d/docs/checks.md#maintained"
      }
    },
    {
      "name": "Code-Review",
      "score": 10,
      "reason": "all changesets reviewed",
      "details": null,
      "documentation": {
        "short": "Determines if the project requires human code review before pull requests (aka merge requests) are merged.",
        "url": "https://github.com/ossf/scorecard/blob/8eaf0d7647a3f50d80615812c72a277fd568fb6d/docs/checks.md#code-review"
      }
    },
    {
      "name": "License",
      "score": 10,
      "reason": "license file detected",
      "details": [
        "Info: License file found in expected location: LICENSE:1",
        "Info: FSF or OSI recognized license: LICENSE:1"
      ],
      "documentation": {
        "short": "Determines if the project has defined a license.",
        "url": "https://github.com/ossf/scorecard/blob/8eaf0d7647a3f50d80615812c72a277fd568fb6d/docs/checks.md#license"
      }
    },
    {
      "name": "CII-Best-Practices",
      "score": 5,
      "reason": "badge detected: passing",
      "details": null,
      "documentation": {
        "short": "Determines if the project has an OpenSSF (formerly CII) Best Practices Badge.",
[...]

PR #12036, opened 2023-10-18
chore(deps): automatically `audit fix` UI deps

agilgur5:
Partial fix for #12031, Vulnerabilities

### Motivation

- 77 vulns -> 1 vuln
  - the 1 remaining is a dep of `swagger-ui-react`, which has it [pinned to a specific version](https://github.com/argoproj/argo-workflows/blob/5c264c094104645a4c917a9a23615424d564d1e4/ui/yarn.lock#L7702)
    - see also https://github.com/argoproj/argo-workflows/pull/11418#issuecomment-1696260188; `swagger-ui-react` is blocked on upgrading due to other devDeps
- Some dependabot PRs have stalled because it's not deduplicating dependencies 😕 
  - See https://github.com/argoproj/argo-workflows/pull/11728#issuecomment-1707477279
- Improve OpenSSF Scorecard Vulnerabilities check
  - all but one of the vulns currently reported are in JS deps (the remaining one [is a vuln in the AWS Go SDK](https://osv.dev/list?ecosystem=&q=GO-2022-0646) with no patch currently available)

### Modifications

- ran `npx yarn-audit-fix && yarn deduplicate`
  - see https://github.com/antongolub/yarn-audit-fix, a Yarn variant for `npm audit fix`

### Verification

- ran `yarn` after, no additional changes
  - `audit fix` will only bump versions in a SemVer compatible way, so no major bumps here; mostly patches

### DIFF
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -58,12 +58,12 @@
     resize-observer-polyfill "^1.5.1"
     throttle-debounce "^5.0.0"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.22.10", "@babel/code-frame@^7.22.5":
-  version "7.22.10"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.22.10.tgz#1c20e612b768fefa75f6e90d6ecb86329247f0a3"
-  integrity sha512-/KKIMG4UEL35WmI9OlvMhurwtytjvXoFcGNrOvyG9zIzA8YmPjVtIZUf7b05+TPO7G7/GEmLHDaoCgACHl9hhA==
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
This should be part of ui lint? `npx yarn-audit-fix && yarn deduplicate`

agilgur5:
> This should be part of ui lint? `npx yarn-audit-fix && yarn deduplicate`

No I don't think so, that's why I didn't add it.
Here's a pointed counter-example to demonstrate why: if you change no dependencies yourself in a PR, `audit fix` may still have changes.
After all, it depends on if there are vulns in your dependencies and a SemVer compatible fix available. Since vulns are discovered in real-time, that is a moving target.

`yarn deduplicate` on the other hand will only make changes if you made changes to the deps. It will effectively no-op when there are no changes to deps (and if the list has been kept deduped as it is within the UI lint check).

terrytangyuan (approved):

PR #12040, opened 2023-10-19
chore: Add script to auto cherry-pick commits for patch release

terrytangyuan:
This would be very useful for chore commits.


hack/auto-cp.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env sh
+
+set -eu
+# this script prints out a list a commits that are on master that you should probably cherry-pick to the release branch
```
agilgur5, line 4: this comment is not valid for this script -- looks this was copied from `what-to-cherry-pick.sh`

hack/auto-cp.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# print all the commits that need cherry-picking
+git log --oneline --grep ${commitGrepPattern} $base...master | while read -r m; do
+  commit=`grep -q "$(echo $m | prNo)" /tmp/prs && echo ${m:0:9}`
+  set +e
+  echo cherry-picking: $commit
+  if ! git cp $commit; then
+    echo failed to cherry-pick $commit
+    git cp --abort
+  fi
```
agilgur5, line 39: this is the main difference, right? I would think that it probably makes more sense to add a flag to `what-to-cherry-pick.sh` instead of copy+pasting a whole script and needing to keep both in sync.

the default behavior would basically be a "dry run" then.

or alternatively you could pipe the results of `what-to-cherry-pick.sh` into this script

docs/releasing.md
```diff
@@ -13,6 +13,14 @@ Then get a list of commits you may want to cherry-pick:
 ./hack/what-to-cherry-pick.sh release-3.3 "ci"
 ```
 
+To automatically cherry-pick, run the following:
+```
```
agilgur5, line 17:
```suggestion
```bash
```

this is failing CI for docs

terrytangyuan:
I meant to keep this as a draft since I would need to consolidate this script with `what-to-cherry-pick.sh`.

agilgur5 (approved):
seems good to me 👍

hack/cherry-pick.sh
```diff
@@ -0,0 +1,47 @@
...
+
+prs $base $br
+prs $base master
+
+# find PRs added to master
+diff /tmp/$br /tmp/master | grep '^> ' | cut -c 3- > /tmp/prs
+
+# print all the commits that need cherry-picking
+git log --oneline --grep ${commitGrepPattern} $base...master | while read -r m; do
+  if [ "$dryRun" = true ]; then
```
agilgur5, line 36:
```suggestion
  if [[ "$dryRun" == "true" ]]; then
```

pretty sure this is the safer syntax IIRC

hack/cherry-pick.sh
```diff
@@ -0,0 +1,47 @@
...
+    grep -q "$(echo $m | prNo)" /tmp/prs && echo $m
+  else
+    commit=`grep -q "$(echo $m | prNo)" /tmp/prs && echo ${m:0:9}`
```
agilgur5, line 39:
```suggestion
    commit=$(grep -q "$(echo "$m" | prNo)" /tmp/prs && echo "${m:0:9}")`
```

shellcheck recommendation as backticks are legacy

hack/cherry-pick.sh
```diff
@@ -0,0 +1,47 @@
...
+    set +e
```
agilgur5, line 40: you should probably add `set -e` at the end of the `if`, as otherwise it lasts for the rest of the shell session

agilgur5: or actually should probably rewrite this to not need `set +e`

hack/cherry-pick.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -eu
+# See releasing.md for instructions.
```
agilgur5, line 3:
```suggestion
# See docs/releasing.md for instructions.
```

as it's in a different directory than this one (`hack/`)

hack/cherry-pick.sh
agilgur5: shellcheck also recommends quoting all the variables as strings, but that's a larger change to the rest of the file. nearly every `$...` reference should be changed to `"$..."`

agilgur5 (commented):
made some small shell recommendations that I missed

agilgur5:
oh I totally missed some recommendations my first review, woops, my bad. first time I looked at it on my phone, I had some suggestions; then when I got to my computer the first time, I completely forgot them 😅

hack/cherry-pick.sh
agilgur5: actually I think the `set +e` _already_ isn't necessary with the `if ! cmd; then` syntax you used since the conditional acts as a trap

terrytangyuan:
Thanks. Here it is https://github.com/argoproj/argo-workflows/pull/12043

agilgur5:
This and #12043 seemed to have been manually copied over to `release-3.5` in 40de363a83be731edf8b1fe014915f877855fc4a
- The docs changes were missing there, so it caused a merge conflict for later backports, such as for #12714
- So I backported the docs change from here in 53d5684b5aab372ad713e275eb44e3da630ba0b8
  - And #12043 became an empty commit in 94b3e91b877c3b86e6313ae54e5f5a883de7b9a4

PR #12046, opened 2023-10-20
refactor(ui): WorkflowsToolbar component from class to functional

juijeong8324:
Partial Fixes for #9810

### Motivation

Use newer, modern functional components compatible with hooks instead of legacy class-based components

### Modifications

1. class fo functional
- `this.props` -> arguments
- methods -> inner functions

<br>

2. Delete `loadWorkflows()` from `WorkflowsToolBar` props
- Beacause `ListWatch()` update workflows when workflows was performed action

<br>

3. `getNumberSelected()` -> `WorkflowsToolbar`'s local variable `numberSelected`
- This function get the number of selected workflows from calling props. However it gets called multiple times each rendering. 
-  So, when they are updated, we can get the number of selected workflows from `numberSelected` _once_ without multiple function call.  

<br>

4. using `popup` and `notifications` with `useContext`

<br>

5. delete `renderActions`
- move rendering Action buttons element to`WorkflowsToolBar`s rendering  
- local variable
   - `actionButtons` -> delete
   - `actions` -> local variable in useMemo
   - `disabled` -> local variable in useMemo
   - `groupActions` -> variable `groupAction` in `WorkflowsToolbar`
       - using `useMemo` with dependency  `props.selectedWorkflows`

<br>

6. modify name in `WorkflowsGroupAction` and variable
- interface `WorkflowsGroupAction` -> `WorkflowsOperation`
   - Beacuse I think it is _ambiguous what 'group' means 'workflows' or 'Action'_
   - WorkflowsOperation is similar with WorkflowOperation
- `groupIsDisabled` -> `isDisabled`
  - I think _'disabled' is more related with "operation"_ not "group", 
  - _Also, `isDiabled` -> `actionIsDiabled` in `WorkflowsToolBarProps`_ 
- `className` -> delete
  - It is same `action.title`
- `groupAction` -> `workflowsAction`

<br>

7. modify function `performActionOnSelectedWorfklows`
- async/await -> delete
  - `await` keyword use in popup that check action  
  - `popup` move to `workflowsAction`  
- _This function is called only when action are performed in the selected workflow._

<br>

### Verification

Pure refactor, no real semantic changes
Tested in Local environment
- Screenshot below, Action of delete workflows
image
image

- Screenshot below, Action of resubmit workflows
image
image

- Screenshot below, action button when workflows running
image

### DIFF
--- a/ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
+++ b/ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
@@ -1,7 +1,8 @@
 import {NotificationType} from 'argo-ui';
 import * as React from 'react';
+import {useContext, useMemo} from 'react';
 import {isArchivedWorkflow, isWorkflowInCluster, Workflow} from '../../../../models';
-import {Consumer, ContextApis} from '../../../shared/context';
+import {Context} from '../../../shared/context';
 import {services} from '../../../shared/services';
 import * as Actions from '../../../shared/workflow-operations-map';
 import {WorkflowOperation, WorkflowOperationAction, WorkflowOperationName} from '../../../shared/workflow-operations-map';
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
 
 interface WorkflowsToolbarProps {
     selectedWorkflows: Map<string, Workflow>;
-    loadWorkflows: () => void;
-    isDisabled: Actions.OperationDisabled;
+    actionsIsDisabled: Actions.OperationDisabled;
     clearSelection: () => void;
 }
 
-interface WorkflowGroupAction extends WorkflowOperation {
-    groupIsDisabled: boolean;
-    className: string;
-    groupAction: () => Promise<any>;
+interface WorkflowsOperation extends WorkflowOperation {
+    isDisabled: boolean;
+    workflowsAction: () => Promise<any>;
 }
 
-export class WorkflowsToolbar extends React.Component<WorkflowsToolbarProps, {}> {
-    constructor(props: WorkflowsToolbarProps) {
-        super(props);
-    }
+export function WorkflowsToolbar(props: WorkflowsToolbarProps) {
+    const {popup, notifications} = useContext(Context);
+    const numberSelected: number = props.selectedWorkflows.size;
 
-    public render() {
-        return (
-            <Consumer>
-                {ctx => (
-                    <div className={`workflows-toolbar ${this.getNumberSelected() === 0 ? 'hidden' : ''}`}>
-                        <div className='workflows-toolbar__count'>
-                            {this.getNumberSelected() === 0 ? 'No' : this.getNumberSelected()}
-                            &nbsp;workflow{this.getNumberSelected() === 1 ? '' : 's'} selected
-                        </div>
-                        <div className='workflows-toolbar__actions'>{this.renderActions(ctx)}</div>
-                    </div>
-                )}
-            </Consumer>
-        );
-    }
+    const groupAction = useMemo<WorkflowsOperation[]>(() => {
+        const actions: any = Actions.WorkflowOperationsMap;
+        const disabled = props.actionsIsDisabled;
 
-    private getNumberSelected(): number {
-        return this.props.selectedWorkflows.size;
-    }
+        return Object.keys(actions).map((actionName: WorkflowOperationName) => {
+            const action = actions[actionName];
+            return {
+                title: action.title,
+                iconClassName: action.iconClassName,
+                isDisabled: disabled[actionName],
+                action,
+                workflowsAction: async () => {
+                    //check for action
+                    const confirmed = await popup.confirm('Confirm', `Are you sure you want to ${action.title.toLowerCase()} all selected workflows?`);
+                    if (!confirmed) {
+                        return;
+                    }
 
-    private async performActionOnSelectedWorkflows(ctx: ContextApis, title: string, action: WorkflowOperationAction): Promise<any> {
-        const confirmed = await ctx.popup.confirm('Confirm', `Are you sure you want to ${title.toLowerCase()} all selected workflows?`);
-        if (!confirmed) {
-            return Promise.resolve(false);
-        }
+                    // check for delete from archived workflows
+                    let deleteArchived = false;
+                    if (action.title === 'DELETE') {
+                        // check for delete workflows from archived workflows
+                        for (const entry of props.selectedWorkflows) {
+                            if (isArchivedWorkflow(entry[1])) {
+                                deleteArchived = await popup.confirm('Confirm', 'Do you also want to delete them from the Archived Workflows database?');
+                                break;
+                            }
+                        }
+                    }
 
-        let deleteArchived = false;
-        if (title === 'DELETE') {
-            for (const entry of this.props.selectedWorkflows) {
-                if (isArchivedWorkflow(entry[1])) {
-                    deleteArchived = await ctx.popup.confirm('Confirm', 'Do you also want to delete them from the Archived Workflows database?');
-                    break;
-                }
-            }
-        }
+                    performActionOnSelectedWorkflows(action.title, action.action, deleteArchived);
 
+                    props.clearSelection();
+                    notifications.show({
+                        content: `Performed '${action.title}' on selected workflows.`,
+                        type: NotificationType.Success
+                    });
+                },
+                disabled: () => false
+            } as WorkflowsOperation;
+        });
+    }, [props.selectedWorkflows]);
+
+    function performActionOnSelectedWorkflows(title: string, action: WorkflowOperationAction, deleteArchived: boolean): Promise<any> {
         const promises: Promise<any>[] = [];
-        this.props.selectedWorkflows.forEach((wf: Workflow) => {
+        props.selectedWorkflows.forEach((wf: Workflow) => {
             if (title === 'DELETE') {
                 // The ones without archivalStatus label or with 'Archived' labels are the live workflows.
                 if (isWorkflowInCluster(wf)) {
                     promises.push(
                         services.workflows.delete(wf.metadata.name, wf.metadata.namespace).catch(reason =>
-                            ctx.notifications.show({
+                            notifications.show({
                                 content: `Unable to delete workflow ${wf.metadata.name} in the cluster: ${reason.toString()}`,
                                 type: NotificationType.Error
                             })
@@ -79,7 +85,7 @@ export class WorkflowsToolbar extends React.Component<WorkflowsToolbarProps, {}>
                 if (deleteArchived && isArchivedWorkflow(wf)) {
                     promises.push(
                         services.workflows.deleteArchived(wf.metadata.uid, wf.metadata.namespace).catch(reason =>
-                            ctx.notifications.show({
+                            notifications.show({
                                 content: `Unable to delete workflow ${wf.metadata.name} in database: ${reason.toString()}`,
                                 type: NotificationType.Error
                             })
@@ -89,8 +95,7 @@ export class WorkflowsToolbar extends React.Component<WorkflowsToolbarProps, {}>
             } else {
                 promises.push(
                     action(wf).catch(reason => {
-                        this.props.loadWorkflows();
-                        ctx.notifications.show({
+                        notifications.show({
                             content: `Unable to ${title} workflow: ${reason.content.toString()}`,
                             type: NotificationType.Error
                         });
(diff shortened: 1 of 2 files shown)


ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
 
-    private async performActionOnSelectedWorkflows(ctx: ContextApis, title: string, action: WorkflowOperationAction): Promise<any> {
-        const confirmed = await ctx.popup.confirm('Confirm', `Are you sure you want to ${title.toLowerCase()} all selected workflows?`);
-        if (!confirmed) {
-            return Promise.resolve(false);
-        }
+                    // check for delete from archived workflows
+                    let deleteArchived = false;
+                    if (action.title === 'DELETE') {
+                        // check for delete workflows from archived workflows
```
agilgur5, line 48:
```suggestion
                        // check if there are archived workflows to delete
```

- be a bit more specific and less redundant

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
+                    if (!confirmed) {
+                        return;
+                    }
 
-    private async performActionOnSelectedWorkflows(ctx: ContextApis, title: string, action: WorkflowOperationAction): Promise<any> {
-        const confirmed = await ctx.popup.confirm('Confirm', `Are you sure you want to ${title.toLowerCase()} all selected workflows?`);
-        if (!confirmed) {
-            return Promise.resolve(false);
-        }
+                    // check for delete from archived workflows
```
agilgur5, line 45:
```suggestion
```

this is duplicative of the below comment

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
-    }
+        return Object.keys(actions).map((actionName: WorkflowOperationName) => {
+            const action = actions[actionName];
+            return {
+                title: action.title,
+                iconClassName: action.iconClassName,
+                isDisabled: disabled[actionName],
+                action,
+                workflowsAction: async () => {
+                    // check for action
```
agilgur5, line 39:
```suggestion
```

this seems redundant to me since the word "confirm" is already used multiple times in the code and in the message

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
-        let deleteArchived = false;
-        if (title === 'DELETE') {
-            for (const entry of this.props.selectedWorkflows) {
-                if (isArchivedWorkflow(entry[1])) {
-                    deleteArchived = await ctx.popup.confirm('Confirm', 'Do you also want to delete them from the Archived Workflows database?');
-                    break;
-                }
-            }
-        }
+                    performActionOnSelectedWorkflows(action.title, action.action, deleteArchived);
```
agilgur5, line 57:
```suggestion
                    await performActionOnSelectedWorkflows(action.title, action.action, deleteArchived);
```

the previous behavior waited for this to complete first

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
+                        content: `Performed '${action.title}' on selected workflows.`,
+                        type: NotificationType.Success
+                    });
+                },
+                disabled: () => false
+            } as WorkflowsOperation;
+        });
+    }, [props.selectedWorkflows]);
 
+    function performActionOnSelectedWorkflows(title: string, action: WorkflowOperationAction, deleteArchived: boolean): Promise<any> {
```
agilgur5, line 70:
```suggestion
    async function performActionOnSelectedWorkflows(title: string, action: WorkflowOperationAction, deleteArchived: boolean): Promise<any> {
```

to hopefully avoid future misses on awaiting this like the above, we could be more explicit that this is an async function (even if it only uses promise chains)

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -101,48 +106,28 @@ export class WorkflowsToolbar extends React.Component<WorkflowsToolbarProps, {}>
...
-                    if (!confirmed) {
-                        return;
-                    }
-
-                    this.props.clearSelection();
-                    ctx.notifications.show({
-                        content: `Performed '${action.title}' on selected workflows.`,
-                        type: NotificationType.Success
-                    });
-                    this.props.loadWorkflows();
```
agilgur5, line 126: so I actually don't think we should remove `loadWorkflows` at this point if we plan to replace it with an optimistic update per #12008. this specific call would be replaced with another call that takes the action as an argument

it would make the future change to an optimistic update more straightforward if we left this in

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
 
 interface WorkflowsToolbarProps {
     selectedWorkflows: Map<string, Workflow>;
-    loadWorkflows: () => void;
-    isDisabled: Actions.OperationDisabled;
+    actionsIsDisabled: Actions.OperationDisabled;
```
agilgur5, line 14:
```suggestion
    disabledActions: Actions.OperationDisabled;
```

that probably makes a bit more sense since it's a mapping

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
 
-    private getNumberSelected(): number {
-        return this.props.selectedWorkflows.size;
-    }
+        return Object.keys(actions).map((actionName: WorkflowOperationName) => {
+            const action = actions[actionName];
+            return {
+                title: action.title,
+                iconClassName: action.iconClassName,
+                isDisabled: disabled[actionName],
```
agilgur5, line 36:
```suggestion
                isDisabled: props.disabledActions[actionName],
```
we can just in-line this as it's only used here. no renames etc needed then

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
     clearSelection: () => void;
 }
 
-interface WorkflowGroupAction extends WorkflowOperation {
-    groupIsDisabled: boolean;
-    className: string;
-    groupAction: () => Promise<any>;
+interface WorkflowsOperation extends WorkflowOperation {
+    isDisabled: boolean;
+    workflowsAction: () => Promise<any>;
```
agilgur5, line 20:
```suggestion
    action: () => Promise<any>;
```

I think we can just overwrite `action`, since it's not used otherwise. the interface is already plural Workflows, so I don't think we need to repeat it again inside of it

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
```diff
@@ -10,66 +11,71 @@ require('./workflows-toolbar.scss');
...
-                            {this.getNumberSelected() === 0 ? 'No' : this.getNumberSelected()}
-                            &nbsp;workflow{this.getNumberSelected() === 1 ? '' : 's'} selected
-                        </div>
-                        <div className='workflows-toolbar__actions'>{this.renderActions(ctx)}</div>
-                    </div>
-                )}
-            </Consumer>
-        );
-    }
+    const groupAction = useMemo<WorkflowsOperation[]>(() => {
```
agilgur5, line 27:
```suggestion
    const groupActions = useMemo<WorkflowsOperation[]>(() => {
```

this should be plural since there are multiple actions in here

agilgur5 (commented):
Thanks for diving into this and your detailed explanations!

The renames and re-ordering took a bit for me to wrap my head around, so I waited a bit to finish the review until I was fully recovered 😅 . It definitely made more sense when I was feeling better! And the namings were definitely confusing before your changes too

I have a few suggestions in-line below:

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
agilgur5: or maybe better, we could rename it to `operations` to match the type names
```suggestion
    const operations = useMemo<WorkflowsOperation[]>(() => {
```

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
juijeong8324: I think it seems redundant and unnecessary, too!

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
juijeong8324: I think it's better to change the name to 'operations' to match the type name. Actually, I was confused that why the type was `Operation` but the name was `groupAction`.

ui/src/app/workflows/components/workflows-toolbar/workflows-toolbar.tsx
juijeong8324: Umm Then, Should we rename the name of`action` in below the code?  

- Before

```typescript
        <div className={`workflows-toolbar ${numberSelected === 0 ? 'hidden' : ''}`}>
            <div className='workflows-toolbar__count'>
                {numberSelected === 0 ? 'No' : numberSelected}
                &nbsp;workflow{numberSelected === 1 ? '' : 's'} selected
            </div>
            <div className='workflows-toolbar__actions'>
                {operations.map(action => {
                    return (
                        <button
                            key={action.title}
                            onClick={() => {
                                action.workflowsAction();
                            }}
                            className={`workflows-toolbar__actions--${action.title} workflows-toolbar__actions--action`}
                            disabled={numberSelected === 0 || action.isDisabled}>
                            <i className={action.iconClassName} />
                            &nbsp;{action.title}
                        </button>
                    );
                })}
            </div>
        </div>
```

- After

```typescript
<div className={`workflows-toolbar ${numberSelected === 0 ? 'hidden' : ''}`}>
            <div className='workflows-toolbar__count'>
                {numberSelected === 0 ? 'No' : numberSelected}
                &nbsp;workflow{numberSelected === 1 ? '' : 's'} selected
            </div>
            <div className='workflows-toolbar__actions'>
                {operations.map(operation => {
                    return (
                        <button
                            key={operation.title}
                            onClick={() => {
                                operation.action();
                            }}
                            className={`workflows-toolbar__actions--${operation.title} workflows-toolbar__actions--action`}
                            disabled={numberSelected === 0 || operation.isDisabled}>
                            <i className={operation.iconClassName} />
                            &nbsp;{operation.title}
                        </button>
                    );
                })}
            </div>
        </div>
```

agilgur5: Yup, that would be required

juijeong8324: Oh I missed it, thank you!

juijeong8324: > Note that with the `onClick` optimization in my [above comment](https://github.com/argoproj/argo-workflows/pull/12046#discussion_r1372339282), those lines become `onClick={operation.action}`

Oh I missed it, thank you!

juijeong8324:
> The renames and re-ordering took a bit for me to wrap my head around, so I waited a bit to finish the review until I was fully recovered 😅 . It definitely made more sense when I was feeling better! And the namings were definitely confusing before your changes too
> 

Thanks for detailed advice!! 
The most difficult in refactoring is obviously naming value and parameter... 😂

agilgur5:
> The most difficult in refactoring is obviously naming value and parameter... 😂

"There are only [two hard things](https://martinfowler.com/bliki/TwoHardThings.html) in Computer Science: cache invalidation and naming things." -- Phil Karlton

naming is indeed a hard thing, no joke 😉

terrytangyuan (approved):

Issue #12059, opened 2023-10-21
UI: Code Split larger and/or less-used paths for smaller initial bundle


agilgur5:
# Summary

Instead of having all the UI pages in one bundle, we can code split out certain pages that are either quite large (e.g. have a lot of code or deps) or less-used.

## Use Cases

- To load the UI faster by loading the most used pages first and then asynchronously loading other pages when needed. For a faster TTFP (time to first paint), TTI (time to interactive), etc. Use less networking on average as well.
- For slower devices such as those described in #11970

### Implementation Details

Nowadays, React supports loading components async with the built-in [`lazy` function](https://legacy.reactjs.org/docs/code-splitting.html), which you normally pass a dynamic import into. Webpack will [automatically code-split dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports).

Two components that I think might be ripe for code-splitting are:
- the Swagger API docs (see also #12058 and #11970) that are used pretty infrequently, contain a whole `swagger-ui` in it [and the Swagger file itself](https://github.com/argoproj/argo-workflows/blob/03a6168da30717eba3a7e585832841350724e90d/ui/src/app/webpack.config.js#L79)
- perhaps the code editing features that rely on `monaco-editor`

Also, code-splitting out the Argo Events pages of the UI could be useful as well as not everyone uses Workflows and Events together.

### Trade-offs

Note that code-splitting is not without its trade-offs as more bundles can have downsides. In particular if those bundles are very small. In this case I think the pros outweigh the cons for certain pages, but may need some more testing for other pages.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
Ran [`webpack-bundle-analyzer`](https://github.com/webpack-contrib/webpack-bundle-analyzer) (which I had wanted to do for a while, figured now's a good time if I wrote this issue already) and here are the results:

![Screenshot 2023-10-21 at 12 24 19 PM](https://github.com/argoproj/argo-workflows/assets/4970083/92ea8f24-5a85-4cc6-8e96-32b3f535adab)

Observations:
1. `monaco-editor` is gigantic and takes up almost half of the bundle. I'm pretty sure the workers are from Monaco as well
    - I expected it to be large per above, but did not quite expect half of the bundle 😬
[...]

PR #12061, opened 2023-10-21
build(ui): code-split `ApiDocs` and `Reports` components

agilgur5:
Partial fix for #12059
  - still more code splitting that can be done

### Motivation

- `swagger-ui-react` was one of the biggest dependencies and only used on the infrequently used `/apiDocs` page
  - and it was the sole importer of large dependencies like `dompurify`, `immutable`, `redux`, etc
  - shave off ~2MB / 826KB minified / 250KB gzipped from the main bundle
- `charts.js` was also one of the bigger deps and similarly only used on the infrequently used `/reports` page
  - shave off ~525KB / 190KB minified / 60KB gzipped from the main bundle

### Modifications

- use `React.lazy` to lazy load them with dynamic imports
  - and `React.Suspense` to add a fallback etc
  - see https://react.dev/reference/react/lazy and https://webpack.js.org/guides/code-splitting/#dynamic-imports
    - also https://v5.reactrouter.com/web/guides/code-splitting, though it's a bit dated now (`lazy` and `Suspense` are built-in now, no need to use `loadable-components`)

- make some small upgrades to support dynamic importing
  - per in-line code comments, `ES2020`+ `module` format is required for dynamic imports (otherwise they get transpiled to regular, synchronous `require`s) (c.f. https://github.com/Microsoft/TypeScript/issues/16675, https://github.com/webpack/webpack/issues/5703#issuecomment-357512412)
    - note that Argo CD [already uses this](https://github.com/argoproj/argo-cd/blob/973565e194c094d163d2fd0873364d0e36951806/ui/src/app/tsconfig.json#L6)
    - this new module format did require updating some legacy code that used the [`import x = require('x')` syntax](https://www.typescriptlang.org/docs/handbook/modules/reference.html#export--and-import--require)
      - update it to use what are nowadays just regular ESM imports
      - remaining `require` usage is all for CSS/SCSS and a handful of packages that have no types (or at least that we haven't imported an `@types/` package for)
        - those should be converted in the future too, but nbd right now
      - there are some new TS/Webpack warnings that pop up as a result of this new module format
        - they're warnings, so not causing any errors just yet -- they should be worked on in the future though
        - seemingly all of these are for exports of types -- they should be moved to the [newer `export type` syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export)
          - problematically, updating to `export type` seems to cause errors in legacy `tslint`... so we may be blocked until that is swapped out for `eslint`
  - add a `default` `export` to `Reports` because this is (still) currently required for lazy loading

- minor stylistic changes while at it
  - consts assigned to anonymous functions -> named functions for better tracing
  - new line between external imports and internal imports for consistency

### Verification

See screenshots bundle analysis after this code splitting was added below:
Screenshot 2023-10-21 at 3 14 45 PM -- after code-split apiDocs and Reports
Screenshot 2023-10-21 at 3 10 22 PM -- after code-split apiDocs and Reports

Tested locally and worked without issues.
I barely even noticed the lazy load as it took less than a second for it to appear, and I was specifically paying attention for it. Results may vary ofc, esp since locally the network is ofc much faster, but per #12059 I still think this is well worth the tiny trade-off given how rarely used these pages are. 

### Future Work

There's still more to code-split per #12059
- these two were particularly quick as they were their own pages and isolated to a single module
- `monaco-editor` would be good to do next; I didn't do it as part of this because it requires more granular changes to two different components/modules -- need to take some more time with that

### DIFF
--- a/ui/src/app/apidocs/components/apiDocs.tsx
+++ b/ui/src/app/apidocs/components/apiDocs.tsx
@@ -1,8 +1,7 @@
 import {Page} from 'argo-ui';
 import * as React from 'react';
-import SwaggerUI from 'swagger-ui-react';
-import 'swagger-ui-react/swagger-ui.css';
 import {uiUrl} from '../../shared/base';
+import {Loading} from '../../shared/components/loading';
 import {useCollectEvent} from '../../shared/components/use-collect-event';
 
 export const ApiDocs = () => {
@@ -10,8 +9,22 @@ export const ApiDocs = () => {
     return (
         <Page title='Swagger'>
             <div className='argo-container'>
-                <SwaggerUI url={uiUrl('assets/openapi-spec/swagger.json')} defaultModelExpandDepth={0} />
+                <SuspenseSwaggerUI url={uiUrl('assets/openapi-spec/swagger.json')} defaultModelExpandDepth={0} />
             </div>
         </Page>
     );
 };
+
+// lazy load SwaggerUI as it is infrequently used and imports very large components (which can be split into a separate bundle)
+const LazySwaggerUI = React.lazy(() => {
+    import('swagger-ui-react/swagger-ui.css');
+    return import('swagger-ui-react');
+});
+
+function SuspenseSwaggerUI(props: any) {
+    return (
+        <React.Suspense fallback={<Loading />}>
+            <LazySwaggerUI {...props} />
+        </React.Suspense>
+    );
+}
--- a/ui/src/app/cron-workflows/components/pretty-schedule.tsx
+++ b/ui/src/app/cron-workflows/components/pretty-schedule.tsx
@@ -1,7 +1,7 @@
-import React = require('react');
-import {WarningIcon} from '../../shared/components/fa-icons';
+import x from 'cronstrue';
+import * as React from 'react';
 
-const x = require('cronstrue');
+import {WarningIcon} from '../../shared/components/fa-icons';
 
 /*
     https://github.com/bradymholt/cRonstrue
@@ -12,7 +12,7 @@ const x = require('cronstrue');
     sometime it'll not work as expected. Therefore, we must let the user know about this.
  */
 
-export const PrettySchedule = ({schedule}: {schedule: string}) => {
+export function PrettySchedule({schedule}: {schedule: string}) {
     try {
         if (schedule.split(' ').length >= 6) {
             throw new Error('cron schedules must consist of 5 values only');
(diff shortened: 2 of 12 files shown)


terrytangyuan (commented):
> still more code splitting that can be done

Maybe do it in one PR?

agilgur5:
The `monaco-editor` ones are a good bit more complex, so idk that I'll get to them soon. I put the PR out so we can get some improvements sooner rather than later

terrytangyuan:
Same comment for this PR https://github.com/argoproj/argo-workflows/pull/12062#issuecomment-1773955081

agilgur5:
> Same comment for this PR [#12062 (comment)](https://github.com/argoproj/argo-workflows/pull/12062#issuecomment-1773955081)

I would not call these changes "trivial" in any definition of that word... 

There is no benefit to a large PR, but smaller PRs and commits are well regarded as best practice. Notably, I've even shown you some examples before in this very codebase where large PRs made things incredibly difficult to trace (and harder to review, among other downsides). I don't think it's a good idea to intentionally make things more difficult, and without a good reason at that.

terrytangyuan:
Sorry I didn't mean to apply exactly the same comment from the other PR. I only meant to say the first part of that comment, "I would appreciate a combined PR". While you are already at this, it might be a good idea to apply the same optimization for more components. I definitely appreciate your time and efforts spent on this.

agilgur5:
> While you are already at this, it might be a good idea to apply the same optimization for more components.

If it were as straightforward as these, I would have done so already 😕 As I mentioned above, given that there are complexities in other components, idk that I will get to them timely, as I mentioned above.
Or, in agile terminology, there are more "unknowns" others (in the code, or in the testing to determine whether a code-split is worthwhile, etc). I very intentionally did the two with the least unknowns here to first get code-splitting working correctly at least and the build system set up properly for it. These are also the most obvious optimizations given how infrequently these pages are used and how isolated their deps are (other components are used across multiple pages or used quite frequently, so there's a bit more care to be taken with them and more testing)

agilgur5:
You also haven't really answered the question of "why wait?".
I don't see any pros to holding off on merging this for the sole purpose of waiting for _even more_ optimizations, only cons. Any optimization is better than the current status

terrytangyuan:
Just waiting for other UI experts to review

toyamagu-2021 (approved):
LGTM. I respect the contributions of both of you.

terrytangyuan (approved):

PR #12064, opened 2023-10-23
ci: Remove coverage report for Windows tests

terrytangyuan:
This causes master branch to fail. We already uploaded coverage in non-Windows unit tests builds which is more correct since we skip certain tests on Windows.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -140,10 +140,6 @@ jobs:
       - run: go test -p 1  -covermode=atomic -coverprofile='coverage.out' $(go list ./... | select-string -Pattern 'github.com/argoproj/argo-workflows/v3/workflow/controller' , 'github.com/argoproj/argo-workflows/v3/server' -NotMatch)
         env:
           KUBECONFIG: /dev/null
-      # engineers just ignore this in PRs, so lets not even run it
-      - name: Upload coverage report
-        if: github.ref == 'refs/heads/master'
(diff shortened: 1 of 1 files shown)


juliev0 (approved):

agilgur5:
Oh the upload _itself_ fails. Apparently the upload shell script [has invalid syntax on Windows](https://github.com/argoproj/argo-workflows/actions/runs/6601833360/job/17933284385) (TIL, I haven't developed natively on a Windows machine in like a decade)

The [Codecov Bash Uploader](https://about.codecov.io/blog/introducing-codecovs-new-uploader/) is actually deprecated, so if that is the case, it would be a good time to replace it with the new cross-platform uploader or just [the GH Action](https://github.com/marketplace/actions/codecov)

> which is more correct since we skip certain tests on Windows

For clarity, I asked for the coverage report in https://github.com/argoproj/argo-workflows/pull/12011#discussion_r1361155317 because we do have some Windows-specific logic. Most coverage providers, [including Codecov](https://docs.codecov.com/docs/merging-reports), will automatically merge reports and create a union to handle scenarios like these.

agilgur5:
> The Codecov Bash Uploader is actually [deprecated](https://about.codecov.io/blog/introducing-codecovs-new-uploader/), so if that is the case, it would be a good time to replace it with the new cross-platform uploader or just [the GH Action](https://github.com/marketplace/actions/codecov)

Resolved this and added back Windows coverage in #12618

PR #12067, opened 2023-10-23
build: Add support for RISC-V

chazapis:
This PR adds RISC-V support, by generating a CLI binary for RISC-V and container images.

### Motivation

The work is part of an ongoing effort to run [K3s on RISC-V](https://github.com/k3s-io/k3s/pull/7778). The main functionality is there, so we are now moving to higher-level services and Argo Workflows is first on the list. We expect Argo Workflows to enable us to seamlessly run workflows in a hybrid hardware environment with amd64, arm64, and riscv64 hosts. 

### Modifications

There are no code changes in this PR, just small additions and modifications in the `Makefile` and release scripts.

A binary for `riscv64` is created by adding the architecture to the `Makefile`.

The container image generation is a bit trickier, as the base images used do not currently support RISC-V. Alpine supports RISC-V in the "edge" branch and distroless has [deferred releasing RISC-V images until the Debian unstable branch is moved into stable](https://github.com/GoogleContainerTools/distroless/issues/1269) (estimated in 2025). For this reason, I turned the base image names into arguments in the `Dockerfile` and use different ones for RISC-V in `release.yaml`.

### Verification

I built container images by running the script in `release.yaml` by hand. They can be found at my organization's [Docker Hub account](https://hub.docker.com/u/carvicsforth). I then modified `install.yaml` to use these container images and tested Argo Workflows in a QEMU setup with a [RISC-V build of K3s](https://github.com/k3s-io/k3s/pull/7778). The installation manifest, as well as test workflow can be found [here](https://github.com/CARV-ICS-FORTH/kubernetes-riscv64/tree/main/argo-workflows).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,15 @@
 ARG GIT_COMMIT=unknown
 ARG GIT_TAG=unknown
 ARG GIT_TREE_STATE=unknown
+ARG BUILDER_IMAGE=golang:1.21-alpine3.18
+ARG UI_IMAGE=node:20-alpine
+ARG DISTROLESS_IMAGE=gcr.io/distroless/static
 
-FROM golang:1.21-alpine3.18 as builder
+FROM $BUILDER_IMAGE as builder
 
-RUN apk update && apk add --no-cache \
+RUN apk update && \
+    (command -v go >/dev/null 2>&1 || apk add --no-cache go) && \
+    apk add --no-cache \
     git \
     make \
     ca-certificates \
@@ -24,9 +29,13 @@ COPY . .
 
 ####################################################################################################
 
-FROM node:20-alpine as argo-ui
+FROM $UI_IMAGE as argo-ui
 
-RUN apk update && apk add --no-cache git
+RUN apk update && \
+    (command -v node >/dev/null 2>&1 || apk add --no-cache nodejs-current) && \
+    (command -v yarn >/dev/null 2>&1 || apk add --no-cache yarn) && \
+    apk add --no-cache \
+    git
 
 COPY ui/package.json ui/yarn.lock ui/
 
@@ -76,7 +85,7 @@ RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache
 
 ####################################################################################################
 
-FROM gcr.io/distroless/static as argoexec
+FROM $DISTROLESS_IMAGE as argoexec
 
 COPY --from=argoexec-build /go/src/github.com/argoproj/argo-workflows/dist/argoexec /bin/
 COPY --from=argoexec-build /etc/mime.types /etc/mime.types
@@ -87,7 +96,7 @@ ENTRYPOINT [ "argoexec" ]
 
 ####################################################################################################
 
-FROM gcr.io/distroless/static as workflow-controller
+FROM $DISTROLESS_IMAGE as workflow-controller
 
 USER 8737
 
@@ -99,7 +108,7 @@ ENTRYPOINT [ "workflow-controller" ]
 
 ####################################################################################################
 
-FROM gcr.io/distroless/static as argocli
+FROM $DISTROLESS_IMAGE as argocli
 
 USER 8737
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        platform: [ linux/amd64, linux/arm64 ]
+        platform: [ linux/amd64, linux/arm64, linux/riscv64 ]
         target: [ workflow-controller, argocli, argoexec ]
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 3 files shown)


server/static/files.go
terrytangyuan, line 1: Is this change necessary?

terrytangyuan:
I don't know much about RISC-V so I'll defer the review to others who are more experienced

server/static/files.go
chazapis: You are right. This somehow slipped into the commit. I reverted the changes to the file.

agilgur5 (commented):
IMO, I don't think we should add support for new architectures that do not have significant user requests. In this case, this seems pre-emptive to user requests (correct me if I'm wrong).

Argo CD supports [a few different archs](https://github.com/argoproj/argo-cd/blob/bc5fb811d660d8f1b4d47b81035147d5e9409882/.github/workflows/release.yaml#L27), but that doesn't include RISC-V. Those different archs do also slow down the build considerably for some rare use-cases.

Adding support for a new arch means a new maintenance burden for that arch. If there are no user requests for that, that maintenance burden seems not worthwhile. Given the unstable status of several upstream projects as you mentioned, that sounds even more maintenance heavy as a result

Dockerfile
```diff
@@ -2,10 +2,15 @@
...
+ARG BUILDER_IMAGE=golang:1.21-alpine3.18
+ARG UI_IMAGE=node:20-alpine
+ARG DISTROLESS_IMAGE=gcr.io/distroless/static
 
-FROM golang:1.21-alpine3.18 as builder
+FROM $BUILDER_IMAGE as builder
 
-RUN apk update && apk add --no-cache \
+RUN apk update && \
+    (command -v go >/dev/null 2>&1 || apk add --no-cache go) && \
```
agilgur5, line 12: we should add versions if we want this to match the regular alpine images.

this also feels very hacky to have different base images, but it sounds like there's no pre-built Alpine image with Go and Node for RISC-V?

Dockerfile
```diff
@@ -24,9 +29,13 @@ COPY . .
 
 ####################################################################################################
 
-FROM node:20-alpine as argo-ui
+FROM $UI_IMAGE as argo-ui
 
-RUN apk update && apk add --no-cache git
+RUN apk update && \
+    (command -v node >/dev/null 2>&1 || apk add --no-cache nodejs-current) && \
+    (command -v yarn >/dev/null 2>&1 || apk add --no-cache yarn) && \
```
agilgur5, line 36: same as above re: versioning

chazapis:
> IMO, I don't think we should add support for new architectures that do not have significant user requests. In this case, this seems pre-emptive to user requests (correct me if I'm wrong).
> 
> Argo CD supports [a few different archs](https://github.com/argoproj/argo-cd/blob/bc5fb811d660d8f1b4d47b81035147d5e9409882/.github/workflows/release.yaml#L27), but that doesn't currently include RISC-V. Those different archs do also slow down the build considerably for some rare use-cases.
> 
> Adding support for a new arch means a new maintenance burden for that arch. If there are no user requests for that, that maintenance burden seems not worthwhile. Given the unstable status of several upstream projects as you mentioned, that sounds even more maintenance heavy as a result

@agilgur5, I agree.

However, I want to stress out that even though the RISC-V community is still small, it is constantly growing. You may not have user requests now, but RISC-V users currently just run what is available (or recompile). If Argo (and Kubernetes/K3s) were already available, I am sure people would find good use for them. A whole range of RISC-V hardware has been announced. We envision these devices to join a Kubernetes control plane and run partial workflows via Argo.

Also, at this point, I think it would be perfectly acceptable to publish a build marked as experimental/untested. In any case, let me know if there is any way I can help.

agilgur5:
As an example of previous work, https://github.com/nodejs/unofficial-builds/ creates experimental builds for rarer architectures like RISC-V. We could potentially have something similar in `argoproj-labs`, which would be very explicitly experimental

juliev0:
Closing stale PRs

chazapis:
@juliev0, this is still relevant. CLI support has been added with #12977, but we need this to run Argo on RISC-V machines.

agilgur5:
@chazapis I'll refer back to my [earlier](https://github.com/argoproj/argo-workflows/pull/12067#pullrequestreview-1697949995) [comments](https://github.com/argoproj/argo-workflows/pull/12067#issuecomment-1783603520), particularly the `unofficial-builds` one, which was not responded to. Not to mention my in-line comments as well.

AllardKrings:
Hello,

as I am using a Kubernetes cluster on RISCV hardware I am now looking for a workflow tool.

I would like to use Argo Workflows on RISCV as soon as it is available.

agilgur5:
The equivalent was said [above](https://github.com/argoproj/argo-workflows/pull/12067#issuecomment-1780726875), but "build and maintain it pre-emptive to an actual need" is not a sustainable maintenance strategy (following that could make infinite permutations of features), and Argo already has [sustainability issues](https://github.com/argoproj/argo-workflows/blob/974d819/community/sustainability_effort.md).

Also neither high skill nor high motivation is needed for an [`unofficial-builds` repo](https://github.com/argoproj/argo-workflows/pull/12067#issuecomment-1783603520), yet no community member [nor corporate entity](https://github.com/argoproj/argo-workflows/issues/12449) has made one, so that argument is also a strawman.

szpeter80:
I understand your point, and thanks for the clarification.

As currently procurable RiscV boards just getting over the RPi3 performance level, things might change in the future as drivers and Kubernetes ecosystem support matures.

agilgur5:
Yea I'd personally love to see an open source ISA like RISC-V gain more traction and adoption.
Just wouldn't make sense for some projects like Argo to support it yet as the maintenance would not be worthwhile / sustainable.

PR #12068, opened 2023-10-23
fix: Revert #11761 to avoid argo-server performance issue

terrytangyuan:
This reverts #11761. See https://github.com/argoproj/argo-workflows/issues/12025 for context.

### DIFF
--- a/server/workflow/workflow_server.go
+++ b/server/workflow/workflow_server.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"sort"
-	"strconv"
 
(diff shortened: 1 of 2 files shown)


juliev0:
Just want to make sure - can you confirm that there are no changes that were made after this change that were dependent on it?

terrytangyuan:
> Just want to make sure - can you confirm that there are no changes that were made after this change that were dependent on it?

Yes that is the last change to https://github.com/argoproj/argo-workflows/blob/master/server/workflow/workflow_server.go

juliev0 (approved):

PR #12069, opened 2023-10-23
ci: Speed up tests on Windows build

terrytangyuan:
The build has been timed out in master branch due to `-p 1`. Removing it to maximize performance ([verified that the build is successful in a test run](https://github.com/argoproj/argo-workflows/actions/runs/6617999161)).

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -5,7 +5,6 @@ on:
       - "master"
       - "release-*"
       - "!release-2.8"
-      - "test-windows-build"
   pull_request:
     branches:
       - "master"
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -5,6 +5,7 @@ on:
       - "master"
       - "release-*"
       - "!release-2.8"
+      - "test-windows-build"
   pull_request:
     branches:
       - "master"
@@ -137,7 +138,7 @@ jobs:
           go-version: "1.21"
           cache: true
       # windows run does not use makefile target because it does a lot more than just testing and is not cross-platform compatible
-      - run: go test -p 1  -covermode=atomic -coverprofile='coverage.out' $(go list ./... | select-string -Pattern 'github.com/argoproj/argo-workflows/v3/workflow/controller' , 'github.com/argoproj/argo-workflows/v3/server' -NotMatch)
+      - run: go test -covermode=atomic -coverprofile='coverage.out' $(go list ./... | select-string -Pattern 'github.com/argoproj/argo-workflows/v3/workflow/controller' , 'github.com/argoproj/argo-workflows/v3/server' -NotMatch)
         env:
           KUBECONFIG: /dev/null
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -137,7 +137,7 @@ jobs:
           go-version: "1.21"
           cache: true
       # windows run does not use makefile target because it does a lot more than just testing and is not cross-platform compatible
-      - run: go test -p 1  -covermode=atomic -coverprofile='coverage.out' $(go list ./... | select-string -Pattern 'github.com/argoproj/argo-workflows/v3/workflow/controller' , 'github.com/argoproj/argo-workflows/v3/server' -NotMatch)
+      - run: go test -covermode=atomic -coverprofile='coverage.out' $(go list ./... | select-string -Pattern 'github.com/argoproj/argo-workflows/v3/workflow/controller' , 'github.com/argoproj/argo-workflows/v3/server' -NotMatch)
```
agilgur5, line 140:
```suggestion
      - run: go test -p 20 -covermode=atomic -coverprofile='coverage.out' $(go list ./... | select-string -Pattern 'github.com/argoproj/argo-workflows/v3/workflow/controller' , 'github.com/argoproj/argo-workflows/v3/server' -NotMatch)
```

The Linux ones use `20`, so I would think we should match them, no?

agilgur5: Also, can't comment on unchanged lines, but should we decrease `timeout-minutes` now? May want to test more to get an average though as GH runs are not all equivalent

agilgur5:
So I mentioned this in https://github.com/argoproj/argo-workflows/pull/12011#discussion_r1364354713 -- @mweibel said there that there were some race conditions when parallelizing? I wonder if this CI run just happened to not hit them or if that only happened locally or something

.github/workflows/ci-build.yaml
terrytangyuan: Let's leave it as is and observe.

terrytangyuan:
> So I mentioned this in [#12011 (comment)](https://github.com/argoproj/argo-workflows/pull/12011#discussion_r1364354713) -- @mweibel said there that there were some race conditions when parallelizing? I wonder if this CI run just happened to not hit them or if that only happened locally or something

Maybe. I tried several times but haven't seen anything yet

agilgur5 (approved):
LGTM if Windows race conditions aren't being hit in CI

mweibel:
> So I mentioned this in [#12011 (comment)](https://github.com/argoproj/argo-workflows/pull/12011#discussion_r1364354713) -- @mweibel said there that there were some race conditions when parallelizing? I wonder if this CI run just happened to not hit them or if that only happened locally or something (since we _do_ have the opposite sometimes, races only hit on CI and not locally)

I ran into them locally. Not sure why it doesn't seem to happen on CI (I think I never tested on CI with p 20). Maybe it's related to windows versions since locally I run on windows 10 and ltsc2022 is IIRC on a similar base as windows 11.

PR #12071, opened 2023-10-24
fix(test): windows tests fixes. Fixes #11994

### LINKED ISSUE #11994: Windows tests not run automatically
### Pre-requisites

While working on reproducing #11810 and contributing the fix in #11993 I noticed a lot of failing tests for windows. Especially also the bug in question would've triggered a failed run when CI would run windows tests automatically. 
However it seems that this is not done currently. 

Running the tests on a windows machine, I get quite a lot of errors, see attached [logs.txt](https://github.com/argoproj/argo-workflows/files/12892852/logs.txt).

What is the reason for not running them automatically? 
Was it always like this or did that change recently? 

What can be done (besides providing windows fixes for those failing tests) to improve this in the future?

### Version

latest

```YAML
not applicable
```

```text
not applicable
```

```text
not applicable
```

mweibel:
Fixes #11994 (or rather adjusts a few things fixed in #12011)

### Motivation

Adjusts changes based on review feedback.

### DIFF
--- a/cmd/argo/lint/formatter_pretty_test.go
+++ b/cmd/argo/lint/formatter_pretty_test.go
@@ -1,9 +1,8 @@
-//go:build !windows
-
 package lint
 
 import (
 	"fmt"
+	"runtime"
 	"testing"
 
 	"github.com/TwiN/go-color"
@@ -41,6 +40,9 @@ func TestPrettyFormat(t *testing.T) {
 			Linted: true,
 		})
 		expected := "\x1b[4mtest1\x1b[0m:\n   \x1b[31m✖\x1b[0m some error\n   \x1b[31m✖\x1b[0m some error2\n\n"
+		if runtime.GOOS == "windows" {
+			expected = "test1:\n   ✖ some error\n   ✖ some error2\n\n"
+		}
 		assert.Equal(t, expected, msg)
 	})
 
@@ -53,6 +55,9 @@ func TestPrettyFormat(t *testing.T) {
 			Linted: true,
 		})
 		expected := "\x1b[4mtest2\x1b[0m:\n   \x1b[31m✖\x1b[0m some error\n\n"
+		if runtime.GOOS == "windows" {
+			expected = "test2:\n   ✖ some error\n\n"
+		}
 		assert.Equal(t, expected, msg)
 	})
 
--- a/pkg/apis/workflow/v1alpha1/workflow_types_test.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types_test.go
@@ -1,9 +1,9 @@
-//go:build !windows
-
 package v1alpha1
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"sort"
 	"testing"
 	"time"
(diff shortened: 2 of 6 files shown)


cmd/argo/lint/formatter_pretty_test.go
```diff
@@ -1,3 +1,4 @@
+// disabled on windows because colorized output is not supported - go-color will return just the text without colors.
 //go:build !windows
```
agilgur5, line 2: Could you change these to a `t.Skip` in `TestMain` instead (as mentioned in https://github.com/argoproj/argo-workflows/pull/12011#discussion_r1369420886)?

cmd/argo/lint/formatter_pretty_test.go
```diff
@@ -1,3 +1,4 @@
+// disabled on windows because colorized output is not supported - go-color will return just the text without colors.
```
agilgur5, line 1: I might be mistaken, but I thought this file tests the output messages themselves too, which include colors. the output message part would still be relevant, no?

agilgur5 (commented):
Thanks for following up on [my review](https://github.com/argoproj/argo-workflows/pull/12011#pullrequestreview-1685827398)!

small comments in-line -- you might be working on those already since this is a draft

cmd/argo/lint/formatter_pretty_test.go
mweibel: fixed, thanks for bringing this to my attention again :)

mweibel:
it took a while - sorry about the delay. I think I incorporated the bug fixes and also noticed a few of the tests I skipped last time seem to work now. Let's see if it works in CI, too.

pkg/apis/workflow/v1alpha1/workflow_types_test.go
```diff
@@ -255,53 +255,60 @@ func TestArtifact_ValidatePath(t *testing.T) {
...
+		assert.Equal(t, filepath.Join("..", ".."), a3.Path)
 
 		a4 := Artifact{Name: "a4", Path: "../etc/passwd"}
 		err = a4.CleanPath()
 		require.NoError(t, err)
-		assert.Equal(t, "../etc/passwd", a4.Path)
+		assert.Equal(t, filepath.Join("..", "etc", "passwd"), a4.Path)
 	})
 
+	ensureRootPathSeparator := func(p string) string {
```
agilgur5, line 266: This probably makes more sense to convert to a plain named function as opposed to a variable assigned to an anonymous function since it doesn't use any scope

mweibel: I don't mind to adjust it, however a Go anon function is not really the same as UI anon functions I'd say. I simply opted for an anon function because it's only relevant to this test and there's no need to pollute the package namespace with that function, otherwise. 

but as said, I don't mind, happy to adjust it.

mweibel: adjusted.

agilgur5 (approved):
Thanks for making the adjustments and handling more tests, LGTM

mweibel:
thanks for retest - it passed 👍

agilgur5:
Right there were some flakey tests here (and in other PRs) that I wanted to make note of for an issue, had been meaning to come back to this for that 😅

PR #12089, opened 2023-10-27
docs: updated self reporting progress example to match recent conventions

mrbig:
This is a proposed update to the self reporting progress documentation at https://argoproj.github.io/argo-workflows/progress/

The current example is not recommended with version >= 3.4 because the pod metadata has been superseded by the workflowtaskresults resource.

With recent versions if you add the `workflows.argoproj.io/progress` annotation on the pod the controller will fall back to the
legacy mode. Also the annotation will revert the output of the executor and the progress does not change during the workflow run.

I have also included some hints how the progress update works, because it took me a while to figure out why it takes so long
for the progress to appear on the UI.

I have tested the original and the updated example on argo workflows version 3.5.0


docs/progress.md
```diff
@@ -56,13 +60,13 @@ spec:
...
       container:
         image: alpine:3.14
         command: [ "/bin/sh", "-c" ]
         args:
           - |
             for i in `seq 1 10`; do sleep 10; echo "$(($i*10))"'/100' > $ARGO_PROGRESS_FILE; done
 ```
+
+Please make sure you have granted `patch` privilege on `workflowtaskresults` resources for the executor
+to be able to update progress on the workflow. See https://argoproj.github.io/argo-workflows/workflow-rbac/
```
agilgur5, line 72:
```suggestion
You must set your [Workflow RBAC](workflow-rbac.md) properly for the executor to be able to update progress.
```

- relative links are used when referring to other docs
- links are preferred to be in-line and contextual, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#links)
- use more direct language, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

docs/progress.md
```diff
@@ -37,7 +37,11 @@ Reporting progress works as follows:
 
 The executor will read this file every 3s and if there was an update,
 patch the pod annotations with `workflows.argoproj.io/progress: N/M`.
-The controller picks this up and writes the progress to the appropriate Status properties.
+The controller picks this up and writes the progress to the appropriate Status properties
+every 1m.
+
+These values cane be fine tuned with the `ARGO_PROGRESS_PATCH_TICK_DURATION` and `ARGO_PROGRESS_FILE_TICK_DURATION`
+environment variables.
```
agilgur5, line 44:
```suggestion
These values can be fine tuned with the `ARGO_PROGRESS_PATCH_TICK_DURATION` and `ARGO_PROGRESS_FILE_TICK_DURATION`
[environment variables](environment-variables.md).
```

- fix typo: "cane" -> "can"
- add link to the env vars doc

docs/progress.md
```diff
@@ -37,7 +37,11 @@ Reporting progress works as follows:
...
 
 Initially the progress of a workflows' pod is always `0/1`. If you want to influence this, make sure to set an initial
 progress annotation on the pod:
```
agilgur5, line 47: here and on line 39 above, "pod annotations" are specifically mentioned. those should be altered as well if we will not be recommending those

agilgur5 (commented):
Nice catch.

There are some additional changes needed if we want to remove mentions of the pod annotation, see in-line comments

docs/progress.md
mrbig: @agilgur5  I'm new to argo workflow, so I'm might not the best person to explain the process. If I understand correctly now the executor is picking up the progress and patches that to the `workflowtaskresults` resource. Then the controller will pick up the values from there. Am I right?
If you can suggest a new explanation for lines 38-39, I would be very glad.
Thank you

agilgur5: I believe your understanding is correct. I'm not an expert in the Executor code yet either actually.

Based on your analysis in the PR description, it sounds like `workflows.argoproj.io/progress` is buggy and actually needs a bugfix for 3.4 to update/initialize `workflowtaskresults` instead of patching the `pod`.

mrbig: I'm not sure if it's a bug or just some conflict between the old and new behavior that could be avoided by proper configuration on the user side.

My experience on v3.5 showed that the executor automatically falls back to patching the pod resource if it has no permission to patch the `workflowtaskresults`, and issues a warning about this.

Meanwhile the controller looks at the pod annotations, and if finds the `workflows.argoproj.io/progress` annotation then the values found in there will override the values from `workflowtaskresults`, and issues the warning `workflow uses legacy/insecure pod patch, see https://argoproj.github.io/argo-workflows/workflow-rbac/`

By following the current documentation I easily navigated me into this situation where the executor was running the new behavior while the controller expecting the old one.

My goal with this PR is to help others to avoid the mistake I have made.

I have  updated the explanation of the process to match the "new" method, please review it.

docs/progress.md
```diff
@@ -36,8 +36,12 @@ Reporting progress works as follows:
 * format of the progress must be `N/M`
 
 The executor will read this file every 3s and if there was an update,
-patch the pod annotations with `workflows.argoproj.io/progress: N/M`.
-The controller picks this up and writes the progress to the appropriate Status properties.
+patch the matching `WorkflowTaskResults` resource with the value `N/M`.
+The controller picks this up and writes the progress to the appropriate Status properties
+every 1m.
```
juliev0, line 41: Thanks for this change. [It looks like](https://github.com/argoproj/argo-workflows/blob/main/workflow/executor/executor.go#L862) it's both patching annotations and patching `WorkflowTaskResults` with this progress information, right?

docs/progress.md
```diff
@@ -36,8 +36,12 @@ Reporting progress works as follows:
...
+
+These values can be fine tuned with the `ARGO_PROGRESS_PATCH_TICK_DURATION` and `ARGO_PROGRESS_FILE_TICK_DURATION`
+[environment variables](environment-variables.md).
```
juliev0, line 44: even better if maybe we can clarify what those values represent?

docs/progress.md
juliev0: Ahh, you're right. I looked at the code too quickly. It's a backup, as you mention in your other comment.

juliev0 (approved):

docs/progress.md
agilgur5: #13260 is potentially related

PR #12097, opened 2023-10-28
refactor(ui): replace `moment-timezone` with native `Intl`

agilgur5:
Closely related to #12059; this replaces a dep instead of code-splitting it out

### Motivation

- `moment-timezone` is a very large dependency, currently the second largest in the codebase
  - `moment` is [deprecated](https://momentjs.com/docs/#/-project-status/) and recommends using `Intl`, which is supported in all modern browsers (c.f. [MDN compatibility matrix](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#browser_compatibility))
    - `moment` was announced as deprecated in Sep 2020; we should be more careful with PRs like #10120 that introduce new dependencies
      - see also the existing dependency policy: https://argoproj.github.io/argo-workflows/CONTRIBUTING/#dependencies
  
  - shave off 775KB / 760KB minified / 36KB gzipped

    
### Modifications

- rather than code-splitting `moment-timezone`, we can just replace it with native code in the one place it's used
  - equivalently get a list of all timezones from the browser
  - equivalently convert ISO strings with TZ (thanks Canada)

- remove `useEffect` code that is no longer needed (and really should've been a `useMemo`)

### Verification

Confirmed that behavior was equivalent on two known TZs (NYC and LA):
Screenshot 2023-10-28 at 11 52 40 AM - NYC time
Screenshot 2023-10-28 at 11 52 13 AM - LA time

Bundle Analysis before/after on removal of `moment-timezone`:
Screenshot 2023-10-27 at 4 15 44 PM - with moment tz
Screenshot 2023-10-27 at 6 07 03 PM -- without moment tz


ui/yarn.lock
```diff
@@ -5825,7 +5825,7 @@ mkdirp@^0.5.1:
   dependencies:
     minimist "^1.2.6"
 
-moment-timezone@^0.5.34, moment-timezone@^0.5.43:
```
agilgur5, line 5828: this is still listed as a dep in [`argo-ui`](https://github.com/argoproj/argo-ui/blob/5ff344ac9692c14dd108468bd3c020c3c75181cb/package.json#L32) although it's actually [unused there](https://github.com/search?q=repo%3Aargoproj%2Fargo-ui%20moment-timezone&type=code), hence why it doesn't show up in the bundle anymore.

I'll have some separate PRs to clean-up more of `argo-ui`, it's uh, got a few things up with its `package.json` (like deps + peerDeps overlapping and also having different versions 😕 )

isubasinghe (approved):

agilgur5:
@terrytangyuan any chance you could merge this soon? It's already been approved and is getting merge conflicts with dependabot

terrytangyuan (approved):

ui/yarn.lock
agilgur5: Made one upstream fix here: https://github.com/argoproj/argo-ui/pull/511. I'll have a few more follow-ups to that as they may merge conflict with each other

agilgur5: Welp that still hasn't been reviewed or merged, but added https://github.com/argoproj/argo-ui/pull/534 as well (which removes the unused `moment-timezone` plus a bunch of other unused deps). Seems to actually not conflict, though I wrapped together a few of my local changes into that one.

PR #12100, opened 2023-10-29
docs: Add an example field to argo cluster-template create command. Fixes #11898

shubhamch71:
Fixes #11898 

### Motivation

Changes to support more detailed information when someone runs the argo cluster-template create

### Modifications

Added examples for some common usages of argo cluster-template create command

### Verification

RUN `argo cluster-template create --help`


agilgur5:
Looks like this was replaced by #12101.

For future reference, please re-use the same PR instead of creating a new one.

PR #12102, opened 2023-10-29
docs: fix description of argo stop --dry-run. Fixes #10782

### LINKED ISSUE #10782: Fix documentation of `argo stop --dry-run`
# Summary

While reading the argo documentation, I ran into the following confusing message:
`--dry-run                      If true, only stop the workflows that would be stopped, without stopping them.`
https://argoproj.github.io/argo-workflows/cli/argo_stop/

Same confusing message here on github: https://github.com/argoproj/argo-workflows/blob/d4a30a556a7093068624dbe16f05b381705dc6e0/cmd/argo/commands/stop.go#L74

I didn't think this merited a bug report, but it would be nice to have a more comprehensible message, such as `If true, only print the workflows that would be stopped, without stopping them.` The equivalent message for `--dry-run` on the `argo delete` page is `Do not delete the workflow, only print what would happen` which also seems like a good description.
https://github.com/argoproj/argo-workflows/blob/d4a30a556a7093068624dbe16f05b381705dc6e0/cmd/argo/commands/delete.go#L90

# Use Cases

More accurate documentation.

---

anastasiiakozlova245:
Updated the description of `--dry-run` flag for `argo stop` command.

Fixes #10782

### Motivation

Improve cli commands description for clarity.

### Modifications

Updated the description of `--dry-run` flag for `argo stop` command.

### Verification

Docs update.


agilgur5 (approved):
Thanks for the quick fix!

terrytangyuan (approved):

Issue #12010, opened 2023-10-16
Some parallell execute Steps may be stucked on running phase when retry workflow manually


jswxstw:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

<img width="380" alt="image" src="https://github.com/argoproj/argo-workflows/assets/18543532/f4d7e89a-4e9e-4f67-8458-e30aa2dd24fb">

As the picture shows, workflow has three parallell execute nodes: step-case(Failed), retry-step-case(Failed), retry-step-group-case(Succeeded).
If I retry this workflow manually, retry-step-group-case(0) will stucked on running phase even if workflow is completed.
<img width="368" alt="image" src="https://github.com/argoproj/argo-workflows/assets/18543532/922aea28-e4e3-4620-8c46-37f7126ad5ae">

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: workflow-fail-with-argument-
spec:
  entrypoint: main
  arguments:
    parameters:
      - name: code
        value: 1
  templates:
  - name: main
    steps:
    - - name: retry-step-group-case
        template: fail-step-group
      - name: retry-step-case
        template: fail-with-argument-with-retry
        arguments:
          parameters:
            - name: code
              value: '{{workflow.parameters.code}}'
      - name: step-case
        template: fail-with-argument
        arguments:
          parameters:
            - name: code
              value: '{{workflow.parameters.code}}'
  - name: fail-with-argument-with-retry
    inputs:
      parameters:
        - name: code
    container:
      image: python:alpine3.6
      command: [python, -c]
      args: ["import sys; sys.exit({{inputs.parameters.code}})"]
    retryStrategy:
      limit: "5"
      backoff:
        duration: "5"
        factor: "2"
        maxDuration: "1m"
  - name: fail-with-argument
    inputs:
      parameters:
        - name: code
    container:
      image: python:alpine3.6
      command: [python, -c]
      args: ["import sys; sys.exit({{inputs.parameters.code}})"]
  - name: fail-with-rate
    container:
      image: python:alpine3.6
[...]

Issue #12009, opened 2023-10-16
Avoid successful nodes in Steps(has RetryStrategy configured) being re-executed


jswxstw:
# Summary

Successful nodes should not be re-executed when Steps(has RetryStrategy configured) are retrying 

## Use Cases

My workflow may use multiple Steps template, but If I configure a RetryStrategy to Steps, nodes in Steps would be re-executed each time it retries.

Howerver, node in Steps may not be idempotent, so it will bring a lot of consistency problems.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

jswxstw:
<img width="214" alt="image" src="https://github.com/argoproj/argo-workflows/assets/18543532/100ed123-99e0-4d00-8dba-3122f61af40b">
As the picture shows, Workflow root is a Retry Node and its child nodes are Steps(retried three times), node(create-pvc) was re-executed each time even if it has been executed successfully which may be unaccecptable.

PR #12105, opened 2023-10-30
fix: Clean up pods of fulfilled nodes when manual retry. Fixes #12028

### LINKED ISSUE #12028: Completed pods were not all cleaned when workflow is succeesed after manual  retry
### Pre-requisites

1. Workflow PodGC strategy is set to OnWorkflowSuccess.
2. Workflow executed failed first time.
3. Retry the failed workflow manually, then workflow executes successfully and changes to succeeded phase.
4. The pods corresponding to the steps that have been successfully executed before will not be deleted. 

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: workflow-fail-with-argument-
spec:
  entrypoint: main
  podGC:
    strategy: OnWorkflowSuccess
  arguments:
    parameters:
      - name: code
        value: 1
  templates:
  - name: main
    steps:
    - - name: print
        template: print-with-argument
        arguments:
          parameters:
            - name: code
              value: '{{workflow.parameters.code}}'
    - - name: check-status
        template: fail-with-argument
        arguments:
          parameters:
            - name: code
              value: '{{workflow.parameters.code}}'
  - name: print-with-argument
    inputs:
      parameters:
        - name: code
    container:
      image: docker/whalesay:latest
      command: [sh, -c]
      args: ["echo {{inputs.parameters.code}}"]
  - name: fail-with-argument
    inputs:
      parameters:
        - name: code
    container:
      image: python:alpine3.6
      command: [python, -c]
      args: ["import sys; sys.exit({{inputs.parameters.code}})"]
```

```text
```

```text
```

jswxstw:
Fixes #12028 

### Motivation
PodGC strategy: OnWorkflowSuccess.
Pods of the previously successful steps were not cleaned up after workflow manual retry.
### Modifications
Pods of fulfilled nodes will be relabeled completed=false when workflow manual retry.
### Verification
Workflow Demo
```yaml
kind: Workflow
metadata:
  generateName: workflow-fail-with-argument-
spec:
  entrypoint: main
  podGC:
    strategy: OnWorkflowSuccess
  arguments:
    parameters:
      - name: code
        value: 1
  templates:
  - name: main
    steps:
    - - name: print
        template: print-with-argument
        arguments:
          parameters:
            - name: code
              value: '{{workflow.parameters.code}}'
    - - name: check-status
        template: fail-with-argument
        arguments:
          parameters:
            - name: code
              value: '{{workflow.parameters.code}}'
  - name: print-with-argument
    inputs:
      parameters:
        - name: code
    container:
      image: docker/whalesay:latest
      command: [sh, -c]
      args: ["echo {{inputs.parameters.code}}"]
  - name: fail-with-argument
    inputs:
      parameters:
        - name: code
    container:
      image: python:alpine3.6
      command: [python, -c]
      args: ["import sys; sys.exit({{inputs.parameters.code}})"]
```
1. Workflow executed failed first time.
image
image

2. Retry the failed workflow manually, then workflow executes successfully and changes to succeeded phase.
image

3. All pods were successfully cleaned up.
image

### DIFF
--- a/manifests/cluster-install/argo-server-rbac/argo-server-clusterole.yaml
+++ b/manifests/cluster-install/argo-server-rbac/argo-server-clusterole.yaml
@@ -29,6 +29,7 @@ rules:
       - list
       - watch
       - delete
+      - patch
   - apiGroups:
       - ""
     resources:
--- a/server/workflowarchive/archived_workflow_server.go
+++ b/server/workflowarchive/archived_workflow_server.go
@@ -17,6 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/argoproj/argo-workflows/v3/persist/sqldb"
 	workflowarchivepkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflowarchive"
@@ -286,7 +287,7 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 	_, err = wfClient.ArgoprojV1alpha1().Workflows(req.Namespace).Get(ctx, wf.Name, metav1.GetOptions{})
 	if apierr.IsNotFound(err) {
 
-		wf, podsToDelete, err := util.FormulateRetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector, req.Parameters)
+		wf, podsToDelete, podsToReset, err := util.FormulateRetryWorkflow(ctx, wf, req.RestartSuccessful, req.NodeFieldSelector, req.Parameters)
 		if err != nil {
 			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
@@ -294,6 +295,20 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 		for _, podName := range podsToDelete {
 			log.WithFields(log.Fields{"podDeleted": podName}).Info("Deleting pod")
 			err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
+			if err != nil && !apierr.IsNotFound(err) {
+				return nil, sutils.ToStatusError(err, codes.Internal)
+			}
+		}
+
+		for _, podName := range podsToReset {
+			log.WithFields(log.Fields{"podReset": podName}).Info("Resetting pod")
+			_, err := kubeClient.CoreV1().Pods(wf.Namespace).Patch(
+				ctx,
+				podName,
+				types.MergePatchType,
+				[]byte(`{"metadata": {"labels": {"workflows.argoproj.io/completed": "false"}}}`),
+				metav1.PatchOptions{},
+			)
 			if err != nil && !apierr.IsNotFound(err) {
 				return nil, sutils.ToStatusError(err, codes.Internal)
 			}
--- a/manifests/namespace-install/argo-server-rbac/argo-server-role.yaml
+++ b/manifests/namespace-install/argo-server-rbac/argo-server-role.yaml
@@ -29,6 +29,7 @@ rules:
       - list
       - watch
       - delete
+      - patch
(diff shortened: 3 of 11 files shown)


manifests/cluster-install/argo-server-rbac/argo-server-clusterole.yaml
```diff
@@ -29,6 +29,7 @@ rules:
       - list
       - watch
       - delete
+      - patch
```
agilgur5, line 32: this is a pretty minor permission given the Server already has `delete`, but I was thinking this logic may make more sense on the Controller actually.

When the Controller detects a retry, it can check the Workflow's child Pods.

Need to think a bit more about how that would work, but that would preserve the existing separation of duties between Server and Controller, where the Server is just a simple intermediary for users that can be bypassed with correct RBAC.

agilgur5: Hmmm I see that `podsToDelete` _already_ broke the separation a bit... we honestly may want to refactor that too...

@terrytangyuan do you have any thoughts on this? Specifically, the current behavior with the Server having Pod modification logic actually invalidates your answer in https://github.com/argoproj/argo-workflows/discussions/12027#discussioncomment-7316157 . I think it ideally _should_ behave the way you described, if possible.

server/workflowarchive/archived_workflow_server.go
```diff
@@ -299,6 +300,20 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req
 			}
 		}
 
+		for _, podName := range podsToReset {
```
agilgur5, line 303: I was gonna say that this may make sense to place into a helper function, but I see that https://github.com/argoproj/argo-workflows/pull/7988#discussion_r815240441 _explicitly_ moved the `kubeClient` logic out of the helper function (when retry was introduced for archived workflows in #7988).

So this matches existing behavior with `podsToDelete` above. If we refactored these into the Controller (per above comment), this would go away though 🤔

agilgur5:
Thanks for taking the time to investigate and fix this!

Per the in-line comments, we may want to refactor some _existing_ logic that this PR highlights as potentially having broken the Controller/Server separation of responsibilities.
I think the current code more or less makes sense within that existing context (the re-processing checks in the Controller are a bit confusing, though that's partially due to the existing logic)

jswxstw:
Retries(both auto or manual) are indeed the most complex parts of the codebase,  which bring a lot of problems like #12009, #12010.
Moving the implementation into Controller really make sense, I'm glad to see this improvement.

manifests/cluster-install/argo-server-rbac/argo-server-clusterole.yaml
terrytangyuan: You are right. RBAC modifications is required for my answer in https://github.com/argoproj/argo-workflows/discussions/12027#discussioncomment-7316157

agilgur5: @terrytangyuan I was actually looking for your thoughts on the approach. I think we should refactor this logic to not have the Server perform anything but a label modification, and then the Controller should actually delete, reset, etc child Pods as needed (as that is the Controller's responsibility, not the Server's)

sarabala1979: @terrytangyuan  I agree with @agilgur5. Server should not update or delete other than `Argo workflow` CRD.  If some user directly updates the WF spec then this will not work. Controller should handle the scenario to clean the workflow pods based on GCStrategy. 
I remember @ishitasequeira was proposed to delete all workflow pods "label based" like `workflowname` deletion if the workflow complete.

agilgur5: Julie agreed in https://github.com/argoproj/argo-workflows/pull/12419#issuecomment-1896588824 as well, so I filed #12538 as a tracking issue for this refactor

terrytangyuan: SGTM

isubasinghe:
Can you please rebase this @jswxstw ?

jswxstw:
Hi @isubasinghe,

This PR is from 2023 and #13734 (merged Oct 2024) changed the retry logic, so a straight rebase won't work.

PR #12076, opened 2023-10-24
docs: Add an example field to argo list command. Fixes #11898

Krunal2017:
Fixes #11898

### Motivation
Changes have been made for a more comprehensive argo list command

### Modifications
Added examples for some common usages of argo list command

### Verification
By running `argo list -h`

PR #12106, opened 2023-10-30
docs: Add an example field to argo cluster-template list command. Fixes #11898

shubhamch71:
Fixes #11898 

### Motivation

Changes to support more detailed information when the user runs the argo cluster-template list command

### Modifications

Added examples for some common usages of argo cluster-template list command

### Verification

RUN `argo cluster-template list --help`

### DIFF
--- a/cmd/argo/commands/clustertemplate/list.go
+++ b/cmd/argo/commands/clustertemplate/list.go
@@ -22,6 +22,15 @@ func NewListCommand() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "list",
 		Short: "list cluster workflow templates",
+		Example: `# List Cluster Workflow Templates:
+  argo cluster-template list
+	
+# List Cluster Workflow Templates in "wide" output format:
+  argo cluster-template list --output wide
+  
+# List Cluster Workflow Templates in "name" output format:
+  argo cluster-template list -o name
+`,
 		Run: func(cmd *cobra.Command, args []string) {
 			ctx, apiClient := client.NewAPIClient(cmd.Context())
 			serviceClient, err := apiClient.NewClusterWorkflowTemplateServiceClient()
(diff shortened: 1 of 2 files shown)


shubhamch71:
@terrytangyuan @agilgur5  Please have a look and let me know if any changes are supposed to be made!!!

cmd/argo/commands/clustertemplate/list.go
```diff
@@ -22,6 +22,15 @@ func NewListCommand() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "list",
 		Short: "list cluster workflow templates",
+		Example: `# List Cluster Workflow Templates:
+  argo cluster-template list
+	
+# List Cluster Workflow Templates in "wide" output format:
```
agilgur5, line 28: This is an example, so we should be more specific and not just reiterate the option.

For example, [this other PR](https://github.com/argoproj/argo-workflows/pull/12076/files#diff-f9df5c1e05ae7460690c91a7d11839e40bf0eba4ee12c8b91e6a3d6037a4b741R85) specified what other information `wide` provides vs the default.
You should similarly test to see what it does and be more specific

cmd/argo/commands/clustertemplate/list.go
```diff
@@ -22,6 +22,15 @@ func NewListCommand() *cobra.Command {
...
+  argo cluster-template list --output wide
+  
+# List Cluster Workflow Templates in "name" output format:
```
agilgur5, line 31: Same as above, this should be more specific. I believe `name` lists names only, but I'm not sure, haven't tried it

shubhamch71: @agilgur5 I have made the modifications and tried to be more specific. Please have a look at it once and also the windows unit tests are failing but I am not able to understand why !! Please throw some light on it as well !!

agilgur5: The Windows tests aren't required to pass CI. They're new and still being worked on

agilgur5 (commented):
two small modifications below to improve directness, otherwise LGTM

agilgur5 (approved):

PR #12111, opened 2023-10-31
fix: Upgrade axios to v1.6.0. Fixes #12085

### LINKED ISSUE #12085: New vulnerability detected in Node dependency axios
Master branch failed with:

```
 Issues with no direct upgrade or patch:
  ✗ Cross-site Request Forgery (CSRF) [High Severity][https://security.snyk.io/vuln/SNYK-JS-AXIOS-6032459] in axios@1.5.1
    introduced by swagger-ui-react@4.19.1 > swagger-client@3.23.1 > @swagger-api/apidom-reference@0.78.0 > axios@1.5.1
  No upgrade or patch available
```

https://security.snyk.io/vuln/SNYK-JS-AXIOS-6032459

There is a documented workaround that we can use or wait for a new axios patch.

terrytangyuan:
Fixes #12085.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -87,6 +87,7 @@
         "yarn-deduplicate": "^6.0.2"
     },
     "resolutions": {
-        "semver": "^7.5.2"
+        "semver": "^7.5.2",
+        "axios": "1.6.0"
     }
 }
--- a/ui/package.json
+++ b/ui/package.json
@@ -87,7 +87,6 @@
         "yarn-deduplicate": "^6.0.2"
     },
     "resolutions": {
-        "semver": "^7.5.2",
-        "axios": "1.6.0"
+        "semver": "^7.5.2"
     }
 }
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -2610,10 +2610,10 @@ autolinker@^3.11.0:
   dependencies:
     tslib "^2.3.0"
 
-axios@^1.4.0:
-  version "1.5.1"
-  resolved "https://registry.yarnpkg.com/axios/-/axios-1.5.1.tgz#11fbaa11fc35f431193a9564109c88c1f27b585f"
-  integrity sha512-Q28iYCWzNHjAm+yEAot5QaAMxhMghWLFVf7rRdwhUI+c2jix2DUXjAHXVi+s1ibs3mjPO/cCgbA++3BjD0vP/A==
+axios@1.6.0, axios@^1.4.0:
+  version "1.6.0"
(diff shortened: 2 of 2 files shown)


ui/package.json
```diff
@@ -87,6 +87,7 @@
         "yarn-deduplicate": "^6.0.2"
     },
     "resolutions": {
-        "semver": "^7.5.2"
+        "semver": "^7.5.2",
+        "axios": "1.6.0"
```
agilgur5, line 91: this doesn't need a `resolution`, as I mentioned in a [previous PR](https://github.com/argoproj/argo-workflows/pull/11641#issuecomment-1693869029). you can change the lockfile directly to update it.

A `resolution` is an override, but we don't need to override this as it's a SemVer-compatible update

```suggestion
        "semver": "^7.5.2"
```

ui/yarn.lock
```diff
@@ -2610,10 +2610,10 @@ autolinker@^3.11.0:
   dependencies:
     tslib "^2.3.0"
 
-axios@^1.4.0:
-  version "1.5.1"
-  resolved "https://registry.yarnpkg.com/axios/-/axios-1.5.1.tgz#11fbaa11fc35f431193a9564109c88c1f27b585f"
-  integrity sha512-Q28iYCWzNHjAm+yEAot5QaAMxhMghWLFVf7rRdwhUI+c2jix2DUXjAHXVi+s1ibs3mjPO/cCgbA++3BjD0vP/A==
+axios@1.6.0, axios@^1.4.0:
+  version "1.6.0"
```
agilgur5, line 2614: `1.6.0` is SemVer-compatible with `^1.4.0`. Also if you notice here, this resolution would actually lock us into `1.6.0` as it is an exact version and not a SemVer range.

```suggestion
axios@^1.4.0:
  version "1.6.0"
```

agilgur5 (changes requested):

agilgur5:
So it looks like this was a breaking change: https://github.com/axios/axios/pull/6028#issuecomment-1783536371
Per [previous comment](https://github.com/argoproj/argo-workflows/issues/12085#issuecomment-17815463380), I don't think a cookie is needed for the API docs, but could you check to confirm that it still works?

Also I might recommend `fix(deps): Update transitive axios to v1.6.0` for the title to be a bit more specific as it's not a dep we use directly

agilgur5:
> but could you check to confirm that it still works?

Checked myself, seems to work fine. So I think this is good to go once the suggested changes are made

PR #12133, opened 2023-11-03
fix: Resource version incorrectly overridden for wfInformer list requests. Fixes #11948

### LINKED ISSUE #11948: Controller: High RAM Usage and Frequent Crashes on Large Scale Workflows
### Pre-requisites

**Description:**
We are observing significant issues with the Argo Workflows Controller while handling a large number of workflows in parallel.

**Environment:**
Argo Workflows version: 3.4.11
Nodes count: 300+
Parallel workflows: 5000+

**What happened:**
The Argo Workflows Controller's memory consumption increases exponentially, sometimes surpassing 100GB. Despite this excessive memory usage, the controller crashes frequently. Notably, despite workflows being archived, they aren't deleted post-archiving, possibly contributing to the memory usage. It does not log any specific error messages prior to these crashes, making it challenging to pinpoint the cause or underlying issue.

**What you expected to happen:**
We expected the Argo Workflows Controller to handle the parallel execution of 5000+ workflows across 300+ nodes without such a drastic increase in RAM consumption. We also expected a more resilient behavior, not prone to unexpected crashes, and better error logging for troubleshooting.

**How to reproduce it (as minimally and precisely as possible):**
Set up an environment with 300+ nodes.
Launch 5000+ workflows in parallel.
Monitor the RAM usage of the Argo Workflows Controller and note any unexpected crashes.

**Additional context:**
Given the scale at which we are operating, it's critical for our operations that Argo can handle such workloads efficiently. Any assistance in resolving this issue or guidance on potential optimizations would be greatly appreciated.

RAM

### Version

v3.4.11

```YAML
-
```

```text
-
```

```text
-
```

terrytangyuan:
Potential fix for #11948

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -266,7 +266,7 @@ func (wfc *WorkflowController) Run(ctx context.Context, wfWorkers, workflowTTLWo
 		WithField("podCleanup", podCleanupWorkers).
 		Info("Current Worker Numbers")
 
-	wfc.wfInformer = util.NewWorkflowInformer(wfc.dynamicInterface, wfc.GetManagedNamespace(), workflowResyncPeriod, wfc.tweakListOptions, indexers)
+	wfc.wfInformer = util.NewWorkflowInformer(wfc.dynamicInterface, wfc.GetManagedNamespace(), workflowResyncPeriod, wfc.tweakListRequestListOptions, wfc.tweakWatchRequestListOptions, indexers)
 	wfc.wftmplInformer = informer.NewTolerantWorkflowTemplateInformer(wfc.dynamicInterface, workflowTemplateResyncPeriod, wfc.managedNamespace)
 	wfc.wfTaskSetInformer = wfc.newWorkflowTaskSetInformer()
 	wfc.artGCTaskInformer = wfc.newArtGCTaskInformer()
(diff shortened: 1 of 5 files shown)


workflow/controller/controller.go
```diff
@@ -817,7 +817,7 @@ func (wfc *WorkflowController) enqueueWfFromPodLabel(obj interface{}) error {
 	return nil
 }
 
-func (wfc *WorkflowController) tweakListOptions(options *metav1.ListOptions) {
+func (wfc *WorkflowController) tweakListRequestListOptions(options *metav1.ListOptions) {
```
Joibel, line 820: Possibly: keep them in step with
```
tweakWatchRequestListOptions(options)
options.ResourceVersion = ""
```

jessesuen:
I compared the watch API call in the controller logs from before the change v3.4.6:

```
GET [https://172.20.0.1:443/apis/argoproj.io/v1alpha1/workflows?allowWatchBookmarks=true&labelSelector=%21workflows.argoproj.io%2Fcontroller-instanceid&resourceVersion=238725385&timeoutSeconds=483&watch=true](https://172.20.0.1/apis/argoproj.io/v1alpha1/workflows?allowWatchBookmarks=true&labelSelector=%21workflows.argoproj.io%2Fcontroller-instanceid&resourceVersion=238725385&timeoutSeconds=483&watch=true) 200 OK
```

and after (v3.4.11):

```
GET [https://172.20.0.1:443/apis/argoproj.io/v1alpha1/workflows?allowWatchBookmarks=true&labelSelector=%21workflows.argoproj.io%2Fcontroller-instanceid&timeoutSeconds=407&watch=true](https://172.20.0.1/apis/argoproj.io/v1alpha1/workflows?allowWatchBookmarks=true&labelSelector=%21workflows.argoproj.io%2Fcontroller-instanceid&timeoutSeconds=407&watch=true) 200 OK
```

Before, in v3.4.6, K8s would set `resourceVersion=238725385` to the watch. In v3.411 it gets omitted entirely because of the change. I agree that this could cause a potential timing issue because kubernetes informer relies on continuing from a state of the world at a specific resourceVersion, but is no longer able to because we omit resourceVersion from the watch call.

So I think the fix may work.

workflow/controller/controller.go
terrytangyuan: I don't have a preference here.

jessesuen (approved):

PR #12145, opened 2023-11-04
fix(server): allow passing loglevels as env vars to Server

agilgur5:
### Motivation

In #11978 I noticed that `ARGO_LOGLEVEL` did not work for the Server
- `--loglevel` (and `--gloglevel`) were only set [at the parent command, `argo`](https://github.com/argoproj/argo-workflows/blob/b2e3676503999e699cd3aaca62ff7bfff64464af/cmd/argo/commands/root.go#L120), and not in [the `argo server` subcommand](https://github.com/argoproj/argo-workflows/blob/b2e3676503999e699cd3aaca62ff7bfff64464af/cmd/argo/commands/server.go#L42) itself
  - however, `viper.AutomaticEnv()` and `viper.BindPFlags(command.Flags())` does not evaluate the parent command (related https://github.com/spf13/viper/issues/646, https://github.com/spf13/viper/issues/375, https://github.com/spf13/viper/issues/801)
    - similarly, the code that actually sets the loglevels (which is in the CLI itself), would not run when an env var was used
    - hence, we have a few options:
      1. import the parent command and bind it in the child or something wacky like that
          - **EDIT**: I tried this and this crashed hard
            - trying `command.Parent().PersistentFlags()` would result in a segfault
              - tried to use `AddFlagSet` to add it in `server.go` and that crashed hard. It crashes without the `AddFlagSet`, just the `Parent().PersistentFlags()` causes it (so `viper.BindPFlags(command.Parent().PersistentFlags())` crashed too etc)
              - <details> <summary> segfault error log </summary>
                
                ```
                server: panic: runtime error: invalid memory address or nil pointer dereference
                server: [signal SIGSEGV: segmentation violation code=0x1 addr=0x188 pc=0xa3d692]
                server:
                server: goroutine 1 [running]:
                server: github.com/spf13/cobra.(*Command).PersistentFlags(0xc000169b00?)
                server:         /home/vscode/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1676 +0x12
                server: github.com/argoproj/argo-workflows/v3/cmd/argo/commands.NewServerCommand()
                server:         /home/vscode/go/src/github.com/argoproj/argo-workflows/cmd/argo/commands/server.go:240 +0xcda
                server: github.com/argoproj/argo-workflows/v3/cmd/argo/commands.NewCommand()
                server:         /home/vscode/go/src/github.com/argoproj/argo-workflows/cmd/argo/commands/root.go:100 +0x23e
                server: main.main()
                server:         /home/vscode/go/src/github.com/argoproj/argo-workflows/cmd/argo/main.go:14 +0x13
                2023/11/04 19:50:17 server: process exited 89278: exit status 2
                2023/11/04 19:50:17 server: context cancelled, stopping process
                server: exit status 2
                ```
                
                </details>
      1. try to bind env vars at the parent level
          - which may not work and would apply to _all_ subcommands, even though `viper` env vars are only used for the Controller & Server
          - parent flags and persistent flags also seem to a work a tad differently in `viper` / `cobra` per https://github.com/spf13/viper/issues/646
          - **EDIT**: this is the option I ended up going with since Option C caused a panic
      1. just add flags for loglevels inside of the `argo server` subcommand
          - this seemed by far the easiest, least complex, and least LoC
          - it is also consistent with the Controller as well
          - and we already have `--log-format` in the `argo server` subcommand as well
          - **EDIT**: this caused a panic unfortunately
            - <details> <summary> gloglevel klog flag redefinition error </summary>

              ```
              server: ./dist/argo flag redefined: log_dir
              server: panic: ./dist/argo flag redefined: log_dir
              server: 
              server: goroutine 1 [running]:
              server: flag.(*FlagSet).Var(0xc0001f8150, {0x314fe28, 0x47c1bf0}, {0x2aae917, 0x7}, {0x2b7e755, 0x52})
              server:         /usr/local/go/src/flag/flag.go:1028 +0x3a5
              server: flag.(*FlagSet).StringVar(...)
              server:         /usr/local/go/src/flag/flag.go:879
              server: k8s.io/klog/v2.InitFlags(0x4?)
              server:         /home/vscode/go/pkg/mod/k8s.io/klog/v2@v2.70.1/klog.go:424 +0x4f
              server: github.com/argoproj/argo-workflows/v3/util/cmd.SetGLogLevel(0xc00001283c?)
              server:         /home/vscode/go/src/github.com/argoproj/argo-workflows/util/cmd/glog.go:14 +0x1a
              server: github.com/argoproj/argo-workflows/v3/cmd/argo/commands.NewServerCommand.func1(0xc00022fd00?, {0x2aa9b6f?, 0x4?, 0x2aa9a8b?})
              server:         /home/vscode/go/src/github.com/argoproj/argo-workflows/cmd/argo/commands/server.go:75 +0x16f
              server: github.com/spf13/cobra.(*Command).execute(0xc0001d7800, {0x4819640, 0x0, 0x0})
              server:         /home/vscode/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x87c
              server: github.com/spf13/cobra.(*Command).ExecuteC(0xc000005200)
              server:         /home/vscode/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3a5
              server: github.com/spf13/cobra.(*Command).Execute(0x0?)
              server:         /home/vscode/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992 +0x13
              server: main.main()
              server:         /home/vscode/go/src/github.com/argoproj/argo-workflows/cmd/argo/main.go:14 +0x18
              2023/11/04 19:42:10 server: process exited 82584: exit status 2
              server: exit status 2
              ```

              </details>
              
            - running [`cmd.SetGLogLevel`](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/util/cmd/glog.go#L13) in `server.go` would cause a panic due to a flag redefinition of `klog`
             - i.e. `SetGLogLevel` can't be called twice
             - but `cli.SetLogLevel` worked fine, so the only issue is `--gloglevel`, to be exact

### Modifications

- ~added the two flags to the `argo server` subcommand next to `--log-format`~
  - ~added the same code to `RunE` that [`argo`](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/cmd/argo/commands/root.go#L128-L129) and [`workflow-controller`](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/cmd/workflow-controller/main.go#L74-L75) run when these flags are set~

**EDIT**: Option C actually did not work, so went with Option B

- so instead of hacking together some logic to get Option C to work, I just added `viper` to `root.go`, where it does capture the env vars from subcommands too
  - same `viper` logic as in other parts of the CLI (like [`server.go`](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/cmd/argo/commands/server.go#L240)), just using `command.PersistentFlags()` instead of `command.Flags()`

  - one caveat is that now env vars can be used for `loglevel`, `gloglevel`, and `verbose` for _all_ of the Argo CLI
    - nbd, IMO, but it is noteworthy to mention as a side-effect, especially when compared to Option C
  
### Verification

  - tested by changing [this init log line](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/server/apiserver/argoserver.go#L272) to a `Debugf` instead of an `Infof` and then changing [the `ARGO_LOGLEVEL` in the kit `tasks.yaml` to `DEBUG`](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/tasks.yaml#L72) as well
    - and then checking for this in the logs:
    ```
    server: time="2023-11-04T20:12:56.151Z" level=debug msg="Argo Server started successfully on http://localhost:2746" url="http://localhost:2746"
    ```

### DIFF
--- a/cmd/argo/commands/root.go
+++ b/cmd/argo/commands/root.go
@@ -141,9 +141,11 @@ If your server is behind an ingress with a path (you'll be running "argo server 
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("ARGO")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
+	// bind flags to env vars (https://github.com/spf13/viper/tree/v1.17.0#working-with-flags)
 	if err := viper.BindPFlags(command.PersistentFlags()); err != nil {
 		log.Fatal(err)
 	}
+	// workaround for handling required flags (https://github.com/spf13/viper/issues/397#issuecomment-544272457)
 	command.PersistentFlags().VisitAll(func(f *pflag.Flag) {
 		if !f.Changed && viper.IsSet(f.Name) {
 			val := viper.Get(f.Name)
--- a/cmd/argo/commands/root.go
+++ b/cmd/argo/commands/root.go
@@ -1,9 +1,14 @@
 package commands
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/argoproj/pkg/cli"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
 
 	"github.com/argoproj/argo-workflows/v3"
 	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands/archive"
@@ -132,6 +137,21 @@ If your server is behind an ingress with a path (you'll be running "argo server 
 	command.PersistentFlags().StringVar(&logLevel, "loglevel", "info", "Set the logging level. One of: debug|info|warn|error")
 	command.PersistentFlags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enabled verbose logging, i.e. --loglevel debug")
+
+	viper.AutomaticEnv()
+	viper.SetEnvPrefix("ARGO")
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
+	if err := viper.BindPFlags(command.PersistentFlags()); err != nil {
+		log.Fatal(err)
+	}
+	command.PersistentFlags().VisitAll(func(f *pflag.Flag) {
+		if !f.Changed && viper.IsSet(f.Name) {
+			val := viper.Get(f.Name)
+			if err := command.PersistentFlags().Set(f.Name, fmt.Sprintf("%v", val)); err != nil {
+				log.Fatal(err)
+			}
+		}
+	})
 
 	return command
 }
--- a/cmd/argo/commands/root.go
+++ b/cmd/argo/commands/root.go
@@ -138,6 +138,7 @@ If your server is behind an ingress with a path (you'll be running "argo server 
 	command.PersistentFlags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enabled verbose logging, i.e. --loglevel debug")
 
+	// set-up env vars for the CLI such that ARGO_* env vars can be used instead of flags
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("ARGO")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
--- a/cmd/argo/commands/server.go
+++ b/cmd/argo/commands/server.go
@@ -13,6 +13,7 @@ import (
 
 	eventsource "github.com/argoproj/argo-events/pkg/client/eventsource/clientset/versioned"
 	sensor "github.com/argoproj/argo-events/pkg/client/sensor/clientset/versioned"
+	"github.com/argoproj/pkg/cli"
 	"github.com/argoproj/pkg/stats"
 	log "github.com/sirupsen/logrus"
 	"github.com/skratchdot/open-golang/open"
@@ -59,6 +60,8 @@ func NewServerCommand() *cobra.Command {
 		kubeAPIQPS               float32
 		kubeAPIBurst             int
 		allowedLinkProtocol      []string
+		logLevel                 string // --loglevel
+		glogLevel                int    // --gloglevel
 		logFormat                string // --log-format
 	)
 
@@ -68,6 +71,8 @@ func NewServerCommand() *cobra.Command {
 		Example: fmt.Sprintf(`
 See %s`, help.ArgoServer),
 		RunE: func(c *cobra.Command, args []string) error {
+			cli.SetLogLevel(logLevel)
+			cmd.SetGLogLevel(glogLevel)
 			cmd.SetLogFormatter(logFormat)
 			stats.RegisterStackDumper()
 			stats.StartStatsTicker(5 * time.Minute)
@@ -233,6 +238,8 @@ See %s`, help.ArgoServer),
 	command.Flags().StringVar(&accessControlAllowOrigin, "access-control-allow-origin", "", "Set Access-Control-Allow-Origin header in HTTP responses.")
 	command.Flags().Uint64Var(&apiRateLimit, "api-rate-limit", 1000, "Set limit per IP for api ratelimiter")
 	command.Flags().StringArrayVar(&allowedLinkProtocol, "allowed-link-protocol", defaultAllowedLinkProtocol, "Allowed link protocol in configMap. Used if the allowed configMap links protocol are different from http,https. Defaults to the environment variable ALLOWED_LINK_PROTOCOL")
+	command.Flags().StringVar(&logLevel, "loglevel", "info", "Set the logging level. One of: debug|info|warn|error")
+	command.Flags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.Flags().StringVar(&logFormat, "log-format", "text", "The formatter to use for logs. One of: text|json")
 	command.Flags().Float32Var(&kubeAPIQPS, "kube-api-qps", 20.0, "QPS to use while talking with kube-apiserver.")
 	command.Flags().IntVar(&kubeAPIBurst, "kube-api-burst", 30, "Burst to use while talking with kube-apiserver.")
--- a/cmd/argo/commands/server.go
+++ b/cmd/argo/commands/server.go
@@ -237,6 +237,7 @@ See %s`, help.ArgoServer),
 	command.Flags().Float32Var(&kubeAPIQPS, "kube-api-qps", 20.0, "QPS to use while talking with kube-apiserver.")
 	command.Flags().IntVar(&kubeAPIBurst, "kube-api-burst", 30, "Burst to use while talking with kube-apiserver.")
 
+	// set-up env vars for the CLI such that ARGO_* env vars can be used instead of flags
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("ARGO")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
--- a/cmd/argo/commands/server.go
+++ b/cmd/argo/commands/server.go
@@ -13,7 +13,6 @@ import (
 
 	eventsource "github.com/argoproj/argo-events/pkg/client/eventsource/clientset/versioned"
 	sensor "github.com/argoproj/argo-events/pkg/client/sensor/clientset/versioned"
-	"github.com/argoproj/pkg/cli"
 	"github.com/argoproj/pkg/stats"
 	log "github.com/sirupsen/logrus"
 	"github.com/skratchdot/open-golang/open"
@@ -60,8 +59,6 @@ func NewServerCommand() *cobra.Command {
 		kubeAPIQPS               float32
 		kubeAPIBurst             int
 		allowedLinkProtocol      []string
-		logLevel                 string // --loglevel
-		glogLevel                int    // --gloglevel
 		logFormat                string // --log-format
 	)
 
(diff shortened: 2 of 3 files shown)


cmd/argo/commands/root.go
```diff
@@ -133,5 +138,20 @@ If your server is behind an ingress with a path (you'll be running "argo server
...
+	}
+	command.PersistentFlags().VisitAll(func(f *pflag.Flag) {
+		if !f.Changed && viper.IsSet(f.Name) {
+			val := viper.Get(f.Name)
+			if err := command.PersistentFlags().Set(f.Name, fmt.Sprintf("%v", val)); err != nil {
+				log.Fatal(err)
+			}
+		}
+	})
+
```
juliev0, line 155: can we add comments to make it clear what all of this is doing?

agilgur5: it's not commented in [other files](https://github.com/argoproj/argo-workflows/blob/a47e46208cd0029ef2e35019207d0b1ee4f85ea4/cmd/argo/commands/server.go#L240), but I could add explanatory comments to all of the files

juliev0:
So, if this happens at both the root and in the server subcommand, then when `argo server` is called, for each possible root flag, it will look for "ARGO_<flag_with_underscores>"  and then for each possible subcommand flag, it will do the same, right?

agilgur5:
yup, that's correct

cmd/argo/commands/root.go
juliev0: that would be great

agilgur5: added comments in the most recent [commit](https://github.com/argoproj/argo-workflows/pull/12145/commits/31793b029a0676706713c7ffaa7a01008822d738)

cmd/argo/commands/root.go
```diff
@@ -133,5 +138,22 @@ If your server is behind an ingress with a path (you'll be running "argo server
 	command.PersistentFlags().IntVar(&glogLevel, "gloglevel", 0, "Set the glog logging level")
 	command.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enabled verbose logging, i.e. --loglevel debug")
 
+	viper.AutomaticEnv()
+	viper.SetEnvPrefix("ARGO")
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
+	// bind flags to env vars (https://github.com/spf13/viper/tree/v1.17.0#working-with-flags)
```
juliev0, line 144: hate to be anal but was thinking we could clarify in a comment that we're enabling somebody to use environment variables instead of flags, prefixed by "ARGO" with character replacement

I know there's an idea that maybe sometimes code can kind of explain itself. But I just think most of these functions people have never seen before, and it would be nice to know what this whole block of code is doing without having to read through each line and look up the function definition. Let me know if you disagree.

cmd/argo/commands/root.go
```diff
@@ -141,9 +141,11 @@ If your server is behind an ingress with a path (you'll be running "argo server
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("ARGO")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
+	// bind flags to env vars (https://github.com/spf13/viper/tree/v1.17.0#working-with-flags)
 	if err := viper.BindPFlags(command.PersistentFlags()); err != nil {
```
pryingbytez, line 145: Wouldn't this remove structured key configuration from viper?
E.G:
```yaml
foo:
     bar

```
 Would result in being saved as in .yaml configuration:
 ```yaml
 bar
```

Comment (agilgur5) [reply cmd/argo/commands/root.go:145]: Are you referring to the fact that `PersistentFlags` is being used?

We don't use structured key configuration and we _don't_ want structured/nested env vars either. So that would be intended behavior.

Per the PR description, this is only intended to be used by `argo server`, but in this case the flag is at the root level.

Comment (pryingbytez) [reply cmd/argo/commands/root.go:145]: Ah okay, i was having the same issues in my project,  I use  structured/nested env vars, which unfortunately does not work with `BindPFlags`, so i was just really curious.  I really hope this will get  resolved in v2.

File: cmd/argo/commands/root.go (hunk 2 above)
Comment (agilgur5) [reply cmd/argo/commands/root.go:144]: mmm I do think the first 3 lines are pretty self-explanatory, which is why I only added comments to the somewhat opaque and fairly specific workarounds below those lines. with this new comment, I think that inference on the first 3 lines becomes even more simple (as env vars are directly mentioned).

also if a contributor is familiar with Argo's [environment variables](https://argoproj.github.io/argo-workflows/environment-variables/), then this lines up pretty neatly when looking at the code.

if a contributor is going to make CLI modifications, then they will likely need at least some cursory knowledge of `cobra`, since the CLI is fairly coupled to `cobra`'s implementation. `pflag` and `viper` are commonly used with `cobra` as well. `viper` is pretty commonly used for config files and/or env vars.

and more generally speaking, there's a bit of a slipper slope there -- why not comment every line of usage of `cobra` in the codebase then? I think it is reasonable that if someone wants a deep understanding or wants to modify a specific functionality, such as env vars in this case, that they will need to have some familiarity with the underlying library that handles that.

for a point of reference, I had actually never written code for a Go CLI before working on Argo. I had seen many Go CLIs use `cobra` though and had heard of `viper` before (though not seen much code with it). When I first saw the `viper` code I immediately thought "ah this must be the env var -> flag mapping". I then checked the `blame` to confirm that and did not look up any of these functions. The `BindPFlags` and `VisitAll` blocks vaguely made sense to me as well at a glance.
I only read through `viper` in more detail for this PR specifically.
(you might say I am perhaps an outlier when it comes to points of reference though 😅)

Comment (juliev0) [reply cmd/argo/commands/root.go:144]: I've worked in software for 2 decades but not on a CLI before. The concept of "flag to environment variable" mapping was not something I was aware of. For anyone like me, I think it should be easy to just read the comment and know what it's for.

Comment (agilgur5) [reply cmd/argo/commands/root.go:144]: The new comment does specifically explain the mapping part though? Not sure if I'm misunderstanding something, but I thought you were referencing the first 3 lines that have no comment currently. The first 3 lines just set up env vars in general

Comment (juliev0) [reply cmd/argo/commands/root.go:144]: Why not make the comments more user friendly?....like: 
```
// enable environment variables to be passed instead of flags
// for each flag, an environment variable capitalized, prefixed by "ARGO", and with periods and dashes replaced with underscores
```

agilgur5: I don't think that spelling out word-for-word _exactly_ what each line does sets a good precedent, but I can meet you somewhere in the middle and give an explanation off what the entire `viper` block as a whole is intended to do

agilgur5: See [new commit](https://github.com/argoproj/argo-workflows/pull/12145/commits/27765a9e6a75ad55ed1ef2528db132cca9a12bf2)

juliev0 (approved):

cmd/argo/commands/root.go
juliev0: Thanks for the compromise. Do you mean the precedent of insisting on more comments from others? Is one of your concerns the maintenance of comments? That we may forget to change a comment when the code changes? I feel we need to balance that with efficient readability.

Re-reading your comments above, I think one difference in our perspectives is that you’re referring to somebody who’s modifying the code here. I am just thinking of the person who’s trying to read it and quickly get a sense of how it works.

agilgur5: > Do you mean the precedent of insisting on more comments from others? Is one of your concerns the maintenance of comments?

No I mean specifically of commenting to explain what _each line_ does. As you know, I heavily document things, so I most certainly believe that comments have their place.
Explaining a confusing workaround -- absolutely, useful in pretty much every scenario I can think of as otherwise it's difficult to tell why code was written the way it was; this should be done more often and I'll say this myself in code reviews too. A summary of a large block of related code, sure, I think that can add value sometimes when quickly reading through; the comment effectively lets you skip through.

_Each line_ though, I hesitate quite a bit with that. Every single statement shouldn't be explained and how do you draw a boundary there? which function calls to which libraries should be commented and which shouldn't?
Like `SetEnvPrefix` and `SetEnvKeyReplacer` are pretty self-explanatory for me and are not confusingly named to me. They are also common use-cases of env vars in general.

> I think one difference in our perspectives is that you’re referring to somebody who’s modifying the code here. I am just thinking of the person who’s trying to read it and quickly get a sense of how it works.

I referred to both. The first time I read it I wasn't modifying the env vars and I got the gist of what the code does because I already knew that Argo allowed env vars instead of flags from [the docs](https://argoproj.github.io/argo-workflows/environment-variables/) and from having used that functionality myself before. The `viper` code was also only in 2 files before this (now 3 files), only the Controller and Server executables specifically.

The gist of it I think is enough for a reader. Knowing specifically what each function call to a library does I think falls more within the realm of a modifier, who should ideally read the library's docs to get familiar with the usage they plan to modify.

juliev0: Yeah, I agree on not commenting every line of code. In my example the comments at the top were for an entire code block. I agree they were pretty detailed. I’m good with your shorter alternative. I think I was going for extra clarity because it explained the interface to the application. I didn’t realize this is explained in the docs as well. Thanks for talking it out and good to know. I think our perspectives aren’t really that different but maybe we both different err differently.

agilgur5: > I think our perspectives aren’t really that different

Indeed that seems to be the case most of the time 🙂

PR #12146, opened 2023-11-04
refactor: invert conditionals for less nesting in `includeScriptOutput`

agilgur5:
### Motivation

- I didn't get a chance to review #12020 before it was merged in, but it added more nesting when there was already some unnecessary nesting
- less nesting = less code complexity

### Modifications

- invert the `boundaryID` conditional from there for less nesting
- and invert the `boundaryNode` conditional as well for less nesting

- **NOTE**: there is a small semantic difference here from the previous code from #11451
  - when there is an error retrieving `boundaryNode`, the error will now be returned
  - previously it was not returned, but the purpose of the PR was to expose these errors as they are indicative of an underlying logic bug
    - and indeed #12020 added the top conditional to avoid an underlying logic bug
    - so I think this change is more within the _intent_ of both PRs, but CC @isubasinghe (who authored #11451)

### Verification

- other than the tiny semantic change above to expose more errors, this is semantically equivalent to the previous behavior

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -3688,29 +3688,30 @@ func (woc *wfOperationCtx) deletePDBResource(ctx context.Context) error {
 // Check if the output of this node is referenced elsewhere in the Workflow. If so, make sure to include it during
 // execution.
 func (woc *wfOperationCtx) includeScriptOutput(nodeName, boundaryID string) (bool, error) {
-	if boundaryID != "" {
-		if boundaryNode, err := woc.wf.Status.Nodes.Get(boundaryID); err == nil {
-			tmplCtx, err := woc.createTemplateContext(boundaryNode.GetTemplateScope())
-			if err != nil {
-				return false, err
-			}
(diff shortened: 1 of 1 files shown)


workflow/controller/operator.go
```diff
@@ -3688,29 +3688,30 @@ func (woc *wfOperationCtx) deletePDBResource(ctx context.Context) error {
...
-			if templateStored {
-				woc.updated = true
-			}
+	if boundaryID == "" {
+		return false, nil
+	}
+	boundaryNode, err := woc.wf.Status.Nodes.Get(boundaryID)
+	if err != nil {
+		woc.log.Errorf("was unable to obtain node for %s", boundaryID)
+		return false, err
```
agilgur5, line 3697: this would've returned `false, nil` [before](https://github.com/argoproj/argo-workflows/pull/11451/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R3610)

juliev0 (approved):

PR #12150, opened 2023-11-05
refactor(ui): code-split gigantic Monaco Editor dep

agilgur5:
Partial fix for #12059 -- this is the largest improvement by a wide margin, though still a little bit left to do
Partial fix for #11740 

### Motivation

- shave off 7.72MB / 3MB minified / 762KB gzipped from main bundle
  - main bundle itself is now only 5.11MB / 2.81MB minified / 553KB gzipped
    - i.e. the _entire_ main bundle is smaller than the Monaco bundle 😮 😬

### Modifications

- wrap `react-monaco-editor` into a lazy loaded component
  - with some tiny gimmicks to it per in-line comments

- in `ObjectEditor`, asynchronously import `languages` as well in one of its effects
  - otherwise all of `monaco-editor` gets imported, per in-line comment
  - also rewrite that effect to use modern async/await instead of promise chains
    - which ends up a good bit simpler than what it would be with a promise chain with the new async import on top of the async fetch etc
    - and invert a conditional as well to reduce nesting / code complexity

- also refactor `ObjectEditor` a bit to correctly use `useRef` instead of `createRef`
  - `createRef` is for legacy class-based components, while `useRef` is for hooks-based functional components

- add [`webpackChunkName` magic comment](https://webpack.js.org/api/module-methods/#webpackchunkname) for clearer names for all of the dynamic imports
  - note that sub-chunks (like `Chart.js`, which is a chunk of `reports`) can't be directly named
    - [`output.chunkFilename`](https://webpack.js.org/configuration/output/#outputchunkfilename) is how it's typically configured, but you can't [access](https://webpack.js.org/configuration/output/#template-strings) the unresolved import name in that
      - `[name]` is equivalent to `[id]` in this case; I tried
      
### Verification

Tested locally, editor seems to work fine in the UI:
Screenshot 2023-11-04 at 11 40 43 PM

#### Bundle verification

Full bundle, `monaco-editor` now split out:
Screenshot 2023-11-05 at 12 27 22 AM - split monaco

Size of remaining main bundle:
Screenshot 2023-11-05 at 12 28 05 AM - main wo monaco

### DIFF
--- a/ui/src/app/shared/components/object-editor/object-editor.tsx
+++ b/ui/src/app/shared/components/object-editor/object-editor.tsx
@@ -44,7 +44,7 @@ export const ObjectEditor = <T extends any>({type, value, buttons, onChange}: Pr
 
     useEffect(() => {
         if (!type || lang !== 'json') {
-            return
+            return;
         }
 
         (async () => {
--- a/ui/src/app/shared/components/object-editor/object-editor.tsx
+++ b/ui/src/app/shared/components/object-editor/object-editor.tsx
@@ -1,12 +1,13 @@
-import {languages} from 'monaco-editor/esm/vs/editor/editor.api';
 import * as React from 'react';
-import {createRef, useEffect, useState} from 'react';
+import {useEffect, useRef, useState} from 'react';
 import MonacoEditor from 'react-monaco-editor';
+
 import {uiUrl} from '../../base';
 import {ScopedLocalStorage} from '../../scoped-local-storage';
 import {Button} from '../button';
 import {parse, stringify} from '../object-parser';
 import {PhaseIcon} from '../phase-icon';
+import {SuspenseMonacoEditor} from '../suspense-monaco-editor';
 
 interface Props<T> {
     type?: string;
@@ -22,11 +23,16 @@ export const ObjectEditor = <T extends any>({type, value, buttons, onChange}: Pr
     const [error, setError] = useState<Error>();
     const [lang, setLang] = useState<string>(storage.getItem('lang', defaultLang));
     const [text, setText] = useState<string>(stringify(value, lang));
+    const editor = useRef<MonacoEditor>(null);
 
     useEffect(() => storage.setItem('lang', lang, defaultLang), [lang]);
     useEffect(() => setText(stringify(value, lang)), [value]);
     useEffect(() => setText(stringify(parse(text), lang)), [lang]);
     useEffect(() => {
+        if (!editor.current) {
+            return;
+        }
+
         // we ONLY want to change the text, if the normalized version has changed, this prevents white-space changes
         // from resulting in a significant change
         const editorText = stringify(parse(editor.current.editor.getValue()), lang);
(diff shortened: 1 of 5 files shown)


ui/src/app/shared/components/suspense-monaco-editor.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import {MonacoEditorProps} from 'react-monaco-editor';
+import MonacoEditor from 'react-monaco-editor';
```
agilgur5, line 3:
```suggestion
import type {MonacoEditorProps} from 'react-monaco-editor';
import type MonacoEditor from 'react-monaco-editor';
```

I originally made these explicit `import type`s, but that made `tslint` error out as it doesn't understand the syntax. Need to migrate to ESLint (as `tslint` has long been deprecated)

ui/src/app/shared/components/object-editor/object-editor.tsx
```diff
@@ -1,12 +1,13 @@
-import {languages} from 'monaco-editor/esm/vs/editor/editor.api';
 import * as React from 'react';
-import {createRef, useEffect, useState} from 'react';
+import {useEffect, useRef, useState} from 'react';
 import MonacoEditor from 'react-monaco-editor';
```
agilgur5, line 3:
```suggestion
import {useEffect, useRef, useState} from 'react';
import type MonacoEditor from 'react-monaco-editor';
```

same as above comment, `import type` unfortunately fails due to `tslint`

ui/src/app/shared/components/suspense-monaco-editor.tsx
```diff
@@ -0,0 +1,22 @@
...
+import {Loading} from './loading';
+
+// lazy load Monaco Editor as it is a gigantic component (which can be split into a separate bundle)
+const LazyMonacoEditor = React.lazy(() => {
+    return import(/* webpackChunkName: "react-monaco-editor" */ 'react-monaco-editor');
+});
+
+// workaround, react-monaco-editor's own default no-op seems to fail when lazy loaded, causing a crash when unmounted
+// react-monaco-editor's default no-op: https://github.com/react-monaco-editor/react-monaco-editor/blob/7e5a4938cd328bf95ebc1288967f2037c6023b5a/src/editor.tsx#L184
+const noop = () => {}; // tslint:disable-line:no-empty
```
toyamagu-2021, line 14: I've reproduced this statement.

toyamagu-2021 (approved):

ui/src/app/shared/components/suspense-monaco-editor.tsx
agilgur5: Filed an issue for it: https://github.com/react-monaco-editor/react-monaco-editor/issues/852. Workaround is trivial though, so I don't plan to investigate that much

ui/src/app/shared/components/suspense-monaco-editor.tsx
agilgur5: Migrated to ESLint in #12163. As mentioned there, there's still a syntax reading error due to the old version of Prettier. But #12163 unblocks upgrading Prettier, which should be a lot more straightforward.

And _then_ we can start using `import type` 😅

terrytangyuan (approved):

ui/src/app/shared/components/suspense-monaco-editor.tsx
agilgur5: Fixed in #12514

ui/src/app/shared/components/object-editor/object-editor.tsx
agilgur5: Fixed in #12514

PR #12158, opened 2023-11-07
refactor(ui): code-split out large `xterm` dep

agilgur5:
Partial fix for #12059 

### Motivation

After the past few code-splits (#12150, #12097, #12061) `xterm` was left as one of the largest deps in the main bundle
- `xterm` is only used by `argo-ui` in its `LogsViewer` component, which is only used by Workflows' `FullHeightLogsViewer` component
- shave off more than half a MB from the main bundle:
  - from `xterm` itself: ~550KB / ~380KB minified / 84KB gzipped from main bundle
  - from `argo-ui`: ~190KB / ~60KB minified / ~17KB gzipped from main bundle
    - part of this went into the `xterm` bundle though, so there is some overlap
  - main bundle itself is now only 3.26MB / 1.64MB minified / ~430KB gzipped
  
### Modifications

- lazy load `LogsViewer` within `FullHeightLogsViewer` in order to code-split it
  - add a [`webpackPrefetch`](https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules) magic comment as logs are _very_ commonly used, at least in my experience
  
- import all `argo-ui` components individually, via `argo-ui/src/components/*`, instead of together via `argo-ui` or `argo-ui/src/index`
  - this makes tree-shaking `argo-ui` actually possible. the overall `argo-ui` import as well as each component imports styles, which are side-effects that make tree-shaking not possible
    - in particular, [this SCSS import](https://github.com/argoproj/argo-ui/blob/5ff344ac9692c14dd108468bd3c020c3c75181cb/src/components/index.ts#L1) made the main import not tree-shakeable
  - this also removed some `argo-ui` components from the bundle that are not used by Workflows

### Verification

UI seems to work as before. Log viewing works as before. <details><summary>Screenshot: </summary>

Screenshot 2024-05-14 at 2 41 04 PM

</details>

#### Bundle analysis

<details><summary>Before:</summary>

Screenshot 2024-05-14 at 2 51 42 PM
Screenshot 2024-05-14 at 2 51 54 PM

</details>

<details><summary>After:</summary>

Screenshot 2024-05-14 at 2 48 41 PM
Screenshot 2024-05-14 at 2 49 32 PM
Screenshot 2024-05-14 at 2 49 15 PM

</details>

### Notes to Reviewers

1. The main change is to [`full-height-logs-viewer.tsx`](https://github.com/argoproj/argo-workflows/pull/12158/files#diff-2954852619de80fb686f957e28ae3cad3090e2a0206e58387b3dc7a273702051), the rest is just import changes

1. I thought of doing all the tree-shaking within a shim import, e.g. `./argo-ui-shim.ts`, but then anything that imports that would get everything that the shim imports (due to its CSS import side-effects). Importing components individually where they're used allows for _more_ splitting, particularly when more pages are split per #12059

### DIFF
--- a/ui/src/app/api-docs/api-docs.tsx
+++ b/ui/src/app/api-docs/api-docs.tsx
@@ -1,4 +1,4 @@
-import {Page} from 'argo-ui';
+import {Page} from 'argo-ui/src/components/page/page';
 import * as React from 'react';
 
 import {uiUrl} from '../shared/base';
--- a/ui/src/app/app-router.tsx
+++ b/ui/src/app/app-router.tsx
@@ -1,5 +1,9 @@
-import {Layout, Notifications, NotificationsManager, NotificationType, Popup, PopupManager, PopupProps} from 'argo-ui';
 import * as H from 'history';
+import {Layout} from 'argo-ui/src/components/layout/layout';
+import {NotificationsManager} from 'argo-ui/src/components/notifications/notification-manager';
+import {Notifications, NotificationType} from 'argo-ui/src/components/notifications/notifications';
+import {PopupManager} from 'argo-ui/src/components/popup/popup-manager';
+import {Popup, PopupProps} from 'argo-ui/src/components/popup/popup';
 
 import * as React from 'react';
 import {useEffect, useState} from 'react';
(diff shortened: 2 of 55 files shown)


agilgur5:
Merged `main` for new PR checks per #13027

tczhao (approved):
LGTM. It was a good read, I learned quite a bit.

PR #12196, opened 2023-11-14
fix: delete pending pod when workflow terminated 

shuangkun:
When I submit a workflow. It didn't work successed for some reason(such as no stock), some of pod is  still Pending.
So I terminate the workflow.  But the pending pod was not deleted.

After awhile, the pending pod is running , but the workflow has failed and caused some waste of resources.

So, we need to delete pending pod when workflow terminated.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -525,7 +525,10 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
 		case terminateContainers:
-			if terminationGracePeriod, err := wfc.signalContainers(namespace, podName, syscall.SIGTERM); err != nil {
+			pod, err := wfc.getPod(namespace, podName)
+			if err == nil && pod != nil && pod.Status.Phase == apiv1.PodPending {
+				wfc.queuePodForCleanup(namespace, podName, deletePod)
+			} else if terminationGracePeriod, err := wfc.signalContainers(namespace, podName, syscall.SIGTERM); err != nil {
 				return err
 			} else if terminationGracePeriod > 0 {
 				wfc.queuePodForCleanupAfter(namespace, podName, killContainers, terminationGracePeriod)
--- a/workflow/controller/controller_test.go
+++ b/workflow/controller/controller_test.go
@@ -1069,3 +1069,37 @@ spec:
 	podCleanupKey := "test/my-wf/labelPodCompleted"
 	assert.Equal(t, 0, controller.podCleanupQueue.NumRequeues(podCleanupKey))
 }
+
+func TestPodCleanupDeletePendingPodWhenTerminate(t *testing.T) {
+	wf := wfv1.MustUnmarshalWorkflow(`
+metadata:
+  name: my-wf
+  namespace: test
+spec:
+  entrypoint: main
+  templates:
+    - name: main
+      container:
+        image: my-image
+  `)
+	cancel, controller := newController(wf)
+	defer cancel()
+
+	ctx := context.Background()
+	assert.True(t, controller.processNextItem(ctx))
+
+	woc := newWorkflowOperationCtx(wf, controller)
+	woc.operate(ctx)
+	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
+	makePodsPhase(ctx, woc, apiv1.PodPending)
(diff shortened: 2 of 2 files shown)


workflow/controller/controller.go
```diff
@@ -512,7 +512,18 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
 		case terminateContainers:
-			if terminationGracePeriod, err := wfc.signalContainers(namespace, podName, syscall.SIGTERM); err != nil {
+			pod, err := wfc.getPod(namespace, podName)
+			if err == nil && pod != nil && pod.Status.Phase == "Pending" {
+				logCtx.Info("deleting pending pod")
```
Joibel, line 517: Please could you add `.WithFields(log.Fields{"name": podName, "namespace": namespace})` to the logging.

Joibel (changes requested):
I think I agree with the overall intention here, it seems reasonable.

It's a gap which isn't caught by any existing mechanism.

workflow/controller/controller.go
shuangkun: ok，I will add it.

Joibel (approved):
Looks good, thank you for adding this. You'll need to fix the linter error before this can be merged though.

workflow/controller/controller.go
```diff
@@ -512,7 +512,18 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
 		case terminateContainers:
-			if terminationGracePeriod, err := wfc.signalContainers(namespace, podName, syscall.SIGTERM); err != nil {
+			pod, err := wfc.getPod(namespace, podName)
+			if err == nil && pod != nil && pod.Status.Phase == "Pending" {
+				logCtx.WithFields(log.Fields{"name": podName, "namespace": namespace}).Info("deleting pending pod")
+				propagation := metav1.DeletePropagationBackground
+				err = pods.Delete(ctx, podName, metav1.DeleteOptions{
```
sarabala1979, line 519: Can we use the Pod cleanup queue delete it?

sarabala1979:
Thanks @Joibel for reviewing.

workflow/controller/controller.go
Joibel: Ah, not something I've come across yet. It sounds cleaner.

shuangkun: > Can we use the Pod cleanup queue to delete it?
OK, I have requeue it to cleanup queue to delete it

shuangkun:
> Looks good, thank you for adding this. You'll need to fix the linter error before this can be merged though.



> Looks good, thank you for adding this. You'll need to fix the linter error before this can be merged though.

Do you know how should i do to fix the linter error? I really didn't chang something about it.

Joibel (approved):
Lovely, thank @shuangkun

sarabala1979:
@shuangkun can you add tests (unit + e2e) ?

shuangkun:
> @shuangkun can you add tests (unit + e2e) ?

I have tested it in my environment，I will try to add some unit tests. This scene is slightly more difficult to construct.

shuangkun:
> @shuangkun can you add tests (unit + e2e) ?

I have add a ut. Pealease have a check.

Joibel:
LGTM, thanks @shuangkun. @sarabala1979, could you re-review and consider merging

workflow/controller/controller.go
```diff
@@ -525,7 +525,10 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
 		case terminateContainers:
-			if terminationGracePeriod, err := wfc.signalContainers(namespace, podName, syscall.SIGTERM); err != nil {
+			pod, err := wfc.getPod(namespace, podName)
+			if err == nil && pod != nil && pod.Status.Phase == "Pending" {
```
terrytangyuan, line 529: Use the constant from common.go?

shuangkun: OK, I use the constant.

shuangkun: > Use the constant from common.go?

Could you have a check, thanks!

Joibel: LGTM. PTAL @terrytangyuan

sarabala1979 (approved):
Thanks for your contribution. LGTM. Please update branch

PR #12197, opened 2023-11-14
fix: return failed instead of success when no container status

shuangkun:
We have a product named ManagedArgoWorkflows in AlibabaCloud use eci (virtual-kubelet).
Sometimes the pod failed for some reason (no stock、timeout、schedulerFailed and more than these)，in these conditions, the pod mightly not has containerstatus/condations/messages,  and then the pod will marked success and caused mistake.

So I think we should return failed when the pod is Failed and we can't find message and status in the pod.

Maybe this problem will also occur in other public cloud platforms.

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1555,6 +1555,10 @@ func (woc *wfOperationCtx) inferFailedReason(pod *apiv1.Pod, tmpl *wfv1.Template
 	sort.Slice(ctrs, func(i, j int) bool { return order(ctrs[i].Name) < order(ctrs[j].Name) })
 	// Init containers have the highest preferences over other containers.
 	ctrs = append(pod.Status.InitContainerStatuses, ctrs...)
+	// When there isn't any containstatus (such as no stock in public cloud), return Failure.
+	if len(ctrs) == 0 {
+		return wfv1.NodeFailed, fmt.Sprintf("can't find failed message for pod %s namespace %s", pod.Name, pod.Namespace)
+	}
 
 	for _, ctr := range ctrs {
 


Joibel:
This feels like a very dangerous assumption to make. How do we know that the status won't get updated later to something more successful?

shuangkun:
> This feels like a very dangerous assumption to make. How do we know that the status won't get updated later to something more successful?

I think if it get update later from failed to successeed.  the workflow will reconicle. if the pod failed, it may be retry. if it marked successed(but actually never run )，we made a mistake. This is what I encountered

workflow/controller/operator.go
```diff
@@ -1524,6 +1524,10 @@ func (woc *wfOperationCtx) inferFailedReason(pod *apiv1.Pod, tmpl *wfv1.Template
 	sort.Slice(ctrs, func(i, j int) bool { return order(ctrs[i].Name) < order(ctrs[j].Name) })
 	// Init containers have the highest preferences over other containers.
 	ctrs = append(pod.Status.InitContainerStatuses, ctrs...)
+	// When there isn't any containstatus (such as no stock in public cloud), return Failure.
+	if len(ctrs) == 0 {
+		return wfv1.NodeFailed, ""
```
Joibel, line 1529: We should fill out the message (second return parameter) here, to pass something to the user. Otherwise they're going to struggle to determine why their node has failed.

Joibel (changes requested):
Ok, I follow your reasoning now and why this is a reasonable change.

workflow/controller/operator.go
shuangkun: ok, I will fill out the message.

juliev0 (approved):

PR #12198, opened 2023-11-14
fix: completed workflow tracking

Joibel:
Problem: The workflow cache can have out of date information in it which will resurrect a completed workflow. In this scenario, with an on-exit handler with `when: "{{workflow.status}} != Succeeded"` it may fire the exit handler erroneously.

Firing enough of this workflow into a real cluster fast enough will reproduce this - tested in 3.4.8, 3.4.9, and master
```
kind: Workflow
metadata:
  generateName: exit-
spec:
  entrypoint: entrypoint
  onExit: exit-handler
  templates:
    - name: entrypoint
      container:
        image: quay.io/pch/whalesay:latest
        command: [ cowsay ]
        args: [ "hello world" ]
        resources:
          limits:
            memory: 32Mi
            cpu: 100m
    - name: ohnoes
      container:
        image: quay.io/pch/whalesay:latest
        command: [ cowsay ]
        args: [ "oh noes" ]
        resources:
          limits:
            memory: 32Mi
            cpu: 100m
    - name: exit-handler
      steps:
        - - name: ohnoes
            template: ohnoes
```

It does not reproduce easily on a toy cluster in k3d. I believe because API latencies are too small and the whole machine starts to fall apart before the problem happens.

It happens rarely in real workloads but if `ohnoes` is replaced by a notification you'll receive a notification that your workflow has failed, even though it has succeeded when you go and look at it.

Story: The workflow controller is busy, and the k8s API is busy. The k8s API is providing workflow updates which run through the workflow informer cache. Normally these cannot go 'backwards'.

Backwards would be where an update to the workflow takes our view of the workflow back to an earlier point in time because we've received a high latency update from the k8s API. We prevent many of these using the UpdateFunc handler which stops us reconciling at the point we've already seen.

The problem is that as a workflow completes we drop it from the informer cache (it no longer meets `reconciliationNeeded`). This allows it to come back into the cache in an old form as a new resource.

If this happens we attempt to reconcile it. In the event the pod representing the node that was running in the form of the reappearing workflow doesn't exist we're going to (correctly) treat the workflow as in error.

So we have a workflow which is in error exiting, and test the onExit `when` clause, decide it's true, and launch the pod to handle it (ohnoes in my example). This runs, regardless of the fact that the workflow controller then discovers that the workflow resourceVersion is out of date and cannot write the current state back to the cluster.

Fixes considered:
* Attempt to ensure we're up to date before actioning the onExit. This seemed like a very convoluted code path for doing special onExit handling and likely to introduce more bugs than it fixed. This would either put much more load on the API by causing the controller to get the latest version, or attempt the write back to see if we were out of date.
* Attempt to prevent old version re-entry in the cache rather than at the reconcile stage. I tried this, but we still have a hole where it'll make it through.

So the fix is to separately track workflows that have been evicted from the cache and not process them at all. This cache has a lifetime of 10 minutes as I could get up to latencies of a few minutes during heavy workload testing.

### Verification

Tested on EKS with 3 4-core nodes and the above workflow injected rapidly.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -77,7 +77,7 @@ type recentlyCompletedWorkflow struct {
 
 type recentCompletions struct {
 	completions []recentlyCompletedWorkflow
-	mutex       gosync.Mutex
+	mutex       gosync.RWMutex
 }
 
 // WorkflowController is the controller for workflow resources
@@ -761,7 +761,7 @@ func (wfc *WorkflowController) processNextItem(ctx context.Context) bool {
 		return true
 	}
 
-	if wfc.checkRecentlyCompleted(wf.ObjectMeta.Name) {
+	if wf.Status.Phase != "" && wfc.checkRecentlyCompleted(wf.ObjectMeta.Name) {
 		log.WithFields(log.Fields{"name": wf.ObjectMeta.Name}).Warn("Cache: Rejecting recently deleted")
 		return true
 	}
@@ -869,6 +869,8 @@ func getWfPriority(obj interface{}) (int32, time.Time) {
 // 10 minutes in the past
 const maxCompletedStoreTime = time.Second * -600
 
+// This is a helper function for expiring old records of workflows
+// completed more than maxCompletedStoreTime ago
 func (wfc *WorkflowController) cleanCompletedWorkflowsRecord() {
 	cutoff := time.Now().Add(maxCompletedStoreTime)
 	removeIndex := -1
@@ -886,35 +888,43 @@ func (wfc *WorkflowController) cleanCompletedWorkflowsRecord() {
 	}
 }
 
+// Records a workflow as recently completed in the list
+// if it isn't already in the list
 func (wfc *WorkflowController) recordCompletedWorkflow(key string) {
 	if !wfc.checkRecentlyCompleted(key) {
 		wfc.recentCompletions.mutex.Lock()
+		defer wfc.recentCompletions.mutex.Unlock()
 		wfc.recentCompletions.completions = append(wfc.recentCompletions.completions,
 			recentlyCompletedWorkflow{
 				key:  key,
 				when: time.Now(),
 			})
-		wfc.recentCompletions.mutex.Unlock()
 	}
 }
 
+// Returns true if the workflow given by key is in the recently completed
+// list. Will perform expiry cleanup before checking.
 func (wfc *WorkflowController) checkRecentlyCompleted(key string) bool {
 	wfc.cleanCompletedWorkflowsRecord()
 	recent := false
-	wfc.recentCompletions.mutex.Lock()
+	wfc.recentCompletions.mutex.RLock()
+	defer wfc.recentCompletions.mutex.RUnlock()
 	for _, val := range wfc.recentCompletions.completions {
 		if val.key == key {
 			recent = true
 			break
 		}
 	}
-	wfc.recentCompletions.mutex.Unlock()
 	return recent
 }
 
 func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context) {
 	wfc.wfInformer.AddEventHandler(
 		cache.FilteringResourceEventHandler{
+			// FilterFunc is called for every operation affecting the
+			// informer cache and can be used to reject things from
+			// the cache. When they are rejected (this returns false)
+			// they will be deleted.
 			FilterFunc: func(obj interface{}) bool {
 				un, ok := obj.(*unstructured.Unstructured)
 				if !ok {
@@ -925,11 +935,12 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context) 
 				if !needed {
 					key, _ := cache.MetaNamespaceKeyFunc(un)
 					wfc.recordCompletedWorkflow(key)
-					wfc.throttler.Remove(key)
 				}
 				return needed
 			},
 			Handler: cache.ResourceEventHandlerFuncs{
+				// This function is called when a new to the informer object
+				// is to be added to the informer
 				AddFunc: func(obj interface{}) {
 					key, err := cache.MetaNamespaceKeyFunc(obj)
 					if err == nil {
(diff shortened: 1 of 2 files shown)


workflow/controller/controller.go
```diff
@@ -857,7 +921,13 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
 					log.Warnf("Workflow FilterFunc: '%v' is not an unstructured", obj)
 					return false
 				}
-				return reconciliationNeeded(un)
+				needed := reconciliationNeeded(un)
+				if !needed {
+					key, _ := cache.MetaNamespaceKeyFunc(un)
+					wfc.recordCompletedWorkflow(key)
+					wfc.throttler.Remove(key)
+				}
```
juliev0, line 929: I notice `wfc.releaseAllWorkflowLocks(obj)` in the DeleteFunc. Do we not need that here too? If we do, can we consolidate the shared code into a single method?

juliev0: Also, can we add comments here? (I for one had no idea that the FilterFunc actually purges from the Informer cache, which should have been commented before you arrived I think)

workflow/controller/controller.go
```diff
@@ -848,6 +866,52 @@ func getWfPriority(obj interface{}) (int32, time.Time) {
...
+	recent := false
+	wfc.recentCompletions.mutex.Lock()
+	for _, val := range wfc.recentCompletions.completions {
+		if val.key == key {
+			recent = true
+			break
+		}
+	}
+	wfc.recentCompletions.mutex.Unlock()
+	return recent
```
juliev0, line 912: can we use an rwMutex?

workflow/controller/controller.go
```diff
@@ -69,6 +70,16 @@ import (
...
 
+type recentlyCompletedWorkflow struct {
+	key  string
+	when time.Time
+}
+
+type recentCompletions struct {
+	completions []recentlyCompletedWorkflow
+	mutex       gosync.Mutex
+}
```
juliev0, line 81: can we use an rwMutex?

workflow/controller/controller.go
```diff
@@ -848,6 +866,52 @@ func getWfPriority(obj interface{}) (int32, time.Time) {
...
+func (wfc *WorkflowController) checkRecentlyCompleted(key string) bool {
+	wfc.cleanCompletedWorkflowsRecord()
+	recent := false
+	wfc.recentCompletions.mutex.Lock()
+	for _, val := range wfc.recentCompletions.completions {
+		if val.key == key {
+			recent = true
+			break
+		}
+	}
```
juliev0, line 910: thinking about how we have an O(n) operation for each workflow reconciliation. Maybe it's not a big deal. I guess n could be however many workflows would have completed in the last 10 minutes, which may not be terrible. I know you probably made it a slice because you can quickly delete the old ones which is nice. Theoretically, you could actually store a `map[string]completedWorfkflow`  alongside your slice in `recentCompletions`. 
When you delete from the slice, you have the key in your `recentlyCompletedWorkflow` to delete from the map.

workflow/controller/controller.go
```diff
@@ -848,6 +866,52 @@ func getWfPriority(obj interface{}) (int32, time.Time) {
...
+	wfc.recentCompletions.mutex.Unlock()
+	return recent
+}
```
juliev0, line 913: can we add a few comments above these functions so people know what they're for?

juliev0:
Oh, just thinking...the key is the Workflow name, right? So, what happens when somebody legitimately creates a new Workflow of the same name as the previous Workflow within 10 minutes of the previous one being deleted?

workflow/controller/controller.go
Joibel: Yes, that's a good call. I'll do that.

workflow/controller/controller.go
Joibel: The slice can be mass deleted, which was why I did it that way. A map would require a one by one deletion, and insertion is more expensive.
I don't think keeping both is a good idea as we have most of the downsides of both in terms of costs. If we want to go to a map, I'd turn GC into a separate go background task which runs around the map handling a few at a time then releasing the mutex before grabbing it again. This is more complex, but perhaps better. Happy to switch if you think it's important.

workflow/controller/controller.go
Joibel: I don't think we do, as I'm imagining that FilterFunc returning false then does a Delete, which should call DeleteFunc.

workflow/controller/controller.go
Joibel: Ok, will do.

Joibel:
> Oh, just thinking...the key is the Workflow name, right? So, what happens when somebody legitimately creates a new Workflow of the same name as the previous Workflow within 10 minutes of the previous one being deleted?

I thought I'd handled this, but obviously not. This went through many iterations, so I'm blaming that!

I've added a check when we may reject to ensure that the Phase is not "". This will allow brand new workflows back in, but any that have already been handled by the controller will go through potential rejection.

workflow/controller/controller.go
juliev0: okay, thanks for the analysis. I'm fine with it as is.

workflow/controller/controller.go
juliev0: True. In that case, why include `wfc.throttler.Remove(key)` in both places then?

workflow/controller/controller.go
isubasinghe: For what it's worth, I think this O(n) still should be plenty fast relative to the rest of workflows even for largish slices (say 1M)

workflow/controller/controller.go
isubasinghe: As long as the workflow didn't run, we shouldn't need to release any locks is my understanding based upon my previous work on the syncManager. @Joibel is correct I believe, but also this means the wfc.throttler.Remove(key) is not needed since the DeleteFunc will get cold

isubasinghe:
> Oh, just thinking...the key is the Workflow name, right? So, what happens when somebody legitimately creates a new Workflow of the same name as the previous Workflow within 10 minutes of the previous one being deleted?

Nice catch! I didn't think about this myself

isubasinghe (commented):
Generally looks good, but I couldn't find the change for:

> I've added a check when we may reject to ensure that the Phase is not "". This will allow brand new workflows back in, but any that have already been handled by the controller will go through potential rejection.

There also seems to be a redundant throttler.Remove(), after those changes I would be happy to approve. 

What this issue opens up though is the question of where else are there similar bugs in the code? Would this be the only interaction with the k8s API where we need to care about this, or is this a bug that would impact other things as well,say for example configmaps (I remember seeing a ConfigMap cache somewhere) ?

isubasinghe (approved):
LGTM

juliev0 (approved):

PR #12199, opened 2023-11-14
feat(server): Support supplying a list of headers when keying IPs for rate limiting

terrytangyuan:
We are currently [using an upstream rate limiter library](https://github.com/argoproj/argo-workflows/blob/1f6b19f3ab9f8758684bb6c93289d57c5dd1d963/server/apiserver/argoserver.go#L326) and should consider adding support for supplying a list of headers to look at instead of only extracting from source IPs. This can avoid some rate limiting issues.


agilgur5 (commented):
Just a quick glance from my phone, some small comments. There's also a merge conflict.

Need to actually read up on the keyfunc before I can approve

terrytangyuan:
Thanks. Resolved conflicts.

server/apiserver/argoserver.go
```diff
@@ -325,16 +326,21 @@ func (as *argoServer) newGRPCServer(instanceIDService instanceid.Service, offloa
 // using grpc-gateway as a proxy to the gRPC server.
 func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServer *artifacts.ArtifactServer) *http.Server {
 	endpoint := fmt.Sprintf("localhost:%d", port)
+	ipKeyFunc := httplimit.IPKeyFunc()
+	if ipKeyFuncHeadersStr := env.GetString("IP_KEY_FUNC_HEADERS", ""); ipKeyFuncHeadersStr != "" {
+		ipKeyFuncHeaders := strings.Split(ipKeyFuncHeadersStr, ",")
+		ipKeyFunc = httplimit.IPKeyFunc(ipKeyFuncHeaders...)
```
isubasinghe, line 332: This IPKeyFunc(https://github.com/sethvargo/go-limiter/blob/v0.7.2/httplimit/middleware.go#L45) will return errors if the ip is not valid. 

I am okay with this behaviour because it's the users fault for supplying bad IPs. 

Just wanted to point this out if you think this is acceptable or not.

isubasinghe (approved):
LGTM made a small comment about possible behaviour on bad IP, imo sort of a user issue in my eyes so happy to go ahead with this one.

terrytangyuan:
> LGTM made a small comment about possible behaviour on bad IP, imo sort of a user issue in my eyes so happy to go ahead with this one. 

Yeah that's fine IMO.

terrytangyuan:
@agilgur5 Would you like to take another look as well before I merge?

server/apiserver/argoserver.go
agilgur5: the `IPKeyFunc` was already used before as well, just with no arguments passed in, so we've already been returning errors in that case. haven't really seen problems with it, so I'm inclined to not change that

> I am okay with this behaviour because it's the users fault for supplying bad IPs.

so that depends on who adds the header -- usually it's a proxy, client, etc. It's not necessarily the "user" in that sense.

but that is generally an input problem / bug in the user's software and I do think it's ok to continue erroring in that case.
our other options are to return a more specific error or to bucket all IP address parse failures together into one bucket for the purpose of rate limitng

docs/environment-variables.md
```diff
@@ -122,6 +122,7 @@ data:
 | `DISABLE_VALUE_LIST_RETRIEVAL_KEY_PATTERN` | `string` | `""`    | Disable the retrieval of the list of label values for keys based on this regular expression.                            |
 | `FIRST_TIME_USER_MODAL`                    | `bool`   | `true`  | Show this modal.                                                                                                        |
 | `FEEDBACK_MODAL`                           | `bool`   | `true`  | Show this modal.                                                                                                        |
+| `IP_KEY_FUNC_HEADERS`                      | `string` | `""`    | List of comma separated headers to key requests for rate limiting. By default, uses hosts in remote addresses.          |
```
agilgur5, line 125:
```suggestion
| `IP_KEY_FUNC_HEADERS`                      | `string` | `""`    | List of comma separated request headers containing IPs to use for rate limiting. For example, "X-Forwarded-For,X-Real-IP". By default, uses the request's remote IP address.          |
```

Added some more details, specificity, and an example

agilgur5 (commented):
Improved the docs a little bit with an example etc. I was unsure myself exactly how to use it, so I think this would be helpful.

LGTM otherwise

terrytangyuan:
Ping @sarabala1979 @juliev0 for approval/merge.

agilgur5 (approved):

PR #12130, opened 2023-11-03
fix: regression in memoization without outputs

Joibel:
The merge with #11451 reverted this, so this commit is just to reinstate that.
The tests included in #11379 failed to catch this, I've raised #12129 for this, but in the interests of matching the documentation and kubecon next week I'm putting this PR in now.

Fixes #12117

PR #12201, opened 2023-11-15
fix: Revert "fix: regression in memoization without outputs (#12130)"

terrytangyuan:
This reverts commit c751e66050a7c16d84ecdfdfe0f7ef59d6ec84c0.

See https://github.com/argoproj/argo-workflows/issues/12165

### DIFF
--- a/workflow/controller/steps.go
+++ b/workflow/controller/steps.go
@@ -170,19 +170,21 @@ func (woc *wfOperationCtx) executeSteps(ctx context.Context, nodeName string, tm
 	if err != nil {
 		return node, err
 	}
-
 	if outputs != nil {
+		node, err := woc.wf.GetNodeByName(nodeName)
+		if err != nil {
+			return nil, err
+		}
 		node.Outputs = outputs
 		woc.addOutputsToGlobalScope(node.Outputs)
 		woc.wf.Status.Nodes.Set(node.ID, *node)
-	}
-
-	if node.MemoizationStatus != nil {
-		c := woc.controller.cacheFactory.GetCache(controllercache.ConfigMapCache, node.MemoizationStatus.CacheName)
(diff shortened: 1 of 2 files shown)


workflow/controller/steps.go
```diff
@@ -170,19 +170,21 @@ func (woc *wfOperationCtx) executeSteps(ctx context.Context, nodeName string, tm
 	if err != nil {
 		return node, err
 	}
-
 	if outputs != nil {
+		node, err := woc.wf.GetNodeByName(nodeName)
+		if err != nil {
+			return nil, err
+		}
```
agilgur5, line 177: per my root cause analysis, we can have a more targeted fix by just reverting this part (in both steps and DAGs), so that we retain the memoization fix below. then I can solve the root cause in a follow-up PR

Joibel: I've done this in #12203

#12130 should not have gone in to 3.4 branch at all (as memoization without outputs was a feat), so this revert should be done on 3.4, but #12203 should be in 3.5.

agilgur5:
@terrytangyuan do you want to re-open this and change the base branch to the 3.4.x branch/patch series as @Joibel suggested [above](https://github.com/argoproj/argo-workflows/pull/12201#discussion_r1393977552)?

terrytangyuan:
Done. @agilgur5 @Joibel

agilgur5 (approved):

agilgur5:
> This reverts commit [c751e66](https://github.com/argoproj/argo-workflows/commit/c751e66050a7c16d84ecdfdfe0f7ef59d6ec84c0).

https://github.com/argoproj/argo-workflows/commit/bdc1b25900f44c194ab36d202821cec01ba96a73 now from the 3.4.13 release (#12203 is for 3.5.1 partial revert)

PR #12208, opened 2023-11-15
fix: interface conversion: interface {} is *unstructured.Unstructured, not *v1alpha1.Workflow

woehrl01:
Fixes crash loop caused by the health check in v3.5.1

Bildschirmfoto 2023-11-15 um 20 19 20

### Modifications

I safeguarded the type conversion.

### DIFF
--- a/pkg/client/listers/workflow/v1alpha1/workflow.go
+++ b/pkg/client/listers/workflow/v1alpha1/workflow.go
@@ -65,7 +65,9 @@ type workflowNamespaceLister struct {
 // List lists all Workflows in the indexer for a given namespace.
 func (s workflowNamespaceLister) List(selector labels.Selector) (ret []*v1alpha1.Workflow, err error) {
 	err = cache.ListAllByNamespace(s.indexer, s.namespace, selector, func(m interface{}) {
-		ret = append(ret, m.(*v1alpha1.Workflow))
+		if _, ok := m.(*v1alpha1.Workflow); ok {
+			ret = append(ret, m.(*v1alpha1.Workflow))
+		}
 	})
 	return ret, err
 }


terrytangyuan (commented):
This file is autogenerated by codegen.

woehrl01:
@terrytangyuan Absolutely, but I'm not sure what I should add more to the PR than fixing the attached error, referenced in the screenshot. I now have noticed that this is a generated file. Do you have any suggestions how to fix this better?

Joibel:
@woehrl01 - do you have an invalid workflow in your cluster. Are your CRD definitions in cluster up to date with the workflow-controller version? I'm not sure that would help if the workflow was uploaded prior to the CRD update though.

woehrl01:
- [ ] @Joibel no, I already cleared all workflows multiple times, crd are in 3.5.1 same as the controller. All the new workflows are creates by the same workflow templates.

Joibel:
Is this a set of workflows that worked under an earlier version? Any chance of attempting to find the workflow that's causing this?

woehrl01:
@Joibel I don't know because, we are starting to use it more intense since 3.5.1. I could definitely give it a try to find the workflow, do you have any ideas how I can find the affected workflow?

Joibel:
I don't know how many workflows were talking about. Can you introduce them to another workflow-controller running in a separate cluster (e.g. k3d/minikube or whatever locally) one at a time until you find the one that crashes it. Or use a bisect on the full list to narrow it down. You don't actually need the workflows to be able to work at all.

Otherwise you could build yourself a controller which can spit out just the metadata.name for each workflow it's about to list, but that's more hassle.

sarabala1979:
@woehrl01  how will workflow get created? Are you using CLI/API? Will minimal CRD created on cluster?
Looks like one or some workflows are malformed.
you can try with Argo CLI to list the workflows to see if any workflows are missing.

tzahush:
Hey guys, I'm getting this as well, are we planning on introducing this fix?
Regardless of what is the issue, its probably still a good solution to add this validation.

agilgur5:
I believe this was fixed by #12353 which changed the source code

PR #12203, opened 2023-11-15
fix: Fix for missing steps in the UI

Joibel:
This is a partial revert of #12130 and fixes #12165

Offered as an alternative to a full revert.

Fixes #12165 after @agilgur5's investigations

### Verification

The ReproduceExample.zip from #12165 goes from 

image

to

image

PR #12214, opened 2023-11-17
test: add test for missing child node

Joibel:
This test is providing a regression test for #12165. As promised in #12203.

It verifies that a child link is correctly made in the case of a step with outputs.

### Verification

This test fails if #12203 is reverted.

### DIFF
--- a/test/e2e/functional/missing-steps.yaml
+++ b/test/e2e/functional/missing-steps.yaml
@@ -1,0 +1,72 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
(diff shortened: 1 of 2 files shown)


terrytangyuan:
Consider merging multiple PRs into one?

Joibel:
> Consider merging multiple PRs into one?

It's not usually good practice, why would you prefer it?

terrytangyuan:
Just reduces the review burden. Currently, you have like 3 PRs for 3 different tests. Just a personal preference for changes like these

terrytangyuan (approved):

PR #12234, opened 2023-11-20
ci: Apply yarn-deduplicate to dependabot tasks

Yaminyam:
### Motivation

For dependabot pr we are currently accepting and merging automatically.
However, there are cases where the yarn-deduplicate operation is not reflected and cannot be merged automatically.
https://github.com/argoproj/argo-workflows/pull/12183

### Modifications

For dependabot pr, run yarn --cwd ui deduplicate and commit and push the changes.


terrytangyuan (commented):
cc @agilgur5 to take a look

Yaminyam:
I think the solution would be to create a bot for signing or create a commit by the maintainer of the repo.

agilgur5:
> However, there are cases where the yarn-deduplicate operation is not reflected and cannot be merged automatically.

Yea I mentioned this in https://github.com/argoproj/argo-workflows/pull/11728#issuecomment-1707477279 and [a previous](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.6i2mpmot82zm) Contributor Meeting (or two).
I thought updating to newer Yarn might fix this but this seems to still be a problem in newer Yarn versions too: https://github.com/yarnpkg/berry/issues/4976.

There's also this Dependabot issue on the topic: https://github.com/dependabot/dependabot-core/issues/5830

> For dependabot pr, run yarn --cwd ui deduplicate and commit and push the changes.

Thanks for the contribution! I was thinking of adding something similar after seeing comments in the Dependabot issue pointing to https://github.com/foxglove/studio/pull/4736 ([current GHA workflow](https://github.com/foxglove/studio/blob/0977e3f553d0fc7bed19eabfad477536c32c6b96/.github/workflows/dependabot-fix.yml))

.github/workflows/ci-build.yaml
```diff
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
-  contents: read
+  contents: write
```
agilgur5, line 17: Per #12035 and #9552, the top-level permissions of a GHA Workflow should _never_ be changed. It should always be read-only.
Only individual jobs should have their permissions changed.
This change will fail OpenSSF / CLOMonitor checks as is.

Also all permissions should have explicit comments about why they're enabled

.github/workflows/ci-build.yaml
```diff
@@ -401,7 +401,15 @@ jobs:
...
-      - run: git diff --exit-code
+      - if: github.actor == 'dependabot[bot]'
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git add ui/yarn.lock
+          git commit -m 'chore: update yarn.lock'
+          git push
+      - if: github.actor != 'dependabot[bot]'
+        run: git diff --exit-code
```
agilgur5, line 412: this should still run for dependabot as well. it should have no diff after the commit

.github/workflows/ci-build.yaml
```diff
@@ -401,7 +401,15 @@ jobs:
       - run: yarn --cwd ui test
       - run: yarn --cwd ui lint
       - run: yarn --cwd ui deduplicate
-      - run: git diff --exit-code
+      - if: github.actor == 'dependabot[bot]'
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git add ui/yarn.lock
+          git commit -m 'chore: update yarn.lock'
```
agilgur5, line 409:
```suggestion
          git commit -s -m 'chore: deduplicate yarn.lock'
```

be more explicit in the message and add a sign-off

agilgur5: this line might [also need](https://github.com/foxglove/studio/blob/0977e3f553d0fc7bed19eabfad477536c32c6b96/.github/workflows/dependabot-fix.yml#L58) a `--allow-empty` or a conditional to otherwise check for empty

actually a conditional might(?) be necessary, otherwise this could cause an infinite loop as written (new commit -> rebuild -> repeat). maybe not if the `github.actor` check recognizes that commit is separate?

Yaminyam:
@agilgur5 It's great to hear that this was an issue that was already being addressed.
In the case of the attached pr, it operates as an independently executed workflow.
However, I added logic to the existing workflow to make execution simpler.

And it would be great if you could give us your opinion on how to solve the sign-off problem!

.github/workflows/ci-build.yaml
agilgur5: It would also be good to ensure Yarn runs without scripts if we're adding write permissions (similar to [this line](https://github.com/foxglove/studio/blob/0977e3f553d0fc7bed19eabfad477536c32c6b96/.github/workflows/dependabot-fix.yml#L25), although the env var might be different from Yarn v1)

.github/workflows/ci-build.yaml
```diff
@@ -401,7 +401,15 @@ jobs:
       - run: yarn --cwd ui test
       - run: yarn --cwd ui lint
       - run: yarn --cwd ui deduplicate
-      - run: git diff --exit-code
+      - if: github.actor == 'dependabot[bot]'
```
agilgur5, line 404: I was thinking of adding this to the [`dependabot-reviewer`](https://github.com/argoproj/argo-workflows/blob/feaed7ac8cf8a0667fc20b8be5bd590bef4da42a/.github/workflows/dependabot-reviewer.yml) Workflow instead, so all dependabot things are in one place. That may be slightly more complicated though.

agilgur5: This whole block should also have a comment describing it and why it exists

agilgur5 (changes requested):

.github/workflows/ci-build.yaml
Yaminyam: If GITHUB_TOKEN, which is provided by default in actions, is used, github actions are not triggered.
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

.github/workflows/ci-build.yaml
Yaminyam: I thought I should skip this because the current method doesn't trigger the actions again with a new commit made by the bot, but since the script is executed based on the last commit made by the bot, it should work like you said.

.github/workflows/ci-build.yaml
Yaminyam: I thought the bot would need this permission to commit, but I was mistaken.

.github/workflows/ci-build.yaml
Yaminyam: In order for this job to work in dependabot-reviewer, I think it is necessary to run change-files like ci-build workflow to determine whether the ui job is running.
But I think that creates too much complexity.

agilgur5: Mmm not so much `changed-files` as we'd need to run parts of the `ui:` job itself in order to install Node, NPM, and some deps. It would indeed be simpler to leave here 🤔

.github/workflows/ci-build.yaml
```diff
@@ -401,6 +401,14 @@ jobs:
       - run: yarn --cwd ui test
       - run: yarn --cwd ui lint
       - run: yarn --cwd ui deduplicate
+      # Depends deduplicate changes for dependabot-reviewer to work
```
agilgur5, line 404:
```suggestion
      # Deduplicate UI deps for dependabot PRs
```

This isn't needed for `dependabot-reviewer`, just for dependabot. That Workflow will pass, but it won't merge because this step of the build fails

agilgur5 (commented):
Sorry for the delay on this one, I had some family medical emergencies in late November that got me really busy.

I just have one tiny change, otherwise LGTM

agilgur5 (approved):

terrytangyuan (approved):

.github/workflows/ci-build.yaml
agilgur5: > this line might [also need](https://github.com/foxglove/studio/blob/0977e3f553d0fc7bed19eabfad477536c32c6b96/.github/workflows/dependabot-fix.yml#L58) a `--allow-empty` or a conditional to otherwise check for empty

Yea the lack of conditional and `--allow-empty` is causing failures in CI when there are no changes, see https://github.com/argoproj/argo-workflows/pull/12880#issuecomment-2036064662

.github/workflows/ci-build.yaml
```diff
@@ -401,6 +401,14 @@ jobs:
...
       - run: yarn --cwd ui lint
       - run: yarn --cwd ui deduplicate
+      # Deduplicate UI deps for dependabot PRs
+      - if: github.actor == 'dependabot[bot]'
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git add ui/yarn.lock
+          git commit -s -m 'chore: deduplicate yarn.lock'
+          git push
```
agilgur5, line 411: Welp, now the `git commit` passes with an `--allow-empty`, but the `git push` line is failing. From [failing CI](https://github.com/argoproj/argo-workflows/actions/runs/8562226554/job/23465109452?pr=12891) in #12891:

```
Run git config --global user.name 'github-actions[bot]'
  git config --global user.name 'github-actions[bot]'
  git config --global user.email 'github-actions[bot]@users.noreply.github.com'
  git add ui/yarn.lock
  git commit -s --allow-empty -m 'chore: deduplicate yarn.lock'
  git push
  shell: /usr/bin/bash -e {0}
  env:
    NODE_OPTIONS: --max-old-space-size=4096
[detached HEAD 4ab78f5] chore: deduplicate yarn.lock
fatal: You are not currently on a branch.
To push the history leading to the current (detached HEAD)
state now, use

    git push origin HEAD:<name-of-remote-branch>

Error: Process completed with exit code 1[2](https://github.com/argoproj/argo-workflows/actions/runs/8561962725/job/23464600770?pr=12880#step:9:2)8.
```

agilgur5: We should be able to use [`$GITHUB_HEAD_REF`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) as the branch name (per the docs, it is set for PRs, and dependabot is always a PR), so something like:

```suggestion
          git push origin HEAD:$GITHUB_HEAD_REF
```

but honestly this needs some more testing given all the errors so far. I would suggest trying this on a PR on a fork and skipping the dependabot `actor` check for testing purposes

agilgur5: I'm testing this in https://github.com/agilgur5/argo-workflows/pull/3. I had to add `permissions.contents: write` as well (which isn't great, but necessary, with no real workarounds for it).

It's still [failing](https://github.com/agilgur5/argo-workflows/actions/runs/8563028220/job/23467388104?pr=3) though, because it thinks it has something to `fetch` from the remote branch, even though it doesn't. So it needs some more work. 

Tbh, this is getting complicated enough that it perhaps should be split out into its own independent action

agilgur5: > It's still [failing](https://github.com/agilgur5/argo-workflows/actions/runs/8563028220/job/23467388104?pr=3) though, because it thinks it has something to `fetch` from the remote branch, even though it doesn't.

I fixed this by adding `fetch-depth` to `actions/checkout`. It passes now.... but the PR won't merge because the new commit doesn't run any checks, since GH won't run actions on changes from other actions (to avoid an infinite loop), per [previous comment](https://github.com/argoproj/argo-workflows/pull/12234#discussion_r1399762107).

So this whole approach isn't workable as checks will never run/pass on the final commit, meaning it will never be mergeable due to branch protections.

.github/workflows/ci-build.yaml
agilgur5: > @Yaminyam if you'd like to work on a more optimal fix that checks if there are any `yarn.lock` before running any of this code, that'd be great!

So I added this logic in https://github.com/agilgur5/argo-workflows/pull/3, so this has been completed.
Also due to the [below comment](https://github.com/argoproj/argo-workflows/pull/12234#discussion_r1552636495), this entire approach is not workable, so I'm going to be reverting this shortly.

.github/workflows/ci-build.yaml
agilgur5: Reverted in https://github.com/argoproj/argo-workflows/pull/12892

Issue #12161, opened 2023-11-07
UI: 3.5 missing name + name prefix filters for archived workflows


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

v3.4.11 had 'name' and 'name prefix' filters in the 'archived workflows' ui, this is extremely useful to filter on workflows starting with a name or an exact name. particularly if its a sensor generated workflow or you just want to jump to a known workflow name without needing to construct a url in the browser address bar (not user friendly)

v3.5.1 has no such option in the UI!

### Version

3.5.1

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

maxisam:
We also need a filter for archive items. We didn't need it because It was a separated view.

agilgur5:
### doable now

This is now doable with the SQLite DB from #13021 / #12736. It previously was not possible for non-archived Workflows as k8s/etcd has no such native filter per https://github.com/argoproj/argo-workflows/issues/12025#issuecomment-1914065370

### implementation details; naming

#### archived vs. non-archived filter

@Adrien-D "State" is a bit ambiguous naming; that part we did actually discuss this in https://github.com/argoproj/argo-workflows/issues/12025#issuecomment-1865458377 et al, where I suggested "Workflow Storage" and "In-cluster" vs. "Archived" as the options. "Active" is a bit ambiguous as well, as was the original terminology of "Live" workflows.
To be fair, you can have an Archive DB in your cluster too, so "In-cluster" is not necessarily fully precise either, but "etcd" is an implementation detail of k8s and so could be both confusing and imprecise on some distros as well. "k8s resources" sounds a bit confusing user-side as well 😕 

Ideally also the checkboxes shouldn't display if you don't have the Workflow Archive enabled.

[...]

PR #12237, opened 2023-11-21
feat: Workflows list page performance improvements

### LINKED ISSUE #12025: 3.5 ListWorkflows causes server to hang when there are lots of archived workflows
### Pre-requisites

We had >200,000 rows in the workflow archive table, and when trying to view the new combined workflow/archived workflow list page in the UI, the server times out

scanning the code, it looks like the `LoadWorkflows` code loads _all_ rows from the archive table, combines them with the k8s results and then applies sorting and limiting.

as a workaround, we've reduced the archive ttl from 14 days to 1 day, and the endpoint now responds before timing out, but is still pretty slow.

### Version

v3.5.0

*--- edits below by agilgur5 to add updates since this is a (very) popular issue ---*

### Updates

- Most of the performance regression part of this issue should have been solved by https://github.com/argoproj/argo-workflows/pull/12068 (which did re-instate a different bug: https://github.com/argoproj/argo-workflows/issues/11715), which was released in [v3.5.1](https://github.com/argoproj/argo-workflows/blob/01936bb54d8c140a28d32bcf3cbe7572fd4d0f01/CHANGELOG.md#v351-2023-11-03)
- Another performance regression was fixed in #12912, which was released in [v3.5.6](https://github.com/argoproj/argo-workflows/blob/49e21acd48b850088a5111b67f372e7d180b15de/CHANGELOG.md#v356-2024-04-19)
- Discussion continues below on other regressions and thoughts on the general merge of the Archived + Live UI in 3.5
  - Please [help test](https://github.com/argoproj/argo-workflows/issues/12025#issuecomment-2035908312) the new in-memory SQLite DB from #12736 and report your results/feedback here!

sunyeongchoi:
Fixes #12025

### Motivation

A performance issue occurred when merging the workflows and archived workflows pages into one place, and the logic was changed to resolve this.

### Modifications

* When archived, an archived label is attached to the k8s resource, so the performance problem caused by the entire search was solved by changing the logic using the labelSelector of k8s.
* To manage the pagination values ​​of workflows and archived workflows separately, a value called paginationOptions was added to both the front and back.

### Verification

1. created a workflows resource with archived enabled.
2. checked that the data was not displayed in duplicate.
3. checked if it operates normally when changing the number of pages and clicking the `next page` and `first page` buttons.
4. checked whether all test codes pass without any problems


tooptoop4:
@sunyeongchoi are u able to include fix for https://github.com/argoproj/argo-workflows/issues/12161 too?

terrytangyuan (commented):
Thank you! Let's wait for others to review as well this time. It's Thanksgiving week and so please be patient while waiting for feedback.

sunyeongchoi:
> @sunyeongchoi are u able to include fix for #12161 too?

Yes. But I will take a look after this PR is completely finished.

sunyeongchoi:
I found a bug in the code I implemented. (Cronjob, event, and cli commands were also using a workflow list, but I realized that this had not been considered in this PR.)

As a result, the current PR is not complete and plan to add additional commit.

However, I would like to receive a review to see if the PR I implemented is in line with the direction of resolving the issue.

for example, when combining k8s workflows and archived workflows to show them on one page, the `PaginationOptions` parameter was added to the parameters, and the names were set to `WfContinue` for k8s workflows continue and `ArchivedContinue` for archived workflows continue.

I would like to receive reviews on whether you think this name is right or if there are other good names.

terrytangyuan:
> When archived, an archived label is attached to the k8s resource, so the performance problem caused by the entire search was solved by changing the logic using the labelSelector of k8s.

Could you elaborate why this would solve the performance problem? If the number of archived workflows is large, it seems like we would still encounter the performance issue.

sunyeongchoi:
> Could you elaborate why this would solve the performance problem? If the number of archived workflows is large, it seems like we would still encounter the performance issue.

Okay. A performance problem occurred because the total number of `archived workflows` was searched.

So the logic I implemented is as follows:

1. If the label "Archived" is attached to `k8s workflows`, it means that it is stored in `archived workflows`. Therefore, by adding the labelSelector syntax to `k8s workflows` `wfLabelSelector += fmt.Sprintf("%s!=%s", common.LabelKeyWorkflowArchivingStatus, "Archived")`, workflows stored in `archived workflows` are not searched in `k8s workflows`. (In other words, the possibility that `k8s workflows` and `archived workflows` may overlap was resolved through kubernetes labelSelector.)

2. If 10 pieces of data are needed on one page, if 6 `k8s workflows` were searched in step 1, 10-6 = 4, that is, there is a possibility of searching 4 more data, so at this time, only search up to 4 `archived workflows`.
```go
if int64(len(mergedWfs)) < limit {
  listOption.Continue = archivedContinue
  listOption.Limit = limit - int64(len(mergedWfs))
  listOption.LabelSelector = labelSelector
  archivedWfList, err = s.wfArchiveServer.ListArchivedWorkflows(ctx, &workflowarchivepkg.ListArchivedWorkflowsRequest{
    ListOptions: listOption,
    NamePrefix:  "",
    Namespace:   req.Namespace,
  })
}
```

3. As described in number 1, deduplication was solved through Kubernetes' labelSelector, and the page limit was solved as described in number 2, so if `k8s workflows` and `archived workflows` manage only the continue value for each pagination separately, the performance problem will be solved because `k8s workflows` and `archived workflows` use their own pagination logic.
```go
res := &wfv1.WorkflowList{ListMeta: metav1.ListMeta{ResourceVersion: wfList.ResourceVersion}, Items: mergedWfs, PaginationOptions: wfv1.WorkflowPaginationOptions{
  WfContinue:       common.CheckNilString(wfList.ListMeta.Continue),
  ArchivedContinue: common.CheckNilString(archivedWfList.ListMeta.Continue),
}}
```

ui/src/models/workflows.ts
```diff
@@ -776,6 +776,10 @@ export interface Condition {
 export type ConditionType = 'Completed' | 'SpecWarning' | 'MetricsError' | 'SubmissionError' | 'SpecError' | 'ArtifactGCError';
 export type ConditionStatus = 'True' | 'False' | 'Unknown';
 
+export interface WorkflowsPaginationOptions {
+    wfContinue: string;
+    archivedContinue: string;
+}
```
agilgur5, line 782:
```suggestion
}

```

missing a new line as is convention in this file between all the types

ui/src/app/workflows/pagination.ts
```diff
@@ -0,0 +1,7 @@
+export interface WorkflowsPagination {
```
agilgur5, line 1: I imagine this could just live next `WorkflowsPaginationOptions` in `models/workflows.ts`, no?

ui/src/app/shared/services/workflows-service.ts
```diff
@@ -36,6 +38,35 @@ export const WorkflowsService = {
...
+            'items.metadata.annotations',
+            'items.status.phase',
+            'items.status.message',
+            'items.status.finishedAt',
+            'items.status.startedAt',
+            'items.status.estimatedDuration',
+            'items.status.progress',
+            'items.spec.suspend',
+            'paginationOptions.wfContinue',
+            'paginationOptions.archivedContinue'
```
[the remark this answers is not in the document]
agilgur5: Oh it looks like you added `paginationOptions` to the `WorkflowsList` struct in `workflow_types.go` [above](https://github.com/argoproj/argo-workflows/pull/12237/files#diff-94d0f1284fafd0a1cb05dc2309b4be043b6fb01d66626ab1867d40c84b573ec2R251), nevermind then

pkg/apiclient/workflow/workflow.proto
```diff
@@ -147,6 +148,10 @@ service WorkflowService {
     option (google.api.http).get = "/api/v1/workflows/{namespace}/{name}";
   }
 
+  rpc ListK8SAndArchivedWorkflows(WorkflowListRequest) returns (github.com.argoproj.argo_workflows.v3.pkg.apis.workflow.v1alpha1.WorkflowList) {
+    option (google.api.http).get = "/api/v1/k8s-archived-workflows/{namespace}";
+  }
+
```
agilgur5, line 154: So I was thinking that this would work more roughly within the constraints of the v3.5 API, as in, implementing it as part of the `workflows` API method just as optional parameters. Would that not work?

agilgur5 (commented):
I only looked through this pretty quickly, so I still need to spend some more time reviewing this, but left a few comments.

Generically speaking, having two cursors (one for archived and one for in-cluster) was my idea to fix this as well, so I think I'm roughly good with this approach, but I need to look through some of the specifics a bit more. The API method, the deduplication approach, and how the cursors are stored in particular stood out.

I think this is roughly in the right direction, thanks for all the work and testing to get this functional!

ui/src/app/workflows/pagination.ts
sunyeongchoi: Oh. that is more natural! I'll reflect this opinion. Thank you :)

pkg/apiclient/workflow/workflow.proto
sunyeongchoi: I also kept thinking about that part.
If so, I will reflect this by branching to parameters within `ListWorkflows` and separating them.

Thank you for your feedback!

tooptoop4 (approved):

tczhao:
hi @sunyeongchoi , could you fix the DCO signature

agilgur5:
> hi @sunyeongchoi , could you fix the DCO signature

This still needs more work in any case, per above review.

Also per my [last comment](https://github.com/argoproj/argo-workflows/issues/12025#issuecomment-1881912722) on the issue, Sun Yeong will be busy a few more days at least

sunyeongchoi:
https://github.com/argoproj/argo-workflows/issues/12025#issuecomment-1949330952

I believe the concept of the issue has completely changed, so if everyone agrees, I will close this PR.

terrytangyuan:
Thank you so much for your efforts so far!

PR #12246, opened 2023-11-22
chore: Update CODEOWNERS file

terrytangyuan:
This has not been useful and redundant to approvers.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*.proto    @terrytangyuan @sarabala1979
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-*.proto    @jessesuen @alexec @terrytangyuan @sarabala1979


sarabala1979:
@terrytangyuan Can you help me to understand for this deletion?

sarabala1979:
I remember Alex and Jessie added this for Proto change needs to be reviewed by leads. Is the configure broken?

agilgur5:
> I remember Alex and Jessie added this for Proto change needs to be reviewed by leads. Is the configure broken?

No it works, but as Terry is currently the only one actively approving, it is not particularly useful right now. It can also give a false impression that there are more active approvers than there actually are.
Also, others are getting notifications despite not responding to them -- that may dilute notifications where we may actually want their input more in a direct mention, for instance.

agilgur5 (approved):

terrytangyuan:
I was thinking we don't need this since only approvers have permission to merge PRs and all of the ones listed in CODEOWNERS are already approvers. Also neither Alex nor Jesse is lead now.

sarabala1979:
> I am not sure if it's a good idea to distinguish proto changes from other changes. There are other changes that are important and might be breaking.
All CRD changes will be changed in the proto. The main idea is to get two approvers(approver and  Leads) for any CRD and API changes. I think it is needed for core files.

terrytangyuan:
@sarabala1979 I changed the PR to update the file instead of deleting it. I removed non-leads from the file.

sarabala1979 (approved):

PR #12258, opened 2023-11-25
feat: store podname in nodestatus

isubasinghe:
Fixes #TODO


agilgur5:
Looks like this was replaced by #12503

PR #12271, opened 2023-11-27
fix: remove deprecated function rand.Seed

szpnygo:
### Motivation

```go


// Seed uses the provided seed value to initialize the default Source to a
// deterministic state. Seed values that have the same remainder when
// divided by 2³¹-1 generate the same pseudo-random sequence.
// Seed, unlike the Rand.Seed method, is safe for concurrent use.
//
// If Seed is not called, the generator is seeded randomly at program startup.
//
// Prior to Go 1.20, the generator was seeded like Seed(1) at program startup.
// To force the old behavior, call Seed(1) at program startup.
// Alternately, set GODEBUG=randautoseed=0 in the environment
// before making any calls to functions in this package.
//
// Deprecated: As of Go 1.20 there is no reason to call Seed with
// a random value. Programs that call Seed with a known value to get
// a specific sequence of results should use New(NewSource(seed)) to
// obtain a local random generator.
func Seed(seed int64) {

```

### Modifications

remove deprecated function rand.Seed

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -63,8 +63,6 @@ linters-settings:
       # G601: Implicit memory aliasing in for loop.
       - G601
 issues:
-  exclude:
-    - "SA1019: rand.Seed has been deprecated since Go 1.20"
   exclude-rules:
     - path: server/artifacts/artifact_server_test.go
       text: "response body must be closed"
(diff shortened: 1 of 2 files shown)


sarabala1979 (approved):
LGTM
Rand is initialized by default
```
var rng = struct {
	sync.Mutex
	rand *rand.Rand
}{
	rand: rand.New(rand.NewSource(time.Now().UnixNano())),
}```

Issue #7291, opened 2021-11-27
Circuit Breakers for CronWorkflows


simster7:
# Summary

What change needs making?

Implement a circuit breaker for cron workflows. For example, if a workflow has had multiple failures in a row, pause the workflow automatically. Or pause the workflow after a certain number of executions (https://github.com/argoproj/argo-workflows/issues/7201).

Either an expression:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  name: hello-world
spec:
  schedule: "* * * * *"
  pauseExpression: "{{cronworkflow.failures}} > 3"
```

Or a field:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  name: hello-world
spec:
  schedule: "* * * * *"
  pauseWhen:
    failures: 2
```

# Use Cases

When would you use this?

Cron workflows are the only CRD that act on their own (i.e. no explicit action is needed to spawn workflows from cron workflows), so a circuit breaker could be useful.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

terrytangyuan:
Duplicate of https://github.com/argoproj/argo-workflows/issues/5659?

simster7:
More like an extension

simster7:
@alexec Any thoughts on the spec proposal or this feature overall? Want to make sure we are all aligned before implementation begins.

alexec:
I think the concept is fine. I'm not sure many people are asking for it, so I don't see the core team working on it soon.

tooptoop4:
@eduardodbr since https://github.com/argoproj/argo-workflows/pull/12305 was merged do u think this issue can be closed?

eduardodbr:
I believe it can be closed. @agilgur5 do you agree?

agilgur5:
Per https://github.com/argoproj/argo-workflows/issues/5659#issuecomment-2381628641, I'm not sure either of these two were resolved, since they both specifically say "pause" and "suspend", which is intentionally different from `stopStrategy` per https://github.com/argoproj/argo-workflows/pull/12696#discussion_r1516670248

Issue #7201, opened 2021-11-10
Limit Cronworkflow Runtime count


sundbing:
# Summary

Add a Max counter to the cronworkflowtemplate that is limiting the amount of reschedules and allows to run a workflow for the given number of Times on a cron schedule. 

# Use Cases

Current cronworkflows are running infint till it is removed. That way running a scheduled workflow one time or a fixed amount of times isn't easy to implement. 
E.g. Run a workflow every hour but only 10 times. 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

NikeNano:
I would be happy to help out with this @simster7 if you haven't started the work?

simster7:
Actually, I'll take this one. I have some ideas with how to fit this request into a bigger feature I'm thinking about

eduardodbr:
closed by #12305

PR #12305, opened 2023-11-30
feat: add stopStrategy to cron workflows

eduardodbr:
This feature allows the configuration of a stopping condition to cron workflows, allowing them to trigger a workflow only a specific amount of times or stop scheduling it after a number of failed attempts. 
Discussed here: #11559.
Could fix: #10620, #5659, #7291, #7201

### Features: 

 - New Spec field `stopStrategy` with a `condition` that is used to stop scheduling workflows. The controller stops scheduling workflows if the stopping condition is achieved:
 ```
 stopStrategy:
    condition: "succeeded >= 1"
 ```
- New status fields:
     - Succeeded - counts the number of workflows that completed in phase `Succeeded`
     - Failed -  counts the number of workflows that failed to submit or completed in phase  `Error` or `Failed` . 
     - Completed - whether the stopping condition was achieved or not (is it worth it to have it as an annotation?) 
- The `completed` status field is used instead of changing the cronworkflow with “suspend=true” to avoid changing the cronworkflow spec
- Keywords allowed for stopping conditions are `failed` and `succeeded`:
      - Failed - uses the status field `failed` 
      - Succeeded - uses the status field `succeeded`
 ```
stopStrategy:
  condition: "failed >= 3"
```
 ```
stopStrategy:
  condition: "succeeded >= 1"
```
-  The stopping condition is checked on every child workflow attempt and after every finished workflow.
-  Since the stopping condition is verified on every schedule, updating the stopping condition after the workflow is completed will continue the normal cron workflow execution.

### Possible Issues:

- Depending on the time it takes to schedule and run a workflow, it is possible that the number of executions is higher than the configured max executions. For example, if you schedule it to run every minute but the WF executes in 90 seconds, it will have 2 completions because when the stopping condition is achieved, there is another workflow running. For that reason, it is advisable to have conditions like `succeeded  >= 1` instead of `succeeded == 1`
- Errors submitting workflows are being counted towards the `failed` status field. We could separate them into:
    - `failed` for counting workflows that terminated in failed/errored state
    - `failedSubmissions` for counting `ConditionTypeSubmissionError` 

### Possible Next steps:
- Allow to configure if the `succeeded` and `failed` counters are consecutive or total. This PR only implements the total approach. We can add this feature to this PR as well if you think it's essential.
- Update UI to handle completed workflows: show and filter by completed cronworkflows, remove the “Next Run” countdown 

### Verification

This feature can be tested with the following CronWorkflow:

```
kind: CronWorkflow
metadata:
  name: test-cron-wf-stop-condition
spec:
  schedule: "* * * * *"
  concurrencyPolicy: "Allow"
  stopStrategy:
    condition: "succeeded >= 1"
  workflowSpec:
    metadata:
      labels:
    podGC:
      strategy: OnPodCompletion
    entrypoint: whalesay
    templates:
      - name: whalesay
        container:
          image: argoproj/argosay:v2
          command: [/argosay]
```

The workflow should be triggered only once. 

## TODO

update the documentation if this feature is accepted.


workflow/cron/operator.go
```diff
@@ -145,7 +153,7 @@ func (woc *cronWfOperationCtx) persistUpdate(ctx context.Context) {
 }
 
 func (woc *cronWfOperationCtx) persistUpdateActiveWorkflows(ctx context.Context) {
```
Joibel, line 155: This function should be renamed now, it's not just *Active* workflows

test/e2e/cron_test.go
```diff
@@ -354,6 +354,75 @@ spec:
...
+      - name: whalesay
+        container:
+            image: argoproj/argosay:v2
+            command: [/argosay]`).
+			When().
+			CreateCronWorkflow().
+			Wait(2 * time.Minute). // wait to be scheduled 2 times, but only runs once
+			Then().
+			ExpectCron(func(t *testing.T, cronWf *wfv1.CronWorkflow) {
+				assert.Equal(t, int64(1), cronWf.Status.Succeeded)
```
Joibel, line 389: Also check `Status.Failed` is as expected.

test/e2e/cron_test.go
```diff
@@ -354,6 +354,75 @@ spec:
...
+      - name: whalesay
+        container:
+          image: argoproj/argosay:v2
+          args: ["exit", "1"]`).
+			When().
+			CreateCronWorkflow().
+			Wait(2 * time.Minute). // wait to be scheduled 2 times, but only runs once
+			Then().
+			ExpectCron(func(t *testing.T, cronWf *wfv1.CronWorkflow) {
+				assert.Equal(t, int64(1), cronWf.Status.Failed)
```
Joibel, line 422: Also check `Status.Succeeded` is as expected.

Joibel (changes requested):
`workflow/cron/controller.go:addCronWorkflowInformerHandler` should also become a `FilteringResourceEventHandler` like in `workflow/controller/controller.go:addWorkflowInformerHandlers` so that it filters out completed workflows.

eduardodbr:
> `workflow/cron/controller.go:addCronWorkflowInformerHandler` should also become a `FilteringResourceEventHandler` like in `workflow/controller/controller.go:addWorkflowInformerHandlers` so that it filters out completed CronWorkflows.

@Joibel 

I maintained the same behavior as the `Suspend` spec field. However, I agree that we can improve performance if we just stop scheduling jobs from a cronworkflow that is completed. To do that we need to remove the job from the cronjob queue.

As such, I think that instead of doing what you proposed, we could just check the stopping condition when the controller is processing a new cron item ( `processNextCronItem` func). If it evaluates to true, the cronjob is removed from the queue and never scheduled again. If a change is made to the cronworkflow spec, the stopping condition is evaluated again, adding the job to the cronjob queue if it evaluates to false.

From my understanding, there are 3 places where we can evaluate the stopping condition:

- The EventHandler as you proposed:  avoids adding the job to the cronjob queue when the condition is already true. However, if the job is already in the cronjob queue (i.e. an update to the spec), new jobs will keep being scheduled. I concluded this from a quick test, correct me if I'm wrong.
- The `processNextCronItem` func:  we can remove the job from the cronjob queue if the condition evaluates to true. This is where we can improve performance by not scheduling and running unnecessary cron jobs.
- when the job that triggers the workflow runs - I believe this one we must have it

I guess we can evaluate the stopping condition on all three phases but I don't know if we gain a lot by doing it at the EventHandler. If the EventHandler has any other responsibility that I'm missing here, please correct me.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -69,6 +78,12 @@ type CronWorkflowStatus struct {
 	LastScheduledTime *metav1.Time `json:"lastScheduledTime" protobuf:"bytes,2,opt,name=lastScheduledTime"`
 	// Conditions is a list of conditions the CronWorkflow may have
 	Conditions Conditions `json:"conditions" protobuf:"bytes,3,rep,name=conditions"`
+	// Succeeded is a counter of how many times the child workflows had success
+	Succeeded int64 `json:"succeeded" protobuf:"varint,4,rep,name=succeeded"`
+	// Failed is a counter of how many times a child workflow terminated in failed or errored state
+	Failed int64 `json:"failed" protobuf:"varint,5,rep,name=failed"`
+	// Completed is a flag that is set to true when the stopping condition is achieved which stops new CronWorkflows from running
+	Completed bool `json:"completed" protobuf:"varint,6,rep,name=completed"`
```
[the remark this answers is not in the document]
eduardodbr: That makes sense. Maybe `active` instead of `running`?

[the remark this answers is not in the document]
Joibel: When would the phase be failed? If `checkStoppingCondition` fails to evaluate? Would the cron continue to fire if that happens?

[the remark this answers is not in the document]
eduardodbr: IMO the cron shouldn't fire on an invalid `condition`

eduardodbr:
> 
> If the EventHandler was a Filtering event handler then if the `FilterFunc` returns false then it should initiate deletion from the informer. I think everywhere else it becomes unnecessary to delete it from the informer, and it's where I'd expect it to happen.
> 
> For the queue, your analysis is correct, I'm talking about the informer instead.
> 
> Yes, it's confusing, we have two things that cache/queue. Once it goes from the informer then it'll disappear from the queue, and you need the guards you have for that. Once it's filtered from the informer it's as though it doesn't exist to the rest of the code.
> 
> I hope I've got this right

I've added the `FilterFunc` to the cron workflow informer event handler. However, I believe it is not working as you would expect: `Once it goes from the informer then it'll disappear from the queue`
When the stop condition is achieved an event is generated that removes it from the informer but It seems that it is still on the queue because the controller is still trying to schedule new jobs so it keeps logging:

`level=info msg="CronWorkflow test-cron-wf-stop-condition is marked as completed since it achieved the stopping condition" namespace=argo workflow=test-cron-wf-stop-condition`

I've pushed my changes so that you can take a look.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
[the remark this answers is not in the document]
Joibel: I'm still not sure when we would enter `phase: failed` then - I'm just trying to determine what these three phases would mean.

Joibel (changes requested):

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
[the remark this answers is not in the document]
eduardodbr: Let's wait for @sarabala1979  input  but I understood it as:
`failure` - if a condition using the `failed` keyword is achieved
`complete` - if any other type of condition is achieved 

However, this increases complexity because we need to check the stopping condition and which type of condition (success/fail) was achieved. 

If you think that it is better to have a phase, for now, I would prefer to have it with `active` or `terminated`.
It could also have a `error` phase to handle the case where `checkStoppingCondition` fails to evaluate, however, IMO it makes more sense to just throw a spec error.

workflow/cron/controller.go
```diff
@@ -204,26 +204,49 @@ func (cc *Controller) processNextCronItem(ctx context.Context) bool {
...
+					log.Warnf("Cron Workflow FilterFunc: '%v' is not an unstructured", obj)
+					return false
+				}
+				cronWf := &v1alpha1.CronWorkflow{}
+				err := util.FromUnstructuredObj(un, cronWf)
+				if err != nil {
+					log.Warnf("Cron Workflow FilterFunc: failed to convert unstructured '%v' to a CronWorkflow: %v", un, err)
+					return false
+				}
+				cronWorkflowOperationCtx := newCronWfOperationCtx(cronWf, cc.wfClientset, cc.metrics, cc.wftmplInformer, cc.cwftmplInformer)
```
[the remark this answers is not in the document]
eduardodbr: I purposely implemented it as a feature. In the same way you can edit a schedule expression or suspend/unsuspend a cronjob, you could also update the stopping condition without having to delete and recreate the cron workflow. 

@sarabala1979  @terrytangyuan  @caelan-io  let us know what do you think as well.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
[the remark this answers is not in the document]
Joibel: I agree with all you've written here. I had assumed we'd want to support `succeeded+failed=5` as a condition, which isn't either failure type, hence the confusion.

I also agree with you on how to handle `checkStoppingCondition` failure to evaluate.

workflow/common/common.go
```diff
@@ -101,6 +101,9 @@ const (
 	// LabelKeyReportOutputsCompleted is a label applied to WorkflowTaskResults indicating whether all the outputs have been reported.
 	LabelKeyReportOutputsCompleted = workflow.WorkflowFullName + "/report-outputs-completed"
 
+	// LabelKeyCronWorkflowCompleted is a label applied to the cron workflow when the configured stopping condition is achieved
+	LabelKeyCronWorkflowCompleted = workflow.CronWorkflowFullName + "/completed"
```
eduardodbr, line 105: let me know if there is a better place to create this label

Joibel:
I'm happy with this. @sarabala1979 could you take another look now please.

terrytangyuan (approved):
LGTM. @sarabala1979 and others: would you like to take another look before merging?

terrytangyuan:
Please also resolve conflicts

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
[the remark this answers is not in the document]
sarabala1979: @eduardodbr Sorry for the long delay for reply. my comment was for only cron workflow `phase`, as per PR cronworkflow is running or not is based on `completed`
instead of `boolean` can we have `phase` with `active` and `stopped` ?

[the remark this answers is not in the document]
sarabala1979:
``` yaml
   phase: active/stopped
```

workflow/cron/controller.go
```diff
@@ -204,26 +204,45 @@ func (cc *Controller) processNextCronItem(ctx context.Context) bool {
...
+					if err == nil {
+						cc.cronWfQueue.Add(key)
+					}
+				},
+			},
+		})
+}
+
+func isCompleted(wf v1.Object) bool {
+	completed, ok := wf.GetLabels()[common.LabelKeyCronWorkflowCompleted]
```
sarabala1979, line 241: is there any reason to check the label instead of phase?

sarabala1979 (commented):
@eduardodbr PR looks good. 
few comments:
1. Cronworkflow didn't set `active` phase when it is starting the first process
2. can we use  the phase as status check for active/completed?

workflow/cron/controller.go
eduardodbr: To use the Phase we would need to change the InformerHandler FilterFunc to convert the *unstructured.Unstructured object to a v1alpha1.CronWorkflow{} to be able to access the `Status.Phase` field. By doing this, malformed workflows were being filtered out by the handler, and no events were being generated to warn the user, thus the test `TestMalformedCronWorkflow` would failed.

eduardodbr: using a label allows to filter based on an unstructured.Unstructured object

sarabala1979: got it. thanks

sarabala1979 (approved):
LGTM

sarabala1979:
@eduardodbr Thanks for your contribution.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -59,6 +59,15 @@ type CronWorkflowSpec struct {
 	Timezone string `json:"timezone,omitempty" protobuf:"bytes,8,opt,name=timezone"`
 	// WorkflowMetadata contains some metadata of the workflow to be run
 	WorkflowMetadata *metav1.ObjectMeta `json:"workflowMetadata,omitempty" protobuf:"bytes,9,opt,name=workflowMeta"`
+	// StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
```
agilgur5, line 62: Similar to https://github.com/argoproj/argo-workflows/pull/12616#discussion_r1495330587, we'd want to put a version specifier here since this pops up in the Field Reference.

```suggestion
	// v3.6 and after: StopStrategy defines if the CronWorkflow should stop scheduling based on a condition
```

  - "involving a number of runs of the workflow" is described in the `condition` definition, so we can leave that out here (it is also a bit verbose)
- consistently use "CronWorkflow"
- I think "scheduling" as you used below in the `Condition` description is clearer wording than "triggering". "Triggering" is a bit ambiguous and not used in the docs, whereas "Schedule" is an existing term

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -59,6 +59,15 @@ type CronWorkflowSpec struct {
...
+	StopStrategy *StopStrategy `json:"stopStrategy,omitempty" protobuf:"bytes,10,opt,name=stopStrategy"`
+}
+
+// StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
```
agilgur5, line 66:
```suggestion
// StopStrategy defines if the CronWorkflow should stop scheduling based on a condition
```
same as above, although I think the version specifier isn't necessary here since the Fields Reference doesn't pull this one

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -69,8 +78,21 @@ type CronWorkflowStatus struct {
 	LastScheduledTime *metav1.Time `json:"lastScheduledTime" protobuf:"bytes,2,opt,name=lastScheduledTime"`
 	// Conditions is a list of conditions the CronWorkflow may have
 	Conditions Conditions `json:"conditions" protobuf:"bytes,3,rep,name=conditions"`
+	// Succeeded is a counter of how many times the child workflows had success
```
agilgur5, line 81:
```suggestion
	// v3.6 and after: Succeeded counts how many times the child workflows succeeded
```
a bit simpler plus a version specifier

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -69,8 +78,21 @@ type CronWorkflowStatus struct {
...
+	Succeeded int64 `json:"succeeded" protobuf:"varint,4,rep,name=succeeded"`
+	// Failed is a counter of how many times a child workflow terminated in failed or errored state
```
agilgur5, line 83:
```suggestion
	// v3.6 and after: Failed counts how many times a child workflow failed
```
- version specifier
- a bit simpler
- ~"terminated" to "completed" -- "Terminate" is an action on Workflows that has a specific meaning. "Completed" has the correct meaning here~.
  - we can just say "failed", which describes this and is similarly an existing term

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -59,6 +59,15 @@ type CronWorkflowSpec struct {
...
 	// WorkflowMetadata contains some metadata of the workflow to be run
 	WorkflowMetadata *metav1.ObjectMeta `json:"workflowMetadata,omitempty" protobuf:"bytes,9,opt,name=workflowMeta"`
+	// StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
+	StopStrategy *StopStrategy `json:"stopStrategy,omitempty" protobuf:"bytes,10,opt,name=stopStrategy"`
+}
+
+// StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
+type StopStrategy struct {
+	// Condition defines a condition that stops scheduling workflows when evaluates to true. Use the
+	// keywords `failed` or `succeeded` to access the number of failed or successful child workflows.
```
agilgur5, line 69:
```suggestion
	// v3.6 and after: Condition is an expression that stops scheduling workflows when true. Use the
	// keywords `failed` or `succeeded` to access the number of failed or successful child workflows.
```
- add version specifier
- specify that this is [an expression](https://argo-workflows.readthedocs.io/en/latest/variables/#expression)

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -69,8 +78,21 @@ type CronWorkflowStatus struct {
 	LastScheduledTime *metav1.Time `json:"lastScheduledTime" protobuf:"bytes,2,opt,name=lastScheduledTime"`
 	// Conditions is a list of conditions the CronWorkflow may have
 	Conditions Conditions `json:"conditions" protobuf:"bytes,3,rep,name=conditions"`
+	// Succeeded is a counter of how many times the child workflows had success
+	Succeeded int64 `json:"succeeded" protobuf:"varint,4,rep,name=succeeded"`
+	// Failed is a counter of how many times a child workflow terminated in failed or errored state
+	Failed int64 `json:"failed" protobuf:"varint,5,rep,name=failed"`
+	// Phase defines the cron workflow phase. It is changed to Stopped when the stopping condition is achieved which stops new CronWorkflows from running
```
agilgur5, line 85: This one was a bit incorrect, as it stops new Workflows from being scheduled, not CronWorkflows

```suggestion
	// v3.6 and after: Phase is an enum of Active or Stopped. It changes to Stopped when stopStrategy.condition is true
```
- add version specifier
- simplify a bit, similar to `StopStrategy` above

agilgur5 (commented):
Thanks for working on the docs for this in #12696. I was going to suggest that similar to https://github.com/argoproj/argo-workflows/pull/12616#pullrequestreview-1889629343, but you beat me to the chase.

I left a few in-line comments here on the comments which populate the Field Reference docs page. Modifications to those probably make sense to add to #12696?

agilgur5:
Above changes have now been implemented in #12696

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
agilgur5: I was wrong, this does get pulled into the Fields Reference and so does need a version specifier. Followed up in https://github.com/argoproj/argo-workflows/pull/12696#discussion_r1508529694

jigar-7:
Is there any way I can do or condition like "succeeded >=1 or failed >=1". which means regardless of success or fail of child workflow it should trigger only once?

eduardodbr:
> Is there any way I can do or condition like "succeeded >=1 or failed >=1". which means regardless of success or fail of child workflow it should trigger only once?

Yes you can do that. However this will only be released on v3.6

agilgur5:
Per https://github.com/argoproj/argo-workflows/pull/13474#issuecomment-2347998579, I think there's some follow-up work now to unify the variables and naming prior to 3.6.0.

For instance, `cronworkflow.` as a prefix and using the field `expression` instead of `condition`, as all other expression based fields use that name (except for `when`, c.f. #7831)

agilgur5:
Alan and I discussed [the above](https://github.com/argoproj/argo-workflows/pull/12305#issuecomment-2381625468) in the [Oct 1st Contributor Meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.gfymrsa8jhi7) and agreed on this. See #13702 as a follow-up

PR #12318, opened 2023-12-05
fix: SSO with Jumpcloud "email_verified" field #12257

### LINKED ISSUE #12257: SSO with Jumpcloud "email_verified" causes failure to authenticate
### Pre-requisites

I have tried to setup Jumpcloud as the OIDC provider.
When one sets the email profile in Jumpcloud, the attributes email (dynamic) and email_verified (statically set to true) are added.
Attempting to login in this scenario causes the following error to occur upon callback:
`time="2023-11-24T18:22:44.257Z" level=error msg="failed to get claims from the id token" error="json: cannot unmarshal string into Go struct field claimAlias.email_verified of type bool"
`
Removing email_verified from the list of attributes solves the issue. 
i have tried the following values for email_verified: `1, true, TRUE, True` but they all fail equally. 

### Version

latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

sonbui00:
Fixes #12257

### Motivation
- Support Jumpcloud

### Modifications
- Support "email_verified" field as string "true" 

### Verification

- Unit test included

### DIFF
--- a/server/auth/types/claims.go
+++ b/server/auth/types/claims.go
@@ -12,7 +12,7 @@ type Claims struct {
 	jwt.Claims
 	Groups                  []string               `json:"groups,omitempty"`
 	Email                   string                 `json:"email,omitempty"`
-	EmailVerified           bool                   `json:"email_verified,omitempty"`
+	EmailVerified           bool                   `json:"-"`
 	Name                    string                 `json:"name,omitempty"`
 	ServiceAccountName      string                 `json:"service_account_name,omitempty"`
 	ServiceAccountNamespace string                 `json:"service_account_namespace,omitempty"`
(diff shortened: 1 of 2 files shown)


sarabala1979 (approved):

server/auth/types/claims.go
```diff
@@ -12,7 +12,7 @@ type Claims struct {
 	jwt.Claims
 	Groups                  []string               `json:"groups,omitempty"`
 	Email                   string                 `json:"email,omitempty"`
-	EmailVerified           bool                   `json:"email_verified,omitempty"`
+	EmailVerified           bool                   `json:"-"`
```
agilgur5, line 15: Hmm I'm thinking the more correct way of handling this would have been to use a custom type like in [this SO question's answers](https://stackoverflow.com/q/30856454/3431180)

Issue #9120, opened 2022-07-07
Unable to resolve output artifact of steps from expression when it refers to its own input artifact


zjgemi:
## Summary

I'm trying to assign an output artifact of steps using `fromExpression`. When the expression contains reference of an input artifact of the same steps, Argo fails to resolve it. Interestingly, using `from` instead of `fromExpression` works.

## Diagnostics

The smallest workflow that reproduces the bug

```yaml
kind: Workflow
metadata:
  generateName: from-expression-bug-
spec:
  entrypoint: main
  arguments:
    artifacts:
    - name: foo
      raw:
        data: |
          Hello
  templates:
    - name: main
      inputs:
        artifacts:
        - name: foo
      steps:
      - - name: hello
          inline:
            container:
              image: alpine:latest
      outputs:
        artifacts:
        - name: result
          # a trivial expression here only for reproducing the bug
          fromExpression: "1 == 1 ? inputs.artifacts.foo : inputs.artifacts.foo"
```

When the workflow is submitted, the steps becomes error with message:
```
unable to resolve outputs from scope: Variable {{&Artifact{Name:result,Path:,Mode:nil,From:,ArtifactLocation:ArtifactLocation{ArchiveLogs:nil,S3:nil,Git:nil,HTTP:nil,Artifactory:nil,HDFS:nil,Raw:nil,OSS:nil,GCS:nil,},GlobalName:,Archive:nil,Optional:false,SubPath:,RecurseMode:false,FromExpression:1 == 1 ? inputs.artifacts.foo : inputs.artifacts.foo,ArtifactGC:nil,Deleted:false,}}} is not an artifact
```
<img width="487" alt="截屏2022-07-07 下午2 30 06" src="https://user-images.githubusercontent.com/27884723/177706703-9b68043d-2144-4099-ae30-548b6af9280c.png">

While the workflow using `from` instead of `fromExpression`
```yaml
kind: Workflow
metadata:
  generateName: from-expression-bug-
spec:
  entrypoint: main
  arguments:
    artifacts:
    - name: foo
      raw:
        data: |
          Hello
  templates:
    - name: main
      inputs:
        artifacts:
        - name: foo
      steps:
      - - name: hello
          inline:
            container:
              image: alpine:latest
      outputs:
        artifacts:
        - name: result
          from: "inputs.artifacts.foo"
```
will not encounter the error.

---

**Message from the maintainers**:

[...]

PR #12320, opened 2023-12-05
fix: resolve output artifact of steps from expression when it refers …

shuangkun:
Solve the question mentioned:
https://github.com/argoproj/argo-workflows/issues/9120


workflow/controller/scope.go
```diff
@@ -103,7 +103,12 @@ func (s *wfScope) resolveArtifact(art *wfv1.Artifact) (*wfv1.Artifact, error) {
 	}
 	valArt, ok := val.(wfv1.Artifact)
 	if !ok {
-		return nil, errors.Errorf(errors.CodeBadRequest, "Variable {{%v}} is not an artifact", art)
+		ptArt, ok := val.(*wfv1.Artifact)
```
terrytangyuan, line 106: Could you explain why this is needed? Seems redundant to the above lines?

shuangkun: Like the workflow mentioned，when i submit the workflow which refers itself input artfacts in fromExpression,  it return a pointer (*wfv1.Artifact)， so i try to transform it to a point when it transform to wfv1.Artifact failed.

kind: Workflow
metadata:
  generateName: from-expression-bug-
spec:
  entrypoint: main
  arguments:
    artifacts:
    - name: foo
      raw:
        data: |
          Hello
  templates:
    - name: main
      inputs:
        artifacts:
        - name: foo
      steps:
      - - name: hello
          inline:
            container:
              image: alpine:latest
      outputs:
        artifacts:
        - name: result
          # a trivial expression here only for reproducing the bug
          fromExpression: "1 == 1 ? inputs.artifacts.foo : inputs.artifacts.foo"

terrytangyuan (commented):
Can you add a comment in the code as well as a test case?

shuangkun:
> Can you add a comment in the code as well as a test case?

I have add a unit test and comment.
Before I add the pr, there is a error:
image
After add the code, the ut pass:

terrytangyuan (approved):
Thank you!

PR #12323, opened 2023-12-06
fix: create dir when input path is not exist in oss

shuangkun:
when down a flie to path, and the path is not exist which mount by a oss volume. It made a mistake. So I want to create the dir path before functione "GetObjectToFile". Be consistent with the function "GetOssDirectory".
<!--

### DIFF
--- a/workflow/artifacts/oss/oss.go
+++ b/workflow/artifacts/oss/oss.go
@@ -127,6 +127,11 @@ func (ossDriver *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string)
 				return !isTransientOSSErr(err), err
 			}
 			objectName := inputArtifact.OSS.Key
+			dirPath := filepath.Dir(path)
+			err = os.MkdirAll(dirPath, 0o700)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


workflow/artifacts/oss/oss.go
```diff
@@ -127,6 +127,11 @@ func (ossDriver *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string)
 				return !isTransientOSSErr(err), err
 			}
 			objectName := inputArtifact.OSS.Key
+			dirPath := filepath.Dir(path)
+			err = os.MkdirAll(dirPath, 0o700)
+			if err != nil {
+				return false, fmt.Errorf("mkdir %s error: %w", dirPath, err)
+			}
 			origErr := bucket.GetObjectToFile(objectName, path)
```
terrytangyuan, line 135: Should this be put around line 152 code block?

shuangkun: It should before "GetObjectToFile", because destination dir is not exist.

terrytangyuan (approved):
LGTM

terrytangyuan:
Could you take a look at the failing test?

PR #12328, opened 2023-12-07
feat: speed up resolve reference

shuangkun:
I have a big and flat workflow( 2000+ ) pod, which have artifacts reference each step. I find it won’t continue running down when it run to around step 100～200. 

After investigation， I find the function "resolveReferences" is too slow. One Step Group node has 1000 step need 300s to resolveReferences and then update the Step Group succeed. After that, then the workflow can run down.  But the default time for woc.operator is 30s.  I have set MAX_OPERATION_TIME to 20 min to succeed the workflow. But I think there is a better solution.

So I want to  speed up resolve reference.

I print the step time when it change to newStepStr use "template.Replace". Maybe I will speed up the "template.Replace"  function future.
image
image

Fixes #TODO

### DIFF
--- a/workflow/controller/steps.go
+++ b/workflow/controller/steps.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Knetic/govaluate"
@@ -397,6 +398,15 @@ func shouldExecute(when string) (bool, error) {
 	return boolRes, nil
 }
 
+func errorFromChannel(errCh <-chan error) error {
+	select {
+	case err := <-errCh:
+		return err
+	default:
+	}
+	return nil
+}
+
 // resolveReferences replaces any references to outputs of previous steps, or artifacts in the inputs
 // NOTE: by now, input parameters should have been substituted throughout the template, so we only
 // are concerned with:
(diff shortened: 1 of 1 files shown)


workflow/controller/steps.go
```diff
@@ -462,13 +462,34 @@ func (woc *wfOperationCtx) resolveReferences(stepGroup []wfv1.WorkflowStep, scop
...
+	for i, step := range stepGroup {
+		parallelStepNum <- step.Name
+		wg.Add(1)
+		go func(i int, step wfv1.WorkflowStep) {
+			defer wg.Done()
+			if err := resolveStepReferences(i, step, newStepGroup); err != nil {
+				woc.log.WithFields(log.Fields{"stepName": step.Name}).WithError(err).Error(err.Error())
+				resolveReferenceError = err
+			}
+			<-parallelStepNum
```
juliev0, line 487: What's the purpose of `parallelStepNum` channel?

shuangkun: > What's the purpose of `parallelStepNum` channel?

Just control the number of concurrencies， like “parallelPodNum” in workflow/controller/operator.go

juliev0: oh, neat. Would you mind adding a comment that that's what you're doing?

shuangkun: Okay,I will add some comments.

juliev0: thank you!

workflow/controller/steps.go
```diff
@@ -461,13 +471,37 @@ func (woc *wfOperationCtx) resolveReferences(stepGroup []wfv1.WorkflowStep, scop
...
+	parallelStepNum := make(chan string, 500)
+	errCh := make(chan error, len(stepGroup)) // contains the error during resolveStepReferences
+	var wg sync.WaitGroup
+	for i, step := range stepGroup {
+		parallelStepNum <- step.Name
+		wg.Add(1)
+		go func(i int, step wfv1.WorkflowStep) {
+			defer wg.Done()
+			if err := resolveStepReferences(i, step, newStepGroup); err != nil {
+				woc.log.WithFields(log.Fields{"stepName": step.Name}).WithError(err).Error(err.Error())
```
juliev0, line 494: Just want to make sure - would we possibly print the error twice? I'm looking at `WithError(err).Error(err.Error())`

shuangkun: Thanks, I changed it!

juliev0 (approved):

PR #12334, opened 2023-12-08
docs: Add Distributed Machine Learning Patterns book

terrytangyuan:
My book [Distributed Machine Learning Patterns from Manning Publication](https://github.com/terrytangyuan/distributed-ml-patterns) uses Argo Workflows and would be a good resource for people who use it for ML use cases.


juliev0 (changes requested):
Should we change the heading "Community Blogs and Presentations" now?

terrytangyuan:
@juliev0 I don't want to change it just for this one addition since this seems to be a common heading across other Argo subprojects and they may be linked by other projects/resources already.

juliev0:
> @juliev0 I don't want to change it just for this one addition since this seems to be a common heading across other Argo subprojects and they may be linked by other projects/resources already.

Okay. If there is a link to it, I wouldn't want to mess that up. Well, let's give it a few days in case anyone else objects. After that I'll let it go. :) (the other option would be to have another heading for this)

juliev0 (approved):

PR #12360, opened 2023-12-14
chore: move documentation to readthedocs

### LINKED ISSUE #11390: Versioned documentation

jmeridth:
Draft until I get the gh-pages PR updated with meta tags

Fixes #11390

### Motivation

Moving to readthedocs so we can have versioned docs easily and standardize like argo-cd and other projects have already.

### Modifications

- [x] add .readthedocs.yml
- [x] add docs/requirements.txt and use it in Makefile
- [x] update urls in code
  - [x] sdk and other files updates were from `make pre-commit -B` locally
- [x] update .spelling with words to exlude during spellcheck
  - [x] alphabetize 
- [x] update gh-pages branch with meta tags to handle redirects #12362
- [x] update /hack script to use new mkdocs built-in validation (example result [here](https://github.com/argoproj/argo-workflows/actions/runs/7210935020/job/19645182282?pr=12360#step:6:148))
- [x] remove the [gh-pages publish action](https://github.com/argoproj/argo-workflows/blob/76f4d1f9ca53b8ac48965e12c15454e544936c49/.github/workflows/docs.yaml#L47) in this PR so that https://github.com/argoproj/argo-workflows/pull/12362 isn't overwritten later

### Verification

https://jm-argo-workflows.readthedocs.io/en/latest/

### Notes

For us to have other versions available we have to mark them as Active in the readthedocs UI

Screenshot 2023-12-14 at 8 02 34 AM

But those versions have to have all of the changes in this PR or they will fail

Screenshot 2023-12-14 at 8 08 21 AM

So this idea of versioned docs may only apply going forward.


jmeridth:
`make test` passes locally.  Gotta research why the UI and Codegen tests are failing in CI.

terrytangyuan:
>  -// Code generated by mockery v2.37.1. DO NOT EDIT.
+// Code generated by mockery v2.26.0. DO NOT EDIT.

The codegen failure is caused by an older version of mockery

jmeridth:
> > -// Code generated by mockery v2.37.1. DO NOT EDIT.
> > +// Code generated by mockery v2.26.0. DO NOT EDIT.
> 
> The codegen failure is caused by an older version of mockery

Yep.  Missed one. :smile:  Fixed and pushed.

Joibel (commented):
Did you manage to work out why all the `.spelling` changes were needed? Or rather why they weren't needed before.

jmeridth:
@Joibel no.  My assumption, still needs verification, is that spelling was only validating modified files and due to the size of my changes, it included many files.  I'm still trying to verify.  The Makefile command [`docs-spellcheck`](https://github.com/argoproj/argo-workflows/blob/f5b6b17c44a8152aaa0a0fdb92adfe1aee5c7991/Makefile#L685) isn't showing much to me except files to exclude.

Joibel:
> @Joibel no. My assumption, still needs verification, is that spelling was only validating modified files and due to the size of my changes, it included many files. I'm still trying to verify. The Makefile command [`docs-spellcheck`](https://github.com/argoproj/argo-workflows/blob/f5b6b17c44a8152aaa0a0fdb92adfe1aee5c7991/Makefile#L685) isn't showing much to me except files to exclude.

The spelling check checks all files regardless of modification, but has some specific ignored files listed. One of your additions is `Ansible` which is only in `docs/README.md` which should be being ignored, so somehow I think your process which has caused you to add stuff to `.spelling` is bypassing the `-not -name` bits of mdspell.

I attempted to reproduce why you'd had to add them but failed:

```
$ git checkout jm-readthedocs
$ git checkout origin/main .spelling
$ mdspell --ignore-numbers --ignore-acronyms --en-us --no-suggestions --report $(find docs -name '*.md' -not -name upgrading.md -not -name README.md -not -name fields.md -not -name upgrading.md -not -name swagger.md -not -name executor_swagger.md -not -path '*/cli/*')
>> 131 files are free from spelling errors
```

I'm all for taking out some of the exclusions from the spelling checker, and making `README.md` and probably `upgrading.md` be spellchecked, it seems reasonable. Especially as `upgrading.md` is in the exclusion list twice. I'm not bothered if it goes in like it is - really just curious.

.spelling
```diff
@@ -220,20 +257,3 @@ workflow-controller-configmap
 WorkflowTemplate
 WorkflowTemplates
 yaml
-idempotence
```
jmeridth, line 223: I alphabetized these.  WHERE THEY BELONG 😄

jmeridth: Would like to alphabetize no matter case. thoughts?

agilgur5:
> But those versions have to have all of the changes in this PR or they will fail

They need to have the build changes, but not necessarily all of the other changes.

> So this idea of versioned docs may only apply going forward.

I think we should create a tracking issue for this and slowly backport changes for v3 at least. That was my plan to backport all docs fixes to older versioned docs as well.

agilgur5 (commented):
I still need to review the build changes (including the `.spelling` discrepancies pointed out above and in https://github.com/argoproj/argo-workflows/pull/12211#issuecomment-1814860395), but I reviewed all the content this look through, only a couple changes needed.

For other reviewers, note that the vast majority of this PR is automated changes to the SDKs which are one-liner comment changes to the new docs link. Can skip all the `sdk/` paths as such

agilgur5 (commented):
We also need to remove the [`gh-pages` publish action](https://github.com/argoproj/argo-workflows/blob/76f4d1f9ca53b8ac48965e12c15454e544936c49/.github/workflows/docs.yaml#L47) in this PR so that #12362 isn't overwritten later

.github/workflows/docs.yaml
```diff
@@ -44,10 +44,3 @@ jobs:
           name: docs
           path: site
           if-no-files-found: error
-      - name: Publish to GH Pages (when on main)
```
agilgur5, line 47: Oh I forgot to mention in my previous comment; with this step removed, we can now take out the `permissions` section on this job (`docs`), which will make us more secure per least-privilege.

docs/requirements.txt
```diff
@@ -0,0 +1,7 @@
+mkdocs==1.5.0
+mkdocs-material==8.2.6
+mkdocs-spellcheck==0.2.1
```
agilgur5, line 3: It doesn't seem like we actually use `mkdocs-spellcheck` as it's not listed in our `mkdocs.yml` [as a plugin](https://github.com/pawamoy/mkdocs-spellcheck/tree/1.0.2?tab=readme-ov-file#usage)
```suggestion
```

docs/requirements.txt
```diff
@@ -0,0 +1,7 @@
...
+markdown_include==0.6.0
+pygments==2.15.0
+jinja2==3.0.3
+markdown==3.3.7
```
agilgur5, line 7: `mkdocs-material` already [depends on](https://github.com/squidfunk/mkdocs-material/blob/43ed6fac8dfc5385a9bdf22d8bf207c868c23a80/requirements.txt#L22) `mkdocs`, `pygments`, `jinja2`, and `markdown`, so we don't need to explicitly specify them in our own `requirements.txt` (and `mkdocs` itself also [depends on](https://github.com/mkdocs/mkdocs/blob/7186f4ce9c88e7c445ef90fee85c065ca297a7c8/pyproject.toml#L35) `jinja2` and `markdown`)

```suggestion
```

docs/requirements.txt
```diff
@@ -0,0 +1,7 @@
+mkdocs==1.5.0
```
agilgur5, line 1: same as above, we don't need to specify `mkdocs` directly as `mkdocs-material` depends on it

```suggestion
```

docs/requirements.txt
```diff
@@ -0,0 +1,7 @@
...
+mkdocs-material==8.2.6
+mkdocs-spellcheck==0.2.1
+markdown_include==0.6.0
```
agilgur5, line 4: `markdown_include` wasn't in the `Makefile` [above](https://github.com/argoproj/argo-workflows/pull/12360/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L715). We don't use it either as it's not in our `mkdocs.yml` [as a plugin](https://github.com/cmacmackin/markdown-include/tree/v0.8.1?tab=readme-ov-file#usage) and this syntax isn't used in the docs afaik

```suggestion
```

mkdocs.yml
```diff
@@ -34,6 +34,13 @@ markdown_extensions:
   - pymdownx.details
   - toc:
       permalink: true
+validation:
+  omitted_files: warn
+  absolute_links: warn
+  unrecognized_links: warn
+exclude_docs: |
+  proposals/
+
```
agilgur5, line 43: nice addition!

agilgur5 (commented):
A few small modifications to the build, otherwise the only leftover thing is the `.spelling` changes as per above

agilgur5:
> The spelling check checks all files regardless of modification, but has some specific ignored files listed. One of your additions is `Ansible` which is only in `docs/README.md` which should be being ignored, so somehow I think your process which has caused you to add stuff to `.spelling` is bypassing the `-not -name` bits of mdspell.

This _might_ be due to differences in GNU `find` vs Mac `find` as JM is currently running natively on Intel Mac

jmeridth:
@agilgur5 made you co-author on latest commit cleaning up `requirements.txt` and docs GitHub Action permissions.  Build succeeded in my readthedocs fork.  Excellent suggestions.  Thank you.

jmeridth:
Testing spelling changes in devcontainer now to see if my OSX Intel `find` command is the issue.

jmeridth:
@Joibel @agilgur5 I have reverted the `.spelling` file changes.  I ran the `make docs-spellcheck` command in the devcontainer with the one from the `main` branch and no issues.  I will be using the devcontainer from now on for local development.  Thank you @agilgur5 for the nudge and tip to build using the CLI.  Was much faster.

.github/workflows/docs.yaml
```diff
@@ -12,9 +12,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-permissions:
-  contents: read
```
agilgur5, line 16: oh wrong permissions; this is required/suggested by OpenSSF Scorecard as least privilege on all top-level GH Workflows. (As one should never override the top-level, only on a per job level)

I meant lines 18-19 below this can be removed, the `write` permissions that were specifically used for the `docs` GH Job, which was only needed to write to the `gh-pages` branch.

Summarizing, the `read` permissions here should be left in, the `write` permissions below should be removed.

jmeridth: @agilgur5 I'll put the read permissions back.  Wont' removing the `write` permissions cause the `Upload Docs Site` step to fail?  [Its README](https://github.com/actions/upload-artifact?tab=readme-ov-file) doesn't make me think so but was curious.  Putting `read` back and removing `write`.

agilgur5: No, `actions/upload-artifact` is CI/GHA specific; i.e. it's writing an artifact to GHA which actually doesn't need extra permissions; similar to how writing to the GHA cache doesn't.

`contents: write` [specifically](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) is the ability to write to the repo _itself_ (GHA's artifact server is independent of that)

agilgur5 (approved):
LGTM. Thanks for getting this done and all the iteration JM! 🚀

agilgur5:
Only thing I'm wondering about is if giscus will end up creating new GH Discussions for the new URLs and each version of the URLs? I don't think giscus is a huge deal either way, but I am wondering how that will work out

terrytangyuan (approved):
LGTM

jmeridth:
It's alive.  https://argo-workflows.readthedocs.io/en/latest/ 

Resyncing #12362 now and will open for merging.

agilgur5:
> Only thing I'm wondering about is if giscus will end up creating new GH Discussions for the new URLs and each version of the URLs? I don't think giscus is a huge deal either way, but I am wondering how that will work out

Giscus is indeed creating new URLs now, and seemingly for each version too 😕 See #12451 as an example of the first Giscus discussion created on RTD.

I wonder if we can configure Giscus to use existing discussions or at least use the same name/version for all new ones instead of different ones per version. It currently seems to naively use the full URL path

agilgur5:
Investigated Giscus, can see the [`mkdocs-material` docs](https://squidfunk.github.io/mkdocs-material/setup/adding-a-comment-system/?h=giscus#giscus-integration) and the [Giscus docs/setup](https://giscus.app/), which is integrated on [this line of code](https://github.com/argoproj/argo-workflows/blob/1968436a6e021adc4785ec287635a6136574c5ac/docs/overrides/main.html#L8).

Unfortunately, Giscus can only use the full `pathname`, the full URL, the full `title`, or a [hash](https://github.com/giscus/giscus/blob/main/ADVANCED-USAGE.md#data-strict), so it won't be able to match the old page discussions out-of-the-box.

That being said, Giscus performs a search on the GH Discussions to match a page to a discussion, so we can modify the old discussions to have the new URLs in them, which should make the search work.

Main problem with that is, well, I don't have permissions to edit Discussion titles or opening comments, as that's only granted with write permissions (c.f. https://github.com/argoproj/argoproj/pull/243).

If someone can grant me those permissions temporarily, I can manually edit all the existing page discussions to match up

.spelling
agilgur5: Automated this in #12521

agilgur5:
> That being said, Giscus performs a search on the GH Discussions to match a page to a discussion, so we can modify the old discussions to have the new URLs in them, which should make the search work.
> 
> Main problem with that is, well, I don't have permissions to edit Discussion titles or opening comments, as that's only granted with write permissions (c.f. [argoproj/argoproj#243](https://github.com/argoproj/argoproj/pull/243)).

As I now have permissions (https://github.com/argoproj/argoproj/issues/277), I did do this for a few discussions, but ran into a couple of issues:
1. When a discussion is edited, it moves to the front page of the discussion as it was "recently updated"
1. There are like 100 Giscus discussions. That's a lot to manually go through (although I could possibly write a script that uses the GH API for it), and editing all of them will clog up like 5 pages of discussions, per above

Given that Giscus discussions are also not necessarily made for Q&A, as there are multiple Q&As in one, no "mark as answer" possible, and not easy to search since there can be dozens of comments, I am considering removing Giscus entirely. It is a source of interactivity for new users, but I don't think a very efficient one for getting questions answered. Pointing to people to search issues and discussions and open a new one if they don't see an answer would be more optimal, but not necessarily that simple UX-wise

agilgur5:
> Given that Giscus discussions are also not necessarily made for Q&A, as there are multiple Q&As in one, no "mark as answer" possible, and not easy to search since there can be dozens of comments, I am considering removing Giscus entirely. It is a source of interactivity for new users, but I don't think a very efficient one for getting questions answered. Pointing to people to search issues and discussions and open a new one if they don't see an answer would be more optimal, but not necessarily that simple UX-wise

PR to remove Giscus and replace with links to search Discussions and Slack: #13138

Issue #12366, opened 2023-12-15
Do not print sensitive information during executor initialization


melden19:
# Summary

Currently, after the executor (argoexec) initialization the logs with the configuration is printed. One of the logged fields is the `template` and potentially it can have env variables with sensitive information.

Example: 
```
time="2023-12-15T10:16:25.427Z" level=info msg="Executor initialized" deadline="0001-01-01 00:00:00 +0000 UTC" includeScriptOutput=false namespace=melden-sandbox podName=hello-world-k5rh5 template="{\"name\":\"whalesay\",\"inputs\":{},\"outputs\":{},\"metadata\":{},\"container\":{\"name\":\"\",\"image\":\"docker/whalesay:latest\",\"command\":[\"cowsay\"],\"args\":[\"hello world\"],\"env\":[{\"name\":\"API_KEY\",\"value\":\"\\u003csensitive_data\\u003e\"}],\"resources\":{}}}" version="&Version{Version:v3.4.4,BuildDate:2023-02-23T18:22:48Z,GitCommit:96276a4816d85ddd06bbe3e3fedcd0f9ad9851be,GitTag:v3.4.4,GitTreeState:clean,GoVersion:go1.18.10,Compiler:gc,Platform:linux/amd64,}"
```
Ideally, it shouldn't happen and all of the sensitive data should be stored as part of the k8s secret so only reference to this secret will be exposed, but in my case, the workflows get triggered using HTTP API and some of the creds come from the request.
Then these logs are collected and stored by the monitoring tool or log scraper. 

IMO the `template` field can be moved to the `debug` loglevel:
```go
	// after
	log.Debugf("template: %s", wfv1.MustMarshallJSON(&wfExecutor.Template))
	log.
		WithField("version", version.String()).
		WithField("namespace", namespace).
		WithField("podName", podName).
		// before
		// WithField("template", wfv1.MustMarshallJSON(&wfExecutor.Template)).
		WithField("includeScriptOutput", includeScriptOutput).
		WithField("deadline", deadline).
		Info("Executor initialized")
```

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

PR #12368, opened 2023-12-15
fix: move log with potential sensitive data to debug loglevel. Fixes: #12366

melden19:
… #12366

Fixes #TODO


cmd/argoexec/commands/root.go
```diff
@@ -129,11 +129,11 @@ func initExecutor() *executor.WorkflowExecutor {
 		progressFileTickDuration,
 	)
 
+	log.Debugf("template: %s", wfv1.MustMarshallJSON(&wfExecutor.Template))
 	log.
 		WithField("version", version.String()).
 		WithField("namespace", namespace).
 		WithField("podName", podName).
-		WithField("template", wfv1.MustMarshallJSON(&wfExecutor.Template)).
```
sarabala1979, line 136: @melden19 Can we just print the Template name information?

cmd/argoexec/commands/root.go
```diff
@@ -129,11 +129,11 @@ func initExecutor() *executor.WorkflowExecutor {
 		progressFileTickDuration,
 	)
 
+	log.Debugf("template: %s", wfv1.MustMarshallJSON(&wfExecutor.Template))
```
terrytangyuan, line 132: Agreed. We probably don’t need this log. Consider removing it completely or just printing of template name.

terrytangyuan (approved):

PR #12394, opened 2023-12-21
docs: Fix incorrect link in edit button

terrytangyuan:
Currently, when click "edit" on any of the docs page from the website, it will direct you to a 404 page since it uses hard-coded "master" branch. This PR fixes that by changing it to "main" branch URL instead.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,6 @@
 site_name: Argo Workflows - The workflow engine for Kubernetes
 repo_url: https://github.com/argoproj/argo-workflows
+edit_uri: https://github.com/argoproj/argo-workflows/edit/main/docs
 strict: true
 theme:
   name: material


agilgur5:
Oh I don't think I ever use that button, good catch!

agilgur5 (approved):

mkdocs.yml
```diff
@@ -1,5 +1,6 @@
 site_name: Argo Workflows - The workflow engine for Kubernetes
 repo_url: https://github.com/argoproj/argo-workflows
+edit_uri: https://github.com/argoproj/argo-workflows/edit/main/docs
```
agilgur5, line 3:
```suggestion
edit_uri: edit/main/docs/
```

this can be simplified a bit per [the `mkdocs` docs](https://www.mkdocs.org/user-guide/configuration/#edit_uri)

also it's supposed to have a `/` at the end by default (although I'm not sure if not having it will break anything or if it'll auto-add it)

agilgur5: I mean that seems like a bit of a straw man to me, a one line change does not significantly change any focus on either this PR or others. I've also been leading review on all the other website changes, and those have lots more comments, so I wouldn't say this affected my focus either.

The one-line change is just DRY and closer to the default per the `mkdocs` docs; I don't really see much reason to reject it 🤷

agilgur5: > also it's supposed to have a `/` at the end by default as it will append the page's subpath to that (although I'm not sure if not having it will break anything or if it'll auto-add it if missing)

it seems to have worked without a `/` as the edit link post-change works now

terrytangyuan: Sorry what I meant originally was to that I knew this exact change works since I use it for other project as well so I wanted to avoid any optimizations that I haven't tried personally to avoid any side effect.

agilgur5: Oh gotcha, thanks for the clarification, that makes more sense. I've never used `mkdocs` extensively before Argo actually, so I just went to the docs and pulled that URI out

PR #12399, opened 2023-12-22
chore: ask bug reporters to search for existing issues

Joibel:
Add a tick box to request searches for duplicates before submitting


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -11,6 +11,8 @@ body:
           required: true
         - label: I can confirm the issues exists when I tested with `:latest`
           required: true
+        - label: I have searched for existing issues which may be the same as this
```
agilgur5, line 14:
```suggestion
        - label: I have searched existing issues and could not find a match for this bug
```

a tad more specific

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -11,6 +11,8 @@ body:
           required: true
         - label: I can confirm the issues exists when I tested with `:latest`
```
agilgur5, line 12:
```suggestion
        - label: I can confirm the issue exists when I tested with `:latest`
```

if you don't mind, there's an existing spelling mistake to fix up here

agilgur5 (commented):
I haven't seen too many duplicate issues so I don't think this is a big problem, but can certainly add this. Note that GH usually prompts users to do this as well.

Also we'll want to add this checkbox to `regression.yaml` as well

.github/ISSUE_TEMPLATE/bug_report.yaml
Joibel: Done

.github/ISSUE_TEMPLATE/bug_report.yaml
Joibel: Done

Joibel:
> I haven't seen too many duplicate issues so I don't think this is a big problem, but can certainly add this. Note that GH usually prompts users to do this as well.

It was just I'd seen one or two, and I'd rather other people put the effort in. The tick box has helped me in the past to avoid dupes, so I thought it worthwhile

> Also we'll want to add this checkbox to `regression.yaml` as well

Done

agilgur5 (approved):

agilgur5:
We do also have the problem that some users completely ignore the checkboxes. The `:latest` checkbox is pretty much always checked but has not actually been tested in many cases (especially when the "Version" reported in the input box below is significantly lower than a currently supported version)

juliev0 (approved):

Joibel:
> We do also have the problem that some users completely ignore the checkboxes. The `:latest` checkbox is pretty much always checked but has not actually been tested in many cases (especially when the "Version" reported in the input box below is significantly lower than a currently supported version)

Yeah, I know. I was going to bring this up at next contrib meeting. I don't have a good answer though.

PR #12400, opened 2023-12-23
fix(controller): option to not watch configmap

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/11657

Without setting this env variable argo-workflows is unusable with k8s >= 1.27 as logs produced are excessive and cpu climbs


docs/environment-variables.md
```diff
@@ -52,6 +52,7 @@ This document outlines environment variables that can be used to customize behav
...
 | `SEMAPHORE_NOTIFY_DELAY`                 | `time.Duration`     | `1s`                                                                                        | Tuning Delay when notifying semaphore waiters about availability in the semaphore                                                                                                                                                                                        |
+| `WATCH_CONTROLLER_CONFIGMAP`                | `bool`              | `true`                                                                                     | Whether controller configMap changes should cause the controller pod to restart.                                                                                                                                                                                                                              |
```
agilgur5, line 55: This doesn't cause it to restart, it actually modifies values at run-time in memory. Also, as I've noted in #11855, this `RetryWatcher` is (currently) also responsible for semaphore ConfigMap updates as well, so it will disable that functionality as well.

As such we should reword this a good bit:
```suggestion
| `WATCH_CONFIGMAPS`                | `bool`              | `true`                                                                                     | Whether to watch the Controller's ConfigMap and semaphore ConfigMaps for run-time changes. When disabled, the Controller will only read ConfigMaps once and will have to be manually restarted to pick up new changes.                                                                                                  |
```

agilgur5 (commented):
> Fixes https://github.com/argoproj/argo-workflows/issues/11657

This is actually a workaround, not a pure fix. #11855 should be the pure fix, but I think this is a good workaround to have in the interim.

Could you please make the suggested changes and make sure CI passes?

> feat(controller): option to not watch configmap

Also let's label this as a `fix` so that we backport it (although yes, technically it's a feature that enables a fix / workaround)

agilgur5:
Please make sure CI passes. The docs build is [still failing](https://github.com/argoproj/argo-workflows/actions/runs/7434206363/job/20228125600?pr=12400) as you'll need to update [`.spelling`](https://github.com/argoproj/argo-workflows/blob/93914261cff4216561c89c1f5f6123e7ad0d5f61/.spelling) with `ConfigMap` and `ConfigMaps`.

tooptoop4:
@agilgur5 updated

agilgur5:
CI is still failing on [missing DCO](https://github.com/argoproj/argo-workflows/pull/12400/checks?check_run_id=20674359667).

Again, please make sure CI passes.

pquadri:
@agilgur5 @tooptoop4 would you mind if i open an MR of the same content and get the CI passing?

tooptoop4:
@pquadri you can

PR #12397, opened 2023-12-22
fix: Add missing 'archived' prop for ArtifactPanel component. Fixes #12331

Garett-MacGowan:
Fixes #12331

### Motivation

A missing `archived` prop in the UI's `ArtifactPanel` component was causing an improper URL to be used to fetch the artifact (assuming the workflow was archived).

### Modifications

Added the missing `archived` prop to the `ArtifactPanel` component.

### Verification

I added tests to the argo server test suite to account for the URL structure which was previously untested. I called this test `GetArtifactByNodeID`, in accordance with naming conventions in similar tests.

For the UI, I ran the `TestArtifactServerArtifactPassing` test (which contains the `GetArtifactByNodeID` test mentioned above) using:
```
make start UI=true PROFILE=postgres
make TestArtifactServerArtifactPassing
```
Once it completed, I manually verified that the artifact in the ArtifactPanel was downloadable.

image

image

(download button)
image

Issue #12331, opened 2023-12-07
Artifact in workflow diagram not working when workflow is archived and deleted from cluster


dmarquez-splunk:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

Artifact URLs used in the workflow diagram's artifact icons seem to no longer work once a workflow is archived and deleted from the cluster (Returns Internal Server Error). However, if you click on the node from which the artifact is output from and attempt to download the artifact from the right-hand bar's input/output's menu then the artifact successfully downloads.

Difference in the 2 links:

link from workflow diagram's artifact icon: `https://<argo-host>/artifact-files/<namespace>/workflows/<workflow-name>/<workflow-name>-1855980756/outputs/result_html`

link from node's input/output menu: `https://<argo-host>/artifact-files/<namespace>/archived-workflows/f7655950-a426-4fb3-a82d-eb1f53dfb948/<workflow-name>-1855980756/outputs/result_html`

As you can see the link in the workflow diagram's artifact icon is still trying to reference the worklow as a non-archived workflow. Whereas the link in the node's input/output menu is now referencing it as an archived-workflow in the API. I suspect this is causing the argo-server to return Internal Server Error when clicking on the link from the workflow diagram's artifact icon.

**NOTE**: In order to reproduce the workflow and its artifacts must be archived AND deleted from the cluster in order to reproduce

### Version

v3.5.2

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: artifact-passing-
spec:
  entrypoint: artifact-example
  templates:
  - name: artifact-example
    steps:
    - - name: generate-artifact
        template: whalesay
    - - name: consume-artifact
        template: print-message
        arguments:
          artifacts:
          # bind message to the hello-art artifact
          # generated by the generate-artifact step
          - name: message
            from: "{{steps.generate-artifact.outputs.artifacts.hello-art}}"

  - name: whalesay
    container:
      image: docker/whalesay:latest
      command: [sh, -c]
      args: ["cowsay hello world | tee /tmp/hello_world.txt"]
    outputs:
[...]

PR #11947, opened 2023-10-05
fix: ArtifactGC Fails for Stopped Workflows. Fixes #11879

Garett-MacGowan:
Fixes #11879

### Motivation

Currently, stopped workflows will happily save artifacts to a storage bucket but the artifacts won't get garbage collected after workflow deletion. The changes make it so that a stopped workflow's artifacts are properly garbage collected upon workflow deletion.

### Modifications

The `operator.go`'s call to `woc.taskResultReconciliation()` previously occurred after Artifact Garbage Collection (`ArtifactGC`) and an early return due to `if woc.wf.Labels[common.LabelKeyCompleted] == "true"`, leading to a scenerio in which garbage collection would fail upon workflow deletion. This in turn would cause workflow deletion to get blocked by the ArtifactGC finalizer. `woc.taskResultReconciliation()` is responsible for updating the node status in the workflow based on the artifact outputs saved during workflow execution with `reportOutputs()` in `executor.go`. 

### Verification

I tested the changes by adding an `e2e` test called `TestStoppedWorkflow` which uses a custom workflow with the following logic.

1) Ensure no artifact exists in the bucket.
2) Start the workflow.
3) Workflow runs and creates a test artifact.
4) Verifiy the artifact exists in the bucket.
4) The workflow submits a CLI-based stop event for itself.
5) The workflow is deleted by the test.
6) Artifacts are verified to have been deleted.
7) Finalizer is removed incase the test failed.

---

I made some additional minor tweaks to other files, most notably, `timeout = v - 30*time.Second + defaultTimeout` in `when.go` which would cause custom timeout settings to be incorrect when using the `WaitForWorkflow()` e2e testing utility.

PR #12402, opened 2023-12-23
fix: fix missing artifacts for stopped workflows. Fixes #12401

### LINKED ISSUE #12401: Stopped archived workflow missing artifacts in UI
### Pre-requisites

When a workflow is stopped and archived, artifacts are missing.

I expect artifacts to be present for stopped & archived workflows.

I already implemented a fix, this issue is here only to reference in the commit and PR.

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: wf-stopped-
spec:
  workflowMetadata:
    labels:
      workflows.argoproj.io/workflow: "wf-stopped"
  entrypoint: wf-stopped-main
  serviceAccountName: argo
  executor:
    serviceAccountName: default
  volumeClaimTemplates:
    - metadata:
        name: artifacts
      spec:
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 1Gi
  templates:
    - name: wf-stopped-main
      dag:
        tasks:
          - name: create-artifact
            template: artifact-creator
          - name: delay-stop-workflow
            template: delay-stop
          - name: stop-workflow
            template: workflow-stopper
            dependencies: [delay-stop-workflow]
    - name: delay-stop
      container:
        image: alpine:latest
        volumeMounts:
          - name: artifacts
            mountPath: /mnt/vol
        command: [sh, -c]
        args:
          - |
            echo "Delaying workflow stop"
            ls /mnt
            x=0
            while [ $x -le 60 ]
            do
              sleep 1
              if [ -f "/mnt/vol/test.txt" ]; then
                echo "Artifact found in shared volume"
                break
              fi
              x=$(( $x + 1 ))
            done
    - name: workflow-stopper
      container:
        image: argoproj/argocli:latest
        args: 
          - stop
          - -l
          - workflows.argoproj.io/workflow=wf-stopped
          - --namespace=argo
          - --loglevel=debug
    - name: artifact-creator
      metadata:
        labels:
          template: "artifact-creator"
      container:
        image: alpine:latest
        volumeMounts:
          - name: artifacts
            mountPath: /mnt/vol
        command: [sh, -c]
        args:
          - |
            echo 'testing' > /mnt/vol/test.txt
            echo "Artifact saved to /mnt/vol/test.txt"
            echo "Pretending to continue to do work."
            ls /mnt
            while :
            do
              sleep 1
            done
      outputs:
        artifacts:
          - name: artifact-creator
            path: /mnt/vol/test.txt
            s3:
              key: artifact-creator
              bucket: my-bucket-3
              endpoint: minio:9000
              insecure: true
              accessKeySecret:
                name: my-minio-cred
                key: accesskey
              secretKeySecret:
                name: my-minio-cred
                key: secretkey
            archive:
              none: {}
```

```text
N/A
```

```text
N/A
```

Garett-MacGowan:
Fixes #12401

### Motivation

Stopped workflows would not have artifacts available in the UI, this can can be particularly annoying when debugging a workflow.

### Modifications

Ensure workflow is archived only once all task results are finished processing.

image

### Verification

Added a argo server test to ensure that the artifact is available. Manually verified that artifacts are now present in the UI.


Garett-MacGowan:
@juliev0 Here's the fix for what I was hypothesizing in #12331

juliev0:
Thanks for being on top of any new regressions that may have been caused by your PR! I glanced the other day but then it sounded like you were saying that it was a UI issue. Did you find that that was not the case after all?

Garett-MacGowan:
> Thanks for being on top of any new regressions that may have been caused by your PR! I glanced the other day but then it sounded like you were saying that it was a UI issue. From going through that Issue now, it looks like this is not a direct fix to that issue, right? In that case, do you want to remove "Fixes" from your description and instead say something like "Related to"? Otherwise, this will automatically close that Issue.

@juliev0 

These are two separate issues! #12331 must have been caused by another PR. This PR fixes @12401, which occurs when you stop a workflow. My previous PR didn't guarantee that artifacts would get archived for stopped workflows, although it fixed the garbage collection.

juliev0:
> > Thanks for being on top of any new regressions that may have been caused by your PR! I glanced the other day but then it sounded like you were saying that it was a UI issue. From going through that Issue now, it looks like this is not a direct fix to that issue, right? In that case, do you want to remove "Fixes" from your description and instead say something like "Related to"? Otherwise, this will automatically close that Issue.
> 
> @juliev0
> 
> These are two separate issues! #12331 must have been caused by another PR. I fixed that issue in #12397. This PR fixes #12401, which occurs when you stop a workflow. My previous PR (#11947) didn't guarantee that artifacts would get archived for stopped workflows, although it fixed the garbage collection.

Oh, okay!

workflow/controller/operator.go
```diff
@@ -802,7 +802,7 @@ func (woc *wfOperationCtx) persistUpdates(ctx context.Context) {
 
 func (woc *wfOperationCtx) checkTaskResultsCompleted() bool {
 	taskResultsCompleted := woc.wf.Status.GetTaskResultsCompleted()
-	if len(taskResultsCompleted) == 0 {
+	if len(taskResultsCompleted) == 0 && woc.wf.Status.Outputs != nil {
```
juliev0, line 805: It does seem like we want to make sure there are actually Outputs.

But now it makes me wonder, when do the Outputs get set? Is it possible to arrive here when `Outputs==nil` because they haven't been set yet?

Garett-MacGowan: @juliev0 Happy Holidays!

Outputs are first initialized and populated (at a node level) in `taskResultReconciliation()`. Outputs are subsequently initialized and populated to the global scope `wf.Status.Outputs` in `woc.podReconciliation()` via `woc.addOutputsToGlobalScope(newState.Outputs)`, assuming `Parameters` or `Artifacts` are defined for any node. Therefore, `Outputs==nil` when either `woc.TaskResultReconciliation()` hasn't been run yet, or `Parameters` or `Artifacts` aren't defined for any node in the workflow.

As long as `checkTaskResultsCompleted()` has been run after `taskResultReconciliation()`, `&& woc.wf.Status.Outputs != nil` is meaningful.

`checkTaskResultsCompleted()` is used in two places:

1) Before `woc.garbageCollectArtifacts(ctx)` (after `taskResultReconciliation()`)
```
	// Do artifact GC if all task results are completed.
	if woc.checkTaskResultsCompleted() {
		if err := woc.garbageCollectArtifacts(ctx); err != nil {
			woc.log.WithError(err).Error("failed to GC artifacts")
			return
		}
	} else {
		woc.log.Debug("Skipping artifact GC")
	}
```

2) Before `woc.deletetaskResults(ctx)` as part of `persistUpdates(ctx)` (called in conjunction with `woc.wf.Status.Phase.Completed()`, which ensures that the workflow has either succeeded, failed, or errored, and therefore after `checkTaskResultsCompleted()` has had a chance to run).
```
	// Make sure the TaskResults are incorporated into WorkflowStatus before we delete them.
	if woc.wf.Status.Phase.Completed() && woc.checkTaskResultsCompleted() {
		if err := woc.deleteTaskResults(ctx); err != nil {
			woc.log.WithError(err).Warn("failed to delete task-results")
		}
	}
```

juliev0: Happy holidays! 

Trying to follow. So, based on your statement about how "Outputs are subsequently initialized and populated to the global scope `wf.Status.Outputs` in `woc.podReconciliation()`", doesn't that mean that `Outputs` could also be nil in the case that `taskResultReconciliation()` had been run but `podReconciliation()` had not?

Garett-MacGowan: Oops, I meant to say:

2) Before `woc.deletetaskResults(ctx)` as part of `persistUpdates(ctx)` (called in conjunction with `woc.wf.Status.Phase.Completed()`, which ensures that the workflow has either succeeded, failed, or errored, and therefore after **`woc.podReconciliation()`** has had a chance to run).

I missed the obvious (case 3), which I just introduced, inside `markWorkflowPhase`, which is called at the end of the operator loop under various conditions.
```
	switch phase {
	case wfv1.WorkflowSucceeded, wfv1.WorkflowFailed, wfv1.WorkflowError:
		// wait for all daemon nodes to get terminated before marking workflow completed & make sure task results are complete as well.
		if phase.Completed() && !woc.hasDaemonNodes() && woc.checkTaskResultsCompleted() {
			woc.log.Info("Marking workflow completed")
```

Good catch, yes, you are right. The case you identified can happen for `woc.garbageCollectArtifacts(ctx)` (case 1), which gets called right after `woc.taskResultReconciliation()`, but before `woc.podReconciliation()`. I think a race condition could occur where the `taskResultInformer` hasn't received an update yet, and therefore garbage collection is allowed to happen because wf.Status.Outputs == nil.

Since `woc.wf.Status.Outputs != nil` is meant to check that outputs are in global scope (for archiving purposes), maybe it's best that we pull that logic up to `if phase.Completed() && !woc.hasDaemonNodes() && woc.checkTaskResultsCompleted() {`.

For example:
```
	switch phase {
	case wfv1.WorkflowSucceeded, wfv1.WorkflowFailed, wfv1.WorkflowError:
		// wait for all daemon nodes to get terminated before marking workflow completed & make sure task results are complete as well.
		if phase.Completed() && !woc.hasDaemonNodes() && (woc.checkTaskResultsCompleted() || woc.wf.Status.Outputs == nil) {
			woc.log.Info("Marking workflow completed")
```

Garett-MacGowan: The above doesn't work. I'm thinking through it and will propose another solution.

Garett-MacGowan: @juliev0 I suggest the following changes. Notice the new `considerStatusOutputs bool` parameter for `checkTaskResultsCompleted`:

```
func (woc *wfOperationCtx) checkTaskResultsCompleted(considerStatusOutputs bool) bool {
	taskResultsCompleted := woc.wf.Status.GetTaskResultsCompleted()

	if considerStatusOutputs && woc.wf.Status.Outputs != nil && len(taskResultsCompleted) == 0 {
		return false
	}
	if !considerStatusOutputs && len(taskResultsCompleted) == 0 {
		return false
	}

	for _, completed := range taskResultsCompleted {
		if !completed {
			return false
		}
	}
	return true
}
```

Case 1)

```
	// Do artifact GC if all task results are completed.
	if woc.checkTaskResultsCompleted(false) {
		if err := woc.garbageCollectArtifacts(ctx); err != nil {
			woc.log.WithError(err).Error("failed to GC artifacts")
			return
		}
	} else {
		woc.log.Debug("Skipping artifact GC")
	}
```

Case 2)
```
	// Make sure the TaskResults are incorporated into WorkflowStatus before we delete them.
	if woc.wf.Status.Phase.Completed() && woc.checkTaskResultsCompleted(true) {
		if err := woc.deleteTaskResults(ctx); err != nil {
			woc.log.WithError(err).Warn("failed to delete task-results")
		}
	}
```

Case 3)
```
	switch phase {
	case wfv1.WorkflowSucceeded, wfv1.WorkflowFailed, wfv1.WorkflowError:
		// wait for all daemon nodes to get terminated before marking workflow completed & make sure task results are complete as well.
		if phase.Completed() && !woc.hasDaemonNodes() && woc.checkTaskResultsCompleted(true) {
			woc.log.Info("Marking workflow completed")
```

juliev0: Sorry, going back....let's start from "what was the original intention of this new check?"

Garett-MacGowan: @juliev0 This is to ensure that the workflow is only marked as complete and archived once the ouputs have been reflected in the status.

Garett-MacGowan: `len(taskResultsCompleted) == 0` can happen if there are no outputs for any nodes, so we can't always say that task results have not completed if `len(taskResultsCompleted) == 0`. If `wf.Status.Outputs` is defined and `len(taskResultsCompleted) == 0`, then we can say that task results have not completed.

If `wf.Status.Outputs` is defined (outputs or parameters do exist in at least one node), and `taskResultsComplete` is an empty map, then we know that the wait containers aren't done executing.

I should have been more clear with the above, the new usage of woc.CheckTaskResultsCompleted(true) in `case wfv1.WorkflowSucceeded, wfv1.WorkflowFailed, wfv1.WorkflowError:` is to prevent the workflow from archiving before the status is updated with the outputs.

Garett-MacGowan: @juliev0 I just updated my last two comments since I did a really poor job in answering the first time.

workflow/controller/operator.go
```diff
@@ -802,7 +802,7 @@ func (woc *wfOperationCtx) persistUpdates(ctx context.Context) {
...
 		return false
 	}
```
juliev0, line 807: Sorry, do you mind if we first go back to the original code and talk about the case you were trying to take care of with `if len(taskResultsCompleted) == 0 {`?

juliev0: BTW, maybe we should rename `taskResultsCompleted` to something like `taskResultCompletionStatus` since it sounds like we're talking about the number completed here but we're not.?

juliev0: As far as I can tell, we're looking at the case in which no taskresults have ever even been initialized as incomplete, right?

Garett-MacGowan: Yeah, we could probably do with a rename.

Yes, we're looking at the case where no task results have been initialized as incomplete yet.

juliev0: Is this the case we're concerned about, or is it more generally speaking the case in which a task result would be initialized in the future and hasn't been yet? As far as I can tell, it's once a Pod exists that a new task result will get added? So, wouldn't it be the same case if say this map has one entry and it's completed, but another task result simply hasn't been initialized yet?

Garett-MacGowan: 💩 Yes, you're right. We should be handling that. I'm going to think on this a bit more & try to come up with something better.

Garett-MacGowan: I've flipped the way I'm looking at this and reformulated the check as `checkTaskResultsInProgress`. When a task result hasn't completed yet, it is added to a `TaskResultsInProgress` map. Once it is completed, it is removed from the map.

```
func (woc *wfOperationCtx) checkTaskResultsInProgress() bool {
	taskResultsInProgress := woc.wf.Status.GetTaskResultsInProgress()
	if len(taskResultsInProgress) != 0 {
		return true
	}
	return false
}
```

In all three cases, `!woc.checkTaskResultsInProgress()` is called. All tests seem to be passing & the code is much easier to understand. I'm re-running the test script from my last PR to ensure the garbage collection race condition doesn't exist. https://github.com/argoproj/argo-workflows/pull/11947#issuecomment-1805074067.

Assuming all the tests pass locally, I'll push and you can review @juliev0

Garett-MacGowan: @juliev0 The case of "no task results have been initialized as incomplete yet" doesn't actually seem to be an issue based on my testing.

workflow/controller/operator.go
```diff
@@ -240,7 +240,7 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 	woc.taskResultReconciliation()
 
 	// Do artifact GC if all task results are completed.
-	if woc.checkTaskResultsCompleted() {
+	if !woc.checkTaskResultsInProgress() {
```
juliev0, line 243: we don't need to check if `Phase.Completed()` for this case?

Garett-MacGowan: Pretty sure we don't need to because of the strategy check, but I see no harm in doing so.

Garett-MacGowan: The artifact gc strategy protects us for this case. We could conceivably add this though. It would make things a bit more efficient.

workflow/controller/operator.go
```diff
@@ -2314,8 +2303,8 @@ func (woc *wfOperationCtx) markWorkflowPhase(ctx context.Context, phase wfv1.Wor
 
 	switch phase {
 	case wfv1.WorkflowSucceeded, wfv1.WorkflowFailed, wfv1.WorkflowError:
-		// wait for all daemon nodes to get terminated before marking workflow completed
-		if markCompleted && !woc.hasDaemonNodes() {
+		// wait for all daemon nodes to get terminated before marking workflow completed & make sure task results are complete as well.
+		if phase.Completed() && !woc.hasDaemonNodes() && !woc.checkTaskResultsInProgress() {
```
juliev0, line 2307: We don't need `podReconciliation()` for this one?

workflow/controller/operator.go
```diff
@@ -788,7 +788,7 @@ func (woc *wfOperationCtx) persistUpdates(ctx context.Context) {
 	}
 
 	// Make sure the TaskResults are incorporated into WorkflowStatus before we delete them.
-	if woc.wf.Status.Phase.Completed() && woc.checkTaskResultsCompleted() {
+	if woc.wf.Status.Phase.Completed() && !woc.checkTaskResultsInProgress() {
```
juliev0, line 791: Do we also need to make sure that the podReconciliation has occurred to incorporate the taskResults into the Workflow's Status, or is that not needed based on where this is being called from?

workflow/controller/operator.go
juliev0: plus, if we have common logic between this and the other calls, we might be able to have a common function, which would be nice

juliev0: I'm actually kind of thinking about a common function which checks for the Workflow being completed, plus compares the number of pods that were run to the number of task results that got incorporated into the Workflow Status, which could be called `Workflow.ReconciliationComplete()` or something like that. Does that make any sense?

This method could be called at any place in the code, and it should work.

juliev0: I know in this case you were only concerned about which taskResults were saved at all, as opposed to which ones had actually gone through `podReconciliation()` as well, so maybe the common function wouldn't be used here if it doesn't make sense.

juliev0: (just editing my comment 2 comments back because it wasn't clear what i was talking about)

juliev0:
Something I alluded to in my other comment, but I'm kind of wondering if there could be a risk of only checking whether there's a taskresult in progress - that requires that it got added to the map in the first place. It may be okay if it's only called from a particular place in the code I suppose.

But I was thinking: isn't the simplest thing to check if the Workflow is complete, plus confirm that all TaskResults have been fully processed (therefore equal to the number of Pods run), which could be a function called from anywhere? 

Just let me know if you disagree! I haven't looked that carefully.

Garett-MacGowan:
> Something I alluded to in my other comment, but I'm kind of wondering if there could be a risk of only checking whether there's a taskresult in progress - that requires that it got added to the map in the first place. It may be okay if it's only called from a particular place in the code I suppose.
> 
> But I was thinking: isn't the simplest thing to check if the Workflow is complete, plus confirm that all TaskResults have been fully processed (therefore equal to the number of Pods run), which could be a function called from anywhere?
> 
> Just let me know if you disagree! I haven't looked that carefully.

> Something I alluded to in my other comment, but I'm kind of wondering if there could be a risk of only checking whether there's a taskresult in progress - that requires that it got added to the map in the first place. It may be okay if it's only called from a particular place in the code I suppose.
> 
> But I was thinking: isn't the simplest thing to check if the Workflow is complete, plus confirm that all TaskResults have been fully processed (therefore equal to the number of Pods run), which could be a function called from anywhere?
> 
> Just let me know if you disagree! I haven't looked that carefully.

What you are saying is probably the safest solution. Here's what I'm thinking now:

As part of `podReconciliation()`, outputs are added to the global scope, then, the node phase is updated based on the pod phase. If we observe that the node phase is fulfilled, then we know that `podReconciliation()` has been run for the last time for that node, and any outputs for the node at that time should be reflected in the global scope. One problem to consider, however, is what happens when outputs are sent to the `taskResultInformer` between when `taskResultReconciliation()` and `podReconciliation()` are called. Outputs sent to the `taskResultInformer` during this time won't be reflected in the node state, and therefore won't be reflected in the global scope. At this point, i'm considering the possibility of adding `woc.addOutputsToGlobalScope()` to `woc.taskResultReconciliation()`

Assuming the above issue is resolved. it should be okay to use something like:
```
func (woc *wfOperationCtx) checkReconciliationComplete() {
	numCompletedNodes := len(woc.wf.Status.Nodes.Filter(func(x wfv1.NodeStatus) interface{} {
		return x.Phase.Completed()
	}))
	return woc.wf.Status.Phase.Completed() && numCompletedNodes == woc.wf.Status.GetTaskResultsCompleted()
}
```

Thoughts?

juliev0:
Thanks for all the flexibility on the approach! I’m on vacation now, so may not get a chance to look at your response until after 1/1 unfortunately. I’ll get back to you then.

Garett-MacGowan:
> Thanks for all the flexibility on the approach! I’m on vacation now, so may not get a chance to look at your response until after 1/1 unfortunately. I’ll get back to you then.

Have a good vacation! I will take a crack at what I've suggested above.

Garett-MacGowan:
@juliev0 I had to make some tweaks the implementation I suggested above, but I think I've got something more along the lines of what you were thinking.

Garett-MacGowan:
Working through the test failures. They are to do with termination as far as I can tell.

Garett-MacGowan:
Unfortunately the number of pods run is not a reliable indicator of wait container execution. If the workflow is terminated as soon as it is created, the wait container does not start up.

I'm going to push an alternative solution.

workflow/controller/taskresult.go
```diff
@@ -86,6 +86,8 @@ func (woc *wfOperationCtx) taskResultReconciliation() {
 			if old.Outputs != nil && newNode.Outputs.ExitCode == nil { // prevent overwriting of ExitCode
 				newNode.Outputs.ExitCode = old.Outputs.ExitCode
 			}
+			// Add outputs to global scope here to ensure that they are reflected in archive.
+			woc.addOutputsToGlobalScope(newNode.Outputs)
```
juliev0, line 90: I see. So, basically you moved the propagation to `WorkflowStatus.Outputs` into `taskResultReconciliation()` so you don't need to have a delay anymore, right? I think that seems good.

We should just make sure that there's no logic between here and `podReconciliation()` that assumes that it hasn't been set or something (probably not but would you mind verifying that?).

cmd/argoexec/commands/wait.go
```diff
@@ -35,6 +35,9 @@ func waitContainer(ctx context.Context) error {
 	defer stats.LogStats()
 	stats.StartStatsTicker(5 * time.Minute)
 
+	// Ensures the LabelKeyReportOutputsCompleted is set to false.
+	wfExecutor.InitializeOutput(bgCtx)
```
juliev0, line 39: is this comment complete? It looks like we're creating a new TaskResult, right? (and that TaskResult has the label?)

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1944,29 +1944,31 @@ type WorkflowStatus struct {
...
-func (ws *WorkflowStatus) GetTaskResultsCompleted() map[string]bool {
-	return ws.TaskResultsCompleted
+
+func (ws *WorkflowStatus) TaskResultsComplete() bool {
+	for _, value := range ws.TaskResultsCompletionStatus {
+		if !value {
+			return false
+		}
+	}
+	return true
```
juliev0, line 1971: I hate to be nit-picky but I wonder if we could instead make this method `TaskResultsInProgress()` and then we can test for `!TaskResultsInProgress()` just because `TaskResultsComplete()` implies that all TaskResults are completed, when in fact some may not have been accounted for yet.

workflow/controller/operator.go
```diff
@@ -800,17 +800,9 @@ func (woc *wfOperationCtx) persistUpdates(ctx context.Context) {
...
-	}
-	for _, completed := range taskResultsCompleted {
-		if !completed {
-			return false
-		}
-	}
-	return true
+func (woc *wfOperationCtx) checkReconciliationComplete() bool {
+	woc.log.Debugf("Task results completion status: %v", woc.wf.Status.TaskResultsCompletionStatus)
+	return woc.wf.Status.Phase.Completed() && woc.wf.Status.TaskResultsComplete()
```
juliev0, line 805: thanks for adding this!

workflow/controller/operator.go
```diff
@@ -1374,7 +1366,6 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 	if x, ok := pod.Annotations[common.AnnotationKeyReportOutputsCompleted]; ok {
 		woc.log.Warn("workflow uses legacy/insecure pod patch, see https://argoproj.github.io/argo-workflows/workflow-rbac/")
 		resultName := pod.GetName()
-		woc.wf.Status.InitializeTaskResultIncomplete(resultName)
 		if x == "true" {
 			woc.wf.Status.MarkTaskResultComplete(resultName)
 		}
```
juliev0, line 1371: do we need to `MarkTestResultIncomplete()` in the `else` case?

cmd/argoexec/commands/wait.go
Garett-MacGowan: I will add to the comment to indicate a new task result is being created.

pkg/apis/workflow/v1alpha1/workflow_types.go
Garett-MacGowan: Will do. Makes sense.

workflow/controller/taskresult.go
Garett-MacGowan: A nice side effect is that it removes the delay, yes, but It's not just for removing the delay.

It fixes the condition where a node status has been set to its final state, but the task result wasn't reconciled on the last pass through the operator, so the outputs aren't reflected in the global scope (and therefore don't show up in the API). This can happen, for example, when the workflow is stopped. I reference the rationale behind this [here](https://github.com/argoproj/argo-workflows/pull/12402#issuecomment-1871710821). Notably, `podReconciliation()` checks if there has been a state change for a node before calling `woc.addOutputsToGlobalScope()`. If the node is marked as `wfv1.NodeFailed`, for example, but reconciliation hasn't completed for the node's outputs yet, then on the next pass through `podReconciliation()`, `newState` will be nil, and `woc.addOutputsToGlobalScope(newState.Outputs)` won't run.

```
	if newState := woc.assessNodeStatus(pod, node); newState != nil {
		woc.addOutputsToGlobalScope(newState.Outputs)
```

**Example flow:**

Stop workflow -> running node marked failed (outputs haven't been written by wait container yet) -> newState (wfv1.NodeFailed) returned by `assessNodeStatus(pod, node)` -> `addOutputsToGlobalScope()` called inside `podReconciliation`, but outputs still not reconciled, so they are not included in global scope -> operator continues looping until `woc.checkReconciliationComplete() && !woc.hasDaemonNodes()` resolves true, and `common.LabelKeyCompleted` is set. -> wait container finishes writing outputs -> `woc.taskResultReconciliation()` deals with them and makes its own call to `woc.addOutputsToGlobalScope(newNode.Outputs)` (remember, this wouldn't get called in `podReconciliation()` because there is no new node state) -> Outputs reconciled, therefore allowing `common.LabelKeyCompleted` to be set to true.

The only other references to Status.Outputs are in `addOutputsToLocalScope` inside `steps.go` and `dag.go` and in the `AddParamToGlobalScope` util, so there is no dependency on Status.Outputs being nil, as far as I can see.

Obviously it's also referenced in `get.go` for the API.

workflow/controller/operator.go
Garett-MacGowan: Yes, good catch.

Garett-MacGowan:
@juliev0 I've addressed all your comments. Let me know if there's anything else.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1944,29 +1944,31 @@ type WorkflowStatus struct {
 	// ArtifactGCStatus maintains the status of Artifact Garbage Collection
 	ArtifactGCStatus *ArtGCStatus `json:"artifactGCStatus,omitempty" protobuf:"bytes,19,opt,name=artifactGCStatus"`
 
-	// Have task results been completed? (mapped by Pod name) used to prevent premature garbage collection of artifacts.
-	TaskResultsCompleted map[string]bool `json:"taskResultsCompleted,omitempty" protobuf:"bytes,20,opt,name=taskResultsCompleted"`
+	// TaskResultsCompletionStatus tracks task result completion status' (mapped by pod name). Used to prevent premature archiving and garbage collection.
```
juliev0, line 1947: there's an extraneous apostrophe

workflow/controller/taskresult.go
juliev0: thanks for checking

pkg/apis/workflow/v1alpha1/workflow_types.go
Garett-MacGowan: Oops, I reached for the possessive here for some reason. Will fix.

Garett-MacGowan: @juliev0 Done.

Garett-MacGowan:
@juliev0 I feel like the tests fail 1/3 times because the argo exec image download keeps timing out. It's super annoying to have to keep submitting empty commits to re run the CI. Is there anything in the works to address this?

juliev0 (approved):

juliev0:
> @juliev0 I feel like the tests fail 1/3 times because the argo exec image download keeps timing out. It's super annoying to have to keep submitting empty commits to re run the CI. Is there anything in the works to address this?

Sorry, I know the feeling. There was a discussion of potentially implementing automatic retries for failures in the CI but somebody decided against it since it would hide actual failures and they preferred to address failures. As for the argoexec image download timing out, I'm not sure why that's happening so frequently. Is it just today or have you seen it in the past too?

juliev0:
@Garett-MacGowan Thanks for keeping an eye on the new incoming bugs. Now that I've merged this, it would be great if you could continue to keep an eye on anything that could be caused by this, just in case, such as related to output parameters/artifacts, or workflows not being cleaned up, etc.

Garett-MacGowan:
> @Garett-MacGowan Thanks for keeping an eye on the new incoming bugs. Now that I've merged this, it would be great if you could continue to keep an eye on anything that could be caused by this, just in case, such as related to output parameters/artifacts, or workflows not being cleaned up, etc.

Thanks for merging! Will do.

Garett-MacGowan:
> > @juliev0 I feel like the tests fail 1/3 times because the argo exec image download keeps timing out. It's super annoying to have to keep submitting empty commits to re run the CI. Is there anything in the works to address this?
> 
> Sorry, I know the feeling. There was a discussion of potentially implementing automatic retries for failures in the CI but somebody decided against it since it would hide actual failures and they preferred to address failures. As for the argoexec image download timing out, I'm not sure why that's happening so frequently. Is it just today or have you seen it in the past too?

This one has been happening for at least the last few days.

I can understand not wanting to automatically retry test failures, but setup failures (like image downloads) fall into a different category. Perhaps we can increase the timeout for that? Maybe it's happening when there is a particularly high load on the image server?

juliev0:
> > > @juliev0 I feel like the tests fail 1/3 times because the argo exec image download keeps timing out. It's super annoying to have to keep submitting empty commits to re run the CI. Is there anything in the works to address this?
> > 
> > 
> > Sorry, I know the feeling. There was a discussion of potentially implementing automatic retries for failures in the CI but somebody decided against it since it would hide actual failures and they preferred to address failures. As for the argoexec image download timing out, I'm not sure why that's happening so frequently. Is it just today or have you seen it in the past too?
> 
> This one has been happening for at least the last few days.
> 
> I can understand not wanting to automatically retry test failures, but setup failures (like image downloads) fall into a different category. Perhaps we can increase the timeout for that? Maybe it's happening when there is a particularly high load on the image server?

Very true. Definitely feel free to create a PR for that if you're motivated! (I for one am only spending a limited amount of time working on this repo these days)

zhangtbj:
Hi all,

We face a regression problem after upgrading to release `v3.5.3`. It should be related to this PR.

`We have a global artifact which is used and exported in two different steps WITH different values:`
- e2e-test-inject-runner-context
- e2e-test-output-testupdate

In the previous release, we can get the current global artifact from the last step (e2e-test-output-testupdate).
But in the new release v3.5.3 , the global artifact will be overwritten. so the artifact will be picked up randomly from these two steps which make our process failed.

I found there is a NEW change here:
https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/taskresult.go#L89-L90
```
  // Add outputs to global scope here to ensure that they are reflected in archive.
  woc.addOutputsToGlobalScope(newNode.Outputs)
```

Which will overwrite the global artifact after every task and should be the reason to make our process failed. We can find the overwriting log here:
```
time="2024-01-19T02:03:53.554Z" level=debug msg="task result name:\ne2e-test" namespace=action-dev2 workflow=e2e-test
time="2024-01-19T02:03:53.554Z" level=debug msg="Marking task result complete e2e-test" namespace=action-dev2 workflow=e2e-test
time="2024-01-19T02:03:53.554Z" level=info msg="overwriting workflow.outputs.artifacts.testUpdate: {testUpdate  <nil>  {<nil> &S3Artifact{S3Bucket:S3Bucket{Endpoint:,Bucket:,Region:,Insecure:nil,AccessKeySecret:nil,SecretKeySecret:nil,RoleARN:,UseSDKCreds:false,CreateBucketIfNotPresent:nil,EncryptionOptions:nil,CASecret:nil,},Key:e2e-test/e2e-test-inject-input-type-instances/testUpdate.tgz,} nil nil nil nil nil nil nil nil}  nil false  false  nil false}" namespace=action-dev2 workflow=e2e-test-1-4065749210214828732
```
I don't know if I found the current code and don't know how to solve that.

Can you please help check if my investigation is correct?

I also posted a comment in the argo-workflow channel:
https://cloud-native.slack.com/archives/C01QW9QSSSK/p1705635020012259

Thanks!

zhangtbj:
Maybe also cc @Garett-MacGowan for help. Thanks!

Garett-MacGowan:
@zhangtbj Do you have a workflow example to reproduce? I can see if I can fix it for next release.

Garett-MacGowan:
@zhangtbj 

> the global artifact will be overwritten. so the artifact will be picked up randomly from these two steps which make our process failed.

This makes it sound like you are running these steps in parallel. If that's the case, then I think you're getting expected behavior. Either way, send a long a workflow that reproduces the issue and I will have a better idea of what you're trying to do.

zhangtbj:
Hi @Garett-MacGowan ,

Thanks for your reply. We are not running these steps in parallel, but it is a not sample workflow. Please check:

```
kind: Workflow
metadata:
  generation: 21
  generateName: test-artifact-12402-
  namespace: action-dev2
spec:
  activeDeadlineSeconds: 28800
  arguments: {}
  entrypoint: test-root
  serviceAccountName: supermaestro
  templates:
  - inputs: {}
    metadata: {}
    name: test-root
    outputs: {}
    steps:
    - - arguments: {}
        name: inject-input-artifact
        template: inject-input-artifact
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testInput}}'
            name: testInput
          - from: '{{workflow.outputs.artifacts.testUpdate}}'
            name: testUpdate
        name: start-entrypoint
        template: main
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testUpload}}'
            name: testUpload
        name: upload-output-type-instances-step
        template: upload-output-type-instances
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testUpdate}}'
            name: testUpdate
        name: upload-update-type-instances-step
        template: upload-update-type-instances

  - inputs: {}
    metadata: {}
    name: main
    outputs: {}
    steps:
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testInput}}'
            name: input
        name: cp
        template: cp
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testUpdate}}'
            name: input-parameters
        name: render-update-ti
        template: main-render-update-ti-template
    - - arguments:
          artifacts:
          - from: '{{steps.cp.outputs.artifacts.upload}}'
            name: testUpload
        name: output-testupload
        template: output-testupload
    - - arguments:
          artifacts:
          - from: '{{steps.render-update-ti.outputs.artifacts.render}}'
            name: testUpdate
        name: output-testupdate
        template: output-testupdate

  - container:
      image: alpine:3.7
      command: [sh, -c]
      args: ["sleep 1; echo -n test input > /testInput.yaml; echo -n test update > /testUpdate.yaml"]
    name: inject-input-artifact
    outputs:
      artifacts:
      - globalName: testInput
        name: testInput
        path: /testInput.yaml
      - globalName: testUpdate
        name: testUpdate
        path: /testUpdate.yaml

  - container:
      command: [sh, -c]
      args: ["sleep 1; cp /input /upload"]
      image: alpine:3.7
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: input
        path: /input
    metadata: {}
    name: cp
    outputs:
      artifacts:
      - name: upload
        path: /upload

  - container:
      image: alpine:3.7
      command: [sh, -c]
      args: ["sleep 1; echo -n change update!!! > /render.yml"]
    metadata: {}
    name: main-render-update-ti-template
    outputs:
      artifacts:
      - name: render
        path: /render.yml

  - container:
      args:
      - sleep 1
      command:
      - sh
      - -c
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpload
        path: /typeinstance
    metadata: {}
    name: output-testupload
    outputs:
      artifacts:
      - globalName: testUpload
        name: testUpload
        path: /typeinstance

  - container:
      args:
      - sleep 1
      command:
      - sh
      - -c
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpdate
        path: /typeinstance
    metadata: {}
    name: output-testupdate
    outputs:
      artifacts:
      - globalName: testUpdate
        name: testUpdate
        path: /typeinstance



  - container:
      args:
      - sleep 1
      command: [sh, -c]
      args: ["sleep 1; cat /upload/typeInstances/testUpload; cat /upload/typeInstances/testUpload > /upload/result.yaml"]
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpload
        path: /upload/typeInstances/testUpload
    metadata: {}
    name: upload-output-type-instances
    outputs:
      artifacts:
      - globalName: uploadresult
        name: uploadresult
        path: /upload/result.yaml

  - container:
      args:
      - sleep 1
      command: [sh, -c]
      args: ["sleep 1; cat /upload/typeInstances/testUpdate"]
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpdate
        path: /upload/typeInstances/testUpdate
    metadata: {}
    name: upload-update-type-instances
    outputs: {}
```

If you run this workflow several times, you will find the final global output artifact will be changed to:
Screenshot 2024-01-19 at 14 59 13

`testUpdate` is refered to `test-artifact-12402-2cntl/test-artifact-12402-2cntl-inject-input-artifact-46224399/testUpdate.tgz`

But acutally, it should use the the artifact from later step, like:
`test-artifact-12402-2cntl/test-artifact-12402-2cntl-output-testupdate-4251288622/testUpdate.tgz`

Please let me know if anything is not clear. Thanks!

zhangtbj:
I am not sure if it is a special case that it is not covered well.
- `start-entrypoint` step is a template with some inner steps, and we update the `testUpdate` global output artifact in these inner step.
- The problem happens in the next step `upload-output-type-instances-step` just after `start-entrypoint`, I am not sure if it load artifact different.

Garett-MacGowan:
@zhangtbj I have a hypothesis for what is wrong. I'm working though how I can properly test it. Can you open an issue with the information above? I will submit a fix linked to that issue. Please mention this issue in the new issue. Thanks!

PR #9058, opened 2022-06-25
feat: add finalizer to workflow pod to prevent "pod deleted". Fixes #8783

rohankmr414:
Signed-off-by: Rohan Kumar <rohan@pipekit.io>

Fixes #8783

PR #12413, opened 2023-12-26
feat: Add finalizer to workflow pod to prevent 'pod deleted'. Fixes #8783 Continuing Work of #9058

### LINKED ISSUE #8783: Use finalizer to prevent `pod deleted`
Like Spark does, we can add a finalizer to workflow pods. This will stop them being deleted by eviction etc and allow to controller to reconcille the without getting a "pod deleted" error.

## Implementation

Add a finalizer here named `workflow.argoproj.io`:

https://github.com/argoproj/argo-workflows/blob/37a3eb375410dd778d44f9965d9ec4465f1011d9/workflow/controller/workflowpod.go#L159

This needs to be removed when the pod is complete:

https://github.com/argoproj/argo-workflows/blob/859ebe99f760c6fb30870993359274a92cec2fb9/workflow/controller/controller.go#L481

And also here:

https://github.com/argoproj/argo-workflows/blob/859ebe99f760c6fb30870993359274a92cec2fb9/workflow/controller/controller.go#L492

sakai-ast:
Fixes #8783 Continuing Work of #9058

### Motivation
Building upon the foundational work of #9058, I am advancing the implementation to address our needs.
see: https://github.com/argoproj/argo-workflows/pull/9058#issuecomment-1869324681

### Verification

ut and e2e tests


workflow/controller/controller.go
```diff
@@ -523,32 +531,49 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	logCtx.Info("cleaning up pod")
 	err := func() error {
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
+		pod, err := pods.Get(ctx, podName, metav1.GetOptions{})
```
sakai-ast, line 534: The SharedInformer pod information is sometimes out of date(see: [191/files#r1399927221](https://github.com/rohankmr414/argo-workflows/pull/191/files#r1399927221)) && pods already have a pod information from the API, so I use it instead of the `getPod` function.

workflow/controller/controller.go
```diff
@@ -523,32 +531,49 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
+			return err
+		}
 		switch action {
 		case terminateContainers:
-			pod, err := wfc.getPod(namespace, podName)
-			if err == nil && pod != nil && pod.Status.Phase == apiv1.PodPending {
-				wfc.queuePodForCleanup(namespace, podName, deletePod)
-			} else if terminationGracePeriod, err := wfc.signalContainers(namespace, podName, syscall.SIGTERM); err != nil {
+			patch := createFinalizerPodStatusRemovalPatchIfExists(pod)
+			if err := applyPatchIfExists(ctx, pods, podName, patch); err != nil {
```
sakai-ast, line 541: Perform patching process on all case sections if finalizer remains.

workflow/controller/controller.go
```diff
@@ -523,32 +531,49 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
 				return err
-			} else if terminationGracePeriod > 0 {
-				wfc.queuePodForCleanupAfter(namespace, podName, killContainers, terminationGracePeriod)
+			}
+			if pod.Status.Phase == apiv1.PodPending {
```
sakai-ast, line 544: `if err == nil && pod != nil` check is no longer needed because pod already checked the error in the previous section.

workflow/controller/controller.go
```diff
@@ -598,9 +629,24 @@ func (wfc *WorkflowController) signalContainers(namespace string, podName string
 		_ = signal.SignalContainer(wfc.restConfig, pod, c.Name, sig)
 	}
 	if pod.Spec.TerminationGracePeriodSeconds == nil {
-		return 30 * time.Second, nil
+		return 30 * time.Second
+	}
+	return time.Duration(*pod.Spec.TerminationGracePeriodSeconds) * time.Second
+}
+
+func (wfc *WorkflowController) getPod(namespace string, podName string) (*apiv1.Pod, error) {
```
sakai-ast, line 637: Moved the `getPod` function for clarity since `processNextPodCleanupItem` no longer uses the `getPod` function.

workflow/controller/controller.go
```diff
@@ -569,27 +594,33 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
+	i := slices.Index(pod.Finalizers, common.FinalizerPodStatus)
+	if i >= 0 {
+		patch = append(patch, PatchOperation{
+			Operation: "remove",
+			Path:      fmt.Sprintf("/metadata/finalizers/%d", i),
+		})
 	}
-	return pod, nil
+	return patch
 }
```
juliev0, line 607: Nit: Can we make this function more generic and pass in which finalizer it is? The side benefit is that we can make the name smaller, e.g.: `createFinalizerRemovalPatchIfExists`

workflow/controller/controller.go
```diff
@@ -569,27 +594,33 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
+		return nil
 	}
+	data, err := json.Marshal(patch)
+	if err != nil {
+		return fmt.Errorf("failed to marshal patch: %w", err)
+	}
+	log.WithField("data", string(data)).Debug("patching pod")
+	_, err = pods.Patch(ctx, podName, types.JSONPatchType, data, metav1.PatchOptions{})
+	return err
+}
```
juliev0, line 621: Nit: should this be named something like `applyPatches()` since there are more than one? (and since it's an array slice, I don't even feel we need "IfExists" in the name, since any array slice can be of 0 length, although okay if you do want that)

workflow/controller/controller.go
juliev0: So, we'll now have 3 or so more API calls per pod I guess? Would you mind summarizing why this doesn't simply cause the informer to be on a delay and for everything to just happen on a delay? (which would seem acceptable)

workflow/controller/controller.go
juliev0: actually, why does this return a slice? why not just one?

workflow/controller/controller.go
```diff
@@ -523,32 +531,49 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
+					wfc.queuePodForCleanupAfter(namespace, podName, killContainers, terminationGracePeriod)
+				}
 			}
 		case killContainers:
-			if _, err := wfc.signalContainers(namespace, podName, syscall.SIGKILL); err != nil {
+			patch := createFinalizerPodStatusRemovalPatchIfExists(pod)
+			if err := applyPatchIfExists(ctx, pods, podName, patch); err != nil {
 				return err
 			}
+			wfc.signalContainers(pod, syscall.SIGKILL)
```
juliev0, line 557: Can I understand the rationale for doing this from `terminateContainers` and `killContainers`? (I'm just guessing `deletePod` would be called later in those cases?)

workflow/controller/controller.go
```diff
@@ -973,6 +1019,22 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
...
+					pods := wfc.kubeclientset.CoreV1().Pods(wfc.GetManagedNamespace())
+					podList, err := pods.List(ctx, metav1.ListOptions{
+						LabelSelector: fmt.Sprintf("%s=%s", common.LabelKeyWorkflow, obj.(*unstructured.Unstructured).GetName()),
+					})
+					if err != nil {
+						log.WithError(err).Error("Failed to list pods")
+					}
+					for _, p := range podList.Items {
+						patch := createFinalizerPodStatusRemovalPatchIfExists(&p)
+						if err := applyPatchIfExists(ctx, pods, p.Name, patch); err != nil {
```
juliev0, line 1033: We always do these two methods together - I wonder if there should be a convenience function like `enablePodForDeletion()`?

workflow/controller/controller.go
```diff
@@ -973,6 +1019,22 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
...
+							log.WithError(err).Error("Failed to remove finalizer from pod")
+						}
```
juliev0, line 1035: Can I understand why we do it here for the case of Workflow having been deleted? I presume we will call the `deletePod` logic anyway (although haven't yet looked where).?

juliev0: or is this perhaps the case where the OwnerReference was automatically deleting the Pods remaining?

workflow/controller/controller.go
alexec: The informer can certainly be out of date. However, 3 APIs calls per pod seems high. Could you have an idempotent patch

juliev0: Sorry, I edited my original comment above, which didn't really make sense, and took a better look at the link you posted in the top of this thread @sakai-ast (apologies that I should have looked at that more closely before). Trying to understand your comment @alexec. Which Patch Operation do you consider not to be idempotent? There's a setting of a label to True, and there's a removal of a Finalizer at a given index. It seems like we could do each of them multiple times; it's just that the latter may give an error if it doesn't exist, but to me that theoretically seems like it should be okay (except that an error message could be misleading).

juliev0:
I'm surprised but I've just become aware that the Argo Server deletes Pods as part of `RetryWorkflow()` ([here](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server.go#L399)). I figured it was something that would only be the role of the Controller to do, but in any case, I guess this is an example where pod deletion is happening outside of these calls. I suppose the `labelPodCompleted` code would happen in any case, but I'm not sure what happens if you try to invoke this on an incomplete Workflow.

workflow/controller/controller.go
sakai-ast: > actually, why does this return a slice? why not just one?

This implementation simply followed the original PR's implementation.
I have addressed this in the following commit.
https://github.com/argoproj/argo-workflows/pull/12413/commits/f96c17f3365bd5b90be08240cda436a70c177472

workflow/controller/controller.go
sakai-ast: I have addressed this in the commit in the following comment.
https://github.com/argoproj/argo-workflows/pull/12413#discussion_r1446980034

workflow/controller/controller.go
sakai-ast: I have addressed this in the commit in the following comment.
https://github.com/argoproj/argo-workflows/pull/12413#discussion_r1446980034

sakai-ast:
>  I suppose the labelPodCompleted code would happen in any case

I agree. In my local tests, retried pods always have the `labelPodCompleted` and the finalizer is removed.

> but I'm not sure what happens if you try to invoke this on an incomplete Workflow.

I believe the feature flag is intended for situations like the one described above.

workflow/controller/controller.go
sakai-ast: From my research, `terminateContainers` appears to be used [here](https://github.com/argoproj/argo-workflows/blob/v3.5.2/workflow/controller/exec_control.go#L48), for example.
As far as I know, `killContainers` is only used in the `case terminateContainers` section.

`deletePod` calls `pods.Delete()`.

workflow/controller/controller.go
sakai-ast: I had assumed this code had already been reviewed on the original PR, so this implementation simply followed it.
But I'm also not sure why we're doing this here.
In fact, tests are successful without this implementation.
If you also believe that this implementation is unnecessary, I would like to remove it.

workflow/controller/controller.go
juliev0: So, I guess my question was whether it's redundant to both remove the finalizer in "terminateContainers" as well as in "deletePod"? In the line you pointed to, won't the "terminateContainers" logic ultimately cause the "deletePod" logic? (I'm also willing to entertain the idea that we do it in both places just in case, but want to understand.)

workflow/controller/controller.go
juliev0: sounds good!

juliev0:
Would you mind adding to the Description more information on how you decided on your approach and your investigation of corner cases?

workflow/controller/controller.go
sakai-ast: As you pointed out, the finalizer removal logic in `terminateContainers` is indeed redundant upon further review. Thank you for highlighting this. I have removed the redundant finalizer removal logic in the commit below.
https://github.com/argoproj/argo-workflows/pull/12413/commits/b73ed375a03210c1fec4045894b28ccb2b6e1841

workflow/controller/controller.go
sakai-ast: I have removed this in the commit below.
https://github.com/argoproj/argo-workflows/pull/12413/commits/de2bc6ce34c6cae6f40a204686fca0b2fcd604c9

workflow/controller/controller.go
juliev0: Can you check if `killContainers` is the same?

juliev0:
Here's another case: let's say that somebody is prematurely deleting their Workflow before it's done. We want to make sure the Finalizer gets removed from the Pods. I'm not sure if this calls "terminateContainers" - where your logic was before - or not, but here is a possible alternative solution: say there's a Finalizer on the Workflow, which causes the Workflow Controller to first remove the finalizer on all the Pods before removing the Finalizer on the Workflow itself. 

Generally speaking, I think our strategy for analyzing this should be to determine what all the ways are for a pod to be deleted, and try to make sure that all of those will be handled. If I search the code I see 3 instances of `Delete()` called for Pods: 
- workflow/controller/controller.go
- server/workflow/workflow_server.go
- server/workflowarchive/archived_workflow_server.go
- implicitly through deletion of the parent Workflow which has an OwnerReference to it

workflow/controller/controller.go
sakai-ast: I believe it's the same and it was removed as well.
Additionally, I also added the executor e2e test with the finalizer option set to true, just to be safe.
https://github.com/argoproj/argo-workflows/pull/12413/commits/cf831c06b7756d901f3e76644fa019c3ab5ab2f5

sakai-ast:
> Would you mind adding to the Description more information on how you decided on your approach

I utilized the approach from the original PR under the assumption that it had already undergone review. 
Other implementations in my approach are based on feedback received during the review of the original PR, 
such as the addition of a feature flag, and also include some tests for the feature flag. 
Would you like me to need to add this explanation to the Description of this PR?

> and your investigation of corner cases?

My verification for this PR primarily consists of e2e and unit tests. 
I haven't identified any additional corner cases beyond those covered in the unit and e2e tests. 

> Here's another case: let's say that somebody is prematurely deleting their Workflow before it's done. We want to make sure the Finalizer gets removed from the Pods. I'm not sure if this calls "terminateContainers" - where your logic was before

I think this is  controlled by the [applyExecutionControl().](https://github.com/argoproj/argo-workflows/blob/v3.5.2/workflow/controller/exec_control.go#L15-L76)  This function does invoke `terminateContainers`. 
In a [previous discussion](https://github.com/argoproj/argo-workflows/pull/12413#discussion_r1439104958), I removed the finalizer removal logic from both `terminateContainers` and `killContainers` based on the understanding that it was redundant. 
However, considering the scenario you described, it seems we might need to revisit this decision and potentially re-add the logic to handle such cases.

> Generally speaking, I think our strategy for analyzing this should be to determine what all the ways are for a pod to be deleted, and try to make sure that all of those will be handled. If I search the code I see 3 instances of Delete() called for Pods:

I concur that the Delete() function is called for Pods in the 3 instances you mentioned above. 
I have been reviewing the code to identify any additional logic or potential corner cases related to pod deletion. 
However, I have yet to find any further instances beyond those already discussed.

If you have concerns that the implementation in this PR may not be comprehensive enough, I would greatly appreciate the involvement of another reviewer with more in-depth knowledge of pod deletion logic to provide additional insights.

But regarding the feature flag, I understand it addresses concerns the finalizer for pods like above, which is currently set to false by default.
After this PR is merging, If any issues arise when this flag is true, my expectation is that we should be address them in a subsequent PR. This is how I envision the purpose and use of the feature flag.

Please let me know if my understanding needs correction or clarification.

juliev0:
Hey @isubasinghe - would you be interested in also reviewing this PR? I believe you may have worked on pod deletion-related PRs unless I’m mistaken?

juliev0:
Thank you @sakai-ast ! I apologize on all the iterations, including potentially needing to add back code that we decided to remove.

I agree on the feature flag really helping to lessen the concerns of scenarios missed. At the same time, it would be great to identify any possible scenarios ahead of time and test anything we're not sure of.

If we think the logic for `terminateContainers` should go back, then maybe re-add that, and then I'll be comfortable if you could just test these 2 cases and make sure the finalizer is removed and pods get deleted:

- use of "Retry Workflow" through the Argo Server (and with that, see if you're able to retry the workflow even before the Workflow has completed)
- premature deletion of a Workflow through e.g. `kubectl delete`

isubasinghe:
@juliev0 Yeah I did do some work on pod deletion I think, I can have a look now.

juliev0:
> @juliev0 Yeah I did do some work on pod deletion I think, I can have a look now.

Appreciate it. Thanks! Note that there is a feature flag for this, so slightly less risk.

sakai-ast:
@juliev0 
I have conducted tests based on your scenarios in the current PR implementation.

## Preparation

To enable `ARGO_POD_STATUS_CAPTURE_FINALIZER=true` in `RUN_MODE=kubernetes` , I added the following to the workflow-controller deployment settings(because in my local, the UI on the browser work properly when the `RUN_MODE=kubernetes`).

```diff
--- a/test/e2e/manifests/mixins/workflow-controller-deployment.yaml
+++ b/test/e2e/manifests/mixins/workflow-controller-deployment.yaml
@@ -8,4 +8,7 @@ spec:
     spec:
       containers:
         - name: workflow-controller
-          imagePullPolicy: Never
\ No newline at end of file
+          imagePullPolicy: Never
+          env:
+          - name: ARGO_POD_STATUS_CAPTURE_FINALIZER
+            value: "true"
```

built the current code into images and start
```
make workflow-controller-image argocli-image IMAGE_NAMESPACE=finalizer-test
make start PROFILE=minimal RUN_MODE=kubernetes IMAGE_NAMESPACE=finalizer-test AUTH_MODE=client STATIC_FILES=false LOG_LEVEL=debug API=true UI=true 
```

## Tests
### Retry Workflow through Argo Server

#### RETRY, SUBMIT, RESUBMIT
I observed that [RetryWorkflow()](https://github.com/argoproj/argo-workflows/blob/v3.5.4/server/workflow/workflow_server.go#L373-L416) is called from the UI's `RETRY` button. The `RETRY` button is enabled only after the workflow completion, meaning we cannot retry a workflow before it completes. 
When retried, all pods already have `workflows.argoproj.io/completed: "true"` and the finalizer is removed. Here are server logs when the `RETRY` button is pressed.
```
time="2024-01-16T11:25:27.063Z" level=info msg="Deleting pod" podDeleted=fail-4wdjs
time="2024-01-16T11:25:27.071Z" level=info msg="Workflow to be dehydrated" Workflow Size=995
time="2024-01-16T11:25:27.077Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=RetryWorkflow grpc.service=workflow.WorkflowService grpc.start_time="2024-01-16T11:25:27Z" grpc.time_ms=20.98 span.kind=server system=grpc
time="2024-01-16T11:25:27.077Z" level=info duration=22.54035ms method=PUT path=/api/v1/workflows/argo/fail-4wdjs/retry size=1799 status=0
```
RETRY
Screenshot 2024-01-16 at 20 29 19
during SUBMIT and RESUBMIT (RETRY button does not appear)
Screenshot 2024-01-16 at 20 42 36

#### TERMINATE
Terminating a running workflow using the `TERMINATE` button worked as expected, with related pods marked completed and finalizers removed.

### Premature Deletion of Workflow via kubectl delete
Deleting a running workflow using `kubectl delete` properly removed the workflow and its pods, even when the pods had finalizers, and the workflow status changed to failed. This appears to be the correct behavior.

## Problem
### On UI
Deleting a running workflow via `DELETE` button removes the workflow, but related pods get stuck in the Terminating state due to remaining finalizers.

```
time="2024-01-16T11:57:18.461Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=DeleteWorkflow grpc.service=workflow.WorkflowService grpc.start_time="2024-01-16T11:57:18Z" grpc.time_ms=18.524 span.kind=server system=grpc
time="2024-01-16T11:57:18.461Z" level=info duration=19.956665ms method=DELETE path=/api/v1/workflows/argo/fail-mp2k4 size=2 status=0
```
Screenshot 2024-01-17 at 10 58 17

This behavior is due to [DeleteWorkflow()](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server.go#L351-L372) not accounting for pods with finalizers. It attempts to delete workflow before the controller manages related pods. I believe an additional logic is needed to address this, like controller reconciliation of pod termination status and finalizer removal when eligible.

### Argo CLI
Using `argo delete` for a running workflow results in the same issue as using the `DELETE` button, with related pods remaining in the Terminating state.

## Conclusion
The tests suggests that merely adding logic to remove finalizers in `terminateContainers` might not be sufficient. Implementing a routine to periodically identify and remove finalizers from pods stuck in the Terminating status, as proposed with the `controller reconciliation of pod termination status and finalizer removal when eligible` , would more reliably prevent situations where pods that should be deleted are not.

Please review this and let me know if additional implementation is needed.

isubasinghe (changes requested):
The code itself looks fine, what I am more concerned about are edge cases in the actual network layer, I already think the semantics differ from what happened previously, would like to know what your opinions are @sakai-ast and @juliev0 ?

juliev0:
Thank you so much for all of the testing @sakai-ast . Do you remember the functionality that was in this PR that we decided to remove in the `DeleteFunc` from controller.go? I wonder if the issue you're seeing would be addressed if that were added back?
And then if we did that, maybe we don't need the `terminateContainers` logic? Does it make sense?

sakai-ast:
I have re-added the finalizer removal logic in the `DeleteFunc`
https://github.com/argoproj/argo-workflows/pull/12413/commits/b1c54ac8f28e855573690330a55610e4dcc49002

and the following issue has resolved.

> Deleting a running workflow via DELETE button removes the workflow, but related pods get stuck in the Terminating state due to remaining finalizers.

> Using argo delete for a running workflow results in the same issue as using the DELETE button, with related pods remaining in the Terminating state.

I'm not sure how it works, but adding the following temporary log settings to the DeleteFunc for debugging purposes and it logs showed the function working properly.
```diff
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -1050,6 +1050,7 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
                                                if err := enablePodForDeletion(ctx, pods, &p); err != nil {
                                                        log.WithError(err).Error("Failed to enable pod for deletion")
                                                }
+                                               log.Info("pod is enabled for pod deletion in DeleteFunc()")
                                        }

                                        key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
```
controller logs(`argo delete` executed)
```
time="2024-01-18T05:26:14.749Z" level=info msg="TaskSet Reconciliation" namespace=argo workflow=sleep-grln9
time="2024-01-18T05:26:14.749Z" level=info msg=reconcileAgentPod namespace=argo workflow=sleep-grln9
time="2024-01-18T05:26:14.749Z" level=info msg="Workflow to be dehydrated" Workflow Size=1155
time="2024-01-18T05:26:14.762Z" level=info msg="Workflow update successful" namespace=argo phase=Running resourceVersion=350104 workflow=sleep-grln9
time="2024-01-18T05:26:16.950Z" level=info msg="pod is enabled for pod deletion in DeleteFunc()"
```
How did you notice this and do you know how to call the deleteFunc?

> And then if we did that, maybe we don't need the terminateContainers logic? Does it make sense?

Yes it makes sense.

workflow/controller/controller.go
juliev0: Thinking about the API calls again. If we decide we still need them, then maybe we only need them for the calls to `labelPodCompleted` and `deletePod`?

sakai-ast: I think so.

workflow/controller/controller.go
```diff
@@ -523,32 +531,37 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	logCtx.Info("cleaning up pod")
 	err := func() error {
 		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
+		pod, err := pods.Get(ctx, podName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
isubasinghe: Yeah I think the solution here is to check if it is a transient error and then requeue.

workflow/controller/controller.go
sakai-ast: I have restored the `terminateContainers`, `killContainers`, and `signalContainers()` logic, and modified it so that the logic for obtaining `pods` and `pod` information from the API is only called by `labelPodCompleted` and `deletePod`. 
This adjustment should lead to a more appropriate number of API calls.
https://github.com/argoproj/argo-workflows/pull/12413/commits/96358f98ec9736f417499c50dd1a3c54b519e8bc

workflow/controller/controller.go
```diff
@@ -991,6 +1053,21 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
 				DeleteFunc: func(obj interface{}) {
 					// IndexerInformer uses a delta queue, therefore for deletes we have to use this
 					// key function.
+
+					// Remove finalizers if they exist before deletion
```
juliev0, line 1057:
```suggestion
					// Remove finalizers from Pods if they exist before deletion
```

workflow/controller/controller.go
```diff
@@ -569,7 +581,15 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
...
 }
 
-func (wfc *WorkflowController) getPod(namespace string, podName string) (*apiv1.Pod, error) {
+func (wfc *WorkflowController) getPodFromAPI(ctx context.Context, namespace string, podName string) (*apiv1.Pod, error) {
+	pod, err := wfc.kubeclientset.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return pod, nil
+}
```
juliev0, line 590: Did we address the issue from @isubasinghe [here](https://github.com/argoproj/argo-workflows/pull/12413#discussion_r1455463263)? What happens when there is a transient error such as one due to rate limiting? If we return an error, do we get a chance to try again later or no? 
@isubasinghe does the original code truly address this type of transient error anyway? If I look at the `labelPodCompleted` line in the original code, if it fails to run Patch(), it returns an error.

juliev0: ahh, okay, thanks. Seems good to me. Let me know if you see any issue @isubasinghe

.github/workflows/ci-build.yaml
```diff
@@ -201,12 +210,24 @@ jobs:
...
             install_k3s_version: v1.25.11-k3s1
             profile: minimal
+            pod-status-finalizer: true
           - test: test-functional
             install_k3s_version: v1.25.11-k3s1
             profile: minimal
+          - test: test-functional
+            install_k3s_version: v1.25.11-k3s1
+            profile: minimal
+            pod-status-finalizer: true
```
juliev0, line 230: @agilgur5 wanted to bring to your attention that this PR will double our e2e tests I believe. This seems like a good thing for being able to test with the environment variable feature flag both on and off, but thought I'd bring it up to you in case you have any concerns?

workflow/controller/controller.go
```diff
@@ -584,8 +604,50 @@ func (wfc *WorkflowController) getPod(namespace string, podName string) (*apiv1.
...
+			Operation: "remove",
+			Path:      fmt.Sprintf("/metadata/finalizers/%d", i),
+		}
+	}
+	return nil
+}
+
+func applyPatches(ctx context.Context, pods typedv1.PodInterface, podName string, patches []PatchOperation) error {
+	if len(patches) == 0 {
+		log.Debug("not patching pod")
```
juliev0, line 637: will this log line include information about which pod it refers to?

juliev0: same comment for line below

sakai-ast: This format follows the original PR. Do I need to make any changes?

juliev0: can you add the Pod name to it?

.github/workflows/ci-build.yaml
juliev0: I wonder if we may want to increase [this](https://github.com/argoproj/argo-workflows/blob/main/.github/workflows/ci-build.yaml#L173) timeout-minutes field at least. Maybe we should double it?

workflow/controller/controller.go
sakai-ast: I've added in the bellow commits.
https://github.com/argoproj/argo-workflows/pull/12413/commits/7d2640d03539df1952e74f3e8afe502980276d66

This format followed
https://github.com/argoproj/argo-workflows/blob/a15587755e250fbbe3b9538b1c15ae581a108f0c/workflow/controller/exec_control.go#L45

When I check the logs, it looks good for me.
```
controller: time="2024-01-26T08:13:32.358Z" level=debug msg="not patching pod" podName=test-pod-cleanup-on-pod-completion-w6f9d
```

```
controller: time="2024-01-26T08:11:24.619Z" level=debug msg="patching pod" data="[{\"op\":\"remove\",\"path\":\"/metadata/finalizers/0\"}]" podName=test-pod-cleanup-on-pod-completion-label-selected-2dxbn
```

.github/workflows/ci-build.yaml
agilgur5: > I wonder if we may want to increase [this](https://github.com/argoproj/argo-workflows/blob/main/.github/workflows/ci-build.yaml#L173) timeout-minutes field at least. Maybe we should double it?

The `timeout-minutes` actually applies to _each_ job in the matrix -- each of them has 30min. So unless this feature makes those take substantially longer, I wouldn't increase it.
When I was working on auto-retrying CI (I still have a branch with that WIP), I actually added `timeout-minutes` to each individual job in the matrix as some _shouldn't_ use the full 30 min (i.e. reducing timeouts rather than increasing them).

agilgur5: > @agilgur5 wanted to bring to your attention that this PR will increase our e2e tests. This seems like a good thing for being able to test with the environment variable feature flag both on and off, but thought I'd bring it up to you in case you have any concerns?

Yea I am a bit concerned by that. In particular:
1. is it really necessary to run _all_ of the E2E test suites with this flag? or does this feature only apply to some of them? I know the E2E test suite organization has been a bit of a black box without Alex, so that may not be easy to answer 😅 
1. we need to draw a boundary as to which feature flags are tested and which ones aren't. generically speaking, a common approach is to run tests with _all_ flags off (or at their defaults, rather) and with all flags on. selective testing is very complicated to manage otherwise and can exponentially increase the test time to an unreasonable amount.
    1. do we _always_ want to run E2E tests on this feature in particular? since this doesn't just run them for this PR, it runs them for all future commits / PRs.

juliev0: Thanks for the input @agilgur5. I know @sakai-ast was feeling this feature was risky (hence the feature flag) and that was the motivation - that this would catch issues of pods not shutting down. 

But now that you mention it, I wonder if it can give us a bunch of confidence simply seeing these pass here in this PR that maybe we don't need to run for every e2e test.? Maybe we could just add it to `test-functional` or something? What do you think @sakai-ast ?

agilgur5: We could also turn it on by default for the E2E tests instead of running it a second time.

workflow/controller/controller.go
juliev0: great, thanks, yeah, I wasn't sure what was already in the `log` - thanks for checking!

workflow/controller/controller.go
juliev0: Okay, I suppose the main question is whether the [isTransient](https://github.com/argoproj/argo-workflows/blob/main/util/errors/errors.go#L31) method includes rate limiting. Maybe it doesn't. But perhaps that was a problem in the old code anyway.

.github/workflows/ci-build.yaml
sakai-ast: Thank you for the discussion.
Turning it on by default for the E2E tests is the best safe direction if it can be done.
Because I believe Argo Workflows implementations after this PR is merged will need to be based on this feature, and the feature flag will be removed in the future.
It should be fixed as follows.

```diff
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -180,19 +180,10 @@ jobs:
         include:
           - test: test-executor
             profile: minimal
-          - test: test-executor
-            profile: minimal
-            pod-status-finalizer: true
-          - test: test-corefunctional
-            profile: minimal
           - test: test-corefunctional
             profile: minimal
-            pod-status-finalizer: true
           - test: test-functional
             profile: minimal
-          - test: test-functional
-            profile: minimal
-            pod-status-finalizer: true
           - test: test-api
             profile: mysql
           - test: test-cli
@@ -210,24 +201,12 @@ jobs:
           - test: test-executor
             install_k3s_version: v1.25.11-k3s1
             profile: minimal
-          - test: test-executor
-            install_k3s_version: v1.25.11-k3s1
-            profile: minimal
-            pod-status-finalizer: true
-          - test: test-corefunctional
-            install_k3s_version: v1.25.11-k3s1
-            profile: minimal
           - test: test-corefunctional
             install_k3s_version: v1.25.11-k3s1
             profile: minimal
-            pod-status-finalizer: true
-          - test: test-functional
-            install_k3s_version: v1.25.11-k3s1
-            profile: minimal
           - test: test-functional
             install_k3s_version: v1.25.11-k3s1
             profile: minimal
-            pod-status-finalizer: true
     steps:
       - name: Install socat (needed by Kubernetes v1.25)
         run: sudo apt-get -y install socat
@@ -291,7 +270,7 @@ jobs:
             LOG_LEVEL=info \
             API=${{matrix.test == 'test-api' || matrix.test == 'test-cli' || matrix.test == 'test-java-sdk' || matrix.test == 'test-python-sdk'}} \
             UI=false \
-            POD_STATUS_CAPTURE_FINALIZER=${{ matrix.pod-status-finalizer || 'false' }}> /tmp/argo.log 2>&1 &
+            POD_STATUS_CAPTURE_FINALIZER=true > /tmp/argo.log 2>&1 &
```

workflow/controller/controller.go
isubasinghe: I think this: https://github.com/argoproj/argo-workflows/blob/main/util/errors/errors.go#L32 should handle rate limiting right?

isubasinghe: The new error handling addresses this but I think this must've meant that this was an issue in the old code as well. 
Do you think its a good idea to open a new issue for that and us make a note there that this PR will fix it? 

Or do you think a separate PR itself is more appropriate? 

Kinda happy to go with the first option but do admit that the second is cleaner.

workflow/controller/controller.go
```diff
@@ -522,10 +530,9 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	logCtx := log.WithFields(log.Fields{"key": key, "action": action})
 	logCtx.Info("cleaning up pod")
 	err := func() error {
-		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
 		switch action {
 		case terminateContainers:
-			pod, err := wfc.getPod(namespace, podName)
+			pod, err := wfc.getPodFromCache(namespace, podName)
```
isubasinghe, line 535: Why is it safe to use the cache here? 
What if pod.Status.Phase is out of sync with what is really going on? What would happen in that case?

isubasinghe: Hmm I guess the only thing that will happen is for the deletePod action to happen, which I think should be okay

isubasinghe (changes requested):
I've added a couple more comments, could you please address them? Just so that we are on the same page.

workflow/controller/controller.go
isubasinghe: Ah no, this is my bad sorry, I for some reason didn't realise that the old code handled this as well. I sort of missed that it was inside a lambda. 


We can resolve this discussion I think.

.github/workflows/ci-build.yaml
sakai-ast: I've added the fix above.
https://github.com/argoproj/argo-workflows/pull/12413/commits/e2b2de4ac9b034ce3485dd90cc232846ee17c9b8

juliev0 (approved):

juliev0 (commented):
Approved before waiting for e2e tests....will approve once they all do

imliuda:
if workflow-controller stopped for a while, eg, maintaining or upgrading, and then workflows deleted by somebody, if this will still work properly? during the gap, some pods completed, but the finalizer have no chance to get removed, how can we ensure that pods can be deleted.

Joibel:
You can't. The deletion will hang until either the controller is started or someone through another means deletes the finalizer words on the pod.

The whole point of this change is to do this, and prevent inconsistency of state between the workflow and pods.

If you are happy with the alternative that a workflow may have unexpectedly missing pods you may delete the finalizer manually.

imliuda:
> You can't ensure the pods will be deleted. The deletion will hang until either the controller is started or someone through another means deletes the finalizer words on the pod. The point of this change is to do this, and prevent inconsistency of state between the workflow and pods.
> 
> If you are happy with the alternative that a workflow may have unexpectedly missing pods you may delete the finalizer manually from the pods. This is scriptable.

It is unnecessary for keeping consistency between workflow and pods, once workflow-controller has finished processing pods (that is node.Fullfilled()), then the pod can be deleted safely.  Or, that may induce a memory leak of kube-apiserver due to large amount pods (kube-controller-manager will do pod gc of Failed and Succeed pods). Since finalizer is added by workflow-controller, then it should have to be responsible for removing it when the corresponding workflow deleted. Removing finalizer by a script is not appropriate.

I can make some optimizations for situation when pod's owner workflow has been deleted, and open a new PR for that.

PR #12415, opened 2023-12-26
docs: Fix path

Jellyfrog:
Correct a path in the docs


terrytangyuan (commented):
Why is this change necessary? The existing path should already work.

sarabala1979 (approved):

PR #12421, opened 2023-12-29
fix: custom columns not supporting annotations

### LINKED ISSUE #12422: Custom columns do not read from annotations like documentation states
### Pre-requisites

Add an annotation to my workflow that I have listed on my `workflow-controller-configmap.yaml` under `columns` with `key: annotation`.

Example in `workflow-controller-configmap.yaml`

```yaml
  columns: |
    - name: Custom Column
      type: annotation
      key: workflows.argoproj.io/custcom-column
```

[Example kustomize overlay](https://github.com/argoproj/argo-workflows/blob/954884d64c7b2d24a2e5fa0641a66692d182fefd/manifests/quick-start/base/overlays/workflow-controller-configmap.yaml#L46) to edit for local testing

### Actual (Current)

`unknown` always shows in custom column

Screenshot 2023-12-28 at 9 47 33 PM

### Expected

value from workflow annotation

### Version

v3.5.2 and latest

```YAML
kind: Workflow
metadata:
  generateName: custom-column-
  namespace: argo
  labels:
    example: 'true'
  annotations:
    argo-events.argoproj.io/customcolumn: custom-column
spec:
  arguments:
    parameters:
      - name: message
        value: hello argo
  entrypoint: argosay
  templates:
    - name: argosay
      inputs:
        parameters:
          - name: message
            value: '{{workflow.parameters.message}}'
      container:
        name: main
        image: argoproj/argosay:v2
        command:
          - /argosay
        args:
          - echo
          - '{{inputs.parameters.message}}'
  ttlStrategy:
    secondsAfterCompletion: 300
  podGC:
    strategy: OnPodCompletion
```

```text
nothing relevant since this is a UI issue
```

```text
Nothing relevant since this is a UI issue
```

jmeridth:
Fixes #12422

### Motivation

We'd like to add a PR column that lists the user's PR URL and using annotations since labels have character limits and don't support special characters.

Builds off #10808

### Modifications

- [x] search annotations for matching key from user's custom column type
  - docs state we can use labels and annotations [here](https://github.com/argoproj/argo-workflows/blob/954884d64c7b2d24a2e5fa0641a66692d182fefd/pkg/apis/workflow/v1alpha1/info.go#L21) and [here](https://github.com/argoproj/argo-workflows/blob/954884d64c7b2d24a2e5fa0641a66692d182fefd/docs/workflow-controller-configmap.yaml#L90)

Screenshot 2023-12-28 at 9 14 42 PM

### Verification

I changed [this file's columns section](https://github.com/argoproj/argo-workflows/blob/954884d64c7b2d24a2e5fa0641a66692d182fefd/manifests/quick-start/base/overlays/workflow-controller-configmap.yaml#L46) from:

```yaml
  columns: |
    - name: Workflow Completed
      type: label
      key: workflows.argoproj.io/completed
```

to

```yaml
  columns: |
    - name: PR
      type: annotation
      key: workflows.argoproj.io/pullrequest
```

and added the follow to a workflow's annotations:

```yaml
metadata:
  annotations:
    workflows.argoproj.io/pullrequest: https://github.com/argoproj/argo-workflows/pull/12421
```

Reference screenshot above for result

### DIFF
--- a/ui/src/app/workflows/components/workflows-row/workflows-row.tsx
+++ b/ui/src/app/workflows/components/workflows-row/workflows-row.tsx
@@ -87,7 +87,7 @@ export function WorkflowsRow(props: WorkflowsRowProps) {
                     <div className='columns small-1'>{isArchivedWorkflow(wf) ? 'true' : 'false'}</div>
                     {(props.columns || []).map(column => {
                         // best not to make any assumptions and wait until this data is filled
-                        const value = (column.type === "label" ? wf?.metadata?.labels?.[column.key] : wf?.metadata?.annotations?.[column.key]) ?? 'unknown';
+                        const value = (column.type === 'label' ? wf?.metadata?.labels?.[column.key] : wf?.metadata?.annotations?.[column.key]) ?? 'unknown';
                         return (
                             <div key={column.name} className='columns small-1'>
                                 {value}
--- a/ui/src/app/workflows/components/workflows-row/workflows-row.tsx
+++ b/ui/src/app/workflows/components/workflows-row/workflows-row.tsx
@@ -87,7 +87,7 @@ export function WorkflowsRow(props: WorkflowsRowProps) {
                     <div className='columns small-1'>{isArchivedWorkflow(wf) ? 'true' : 'false'}</div>
                     {(props.columns || []).map(column => {
                         // best not to make any assumptions and wait until this data is filled
-                        const value = wf?.metadata?.annotations?.[column.key] ?? wf?.metadata?.labels?.[column.key] ?? 'unknown';
+                        const value = (column.type === "label" ? wf?.metadata?.labels?.[column.key] : wf?.metadata?.annotations?.[column.key]) ?? 'unknown';
                         return (
                             <div key={column.name} className='columns small-1'>
                                 {value}
--- a/ui/src/app/workflows/components/workflows-row/workflows-row.tsx
+++ b/ui/src/app/workflows/components/workflows-row/workflows-row.tsx
@@ -87,7 +87,7 @@ export function WorkflowsRow(props: WorkflowsRowProps) {
                     <div className='columns small-1'>{isArchivedWorkflow(wf) ? 'true' : 'false'}</div>
                     {(props.columns || []).map(column => {
                         // best not to make any assumptions and wait until this data is filled
-                        const value = wf?.metadata?.labels?.[column.key] ?? 'unknown';
+                        const value = wf?.metadata?.annotations?.[column.key] ?? wf?.metadata?.labels?.[column.key] ?? 'unknown';
                         return (
                             <div key={column.name} className='columns small-1'>
                                 {value}


ui/src/app/workflows/components/workflows-row/workflows-row.tsx
```diff
@@ -87,7 +87,7 @@ export function WorkflowsRow(props: WorkflowsRowProps) {
                     <div className='columns small-1'>{isArchivedWorkflow(wf) ? 'true' : 'false'}</div>
                     {(props.columns || []).map(column => {
                         // best not to make any assumptions and wait until this data is filled
-                        const value = wf?.metadata?.labels?.[column.key] ?? 'unknown';
+                        const value = wf?.metadata?.annotations?.[column.key] ?? wf?.metadata?.labels?.[column.key] ?? 'unknown';
```
agilgur5, line 90: we should use `column.type` directly here
```suggestion
                        const value = (column.type === "label" ? wf?.metadata?.labels?.[column.key] : wf?.metadata?.annotations?.[column.key]) ?? 'unknown';
```

I'm not the biggest fan of large one-liners though

jmeridth: I'm not a fan of this. You feel strongly about this? Am I missing a possible issue without the type check explicitly?

agilgur5: > Am I missing a possible issue without the type check explicitly?

Yes, for example, you can capture a label even when the user/operator specified the `annotation` type

agilgur5:
```suggestion
                        const keys = column.type === "label" ? wf?.metadata?.labels : wf?.metadata?.annotations;
                        const value = keys?.[column.key] ?? 'unknown';
```

although I'd prefer we specify `label` and `annotation` explicitly too.... hmmm...

```suggestion
                        const keys =  wf?.metadata?.[`${column.type}s`];
                        const value = types?.[column.key] ?? 'unknown';
```

well that could be one-lined as dynamic af:

```suggestion
                        const value = wf?.metadata?.[`${column.type}s`]?.[column.key] ?? 'unknown';
```

jmeridth: Ah, nice catch.  Example for other readers.  We have the following in the workflow:

```yaml
labels:
  workflows.argoproj.io/custom-column: test-label
```

and our configmap is only looking for annotations.

Gonna look at

agilgur5: did you not want to use the dynamic version in my last suggestion [above](https://github.com/argoproj/argo-workflows/pull/12421#discussion_r1437976854)?

jmeridth: I'd do one of the following for clean up

(yours but rename `keys` to `metadata`)

```javascript
const metadata = column.type === "label" ? wf?.metadata?.labels : wf?.metadata?.annotations;
const value = metadata?.[column.key] ?? 'unknown';
```

or VERY EXPLICIT

```javascript
let value = 'unknown';
if (column.key === 'labels') {
    value = wf?.metadata?.labels?.[column.key] ?? 'unknown';
}
if (column.key === 'annotations') {
    value = wf?.metadata?.annotations?.[column.key] ?? 'unknown';
}
```

Not a fan of of the `${column.type}s` stuff. 😢

agilgur5 (approved):
LGTM, nice catch on the code missing an annotation check despite the docs saying it!

terrytangyuan (approved):
LGTM

PR #12438, opened 2024-01-01
fix: update MinIO Helm URL

SriramKeerthi:
Fixes #TODO

### Motivation

MinIO changed the URL from https://helm.min.io/ to https://charts.min.io/ as per https://github.com/minio/minio/tree/master/helm/minio

### Modifications

Changed installation example URL from https://helm.min.io/ to https://charts.min.io/

### Verification

Ran the command with the new URL and it works.


agilgur5 (approved):
Thanks for the update! 

Any reason this is still in draft? This change seems highly requested per #10320, so would love to get this in 🙂

agilgur5:
Replaced by #12552

PR #12362, opened 2023-12-14
chore: Update GitHub pages to redirect to readthedocs

jmeridth:
Previously was #12212

Relates to #12360

### Motivation

Change each GitHub page to utilize meta tag to redirect to new location on readthedocs

Since we don't host our previous documentation we can't do a real HTTP redirect.  Meta tag is the next best thing

https://www.w3.org/TR/WCAG20-TECHS/H76.html

### Modifications

- [x] setup redirects to [readthedocs](https://argo-workflows.readthedocs.io/en/latest/) + previus page suffix (once #12360 merges and builds)

[Python script used to update the files](https://gist.github.com/jmeridth/94cd15d2b11d444fdf91913aa59f34ab)

PR #12443, opened 2024-01-02
chore: Update GitHub pages to redirect to readthedocs

jmeridth:
Previously was #12212 and #12362

Relates to #12360

### Motivation

Change each GitHub page to utilize meta tag to redirect to new location on readthedocs

Since we don't host our previous documentation we can't do a real HTTP redirect.  Meta tag is the next best thing

https://www.w3.org/TR/WCAG20-TECHS/H76.html

### Modifications

- [x] setup redirects to [readthedocs](https://argo-workflows.readthedocs.io/en/latest/) + previus page suffix (once #12360 merges and builds)

[Python script used to update the files](https://gist.github.com/jmeridth/94cd15d2b11d444fdf91913aa59f34ab)


jmeridth:
I made the mistake of recreating the branch and force pushing #12362 hence the PR closing.  My bad.

terrytangyuan (approved):

terrytangyuan:
thanks!

agilgur5:
Having the same CSS from https://github.com/argoproj/argo-workflows/pull/12362#discussion_r1433264020 is a pretty nice experience; it looks like basically the same thing to me in the half a second before I get redirected from https://argoproj.github.io/argo-workflows/ 🙂

agilgur5:
Also @terrytangyuan can you change the "About" link on GitHub to point to the new docs as well?

Right now it's still pointing to GH Pages:
Screenshot 2024-01-03 at 9 29 37 PM

terrytangyuan:
done

CONTRIBUTING/index.html
```diff
@@ -1,4182 +1,63 @@
...
-        data-input-position="bottom"
-        data-theme="light"
-        data-lang="en"
-        data-loading="lazy"
-        crossorigin="anonymous"
-        async>
-</script>
-
-
+              <h1>Contributing - Argo Workflows - The workflow engine for Kubernetes</h1>
```
agilgur5, line 42: I didn't have a chance to review this follow-up PR, but this seems to have used an older version of the script? As this should just be "Contributing"

This seems to be the case for all pages currently live, which is how I noticed it (clicking on an old link). i.e. `heading == title` in this PR when it shouldn't

agilgur5: Looking at the script, it looks to have the erroneous code currently. The [revision](https://gist.github.com/jmeridth/94cd15d2b11d444fdf91913aa59f34ab/revisions) 2 diffs ago (GH doesn't seem to have a direct link) has this diff, which should be reverted:

```diff
-                page_h1 = line.replace(OPEN_TITLE_HTML, '').replace(CLOSE_TITLE_HTML, '').replace(f" - {BASE_TITLE}", '').strip()
+                page_h1 = line.replace(OPEN_TITLE_HTML, '').replace(CLOSE_TITLE_HTML, '').strip()
```

CONTRIBUTING/index.html
```diff
@@ -1,4182 +1,63 @@
...
-  <script async src="https://www.googletagmanager.com/gtag/js?id=G-5Z1VTPDL73"></script>
-
-
-    
-    
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <link rel="canonical" href="https://argo-workflows.readthedocs.io/en/latest/CONTRIBUTING/" />
+    <meta http-equiv="refresh" content="0;URL=https://argo-workflows.readthedocs.io/en/latest/CONTRIBUTING/" />
+    <title>Contributing - Argo Workflows - The workflow engine for Kubernetes - Argo Workflows - The workflow engine for Kubernetes</title>
```
agilgur5, line 8: some have duplicated titles as well, this has "- Argo Workflows - The workflow engine for Kubernetes" twice

jmeridth: I'll get PR up shortly.

agilgur5: I got a branch with all the changes done now, so I think we should be covered. If you want to edit the script for posterity, that could be good

agilgur5: Fixed in #12500

CONTRIBUTING/index.html
agilgur5: Fixed in #12500

PR #12446, opened 2024-01-02
fix: documentation links

### LINKED ISSUE #12444: Documentation links in main are broken
### Pre-requisites

Actual:

Current links have prefix <https://argo-workflows.readthedocs.io/en/stable/> (from #12360 merge)

Expected:

Link suffix should be <https://argo-workflows.readthedocs.io/en/latest/> since that is what `main` branch is, it is latest and it is what we are redirecting all old GitHub pages to (from #12443 merge)

### Version

latest

```YAML
Not relevant
```

```text
Not relevant
```

```text
Not relevant
```

jmeridth:
Fixes #12444 

### Motivation

Current links have prefix https://argo-workflows.readthedocs.io/en/stable/ (from #12360 merge)

Expected:

Link suffix should be https://argo-workflows.readthedocs.io/en/latest/ since that is what main branch is, it is latest and it is what we are redirecting all old GitHub pages to (from #12443 merge)

[Where we switched stable to latest](https://github.com/argoproj/argo-workflows/pull/12362#issuecomment-1860718112)

### Modifications

Change doc links to latest readthedocs instead of stable (for main)


terrytangyuan (approved):

docs/workflow-controller-configmap.yaml
```diff
@@ -1,3 +1,6 @@
+#######
+# NEW URL FOR THIS FILE: https://argo-workflows.readthedocs.io/en/latest/workflow-controller-configmap.yaml
+#######
```
agilgur5, line 3: This was supposed to be only on the `gh-pages` branch since we couldn't redirect this file. But this PR was to `main`. This shouldn't be on `main`

agilgur5: Fixed in #12654

Issue #12414, opened 2023-12-26
Backport doc build changes for versioned docs


jmeridth:
# Summary

Keep track of doc build changes that need to be backported to prior versions so they can show as versioned docs on readthedocs.io (once #12360 merges)

Setup changes that need to be backported:

- [.readthedocs.yaml](https://github.com/argoproj/argo-workflows/pull/12360/files#diff-cde814ef2f549dc093f5b8fc533b7e8f47e7b32a8081e0760e57d5c25a1139d9)
- [Makefile](https://github.com/argoproj/argo-workflows/pull/12360/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)
- [mkdocs.yml](https://github.com/argoproj/argo-workflows/pull/12360/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2)
- [hack/check-mkdocs.sh](https://github.com/argoproj/argo-workflows/pull/12360/files#diff-5c6d522ef188e70337bce37f5e67c94b0e734e06a027d621d73509b23bd7880b)
- [docs/requirements.txt](https://github.com/argoproj/argo-workflows/blob/main/docs/requirements.txt)

We'd still want the URL changes also from #12360 but instead of `/latest/` suffix it'd be the semver (e.g., `/release-[major].[minor]/` suffix)

- [x] release-3.5 branch #12462 
  - https://argo-workflows.readthedocs.io/en/release-3.5/ is live
- [ ] release-3.4 branch #12473 
  - https://argo-workflows.readthedocs.io/en/release-3.4/ not generated yet

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

jmeridth:
@agilgur5 @terrytangyuan Working on the first PR for this.  The files mentioned in the description above make sense and confirming those on my forked project in readthedocs but I also realized we need to update the 461 files that still have github pages links in them.  This brings up a few questions:

1. For backporting we need to switch to readthedocs links but do we, for the backports, want to change to https://argo-workflows.readthedocs.io/en/[version] (e.g., https://argo-workflows.readthedocs.io/en/v3.5.2) or have all backports point to latest (e.g., https://argo-workflows.readthedocs.io/en/latest)?  I assume the version in the URL and we can point the version build in the readthedocs UI.
[...]

PR #12450, opened 2024-01-03
chore: backport documentation changes to v3.5.2

jmeridth:
### Motivation

relates to #12414

### Modifications

- [x] move over files to help with readthedocs build
  - [x] .readthedocs.yaml
  - [x] Makefile
  - [x] mkdocs.yml
  - [x] hack/check-mkdocs.sh
  - [x] docs/requirements.txt
- [x] change links from github pages to readthedocs v3.5.2


terrytangyuan (approved):

terrytangyuan:
LGTM

jmeridth:
@terrytangyuan will we move the tag for v3.5.2 to this new SHA (aka HEAD of release-3.5.2 branch)?

terrytangyuan:
I don't think that's possible

terrytangyuan:
We should only create new tags instead of modifying existing ones

agilgur5:
Ah looks like Argo CD has the `release-3.5` -style branch for each minor that I think it would make sense for us to use as well. See my comment in https://github.com/argoproj/argo-workflows/issues/12414#issuecomment-1876216461 for screenshots/examples of that

README.md
```diff
@@ -20,11 +20,11 @@ Argo is a [Cloud Native Computing Foundation (CNCF)](https://cncf.io/) graduated
 
 ## Use Cases
 
-* [Machine Learning pipelines](https://argoproj.github.io/argo-workflows/use-cases/machine-learning/)
-* [Data and batch processing](https://argoproj.github.io/argo-workflows/use-cases/data-processing/)
-* [Infrastructure automation](https://argoproj.github.io/argo-workflows/use-cases/infrastructure-automation/)
-* [CI/CD](https://argoproj.github.io/argo-workflows/use-cases/ci-cd/)
-* [Other use cases](https://argoproj.github.io/argo-workflows/use-cases/other/)
+* [Machine Learning pipelines](https://argo-workflows.readthedocs.io/en/v3.5.2/use-cases/machine-learning/)
```
agilgur5, line 23: Ah I see, these changes will impact people who navigate to the `release-3.5` branch directly for the examples and docs.

The source code changes would only affect the next version of the binary though, which would be v3.5.3 at that time. Hmmm... using `release-3.5` or `3.5.x` for the version number in the docs permalink may make the most sense then

terrytangyuan:
We do have those release branches. The problem is modifying an existing tag.

terrytangyuan:
See https://github.com/argoproj/argo-workflows/branches/all?query=release

agilgur5:
> See https://github.com/argoproj/argo-workflows/branches/all?query=release

It looks like we're currently using `release-3.5` as tracking v3.5.0 specifically though (the `.0` patch as in), as opposed to the minor as a whole, since it is behind `release-3.5.2`

terrytangyuan:
Can we use a branch? Then it is as simple as syncing release-3.5 branch to be up-to-date with 3.5.2

agilgur5:
Looks like we can indeed use a branch:

Screenshot 2024-01-03 at 9 50 07 PM

jmeridth:
Yes we can but I need to issue another PR to change the URL to `/release-3.5.2` instead of `/v3.5.2`. Why can't we move the tag? I'm not understanding that. Shouldn't the tag point to the head of its release branch?

jmeridth:
Nm. I see @agilgur5's comment [here](https://github.com/argoproj/argo-workflows/issues/12414#issuecomment-1876216461). We should match ArgoCD

agilgur5:
> Why can't we move the tag? I'm not understanding that.

A tag is _supposed_ to be a stable reference, as opposed to a branch. Some artifact registries don't allow changing tags either (like NPM, for instance).
Git/Go and Docker/OCI registries actually have _mutable_ references however, which is why using a SHA or digest in addition to the version number is a general recommendation for supply chain security purposes for those registries.

But for _most_ intents, you typically want tags to be **immutable**.
This is an edge case where I'd support moving it, although as we're not making a new GH release of the binary etc, I'd rather keep things consistent where possible. And if we're just doing the minor branch then we can leave the tag as immutable.

The branch won't cover all the bases (i.e. old versions of the binary and old release tags will still point to GH Pages), but it'll cover all the most important ones, which is "good enough" IMO. Everything else will get redirected to `latest`, which matches the previous behavior anyway

agilgur5:
Follow-up PR in #12460

PR #12465, opened 2024-01-04
fix(docs): handle `fields` examples with `md_in_html`

### LINKED ISSUE #12453: RTD Docs: Fields Reference example links are broken
### Pre-requisites

In the Fields Reference on RTD at https://argo-workflows.readthedocs.io/en/latest/fields/ all of the "Examples" dropdowns are currently not rendering markdown properly and are displaying raw text. See below screenshot:

Screenshot 2024-01-03 at 11 19 35 PM

As a result the list is not displaying correctly and you can't directly click the links either.

I'm actually not yet sure why this is happening, which is why I filed an issue instead of a PR currently

Regression from #12360, although again not entirely sure how/why, potentially due to `mkdocs` dep version changes?

### Version

https://argo-workflows.readthedocs.io/en/latest/fields/

```YAML
n/a this is specific to the docs
```

```text
n/a this is specific to the docs
```

```text
n/a this is specific to the docs
```

agilgur5:
Fixes #12453 

### Motivation

- previously this used a hacky script, `hack/parse_examples.go` to make markdown work inside of the HTML `details` element
  - this hack is not run [on RTD's build system](https://github.com/argoproj/argo-workflows/blob/1968436a6e021adc4785ec287635a6136574c5ac/.readthedocs.yml), and so that meant the markdown ended up rendered as raw text

- instead of using hacky workarounds, use the `md_in_html` extension for `mkdocs`, which will render the markdown inside of the `details` elements correctly
  - this should also work on RTD seamlessly, as it's not a Go script exclusive to our `Makefile`

### Modifications

- add a `markdown` attribute to the `details` element in `hack/docgen.go` per https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown/#markdown-in-html
  - also remove new line that added too much space and is no longer necessary
    - might've been necessary with the hack as I don't remember it creating extra space before

- remove `hack/parse_examples.go` and all references to it

- plus make sure extensions are alphabetized
  - `pymdownx.details` should be before `pymdownx.superfences`

### Verification

1. Ran `make codegen` (`make docs/fields.md` more specifically) to get the changes to `docs/fields.md` that are here in this diff
1. Then ran `mkdocs build --strict` to replicate the RTD build (i.e. _not_ `make docs`)
1. Then `open site/fields/index.html` and verified the examples worked correctly as in below screenshot:

Before:
Screenshot 2024-01-03 at 11 19 35 PM

After:
Screenshot 2024-01-04 at 4 37 39 PM

Still works properly inside of GH Markdown as well -- see [rendered MD](https://github.com/argoproj/argo-workflows/blob/06796c5aeb5d4a35cd1f6509097aab1171b10f61/docs/fields.md) on this PR


jmeridth (approved):
LGTM

terrytangyuan (approved):

PR #12471, opened 2024-01-05
docs: Mention #argo-wf-sustainability in sustainability efforts

terrytangyuan:
This is a new dedicated channel for this effort.

### DIFF
--- a/community/sustainability_effort.md
+++ b/community/sustainability_effort.md
@@ -26,7 +26,7 @@ There are different expectations depending on the role:
 
 Note that the hours per week listed above is an average over time; it's fine to have weeks of no activity so long as the average meets expectations.
 
-Participants should join the [`#argo-wf-contributors`](https://cloud-native.slack.com/archives/C0510EUH90V) Slack channel.
+Participants should join the [`#argo-wf-contributors`](https://cloud-native.slack.com/archives/C0510EUH90V) Slack channel and ask to be invited to the private `#argo-wf-sustainability` channel.
 
 ### Reviewing PRs
 
--- a/community/sustainability_effort.md
+++ b/community/sustainability_effort.md
@@ -26,7 +26,7 @@ There are different expectations depending on the role:
 
 Note that the hours per week listed above is an average over time; it's fine to have weeks of no activity so long as the average meets expectations.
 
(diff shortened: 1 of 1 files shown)


juliev0:
I was thinking about this as well. If I leave it as a Private channel, how do you think they should make the request? Through the `argo-wf-contributors` slack channel? Or maybe I should make it a Public channel and have them join themselves?

terrytangyuan:
It should remain as private channel. They can submit a PR or ask us in wf contributors channel.

juliev0:
> It should remain as private channel (I think only CNCF has permission to change the visibility). They can submit a PR or ask us in wf contributors channel.

Well, they will be submitting a PR anyway to add their name to the document. So, as a matter of process we could add them when we merge that? (assuming we don't forget)

terrytangyuan:
Well currently it points to wf-contributors channel which is not correct (they don't necessarily need to join this channel). 

This could serve as a good reminder for us as well.

juliev0:
> Well currently it points to wf-contributors channel which is not correct (they don't necessarily need to join this channel).
> 
> This could serve as a good reminder for us as well.

Yeah, it's a good question as to whether they should still join the `argo-wf-contributors` channel. Maybe it's good for them to be part of that anyway?

community/sustainability_effort.md
```diff
@@ -26,7 +26,7 @@ There are different expectations depending on the role:
 
 Note that the hours per week listed above is an average over time; it's fine to have weeks of no activity so long as the average meets expectations.
 
-Participants should join the [`#argo-wf-contributors`](https://cloud-native.slack.com/archives/C0510EUH90V) Slack channel.
+Participants should join the [`#argo-wf-contributors`](https://cloud-native.slack.com/archives/C0510EUH90V) Slack channel and ask to be invited to the private `#argo-wf-sustainability` channel.
```
juliev0, line 29:
```suggestion
Participants should join the [`#argo-wf-contributors`](https://cloud-native.slack.com/archives/C0510EUH90V) Slack channel. They should also be invited to the private `#argo-wf-sustainability` channel.
```

juliev0: what do you think of something like that?

juliev0: I feel like if I saw this and somebody forgot to invite me, I would ask ?

juliev0 (approved):

PR #11410, opened 2023-07-20
build: upgrade to use nodev20

chenrui333:
👋 node 16 is gonna be  EOL on 11 Sep 2023, right now, building with node v20 would have [`digital envelope routines::unsupported` issue](https://github.com/argoproj/argo-workflows/issues/11404), thus updating the build to use v20.

closes #11404
relates to https://github.com/Homebrew/homebrew-core/pull/137118

cc @templarfelix @JPZ13

PR #12480, opened 2024-01-07
build(ui): remove no longer needed `--openssl-legacy-provider`

agilgur5:
### Motivation

- this was needed for legacy Webpack v4 support, but as of #11628, we are on Webpack v5 and so this is no longer necessary
  - and this is more secure too, as the legacy option was insecure
  
Remembered that this was no longer needed after https://github.com/argoproj/argo-ui/pull/432#issuecomment-1879816619, I must've forgot I could remove this earlier (or that the flags were even there)

### Modifications

- basically undoes the parts of #11410 that added `--openssl-legacy-provider`
  - removes all `--openssl-legacy-provider` flags

### Verification

1. `yarn build` still works fine
1. `yarn start` still works fine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ COPY api api
 
 RUN --mount=type=cache,target=/root/.yarn \
   YARN_CACHE_FOLDER=/root/.yarn JOBS=max \
-  NODE_OPTIONS="--openssl-legacy-provider --max-old-space-size=2048" JOBS=max yarn --cwd ui build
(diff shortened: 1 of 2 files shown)


alexec:
I’m trying to remember why this was added, but I think if the build passes, this is good to go.

agilgur5:
I wrote why it was added in my PR description

jmeridth (approved):

terrytangyuan (approved):

Issue #10426, opened 2023-01-30
Memoization doesn't work in templates that are dag or use retries


gmeshkov:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

When I add memoization to a step with `retryStrategy`, the memoization doesn't work.
Checked with `retryStrategy: Always` and the default `OnFailure` value.

Also, the memoization doesn't work for a step which is a `dag` containing other steps.

From the logs of the controller I can conclude that the entry in ConfigMap is not created for some reason after the retry/dag steps finish.

### Version

3.4.4

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
# This test shows that memoization doesn't work for templates with retryStrategy and dag templates
# `regular-1` and `regular-2` tasks show that memoization works for regular templates
# `with-retries-1` and `with-retries-2` show that if you add the retryStrategy to the template, memoization no longer works
# `with-dag-1` and `with-dag-2` show that memoization doesn't work if a memoized template contains multiple steps (is a dag)
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: memoized-workflow-test
spec:
  entrypoint: main
#  podGC:
#    strategy: OnPodCompletion
  templates:
  - name: main
    dag:
      tasks:
      - name: regular-1
        template: run
        arguments:
          parameters:
          - name: id
            value: 1
          - name: cache-key
            value: '{{workflow.name}}'
      - name: regular-2
        template: run
        depends: regular-1.Succeeded
        arguments:
          parameters:
          - name: id
            value: 2
          - name: cache-key
            value: '{{workflow.name}}'
      - name: with-retries-1
        template: run-with-retries
        arguments:
          parameters:
          - name: id
            value: 3
          - name: cache-key
            value: '{{workflow.name}}'
      - name: with-retries-2
        template: run-with-retries
        depends: with-retries-1.Succeeded
        arguments:
          parameters:
          - name: id
            value: 4
          - name: cache-key
            value: '{{workflow.name}}'
      - name: with-dag-1
        template: run-with-dag
        arguments:
          parameters:
          - name: id
            value: 5
          - name: cache-key
[...]

PR #12491, opened 2024-01-09
fix: cache configmap don't create with workflow has retrystrategy. Fi…

### LINKED ISSUE #12490: Memoized didn't created when there is retryStrategy
### Pre-requisites

I want a configmap named cache-config but not found.

I think the root cause is when the workflow have retrystrategy, memoizationStatus will only in ParentNode instead of children node.

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: memoized-
spec:
  entrypoint: whalesay
  templates:
  - name: whalesay
    memoize:
      key: "cache-key"
      cache:
        configMap:
          name: cache-config
    container:
      image: docker/whalesay:latest
      command: [sh, -c]
      args: ["cowsay world > /tmp/hello_world.txt"]
    outputs:
      parameters:
      - name: hello
        valueFrom:
          path: /tmp/hello_world.txt
---
workflowDefaults: |
    spec:
      # must complete in 8h (28,800 seconds)
      activeDeadlineSeconds: 28800
      # keep workflows for 1d (86,400 seconds)
      ttlStrategy:
        secondsAfterCompletion: 86400
        secondsAfterFailure: 86400
        secondsAfterSuccess: 86400
      retryStrategy:
        limit: 4
```

```text
```

```text
```

shuangkun:
I think MemoizationStatus should like inputs, if the executeTmpl has Memoize, the node should have MemoizationStatus.

I test it and if i Set the MemoizationStatus,  the ut pass. If did't，the ut will failed.

Fixes #12490


Joibel:
Would also fix the second part of #10426

shuangkun:
> I'm going to have a build and test of this tomorrow, but the code looks like it does the right thing. Thanks @shuangkun.

Thanks for your review!

Joibel:
This breaks the workflow example from #10426. Without this change the workflow doesn't correctly memoize but does run correctly to completion. With this change you get
```
level=error msg="Recovered from panic" namespace=argo r="runtime error: invalid memory address or nil pointer dereference" stack="goroutine 307 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:24 +0x5e\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).operate.func2()\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:200 +0xb4\npanic({0x20c1e00?, 0x39f0490?})\n\t/usr/local/go/src/runtime/panic.go:920 +0x270\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).runOnExitNode(0xc00150a0c0, {0x27b8560, 0x3a5c2c0}, 0x0, 0xc000d07980, {0xc0015265a0, 0x1b}, 0xc000fe9650?, {0xc000aa1db8, 0x14}, ...)\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/exit_handler.go:20 +0xa6\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).executeDAGTask(0xc00150a0c0, {0x27b8560, 0x3a5c2c0}, 0xc00034b490, {0xc000b92a90, 0xe})\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/dag.go:614 +0x2232\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).executeDAG(0xc00150a0c0, {0x27b8560, 0x3a5c2c0}, {0xc001526540, 0x1b}, 0xc000404b80, {0xc000acab40, 0x21}, 0xc000adfd40, {0x27bbe20, ...}, ...)\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/dag.go:269 +0x42a\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).executeTemplate(0xc00150a0c0, {0x27b8560, 0x3a5c2c0}, {0xc001526540, 0x1b}, {0x27bbe20, 0xc000b46000?}, 0x46add3?, {{0x0, 0x0, ...}, ...}, ...)\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:2120 +0x2d4a\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).operate(0xc00150a0c0, {0x27b8560?, 0x3a5c2c0})\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:364 +0x18cb\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*WorkflowController).processNextItem(0xc00042f900, {0x27b8560, 0x3a5c2c0})\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/controller.go:802 +0x687\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*WorkflowController).runWorker(0xc000d846a0?)\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/controller.go:719 +0x88\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0x30?)\n\t/home/vscode/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:155 +0x33\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil(0x0?, {0x27909c0, 0xc000d0a870}, 0x1, 0xc0006b6240)\n\t/home/vscode/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:156 +0xaf\nk8s.io/apimachinery/pkg/util/wait.JitterUntil(0x0?, 0x3b9aca00, 0x0, 0x0?, 0x0?)\n\t/home/vscode/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:133 +0x7f\nk8s.io/apimachinery/pkg/util/wait.Until(0x0?, 0x0?, 0x0?)\n\t/home/vscode/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:90 +0x1e\ncreated by github.com/argoproj/argo-workflows/v3/workflow/controller.(*WorkflowController).Run in goroutine 50\n\t/home/vscode/go/src/github.com/argoproj/argo-workflows/workflow/controller/controller.go:339 +0x185a\n" workflow=memoized-workflow-testgbx2b
```

That's how it comes out in the logs, sorry it's horridly formatted.

Could you look into this?

Joibel (changes requested):
This causes the workflow given as an example in #10426 to crash the controller.

shuangkun:
> This causes the workflow given as an example in #10426 to crash the controller.

Thanks for your good find.   I test it and find the root cause is when the node hit the memoize, it won't has children even though it is a retry type node.  So i think we should change node output from children when it didn't hit memoize.

workflow/controller/exit_handler.go
```diff
@@ -15,7 +15,7 @@ import (
 
 func (woc *wfOperationCtx) runOnExitNode(ctx context.Context, exitHook *wfv1.LifecycleHook, parentNode *wfv1.NodeStatus, boundaryID string, tmplCtx *templateresolution.Context, prefix string, scope *wfScope) (bool, *wfv1.NodeStatus, error) {
 	outputs := parentNode.Outputs
-	if parentNode.Type == wfv1.NodeTypeRetry {
+	if parentNode.Type == wfv1.NodeTypeRetry && !(parentNode.MemoizationStatus != nil && parentNode.MemoizationStatus.Hit) {
```
Joibel, line 18: This is a complex thing to get your head around (not that it wasn't before your change). I agree it does the right thing, but I'd like to wrap up this if and the subsequent `getChildNodeIndex()` call into a single helper function.
It feels like it may result in breakage in the future unless we do this.

```
// Check if we have a retry node which wasn't memoized and return that if we do
func (woc *wfOperationCtx) possiblyGetRetryChildNode(node *wfv1.NodeStatus) *wfv1.NodeStatus {
	if node.Type == wfv1.NodeTypeRetry && !(node.MemoizationStatus != nil && node.MemoizationStatus.Hit) {
		return getChildNodeIndex(node, woc.wf.Status.Nodes, -1)
	}
	return nil
}
```

and then this becomes
```
if lastChildNode := possiblyGetRetryChildNode(node); lastChildNode != nil {
  outputs = lastChildNode.Outputs
}
```

workflow/controller/hooks.go
```diff
@@ -75,7 +75,7 @@ func (woc *wfOperationCtx) executeTmplLifeCycleHook(ctx context.Context, scope *
 		// executeTemplated should be invoked when hookedNode != nil, because we should reexecute the function to check mutex condition, etc.
 		if execute || hookedNode != nil {
 			outputs := parentNode.Outputs
-			if parentNode.Type == wfv1.NodeTypeRetry {
+			if parentNode.Type == wfv1.NodeTypeRetry && !(parentNode.MemoizationStatus != nil && parentNode.MemoizationStatus.Hit) {
```
Joibel, line 78: And we can use that here

workflow/controller/operator.go
```diff
@@ -2978,7 +2988,7 @@ func (woc *wfOperationCtx) requeueIfTransientErr(err error, nodeName string) (*w
 func (woc *wfOperationCtx) buildLocalScope(scope *wfScope, prefix string, node *wfv1.NodeStatus) {
 	// It may be that the node is a retry node, in which case we want to get the outputs of the last node
 	// in the retry group instead of the retry node itself.
-	if node.Type == wfv1.NodeTypeRetry {
+	if node.Type == wfv1.NodeTypeRetry && !(node.MemoizationStatus != nil && node.MemoizationStatus.Hit) {
```
Joibel, line 2991: And here.

Joibel (changes requested):

workflow/controller/exit_handler.go
shuangkun: Thanks! I agree with you！ I have changed all.

Joibel (approved):
Nice, thank you.
LGTM, @terrytangyuan could you take a look please.

terrytangyuan (commented):
Can you fix the PR title?

shuangkun:
> Can you fix the PR title?

Yes，Is this okay?

terrytangyuan (approved):

PR #12492, opened 2024-01-09
fix(resources): improve ressource accounting. Fixes #12468

### LINKED ISSUE #12468: Resource duration is overestimated for cpu
### Pre-requisites

The resourcesDuration field in a completely workflow currently greatly overestimates the cpu consumption of the workflows. The example workflow below, for instance, ends with ~125s of consumed CPU, while we would expect it to be closer to 60s.

The issue comes from this line: https://github.com/argoproj/argo-workflows/blob/main/util/resource/summary.go#L33

When the `q.Value()` function is called, small decimal quantities are rounded up. As a result, the `wait` container that requests `10m` cores, is counted as if it requested a full core.

See the snippet below for a sample that shows the issue with the `Value()` function:
```go
import (
	"fmt"

	"k8s.io/apimachinery/pkg/api/resource"
)

func main() {
	a := resource.MustParse("10m")
	fmt.Println(a.Value())      // 1
	fmt.Println(a.MilliValue()) // 10
	fmt.Println(a.AsDec())      // 0.010

	a = resource.MustParse("1000m")
	fmt.Println(a.Value())      // 1
	fmt.Println(a.MilliValue()) // 1000
	fmt.Println(a.AsDec())      // 1
}
```
Because of this, small CPU requests are rounded up to one core, and the cpu resources consumed by workflows are greatly overestimated.

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: resources-test-
  namespace: workflows
spec:
  serviceAccountName: server
  entrypoint: sleep
  templates:
    - name: sleep
      container:
        image: ubuntu:22.04
        command: [ sleep ]
        args: [ "60" ]
        resources:
          requests:
            memory: 128Mi
            cpu: 1
```

```text
Nothing relevant
```

```text
Nothing relevant
```

AloysAqemia:
Fixes #12468

### Motivation

Fix cpu ressources consumption estimation.

### Modifications

Change the use of the `.Values()` function (which rounds ressources up to an integer) to `.MilliValues()` which is more accurate for CPU resources, which can be fractional.
The `MilliValues()` documentation mentions that this function may overflow, but the return value is stored in an int64, so the overflow will happen if a container requests more than 9*10^(15) of a ressource. I assume this should be fine for a while.

### Verification

Tested by submitting the workflow attached to the issue this PR fixes. The result is now what is expected:
```
      resourcesDuration:
        cpu: 63
        memory: 101
```

### DIFF
--- a/util/resource/summary.go
+++ b/util/resource/summary.go
@@ -28,9 +28,14 @@ func (ss Summaries) Duration() wfv1.ResourcesDuration {
 	// Add container states.
 	d := wfv1.ResourcesDuration{}
 	for _, s := range ss {
+		// age is converted to seconds, otherwise the multiplication below is very likely to overflow
 		age := int64(s.age().Seconds())
 		for n, q := range s.ResourceList {
-			d = d.Add(wfv1.ResourcesDuration{n: wfv1.NewResourceDuration(time.Duration(q.Value() * age / wfv1.ResourceQuantityDenominator(n).Value() * int64(time.Second)))})
+			d = d.Add(wfv1.ResourcesDuration{
+				n: wfv1.NewResourceDuration(time.Duration(
+					q.MilliValue()*age/wfv1.ResourceQuantityDenominator(n).MilliValue(),
+				) * time.Second),
+			})
 		}
 	}
 	return d
--- a/util/resource/duration_test.go
+++ b/util/resource/duration_test.go
@@ -44,7 +44,7 @@ func TestDurationForPod(t *testing.T) {
 			corev1.ResourceCPU:    wfv1.NewResourceDuration(2 * time.Minute),
 			corev1.ResourceMemory: wfv1.NewResourceDuration(1 * time.Minute),
 		}},
-		{"ContainerWithCPURequest", &corev1.Pod{
+		{"ContainerWithGPULimit", &corev1.Pod{
 			Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "main", Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
 					corev1.ResourceCPU: resource.MustParse("2000m"),


agilgur5 (commented):
Code LGTM, nice catch!

I think this doesn't require any changes to [the docs](https://argo-workflows.readthedocs.io/en/latest/resource-duration/) either

The unit tests need some modifications now though as they are [failing](https://github.com/argoproj/argo-workflows/actions/runs/7463449500/job/20308168312?pr=12492) due to different expected values.
The E2E test failure seems like a flake at a glance

AloysAqemia:
Thanks for taking a look @agilgur5 ! 
I think there's an issue in the code, the memory shouldn't be 0. I'll look into it.

AloysAqemia:
It turns out removing the conversion of age into seconds caused an overflow as `time.Second == 1e9`. This is now fixed, this PR should be good now.

util/resource/summary.go
```diff
@@ -28,9 +28,14 @@ func (ss Summaries) Duration() wfv1.ResourcesDuration {
 	// Add container states.
 	d := wfv1.ResourcesDuration{}
 	for _, s := range ss {
+		// age is converted to seconds, otherwise the multiplication below is very likely to overflow
 		age := int64(s.age().Seconds())
 		for n, q := range s.ResourceList {
-			d = d.Add(wfv1.ResourcesDuration{n: wfv1.NewResourceDuration(time.Duration(q.Value() * age / wfv1.ResourceQuantityDenominator(n).Value() * int64(time.Second)))})
+			d = d.Add(wfv1.ResourcesDuration{
+				n: wfv1.NewResourceDuration(time.Duration(
+					q.MilliValue()*age/wfv1.ResourceQuantityDenominator(n).MilliValue(),
```
agilgur5, line 36:
```suggestion
					q.MilliValue() * age / wfv1.ResourceQuantityDenominator(n).MilliValue(),
```
spaces would definitely make this a little easier on the eyes 😅 

Or did `gofmt` make it have no spaces?

util/resource/summary.go
```diff
@@ -28,9 +28,14 @@ func (ss Summaries) Duration() wfv1.ResourcesDuration {
...
+				) * time.Second),
```
agilgur5, line 37: so is the previous `int64` conversion on `time.Second` here no longer necessary?

agilgur5:
Oh, good find! I would not have expected an overflow and did not know that `time.Second` is that big 😮

agilgur5 (commented):
just a small style comment and a question for clarification

util/resource/summary.go
AloysAqemia: It is still necessary, I've just moved it out to the next line to avoid the cast to `int64`, and to make sure that the division by the `ResourceQuantityDenominator.MilliValue()` (which is `1e11` for memory) is done before the multiplication by `time.Second` to ensure we don't overflow.

util/resource/summary.go
AloysAqemia: I agree with you, `gofmt` did that unfortunately :/

util/resource/summary.go
AloysAqemia: Sorry, looking back at this I think I read your question too fast previously. 
The cast to `int64` is no longer necessary, since everything on line 36 is casted to a `time.Duration`, we can multiply directly with `time.Second` which is also of type `time.Duration`.

agilgur5: Gotcha thanks for the clarification there, just wanted to make sure.

Rephrasing, before it was multiplied inside of the `time.Duration(...)` and now it's outside of that to avoid the overflow as you mentioned. Since it's outside now, it no longer needs to cast to an `int64`

agilgur5 (approved):

terrytangyuan (approved):

PR #12500, opened 2024-01-13
fix(docs): correct duplicated title in `h1` and `title` tags in redirecting `gh-pages` branch

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/12443#discussion_r1451024296 and https://github.com/argoproj/argo-workflows/pull/12443#discussion_r1451032764

### Motivation

- `h1` erroneously had `page_heading - Argo Workflows - The workflow engine for Kubernetes` when it should just be `page_heading`
- `title` erroneously had the above plus another ` - Argo Workflows - The workflow engine for Kubernetes` after it

- also, `proposals` were previously unintentionally included in the docs site, but they should only be on GitHub
  - they're explicitly excluded in future builds (per #12360 ([this line](https://github.com/argoproj/argo-workflows/pull/12360/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R42)))
  - they also weren't configured to show up anywhere in the ToC etc in the `mkdocs.yml`
    - but they would show up in search, unintentionally, as they did exist on the old docs site 

### Modifications

1. `Find+Replace` on " - Argo Workflows - The workflow engine for Kubernetes\</h1>" with "\</h1>"
2. `Find+Replace` on " - Argo Workflows - The workflow engine for Kubernetes - Argo Workflows - The workflow engine for Kubernetes" with " - Argo Workflows - The workflow engine for Kubernetes"
3. `rm -rf proposals/`

### Verification

- Checked a handful of individual docs manually
- Made sure that the previous phrases didn't exist anymore with another `Find`

PR #12501, opened 2024-01-13
docs: add internal redirects for moved or renamed pages

agilgur5:
Follow-up to #11516, #12042, https://github.com/argoproj/argo-workflows/issues/11390#issuecomment-1876274288

### Motivation

- This will allow us to move and rename pages in the docs, with two existing uses:
  - `webhdfs.md` was under `use-cases/`, though that it is for generic things like "Machine Learning"
    - move it under `patterns` in the ToC instead, which is more appropriate

  - `manually-create-secrets.md` had a confusing sub-path name, renamed to match the heading: `service-account-secrets.md`
    - the new name matches its current heading

### Modifications

- add [`mkdocs-redirects`](https://github.com/mkdocs/mkdocs-redirects) plugin and configuration
  - also add the default plugin `search` to the list as it gets overridden (this is the first plugin we're using actually (vs theme or markdown extension))
    - as per the `mkdocs-redirects` [README](https://github.com/mkdocs/mkdocs-redirects/blob/3b36c72dc50c41da5fcf9604be845648d415db0d/README.md?plain=1#L29) and the `mkdocs` [config docs](https://www.mkdocs.org/user-guide/configuration/#plugins)
  
- convert `webhdfs.md` page's own links as it is no longer under a subdirectory (`../` -> ``)
- convert 1 link to `manually-create-secrets.md` to its new name

### Verification

1. `mkdocs build`
1. `open site/use-cases/webhdfs/index.html` causes a redirect (it uses both the `meta` `refresh` tag as well as a back-up `script`
    - new generated code from the plugin is:
    ```html
    <!doctype html>
    <html lang="en">
    <head>
        <meta charset="utf-8">
        <title>Redirecting...</title>
        <link rel="canonical" href="../../webhdfs/">
        <meta name="robots" content="noindex">
        <script>var anchor=window.location.hash.substr(1);location.href="../../webhdfs/"+(anchor?"#"+anchor:"")</script>
        <meta http-equiv="refresh" content="0; url=../../webhdfs/">
    </head>
    <body>
    Redirecting...
    </body>
    </html>
    ```
      - interestingly, the back-up `script` does take into consideration anchors, which I mentioned in https://github.com/argoproj/argo-workflows/pull/12362#issuecomment-1871470972
1. same for `site/manually-create-secrets/index.html`

Screenshot of `webhdfs.md` correctly placed in the ToC under "Patterns" and _not_ "Use Cases":
Screenshot 2024-01-12 at 7 58 40 PM


jmeridth:
> interestingly, the back-up script does take into consideration anchors, which I mentioned in https://github.com/argoproj/argo-workflows/pull/12362#issuecomment-1871470972

Do we want to add this to #12500 or a future PR or just stating an observation?

jmeridth (approved):
Nice find on the plugin.

agilgur5:
> Do we want to add this to #12500 or a future PR or just stating an observation?

I haven't tested it yet, but if it does properly redirect anchors (i.e. if the JS goes first before the HTML redirect), I do think it would be good to add in a future PR.

isubasinghe (approved):

agilgur5:
Backported cleanly to `release-3.5` as 72504d1466e9ebd47c51b66ea773d17571bd3f15

PR #12502, opened 2024-01-13
docs: add unicode char next to external links in header/ToC

agilgur5:
Follow-up to https://github.com/argoproj/argo-workflows/issues/11390#issuecomment-1876274288 
- I noticed this while looking at CD's `mkdocs.yml` for differences

### Motivation

- [CD](https://github.com/argoproj/argo-cd/blob/d9df2525c57d4870215a6ce149dbedd08ae05fdb/mkdocs.yml#L222) and [Rollouts](https://github.com/argoproj/argo-rollouts/blob/bc992a3fe082e6dc19529194b331d7134928257d/mkdocs.yml#L147) both have this unicode character "⧉" to indicate that a link is external
  - it's also a nice touch to indicate an external link (vs. an internal link)
  - here's a screenshot of CD's ToC: 
Screenshot 2024-01-12 at 9 14 58 PM

  
### Modifications

- add this to Workflows's `mkdocs.yml`'s headers/ToC as well for consistency

### Verification

1. `mkdocs build` passes
1. Screenshot: Screenshot 2024-01-12 at 9 13 51 PM

### Future Work

It may be a good idea to add this to the UI or inside the docs themselves
- CD has this in exactly two other places [internally in its docs](https://github.com/search?q=repo%3Aargoproj%2Fargo-cd%20%E2%A7%89&type=code) for reference

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -267,7 +267,7 @@ nav:
       - releasing.md
       - survey-data-privacy.md
   - Roadmap: roadmap.md
-  - Blog: https://blog.argoproj.io/
-  - Slack: https://argoproj.github.io/community/join-slack
-  - Twitter: https://twitter.com/argoproj
-  - LinkedIn: https://www.linkedin.com/company/argoproj/
+  - Blog ⧉: https://blog.argoproj.io/
+  - Slack ⧉: https://argoproj.github.io/community/join-slack
+  - Twitter ⧉: https://twitter.com/argoproj
+  - LinkedIn ⧉: https://www.linkedin.com/company/argoproj/


jmeridth (approved):

terrytangyuan (commented):
This doesn't look appearing though

agilgur5:
> This doesn't look appearing [sic] though

Do you mean the "⧉" character specifically, or _anything_ next to external links?

I do think the indication is a nice touch, as mentioned above, but I agree that this is not an ideal character as it is not quite the same as the actual [external link icon](https://qr.ae/pKdSJH).
I'm not sure if we have fontawesome (or similar) available already in the docs or not, the proper icon would be more ideal, but I think this is better than nothing

terrytangyuan (approved):

agilgur5:
> I'm not sure if we have fontawesome (or similar) available already in the docs or not

There are indeed [a few icon packs](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/) available with the material theme. `:octicons-link-external-16:` looks like the closest thing.
But it seems like icons [in headings, sidebars, & tabs](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/#icons-emojis-in-sidebars) may only be available for sponsors of the theme, so that may not apply for this specifically.

Issue #10107, opened 2022-11-25
Invalid node IDs in workflow.status


mweibel:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

Since #8748 pod names use the V2 naming which contains the template name in the pod name. However the implementation did not update the Workflow.Status.Nodes map to contain the correct pod name anymore. There's a disconnect between NodeIDs and pod names which wasn't the case before. This makes it impossible to look at a argo workflow status, take the nodeID and use it to know which pod it belongs to. 

This is a follow-up of #9906. I initially thought this would be the same case but apparently is not. 

The below workflow triggers the following workflow:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  annotations:
    workflows.argoproj.io/pod-name-format: v2
  creationTimestamp: "2022-11-25T11:33:41Z"
  generateName: nodename-
  generation: 3
  labels:
    workflows.argoproj.io/phase: Running
  name: nodename-bvd45
  namespace: argo
  resourceVersion: "15649"
  uid: ea233eef-210d-4394-a238-ef847b104458
spec:
  activeDeadlineSeconds: 300
  arguments: {}
  entrypoint: render
  podSpecPatch: |
    terminationGracePeriodSeconds: 3
  templates:
  - inputs: {}
    metadata: {}
    name: render
    outputs: {}
    steps:
    - - arguments:
          parameters:
          - name: frames
            value: '{{item.frames}}'
        name: run-blender
        template: blender
        withItems:
        - frames: 1
  - container:
      args:
      - /argosay echo 0/100 $ARGO_PROGRESS_FILE && /argosay sleep 10s && /argosay
        echo 50/100 $ARGO_PROGRESS_FILE && /argosay sleep 10s
      command:
      - /bin/sh
      - -c
      image: argoproj/argosay:v2
      name: ""
      resources: {}
    inputs:
      parameters:
      - name: frames
    metadata: {}
    name: blender
    outputs: {}
    retryStrategy:
      limit: 2
      retryPolicy: Always
status:
  artifactGCStatus:
    notSpecified: true
  artifactRepositoryRef:
    artifactRepository:
      archiveLogs: true
      s3:
[...]

Issue #9193, opened 2022-07-20
`retryStrategy` with `nodeAntiAffinity` is not working as expected.


cwichka:
## Summary

What happened/what you expected to happen?

I wanted to use `retryStrategy` with `nodeAntiAffinity` in order to prevent retrials from running on the same hosts. I was using the following small workflow for testing, but what happens is all retrials were started on the same host (node).
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: random-fail-
spec:
  entrypoint: random
  templates:
  - name: random
    retryStrategy:
      limit: 10
      retryPolicy: "Always"
      affinity:
        nodeAntiAffinity: {}
    script:
      image: python:alpine3.6
      command: [python]
      source: |
        import random
        import time
        random.seed(time.time())
        i = random.randint(0, 10)
        print(i)
        exit(i)`
```
Not sure if this is an expected behavior and in my case i should use `RetryNodeAntiAffinity` which is, as mentioned in the documentation, is a placeholder for future expansion.

What version are you running?
Tested with v3.2.9 and v3.3.3

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

alexec:
#4679

alexec:
@goock could you look into this pls?

goock:
> @goock could you look into this pls?

I run the example from the ticket, and from above 30K feet, it looks like the controller stopped adding `Affinity` to the pod's spec, or it's removed at some point, I'll start debugging and find the cause.

goock:
@alexec I started working on this. 
In this function https://github.com/argoproj/argo-workflows/blob/9d66b69f0bca92d7ef0c9aa67e87a2e334797530/workflow/controller/retry_tweak.go#L15 I'm taking the template from the boundary node and then finding the retry node with the same template, which works for more complicated scenarios but does not work for very simple ones from this ticket - the `BoundaryID` is empty.

I see four possible solutions to the issue, but I need advice on this:

1. Modify `FindRetryNode`; if the `BoundaryID` is empty, take the template from the node and then find the retry node. I would need confirmation if `TemplateName` is not empty in all cases or at least when `BoundaryID` is empty.

2. Create another traverse function (like this one https://github.com/argoproj/argo-workflows/blob/master/workflow/util/retry/retry.go#L10), walk over the nodes tree and find the retry node. This solution would be clumsy and inefficient.

[...]

PR #12503, opened 2024-01-13
feat: store podname in nodestatus

### LINKED ISSUE #12528: Store podname in Node status to ensure single source of truth for PodName
# Summary

The PodName should be stored in the NodeStatus to reduce bugs that have been occurring relating to incorrect PodNames.
This issue was opened as a result of this comment: https://github.com/argoproj/argo-workflows/issues/10267#issuecomment-1630308619

## Use Cases
N/A

isubasinghe:
Fixes #12528

### Modifications

The changes are relatively simple, I reduced the pod name generation code as much as possible and stored the PodnName in
the node itself.

### Verification

Verified via testing, this change introduces a slight change to existing behaviour however. The change here is that we must have created a pod before a podName is assigned.


sarabala1979:
@isubasinghe why do we need the pod name in node status? Can you provide more details about usecase/scenario?
This will increase the workflow object size that will reduce the  number of steps.
if the usecase is to find the pod for that particular node, node name can be added as label in the pod object.

cmd/argo/commands/common/get.go
```diff
@@ -477,23 +477,28 @@ func renderChild(w *tabwriter.Writer, wf *wfv1.Workflow, nInfo renderNode, depth
...
 	duration := humanize.RelativeDurationShort(node.StartedAt.Time, node.FinishedAt.Time)
 	if node.Type == wfv1.NodeTypePod {
-		podName := util.GeneratePodName(wfName, nodeName, templateName, node.ID, podNameVersion)
+		podName := ""
+		if node.PodName != nil {
+			podName = *node.PodName
+		} else {
+			expectedPodName := util.GeneratePodName(wfName, node.Name, templateName, node.ID, podNameVersion)
+			panic("podName absent expected " + expectedPodName + " for " + node.Name)
+		}
```
isubasinghe, line 501: This is probably not the desirable behaviour, there might be cases where the pod has not been created yet.
What should we do in this case?

agilgur5 (commented):
I only took a quick glance at this, but doesn't this need to modify [the UI code as well](https://github.com/argoproj/argo-workflows/blob/3652241a42be8e1cc699719f97ff7af2c7ca86c5/ui/src/app/shared/pod-name.ts)?

Also, for backward-compatibility, we'd still have to generate the pod names for Workflows created before this change.

terrytangyuan (commented):
We'll need to test this thoroughly this time

isubasinghe:
> I only took a quick glance at this, but doesn't this need to modify [the UI code as well](https://github.com/argoproj/argo-workflows/blob/3652241a42be8e1cc699719f97ff7af2c7ca86c5/ui/src/app/shared/pod-name.ts)?

I thought I replied to this, must have just not pressed the comment button I suppose, this was on purpose, wanted to make the backend changes first and follow it up with the frontend changes, but happy to make the UI changes here as well if you'd like me to.

> Also, for backward-compatibility, we'd still have to generate the pod names for Workflows created before this change.

Hmmm yeah you are right. What I might do is also keep the previous pod name generation code. 
We can than then also use that to test if these changes kept the original behaviour.

HumairAK:
Bumping this, @isubasinghe do you think you can rebase and resolve the conflicts? 

@terrytangyuan  this is something we use in kubeflow pipelines to retrieve pod names (we are [currently planning ](https://github.com/kubeflow/pipelines/pull/10568) to use a workaround [you prescribed here](https://github.com/argoproj/argo-workflows/issues/10107#issuecomment-1540064995)), we are very much interested in seeing this merged, so we can rely on Argo to provide this value, what can we do to get this some more traction?

terrytangyuan:
We can discuss this in the upcoming contributors meeting to prioritize.

agilgur5:
@HumairAK this is currently a feature, so it wouldn't land until 3.6. If KFP is only bumping to 3.4, this PR wouldn't solve your problem

agilgur5 (commented):
So this certainly simplifies things moving forward (not during a [backward-compat window](https://github.com/argoproj/argo-workflows/pull/12503#pullrequestreview-1828140731) though), but in recent months I had found older comments from Alex C about trying to store as little as possible in the `status` due to our existing issues with large Workflows ([`nodeStatusOffload`](https://argo-workflows.readthedocs.io/en/release-3.5/offloading-large-workflows/) + general memory usage).

I couldn't find his comments in a search, but they were of the form "storing this in `status` would take `n` space for _each_ node". **EDIT**: Here's one of Alex's comments on the topic: https://github.com/argoproj/argo-workflows/issues/9193#issuecomment-1618952357
I'm wondering if that was perhaps the original rationale behind some of the derivations in the codebase -- derive the field instead of storing it in `status`.

Effectively, this is a trade-off between storage and determinism. It actually reminds me quite heavily of the evolution of lockfiles in the JS ecosystem.
There though, determinism is simple to favor over storage, as hard disk/FS storage is cheap and plentiful (although VCS storage is a bit more complicated), compared to in-memory or etcd storage, where we have a hard limit

agilgur5:
@HumairAK ok. That does affect our prioritization though, as that means KFP won't be using this functionality in the short or likely near-term.

agilgur5:
> Effectively, this is a trade-off between storage and determinism.

To be clear, I don't have a strong opinion on that per se; but I think we should be explicit about that -- larger workflows would be affected by this feature.

I probably lean toward agreeing with this feature, as it would simplify areas of the codebase where we do have some bugs / edge-cases currently that are handled in different ways in different places.

> (not during a [backward-compat window](https://github.com/argoproj/argo-workflows/pull/12503#pullrequestreview-1828140731) though)

The deprecation window is actually a bit complicated now that I think about it. If we ever want to remove the pod name generation code, we would break backward-compat with Workflows that were created in a version that doesn't contain the pod name in the `status`.

For instance, say 3.6 puts pod names into the `status` by default and has backward-compat to derive pod names when not present in the status. That backward-compat means the `POD_NAMES` env var still has to be set.

Then say, in 3.7 we were to remove the backward-compat -- all Workflows created with Argo <3.6 would no longer be processable, e.g. for retries or Pod logs in the UI and possibly other features.

With that in mind, the pod name derivation code is going to still have to be around for a while. Defaulting to this feature means that'd we could eliminate bugs for newer Workflows, but might still have them present in older Workflows. So the frequency of the old bugs would decrease, but we may want to still fix them.
Or in other words, the backward-compat requirement decreases some of the benefit of this feature.

terrytangyuan (commented):
One concern is that this will increase the object size.

agilgur5:
Discussed in [today's Contributor Meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.shpizyln62nu). The storage concern that I brought up [earlier](https://github.com/argoproj/argo-workflows/pull/12503#pullrequestreview-2004004741) was still relevant, especially as we've had more users recently reporting etcd getting full (particularly on managed k8s providers where you cannot configure your etcd space), such as #12802.

But given that I brought it up and I personally think the trade-off is worthwhile in this case -- determinism over storage and treat storage as a separate problem to optimize independent of any one field -- this has the green light to go.

We do need to implement the [backward compat](https://github.com/argoproj/argo-workflows/pull/12503#issuecomment-2059625518) in this PR as well as decide whether pod names will be on all nodes or just pod nodes.

HumairAK:
> this has the green light to go.

hmm would still be interested in seeing this change 

@isubasinghe do you have the bandwidth to continue work on this?

isubasinghe:
@HumairAK will continue on this when I get the chance

PR #12514, opened 2024-01-14
refactor(ui): use `import type` syntax where possible

agilgur5:
Unblocks #12516
Follow-up to #12163 & #12290, which unblocked this
Follow-up to https://github.com/argoproj/argo-workflows/pull/12150#discussion_r1382510455 
Same changes can be made for #12158 and the `argo-ui` warnings introduced in #12061

### Motivation

- now that we've migrated from archived `tslint` to `eslint` and updated `prettier` from v1 to v3, we can _finally_ use this [newer TS syntax from TS 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export)
  - this fixes a few warnings in the build as well as makes things more explicit and efficient
    - **note**: there are more warnings to fix around `export type`, but those are all from `argo-ui`'s source code and so can't be fixed from within this repo

### Modifications

- use `import type` syntax in a few files where it is only types that are imported

### Verification

1. `yarn build` works fine, with a few less warnings now
2. `make start UI=true` and tested UI, esp. `monaco-editor`, all still works fine

PR #12516, opened 2024-01-14
build(ui): improve performance with `esbuild-loader`

agilgur5:
Follow-up to #11628

### Motivation

- similar to [Argo CD](https://github.com/argoproj/argo-cd/blob/d9df2525c57d4870215a6ce149dbedd08ae05fdb/ui/src/app/webpack.config.js#L37) and many other projects, we can use `esbuild` instead of TS etc to significantly speed up build times
  - (as `esbuild` is written in Go and compiled instead of using interpreted TS/JS)
  - we don't use any Babel or TS plugins, so I think this codebase is a good candidate for that
  - it also simplifies the build process as well, removing & replacing many other deps
  - caveats:
    - `esbuild` does not do type-checking, so we have to run `tsc --noEmit` as part of the `lint` phase
    - `esbuild-loader` does not currently support React Fast Refresh (c.f. https://github.com/privatenumber/esbuild-loader/issues/250, https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/366) <details>
      - `react-hot-loader` [was replaced](https://github.com/gaearon/react-hot-loader/tree/v4.13.1?tab=readme-ov-file#moving-towards-next-step) by [React Fast Refresh](https://github.com/pmmmwh/react-refresh-webpack-plugin) and is deprecated / hasn't received an update in 2 years
      - that being said, `esbuild` is significantly faster than using Babel or TS for builds, so hot reloading is nice but not necessary
        - also hot reloading can be buggy
        - plus, in my experience, the dev build often falls back to a full page refresh anyway

</details>

- resolve various warnings in the build, during the install, and in the console

### Modifications

- remove `babel-loader`, `ts-loader`, `react-hot-loader`, and `source-map-loader` <details>
  - note that `source-map-loader` is still present in the `yarn.lock` as it is used under-the-hood by other deps
    - but the warnings from directly using `source-map-loader` are no longer present in the build
    - and the bundle size shrank a bit, seemingly bc there is no `autolinker` etc dep in the bundle, but potentially other reasons (I did not check in-depth)

</details>

- add `esbuild-loader`

- add `tsc --noEmit` to the `lint` script
- add [`isolatedModules`](https://www.typescriptlang.org/tsconfig#isolatedModules) config to [ensure compatibility](https://esbuild.github.io/content-types/#isolated-modules) with `esbuild`'s compiler
  - since it has no type-checking, it can't do full program analysis, meaning there are certain types it doesn't support
  - same as when [using Babel to compile TS](https://babeljs.io/docs/babel-plugin-transform-typescript#typescript-compiler-options)

- remove `mdl.hot` etc usage of `react-hot-loader`

#### Related config / performance changes

- further optimize dev builds by using cheaper `eval` sourcemaps as [recommended by Webpack for performance](https://webpack.js.org/configuration/devtool/)
  - prod builds should always have highest quality sourcemaps, so do not change those
  - this was also discussed in https://github.com/argoproj/argo-workflows/pull/11716#discussion_r1309661376

- remove unused `react-paginate` `exclude`
  - it's not a dep we currently use; it doesn't exist in [the lockfile](https://github.com/argoproj/argo-workflows/blob/fa27edb845d3e4407f27f4f383d3d78620e11ee3/ui/yarn.lock)
    - (nor in `argo-ui`'s [lockfile](https://github.com/argoproj/argo-ui/blob/c65a9520366b0c0cf0ac585618d029e60041a94d/yarn.lock), of which all prod deps would be in Workflows's lockfile as well)

### Verification

1. `yarn build` works fine
1. `make start UI=true` works fine and UI seems to function as expected

#### Performance Testing

some simple performance testing when running natively on macOS, taken directly from Webpack's reported stats:
- before:
  - prod build: ~65s
  - dev build: ~22s
  - dev rebuild: ~1.8s
- after:
  - prod build: ~50s (~25% faster)
  - dev build: ~16s (~25% faster)
  - dev rebuild: ~.8s (~50% faster)

#### Future Work

these are not as good improvements as I would have liked, which mean there are bigger bottlenecks elsewhere
  - e.g. Webpack itself (analysis, source maps, running loaders), `sass-loader`, etc
    - `sass` has been slow in the past and is ripe to be replaced with more modern PostCSS as well
    - switching to `esbuild` directly may further improve these stats
      - i.e. instead of Webpack + `esbuild-loader`
  - independently building `argo-ui` instead of as part of the Workflows (and CD) builds per https://github.com/argoproj/argo-ui/issues/453
  - optimizing more deps per #12059 and related comments, such as removing `moment`, consolidating similar deps, etc


agilgur5:
Gah, this already got a merge conflict with a Dependabot PR #12511. Adding `prioritized-review` label to this and #12487 as a result

agilgur5:
Oh, interesting, due to the new `isolatedModules` config, the build is [failing](https://github.com/argoproj/argo-workflows/actions/runs/7516274931/job/20461171832?pr=12516) due to missing `import type`. i.e. this PR actually depends on #12514

ui/package.json
```diff
@@ -8,7 +8,7 @@
     "scripts": {
         "build": "rm -Rf dist && NODE_OPTIONS='--openssl-legacy-provider' NODE_ENV=production webpack --mode production --config ./src/app/webpack.config.js",
         "start": "NODE_OPTIONS='--no-experimental-fetch --openssl-legacy-provider' webpack-dev-server --config ./src/app/webpack.config.js",
-        "lint": "eslint --fix ./src/app",
+        "lint": "eslint --fix ./src/app && tsc --noEmit",
```
isubasinghe, line 11: Just to clarify any type checking failures will not be caught until lint is run now?

isubasinghe: I am wondering if SWC is a better option than esbuild if thats the case?

agilgur5: Build won't type-check anymore, correct. This adds it to the `lint` script and the tests do still run with `ts-jest` (but they have very little coverage, and it's ripe to make faster with a different plugin too).
This has basically emerged as the most common pattern for TS codebases in recent years. Build with Babel or `esbuild`, type-check with `tsc`. Since I maintained TSDX I was at the forefront of that a few years ago and still maintain `rollup-plugin-typescript2`, which hasn't had as much usage recently in part due to this pattern (and Vite & `esbuild` popularity in general).

Re: SWC, it also does not support type-checking, although there is an up and coming Rust library for that: https://github.com/swc-project/swc/issues/571.
So basically type-checking is currently limited to the TS compiler itself (and its integrations); all other TS-compatible build tools do not perform type-checking.

I also currently prefer `esbuild` over SWC as it tries to read `tsconfig.json` whereas SWC has its own configuration (it is similar to Babel's config in principle, but not compatible). (`esbuild` is also more popular and has more features currently, afaik)

isubasinghe: Hmm okay, thanks for the explanation. I personally would like to get type checking but if that is the current pattern, I am happy to go with it. Perhaps  we could add an option to make it type check as well in the future, but happy for that to be a new feature in the future.

isubasinghe (approved):
Nice work, this is so much faster.
Works as expected

ui/package.json
agilgur5: To be explicit, we do have type-checking, just not during `yarn build` specifically. 

I don't really have a strong opinion on the pattern (I didn't in the above discussions a few years ago either), but the thought process makes sense, where you get your code working first and then you make sure your types, static analysis, tests, etc all pass. Although certainly you can get bugs because those are failing.

For full incremental development, some people also run 4 separate terminals simultaneously: dev build in watch mode, type checker in watch mode, linter in watch mode, and tests in watch mode. Separating out all the concerns explicitly that way.

Also we actually need to improve the strictness of the `tsconfig.json` first, as it's pretty loose. Unfortunately it does require independently building `argo-ui` before we can do that, since the settings have to match right now per https://github.com/argoproj/argo-ui/issues/453. That would also probably result in a large performance improvement for builds similar to this PR.

terrytangyuan (approved):

terrytangyuan:
Build failed

isubasinghe:
@terrytangyuan you need to merge this one first: https://github.com/argoproj/argo-workflows/pull/12514

agilgur5:
Fixed another merge conflict in the `yarn.lock`. Still waiting for #12514 to be merged so that this can be unblocked and merged

juliev0:
@isubasinghe are you comfortable being the Assignee for this one, so you may be able to merge it once you officially have access to?

agilgur5:
Merged `main` since #12514 was merged, but getting a test/build [failure](https://github.com/argoproj/argo-workflows/actions/runs/7768757176/job/21187052202?pr=12516) in `argo-ui` now 😕 . I was trying to partly workaround this in #12158, but otherwise this requires a few changes upstream in `argo-ui` (where [my PRs](https://github.com/argoproj/argo-ui/pulls/agilgur5) have been open even longer 😕)

I guess I'll kick this back to draft as a result -- but uh may want my `argo-ui` PRs to be prioritized 😅

juliev0:
> Merged `main` since #12514 was merged, but getting a [test failure](https://github.com/argoproj/argo-workflows/actions/runs/7768757176/job/21187052202?pr=12516) in `argo-ui` now 😕 . Interestingly enough, the build passes fine (which I had tested), it's only the tests' build that's failing, I think because the tooling is different (`ts-jest` uses the TS compiler directly, `esbuild-loader` uses `esbuild`)?
> 
> I was trying to partly workaround a CSS side-effect import in #12158 for similar reasons to this, but otherwise this requires a few changes upstream in `argo-ui` (where [my PRs](https://github.com/argoproj/argo-ui/pulls/agilgur5) have been open [even longer](https://cloud-native.slack.com/archives/C020XM04CUW/p1705097023721029) 😕)
> 
> I guess I'll kick this back to draft as a result since it is pretty blocked on `argo-ui` -- but uh may want my `argo-ui` PRs to be prioritized 😅 (the label does not exist there ofc)

Okay, yes!

isubasinghe:
> @isubasinghe are you comfortable being the Assignee for this one, so you may be able to merge it once you officially have access to?

Yup happy to merge when the tests pass

agilgur5:
I might be able to get this working again now that https://github.com/argoproj/argo-ui/pull/509 and https://github.com/argoproj/argo-ui/pull/552 have been merged

agilgur5:
Guh, this is now [failing to typecheck](https://github.com/argoproj/argo-workflows/actions/runs/9045160674/job/24854615361?pr=12516) due to another dep, this time very transitive: https://github.com/ramda/ramda/issues/3415#issuecomment-1757260119, https://github.com/swagger-api/swagger-js/issues/3176, https://github.com/swagger-api/apidom/issues/3279

agilgur5:
Ok had to fix that by pinning `types-ramda` as it requires TS v5 per  https://github.com/ramda/ramda/issues/3415#issuecomment-1758719123 and https://github.com/swagger-api/swagger-js/issues/3176#issuecomment-1759429014

agilgur5:
> ### Verification
> 1. `yarn build` works fine
> 2. `make start UI=true` works fine and UI seems to function as expected

Re-verified, seems to all be fine

agilgur5:
I'm gonna go ahead and merge this since I only made the changes necessary to fix CI since this was last approved. It also fixes a bunch of warnings and improves performance etc, so it's pretty darn nice to have sooner rather than later

PR #12518, opened 2024-01-14
docs: update SwaggerUI version

agilgur5:
Follow-up to #10923, where the version of Swagger UI added there was already a year+ old at the time of the PR

### Motivation

- use latest [5.11.0](https://github.com/swagger-api/swagger-ui/releases/tag/v5.11.0) instead of the 2 years outdated [4.5.0](https://github.com/swagger-api/swagger-ui/releases/tag/v4.5.0)
  - it seems like the version was taken directly from [the SwaggerUI installation docs](https://github.com/swagger-api/swagger-ui/blob/v5.11.0/docs/usage/installation.md#unpkg), which still has an old version listed
    - **EDIT**: I sent a PR upstream to update it: https://github.com/swagger-api/swagger-ui/pull/9489
    - in fact, the entire code block was taken from the doc
      - fix invalid inline HTML in this code block (see below) 

- this seems to make it slightly more performant, which is our main issue, but also has other benefits of using a current version of a dep

### Modifications

- update the SwaggerUI dep loaded in a `script` tag on the docs site's Swagger page

- significantly simplify the HTML code block as it renders inside of existing MD/HTML
  - a second `DOCTYPE`, `html`, `head`, `body`, etc tags are actually invalid HTML, but either `mkdocs` ignored some of those or browsers just ignore some of those
    - fixing invalid HTML makes this doc page more robust
  - also fix the indentation in the code block too, which also did not meet style guidelines

- add a `markdownlint-disable` comment to allow the inline HTML
  - add a comment describing why it's necessary
    - similarly add a comment to the PR template explaining why that one is necessary too
  - remove the `Makefile`'s ignore on `swagger.md` for docs linting et al

### Verification

1. `make docs` still passes
    - including `docs-lint`, `docs-spellcheck`, etc
1. `open site/swagger/index.html` renders fine


terrytangyuan:
> this seems to make it slightly more performant, which is our main issue

What performance issue?

terrytangyuan (commented):
What performance issue?

agilgur5:
> What performance issue?

Try opening any API method in the Swagger UI and watch it lag, freeze, or straight up crash the browser window.

It's a very prominent issue and was even mentioned in #10923, the original PR that added this (linked at the top of the description). JM briefly mentioned it [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1703917499533209?thread_ts=1703894735.423229&cid=C0510EUH90V) too and I agreed there that it is very hard to use in its current state due to the lag etc.

As I wrote, this seems to make it better, but does not fix it. It likely needs more performance related fixes upstream, but those will only come in v5 regardless

isubasinghe (approved):
Nice work, this actually makes the UI a lot more usable.

agilgur5:
Fixed merge conflict

agilgur5:
Gonna merge this in as it has already been approved and is a nice perf boost for the docs. I also already upgraded the UI to v5 in #12540, which required a more careful process.

agilgur5:
Backported this update into `release-3.5` in https://github.com/argoproj/argo-workflows/commit/f4470ed61da651d6a5a3859252b14b7dc125fa4e. Will probably do the same for `release-3.4`

PR #12520, opened 2024-01-14
Upgrade nix pkgs

isubasinghe:
Fixes #TODO


terrytangyuan (approved):

hack/patch-nix-hash.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+nix --log-format raw build ./dev/nix/
```
terrytangyuan, line 3: Is installing nix part of the Makefile yet?

isubasinghe: No it isn't, don't really want to do that in the Makefile because the installation is not straightforward (see this: https://nixos.org/download). 


Do you think we should add it or is it okay to leave it out?

caelan-io:
Are we okay to merge this one?

hack/patch-nix-hash.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ $? -eq 0 ]; then
+  echo OK
+else
+  # this is reliant on the vendorHash being on line 195
+  sed -i '195s/vendorHash = \"\([^\"]*\)\"/vendorHash = ""/g' ./dev/nix/flake.nix
+  export NIX_HASH=$(nix --log-format raw build ./dev/nix 2>&1 | grep  "got: " | awk '{ split($0,a,"got:    "); print a[2] }')
+  sed -i '195s|vendorHash = \"\([^\"]*\)\"|vendorHash = "'$NIX_HASH'"|g' ./dev/nix/flake.nix
+  echo $NIX_HASH
+fi
```
agilgur5, line 12:
```suggestion
if [ $? -eq 0 ]; then
  echo "No changes to Nix hash needed"
  return
fi

# this is reliant on the vendorHash being on line 195
sed -i '195s/vendorHash = \"\([^\"]*\)\"/vendorHash = ""/g' ./dev/nix/flake.nix
NIX_HASH=$(nix --log-format raw build ./dev/nix 2>&1 | grep  "got: " | awk '{ split($0,a,"got:    "); print a[2] }')
sed -i '195s|vendorHash = \"\([^\"]*\)\"|vendorHash = "'$NIX_HASH'"|g' ./dev/nix/flake.nix
echo "Changed Nix hash to: $NIX_HASH"
```

- can remove the nesting
- can be more explicit in the logging
- I don't think this needs `export`ing? it's only used in the script itself, so it's a local variable, as far as I can tell from this diff

Makefile
```diff
@@ -255,11 +255,15 @@ argoexec-image:
 	if [ $(DOCKER_PUSH) = true ] && [ $(IMAGE_NAMESPACE) != argoproj ] ; then docker push $(IMAGE_NAMESPACE)/$*:$(VERSION) ; fi
 
 .PHONY: codegen
-codegen: types swagger manifests $(GOPATH)/bin/mockery docs/fields.md docs/cli/argo.md
+codegen: types swagger manifests nix-hash $(GOPATH)/bin/mockery docs/fields.md docs/cli/argo.md
```
agilgur5, line 258: I think this makes more sense inside of the [`lint`](https://github.com/argoproj/argo-workflows/blob/42262690fcbab0ed6106d3f9069e128655b3339f/Makefile#L448) job, which already lints deps for both Go and Node/JS.

The script can be one-lined in there without a new `make` job as well

```suggestion
codegen: types swagger manifests $(GOPATH)/bin/mockery docs/fields.md docs/cli/argo.md
```

Makefile
```diff
@@ -255,11 +255,15 @@ argoexec-image:
...
 	go generate ./...
 	make --directory sdks/java USE_NIX=$(USE_NIX) generate
 	make --directory sdks/python USE_NIX=$(USE_NIX) generate
 
+.PHONY: nix-hash
+nix-hash:
+	./hack/patch-nix-hash.sh
+
```
agilgur5, line 266:
```suggestion
```

see above comment around one-lining this into the `lint` job

.github/workflows/ci-build.yaml
```diff
@@ -336,6 +336,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
+      - uses: cachix/install-nix-action@v22
```
agilgur5, line 339: similarly as the above comments, this should be moved into `lint`
```suggestion
```

hack/patch-nix-hash.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
```
agilgur5, line 2: we should have a comment at the top of this script describing what it does / why it exists / needs to exist

.github/workflows/ci-build.yaml
agilgur5: The patch script also needs to be added to `changed-files`

agilgur5 (changes requested):
I think this makes more sense inside of the `lint` job, rather than `codegen`. We already have deps linting there and this does not generate code.
The `patch-nix-hash` script also needs to be added to `changed-files` in CI

Also made some small improvements to the script

> This PR will likely result in someone needing to run `make codegen` on dependabot PRs relating to golang dependencies.
> I would be happy to do this.

We can automate this similarly to #12234 . Maybe inside of the `lint` CI job? not sure what the best place might be

hack/patch-nix-hash.sh
agilgur5, line 3: We can also store this as a variable here instead of re-running the same command below. Can just parse the output below instead

hack/patch-nix-hash.sh
```diff
@@ -0,0 +1,12 @@
...
+nix --log-format raw build ./dev/nix/
+if [ $? -eq 0 ]; then
+  echo OK
+else
+  # this is reliant on the vendorHash being on line 195
+  sed -i '195s/vendorHash = \"\([^\"]*\)\"/vendorHash = ""/g' ./dev/nix/flake.nix
```
agilgur5, line 8:
```suggestion
```

we can just replace it in the next `sed` command; no need to run a similar command twice.

agilgur5 (commented):
Can optimize the script a bit as well to be more DRY and not repeat some commands / calculations

agilgur5:
> feat:

This probably makes more sense as a `build:` commit as well

isubasinghe:
> I think this makes more sense inside of the `lint` job, rather than `codegen`. We already have deps linting there and this does not generate code. The `patch-nix-hash` script also needs to be added to `changed-files` in CI
> 
> Also made some small improvements to the script
> 
> > This PR will likely result in someone needing to run `make codegen` on dependabot PRs relating to golang dependencies.
> > I would be happy to do this.
> 
> We can automate this similarly to #12234 . Maybe inside of the `lint` CI job? not sure what the best place might be

haha god, that works I suppose, not a fan but defs better than manual effort I suppose.

hack/patch-nix-hash.sh
isubasinghe: Thanks, will change this

hack/patch-nix-hash.sh
isubasinghe: So this ordering actually matters, this is because the error message for a non present nix hash can be different to the error message for an empty hash.

This means that we can't really get rid of this nor store the output as a variable.

.github/workflows/ci-build.yaml
isubasinghe: This is to determine if the lint job should run right? 
any file added there will trigger the lint job to run? 
I think in this case, the go.mod and go.sum files should also be added.

hack/patch-nix-hash.sh
agilgur5: But do we actually care to differentiate the error? If there's an error, it needs a manual fix.
I would think we can just do the replace, and if that fails, print a nicer error message that it needs a manual fix and `exit 1`

.github/workflows/ci-build.yaml
agilgur5: > This is to determine if the lint job should run right?
> any file added there will trigger the lint job to run?

Correct, yes.

> I think in this case, the go.mod and go.sum files should also be added.

They're already there, [under `tests`](https://github.com/argoproj/argo-workflows/blob/13444e663387e3b5b331c278cd9e79fc88968d7e/.github/workflows/ci-build.yaml#L54), which is [inherited by `lint`](https://github.com/argoproj/argo-workflows/blob/13444e663387e3b5b331c278cd9e79fc88968d7e/.github/workflows/ci-build.yaml#L87)

hack/patch-nix-hash.sh
isubasinghe: Yeah we do actually care to differentiate the error, only a certain error msg contains the correct nix hash. 
We force the correct error output via the first sed command.

.github/workflows/ci-build.yaml
isubasinghe: Ah okay thanks

hack/patch-nix-hash.sh
isubasinghe: I've reduced the if check, and forced this ordering by always doing the first sed.

.github/workflows/ci-build.yaml
```diff
@@ -102,6 +102,7 @@ jobs:
               - .mlc_config.json
               - .spelling
               - mkdocs.yml
+              - hack/patch-nix-hash.sh
```
agilgur5, line 105: this list is categorized by comments -- this is currently placed under the docs scripts, which wouldn't be correct.

I'd move it up to after `.golangci.yml` and add its own comment

agilgur5 (commented):
mostly nits, although the "check if Nix is installed before running" comment below is pretty important

This is back to draft now, but the [commit prefix change](https://github.com/argoproj/argo-workflows/pull/12520#issuecomment-1920266490) and the [dependabot automation](https://github.com/argoproj/argo-workflows/pull/12520#pullrequestreview-1854934639) is also still needed.

hack/patch-nix-hash.sh
agilgur5: > Yeah we do actually care to differentiate the error, only a certain error msg contains the correct nix hash.

We can just print a nicer error ourselves though rather than relying on `sed`'s error, which is not as user-friendly / specific to the script. Unless I'm misunderstanding something?

The current version seems to also unconditionally print the change even if nothing was changed?

hack/patch-nix-hash.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+#This script is a hack to automatically update the Nix hash.
+# This is because dependabot will automatically 
+# update go dependencies.
```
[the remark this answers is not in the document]
isubasinghe: Changed

hack/patch-nix-hash.sh
isubasinghe: Yeah we aren't relying on sed's error. Sed is used to force an ordering which gets the correct error output from the nix build command. 

Yes the current version unconditionally changes the hash, this is the only way to reduce the number of "nix build" commands issued. We can't just store it unfortunately.

agilgur5: OH you mean you're intentionally making the `nix build` command fail in order to pull the `vendorHash` from its output??
Oh that is wacky I did not expect that.

> Yes the current version unconditionally changes the hash

I actually didn't mean the hash, I meant the print / `echo`. It always says "Changed" even if it didn't change, which is a bit confusing.

Are you not able to:
1. `grep` current hash
1. `nix build` 
1. `grep` new hash
1. Check if it changed, if so, `echo`

Or is there a reason you can't?

.github/workflows/ci-build.yaml
```diff
@@ -377,12 +379,15 @@ jobs:
     timeout-minutes: 15 # must be strictly greater than the timeout in .golangci.yml
     env:
       GOPATH: /home/runner/go
+      USE_NIX: true
```
agilgur5, line 382: mmm this changes if/how/when things get installed, I don't think it should be on during CI.

In the `Makefile` you can just check if Nix exists before running `./hack/patch-nix-hash.sh`. Like check if the `nix` command exists, e.g. `command -v nix`. Much simpler check 😅

.github/workflows/ci-build.yaml
```diff
@@ -377,12 +379,15 @@ jobs:
...
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version: "1.21"
           cache: true
+      - uses: cachix/install-nix-action@6004951b182f8860210c8d6f0d808ec5b1a33d28 #v25 -> nix 2.19.2
```
agilgur5, line 389:
```suggestion
      - uses: cachix/install-nix-action@6004951b182f8860210c8d6f0d808ec5b1a33d28 # v25 -> nix 2.19.2
```

nit: Consistent space after `#`
(I'm also not sure how many different permutations of the comment dependabot handles)

Joibel:
@isubasinghe - do you want to keep this PR?

isubasinghe:
No, I have something better now.

PR #12522, opened 2024-01-14
build(docs): don't install NPM global deps if they're already installed

agilgur5:
### Motivation

Only install NPM global deps if they're not already installed
- previously the `make` commands would unconditionally install these global deps, which was unnecessary & inefficient the vast majority of the time
  - it would reinstall these every single time I ran these commands

### Modifications

- use `npm list -g` to check if they're installed first
  - this exits with error code 1 if it doesn't exist

- note: that was the purpose of the `/usr/local/bin/` part in the `Makefile`, but this is not universally installed in the same place
  - can run `npm prefix -g` and concat `/bin/` to that to get the path to your global installs
    - on my Mac with versioned node (using [`rtx`](https://github.com/jdx/rtx)), the path is: `/Users/<my-user>/.local/share/rtx/installs/node/20.6.0/bin/`
    - in the `devcontainer` with versioned node (uses `nvm`), the path is: `/usr/local/share/nvm/versions/node/v20.5.1/bin`

### Verification

1. Manually ran the commands with the dependency, confirmed it skipped the install
1. Manually ran the commands without the dependency, confirmed it performed the install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -678,7 +678,7 @@ docs/cli/argo.md: $(CLI_PKGS) go.sum server/static/files.go hack/cli/main.go
 /usr/local/bin/mdspell:
 # update this in Nix when upgrading it here
 ifneq ($(USE_NIX), true)
-	npm i -g markdown-spellcheck@1.3.1
+	npm list -g markdown-spellcheck@1.3.1 > /dev/null || npm i -g markdown-spellcheck@1.3.1
 endif
 
 .PHONY: docs-spellcheck
@@ -689,7 +689,7 @@ docs-spellcheck: /usr/local/bin/mdspell
 /usr/local/bin/markdown-link-check:
 # update this in Nix when upgrading it here
 ifneq ($(USE_NIX), true)
-	npm i -g markdown-link-check@3.11.1
+	npm list -g markdown-link-check@3.11.1 > /dev/null || npm i -g markdown-link-check@3.11.1
 endif
 
 .PHONY: docs-linkcheck
@@ -700,7 +700,7 @@ docs-linkcheck: /usr/local/bin/markdown-link-check
 /usr/local/bin/markdownlint:
 # update this in Nix when upgrading it here
 ifneq ($(USE_NIX), true)
-	npm i -g markdownlint-cli@0.33.0
+	npm list -g markdownlint-cli@0.33.0 > /dev/null || npm i -g markdownlint-cli@0.33.0
 endif
 
 


agilgur5:
@isubasinghe this might make the `USE_NIX` check no longer necessary? If Nix is installing as a global dep, then this should work for it as well. If so, I can remove the checks from these deps

isubasinghe (approved):
This seems sensible to me

isubasinghe:
@agilgur5 The dependency isn't really installed in Nix via a `npm install -g` equivalent, so `npm list -g` will fail (I tested this out), 
so the Makefile will have to remain as it is. 

I don't think this is necessarily a negative though, it does make the Makefile more consistent at least.

agilgur5:
> @agilgur5 The dependency isn't really installed in Nix via a `npm install -g` equivalent, so `npm list -g` will fail (I tested this out)

Thanks for checking!
 I thought `node2nix` might create a similar environment or that the Nix shell would register it an equivalent way, especially if it puts it on `$PATH` for the `make` commands to work.
But I guess not, here's the line that [modifies `$PATH`](https://github.com/argoproj/argo-workflows/blob/de53da84646f5be31e2c7b7457574ff417126add/dev/nix/node-env.nix#L681). Potentially a limitation of `node2nix` or how it was configured, since for reproducibility it _should_ behave similarly to `npm i -g`

isubasinghe:
> > @agilgur5 The dependency isn't really installed in Nix via a `npm install -g` equivalent, so `npm list -g` will fail (I tested this out)
> 
> Thanks for checking! I thought `node2nix` might create a similar environment or that the Nix shell would register it an equivalent way, especially if it puts it on `$PATH` for the `make` commands to work. But I guess not, here's the line that [modifies `$PATH`](https://github.com/argoproj/argo-workflows/blob/de53da84646f5be31e2c7b7457574ff417126add/dev/nix/node-env.nix#L681). Potentially a limitation of `node2nix` or how it was configured, since for reproducibility it _should_ behave similarly to `npm i -g`

No worries, I don't quite know how it works but I believe it might be due to not assuming a global node installation. I think its a node2nix limitation, really I should replace that tool with raw Nix.

agilgur5:
> No worries, I don't quite know how it works but I believe it might be due to not assuming a global node installation.

Ah that would make sense if so. Ideally these shouldn't be globally installed either -- or any of the build deps for that matter -- but we've still got a fair bit of changes necessary before we can get there.

PR #11358, opened 2023-07-14
docs: Update README.md for home page

terrytangyuan:
https://argoproj.github.io/argo-workflows/ is super outdated and does not contain much information. It should be the same as the root level README. Ideally we should use the root level README but I am not sure if it works  with mkdocs when working outside of `docs` folder: https://github.com/argoproj/argo-workflows/blob/master/mkdocs.yml#L37

PR #12531, opened 2024-01-17
docs: Additional badges

terrytangyuan:
The categorization is to be consistent with Argo CD. Also added release badge and LinkedIn badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,11 @@
 <!-- markdownlint-disable-next-line MD041 -->
-**Code:** 
 [![Security Status](https://github.com/argoproj/argo-workflows/workflows/Snyk/badge.svg)](https://github.com/argoproj/argo-workflows/actions/workflows/snyk.yml?query=branch%3Amain)
 [![OpenSSF Best Practices](https://bestpractices.coreinfrastructure.org/projects/3830/badge)](https://bestpractices.coreinfrastructure.org/projects/3830)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/argoproj/argo-workflows/badge)](https://api.securityscorecards.dev/projects/github.com/argoproj/argo-workflows)
-[![FOSSA License Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fargoproj%2Fargo-workflows.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fargoproj%2Fargo-workflows?ref=badge_shield)<br />
-**Social:**
+[![FOSSA License Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fargoproj%2Fargo-workflows.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fargoproj%2Fargo-workflows?ref=badge_shield)
 [![Slack](https://img.shields.io/badge/slack-argoproj-brightgreen.svg?logo=slack)](https://argoproj.github.io/community/join-slack)
 [![Twitter Follow](https://img.shields.io/twitter/follow/argoproj?style=social)](https://twitter.com/argoproj)
-[![LinkedIn](https://img.shields.io/badge/LinkedIn-argoproj-blue.svg?logo=linkedin)](https://www.linkedin.com/company/argoproj/)<br />
-**Releases:**
+[![LinkedIn](https://img.shields.io/badge/LinkedIn-argoproj-blue.svg?logo=linkedin)](https://www.linkedin.com/company/argoproj/)
 [![Release Version](https://img.shields.io/github/v/release/argoproj/argo-workflows?label=argo-workflows)](https://github.com/argoproj/argo-workflows/releases/latest)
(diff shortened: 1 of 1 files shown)


terrytangyuan:
Is there a way to use the root level README file in the docs website? e.g. https://github.com/argoproj/argo-workflows/blob/main/docs/README.md  is pretty redundant and we don't want to update both places every time.

README.md
```diff
@@ -1,11 +1,18 @@
 <!-- markdownlint-disable-next-line MD041 -->
-[![slack](https://img.shields.io/badge/slack-argoproj-brightgreen.svg?logo=slack)](https://argoproj.github.io/community/join-slack)
+**Releases:**
+[![Release Version](https://img.shields.io/github/v/release/argoproj/argo-workflows?label=argo-workflows)](https://github.com/argoproj/argo-workflows/releases/latest)
+[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/argo-workflows)](https://artifacthub.io/packages/helm/argo/argo-workflows)
+
+**Code:**
```
agilgur5, line 6:
```suggestion
[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/argo-workflows)](https://artifacthub.io/packages/helm/argo/argo-workflows)<br />
**Code:** 
```

This would make the new line in between the sections half the size. I.e. it would be on the next-line instead of double-spaced.

README.md
```diff
@@ -1,11 +1,18 @@
...
+[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/argo-workflows)](https://artifacthub.io/packages/helm/argo/argo-workflows)
+
+**Code:** 
 [![Security Status](https://github.com/argoproj/argo-workflows/workflows/Snyk/badge.svg)](https://github.com/argoproj/argo-workflows/actions/workflows/snyk.yml?query=branch%3Amain)
 [![OpenSSF Best Practices](https://bestpractices.coreinfrastructure.org/projects/3830/badge)](https://bestpractices.coreinfrastructure.org/projects/3830)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/argoproj/argo-workflows/badge)](https://api.securityscorecards.dev/projects/github.com/argoproj/argo-workflows)
 [![FOSSA License Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fargoproj%2Fargo-workflows.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fargoproj%2Fargo-workflows?ref=badge_shield)
-[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/argo-workflows)](https://artifacthub.io/packages/helm/argo/argo-workflows)
+
+**Social:**
```
agilgur5, line 12: same as above, although GitHub isn't letting make me a suggestion on "removed code"

README.md
```diff
@@ -1,11 +1,18 @@
...
 [![Twitter Follow](https://img.shields.io/twitter/follow/argoproj?style=social)](https://twitter.com/argoproj)
+[![Slack](https://img.shields.io/badge/slack-argoproj-brightgreen.svg?logo=slack)](https://argoproj.github.io/community/join-slack)
```
agilgur5, line 14:
```suggestion
[![Slack](https://img.shields.io/badge/slack-argoproj-brightgreen.svg?logo=slack)](https://argoproj.github.io/community/join-slack)
[![Twitter Follow](https://img.shields.io/twitter/follow/argoproj?style=social)](https://twitter.com/argoproj)
```

Should probably place Slack first, since that is our primary communication channel

agilgur5 (commented):
I actually don't like the way Argo CD has badges with a lot of spacing and explicit categories, because that kind of defeats the purpose of badges as being small, aesthetically pleasing, in-line links.

Screenshot 2024-01-19 at 2 53 35 PM

If we wanted to do something similar, I would at least remove the double-spaced new-lines as suggested below, but I still think this is not quite how badges are intended to be used.

It also violates `markdownlint` & markdown conventions as well

README.md
```diff
@@ -1,11 +1,18 @@
 <!-- markdownlint-disable-next-line MD041 -->
```
agilgur5, line 1: this ignore is here because the first line is supposed to be an `H1`. If there's a line of badges first, I think that's ok, but putting several lines of text + badges would be an anti-pattern and would fail `markdownlint` multiple times. By convention, any text should be inside of / below headings

agilgur5:
> Is there a way to use the root level README file in the docs website? e.g. https://github.com/argoproj/argo-workflows/blob/main/docs/README.md is pretty redundant and we don't want to update both places every time.

If we want them to be equivalent, yes, I'm pretty sure there's a few ways to do it. They weren't the same until your #11358 I think, so that is a more recent change. I can handle that

terrytangyuan:
Addressed comments

terrytangyuan:
Agreed that this looks ugly. I re-purposed the PR to just add the additional badges.

agilgur5 (approved):

agilgur5:
> If we want them to be equivalent, yes, I'm pretty sure there's a few ways to do it. They weren't the same until your #11358 I think, so that is a more recent change. I can handle that

I added a script for this in #12752. It's not quite a straight copy+paste because of needing to change relative and absolute links.

PR #12459, opened 2024-01-04
chore: bump github.com/Azure/azure-sdk-for-go/sdk/storage/azblob

roelarents:
Bumps [github.com/Azure/azure-sdk-for-go/sdk/storage/azblob](https://github.com/Azure/azure-sdk-for-go) from 1.0.0 to 1.2.1.
- [Release notes](https://github.com/Azure/azure-sdk-for-go/releases)
- [Changelog](https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/release.md)
- [Commits](https://github.com/Azure/azure-sdk-for-go/compare/sdk/storage/azblob/v1.0.0...sdk/storage/azblob/v1.2.1)

Fixes: n/a. I didn't create an issue.

### Motivation

Update azure-sdk-for-go/sdk/storage/azblob to the latest version.

Dependabot would normally do this, but it has been explicitly [disabled](https://github.com/argoproj/argo-workflows/pull/10509#issuecomment-1427081626) for this dep by @alexec. Perhaps it could/should be enabled again, since @brianloss manually updated to the 1.0 version.

After updating argo from v3.4.8 to v3.5.2 we ran into [this issue](https://github.com/Azure/azure-sdk-for-go/issues/20735) in the azure sdk with uploading artifacts with a particular size in which case the upload part size is not correctly calculated.

### Modifications

A version update in go.mod and .sum

Also tagged on the expansion of the debug message for `executeTemplate`, to help investigating for this (unrelated) [issue with the recently introduced recursion limit](https://github.com/argoproj/argo-workflows/issues/12162)

### Verification

Tested by:
* updating this dep in 3.5.2
* running the unit tests
* building the argoexec image locally
* running the same workflow that failed on it before

PR #12532, opened 2024-01-17
fix: disable concurrency when downloading azure files

roelarents:
workaround for https://github.com/Azure/azure-sdk-for-go/issues/22156

### Motivation

Recently [bumped](https://github.com/argoproj/argo-workflows/pull/12459) the azure sdk for go for blob storage. But it has an [issue](https://github.com/Azure/azure-sdk-for-go/issues/22156) with concurrent downloads.

### Modifications

Set concurrency when downloading artifacts/files/blobs from azure blob storage to 1 (instead of 5)

### Verification

Tested in production with a patch on 3.5.4. We [suffered](https://github.com/Azure/azure-sdk-for-go/issues/22156#issuecomment-1883653523) from this issue before (hanging workflow steps) and now not anymore.

### DIFF
--- a/workflow/artifacts/azure/azure.go
+++ b/workflow/artifacts/azure/azure.go
@@ -170,8 +170,8 @@ func DownloadFile(containerClient *container.Client, blobName, path string) erro
 		}
 	}()
 
-	o := blob.DownloadFileOptions{Concurrency: 1} // workaround https://github.com/Azure/azure-sdk-for-go/issues/22156
-	_, err = blobClient.DownloadFile(context.TODO(), outFile, &o)
+	opts := blob.DownloadFileOptions{Concurrency: 1} // workaround for https://github.com/Azure/azure-sdk-for-go/issues/22156
+	_, err = blobClient.DownloadFile(context.TODO(), outFile, &opts)
 	return err
 }
 
--- a/workflow/artifacts/azure/azure.go
+++ b/workflow/artifacts/azure/azure.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


agilgur5:
This sounds like it's going to be [fixed soon upstream](https://github.com/Azure/azure-sdk-for-go/issues/22156#issuecomment-1893121130)? So maybe we should just wait for the fix and update the dep rather than adding and removing a workaround. I would think the Azure SDK would have a release before Argo as well

roelarents:
True, if they fix it before the next argo release then this is obsolete. If not, I do suggest this temporary workaround. Because there is [another issue](https://github.com/Azure/azure-sdk-for-go/issues/20735) (that we also encountered in production) in the azure blob sdk since 1.0 (which is in argo 3.5). That has been fixed upstream already. I [bumped](https://github.com/argoproj/argo-workflows/pull/12459) the sdk to 1.2.1 for it. But that introduced this issue. (Probably/possibly there is a version in between with the fix for the one and not yet introducing the next. But I think using the latest version is easier.)

tanyasethi-msft:
This fix constraints the concurrency to 1, which means parallel downloads not supported. 
We will be fixing the issue in our next release, Please use [azblob v1.1.0](https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fstorage%2Fazblob%2Fv1.1.0) for temporary mitigation.

roelarents:
yep, they fixed it. this can be closed.

agilgur5:
Looks like it was fixed in https://github.com/Azure/azure-sdk-for-go/pull/22334 which [was released in](https://github.com/Azure/azure-sdk-for-go/issues/22156#issuecomment-1941895878) `azblob` v1.3.0

PR #12537, opened 2024-01-17
fix: make sure taskresult completed when mark node succeed when it has outputs

shuangkun:
When my cluster has lots of workflows, I meet some errors.
```
="Mark error node" error="failed to evaluate expression: cannot fetch steps-init-artifact from <nil> (1:6)\n | steps['init-artifact'].outputs.parameters['workflow_artifact_key']\n | .....^" namespace=argo nodeName="workflow-bhr9k[3].energy(0:0)[1].energy-steps(0:0)[3].comp-binding-energy-steps(0:0)[15]" workflow=workflow-bhr9k
```
When the number of workflows is not large, there is no such error. 

My workflow has lots of template like this, the next step refer the output of the previous step. Like hello2a refer hello1 in parameter steps['hello1'].outputs.parameters['workflow_artifact_key'].
```
kind: Workflow
metadata:
  generateName: steps-
spec:
  entrypoint: hello-hello-hello
  arguments:
    parameters:
    - name: message1
      value: hello world
    - name: message2
      value: foobar
  templates:
  - name: hello-hello-hello
    steps:
        continueOn: {}
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "hello1"
          - name: workflow_artifact_key
            value: "{{ workflow.parameters.message2}}"
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "{{=steps['hello1'].outputs.parameters['workflow_artifact_key']}}"

  - name: whalesay
    metadata:
      annotations:
        k8s.aliyun.com/eci-spot-strategy: "SpotAsPriceGo"
    inputs:
      parameters:
      - name: message
    outputs:
      parameters:
      - name: workflow_artifact_key
        value: '{{workflow.name}}'
    script:
      image: python:alpine3.6
      command: [python]
      source: |
        import random
        i = random.randint(1, 100)
        print(i)
```

When I search the logs. I find the time of preStep(hello1)‘s “node changed” to succeed are earlier than "task-result changed". And
this cause the hello2a's  evaluate expression error. So I want to make sure taskresult completed when mark node succeed when it has outputs.


workflow/controller/operator.go
```diff
@@ -312,6 +312,12 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 			woc.failSuspendedAndPendingNodesAfterDeadlineOrShutdown()
 		}
 
+		if err != nil && strings.Contains(err.Error(), "need reconcil workflowtaskresults") {
+			woc.log.WithError(err).WithField("workflow", woc.wf.ObjectMeta.Name).Error("need reconcil workflowtaskresults")
+			woc.requeue()
+			return
+		}
```
juliev0, line 319: I don't think we should ever rely on the specific wording of an error - if needed, we could have specific error types that can be checked or something

juliev0:
So, it sounds like previously the execution of a Workflow was allowed to continue even if the previous Step's Outputs weren't reconciled? Are you essentially preventing the next Step from running yet in that case? 

Looking at where `podReconciliation()` is called, are you basically hitting [this](https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/operator.go#L316) error line?

shuangkun:
> So, it sounds like previously the execution of a Workflow was allowed to continue even if the previous Step's Outputs weren't reconciled? Are you essentially preventing the next Step from running yet in that case? 
> 
> 
> 
> 

Yes, I want to prevent the next step from running.

shuangkun:
Yes, I think this was the case before, although under normal circumstances output will be processed before pod status normally, because this resource is indeed created earlier. But on a large scale, these two processing orders may be caused by high pressure Events arrive at APIserver in different order

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1971,6 +1971,14 @@ func (ws *WorkflowStatus) TaskResultsInProgress() bool {
 	return false
 }
 
+func (ws *WorkflowStatus) IsTaskResultInCompleted(name string) bool {
```
juliev0, line 1974: assuming we keep this method, let's rename it to `IsTaskResultIncomplete()` (current name sounds a bit like the opposite)

workflow/controller/operator.go
juliev0: also, I see you're logging an error - but it's not actually an error, right?

workflow/controller/operator.go
```diff
@@ -1160,6 +1179,12 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 
 	wg.Wait()
 
+	// If true, it means there are some nodes which have outputs want to be mark succeed, but the node's taskresults didn't completed.
+	// We should make sure it the taskresults processing is complete as it will be possible to reference it in the next step.
+	if needReconcilTaskResult {
+		return fmt.Errorf("need reconcil workflowtaskresults")
+	}
```
juliev0, line 1186: if this isn't an error, we should return it as something else, not an error

pkg/apis/workflow/v1alpha1/workflow_types.go
shuangkun: ok，I will rename it.

juliev0: Actually, to match the name above, it should probably be `IsTaskResultInProgress()`

juliev0:
Is it possible to see if this worked on an older version of code? I'm curious if something broke this. It seems like core functionality.

workflow/controller/operator.go
```diff
@@ -1099,6 +1105,7 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 	seenPodLock := &sync.Mutex{}
 	wfNodesLock := &sync.RWMutex{}
 	podRunningCondition := wfv1.Condition{Type: wfv1.ConditionTypePodRunning, Status: metav1.ConditionFalse}
+	needReconcilTaskResult := false
```
juliev0, line 1108: nit: can you fix the spelling of all "reconcil" to "reconcile"?

juliev0:
> Yes, I think this was the case before, although under normal circumstances output will be processed before pod status normally, because this resource is indeed created earlier. But on a large scale, these two processing orders may be caused by high pressure Events arrive at APIserver in different order

I see. So, maybe this is a good enough answer to my request that you test on an older version - perhaps this case is just an unusual one? I am kind of curious if other people have logged similar bugs.

workflow/controller/operator.go
shuangkun: OK，I will change it.

workflow/controller/operator.go
shuangkun: OK, I will,thanks!

shuangkun:
> Is it possible to see if this worked on some older versions of code? I'm curious if something broke this. It seems like core functionality.

I think this may be related to the introduction of taskresult resources since 3.4.   Maybe it is hard to support old， because there is a lack of records recording whether the taskresult was processed（Originally I needed to add this record, but found that it was included in the latest version.)

shuangkun:
> > Yes, I think this was the case before, although under normal circumstances output will be processed before pod status normally, because this resource is indeed created earlier. But on a large scale, these two processing orders may be caused by high pressure Events arrive at APIserver in different order
> 
> I see. So, maybe this is a good enough answer to my request that you test on an older version - perhaps this case is just an unusual one? I am kind of curious if other people have logged similar bugs.



> > Yes, I think this was the case before, although under normal circumstances output will be processed before pod status normally, because this resource is indeed created earlier. But on a large scale, these two processing orders may be caused by high pressure Events arrive at APIserver in different order
> 
> I see. So, maybe this is a good enough answer to my request that you test on an older version - perhaps this case is just an unusual one? I am kind of curious if other people have logged similar bugs.

Yes，

> failed to evaluate expression



> > Yes, I think this was the case before, although under normal circumstances output will be processed before pod status normally, because this resource is indeed created earlier. But on a large scale, these two processing orders may be caused by high pressure Events arrive at APIserver in different order
> 
> I see. So, maybe this is a good enough answer to my request that you test on an older version - perhaps this case is just an unusual one? I am kind of curious if other people have logged similar bugs.

Yes, I tested it on version 3.4.12 for few weeks. Looks well. There will be no "failed to evaluate expression" error like before.

juliev0:
@Garett-MacGowan do you want to look at this too?

Garett-MacGowan:
> @Garett-MacGowan do you want to look at this too?

I just skimmed it. I can take a proper look after I 😴. In general, if we're proceeding to next steps before outputs are reconciled, it seems important that we add the wait behavior. As you said, it seems like core functionality, so I'm surprised if it's not already accounted for.

I'm wondering if this can be tested.

workflow/controller/operator.go
```diff
@@ -1117,6 +1124,18 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
...
 		if err == nil {
 			if newState := woc.assessNodeStatus(pod, node); newState != nil {
+				if newState.Succeeded() {
+					tmpl, err := woc.GetNodeTemplate(node)
+					if err != nil {
+						return
+					}
+					// Check whether the node has output and whether its taskresult is in an incompleted state.
+					if tmpl.HasOutputs() && woc.wf.Status.IsTaskResultInCompleted(node.ID) && woc.wf.Status.IsTaskResultInCompleted(pod.Name) {
+						woc.log.WithFields(log.Fields{"nodeID": newState.ID}).WithError(err).Error("Taskresult of the node not yet completed")
```
Garett-MacGowan, line 1134: Same comment along the lines of what @juliev0 was saying, I don't think this is an error. We just need to flag `needReconcileTaskResult`. Could maybe just log it normally if you want the log?

workflow/controller/controller_test.go
```diff
@@ -510,6 +510,13 @@ func makePodsPhase(ctx context.Context, woc *wfOperationCtx, phase apiv1.PodPhas
 			if err != nil {
 				panic(err)
 			}
+			if phase == apiv1.PodSucceeded {
+				if woc.wf.Status.TaskResultsCompletionStatus == nil {
```
Garett-MacGowan, line 514: `workflow_types.go` has `MarkTaskResultIncomplete` and `markTaskResultComplete` convenience methods that handle initialization of `TaskResultsCompletionStatus`.

workflow/controller/operator.go
juliev0: right, probably a Debug line

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1971,6 +1971,14 @@ func (ws *WorkflowStatus) TaskResultsInProgress() bool {
...
+	value, found := ws.TaskResultsCompletionStatus[name]
```
Garett-MacGowan, line 1975: We should probably fix + update the comment for `ws.TaskResultsCompletionStatus` in `workflow_types.go`.

```
// TaskResultsCompletionStatus tracks task result completion status (mapped by pod name). Used to prevent premature archiving and garbage collection.
```

Maybe we could change it to the following?
```
// TaskResultsCompletionStatus tracks task result completion status (mapped by node ID). Used to prevent premature archiving, garbage collection, and workflow progression.
```

cmd/argoexec/commands/resource.go
```diff
@@ -34,6 +34,7 @@ func NewResourceCommand() *cobra.Command {
 func execResource(ctx context.Context, action string) error {
 	wfExecutor := initExecutor()
 	defer wfExecutor.HandleError(ctx)
+	defer wfExecutor.FinalizeOutput(ctx)
```
juliev0, line 37: good catch - I just saw in the code that there's no wait container for "argoexec resource", huh?

workflow/controller/operator.go
juliev0: By the way, do we need to call it for both `woc.wf.Status.IsTaskResultInCompleted(node.ID) && woc.wf.Status.IsTaskResultInCompleted(pod.Name)`?

Garett-MacGowan: Oh yeah, I was thinking this but had to step away and forgot to ask. I think it should just be `tmpl.HasOutputs() && woc.wf.Status.IsTaskResultInCompleted(node.ID)`

Garett-MacGowan: Maybe confusion from the comment [here](https://github.com/argoproj/argo-workflows/pull/12537#discussion_r1464003132)

juliev0: seeing that comment about the comment made me think of it :)

workflow/controller/controller_test.go
shuangkun: Yes，I can change it.

cmd/argoexec/commands/resource.go
shuangkun: Yes, it didn't have wait container.

workflow/controller/operator.go
shuangkun: > By the way, do we need to call it for both `woc.wf.Status.IsTaskResultInCompleted(node.ID) && woc.wf.Status.IsTaskResultInCompleted(pod.Name)`?

Yes, I thought about this problem at first. But there is a problem. If the outputs are in pod annotations or in taskresult, the key values ​​are different. Maybe we can unify to podName or NodeId. I think nodeId is better, how about you?

May be I can add a func named pod.GetNodeId()

```
	if x, ok := pod.Annotations[common.AnnotationKeyReportOutputsCompleted]; ok {
		woc.log.Warn("workflow uses legacy/insecure pod patch, see https://argo-workflows.readthedocs.io/en/latest/workflow-rbac/")
		resultName := pod.GetName()
		if x == "true" {
		} else {
			woc.wf.Status.MarkTaskResultIncomplete(resultName)
		}
	}

File: pkg/apis/workflow/v1alpha1/workflow_types.go (hunk 7 above)
Comment (shuangkun) [reply pkg/apis/workflow/v1alpha1/workflow_types.go:1975]: Yes, may we can unified to nodeId.

File: cmd/argoexec/commands/resource.go (hunk 8 above)
Comment (shuangkun) [reply cmd/argoexec/commands/resource.go:37]: We need to mark the tasksult to true for these templates that do not have a wait container but have outputs. https://github.com/argoproj/argo-workflows/pull/12544

Comment (juliev0) [reply cmd/argoexec/commands/resource.go:37]: thanks. yeah, I added a couple of comments over there that would be related here as well

File: workflow/controller/operator.go (hunk 1 above)
Comment (juliev0) [reply workflow/controller/operator.go:319]: just wanted to make sure you saw the second comment above as well - if it's not an error, let's just log it as a debug line

File: workflow/controller/operator.go [hunk 9]
```diff
@@ -1160,6 +1179,12 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 
 	wg.Wait()
 
+	// If true, it means there are some nodes which have outputs want to be mark succeed, but the node's taskresults didn't completed.
```
juliev0, line 1182:
```suggestion
	// If true, it means there are some nodes which have outputs we wantedto be mark succeed, but the node's taskresults didn't completed.
```

workflow/controller/operator.go
```diff
@@ -1160,6 +1179,12 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
...
+	// We should make sure it the taskresults processing is complete as it will be possible to reference it in the next step.
```
juliev0, line 1183:
```suggestion
	// We should make sure the taskresults processing is complete as it will be possible to reference it in the next step.
```

workflow/controller/operator.go
```diff
@@ -1117,6 +1124,18 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 		node, err := woc.wf.Status.Nodes.Get(nodeID)
 		if err == nil {
 			if newState := woc.assessNodeStatus(pod, node); newState != nil {
+				if newState.Succeeded() {
+					tmpl, err := woc.GetNodeTemplate(node)
+					if err != nil {
+						return
```
juliev0, line 1130: would this be an error that we should log?

workflow/controller/operator.go
```diff
@@ -1117,6 +1124,18 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
...
+					}
+					// Check whether the node has output and whether its taskresult is in an incompleted state.
+					if tmpl.HasOutputs() && woc.wf.Status.IsTaskResultIncomplete(node.ID) {
```
juliev0, line 1133: I'm wondering if it's necessary to include `tmpl.HasOutputs()` in the if statement?

workflow/controller/operator.go
Garett-MacGowan: Yes, just unify to Node ID.

workflow/controller/operator.go
shuangkun: ok,I don't need a error. But I want to make it a info. This information is not common. I can search for this information to see if my cluster is facing this "need reconcile workflowtaskresults" problem. Now I do the same，help me determine the pressure on the cluster at this time.
image

workflow/controller/operator.go
shuangkun: OK

workflow/controller/operator.go
shuangkun: I think it is necessary. We can't judge a node IsTaskResultIncomplete only by "woc.wf.Status.IsTaskResultIncomplete", If there no key in TaskResultsCompletionStatus, Maybe the taskresult has not been received yet, or maybe the taskresult is not needed at all.

workflow/controller/operator.go
shuangkun: OK

workflow/controller/operator.go
Garett-MacGowan: maybe im mossong something, but if `HasOutputs()` evaluates to true, it could still be possible that outputs haven't been initialized and therefore no key exists in `TaskResultCompletionStatus`. Does this actually give us more protection? This is only here for when task results are incomplete, so it depends on keys being in `TaskResultsCompletionStatus` anyway.

workflow/controller/operator.go
juliev0: okay, I'm good with "Info" level

workflow/controller/operator.go
juliev0: As far as I understand, every TaskResult that's received will be marked [incomplete](https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/taskresult.go#L72) and then marked [complete](https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/taskresult.go#L69), regardless of whether there are outputs defined, right? And it looks like `IsTaskResultIncomplete()` will return true if there's no entry in the map for the node yet. So, it seems like if we could just test for that, no?

shuangkun: Yes, I overlooked something. Some tasks do not define output. We also created taskresult. Can we add some restrictions? For tasks without output, cancel
```
wfExecutor.InitializeOutput(bgCtx)
defer wfExecutor.HandleError(bgCtx), 
```
because after all, this increases the pressure on the apiserver.

juliev0: So, 2 fewer API calls per Pod that doesn't have "Outputs"? I can see that. I do, on the other hand, kind of like the simplicity of just doing it across the board. And also, that label essentially means that all outputs of any kind are done, including [Logs](https://github.com/argoproj/argo-workflows/blob/main/cmd/argoexec/commands/wait.go#L66).

juliev0: Actually, we probably need to keep it the way it is - `checkReconciliationComplete()` will verify that all TaskResults of any kind are completely processed before allowing Workflow to be marked completed ([here](https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/operator.go#L2319))

shuangkun: > So, 2 fewer API calls per Pod that doesn't have "Outputs"? I can see that. I do, on the other hand, kind of like the simplicity of just doing it across the board and not having to have special if statements on the Controller side. And also, that label essentially means that all outputs of any kind are done, including [Logs](https://github.com/argoproj/argo-workflows/blob/main/cmd/argoexec/commands/wait.go#L66).

OK, If each node has taskresult. I  will change it.

workflow/controller/operator.go
```diff
@@ -307,11 +307,17 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 		woc.wf.Status.EstimatedDuration = woc.estimateWorkflowDuration()
 	} else {
 		woc.workflowDeadline = woc.getWorkflowDeadline()
-		err = woc.podReconciliation(ctx)
+		err, needReconcileTaskResult := woc.podReconciliation(ctx)
```
juliev0, line 310: Nit: can we make the boolean value that's returned generically represent whether pod reconciliation finished rather than being specific to task results? I think that will improve readability. And flip it so true means "pod reconciliation finished". So, this could basically read something like: `err, podReconciliationCompleted := woc.podReconciliation(ctx)` (let me know if my naming is bad)

juliev0: Note: I do see that this means your log line directly below would have to be more generic as well, but to me, that seems okay given that you have a more detailed log line at Info level in the `podReconciliation()` function

shuangkun: > Nit: can we make the boolean value that's returned generically represent whether pod reconciliation finished rather than being specific to task results? I think that will improve readability. And flip it so true means "pod reconciliation finished". So, this could basically read something like: `err, podReconciliationCompleted := woc.podReconciliation(ctx)` (let me know if my naming is bad)

OK，i think so ， i will change it

juliev0: Really appreciate the iterations!

workflow/controller/operator.go
```diff
@@ -1117,6 +1124,12 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 		node, err := woc.wf.Status.Nodes.Get(nodeID)
 		if err == nil {
 			if newState := woc.assessNodeStatus(pod, node); newState != nil {
+				// Check whether its taskresult is in an incompleted state.
+				if newState.Succeeded() && woc.wf.Status.IsTaskResultIncomplete(node.ID) {
+					woc.log.WithFields(log.Fields{"nodeID": newState.ID}).Debug("Taskresult of the node not yet completed")
+					podReconciliationCompleted = true
```
juliev0, line 1130: Unfortunately, you will need to flip all of the true/false here and where it's being checked now - this should represent `podReconciliationCompleted = false`, right?

juliev0: Actually, here I think it would be fine to just name it back to `taskResultIncomplete` - maybe I can try do it inline everywhere and you can see what I mean. (Sorry for the trouble!)

workflow/controller/operator.go
```diff
@@ -1090,15 +1096,16 @@ func (woc *wfOperationCtx) processNodeRetries(node *wfv1.NodeStatus, retryStrate
 // pods and update the node state before continuing the evaluation of the workflow.
 // Records all pods which were observed completed, which will be labeled completed=true
 // after successful persist of the workflow.
-func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
+func (woc *wfOperationCtx) podReconciliation(ctx context.Context) (error, bool) {
```
juliev0, line 1099:
```suggestion
// returns whether pod reconciliation successfully completed
func (woc *wfOperationCtx) podReconciliation(ctx context.Context) (error, bool) {
```

workflow/controller/operator.go
```diff
@@ -1090,15 +1096,16 @@ func (woc *wfOperationCtx) processNodeRetries(node *wfv1.NodeStatus, retryStrate
...
 	podList, err := woc.getAllWorkflowPods()
 	if err != nil {
-		return err
+		return err, false
 	}
 	seenPods := make(map[string]*apiv1.Pod)
 	seenPodLock := &sync.Mutex{}
 	wfNodesLock := &sync.RWMutex{}
 	podRunningCondition := wfv1.Condition{Type: wfv1.ConditionTypePodRunning, Status: metav1.ConditionFalse}
+	podReconciliationCompleted := false
```
juliev0, line 1108:
```suggestion
	taskResultIncomplete := false
```

workflow/controller/operator.go
juliev0, line 1130:
```suggestion
					taskResultIncomplete = true
```

workflow/controller/operator.go
```diff
@@ -1160,6 +1173,12 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 
 	wg.Wait()
 
+	// If true, it means there are some nodes which have outputs we wanted to be marked succeed, but the node's taskresults didn't completed.
+	// We should make sure the taskresults processing is complete as it will be possible to reference it in the next step.
+	if podReconciliationCompleted {
+		return nil, podReconciliationCompleted
+	}
```
juliev0, line 1180:
```suggestion
	if taskResultIncomplete {
		return nil, false
	}
```

workflow/controller/operator.go
```diff
@@ -1201,7 +1220,7 @@ func (woc *wfOperationCtx) podReconciliation(ctx context.Context) error {
 			woc.markNodePhase(node.Name, wfv1.NodeError, "pod deleted")
 		}
 	}
-	return nil
+	return nil, podReconciliationCompleted
```
juliev0, line 1223:
```suggestion
	return nil, !taskResultIncomplete
```

workflow/controller/operator.go
```diff
@@ -307,11 +307,17 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
+		err, podReconciliationCompleted := woc.podReconciliation(ctx)
 		if err == nil {
 			woc.failSuspendedAndPendingNodesAfterDeadlineOrShutdown()
 		}
 
+		if podReconciliationCompleted {
+			woc.log.WithField("workflow", woc.wf.ObjectMeta.Name).Info("need reconcile workflowtaskresults")
+			woc.requeue()
+			return
+		}
```
juliev0, line 319:
```suggestion
		if !podReconciliationCompleted {
			woc.log.WithField("workflow", woc.wf.ObjectMeta.Name).Info("pod reconciliation didn't complete, will retry")
			woc.requeue()
			return
		}
```

workflow/controller/operator.go
juliev0: Okay, I just added inline suggestions everywhere - what do you think? Does it make sense? Hopefully, I don't have any errors.

shuangkun: > Okay, I just added inline suggestions everywhere - what do you think? Does it make sense? Hopefully, I don't have any errors.

Yes, you are right! I have accept the suggestions.

workflow/controller/operator.go
```diff
@@ -307,11 +307,17 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
+		err, podReconciliationCompleted := woc.podReconciliation(ctx)
 		if err == nil {
 			woc.failSuspendedAndPendingNodesAfterDeadlineOrShutdown()
 		}
 
+		if !podReconciliationCompleted {
+			woc.log.WithField("workflow", woc.wf.ObjectMeta.Name).Info("pod reconciliation didn't complete, will retry")
+			woc.requeue()
+			return
+		}
```
juliev0, line 319: Sorry, I just realized that we probably need to move the `if err != nil` clause above the `if !podReconciliationCompleted {`, since we can return `err, false`

juliev0: hopefully after that we should be good, thank you for the iterations!

shuangkun: Yes, I changed it.

juliev0 (approved):

PR #12544, opened 2024-01-18
fix: Mark resource && data template report-outputs-completed true

shuangkun:
I run two example(use latest image):
 https://github.com/argoproj/argo-workflows/blob/main/test/e2e/testdata/resource-templates/outputs.yaml 
https://github.com/argoproj/argo-workflows/blob/main/test/e2e/testdata/data-transformation.yaml
I find the workflow succeed but the workflowtaskresult have two is not marked true.
The two is resource && data template. So I want to Mark resource && data report-outputs-completed true.

```
tianshuangkun@B-QGHAMD6M-2124 myworkflow % argo list
NAME                        STATUS      AGE   DURATION   PRIORITY
data-transformation-ff287   Succeeded   2h    2h         0
outputs-fww7h               Succeeded   6h    6h         0
tianshuangkun@B-QGHAMD6M-2124 myworkflow % kubectl get workflowtaskresults -o yaml | grep workflows.argoproj.io/report-outputs-completed
      workflows.argoproj.io/report-outputs-completed: "true"
      workflows.argoproj.io/report-outputs-completed: "true"
      workflows.argoproj.io/report-outputs-completed: "true"
      workflows.argoproj.io/report-outputs-completed: "true"
      workflows.argoproj.io/report-outputs-completed: "true"
      workflows.argoproj.io/report-outputs-completed: "false"
      workflows.argoproj.io/report-outputs-completed: "true"
      workflows.argoproj.io/report-outputs-completed: "false"
tianshuangkun@B-QGHAMD6M-2124 myworkflow %
```

### DIFF
--- a/cmd/argoexec/commands/data.go
+++ b/cmd/argoexec/commands/data.go
@@ -24,12 +24,12 @@ func NewDataCommand() *cobra.Command {
 
 func execData(ctx context.Context) error {
 	wfExecutor := initExecutor()
-	defer wfExecutor.HandleError(ctx)
 
 	// Don't allow cancellation to impact capture of results, parameters, artifacts, or defers.
 	bgCtx := context.Background()
 	// Create a new empty (placeholder) task result with LabelKeyReportOutputsCompleted set to false.
 	wfExecutor.InitializeOutput(bgCtx)
+	defer wfExecutor.HandleError(bgCtx)
 	defer wfExecutor.FinalizeOutput(bgCtx) //Ensures the LabelKeyReportOutputsCompleted is set to true.
 
 	err := wfExecutor.Data(ctx)
--- a/cmd/argoexec/commands/data.go
+++ b/cmd/argoexec/commands/data.go
@@ -12,7 +12,7 @@ func NewDataCommand() *cobra.Command {
 		Use:   "data",
 		Short: "Process data",
 		Run: func(cmd *cobra.Command, args []string) {
-			ctx := cmd.Context()
+			ctx := context.Background()
 			err := execData(ctx)
 			if err != nil {
 				log.Fatalf("%+v", err)
--- a/cmd/argoexec/commands/data.go
+++ b/cmd/argoexec/commands/data.go
@@ -12,7 +12,7 @@ func NewDataCommand() *cobra.Command {
 		Use:   "data",
 		Short: "Process data",
 		Run: func(cmd *cobra.Command, args []string) {
-			ctx := context.Background()
+			ctx := cmd.Context()
 			err := execData(ctx)
 			if err != nil {
 				log.Fatalf("%+v", err)
@@ -26,9 +26,11 @@ func execData(ctx context.Context) error {
 	wfExecutor := initExecutor()
 	defer wfExecutor.HandleError(ctx)
 
+	// Don't allow cancellation to impact capture of results, parameters, artifacts, or defers.
+	bgCtx := context.Background()
 	// Create a new empty (placeholder) task result with LabelKeyReportOutputsCompleted set to false.
-	wfExecutor.InitializeOutput(ctx)
-	defer wfExecutor.FinalizeOutput(ctx) //Ensures the LabelKeyReportOutputsCompleted is set to true.
+	wfExecutor.InitializeOutput(bgCtx)
+	defer wfExecutor.FinalizeOutput(bgCtx) //Ensures the LabelKeyReportOutputsCompleted is set to true.
 
 	err := wfExecutor.Data(ctx)
 	if err != nil {
--- a/cmd/argoexec/commands/data.go
+++ b/cmd/argoexec/commands/data.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"context"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/argoexec/commands/data.go
```diff
@@ -3,18 +3,37 @@ package commands
...
 	return &command
 }
+
+func execData(ctx context.Context) error {
+	wfExecutor := initExecutor()
+	defer wfExecutor.HandleError(ctx)
+
+	// Create a new empty (placeholder) task result with LabelKeyReportOutputsCompleted set to false.
+	wfExecutor.InitializeOutput(ctx)
+	defer wfExecutor.FinalizeOutput(ctx) //Ensures the LabelKeyReportOutputsCompleted is set to true.
```
juliev0, line 31: One thing I'm looking at is the `ctx` in both this and resource.go. If you look at wait.go, there is [this](https://github.com/argoproj/argo-workflows/blob/main/cmd/argoexec/commands/wait.go#L30) comment about not allowing the standard `ctx` object to prevent capture of results, parameters, artifacts, or defers. That is why that code uses `context.Background()`. I believe if we use a context that gets cancelled, then the defer to `FinalizeOutput()` will not do what it needs to do.

juliev0:
Great job catching this ahead of another Issue that found it! And resolving it as well.

cmd/argoexec/commands/resource.go
```diff
@@ -34,11 +34,13 @@ func NewResourceCommand() *cobra.Command {
 func execResource(ctx context.Context, action string) error {
 	wfExecutor := initExecutor()
 	defer wfExecutor.HandleError(ctx)
+	defer wfExecutor.FinalizeOutput(ctx) //Ensures the LabelKeyReportOutputsCompleted is set to true.
```
juliev0, line 37: Was it an oversight not to include `wfExecutor.InitializeOutput()` here?

juliev0:
@Garett-MacGowan will you also review this one?

cmd/argoexec/commands/data.go
Garett-MacGowan: Yeah, I think we need `context.Background()`

Garett-MacGowan:
Once comments are resolved, LGTM.

cmd/argoexec/commands/resource.go
juliev0: yes

cmd/argoexec/commands/data.go
shuangkun: Yes, I agree with you! Thanks

Garett-MacGowan:
Just to be clear, we were already initializing outputs (creating a task result) somewhere in order to get `workflows.argoproj.io/report-outputs-completed: "false"`, right? Where was this happening?

Garett-MacGowan:
> Just to be clear, we were already initializing outputs (creating a task result) somewhere in order to get `workflows.argoproj.io/report-outputs-completed: "false"`, right? Where was this happening? If its the wait container running in the main container, why does the defer not run? Do we actually need to initialize outputs here? Aren't we double initializing?

Nevermind, ReportOutputs does an upsert, so I think its necessary to initialize here.

cmd/argoexec/commands/data.go
```diff
@@ -3,18 +3,39 @@ package commands
...
+				log.Fatalf("%+v", err)
+			}
 		},
 	}
 	return &command
 }
+
+func execData(ctx context.Context) error {
+	wfExecutor := initExecutor()
+	defer wfExecutor.HandleError(ctx)
```
juliev0, line 27: in wait.go, `HandleError()` uses `bgCtx` - do you see a need to do it differently?

juliev0: (I do see that resource.go is using `ctx` but maybe that's just because it was never modified to use `bgctx`?

juliev0: Oh, I guess it looks like `HandleError()` isn't even using the context that's passed in at all. I'm surprised that the linter is okay with that. I kind of wonder if we should remove it from the method parameter, or if not, make it consistently passed in between all in case it's used in the future.

shuangkun: Good discovery,
I have unified to bgCtx.

juliev0:
> Just to be clear, we were already initializing outputs (creating a task result) somewhere in order to get `workflows.argoproj.io/report-outputs-completed: "false"`, right? Where was this happening? If its the wait container running in the main container, why does the defer not run? Do we actually need to initialize outputs here? Aren't we double initializing?
> 
> Edit: I should have looked deeper before commenting. This was a shower thought.

I can relate to the "shower thought" :)

juliev0 (approved):

cmd/argoexec/commands/data.go
```diff
@@ -3,18 +3,39 @@ package commands
...
 func NewDataCommand() *cobra.Command {
 	command := cobra.Command{
 		Use:   "data",
 		Short: "Process data",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			ctx := context.Background()
-			wfExecutor := initExecutor()
-			return wfExecutor.Data(ctx)
+		Run: func(cmd *cobra.Command, args []string) {
+			ctx := cmd.Context()
```
agilgur5, line 15: this specific line change seems to effectively follow-up to https://github.com/argoproj/argo-workflows/pull/10523#discussion_r1104603180

PR #12547, opened 2024-01-19
docs: clarifications to Sustainability Effort process

juliev0:
Fixes #TODO

### DIFF
--- a/community/sustainability_effort.md
+++ b/community/sustainability_effort.md
@@ -41,7 +41,9 @@ For general guidance regarding PR Reviews, see [the Contributing docs](../CONTRI
 
 ### Authoring PRs
 
-Participants can apply a `prioritized-review` label to any PRs they author.
+Members of the repository have permission to apply labels to PRs and can therefore apply a `prioritized-review` label to any PRs they author.
+Those who are not yet members can request to have the label added by posting to the `#argo-wf-sustainability` channel.
+Feel free to give a gentle nudge in that channel if a PR you've authored has almost reached a week without an Assignee.
 
 ## Participants
 


community/sustainability_effort.md
```diff
@@ -41,7 +41,9 @@ For general guidance regarding PR Reviews, see [the Contributing docs](../CONTRI
 
 ### Authoring PRs
 
-Participants can apply a `prioritized-review` label to any PRs they author.
+Members of the repository have permission to apply labels to PRs and can therefore apply a `prioritized-review` label to any PRs they author.
+Those who are not yet members can request to have the label added by posting to the `#argo-wf-sustainability` channel.
+Feel free to give a gentle nudge in that channel if a PR you've authored has almost reached a week without an Assignee.
```
agilgur5, line 46:
```suggestion
Participants can apply a `prioritized-review` label to any PRs they author.
If you do not have permissions to add labels yet, you can request the label in the `#argo-wf-sustainability` channel.

If a PR has reached a week without an Assignee, you can request an assignment in the channel.
```

- simpler and more direct wording, [per k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - "Members" are of the org, not of the repository
  - We don't need to describe permissions and how they work here, the org-level roles document does that
    - We can just tell them what to do if they don't have that permission, which is substantially simpler and more direct.
    
 - add a new line for the "nudge" as that's a separate topic from labeling.
 - make the language for the "nudge" the same as the label for consistency and simplicity -- you can "request" both.

terrytangyuan (approved):

PR #12550, opened 2024-01-19
feat:  Allow users to selectively retry specific failed nodes . Fixes #12543

### LINKED ISSUE #12543: Retrying specific failed node does not work
# Suggested Enhancement

## Use Cases

image

mio4kon:
Fixes #12543 

### Motivation

### Modifications

Removed the restriction that required the simultaneous use of `--node-field-selector` and `--restart-successful`. Now, using `--node-field-selector` alone allows for individual retries of specific failed nodes, instead of retrying all failures.

### Verification
image

_--node-field-selector can be used independently._
`./dist/argo retry fail-24ptx --node-field-selector name=fail-24ptx.BB -v `

_Regressively used in combination._
`./dist/argo retry fail-mz9c4 --restart-successful --node-field-selector name=fail-mz9c4.A`

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -873,7 +873,7 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 
 	onExitNodeName := wf.ObjectMeta.Name + ".onExit"
 	// Get all children of nodes that match filter
-	nodeIDsToReset, err := getNodeIDsToReset(restartSuccessful, nodeFieldSelector, wf.Status.Nodes)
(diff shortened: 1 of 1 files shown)


mio4kon (commented):
Please help me review this PR

shuangkun:
Looks good, Can you add a ut?

mio4kon:
> Looks good, Can you add a ut?

OK

agilgur5:
Replaced by #12553.

For future reference, please use the original PR instead of closing and creating new, duplicate PRs.

PR #11005, opened 2023-04-28
feat: Make retryPolicy saner in the presence of an expression

Joibel:
Most of the explanation for this change is in #10980. When an `expression` is specified in a `retryStrategy` I think it is confusing that the default `retryPolicy` of `OnFailure` still applies. In that event and that only, the default `retryPolicy` becomes `Always` so the `expression` chooses whether or not a retry happens.

If both a `retryPolicy` and `expression`  are provided the previous behavior remains. If no `retryPolicy` and no `expression` are supplied then the default `retryPolicy` of `OnFailure`, the previous behavior, also remains.

Test cases are added to show this works.

I have clarified the current behavior in one commit, so in the event we don't want this change I indent to submit that as a PR.

Fixes #10980

PR #12553, opened 2024-01-19
fix:  Allow users to selectively retry specific failed nodes . Fixes #12543

### LINKED ISSUE #12543: Retrying specific failed node does not work

mio4kon:
Fixes #12543 

### Motivation

### Modifications

Removed the restriction that required the simultaneous use of `--node-field-selector` and `--restart-successful`. Now, using `--node-field-selector` alone allows for individual retries of specific failed nodes, instead of retrying all failures.

### Verification
image

_--node-field-selector can be used independently._
`./dist/argo retry fail-24ptx --node-field-selector name=fail-24ptx.BB -v `

_Regressively used in combination._
`./dist/argo retry fail-mz9c4 --restart-successful --node-field-selector name=fail-mz9c4.A`


workflow/util/util_test.go
```diff
@@ -1367,6 +1367,37 @@ func TestFormulateRetryWorkflow(t *testing.T) {
 			}
 		}
 	})
+
+	t.Run("Nested DAG with Non-group Node Selected, NO restartSuccessful", func(t *testing.T) {
```
agilgur5, line 1371: this should be worded the same way as the other test cases
```suggestion
	t.Run("Retry failed workflow with only nodeFieldSelector", func(t *testing.T) {
```

workflow/util/util.go
```diff
@@ -1047,12 +1056,11 @@ func GetTemplateFromNode(node wfv1.NodeStatus) string {
...
 
-func getNodeIDsToReset(restartSuccessful bool, nodeFieldSelector string, nodes wfv1.Nodes) (map[string]bool, error) {
+func getNodeIDsToReset(nodeFieldSelector string, nodes wfv1.Nodes) (map[string]bool, error) {
 	nodeIDsToReset := make(map[string]bool)
-	if !restartSuccessful || len(nodeFieldSelector) == 0 {
+	if len(nodeFieldSelector) == 0 {
 		return nodeIDsToReset, nil
 	}
-
 	selector, err := fields.ParseSelector(nodeFieldSelector)
```
agilgur5, line 1064: this should have different behavior based on `restartSuccessful` -- when it's false, successful nodes shouldn't be reset

workflow/util/util.go
```diff
@@ -960,6 +960,15 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 				log.Debugf("Reset %s node %s since it's a group node", node.Name, string(node.Phase))
 				continue
 			} else {
+				// If restartSuccessful flag is unset and nodeFieldSelector is set, retry the failed node specified by nodeFieldSelector
```
agilgur5, line 963:
```suggestion
				// If restartSuccessful is unset and nodeFieldSelector is set, only retry the specified failed node
```

agilgur5 (commented):
Something feels off here -- `nodeIdsToReset` should be different if `--restart-successful` is set or not. This is making it unconditionally restart all successful nodes now. This might need more semantic refactoring

workflow/util/util.go
mio4kon: @agilgur5 Submitted a new commit, please help me review it

workflow/util/util.go
```diff
@@ -1049,16 +1058,18 @@ func GetTemplateFromNode(node wfv1.NodeStatus) string {
...
 	}
-
 	selector, err := fields.ParseSelector(nodeFieldSelector)
 	if err != nil {
 		return nil, err
 	} else {
 		for _, node := range nodes {
 			if SelectorMatchesNode(selector, node) {
+				if !restartSuccessful && node.Phase == wfv1.NodeSucceeded {
+					return nil, errors.Errorf(errors.CodeBadRequest, "cannot restart successful node %s , plz add --restart-successful", node.Name)
```
agilgur5, line 1071: this is a helper function, it shouldn't error out when you pass 1 of 2 options (it's a boolean) of an argument to it.

this should just `continue` if that's the case, which properly handles both conditions
```suggestion
					continue
```

also this is a formal error message, it should not use language like "plz"
and it's an API error message, while `--restart-successful` is specific to the CLI

agilgur5: the `continue` is a correctness fix, but this would be better consolidated if it actually used the selector to filter out this condition instead

agilgur5:
> feat: [...]

Please re-title this PR as a `fix:`, since per https://github.com/argoproj/argo-workflows/issues/12543#issuecomment-1900910489 this very much seems like a bug and not intended behavior

mio4kon:
> > feat: [...]
> 
> Please re-title this PR as a `fix:`, since per [#12543 (comment)](https://github.com/argoproj/argo-workflows/issues/12543#issuecomment-1900910489) this very much seems like a bug and not intended behavior

done

mio4kon:
> > feat: [...]
> 
> Please re-title this PR as a `fix:`, since per [#12543 (comment)](https://github.com/argoproj/argo-workflows/issues/12543#issuecomment-1900910489) this very much seems like a bug and not intended behavior

@agilgur5  hello，Will this MR be merged into the trunk in the future?

workflow/util/util.go
```diff
@@ -1049,16 +1058,18 @@ func GetTemplateFromNode(node wfv1.NodeStatus) string {
...
 		return nodeIDsToReset, nil
 	}
-
 	selector, err := fields.ParseSelector(nodeFieldSelector)
 	if err != nil {
 		return nil, err
 	} else {
 		for _, node := range nodes {
 			if SelectorMatchesNode(selector, node) {
+				if !restartSuccessful && node.Phase == wfv1.NodeSucceeded {
```
agilgur5, line 1070: I mentioned this [before](https://github.com/argoproj/argo-workflows/pull/12553#discussion_r1461432118), but the latter condition would be better as part of the selector if possible

workflow/util/util.go
agilgur5: Above comment followed up on [below](https://github.com/argoproj/argo-workflows/pull/12553#discussion_r1483696538)

isubasinghe (changes requested):
Can you add two e2e tests please? 
One that is a dag and one that is not.

Don't stress about invoking the server if there isn't infrastructure for those kind of tests (although I you should be able to use REST) already.

Just add those two tests with comments linking them to this issue and PR.

tooptoop4:
:shipit:

workflow/util/util.go
mio4kon: Hello. Here, I want to confirm whether restartSuccessful and selector have the highest priority of selector. If the selector selector selects a successful node but does not specify restartSuccessful, whether to re-execute the successful node. In this case, the restartSuccessful parameter doesn't feel very meaningful.

agilgur5: The diagram looks correct to me

mio4kon: > The diagram looks correct to me

However, the above diagram may be a bit in conflict with the previous logic, which may cause problems with the previous e2e test. For example, `TestFormulateRetryWorkflow/Nested_DAG_with_Non-group_Node_Selected`. Do you think it's better to fix only the previous red logic in the figure below and keep the previous blue logic?

image

mio4kon:
> Can you add two e2e tests please? One that is a dag and one that is not.
> 
> Don't stress about invoking the server if there isn't infrastructure for those kind of tests (although I suspect you should be able to use REST) already.
> 
> Just add those two tests with comments linking them to this issue and PR.

add e2e tests : `TestRetryWorkflowWithStepsWithSelectedFailNodes`  and `TestRetryWorkflowWithDAGWithSelectedFailNodes`, Please help to check it out

workflow/util/util.go
agilgur5: Ahhhh -- based on the issue and your fix I strongly suspected there was something deeper incorrect with the previous logic. I actually [asked in the contributors channel](https://cloud-native.slack.com/archives/C0510EUH90V/p1707434418760689) if someone more familiar with the retry logic could check this PR because of that suspicion. 

> Do you think it's better to fix only the previous red logic in the figure below and keep the previous blue logic?

Ah is that what you were doing with your initial/current fix?
I honestly think we should change it all to work as expected (the black boxes) since the previous logic feels very confusing (and has confused users many a time before).

Big thanks for drawing the diagrams here, those are super helpful! We may want to add a similar `mermaid` flowchart of this to the docs as well

agilgur5: If we rewrite this, we may want to release it as a breaking change similar to `retryStrategy` fixes from https://github.com/argoproj/argo-workflows/pull/11005

(different retries, but similar concept that they both behaved unexpectedly)

mio4kon: > If we rewrite this, we may want to release it as a breaking change similar to `retryStrategy` fixes from #11005
> 
> (different retries, but similar concept that they both behaved unexpectedly)

By the way, do you have any plans to rewrite this part of the logic? Right now, our business needs to be able to retry a single error node. The changes to this issue are to minimize the impact on existing API logic and to provide the ability to retry a single faulty node. Of course, a broader refactoring might provide a more elegant solution. Do you know if the official team has any plans in this regard? 😃

JasonChen86899:
Hi, I review the logic and this modification may cause problems. The current logic of the `mian` branch is to not retry the error node with successful child nodes. This modification will result in some errors

shuangkun:
Hi, will this question be updated again? I encountered a new problem. I want to retry a hanging node while the workflow is running.

isubasinghe:
Just an update of the retry logic. I had to rewrite it from scratch, will create a PR very soon (within the day).

agilgur5:
For reference, I mentioned this PR in https://github.com/argoproj/argo-workflows/issues/13692#issuecomment-2403869711

> Closing this as a result, the current retry logic is very broken.

I assume Isitha's refactor will fix the diagram @mio4kon made [above](https://github.com/argoproj/argo-workflows/pull/12553#discussion_r1493314678), which illustrates some of the broken logic

workflow/util/util.go
agilgur5: > Of course, a broader refactoring might provide a more elegant solution. Do you know if the official team has any plans in this regard? 😃

Should be happening shortly per [below](https://github.com/argoproj/argo-workflows/pull/12553#issuecomment-2404005416)

agilgur5:
> I had to rewrite it from scratch, will create a PR very soon (within the day).

See #13734

shuangkun:
Can nodes that did not fail also be retried？

PR #12559, opened 2024-01-20
fix: Global Artifact Passing. Fixes #12554

### LINKED ISSUE #12554: The global output cannot be used correctly after upgrade v3.5.3
### Pre-requisites

Hi all,

We face a regression problem after upgrading to release `v3.5.3`. It should be related to this PR.

`We have a global artifact which is used and exported in two different steps WITH different values:`
- e2e-test-inject-runner-context
- e2e-test-output-testupdate

In the previous release, we can get the current global artifact from the last step (e2e-test-output-testupdate).
But in the new release v3.5.3 , the global artifact will be overwritten. so the artifact will be picked up randomly from these two steps which make our process failed.

I found there is a NEW change here:
https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/taskresult.go#L89-L90
```
  // Add outputs to global scope here to ensure that they are reflected in archive.
  woc.addOutputsToGlobalScope(newNode.Outputs)
```

Which will overwrite the global artifact after every task and should be the reason to make our process failed. We can find the overwriting log here:
```
time="2024-01-19T02:03:53.554Z" level=debug msg="task result name:\ne2e-test" namespace=action-dev2 workflow=e2e-test
time="2024-01-19T02:03:53.554Z" level=debug msg="Marking task result complete e2e-test" namespace=action-dev2 workflow=e2e-test
time="2024-01-19T02:03:53.554Z" level=info msg="overwriting workflow.outputs.artifacts.testUpdate: {testUpdate  <nil>  {<nil> &S3Artifact{S3Bucket:S3Bucket{Endpoint:,Bucket:,Region:,Insecure:nil,AccessKeySecret:nil,SecretKeySecret:nil,RoleARN:,UseSDKCreds:false,CreateBucketIfNotPresent:nil,EncryptionOptions:nil,CASecret:nil,},Key:e2e-test/e2e-test-inject-input-type-instances/testUpdate.tgz,} nil nil nil nil nil nil nil nil}  nil false  false  nil false}" namespace=action-dev2 workflow=e2e-test-1-4065749210214828732
```

I am not sure if it is a special case that it is not covered well.
- `start-entrypoint` step is a template with some inner steps, and we update the `testUpdate` global output artifact in these inner steps.
- The problem happens in the next step `upload-output-type-instances-step` just after `start-entrypoint`, I am not sure if it loads artifacts differently or delays （But it works correctly before).

Before step `upload-output-type-instances-step`, the status is:
Screenshot 2024-01-19 at 15 12 58

After step `upload-output-type-instances-step`, sometimes the status is changed to:
Screenshot 2024-01-19 at 15 15 16

Thanks!

### Version

v3.5.3

```YAML
kind: Workflow
metadata:
  generation: 21
  generateName: test-artifact-12402-
  namespace: action-dev2
spec:
  activeDeadlineSeconds: 28800
  arguments: {}
  entrypoint: test-root
  serviceAccountName: test
  templates:
  - inputs: {}
    metadata: {}
    name: test-root
    outputs: {}
    steps:
    - - arguments: {}
        name: inject-input-artifact
        template: inject-input-artifact
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testInput}}'
            name: testInput
          - from: '{{workflow.outputs.artifacts.testUpdate}}'
            name: testUpdate
        name: start-entrypoint
        template: main
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testUpload}}'
            name: testUpload
        name: upload-output-type-instances-step
        template: upload-output-type-instances
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testUpdate}}'
            name: testUpdate
        name: upload-update-type-instances-step
        template: upload-update-type-instances

  - inputs: {}
    metadata: {}
    name: main
    outputs: {}
    steps:
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testInput}}'
            name: input
        name: cp
        template: cp
    - - arguments:
          artifacts:
          - from: '{{workflow.outputs.artifacts.testUpdate}}'
            name: input-parameters
        name: render-update-ti
        template: main-render-update-ti-template
    - - arguments:
          artifacts:
          - from: '{{steps.cp.outputs.artifacts.upload}}'
            name: testUpload
        name: output-testupload
        template: output-testupload
    - - arguments:
          artifacts:
          - from: '{{steps.render-update-ti.outputs.artifacts.render}}'
            name: testUpdate
        name: output-testupdate
        template: output-testupdate

  - container:
      image: alpine:3.7
      command: [sh, -c]
      args: ["sleep 1; echo -n test input > /testInput.yaml; echo -n test update > /testUpdate.yaml"]
    name: inject-input-artifact
    outputs:
      artifacts:
      - globalName: testInput
        name: testInput
        path: /testInput.yaml
      - globalName: testUpdate
        name: testUpdate
        path: /testUpdate.yaml

  - container:
      command: [sh, -c]
      args: ["sleep 1; cp /input /upload"]
      image: alpine:3.7
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: input
        path: /input
    metadata: {}
    name: cp
    outputs:
      artifacts:
      - name: upload
        path: /upload

  - container:
      image: alpine:3.7
      command: [sh, -c]
      args: ["sleep 1; echo -n change update!!! > /render.yml"]
    metadata: {}
    name: main-render-update-ti-template
    outputs:
      artifacts:
      - name: render
        path: /render.yml

  - container:
      args:
      - sleep 1
      command:
      - sh
      - -c
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpload
        path: /typeinstance
    metadata: {}
    name: output-testupload
    outputs:
      artifacts:
      - globalName: testUpload
        name: testUpload
        path: /typeinstance

  - container:
      args:
      - sleep 1
      command:
      - sh
      - -c
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpdate
        path: /typeinstance
    metadata: {}
    name: output-testupdate
    outputs:
      artifacts:
      - globalName: testUpdate
        name: testUpdate
        path: /typeinstance



  - container:
      args:
      - sleep 1
      command: [sh, -c]
      args: ["sleep 1; cat /upload/typeInstances/testUpload; cat /upload/typeInstances/testUpload > /upload/result.yaml"]
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpload
        path: /upload/typeInstances/testUpload
    metadata: {}
    name: upload-output-type-instances
    outputs:
      artifacts:
      - globalName: uploadresult
        name: uploadresult
        path: /upload/result.yaml

  - container:
      args:
      - sleep 1
      command: [sh, -c]
      args: ["sleep 1; cat /upload/typeInstances/testUpdate"]
      image: alpine:3.18.4
      name: ""
      resources: {}
    inputs:
      artifacts:
      - name: testUpdate
        path: /upload/typeInstances/testUpdate
    metadata: {}
    name: upload-update-type-instances
    outputs: {}
```

If you run this workflow several times, you will find the final global output artifact will be changed to:
Screenshot 2024-01-19 at 14 59 13

`testUpdate` is refered to `test-artifact-12402-2cntl/test-artifact-12402-2cntl-inject-input-artifact-46224399/testUpdate.tgz`

But acutally, it should use the the artifact from later step, like:
`test-artifact-12402-2cntl/test-artifact-12402-2cntl-output-testupdate-4251288622/testUpdate.tgz`

Please let me know if anything is not clear. Thanks!
```


```text


time="2024-01-19T02:03:53.554Z" level=debug msg="task result name:\ne2e-test" namespace=action-dev2 workflow=e2e-test
time="2024-01-19T02:03:53.554Z" level=debug msg="Marking task result complete e2e-test" namespace=action-dev2 workflow=e2e-test
time="2024-01-19T02:03:53.554Z" level=info msg="overwriting workflow.outputs.artifacts.testUpdate: {testUpdate  <nil>  {<nil> &S3Artifact{S3Bucket:S3Bucket{Endpoint:,Bucket:,Region:,Insecure:nil,AccessKeySecret:nil,SecretKeySecret:nil,RoleARN:,UseSDKCreds:false,CreateBucketIfNotPresent:nil,EncryptionOptions:nil,CASecret:nil,},Key:e2e-test/e2e-test-inject-input-type-instances/testUpdate.tgz,} nil nil nil nil nil nil nil nil}  nil false  false  nil false}" namespace=action-dev2 workflow=e2e-test-1-4065749210214828732
```
```


```text

No error in the wait container
```

Garett-MacGowan:
Fixes #12554

### Motivation

An issue with global artifact passing was introduced in https://github.com/argoproj/argo-workflows/pull/12402.

### Modifications

The fix is to protect the call to `addOutputsToGlobalScope()` in `taskResultReconciliation` with `checkReconciliationComplete()`. This allows reconciliation to promote outputs to the global scope, but only when the workflow is complete.

This ensures that artifacts make it into the archive, while also ensuring that the order of global artifact availability in the workflow isn't disrupted.

### Verification

Added multiple global artifact passing tests. Notably, `complex-global-artifact-passing.yaml` would frequently (but not always) fail before the fix made in this PR.

### DIFF
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -1051,12 +1051,20 @@ func (s *ArgoServerSuite) TestArtifactServerArchivedStoppedWorkflow() {
 			nodeID = status.Nodes.FindByDisplayName("create-artifact").ID
 		})
 
-	s.Run("GetArtifactByNodeID", func() {
-		s.e().GET("/artifact-files/argo/archived-workflows/{uid}/{nodeID}/outputs/artifact-creator", uid, nodeID).
+	s.Run("GetLocalArtifactByNodeID", func() {
+		s.e().GET("/artifact-files/argo/archived-workflows/{uid}/{nodeID}/outputs/local-artifact", uid, nodeID).
 			Expect().
 			Status(200).
 			Body().
 			Contains("testing")
+	})
+
+	s.Run("GetGlobalArtifactByNodeID", func() {
+		s.e().GET("/artifact-files/argo/archived-workflows/{uid}/{nodeID}/outputs/global-artifact", uid, nodeID).
+			Expect().
+			Status(200).
+			Body().
+			Contains("testing global")
 	})
 }
 
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -63,60 +63,63 @@ func (s *ArtifactsSuite) TestArtifactPassing() {
 }
 
 type expectedArtifact struct {
-	key                   string
-	bucketName            string
-	value 	              string
+	key        string
+	bucketName string
+	value      string
 }
 
 func (s *ArtifactsSuite) TestGlobalArtifactPassing() {
 	for _, tt := range []struct {
-		workflowFile string
+		workflowFile     string
 		expectedArtifact expectedArtifact
 	}{
 		{
 			workflowFile: "@testdata/global-artifact-passing.yaml",
 			expectedArtifact: expectedArtifact{
-				key: "globalArtifact",
+				key:        "globalArtifact",
 				bucketName: "my-bucket-3",
-				value: "01",
+				value:      "01",
 			},
 		},
 		{
 			workflowFile: "@testdata/complex-global-artifact-passing.yaml",
 			expectedArtifact: expectedArtifact{
-				key: "finalTestUpdate",
+				key:        "finalTestUpdate",
 				bucketName: "my-bucket-3",
-				value: "Updated testUpdate",
+				value:      "Updated testUpdate",
 			},
 		},
 	} {
 		then := s.Given().
-		Workflow(tt.workflowFile).
-		When().
-		SubmitWorkflow().
-		WaitForWorkflow(fixtures.ToBeSucceeded, time.Minute*2).
-		Then().
-		ExpectWorkflow(func(t *testing.T, objectMeta *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
-			// Check the global artifact value and see if it equals the expected value.
-			c, err := minio.New("localhost:9000", &minio.Options{
-				Creds: credentials.NewStaticV4("admin", "password", ""),
+			Workflow(tt.workflowFile).
+			When().
+			SubmitWorkflow().
+			WaitForWorkflow(fixtures.ToBeSucceeded, time.Minute*2).
+			Then().
+			ExpectWorkflow(func(t *testing.T, objectMeta *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+				// Check the global artifact value and see if it equals the expected value.
+				c, err := minio.New("localhost:9000", &minio.Options{
+					Creds: credentials.NewStaticV4("admin", "password", ""),
+				})
+
+				if err != nil {
+					t.Error(err)
+				}
+
+				object, err := c.GetObject(context.Background(), tt.expectedArtifact.bucketName, tt.expectedArtifact.key, minio.GetObjectOptions{})
+				if err != nil {
+					t.Error(err)
+				}
+
+				buf := new(bytes.Buffer)
+				_, err = buf.ReadFrom(object)
+				if err != nil {
+					t.Error(err)
+				}
+				value := buf.String()
+
+				assert.Equal(t, tt.expectedArtifact.value, value)
 			})
-
-			if err != nil {
-				t.Error(err)
-			}
-
-			object, err := c.GetObject(context.Background(), tt.expectedArtifact.bucketName, tt.expectedArtifact.key, minio.GetObjectOptions{})
-			if err != nil {
-				t.Error(err)
-			}
-
-			buf := new(bytes.Buffer)
-			buf.ReadFrom(object)
-			newStr := buf.String()
-
-			assert.Equal(t, tt.expectedArtifact.value, newStr)
-		})
 
 		then.
 			When().
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -238,7 +238,7 @@ func (s *ArtifactsSuite) TestDeleteWorkflow() {
 
 	when.WaitForWorkflowDeletion()
 
-	when = when.RemoveFinalizers(false)
+	when.RemoveFinalizers(false)
 }
 
 func (s *ArtifactsSuite) TestArtifactGC() {
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -4,9 +4,11 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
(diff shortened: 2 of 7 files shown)


Garett-MacGowan:
@zhangtbj Here's the fix. Will mark as ready for review assuming all the tests pass.

Garett-MacGowan:
@juliev0 Should I still be requesting reviews from you for artifact issues?

zhangtbj:
Thanks @Garett-MacGowan , 👍 

I can also help test once I get the image.

juliev0:
> @juliev0 Should I still be requesting reviews from you for artifact issues?

Thanks for asking. Since I have the history, that can work.

juliev0:
Thanks for the immediate fix! Seems like it's probably a good fix, but can you help me to understand exactly what was happening? We should've only been making that call when the node and taskResult were complete, right?

Garett-MacGowan:
> Thanks for the immediate fix! Seems like it's probably a good fix, but can you help me to understand exactly what was happening? We should've only been making that call when the node and taskResult were complete, right?

Yeah, if a task result was reconciled for an old step during the processing of a subsequent step, then the global artifact could be overwritten with the old global artifact, whereas it should have been using the global artifact update from the prior step.

Take this flow for example, which is ~zhangtbj's workflow:
1) Generate a starting global artifact
2) Pass it to a step which updates it
2.5) Artifact reconciliation for 1 happens, overwriting the global artifact which was updated in 2.
3) Save the global artifact.

Now, task reconciliation won't set the global artifact until the workflow is complete, meaning we're not at risk of messing with global artifacts during the workflow.

Does this make sense?

juliev0:
> > Thanks for the immediate fix! Seems like it's probably a good fix, but can you help me to understand exactly what was happening? We should've only been making that call when the node and taskResult were complete, right?
> 
> Yeah, if a task result was reconciled for an old step during the processing of a subsequent step, then the global artifact could be overwritten with the old global artifact, whereas it should have been using the global artifact update from the prior step.
> 
> Take this flow for example, which is ~zhangtbj's workflow:
> 
> 1. Generate a starting global artifact
> 2. Pass it to a step which updates it
>    2.5) Artifact reconciliation for 1 happens, overwriting the global artifact which was updated in 2.
> 3. Save the global artifact.
> 
> After the fix, task reconciliation won't set the global artifact until the workflow is complete, meaning we're not at risk of messing with global artifacts during the workflow.
> 
> Does this make sense?

Thanks for this. I looked over the Workflow that was pasted, and this makes sense.

I am revisiting your original [rationale](https://github.com/argoproj/argo-workflows/pull/12402/files#r1440021917) for adding this line here. I need to look at this more closely wrt the fact that we call `woc.addOutputsToGlobalScope()` in so many places and wonder if there can be any simplification. It may be fine as is, but want to check.

shuangkun:
I think the problem cause by the processing order of taskresults. Maybe we can delete "addOutputsToGlobalScope" in taskresult（podReconciliation will call it soon），and we can control the  order "addOutputsToGlobalScope"  in podReconciliation.  https://github.com/argoproj/argo-workflows/pull/12537/commits

Garett-MacGowan:
The problem this solves is that the node status can be marked as `wfv1.NodeFailed` before `taskResultReconciliation` completes for an artifact. This means that the call to `addOutputsToGlobalScope` in `podReconciliation` won't execute because there hasn't been a node status change per `assessNodeStatus`.

Garett-MacGowan:
@juliev0 I think I confused myself somewhere with respect to the need to call `addArtifactToGlobalScope` in `taskResultReconciliation` for the specific issue addressed in #12402. However, I took another look at the podReconciliation logic and I still think that artifacts won't get added to the global scope in the event that the node status changes to its final state (e.g. `wfv1.NodeFailed`) before it is reconciled.

I went ahead and ran `TestArtifactServerArchivedStoppedWorkflow` locally after removing the call to `woc.addOutputsToGlobalScope` and it works fine (as expected, there is no global artifact used in that test).

I think I will add some more tests around global scope artifacts tomorrow & see if this change is necessary. If it proves unnecessary, we can remove it entirely.

Garett-MacGowan:
> I think the problem cause by the processing order of taskresults. Maybe we can delete "addOutputsToGlobalScope" in taskresult.go（podReconciliation will call it soon），and we can control the order "addOutputsToGlobalScope" in podReconciliation. https://github.com/argoproj/argo-workflows/pull/12537/commits

I think you're right about not needing to call `addOutputsToGlobalScope` in `taskResultReconciliation`. I'm not sure we need to change anything else though. The use case I was thinking of is moot. If a workflow finishes or fails, i'm not sure if there is any use in making sure artifacts are in the global scope at that point.

juliev0:
> > I think the problem cause by the processing order of taskresults. Maybe we can delete "addOutputsToGlobalScope" in taskresult.go（podReconciliation will call it soon），and we can control the order "addOutputsToGlobalScope" in podReconciliation. https://github.com/argoproj/argo-workflows/pull/12537/commits
> 
> I think you're right about not needing to call `addOutputsToGlobalScope` in `taskResultReconciliation`. I'm not sure we need to change anything else though. The use case I was thinking of is moot. If a workflow finishes or fails, i'm not sure if there is any use in making sure artifacts are in the global scope at that point.

Great! That would be great if we don't need it at all. I admittedly haven't gone through the work of reaching that conclusion, but if you have and don't feel we need it, it will great not to have it there too.

Garett-MacGowan:
> > > I think the problem cause by the processing order of taskresults. Maybe we can delete "addOutputsToGlobalScope" in taskresult.go（podReconciliation will call it soon），and we can control the order "addOutputsToGlobalScope" in podReconciliation. https://github.com/argoproj/argo-workflows/pull/12537/commits
> > 
> > 
> > I think you're right about not needing to call `addOutputsToGlobalScope` in `taskResultReconciliation`. I'm not sure we need to change anything else though. The use case I was thinking of is moot. If a workflow finishes or fails, i'm not sure if there is any use in making sure artifacts are in the global scope at that point.
> 
> Great! That would be great if we don't need it at all. I admittedly haven't gone through the work of reaching that conclusion, but if you have and don't feel we need it, it will great not to have it there too.

Yeah, I added to the test to verify that artifacts marked with a globalName are still archived. They are, so it shouldn't matter.

juliev0 (approved):

juliev0:
Since it sounds like you've taken a look at this PR @shuangkun - do you have any objections to merging it as is?

shuangkun:
> Since it sounds like you've taken a look at this PR @shuangkun - do you have any objections to merging it as is?


I have no objection, thanks.

PR #12560, opened 2024-01-22
docs(quick-start): consolidate UI & port-forward sections

agilgur5:
### Motivation

- these sections list the same thing and are duplicative, so combine them / remove one
  - leave the submitting example one as that's specifically written to show the UI vs. the CLI
  - having the `port-forward` part when using the CLI doesn't make sense since you don't need it for the CLI

### Modifications

- combine the `port-forward` section with the UI submit example section

- <details><summary>also fix some docs issues in this section</summary>

  - grammar issues:
    - "Submitting" -> "Submit" -- use present tense, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-present-tense)
      - also, the CLI and UI subsections had "Submit" but the top-level had "Submitting", which is inconsistent grammar
    - "port-forward" is a k8s verb, not a noun, so "Open a port-forward" is grammatically incorrect
      - replace with "Forward the Server's port"

  - markdown issues:
    - UI directions should be an ordered list, not unordered, since they are in-order
      - the `port-forward` command should be inside of the list, not outside
    - the `http` vs. `https` note isn't a quotation and so shouldn't be in a blockquote
      - if anything we could use a `!!! note`, but now that it's within a list, I think that would take up too much space isn't necessary

  - style issues:
    - "GUI" -> "UI", which is how it is referred to elsewhere, including 1 line below the "GUI" text
    - "an example workflow" -> "via the UI/CLI" -- the `h2` is already "Submit an example workflow" and should not be duplicated
    - be more direct in the `https` note, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

  </details>

### Verification

1. `mkdocs build` passes
1. Looks as expected
    - Before: 
Screenshot 2024-01-21 at 10 02 42 PM

    - After:
Screenshot 2024-01-21 at 10 06 33 PM

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -27,25 +27,13 @@ kubectl create namespace argo
 kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v<<ARGO_WORKFLOWS_VERSION>>/quick-start-minimal.yaml
 ```
 
-### Port-forward the UI
-
-Open a port-forward so you can access the UI:
-
-```bash
-kubectl -n argo port-forward deployment/argo-server 2746:2746
-```
-
-This will serve the UI on <https://localhost:2746>. Due to the self-signed certificate, you will receive a TLS error which you will need to manually approve.
-
-> Pay close attention to the URI. It uses `https` and not `http`. Navigating to `http://localhost:2746` results in a server-side error that breaks the port forwarding.
-
 ## Install the Argo Workflows CLI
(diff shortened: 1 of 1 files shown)


Joibel (approved):

jmeridth (approved):

agilgur5:
Fixed merge conflicts

PR #12561, opened 2024-01-22
docs(install): use material admonition instead of unicode caution

agilgur5:
### Motivation

- use [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) that come with the material theme instead of a unicode caution sign
  - consistent with the design of the docs 

### Modifications

- replace unicode caution signs with admonitions

- be more specific in the notes as well
  - using a remote base is not inherently "dangerous" as the previous sentence said without elaboration
    - instead specify that you should use a full hash when using a remote base
    - and add a link to Kustomize's remote base docs
  - similarly in the second note, specify what "tip" means
    - and specify what you should do instead of using `latest`

### Verification

1. `mkdocs build` passes
1. Looks as expected:
    - Before:
Screenshot 2024-01-21 at 10 20 47 PM

    - After:
Screenshot 2024-01-21 at 10 22 31 PM

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -14,9 +14,12 @@ To install Argo Workflows, navigate to the [releases page](https://github.com/ar
 
 You can use Kustomize to patch your preferred [configurations](managed-namespace.md) on top of the base manifest.
 
-⚠️ If you are using GitOps, never use Kustomize remote base: this is dangerous. Instead, copy the manifests into your Git repo.
+!!! Note "Use a full hash"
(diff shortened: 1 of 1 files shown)


jmeridth (approved):

terrytangyuan (approved):

agilgur5:
Backported cleanly to `release-3.5` as 0d4bb58deca980458f9309315d01dc668149fd3b

PR #12567, opened 2024-01-24
fix: make etcd errors transient

Joibel:
These errors occasionally occur on LKE (Linode hosted kubernetes) in response to normal actions, and are perfectly retriable. It isn't unreasonable to have these errors exposed to users, even if it doesn't appear to happen on any other cluster types I've tried.

We have tried to replicate with toy clusters.

### Motivation

Put the errors in the code rather than expect users to have to do this themselves.

### Verification

Verified using the TRANSIENT_ERROR_PATTERN by a customer as I don't have access to an LKE cluster.

### DIFF
--- a/util/errors/errors.go
+++ b/util/errors/errors.go
@@ -28,7 +28,15 @@ func IsTransientErr(err error) bool {
 		return false
 	}
 	err = argoerrs.Cause(err)
-	isTransient := isExceededQuotaErr(err) || apierr.IsTooManyRequests(err) || isResourceQuotaConflictErr(err) || isResourceQuotaTimeoutErr(err) || isTransientNetworkErr(err) || apierr.IsServerTimeout(err) || apierr.IsServiceUnavailable(err) || matchTransientErrPattern(err) ||
+	isTransient := isExceededQuotaErr(err) ||
+		apierr.IsTooManyRequests(err) ||
+		isResourceQuotaConflictErr(err) ||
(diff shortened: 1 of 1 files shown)


util/errors/errors.go
```diff
@@ -61,6 +69,16 @@ func isResourceQuotaTimeoutErr(err error) bool {
 	return apierr.IsInternalError(err) && strings.Contains(err.Error(), "resource quota evaluation timed out")
 }
 
+func isTransientEtcdErr(err error) bool {
+	// Some clusters expose these (transient) etcd errors to the caller
+	if strings.Contains(err.Error(), "etcdserver: leader changed") {
+		return true
+	} else if strings.Contains(err.Error(), "etcdserver: request timed out") {
+		return true
```
terrytangyuan, line 77: Can this be merged into isTransientNetworkErr?

Joibel: I did consider it, but it isn't strictly a network error. We're successfully talking to the k8s API, which is telling us to go away because it's not ready right now.

I think the network errors are errors between the workflow-controller (or executor because it also uses this logic) and the k8s API.

That said, I'd rather get this in somewhere than have this block the issue, so if you're sure you want it there I'm happy to change it.

terrytangyuan (approved):
LGTM given that this is etcd. For other errors, let's encourage people to use `TRANSIENT_ERROR_PATTERN`

Issue #11849, opened 2023-09-20
Feature Request: New Canceled status


jmeridth:
# Summary

new `Canceled` status

NOTE: debates can ensue regarding one l (American English) or two l (British English).  Looks like CircleCI and TravisCI use one.  They could be doing i18n and because I'm in the USA it renders as one. 🤷 

## Use Cases

Since many consumers of argo-workflows are using it for Continuous Integration (CI), when the following sequence occurs, a `Cancelled` status would be nice:

- pull request created
- workflow starts CI run
- new commit(s) pushed to pull request
- if the previous workflow is still running, it should be terminated and status set to `Canceled` (currently when terminated its status is set to `Failed`)
- new workflow starts CI run, including new commits

This will allow for the analytics to be accurate on which workflows actually `Failed` 😄 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

jmeridth:
We currently determine previous workflows by having passed GitHub org, repo and pr number as labels and before we kick off a new workflow (via argo events to a custom service) we call the argo-workflows API to find any previously running workflows with those labels and terminate them (resulting in a `❌  Failed` status with message `Stopped with strategy 'Terminate'`

terrytangyuan:
Some comments in https://github.com/argoproj/argo-workflows/pull/11992?notification_referrer_id=NT_kwDOAEEnSrI4MDQxMDkwOTM0OjQyNjk4OTg#discussion_r1370185782

agilgur5:
> There's also a question here of if we should add a `Cancelled` `NodeStatus` as well. I believe that currently the running node is marked as `Failed` and all later nodes are marked as `Skipped`, both of which require a bit of interpretation right now as they are not the same as regular failures or regular skips. This popped up in [a code review](https://github.com/argoproj/argo-workflows/pull/11493#discussion_r1280723213) too

This was changed in https://github.com/argoproj/argo-workflows/pull/13214 to `Failed`, but `Cancelled` would still be easier to understand/interpret

PR #12574, opened 2024-01-25
fix: ensure workflowtaskresults complete before mark workflow completed status. Fixes: #12615

### LINKED ISSUE #12615: Workflow in `Running` but `workflows.argoproj.io/completed` is `true`
### Pre-requisites

I want when I retry a workflow, if it retry success, it will be Running and the workflow continues processing.
But in this case, the workflow is stuck.
The issue is to record https://github.com/argoproj/argo-workflows/pull/12574

### Version

latest

This problem is relatively difficult to reproduce. This situation can only occur if the user performs a retry during the last “Processing workflow”. This may occur when very large workflows and API Servers are under pressure.

This is difficult to reproduce in unit testing and e2e testing. But I designed a reasonable experiment to reproduce it and added it to the test for reference.

First, add the code to func "taskResultReconciliation". （I think this is possible, when the Wokflow is large and has complex outputs, processing the workflow may take more than 2s.）
```go
if woc.checkReconciliationComplete() {
	time.Sleep(time.Second*2)
}
```
Second,  run the test. Retry the failed workflow when "Processing workflow". The test is in https://github.com/argoproj/argo-workflows/pull/12574
```sh
make TestRetryStoppedButIncompleteWorkflow
```

Finally, you will get a workflow in Running status but its labelCompleted is true.
```
tianshuangkun@B-QGHAMD6M-2124 logs % kubectl get workflow -o yaml -w | grep workflows.argoproj.io/completed
    workflows.argoproj.io/completed: "false"
    workflows.argoproj.io/completed: "false"
    workflows.argoproj.io/completed: "true"
    workflows.argoproj.io/completed: "true"
^C
tianshuangkun@B-QGHAMD6M-2124 logs % argo list
NAME                     STATUS    AGE   DURATION   PRIORITY
wf-retry-stopped-b57pd   Running   8s    5s         0
```
```


```text
```

```text
```

shuangkun:
Fixes #12615

I encountered a problem：

My workflow failed but don‘t have key LabelKeyCompleted. So enter “Processing workflow”， at the same time，my user find the workflow failed and retry it.
 
At Last, the workflow is Running but have LabelKeyCompleted, so it will never be dealt with again.

So I think we need to ensure workflowtaskresults complete before mark workflow completed status

```
time="2024-01-25T02:58:46.957Z" level=info msg="Updated phase Running -> Failed" namespace=argo workflow=dongdong
time="2024-01-25T03:00:25.693Z" level=info msg="Workflow update successful" namespace=argo phase=Failed resourceVersion=41288141 workflow=dongdong
time="2024-01-25T03:00:31.005Z" level=info msg="Processing workflow" namespace=argo workflow=dongdong
time="2024-01-25T03:03:08.430Z" level=warning msg="Error updating workflow: Operation cannot be fulfilled on workflows.argoproj.io \"dongdong\": the object has been modified; please apply your changes to the latest version and try again Conflict" namespace=argo workflow=dongdong
```

### Reproduce

This problem is relatively difficult to reproduce. This situation can only occur if the user performs a retry during the last “Processing workflow”. This may occur when very large workflows and API Servers are under pressure.

This is difficult to reproduce in unit testing and e2e testing. But I designed a reasonable experiment to reproduce it and added it to the test for reference.

First, add the code to func "taskResultReconciliation". （I think this is possible, when the Wokflow is large and has complex outputs, processing the workflow may take more than 2s.）
```go
if woc.checkReconciliationComplete() {
	time.Sleep(time.Second*2)
}
```
Second,  run the test. Retry the failed workflow when "Processing workflow".
```
make TestRetryStoppedButIncompleteWorkflow
```

Finally, you will get a workflow in Running status but its labelCompleted is true.
```
tianshuangkun@B-QGHAMD6M-2124 logs % kubectl get workflow -o yaml -w | grep workflows.argoproj.io/completed
    workflows.argoproj.io/completed: "false"
    workflows.argoproj.io/completed: "false"
    workflows.argoproj.io/completed: "true"
    workflows.argoproj.io/completed: "true"
^C
tianshuangkun@B-QGHAMD6M-2124 logs % argo list
NAME                     STATUS    AGE   DURATION   PRIORITY
wf-retry-stopped-b57pd   Running   8s    5s         0
```

### Solution

I hope to define a state to describe the stage of workflow from reaching completion (Failed, Error, Succedd) to truly Completed,  named Completing.


Joibel:
As a simple code change, this feels wrong. 
It feels as though we are heaping workarounds on top of workarounds in the hope that we'll end up with something that works, and that there is an underlying problem with the design that needs fixing instead.

shuangkun:
> As a simple code change, this feels wrong. It feels as though we are heaping workarounds on top of workarounds in the hope that we'll end up with something that works, and that there is an underlying problem with the design that needs fixing instead.

Yes, I think I should change markWorkflowPhase logic, it is the root cause, shouldn't mark it completed status (Failed、Error) when the workflowtaskresults Incompleted.

Joibel:
I agree, so I'd prefer to reject and close this PR. Are you happy to do that?

shuangkun:
> As a simple code change, this feels wrong. It feels as though we are heaping workarounds on top of workarounds in the hope that we'll end up with something that works, and that there is an underlying problem with the design that needs fixing instead.

I think so, it is not a good 

> I agree, so I'd prefer to reject and close this PR. Are you happy to do that?



> I agree, so I'd prefer to reject and close this PR. Are you happy to do that?

No need to close it yet. I will submit a new commit in the branch soon.（My customer encountered this problem again. He has a large workflow (8000 steps), but after encountering this problem, he cannot retry, stop, terminate, and can hardly perform any operations. I did a lot of non-standard operations today and finally helped him retry this workflow. And I realized that the final solution should be shouldn't mark it completed status (Failed、Error) when the workflowtaskresults Incompleted. ） 
We can review it here, because it solve the same question. Thanks！

juliev0:
CC @Garett-MacGowan

juliev0:
Just read this from the original description: "So we need to ensure workflowtaskresults complete before mark workflow completed status". 
But that seems to be what we're doing [here](https://github.com/argoproj/argo-workflows/blob/6abe8a9503dcde3de7057922c6d7688c68ae4957/workflow/controller/operator.go#L2339-L2346), no?

shuangkun:
> Just read this from the original description: "So we need to ensure workflowtaskresults complete before mark workflow completed status". But that seems to be what we're doing [here](https://github.com/argoproj/argo-workflows/blob/6abe8a9503dcde3de7057922c6d7688c68ae4957/workflow/controller/operator.go#L2339-L2346), no?

I think we should check a little earlier, because if the code is executed here, the status of the workflow has been set to completed (error、Failed、successd), but it is still possible that common.LabelKeyCompleted is not set because the check fails.

Garett-MacGowan:
Will take a look at this after 😴

juliev0:
> > Just read this from the original description: "So we need to ensure workflowtaskresults complete before mark workflow completed status". But that seems to be what we're doing [here](https://github.com/argoproj/argo-workflows/blob/6abe8a9503dcde3de7057922c6d7688c68ae4957/workflow/controller/operator.go#L2339-L2346), no?
> 
> I think we should check it a little earlier, because if the code is executed here, the status of the workflow has been set to completed (error、Failed、successd), but it is still possible that common.LabelKeyCompleted is not set because the check fails. And then，the user may see its status is completed and to make some action update the workflow. At the sametime, the workflow-controller still processing the workflow because of no "common.LabelKeyCompleted" and may update the workflow's common.LabelKeyCompleted. At Last, the two updates will be Conflict and be merged，the workflow may be Running but have LabelKeyCompleted finnally.

Thanks for the explanation.

Garett-MacGowan:
Does a restarted workflow not get a new ID? @juliev0

Garett-MacGowan:
This seems to imply that `common.LabelKeyCompleted` isn't considered in the UI, right?

We can't mark common.LabelKeyCompleted before tasks results have reconciled.

I'll take a better look once I'm at my computer.

juliev0:
> Does a restarted workflow not get a new ID? @juliev0

A new name? As far as I can tell no, looking at the [code](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server.go#L383).

Garett-MacGowan:
@shuangkun Just to clarify, is the issue that user(s) may update workflows because they think they are fully complete, when in reality, the controller is still working towards reconciling all the outputs, thus leading to update conflicts if the user retries the workflow too early?

If you wait a sufficient amount of time for all outputs to be reconciled before resubmitting the workflow, does the issue still persist?

Garett-MacGowan:
@shuangkun do you also see `Failed to re-apply update`?
Do you happen to see `Error updating workflow with size error`?

I think the best way forward with this is to send a redacted workflow controller log file or create a test that reproduces the issue.

workflow/controller/operator.go
```diff
@@ -807,6 +812,16 @@ func (woc *wfOperationCtx) checkReconciliationComplete() bool {
...
 
+func (woc *wfOperationCtx) checkWorkflowComplete() bool {
+	woc.log.Debugf("Task results completion status: %v", woc.wf.Status.TaskResultsCompletionStatus)
+	return woc.wf.Labels[common.LabelKeyCompleted] == "true" && !woc.wf.Status.TaskResultsInProgress()
+}
+
+func (woc *wfOperationCtx) checkTaskResultsComplete() bool {
+	woc.log.Debugf("Task results completion status: %v", woc.wf.Status.TaskResultsCompletionStatus)
+	return !woc.wf.Status.TaskResultsInProgress()
+}
```
juliev0, line 823: we can't call this `checkTaskResultsComplete()` since this can be true when some TaskResults haven't started

workflow/controller/operator.go
```diff
@@ -2257,6 +2272,34 @@ func (woc *wfOperationCtx) checkTemplateTimeout(tmpl *wfv1.Template, node *wfv1.
...
+		if woc.wf.Status.Message != message {
+			woc.log.Infof("Updated message %s -> %s", woc.wf.Status.Message, message)
+			woc.wf.Status.Message = message
+		}
+		woc.wf.ObjectMeta.Labels[common.LabelKeyPhase] = string(phase)
+		woc.wf.Status.Phase = wfv1.WorkflowCompleting
+		if woc.wf.ObjectMeta.Labels == nil {
+			woc.wf.ObjectMeta.Labels = make(map[string]string)
+		}
+		woc.wf.ObjectMeta.Labels[common.LabelKeyPhase] = string(phase)
```
juliev0, line 2292: this line seems to be a few lines above as well

workflow/controller/operator.go
shuangkun: Ok, I have changed it.Thanks

Garett-MacGowan:
I see what you're doing here 👍

Garett-MacGowan:
I was hesitant to do this in my test attempt in #12595, but do you think it is possible to reproduce in a test without adding `time.Sleep(time.Second*2)` to `taskResultReconciliation` if we make a test workflow with a large number of outputs? e.g. 100 outputs, or however many it takes to get `taskResultReconciliation` to chug? Unfortunately this will be flakey if the test runners change (e.g if they run on faster CPUs in the future), but it might be better than nothing. It would be a balancing act between testing the edge case & not having the test take too long.

Thoughts @juliev0?

I'd like the tests to fail if someone changes how this works in the future.

shuangkun:
> I was hesitant to do this in my test attempt in #12595, but do you think it is possible to reproduce in a test without adding `time.Sleep(time.Second*2)` to `taskResultReconciliation` if we make a test workflow with a large number of outputs? e.g. 100 outputs, or however many it takes to get `taskResultReconciliation` to chug? Unfortunately this will be flakey if the test runners change (e.g if they run on faster CPUs in the future), but it might be better than nothing. It would be a balancing act between testing the edge case & not having the test take too long.
> 
> 
> 
> Thoughts @juliev0?
> 
> 
> 
> I'd like the tests to fail if someone changes how this works in the future.

I once wanted to do this too, but I found it is very difficult. I have tried to make the outputs more complex and large in number. But the processing speed is still very fast and it is difficult to control the timing of retry. It should only be possible to reproduce larger workflows in production, indeed. My problem is also in this scenario.

Garett-MacGowan:
> I once wanted to do this too, but I found it is very difficult. I have tried to make the outputs more complex and large in number. But the processing speed is still very fast and it is difficult to control the timing of retry. It should only be possible to reproduce larger workflows in production, indeed. My problem is also in this scenario.

That's unfortunate. Thanks for trying!

pkg/apis/workflow/v1alpha1/workflow_phase.go
```diff
@@ -4,17 +4,18 @@ package v1alpha1
...
-	WorkflowPending   WorkflowPhase = "Pending" // pending some set-up - rarely used
-	WorkflowRunning   WorkflowPhase = "Running" // any node has started; pods might not be running yet, the workflow maybe suspended too
-	WorkflowSucceeded WorkflowPhase = "Succeeded"
-	WorkflowFailed    WorkflowPhase = "Failed" // it maybe that the workflow was terminated
-	WorkflowError     WorkflowPhase = "Error"
+	WorkflowUnknown    WorkflowPhase = ""
+	WorkflowPending    WorkflowPhase = "Pending" // pending some set-up - rarely used
+	WorkflowRunning    WorkflowPhase = "Running" // any node has started; pods might not be running yet, the workflow maybe suspended too
+	WorkflowSucceeded  WorkflowPhase = "Succeeded"
+	WorkflowFailed     WorkflowPhase = "Failed" // it maybe that the workflow was terminated
```
juliev0, line 11:
```suggestion
	WorkflowRunning    WorkflowPhase = "Running" // any node has started; pods might not be running yet, the workflow may be suspended too
	WorkflowSucceeded  WorkflowPhase = "Succeeded"
	WorkflowFailed     WorkflowPhase = "Failed" // it may be that the workflow was terminated
```

juliev0: (I know you didn't write this, but would be nice if we can fix it anyway?)

pkg/apis/workflow/v1alpha1/workflow_phase.go
```diff
@@ -4,17 +4,18 @@ package v1alpha1
...
+	WorkflowError      WorkflowPhase = "Error"
+	WorkflowCompleting WorkflowPhase = "Completing"
```
juliev0, line 13:
```suggestion
	WorkflowCompleting WorkflowPhase = "Completing" // the Workflow completed running but hasn't been fully reconciled yet
```

pkg/apis/workflow/v1alpha1/workflow_phase.go
```diff
@@ -4,17 +4,18 @@ package v1alpha1
...
 )
 
 func (p WorkflowPhase) Completed() bool {
 	switch p {
-	case WorkflowSucceeded, WorkflowFailed, WorkflowError:
+	case WorkflowSucceeded, WorkflowFailed, WorkflowError, WorkflowCompleting:
```
juliev0, line 18: Did you go through all of the calls to `WorkflowPhase.Completed()` (including all of the calls to `WorkflowStatus.Fulfilled()` which calls it) to make sure all of them would be fine with this addition?

workflow/controller/operator.go
```diff
@@ -2365,6 +2365,12 @@ func (woc *wfOperationCtx) markWorkflowPhase(ctx context.Context, phase wfv1.Wor
 				}
 			}
 			woc.updated = true
+		} else {
+			woc.log.Info("Marking workflow completing")
+			woc.log.WithFields(log.Fields{"fromPhase": woc.wf.Status.Phase, "toPhase": wfv1.WorkflowCompleting}).
+				Debug("taskresults of workflow is incomplete or still has daemon nodes, so can't mark workflow completed")
```
juliev0, line 2371:
```suggestion
				Debug("taskresults of workflow are incomplete or still have daemon nodes, so can't mark workflow completed")
```

workflow/controller/operator.go
```diff
@@ -2365,6 +2365,12 @@ func (woc *wfOperationCtx) markWorkflowPhase(ctx context.Context, phase wfv1.Wor
 				}
 			}
 			woc.updated = true
+		} else {
+			woc.log.Info("Marking workflow completing")
+			woc.log.WithFields(log.Fields{"fromPhase": woc.wf.Status.Phase, "toPhase": wfv1.WorkflowCompleting}).
+				Debug("taskresults of workflow are incomplete or still have daemon nodes, so can't mark workflow completed")
+			woc.wf.Status.Phase = wfv1.WorkflowCompleting
+			woc.updated = true
```
juliev0, line 2373: just to note that there is a very tiny possibility that we could set the `woc.wf.Status.Phase` to Success/Failed/Error up on line 2299, and then the Controller could go down before it gets overwritten here but that may be a remote enough possibility that we not worry about it

juliev0: Well, I am kind of wondering if there's a bit of restructuring that can be done. The other thing I see is that we're setting these Events even though we haven't made it the official new Phase yet:
```
switch phase {
		case wfv1.WorkflowRunning:
			woc.eventRecorder.Event(woc.wf, apiv1.EventTypeNormal, "WorkflowRunning", "Workflow Running")
		case wfv1.WorkflowSucceeded:
			woc.eventRecorder.Event(woc.wf, apiv1.EventTypeNormal, "WorkflowSucceeded", "Workflow completed")
			woc.eventRecorder.Event(woc.wf, apiv1.EventTypeWarning, "WorkflowFailed", message)
		}
```

juliev0: Would you mind making a change so that we only set `woc.wf.Status.Phase = phase` and the Events when we should be?

pkg/apis/workflow/v1alpha1/workflow_phase.go
```diff
@@ -4,17 +4,18 @@ package v1alpha1
...
-	WorkflowSucceeded WorkflowPhase = "Succeeded"
-	WorkflowFailed    WorkflowPhase = "Failed" // it maybe that the workflow was terminated
-	WorkflowError     WorkflowPhase = "Error"
+	WorkflowUnknown    WorkflowPhase = ""
+	WorkflowPending    WorkflowPhase = "Pending" // pending some set-up - rarely used
+	WorkflowRunning    WorkflowPhase = "Running" // any node has started; pods might not be running yet, the workflow may be suspended too
+	WorkflowSucceeded  WorkflowPhase = "Succeeded"
+	WorkflowFailed     WorkflowPhase = "Failed" // it may be that the workflow was terminated
+	WorkflowError      WorkflowPhase = "Error"
+	WorkflowCompleting WorkflowPhase = "Completing" // the Workflow completed running but hasn't been fully reconciled yet
```
agilgur5, line 13: So adding a new `phase` requires changes to _many_ places. That's one reason why we have yet to implement a new `status` like `Cancelled` (see #11849), which has very similar ramifications.

For instance, there are no UI changes in this PR currently, but this does [affect how the UI displays Workflows](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/ui/src/app/shared/utils.ts#L10) -- I don't even know what icon should represent "Completing" as it's not a common UX state.

I did see that there was discussion on whether this should even be user-facing in #12615, as it seems that this state can be derived without a new `phase`

juliev0: thanks for pointing this out @agilgur5

agilgur5: If Go had algebraically typed `match` statements, many missed would be picked up by type-checker, but we also have `default`s or `else`s in many places that don't necessarily apply to "Completing" (the UI one is one of those).

shuangkun:
@juliev0  Can you give me some advice? Thanks！
Here are three possible ways to resolve this issue.
1) Limit during retry. If the workflow is Faild but Completed is not true, an error will be returned.
2) Add a label to describe the workflow from Failed to Completed. For example workflows.argoproj.io/completing: true
3) Add phase to describe the workflow from Failed to Completed. For example phase Completing
Which one do you think is better?

juliev0:
> @juliev0 Can you give me some advice? Thanks！ Here are three possible ways to resolve this issue.
> 
> 1. Limit during retry. If the workflow is Faild but Completed is not true, an error will be returned.
> 2. Add a label to describe the workflow from Failed to Completed. For example workflows.argoproj.io/completing: true
> 3. Add phase to describe the workflow from Failed to Completed. For example phase Completing
>    Which one do you think is better?

So, the 3rd one is the one you have now, right? (But then you'd have to make sure this phase is recognized everywhere in the UI, which sounds like it could be challenging.)

What you have now essentially makes it so that LabelKeyCompleted happens at the same time as the phase being in a Completed state, right? I think it would be good to keep that aspect of your change. Does your # 2 help you to achieve that?

shuangkun:
> > @juliev0 Can you give me some advice? Thanks！ Here are three possible ways to resolve this issue.
> > 
> > 1. Limit during retry. If the workflow is Faild but Completed is not true, an error will be returned.
> > 2. Add a label to describe the workflow from Failed to Completed. For example workflows.argoproj.io/completing: true
> > 3. Add phase to describe the workflow from Failed to Completed. For example phase Completing
> >    Which one do you think is better?
> 
> So, the 3rd one is the one you have now, right? (But then you'd have to make sure this phase is recognized everywhere in the UI, which sounds like it could be challenging.)
> 
> What you have now essentially makes it so that LabelKeyCompleted happens at the same time as the phase being in a Completed (Failed, Succeeded, Error) state, right? I think it would be good to keep that aspect of your change. Does your # 2 help you to achieve that?

Yes，make sure this phase is recognized everywhere in the UI looks like it could be challenging， I will try it.
2rd may also can achieve,   WorkflowRunning + workflows.argoproj.io/completing: true = WorkflowComleting，But I don't know if it's reasonable. This results in the phase of the workflow not showing the real status.

juliev0:
> But I don't know if it's reasonable. This results in the phase of the workflow not showing the real status.

Sorry, I may not know what you mean. Do you mean for the period of time when the workflow has technically completed but the results haven't been reconciled, then the workflow will appear as "Running"? Is it a problem? I'm wondering if it's reasonable to call it "Running" if it's not fully reconciled? (I'm not a heavy Workflows user so I may be missing something)

agilgur5:
> Here are three possible ways to resolve this issue.

I am wondering if https://github.com/argoproj/argo-workflows/issues/12615#issuecomment-1926854738 is possible; that would limit the amount of new labels/phases etc added.

> I'm wondering if it's reasonable to call it "Running" if it's not fully reconciled?

I think that would be reasonable IMO. A bit confusing, but not very. It wouldn't last too long and reconciliation is a normal part of operator logic -- though admittedly not all users are aware of that second part.

shuangkun:
> > But I don't know if it's reasonable. This results in the phase of the workflow not showing the real status.
> 
> Sorry, I may not know what you mean. Do you mean for the period of time when the workflow has technically completed but the results haven't been reconciled, then the workflow will appear as "Running"? Is it a problem? I'm wondering if it's reasonable to call it "Running" if it's not fully reconciled? (I'm not a heavy Workflows user so I may be missing something)

I think users won’t think this is a problem, only we need to pay attention to this when dealing with it.

juliev0:
> > > Here are three possible ways to resolve this issue.
> > 
> > 
> > I am wondering if [#12615 (comment)](https://github.com/argoproj/argo-workflows/issues/12615#issuecomment-1926854738) is possible; that would limit the amount of new labels/phases etc added.
> 
> I think it is possible. But if we record the workflow phase use labels or workflow phase, it will help us process workflow in next reconcile（artifacts gc、execTemplates、or any other need to care about the workflow fullfilled）.

I too like the idea of this kind of simplicity. But I may not understand the complexities. Are you basically saying that you'll be able to process certain things sooner than later? Would you still be able to process those things eventually?

shuangkun:
Do you mean that you also hope that labelcompleted and completed status can occur at the same time without adding any label or phase? Anyway, I will try to solve this problem as soon as possible.

shuangkun:
> process 

Do you mean that you also hope that labelcompleted and completed status can occur at the same time without adding any label or phase? Anyway, I will try to solve this problem as soon as possible.

shuangkun:
> > > > Here are three possible ways to resolve this issue.
> 
> > > 
> 
> > > 
> 
> > > I am wondering if [#12615 (comment)](https://github.com/argoproj/argo-workflows/issues/12615#issuecomment-1926854738) is possible; that would limit the amount of new labels/phases etc added.
> 
> > 
> 
> > I think it is possible. But if we record the workflow phase use labels or workflow phase, it will help us process workflow in next reconcile（artifacts gc、execTemplates、or any other need to care about the workflow fullfilled）.
> 
> 
> 
> I too like the idea of this kind of simplicity. But I may not understand the complexities. Are you basically saying that you'll be able to process certain things sooner than later? Would you still be able to process those things eventually?

Do you mean that you also hope that labelcompleted and completed status can occur at the same time without adding any label or phase? Anyway, I will try to solve this problem as soon as possible.

juliev0:
> > > > > Here are three possible ways to resolve this issue.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > I am wondering if [#12615 (comment)](https://github.com/argoproj/argo-workflows/issues/12615#issuecomment-1926854738) is possible; that would limit the amount of new labels/phases etc added.
> > 
> > 
> > > 
> > 
> > 
> > > I think it is possible. But if we record the workflow phase use labels or workflow phase, it will help us process workflow in next reconcile（artifacts gc、execTemplates、or any other need to care about the workflow fullfilled）.
> > 
> > 
> > I too like the idea of this kind of simplicity. But I may not understand the complexities. Are you basically saying that you'll be able to process certain things sooner than later? Would you still be able to process those things eventually?
> 
> Do you mean that you also hope that labelcompleted and completed status can occur at the same time without adding any label or phase? Anyway, I will try to solve this problem as soon as possible.

That would be great, if it's not too difficult. I'm gathering the concern is that [WorkflowPhase.Completed()](https://github.com/argoproj/argo-workflows/pull/12574/files#diff-77c9e61c13a2de8af46df3f5f6d159bbd754bb1698fbe073531efa1f055b0cf8R16) would no longer be true in the next reconciliation? (and that gets called from a bunch of places) So, I guess the question is just: would it at least eventually be true for a subsequent reconciliation and clean everything up properly?

Or is there a concern that there's some call to `WorkflowPhase.Completed()` that would return false and then do something that presumes the Workflow is actually running that would be harmful?

shuangkun:
> > > > > > Here are three possible ways to resolve this issue.
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > > I am wondering if [#12615 (comment)](https://github.com/argoproj/argo-workflows/issues/12615#issuecomment-1926854738) is possible; that would limit the amount of new labels/phases etc added.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > I think it is possible. But if we record the workflow phase use labels or workflow phase, it will help us process workflow in next reconcile（artifacts gc、execTemplates、or any other need to care about the workflow fullfilled）.
> > > 
> > > 
> > > I too like the idea of this kind of simplicity. But I may not understand the complexities. Are you basically saying that you'll be able to process certain things sooner than later? Would you still be able to process those things eventually?
> > 
> > 
> > Do you mean that you also hope that labelcompleted and completed status can occur at the same time without adding any label or phase? Anyway, I will try to solve this problem as soon as possible.
> 
> That would be great, if it's not too difficult. I'm gathering the concern is that [WorkflowPhase.Completed()](https://github.com/argoproj/argo-workflows/pull/12574/files#diff-77c9e61c13a2de8af46df3f5f6d159bbd754bb1698fbe073531efa1f055b0cf8R16) would no longer be true in the next reconciliation? (and that gets called from a bunch of places) So, I guess the question is just: would it at least eventually be true for a subsequent reconciliation and clean everything up properly?
> 
> Or is there a concern that there's some call to `WorkflowPhase.Completed()` that would return false and then do something that presumes the Workflow is actually running that would be harmful?

Yes, there may be some problems here. Completed status is used in many places. I will try to go in this direction first.

juliev0:
@Garett-MacGowan Since you've also worked on shutdown logic, wanted to also give you an opportunity to chime in if you see challenges as far as the idea of simply just not setting Phase to any kind of Completed/Completing state until the time when we set LabelKeyCompleted. Basically, any time that [WorkflowPhase.Completed()](https://github.com/argoproj/argo-workflows/pull/12574/files#diff-77c9e61c13a2de8af46df3f5f6d159bbd754bb1698fbe073531efa1f055b0cf8R16) is called, it would return false until that time.

juliev0:
> > @Garett-MacGowan Since you've also worked on shutdown logic, wanted to also give you an opportunity to chime in if you see challenges as far as the idea of simply just not setting Phase to any kind of Completed/Completing state until the time when we set LabelKeyCompleted. Basically, any time that [WorkflowPhase.Completed()](https://github.com/argoproj/argo-workflows/pull/12574/files#diff-77c9e61c13a2de8af46df3f5f6d159bbd754bb1698fbe073531efa1f055b0cf8R16) is called, it would return false until that time.
> 
> It's been a bit. Took a quick look just now.
> 
> Do you mean adding a condition to `func (p WorkflowPhase) Completed() bool ` that checks if `woc.wf.ObjectMeta.Labels[common.LabelKeyCompleted] == "true"`?
> 
> If so, we need to change below to keep old Phase.Completed() functionality.
> 
> ```
> func (woc *wfOperationCtx) checkReconciliationComplete() bool {
> 	woc.log.Debugf("Task results completion status: %v", woc.wf.Status.TaskResultsCompletionStatus)
> 	return woc.wf.Status.Phase.Completed() && !woc.wf.Status.TaskResultsInProgress()
> }
> ```
> 
> Every other use of `Completed()` can require `woc.wf.ObjectMeta.Labels[common.LabelKeyCompleted] == "true"`, I think...
> 
> Maybe we could copy the existing `Completed()` function, give it a new name `Completing()` and then modify and add the `LabelKeyCompleted` requirement to `Completed()`?
> 
> ```
> func (p WorkflowPhase) Completing() bool {
> 	switch p {
> 	case WorkflowSucceeded, WorkflowFailed, WorkflowError:
> 		return true
> 	default:
> 		return false
> 	}
> }
> ```

What I meant was keeping the `Completed()` function as it is in master:  only returning true for phase={WorkflowSucceeded, WorkflowFailed, WorkflowError}. But basically holding off on setting phase={WorkflowSucceeded, WorkflowFailed, WorkflowError} until that time when we set LabelKeyCompleted=true (when all task results have been reconciled, etc). So, essentially the phase would still be "Running" for a little extra longer.

juliev0:
Oh, now i see the problem with what I said. 

1. `checkReconciliationComplete()` is dependent on `woc.wf.Status.Phase.Completed()`.
```
func (woc *wfOperationCtx) checkReconciliationComplete() bool {
	return woc.wf.Status.Phase.Completed() && !woc.wf.Status.TaskResultsInProgress()
}
```
(We had to include `woc.wf.Status.Phase.Completed()` in there to know that the Workflow was in fact complete; otherwise we had no way of knowing that all TaskResults had necessarily finished, due to the fact that there could be other TaskResults in the future. )

2. And we only set LabelKeyCompleted=true once `checkReconciliationComplete()`

juliev0:
@Garett-MacGowan sorry I didn't read your [message](https://github.com/argoproj/argo-workflows/pull/12574#issuecomment-1974163559) thoroughly before responding. This is an interesting solution you've proposed. Maybe that can work.

workflow/controller/operator.go
juliev0: Why did you end up putting this back? (per previous comment)

juliev0:
This looks like a nice clean solution to the problem as far as I can tell so far. 
Only once the workflow has ended and its reconciliation tasks are done, the Status.Phase changes to a Completed state, and its LabelKeyCompleted is set. 

I will dig deeper when I get a chance. Thanks!

workflow/controller/operator.go
```diff
@@ -817,6 +814,11 @@ func (woc *wfOperationCtx) checkReconciliationComplete() bool {
 	return woc.wf.Status.Phase.Completed() && !woc.wf.Status.TaskResultsInProgress()
 }
 
+func (woc *wfOperationCtx) checkTaskResultsComplete() bool {
+	woc.log.Debugf("Task results completion status: %v", woc.wf.Status.TaskResultsCompletionStatus)
+	return !woc.wf.Status.TaskResultsInProgress()
+}
```
juliev0, line 820: I was always a little reluctant to call this `checkTaskResultsComplete()` since it could imply that all TaskResults that would ever exist are complete. So, my preference was to say `checkTaskResultsInProgress()` and flip the boolean. Would it be okay to do that?

shuangkun:
> This looks like a nice clean solution to the problem as far as I can tell so far. Only once the workflow has ended and its reconciliation tasks are done, the Status.Phase changes to a Completed state, and its LabelKeyCompleted is set.
> 
> I will dig deeper when I get a chance. Thanks!

Thanks, I have solved the related errors and look forward to your review

workflow/controller/operator.go
Joibel: In the event of controller death none of the changes to wf.Status.Phase will make it out of the controller until Status is persisted to kubernetes at the end of reconciliation, in [persistUpdates](https://github.com/argoproj/argo-workflows/blob/d805b7fa6c36eac63d3dc24a84f2348a48f66be7/workflow/controller/operator.go#L718).

Events are queued, so when they are exposed to the outside world is also uncertain in a controller death scenario.

It does feel like this function is quite cognitively heavy, so attempting to make it simpler to understand is useful though.

workflow/controller/operator.go
```diff
@@ -1704,7 +1706,7 @@ func (woc *wfOperationCtx) deletePVCs(ctx context.Context) error {
 
 	switch gcStrategy {
 	case wfv1.VolumeClaimGCOnSuccess:
-		if woc.wf.Status.Phase == wfv1.WorkflowError || woc.wf.Status.Phase == wfv1.WorkflowFailed {
+		if woc.wf.Status.Phase == wfv1.WorkflowError || woc.wf.Status.Phase == wfv1.WorkflowFailed || woc.wf.Status.Phase == wfv1.WorkflowRunning {
```
juliev0, line 1709: this makes me wonder why the original logic wasn't just simply `woc.wf.Status.Phase != wfv1.WorkflowSucceeded`

juliev0: I guess the original function was written with the assumption that it would only ever be called if the Workflow were in fact Completed. If we want to make it safer to call this even when the Workflow isn't completed, then we should also add a check to the `wfv1.VolumeClaimGCOnCompletion` case.

workflow/controller/operator.go
```diff
@@ -513,7 +510,7 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 		woc.markWorkflowError(ctx, err)
 	}
 
-	if woc.execWf.Spec.Metrics != nil {
+	if woc.execWf.Spec.Metrics != nil && woc.wf.Status.Fulfilled() {
```
juliev0, line 513: nit: If it were previously assumed that this line until the end of the function should only be invoked after everything was officially completed, then perhaps we just add a line right above here to terminate early if we're done instead of having the condition here. What do you think? This function is so long and windy that I find myself looking for any way to improve readability and maintainability. :)

juliev0:
You know, I'm seeing we may no longer need this method:
```
func (woc *wfOperationCtx) checkReconciliationComplete() bool {
	woc.log.Debugf("Task results completion status: %v", woc.wf.Status.TaskResultsCompletionStatus)
	return woc.wf.Status.Phase.Completed() && !woc.wf.Status.TaskResultsInProgress()
}
```

If `woc.wf.Status.Phase.Completed()`, then it's impossible for `woc.wf.Status.TaskResultsInProgress()`, right? So, should we remove this method and replace any calls with a simple check to see if Status.Phase is completed?

shuangkun:
> checkReconciliationComplete

OK, I remove it.

juliev0 (approved):

juliev0:
This looks great! Thanks for going through the iterations. 

Before merging, @Garett-MacGowan would you mind taking a look as well? Now only once the workflow has ended and its reconciliation tasks are done, the Status.Phase changes to a Completed state, and its LabelKeyCompleted is set. Given that, `checkReconciliationComplete()` was no longer needed as a method per this [comment](https://github.com/argoproj/argo-workflows/pull/12574#issuecomment-1984961640) and was removed.

Garett-MacGowan:
Looks good to me! Thanks @shuangkun & @juliev0

agilgur5:
Thanks everyone for all the efforts here! This iteration looks simpler and cleaner too

agilgur5:
Backported cleanly to `release-3.5` as 68c089d49

PR #12575, opened 2024-01-25
fix: set burst and qps for auth.kubeclient

shuangkun:
When I submit or access a large number of workflows at the same time in argo-server URL，I encountered many client-side throttling  which let my request slow， I find the root cause is the client in workflow_server.go 's burst and qps is 0. So I want to increase it. After I added it, it was fine. There is no throttling.
```
wfClient := auth.GetWfClient(ctx)
kubeClient := auth.GetKubeClient(ctx)
```
some log in argo server

```
I0123 08:01:37.531986       1 request.go:601] Waited for 1.196145076s due to client-side throttling, not priority and fairness, request: DELETE:https://192.168.1.138:6443/api/v1/namespaces/default/pods/steps-bszqb-whalesay-600069335
I0123 08:01:47.731411       1 request.go:601] Waited for 11.379013988s due to client-side throttling, not priority and fairness, request: DELETE:https://192.168.1.138:6443/api/v1/namespaces/default/pods/steps-bszqb-whalesay-4291585860
I0123 08:01:57.930750       1 request.go:601] Waited for 21.53704195s due to client-side throttling, not priority and fairness, request: DELETE:https://192.168.1.138:6443/api/v1/namespaces/default/pods/steps-bszqb-whalesay-4209244865
I0123 08:02:07.930913       1 request.go:601] Waited for 31.536611828s due to client-side throttling, not priority and fairness, request: DELETE:https://192.168.1.138:6443/api/v1/namespaces/default/pods/steps-bszqb-whalesay-163250805
```


server/auth/gatekeeper.go
```diff
@@ -286,7 +286,12 @@ func (s *gatekeeper) getClientsForServiceAccount(ctx context.Context, claims *ty
 	if err != nil {
 		return nil, err
 	}
-	_, clients, err := s.clientForAuthorization(authorization)
+	var clients *servertypes.Clients
+	if s.restConfig != nil {
+		_, clients, err = s.clientForAuthorization(authorization, s.restConfig.QPS, s.restConfig.Burst)
+	} else {
+		_, clients, err = s.clientForAuthorization(authorization, 5, 10)
```
tooptoop4, line 293: is 5, 10 lower than existing default?

shuangkun: I print the default value，it is 0. In kubernetes, if the config's burst and qps valus is 0, it will be changed to 10,5.

agilgur5: can you leave a comment with a permalink to the source of that? that way we know where it was taken from and where to find the default if it is updated in the future.

though ideally we'd fallback to the default implicitly instead of needing to copy it ourselves.

agilgur5:
For reference for other readers, this PR takes the burst and QPS from [the Server's configuration in the CLI](https://github.com/argoproj/argo-workflows/blob/a15587755e250fbbe3b9538b1c15ae581a108f0c/cmd/argo/commands/server.go#L82) and passes it through to the `auth.kubeclient` of the Server. Notably [the defaults there](https://github.com/argoproj/argo-workflows/blob/a15587755e250fbbe3b9538b1c15ae581a108f0c/cmd/argo/commands/server.go#L237) are 20 QPS, 30 burst, so this does also implicitly increase the defaults of the `auth.kubeclient` 

I think that makes sense to do, but it wasn't clear from the PR where the `restConfig` values come from

server/auth/gatekeeper.go
```diff
@@ -337,11 +342,13 @@ func addClaimsLogFields(claims *types.Claims, fields log.Fields) log.Fields {
 	return fields
 }
 
-func DefaultClientForAuthorization(authorization string) (*rest.Config, *servertypes.Clients, error) {
+func DefaultClientForAuthorization(authorization string, qps float32, burst int) (*rest.Config, *servertypes.Clients, error) {
```
agilgur5, line 345: this is the "default" client, so I don't think it should be configurable, by definition

agilgur5: oh I see, this won't type-check otherwise (or necessarily work properly).... hmmm... I feel like we shouldn't have two `restConfig`s, which would simplify this. we can merge them, although that doesn't quite fix the type-check issue here... I wonder if the merge should be done in `server.go` instead... 🤔

agilgur5: at minimum, I think it would make sense for this to accept a `rest.Config` instead of `qps` and `burst` in isolation. then we just pass through `s.restConfig`.

whether that actually makes sense for a "default" to have a configuration could be refactored later, I suppose. I need to look at the initialization chain a bit more

server/auth/gatekeeper.go
```diff
@@ -286,7 +286,12 @@ func (s *gatekeeper) getClientsForServiceAccount(ctx context.Context, claims *ty
 	if err != nil {
 		return nil, err
 	}
-	_, clients, err := s.clientForAuthorization(authorization)
+	var clients *servertypes.Clients
+	if s.restConfig != nil {
```
agilgur5, line 290: is there a reason you only checked this condition in `getClientsForServiceAccount` and not in `getClients` as well?

I think effectively it will never be `nil` except in potentially a test case though

server/auth/gatekeeper.go
agilgur5: ok it seems like the `clientForAuthorization` is only exposed for testing purposes, otherwise the default is always used, so yea I guess we'd merge `restConfig`s here

agilgur5:
> this PR takes the burst and QPS from [the Server's configuration in the CLI](https://github.com/argoproj/argo-workflows/blob/a15587755e250fbbe3b9538b1c15ae581a108f0c/cmd/argo/commands/server.go#L82) and passes it through to the auth `kubeClient` of the Server

If this is just pass through, this might make sense as a `fix` rather than a `feat`, since one would think it would pass through already

server/auth/gatekeeper.go
shuangkun: Yes，it won't. Because a UT，maybe I can change the UT

shuangkun:
https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/server.go 
'''
config.Burst = kubeAPIBurst
config.QPS = kubeAPIQPS
'''

shuangkun:
> > this PR takes the burst and QPS from [the Server's configuration in the CLI](https://github.com/argoproj/argo-workflows/blob/a15587755e250fbbe3b9538b1c15ae581a108f0c/cmd/argo/commands/server.go#L82) and passes it through to the auth `kubeClient` of the Server
> 
> 
> 
> If this is just pass through, this might make sense as a `fix` rather than a `feat`, since one would think it would pass through already

OK i will change it

server/auth/gatekeeper.go
agilgur5: @shuangkun you might've misunderstood what I said here (that's partly my fault, I had some scattered thoughts) -- I think it would make more sense to accept a `rest.Config` instead of a `qps` and `burst` in isolation

shuangkun: Oh, do you mean this
```
type ClientForAuthorization func(authorization string, config *rest.Config) (*rest.Config, *servertypes.Clients, error)
```
I have change it.

agilgur5: Yea that's what I meant, although I also meant merging the entirety of the `rest.Config` as well, see below comment

server/auth/gatekeeper.go
```diff
@@ -337,11 +337,13 @@ func addClaimsLogFields(claims *types.Claims, fields log.Fields) log.Fields {
...
 }
 
-func DefaultClientForAuthorization(authorization string) (*rest.Config, *servertypes.Clients, error) {
+func DefaultClientForAuthorization(authorization string, config *rest.Config) (*rest.Config, *servertypes.Clients, error) {
 	restConfig, err := kubeconfig.GetRestConfig(authorization)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to create REST config: %w", err)
 	}
+	restConfig.QPS = config.QPS
+	restConfig.Burst = config.Burst
```
agilgur5, line 346: I mentioned above, but I think it might be better to merge the configs together. For instance, the Server's `rest.Config` also has [a user agent](https://github.com/argoproj/argo-workflows/blame/42262690fcbab0ed6106d3f9069e128655b3339f/cmd/argo/commands/server.go#L81).

Those are the only 3 configurations right now, so maybe 1 line for each is fine, but a merge could be more resilient to change if other configurations are added to the Server

server/auth/gatekeeper.go
```diff
@@ -370,3 +371,10 @@ func DefaultClientForAuthorization(authorization string) (*rest.Config, *servert
 		Kubernetes:  kubeClient,
 	}, nil
 }
+
+func mergeServerRestConfig(argoServerConfig *rest.Config, newConfig *rest.Config) *rest.Config {
```
agilgur5, line 375: oh I meant using a struct merge (which may require a library), but this is fine too

server/auth/gatekeeper.go
```diff
@@ -370,3 +371,10 @@ func DefaultClientForAuthorization(authorization string) (*rest.Config, *servert
...
+	newConfig.Burst = argoServerConfig.Burst
+	newConfig.QPS = argoServerConfig.QPS
+	// TO DO: Merge other common configurations，such as RateLimiter.
```
agilgur5, line 378: the user agent was the other one I [mentioned](https://github.com/argoproj/argo-workflows/pull/12575#discussion_r1473515842) from the Server. that's the only other one currently I think, no?

agilgur5 (approved):

shuangkun:
Thanks!

PR #12577, opened 2024-01-26
docs: add an example for nested `withSequence`

shuangkun:
Add an exampel for withSequence recursion.

A user want to know how to combine recursion and a withSequence in a Workflow to achieve this. I searched with sequence and didn't find similar, so I wrote it.
image


examples/hello-recursion.yaml
```diff
@@ -0,0 +1,43 @@
+# The following workflow executes a workflow which the number of next level depends on the output of the previous step.
```
agilgur5, line 1:
```suggestion
# This example shows how to use recursion in a Workflow.
# The number of executions of the next step depends on the output of the previous step.
```

- "the following workflow executes a workflow" is redundant, be more simple and direct [per k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- "level" is not a term in Workflows and "number of next level" is missing a concrete subject -- let's use the word "executions" here

examples/hello-recursion.yaml
```diff
@@ -0,0 +1,43 @@
...
+# A is the first step. After a is executed, b is executed (b generates n b based on the result of a),
+# and after b is executed, c is executed (c generates n c based on the result of each b).
```
agilgur5, line 3:
```suggestion
# A is the first step. A's output determines how many times B is executed.
# B's output then determines how many times C is executed.
```

- more simple and direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - no parentheses, no "n"
- consistently use the same capitalization (capital A, B, C)
- fix run-on sentence

agilgur5 (commented):
I made some simplifications to the example in-line below.

Otherwise, I'm not sure we really need this example? There is already [a recursion example in the docs](https://argo-workflows.readthedocs.io/en/release-3.5/walk-through/recursion/) and I think that one is simpler to understand as well (since it's a boolean rather than a number).

We could keep this as a `withSequence` + recursion example, but we may want to rename it then

> exampel

Also you have a typo in your PR title and description: "exampel" -> "example"

shuangkun:
Yes, withsequence -recursion is better！ This question came from my client, I searched with sequence and didn't find anything similar, so I wrote it

examples/withsequence-recursion.yaml
```diff
@@ -0,0 +1,43 @@
+# This example shows how to use recursion in a Workflow.
```
agilgur5, line 1:
```suggestion
# This example shows how to combine recursion and a withSequence loop in a Workflow.
```

since you modified the name now, we should probably modify the comment a little

agilgur5 (commented):
Just one small change since the name was modified, otherwise LGTM 🙂

agilgur5:
> This question came from my client, I searched with sequence and didn't find anything similar, so I wrote it

By the way, this is good context to have. I would encourage writing some of this rationale in your PR descriptions as it helps to understand the background. Same way that I might link to a Slack conversation from a user in one of my PRs

agilgur5 (approved):

juliev0:
Is it considered recursion? I thought I would see a template invoke itself or a template invoke something else which invoked itself, but I don't. Sorry, I may be missing something.

shuangkun:
> Is it considered recursion? I thought I would see a template invoke itself or a template invoke something else which invoked itself, but I don't. Sorry, I may be missing something.

It feels a bit like it. I can't think of a more suitable word to describe it right now.

agilgur5:
> Is it considered recursion?

Oh, you're right, it's not. I didn't even pay attention, good catch!

All the templates eventually call `hello`, so indeed there is no recursion or mutual recursion

agilgur5:
> It feels a bit like it. I can't think of a more suitable word to describe it right now.

Yea it is certainly a bit "meta" to use the outputs of a step to determine how many times to run the next step, although I also can't quite think of a term for that off the top of my head (if there is one) 🤔

agilgur5:
Simplifying this in code I think makes it easier to understand:

```python
import random

def entrypoint():
  for i in range(hello()):
    for j in range(hello()):
      hello()
    for j in range(hello()):
      hello()

def hello():
  result = random.randint(0,5)
  print(result)
  return result
```

In Argo's case there is a potentially interesting nuance that `stdout` is used as the `result` of a template, but otherwise it isn't as interesting in code

agilgur5:
Looking at the code makes it a bit easier to identify -- maybe we just call this `withSequence-nested`?

agilgur5:
> May be it need a withSequence example.

You're right it is missing a `withSequence` example.
I'm not sure about including this one there however, as the walk through should have primarily simple examples, while this one takes a bit to comprehend.

shuangkun:
> > May be it need a withSequence example.
> 
> 
> 
> You're right it is missing a `withSequence` example.
> 
> I'm not sure about including this one there however, as the walk through should have primarily simple examples, while this one takes a bit to comprehend.

I add a simple example, can you take a look? Thanks

examples/withsequence-nested-result.yaml
```diff
@@ -0,0 +1,43 @@
+# This example shows how to use recursion in a Workflow.
```
agilgur5, line 1:
```suggestion
# This example shows how to nested withSequence loops in a Workflow.
```

Since this isn't recursion, the description should be rewritten as well.

docs/walk-through/loops.md
```diff
@@ -2,12 +2,13 @@
...
 When writing workflows, it is often very useful to be able to iterate over a set of inputs, as this is how argo-workflows can perform loops.
 
-There are two basic ways of running a template multiple times.
+There are three basic ways of running a template multiple times.
 
 - `withItems` takes a list of things to work on. Either
     - plain, single values, which are then usable in your template as '{{item}}'
     - a JSON object where each element in the object can be addressed by it's key as '{{item.key}}'
 - `withParam` takes a JSON array of items, and iterates over it - again the items can be objects like with `withItems`. This is very powerful, as you can generate the JSON in another step in your workflow, so creating a dynamic workflow.
+- `withSequence` take a Sequence, and convert it into a numeric sequence.
```
agilgur5, line 11: this is arguably simpler than `withItems`, so maybe we should move this to be the first of the three in this list and in the examples below.

docs/walk-through/loops.md
```diff
@@ -2,12 +2,13 @@
...
 
 ## `withItems` basic example
```
agilgur5, line 13: on line 15 below (GitHub won't let me comment on it, unfortunately), `withItems` would no longer be the simplest, so that line should be reworded.
k8s style guide also says to [avoid words that assume a level of understanding](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding), so it shouldn't say "simplest" at all (nor "basic").

I'd remove the first sentence entirely and simplify the second one (which is not in [present tense](https://kubernetes.io/docs/contribute/style/style-guide/#use-present-tense), [active voice](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice) etc)

docs/walk-through/loops.md
agilgur5: also the definition you gave `withSequence` is currently recursive "Sequence [...] convert [...] sequence". it also has grammar and casing issues.

```suggestion
- `withSequence` iterates over a sequence of numbers.
```

- also simpler and more direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

docs/walk-through/loops.md
```diff
@@ -177,6 +178,32 @@ spec:
       args: ["echo sleeping for {{inputs.parameters.seconds}} seconds; sleep {{inputs.parameters.seconds}}; echo done"]
 ```
 
+## `withSequence` example
```
agilgur5, line 181: should probably make this the first example, per above comment

docs/walk-through/loops.md
```diff
@@ -177,6 +178,32 @@ spec:
...
+
+This example run a simple template multiple times using `withSequence`.
```
agilgur5, line 183:
```suggestion
This runs a template multiple times using `withSequence`.
```

- avoid redundant words, simpler and more direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- avoid assuming a level with "simple", per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding)

docs/walk-through/loops.md
```diff
@@ -177,6 +178,32 @@ spec:
...
+    steps:
+    - - name: run-pod
+        template: run-pod
+        withSequence:
+          count: "5"
+
+  - name: run-pod
+    container:
+      image: argoproj/argosay:v2
+      args: [sleep, 1s]
```
agilgur5, line 204: this should probably do something other than sleep as an example. the output should confirm that the sequence is running. for instance have it print "hello world". can use `whalesay` like the `withItems` example

docs/walk-through/loops.md
```diff
@@ -177,6 +178,32 @@ spec:
...
+metadata:
+  generateName: loop-sequence-
+spec:
+  entrypoint: loop-sequence-example
+
+  templates:
+  - name: loop-sequence-example
+    steps:
+    - - name: run-pod
+        template: run-pod
```
agilgur5, line 197: this should have more specific naming than `run-pod`. if the command is changed to print "hello world", then this should be "hello-world" or similar
```suggestion
    - - name: hello-world-x5
        template: hello-world
```

docs/walk-through/loops.md
shuangkun: OK, I move it to first

docs/walk-through/loops.md
```diff
@@ -2,13 +2,41 @@
...
+
+  templates:
+  - name: loop-sequence-example
+    steps:
+    - - name: hello-world-x5
+        template: hello-world
+        withSequence:
+          count: "5"
+
+  - name: run-pod
```
agilgur5, line 33:
```suggestion
  - name: hello-world
```

the template name was renamed above (per my [previous comment](https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1495375095)), so should be renamed here as well

docs/walk-through/loops.md
```diff
@@ -2,13 +2,41 @@
...
+    container:
+      image: argoproj/argosay:v2
+      command: ["sh", -c]
+      args: ["echo hello world!"]
+```
+
 ## `withItems` basic example
 
 This example is the simplest. We are taking a basic list of items and iterating over it with `withItems`. It is limited to one varying field for each of the workflow templates instantiated.
```
agilgur5, line 42:
```suggestion
This iterates over a list of items with `withItems`, substituting a string for each instantiated template.
```

this line was what I mentioned in my [previous comment](https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1495365093). it's no longer the "simplest example" and there are other changes that should be made here according to the style guide

examples/withsequence-nested-result.yaml
```diff
@@ -0,0 +1,43 @@
+# This example shows how to nested withSequence loops in a Workflow.
```
agilgur5, line 1:
```suggestion
# This example shows how to nest withSequence loops in a Workflow.
```

sorry I had a typo in my [previous comment](https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1495359070), fixed here

docs/walk-through/loops.md
```diff
@@ -2,13 +2,41 @@
...
+    - - name: hello-world-x5
+        template: hello-world
+        withSequence:
+          count: "5"
+
+  - name: run-pod
+    container:
+      image: argoproj/argosay:v2
+      command: ["sh", -c]
+      args: ["echo hello world!"]
```
agilgur5, line 37:
```suggestion
      image: docker/whalesay:latest
      command: [cowsay]
      args: ["hello world!"]
```

I mentioned using `whalesay` in my [previous comment](https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1495373464) for consistency with the `withItems` example.

Also `argosay` prints things (it uses [`cowsay`](https://github.com/argoproj/argo-workflows/blob/a68ea0feabc87c09d5e13d12e6f0d1a61adc5b16/test/e2e/images/argosay/v1/Dockerfile#L17), so it doesn't need a `sh -c echo`.

agilgur5 (commented):
A few small follow-ups, otherwise LGTM

docs/walk-through/loops.md
agilgur5: Follow-up comment: https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1498165392

examples/withsequence-nested-result.yaml
agilgur5: I had a typo here, sorry. Follow-up comment: https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1498172887

docs/walk-through/loops.md
agilgur5: This was about a different line, see follow-up comment: https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1498170713

docs/walk-through/loops.md
agilgur5: Follow-up to use `whalesay` for consistency: https://github.com/argoproj/argo-workflows/pull/12577#discussion_r1498185891

agilgur5 (approved):

shuangkun:
Thank you very much！

PR #7114, opened 2021-11-01
docs(ui): document wftemplate enum dropdown. Fixes #6824

tczhao:
Fixes #6824

Make the workflow template UI dropdown selection example more obvious to the user

---

Don't bother creating a PR until you've done this:

* [ ] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it
  does not need to pass.
* Once required tests have passed, you can make it "Ready for review".
* Say how how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* If changes were requested, and you've made them, then dismiss the review to get it looked at again.
* Add you organization to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md) if you like.
* You can ask for help!

PR #4365, opened 2020-10-24
feat(ui): Add parameter value enum support to the UI. Fixes #4192

basanthjenuhb:
Feature: Add UI support for enums
Discussed in https://github.com/argoproj/argo/issues/4192

Signed-off-by: basanthjenuhb <basanthjenuhb@gmail.com>

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).
* [x] I've signed the CLA.
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not.

PR #4314, opened 2020-10-17
feat(controller): add enum support in parameters (fixes #4192)

basanthjenuhb:
Enhancement proposal - https://github.com/argoproj/argo/issues/4192
UI PR coming up next

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).
* [x] I've signed the CLA.
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not.

PR #12578, opened 2024-01-26
add instruction on enum parameter in "getting started"

tylerbrewer2:
Fixes #TODO

### Motivation

I didn't see anything in the docs about using `enum` outside of `Intermediate Parameters`

### Modifications

Adds instructions on how to use the `enum` type in the `Getting Started` section.

### Verification

Viewed the markdown preview.

### DIFF
--- a/docs/walk-through/parameters.md
+++ b/docs/walk-through/parameters.md
@@ -89,3 +89,34 @@ spec:
 ```
 
 In this workflow, both steps `A` and `B` would have the same log-level set to `INFO` and can easily be changed between workflow submissions using the `-p` flag.
+
+If using the UI, the `enum` parameter type may also be used to restrict the parameter to certain values.
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: hello-world-enum-parameters-
+spec:
+  entrypoint: whalesay
+  arguments:
+    parameters:
+    - name: message
+      default: hello
+      enum:
+      - hello
+      - goodbye
+
+  templates:
+  - name: whalesay
+    inputs:
+      parameters:
+      - name: message       # parameter declaration
+    container:
+      # run cowsay with that message input parameter as args
+      image: docker/whalesay
+      command: [cowsay]
+      args: ["{{inputs.parameters.message}}"]
+```
+
+`enum` may also be used to create [Intermediate Parameters](../intermediate-inputs.md).
--- a/docs/walk-through/parameters.md
+++ b/docs/walk-through/parameters.md
@@ -89,34 +89,3 @@ spec:
 ```
 
 In this workflow, both steps `A` and `B` would have the same log-level set to `INFO` and can easily be changed between workflow submissions using the `-p` flag.
-
-If using the UI, the `enum` parameter type may also be used to restrict the parameter to certain values.
-
-```yaml
-apiVersion: argoproj.io/v1alpha1
-kind: Workflow
-metadata:
-  generateName: hello-world-enum-parameters-
-spec:
-  entrypoint: whalesay
-  arguments:
-    parameters:
-    - name: message
-      default: hello
-      enum:
-      - hello
-      - goodbye
-
-  templates:
-  - name: whalesay
-    inputs:
-      parameters:
-      - name: message       # parameter declaration
-    container:
-      # run cowsay with that message input parameter as args
-      image: docker/whalesay
-      command: [cowsay]
-      args: ["{{inputs.parameters.message}}"]
-```
-
-`enum` may also be used to create [Intermediate Parameters](../intermediate-inputs.md).
--- a/docs/intermediate-inputs.md
+++ b/docs/intermediate-inputs.md
@@ -1,4 +1,4 @@
-# Intermediate Parameters
+# `enum` and Intermediate Parameters
 
 > v3.4 and after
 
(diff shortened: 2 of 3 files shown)


agilgur5:
> I didn't see anything in the docs about using `enum` outside of `Intermediate Parameters`

Yea I've been thinking of adding a link to this to the [Suspending page](https://argo-workflows.readthedocs.io/en/release-3.5/walk-through/suspending/), it's not that well known.

That being said, since it is UI-specific, I'm not sure if we should include it here specifically. It's currently under "UI Features" in the "User Guide", which makes a lot of sense in the current organization of the docs.

docs/walk-through/parameters.md
```diff
@@ -89,3 +89,34 @@ spec:
...
+kind: Workflow
+metadata:
+  generateName: hello-world-enum-parameters-
+spec:
+  entrypoint: whalesay
+  arguments:
+    parameters:
+    - name: message
+      default: hello
+      enum:
```
agilgur5, line 106: huh I've actually never used this as a top-level Workflow parameter 👀  this use-case is indeed undocumented currently then... so I guess we should document it here as there's no other doc on parameters

docs/walk-through/parameters.md
```diff
@@ -89,3 +89,34 @@ spec:
 ```
 
 In this workflow, both steps `A` and `B` would have the same log-level set to `INFO` and can easily be changed between workflow submissions using the `-p` flag.
+
```
agilgur5, line 92:
```suggestion

# `enum` support

> v3.0 and after

```

Let's put this in a new section since it is distinct from the rest.

We can also add a version specifier for consistency with the rest of the docs. #4314 was released as part of [v2.12](https://github.com/argoproj/argo-workflows/blob/v3.5.4/CHANGELOG.md#v2120-rc1-2020-11-06), while #4365 was released as part of [v3.0](https://github.com/argoproj/argo-workflows/blob/v3.5.4/CHANGELOG.md#v300-rc1-2021-02-08)

docs/walk-through/parameters.md
```diff
@@ -89,3 +89,34 @@ spec:
...
+If using the UI, the `enum` parameter type may also be used to restrict the parameter to certain values.
```
agilgur5, line 93:
```suggestion
In the UI, the `enum` field can be used to restrict a parameter to certain values.
```

- more simple and direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- terminology: it's a "field", not a "parameter type"
- grammar: "a" parameter, not specific to "the"

docs/walk-through/parameters.md
```diff
@@ -89,3 +89,34 @@ spec:
...
+      parameters:
+      - name: message       # parameter declaration
+    container:
+      # run cowsay with that message input parameter as args
+      image: docker/whalesay
+      command: [cowsay]
+      args: ["{{inputs.parameters.message}}"]
+```
+
+`enum` may also be used to create [Intermediate Parameters](../intermediate-inputs.md).
```
agilgur5, line 122:
```suggestion
The `enum` field can also be used with [Intermediate Parameters](../intermediate-inputs.md).
```

- consistency: refer to `enum` the same way, use "can"
- I'm actually not sure that `enum` is necessarily required for that feature; `valueFrom.supplied` is the main thing I think. In general, it could use some more details. so let's just say it can be used with that feature

agilgur5 (commented):
Thanks for noticing this and adding some docs in an appropriate place! I left some small suggestions below 

Also you have [a](https://github.com/argoproj/argo-workflows/actions/runs/7674179336/job/20918271630?pr=12578) [few](https://github.com/argoproj/argo-workflows/pull/12578/checks?check_run_id=20918242237) CI checks [failing](https://github.com/argoproj/argo-workflows/actions/runs/7674169192/job/20918243038?pr=12578) -- those will all need to be fixed

docs/walk-through/parameters.md
agilgur5: Oh, this _is_ documented elsewhere; #7114 added this under the `WorkflowTemplate` page. That partly makes sense since top-level parameters would only be available with `WorkflowTemplates` -- if you submit a Workflow directly, you wouldn't have access to the UI selection menu. The walk through actually doesn't mention `WorkflowTemplates` at all, so back to square one of me thinking this doesn't quite fit here

I think one of the problems with the current placement is that it's under "Managing WorkflowTemplates" but this feature is not specific to maintaining them, it is for using them. I see that there's no real other place for `WorkflowTemplate` docs currently though.

Maybe we just link to those docs instead then

agilgur5:
See [above comment](https://github.com/argoproj/argo-workflows/pull/12578#discussion_r1468235830), this is actually an incorrect example as is, since using this feature at the top level (and without a `suspend`) require submitting via a `WorkflowTemplate`

docs/walk-through/parameters.md
agilgur5:
```suggestion
With `WorkflowTemplates`, you [can use](../workflow-templates.md#UI) the `enum` field to restrict a parameter to certain values in the UI.
```

After realizing https://github.com/argoproj/argo-workflows/pull/12578#discussion_r1468235830, I think it might be better to remove the example entirely from here and just link to the existing example in the `WorkflowTemplate` docs as that is advanced usage (the walk through is mostly for beginners to / new users of Workflows).

If you'd like, you can also improve that example as part of this PR (e.g. adding versions, improving the wording, perhaps moving it out of its current section)

docs/walk-through/parameters.md
tylerbrewer2: I was just looking back at how I used `enum` it was in a workflow template, so this example is definitely wrong.

I do think I would have found the docs if they were in the UI section. We could move [this section ](https://github.com/Hadrian-MTV/charts/blob/main/applications/flow/factory-execution/values-prod.yaml#L12)to `UI Features` and add screenshots? And then reference the new page within the `WorkflowTemplate` page.

agilgur5: > We could move [this section](https://github.com/Hadrian-MTV/charts/blob/main/applications/flow/factory-execution/values-prod.yaml#L12)

Your link above 404s for me; that seems to be a link to a private repo's Helm values?


> to `UI Features` and add screenshots?

The [Intermediate Parameters](https://argo-workflows.readthedocs.io/en/release-3.5/intermediate-inputs/) page is already under UI Features. Usage with WorkflowTemplates could perhaps be a prelude to that all on one page

tylerbrewer2: Sorry! [This](https://argo-workflows.readthedocs.io/en/latest/workflow-templates/#ui) is where I meant to link to.

And that makes sense to me.

agilgur5: Ok I thought that was what you meant 😅.
I think we can move that to the Intermediate Parameters page and rename the title there a bit

PR #9805, opened 2022-10-13
feat: Workflow title/description in workflow list view. Fixes #6529

jmeridth:
Screen Shot 2022-10-16 at 10 52 08 PM

Fixes #6529 (partially, no markdown)

- [x] title and description in list view

TL;DR
Add the following to your workflow annotations to get a title/description to show on the workflow list page:

```yaml
spec:
  annotations:
    workflows.argoproj.io/title: 'this is a title'
    workflows.argoproj.io/description: 'this is a description'

```

Signed-off-by: jmeridth <jmeridth@gmail.com>

Fixes #TODO

Please do not open a pull request until you have checked ALL of these:

* [x] Create the PR as draft .
* [x] Run `make pre-commit -B` to fix codegen and lint problems.
* [x] Sign-off your commits (otherwise the DCO check will fail).
* [x] Use [a conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) (otherwise the commit message check will fail).
* [x] "Fixes #" is in both the PR title (for release notes) and this description (to automatically link and close the issue).
* [x] Add unit or e2e tests. Say how you tested your changes. If you changed the UI, attach screenshots.
* [x] Github checks are green.
* [x] Once required tests have passed, mark your PR "Ready for review".

If changes were requested, and you've made them, dismiss the review to get it reviewed again.

PR #12580, opened 2024-01-28
fix(ui): code-split markdown title + desc, fix row linking, etc

agilgur5:
Follow-up to #10553 issues
-  Partial fix for https://github.com/argoproj/argo-workflows/pull/10553#pullrequestreview-1799370952 -- this fixes the UI issues, will have some separate PRs for docs. **EDIT**: Docs PR: #12582
    - part of the docs changes need backporting as well as the title / desc feature is older per https://github.com/argoproj/argo-workflows/pull/10553#discussion_r1438476652 (it is from #9805)

Related / partial fix for to #12059, although the PR was merged after that issue was created

### Motivation

- `react-markdown` and `remark-gfm` are large deps, so we should code-split / lazy load them
  - and we should only load them if a user is using this feature
    - can't quite detect if a string is markdown without a dep, but can at least check if the annotations are used

- CSS had several issues
  - was imported from the wrong component
  - `overlay` was not sufficiently unique (as we currently use global CSS and not something like CSS Modules)
  - some unnecessary CSS attributes
  - most importantly, the positioning of rows broke and did not flex to the last column
    - it caused columns to not match the heading very confusingly and a lot of empty space on the right as well
    - that also broke some overflow properties (it re-broke #11908)
    - and it broke the details drawer as well
      - I couldn't get it to open half the time
      - when I did open it, the first two columns were empty, and with the last few columns empty too, it looked like it was just centered horizontally (unintentionally)

- fix nested link issues
  - overlay was unintuitive and hacky
    - can instead use JS to capture the event & explicitly prevent bubble up
      - semantically, the JS is still hacky to have nested links, but it less hacky
        - nested linking is also an optional feature that happens only if a user has an annotation, with embedded markdown, and with a link in the markdown (i.e. an optional feature of an optional feature of an optional feature), so I don't mind that hack as much now

- remove `Link` for the entire row as it was unintuitive, especially with the drawer
  - the drawer click conflicted, but superseded the link click
  - the whole drawer would be highlighted on hover since the whole row was a link, including the drawer
  - one could also not copy+paste things from the row because it caused a click
    - in particular, there were several times I wanted to copy something from the details drawer and would instead end up on the details page
  - only link the workflow name now

    
### Modifications

- code-split `react-markdown` and `remark-gfm`
  - only load them when annotations are present
- revert the split of the `workflows-row-name.tsx` file and merge it back to where it was
  - import the CSS from `workflows-row.tsx`, not `workflows-row-name.tsx`
    - remove most of the CSS as the Link was changed and because some of it was not necessary
- fix several issues with the links
  - the `overlay` anchor element was a workaround for nested linking (which could optionally be used with the markdown feature), remove it and use JS to handle nested links instead
    - can explicitly control event bubbling in JS 
    - the overlay caused problems with the drawer and removed the highlighting that an anchor element has on hover
  - only link the `name` column and not the entire row (which included the drawer before, both open and closed)

#### Rendered UI itself

Manually tested the UI:
<details> <summary>Before had several issues (as listed in <a href="https://github.com/argoproj/argo-workflows/pull/10553#discussion_r1461385314">#10553 (review)</a>): </summary>

Misplaced columns; row does not flex all the way to the right:
Screenshot 2024-01-22 at 1 05 49 AM

Details drawer has empty space on both left and right:
Screenshot 2024-01-28 at 12 21 52 PM
Screenshot 2024-01-28 at 12 22 46 PM

Overflow that was fixed in #11908 was re-broken:
Screenshot 2024-01-28 at 12 23 53 PM

</details>

After:
Screenshot 2024-01-27 at 9 26 51 PM

#### UI deps code-split

Shaved off ~575KB / ~140KB minified / ~41KB gzipped from the main entrypoint:
Screenshot 2024-01-28 at 1 39 49 PM
Screenshot 2024-01-28 at 1 40 01 PM


ui/src/app/workflows/components/workflows-row/workflows-row.tsx
```diff
@@ -23,7 +27,11 @@ interface WorkflowsRowProps {
 export function WorkflowsRow(props: WorkflowsRowProps) {
     const [hideDrawer, setHideDrawer] = useState(true);
     const wf = props.workflow;
-    const workflowLink = uiUrl(`workflows/${wf.metadata.namespace}/${wf.metadata.name}?uid=${wf.metadata.uid}`);
+    // title + description vars
```
jmeridth, line 30: Can we remove? Wdyt?

jmeridth (approved):
Nice use of suspense. Just one minor nit.

ui/src/app/workflows/components/workflows-row/workflows-row.tsx
agilgur5: I added it bc we have some long lists of vars elsewhere that have some comments. A comment makes those lists a little easier on the eyes with some visual separation / categorization.

jmeridth: I'd prefer a newline over a redundant comment.

agilgur5: I'd rather specify a category 🤷 I don't think it's entirely redundant either given the last two variables are different by name at least (from a quick glance I pay attention to the names much more than the logic)

jmeridth: 👎 is that I disagree but not enough to block this PR. Maybe I can learn something from this perspective. 🤷

agilgur5: honestly I can't say I care too much about it either, but enough to prefer it 😅

juliev0 (approved):

PR #12582, opened 2024-01-28
docs(ui): full copy-edit of title & description page

agilgur5:
Partial fix for https://github.com/argoproj/argo-workflows/pull/10553#pullrequestreview-1799370952 -- this fixes the docs & example issues, while #12580 fixes the UI issues

### Motivation

<details> <summary>Significantly improve the "Title and Description" page (and related example) from #10553 </summary>

- It was missing version specifiers (as was the example)
- It only showed the markdown variant, but the non-markdown variant actually existed earlier after #9805 (released in [v3.4.4](https://github.com/argoproj/argo-workflows/blob/v3.5.4/CHANGELOG.md#v344-2022-11-28))
- It only had one complex example, and not simple examples
- Complex example had verbal descriptions, but not the raw code
- The example image's alt `title` was not very accessible
- Had an outdated link to Couler
- Had a few k8s docs style guide issues
- Had a few typos
- Had a markdown issue
- etc

</details>

### Modifications

- <details> <summary>split the page into sections with and without markdown </summary>
 
  - add a non-markdown example and description
  - add a simple markdown example of just one row
  - use the same width images for both of these examples for visual consistency with the larger example
 
 </details>
 
- add version specifiers to both of these sections

- link to markdown docs when referencing Markdown
- <details> <summary>improve example descriptions </summary>

  - fix missing end parenthesis
  - be more specific, for instance, one of the examples is multi-line

</details>

- <details> <summary>instead of having loose descriptions of rows in a list, actually write out the (partial) manifest that results in those renderings </summary> 

  - I "reverse-engineered" what they would be based on the image
  - remove extra / duplicated example at the end (?)
  - use one code block instead of multiple to make it easier to read

</details>

<details> <summary> accessibility improvements: </summary>

- better describe the example image(s), specify that they are examples
  - previously, it just said "Workflow Title and Description", which would not really explain the image for someone using a screen reader, as that's the same as the title of the page

</details>

<details> <summary>prose style improvements: </summary>

- use present tense and active voice, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-present-tense)
- be more simple and direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- avoid jargon / idioms, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-jargon-and-idioms)
  - "auto turn" -> "convert"
- be specific, instead of "title and description annotations", specify what the annotations are

[...]

PR #12583, opened 2024-01-28
build: parallelize `Makefile` jobs by default

agilgur5:
### Motivation

- optimize some of the `make` jobs by parallelizing them by default when possible
  - this way you don't have to specify, for instance, `make start -j4 UI=true` etc

### Modifications

- `-j` flags allows it to start more processes
  - e.g. `-j4` means it can start 4

- use `nprocs` to get number of available CPUs

### Verification

Used the `Makefile` with this unstaged change for a few PRs, confirmed some things ran in parallel
- the logs are interleaved for parallel jobs by default, so fairly simple to tell

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ USE_NIX := false
 MAKEFLAGS += --no-builtin-rules
 .SUFFIXES:
 
+# parallelize jobs by number of CPUs available
+NPROCS = $(shell nproc || printf 1)
+MAKEFLAGS += -j$(NPROCS)
+
 BUILD_DATE            := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 # copied verbatim to release.yaml
 GIT_COMMIT            := $(shell git rev-parse HEAD || echo unknown)


agilgur5:
Well that's interesting, CI `codegen` [failed](https://github.com/argoproj/argo-workflows/actions/runs/7689244513/job/20951411465?pr=12583). I did actually run `make codegen` with this change (for #12582), but probably wasn't fresh like it is on GH Actions. Might be an ordering issue in the `Makefile` then

agilgur5:
Ok everything else passed, so it's just `codegen` failing. I think that might be because of the `rm -rf v3` lines -- there are several individual jobs that have that and if they ran while another `codegen` job was running, that would conflict.

Either need to remove those (replace with only one at the end of `codegen` for instance) or make `codegen` sequential (not as ideal).

Kicking this back into draft for now as I need to test that a bit and am working on some higher priority things (P1 fixes)

Joibel:
I think I know how to solve the v3 link thing relatively painlessly, do you mind if I take over this PR?

I don't promise to keep digging until it's all parallel safe, but at least try to progress that bit.

agilgur5:
Sure, go for it

PR #12595, opened 2024-01-31
test: Manual Retry (CLI) Failed Workflow

Garett-MacGowan:
### Motivation

~As part of my investigation and testing for https://github.com/argoproj/argo-workflows/pull/12574, I noticed that workflow update conflicts did not result in the workflow being marked as errored. I believe workflows should be marked as errored upon an update conflict. Additionally, marking workflows with conflicting updates as errored makes it easier to test the issue described in #12574.~

**Edit:** 
As Joibel pointed out, I was mistaken in marking the workflow as errored upon conflict. I didn't notice the `!` infront of `apierr.IsConflict()`, so I interpreted it wrong.

I think the test is still useful.

### Modifications

~Added `persistWorkflowConflictErr(ctx, wfClient, err)` and a call to it in `persistUpdates`~

### Verification

Added test for manual workflow retry (in attempt to reproduce the issue described in #12574). This test is unable to reproduce the issue, though it is still a useful test.

### DIFF
--- a/test/e2e/cli_test.go
+++ b/test/e2e/cli_test.go
@@ -866,6 +866,20 @@ func (s *CLISuite) TestWorkflowRetry() {
 		})
 }
 
+func (s *CLISuite) TestWorkflowRetryFailedWorkflow() {
+	failFirstPassSecondWorkflowWhen := s.Given().
+		Workflow("@testdata/fail-first-pass-second.yaml").
+		When()
+
+	failFirstPassSecondWorkflowWhen.
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeFailed).
+		RunCli([]string{"retry", "-l", "workflows.argoproj.io/workflow=fail-first-pass-second-workflow", "--namespace=argo"}, func(t *testing.T, output string, err error) {
+			assert.NoError(t, err, output)
+		}).
+		WaitForWorkflow(fixtures.ToBeSucceeded)
+}
+
 func (s *CLISuite) TestWorkflowRetryNestedDag() {
 	s.Given().
 		Workflow("@testdata/retry-nested-dag-test.yaml").
--- a/test/e2e/testdata/fail-first-pass-second.yaml
+++ b/test/e2e/testdata/fail-first-pass-second.yaml
@@ -1,0 +1,37 @@
+metadata:
+  name: fail-first-pass-second-workflow
+  labels:
+    workflows.argoproj.io/workflow: "fail-first-pass-second-workflow"
+spec:
+  entrypoint: main
+  volumeClaimTemplates:
+    - metadata:
+        name: artifacts
+      spec:
+        accessModes: [ "ReadWriteOnce" ]
+        resources:
+          requests:
+            storage: 0.1Gi
+  templates:
+    - name: main
+      container:
+        name: main
+        image: 'argoproj/argosay:v2'
+        volumeMounts:
+          - name: artifacts
+            mountPath: /tmp
+        command: [sh, -c]
+        args:
+          - |
+            if [ -s /tmp/artifact.txt ]
+            then
+              echo "Successfully retried failing workflow."
+              exit 0
+            else
+              echo "test artifact" > /tmp/artifact.txt
+              exit 1
+            fi
+      outputs:
+        artifacts:
+          - name: artifact
+            path: /tmp/artifact.txt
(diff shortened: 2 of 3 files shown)


Garett-MacGowan:
@juliev0 Let me know what you think of this.

@shuangkun This may be of use in your investigation.

Joibel:
This doesn't feel correct. A conflict is an expected event in our scenario where we have both the argo-server and workflow-controller potentially writing to the same Workflow object concurrently. Marking those workflows as errored feels likely to cause inadvertant errors.

The most obvious case to me would be a Stop or Terminate from the argo-server.

Garett-MacGowan:
> This doesn't feel correct. A conflict is an expected event in our scenario where we have both the argo-server and workflow-controller potentially writing to the same Workflow object concurrently. Marking those workflows as errored feels likely to cause inadvertant errors.
> 
> The most obvious case to me would be a Stop or Terminate from the argo-server.

Thanks for the feedback. I'll look into this.

shuangkun:
> @juliev0 Let me know what you think of this.
> 
> @shuangkun This may useful in your investigation.


I think there is a problem with this. This change cannot solve my current problem. The best way is that the workflow has been truly completed when the customer can retry.

Garett-MacGowan:
> > @juliev0 Let me know what you think of this.
> > @shuangkun This may useful in your investigation.
> 
> I think there are something wrong with this. This change cannot solve my current problem. The best way is that the workflow has been truly completed when the customer can retry.

I'm not suggesting that this fixes your issue. The test came however, attempts to reproduce your issue. I haven't been able to so far.

shuangkun:
> > > @juliev0 Let me know what you think of this.
> > > @shuangkun This may useful in your investigation.
> > 
> > 
> > I think there are something wrong with this. This change cannot solve my current problem. The best way is that the workflow has been truly completed when the customer can retry.
> 
> @shuangkun I'm not suggesting that this fixes your issue. The test case attempts to reproduce your issue. I haven't been able to so far.

Okay, thanks. I'll take a look

Garett-MacGowan:
@Joibel Oops, I missed the `!` In `!apierr.IsConflict(err)`. What you're saying makes sense now. I see it performs a retry. Will remove.

@shuangkun's

shuangkun:
> @Joibel Oops, I missed the `!` In `!apierr.IsConflict(err)`. What you're saying makes sense now. I see it performs a retry. Will remove.
> 
> @shuangkun update retry is limited. Are you seeing a `woc.log.Warnf("Update retry attempt %d failed: %v", attempt, err)` log

Yes, I have

Garett-MacGowan:
> It's very creative to run the CLI in a Workflow. But is there something about what you're testing that it makes more sense to run it in a Workflow as opposed to testing the CLI directly? (which I notice we do in various tests [here](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/cli_test.go#L829))

Apologies for being slow to respond. Been busy.

There's no reason other than it was the easiest for me to do since I didn't see the CLI testing utilities you've pointed out here. I can change if you think it makes sense to do so.

juliev0:
> > It's very creative to run the CLI in a Workflow. But is there something about what you're testing that it makes more sense to run it in a Workflow as opposed to testing the CLI directly? (which I notice we do in various tests [here](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/cli_test.go#L829))
> 
> Apologies for being slow to respond. Been busy.
> 
> There's no reason other than it was the easiest for me to do since I didn't see the CLI testing utilities you've pointed out here. I can change if you think it makes sense to do so.

I hate to bug you with that but I guess there is something to be said for consistency. Would you be up for that?

Garett-MacGowan:
> > > It's very creative to run the CLI in a Workflow. But is there something about what you're testing that it makes more sense to run it in a Workflow as opposed to testing the CLI directly? (which I notice we do in various tests [here](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/cli_test.go#L829))
> > 
> > 
> > Apologies for being slow to respond. Been busy.
> > There's no reason other than it was the easiest for me to do since I didn't see the CLI testing utilities you've pointed out here. I can change if you think it makes sense to do so.
> 
> I hate to bug you with that but I guess there is something to be said for consistency. Would you be up for that?

No problem. I'll make the change. Might not be for a little bit, but I'll get around to it!

Garett-MacGowan:
@juliev0 `retry_test.go:168: You cannot use RunCli for tests that are not in TestCLISuite`. Permission to run CLI inside the container?

Garett-MacGowan:
@juliev0 Let me know if this test works for you now. It has been greatly simplified.

juliev0 (approved):

juliev0:
that works, thanks for this!

PR #12598, opened 2024-01-31
fix: Fixes git over azure devops Fixes #11705

### LINKED ISSUE #11705: Authentication issue on Azure devops with the git artifact example
Hi all,
the issue 10600 is closed but I still have an authentication issue on Azure devops with the git artifact example

https://github.com/argoproj/argo-workflows/blob/master/examples/input-artifact-git.yaml

`artifact argo-source failed to load: failed to clone "https://<org>@dev.azure.com/<org>/folder/_git/customerdata":
          unexpected client error: unexpected requesting "https://<org>@dev.azure.com/<org>/folder/_git/customerdata/git-upload-pack"
          status code: 400`

This error is thrown using Argo Workflow 3.4.10: any idea?

Thanks

_Originally posted by @MarcoFabbri-orienteed in https://github.com/argoproj/argo-workflows/issues/10600#issuecomment-1673171748_

clkamp:
Fixes #11705 

### Verification

Tested it in a local dev environment, the fix is the one suggested in https://github.com/argoproj/argo-workflows/issues/11705#issuecomment-1793778023

### DIFF
--- a/workflow/artifacts/git/git.go
+++ b/workflow/artifacts/git/git.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/protocol/packp/capability"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	ssh2 "github.com/go-git/go-git/v5/plumbing/transport/ssh"
@@ -84,6 +85,19 @@ func (g *ArtifactDriver) Delete(s *wfv1.Artifact) error {
 
 func (g *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string) error {
 	a := inputArtifact.Git
+
+	// Azure DevOps requires multi_ack* capabilities which go-git does not currently support
+	// Workaround: removing these from UnsupportedCapabilities allows clones to work (see https://github.com/go-git/go-git/pull/613)
+	// TODO: remove this once https://github.com/go-git/go-git/issues/64 is fully fixed
+	var newCaps []capability.Capability
+	for _, c := range transport.UnsupportedCapabilities {
+		if c == capability.MultiACK || c == capability.MultiACKDetailed {
+			continue
+		}
+		newCaps = append(newCaps, c)
+	}
+	transport.UnsupportedCapabilities = newCaps
+
 	sshUser := GetUser(a.Repo)
 	closer, auth, err := g.auth(sshUser)
 	if err != nil {


eduardodbr:
Hey @clkamp ,

the CI is failing due to some flaky tests that were already fixed and merged to master, can you please merge them?

clkamp:
@agilgur5 Could you review and possibly merge this?

workflow/artifacts/git/git.go
```diff
@@ -84,6 +85,14 @@ func (g *ArtifactDriver) Delete(s *wfv1.Artifact) error {
 
 func (g *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string) error {
 	a := inputArtifact.Git
+
+	// Azure DevOps requires capabilities multi_ack / multi_ack_detailed,
+	// which are not fully implemented and by default are included in
+	// transport.UnsupportedCapabilities.
+	transport.UnsupportedCapabilities = []capability.Capability{
+		capability.ThinPack,
+	}
```
agilgur5, line 94: so as I understand, this currently resets a global for all hosts. we should limit this to Azure repos, as this is otherwise advertising incorrect capabilities to other hosts as well (which can have some unexpected consequences / side-effects).

Ideally we should only set it for the request and then set it back, but I'm not sure if there's a way to pass the `transport` as a variable (as relying on a global would not be thread-safe).

Also we should be a bit more explicit in the comment. And I think the approach from https://github.com/kluctl/kluctl/pull/637 is a bit more future proof -- remove those two capabilities instead of resetting:
```suggestion
	// Azure DevOps requires multi_ack* capabilities which go-git does not currently support
	// Workaround: removing those from UnsupportedCapabilities allows clones to work (see https://github.com/go-git/go-git/pull/613)
	// TODO: remove this once https://github.com/go-git/go-git/issues/64 is fully fixed
    var newCaps []capability.Capability
	for _, c := range transport.UnsupportedCapabilities {
		if c == capability.MultiACK || c == capability.MultiACKDetailed {
			continue
		}
		newCaps = append(newCaps, c)
	}
	transport.UnsupportedCapabilities = newCaps
```

agilgur5 (commented):
Sorry I've been passing a kidney stone the past month per my status (think I'm finally in the clear now...) among other things.

I left a comment in-line with some suggestions. Ideally we should limit this to Azure DevOps repos and only set it per request as well.

Could you also add a unit test for this?

clkamp:
Thanks for your review and sorry for my late reply.
For the moment I just included your suggestion, you are right its much more future proof. I'm not sure how to test for azure devops Repos and all other places that applied a fix for this also did not check for it.
Would you be fine with accepting it in this state or should I further investigate how to limit the fix to azure repos?

agilgur5:
> Would you be fine with accepting it in this state [...]?

No, as it can cause issues for and unexpected behavior for other providers.
If it's running on all repos, it should at the very least be behind an environment variable and not the default behavior.

> I'm not sure how to test for azure devops Repos

You can check the URL in `a.Repo` for an Azure DevOps URL, no?

> and only set it per request as well.
> 
> Could you also add a unit test for this?

You also missed these two other parts from my [previous review comment](https://github.com/argoproj/argo-workflows/pull/12598#pullrequestreview-1904465296)

SDEBEUL:
Hi all any news on getting this released? 
I am also waiting on this to be fixed :).

br

PR #12600, opened 2024-01-31
Update README.md

Talibiilm:
Fixes #TODO


docs/README.md
```diff
@@ -1,4 +1,4 @@
-# Argo Workflows
+# Argo Workflows
```
agilgur5, line 1: unnecessary and fails `markdownlint` too

PR #12604, opened 2024-02-01
docs: document windows symbolic link config for local dev

### LINKED ISSUE #12603: Make argoexec-image returns error in Windows `expected 'package', found signal_darwin`
### Pre-requisites

`make start` `make argoexec image` used to work in windows with windows file system and devcontainer prior to 3.4.x

Now it returns
```
$ make dist/argoexec 
GIT_COMMIT=877c5523066e17687856fe3484c9b2d398e986f5 GIT_BRANCH=release-3.5.1 GIT_TAG=v3.5.1 GIT_TREE_STATE=dirty RELEASE_TAG=true DEV_BRANCH=false VERSION=v3.5.1
KUBECTX= DOCKER_DESKTOP=false K3D=false DOCKER_PUSH=false
RUN_MODE=local PROFILE=minimal AUTH_MODE=hybrid SECURE=false STATIC_FILES=true ALWAYS_OFFLOAD_NODE_STATUS=false UPPERIO_DB_DEBUG=0 LOG_LEVEL=debug NAMESPACED=trueCGO_ENABLED=0 go build -v -gcflags '' -ldflags '-X github.com/argoproj/argo-workflows/v3.version=v3.5.1 -X github.com/argoproj/argo-workflows/v3.buildDate=2024-01-01T17:15:27Z -X github.com/argoproj/argo-workflows/v3.gitCommit=877c5523066e17687856fe3484c9b2d398e986f5 -X github.com/argoproj/argo-workflows/v3.gitTreeState=dirty -X github.com/argoproj/argo-workflows/v3.gitTag=v3.5.1 -extldflags -static' -o dist/argoexec ./cmd/argoexec
workflow\executor\os-specific\signal_linux.go:1:1: expected 'package', found signal_darwin
make: *** [Makefile:237: dist/argoexec] Error 1
```

### Version

3.5.0+

```YAML
.
```

```text
```

```text
```

tczhao:
Fixes #12603 

In Windows, setting symlink=true in Git automatically trigger powershell command `mklink` on all symbolic link files, so they connect properly in windows file system

Setting autocrlf to false prevents line return getting modified causing bash issue

### DIFF
--- a/docs/running-locally.md
+++ b/docs/running-locally.md
@@ -33,7 +33,7 @@ Note:
     * Configure [`.wslconfig`](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configuration-setting-for-wslconfig) to limit memory usage by the WSL2 to prevent VSCode OOM.
 
 * For **Windows File System**
-    * Make sure git symbolic link is set to True when cloning the repo, clone the repo with `git clone -c core.symlinks=true core.autocrlf=false <URL>`
+    * Clone the repo with `git clone -c core.symlinks=true core.autocrlf=false https://github.com/<my-fork>/argo-workflows.git`
 
 * For **Linux**
     * Use [Docker Desktop](https://docs.docker.com/desktop/linux/install/) instead of [Docker Engine](https://docs.docker.com/engine/install/) to prevent incorrect network configuration by k3d.
--- a/docs/running-locally.md
+++ b/docs/running-locally.md
@@ -32,6 +32,9 @@ Note:
 * For **Windows WSL2**
     * Configure [`.wslconfig`](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configuration-setting-for-wslconfig) to limit memory usage by the WSL2 to prevent VSCode OOM.
 
+* For **Windows File System**
+    * Make sure git symbolic link is set to True when cloning the repo, clone the repo with `git clone -c core.symlinks=true core.autocrlf=false <URL>`
+
 * For **Linux**
     * Use [Docker Desktop](https://docs.docker.com/desktop/linux/install/) instead of [Docker Engine](https://docs.docker.com/engine/install/) to prevent incorrect network configuration by k3d.
 


docs/running-locally.md
```diff
@@ -32,6 +32,9 @@ Note:
 * For **Windows WSL2**
     * Configure [`.wslconfig`](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configuration-setting-for-wslconfig) to limit memory usage by the WSL2 to prevent VSCode OOM.
 
+* For **Windows File System**
+    * Make sure git symbolic link is set to True when cloning the repo, clone the repo with `git clone -c core.symlinks=true core.autocrlf=false <URL>`
```
agilgur5, line 36: This is fairly redundant as written with "when cloning the repo, clone the repo with". The description also restates the command in less clear terms, IMO

```suggestion
    * Clone the repo with `git clone -c core.symlinks=true core.autocrlf=false https://github.com/<my-fork>/argo-workflows.git`
```

- simple and direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- specify most of the URL

agilgur5 (commented):
I believe these can be set automatically in the [`.gitatrributes`](https://github.com/argoproj/argo-workflows/blob/6a39edf366319a40d37ccf406fe27dcee3d15705/.gitattributes) if necessary. 

`autocrlf` is usually recommended as `auto` though

tczhao:
> I believe these can be set automatically in the [`.gitatrributes`](https://github.com/argoproj/argo-workflows/blob/6a39edf366319a40d37ccf406fe27dcee3d15705/.gitattributes) if necessary, which would be preferable to docs changes.
> 
> `autocrlf` is usually recommended as `auto` though

- For the `symlink`, if we configure it using `gitattributes`, we have to specify every single symbolic files but with `git config core.symlinks`, it will automatically set up symlink for all files, and we don't need to maintain this file list.

- For `autocrlf`, I think setting to `auto` if only useful if we develop frontend without running the backend. Running the argo-workflow backend in Windows requires `dev container` or `wsl2`; both require line return using `LF`. Otherwise, `bash` commands won't work. 
Since in the running-locally guide we always run the backend and frontend together, I think setting `autocrlf=false` for this repo is a valid approach.

agilgur5:
> * For the `symlink`, if we configure it using `gitattributes`, we have to specify every single symbolic files but with `git config core.symlinks`, it will automatically set up symlink for all files, and we don't need to maintain this file list.

Looks like `core.symlinks` is `true` [by default](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks)?

> * [...] Since in the running-locally guide we always run the backend and frontend together, I think setting `autocrlf=false` for this repo is a valid approach.

So it would make sense to set this in the `.gitattributes` then, no?
Also, I do sometimes run the front-end build or install separately (not the devserver), but I'm on Mac.

agilgur5:
> Looks like `core.symlinks` is `true` [by default](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks)?

Stumbled upon on an old thread I hadn't finished reading -- another contributor there also had to [enable `core.symlinks`](https://github.com/argoproj/argo-workflows/issues/11879#issuecomment-1739987441). I'm wondering if it's maybe disabled on Windows installations by default? (though that might depend on where you installed from too)

tczhao:
> > Looks like `core.symlinks` is `true` [by default](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks)?
> 
> Stumbled upon on an old thread I hadn't finished reading -- another contributor there also had to [enable `core.symlinks`](https://github.com/argoproj/argo-workflows/issues/11879#issuecomment-1739987441). I'm wondering if it's maybe disabled on Windows installations by default? (though that might depend on where you installed from too)

I had a quick look, and the latest Git installation in Windows defaults the symlink to `enabled`. I think we only see this issue if people have been using Git on Windows and never reinstalled it.
Perhaps we could make a comment and workaround label in the issue and close it instead

agilgur5:
Superseded by #13534

PR #12612, opened 2024-02-03
ci: add `mkdocs` Python deps to dependabot security updates 

agilgur5:
Follow-up to #12360 that added `docs/requirements.txt` (glad to have a proper dep file for that!)
- Follows #12487, which merge conflicted with this, so I waited to make the PR for this one (can see that the original commits were made at the same time)

### Motivation

- same as our other deps, for security updates only
  - in particular, a scenario I can see possible is if `mkdocs` has an XSS vuln or something that would impact the docs site

### Modifications

- Add `docs/` `pip` section to `dependabot.yml`

style: reorder dep sections
- put Go, then NPM, then docs, then GH Actions
  - i.e. the most important prod impacting deps, then build deps

### Verification

n/a, this can only be tested after updating

### Notes to Reviewers

I was thinking about some of the NPM docs deps too, but those don't make it into prod, so I don't think they necessarily need to have security updates, but that wouldn't necessarily be bad practice to safeguard devDeps too.
- they also need a proper dep file as well, currently they're installed directly in the `Makefile` as well as in the Nix flakes.

`mkdocs` is technically _also_ a devDep, but the code it produces goes into the prod docs site, per the "Motivation" section above. So if `mkdocs` itself has a vuln, not a big issue, but if its generated code has a vuln, potentially an issue. There's no real way to distinguish between those two scenarios though, unfortunately.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,14 @@ updates:
     open-pull-requests-limit: 0
 
   # build / CI dependencies
+  - package-ecosystem: "pip"
+    directory: "/docs"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+    # ignore all non-security updates: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit
+    open-pull-requests-limit: 0
+
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,6 @@
 version: 2
 updates:
+  # prod dependencies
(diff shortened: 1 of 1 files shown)


jmeridth (approved):

isubasinghe:
Opinion on some kind of CI check to ensure the dependencies in Nix and otherwise are the same? 
Should be relatively easy to do I think. 

Just print out the version and diff them. Can be as a part of a later PR.

isubasinghe (approved):
Ah the diff was annoying to read, but it looks good otherwise.

agilgur5:
> Ah the diff was annoying to read, but it looks good otherwise.

Oh yea, the re-ordering made the diff look real wacky since some of the config is the same for each ecosystem. Sorry I forgot to mention that!

Issue #9734, opened 2022-10-03
allow cronworkflow's schedule field accept multiple strings


tooptoop4:
ie schedule: ["*/20 * 1-5 * *", "*/15 * 25-31 * *"]
so that the cronworkflow is run on any of these schedules

this is less duplication than making 2 different cronworkflows with one schedule each

sarabala1979:
@tooptoop4 is a good idea. Do you like to contribute to it?

shraddhazpy:
@sarabala1979 I will want to take this up!

PR #12616, opened 2024-02-04
feat: allow multiple schedules in a cron workflow

### LINKED ISSUE #12493: Support Complex schedule expression in CronWorkflow
# Summary

Improve CronWorkflow to support multiple complex cron schedules.

## What change needs making?

Currently, a single CronWorkflow can only be configured with a single cron expression. This makes it hard to configure complex schedules that can't be expressed with a single cron expression. To resolve this, I suggest that we improve CronWorkflow to allow the usage of multiple cron expressions in a single CronWorkflow resource.

## Use Cases:

I have a script that I need to run on a complex schedule, which isn't expressible with a single cron expression. The schedule changes within the day, week, and month as follows:

```
0  11    *  *  1-5 /path/to/some_process
20 11-13 *  *  1-5 /path/to/some_process
50 11-18 21 09 *   /path/to/some_process
50 11-18 24 10 *   /path/to/some_process
```

With the current implementation, I'd need to create a separate CronWorkflow for each schedule. It comes with some inconveniences: having to manage multiple CronWorkflows and the possibility of the script executed by different CronWorkflows at the same time, causing overlapping execution.

By allowing a single CronWorkflow to take multiple cron expressions, it will ease the management of schedules and ensure that the same script is not run by different CronWorkflows simultaneously.

---

eduardodbr:
Fixes: #12493
Fixes: #9734

### Motivation

This PR introduces a new CronWorkflow spec field `Schedules`, a list of cron schedules for a workflow to execute. It maintains backward compatibility with `Spec.Schedule` and adds a validation that ensures that both fields cannot be configured simultaneously. 

### Controller

1 - This implementation tries to maintain the same behavior for the controller and operator. This biggest change was made to the `cronFacade`where a key can hold multiple cron entries instead of only one. This involved changing the signature for the method `Load` to return an array of cronWfOperationCtx since a key can hold multiple entries:

```
 func (f *cronFacade) Load(key string) ([]*cronWfOperationCtx, error). 
```

This method doesn't seem to be used anywhere and cronFacede is not exposed so I don't think there is any issue with this change. Let me know if you think otherwise. 

2 - The controller is currently using the label  `cronworkflows.argoproj.io/last-used-schedule` to store the current cron expression. This label is only used to check whether the cron expression has been updated since the last time the WF ran, so this PR changes it to be a comma-separated list of the schedules (if applicable). Whenever any of the schedules is updated the label will be changed. 

### Possible Issues 

If two cron expressions overlap (for example the expressions `*/2 * * * *` and  `*/3 * * * *` will overlap at minute 6,12,18,etc)  only one WF will be executed because the second will be seen as a duplicate submission. I don't think there is an issue with this, but let me know if you think otherwise. This makes it a bit harder to create an E2E test that is deterministic. Because of that, I decided to not add an E2E test for this feature.

### CLI
`argo cron list` was updated to show all schedules:

image

`argo cron get` was updated:

image

### UI

**NOTE: this is my first time ever doing react /typescript it is expected that the code lacks quality. Please suggest changes.** 

Cron Workflow list was updated to show all schedules:
image

Cron Workflow status viewer was also updated:

image

Cron Spec editor:

image

### Verification

This change can be tested with the following cron workflow:

```
kind: CronWorkflow
metadata:
  name: test-cron
spec:
  schedules:
  - "*/2 * * * *"
  - "*/3 * * * *"
  concurrencyPolicy: "Allow"
  startingDeadlineSeconds: 0
  workflowSpec:
    entrypoint: whalesay
    templates:
    - name: whalesay
      container:
        image: alpine:3.6
        command: [sh, -c]
        args: ["date; sleep 1"]
```


eduardodbr:
CI is failing because of #12596

tooptoop4 (approved):

tooptoop4:
:shipit:

terrytangyuan (commented):
Great work!

cmd/argo/commands/cron/util.go
```diff
@@ -11,9 +11,18 @@ import (
 // GetNextRuntime returns the next time the workflow should run in local time. It assumes the workflow-controller is in
 // UTC, but nevertheless returns the time in the local timezone.
 func GetNextRuntime(cwf *v1alpha1.CronWorkflow) (time.Time, error) {
```
terrytangyuan, line 13: Some unit tests on this would be great

terrytangyuan (approved):
LGTM once conflicts are resolved

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
```diff
@@ -152,19 +152,35 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
...
-                                            <PrettySchedule schedule={w.spec.schedule} />
+                                            {w.spec.schedule != '' ? (
+                                                <PrettySchedule schedule={w.spec.schedule} />
+                                            ) : (
+                                                <>
+                                                    {w.spec.schedules.map(schedule => (
+                                                        <>
+                                                            <PrettySchedule schedule={schedule} />
+                                                            <br />
+                                                        </>
```
agilgur5, line 174: I'm pretty sure that most of these instances can be simplified with a `<p>` paragraph element instead of a `<>` fragment + a `<br />` line break

```suggestion
                                                        <p><PrettySchedule schedule={schedule} /></p>
```

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
```diff
@@ -183,3 +199,18 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
         </Page>
     );
 }
+
+function getCronNextScheduledTime(spec: CronWorkflowSpec): Date {
```
agilgur5, line 203: this naming is quite confusing -- `getNextScheduledTime` already operates on cron schedules (and uses `cron-parser` directly and is imported from `shared/cron.ts`) so adding a function with `Cron` in it sounds identical.

It looks like you wrote "Cron" because it operates on the `CronWorkflowSpec` -- "Spec" would be better naming in that case.
Or "Multi" can differentiate it from the other one, since this operates on multiple schedules

```suggestion
function getSpecNextScheduledTime(spec: CronWorkflowSpec): Date {
```

ui/src/models/cron-workflows.ts
```diff
@@ -21,6 +21,7 @@ export interface CronWorkflowSpec {
     successfulJobsHistoryLimit?: number;
     failedJobsHistoryLimit?: number;
     timezone?: string;
+    schedules?: string[];
```
agilgur5, line 24: this makes more sense on line 18 right below `schedule`

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
```diff
@@ -152,19 +152,35 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
...
                                         <div className='columns small-1'>{w.spec.timezone}</div>
-                                        <div className='columns small-1'>{w.spec.schedule}</div>
+                                        <div className='columns small-1'>
+                                            {w.spec.schedule != ''
+                                                ? w.spec.schedule
+                                                : w.spec.schedules.map(schedule => (
+                                                      <>
+                                                          {schedule}
+                                                          <br />
+                                                      </>
```
agilgur5, line 162:
```suggestion
                                                      <p>{schedule}</p>
```

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
agilgur5: Also how does this look when `schedules` only contains a single schedule? It seems like there may be an unnecessary new line in that case since it is currently treated differently from `schedule`

ui/src/app/cron-workflows/components/cron-workflow-status-viewer.tsx
```diff
@@ -19,7 +19,18 @@ export function CronWorkflowStatusViewer({spec, status}: {spec: CronWorkflowSpec
...
+                                {spec.schedule != '' ? (
+                                    <>
+                                        <code>{spec.schedule}</code> <PrettySchedule schedule={spec.schedule} />
+                                    </>
+                                ) : (
+                                    spec.schedules.map(schedule => (
+                                        <>
+                                            <code>{schedule}</code> <PrettySchedule schedule={schedule} />
+                                            <br />
+                                        </>
```
agilgur5, line 31:
```suggestion
                                        <p><code>{schedule}</code> <PrettySchedule schedule={schedule} /><p/>
```

agilgur5 (commented):
> **NOTE: this is my first time ever doing react /typescript it is expected that the code lacks quality. Please suggest changes**.

This seems to have not gotten a UI SME before merge. It's mostly good, but I left a few comments on improvements below. In particular, the possible unnecessary new line when `schedules` contains only a single schedule is something to check

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
agilgur5, line 1: Line 138 should probably also be `Schedules` now. I can't comment on the line since GH does not allow commenting on unchanged lines that are not close to a diff

cmd/argo/commands/cron/list.go
agilgur5, line 1: Similar on line 69 below, `SCHEDULE` -> `SCHEDULES`

cmd/argo/commands/cron/get.go
```diff
@@ -73,7 +73,7 @@ func getCronWorkflowGet(cwf *wfv1.CronWorkflow) string {
 	out += fmt.Sprintf(fmtStr, "Name:", cwf.ObjectMeta.Name)
 	out += fmt.Sprintf(fmtStr, "Namespace:", cwf.ObjectMeta.Namespace)
 	out += fmt.Sprintf(fmtStr, "Created:", humanize.Timestamp(cwf.ObjectMeta.CreationTimestamp.Time))
-	out += fmt.Sprintf(fmtStr, "Schedule:", cwf.Spec.Schedule)
+	out += fmt.Sprintf(fmtStr, "Schedule:", cwf.Spec.GetScheduleString())
```
agilgur5, line 76:
```suggestion
	out += fmt.Sprintf(fmtStr, "Schedules:", cwf.Spec.GetScheduleString())
```

similar to above

agilgur5 (commented):
Also clients (CLI and UI) should probably have updated labels: `Schedule` -> `Schedules`.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -61,6 +63,8 @@ type CronWorkflowSpec struct {
 	WorkflowMetadata *metav1.ObjectMeta `json:"workflowMetadata,omitempty" protobuf:"bytes,9,opt,name=workflowMeta"`
 	// StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
 	StopStrategy *StopStrategy `json:"stopStrategy,omitempty" protobuf:"bytes,10,opt,name=stopStrategy"`
+	// Schedules is a list of schedules to run the Workflow in Cron format
```
agilgur5, line 66:
```suggestion
	// v3.6 and after: Schedules is a list of schedules to run the Workflow in Cron format
```

This should ideally have a version specifier

agilgur5 (commented):
The [`CronWorkflow`s docs](https://argo-workflows.readthedocs.io/en/latest/cron-workflows/) also need updating around multiple schedules, with version specifiers (see #12581 for examples)

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
eduardodbr: I tested this suggestion and it seems that replacing `<> <br /> </>` with:
```
  <p>
      <PrettySchedule schedule={schedule} />
  </p>
```
 seems to be adding an unnecessary new line after the last schedule, as you can see here:

image

while the current behavior is shown as:

image

no new line

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
eduardodbr: This func calculates the next scheduled time for the combination of cron expressions configured in the cron workflow spec, it handles both"schedule" and "schedules". Changing it to `getSpecNextScheduledTime` seems fine to me.

agilgur5: Yea the word "Cron" was ambiguous in this case, whereas "Spec" is very specific

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
agilgur5: Hmm we use `<p>` elements in the [new markdown title & description feature](https://github.com/argoproj/argo-workflows/pull/12580/files#diff-b9b22a6a512902e9b276dae962c848ad6b87d55b8a4daf66252e9524957274e4R9). Might need to add [similar CSS](https://github.com/argoproj/argo-workflows/pull/12580/files#diff-ab3fd4b11164b06bb926ee013f016665956ebb2d569433d57b5ad177ff3d3470R3) here (`line-height`, `display`, and `vertical-align` in particular). That and it might not work properly if one column uses `<p>` while the other uses `<br />`.

Try that; if it doesn't work, I can take a look on my side

agilgur5: Also were you able to check `schedules` with only one schedule as I mentioned [above](https://github.com/argoproj/argo-workflows/pull/12616#discussion_r1495318106)?

ui/src/app/cron-workflows/components/cron-workflow-status-viewer.tsx
```diff
@@ -19,7 +19,18 @@ export function CronWorkflowStatusViewer({spec, status}: {spec: CronWorkflowSpec
                         title: 'Schedule',
```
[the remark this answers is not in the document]
agilgur5: Yes I considered that being the case. My thought process included:
- we might eventually deprecate and remove `schedule` if we have `schedules`, to simplify the code / remove tech debt etc.
- Argo and k8s already have fields that are plural even if there is only 1 element in the list, for instance: `labels`, `volumes`, `volumeMounts`, `sidecars`, `templates`, `tasks`, `steps`, `nodes`, etc

"Schedule(s)" would be an in between option, although it is inconsistent with other 1 element lists, per above

cmd/argo/commands/cron/list.go
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-abb79cf2d705b99727544690a5ab07807f39a5cb145b5c58f5a2859b5ab767ab))

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-b05cc1f792ee440442970a2925db373864b61fb061bfa49df7e573ba0b99f9bc))

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-b7c27449b2cd699a40591fe01eb016f3532a68103e1617ac0e71f357e274d0c5))

ui/src/app/cron-workflows/components/cron-workflow-list/cron-workflow-list.tsx
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-b7c27449b2cd699a40591fe01eb016f3532a68103e1617ac0e71f357e274d0c5))

ui/src/app/cron-workflows/components/cron-workflow-status-viewer.tsx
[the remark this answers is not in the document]
agilgur5: Fixed in #13369 ([this line](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-c218a12c11e5b521af50e1abd2e003c74afe031dbe90c9ce09a6f9d7451315d8))

ui/src/app/cron-workflows/components/cron-workflow-spec-editior.tsx
```diff
@@ -12,8 +12,22 @@ export function CronWorkflowSpecEditor({onChange, spec}: {spec: CronWorkflowSpec
                 <div className='row white-box__details-row'>
                     <div className='columns small-3'>Schedule</div>
```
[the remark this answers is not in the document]
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-b3e6d00a8da34f2febac2977b1c3c3d5ce62c279f070627ac63366ae2657de5f))

ui/src/models/cron-workflows.ts
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-4393829fdba44d9c8f01c65f50a8dbe38c3e07f8306563d1701e26f4225a9cc5))

cmd/argo/commands/cron/get.go
agilgur5: Fixed in #13369 ([this file](https://github.com/argoproj/argo-workflows/pull/13369/files#diff-85b5b6e77fc3dd06d616bb8a44ba63c95f7bc8193c993e7e0ed04cc203ad6aeb))

PR #12622, opened 2024-02-05
fix: controller option to not watch configmap

pquadri:
Fixes https://github.com/argoproj/argo-workflows/issues/11657

Original PR by @tooptoop4 here: https://github.com/argoproj/argo-workflows/pull/12400


tooptoop4:
:shipit:

pquadri:
not sure about the failed test, probably a flaky one but i can't retry the action :/

agilgur5:
I gave it a retry. That was an Executor test, so sounds unrelated, but I'm not sure I've seen that specific test flake before (though some related tests have before `TestStopBehavior`) 🤔

docs/environment-variables.md
```diff
@@ -53,6 +53,7 @@ This document outlines environment variables that can be used to customize behav
...
 | `WORKFLOW_GC_PERIOD`                     | `time.Duration`     | `5m`                                                                                        | The periodicity for GC of workflows.                                                                                                                                                                                                                                     |
 | `SEMAPHORE_NOTIFY_DELAY`                 | `time.Duration`     | `1s`                                                                                        | Tuning Delay when notifying semaphore waiters about availability in the semaphore                                                                                                                                                                                        |
+| `WATCH_CONFIGMAPS` | `bool` | `true` | Whether to watch the Controller's ConfigMap and semaphore ConfigMaps for run-time changes. When disabled, the Controller will only read ConfigMaps once and will have to be manually restarted to pick up new changes. |
```
agilgur5, line 56:
```suggestion
| `WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS` | `bool` | `true` | Whether to watch the Controller's ConfigMap and semaphore ConfigMaps for run-time changes. When disabled, the Controller will only read these ConfigMaps once and will have to be manually restarted to pick up new changes. |
```

just a small clarification. there are also memoization configmaps, `valueFrom` configmaps, `artifact-repos` configmaps, plugin configmaps, etc, so I figure we should be more specific

workflow/controller/controller.go
```diff
@@ -308,7 +309,10 @@ func (wfc *WorkflowController) Run(ctx context.Context, wfWorkers, workflowTTLWo
 		log.Fatal(err)
 	}
 
-	go wfc.runConfigMapWatcher(ctx.Done())
+	if os.Getenv("WATCH_CONFIGMAPS") != "false" {
```
agilgur5, line 312:
```suggestion
	if os.Getenv("WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS") != "false" {
```

same as above

agilgur5 (commented):
Took a look at this in more depth now that it passes. Just one small addition to #12400, otherwise LGTM. Thanks for finishing this up!

agilgur5 (approved):

tooptoop4:
@xrafhue the os variable should be WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS

xrafhue:
@tooptoop4 , after 2 days in all our clusters the logs are more generated. Thanks a lot.

PR #12623, opened 2024-02-05
fix: argo server graceful restart when configmap changed

shuangkun:
Argo server graceful restart when configmap changed


server/apiserver/argoserver.go
```diff
@@ -259,7 +268,8 @@ func (as *argoServer) Run(ctx context.Context, port int, browserOpenFunc func(st
 	httpL := tcpm.Match(cmux.HTTP1Fast())
 	grpcL := tcpm.Match(cmux.Any())
 
-	go eventServer.Run(as.stopCh)
+	go as.runConfigMapWatcher(as.stopCh)
```
tooptoop4, line 271: this will cause lots of logs on k8s 1.27+ like https://github.com/argoproj/argo-workflows/issues/11657

shuangkun: Are there any solutions already? Should we change it to informer? I am referring to controller

agilgur5: Supposedly changing to an Informer will fix it. It's on my to-do list as a contributor abandoned a PR I was reviewing. Should hopefully get to it this week (per my status, the, uh, kidney stone has made things a little harder)

agilgur5: > this will cause lots of logs on k8s 1.27+ like #11657

also, to be clear, it's the `RetryWatcher` on line 459 below that will cause this more specifically

agilgur5 (changes requested):
So this is a fix for one of the very first issues I filed: #10970.

But I don't think a graceful restart is a good approach, since it causes unexpected downtime. Even in an HA scenario with multiple replicas, all of them will go down simultaneously with this code. Comparatively, the user-land workaround I mentioned in the issue of using Reloader is safer and more stable, as it will cause a rolling deploy.

server/apiserver/argoserver.go
shuangkun: > Supposedly changing to an Informer will fix it. It's on my to-do list as a contributor abandoned a PR I was reviewing. Should hopefully get to it this week (per my status, the, uh, kidney stone has made things a little harder, pun unintended)

Okay, I'm looking forward to your PR. Will your PR solve the problem of argo server re-downloading configmap?

agilgur5: No, just #11657 for the Controller

shuangkun:
> So this is a fix for one of the very first issues I filed: #10970.
> 
> But I don't think a graceful restart is a good approach, since it causes unexpected downtime. Even in an HA scenario with multiple replicas, all of them will go down simultaneously with this code. Comparatively, the user-land workaround I mentioned in the issue of using Reloader or Helm is safer and more stable, as it will cause a rolling deploy.

Yes, it would be best if it could be reloaded like the controller. Now I need to restart Argo Server if I want the relevant configuration to take effect, which makes users confused and will also make the service unavailable.

agilgur5:
> will also make the service unavailable.

The user-land workarounds don't have this issue as they trigger rolling deploys, so an HA Server set-up with appropriate `maxUnavailable` etc should not become unavailable.

That is strictly better than the current approach of the PR; I'd prefer to point folks to a zero-downtime workaround than supporting a (unflagged) feature that causes downtime.

> Yes, it would be best if it could be reloaded like the controller.

Yea it requires some non-trivial refactoring as I mentioned in the issue. Otherwise I would have added that feature already. The workarounds are fairly simple in the interim.

agilgur5:
@shuangkun I think we should close this PR as the current approach is incompatible with HA and this hasn't received updates in ~2 weeks. If you're interested in refactoring the Server init to support in-process, zero-downtime hot reloading, that might be best as a separate PR too

shuangkun:
> @shuangkun I think we should close this PR as the current approach is incompatible with HA and this hasn't received updates in ~2 weeks. If you're interested in refactoring the Server init to support in-process, zero-downtime hot reloading, that might be best as a separate PR too

Sorry for not updating for a long time, I am on a long vacation. I will study how to perform a hot restart later. Thanks.

agilgur5:
Ah ok, take your time and enjoy your vacation!

PR #12419, opened 2023-12-27
feat: delete pods in parallel to speed up retryworkflow

shuangkun:
I have a big worklfow. When it failed and retry, it will be slowly. Latency increases with the number of pods. Always letting my automated system（interact with argo server use https) time out.

```
time="2023-12-26T13:02:01.560Z" level=info duration=1m18.730806898s method=PUT path=/api/v1/workflows/default/catch-workflow-8c48s/retry size=194 status=409
```

I found that the main time spent was deleting pods. 

```
time="2023-12-26T13:02:00.126Z" level=info msg="Deleting pod" podDeleted=catch-workflow-8c48s-read-ligands-from-vswdb-op-bmvou-9dzxr-2538960512
time="2023-12-26T13:02:00.328Z" level=info msg="Deleting pod" podDeleted=catch-workflow-8c48s-read-ligands-from-vswdb-op-bmvou-9dzxr-2824246174
time="2023-12-26T13:02:00.527Z" level=info msg="Deleting pod" podDeleted=catch-workflow-8c48s-read-ligands-from-vswdb-op-bmvou-9dzxr-3817503941
time="2023-12-26T13:02:00.731Z" level=info msg="Deleting pod" podDeleted=catch-workflow-8c48s-read-ligands-from-vswdb-op-bmvou-9dzxr-1699455227
time="2023-12-26T13:02:00.925Z" level=info msg="Deleting pod" podDeleted=catch-workflow-8c48s-read-ligands-from-vswdb-op-bmvou-9dzxr-4168132452
time="2023-12-26T13:02:01.133Z" level=info msg="Deleting pod" podDeleted=catch-workflow-8c48s-read-ligands-from-vswdb-op-bmvou-9dzxr-1982889595
```

So I want make deleting pods parallel.

Issue #12538, opened 2024-01-17
Move `retry` Pod deletions out of Server and into Controller for proper separation of duties


agilgur5:
# Summary

Currently, the Server and Controller are architected & intended to be independent, with the Server not being strictly necessary for any operations. Most of the Server's functionality is to be a simple CRUD wrapper where a user could replicate that functionality themselves via `kubectl`. When the Server has to communicate with the Controller, it typically signals to it by adding a label to a Workflow.

This separation of duties is important to keep consistent and is currently true for all but one case: the `retry` operation currently has the Server delete Pods of a Workflow, which is something that the Controller should do instead. The Server shouldn't need permissions to delete Pods either as it [currently does](https://github.com/argoproj/argo-workflows/blob/1202ae473a9f047cccef820a80ecb97c59a02b92/manifests/cluster-install/argo-server-rbac/argo-server-clusterole.yaml#L31).

## Use Cases

I (and then others) noticed this in https://github.com/argoproj/argo-workflows/pull/12105#discussion_r1376399199 and https://github.com/argoproj/argo-workflows/pull/12419#issuecomment-1896552339 and were pretty surprised when we saw this. 

Removing this functionality from the Server will make it more secure by not having `delete` `pods` permissions which the Controller [already has](https://github.com/argoproj/argo-workflows/blob/1202ae473a9f047cccef820a80ecb97c59a02b92/manifests/cluster-install/workflow-controller-rbac/workflow-controller-clusterrole.yaml#L18).

It will also make it possible to do a `retry` with just `kubectl` by adding a label to the Workflow CR, as is intended and as was thought as possible per https://github.com/argoproj/argo-workflows/discussions/12027#discussioncomment-7316157.

## Implementation details

1. The Server should only label the Workflow
1. The Server should no longer need `delete` `pods` permissions
1. The Controller should detect that label as a trigger for the retry
1. The Controller should perform the Pod deletion and then initiate the retry
1. The Controller should handle bugs / missing functionality such as that of #12105 / #12028

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritize the proposals with the most 👍.

terrytangyuan:
SGTM

Joibel:
This would be covered by the second part of #16741

PR #12624, opened 2024-02-05
feat: speed up retry archived workflow

shuangkun:
https://github.com/argoproj/argo-workflows/pull/12419 
Since I accelerated the deletion of pods during retry workflow, it works well and is much faster. I want to speed up the retry archived workflow too.

### Motivation

Speed up retry archived workflow. Let a large archived workflow (more than 8000 pods) can be successfully retryed within 1 minute.

### Modifications

Clean up failed nodes in parallel.

### Verification

e2e test.

### DIFF
--- a/server/workflowarchive/archived_workflow_server.go
+++ b/server/workflowarchive/archived_workflow_server.go
@@ -25,7 +25,6 @@ import (
 	workflowarchivepkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflowarchive"
 	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow"
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
-	"github.com/argoproj/argo-workflows/v3/server/apiserver"
 	"github.com/argoproj/argo-workflows/v3/server/auth"
 	sutils "github.com/argoproj/argo-workflows/v3/server/utils"
 	"github.com/argoproj/argo-workflows/v3/workflow/util"
@@ -294,6 +293,13 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
 
+		var backoff = wait.Backoff{
+			Steps:    5,
+			Duration: 500 * time.Millisecond,
+			Factor:   1.0,
+			Jitter:   0.1,
+		}
+
 		errCh := make(chan error, len(podsToDelete))
 		var wg sync.WaitGroup
 		parallelPodNum := make(chan string, 500)
@@ -303,7 +309,7 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 			wg.Add(1)
 			go func(podName string) {
 				defer wg.Done()
-				err := wait.ExponentialBackoff(apiserver.Backoff, func() (bool, error) {
+				err := wait.ExponentialBackoff(backoff, func() (bool, error) {
 					err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
 					if err != nil && !apierr.IsNotFound(err) {
 						klog.Errorf("Failed to delete pod %s: %v", podName, err)
--- a/server/workflowarchive/archived_workflow_server.go
+++ b/server/workflowarchive/archived_workflow_server.go
@@ -3,6 +3,8 @@ package workflowarchive
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 	"os"
 	"regexp"
 	"sort"
@@ -23,10 +25,10 @@ import (
 	workflowarchivepkg "github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflowarchive"
 	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow"
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+	"github.com/argoproj/argo-workflows/v3/server/apiserver"
 	"github.com/argoproj/argo-workflows/v3/server/auth"
-	"github.com/argoproj/argo-workflows/v3/workflow/util"
-
 	sutils "github.com/argoproj/argo-workflows/v3/server/utils"
+	"github.com/argoproj/argo-workflows/v3/workflow/util"
 )
 
 const disableValueListRetrievalKeyPattern = "DISABLE_VALUE_LIST_RETRIEVAL_KEY_PATTERN"
@@ -294,16 +296,25 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 
 		errCh := make(chan error, len(podsToDelete))
 		var wg sync.WaitGroup
-		wg.Add(len(podsToDelete))
+		parallelPodNum := make(chan string, 500)
 		for _, podName := range podsToDelete {
 			log.WithFields(log.Fields{"podDeleted": podName}).Info("Deleting pod")
+			parallelPodNum <- podName
+			wg.Add(1)
 			go func(podName string) {
 				defer wg.Done()
-				err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
-				if err != nil && !apierr.IsNotFound(err) {
+				err := wait.ExponentialBackoff(apiserver.Backoff, func() (bool, error) {
+					err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
+					if err != nil && !apierr.IsNotFound(err) {
+						klog.Errorf("Failed to delete pod %s: %v", podName, err)
+						return false, nil
+					}
+					return true, nil
+				})
+				if err != nil {
 					errCh <- err
-					return
 				}
+				<-parallelPodNum
 			}(podName)
 		}
 		wg.Wait()
--- a/server/workflowarchive/archived_workflow_server.go
+++ b/server/workflowarchive/archived_workflow_server.go
@@ -8,11 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
-
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog/v2"
 
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
@@ -294,39 +290,7 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
 
-		var backoff = wait.Backoff{
-			Steps:    5,
-			Duration: 500 * time.Millisecond,
-			Factor:   1.0,
-			Jitter:   0.1,
-		}
-
-		errCh := make(chan error, len(podsToDelete))
-		var wg sync.WaitGroup
-		parallelPodNum := make(chan string, 500)
-		for _, podName := range podsToDelete {
-			log.WithFields(log.Fields{"podDeleted": podName}).Info("Deleting pod")
-			parallelPodNum <- podName
-			wg.Add(1)
-			go func(podName string) {
-				defer wg.Done()
-				err := wait.ExponentialBackoff(backoff, func() (bool, error) {
-					err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
-					if err != nil && !apierr.IsNotFound(err) {
-						klog.Errorf("Failed to delete pod %s: %v", podName, err)
-						return false, nil
-					}
-					return true, nil
-				})
-				if err != nil {
-					errCh <- err
-				}
-				<-parallelPodNum
-			}(podName)
-		}
-		wg.Wait()
-
-		err = util.ErrorFromChannel(errCh)
+		err = util.DeletePodsInParallel(ctx, podsToDelete, kubeClient, wf.Namespace)
 		if err != nil {
 			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
--- a/server/workflowarchive/archived_workflow_server.go
+++ b/server/workflowarchive/archived_workflow_server.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -291,12 +292,25 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req 
 			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
 
+		errCh := make(chan error, len(podsToDelete))
+		var wg sync.WaitGroup
+		wg.Add(len(podsToDelete))
 		for _, podName := range podsToDelete {
 			log.WithFields(log.Fields{"podDeleted": podName}).Info("Deleting pod")
-			err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
-			if err != nil && !apierr.IsNotFound(err) {
-				return nil, sutils.ToStatusError(err, codes.Internal)
-			}
+			go func(podName string) {
+				defer wg.Done()
+				err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
+				if err != nil && !apierr.IsNotFound(err) {
+					errCh <- err
+					return
+				}
+			}(podName)
+		}
+		wg.Wait()
+
+		err = util.ErrorFromChannel(errCh)
+		if err != nil {
+			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
 
 		wf.ObjectMeta.ResourceVersion = ""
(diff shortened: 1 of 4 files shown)


server/workflowarchive/archived_workflow_server.go
```diff
@@ -291,12 +292,25 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req
...
-				return nil, sutils.ToStatusError(err, codes.Internal)
-			}
+			go func(podName string) {
+				defer wg.Done()
+				err := kubeClient.CoreV1().Pods(wf.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
+				if err != nil && !apierr.IsNotFound(err) {
+					errCh <- err
+					return
+				}
+			}(podName)
```
isubasinghe, line 307: @agilgur5 do you have any opinions on this? My concern here is that this might very easily hit rate limitations. 
I see a similar PR has already been merged in though. 

I believe there are some timing related helper functions that can be used here, maybe that would help?  
I am concerned because I think some k8s platforms have relatively small QPS and for a large enough workflow, we might hit that easily.

server/workflowarchive/archived_workflow_server.go
```diff
@@ -291,12 +292,25 @@ func (w *archivedWorkflowServer) RetryArchivedWorkflow(ctx context.Context, req
 			return nil, sutils.ToStatusError(err, codes.Internal)
 		}
 
+		errCh := make(chan error, len(podsToDelete))
+		var wg sync.WaitGroup
+		wg.Add(len(podsToDelete))
 		for _, podName := range podsToDelete {
```
Joibel, line 298: It feels like this and the partner could be extracted out to a common helper function.

Joibel (changes requested):

Joibel:
Both this and the other speed up for plain retry feel like they will only be effective speedups in the situation where the controller isn't busy. I share @isubasinghe concerns over this effectively stealing more bandwidth.

I am unsure why all of these pod deletions can't just be put on the pod deletion queue.

server/workflowarchive/archived_workflow_server.go
shuangkun: Yes, this is possible. I can extract a function.

shuangkun:
> Both this and the other speed up for plain retry feel like they will only be effective speedups in the situation where the controller isn't busy. I share @isubasinghe concerns over this effectively stealing more bandwidth.
> 
> 
> 
> I am unsure why all of these pod deletions can't just be put on the pod deletion queue.

Yes, I will adjust the concurrency and burst

agilgur5:
> Both this and the other speed up for plain retry feel like they will only be effective speedups in the situation where the controller isn't busy. I share @isubasinghe concerns over this effectively stealing more bandwidth.
> 
> I am unsure why all of these pod deletions can't just be put on the pod deletion queue.

Regarding these concerns, note that these changes are currently in the Server. The pod deletion logic really shouldn't be in the Server to _begin with_, see #12538.

Moving the logic to the Controller will allow for some more advanced techniques, like using the queue. Note that things like retries should probably be prioritized over other GC, as they are interactive with a user and so a long wait would not necessarily be optimal.

isubasinghe:
> That would be the root cause fix here.

agreed but I think for now, probably better off not doing this regardless? Or perhaps do it with exponential backoff.

shuangkun:
What should we do now. For a large workflow with more than 4,000 nodes, retry often takes more than a minute. After acceleration, it takes 10 seconds. I think we can adjust the concurrency to 500, if we are afraid of occupying bandwidth.

isubasinghe:
> What should we do now. For a large workflow with more than 4,000 nodes, retry often takes more than a minute. After acceleration, it takes 10 seconds. I think we can adjust the concurrency to 500, if we are afraid of occupying bandwidth.

Have you seen the wait package in the k8s api? I think it might be a good idea to use the exponential backoff from that?
Regardless, probably not a good idea to launch this many go routines, you likely aren't going to see much benefit after the point where you've saturated your networking hardware.

shuangkun:
> > What should we do now. For a large workflow with more than 4,000 nodes, retry often takes more than a minute. After acceleration, it takes 10 seconds. I think we can adjust the concurrency to 500, if we are afraid of occupying bandwidth.
> 
> Have you seen the wait package in the k8s api? I think it might be a good idea to use the exponential backoff from that? Regardless, probably not a good idea to launch this many go routines, you likely aren't going to see much benefit after the point where you've saturated your networking hardware.

Thanks, Do you mean use exponential to Reduce stress when delete pod error? I add exponential backoff and can you have a look?

juliev0:
Is anyone up for being Assignee on this one? (I see multiple people have reviewed) and it’s a “prioritized-review”.  Thanks

Joibel:
I think we should put the effort into doing the right thing, and move these deletes to a prioritised queue in the controller as suggested by Anton.

Even with the exponential backoff code this is still attempting to
* hammer the kubernetes API using go routines to dispatch massive numbers of deletes at once
* is in the server not the controller

I also really don't understand why an archived workflow still has pods around. I must admit to not having looked or tried this

Surely, due to owner references at the point a workflow is archived the workflow is deleted, and the pods it owns should be deleted.

shuangkun:
> I think we should put the effort into doing the right thing, and move these deletes to a prioritised queue in the controller as suggested by Anton.
> 
> Even with the exponential backoff code this is still attempting to
> 
> * hammer the kubernetes API using go routines to dispatch massive numbers of deletes at once
> * is in the server not the controller
> 
> I also really don't understand why an archived workflow still has pods around. I must admit to not having looked or tried this
> 
> Surely, due to owner references at the point a workflow is archived the workflow is deleted, and the pods it owns should be deleted.

Once "archive: true" is set, the workflow will be archived when completed, regardless of whether it is deleted. We have no control over whether the user uses retryWorkflow or retryArchiveWorkflow.

I don't think we need to worry too much about the concurrency pressure of 500, it's a very simple operation. We make heavy use of concurrency in our code.

As for moving the code to the controller, it may be the final solution, but it will take time to refactor.
This PR does not introduce new issues, but it can solve the problem of too slow retry for large workflows(> 4000).

Joibel:
> > I think we should put the effort into doing the right thing, and move these deletes to a prioritised queue in the controller as suggested by Anton.
> > Even with the exponential backoff code this is still attempting to
> > 
> > * hammer the kubernetes API using go routines to dispatch massive numbers of deletes at once
> > * is in the server not the controller
> > 
> > I also really don't understand why an archived workflow still has pods around. I must admit to not having looked or tried this
> > Surely, due to owner references at the point a workflow is archived the workflow is deleted, and the pods it owns should be deleted.
> 
> Once "archive: true" is set, the workflow will be archived when completed, regardless of whether it is deleted. We have no control over whether the user uses retryWorkflow or retryArchiveWorkflow.

Indeed, but this PR is called "speed up retry archived workflow". I know the code prior to this commit attempts to delete a lot of pods, but I'm wondering why. Once archiving has happened, why are there any pods in the cluster? I'm questioning why the code exists at all - it will take time to fail to delete 4000 pods that don't exist.

> I don't think we need to worry too much about the concurrency pressure of 500, it's a very simple operation. We make heavy use of concurrency in our code.

I am not at all worried about the use of go routines.

I am worried about being a good kubernetes citizen. Elsewhere in the code we attempt to rate limit how hard we push the kubernetes API. This code change attempts to be selfish and greedy with API usage in order to achieve the goal of achieving a goal on behalf of one user. It is doing this at the expense of other users in the system.

Kubernetes API is often deliberately rate limited, and often rate limited by external actors. This code change attempts to use as much of this bandwidth as possible for this one thing. On any busy system this must do that at the expense of other users (whether human or computer) of the system. This is what I mean by being a bad citizen. What I would expect if we adopted this code is that we will get issues raised by people who object to this "denial of service" to other part of their kubernetes system. They have no option not to have this behavior, but I don't think making this optional is the right approach either.

> As for moving the code to the controller, it may be the final solution, but it will take time to refactor. This PR does not introduce new issues, but it can solve the problem of too slow retry for large workflows(> 4000).

Although a goal of making large workflow retries work faster is a good goal, I don't think that being selfish about doing it is the right approach. I think we need to take time to refactor, and perhaps along side this provide better user feedback on the retry.

shuangkun:
> > > I think we should put the effort into doing the right thing, and move these deletes to a prioritised queue in the controller as suggested by Anton.
> > > Even with the exponential backoff code this is still attempting to
> > > 
> > > * hammer the kubernetes API using go routines to dispatch massive numbers of deletes at once
> > > * is in the server not the controller
> > > 
> > > I also really don't understand why an archived workflow still has pods around. I must admit to not having looked or tried this
> > > Surely, due to owner references at the point a workflow is archived the workflow is deleted, and the pods it owns should be deleted.
> > 
> > 
> > Once "archive: true" is set, the workflow will be archived when completed, regardless of whether it is deleted. We have no control over whether the user uses retryWorkflow or retryArchiveWorkflow.
> 
> Indeed, but this PR is called "speed up retry archived workflow". I know the code prior to this commit attempts to delete a lot of pods, but I'm wondering why. Once archiving has happened, why are there any pods in the cluster? I'm questioning why the code exists at all - it will take time to fail to delete 4000 pods that don't exist.
> 
> > I don't think we need to worry too much about the concurrency pressure of 500, it's a very simple operation. We make heavy use of concurrency in our code.
> 
> I am not at all worried about the use of go routines.
> 
> I am worried about being a good kubernetes citizen. Elsewhere in the code we attempt to rate limit how hard we push the kubernetes API. This code change attempts to be selfish and greedy with API usage in order to achieve the goal of achieving a goal on behalf of one user. It is doing this at the expense of other users in the system.
> 
> Kubernetes API is often deliberately rate limited, and often rate limited by external actors. This code change attempts to use as much of this bandwidth as possible for this one thing. On any busy system this must do that at the expense of other users (whether human or computer) of the system. This is what I mean by being a bad citizen. What I would expect if we adopted this code is that we will get issues raised by people who object to this "denial of service" to other part of their kubernetes system. They have no option not to have this behavior, but I don't think making this optional is the right approach either.
> 
> > As for moving the code to the controller, it may be the final solution, but it will take time to refactor. This PR does not introduce new issues, but it can solve the problem of too slow retry for large workflows(> 4000).
> 
> Although a goal of making large workflow retries work faster is a good goal, I don't think that being selfish about doing it is the right approach. I think we need to take time to refactor, and perhaps along side this provide better user feedback on the retry.



> Both this and the other speed up for plain retry feel like they will only be effective speedups in the situation where the controller isn't busy. I share @isubasinghe concerns over this effectively stealing more bandwidth.
> 
> I am unsure why all of these pod deletions can't just be put on the pod deletion queue.



> > > I think we should put the effort into doing the right thing, and move these deletes to a prioritised queue in the controller as suggested by Anton.
> > > Even with the exponential backoff code this is still attempting to
> > > 
> > > * hammer the kubernetes API using go routines to dispatch massive numbers of deletes at once
> > > * is in the server not the controller
> > > 
> > > I also really don't understand why an archived workflow still has pods around. I must admit to not having looked or tried this
> > > Surely, due to owner references at the point a workflow is archived the workflow is deleted, and the pods it owns should be deleted.
> > 
> > 
> > Once "archive: true" is set, the workflow will be archived when completed, regardless of whether it is deleted. We have no control over whether the user uses retryWorkflow or retryArchiveWorkflow.
> 
> Indeed, but this PR is called "speed up retry archived workflow". I know the code prior to this commit attempts to delete a lot of pods, but I'm wondering why. Once archiving has happened, why are there any pods in the cluster? I'm questioning why the code exists at all - it will take time to fail to delete 4000 pods that don't exist.
> 
> > I don't think we need to worry too much about the concurrency pressure of 500, it's a very simple operation. We make heavy use of concurrency in our code.
> 
> I am not at all worried about the use of go routines.
> 
> I am worried about being a good kubernetes citizen. Elsewhere in the code we attempt to rate limit how hard we push the kubernetes API. This code change attempts to be selfish and greedy with API usage in order to achieve the goal of achieving a goal on behalf of one user. It is doing this at the expense of other users in the system.
> 
> Kubernetes API is often deliberately rate limited, and often rate limited by external actors. This code change attempts to use as much of this bandwidth as possible for this one thing. On any busy system this must do that at the expense of other users (whether human or computer) of the system. This is what I mean by being a bad citizen. What I would expect if we adopted this code is that we will get issues raised by people who object to this "denial of service" to other part of their kubernetes system. They have no option not to have this behavior, but I don't think making this optional is the right approach either.
> 
> > As for moving the code to the controller, it may be the final solution, but it will take time to refactor. This PR does not introduce new issues, but it can solve the problem of too slow retry for large workflows(> 4000).
> 
> Although a goal of making large workflow retries work faster is a good goal, I don't think that being selfish about doing it is the right approach. I think we need to take time to refactor, and perhaps along side this provide better user feedback on the retry.

After the workflow is archived, there may still be pods in the cluster, if the workflow is not deleted and no pod recycling policy is set.

Thanks，maybe you are right, but I still think that the problem needs to be fixed step by step, and it is difficult to solve all the problems at once. If the majority of people think that solving the problem requires refactoring this piece of code, I will do it. We really need this feature.

In the server, there are still some places that use this method to accelerate，For example:
https://github.com/argoproj/argo-workflows/blob/88304568af5b8222344cde8e5b96d1a8484a50f8/util/logs/pods-logger.go#L74

Joibel:
> After the workflow is archived, there may still be pods in the cluster, if the workflow is not deleted and no pod recycling policy is set.

I don't know of how you get archiving not to delete the workflow. It would seem very odd not to have it in both the cluster and the archive.

I also don't think pod GC has any effect here, because pods are owned by the workflow, so will be deleted.

I'm asking for proof that what you're saying is possible, I just don't know a way, but I haven't looked in detail.

> Thanks，maybe you are right, but I still think that the problem needs to be fixed step by step, and it is difficult to solve all the problems at once. If the majority of people think that solving the problem requires refactoring this piece of code, I will do it. We really need this feature.

@juliev0, @agilgur5 and @isubasinghe have commented in this thread, so I'll ping them here.

> In the server, there are still some places that use this method to accelerate，For example:
> 
> https://github.com/argoproj/argo-workflows/blob/88304568af5b8222344cde8e5b96d1a8484a50f8/util/logs/pods-logger.go#L74

I know. In this case it is justified, and is not so evil because it's only hitting live pods.

shuangkun:
> > After the workflow is archived, there may still be pods in the cluster, if the workflow is not deleted and no pod recycling policy is set.
> 
> I don't know of how you get archiving not to delete the workflow. It would seem very odd not to have it in both the cluster and the archive.
> 
> I also don't think pod GC has any effect here, because pods are owned by the workflow, so will be deleted.
> 
> I'm asking for proof that what you're saying is possible, I just don't know a way, but I haven't looked in detail.

I did't test in latest:
I can see workflow and pod in cluster:
```
tianshuangkun@B-QGHAMD6M-2124 test % argo list
NAME                    STATUS      AGE   DURATION   PRIORITY
test-workflow-gcszp     Failed      2m    2m         0
sparkly-poochenheimer   Succeeded   8m    42s        0
tianshuangkun@B-QGHAMD6M-2124 test % kubectl get pod
NAME                    READY   STATUS      RESTARTS   AGE
sparkly-poochenheimer   0/2     Completed   0          8m41s
test-workflow-gcszp     0/2     Error       0          2m47s
```
and in archive:
image
Maybe the logic has changed recently？

> > Thanks，maybe you are right, but I still think that the problem needs to be fixed step by step, and it is difficult to solve all the problems at once. If the majority of people think that solving the problem requires refactoring this piece of code, I will do it. We really need this feature.
> 
> @juliev0, @agilgur5 and @isubasinghe have commented in this thread, so I'll ping them here.
> 
> > In the server, there are still some places that use this method to accelerate，For example:
> > https://github.com/argoproj/argo-workflows/blob/88304568af5b8222344cde8e5b96d1a8484a50f8/util/logs/pods-logger.go#L74
> 
> I know. In this case it is justified, and is not so evil because it's only hitting live pods.

If pod didn't gc, fetch log will hit all pod?

agilgur5:
> > Thanks，maybe you are right, but I still think that the problem needs to be fixed step by step, and it is difficult to solve all the problems at once. If the majority of people think that solving the problem requires refactoring this piece of code, I will do it. We really need this feature.
> 
> @juliev0, @agilgur5 and @isubasinghe have commented in this thread, so I'll ping them here.

Yes, I definitely think it should be refactored and would definitely prefer that, that's why I said it. I did choose the "comment" code review option intentionally though -- I don't really have a strong opinion on a "band-aid" like this, but I generally prefer long-term, root cause fixes rather than adding tech debt. If someone is willing to approve, I don't necessarily mind too much, but I will admit it is a bit concerning when multiple PRs happen to the same place that should all be refactored -- at the point it may even take the same amount of time to refactor.

agilgur5:
Regarding the Workflow Archive -- there are race conditions when a Workflow is both in the Archive and still a resource in the cluster. Even in the Archive process itself, the Workflow must exist while being Archived.

But also in general, the Archive does not require that a Workflow be deleted, that is up to the TTL/retention etc. Workflows are labeled for archiving [when they are completed](https://github.com/argoproj/argo-workflows/blob/88304568af5b8222344cde8e5b96d1a8484a50f8/workflow/controller/operator.go#L2362). This label is then [picked up by the Informer](https://github.com/argoproj/argo-workflows/blob/88304568af5b8222344cde8e5b96d1a8484a50f8/workflow/controller/controller.go#L1070-L1084) which kicks off archiving.

agilgur5 (commented):
@shuangkun it looks like you're going to address #12538 and refactor the Server/Controller logic properly? I think that should be a separate PR, as nearly everything here is about the archived retry logic, including your PR description. I would leave this PR in its previous state for posterity and create a new PR.

Refactoring should also modify the manifests -- the Server should no longer need writeable RBAC for Pods, per #12538 1. The other implementation details listed there are also relevant. Changes to that approach -- especially spec changes -- should be discussed.

shuangkun:
> @shuangkun it looks like you're going to address #12538 and refactor the Server/Controller logic properly? I think that should be a separate PR, as nearly everything here is about the archived retry logic, including your PR description. I would leave this PR in its previous state for posterity and create a new PR.
> 
> Refactoring should also modify the manifests -- the Server should no longer need writeable RBAC for Pods, per #12538 1. The other implementation details listed there are also relevant. Changes to that approach -- especially spec changes -- should be discussed.

OK, thanks,  I will open a new PR.

PR #12651, opened 2024-02-09
chore(deps): upgrade argoproj/pkg version

### LINKED ISSUE #12650: Support EKS Pod Identity to access AWS s3
# Summary

Argo Workflow to support accessing AWS s3 using EKS Pod Identity , this is a additional way of assuming a IAM role than using IRSA. 

The change is a bump in the version of to the aws-sdk-go-v2 minimum packages credentials->v.1.16.0 and config->v1.25.0

These dependencies are already picked up in latest version of argoproj/pkg so the change is to update go.mod/go.sum to pickup the new version https://github.com/argoproj/argo-workflows/blob/main/go.mod#L16

## Use Cases

Instead of configuring service account with annotation (IRSA), now the configuration of associating an IAM role to a sepecific service account is done on the AWS API level nothing to be done in kubernetes API.

---

csantanapr:
Fixes #12650 

### Motivation

Update aws-sdk version for EKS Pod Identity

### Modifications

update go deps


terrytangyuan (approved):
Thanks!

go.mod
```diff
@@ -47,15 +47,15 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966
 	github.com/soheilhy/cmux v0.1.5
-	github.com/spf13/cobra v1.7.0
+	github.com/spf13/cobra v1.8.0
```
agilgur5, line 50: this and the `crypto` on line 58 below seem to be unrelated changes?

rafilkmp3:
Can someone confirm if it is included in release 3.5.7

agilgur5:
The answer is the same as https://github.com/argoproj/argo-workflows/issues/12949#issuecomment-2073712930 for the same reasons

PR #6767, opened 2021-09-20
feat: Add env vars config for argo-server and workflow-controller

prichrd:
This is a proposal for https://github.com/argoproj/argo-workflows/issues/5914. It uses https://github.com/spf13/viper to parse environment variables from the flag names.

I tested this like this:
image
image

PR #6596, opened 2021-08-24
fix: Fix `gosec` warnings, disable pprof by default. Fixes #6594

alexec:
Fixes #6594

PR #12652, opened 2024-02-10
fix(build): check for env vars in all dirs

agilgur5:
### Motivation

- previously, some of the `cmd/` subdirs were missed
- and `server/` was missed in the `check-documented` func
  - instead use a variable to have consistent directories for both
- `env.Get*(` calls were missed as well

I was responding to [this Slack thread](https://cloud-native.slack.com/archives/C04SURUPDL2/p1707383452116839?thread_ts=1707319125.512419&cid=C04SURUPDL2) in `#argo-sig-scalability` and reviewing https://github.com/argoproj/argo-cd/pull/17068 when I re-checked the Server's `pprof` code and found an undocumented flag. So began the whole process of fixing the reason why it and other env vars were undocumented etc

### Modifications

- <details> <summary> add missing env vars to doc </summary>

  - `ARGO_DEBUG_PAUSE_*`, `ALLOWED_LINK_PROTOCOL`, `ARGO_ARTIFACT_SERVER`, `ARGO_PPROF`, `ARGO_SERVER_METRICS_AUTH`, `BASE_HREF`
  - link to `pprof` docs when referencing it
  - also consistently alphabetize the Server env vars -- `GRPC_MESSAGE_SIZE` is not last alphabetically

</details>

- remove some no longer used env vars
  - `ARGO_SERVER_PPROF` was no longer used, meanwhile `ARGO_PPROF` was undocumented for the Server (was only documented for the Controller)
    - it was added in #6004 and then replaced in #6596 but this code was leftover

- <details> <summary> refactor <code>check-env-doc.sh</code> </summary>

  - consistent 2 space indentation, not mixed 2 or 4 space
  - UX: list out all missed variables before exiting
    - previously it would exit after the first miss, meaning you'd have to fix -> run -> fix -> run until you got everything - and it's not the fastest script given how many files there are to parse through, so it's a productivity increase
  - fix some shellcheck issues with loops / arrays: https://www.shellcheck.net/wiki/SC2031 and https://www.shellcheck.net/wiki/SC2207
  - something something "portable shell code" something something "maybe use Python instead next time"

</details>

<details> <summary> <h3>Verification</h3> </summary>

#### Script
1. In my `devcontainer` (for a somewhat standardized version of `bash`), ran: `bash hack/check-env-doc.sh`
   - that's how I got this list
1. Also tested removing a few env vars and confirmed it would list all of them in one go now. 

#### Docs
1. `mkdocs build`
1. `open site/environment-variables/index.html`
1. Screenshot with new env vars in the tables: 
Screenshot 2024-02-10 at 6 57 52 PM

</details>

### Notes to Reviewers

1. Can skip over all the `docs/cli` changes as it's all generated one-liner changes due to the `conn.go` description change
1. Can ["hide whitespace"](https://github.com/argoproj/argo-workflows/pull/12652/files?w=1) for the `hack/check-env-doc.sh` refactor to (mostly) ignore the indentation fixes

### Future Work

1. Still not quite all env vars are found by this script; in particular, dynamically named env vars are missing
    - such as those in [`common.go`](https://github.com/argoproj/argo-workflows/blob/7d70fe264fb0dbf42fbdf64cf94539408806a76d/workflow/common/common.go#L125), although there are others too
1. Would be good to distinguish the section of the docs as well, since there have been some env vars misplaced in the docs before (e.g. `GRPC_MESSAGE_SIZE` I moved in #12148) and some env vars that are used by multiple Argo components/binaries.
1. Remove some no longer needed env vars
    - would be a breaking change, but `BASE_HREF` and `ALLOWED_LINK_PROTOCOL` are no longer needed since all flags can be env vars since #6767 (`ARGO_BASE_HREF` and `ARGO_ALLOWED_LINK_PROTOCOL` can be used instead). might be some others.
      - **EDIT**: See #12653 as this breaking change


cmd/argo/commands/client/conn.go
```diff
@@ -45,7 +45,7 @@ func AddAPIClientFlagsToCmd(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&instanceID, "instanceid", os.Getenv("ARGO_INSTANCEID"), "submit with a specific controller's instance id label. Default to the ARGO_INSTANCEID environment variable.")
 	// "-s" like kubectl
 	cmd.PersistentFlags().StringVarP(&ArgoServerOpts.URL, "argo-server", "s", os.Getenv("ARGO_SERVER"), "API server `host:port`. e.g. localhost:2746. Defaults to the ARGO_SERVER environment variable.")
-	cmd.PersistentFlags().StringVar(&ArgoServerOpts.Path, "argo-base-href", os.Getenv("ARGO_BASE_HREF"), "An path to use with HTTP client (e.g. due to BASE_HREF). Defaults to the ARGO_BASE_HREF environment variable.")
+	cmd.PersistentFlags().StringVar(&ArgoServerOpts.Path, "argo-base-href", os.Getenv("ARGO_BASE_HREF"), "Path to use with HTTP client due to BASE_HREF. Defaults to the ARGO_BASE_HREF environment variable.")
```
isubasinghe, line 48: nit: I'd personally prefer if we kept PRs atomic. I think documentation correction like this can be fixed in it's own separate PR even if it's minor.

isubasinghe: It just makes it a bit harder to review for the issue in question itself, which in this case is checking for missing env vars.

isubasinghe (approved):
Approved although I have a minor nit

isubasinghe:
@agilgur5 we seem to just generally be doing a lot of programming as a part of the build process via the hack directory, wonder if we should maybe just use something like this: https://magefile.org/ and get rid of the hacks alltogether?

cmd/argo/commands/client/conn.go
agilgur5: Generally, I agree, which is why the vast majority of my PRs are small and split into independent chunks. (there are some that carry a different opinion though).

I changed this line because I had to create the docs for the `BASE_HREF` env var for this PR and so referenced these and found the mistakes. So it is related to the purpose, and this PR does already change/add docs for `BASE_HREF` specifically.

Unfortunately this one-liner causes a fairly large diff due to the codegen.

Also I'm not sure if "atomicity" is the right term here. As technically all commits and PRs are atomic. "Small, independent, singular purpose" are all terms I've used to describe this.

agilgur5:
> I guess my reasoning here is that I find the sed/awk/grep combos a bit hard to parse and understand. I'd rather write more easy to understand code instead.

Oh I agree 100%, and I say this having had to write a ton of bash at my last two jobs for fleet management... and also having written proposals for two of those teams to "use a programming language instead of hard-to-maintain shell hackery" (there's _a lot_ more issues with shell than it being hard to read, including directory awareness, arg parsing, (lack of) dependency management, portability, etc etc etc etc). In the PR description I did also literally write: "something something 'portable shell code' something something 'maybe use Python instead next time'"

> @agilgur5 we seem to just generally be doing a lot of programming as a part of the build process via the hack directory, wonder if we should maybe just use something like this: https://magefile.org/ and get rid of the hacks alltogether? [sic]

A few things:
1. This wouldn't get rid of the hacks, it might just move some of the code to a different place. We also already [have](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/hack/diagram/main.go) [several](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/hack/jsonschema/main.go) [hacks](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/hack/swagger/main.go) [written](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/hack/crds.go) [in](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/hack/docgen.go) Go. Magefile isn't necessary for that piece, it's more a `make` substitute (and I think there are better substitutes).
    - `hacks/` is also a pretty conventional Go directory, have seen it many times
    - The larger a codebase gets and the more it does, the more necessary build programming becomes (part of the reason why I've worked in DevOps for many years). This codebase in particular has a ton of codegen hacked together.
1. I wouldn't call Go the most readable or ergonomic scripting language either, would probably prefer Python
   - And Go has no real way of separating build deps from prod deps 😕 one of the reason why there are some build / compilation hacks and why it takes a bit to run. We do already have Python and JS build deps.

Issue #6923, opened 2021-10-13
[Regression] Argo v3.2.0 UI ignores BASE_HREF


rijos:
Argo has the ability to set the base href using the BASE_HREF environment variable in argo-server.
This value is ignored in Argo v3.2.0

Argo v3.2.0
![image](https://user-images.githubusercontent.com/23135242/137121471-3d9e5527-a08c-4f63-b64a-bb68a0446b7d.png)

![image](https://user-images.githubusercontent.com/23135242/137120280-f84f1ccb-3c37-4fac-bc21-ac0ce46c5467.png)

Argo v3.1.13
![image](https://user-images.githubusercontent.com/23135242/137120982-6382736b-b97c-4fc6-8b4e-d1595fd7a4fb.png)

Part of the kustomization config:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: argo-server
spec:
  selector:
    matchLabels:
      app: argo-server
  template:
    metadata:
      labels:
        app: argo-server
    spec:
      containers:
      - name: argo-server
        env:
          - name: BASE_HREF
            value: /argo/
```

agilgur5:
Hmm #6767 could have been listed as a breaking change instead -- `BASE_HREF` was replaced by `ARGO_BASE_HREF`. Now both can be used as well as the flag

PR #12653, opened 2024-02-11
chore!: remove duplicate Server env vars, plus `--basehref` -> `--base-href`

agilgur5:
Follow-up to #12652 and https://github.com/argoproj/argo-workflows/issues/6923#issuecomment-1937357156 on findings of duplicate env vars

### Motivation

- every Server CLI flag has had an equivalent env var since #6767
  - `ALLOWED_LINK_PROTOCOL` and `BASE_HREF` each had a duplicate env var as the odd ones out
- The Server flag `--basehref` was also inconsistent with the env var, other flags, and the client, change it to `--base-href` for consistency

### Modifications

- remove the two env vars in a breaking change, and refer users to their `ARGO_*` equivalents in the upgrade notes
- replace `--basehref` with `--base-href` as a breaking change and mention it in the upgrade notes
- modify the docs and descriptions that referred to the two env vars
  - plus some small grammar and consistency fixes therein

### Verification

Docs build & lint pass.

### Notes to Reviewers

Can skip `docs/cli/*` as it's all auto-generated from the CLI changes


cmd/argo/commands/server.go
```diff
@@ -191,18 +190,8 @@ See %s`, help.ArgoServer),
 		},
 	}
 
-	defaultBaseHRef := os.Getenv("BASE_HREF")
```
tczhao, line 194: Please correct me if I'm wrong.
I think we want to keep this right? but use `defaultBaseHRef := os.Getenv("ARGO_BASE_HREF")` instead?
otherwise setting environmental variable `ARGO_BASE_HREF` wouldn't have an effect

tczhao (changes requested):
I think we also need to change
https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/root.go#L87C102-L87C111
to ARGO_BASE_HREF

cmd/argo/commands/server.go
agilgur5: See the `viper` usage [below](https://github.com/argoproj/argo-workflows/blob/8571e1f1e8b78776e02a5d48013dd3c1f4ecd853/cmd/argo/commands/server.go#L215) that. That was added in #6767 as I referenced in the PR description. This is listed [in the docs](https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/docs/environment-variables.md?plain=1#L140) as well

agilgur5:
> I think we also need to change https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/root.go#L87C102-L87C111 to ARGO_BASE_HREF

See my [in-line comment](https://github.com/argoproj/argo-workflows/pull/12653#discussion_r1487240445), I mentioned this in the PR description.

tczhao:
Thanks for the links. They make a lot of sense!
I still think we need to change https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/root.go#L87C102-L87C111
from 
`If your server is behind an ingress with a path (you'll be running "argo server --basehref /...) or "BASE_HREF=/... argo server"):`
to
`If your server is behind an ingress with a path (you'll be running "argo server --basehref /...) or "ARGO_BASE_HREF=/... argo server"):`
or in any style that make sense, the `BASE_HREF=/... argo server` seems wrong since the env var now has a `ARGO_` preifx

agilgur5:
Ohhh, good catch! I totally missed that in the `root.go`, especially since it says `ARGO_BASE_HREF` on the very next line. Sorry I misread your first comment!

Yea let me modify that too. I could've sworn I searched for all references to `BASE_HREF` so let me double check that too if I missed this one

agilgur5:
> https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/root.go#L87C102-L87C111

Ah I totally missed this column number in the link -- I was on mobile where it's far off screen to the right so I didn't see the column highlight on first glance

agilgur5:
> https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/root.go#L87C102-L87C111

> > Yea let me modify that too. I could've sworn I searched for all references to `BASE_HREF` so let me double check that too if I missed this one

Fixed this one plus a few others. That should be all of them now

cmd/argo/commands/client/conn.go
```diff
@@ -45,7 +45,7 @@ func AddAPIClientFlagsToCmd(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&instanceID, "instanceid", os.Getenv("ARGO_INSTANCEID"), "submit with a specific controller's instance id label. Default to the ARGO_INSTANCEID environment variable.")
 	// "-s" like kubectl
 	cmd.PersistentFlags().StringVarP(&ArgoServerOpts.URL, "argo-server", "s", os.Getenv("ARGO_SERVER"), "API server `host:port`. e.g. localhost:2746. Defaults to the ARGO_SERVER environment variable.")
-	cmd.PersistentFlags().StringVar(&ArgoServerOpts.Path, "argo-base-href", os.Getenv("ARGO_BASE_HREF"), "Path to use with HTTP client due to BASE_HREF. Defaults to the ARGO_BASE_HREF environment variable.")
+	cmd.PersistentFlags().StringVar(&ArgoServerOpts.Path, "argo-base-href", os.Getenv("ARGO_BASE_HREF"), "Path to use with HTTP client due to Base HREF. Defaults to the ARGO_BASE_HREF environment variable.")
```
agilgur5, line 48: this is referring to the ["Base HREF" feature](https://argo-workflows.readthedocs.io/en/release-3.5/argo-server/#base-href) now and not the specific env var

tczhao:
> > https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/root.go#L87C102-L87C111
> 
> Ah I totally missed this column number in the link -- I was on mobile where it's far off screen to the right so I didn't see the column highlight on first glance
> 
> (Screenshot for reference: )

No worries. It happens

tczhao (approved):
LGTM

cmd/argo/commands/server.go
```diff
@@ -191,18 +190,8 @@ See %s`, help.ArgoServer),
...
-		defaultBaseHRef = "/"
-	}
-
-	defaultAllowedLinkProtocol := []string{"http", "https"}
-	if protocol := os.Getenv("ALLOWED_LINK_PROTOCOL"); protocol != "" {
-		defaultAllowedLinkProtocol = strings.Split(protocol, ",")
-	}
-
 	command.Flags().IntVarP(&port, "port", "p", 2746, "Port to listen on")
-	command.Flags().StringVar(&baseHRef, "basehref", defaultBaseHRef, "Value for base href in index.html. Used if the server is running behind reverse proxy under subpath different from /. Defaults to the environment variable BASE_HREF.")
```
sarabala1979, line 205: @agilgur5 This is a breaking change. `Need to be marked` as breaking change.
where does the code refer to `ARGO_BASE_HREF` from Viber to make it base href? 
Am I missing any change?

agilgur5: It is marked as a breaking change -- the PR has a `!`, the description says so, and the docs for breaking changes was updated in this PR.

Regarding Viper, that too is described in the PR and was described in my [previous comment](https://github.com/argoproj/argo-workflows/pull/12653#discussion_r1487240445) as well

sarabala1979: Thanks @agilgur5  sorry I missed the description and `!`. 


As per code `ARGO_BASE_HREF` will not considered during `argo server` starting. 
This code was used to get `BASE_HREF` from `env` and updated during server start. It could be a bug in existing code.  This [code](https://github.com/argoproj/argo-workflows/blob/986b069f34ed5336b6f754cfc6709c30b302c768/cmd/argo/commands/client/conn.go#L57) was never called during the server start.
```go
	defaultBaseHRef := os.Getenv("BASE_HREF")
	if defaultBaseHRef == "" {
		defaultBaseHRef = "/"
	}
```
I tried in my local base href didn't update. Please try on your env set `ARGO_BASE_HREF` and start the `argo-server` to check `baseHRef` updating
```log
(base) ➜  argo-workflows git:(main) ✗ export ARGO_BASE_HREF="/test"
(base) ➜  argo-workflows git:(main) ✗ dist/argo server             
INFO[2024-02-18T20:51:52.092Z] not enabling pprof debug endpoints           
INFO[2024-02-18T20:51:52.096Z]                                               authModes="[client]" baseHRef=/ managedNamespace= namespace=argo secure=true ssoNamespace=argo
INFO[2024-02-18T20:51:52.096Z] Generating Self Signed TLS Certificates for Secure Mode 
INFO[2024-02-18T20:51:52.096Z] SSO disabled                                 
INFO[2024-02-18T20:51:52.109Z] Starting Argo Server                          instanceID= version=untagged
INFO[2024-02-18T20:51:52.109Z] Creating event controller                     asyncDispatch=false operationQueueSize=16 workerCount=4
INFO[2024-02-18T20:51:52.111Z] GRPC Server Max Message Size, MaxGRPCMessageSize, is set  GRPC_MESSAGE_SIZE=104857600
INFO[2024-02-18T20:51:52.111Z] Argo Server started successfully on https://localhost:2746  url="https://localhost:2746"
```

agilgur5: > This code was used to get `BASE_HREF` from `env` and updated during server start.

Yes, I'm aware, I intentionally removed it since Viper does that automatically.

> This [code](https://github.com/argoproj/argo-workflows/blob/986b069f34ed5336b6f754cfc6709c30b302c768/cmd/argo/commands/client/conn.go#L57) was never called during the server start.

That's client code, not Server code. That's used for the CLI commands that connect to the Server

> I tried in my local base href didn't update. Please try on your env set `ARGO_BASE_HREF` and start the `argo-server` to check `baseHRef` updating

👀  let me double-check that. I'm pretty sure I've used this a few times before

> It could be a bug in existing code.

Maybe, I'll check

agilgur5: Ok apparently, there's a bit of a typo in the old code... it's [`--basehref`](https://github.com/argoproj/argo-workflows/blob/986b069f34ed5336b6f754cfc6709c30b302c768/cmd/argo/commands/server.go#L205) in the Server, not `--base-href`, so the env var is `ARGO_BASEHREF`.

This is pretty confusing given that the client uses [`--argo-base-href`](https://github.com/argoproj/argo-workflows/blob/986b069f34ed5336b6f754cfc6709c30b302c768/cmd/argo/commands/client/conn.go#L48).

I would like to make them both use dash-case/kebab-case, which would convert to `ARGO_BASE_HREF`, but that would be another breaking change 😕

agilgur5: Ok I changed it to `--base-href` for consistency and added another breaking change note

sarabala1979 (commented):
LGTM
Before approve, @terrytangyuan @juliev0 can you take a quick look because of it is breaking change?

rijos:
@agilgur5, seems this change will break the helm chart, it still refers to BASE_HREF. See: https://github.com/argoproj/argo-helm/blob/2f82fb5992fe1e390d1ebdbc4be6d5d6c6549a37/charts/argo-workflows/templates/server/server-deployment.yaml#L98

agilgur5:
Yes, all downstream consumers that use this, like the Helm Charts, will need to be updated. That is the nature of breaking changes.

agilgur5:
Merged `main` for new checks per #13027

agilgur5:
Going to merge this as Tianchu approved and Bala LGTM'd and there have been no reviews from others in the past ~4 months

agilgur5:
> @agilgur5, seems this change will break the helm chart, it still refers to BASE_HREF. See: https://github.com/argoproj/argo-helm/blob/2f82fb5992fe1e390d1ebdbc4be6d5d6c6549a37/charts/argo-workflows/templates/server/server-deployment.yaml#L98

Since `ARGO_BASE_HREF` already exists, I addressed this in the chart for forward compat: https://github.com/argoproj/argo-helm/pull/2756

agilgur5:
> Since `ARGO_BASE_HREF` already exists, I addressed this in the chart for forward compat: [argoproj/argo-helm#2756](https://github.com/argoproj/argo-helm/pull/2756)

Annnd I didn't re-read my own release notes or this PR (it's 4 months old, so quite a time delay) and caused a regression downstream in Argo Helm. Reverted that change in https://github.com/argoproj/argo-helm/pull/2770
Sorry I screwed that up, I meant to be proactive there, but the time delay meant I forgot how my own change worked

PR #12654, opened 2024-02-11
fix(docs): remove `workflow-controller-configmap.yaml` self reference

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/12446#discussion_r1482271109

### Motivation

- this was erroneously added to the `main` branch in https://github.com/argoproj/argo-workflows/pull/12446#discussion_r1482271109
  - it should only be on the `gh-pages` branch, for which all HTML pages redirect to ReadTheDocs now, but this YAML could not be, so we added this comment (see https://github.com/argoproj/argo-workflows/pull/12362#discussion_r1433271044)

- on the `main` branch, this is a confusing as it is a self-reference
  - https://argo-workflows.readthedocs.io/en/latest/workflow-controller-configmap.yaml currently points to itself in circular fashion

### Modifications

- remove self-reference comment at the top of `docs/workflow-controller-configmap.yaml` on `main`

### Verification

n/a, as it's rendered raw in the docs

I double-checked that this was not present in the [`release-3.4`](https://github.com/argoproj/argo-workflows/blob/release-3.4/docs/workflow-controller-configmap.yaml#L1) and [`release-3.5`](https://github.com/argoproj/argo-workflows/blob/release-3.5/docs/workflow-controller-configmap.yaml#L1) docs, so no back-porting needed

### DIFF
--- a/docs/workflow-controller-configmap.yaml
+++ b/docs/workflow-controller-configmap.yaml
@@ -1,6 +1,3 @@
-#######
-# NEW URL FOR THIS FILE: https://argo-workflows.readthedocs.io/en/latest/workflow-controller-configmap.yaml
-#######
 # This file describes the config settings available in the workflow controller configmap
 apiVersion: v1
 kind: ConfigMap


jmeridth (approved):

isubasinghe (approved):
LGTM

PR #12666, opened 2024-02-14
fix(controller): add missing namespace index from workflow informer

ryancurrah:
Due to the namespace index missing from the workflow informer the healthz endpoint could not list namespace indexes. 

Leading to the following warning message: 
```
listers.go:79] can not retrieve list of objects using index : Index with name namespace does not exist
```
Everytime the healthz endpoint is called.

This fix has been verified in our environment by building the `workflow-controller` image from scratch and deploying it. 

Related Slack thread: https://cloud-native.slack.com/archives/C01QW9QSSSK/p1707918209805259

### Fixes

Fixes healthz endpoint not being able to list namespace indexes.

### Motivation

These changes are similar to what others have done to resolve the issue.

- https://github.com/prometheus-operator/prometheus-operator/pull/1697
- https://github.com/argoproj/argo-cd/pull/2761

### Modifications

Added the namespace index to the workflow informer.

### Verification

Deployed to our environment hit the healthz endpoint with `curl http://127.0.0.1:6060/healthz` and checked for the log message. It no longer appears.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -265,6 +265,7 @@ var indexers = cache.Indexers{
 	indexes.WorkflowPhaseIndex:           indexes.MetaWorkflowPhaseIndexFunc(),
 	indexes.ConditionsIndex:              indexes.ConditionsIndexFunc,
 	indexes.UIDIndex:                     indexes.MetaUIDFunc,
+	cache.NamespaceIndex:                 cache.MetaNamespaceIndexFunc,
 }
 
 // Run starts an Workflow resource controller


Joibel (approved):
LGTM.

This is a common pattern and appears to do the right thing.

agilgur5 (approved):
Makes sense to me and tests continue to pass, so does not seem to cause any regressions either, and should improve performance.

agilgur5:
Thanks for root causing this and fixing it!

ryancurrah:
No problem @agilgur5!

PR #12669, opened 2024-02-15
docs: Add PayIt to USERS.md

menzenski:
### Motivation

We've been running Argo Workflows for about a year now, and we'd like to make it official 😄


agilgur5 (approved):
Thanks for listing your org 🙂

PR #12678, opened 2024-02-18
docs: Add notes on issue assignment in CONTRIBUTING.md

terrytangyuan:
There have been many cases where new contributors asked to get assigned but never worked on it and these issues were kept getting re-assigned. We should consider only assign issues to regular contributors and maintainers going forward. For new contributors, they can just work on a PR directly without being assigned.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -123,7 +123,7 @@ Bugs can be [sorted by "👍"](https://github.com/argoproj/argo-workflows/issues
 If the issue is determined to be a user error and not a bug, remove the `bug` label (and the `regression` label, if applicable) and replace it with the `support` label.
 If more information is needed from the author to diagnose the issue, then apply the `more information needed` label.
 
-Please only assign issues to regular contributors and maintainers.
+Please only assign issues to regular contributors and maintainers. New contributors are encouraged to work on a PR directly without being assigned.
 
 ##### Staleness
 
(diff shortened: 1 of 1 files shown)


sarabala1979 (approved):
LGTM

docs/CONTRIBUTING.md
```diff
@@ -123,6 +123,8 @@ Bugs can be [sorted by "👍"](https://github.com/argoproj/argo-workflows/issues
 If the issue is determined to be a user error and not a bug, remove the `bug` label (and the `regression` label, if applicable) and replace it with the `support` label.
 If more information is needed from the author to diagnose the issue, then apply the `more information needed` label.
 
+Please only assign issues to regular contributors and maintainers. New contributors are encouraged to work on a PR directly without being assigned.
```
agilgur5, line 126: So there is already a line about assignment under the "Authoring PRs" section [above](https://github.com/argoproj/argo-workflows/blob/bfed12b68f5747b67f6c154fd3ef26f43149c838/docs/CONTRIBUTING.md?plain=1#L52). You actually added that yourself in #11805 😅 

I'm not sure if we need more on this, but if we do want more clarifications, I'm not sure it would go under "Triaging bugs", since assignment is not part of the current bug triage process.

Additionally, the word "maintainer" is not defined here and is perhaps too selective -- I would use "Member", which is defined and broader.

terrytangyuan: "Authoring PRs" is more for new contributors. 

This is more for maintainers to triage issues and assign folks accordingly.

Issue #12659, opened 2024-02-12
controller sending many pod delete requests that result in 404 response


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

although this seems to have no effect on the functioning of argoworkflows it could be potential stability/performance/k8s log cost issue at scale. it must be also filling up the cleanup queue so delaying cleanup of pods that really do exist

from checking the k8s api server the wfcontroller seems to be sending delete pod request for ```<podname from a step>-agent``` and getting not found response. i am using standard workflows like whalesay example. not sure what significance of agent suffix is (i did see https://github.com/argoproj/argo-workflows/blob/66680f1c9bca8b47c40ce918b5d16714058647cb/workflow/controller/agent.go#L25)

seeing 1000s of these, seems for every pod run its sending this unrequired delete request for pod with -agent suffix?

### Version

3.4.11

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
n/a
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
n/a
```

agilgur5:
For reference, this was previously noted [in Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1706902459768879?thread_ts=1706902459.768879&cid=C01QW9QSSSK)

agilgur5:
> not sure what significance of agent suffix is

As far as I understand, the "agent" is a piece of the Executor that runs for certain built-in, non-`container` template types (e.g. `resource`, `http`, `data`, etc). Anyone else please correct me if I'm wrong; the historical references to agent didn't quite have a standard definition.


> seeing 1000s of these, seems for every pod run its sending this unrequired delete request for pod with -agent suffix?

That sounds like it might be accidentally assuming that each Pod has an agent, when there are only certain types that do 🤔

tooptoop4:
[...]

PR #12679, opened 2024-02-18
fix: only cleanup agent pod if exists

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/12659

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2366,7 +2366,6 @@ func (woc *wfOperationCtx) markWorkflowPhase(ctx context.Context, phase wfv1.Wor
 			}
 			woc.updated = true
 		}
-		hasPluginNodes
 		if woc.hasPluginNodes() {
 			woc.controller.queuePodForCleanup(woc.wf.Namespace, woc.getAgentPodName(), deletePod)
 		}
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2366,7 +2366,7 @@ func (woc *wfOperationCtx) markWorkflowPhase(ctx context.Context, phase wfv1.Wor
 			}
 			woc.updated = true
 		}
-		if woc.hasPluginNodes() {
+		if woc.hasNodeWithAgentPod() {
 			woc.controller.queuePodForCleanup(woc.wf.Namespace, woc.getAgentPodName(), deletePod)
 		}
 	}
@@ -2388,9 +2388,9 @@ func (woc *wfOperationCtx) estimateNodeDuration(nodeName string) wfv1.EstimatedD
 	return woc.getEstimator().EstimateNodeDuration(nodeName)
 }
 
-func (woc *wfOperationCtx) hasPluginNodes() bool {
+func (woc *wfOperationCtx) hasNodeWithAgentPod() bool {
 	for _, node := range woc.wf.Status.Nodes {
-		if node.Type == wfv1.NodeTypePlugin {
+		if node.Type == wfv1.NodeTypePlugin || node.Type == wfv1.NodeTypeHTTP {
 			return true
 		}
 	}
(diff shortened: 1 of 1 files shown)


workflow/controller/operator.go
```diff
@@ -2366,7 +2366,9 @@ func (woc *wfOperationCtx) markWorkflowPhase(ctx context.Context, phase wfv1.Wor
 			}
 			woc.updated = true
 		}
-		woc.controller.queuePodForCleanup(woc.wf.Namespace, woc.getAgentPodName(), deletePod)
+		if woc.hasPluginNodes() {
```
agilgur5, line 2369: The agent isn't just for plugin nodes, it's also for other built-in template types like `http` etc as I mentioned in https://github.com/argoproj/argo-workflows/issues/12659#issuecomment-1939189088

tooptoop4: @alexec can u confirm what node types have agent pod? i saw https://github.com/argoproj/argo-workflows/pull/7256 introduced this delete agent pod line and that was the same PR that introduced plugin node

workflow/controller/operator.go
```diff
@@ -2386,6 +2388,15 @@ func (woc *wfOperationCtx) estimateNodeDuration(nodeName string) wfv1.EstimatedD
 	return woc.getEstimator().EstimateNodeDuration(nodeName)
 }
 
+func (woc *wfOperationCtx) hasNodeWithAgentPod() bool {
+	for _, node := range woc.wf.Status.Nodes {
+		if node.Type == wfv1.NodeTypePlugin || node.Type == wfv1.NodeTypeHTTP {
```
agilgur5, line 2393: I don't believe it's just `plugin` and `http` either, per https://github.com/argoproj/argo-workflows/issues/12659#issuecomment-1939189088.

`controller/agent.go` actually does not seem to define the the template types exclusively, so maybe we should instead check if the Pod contains the agent -- the `isAgentPod` function already exists there.

jswxstw: Each workflow involving http/plugin template has an agent pod, see [#12659 comment].(https://github.com/argoproj/argo-workflows/issues/12659#issuecomment-1956125777)
The `isAgentPod` function determine whether a pod is the agent type based on its name, so it can not be used here.

jswxstw:
The agent pod deletion logic is currently performed in `markWorkflowPhase` and the PodGC strategy is invalid to the agent pod, so I think it is more appropriate to put it into `queuePodsForCleanup`.
https://github.com/argoproj/argo-workflows/blob/5c8062e55d975aab117e37c7592c0a648a9e9860/workflow/controller/operator.go#L2291
https://github.com/argoproj/argo-workflows/blob/5c8062e55d975aab117e37c7592c0a648a9e9860/workflow/controller/pod_cleanup.go#L13

jswxstw:
> The agent pod deletion logic is currently performed in `markWorkflowPhase` and the PodGC strategy is invalid to the agent pod, so I think it is more appropriate to put it into `queuePodsForCleanup`.
> 
> https://github.com/argoproj/argo-workflows/blob/5c8062e55d975aab117e37c7592c0a648a9e9860/workflow/controller/operator.go#L2291
> 
> 
> https://github.com/argoproj/argo-workflows/blob/5c8062e55d975aab117e37c7592c0a648a9e9860/workflow/controller/pod_cleanup.go#L13

One thing to note is that when the workflow is completed, the agent pod is acturally still running, so it is necessary to stop the agent pod first if it is not going to be deleted temporarily.

workflow/controller/operator.go
jswxstw: This function can be used to determine whether the creation of `AgentPod` is necessary.
https://github.com/argoproj/argo-workflows/blob/7ba20fea08e40a5792a1d25011ee8c12889960c5/workflow/controller/taskset.go#L59-L61

agilgur5:
Superseded by #13016

workflow/controller/operator.go
agilgur5: The above done in #13294

workflow/controller/operator.go
agilgur5: Follow-up thread [below](https://github.com/argoproj/argo-workflows/pull/12679#discussion_r1495400802). Eventually resolved in #13294

Issue #12391, opened 2023-12-20
quick start fails out of the box


cardoe:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

Able to run the quick start CLI with success.

My environment was a macOS M2 machine running Docker Desktop with kind. The steps are straight from https://argoproj.github.io/argo-workflows/quick-start/

```bash
kind create cluster
# cluster created with 1.27.3
kubectl create namespace argo
kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.5.2/install.yaml
kubectl patch deployment \
  argo-server \
  --namespace argo \
  --type='json' \
  -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/args", "value": [
  "server",
  "--auth-mode=server"
]}]'
```

At this point I ran:

```bash
argo submit -n argo --watch https://raw.githubusercontent.com/argoproj/argo-workflows/main/examples/hello-world.yaml
```

And the output was:

```
Name:                hello-world-g2csp
Namespace:           argo
ServiceAccount:      unset (will run with the default ServiceAccount)
Status:              Error
Message:             Error (exit code 1): pods "hello-world-g2csp" is forbidden: User "system:serviceaccount:argo:default" cannot patch resource "pods" in API group "" in the namespace "argo"
Conditions:
 PodRunning          False
 Completed           True
Created:             Wed Dec 20 10:25:35 -0600 (9 minutes ago)
Started:             Wed Dec 20 10:25:35 -0600 (9 minutes ago)
Finished:            Wed Dec 20 10:26:05 -0600 (9 minutes ago)
Duration:            30 seconds
Progress:            0/1
ResourcesDuration:   13s*(1 cpu),13s*(100Mi memory)

STEP                  TEMPLATE  PODNAME            DURATION  MESSAGE
 ⚠ hello-world-g2csp  whalesay  hello-world-g2csp  21s       Error (exit code 1): pods "hello-world-g2csp" is forbidden: User "system:serviceaccount:argo:default" cannot patch resource "pods" in API group "" in the namespace "argo"
```

### Version

v3.5.2

[...]

PR #12680, opened 2024-02-18
docs: link to RBAC from installation and add notes

rnathuji:
These doc improvements are geared towards guiding users to RBAC details during installation and adding useful notes therein.

### Motivation

These doc changes were motivated by conversations in #12391 

### Verification

The added notes to the `Workflow RBAC` page were confirmed via the following where `rbac.yaml` included definitions for the `Role`, `ServiceAccount`, and `RoleBinding` in the docs:

```bash
$ minikube start
$ kubectl create namespace argo
$ kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.5.4/install.yaml
$ kubectl apply -f rbac.yaml
$ argo submit --serviceaccount hello-world --watch https://raw.githubusercontent.com/argoproj/argo-workflows/main/examples/hello-world.yaml
```

### DIFF
--- a/docs/workflow-rbac.md
+++ b/docs/workflow-rbac.md
@@ -48,3 +48,28 @@ Warning: For many organizations, it may not be acceptable to give a workflow the
 
 If you are not using the emissary, you'll need additional permissions.
 See [executor](https://github.com/argoproj/argo-workflows/tree/main/manifests/quick-start/base/executor) for suitable permissions.
+
+**Note**: Prior to `v3.5.2`, simple workflows such as [hello-world](https://raw.githubusercontent.com/argoproj/argo-workflows/main/examples/hello-world.yaml) may have run successfully without RBAC. As of `v3.5.2`, these workflows will fail without at least the permissions above via associated [Service Accounts](service-accounts.md). For example:
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: hello-world
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: hello-world-executor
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):

terrytangyuan:
Thanks!

docs/workflow-rbac.md
```diff
@@ -48,3 +48,28 @@ Warning: For many organizations, it may not be acceptable to give a workflow the
...
+  kind: Role
+  name: executor
+subjects:
+- kind: ServiceAccount
+  name: hello-world
+```
+
+```bash
+argo submit --serviceaccount hello-world --watch https://raw.githubusercontent.com/argoproj/argo-workflows/main/examples/hello-world.yaml
+```
```
agilgur5, line 75: I don't think this is necessary. The docs have never said otherwise that Workflow RBAC was not required, that was an incidental technicality that simple steps could work without it.
It's something we can mention in upgrade notes, not in this page. Also, realistically it only impacts a small percentage of production users.

Additionally
1. SAs are already mentioned at the top of this page
1. The example can be useful to demonstrate combining everything in this page, but for a footnote, this makes it seem like required reading for everyone, even though it likely will not impact them. A footnote shouldn't need an example.
1. Versioning is specified with `> v3.5.2 and after` per #12581 
1. Should be simpler and more direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

rnathuji: Addressed in this [follow-up PR](https://github.com/argoproj/argo-workflows/pull/12686)

Issue #12645, opened 2024-02-09
Server OOM when too many executions of CronWorkflows or WorkflowTemplates


panicboat:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

This occurs when trying to view the details of each cron-workflows or workflow-template UI, which has a lot of history.
There was nothing like that in the argo-server logs.
![image](https://github.com/argoproj/argo-workflows/assets/4182311/2f023012-db94-4821-98e0-c8cfbbf5c50a)

### Version

v3.5.4

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
Any workflow is acceptable.
```

### Logs from the workflow controller
workflow-controller does not seem to be involved.

```text
Found 2 pods, using pod/workflow-controller-7bbd96c558-qss82
time="2024-02-09T09:40:00.290Z" level=info msg="Processing workflow" Phase= ResourceVersion=646930323 namespace=develop-workflow workflow=**********-1707471600
time="2024-02-09T09:40:00.691Z" level=info msg="Task-result reconciliation" namespace=develop-workflow numObjs=0 workflow=**********-1707471600
time="2024-02-09T09:40:00.790Z" level=info msg="Updated phase  -> Running" namespace=develop-workflow workflow=**********-1707471600
time="2024-02-09T09:40:01.690Z" level=info msg="Created PDB resource for workflow." namespace=develop-workflow workflow=**********-1707471600
time="2024-02-09T09:40:01.690Z" level=warning msg="Node was nil, will be initialized as type Skipped" namespace=develop-workflow workflow=**********-1707471600
time="2024-02-09T09:40:01.691Z" level=info msg="was unable to obtain node for , letting display name to be nodeName" namespace=develop-workflow workflow=**********-1707471600
time="2024-02-09T09:40:02.290Z" level=info msg="Pod node **********-1707471600 initialized Pending" namespace=develop-workflow workflow=**********-1707471600
time="2024-02-09T09:40:02.392Z" level=info msg="Created pod: **********-1707471600 (**********-1707471600)" namespace=develop-workflow workflow=**********-1707471600
[...]

PR #12681, opened 2024-02-19
fix: Add limit to number of Workflows in CronWorkflow history

panicboat:
Fixes https://github.com/argoproj/argo-workflows/issues/12645

### Motivation

I want to avoid argo-server going OOM when there is a lot of execution history.

### Modifications

Limited the number of histories to 50.
image
image

### Verification

- [x] No completed cron workflows
- [x] Cron workflow with the number of cases exceeding the limit
- [x] Cron workflow under limit

### DIFF
--- a/ui/src/app/cron-workflows/cron-workflow-details.tsx
+++ b/ui/src/app/cron-workflows/cron-workflow-details.tsx
@@ -9,6 +9,8 @@ import {uiUrl} from '../shared/base';
 import {ErrorNotice} from '../shared/components/error-notice';
 import {openLinkWithKey} from '../shared/components/links';
 import {Loading} from '../shared/components/loading';
+import {Pagination, parseLimit} from '../shared/pagination';
+import {ScopedLocalStorage} from '../shared/scoped-local-storage';
 import {useCollectEvent} from '../shared/use-collect-event';
 import {ZeroState} from '../shared/components/zero-state';
 import {Context} from '../shared/context';
@@ -22,12 +24,21 @@ import {CronWorkflowEditor} from './cron-workflow-editor';
 import '../workflows/components/workflow-details/workflow-details.scss';
 import './cron-workflow-details.scss';
 
+const storage = new ScopedLocalStorage('ListOptions');
+
 export function CronWorkflowDetails({match, location, history}: RouteComponentProps<any>) {
     // boiler-plate
     const {navigation, notifications, popup} = useContext(Context);
     const queryParams = new URLSearchParams(location.search);
 
     const [namespace] = useState(match.params.namespace);
+    const [pagination] = useState<Pagination>(() => {
+        const savedPaginationLimit = storage.getItem('options', {}).paginationLimit || undefined;
+        return {
+            offset: queryParams.get('offset') || undefined,
+            limit: parseLimit(queryParams.get('limit')) || savedPaginationLimit || 50
+        };
+    });
     const [name] = useState(match.params.name);
     const [sidePanel, setSidePanel] = useState(queryParams.get('sidePanel'));
     const [tab, setTab] = useState(queryParams.get('tab'));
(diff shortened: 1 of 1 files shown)


ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -21,12 +23,21 @@ import {CronWorkflowEditor} from '../cron-workflow-editor';
...
     const queryParams = new URLSearchParams(location.search);
 
     const [namespace] = useState(match.params.namespace);
+    const [pagination] = useState<Pagination>(() => {
+        const savedPaginationLimit = storage.getItem('options', {}).paginationLimit || undefined;
+        return {
+            offset: queryParams.get('offset') || undefined,
+            limit: parseLimit(queryParams.get('limit')) || savedPaginationLimit || 50
+        };
+    });
```
agilgur5, line 40: I see that you copied this from [the code](https://github.com/argoproj/argo-workflows/blob/986b069f34ed5336b6f754cfc6709c30b302c768/ui/src/app/workflows/components/workflows-list/workflows-list.tsx#L56) in the Workflows List, however this does not make sense here.
This is currently creating `state` that is not used -- meaning it isn't stateful to begin with.

I notice that `namespace` and `name` in this function aren't stateful either, those should be converted to plain `const`s as well.

More specifically, this code allows the initial pagination to be set via `localStorage` and query params, but there is no way to change it after. In the Workflows List page, there is the `PaginationPanel` that can change those.

We also don't want it to be changeable -- it should be a hard-coded `50`

```suggestion
```

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -21,12 +23,21 @@ import {CronWorkflowEditor} from '../cron-workflow-editor';
 import '../../../workflows/components/workflow-details/workflow-details.scss';
 import './cron-workflow-details.scss';
 
+const storage = new ScopedLocalStorage('ListOptions');
+
```
agilgur5, line 27:
```suggestion
```

same as above, this is not needed

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -8,6 +8,8 @@ import {uiUrl} from '../../../shared/base';
 import {ErrorNotice} from '../../../shared/components/error-notice';
 import {openLinkWithKey} from '../../../shared/components/links';
 import {Loading} from '../../../shared/components/loading';
+import {Pagination, parseLimit} from '../../../shared/pagination';
+import {ScopedLocalStorage} from '../../../shared/scoped-local-storage';
```
agilgur5, line 12:
```suggestion
```

same as above, not needed

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
```diff
@@ -71,7 +82,7 @@ export function CronWorkflowDetails({match, location, history}: RouteComponentPr
 
     useEffect(() => {
         (async () => {
-            const workflowList = await services.workflows.list(namespace, null, [`${models.labels.cronWorkflow}=${name}`], null);
+            const workflowList = await services.workflows.list(namespace, null, [`${models.labels.cronWorkflow}=${name}`], pagination);
```
agilgur5, line 85:
```suggestion
            const workflowList = await services.workflows.list(namespace, null, [`${models.labels.cronWorkflow}=${name}`], {limit: 50});
```

this should just be hard-coded here

agilgur5 (changes requested):
Some unnecessary and incorrect code was added here with regard to `state`, `localStorage`, and query params. See in-line comments below:

ui/src/app/cron-workflows/components/cron-workflow-details/cron-workflow-details.tsx
panicboat: I see, thanks.
Now I understand why you did not add the pagination feature to this page.
I may issue an additional PR in the future when I get a chance, but for now, please let this be the response.

agilgur5: A PR to add a link at the bottom would be good!

agilgur5 (approved):
Confirmed this is working as intended. Thanks for the fix!

agilgur5:
Fixed merge conflicts with the added then removed imports.

Issue #6945, opened 2021-10-15
UI: User interface plugins


alexec:
Currently, to add a new resources to the UI you need to implement both backend and frontend changes:

* Add new proto.
* Write some Golang code.
* Implement the UI changes.

Unlike workflows APIs, the events and dataflow APIs are generic, we could replace them with a "dynamic" API that uses the `dynamic.NewClient`. That would open up a bunch of APIs that could be used to support any resource we wanted.

This dynamic API would make the events and dataflow APIs redundant.

Argo CD has a UI extension mechanism to load the rollouts into the UI that is proven. We can lift inspiration from this for a similar feature in workflows.

Et voila - we have a user interface with extensibility.

Other needs:

* Add new routes and left-hand nav items to the UI.
* Embed UI elements into other pages to add use case specific features.
* A way to easily package and publish UI extensions.

Workflow APIs are too complex to generalise, e.g. there is as `/suspend` API with logic behind it. So we need a way for users to plugin logic too (TBD).

DhairyaBahl:
Hello @alexec ! I am willing to work on this issue but I am not familiar with cncf projects. I am facing a small issue that resources are more than enough and I am feeling a bit lost in all these repos and all. Kindly help me out a bit. I would also love to start contributing on this.

If this issue is available then kindly assign this to me.

alexec:
Hi @DhairyaBahl this issue is currently being held for Google Summer of Code. It’s a really big issue and wide ranging issue, with potential for massive impact, and would require some time and dedication. We do have a lot of issues marked “good first issue” which are great places to start contributing (if you have not done so before). Users vote for issues with 👍 so you can sort issues by 👍 to find popular issues.

momom-i:
Hi, is this issue opened and could anyone work on it now?

agilgur5:
I had been thinking of a relatively simple (and not necessarily the most efficient or secure method) way to do UI plugins in https://github.com/argoproj/argo-workflows/pull/12682#discussion_r1537072401 and https://github.com/argoproj/argo-workflows/issues/12943#issuecomment-2065649664.

If we have a "Settings" page (or section on the UserInfo page or something), we could potentially just allow users to add arbitrary JS links there. Like all settings, they'd be stored in `localStorage` and so persisted across sessions.
That way we could load a completely arbitrary JS script on page load and it would be entirely user preference what they add.

[...]

Issue #6943, opened 2021-10-15
Pluggability


alexec:
# Summary

Issues:

* [x] #4355
* [x] #5201
* [x] #7154
* [ ] #7177
* [ ] #7710
* [ ] #6945
* [x] #5862
* [ ] #6867
* [x] #7073
* [x] #6313
* [ ] #4162 
* [ ] #4163
* [ ] #7743

Related:

* [ ] #6615 
* [ ] #4158

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

agilgur5:
WASM / WASI could be a huge boon for plugins as it allows for code from almost any language to be executed in a secure, sandboxed way.

There is already [the WASM Executor plugin](https://github.com/Shark/wasm-workflows-plugin), but we could potentially have built-in support for it and utilize WASM in several different places, including in the Controller and Server themselves, not just in ephemeral executors.

We could detect a WASM binaries on the filesystem and load them for different plugins, so a user could then have a custom image or volume mount with their binaries.

Implementation-wise, there do exist some great tools for this already, [Exitsm](https://github.com/extism/) in particular is an SDK for WASM-powered plugin systems and has an existing [Go SDK](https://github.com/extism/go-sdk).

PR #12682, opened 2024-02-20
feat: add i18n for pages which include button, label, and title

### LINKED ISSUE #10103: UI: i18n Proposal
# Summary

It would be good to have i18n support on the UI page.

# Use Cases

As a Chinese developer, I'd like to contribute to the translation of Chinese. Just in case anyone might want to know more about me. I'm a maintainer of the [Jenkins Chinese plugin](https://github.com/jenkinsci/localization-zh-cn-plugin).

---

LinuxSuRen:
Fixes #10103

### Motivation

Make the UI be easier for non-English speakers. I didn't finish all the translation work, please let me know if you have any ideas or suggestions.

### Modifications

image

### Verification

Manually verify.


ui/src/app/index.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import {App} from './app';
+import './i18n';
```
agilgur5, line 4:
```suggestion

import './i18n';
```

separate out the side-effect

ui/src/app/login/components/login.tsx
```diff
@@ -5,6 +5,7 @@ import {uiUrl, uiUrlWithParams} from '../../shared/base';
 import {useCollectEvent} from '../../shared/components/use-collect-event';
 
 import './login.scss';
+import {useTranslation} from 'react-i18next';
```
agilgur5: Please follow the existing conventions within these files -- this import should be at the top of this file, where all the external imports are

ui/src/app/login/login.tsx
```diff
@@ -25,8 +26,9 @@ function getRedirect(): string {
 
 export function Login() {
     useCollectEvent('openedLogin');
+    const { t } = useTranslation();
```
agilgur5:
```suggestion
    const { t } = useTranslation();
    
    useCollectEvent('openedLogin');

```


This would follow the conventions we have elsewhere -- variable declarations at the top, new line, effect, new line, render.

ui/src/app/workflow-templates/workflow-template-list.tsx
```diff
@@ -26,6 +26,7 @@ import {WorkflowTemplateCreator} from './workflow-template-creator';
 import {WorkflowTemplateFilters} from './workflow-template-filters';
 
 import './workflow-template-list.scss';
+import {useTranslation} from 'react-i18next';
```
agilgur5, line 29: same as above, move this to the top of the file with the other external imports

ui/src/app/plugins/plugin-list.tsx
```diff
@@ -8,10 +8,12 @@ import {useCollectEvent} from '../shared/use-collect-event';
 import {ZeroState} from '../shared/components/zero-state';
 import {historyUrl} from '../shared/history';
 import {Utils} from '../shared/utils';
+import {useTranslation} from 'react-i18next';
```
agilgur5, line 11: same as above, move this to the top of the file with the other external imports

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -62,16 +63,17 @@ export function SubmitWorkflowPanel(props: Props) {
         }
     }
 
+    const { t } = useTranslation();
```
agilgur5: this should be above the function declarations, per convention

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -43,6 +43,7 @@ import {SuspendInputs} from './suspend-inputs';
 import {WorkflowResourcePanel} from './workflow-resource-panel';
 
 import './workflow-details.scss';
+import {useTranslation} from 'react-i18next';
```
agilgur5, line 46: same as above, this should be at the top of the file with the other external imports

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -182,6 +183,7 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
     }, [nodeId, workflow]);
 
     const parsedSidePanel = parseSidePanelParam(sidePanel);
+    const { t } = useTranslation();
```
agilgur5: this should similarly be with the other variable declarations above

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
```diff
@@ -12,6 +12,7 @@ import {TagsInput} from '../../../shared/components/tags-input/tags-input';
 import {services} from '../../../shared/services';
 
 import './workflow-filters.scss';
+import {useTranslation} from 'react-i18next';
```
agilgur5, line 15: same as above

ui/src/app/workflows/components/workflow-summary-panel.tsx
```diff
@@ -13,6 +13,7 @@ import {ResourcesDuration} from '../../shared/resources-duration';
 import {WorkflowCreatorInfo} from './workflow-creator-info/workflow-creator-info';
 import {WorkflowFrom} from './workflow-from';
 import {WorkflowLabels} from './workflow-labels/workflow-labels';
+import {useTranslation} from "react-i18next";
```
agilgur5, line 16: same as above

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -26,6 +26,7 @@ import {WorkflowFilters} from '../workflow-filters/workflow-filters';
 import {WorkflowsRow} from '../workflows-row/workflows-row';
 import {WorkflowsSummaryContainer} from '../workflows-summary-container/workflows-summary-container';
 import {WorkflowsToolbar} from '../workflows-toolbar/workflows-toolbar';
+import {useTranslation} from 'react-i18next';
```
agilgur5, line 29: same as above

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -158,6 +159,7 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
     }, [namespace, phases.toString(), labels.toString(), pagination.limit, pagination.offset]); // referential equality, so use values, not refs
 
     useCollectEvent('openedWorkflowList');
+    const { t } = useTranslation();
```
agilgur5: same as above, this should be with the other `use` variable declarations above

ui/src/app/workflows/components/workflows-summary-container/workflows-summary-container.tsx
```diff
@@ -5,6 +5,7 @@ import {useEffect, useState} from 'react';
 import {Workflow} from '../../../../models';
 
 import './workflows-summary-container.scss';
+import {useTranslation} from 'react-i18next';
```
agilgur5, line 8: same as above

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -7,6 +7,7 @@ import {ParametersInput} from '../../shared/components/parameters-input';
 import {TagsInput} from '../../shared/components/tags-input/tags-input';
 import {services} from '../../shared/services';
 import {Utils} from '../../shared/utils';
+import {useTranslation} from "react-i18next";
```
agilgur5, line 10: same as above. this will also fail lint -- should be single quote

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -62,16 +63,17 @@ export function SubmitWorkflowPanel(props: Props) {
         }
     }
 
+    const {t} = useTranslation();
```
agilgur5, line 66: same as above, this should be with the variable declarations at the top of the component

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -182,6 +183,7 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
     }, [nodeId, workflow]);
 
     const parsedSidePanel = parseSidePanelParam(sidePanel);
+    const {t} = useTranslation();
```
agilgur5, line 186: same as above, should be with the `use` variable declarations at the top of the component

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -158,6 +159,7 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
     }, [namespace, phases.toString(), labels.toString(), pagination.limit, pagination.offset]); // referential equality, so use values, not refs
 
     useCollectEvent('openedWorkflowList');
+    const {t} = useTranslation();
```
agilgur5, line 162: same as above, this should be before all the `useEffect` with the other variable declarations at the beginning of the component

ui/src/app/workflows/components/workflows-summary-container/workflows-summary-container.tsx
```diff
@@ -18,6 +19,7 @@ export function WorkflowsSummaryContainer(props: {workflows: Workflow[]}) {
             setWfSummary(summary);
         }
     }, [props.workflows]);
+    const {t} = useTranslation();
```
agilgur5, line 22: same as above, this should be before the `useEffect` with the other variable declarations above

ui/public/locales/zh/translation.json
```diff
@@ -0,0 +1,34 @@
+{
+  "title": {
```
agilgur5, line 2: I wonder if we should just leave these all uncategorized as a single list. Since things like `entrypoint` and `parameters` are not necessarily titles and that can get very semantic very fast.

at the very least, I do think every entry should be alphabetized

agilgur5 (commented):
Thanks for taking a stab at this! I think it's a good start if we have people available to maintain this (I'm not sure we do per https://github.com/argoproj/argo-workflows/issues/10103#issuecomment-1327717345. Also the docs are only in English right now).

There's a number of stylistic issues that I've left comments on below. In general, when editing a file, please try to stay consistent with the existing style in that file

ui/public/locales/zh/translation.json
LinuxSuRen: Sure.

ui/src/app/login/login.tsx
```diff
@@ -25,8 +26,9 @@ function getRedirect(): string {
 
 export function Login() {
     useCollectEvent('openedLogin');
+    const {t} = useTranslation();
```
agilgur5, line 29:
```suggestion
    const {t} = useTranslation();

    useCollectEvent('openedLogin');

```

same as [previous comment](https://github.com/argoproj/argo-workflows/pull/12682#discussion_r1495337679)

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -94,6 +95,7 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
 
     const filteredWorkflows = workflows?.filter(w => nullSafeTimeFilter(createdAfter, finishedBefore, w)).slice(0, pagination.limit || 999999);
     const counts = countsByCompleted(filteredWorkflows);
+    const {t} = useTranslation();
```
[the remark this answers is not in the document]
LinuxSuRen: You mean line 85?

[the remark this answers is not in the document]
agilgur5: Yep

shuangkun:
There are many Chinese users. This is a good beginning！

LinuxSuRen:
> There are many Chinese users. This is a good beginning！

Please feel free to give it a review. And thanks for your response.

shuangkun:
Do we need to focus these? Or you will follow here later. 
image
image
image
image

shuangkun:
> Consider this is the first PR of i18n. The most important meaning of it is having the i18n feature. And we will add all the Chinese words evantually.

OK. For users, consistency in UI style is very important.

LinuxSuRen:
> OK. For users, consistency in UI style is very important.

Sure.

ui/yarn.lock
```diff
@@ -969,10 +969,10 @@
...
-  version "7.23.8"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.23.8.tgz#8ee6fe1ac47add7122902f257b8ddf55c898f650"
-  integrity sha512-Y7KbAP984rn1VGMbGqKmBLio9V7y5Je9GvU4rQPCPinCyNfUcToxIXl06d59URp/F3LwinvODxab5N/G6qggkw==
+"@babel/runtime@^7.1.2", "@babel/runtime@^7.10.1", "@babel/runtime@^7.10.4", "@babel/runtime@^7.11.1", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.1", "@babel/runtime@^7.12.5", "@babel/runtime@^7.16.7", "@babel/runtime@^7.18.0", "@babel/runtime@^7.18.3", "@babel/runtime@^7.20.0", "@babel/runtime@^7.20.7", "@babel/runtime@^7.21.0", "@babel/runtime@^7.22.5", "@babel/runtime@^7.23.2", "@babel/runtime@^7.23.4", "@babel/runtime@^7.23.9", "@babel/runtime@^7.3.1", "@babel/runtime@^7.8.4", "@babel/runtime@^7.9.2":
+  version "7.23.9"
+  resolved "https://registry.npmmirror.com/@babel/runtime/-/runtime-7.23.9.tgz#47791a15e4603bb5f905bc0753801cf21d6345f7"
```
agilgur5, line 974: we shouldn't change the registry; IIRC you can configure that in your local `~/.yarnrc`

ui/src/app/i18n.js
```diff
@@ -0,0 +1,11 @@
+import i18n from 'i18next';
+import Backend from 'i18next-http-backend';
+import LanguageDetector from 'i18next-browser-languagedetector';
+import {initReactI18next} from 'react-i18next';
+
+i18n.use(Backend).use(LanguageDetector).use(initReactI18next).init({
+    fallbackLng: 'en',
+    debug: false
+});
```
agilgur5, line 9: two problems I see with this approach:

1. There's no way to switch languages in the UI right now. You're forced into using the language your browser detects for your locale, even if you don't like or don't prefer that translation.
1. Translations are currently hosted in core in this approach, but core maintainers may not be able to maintain these translations (per https://github.com/argoproj/argo-workflows/pull/12682#pullrequestreview-1889637846 and https://github.com/argoproj/argo-workflows/issues/10103#issuecomment-1327717345).
    - I think a plugin based approach, similar to the Jenkins Chinese plugin that you maintain as you mentioned in [the issue](https://github.com/argoproj/argo-workflows/issues/10103) would be easier to maintain and be more explicit about the level of support (since translations would be done by contributors). We could host translations in `argoproj-labs` too if desired.
        - How to implement the plugin based approach is a bit more complicated as UI plugins don't yet exist (see also #6943. The existing UI subtask #6945 also does not outline an approach for "UI extensions" such as this that could modify the entire UI, as opposed to just add a new page).
           - A very simple, naive approach would be to allow users to add a URL to a translation file that is fetched. Then we can store that in `localStorage`
           - Another approach would be for the API to serve translation files located in a specific directory on disk. Then one can add a volume with those or make a custom image with them.
               - This roughly follows how [Argo CD loads UI extensions](https://argo-cd.readthedocs.io/en/stable/developer-guide/extensions/ui-extensions/) and is more secure as well (it requires an operator to add the config and is served internally instead of externally, the latter of which could violate the CSP if not properly configured)
            - The other common approach, particularly in corporate settings, is to host translations on a separate website with a translation CMS of sorts. There are OSS tools like [Weblate](https://github.com/WeblateOrg/weblate) and others that are also often free to use for other OSS projects.
              - This is probably the most robust and translation-specific approach

ui/public/locales/zh/translation.json
```diff
@@ -0,0 +1,27 @@
...
+  "cronWorkflow": "定时工作流",
+  "entrypoint": "入口",
+  "error": "错误",
+  "failed": "失败",
+  "finishedBefore": "结束",
+  "label": "标签",
+  "labels": "标签",
+  "login": "登录",
+  "logs": "日志",
+  "namespace": "命名空间",
```
agilgur5, line 12: curious how you chose some of these translations -- did you use the official k8s docs translations as a starting point? Ideally we should match those where possible

I ask because most of these words are technical terms and so they don't translate easily. In other languages I know well enough (e.g. Russian and Spanish), almost none of these translate easily, and many technical words are loan words from English too.

Admittedly I do not have much experience as a localization writer/editor, so I don't really know the norms or best practices therein (I am familiar with adding _support_ for i18n like with React components as you did here, but not in writing translations myself).
I know there are some localization tools that allow users to "rate" translations and suggest alternatives as a feedback mechanism, which can help the iteration loop.

ui/src/app/i18n.js
agilgur5, line 1: this should be a `.ts` file (and add any typings if necessary) -- the whole codebase is TypeScript after all

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -170,7 +172,7 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
                 actionMenu: {
                     items: [
                         {
-                            title: 'Submit New Workflow',
+                            title: t('submitNewWorkflow'),
```
agilgur5, line 175: Looking at this again, I feel like the translations would be easier to read/edit/write if they were 1-to-1 with the English, no? Since that's what we're translating from. The current approach effectively creates a "code word" for each translation that loses spacing and capitalization information.

i.e. if the inputs were 1-to-1:
```suggestion
                            title: t('Submit New Workflow'),
```

ui/src/app/login/login.tsx
```diff
@@ -24,9 +25,11 @@ function getRedirect(): string {
 }
 
 export function Login() {
+    const {t} = useTranslation();
+
     useCollectEvent('openedLogin');
     return (
-        <Page title='Login' toolbar={{breadcrumbs: [{title: 'Login'}]}}>
+        <Page title={t('login')} toolbar={{breadcrumbs: [{title: 'Login'}]}}>
```
agilgur5, line 32:
```suggestion
        <Page title={t('login')} toolbar={{breadcrumbs: [{title: t('Login')}]}}>
```

This might not matter for the login page, but in some cases the breadcrumbs are visible text

agilgur5 (commented):
A few comments on the general approach -- in particular, where to host translations and how to change your language in the UI.

Plus some other small changes

agilgur5:
cc @terrytangyuan as a Lead (this would be a pretty big feature) and having responded to the original issue #10103.

ui/yarn.lock
LinuxSuRen: Sure.

ui/src/app/i18n.js
LinuxSuRen: I can add a drop-down menu for language switch.

Having a localization plugin will make it be easier to maintain and publish. Currently I am able to help on the plugin. A very important thing is that the particular language (Chinese for this one) committer is necessary. There is no one willing to do that for the Chinese localization, I prefer to pause this PR.

ui/public/locales/zh/translation.json
LinuxSuRen: Yes, as you mentioned. The Kubernetes dashboard will be a good example which we can learn from. The `namespace` usually be translated to `命名空间`.

ui/src/app/i18n.js
agilgur5: > A very important thing is that the particular language (Chinese for this one) committer is necessary. There is no one willing to do that for the Chinese localization, I prefer to pause this PR.

Yes that's exactly what I meant! I'm not sure if we have anyone able & willing to maintain it actively.

And we definitely couldn't do that for all localizations, so community plugins would make more sense _overall_ as an approach for _all_ localizations.

For Chinese, for instance, you would be able to maintain that specific localization in its own repo (or in a shared `argo-workflows-ui-localizations` repo or something, split by `CODEOWNERS` similar to `argo-helm` for instance).
We could also potentially package some more maintained ones into the UI by default (with the caveat that the default ones could quickly go out-of-date).
If we were to use a hosted site, we could add contributors as editors for different localizations.

Issue #12633, opened 2024-02-06
inline task in DAG makes workflow fail: `failed to get a template`


partomatl:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

With v3.4.6, the attached workflow works reliably.
From v3.4.7 and upwards, the workflow randomly fails with error `failed to get a template` :

<img width="1197" alt="screen" src="https://github.com/argoproj/argo-workflows/assets/14841318/20dbbaed-4947-4753-8772-8e4e870cc1e7">

I don't know if this PR that was merged for v3.4.7 is related : https://github.com/argoproj/argo-workflows/pull/10786

### Version

v3.5.4

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: wftmplcontainer
spec:
  templates:
    - name: print
      inputs:
        parameters:
          - name: message
      container:
        image: alpine
        command: [echo, "{{inputs.parameters.message}}"]
---
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: wftmpl
spec:
  templates:
    - name: dagtmpl
      dag:
        tasks:
          - name: message
            inline:
              script:
                image: python:alpine
                command: [python]
                source: |
                  print("hello world")

          - name: echo
            arguments:
              parameters:
                - name: message
                  value: "{{tasks.message.outputs.result}}"
            dependencies: [message]
            templateRef:
              name: wftmplcontainer
              template: print
---
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  name: cwf
spec:
  schedule: "* * * * *"
  successfulJobsHistoryLimit: 30
  failedJobsHistoryLimit: 30
  workflowSpec:
    entrypoint: default
    templates:
      - name: default
        steps:
          - - name: first
              templateRef:
                name: wftmpl
                template: dagtmpl
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: workflow-controller-configmap
data:
  workflowDefaults: |
    spec:
[...]

PR #12683, opened 2024-02-20
fix: find correct retry node when using `templateRef`. Fixes: #12633

shuangkun:
The root cause is pod have `retryStrategy`, the boundaryNode didn't have template (use `templateRef`), so when find retryNode,  probabilistically found others and cause `"failed to get a template"`.

Some info help review:
Failed workflow(find the wrong retry node):
image
Succeed worklfow:
image

Fixes: #12633

### DIFF
--- a/workflow/controller/retry_tweak.go
+++ b/workflow/controller/retry_tweak.go
@@ -15,12 +15,23 @@ type RetryTweak = func(retryStrategy wfv1.RetryStrategy, nodes wfv1.Nodes, pod *
 func FindRetryNode(nodes wfv1.Nodes, nodeID string) *wfv1.NodeStatus {
 	boundaryID := nodes[nodeID].BoundaryID
 	boundaryNode := nodes[boundaryID]
-	templateName := boundaryNode.TemplateName
-	for _, node := range nodes {
-		if node.Type == wfv1.NodeTypeRetry && node.TemplateName == templateName {
-			return &node
+	if boundaryNode.TemplateName != "" {
+		templateName := boundaryNode.TemplateName
+		for _, node := range nodes {
+			if node.Type == wfv1.NodeTypeRetry && node.TemplateName == templateName {
+				return &node
+			}
 		}
 	}
+	if boundaryNode.TemplateRef != nil {
+		templateRef := boundaryNode.TemplateRef
+		for _, node := range nodes {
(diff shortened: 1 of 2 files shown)


workflow/controller/retry_tweak.go
```diff
@@ -15,12 +15,23 @@ type RetryTweak = func(retryStrategy wfv1.RetryStrategy, nodes wfv1.Nodes, pod *
...
+		for _, node := range nodes {
+			if node.Type == wfv1.NodeTypeRetry && node.TemplateName == templateName {
+				return &node
+			}
 		}
 	}
+	if boundaryNode.TemplateRef != nil {
+		templateRef := boundaryNode.TemplateRef
+		for _, node := range nodes {
+			if node.Type == wfv1.NodeTypeRetry && node.TemplateRef != nil && node.TemplateRef.Name == templateRef.Name && node.TemplateRef.Template == templateRef.Template {
```
Joibel, line 29: Why do we check a match of both `TemplateRef.Name` and `TemplateRef.Template` here?

Joibel: No, you're right. I was looking in the wrong place at this.

Joibel (approved):
LGTM, thanks for digging down to this.

workflow/controller/retry_tweak_test.go
```diff
@@ -59,6 +59,24 @@ func TestFindRetryNode(t *testing.T) {
...
 			TemplateName: "tmpl2",
 		},
+		"E1": wfv1.NodeStatus{
+			ID:         "E1",
+			Type:       wfv1.NodeTypeRetry,
+			Phase:      wfv1.NodeRunning,
+			BoundaryID: "A1",
+			Children:   []string{},
+			TemplateRef: &wfv1.TemplateRef{
+				Name: "tmpl1",
```
agilgur5, line 69: shouldn't this have a `template` field as well? the test is passing with only `name`?

shuangkun: Ok,I have added it.

agilgur5: > the test is passing with only `name`?

Oh I guess the test passed bc `nil == nil`. The code checks whether the `templateRef` struct isn't `nil`, but not the individual fields.
They should never be `nil` in a real use-case, so I suppose that is a test edge-case

agilgur5 (approved):
LGTM, thanks for root causing and fixing this!

PR #12617, opened 2024-02-04
docs: consolidate expression links

agilgur5:
Follow-up to https://github.com/argoproj/argo-workflows/pull/12469#discussion_r1443145079 and https://github.com/argoproj/argo-workflows/pull/12573#discussion_r1471449979

### Motivation

- most of these should go to the definition in Argo's own docs in `variables.md#expression`
- use https://expr-lang.org/docs/language-definition instead of GH as it has nicer UX

### Modifications

- change internal docs references to expressions to go to `variables.md#expression`, the internal docs definition of it
- change external references to go to https://expr-lang.org/docs/language-definition

- while at it, make some small docs changes in the same place
  - more simple and direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
    - some of these, especially the `data` one, were incredibly overly verbose and have been greatly simplified
  - use present tense, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-present-tense)
  - prefer 1 sentence per line, per [style guide](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/)

### Verification

1. `make docs` passes

PR #12696, opened 2024-02-24
docs: add CronWorkflow stopStrategy docs 

eduardodbr:
Docs for `stopStrategy` implemented in #12305 to be released on v3.6

### DIFF
--- a/docs/cron-workflows.md
+++ b/docs/cron-workflows.md
@@ -122,6 +122,13 @@ stopStrategy:
   condition: succeeded >= 1
 ```
 
+It is also possible to stop scheduling new workflows after three failures with:
+
+```yaml
+stopStrategy:
+  condition: failed >= 3
+```
+
 **NOTE:** Depending on the time it takes to schedule and run a workflow, it is possible that the total number of executions of a `CronWorkflow` is higher than the configured max executions. For example, if you configure the `CronWorkflow` to execute every minute with a stopping condition: `succeeded >= 1` and the `Workflow` executes in 90 seconds, the `CronWorkflow` will be completed with 2 executions, because when the stopping condition is achieved, there is another workflow running and the controller is not canceling running workflows. For that reason, it is advisable to have conditions like `succeeded  >= 1` instead of `succeeded == 1`.
 
 ## Managing `CronWorkflow`
--- a/docs/cron-workflows.md
+++ b/docs/cron-workflows.md
@@ -108,7 +108,7 @@ For example, with timezone set at `America/Los_Angeles`, we have daylight saving
 
 > v3.6 and after
 
-It is possible to stop the execution of new workflows from a `CronWorkflow` based on the number of previous successes or failures. The spec field `stopStrategy` accepts a `condition` that stops triggering new workflows when it evaluates to true.
+It is possible to stop the execution of new workflows from a `CronWorkflow` based on the number of previous successes or failures. The spec field `stopStrategy` accepts a `condition` which is an [expression](variables.md#expression) that stops triggering new workflows when it evaluates to true.
 
 Keywords allowed for stopping conditions are `failed` and `succeeded`:
 
--- a/docs/cron-workflows.md
+++ b/docs/cron-workflows.md
@@ -48,7 +48,7 @@ The resulting `Workflow` name will be a generated name based on the `CronWorkflo
 | `startingDeadlineSeconds`    |           `0`          | Number of seconds after the last successful run during which a missed `Workflow` will be run                                                                                                                                            |
 | `successfulJobsHistoryLimit` |           `3`          | Number of successful `Workflows` that will be persisted at a time                                                                                                                                                                       |
 | `failedJobsHistoryLimit`     | `1`                    | Number of failed `Workflows` that will be persisted at a time                                                                                                                                                                           |
-| `stopStrategy`     |         `nil`            | Defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow                                                                                                                                                                           |
+| `stopStrategy`               |         `nil`          | v3.6 and after: defines if the CronWorkflow should stop scheduling based on a condition                                                                                                                                                                      |
 
 ### Cron Schedule Syntax
 
@@ -119,14 +119,14 @@ For example, if you want a workflow to be executed only once you can define the 
 
 ```yaml
 stopStrategy:
-  condition: succeeded >= 1
+  condition: "succeeded >= 1"
 ```
 
 It is also possible to stop scheduling new workflows after three failures with:
 
 ```yaml
 stopStrategy:
-  condition: failed >= 3
+  condition: "failed >= 3"
 ```
 
 **NOTE:** Depending on the time it takes to schedule and run a workflow, it is possible that the total number of executions of a `CronWorkflow` is higher than the configured max executions. For example, if you configure the `CronWorkflow` to execute every minute with a stopping condition: `succeeded >= 1` and the `Workflow` executes in 90 seconds, the `CronWorkflow` will be completed with 2 executions, because when the stopping condition is achieved, there is another workflow running and the controller is not canceling running workflows. For that reason, it is advisable to have conditions like `succeeded  >= 1` instead of `succeeded == 1`.
--- a/docs/cron-workflows.md
+++ b/docs/cron-workflows.md
@@ -48,6 +48,7 @@ The resulting `Workflow` name will be a generated name based on the `CronWorkflo
 | `startingDeadlineSeconds`    |           `0`          | Number of seconds after the last successful run during which a missed `Workflow` will be run                                                                                                                                            |
 | `successfulJobsHistoryLimit` |           `3`          | Number of successful `Workflows` that will be persisted at a time                                                                                                                                                                       |
 | `failedJobsHistoryLimit`     | `1`                    | Number of failed `Workflows` that will be persisted at a time                                                                                                                                                                           |
+| `stopStrategy`     |         `nil`            | Defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow                                                                                                                                                                           |
 
 ### Cron Schedule Syntax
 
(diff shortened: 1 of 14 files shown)


docs/cron-workflows.md
```diff
@@ -103,6 +104,26 @@ For example, with timezone set at `America/Los_Angeles`, we have daylight saving
...
+
+- `failed` - counts the number of workflows that failed to submit or completed in phase `Error` or `Failed`
+- `succeeded` - counts the number of workflows that completed in phase `Succeeded`
+
+For example, if you want a workflow to be executed only once you can define the spec as:
+
+```yaml
+stopStrategy:
+  condition: succeeded >= 1
+```
```
blkperl, line 123: Can we add the example for failed as well?

eduardodbr: done thanks

docs/cron-workflows.md
```diff
@@ -103,6 +104,33 @@ For example, with timezone set at `America/Los_Angeles`, we have daylight saving
     |            | 2        | 2020-11-02 02:01:00 -0800 PST |
     |            | 3        | 2020-11-03 02:01:00 -0800 PST |
 
+### Stopping a `CronWorkflow`
+
+> v3.6 and after
```
terrytangyuan, line 109: We now have versioned docs so I am not sure if this is needed. @jmeridth and others - any thoughts?

agilgur5: We did talk about this in a Contributor Meeting before we even had versioned docs, [in August I think?](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.qogh82i9yij0), but it's not written in the meeting notes/minutes. The [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future) says to use notation for future versions.

But more importantly, users assume a feature is usable unless it says otherwise. We've had that before versioned docs (such as [with `deleteDelayDuration`](https://github.com/argoproj/argo-workflows/pull/11445) and others) and also after versioned docs -- the markdown title & description feature had [several users](https://github.com/argoproj/argo-workflows/pull/12582#issuecomment-1944824694) asking about it on Slack and that was the primary reason I had mentioned my docs PR on [the contributors channel](https://cloud-native.slack.com/archives/C0510EUH90V/p1708533114288009).
So user behavior means this is 100% needed

agilgur5 (commented):
Thanks for updating based off my comments in https://github.com/argoproj/argo-workflows/pull/12305#pullrequestreview-1899700002 !

There's a few more modifications on top of that needed.

Otherwise I haven't had a chance to go through the docs section itself yet. From a glance, it will need some [simplification](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language) and [active voice](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice) changes. Also the expression docs should be linked here (see #12617 for examples)

docs/fields.md
```diff
@@ -1943,12 +1943,12 @@ SynchronizationStatus stores the status of semaphore and mutex.
 
 ## StopStrategy
 
-StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
+StopStrategy defines if the CronWorkflow should stop scheduling based on a condition
```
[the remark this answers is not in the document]
eduardodbr: done, thanks for the review @agilgur5

docs/cron-workflows.md
```diff
@@ -103,6 +104,33 @@ For example, with timezone set at `America/Los_Angeles`, we have daylight saving
...
+```
+
+You can also stop scheduling new workflows after three failures with:
+
+```yaml
+stopStrategy:
+  condition: "failed >= 3"
+```
+
+**NOTE:** Depending on the time it takes to schedule and run a workflow, it is possible that the total number of executions of a `CronWorkflow` is higher than the configured max executions. For example, if you configure the `CronWorkflow` to execute every minute with a stopping condition: `succeeded >= 1` and the `Workflow` executes in 90 seconds, the `CronWorkflow` will be completed with 2 executions, because when the stopping condition is achieved, there is another workflow running and the controller is not canceling active workflows. For that reason, it is advisable to have conditions like `succeeded  >= 1` instead of `succeeded == 1`.
```
agilgur5, line 132: This should use a material admonition (see also #12561), and really should be a warning as it's pretty important.

```suggestion
!!! Warning Scheduling vs. Completions
     Depending on the time it takes to schedule and run a workflow, the number of completions can exceed the configured maximum.
     
     For example, say you configure a `CronWorkflow` to schedule every minute (`* * * * *`) and stop after one success (`succeeded >= 1`).
     If the `Workflow` takes 90 seconds to run, the `CronWorkflow` will actually stop after two completions.
     This is because when the stopping condition is achieved, there is _already_ another `Workflow` running.
     For that reason, prefer conditions like `succeeded  >= 1` over `succeeded == 1`.
```
- prefer 1 sentence per line of markdown, per [style guide](https://argo-workflows.readthedocs.io/en/latest/doc-changes/).
  - "it is possible that the total" can be simplified
  - "is higher than" -> "exceed", "max executions" -> "maximum"
  
- be consistent with terminology and numerics
  - "schedule" over "execute", same as above (and [previous suggestions](https://github.com/argoproj/argo-workflows/pull/12305#discussion_r1501842596))
  - "executions" -> "completions" -- as "complete" is an existing term the docs already use
  - "will be completed" -> "will stop" -- as this is `stopStrategy` and this section is on "stopping"
  - show both the condition and schedule in their respective YAML notation
  - "two" instead of "2" as you have "three" and "one" previously
  - "`Workflow`" to be consistent with in this note
  
- re-word the last sentence for simplicity and active voice, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice)

- "the controller" is an architectural implementation detail that doesn't need to be stated here. that running workflows are not cancelled is also implied by the very nature of this note, so I don't think that's necessary either.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -61,15 +61,15 @@ type CronWorkflowSpec struct {
...
 	// Schedules is a list of schedules to run the Workflow in Cron format
 	Schedules []string `json:"schedules,omitempty" protobuf:"bytes,11,opt,name=schedules"`
 }
 
-// StopStrategy defines if the cron workflow will stop being triggered once a certain condition has been reached, involving a number of runs of the workflow
+// v3.6 and after: StopStrategy defines if the CronWorkflow should stop scheduling based on a condition
 type StopStrategy struct {
-	// Condition defines a condition that stops scheduling workflows when evaluates to true. Use the
+	// v3.6 and after: Condition is an expression that stops scheduling workflows when true. Use the
 	// keywords `failed` or `succeeded` to access the number of failed or successful child workflows.
```
agilgur5, line 73:
```suggestion
	// variables `failed` or `succeeded` to access the number of failed or successful child workflows.
```
same as above, "variables" is the existing term used in the docs

agilgur5 (commented):
Couple of simplifications and clarifications in-line below

docs/cron-workflows.md
```diff
@@ -103,6 +104,33 @@ For example, with timezone set at `America/Los_Angeles`, we have daylight saving
     |            | 2        | 2020-11-02 02:01:00 -0800 PST |
     |            | 3        | 2020-11-03 02:01:00 -0800 PST |
 
+### Stopping a `CronWorkflow`
```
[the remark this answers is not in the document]
eduardodbr: The cronworkflow is not set as "suspended", it is just removed from the queue and cannot become active again, so I think it is best to be different.

eduardodbr:
@agilgur5  do you think this is good to merge?

docs/cron-workflows.md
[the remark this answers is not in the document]
agilgur5: Oh word, I misunderstood the feature, you're right, "stop" is the correct wording then.
Huh I would think it would just auto-delete after that so it doesn't take up space in etcd 🤔 (although I can see that keeping it _somewhere_ might be useful for historical/archive/compliance purposes, although you already have the Workflows themselves)

docs/variables.md
```diff
@@ -276,6 +276,17 @@ For `Template`-level metrics:
 | `workflow.status` | Workflow status. One of: `Succeeded`, `Failed`, `Error` |
 | `workflow.failures` | A list of JSON objects containing information about nodes that failed or errored during execution. Available fields: `displayName`, `message`, `templateName`, `phase`, `podName`, and `finishedAt`. |
 
+### Stop Strategy
```
agilgur5, line 279:
```suggestion
### `stopStrategy`
```

We can probably mimic `retryStrategy` in naming

docs/variables.md
```diff
@@ -276,6 +276,17 @@ For `Template`-level metrics:
...
+
+> v3.6 and after
+
+When using the `condition` field within `stopStrategy` in `CronWorkflows`, special variables are available.
```
agilgur5, line 283:
```suggestion
When using the `condition` field within the [`stopStrategy` of a `CronWorkflow`](cron-workflows.md#automatically-stopping-a-cronworkflow), special variables are available.
```

- fix grammar a bit
- link back to the Cron docs

agilgur5 (commented):
Sorry for the long delay here, my to-do list is pretty long and I'm particularly behind on longer PRs as I need a longer blocks of concentrated time for them. This one actually isn't that long though, might've misremembered that.

I also deprioritized this since it's for v3.6, which has been [delayed, as you know](https://cloud-native.slack.com/archives/C0510EUH90V/p1718116911755969) (we've had a lot of high priority fixes for 3.5 recently).
Although some version references were missing before which is kinda important to prevent people thinking they exist already, so would've been great if I had addressed that part at least earlier 😅 

Also I do still wish GH code review had hunk checkmarks, as I've definitely re-review the same hunk multiple times without that 😕 

This should be the last set of changes though!

docs/cron-workflows.md
```diff
@@ -103,6 +104,35 @@ For example, with timezone set at `America/Los_Angeles`, we have daylight saving
...
+You can also stop scheduling new workflows after three failures with:
+
+```yaml
+stopStrategy:
+  condition: "failed >= 3"
+```
+
+!!! Warning "Scheduling vs. Completions"
+    Depending on the time it takes to schedule and run a workflow, the number of completions can exceed the configured maximum.
+    For example, if you configure the `CronWorkflow` to schedule every minute (`* * * * *`) and stop after one success (`succeeded >= 1`).
```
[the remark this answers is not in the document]
agilgur5: Ah I see, it renders fine (see screenshot below), but fails `markdownlint`. We can disable that error for this section.

Screenshot 2024-06-23 at 6 34 58 PM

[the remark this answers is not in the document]
agilgur5: updated that myself in https://github.com/argoproj/argo-workflows/pull/12696/commits/430b4874dd2b57d944514b0798ce7fb3bad10ae0

agilgur5 (approved):
Made the requested changes myself, so LGTM now. Thanks for your patience and iteration here!

Issue #12317, opened 2023-12-04
ternary expression is not evaluated when using a workflow parameter


ndamclean:
### Pre-requisites

- [X] I have double-checked my configuration
- [x] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

When I reference a workflow parameter in an expression `{{= ... }}` the entire expression is not parsed and is treated as a string.

For example, the following workflow uses a ternary expression to determine whether to run `true` or `false` based on a workflow parameter called `succeed`.

I would expect the attached workflow to exit successfully (exit code `0`) if `succeed` is set to `true` and to fail if it is set to `false`.

However, Argo treats the entire expression as a string without parsing it, resulting in an error message:
```
Error: failed to find name in PATH: exec: "{{= workflow.parameters['succeed'] ? 'true' : 'false' }}": executable file not found in $PATH
failed to find name in PATH: exec: "{{= workflow.parameters['succeed'] ? 'true' : 'false' }}": executable file not found in $PATH
```

If I replace `workflow.parameters['succeed']` with a hard-coded `true` or `false` value, the workflow works as expected.

Furthermore, if Argo is unable to parse an expression, I would expect `argo submit` or `argo lint` to give an error instead of simply treating the unparsed expression as a string. This behaviour makes it very difficult to debug expressions in workflows.

### Version

v3.4.3

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
---
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: test-workflow-
  labels:
    name: test-workflow
spec:
  arguments:
    parameters:
      - name: succeed
        value: true
  entrypoint: main
  templates:
    - name: main
      container:
        image: "alpine:3.18"
        imagePullPolicy: "Always"
        command: 
            - "{{= workflow.parameters['succeed'] ? 'true' : 'false' }}"
```

### Logs from the workflow controller

```text
time="2023-12-04T19:48:18.108Z" level=info msg="Processing workflow" namespace=resilience-data-eng workflow=test-workflowscj9x
time="2023-12-04T19:48:18.113Z" level=info msg="Updated phase  -> Running" namespace=resilience-data-eng workflow=test-workflowscj9x
[...]

PR #12699, opened 2024-02-26
fix: return correct error information when use ternary expression. Fixes: #12317

shuangkun:
Fixes: #12317 

### Motivation

Return correct error information to help customers locate the problem

### Modifications

When parsing an expression, if an "interface conversion" error is encountered, return an error instead of ignoring it.

### Verification

e2e tests

### DIFF
--- a/util/template/expression_template.go
+++ b/util/template/expression_template.go
@@ -160,7 +160,7 @@ func resolveExpression(expression string, env map[string]interface{}, allowUnres
 	// it allows for errors that are obviously
 	// not failed reference checks to also pass
 	if err != nil && !allowUnresolved {
-		return expression, fmt.Errorf("failed to evaluate expression: %w", err)
+		log.WithError(err).Debug("failed to evaluate expression: %w", err)
 	}
 	result, err := expr.Run(program, env)
 	if (err != nil || result == nil) && allowUnresolved {
--- a/util/template/expression_template.go
+++ b/util/template/expression_template.go
@@ -24,14 +24,14 @@ func expressionReplace(w io.Writer, expression string, env map[string]interface{
 	if isTernaryExpression(expression) {
 		splits := strings.SplitN(expression, "?", 2)
 		cond := splits[0]
-		result, err := runExpression(cond, env, allowUnresolved)
+		result, err := resolveExpression(cond, env, allowUnresolved)
 		if err != nil {
 			return w.Write([]byte(fmt.Sprintf("{{%s%s}}", kindExpression, expression)))
 		}
 		expression = fmt.Sprintf("%s?%s", result, splits[1])
 	}
 
-	result, err := runExpression(expression, env, allowUnresolved)
+	result, err := resolveExpression(expression, env, allowUnresolved)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


util/template/expression_template.go
```diff
@@ -21,53 +21,25 @@ func init() {
...
-	// This issue doesn't happen to other template parameters since `workflow.status` and `workflow.failures` only exist in the env
-	// when the exit handlers complete.
-	if ((hasWorkflowStatus(unmarshalledExpression) && !hasVarInEnv(env, "workflow.status")) ||
-		(hasWorkflowFailures(unmarshalledExpression) && !hasVarInEnv(env, "workflow.failures"))) &&
-		allowUnresolved {
-		return w.Write([]byte(fmt.Sprintf("{{%s%s}}", kindExpression, expression)))
+	if isTernaryExpression(expression) {
+		splits := strings.SplitN(expression, "?", 2)
+		cond := splits[0]
+		result, err := resolveExpression(cond, env, allowUnresolved)
```
agilgur5, line 27: doing a partial eval here sounds like it could potentially cause a whole lot of issues, I don't think we should be splitting up expressions like that as `expr` was not meant to be used that way. Special casing ternaries also feels potentially problematic.

could you solve this a different way without either of those?

util/template/expression_template.go
```diff
@@ -151,3 +123,62 @@ func hasVarInEnv(env map[string]interface{}, parameter string) bool {
 	_, ok := flattenEnv[parameter]
 	return ok
 }
+
+// isTernaryExpression check if the expression is a ternary expression
+func isTernaryExpression(expression string) bool {
+	if strings.Contains(expression, "?") && strings.Contains(expression, ":") && len(expression) >= 5 {
```
agilgur5, line 129: there's definitely some false positives here, for instance `workflow.parameters?.json ?? ['thing1', 'thing2'][1:]`

that's one reason why we should avoid parsing `expr` expressions on our own, as I mentioned above. a naive parse can cause a lot of issues

agilgur5 (changes requested):
this approach has a few issues with it, see in-line below

util/template/expression_template.go
shuangkun: Yes, I think split is not a good solution, may causing other issues. May be we can report correct error for user to debug, if have "interface conversion" error when resolveExpression.

util/template/expression_template.go
```diff
@@ -61,6 +61,9 @@ func expressionReplace(w io.Writer, expression string, env map[string]interface{
 	if (err != nil || result == nil) && allowUnresolved {
 		//  <nil> result is also un-resolved, and any error can be unresolved
 		log.WithError(err).Debug("Result and error are unresolved")
+		if err != nil && strings.Contains(err.Error(), "interface conversion") {
+			return 0, fmt.Errorf("failed to evaluate expression: %w", err)
+		}
```
isubasinghe, line 66: @agilgur5 not sure if I am a fan of this kind of string based error matching, this could easily break if expr lang got updated. 

Do you have opinions?

I think I'd rather this be refactored so that it's more explicit that it belongs to expr-lang.

Probably util/expr/argoexpr/error.go ?

PR #12700, opened 2024-02-26
feat: support dag and steps level scheduling constraints. Fixes: #12568

### LINKED ISSUE #12568: Workflow ignoring nodeSelector and tolerations when defined in WorkflowTemplate
### Pre-requisites

I defined a WorkflowTemplate containing `nodeSelector` and `tolerations`. When I manually submit the WorkflowTemplate from the UI both fields are ignored and the workflow is running on any available node in the cluster.

### Version

v3.5.4

```YAML
kind: WorkflowTemplate
metadata:
  name: benchmarks
spec:
  entrypoint: main
  serviceAccountName: workflow
  templates:
  - dag:
      tasks:
      - arguments:
          parameters:
          - name: msg
            value: 'hello'
        name: benchmark
        template: benchmark
    name: main
    nodeSelector:
      pool: workflows
    tolerations:
    - key: pool
      operator: Equal
      value: workflows
  - inputs:
      parameters:
      - name: msg
    name: benchmark
    script:
      command:
      - python
      image: python:latest
      source: |
        print("{{inputs.parameters.msg}}")
```

```text
```

```text
```

shuangkun:
Fixes #12568

### Motivation

I want to Take effect dag and steps level scheduling constraints. Not just workflow level and template level.

### Modifications

Find the pod's bounaryID's Template and add the scheduling constraints to pod if exists.

### Verification

ut and e2e test

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -826,26 +826,35 @@ func (woc *wfOperationCtx) addSchedulingConstraints(pod *apiv1.Pod, wfSpec *wfv1
 	}
 }
 
-// GetBoundaryTemplate get a template through the node's BoundaryID.
+// GetBoundaryTemplate get a template through the nodeName
 func (woc *wfOperationCtx) GetBoundaryTemplate(nodeName string) (*wfv1.Template, error) {
 	node, err := woc.wf.GetNodeByName(nodeName)
 	if err != nil {
 		woc.log.Warnf("couldn't retrieve node for nodeName %s, will get nil templateDeadline", nodeName)
 		return nil, err
 	}
-	boundaryNode, err := woc.wf.Status.Nodes.Get(node.BoundaryID)
-	if err != nil {
-		return nil, err
-	}
-	tmplCtx, err := woc.createTemplateContext(boundaryNode.GetTemplateScope())
-	if err != nil {
-		return nil, err
-	}
-	_, boundaryTmpl, _, err := tmplCtx.ResolveTemplate(boundaryNode)
+	boundaryTmpl, _, err := woc.GetTemplateByBoundaryID(node.BoundaryID)
 	if err != nil {
 		return nil, err
 	}
 	return boundaryTmpl, nil
+}
+
+// GetTemplateByBoundaryID get a template through the node's BoundaryID.
+func (woc *wfOperationCtx) GetTemplateByBoundaryID(boundaryID string) (*wfv1.Template, bool, error) {
+	boundaryNode, err := woc.wf.Status.Nodes.Get(boundaryID)
+	if err != nil {
+		return nil, false, err
+	}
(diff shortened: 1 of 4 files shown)


workflow/controller/workflowpod.go
```diff
@@ -815,6 +826,28 @@ func addSchedulingConstraints(pod *apiv1.Pod, wfSpec *wfv1.WorkflowSpec, tmpl *w
 	}
 }
 
+// GetBoundaryTemplate get a template through the node's BoundaryID.
+func (woc *wfOperationCtx) GetBoundaryTemplate(nodeName string) (*wfv1.Template, error) {
```
Joibel, line 830: Can something like this function also be used from https://github.com/argoproj/argo-workflows/blob/e00abd1ded9e8f4787093b84f6cb26ee0870ec7c/workflow/controller/operator.go#L2740

The two bits of code are very similar to each other, and it's good to extract common functionality whilst we're doing this kind of work.

This function could take a *NodeStatus instead

workflow/controller/workflowpod.go
```diff
@@ -764,22 +764,33 @@ func (woc *wfOperationCtx) addMetadata(pod *apiv1.Pod, tmpl *wfv1.Template) {
 }
 
 // addSchedulingConstraints applies any node selectors or affinity rules to the pod, either set in the workflow or the template
-func addSchedulingConstraints(pod *apiv1.Pod, wfSpec *wfv1.WorkflowSpec, tmpl *wfv1.Template) {
+func (woc *wfOperationCtx) addSchedulingConstraints(pod *apiv1.Pod, wfSpec *wfv1.WorkflowSpec, tmpl *wfv1.Template, nodeName string) {
```
Joibel, line 767: Potentially switching to nodeID instead of nodeName makes some of this slightly simpler

Joibel (changes requested):
This is good stuff.

I'd just like to rework some of the function calls here to allow a bit more code reuse.

workflow/controller/workflowpod.go
shuangkun: OK, thanks I will.

workflow/controller/workflowpod.go
shuangkun: OK， I will try it.

shuangkun: I have made some reuses, please take a look，Thanks.

workflow/controller/workflowpod.go
shuangkun: I find it better to use nodeName to find the Boundary because we have GetNodeByName.

Joibel: Ok, what you've done with two functions below makes sense. I like it.

Joibel (approved):
LGTM, thanks @shuangkun

juliev0 (approved):

PR #12701, opened 2024-02-26
fix: nodeAntiAffinity is not working as expected when boundaryID is empty. Fixes: #9193

### LINKED ISSUE #9193: `retryStrategy` with `nodeAntiAffinity` is not working as expected.

shuangkun:
Fixes #9193

### Motivation

The root cause is the workflow's boundaryID is empty. So did't find the retry node to fetch the HostNodeName. 
So I want to find the retry node and fetch the HostNodeName as the new pod‘s nodeAntiAffinity.

### Modifications

Find the correct retrynode when boundaryID is empty.

### Verification

e2e tests  and ut

### DIFF
--- a/test/e2e/retry_test.go
+++ b/test/e2e/retry_test.go
@@ -218,10 +218,17 @@ spec:
 		WaitForWorkflow(time.Second * 90).
 		Then().
 		ExpectWorkflow(func(t *testing.T, _ *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
-			assert.Equal(t, wfv1.WorkflowPhase("Failed"), status.Phase)
-			nodeStatus := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(0)")
-			nodeStatusRetry := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(1)")
-			assert.NotEqual(t, nodeStatus.HostNodeName, nodeStatusRetry.HostNodeName)
+			if status.Phase == wfv1.WorkflowFailed {
+				nodeStatus := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(0)")
+				nodeStatusRetry := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(1)")
+				assert.NotEqual(t, nodeStatus.HostNodeName, nodeStatusRetry.HostNodeName)
+			}
+			if status.Phase == wfv1.WorkflowRunning {
+				nodeStatus := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(0)")
+				nodeStatusRetry := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(1)")
+				assert.Contains(t, nodeStatusRetry.Message, "1 node(s) didn't match Pod's node affinity/selector")
+				assert.NotEqual(t, nodeStatus.HostNodeName, nodeStatusRetry.HostNodeName)
+			}
 		})
 }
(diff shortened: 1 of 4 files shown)


workflow/controller/retry_tweak.go
```diff
@@ -14,6 +14,13 @@ type RetryTweak = func(retryStrategy wfv1.RetryStrategy, nodes wfv1.Nodes, pod *
 // FindRetryNode locates the closes retry node ancestor to nodeID
 func FindRetryNode(nodes wfv1.Nodes, nodeID string) *wfv1.NodeStatus {
 	boundaryID := nodes[nodeID].BoundaryID
+	if boundaryID == "" {
+		for _, node := range nodes {
```
terrytangyuan, line 18: This is yet another loop. Can we somehow combine it with the loop below?

shuangkun: modifid, please take a look, thanks!

test/e2e/retry_test.go
```diff
@@ -194,6 +194,45 @@ spec:
...
+			if status.Phase == wfv1.WorkflowFailed {
+				nodeStatus := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(0)")
+				nodeStatusRetry := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(1)")
+				assert.NotEqual(t, nodeStatus.HostNodeName, nodeStatusRetry.HostNodeName)
+			}
+			if status.Phase == wfv1.WorkflowRunning {
+				nodeStatus := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(0)")
+				nodeStatusRetry := status.Nodes.FindByDisplayName("test-nodeantiaffinity-strategy(1)")
+				assert.Contains(t, nodeStatusRetry.Message, "1 node(s) didn't match Pod's node affinity/selector")
+				assert.NotEqual(t, nodeStatus.HostNodeName, nodeStatusRetry.HostNodeName)
```
tczhao, line 231: with these assertions tests,
I think we could wait for workflow completed then the check phase Failed.
instead of wait 5sec have a chance of the workflow still running or failed.

What do you think?

workflow/controller/operator_test.go
```diff
@@ -7575,6 +7575,112 @@ func TestRetryOnDiffHost(t *testing.T) {
...
+    script:
+      image: python:alpine3.6
+      command: [python]
+      source: |
+        import random
+        import time
+        random.seed(time.time())
+        i = random.randint(1, 10)
+        print(i)
+        exit(i)
```
tczhao, line 7601: like the e2e test, we can shorten by just using exit, right?

terrytangyuan (approved):

PR #12705, opened 2024-02-28
fix!: only run specified plugin. Fixes #10304

### LINKED ISSUE #10304: The argo agent will request two plugin interfaces at the same time for the same task
If I have two plugins set up at the same time, on ports 4355 and 5678, then the argo agent pod will request the interface of each of these two containers for each task. Is there any way to circumvent this?

```yaml
kind: Workflow
metadata:
  name: test-wf-argo-plugin
  namespace: argo
spec:
  arguments: {}
  entrypoint: hello
  podGC:
    strategy: OnPodCompletion
  serviceAccountName: argo
  templates:
  - dag:
      tasks:
      - arguments:
          parameters:
          - name: taskName
            value: "aaa"
        name: task-packing
        template: test-plugin-one
      - arguments:
          parameters:
          - name: args
            value: "aaa"
        name: async-job-a
        template: test-plugin-two
    inputs: {}
    metadata: {}
    name: hello
    outputs: {}
  - inputs:
      parameters:
      - name: taskName
    metadata: {}
    name: test-plugin-one
    outputs: {}
    plugin:
      test-plugin-one:
        taskName: '{{inputs.parameters.taskName}}'
  - inputs:
      parameters:
      - name: args
    metadata: {}
    name: test-plugin-two
    outputs: {}
    plugin:
      test-plugin-two:
        args: '{{inputs.parameters.args}}'
```
test-plugin-two was listening on port 6789, but instead he requested 4355
image

jswxstw:
Fixes #10304

### Motivation

Multiple plugins in `AgentPod` will attempt to execute the same task, but the required plugin has already been specified in the template.

### Modifications

1. Keep the sidecar container name consistent with the plugin name
2. Maintain plugins as `map[string]executorplugins.TemplateExecutor` and try to select the plugin specified in the template to execute the task, and fallback to the original logic if it was not found.

### Verification

UT and local test.

### DIFF
--- a/cmd/argoexec/commands/agent.go
+++ b/cmd/argoexec/commands/agent.go
@@ -110,7 +110,7 @@ func initAgentExecutor() *executor.AgentExecutor {
 
 	addresses := getPluginAddresses()
 	names := getPluginNames()
-	var plugins []executorplugins.TemplateExecutor
+	plugins := make(map[string]executorplugins.TemplateExecutor, len(names))
 	for i, address := range addresses {
 		name := names[i]
 		filename := tokenFilename(name)
@@ -121,7 +121,7 @@ func initAgentExecutor() *executor.AgentExecutor {
 		if err != nil {
 			log.Fatal(err)
 		}
-		plugins = append(plugins, rpc.New(address, string(data)))
+		plugins[name] = rpc.New(address, string(data))
 	}
 
 	return executor.NewAgentExecutor(clientSet, restClient, config, namespace, workflowName, workflowUID, plugins)
--- a/pkg/apis/workflow/v1alpha1/plugin_types.go
+++ b/pkg/apis/workflow/v1alpha1/plugin_types.go
@@ -17,13 +17,34 @@ func (p *Plugin) UnmarshalJSON(value []byte) error {
 	}
 	// by validating the structure in UnmarshallJSON, we prevent bad data entering the system at the point of
 	// parsing, which means we do not need validate
-	m := map[string]interface{}{}
-	if err := json.Unmarshal(p.Object.Value, &m); err != nil {
-		return err
+	_, err := p.mapValue()
+	return err
+}
+
+// Name returns the user-specified plugin name
+func (p *Plugin) Name() (string, error) {
+	if p.Object.Value == nil {
+		return "", fmt.Errorf("plugin value is empty")
 	}
-	numKeys := len(m)
+	mapValue, err := p.mapValue()
+	if err != nil {
+		return "", err
+	}
+	for key := range mapValue {
+		return key, nil
+	}
+	return "", nil
+}
+
+// mapValue transforms the plugin value to a map of exactly one key, return err if failed
+func (p *Plugin) mapValue() (map[string]interface{}, error) {
+	mapValue := map[string]interface{}{}
+	if err := json.Unmarshal(p.Object.Value, &mapValue); err != nil {
+		return nil, err
+	}
+	numKeys := len(mapValue)
 	if numKeys != 1 {
-		return fmt.Errorf("expected exactly one key, got %d", numKeys)
+		return nil, fmt.Errorf("expected exactly one key, got %d", numKeys)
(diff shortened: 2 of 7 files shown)


dgolja:
I encountered a similar issue while working on https://github.com/argoproj/argo-workflows/issues/13026#issuecomment-2120222234 and this PR nicely solve the issue, however I am worried about the back compatibility. 

If I understand this correctly the `name` in the `sidecar.container` defines the identifier of the plugin which is use to load the right plugin. Problem is a lot of current plugins will break. For example all plugins from the [plugin-directory](https://argo-workflows.readthedocs.io/en/latest/plugin-directory/) may not work properly anymore, because for name they use `<PLUGIN_NAME>-executor-plugin`. 

Also there is no strict name convention for the name value, so some plugins took the liberty to name it a bit differently ([see 1](https://github.com/greenpau/argo-workflows-aws-plugin/blob/main/assets/plugin.yaml#L107)). Also all private plugins may break if they are not using the correct `name`.

jswxstw:
> For example all plugins from the [plugin-directory](https://argo-workflows.readthedocs.io/en/latest/plugin-directory/) may not work properly anymore, because for name they use <PLUGIN_NAME>-executor-plugin.

That's a good question which I didn't think aboud it before. One way to fix it is that replacing the sidecar name with plugin name when loading.

dgolja:
>That's a good question which I didn't think about it before. One way to fix it is that replacing the sidecar name with plugin name when loading.

I was thinking if we really want to provide back compatibility, but also a way forward we could extend `ExecutorPlugin` spec by adding `sidecar.provides []string`. Optionally you would list all the RPC the plugin provide and use that values as a filter. In case no value is provided load the plugin regardless if its used or not.

Another problem is that as it is now you could theoretically write an plugin which can provide more than one RPC, that's why `provides` should be a list.

jswxstw:
> Optionally you would list all the RPC the plugin provide and use that values as a filter. In case no value is provided load the plugin regardless if its used or not.

In fact, compatibility can be ensured without extending the spec. For example, prioritize running the specified plugin, and if it is not found, then fallback to the existing polling logic.
However, if we want to load only the plugins required for the current workflow for the agent pod, it seems impossible without extending the spec, as https://github.com/argoproj/argo-workflows/issues/10304#issuecomment-1928702804 talked about.

> Another problem is that as it is now you could theoretically write an plugin which can provide more than one RPC, that's why provides should be a list.

It seems like you want the same plugin to perform multiple functions, but this can actually be achieved within a single RPC without extending the entire executor plugins runtime framework.

dgolja:
>However, if we want to load only the plugins required for the current workflow for the agent pod, it seems impossible without extending the spec, as https://github.com/argoproj/argo-workflows/issues/10304#issuecomment-1928702804 talked about.

That would be the real end goal for me. As it is now I find it really impractical and resource wasteful to load all the plugins and eventually unusable of you have a lot of plugins.

>It seems like you want the same plugin to perform multiple functions, but this can actually be achieved within a single RPC without extending the entire executor plugins runtime framework.

Personally I would have one plugin for one RPC/task, but what I tried to say is that some users may already have plugins which provide multiple functionality.

For example no one is stopping you that in your plugin `/api/v1/template.execute` can work for multiple `plugin(s)`. Having this in mind we should be careful to not break the back compatibility.

jswxstw:
> That would be the real end goal for me. As it is now I find it really impractical and resource wasteful to load all the plugins and eventually unusable of you have a lot of plugins.

I couldn't agree more, but I currently can't find a elegant way to achieve this goal.

> For example no one is stopping you that in your plugin /api/v1/template.execute can work for multiple plugin(s). Having this in mind we should be careful to not break the back compatibility.

So, the key point is how to declare which plugins the current workflow requires and which plugin should be specified to execute the current template. 
Based on the existing logic, it seems that the design is aimed at achieving the above goal based on plugin names. However, there is no strong restriction, which makes it easy to introduce compatibility issues once modifications are made.

PR #12712, opened 2024-02-29
feat: use `VirtualHosted` addressing style as default for s3-compatible object storage. 

### LINKED ISSUE #10851: Support virtual addressing for S3 artifacts storage
# Summary

Many years ago, Amazon introduced [a new way of addressing buckets and S3 endpoints](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html), known as virtual-hosted-style. The old way, called path-style, is being deprecated. Unfortunately, Argo currently does not support S3 endpoints with virtual addressing style only. One partial workaround is to specify artifacts bucket in the endpoint and use an arbitrary subfolder as a bucket name:
```
# Not working if my-s3-endpoint.com supports only virtual addressing style
s3:
  bucket: my-awesome-bucket
  endpoint: my-s3-endpoint.com

# Partial workaround
s3:
  bucket: artifacts
  endpoint: my-awesome-bucket.my-s3-endpoint.com
```
With such S3 config, Argo can upload artifacts to s3://my-awesome-bucket/artifacts/, but in this configuration, Argo cannot list the bucket. Therefore, with virtual addressing style:

1. Showing logs in UI is broken.
2. Showing artifacts content in UI is broken.

Almost all popular S3-clients and SDKs have supported virtual addressing, including Minio SDK, which is used internally in Argo for accessing S3. Minio calls it DNS-bucket-lookup for virtual style and Path-bucket-lookup for path style. Therefore, the only change needed to support virtual addressing in Argo is a new config option for S3 to specify the desired way of addressing and then pass this value down to the Minio client.

I have already prepared a pull request and successfully tested it with an S3-storage I use. I look forward to comments from the team and some guidance as it's my first contribution to the project.

# Use Cases

According to [this article](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/), Amazon has already switched to using virtual addressing for buckets created after 2020. However, Argo users were still safe because the Minio client automatically detects the proper way of addressing for Amazon S3 and Google Cloud Platform S3.
Unfortunately, there are other S3-storage providers in the world, so Argo still needs to add an explicit support of this feature for users with other, less popular provides with virtual addressing (users like I am).

---

hustclf:
…le object storage.

Fixes #10851 

### Motivation
Due to the long-term plan to [deprecate Path-style requests in S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html), some new S3 protocol-compatible object storage services generally either support both virtual-hosts and Path-style types, or only support virtual-hosts（ . However, in Minio's default logic, if it's not an AWS, Google cloud or aliyun domain, it will [default to using the Path-style protocol](https://github.com/minio/minio-go/blob/master/pkg/s3utils/utils.go#L68).

In Argo Workflows, we should explicitly specify to use virtual-hosts as defult addressing style for new S3 protocol-compatible object storage service.

### Modifications
-  workflow/artifacts/s3/s3.go

### Verification
- github unit test for minio


.github/workflows/ci-build.yaml
```diff
@@ -261,11 +262,14 @@ jobs:
         run: docker load < /tmp/argoexec_image.tar
       - name: Set-up /etc/hosts
         run: |
-          echo '127.0.0.1 dex'      | sudo tee -a /etc/hosts
-          echo '127.0.0.1 minio'    | sudo tee -a /etc/hosts
-          echo '127.0.0.1 postgres' | sudo tee -a /etc/hosts
-          echo '127.0.0.1 mysql'    | sudo tee -a /etc/hosts
-          echo '127.0.0.1 azurite'  | sudo tee -a /etc/hosts
+          echo '127.0.0.1 dex'               | sudo tee -a /etc/hosts
```
Joibel, line 265: These changes haven't been applied to [devcontainer](https://github.com/argoproj/argo-workflows/blob/main/.devcontainer/devcontainer.json) nor explained in [the developer documentation](https://github.com/argoproj/argo-workflows/blob/main/docs/running-locally.md).

It isn't clear that they can be applied there universally either.

Joibel (changes requested):
As it stands this is a breaking change and would need adding to [the upgrading docs](https://github.com/argoproj/argo-workflows/blob/main/docs/upgrading.md)

I think it would be preferred if this was a configurable system option rather than an enforced change, defaulting to the old behaviour. This would allow for smoother upgrades and for this not to be marked as a breaking change when it doesn't need to be.

zerosign:
Hi, this MR is already 8 months, is there any issues that prevents this changes to be mainlined ? I'm willing to give a try to complete anything to mainline this MR if necessary. 

Thank you. 

bcc: @hustclf @Joibel

Joibel:
@zerosign please feel free to take over this PR by creating one of your own and addressing my comments. The comments are what are blocking me approving it. Assuming you reuse @hustclf's code in any way please attribute them with a `Co-authored-by` line. The easiest way to do that is to keep the commit as is and put your own on top of it.

hustclf:
> I think it would be preferred if this was a configurable system option rather than an enforced change, defaulting to the old behaviour. 
Sorry for late reply, and thanks for @Joibel 's advice,  It is better to offer an configurable option. 

There are many code modifications to make the e2e test pass. (For example, the configuration of coredns has been updated so that minio can work in virtualhost mode). And there is also a lack of necessary documentation.

Feel free to continue working here :).  @zerosign

blakepettersson:
@nitishfy has offered to move this to the finish line. Thank you @nitishfy!!

blakepettersson:
Superseded by #15734.

Issue #12592, opened 2024-01-30
Document & improve process for release rotation


terrytangyuan:
# Summary

Purpose: get more people involved in releases and improve the overall process.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritize the proposals with the most 👍.

terrytangyuan:
Since we have two additional approvers, I would suggest one of you (@agilgur5 @isubasinghe) can try follow the instructions in https://github.com/argoproj/argo-workflows/blob/main/docs/releasing.md and see what is missing so we can improve the docs. I don't think we need any separate documentation for this. WDYT? Any volunteers?

terrytangyuan:
For others without write access, you can still send PRs to release branch to help resolve any conflicts and then other approvers can review.

isubasinghe:
Hmm probably should note in the docs that there was an implicit alias in the previous releases.

It would be nice if the "true" "false" options for the script itself were documented, I had to look into the script itself to figure out what it was doing.

If I am correct the new commits for v3.3.5 should be based upon the HEAD of 3.3 instead of 3.3.4 ?
I find that a bit confusing. 


This release process seems like quite a bit of work, wonder if we can automate some of this effort.

terrytangyuan:
> If I am correct the new commits for v3.3.5 should be based upon the HEAD of 3.3 instead of 3.3.4 ?

Yes, see the top of the document: "Please make sure that all patch releases (e.g. v3.3.5) should be released from their associated minor release branches (e.g. release-3.3) to work well with our versioned website."

> It would be nice if the "true" "false" options for the script itself were documented, I had to look into the script itself to figure out what it was doing.

Well, the document mentions "get a list of commits you may want to cherry-pick" and "to automatically cherry-pick" with two separate code blocks already but maybe explicitly call out the flag would be helpful.

> This release process seems like quite a bit of work, wonder if we can automate some of this effort.

Feel free to propose any improvements.

isubasinghe:
> Yes, see the top of the document: "Please make sure that all patch releases (e.g. v3.3.5) should be released from their associated minor release branches (e.g. release-3.3) to work well with our versioned website."

Yeah I saw this, but what I was trying to say is that I would like it to be even more explicit, just to remove any confusion. I guess I want to understand " to work well with our versioned website" this in more depth as well. 


> but maybe explicitly call out the flag would be helpful.

Yeah I think that might be nicer. 

[...]

Issue #11997, opened 2023-10-14
Release v3.5 patch releases discussion


terrytangyuan:
This issue tracks commits for 3.5 patch releases.

terrytangyuan:
https://github.com/argoproj/argo-workflows/issues/12025 needs to be fixed for this

terrytangyuan:
I am thinking about releasing v3.5.1 next week. The most important fix would be https://github.com/argoproj/argo-workflows/pull/12068 since it blocks people from upgrading to v3.5.

terrytangyuan:
Security fix: https://github.com/argoproj/argo-workflows/pull/12111

terrytangyuan:
https://github.com/argoproj/argo-workflows/pull/12130

terrytangyuan:
https://github.com/argoproj/argo-workflows/releases/tag/v3.5.1

agilgur5:
Would be good to get my code-splitting / bundle reductions changes into the next patch release once they're all merged in (#12061, #12097, #12150). They are primarily refactors, but they do fix long wait times for the UI on slow networks or slow devices as described in #11970

terrytangyuan:
https://github.com/argoproj/argo-workflows/pull/12133

terrytangyuan:
https://github.com/argoproj/argo-workflows/pull/12203

terrytangyuan:
TODO: @terrytangyuan to post a list of commits to cherry-pick here

terrytangyuan:
The plan is to cherry-pick the following commits (unless there's conflict)

Fixes:

- [ ] 9eadf2def fix: leak stream (#12193)
- [ ] 206c901c6 fix: Fix for missing steps in the UI (#12203)
- [ ] 939ce4029 fix(server): allow passing loglevels as env vars to Server (#12145)
- [ ] 222d53cdf fix: Resource version incorrectly overridden for wfInformer list requests. Fixes #11948 (#12133)
- [ ] f95a5fc6b fix: retry S3 on RequestError. Fixes #9914 (#12191)
- [ ] 6805c9132 fix: ArtifactGC Fails for Stopped Workflows. Fixes #11879 (#11947)
- [ ] 873bacbaf fix(ui): Cost Opt should only apply to live Workflows (#12170)

Chore(deps):

- [ ] 75d7eb190 chore(deps): bump github.com/aliyun/credentials-go from 1.3.1 to 1.3.2 (#12227)
[...]

PR #12714, opened 2024-02-29
fix(hack): various fixes & improvements to cherry-pick script

agilgur5:
### Motivation

As Isitha and I now have Approver permissions and were working on the release rotation per #12592, we both used the script and both found some issues, one of which was mentioned in https://github.com/argoproj/argo-workflows/issues/11997#issuecomment-1970241749

Fix a few bugs in the script and make it more ergonomic

#### bugfixes

- error out when the current branch that is checked out is not equivalent to the branch we're doing a cherry-pick on, as you can't cherry-pick onto other branches (see also this [SO question](https://stackoverflow.com/q/13878904/3431180))
- correctly order the commits in reverse chronological order -- the commits should be cherry-picked in the same order they were added
  - add a simple `tac` (reverse `cat`) to achieve this
- like the dry-run, commits that aren't going to be cherry-picked should be skipped
  - previously this would error out with no error message, which was _very_ confusing

#### misc improvements

- enable dry-run by default, for safety and simplicity (the arg is now optional)
- add flags to [`git cherry-pick`](https://git-scm.com/docs/git-cherry-pick)
  - use [`-Xpatience`](https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt--Xltoptiongt) to run the slower [`patience` diff strategy](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-algorithmpatienceminimalhistogrammyers) for hopefully less merge conflicts
  - use [`-x`](https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt--x) to add "(cherry picked from commit ...)"
    - this flag is very useful for posterity for backports as it creates a backlink to the original commit

- <details> <summary> other improvements: </summary>
  
  - remove redundant `set -eu` -- this only needs to be set once in the script
  - rename variables and functions to be more explicit etc
    - e.g. use conventional `get*` functions
  - reduce nesting by removing the `else`

</details>

### Verification

Run `./hack/cherry-pick.sh release-3.5 fix` in a few variations
- on a different branch
- with and without dry run

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -9,10 +9,10 @@ to work well with our versioned website.
 Then get a list of commits you may want to cherry-pick:
 
 ```bash
-./hack/cherry-pick.sh release-3.3 "fix" true
-./hack/cherry-pick.sh release-3.3 "chore(deps)" true
-./hack/cherry-pick.sh release-3.3 "build" true
-./hack/cherry-pick.sh release-3.3 "ci" true
+./hack/cherry-pick.sh release-3.3 "fix"
+./hack/cherry-pick.sh release-3.3 "chore(deps)"
+./hack/cherry-pick.sh release-3.3 "build"
+./hack/cherry-pick.sh release-3.3 "ci"
 ```
 
 To automatically cherry-pick, run the following:
--- a/hack/cherry-pick.sh
+++ b/hack/cherry-pick.sh
@@ -32,7 +32,7 @@ prs "$base" main
 diff "/tmp/$br" /tmp/main | grep "^> " | cut -c 3- > /tmp/prs
 
 # print all the commits that need cherry-picking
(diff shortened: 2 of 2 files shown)


terrytangyuan (approved):

agilgur5:
For reference, since this script is not on the `release-3.5` branch (or `release-3.4` or earlier for that matter), I copied the script to `/tmp`, checked out the release branch, and then ran the `/tmp` script:

```sh
cp hack/cherry-pick.sh /tmp/cherry-pick.sh
git checkout release-3.5
. /tmp/cherry-pick.sh
```

agilgur5:
Due to the above, I backported this to `release-3.5` as c956d70eead3cedf2f8c1422c028e26fe4b45683

PR #12722, opened 2024-03-01
feat: allow lastRetry variables in templates

### LINKED ISSUE #10364: Allow use of lastRetry in podSpecPatch resources request
# Summary

What change needs making?
This is related to https://github.com/argoproj/argo-workflows/issues/10362, and builds on the example there.
I would like to be able to able to specify something like the following:

|```
    retryStrategy:
      limit: "5"
      retryPolicy: "Always"
      expression: 'lastRetry.status == "Error" or (lastRetry.status == "Failed" and asInt(lastRetry.exitCode) not in [1,127])'
   podSpecPatch: '{"containers":[{"name":"main", "resources":{"requests":{"memory": "{{=(asInt(lastRetry.statusCode==137) ? (2 ** asInt(retries)) : 1) * 64}}Mi" }}}]}'
```
This currently throws the following error:

```
podSpecPatch: '{"containers":[{"name":"main", "resources":{"requests":{"memory": "{{=(asInt(lastRetry.statusCode==137) ? (2 ** asInt(retries)) : 1) * 64}}Mi" }}}]}'
```

I think this might be because the `lastRetry` might not currently be supported in the `podSpecPatch` ?

# Use Cases

When would you use this?
This would be used to automatically increase memory resource requests on retries, but only if the lastRetry failed due to memory problems.

---
<!-- Issue Author: Don't delete this message to encourage other users to support your issue! -->

### PR DESCRIPTION
**Pull request by eduardodbr**
Fixes #10364

### Verification

```yaml
kind: Workflow
metadata:
  name: pod-spec-patch
spec:
  entrypoint: whalesay
  templates:
  - name: whalesay
    retryStrategy:
      limit: 5
      retryPolicy: Always
    podSpecPatch: '{"containers":[{"name":"main", "resources":{"requests":{"memory": "{{=(sprig.int(lastRetry.exitCode)==1 ? int(retries) : 1)* 10}}Mi" }}}]}'
    container:
      image: python:alpine3.6
      command: ["python", -c]
      args: ["import sys; sys.exit(1)"]
```

uses the exit code to know when to increase memory. This should use 10MB,20MB,30MB,40MB,50MB in each iteration


workflow/controller/operator.go
```diff
@@ -2128,6 +2128,24 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
...
+			}
+			lastRetryStatus = string(lastChildNode.Phase)
+			lastRetryDuration = fmt.Sprint(lastChildNode.GetDuration().Seconds())
+			lastRetryMessage = lastChildNode.Message
+		}
+		localParams[common.LocalVarRetriesLastExitCode] = lastRetryExitCode
+		localParams[common.LocalVarRetriesLastDuration] = lastRetryDuration
+		localParams[common.LocalVarRetriesLastStatus] = lastRetryStatus
+		localParams[common.LocalVarRetriesLastMessage] = lastRetryMessage
+
```
juliev0, line 2148: Since this function is already so long, can we separate this into a function to build the local scope from `lastChildNode`?

workflow/controller/operator.go
```diff
@@ -2128,6 +2128,24 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 		// Inject the retryAttempt number
 		localParams[common.LocalVarRetries] = strconv.Itoa(retryNum)
```
juliev0, line 2129: Since we're already using `LocalVarRetries` in this scope, it does seem reasonable to me to use these other variables as well.

juliev0: Actually, I haven't looked at it long enough, but I am kind of wondering if there's a way to do this and not have to be setting these same variables twice. Should `processNodeRetries()` possibly take a `localScope` and then be able to modify it and return it to the caller? You have probably looked at the code longer than I have - what do you think?

gdhagger:
This PR seems to have stalled - would love to be able to use these vars, what needs to happen to get it moving again?

caelan-io:
Thank you for contributing this, @eduardodbr ! Isitha picked it up and rounded it out to merge the feature in https://github.com/argoproj/argo-workflows/pull/14450 and it's released in v3.7, so we'll close this PR out. cc @gdhagger and @EladProject for visibility.

PR #12734, opened 2024-03-04
refactor: change the logic of delete pod during retry. Fixes: #12538

### LINKED ISSUE #12538: Move `retry` Pod deletions out of Server and into Controller for proper separation of duties

shuangkun:
Fixes: #12538
Refactor the logic of deleting pods during retry to speed up retry a workflow.

### Motivation

Speed up retry a workflow. Let a large archived workflow (more than 8000 pods) can be successfully retryed within 1 minute.

### Modifications

As Anton and Joibel suggested, I moved the delete logic to the controller.
Add spec retry to workflow spec.
Adding spec is just one way, other good ways are also possible. For example, pass label to trigger retry. If you think it's better to pass the label, I will change it.

### Verification

e2e and units

reference:
https://github.com/argoproj/argo-workflows/pull/12624
https://github.com/argoproj/argo-workflows/pull/12419


workflow/util/util.go
```diff
@@ -827,6 +827,29 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow,
 	return newWF, resetParentGroupNodes
 }
 
+// RteryWorkflow retry a workflow by setting spec.retry and controller will retry it.
+func RetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string, parameters []string) (*wfv1.Workflow, error) {
```
tczhao, line 831: We could use a better name here, `retryworkflow` can be confusing,
The word on top of my head is `markWorkflowForRetry`, there must be a better term

workflow/util/util.go
```diff
@@ -827,6 +827,29 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow,
...
+	}
+
+	wf.Spec.Retry = &wfv1.RetryConfig{
+		Retried:           false,
+		RestartSuccessful: restartSuccessful,
+		NodeFieldSelector: nodeFieldSelector,
+		Parameters:        parameters,
+	}
+
+	delete(wf.Labels, common.LabelKeyCompleted)
```
tczhao, line 849: the delete label looks formulation of workflow, I think we can let `formulateRetryWorkflow` handle this instead.
so that server only 
- check client request error
- and update for retry

workflow/controller/operator.go
```diff
@@ -3832,6 +3843,44 @@ func (woc *wfOperationCtx) retryStrategy(tmpl *wfv1.Template) *wfv1.RetryStrateg
 	return woc.execWf.Spec.RetryStrategy
 }
 
+func (woc *wfOperationCtx) shouldRetry() bool {
+	retryConfig := woc.execWf.Spec.Retry
+	if retryConfig == nil {
+		return false
+	}
+	if woc.IsRetried() {
+		if woc.controller.podCleanupQueue.Len() == 0 {
```
tczhao, line 3852: what is the purpose of this line here?

tczhao (changes requested):

shuangkun:
Thank you for your reply. I would like to confirm a question. Do you think it is reasonable to pass the retry parameter through the spec? This will increase the amount of code changes. But at that time, I thought it was not appropriate to just use label to pass it, so I put it in the spec with reference to suspend.

workflow/controller/operator.go
shuangkun: I just want to confirm that all the pod has been cleaned up. If exists is not reported when creating a new pod at this time, this should be removed.

tczhao:
> Thank you for your reply. I would like to confirm a question. Do you think it is reasonable to pass the retry parameter through the spec? This will increase the amount of code changes. But at that time, I thought it was not appropriate to just use label to pass it, so I put it in the spec with reference to suspend.

Here's some context
[spec and status](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)
[annotation](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
[label](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)

Both `annotation` and `spec` make sense in their own way.
- annotation _Directives from the end-user to the implementations to modify behavior or engage non-standard features._
- The spec is _a complete description of the desired state, including configuration settings provided by the user_

I would vote for `spec` thinking in a way that, when we retry, we update the desired state of the workflow, and then the controller works on it towards the desired state

Hi @agilgur5, what would be your thoughts around this?

workflow/controller/operator.go
tczhao: Do you mean something like
make sure all pods deleted before process retry workflow

shuangkun: Yes, this is what I want.

tczhao: I think maybe we should check nodestatus instead.
Other workflows would queue up the podCleanupQueue causing issue

shuangkun:
Hi @agilgur5 , can you give me your vote so that I can better solve these comments? Thanks.

workflow/util/util.go
shuangkun: If delete it from here, reconceil may not be triggered. Do you think  change "reconciliationNeeded" function better than delete label here? If so, I will modify reconciliationNeeded function. 
```
func reconciliationNeeded(wf metav1.Object) bool {
	return wf.GetLabels()[common.LabelKeyCompleted] != "true" || slices.Contains(wf.GetFinalizers(), common.FinalizerArtifactGC)
}
```

workflow/util/util.go
shuangkun: Changed, Thanks!

workflow/util/util.go
```diff
@@ -827,6 +827,29 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow,
...
+func RetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string, parameters []string) (*wfv1.Workflow, error) {
+	switch wf.Status.Phase {
+	case wfv1.WorkflowFailed, wfv1.WorkflowError:
+	case wfv1.WorkflowSucceeded:
+		if !(restartSuccessful && len(nodeFieldSelector) > 0) {
+			return nil, errors.Errorf(errors.CodeBadRequest, "To retry a succeeded workflow, set the options restartSuccessful and nodeFieldSelector")
+		}
+	default:
+		return nil, errors.Errorf(errors.CodeBadRequest, "Cannot retry a workflow in phase %s", wf.Status.Phase)
+	}
```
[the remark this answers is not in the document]
shuangkun: Yes, removed.

workflow/util/util.go
```diff
@@ -827,22 +827,34 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow,
 	return newWF, resetParentGroupNodes
 }
 
-// FormulateRetryWorkflow formulates a previous workflow to be retried, deleting all failed steps as well as the onExit node (and children)
-func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string, parameters []string) (*wfv1.Workflow, []string, error) {
+// MarkWorkflowForRetry mark a workflow's spec.retry and controller will retry it.
+func MarkWorkflowForRetry(ctx context.Context, wf *wfv1.Workflow, restartSuccessful bool, nodeFieldSelector string, parameters []string) (*wfv1.Workflow, error) {
 	switch wf.Status.Phase {
 	case wfv1.WorkflowFailed, wfv1.WorkflowError:
 	case wfv1.WorkflowSucceeded:
```
sarabala1979, line 834: Why do user need to retry the succeeded workflow?

agilgur5 (commented):
> Adding spec is just one way, other good ways are also possible. For example, pass label to trigger retry. If you think it's better to pass the label, I will change it.

> Hi @agilgur5 , can you give me your vote so that I can better solve these comments? Thanks.

Hey sorry, I've had a bit too much on my plate recently. (can you tell by how late I've been up in EDT? 🙃)

So I mentioned using a label in #12538 as we pick up on various labels already in the Controller, so that would be consistent. It would also make it possible for a user to just add a label and not need to go through the Server or CLI (such as in https://github.com/argoproj/argo-workflows/discussions/12027#discussioncomment-7316157), which would also make for a consistent UX.
In https://github.com/argoproj/argo-workflows/pull/12624#pullrequestreview-1913123800, I also suggested _against_ spec changes as those are, well, part of the spec, and so become (much) harder to change.

In particular, per https://github.com/argoproj/argo-workflows/issues/6490#issuecomment-1961246329, we may want to create (optional?) CRs for imperative actions like retry and resubmit. Per https://github.com/argoproj/argo-workflows/issues/6490#issuecomment-1954915041, I also discussed that in SIG Security during two meetings, and they were supportive of that and apparently Argo CD had been thinking of implementing something similar to that.

With that in mind, in particular, I wouldn't want us to implement a spec that _will_ change that we then will have trouble to remove due to breaking spec changes.

> Both `annotation` and `spec` make sense in their own way.
> 
> * annotation _Directives from the end-user to the implementations to modify behavior or engage non-standard features._
> * The spec is _a complete description of the desired state, including configuration settings provided by the user_
> 
> I would vote for `spec` thinking in a way that, when we retry, we update the desired state of the workflow, and then the controller works on it towards the desired state
> 
> Hi @agilgur5, what are your thoughts on this?

Thanks for summarizing @tczhao.
So a retry is an _imperative action_, not a declarative state. So "desired state" is not really accurate to describe a retry -- it doesn't describe state at all, it describes an action. It also only affects the `status` and not the actual Workflow `spec`.
"non-standard feature" could be considered as a roughly accurate description of how retries currently work, as they are not supported via k8s directly, only via the Argo API. And as they are imperative, that is also non-standard in k8s.
A "directive from the end-user [...] to modify behavior"  could be considered indicative of an action and it does modify behavior; it tells the Workflow to reprocess some nodes.

Perhaps an example would be good -- you wouldn't specify a retry within your spec when you submit your Workflow initially, as it's something you do manually after the fact. Whereas you do specify a `retryStrategy` beforehand for automatic retries. Also `retryConfig` needing to be disambiguated even more from `retryStrategy` I think would add confusion (it's a bit confusing even now, but at least it is relatively self-evident that one is manual whereas the other is automatic).

They are not entirely mutually exclusive definitions though, there is certainly some overlap.

#### implementation details 

When I first looked at this implementation (it's been a month since then), my bigger question was actually, "can this even be properly implemented in a label or annotation?" as it's a primitive string -- with a limited length no less -- and not a more complex datatype.
`parameters` in particular seem a bit non-trivial to marshal.

Although retrying with new `parameters` actually seems to break the declarative state in general if I'm not mistaken -- and so may be an anti-pattern in and of itself. For instance, you could have some nodes with one set of parameters and others nodes with a different set. Meanwhile there is only one set of parameters in the actual Workflow `spec`. **EDIT**: This appears to have been mentioned in https://github.com/argoproj/argo-workflows/pull/9141#issuecomment-1184904953 as well

shuangkun:
@agilgur5  Thank you for your reply. I have modified a version. Can you help me take a look?

tczhao (changes requested):

tooptoop4:
🦗

shuangkun:
> 🦗

This PR needs some updates

PR #12736, opened 2024-03-04
feat: add sqlite-based memory store for live workflows. Fixes #12025

### LINKED ISSUE #12025: 3.5 ListWorkflows causes server to hang when there are lots of archived workflows

jiachengxu:
Fixes: https://github.com/argoproj/argo-workflows/issues/12025
Implementation of the proposal: https://github.com/argoproj/argo-workflows/issues/12025#issuecomment-1949330952

This PR implements a sqlite-based in-memory store that the workflow server will use to list live workflows, it provides:
- more unified query availability, including ordering & filtering by date, filtering by labels, archived-workflow-like pagination. (Note that `nameprefix` is not added in this PR).
- using List and Watch reflector so that workflow server won't tax Kubernetes per every user request.
- new pagination method gets rid of the merging of live and archived workflows logic, and it doesn't require merging both live and archived workflows every time.

For testing the change of this PR, I have created a image and pushed to quay.io: https://quay.io/repository/jiachengxu/argocli, and you can replace the `image` of the `argo-server` container in the `argo-server` Deployment with `quay.io/jiachengxu/argocli:fix-v6`


server/workflow/store/sqlite_store.go
```diff
@@ -0,0 +1,421 @@
+package store
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/upper/db/v4"
+	"github.com/upper/db/v4/adapter/sqlite"
```
jiachengxu, line 10: Using `"github.com/upper/db/v4/adapter/sqlite"` will include `github.com/mattn/go-sqlite3` as a dependency which requires cgo, so I have to set `CGO_ENABLED=1` and install `build-base` in the builder docker image for argocli, also I feel like this might break cross-platform compilation.

One way to avoid this is to use some cgo-free sqlite lib, for example, https://github.com/zombiezen/go-sqlite, but I am unsure if there is a better way of handling it.

Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN apk update && apk add --no-cache \
     curl \
     gcc \
     bash \
-    mailcap
+    mailcap \
+    build-base
```
ryancurrah, line 17: gcc is already installed. Build-base provides gcc, g++, make, musl-dev and binutils. You can remove gcc as it is duplicated by build-base or maybe you don't really need build-base if all you were after was gcc.

server/workflow/store/sqlite_store.go
ryancurrah: Theres also https://gitlab.com/cznic/sqlite if you want a `database/sql` driver.

Makefile
```diff
@@ -193,20 +193,20 @@ dist/argo-windows-%.gz: dist/argo-windows-%
 	gzip --force --keep dist/argo-windows-$*.exe
 
 dist/argo-windows-%: server/static/files.go $(CLI_PKGS) go.sum
-	CGO_ENABLED=0 $(GOARGS) go build -v -gcflags '${GCFLAGS}' -ldflags '${LDFLAGS} -extldflags -static' -o $@.exe ./cmd/argo
+	CGO_ENABLED=1 $(GOARGS) go build -v -gcflags '${GCFLAGS}' -ldflags '${LDFLAGS} -extldflags -static' -o $@.exe ./cmd/argo
```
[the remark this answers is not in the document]
agilgur5: Seconding this, would definitely prefer to avoid `CGO` usage, especially as this will likely be released as a patch to 3.5.x. Wouldn't want to add anything that is potentially breaking

server/workflow/store/sqlite_store.go
jiachengxu: I managed to implement the same by using the https://github.com/zombiezen/go-sqlite, and now it is cgo-free.

Dockerfile
jiachengxu: I reverted this back since now it is cgo-free (https://github.com/argoproj/argo-workflows/pull/12736#discussion_r1519971792)

Makefile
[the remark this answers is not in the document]
jiachengxu: I reverted this back since now it is cgo-free (https://github.com/argoproj/argo-workflows/pull/12736#discussion_r1519971792)

server/workflow/store/sqlite_store.go
agilgur5: > Theres also https://gitlab.com/cznic/sqlite if you want a `database/sql` driver.

> I managed to implement the same by using the https://github.com/zombiezen/go-sqlite

`zombiezen/go-sqlite` uses `cznic/sqlite` under the hood, for reference.

Also TIL about https://gitlab.com/cznic/ccgo, a C to Go compiler 😮 🤯

jiachengxu:
@agilgur5 I guess the idea that this PR is trying to implement is the direction we would like to go? Could I get some reviews for this PR from you or other community members?

terrytangyuan (commented):
We need more eyes on this and prioritize getting this merged soon.

server/workflow/store/sqlite_store.go
```diff
@@ -0,0 +1,306 @@
...
+  value varchar(63) not null,
+  primary key (uid, name, value),
+  foreign key (uid) references argo_workflows (uid) on delete cascade
+);
+create index if not exists idx_name_value on argo_workflows_labels (name, value);
+`
+	insertWorkflowQuery      = `insert into argo_workflows (uid, instanceid, name, namespace, phase, startedat, finishedat, workflow) values (?, ?, ?, ?, ?, ?, ?, ?)`
+	insertWorkflowLabelQuery = `insert into argo_workflows_labels (uid, name, value) values (?, ?, ?)`
+	deleteWorkflowQuery      = `delete from argo_workflows where uid = ?`
+)
```
jessesuen, line 46: 👍

persist/sqldb/selector.go
```diff
@@ -0,0 +1,89 @@
...
+		clauses = append(clauses, db.Raw("namespace = ?", options.Namespace))
+	}
+	if options.Name != "" {
+		clauses = append(clauses, db.Raw("name = ?", options.Name))
+	}
+	if options.NamePrefix != "" {
+		clauses = append(clauses, db.Raw("name like ?", options.NamePrefix+"%"))
+	}
+	if !options.MinStartedAt.IsZero() {
+		clauses = append(clauses, db.Raw("startedat > ?", options.MinStartedAt))
```
jessesuen, line 48: Does this need to be `>=`

persist/sqldb/selector.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if !options.MaxStartedAt.IsZero() {
+		clauses = append(clauses, db.Raw("startedat < ?", options.MaxStartedAt))
```
jessesuen, line 51: `<=` ?

jessesuen (commented):
LGTM, but wondering if we have an off-by-one bug?

persist/sqldb/selector.go
jiachengxu: originally I just used the same query as what we are currently doing for archived workflows: https://github.com/argoproj/argo-workflows/blob/main/persist/sqldb/workflow_archive.go#L254-L266, but I think you are right, this is a bug, and I fixed it in https://github.com/argoproj/argo-workflows/pull/12736/commits/d1f0ace1be8810651263b5d9f6fbe22b2ddd8b62 for both archived and live workflows.

persist/sqldb/selector.go
jiachengxu: the same as https://github.com/argoproj/argo-workflows/pull/12736#discussion_r1578743383

jessesuen (approved):
Great work!

terrytangyuan (approved):
Looks great! Thanks for the hardwork @jiachengxu!

sarabala1979 (approved):
LGTM

Joibel:
@jiachengxu unfortunately I've reverted this commit in #13018
The CI was not correctly testing the argocli image as part of the PR, and `ListWorkflows` in some of the tests using `argocli` is returning an empty list (@isubasinghe has demonstrated this).
It'll need fixing up and resubmitting on top of #13018 which should test the CLI properly from now on.

pkg/apiclient/argo-kube-client.go
agilgur5, line 1: so this file is the only one that affects the non-Server CLI commands, so I imagine the bug must be here somewhere

pkg/apiclient/argo-kube-client.go
```diff
@@ -84,13 +86,16 @@ func newArgoKubeClient(ctx context.Context, clientConfig clientcmd.ClientConfig,
 	if err != nil {
 		return nil, nil, err
 	}
-	return ctx, &argoKubeClient{instanceIDService}, nil
+	wfStore, err := store.NewSQLiteStore(instanceIDService)
+	if err != nil {
+		return nil, nil, err
+	}
```
agilgur5, line 92: in particular, this looks potentially off to me, the _client_ shouldn't need a store, only the Server should need that

agilgur5 (commented):
Follow-up to Alan's comments above and my own root cause analysis that led up to the revert, I suspect a very tiny piece of this PR was causing the CLI bug

pkg/apiclient/argo-kube-client.go
agilgur5: @jiachengxu confirmed this [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1715189885251799?thread_ts=1714810708.644859&cid=C0510EUH90V). We're still figuring out the appropriate fix for this though

jiachengxu: I managed to fix the issue with a bit refactoring and abstraction of the code, this commit fixes the issue: https://github.com/argoproj/argo-workflows/pull/13021/commits/14a11367a653d7c312b490f0e0526d8e481e250f
You are right about the issue, in the fix, the sqlite store is only used in server mode and not in cli mode, for cli mode, we still talk to kubernetes to get the live workflows.

agilgur5: > for cli mode, we still talk to kubernetes to get the live workflows.

Or the Argo Server instead of k8s in the CLI's `HTTP` mode

agilgur5: Resolved in #13021

agilgur5:
Cherry-picked into `release-3.5` as 50dc580ba0172427d15517dd7aa8454e35a25857
- Fixed several merge conflicts in the deps -- mostly removed changes in adjacent deps to be as minimal as needed for this PR (lot of the conflicts were from other dep upgrades on `main`, such as #12847 and #12860 etc)
- Fixed one small conflict in the imports in `workflow_server.go`

server/workflowarchive/archived_workflow_server.go
```diff
@@ -43,108 +40,42 @@ func NewWorkflowArchiveServer(wfArchive sqldb.WorkflowArchive, offloadNodeStatus
 }
 
 func (w *archivedWorkflowServer) ListArchivedWorkflows(ctx context.Context, req *workflowarchivepkg.ListArchivedWorkflowsRequest) (*wfv1.WorkflowList, error) {
-	options := req.ListOptions
-	namePrefix := req.NamePrefix
-	if options == nil {
-		options = &metav1.ListOptions{}
-	}
```
agilgur5, line 50: I believe this removal is causing #13804. to be fixed in #13807

PR #12730, opened 2024-03-01
test: fix operator metric tests

blkperl:
### Motivation

Tests were passing when https://github.com/argoproj/argo-workflows/commit/33ad82f689b83875a3b864298221ec3872379a4a was merged but new PRs are failing unit tests now. 

### Modifications

Changed the getMetricStringValue helper to call the String() function

### Verification

Unit tests will pass

PR #12702, opened 2024-02-26
chore(deps): bump prometheus dependencies

blkperl:
Signed-off-by: william.vanhevelingen <william.vanhevelingen@acquia.com>

fixes #12429, fixes #12272, fixes #11957

PR #12737, opened 2024-03-04
test: fix intermittent operator metrics test failure

blkperl:
### Motivation

Fixes intermittent unit test failures

### DIFF
--- a/workflow/controller/operator_metrics_test.go
+++ b/workflow/controller/operator_metrics_test.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


agilgur5:
Follow-up to #12702 and #12730, for reference

workflow/controller/operator_metrics_test.go
```diff
@@ -237,7 +238,11 @@ func getMetricStringValue(metric prometheus.Metric) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("%v", metricString.String()), nil
+
+	// Workaround for https://github.com/prometheus/client_model/issues/83
+	space := regexp.MustCompile(`\s+`)
```
agilgur5, line 243: small optimization: this can be moved outside of the function as it's a constant and this function is called multiple times

terrytangyuan (approved):
Thank you!

workflow/controller/operator_metrics_test.go
blkperl: Fixed

SuperQ:
Due to the intentionally breaking changes in upstream protobuf, I would recommend refactoring your tests to render to proto and use the proto compare tooling.

agilgur5 (approved):

agilgur5:
> Due to the intentionally breaking changes in upstream protobuf, I would recommend refactoring your tests to render to proto and use the proto compare tooling.

See the more detailed comment in the upstream issue: https://github.com/prometheus/client_model/issues/83#issuecomment-1978269460

I merged this for now as it was blocking CI intermittently, but indeed, that sounds like we may want to go with that route if it's more accurate. We'd have to analyze if it's worthwhile or if this workaround is "good enough" if we have to add more deps etc to do that

PR #12752, opened 2024-03-06
build(docs): add script to sync docs/README.md with README.md

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/12531#issuecomment-1901031142

### Motivation

These two READMEs are now almost entirely equivalent; instead of changing both (and remembering to) all the time we should keep them in sync with automation.

### Modifications

- wrote a small script `hack/copy-readme.sh` and added it to the `Makefile` and `ci-build.yaml`
  - can't quite just do a `cp` or a symlink because some of the URLs need modifications
    - so add a few replacements using `sed`

- also update `docs/README.md` to have the latest changes from README.md

### Verification

Ran the script to create the changes seen in this PR

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -99,6 +99,7 @@ jobs:
               # proposals live only on GH as pure markdown
               - '!docs/proposals/**'
               # docs scripts & tools from `make docs`
+              - hack/copy-readme.sh
               - hack/check-env-doc.sh
               - .markdownlint.yaml
               - .mlc_config.json
--- a/Makefile
+++ b/Makefile
@@ -724,6 +724,8 @@ docs: /usr/local/bin/mkdocs \
 	docs-lint \
 	# TODO: This is temporarily disabled to unblock merging PRs.
 	# docs-linkcheck
+	# copy README.md to docs/README.md
+	./hack/copy-readme.sh
 	# check environment-variables.md contains all variables mentioned in the code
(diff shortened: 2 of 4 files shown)


agilgur5:
Guh, docs CI is [failing](https://github.com/argoproj/argo-workflows/actions/runs/8178894030/job/22363787181?pr=12752) with:
```
sed: can't read s/(https:\/\/argo-workflows\.readthedocs\.io\/en\/latest\/\(.*\)\/)/(\1\.md)/: No such file or directory
```

I thought that might happen because macOS `sed` and GNU `sed` are different 😕 Ugh, shell portability is a pain... might rewrite this in Python...

agilgur5:
Was able to make it portable relatively quickly with a bit of refactoring in [this commit](https://github.com/argoproj/argo-workflows/pull/12752/commits/312b0c2fecc2b45fffbc2f31a458c48d334dc248), so will leave it in shell for now.
Isitha and I have discussed writing scripts in a full language for ease of use / review / writing, which I agree with, so maybe we'll convert all to Python (or Go or something) later. I also say that as someone who's had to write and edit a lot of shell scripts -- it's not fun 🙃

terrytangyuan (approved):

agilgur5:
Backported to `release-3.5` as d9f2184acac1fd390f68f36aefc59706885003d8
- Fixed merge conflicts in the `docs/README.md` -- automated the changes with the script from this PR too 😉

PR #12754, opened 2024-03-07
build: support local arm64 and amd64 image

### LINKED ISSUE #12750: local image build: support arm and amd64
# Summary

What change needs making?
Now， if we build controller image, the imgae Architecture will be same as your machine, it can not use on arm and amd64 at the same time. so we want to change.
## Use Cases

When would you use this?

---

heidongxianhua:
Fixes #12750


Joibel:
I'd like by default not to build both images. In #12750 I asked for:

> We'd accept a PR for formalizing this if you want to do it, so long as it only built for the "same as your machine" in the default case. You could set an environment variable to override this behavior.

Makefile
```diff
@@ -97,6 +97,7 @@ endif
 
 ALWAYS_OFFLOAD_NODE_STATUS := false
 POD_STATUS_CAPTURE_FINALIZER ?= true
+TARGETPLATFORM := linux/amd64
```
agilgur5, line 100: As Alan wrote twice now:
> so long as it only built for the "same as your machine" in the default case

This is not "same as your machine"

Also please properly separate words in capital case: 
```suggestion
TARGET_PLATFORM := linux/amd64
```

Makefile
```diff
@@ -243,6 +244,7 @@ argoexec-image:
 %-image:
 	[ ! -e dist/$* ] || mv dist/$* .
 	docker buildx build \
+		--platform $(TARGETPLATFORM) \
```
agilgur5, line 247:
```suggestion
		--platform $(TARGET_PLATFORM) \
```

same as above regarding separating words. otherwise, this line seems correct

agilgur5 (changes requested):

Joibel:
I would also like there to be a line in the documentation about this feature, documenting both single and multiple platform builds.

Joibel (approved):
Thanks.

I'd prefer an approach that didn't assume amd64 in non arm cases, but this is good enough for now and can iterate later.

Joibel:
@agilgur5, could you take another look please.

Makefile
```diff
@@ -97,7 +97,9 @@ endif
 
 ALWAYS_OFFLOAD_NODE_STATUS := false
 POD_STATUS_CAPTURE_FINALIZER ?= true
+TARGET_PLATFORM := $(shell [ `uname -m` = arm64 ] && echo linux/arm64 || echo linux/amd64)
```
agilgur5, line 100: The env vars are partly categorized -- this should be placed above after `DOCKER_PUSH` or `DEV_IMAGE`

Makefile
```diff
@@ -97,7 +97,9 @@ endif
...
 
+$(info TARGET_PLATFORM=$(TARGET_PLATFORM))
 $(info GIT_COMMIT=$(GIT_COMMIT) GIT_BRANCH=$(GIT_BRANCH) GIT_TAG=$(GIT_TAG) GIT_TREE_STATE=$(GIT_TREE_STATE) RELEASE_TAG=$(RELEASE_TAG) DEV_BRANCH=$(DEV_BRANCH) VERSION=$(VERSION))
 $(info KUBECTX=$(KUBECTX) DOCKER_DESKTOP=$(DOCKER_DESKTOP) K3D=$(K3D) DOCKER_PUSH=$(DOCKER_PUSH))
```
agilgur5, line 104: these are also partly categorized:
```suggestion
$(info GIT_COMMIT=$(GIT_COMMIT) GIT_BRANCH=$(GIT_BRANCH) GIT_TAG=$(GIT_TAG) GIT_TREE_STATE=$(GIT_TREE_STATE) RELEASE_TAG=$(RELEASE_TAG) DEV_BRANCH=$(DEV_BRANCH) VERSION=$(VERSION))
$(info KUBECTX=$(KUBECTX) DOCKER_DESKTOP=$(DOCKER_DESKTOP) K3D=$(K3D) DOCKER_PUSH=$(DOCKER_PUSH) TARGET_PLATFORM=$(TARGET_PLATFORM))
```

docs/running-locally.md
```diff
@@ -101,6 +101,17 @@ If you made changes to the executor, you need to build the image:
 make argoexec-image
 ```
 
+Or you can use parameters to compile images that support specific platforms or multiple platforms with `TARGET_PLATFORM`:
+
```
agilgur5, line 105: These are environment variables, not parameters
```suggestion
You can use the `TARGET_PLATFORM` environment variable to compile images for specific platforms:
```
- can also simplify this, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

docs/running-locally.md
```diff
@@ -101,6 +101,17 @@ If you made changes to the executor, you need to build the image:
...
+```bash
+# only support arm64
+make argoexec-image TARGET_PLATFORM=linux/arm64
+# only support amd64
+make argoexec-image TARGET_PLATFORM=linux/amd64
+# support arm64 and amd64
```
agilgur5, line 111:
```suggestion
# compile for both arm64 and amd64
```

- "compile", not "support"
- I think the last example is sufficient for all purposes, as it should be intuitive that you can remove the comma if you only need one, IMO. unless you think otherwise @Joibel

agilgur5 (commented):
Some small adjustments in-line below

> I'd prefer an approach that didn't assume amd64 in non arm cases, but this is good enough for now and can iterate later.

Agreed

docs/running-locally.md
heidongxianhua: yeah, thanks for your comment, i have fix it, could you have a look again ? @agilgur5

agilgur5: For future reference, you can use the "re-request a review" button on GH, then it also pops up in my notifications as a review request specifically

agilgur5 (approved):
LGTM

Issue #12733, opened 2024-03-04
v3.5.5: Workflow does not complete even though Pod completed


panicboat:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

Workflow is submitted and waits for a while, but does not finish.
Any Workflow looks the same.
When I check the status of the pod, it says Completed and the logs appear to be complete.
Also, even though the pod does not appear to be running in the UI, the logs can still be retrieved in the UI.
It appears that we can no longer track the status of the Pod.

![image](https://github.com/argoproj/argo-workflows/assets/4182311/666f6bff-d9af-44bf-ac77-61b937c5b801)
```console
01054855@CF0313 minikube % k get po -n develop-workflow                               
NAME                                                                READY   STATUS      RESTARTS   AGE
p3-deletion-policy-mask-or-delete-user-data-66xf6-main-2457154088   0/2     Completed   0          6h12m
shinkobetsu-check-expired-enrollments-1709510400-main-933938806     0/2     Completed   0          7h45m
wonderful-rhino-argosay-67614653                                    0/2     Completed   0          2m23s
01054855@CF0313 minikube % k logs -n develop-workflow wonderful-rhino-argosay-67614653
time="2024-03-04T07:43:36.342Z" level=info msg="capturing logs" argo=true
hello argo
time="2024-03-04T07:43:37.343Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

### Version

v3.5.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
metadata:
  name: wonderful-rhino
  namespace: develop-workflow
  labels:
    example: 'true'
spec:
  arguments:
    parameters:
      - name: message
        value: hello argo
  entrypoint: argosay
  templates:
    - name: argosay
      inputs:
        parameters:
          - name: message
            value: '{{workflow.parameters.message}}'
      container:
        name: main
        image: argoproj/argosay:v2
        command:
          - /argosay
        args:
          - echo
          - '{{inputs.parameters.message}}'
  ttlStrategy:
    secondsAfterCompletion: 300
  podGC:
[...]

PR #12755, opened 2024-03-07
fix: mark task result completed use nodeId instead of podname. Fixes: #12733

shuangkun:
Fixes: https://github.com/argoproj/argo-workflows/issues/12733
mark task result completed use nodeId instead of podname


agilgur5:
> Fixes: #12733

For context, see the root cause analysis in https://github.com/argoproj/argo-workflows/issues/12733#issuecomment-1981895649 and above. A merge conflict in 3.5.5 was incorrectly resolved; this PR fixes that to the correct resolution (note that this PR is to the `release-3.5` branch specifically)

agilgur5 (approved):
Thanks for tracking this down and fixing it!

As this is a backport PR, going to merge this in cleanly as-is, but a regression test for this would be good to have per https://github.com/argoproj/argo-workflows/issues/12733#issuecomment-1992732678

agilgur5:
Yea it'll be released with 3.5.6 -- the `release-3.5` branch is where all 3.5.x releases tag off. Also was briefly discussed [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1709720628082599?thread_ts=1709512568.671069&cid=C0510EUH90V)

Issue #12329, opened 2023-12-07
Workflow stuck in running state when using activeDeadlineSeconds on template level


n1klasD:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

When `activeDeadlineSeconds` is set on the template level of a workflow as described [here](https://argoproj.github.io/argo-workflows/walk-through/timeouts/), it does not transition to the `failed` state after the specified time. The workflow is stuck in the `running` state and it´s pods are not cleaned up. This seems to happen, when the deadline of `activeDeadlineSeconds` is exceeded while the workflow pods are still in `pending` state. To reproduce this using the provided workflow, you might need to adjust `activeDeadlineSeconds`, so that the deadline is exceeded while the fan-out steps are still in `pending` state.

The provided workflow stays in the running state, after the deadline is exceeded (it was manually stopped after two hours). 
The fan-out steps stay in the `Pending` state, but display "Pod was active on the node longer than the specified deadline":

```
Name:                active-deadline-fanout-template-level-pprjp
Namespace:           XXXX
ServiceAccount:      unset
Status:              Running
Conditions:
 PodRunning          False
Created:             Thu Dec 07 11:44:48 +0100 (2 hours ago)
Started:             Thu Dec 07 11:44:48 +0100 (2 hours ago)
Duration:            2 hours 54 minutes
Progress:            0/4

STEP                                            TEMPLATE    PODNAME                                                      DURATION  MESSAGE
 ● active-deadline-fanout-template-level-pprjp  entrypoint

 └─┬─◷ fanout(0:1)                              echo        active-deadline-fanout-template-level-pprjp-echo-628187050   2h        Pod was active on the node longer than the specified deadline
   ├─◷ fanout(1:2)                              echo        active-deadline-fanout-template-level-pprjp-echo-2925824112  2h        Pod was active on the node longer than the specified deadline
   ├─◷ fanout(2:3)                              echo        active-deadline-fanout-template-level-pprjp-echo-2656767666  2h        Pod was active on the node longer than the specified deadline
   └─◷ fanout(3:4)                              echo        active-deadline-fanout-template-level-pprjp-echo-606271796   2h        Pod was active on the node longer than the specified deadline
```

[...]

PR #12761, opened 2024-03-07
fix: workflow stuck in running state when using activeDeadlineSeconds on template level. Fixes: #12329

shuangkun:
Fixes: #12329   
Root cause is when pod set faied from pending, because wait container's Terminated is nil, So this node is reset to pending and causing workflow stuck.

### Motivation

Let workflow Failed when activeDeadlineSeconds exceed.

### Modifications

Set pod Failed when pod set Failed from Pending. (Let workflow failed)

### Verification

Local test and add some unit test and e2e

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1447,7 +1447,7 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is still Pending) because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
-		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Waiting == nil && new.Phase.Completed()) && !initContainerFailed {
+		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Running != nil && new.Phase.Completed()) && !initContainerFailed {
 			woc.log.WithField("new.phase", new.Phase).Info("leaving phase un-changed: wait container is not yet terminated ")
 			new.Phase = old.Phase
 		}
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1441,7 +1441,7 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless wait contain is still Pending) because it may be busy saving outputs, and these
+	// We cannot fail the node if the wait container is still running because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
 		if c.Name == common.WaitContainerName && c.State.Running != nil && new.Phase.Completed() {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1433,21 +1433,18 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		new.Outputs.ExitCode = pointer.StringPtr(fmt.Sprint(*exitCode))
 	}
 
-	// If the init container failed, we should mark the node as failed.
-	var initContainerFailed bool
 	for _, c := range pod.Status.InitContainerStatuses {
 		if c.State.Terminated != nil && int(c.State.Terminated.ExitCode) != 0 {
 			new.Phase = wfv1.NodeFailed
-			initContainerFailed = true
 			woc.log.WithField("new.phase", new.Phase).Info("marking node as failed since init container has non-zero exit code")
 			break
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is still Pending) because it may be busy saving outputs, and these
+	// We cannot fail the node until the wait container is finished (unless wait contain is still Pending) because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
-		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Running != nil && new.Phase.Completed()) && !initContainerFailed {
+		if c.Name == common.WaitContainerName && c.State.Running != nil && new.Phase.Completed() {
 			woc.log.WithField("new.phase", new.Phase).Info("leaving phase un-changed: wait container is not yet terminated ")
 			new.Phase = old.Phase
 		}
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1444,7 +1444,7 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is not Pending) because it may be busy saving outputs, and these
+	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is still Pending) because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
 		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Waiting == nil && new.Phase.Completed()) && !initContainerFailed {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1444,10 +1444,10 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless any init container has failed) because it may be busy saving outputs, and these
+	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is not Pending) because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
-		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && new.Phase.Completed()) && !initContainerFailed {
+		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Waiting == nil && new.Phase.Completed()) && !initContainerFailed {
 			woc.log.WithField("new.phase", new.Phase).Info("leaving phase un-changed: wait container is not yet terminated ")
 			new.Phase = old.Phase
 		}
--- a/test/e2e/workflow_test.go
+++ b/test/e2e/workflow_test.go
@@ -4,15 +4,19 @@
 package e2e
 
 import (
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
+	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v3/test/e2e/fixtures"
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
 )
 
 type WorkflowSuite struct {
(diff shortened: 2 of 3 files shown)


juliev0:
I've started looking at this, so I assigned it to myself, but if anyone feels they know the code well here and prefers to reassign it to themselves, I'm good. :)

juliev0:
Can you help me understand the sequence of events that led to the [Issue](https://github.com/argoproj/argo-workflows/issues/12329)? So, the containers were in "Pending" state and then the Pod's activeDeadlineSeconds was reached. I see at that point from the documentation that Kubernetes "will actively try to mark it failed and kill associated containers". Do we understand what's preventing Kubernetes from marking it Terminated at that point?

shuangkun:
Kubernetes has marked it as Failed, but Argo observed Failed when recording but still reset it from Faild to Pending because its wait container teminting is nil.

shuangkun:
I think it’s because the waiting container has never been running, so it cannot receive termination.

shuangkun:
https://github.com/kubernetes/kubernetes/issues/108733

https://github.com/kubernetes/kubernetes/pull/108366

juliev0:
> [kubernetes/kubernetes#108733](https://github.com/kubernetes/kubernetes/issues/108733)
> 
> [kubernetes/kubernetes#108366](https://github.com/kubernetes/kubernetes/pull/108366)

Thanks for the links. Okay, I see from the first one that it links to [this](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-terminated) official documentation which says "A container in the Terminated state began execution and then either ran to completion or failed for some reason." So, you're right that if it didn't begin execution, it seems it should just stay Pending I guess.

juliev0:
Part of me is wondering what Kubernetes will do in this case - it can't just leave it in the Pending state, can it?

In any case, though, maybe your change is reasonable - if the purpose of that code is to make sure we save off any outputs from the wait container, but we hit the activeDeadlineSeconds (or terminated for some other reason), then maybe we don't want to allow the container to get into the Running state in the first place I suppose.

workflow/controller/operator.go
```diff
@@ -1444,10 +1444,10 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless any init container has failed) because it may be busy saving outputs, and these
+	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is still Pending) because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
-		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && new.Phase.Completed()) && !initContainerFailed {
+		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Waiting == nil && new.Phase.Completed()) && !initContainerFailed {
```
juliev0, line 1450: this could probably be simplified to change `c.State.Terminated == nil && c.State.Waiting == nil` to `c.State.Running != nil`, right?

juliev0:
> Part of me is wondering what Kubernetes will do in this case - should it just leave it in the Pending state?
> 
> In any case, though, maybe your change is reasonable - if the purpose of that code is to make sure we save off any outputs from the wait container, but we hit the activeDeadlineSeconds (or terminated for some other reason), then maybe we don't want to allow the container to get into the Running state in the first place I suppose.

Although, I guess this code doesn't prevent it from getting into a Running state, does it? It just presumes that it won't get into a Running state, right? (and if it did, presumably we won't wait to save the outputs)

juliev0:
Can you help point me to where in the code after this point that this `node.Phase` can enable vs prevent workflow shutdown?

shuangkun:
> Can you help point me to where in the code after this point that this `node.Phase` can enable vs prevent workflow shutdown?

If the pod Failed, the kubernetes wont't update it. 
The containerstatus is always:
```
  - image: quay.io/argoproj/argoexec:latest
    imageID: ""
    lastState: {}
    name: wait
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        reason: PodInitializing
  hostIP: 172.18.0.2
  message: Pod was active on the node longer than the specified deadline
```
So if argo did't update node phase, the node is always show pending for workflow, and the workflow stuck.

workflow/controller/operator.go
shuangkun: I think this is okay

shuangkun:
> Although, I guess this code doesn't prevent it from getting into a Running state, does it? It just presumes that it won't get into a Running state, right? (and if it did, presumably we won't wait to save the outputs)



> > Part of me is wondering what Kubernetes will do in this case - should it just leave it in the Pending state?
> > In any case, though, maybe your change is reasonable - if the purpose of that code is to make sure we save off any outputs from the wait container, but we hit the activeDeadlineSeconds (or terminated for some other reason), then maybe we don't want to allow the container to get into the Running state in the first place I suppose.
> 
> Although, I guess this code doesn't prevent it from getting into a Running state, does it? It just presumes that it won't get into a Running state, right? (and if it did, presumably we won't wait to save the outputs)

Yes， it does not prevent pod Running.

workflow/controller/operator.go
```diff
@@ -1445,10 +1445,10 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless any init container has failed) because it may be busy saving outputs, and these
+	// We cannot fail the node until the wait container is finished (unless any init container has failed or wait contain is still Pending) because it may be busy saving outputs, and these
 	// would not get captured successfully.
 	for _, c := range pod.Status.ContainerStatuses {
-		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && new.Phase.Completed()) && !initContainerFailed {
+		if (c.Name == common.WaitContainerName && c.State.Terminated == nil && c.State.Running != nil && new.Phase.Completed()) && !initContainerFailed {
```
juliev0, line 1451: So, I was actually saying we can use `c.State.Running != nil` and thus remove `c.State.Terminated == nil` per this description of `ContainerState`:

```
// ContainerState holds a possible state of container.
// Only one of its members may be specified.
// If none of them is specified, the default one is ContainerStateWaiting.
type ContainerState struct {
	// Details about a waiting container
	// +optional
	Waiting *ContainerStateWaiting `json:"waiting,omitempty" protobuf:"bytes,1,opt,name=waiting"`
	// Details about a running container
	// +optional
	Running *ContainerStateRunning `json:"running,omitempty" protobuf:"bytes,2,opt,name=running"`
	// Details about a terminated container
	// +optional
	Terminated *ContainerStateTerminated `json:"terminated,omitempty" protobuf:"bytes,3,opt,name=terminated"`
}
```

The comment says "Only one of its members may be specified."

juliev0: Also, I'm thinking about whether we really need `!initContainerFailed`. If the wait container is running then the Init Container must not have failed, would it? (It may be worth trying to understand why that was put in there)

shuangkun: Good catch, according to this description we may indeed only need "Running != nil".

juliev0:
Okay, so the code being changed isn't particular to failed pods, but for any pod for which `new.Phase.Completed()==true`. So, we should make sure this will be okay for any of those cases. Have you verified that?

If I try to look at those myself, I believe:

- [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1318) is the case of all containers having Succeeded, marking the Pod as Succeeded
- [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1325) is Phase.Failed which Kubernetes says means "All containers in the Pod have terminated, and at least one container has terminated in failure. That is, the container either exited with non-zero status or was terminated by the system."
- [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1356) one wouldn't actually happen
- [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1409) one seems like some json unmarshalling error, which I don't think would happen
- [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1441) one is the InitContainer failing

If I look at the original bug, I see the InitContainer succeeded, and the wait and main containers are in Pending. So, how did we get to new.Phase.Completed?

shuangkun:
> Okay, so the code being changed isn't particular to failed pods, but for any pod for which `new.Phase.Completed()==true`. So, we should make sure this will be okay for any of those cases. Have you verified that?
> 
> If I try to look at those myself, I believe:
> 
> * [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1318) is the case of all containers having Succeeded, marking the Pod as Succeeded
> * [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1325) is Phase.Failed which Kubernetes says means "All containers in the Pod have terminated, and at least one container has terminated in failure. That is, the container either exited with non-zero status or was terminated by the system."

There seems to be something wrong with this description.

> * [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1356) one wouldn't actually happen
> * [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1409) one seems like some json unmarshalling error, which I don't think would happen
> * [this](https://github.com/argoproj/argo-workflows/pull/12761/files#diff-f321d4af83fcf8311dc80c0d50c59ac4c240f761206e7bb652709870eb9feb43R1441) one is the InitContainer failing
> 
> If I look at the original bug, I see the InitContainer succeeded, and the wait and main containers are in Pending. So, how did we get to new.Phase.Completed?

The Pod from yesterday is still in this state now.  K8s version is v1.25.3, Argo is latest.
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2024-03-13T05:34:26Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2024-03-13T05:34:20Z"
    reason: PodFailed
    status: "False"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2024-03-13T05:34:20Z"
    reason: PodFailed
    status: "False"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2024-03-13T05:34:20Z"
    status: "True"
    type: PodScheduled
  containerStatuses:
  - image: alpine:latest
    imageID: ""
    lastState: {}
    name: main
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        reason: PodInitializing
  - image: quay.io/argoproj/argoexec:latest
    imageID: ""
    lastState: {}
    name: wait
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        reason: PodInitializing
  hostIP: 172.18.0.2
  initContainerStatuses:
  - containerID: containerd://66eb4112a8061ff6da5f48b0c15b9c5c5ccf70f8a4ffa079d258b4315b4bbf74
    image: quay.io/argoproj/argoexec:latest
    imageID: quay.io/argoproj/argoexec@sha256:61c8e55d00437565a2823c802cece0b3323aed758e870a5055768337a87d3546
    lastState: {}
    name: init
    ready: true
    restartCount: 0
    state:
      terminated:
        containerID: containerd://66eb4112a8061ff6da5f48b0c15b9c5c5ccf70f8a4ffa079d258b4315b4bbf74
        exitCode: 0
        finishedAt: "2024-03-13T05:34:25Z"
        reason: Completed
        startedAt: "2024-03-13T05:34:21Z"
  message: Pod was active on the node longer than the specified deadline
  phase: Failed
  podIP: 10.244.0.123
  podIPs:
  - ip: 10.244.0.123
  qosClass: Burstable
  reason: DeadlineExceeded
  startTime: "2024-03-13T05:34:20Z"
```

juliev0:
> 
> There seems to be something wrong with this description.
> 

Interesting. Thanks for sending that. Okay, so the Pod is clearly in a failed state. 

So, going through my cases above of Phase==Complete:

- If the Pod is Failed, we'll hold off if the wait container is in Running.
- If the Pod is Succeeded, presumably all of the containers succeeded, so we won't ever hold off.
- If the InitContainer failed (and perhaps Pod is not necessarily considered Failed itself), we'll hold off if the wait container is for some reason Running. 

We won't hold off any longer if for some reason the wait container is in Pending. I can't really think of a case where this would be a problem. I will probably approve this - may just want to take another look later with fresh eyes.

workflow/controller/operator.go
```diff
@@ -1433,21 +1433,18 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
...
 		if c.State.Terminated != nil && int(c.State.Terminated.ExitCode) != 0 {
 			new.Phase = wfv1.NodeFailed
-			initContainerFailed = true
 			woc.log.WithField("new.phase", new.Phase).Info("marking node as failed since init container has non-zero exit code")
 			break
 		}
 	}
 
-	// We cannot fail the node until the wait container is finished (unless any init container has failed) because it may be busy saving outputs, and these
+	// We cannot fail the node until the wait container is finished (unless wait contain is still Pending) because it may be busy saving outputs, and these
```
juliev0, line 1444:
```suggestion
	// We cannot fail the node if the wait container is still running because it may be busy saving outputs, and these
```

juliev0 (approved):

shuangkun:
Thanks！

agilgur5:
Backported cleanly to `release-3.5` as 19a7edebb

PR #12762, opened 2024-03-07
feat: add pendingTimeout for non-deadline timeout  Fixes #10341

### LINKED ISSUE #10341: Add a `pendingTimeout` parameter
# Summary

See https://github.com/argoproj/argo-workflows/issues/3572 for context.

Some of our workflows fails to schedule a k8s node because sometimes there are errors in the configuration that is responsible to execute a workflow.

The currently available option `activeDeadlineSeconds` considers both the `pending` phase and also the `execution` phase. We would need an option that only consider the `pending` phase so our failing pending workflow would be marked as failed after xxx seconds.

This new option could be `pendingTimeoutSeconds` or `pendintDeadlineSeconds`.

---

drawlerr:
Fixes #10341

### Motivation

Current Template timeout parameter can handle nodes spending too long in pending phase, but if the node happens to progress to running this timeout parameter transfers to activeDeadlineSeconds, leaving no way to set a timeout that covers _just_ the pending phase.

### Modifications

I added a new `pendingTimeout` field to Template that should only apply while a node is in `pending` phase.  Existing logic for timeout was re-used and expanded, and requeueAfter was added to help ensure more timely deadline enforcement.

### Verification

Added TestCheckTemplateTimeouts to operator_test, performed manual verifications with a workflow with sleeping + unschedulable pods with combinations of pendingTimeout and timeout

## Summary by CodeRabbit

* **New Features**
  * Added `pendingTimeout` for templates to cap how long a node’s pod may remain in `Pending` before the workflow fails.
  * Enforced separately from the overall `timeout`, based on controller-observed pod state.

* **Bug Fixes**
  * Controller now marks nodes failed on the pending deadline and attempts to delete the timed-out pending pod to free resources.
  * Validation prevents unsupported `pendingTimeout` usage (e.g., Step/DAG or other non-leaf templates).

* **Documentation**
  * Updated CRD/API schemas and executor/field docs to describe `pendingTimeout` semantics and limits.

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2241,7 +2241,9 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 		if node.Type == wfv1.NodeTypePod {
 			// delete the timed-out pod so the resources it was waiting for are freed.
 			// Deletion is by UID so a pod recreated by a retry cannot be affected.
-			if pod, exists, podErr := woc.podExists(node.ID); podErr == nil && exists {
+			if pod, exists, podErr := woc.podExists(node.ID); podErr != nil {
+				woc.log.WithError(podErr).Warn(ctx, "failed to check pod existence while cleaning up timed-out node")
+			} else if exists {
 				woc.controller.PodController.DeletePodByUID(ctx, pod.Namespace, pod.Name, string(pod.UID))
 			}
 		}
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2232,15 +2232,25 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 		return node, ErrDeadlineExceeded
 	}
 
-	// Check the template deadline for Pending nodes
-	// This check will cover the resource forbidden, synchronization scenario,
-	// In above scenario, only Node will be created in pending state
-	deadline, _, err := woc.checkTemplateTimeouts(processedTmpl, node)
+	// Check the timeout and pendingTimeout deadlines for Pending nodes.
+	// This also covers the resource-forbidden and synchronization scenarios,
+	// where only the node exists, in a pending state, with no pod created.
+	deadline, pendingDeadline, err := woc.checkTemplateTimeouts(processedTmpl, node)
 	if err != nil {
 		woc.log.WithField("template", processedTmpl.Name).Warn(ctx, "Template exceeded its deadline")
+		if node.Type == wfv1.NodeTypePod {
+			// delete the timed-out pod so the resources it was waiting for are freed.
+			// Deletion is by UID so a pod recreated by a retry cannot be affected.
+			if pod, exists, podErr := woc.podExists(node.ID); podErr == nil && exists {
+				woc.controller.PodController.DeletePodByUID(ctx, pod.Namespace, pod.Name, string(pod.UID))
+			}
+		}
 		return woc.markNodePhase(ctx, nodeName, wfv1.NodeFailed, err.Error()), err
 	}
-	// Ensure that we will check again soon after the deadline
+	// Ensure that we will check again soon after the earliest deadline
+	if deadline == nil || (pendingDeadline != nil && pendingDeadline.Before(*deadline)) {
+		deadline = pendingDeadline
+	}
 	if deadline != nil && time.Now().Before(*deadline) {
 		woc.requeueAfter(time.Until(*deadline))
 	}
@@ -2601,7 +2611,7 @@ func (woc *wfOperationCtx) handleNodeFulfilled(ctx context.Context, nodeName str
 }
 
 func getTimeoutAsDeadline(startedAt *time.Time, timeoutVal string) (*time.Time, error) {
-	tmplTimeout, err := time.ParseDuration(timeoutVal)
+	tmplTimeout, err := wfv1.ParseStringToDuration(timeoutVal)
 	if err != nil {
 		return nil, fmt.Errorf("invalid timeout format %w", err)
 	}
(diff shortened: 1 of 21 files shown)


workflow/controller/workflowpod.go
```diff
@@ -503,13 +508,18 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
...
 	}
 
-	if templateDeadline != nil && (pod.Spec.ActiveDeadlineSeconds == nil || time.Since(*templateDeadline).Seconds() < float64(*pod.Spec.ActiveDeadlineSeconds)) {
-		newActiveDeadlineSeconds := int64(time.Until(*templateDeadline).Seconds())
-		if newActiveDeadlineSeconds <= 1 {
-			return nil, fmt.Errorf("%s exceeded its deadline", nodeName)
+	if templateDeadline != nil {
+		if !pendingOnly && (pod.Spec.ActiveDeadlineSeconds == nil || time.Since(*templateDeadline).Seconds() < float64(*pod.Spec.ActiveDeadlineSeconds)) {
+			newActiveDeadlineSeconds := int64(time.Until(*templateDeadline).Seconds())
+			if newActiveDeadlineSeconds <= 1 {
```
Joibel, line 514: It seems slightly odd that we cut off with 1 second remaining on the deadline - any reason this is not zero?

Joibel:
If my workflow fails because of `pendingTimeout`, the workflow is marked as failed, but the pod is not deleted. This is surprising, is that what you see @drawlerr 

I merged in main for #13214, but that's not helped.

tooptoop4:
🚢

workflow/controller/workflowpod.go
drawlerr: This is from main but re-indented, presumably because there is a minimum valid activeDeadlineSeconds?

drawlerr:
@Joibel Sorry for the long hibernation, I've refreshed with main.

re: failed workflow vs pod deletion, I think "failed" is generally actually proper since we already have podGc config to handle pods in that case.

Joibel:
### Claude Code review

Found 7 issues:

1. The new `pendingTimeout` field has no parity validation — the existing `Timeout` field at `workflow/validate/validate.go` rejects non-leaf templates and bare-integer durations, but neither check was added for `pendingTimeout`. The field's own godoc says "This duration may not be applied to Step or DAG templates", but nothing enforces it; bad input only fails at runtime in `time.ParseDuration` once a pod goes pending.

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/validate/validate.go#L510-L521

2. The `Timeout` branch of the new `checkTemplateTimeouts` silently drops the `node.Phase == NodePending` guard that the original `checkTemplateTimeout` had. Previously a Running pod with an exceeded `Timeout` was killed by kubelet via `activeDeadlineSeconds`; now the controller will also call `markNodePhase(NodeFailed)` for Running nodes, racing with kubelet termination. This is a behavior change for the existing `Timeout` field that isn't called out anywhere in the PR.

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/controller/operator.go#L2577-L2588

3. When `pendingTimeout` fires, the workflow node is marked Failed but the pending pod is not deleted (Joibel raised this in review). Unlike `Timeout`, `pendingTimeout` deliberately does not set `pod.Spec.ActiveDeadlineSeconds`, so kubelet won't reap the pod either. Deferring to PodGC is optional — if a user's PodGC isn't configured to clean these up, the unschedulable pod persists indefinitely and consumes namespace quota, defeating the feature's purpose (freeing resources for unschedulable pods).

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/controller/operator.go#L2185-L2191

4. When both `pendingTimeout` and `timeout` are set and the node is Pending, the `pendingTimeout` block returns early and the `Timeout` branch is never evaluated. If `pendingTimeout: 60s, timeout: 30s`, a node pending for 35s won't be marked Failed until the 60s pending deadline fires — `Timeout`'s "covers entire node lifecycle" semantics are violated. The function should evaluate both and return the earlier deadline.

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/controller/operator.go#L2565-L2588

5. The new `checkTemplateTimeouts` call inside the `if exists` early-return branch of `createWorkflowPod` is redundant — `executeTemplate` at `operator.go:2186` already invoked the same check moments earlier. The duplicate call discards the deadline (no `requeueAfter`) and returns a bare `err` without `markNodePhase`, so error handling is inconsistent with the upstream check. Removing it would also avoid the need to hoist `GetNodeByName` to line 125 (and the consequent `node` → `progressNode` rename downstream).

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/controller/workflowpod.go#L129-L138

6. The pre-existing comment "Check the template deadline for Pending nodes / This check will cover the resource forbidden, synchronization scenario, / In above scenario, only Node will be created in pending state" no longer matches what `checkTemplateTimeouts` does — after this PR it also returns `ErrTimeout` for non-pending nodes via the `Timeout` branch (see issue 2). The comment should be updated or removed.

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/controller/operator.go#L2181-L2189

7. `getTimeoutAsDeadline` calls `time.ParseDuration` directly, but the project already has `wfv1.ParseStringToDuration` (used by retry backoff, suspend duration, podGC delay, etc.) which additionally accepts bare integers as seconds. As written, `pendingTimeout: 30` would error here while `retryStrategy.backoff.duration: 30` works elsewhere — a third duration-parsing convention. (Original `Timeout` had the same gap, so this is consistency rather than a regression, but worth fixing while the helper is fresh.)

https://github.com/argoproj/argo-workflows/blob/6cf6b86347ad281d356f6124ebb3fbea9e2b531c/workflow/controller/operator.go#L2550-L2558

Joibel:
I've addressed my concerns and rebased on main.

workflow/controller/operator.go
```diff
@@ -2232,14 +2232,28 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
...
+	deadline, pendingDeadline, err := woc.checkTemplateTimeouts(processedTmpl, node)
 	if err != nil {
 		woc.log.WithField("template", processedTmpl.Name).Warn(ctx, "Template exceeded its deadline")
+		if node.Type == wfv1.NodeTypePod {
+			// delete the timed-out pod so the resources it was waiting for are freed.
+			// Deletion is by UID so a pod recreated by a retry cannot be affected.
+			if pod, exists, podErr := woc.podExists(node.ID); podErr == nil && exists {
+				woc.controller.PodController.DeletePodByUID(ctx, pod.Namespace, pod.Name, string(pod.UID))
+			}
+		}
```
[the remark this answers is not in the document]
Joibel: Your proposed fix enhances the logging without any behaviour change, yet you wanted behaviour change.

Joibel (approved):

PR #12776, opened 2024-03-10
feat: add template/cron labels for pods. Fixes: #7611

### LINKED ISSUE #7611: Label pods with template/cron labels
Today, we label pods with the name of the workflow, but not the name of the workflow's template or cron that it was submitted from. It maybe useful to add these labels, so tools like Kubecost (which cannot lookup the owner) can track changes in the cost of running workflow over time.

shuangkun:
Fixes #7611

### Motivation

add template/cron labels for pods to facilitate cost analysis on the cloud for cron and template

### Verification
local test and e2e test.

### DIFF
--- a/workflow/common/common.go
+++ b/workflow/common/common.go
@@ -79,6 +79,8 @@ const (
 	LabelKeyWorkflowArchivingStatus = workflow.WorkflowFullName + "/workflow-archiving-status"
 	// LabelKeyWorkflow is the pod metadata label to indicate the associated workflow name
 	LabelKeyWorkflow = workflow.WorkflowFullName + "/workflow"
+	// LabelKeyTemplate is the pod metadata label to indicate the associated template name
+	LabelKeyTemplate = workflow.WorkflowFullName + "/template"
 	// LabelKeyComponent determines what component within a workflow, intentionally similar to app.kubernetes.io/component.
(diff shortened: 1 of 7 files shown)


workflow/common/common.go
```diff
@@ -79,6 +79,8 @@ const (
 	LabelKeyWorkflowArchivingStatus = workflow.WorkflowFullName + "/workflow-archiving-status"
 	// LabelKeyWorkflow is the pod metadata label to indicate the associated workflow name
 	LabelKeyWorkflow = workflow.WorkflowFullName + "/workflow"
+	// LabelKeyTemplate is the pod metadata label to indicate the associated template name
+	LabelKeyTemplate = workflow.WorkflowFullName + "/template"
```
agilgur5, line 83: Hmm I thought the issue meant `WorkflowTemplate`, for which the label key already exists on line 92 below.

`template` names are also not necessarily unique, unlike a `Workflow`, `CronWorkflow`, or `WorkflowTemplate` name. Though if the Pod has all the labels you can uniquely identify it from a combination

shuangkun: Thanks, I added WorkflowTemplate. Maybe Template also has some application scenarios. For example, what we are doing is automatically adjusting the CPU and memory for the new workflow based on the resource usage of the past workflow. It needs to correspond to the resource usage of a template and pod. If there is a label of the template, it will Much more convenient.

PR #12785, opened 2024-03-12
docs: update Helm values for SSO config

omerlh:
### Motivation

Thanks for the docs! They were really helpful - I fixed a small issues in the values.yaml I encountered when following them

### DIFF
--- a/docs/argo-server-sso-argocd.md
+++ b/docs/argo-server-sso-argocd.md
@@ -136,7 +136,9 @@ data:
 
 ```yaml
      server:
-       authModes: ['sso']
+       # Chart version 0.39.0 and after
+       authModes:
+         - sso
        sso:
          enabled: true
          issuer: https://argo-cd.mydomain.com/api/dex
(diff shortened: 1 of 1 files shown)


docs/argo-server-sso-argocd.md
```diff
@@ -136,9 +136,9 @@ data:
 
 ```yaml
      server:
-       extraArgs:
-         - --auth-mode=sso
+       authModes: ['sso']
```
agilgur5, line 139: > I fixed a small issues in the values.yaml

`authModes` is actually a newer feature in the Helm Chart. it sets the `args` under-the-hood, per https://github.com/argoproj/argo-helm/pull/2291 / https://github.com/argoproj/argo-helm/pull/2336. So it is functionally equivalent.
This feature does require a newer Chart version, >= [0.39.0](https://github.com/argoproj/argo-helm/pull/2336/files#diff-7c1bdf65c82164b5472073b2892bbb5d3bbe8f2ea4b2da6b600163b412fb4b90R6), so we should specify that, similar to how the docs specify Workflows's version in various places (#12581) 

```suggestion
       # Chart version 0.39.0 and after
       authModes:
         - sso
```
- also use consistent array syntax

PR #12713, opened 2024-02-29
chore(deps): bump github.com/vektra/mockery from v2.26.0 to v2.42.0

blkperl:
### Motivation

This is part of a series of PRs to update build dependencies until the build no longer panics on Golang 1.22

### Modifications

The new version of mockery generated some new files and no other changes were made. 

### Verification

No verification, the tests will fail if the mock changes do not work.

Issue #11159, opened 2023-06-01
Dropping of k8sapi executor makes upgrade from 3.2.6 to 3.4.8 not feasible with private images and `script` templates


vatine:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I expected an ugrade from 3.2.6 to 3.4.8 to be (mostly) "specify new images". It also required a small amount of tweaking RBAC roles.

I did not expect it to require reconfiguring every workflow (many of our workflows use custom, private, images; using the `script`functionality). With the upgrade primarily being motivated by "we want the fresh ssh_known_host file" (as opposed to having to use `insecureIgnoreHostKey`), the amount of work needed to switch from the k8sapi to the emissary executor was completely unexpected and disappointing, as for the `script`-type nodes, it really should be possible for Argo/emissary to figure out what the right command is (and bordering on hard for a human to do).

We do not want the Argo workflow controller to have access to our private registries (it should not need that access, the kubernetes cluster has the required image pull capabilities), but it is not obvious what the auto-generated command for a "script" will be (so the entrypoint cannot be specified neither as a `command` nor statically in a configuration file).

### Version

v3.4.8

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a (this is all private images)
```

### Logs from the workflow controller

```text
time="2023-06-01T07:18:47.310Z" level=warning msg="Non-transient error: failed to look-up entrypoint/cmd for image \"ELIDED\", you must either explicitly specify the command, or list the image's command in the index: https://argoproj.github.io/argo-workflows/workflow-executors/#emissary-emissary: GET https://ELIDED: UNAUTHORIZED: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: ELIDED"
```

### Logs from in your workflow's wait container

```text
n/a, the pod is never created
```

vatine:
One possible solution here is that for "script" actions, the emissary executor synthesises its own "command" setting. That would probably solve 98% of our incompatibility issues, and for sure the specific thing being executed is 100% within the control of the executor, so it should be relatively straight-forward.

agilgur5:
[...]

PR #12787, opened 2024-03-12
fix(executor): add executable permission to staged `script`

### LINKED ISSUE #3496: Error: `exec: "/argo/staging/script": permission denied`
# Summary

I want to stop this error:

```
failed to create containerd task: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/argo/staging/script\": permission denied": unknown
```

# Motivation

It is annoying to need to specify `command: [sh]` for scripts, when it could just be Unix executable.

```yaml
      script:
        image: 'docker/whalesay:latest'
        command: [sh]
        source: |
          true
```

```yaml
      script:
        image: 'docker/whalesay:latest'
        source: |
          #!/bin/sh
          true
```

# Proposal

Using chmod.

---

vatine:
Fixes #3496 

### Motivation

As-was, staged scripts end up with the permission `0644` `(rw-r--r--)` which is less than useful for shell scripts. With this modification in place, the scripts are executable.

### Modifications

Changed file permissions for the staged script.

### Verification

Built a docker image of `argoexec`, changed the workflow-controller flags to use the private image and verified that pod logs no longer had a `/argo/staging/script: permission denied` in them


agilgur5:
> I must say that I am not entirely sure how the test failures relate to this change. They seem to relate to completely different things.

Hmm seems to be the same one just mentioned [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1710260573672209) in #12713. Indeed seems unrelated

agilgur5 (commented):
> verified that pod logs no longer had a `/argo/staging/script: permission denied` in them

interestingly I've not seen this error before -- I wonder why it just popped up for you and hasn't popped up before?

vatine:
Even more interestingly, I get the permission error for some images, but not for others (for the ones where this change was required, I also needed to include "shebang" lines listing an interpreter). I have not managed to nail down what the actual difference between these are.

I can have a look at making the permission/mode value variable and setting it to 0o755 only for scripts.

workflow/executor/executor.go
```diff
@@ -283,7 +283,7 @@ func (we *WorkflowExecutor) StageFiles() error {
 	default:
 		return nil
 	}
-	err := os.WriteFile(filePath, body, 0o644)
+	err := os.WriteFile(filePath, body, 0o755)
```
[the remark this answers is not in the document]
vatine: I have just made a change, making this conditioned on "is a resource or script".

agilgur5:
> Even more interestingly, I get the permission error for some images, but not for others (for the ones where this change was required, I also needed to include "shebang" lines listing an interpreter)

This lines up with what I'm suspecting, which is that this might be a [shell form vs. exec form](https://docs.docker.com/reference/dockerfile/#shell-and-exec-form) instance or something similar. Like if you have a script without executable permissions, you can still run it with, e.g. `bash my-script.sh`, but `. my-script.sh` will fail.

Looking at the code, the path to the `script` `source` file is passed as [an arg](https://github.com/argoproj/argo-workflows/blob/5b3909b9567a5c6fc4fef81ac0785cfddbffb291/workflow/controller/operator.go#L3002). I'm not entirely sure if `command` is set elsewhere, so it may very well depend on your container image's default `cmd` -- I think that would add up.

vatine:
Certainly, the images that I can recall failing are on private repos (and, thus, as detailed in #11159 , the workflow controller cannot read the command from the image when constructing the pod), I have tried both `bash` and `bash -c` as the configured command in the configmap, with identical results. I can check what the pod command setting is. What I do know is that with this executable permission, it works. In all cases where this has failed, it is also not the first step in a workflow.

workflow/executor/executor.go
```diff
@@ -267,11 +267,13 @@ func (we *WorkflowExecutor) LoadArtifacts(ctx context.Context) error {
 func (we *WorkflowExecutor) StageFiles() error {
 	var filePath string
 	var body []byte
+	mode := 0o644
```
agilgur5, line 270: this is failing to build due to a typecheck error (see [lint](https://github.com/argoproj/argo-workflows/actions/runs/8254003171/job/22577149813?pr=12787) failure etc).

I think this is supposed to be written as:
```suggestion
	mode := os.FileMode(0o644)
```

I was hoping there might be some constants for these modes, but seems like [no](https://go.dev/src/io/fs/fs.go) (note that [`os.FileMode = fs.FileMode`](https://pkg.go.dev/os#FileMode))

workflow/executor/executor.go
```diff
@@ -267,11 +267,13 @@ func (we *WorkflowExecutor) LoadArtifacts(ctx context.Context) error {
...
 	switch we.Template.GetType() {
 	case wfv1.TemplateTypeScript:
 		log.Infof("Loading script source to %s", common.ExecutorScriptSourcePath)
 		filePath = common.ExecutorScriptSourcePath
 		body = []byte(we.Template.Script.Source)
+		mode = 0o755
```
agilgur5, line 276:
```suggestion
		mode = os.FileMode(0o755)
```

same as above

vatine:
This is the command and args from the relevant pod:

```
$ kubectl get pod -n argo <pod name> -o json | jq '.spec.containers[1] | (.command,.args)'
[
  "/var/run/argo/argoexec",
  "emissary",
  "--loglevel",
  "debug",
  "--log-format",
  "text",
  "--"
]
[
  "/argo/staging/script"
]
```

I still cannot explain why it works in some cases and fail in others.

agilgur5:
> I still cannot explain why it works in some cases and fail in others.

Sorry for the delayed response; I wanted to take some time to see if we could root cause this and have been pretty overloaded. Thanks for looking more into this.

> I have tried both `bash` and `bash -c` as the configured command in the configmap, with identical results.

This is surprising... based on this and your main container's `command`, it almost seems like it's not picking that `command` up from your configmap? If it works in some but not others, I might guess some images are not specified?

> In all cases where this has failed, it is also not the first step in a workflow.

This shouldn't matter, it has no impact on the `command`.

> What I do know is that with this executable permission, it works.

Since we don't know the mode of the `command` beforehand, I think this might still be a correct change; some `command`s will need this permission.
Since it's a permission though, I want to be pretty careful to not extend it in a way that could cause a vuln. So far, most threat models I've thought of are not particularly impactful since the `script` is user input anyway, but I'm not sure I've covered all the bases.

vatine:
This is what Argo puts in the pod's "args:" and "command:" (note this probably has some relevance to issue 11159, as I see no actual trace of the "command" or "entrypoint" in the argoexec call; for those where t just seems to work,  can see a terminal "/bin/sh" on the argoexec command line).

``` 
$ kubectl get pods -n argo pr-run-e-e-tes-202404191144-docker-1590034149 -o json | jq '.spec.containers[1] | .command, .args'
[
  "/var/run/argo/argoexec",
  "emissary",
  "--loglevel",
  "info",
  "--log-format",
  "text",
  "--"
]
[
  "/argo/staging/script"
]
```

This is with the configmap configured to have that image mapped to:
```
cmd:
  - /bin/bash
```

agilgur5:
> Based on my analysis in [#11159 (comment)](https://github.com/argoproj/argo-workflows/issues/11159#issuecomment-2066761200), since we don't know the mode of the `command` beforehand, I think this might still be a correct change; some `command`s will need this permission.

Apparently, per #13549, there was a separate issue filed exclusively for the script exec perms in #3496. Given both of those and the above, this is probably worthwhile to have in its own right, regardless of the Emissary nuances

agilgur5 (approved):
Thanks for finding, fixing, and iterating on this again!

agilgur5:
> > I have tried both `bash` and `bash -c` as the configured command in the configmap, with identical results.
> 
> This is surprising... based on this and your main container's `command`, it almost seems like it's not picking that `command` up from your configmap?

> This is with the configmap configured to have that image mapped to:
> 
> ```yaml
> cmd:
>   - /bin/bash
> ```

I found the problem here. `cmd` in Docker is [equivalent to `args`](https://github.com/argoproj/argo-workflows/blob/54621cc60117cf68183be24322119d85a80bb650/workflow/controller/workflowpod.go#L403) in k8s. There's no example of this in the docs, but you could [set `entrypoint` instead](https://github.com/argoproj/argo-workflows/blob/a0ba3c70d18227e74bed1d77aa86c4ad4212d159/workflow/controller/entrypoint/config_index.go#L16) in your `images` config.

The bug is that for `script` templates, `cmd` seems to be accidentally ignored. The `entrypoint` + `cmd` lookup only [sets `args` if they are not already set](https://github.com/argoproj/argo-workflows/blob/54621cc60117cf68183be24322119d85a80bb650/workflow/controller/workflowpod.go#L402) in the `container` spec. And for `script` templates, the `ScriptSourcePath`, `/argo/staging/script` currently, is [appended to the `args`](https://github.com/argoproj/argo-workflows/blob/6b16f24c97e4b9c792e18835304e517f2d0ddd0f/workflow/controller/operator.go#L3023) just before Pod creation -> entrypoint lookup occurs.

This is fixable by changing the ordering, but may be a breaking change for `script` templates, as existing ones would unexpectedly have their `cmd` added.

For clarity, the full `command` + `args` for `script` templates would end up as:
```yaml
[
  "/var/run/argo/argoexec",
  "emissary",
  "--loglevel",
  "info",
  "--log-format",
  "text",
  "--",
  "<entrypoint>"
],
[
  "<cmd>",
  "/argo/staging/script"
]
```

where `cmd` is currently ignored.

PR #12775, opened 2024-03-09
chore(deps): upgrade `actions/cache` and `create-pull-request` to Node v20

agilgur5:
### Motivation

<details> <summary>Both of these were giving warnings in GH Actions runs during <a href="https://github.com/argoproj/argo-workflows/issues/11997#issuecomment-1972029716">the 3.5.5 release</a> </summary>

  - [`release.yaml` run](https://github.com/argoproj/argo-workflows/actions/runs/8102509236)
    - > Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/cache@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
    - Screenshot for posterity: Screenshot 2024-03-09 at 3 30 43 AM
    
  - [`changelog.yaml` run](https://github.com/argoproj/argo-workflows/actions/runs/8102509233)
    - > Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: peter-evans/create-pull-request@v5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
    - Screenshot for posterity:  Screenshot 2024-03-09 at 3 31 26 AM

</details>

### Modifications

- `actions/cache` [v4](https://github.com/actions/cache/releases/tag/v4.0.0) updated to Node 20
  - no other breaking changes
  - used latest [v4.0.1](https://github.com/actions/cache/releases/tag/v4.0.1) (v3.3.3 -> v4.0.1)

- `peter-evans/create-pull-request` [v6](https://github.com/peter-evans/create-pull-request/releases/tag/v6.0.0) updated to Node 20
  - other breaking changes do not affect our usage:
    - we don't use `push-to-fork`
    - the `author` and `committer` changes are to the automated bot and just make it more accurate
  - used latest [v6.0.1](https://github.com/peter-evans/create-pull-request/releases/tag/v6.0.1) (v5.0.2 -> v6.0.1)

### Verification

This only runs during releases, so a bit hard to test 😕 

### Future Work

A few more actions need updating as well, but they have some blockers that need resolving:
1. `actions/upload-artifact` and `actions/download-artifact` blocked by intermittent failures: #12455 
1. `Azure/docker-login` has no update / seems unmaintained in general: https://github.com/Azure/docker-login/pull/64
    - should probably switch all these to `docker/login-action` which we already use in the very same file; they _should_ work on Windows too since actions are all Node-based

PR #4127, opened 2020-09-24
chore(build): Pre-push temporary manifest with linux/amd64. Fixes #4062

lippertmarkus:
This is pre-pushing a temporary manifest to DockerHub referencing the linux/amd64 images right after those are built so we can release earlier without having to wait for the long-running arm64 image builds. After all images finished building the temporary manifest is overwritten with a new one with references to all images for amd64, arm64 and windows.

Details: 
1. Building and pushing all images for linux/amd64, e.g. `argoproj/argoexec:v2.11-linux-amd64` - this will only take around 6minutes
2. Create and push the manifest and reference only the linux/amd64 images yet, e.g. `argoproj/argoexec:v2.11` referencing only `argoproj/argoexec:v2.11-linux-amd64` for now
3. When all other images for Windows and linux/arm64 finished building after 40mins, we update and push the manifest overwriting the one created before to reference all images now instead of only the linux/amd64 ones, e.g. override `argoproj/argoexec:v2.11` now referencing `argoproj/argoexec:v2.11-linux-amd64`, `argoproj/argoexec:v2.11-linux-arm64`, `argoproj/argoexec:v2.11-windows`

Example run: https://github.com/lippertmarkus/argo/actions/runs/270351664

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or **(c) this is a chore**.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed the CLA.
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged. **Not needed.**
* [x] My builds are green. Try syncing with master if they are not. 
* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).

PR #3291, opened 2020-06-24
chore: Use GitHub Actions to build Docker Images to allow publishing Windows Images

lippertmarkus:
As mentioned in #2747 Docker Hub Automated Build Hooks don't allow to build Windows images. As v2.9.0 comes with Windows support for argoexec and as you migrated CI to GitHub Actions also, it would make sense to use GitHub Actions also for building the Docker images.

I added a workflow file which builds and pushes the images for Linux (workflow-controller, argoexec, argocli) and Windows (argoexec only) whenever a new tag `v*` is pushed. I used the existing Docker Hub Hooks (https://github.com/argoproj/argo/blob/master/hooks/build) as a base for the scripts. The jobs building and pushing the images are appending `-linux` or `-windows` to the tag, a third job is creating a Manifest and pushes it to allow for multiarch. Example:

1. Tag `v2.9.0` is pushed
1. Job `build-linux` is building and pushing `argoproj/workflow-controller:v2.9.0-linux`, `argoproj/argoexec:v2.9.0-linux` `argoproj/argocli:v2.9.0-linux`
1. Job `build-windows` is building and pushing `argoproj/argoexec:v2.9.0-windows`
1. Job `push-images` is creating and pushing the following image manifest:
   - `argoproj/workflow-controller:v2.9.0` referencing `argoproj/workflow-controller:v2.9.0-linux`
   - `argoproj/argoexec:v2.9.0` referencing `argoproj/argoexec:v2.9.0-linux` and `argoproj/argoexec:v2.9.0-windows` (multiarch)
   - `argoproj/argocli:v2.9.0` referencing `argoproj/argocli:v2.9.0-linux`

You can find an example run for a tag `vtest17` here: https://github.com/lippertmarkus/argo/runs/804344322
Resulting image, e.g. for argoexec: https://hub.docker.com/repository/docker/lippertmarkus/argoexec/tags

You only need to add the following two secrets like described [in the GitHub Actions Help](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets-for-a-repository) before merging:
- `DOCKERIO_USERNAME` = `argoproj`
- `DOCKERIO_PASSWORD` = `<a-docker-hub-pat-token>`

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or **(c) this is a chore**.
[...]

PR #12791, opened 2024-03-12
chore(deps): use `docker/login-action` consistently instead of `Azure/docker-login`

agilgur5:
Follow-up to #12775

### Motivation

- `Azure/docker-login` seems to have been the first one to have been added in #3291, then `docker/login-action` was added after in #4127, but currently _both_ are being used without consistency

- [`Azure/docker-login`](https://github.com/Azure/docker-login) is not maintained much, with the [last release](https://github.com/Azure/docker-login/releases/tag/v1.0.1) ~1.5 years ago and no update to Node v20 as of yet, despite issues and PRs on it from a few months ago (https://github.com/Azure/docker-login/issues/63, https://github.com/Azure/docker-login/pull/64)

### Modifications

- consistently use [`docker/login-action`](https://github.com/docker/login-action), which is more maintained, more centralized, more commonly used, and not on EoL Node
    - [last release](https://github.com/docker/login-action/releases/tag/v3.0.0) a few months ago, [last commit](https://github.com/docker/login-action/commit/566711b9b940f3a4aea3c883c1efeb76ab18c145) a few weeks ago

### Verification

This only runs during releases, so a bit hard to test 😕


terrytangyuan (commented):
Why not combining with https://github.com/argoproj/argo-workflows/pull/12775?

agilgur5:
> Why not combining with #12775?

Well, to begin with, that one was made a few days beforehand (both made 2+ weeks ago), and has now been merged earlier too.

Otherwise there are purposes are different, per PR title and the "Motivation" section in each. That one is an update, this one is changing the action itself to a different one (which happens to be more updated). That also entails different risk levels for each.

Similar to mine and others' PRs in the past, they are also independent of each other and so don't need to be tied together.

agilgur5:
Merged `main` for new PR checks per #13027

terrytangyuan (approved):

agilgur5:
Huh so this did partially [fail](https://github.com/argoproj/argo-workflows/actions/runs/9409633233/job/25920972244) on release on `main`:

```
Run echo $(jq -c '. + { "experimental": "enabled" }' ${DOCKER_CONFIG}/config.json) > ${DOCKER_CONFIG}/config.json
jq: error: Could not open file /config.json: No such file or directory
/home/runner/work/_temp/3b2a11f7-f9e3-4c1e-8ab2-431101712831.sh: line 1: /config.json: Permission denied
```

All the builds and pushes of images worked, it's the manifest push that failed; i.e. [this line](https://github.com/argoproj/argo-workflows/blob/bda02806cdc27216cd0ce33418997bf5a4215ae1/.github/workflows/release.yaml#L175), which I didn't change in the PR.

It seems like `${DOCKER_CONFIG}` is empty per the error message 🤔 So it may have only been set by `Azure/docker-login`, which means the code was relying on an env var side effect, if I'm understanding correctly.
The default is `~/.docker/config.json` so this may be easily replaceable

agilgur5:
> The default is `~/.docker/config.json` so this may be easily replaceable

Added this as a default in #13155

Issue #11984, opened 2023-10-11
Workflow processing has been postponed because too many workflows are already running


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

i have parallelism set to 30 in configmap of workflow controller but even with something like 17 workflows running (confirmed from ui and cli) it can sometimes keep workflows in pending state with message 'Workflow processing has been postponed because too many workflows are already running' for a long time

### Version

3.4.11

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

tooptoop4:
they only rely on parallelism but there were other workflows waiting on a mutex:
```
NAME                                               STATUS      AGE    MESSAGE
aaaa-dkntk                                         Running     32m
redact-20231010t044555074ztkhzp                    Running     11m
redact-20231010t044558032z9q6xp                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
redact-20231010t044602469zh999j                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
redact-20231010t044608533zhn6g4                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
redact-20231010t044608554ztjsfn                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
redact-20231010t044614762ztqslg                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
redact-20231010t044615441zxngwl                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
redact-20231010t044620131zrsghg                    Pending            Waiting for auth/Mutex/tbl0/ lock. Lock status: 0/1
[...]

PR #12792, opened 2024-03-13
fix: terminate workflow should not get throttled Fixes #12778

### LINKED ISSUE #12778: Cronworkflow with concurrencyPolicy: Replace/Forbid should not queue up pending workflows
### Pre-requisites

When argo-workflow-controller hits parallelism limit,
the new scheduled cronworkflow will all be in pending regardless of the `concurrencyPolicy`

[Concurrent policy](https://argo-workflows.readthedocs.io/en/latest/cron-workflows/#cronworkflow-options)
Policy that determines what to do if multiple Workflows are scheduled at the same time. Available options: Allow: allow all, Replace: remove all old before scheduling a new, Forbid: do not allow any new while there are old

image

Although when the parallelism slot is available, the controller will clear up the pending cronworkflow according to `concurrencyPolicy`
There still exists a problem if we have alerts that depend on the number of pending workflows, we could end up with hundreds of pending workflow that triggers the alert.
I think in an ideal scenario, when a cronworkflow is scheduled, It should check existing scheduled pending workflow as well and operate accordingly.

### Version

v3.5.5

```YAML
metadata:
  name: omniscient-rhino
  namespace: argo
spec:
  schedule: '* * * * *'
  concurrencyPolicy: "Replace"
  workflowSpec:
    entrypoint: argosay
    templates:
        container:
          name: main
          image: argoproj/argosay:v2
          command:
            - /argosay
          args:
            - echo
            - hello
```

```text
```

```text
```

tczhao:
Fixes #12778 

### Motivation

To prevent cronworkflow from queueing up pending workflows when
- reached parallelism limit 
- concurrenpolicy-replace already update the spec of the pending workflow to `shutdown: terminated`

### Modifications

Allow workflow in the queue to be processed (operated) regardless of parallelism limit.

We have scenarios
- reach parallelism limit, terminate running workflow (no issue)
- reach parallelism limit, terminate pending workflow (we should not block the queue from processing these workflow, just like semaphore doesn't block the queue from processing pending workflow that got terminated https://github.com/argoproj/argo-workflows/blob/5b3909b9567a5c6fc4fef81ac0785cfddbffb291/workflow/controller/operator.go#L528)

### Verification

Add test,
test fail
Add code change
test succeed

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -834,7 +834,7 @@ func (wfc *WorkflowController) processNextItem(ctx context.Context) bool {
 
 	woc := newWorkflowOperationCtx(wf, wfc)
 
-	if !wfc.throttler.Admit(key.(string)) {
+	if !(woc.GetShutdownStrategy().Enabled() && woc.GetShutdownStrategy() == wfv1.ShutdownStrategyTerminate) && !wfc.throttler.Admit(key.(string)) {
 		log.WithField("key", key).Info("Workflow processing has been postponed due to max parallelism limit")
 		if woc.wf.Status.Phase == wfv1.WorkflowUnknown {
 			woc.markWorkflowPhase(ctx, wfv1.WorkflowPending, "Workflow processing has been postponed because too many workflows are already running")
--- a/workflow/controller/controller_test.go
+++ b/workflow/controller/controller_test.go
@@ -692,10 +692,22 @@ spec:
       container:
         image: my-image
 `),
+				wfv1.MustUnmarshalWorkflow(`
+metadata:
+  name: my-wf-2
+spec:
+  shutdown: Terminate
+  entrypoint: main
+  templates:
+    - name: main
+      container:
+        image: my-image
+`),
 				f,
 			)
 			defer cancel()
 			ctx := context.Background()
+			assert.True(t, controller.processNextItem(ctx))
 			assert.True(t, controller.processNextItem(ctx))
 			assert.True(t, controller.processNextItem(ctx))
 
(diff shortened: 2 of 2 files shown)


agilgur5:
> * concurrenpolicy-replace already update the spec of the pending workflow to `shutdown: terminated`

oh interesting, I didn't know that. that does make the fix much simpler.

> * reach parallelism limit, terminate running workflow (no issue)
> * reach parallelism limit, terminate pending workflow (we should not block the queue from processing these workflow, just like semaphore doesn't block the queue from processing pending workflow that got terminated

this makes sense to me and sounds consistent, so I'll approve, but I'm not an expert in the synchronization area of the codebase so would appreciate if someone else could take a look. in particular, technically speaking this change means that terminating Workflows are no longer counted toward parallelism limits, right?

agilgur5 (approved):

agilgur5:
Also if you're interested, another synchronization race condition issue is #11984

tczhao:
> technically speaking this change means that terminating Workflows are no longer counted toward parallelism limits, right?

Correct,
We either
- terminate running workflow (in parallelism limit, release it from parallelism)
- terminate pending/unknown workflow (bypass parallelism)
- terminate finished workflow (no effect)

workflow/controller/controller.go
```diff
@@ -834,7 +834,7 @@ func (wfc *WorkflowController) processNextItem(ctx context.Context) bool {
 
 	woc := newWorkflowOperationCtx(wf, wfc)
 
-	if !wfc.throttler.Admit(key.(string)) {
+	if !(woc.GetShutdownStrategy().Enabled() && woc.GetShutdownStrategy() == wfv1.ShutdownStrategyTerminate) && !wfc.throttler.Admit(key.(string)) {
```
sarabala1979, line 837: @tczhao Quick suggestion:
Can we handle this state workflow in the informer filter section so, the workflow will not even get into the throttler queue?
https://github.com/argoproj/argo-workflows/blob/59d7e5b129d736ae1f34c51eaccb96dd3cb0d7be/workflow/controller/controller.go#L992

tczhao: The informer filter only exclude the workflow.
We want to include the shutdown workflow in the wfqueue, and process it.
Here we skip throttle/parallelism check if the workflow is marked shutdown

sarabala1979: oh yeah got it thanks.  I thought opposite way. make sense

sarabala1979 (approved):
LGTM

agilgur5:
Backported cleanly to `release-3.5` as c9dd50d35

PR #12798, opened 2024-03-14
fix: allow `workflow.failures` variable in exit hooks. Fixes #12330

### LINKED ISSUE #12330: Failed to resolve `{{workflow.failures}}` in exit handler
### Pre-requisites

according to the documentation here: https://argoproj.github.io/argo-workflows/variables/#exit-handler

workflow.failures supposed to be available to use, but it isnt. i tried to submit a workflow template and got the following: `templates.send-slack: failed to resolve {{ workflow.failures.message }}`

### Version

v3.5.1

```YAML
# An exit handler is a template reference that executes at the end of the workflow
# irrespective of the success, failure, or error of the primary workflow. To specify
# an exit handler, reference the name of a template in 'spec.onExit'.
# Some common use cases of exit handlers are:
# - sending notifications of workflow status (e.g. e-mail/slack)
# - posting the pass/fail status to a webhook result (e.g. github build result)
# - cleaning up workflow artifacts
# - resubmitting or submitting another workflow
kind: Workflow
metadata:
  generateName: exit-handlers-
spec:
  entrypoint: intentional-fail
  onExit: exit-handler
  templates:
  # primary workflow template
  - name: intentional-fail
    container:
      image: alpine:latest
      command: [sh, -c]

  # exit handler related templates
  # After the completion of the entrypoint template, the status of the
  # workflow is made available in the global variable {{workflow.status}}.
  # {{workflow.status}} will be one of: Succeeded, Failed, Error
  - name: exit-handler
    steps:
    - - name: notify
        template: send-email
      - name: celebrate
        template: celebrate
        when: "{{workflow.status}} == Succeeded"
      - name: cry
        template: cry
  - name: send-email
    container:
      image: alpine:latest
      command: [sh, -c]
      # Tip: {{workflow.failures}} is a JSON list. If you're using bash to read it, we recommend using jq to manipulate
      # it. For example:
      #
      # echo "{{workflow.failures}}" | jq -r '.[] | "Failed Step: \(.displayName)\tMessage: \(.message)"'
      #
      # Will print a list of all the failed steps and their messages. For more info look up the jq docs.
      # Note: jq is not installed by default on the "alpine:latest" image, however it can be installed with "apk add jq"
      args: ["echo send e-mail: {{workflow.name}} {{workflow.status}} {{workflow.duration}}. Failed steps {{workflow.failures}}"]
  - name: celebrate
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo hooray!"]
  - name: cry
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo boohoo!"]
```

```text
```

```text
```

eduardodbr:
Fixes #12330

### Motivation

The [Lifecycle-Hook docs](https://argo-workflows.readthedocs.io/en/latest/lifecyclehook/) say `workflow.failures` is available in exit hooks but it is failing during validation, check https://github.com/argoproj/argo-workflows/issues/12330#issuecomment-1995868147. 
This seems to be a regression but I couldn't find the PR that introduced this issue so I just fixed it. 

### Modifications

The variable `workflow.failures` was added to the scope when validating the template used in exit hook like:

```yaml 
hooks:
    exit:
      template: fail
```
### Verification

```yaml
kind: Workflow
metadata:
  name: exit-handlers
spec:
  entrypoint: pass
  hooks:
    exit:
      template: fail
  templates:
  - name: pass
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["exit 1"]
  - name: fail
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo {{workflow.failures}}"]
```


workflow/validate/validate.go
```diff
@@ -282,6 +282,21 @@ func ValidateWorkflow(wftmplGetter templateresolution.WorkflowTemplateNamespaced
...
+		tmplHolder = &wfv1.WorkflowStep{Template: hook.Template}
+		if hook.TemplateRef != nil {
+			tmplHolder = &wfv1.WorkflowStep{TemplateRef: hook.TemplateRef}
+		}
+		ctx.globalParams[common.GlobalVarWorkflowFailures] = placeholderGenerator.NextPlaceholder()
+		_, err = ctx.validateTemplateHolder(tmplHolder, tmplCtx, &hook.Arguments, opts.WorkflowTemplateValidation)
+		if err != nil {
+			return err
+		}
+	}
```
tczhao, line 299: I think we better have this as part of `validateHooks` function.
So we validates the hooks at both the workflow and template level

workflow/validate/validate_test.go
```diff
@@ -1082,6 +1082,79 @@ func TestExitHandler(t *testing.T) {
...
+      command: [sh, -c]
+      args: ["exit 0"]
+  - name: fail
+    container:
+      image: alpine:latest
+      command: [sh, -c]
+      args: ["echo {{workflow.failures}}"]
+`
+
+func TestExitHandlerWorkflowFailures(t *testing.T) {
```
tczhao, line 1144: need a test for template level exit hook as well

tczhao (changes requested):
Thanks for the contribution

workflow/validate/validate.go
eduardodbr: I think validating the exit hook at both template level and workflow level makes sense. However, `workflow.failures` is not on the scope of a step's exit hook, so the validation should fail, correct?

agilgur5:
Does #13246's approach make more sense here?

workflow/validate/validate.go
agilgur5: Yes, it should fail. 

There is an edge case of the last template of a Workflow, which could technically be equivalent to a Workflow-level exit hook, but I think we can more-or-less ignore that scenario as the user intent is not there

Bo0km4n:
@agilgur5 @tczhao @eduardodbr 
Hi, I'm currently struggling with this issue, is it scheduled to be merged?

agilgur5:
There are still changes requested here

Bo0km4n:
@agilgur5 If @eduardodbr will not response soon, can I do fix the changes requested on this PR?

agilgur5:
Yes, that's an option with any stale PR, just make sure to credit the original author (easiest way is to add your changes on top of their commits in your own PR and to reference the old PR in the description)

Bo0km4n:
@agilgur5 @tczhao I added template level tests on this PR #13492. Please review this 🙏

eduardodbr:
> @agilgur5, @tczhao  If @eduardodbr will not response soon, can I do fix the changes requested on this PR?

I am currently on PTO so can't help right now. Feel free to contribute. Thanks!

PR #12799, opened 2024-03-14
fix: re-enqueue parallelism when wf block by wf-level synchronization. Fixes #11984

### LINKED ISSUE #11984: Workflow processing has been postponed because too many workflows are already running

tczhao:
Fixes #11984

### Motivation

Given a certain parallelism limit.
If a workflow couldn't execute due to workflow level synchronization lock (semaphore or mutex),
the blocked workflow will release the parallelism lock to allow other workflows to execute.

### Modifications

We add an additional `inProgressNoLimit` map in parallelism Throttler,
Workflows blocked by workflow-level mutex/semaphore will
- get removed from `throttler.inProgress` (so parallelism limit can admit other workflows in `throttler.pending`)
- added to `throttler.inProgressNoLimit`, allows `throttler` to `Admit` the workflow in the future

The number of parallelism (concurrent workflow) still holds since the workflow block by mutex won't start until another workflow is finished and the mutex is released.

### Verification

Add test
test fail (verify original issue)
Add changes
test succeed


tczhao:
There's an issue with the current approach,
Original synchronization(parallelism) after block by mutex, it doesn't store any item in `wfQueue`, only enqueue `wfQueue` after mutex is release, doesn't waste any process
but the current PR approach (throttle.remove + add when block by mutex) results in an infinite enqueue/dequeue to/from `wfQueue` until block by mutex, wasting a lot of processing power.
---
working on a new approach now

tczhao:
Updated with new approach, ready for review

agilgur5:
> There's an issue with the current approach,

Ah your original approach was surprisingly simple, I guess it was _too_ simple 😅

> * results in an infinite loop of enqueue/dequeue to/from `wfQueue` until mutex release, wasting a lot of resource.

Yea that makes sense. Although the queue in general has this a bit whenever a reconcile is not yet necessary. Was it a significant uptick in CPU?

The optimal approach would effectively be to synchronize the two, er, synchronization primitives, i.e. synchronize the synchronization 😅

tooptoop4 (approved):

tczhao:
> Was it a significant uptick in CPU? If it follows `DEFAULT_REQUEUE_TIME`, it should take 10s between each, so I wouldn't think that would be significant

In the original approach, every time we `throttle.Remove/Add` it will check if there's a slot available then pop the key from `throttle.pending` and add to `wfQueue` with `DefaultControllerRateLimiter`. With this, we will see the same mutex blocked workflow getting processed 10+ times in a 1 min interval

tooptoop4:
:shipit:

tooptoop4:
🙏

tooptoop4:
@tczhao do u think this patch will work on 3.4.x line?

tczhao:
> @tczhao do u think this patch will work on 3.4.x line?

Hi @tooptoop4 , this should work on 3.4 without issue

agilgur5:
@tczhao see the test failure are unrelated, see #13002 and [Slack thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1714810708644859?thread_ts=1714810708.644859&cid=C0510EUH90V) where Isitha, Alan, and I were investigating it but haven't found the root cause yet

workflow/sync/throttler.go
```diff
@@ -38,12 +40,13 @@ var NamespaceBucket BucketFunc = func(key Key) BucketKey {
...
-	queue       QueueFunc
-	bucketFunc  BucketFunc
-	inProgress  buckets
-	pending     map[BucketKey]*priorityQueue
-	lock        *sync.Mutex
-	parallelism int
+	queue             QueueFunc
+	bucketFunc        BucketFunc
+	inProgress        buckets
+	inProgressNoLimit buckets
```
sarabala1979, line 46: Will this give the user the wrong count?
If Parallelism is set to 10, 5 WFs await the lock. Given time, the user may see 15 WFs are running state.
if N WFs are waiting for the Lock then N+10 workflows are in running state.
Is it expected behavior?

sarabala1979: Just a proposal: Need to discuss 
Can we completely exclude workflows which have locks on this from parallisam queue and let mutex will control it?
I believe that parallelism and mutex locks tackle the same rate limiting problem.
@tczhao @agilgur5 @terrytangyuan @juliev0

agilgur5: > I believe that parallelism and mutex locks tackle the same rate limiting problem.

They do, but there are use-cases where both overlap and you'd want to use both simultaneously.
For instance, I have only the resources to run 10 Workflows, but this set of Workflows cannot run in parallel and so needs a mutex. I still don't want more than 10 in that case

agilgur5: > Will this give the user the wrong count?
> If Parallelism is set to 10, 5 WFs await the lock. Given time, the user may see 15 WFs are running state.
> if N WFs are waiting for the Lock then N+10 workflows are in running state.

Oh. I think you're right. With a Workflow-level mutex, it might still be `Pending`, but a template-level mutex would remain in `Running` even if it's waiting on a lock.
So it creates the inverse problem of the issue, although only from a visibility and queue perspective, not from an "actually running" perspective

agilgur5: Haven't thought about it in-depth yet, but this might require a new state/phase to correctly implement this, i.e. "Locked" or "Waiting" (or something else better named). Which is not the easiest to implement as it requires changes in _a lot_ of places and almost certainly crosses the threshold into a feature as such

sarabala1979: >Haven't thought about it in-depth yet, but this might require a new state/phase to correctly implement this, i.e. "Locked" or "Waiting" (or something else better named). Which is not the easiest to implement as it requires changes in a lot of places (see also #11849) and almost certainly crosses the threshold into a feature as such

Change on state machine will be a big change. I won't recommend it. It may affect the stability of the product.
Suggestions:
I would recommend adding a sub-state which is just info to the user.
```yaml
Phase: Running
SubPhase: lockWaiting
```

agilgur5: > I won't recommend it. It may affect the stability of the product.

We already have needs for new phases, so it's not new in that sense. But yes, it requires a lot of care to implement correctly (as I myself said in various places).

> I would recommend adding a sub-state which is just info to the user.

Mmmm I wouldn't recommend this:
1. We already have _pseudo_-sub-states in the `message` and in a few other forms (e.g. shutdown, hooked, retried, maybe more). Info to the user can already be provided via the `message` and is already used for that purpose, such as for shutdown
1. "Info to the user" is not enough for this if we're going to change counts
  1. So that effectively becomes a nested state, which is generally a _more_ complex structure than flattened, as well as requires a new spec (as opposed to a new enum for the existing state)
    1. In general, various bugs are due to existing psuedo-sub-states. We should prefer correctness to the state machine rather than workarounds on workarounds; we already have an overreliance on such things like derivations too.
    1. Though if nesting makes more sense correctness-wise or complexity wise, then I'm not against that per se; just against workarounds over correctness

tooptoop4:
cc @Joibel

workflow/sync/throttler.go
brokenjacobs: It would seem a rewrite is necessary anyway as these primatives fail under load in the current releases.

PR #12808, opened 2024-03-15
docs: expand documentation for withSequence

cardoe:
Expanded documentation for withSequence in the loop documentation.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>

### DIFF
--- a/docs/walk-through/loops.md
+++ b/docs/walk-through/loops.md
@@ -37,6 +37,11 @@ spec:
       args: ["hello world!"]
 ```
 
+You can optionally specify a `start` and use `end` instead of `count`.
+And even supply a `format` string to customize the sequence value passed
+to the template. You do not have to define the `item` as an input parameter
+if you do not need it. For more details see the [Sequence Field Definition](../fields.md#sequence).
+
 ## `withItems` basic example
 
 This iterates over a list of items with `withItems`, substituting a string for each instantiated template.


cardoe:
I originally wrote my version looking at the stable docs and not seeing `withSequence` documented. But it appears someone added it in the latest docs which I should have checked. Submitting my delta anyway since it provides a bit more information.

agilgur5:
Yea it was added recently in #12577, specifically after it was noted missing from the docs (around here: https://github.com/argoproj/argo-workflows/pull/12577#issuecomment-1915192830).

I can backport that to the stable docs as it's an existing feature, just didn't have docs before.

agilgur5 (commented):
Adding (more) examples here would be a lot easier to interpret than descriptions

cardoe:
Well help me understand the best place because I've been confused about the docs layout. There's examples of Sequence and all the options s in the examples directory. A lot of times the features in the examples directory are more detailed than the docs.

Then there's the user guide section which I would assume to be the all encompassing place for feature docs but that's not the case for some features like looping which is in the walk through.

Should we link to the examples directory? Should examples live in the docs? User Guide or Walk Through?

I'm happy to contribute, I just don't know where.

agilgur5:
> Then there's the user guide section which I would assume to be the all encompassing place for feature docs but that's not the case for some features like looping which is in the walk through.

Yea some of the separation is indeed confusing. I mentioned this in #11450 / #11500 / #11501 too. Now that #12501 has been merged, we can change up some of the layout without breaking existing links.

> A lot of times the features in the examples directory are more detailed than the docs.

I'm not sure how it was originally designed, but in its current state it effectively acts as a deeper dive on certain options -- in particular, all examples are automatically linked in the [Fields Reference](https://argo-workflows.readthedocs.io/en/latest/fields/).
Some examples are also duplicated in the docs though, which should really use [embeds](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files).

I could definitely use help improving the docs in lots of ways! I've been the main person doing substantial rewrites in recent times.

> Should we link to the examples directory? Should examples live in the docs? User Guide or Walk Through?

If there are existing examples that cover all of the options you're trying to describe, then linking to them or using embeds would be preferred.
Looping isn't covered in the User Guide at all, so let's leave it in the Walk Through as you have now.

We may want to move the whole page into the "Features" section of the User Guide later. I'm not totally sure what to do with the Walk Through yet; if it should link to advanced topics in the User Guide or if it should just be completely merged with the User Guide in some way.

cardoe:
I'll get it fixed up

PR #12809, opened 2024-03-15
docs: describe containerset retrystrategy and verify it works. Fixes: #11502

### LINKED ISSUE #11502: ContainerSet Retry strategy does nothing?
### Pre-requisites

Hi,

According to [this closed issue](https://github.com/argoproj/argo-workflows/issues/7290) we are supposed to be able to configure a retry strategy that applies to containers in a `ContainerSet`, not to the whole `Pod`.

After reading the documentation, i tried the following two configuration:

1. specifying a `retryStategy` with a `limit` of `2` at the template level (same level as `volumes`, `containerset`, `metadata`, etc.)
2. specifying a`retryStategy` with a `retries` of `2` at the containerset level (as specified in the above mentioned issue)

In case 1, the retry strategy applies to the entire `Pod`, meaning that the pod is restarted as any other container templates. The consequence is that all containers within this containerset are restarted, not just the container that failed.
image

In case 2, no retry strategy is happening.
image

So, while the case 1 seems to work as expected (retrying all containers), the case 2 seems to retry nothing.
Checking logs, a single error log is logged, which probably validate the fact that no retry is happening.

Questions:

- do you really support restarting / retry strategy on a per container basis for `containerset` ?
- If yes, could you please indicate some clear documentation or workflow example that illustrate it ?

### Version

3.4.8

```YAML
kind: WorkflowTemplate
metadata:
  name: bug
  annotations:
      This workflow template is used to reproduce a bug.
spec:
  entrypoint: test
  templates:

    - name: test
      containerSet:
        retryStrategy:
          retries: "2"
        containers:
          - name: c1
            image: python:alpine3.6
            command:
              - python
              - -c
            args:
              - |
                print("hi")
          - name: c2
            image: python:alpine3.6
            command:
              - invalid
              - command
```

```text
```

```text
```

shuangkun:
Fixes #11502 

### Motivation

Show how containerset retrystrategy works. 

### Verification

e2e test


tczhao (changes requested):
Maybe add a note here
- https://github.com/argoproj/argo-workflows/blob/v3.5.5/pkg/apis/workflow/v1alpha1/container_set_template_types.go#L16
- https://github.com/argoproj/argo-workflows/blob/v3.5.5/docs/container-set-template.md

so that user can understand the behaviour from docs instead of going through the issue and PR

shuangkun:
> Maybe add a note here
> 
> * https://github.com/argoproj/argo-workflows/blob/v3.5.5/pkg/apis/workflow/v1alpha1/container_set_template_types.go#L16
> * https://github.com/argoproj/argo-workflows/blob/v3.5.5/docs/container-set-template.md
> 
> so that user can understand the behaviour from docs instead of going through the issue and PR

Yes, it needs to be described more clearly in the comments and documentation. I will add more later.  🙏

docs/container-set-template.md
```diff
@@ -116,3 +116,11 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
 Can you see the problem here? `a` only has small requests, but the container set will use the  total of all requests. So it's as if you're using all that GPU for 10h. This will be expensive.
 
 Solution: do not use container set when you have lopsided requests.
+
+## Container Set Retries
```
tczhao, line 120: we should have something here stating this is available from v3.3
https://github.com/argoproj/argo-workflows/commit/808c561f1c4a56668c32caa69be5b0441d372610

and give a little example would be good.

docs/container-set-template.md
```diff
@@ -116,3 +116,11 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+
+Container Set Retry policies describes how to retry a container nodes in the container set if it fails.
+
+Number of retries(default 0) and sleep duration between retries(default 0s, instant retry) can be set.
+
+The shell command should be normal otherwise it will not be retried.
```
tczhao, line 126: I feel the more precise description is: the container won't retry if it's unable to locate the command. What do you think?

docs/fields.md
```diff
@@ -2411,7 +2411,7 @@ _No description available_
 | Field Name | Field Type | Description   |
 |:----------:|:----------:|---------------|
 |`containers`|`Array<`[`ContainerNode`](#containernode)`>`|_No description available_|
-|`retryStrategy`|[`ContainerSetRetryStrategy`](#containersetretrystrategy)|RetryStrategy describes how to retry a container nodes in the container set if it fails. Nbr of retries(default 0) and sleep duration between retries(default 0s, instant retry) can be set.|
+|`retryStrategy`|[`ContainerSetRetryStrategy`](#containersetretrystrategy)|RetryStrategy describes how to retry a container nodes in the container set if it fails. Nbr of retries(default 0) and sleep duration between retries(default 0s, instant retry) can be set. The shell command should work fine otherwise it won't be retried.|
```
tczhao, line 2414: I feel the more precise description is: the container won't retry if it's unable to locate the command. What do you think?

tczhao (changes requested):

docs/container-set-template.md
shuangkun: OK，I add a example and version.Thanks!

docs/fields.md
shuangkun: OK, I change it.

shuangkun:
> I should've caught this earlier, I think the test should be somewhere else instead of `cli`. Imagine in CI when there's failure in `cli test` but totally unrelated to `cli`, this doesn't sound right. What do you think?

Yes, for this test, I was not in the cli at earlier, but it was not easy to get the logs. Apart from the logs, there seemed to be no other good way to judge whether a retry was performed. So I added it here, I will think of other ways.

tczhao (approved):

shuangkun:
Hi @terrytangyuan I see the issue you have seen before. Can you help me take a look at this PR? Thanks.

agilgur5:
> Hi @terrytangyuan I see the issue you have seen before. Can you help me take a look at this PR? Thanks.

I can take a look as I had glanced over this and the issue last week and so have some context

docs/container-set-template.md
```diff
@@ -116,3 +116,26 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+
+> v3.3 and after
+
+Container Set Retry policies describes how to retry a container nodes in the container set if it fails.
+
+Number of retries(default 0) and sleep duration between retries(default 0s, instant retry) can be set.
+
+The container won't retry if it's unable to locate the command.
+
+Here is an example of a Container Set Template with `retryStrategy`:
```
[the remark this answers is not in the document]
agilgur5: Ok I had to re-read things as the incorrect example (see below) changes quite a bit:

```suggestion
You can set a `retryStrategy` to apply to all containers of a container set.

This currently works differently from [retries](retries.md) for other template types.
You can only set the `duration` between each retry and the total number of `retries`.

See an example below:
```

More clearly explain the difference, but let the example do the talking.

Also I see you copied the comments from the spec here -- generally reference-level comments are more specific to low-level implementation details (like defaults), whereas user-facing docs are higher-level describing general usage.

docs/container-set-template.md
```diff
@@ -116,3 +116,26 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+```yaml
+      containerSet:
+        containers:
+          - name: retry-containerset
+            image: alpine:latest
+            retryStrategy:
+              limit: "3"
+            command: [ sh, -c ]
+            args: [ "echo intentional failure; exit 1" ]
+```
```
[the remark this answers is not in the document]
agilgur5: Modifying this since the example was incorrect:

```suggestion
\```

!!! Note
    A container set will not be retried if a container's `command` cannot be located.
    As it will fail each time, the retry logic is short-circuited.
```
- should specify that the _entire_ container set is not retried if a _single_ container's `command` cannot be located. That should answer the question in the issue

pkg/apis/workflow/v1alpha1/container_set_template_types.go
```diff
@@ -14,6 +14,7 @@ type ContainerSetTemplate struct {
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty" protobuf:"bytes,3,rep,name=volumeMounts"`
 	// RetryStrategy describes how to retry a container nodes in the container set if it fails.
 	// Nbr of retries(default 0) and sleep duration between retries(default 0s, instant retry) can be set.
+	// The container won't retry if it's unable to locate the command.
```
agilgur5, line 17:
```suggestion
	// RetryStrategy describes how to retry container nodes if the container set fails.
	// Note that the container set will not be retried if a container's `command` cannot be located, as it will fail each time.
```

- This should match the other template types' `retryStrategy` [description](https://github.com/argoproj/argo-workflows/blob/24f791b9f5d9c3d946670366e1c86d802ea48ce1/pkg/apis/workflow/v1alpha1/workflow_types.go#L687)
- sleep and number of retries are described in the struct below and so are not necessary here
- be more specific as to what won't retry (it's the _entire_ container set)
  - also make that into a "Note" as it's more an aside than a key detail (and is not a spec nuance either)
  - same as above

pkg/apis/workflow/v1alpha1/container_set_template_types.go
```diff
@@ -14,6 +14,7 @@ type ContainerSetTemplate struct {
...
 	RetryStrategy *ContainerSetRetryStrategy `json:"retryStrategy,omitempty" protobuf:"bytes,5,opt,name=retryStrategy"`
 }
```
agilgur5, line 20:
```suggestion

// ContainerSetRetryStrategy provides controls on how to retry a container set
```

A comment should be added above this struct too as it will appear in the Fields reference as well -- it's [currently empty](https://argo-workflows.readthedocs.io/en/release-3.5/fields/#containersetretrystrategy):
Screenshot 2024-03-26 at 3 20 33 AM

This comment should similarly match the other template types' `retryStrategy` struct [description](https://github.com/argoproj/argo-workflows/blob/24f791b9f5d9c3d946670366e1c86d802ea48ce1/pkg/apis/workflow/v1alpha1/workflow_types.go#L2030C4-L2030C67)

agilgur5: Should also modify line 25 `// Nbr of retries` to match the [description of `limit`](https://github.com/argoproj/argo-workflows/blob/24f791b9f5d9c3d946670366e1c86d802ea48ce1/pkg/apis/workflow/v1alpha1/workflow_types.go#L2034).

Honestly it should also be renamed, but that's a whole separate discussion -- the entire spec should ideally be identical to other template types instead of being its own separate thing.

agilgur5 (changes requested):
The example here is both [incorrect](https://github.com/argoproj/argo-workflows/pull/12809#discussion_r1538666447) and does not really address the issue as it does not show multiple containers.

The docs also need a good bit of copy edits to properly address the issue, as copying the spec comments / Field Reference is both insufficient (those already existed) and unclear. More generally speaking, merely copying comments is not sufficient for docs.

The incorrect example and insufficient descriptions also horribly confused me (see all the follow-up comments below), which demonstrates just how confusing this is, to the credit of #11502 and its author. `containerSets` are already confusing to many users, so this really adds to the problem.

<hr> 
Honestly, this whole spec should probably be rewritten at some point to match the other template types' `retryStrategy`. That would be a breaking change though, but worthwhile if we keep `containerSets` around (based on all the user confusion, I'm not sure they solve the problem that well in general -- #8786 is perhaps a more apt solution?). We can also do that in a backward-compatible way and deprecate the old `retryStrategy` as none of the fields are named the same (except for `retryStrategy` itself, unfortunately -- might want to consider a `retryStrategyV2` or something in this case).

docs/container-set-template.md
```diff
@@ -116,3 +116,52 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+See an example below:
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  name: containerset-with-retrystrategy
+  annotations:
+    workflows.argoproj.io/description: |
+      This workflow template is used to create a workflow with containerset with retrystrategy.
```
agilgur5, line 138:
```suggestion
      This workflow creates a container set with a retryStrategy.
```

- This isn't a `WorkflowTemplate`
- Also should use the same spelling as elsewhere in this doc for consistency:
  - "containerset" -> "container set"
  - "retrystrategy" -> "retryStrategy"

Although I'm not sure this annotation is needed at all, since this isn't a standalone example manifest, it's in the context of docs that already explain this.

agilgur5: I see you just copied your testdata manifest here, but they shouldn't necessarily be the same, as the purposes are different. One is for explanation / demonstration, the other is for testing.

test/e2e/testdata/workflow-template-with-containerset.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: containerset-with-retrystrategy
+  annotations:
+    workflows.argoproj.io/description: |
+      This workflow template is used to create a workflow with containerset.
```
agilgur5, line 7:
```suggestion
      This workflow template creates a container set with retryStrategy.
```

- same as above

docs/container-set-template.md
```diff
@@ -116,3 +116,52 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+spec:
+  entrypoint: containerset-retrystrategy-example
+  templates:
+    - name: containerset-retrystrategy-example
+      containerSet:
+        retryStrategy:
+          retries: "2"
```
agilgur5, line 145: this should include both `retries` and `duration` and have comments describing those two options:

```suggestion
        retryStrategy:
          retries: "2" # if fails, retry at most twice
          duration: 30s # retry for at most 30s
```

docs/container-set-template.md
```diff
@@ -116,3 +116,52 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+                print("hi")
+          - name: fail-retry
+            image: alpine:latest
+            command: [ sh, -c ]
+            args: [ "echo intentional failure; exit 1" ]
+          - name: invalic-command
+            image: python:alpine3.6
+            command:
+              - invalid
+              - command
```
agilgur5, line 163:
```suggestion
```

This is the edge case (that has a note _below_ the example) and not expected behavior, so we actually shouldn't show that. We want to demonstrate the normal, expected usage.

docs/container-set-template.md
```diff
@@ -116,3 +116,52 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+        containers:
+          - name: success
+            image: python:alpine3.6
+            command:
+              - python
+              - -c
+            args:
+              - |
+                print("hi")
+          - name: fail-retry
```
agilgur5, line 155:
```suggestion
          # this container will fail, causing the whole container set to get retried twice (both of which will also fail)
          - name: fail-retry
```
explain what this does

docs/container-set-template.md
```diff
@@ -116,3 +116,52 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+      This workflow template is used to create a workflow with containerset with retrystrategy.
+spec:
+  entrypoint: containerset-retrystrategy-example
+  templates:
+    - name: containerset-retrystrategy-example
+      containerSet:
+        retryStrategy:
+          retries: "2"
+        containers:
+          - name: success
```
agilgur5, line 147:
```suggestion
          # this container completes successfully, but it will be retried as another container in the container set fails
          - name: success
```

similarly explain _why_ this will retry as well

docs/container-set-template.md
agilgur5: on second thought, intentional failure is maybe not a good example, as it will never show what a successful retry looks like. the `retryStrategy` docs actually do [a probabilistic failure](https://github.com/argoproj/argo-workflows/blob/24f791b9f5d9c3d946670366e1c86d802ea48ce1/docs/retries.md?plain=1#L22) --  let's do the same here, `retries: 10` (same as the `limit` there) and the same code for the container

docs/container-set-template.md
```diff
@@ -116,3 +116,52 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+            args: [ "echo intentional failure; exit 1" ]
+          - name: invalic-command
+            image: python:alpine3.6
+            command:
+              - invalid
+              - command
+```
+
+!!! NOTE
+    A container set will not be retried if a container's `command` cannot be located. As it will fail each time, the retry logic is short-circuited.
```
agilgur5, line 167:
```suggestion
!!! Note
    A container set will not be retried if a container's `command` cannot be located.
    As it will fail each time, the retry logic is short-circuited.
```

same as my [previous comment](https://github.com/argoproj/argo-workflows/pull/12809/files#r1538681610)
- Use title case "Note" for consistency with other areas of the docs
- Prefer 1 sentence per line of markdown, per [style guide](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/)

pkg/apis/workflow/v1alpha1/container_set_template_types.go
```diff
@@ -12,16 +12,17 @@ import (
...
 
+// ContainerSetRetryStrategy provides controls on how to retry a container set
 type ContainerSetRetryStrategy struct {
 	// Duration is the time between each retry, examples values are "300ms", "1s" or "5m".
 	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 	Duration string `json:"duration,omitempty" protobuf:"bytes,1,opt,name=duration"`
-	// Nbr of retries
+	// Retries is the number of retry attempts when retrying a container.
```
agilgur5, line 25: `Limit` has a more detailed and precise description, that's why I linked it before:

```suggestion
	// Retries is the maximum number of retry attempts for the container set. It does not include the
	// first, original attempt; the maximum number of total attempts will be `retries + 1`.
```

agilgur5 (commented):
This is _a lot_ better. The example could use more improvements with comments and demonstrations of expected behavior

docs/container-set-template.md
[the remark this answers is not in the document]
agilgur5: Follow-up [comment](https://github.com/argoproj/argo-workflows/pull/12809#discussion_r1539771062) on style issues

pkg/apis/workflow/v1alpha1/container_set_template_types.go
agilgur5: Follow up [comment](https://github.com/argoproj/argo-workflows/pull/12809#discussion_r1539774685) for `retries` description

shuangkun:
Hi, @agilgur5  I made some modifications, can you check it last time, thanks!

docs/container-set-template.md
```diff
@@ -116,3 +116,51 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+              - python
+              - -c
+            args:
+              - |
+                print("hi")
+          # if fails, it will retry at most ten times.    
+          - name: fail-retry
+            image: python:alpine3.6
+            command: ["python", -c]
+            args: ["import random; import sys; exit_code = random.choice([0, 1, 1]); sys.exit(exit_code)"]
```
agilgur5, line 161:
```suggestion
            # fail with a 66% probability
            args: ["import random; import sys; exit_code = random.choice([0, 1, 1]); sys.exit(exit_code)"]
```

the [same comment](https://github.com/argoproj/argo-workflows/blob/24f791b9f5d9c3d946670366e1c86d802ea48ce1/docs/retries.md?plain=1#L21) from the `retryStrategy` doc is helpful here too to summarize the one-liner

agilgur5 (commented):
Two small changes + a clarifying question on successful containers that is important to eliminate remaining confusion from the issue

docs/container-set-template.md
```diff
@@ -116,3 +116,51 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+spec:
+  entrypoint: containerset-retrystrategy-example
+  templates:
+    - name: containerset-retrystrategy-example
+      containerSet:
+        retryStrategy:
+          retries: "10" # if fails, retry at most ten times
+          duration: 30s # retry for at most 30s
+        containers:
+          # this container completes successfully, it won't retried.
```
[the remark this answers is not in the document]
shuangkun: Yes according to my test there is no new node here， just retry in emissary.  This may be what confuses a lot of people, hopefully this PR will explain it.

The c3 retry twice, and c1 c2 no retry.
```
time="2024-03-29T06:40:25 UTC" level=info msg="capturing logs" argo=true
intentional failure
time="2024-03-29T06:40:25 UTC" level=debug msg="ignore signal child exited" argo=true
time="2024-03-29T06:40:26 UTC" level=info msg="capturing logs" argo=true
time="2024-03-29T06:40:26 UTC" level=debug msg="ignore signal urgent I/O condition" argo=true
intentional failure
time="2024-03-29T06:40:26 UTC" level=debug msg="ignore signal child exited" argo=true
time="2024-03-29T06:40:26 UTC" level=debug msg="forwarding signal terminated" argo=true
time="2024-03-29T06:40:27 UTC" level=info msg="sub-process exited" argo=true error="<nil>"
time="2024-03-29T06:40:27 UTC" level=info msg="not saving outputs - not main container" argo=true
Error: exit status 1
```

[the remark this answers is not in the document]
shuangkun: The UI is:
image

shuangkun:
Test TestStoppedWorkflow is really unstable, so I have to trigger the test repeatedly.

docs/container-set-template.md
agilgur5: apparently this doesn't retry? see follow-up discussion: https://github.com/argoproj/argo-workflows/pull/12809#discussion_r1544134648

docs/container-set-template.md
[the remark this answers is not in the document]
agilgur5: > The UI is:

Ohhh I got confused by the screenshots in the [issue](https://github.com/argoproj/argo-workflows/issues/11502) that have new nodes, but that's because they used a regular template-level `retryStrategy` in their first screenshot:

> 1. specifying a `retryStategy` with a `limit` of `2` at the template level (same level as `volumes`, `containerset`, `metadata`, etc.)

Thanks for clarifying this!

> Yes according to my test there is no new node here， just retry in emissary. 

It could be useful if Emissary could report the retried `node`s although that may technically require a new node type, since the `Container` itself does not get retried, the _process_ does.

> This may be what confuses a lot of people, hopefully this PR will explain it.

Yes, I think we should note this behavior explicitly in the docs. Let me think about how to best phrase that. Showing the logs as you did here may perhaps be the best way to visualize that (while also giving example logs for a user to check against)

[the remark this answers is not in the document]
shuangkun: OK,Thanks!

agilgur5:
> Test TestStoppedWorkflow is really unstable, so I have to trigger the test repeatedly.

I wrote up #12864 for us to enable a `/retest` comment that will more efficiently only re-run failed GH jobs. In the interim, an Approver+ can manually re-run failed jobs.
So for this PR, for which the logic all passes, if E2E tests fail again, I can rerun myself, no need to open/close to retrigger. It [seems to always pass](https://cloud-native.slack.com/archives/C0510EUH90V/p1711254624694649?thread_ts=1710260573.672209&cid=C0510EUH90V) on the first rerun as well.
Same for any other PRs I'm reviewing, I can do that and have been doing that when I notice it.

docs/container-set-template.md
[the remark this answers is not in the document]
shuangkun: I add a note to describe the behavior, can you have a finnal check, thanks! @agilgur5

agilgur5:
> /retest

#12864 has not been implemented yet, so that won't work yet.

> @[shuangkun](https://github.com/shuangkun) shuangkun closed this [4 days ago](https://github.com/argoproj/argo-workflows/pull/12809#event-12302969990)
> @[shuangkun](https://github.com/shuangkun) shuangkun reopened this [4 days ago](https://github.com/argoproj/argo-workflows/pull/12809#event-12302970735)

As I said in my [previous comment](https://github.com/argoproj/argo-workflows/pull/12809#issuecomment-2028441339), I will retry CI myself when this is ready to merge. No need to spend time getting E2Es to pass in the mean-time. Just need to make sure docs CI passes with each docs change.

docs/container-set-template.md
```diff
@@ -116,3 +116,54 @@ Example B: Lopsided requests, e.g. `a -> b` where `a` is cheap and `b` is expens
...
+            command: ["python", -c]
+            # fail with a 66% probability
+            args: ["import random; import sys; exit_code = random.choice([0, 1, 1]); sys.exit(exit_code)"]
+```
+
+!!! Note
+    A container set will not be retried if a container's `command` cannot be located.
+    As it will fail each time, the retry logic is short-circuited.
+    Since it is only a process-level retry, so won't create new node.
+    You can confirm whether a retry has been performed by checking the container's log.
```
agilgur5, line 169:
```suggestion
!!! Note Template-level `retryStrategy` vs Container Set `retryStrategy`
    `containerSet.retryStrategy` works differently from [template-level retries](retries.md):
    1. Your `command` will be re-ran by the Executor inside the same container if it fails.
        - As no new containers are created, the nodes in the UI remain the same, and the retried logs are appended to original container's logs. For example, your container logs may look like:
        ```
        time="2024-03-29T06:40:25 UTC" level=info msg="capturing logs" argo=true
        intentional failure
        time="2024-03-29T06:40:25 UTC" level=debug msg="ignore signal child exited" argo=true
        time="2024-03-29T06:40:26 UTC" level=info msg="capturing logs" argo=true
        time="2024-03-29T06:40:26 UTC" level=debug msg="ignore signal urgent I/O condition" argo=true
        intentional failure
        time="2024-03-29T06:40:26 UTC" level=debug msg="ignore signal child exited" argo=true
        time="2024-03-29T06:40:26 UTC" level=debug msg="forwarding signal terminated" argo=true
        time="2024-03-29T06:40:27 UTC" level=info msg="sub-process exited" argo=true error="<nil>"
        time="2024-03-29T06:40:27 UTC" level=info msg="not saving outputs - not main container" argo=true
        Error: exit status 1
        ```
    1. If a container's `command` cannot be located, it will not be retried.
        - As it will fail each time, the retry logic is short-circuited.
```

- add a title to the note as there's a common subject now between these
- move the "different from template-level retries" part into this note, including the link to `retries.md`
- be more specific about the "nodes" in the UI / `status` (many users don't know what "nodes" mean as it's part of the `status` rather than user-facing per se) as well as what happens with the logs
  - add an example of the logs for optimal clarity

pkg/apis/workflow/v1alpha1/container_set_template_types.go
```diff
@@ -12,16 +12,18 @@ import (
 type ContainerSetTemplate struct {
 	Containers   []ContainerNode      `json:"containers" protobuf:"bytes,4,rep,name=containers"`
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty" protobuf:"bytes,3,rep,name=volumeMounts"`
-	// RetryStrategy describes how to retry a container nodes in the container set if it fails.
-	// Nbr of retries(default 0) and sleep duration between retries(default 0s, instant retry) can be set.
+	// RetryStrategy describes how to retry container nodes if the container set fails.
+	// Note that the container set will not be retried if a container's `command` cannot be located, as it will fail each time.
```
agilgur5, line 16:
```suggestion
	// Note that this works differently from the template-level `retryStrategy` as it is a process-level retry that does not create new Pods or containers.
```

changing this to account for the other differences. we can include more implementation details like "process-level retry" in API docs/comments. And don't need to be very specific here, as the code and the "Note" in the docs go into the rest

agilgur5 (commented):
A few final changes here to account for the differences between a template-level `retryStrategy` and the inner `containerSet.retryStrategy`

docs/container-set-template.md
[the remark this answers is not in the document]
agilgur5: Follow-up (hopefully final) review [below](https://github.com/argoproj/argo-workflows/pull/12809#pullrequestreview-1970042774)

shuangkun:
I modified it and fixed errors. Thanks!

agilgur5:
I pushed [a commit](https://github.com/argoproj/argo-workflows/pull/12809/commits/02ce8daf86fda50c244f4e0a387942cec422b1d7) myself to fix the `mkdocs-material` admonition syntax with markdown inside of it -- it requires a `markdownlint-ignore` plus some proper newlines. That and forgot that the "Note"'s title has to be in quotes. It looks like this now (as intended):

Screenshot 2024-04-04 at 1 44 46 AM

agilgur5 (approved):
Thanks for adding these docs and iterating on this!

shuangkun:
> I pushed [a commit](https://github.com/argoproj/argo-workflows/pull/12809/commits/02ce8daf86fda50c244f4e0a387942cec422b1d7) myself to fix the `mkdocs-material` admonition syntax with markdown inside of it as it's a bit complicated -- it requires a `markdownlint-disable` comment plus some proper newlines. That and I forgot that the "Note"'s title has to be in quotes. It looks like this now (as intended):
> 
> Screenshot 2024-04-04 at 1 44 46 AM

OK,Thanks！

Issue #11395, opened 2023-07-19
DAG with continueOn in error after retry


fstaudt:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I created following DAG with some tasks that should be bypassed if they fail.\
I used `continueOn` keyword on DAG task for this purpose.\
I also tried to use `depends` keyword on DAG tasks and got the same result.

First execution of workflow looks like this (it is the expected result):
![image](https://github.com/argoproj/argo-workflows/assets/30255959/eb551a76-9a60-4c97-83a4-d1cfd303c80d)

When I retry this workflow, the DAG workflow is in Error with following message:\
`Ancestor task node continue not found`

Workflow after retry looks like this (not expected):
![image](https://github.com/argoproj/argo-workflows/assets/30255959/907ee6e2-0346-48d8-9174-7293ea9dc358)

What I expect after retry is that:
- failed tasks with dependent tasks already executed remain failed in DAG
- failed tasks without dependent tasks already executed are retried

### Version

v3.4.7

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: "dag-to-retry"
spec:
  podMetadata:
    annotations:
      sidecar.istio.io/inject: "false"
  entrypoint: dag
  templates:
    - name: step
      inputs:
        parameters:
          - name: exitCode
      container:
        image: alpine:3.7
        command: [ sh, "-c", "exit {{inputs.parameters.exitCode}}" ]
    - name: dag
      dag:
        failFast: false
        tasks:
          - name: success
            template: step
            arguments:
              parameters:
                - name: exitCode
                  value: 0
          - name: failure
            template: step
            dependencies:
              - success
            arguments:
              parameters:
                - name: exitCode
                  value: 1
          - name: task-after-failure
            template: step
            dependencies:
              - failure
            arguments:
              parameters:
                - name: exitCode
                  value: 0
          - name: continue
            template: step
            continueOn:
              failed: true
            dependencies:
              - success
            arguments:
              parameters:
                - name: exitCode
                  value: 2
          - name: task-after-continue
[...]

Issue #12156, opened 2023-11-06
Retry a stopped workflow report Ancestor node not found


hustclf:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

**What happened**
I created a DAG workflow with four tasks (A, B, C, D). When tasks A and B were successfully completed, I stopped the workflow during the execution of task C. After retrying the workflow, I noticed that the workflow's phase immediately changed to 'Error' and displayed an error message stating that the ancestor node 'C' was not found. 

<img width="1717" alt="image" src="https://github.com/argoproj/argo-workflows/assets/13963737/0713b965-0c23-447a-b3d4-7ffb942b2f2b">

**what you expected to happen?**
I would like to be able to successfully retry a stopped workflow.

### Version

v3.4.13

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: retry-stopped-dag
spec:
  entrypoint: diamond
  templates:
  - name: echo
    inputs:
      parameters:
      - name: message
    container:
      image: alpine:3.7
      command: [echo, "{{inputs.parameters.message}}"]
  - name: sleep
    inputs:
      parameters:
      - name: message
    container:
      image: alpine:3.7
      command: [sleep, 36000s]
  - name: diamond
    dag:
      tasks:
      - name: A
        template: echo
        arguments:
          parameters: [{name: message, value: A}]
      - name: B
        depends: A.Succeeded
        template: echo
        arguments:
          parameters: [{name: message, value: B}]
      - name: C
        depends: B.Succeeded
        template: sleep
        arguments:
          parameters: [{name: message, value: C}]
      - name: D
        depends: C.Succeeded || C.Failed
        template: echo
        arguments:
          parameters: [{name: message, value: D}]
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}

run workflow and stop it.

time="2023-11-06T17:08:09.524Z" level=info msg="cleaning up pod" action=labelPodCompleted key=argo/retry-stopped-dag-sleep-189432235/labelPodCompleted
[...]

PR #12817, opened 2024-03-18
fix: DAG with continueOn in error after retry. Fixes: #11395

shuangkun:
fix: DAG with continueOn in error after retry

 Fixes: #11395

### Verification

local test and e2e test.

After retry:

Before fix: Error and lose some nodes.
image
After fix: Failed and not lose nodes
image

### DIFF
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -1572,44 +1572,6 @@ func (s *ArgoServerSuite) TestRetryStoppedButIncompleteWorkflow() {
 	})
 }
 
-func (s *ArgoServerSuite) TestRetryWorkflowWithContinueOn() {
-	var workflowName string
-	s.Given().
-		Workflow(`@testdata/retry-workflow-with-continueon.yaml`).
-		When().
-		SubmitWorkflow().
-		WaitForWorkflow(fixtures.ToBeFailed).
-		Then().
-		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
-			workflowName = metadata.Name
-			assert.Equal(t, 6, len(status.Nodes))
-		})
-
-	s.Run("Retry", func() {
-		s.e().PUT("/api/v1/workflows/argo/{workflowName}/retry", workflowName).
-			Expect().
-			Status(200).
-			JSON().
-			Path("$.metadata.name").
-			NotNull()
-	})
-
-	s.Given().
-		When().
-		WaitForWorkflow(fixtures.ToBeCompleted).
-		Then().
-		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
-			workflowName = metadata.Name
-			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
-			assert.Equal(t, 6, len(status.Nodes))
-		}).
-		ExpectWorkflowNode(func(status wfv1.NodeStatus) bool {
-			return strings.Contains(status.Name, "retry-workflow-with-continueon.success")
-		}, func(t *testing.T, status *wfv1.NodeStatus, pod *corev1.Pod) {
-			assert.Equal(t, 2, len(status.Children))
-		})
-}
-
 func (s *ArgoServerSuite) TestWorkflowTemplateService() {
 	s.Run("Lint", func() {
 		s.e().POST("/api/v1/workflow-templates/argo/lint").
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -1582,6 +1582,7 @@ func (s *ArgoServerSuite) TestRetryWorkflowWithContinueOn() {
 		Then().
 		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
 			workflowName = metadata.Name
+			assert.Equal(t, 6, len(status.Nodes))
 		})
 
 	s.Run("Retry", func() {
@@ -1603,7 +1604,7 @@ func (s *ArgoServerSuite) TestRetryWorkflowWithContinueOn() {
 			assert.Equal(t, 6, len(status.Nodes))
 		}).
 		ExpectWorkflowNode(func(status wfv1.NodeStatus) bool {
-			return strings.Contains(status.Name, "dag-to-retry.success")
+			return strings.Contains(status.Name, "retry-workflow-with-continueon.success")
 		}, func(t *testing.T, status *wfv1.NodeStatus, pod *corev1.Pod) {
 			assert.Equal(t, 2, len(status.Children))
 		})
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -1572,6 +1572,43 @@ func (s *ArgoServerSuite) TestRetryStoppedButIncompleteWorkflow() {
 	})
 }
 
+func (s *ArgoServerSuite) TestRetryWorkflowWithContinueOn() {
+	var workflowName string
+	s.Given().
+		Workflow(`@testdata/retry-workflow-with-continueon.yaml`).
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeFailed).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			workflowName = metadata.Name
+		})
+
+	s.Run("Retry", func() {
+		s.e().PUT("/api/v1/workflows/argo/{workflowName}/retry", workflowName).
+			Expect().
+			Status(200).
+			JSON().
+			Path("$.metadata.name").
+			NotNull()
+	})
+
+	s.Given().
+		When().
+		WaitForWorkflow(fixtures.ToBeCompleted).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			workflowName = metadata.Name
+			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
+			assert.Equal(t, 6, len(status.Nodes))
+		}).
+		ExpectWorkflowNode(func(status wfv1.NodeStatus) bool {
+			return strings.Contains(status.Name, "dag-to-retry.success")
+		}, func(t *testing.T, status *wfv1.NodeStatus, pod *corev1.Pod) {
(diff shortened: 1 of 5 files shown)


workflow/util/util.go
```diff
@@ -772,6 +772,20 @@ func getDescendantNodeIDs(wf *wfv1.Workflow, node wfv1.NodeStatus) []string {
 	return descendantNodeIDs
 }
 
+func isDescendantNodeSuccessfully(wf *wfv1.Workflow, node wfv1.NodeStatus) bool {
```
tczhao, line 775: `isDescendantNodeSucceeded` probably makes more sense

workflow/util/util.go
```diff
@@ -960,6 +974,11 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 				log.Debugf("Reset %s node %s since it's a group node", node.Name, string(node.Phase))
 				continue
 			} else {
+				if isDescendantNodeSuccessfully(wf, node) {
+					log.Debugf("Node %s remains as is since it has succeed nodes.", node.Name)
```
tczhao, line 978: `log.Debugf("Node %s remains as is since it has succeed child nodes.", node.Name)`

tczhao (changes requested):
Could we add something in `util_test.go`?
I feel we rely too much on e2e, would this be an issue?
Remember to test if the correct node is retried, it's not included in the current test

shuangkun:
> Could we add something in `util_test.go`? I feel we rely too much on e2e, would this be an issue? Remember to test if the correct node is retried, it's not included in the current test

I add a ut in util_test.go and test correct node retried. thanks!

shuangkun:
Hi, @tczhao  can you take a look again? Thanks!

test/e2e/argo_server_test.go
```diff
@@ -1572,6 +1572,43 @@ func (s *ArgoServerSuite) TestRetryStoppedButIncompleteWorkflow() {
...
+			NotNull()
+	})
+
+	s.Given().
+		When().
+		WaitForWorkflow(fixtures.ToBeCompleted).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			workflowName = metadata.Name
+			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
```
tczhao, line 1602: note: this validates the workflow is not errored (the original issue)

test/e2e/argo_server_test.go
```diff
@@ -1572,6 +1572,43 @@ func (s *ArgoServerSuite) TestRetryStoppedButIncompleteWorkflow() {
...
+func (s *ArgoServerSuite) TestRetryWorkflowWithContinueOn() {
+	var workflowName string
+	s.Given().
+		Workflow(`@testdata/retry-workflow-with-continueon.yaml`).
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeFailed).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			workflowName = metadata.Name
```
tczhao, line 1584: shall we validate the node count here as well, so we know retry produces an identical result?

test/e2e/argo_server_test.go
tczhao, line 1: the existing retry tests are in
`cli_test.go` and `retry_test.go`.

I think this test makes the most sense to be in the `cli_test.go`, since we manually retry the workflow

test/e2e/testdata/retry-workflow-with-continueon.yaml
tczhao, line 1: we need to clean up the testdata here,
there are a few irrelevant lines from example on original github issue 

---

Also, see https://github.com/argoproj/argo-workflows/blob/main/test/e2e/testdata/retry-nested-dag-test.yaml as an example as well,
I feel writing in this way is much cleaner, the reduced line per task make it much quicker to understand what the dag is doing, what do you think?

test/e2e/testdata/retry-workflow-with-continueon.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: "dag-to-retry"
```
tczhao, line 4: I would rename this to align with the filename

test/e2e/testdata/retry-workflow-with-continueon.yaml
```diff
@@ -0,0 +1,61 @@
...
+spec:
+  podMetadata:
+    annotations:
+      sidecar.istio.io/inject: "false"
```
tczhao, line 8: 🚮

workflow/util/util_test.go
```diff
@@ -2102,3 +2102,367 @@ func TestRetryWorkflowWithNestedDAGsWithSuspendNodes(t *testing.T) {
 	assert.Equal(t, wfv1.NodeSucceeded, wf.Status.Nodes.FindByName("fail-two-nested-dag-suspend.dag1-step4").Phase)
 	assert.Equal(t, 1, len(podsToDelete))
 }
+
+var failedWorkflowWithContinueOn = `
```
tczhao, line 2106: similar to the e2e test data, we want to clean up the test data here.
see the other example from the same file, https://github.com/argoproj/argo-workflows/blob/main/workflow/util/util_test.go#L766
we want a minimise version. (e.g. the artifactory has no use here)

workflow/util/util_test.go
```diff
@@ -2102,3 +2102,367 @@ func TestRetryWorkflowWithNestedDAGsWithSuspendNodes(t *testing.T) {
...
+      outputs: {}
+    workflowTemplateRef:
+      name: dag-to-retry
+  taskResultsCompletionStatus:
+    dag-to-retry-tb7r7-1325528633: true
+    dag-to-retry-tb7r7-1670055836: true
+    dag-to-retry-tb7r7-1819567448: true
+    dag-to-retry-tb7r7-1987291523: true`
+
+func TestRetryWorkflowWithContinurOnSucceededNodes(t *testing.T) {
```
tczhao, line 2448: There's a existing TestFormulateRetryWorkflow test with many different scenarios,
we should refactor this and add to it https://github.com/argoproj/argo-workflows/blob/5ca8b69fe7d8e890d1436ee5bac7dbd87ff10a88/workflow/util/util_test.go#L1340-L1366

tczhao (changes requested):
👍

test/e2e/argo_server_test.go
shuangkun: Yes，no Error after retry.

test/e2e/argo_server_test.go
shuangkun: Oh, what I often encounter is testing by clicking on the UI.

test/e2e/argo_server_test.go
shuangkun: OK

workflow/util/util_test.go
shuangkun: I deleted artifactory related

test/e2e/testdata/retry-workflow-with-continueon.yaml
shuangkun: renamed.

test/e2e/argo_server_test.go
shuangkun: And the verification of this test is relatively intuitive on the UI.

tczhao: sorry I meant the file location of this test,
I think we should group similar test into the same file

workflow/util/util_test.go
tczhao: My thinking is that,
for real workflow, we already have e2e test to handle it,
and here we use a simpler form more unit test-like
> In TestFormulateRetryWorkflow, The wf are all like this:

I think this works. Let me know what you think and if you need assistance

tczhao (changes requested):

shuangkun:
Thanks! Modified it!

tczhao (approved):
lgtm

terrytangyuan (approved):
Nice tests. Thanks!

agilgur5 (commented):
just looked this over, small comment.

@shuangkun you might want to look at #12156. I feel like there's potentially a deeper root cause here with the retry logic being buggy -- see also https://github.com/argoproj/argo-workflows/pull/12553#discussion_r1493314678

also wanted to say thanks to @tczhao for the initial review -- you might've noticed earlier but I thumbs-up'd pretty much all of your comments 🙂

workflow/util/util.go
```diff
@@ -772,6 +772,21 @@ func getDescendantNodeIDs(wf *wfv1.Workflow, node wfv1.NodeStatus) []string {
 	return descendantNodeIDs
 }
 
+func isDescendantNodeSucceeded(wf *wfv1.Workflow, node wfv1.NodeStatus, nodeIDsToReset map[string]bool) bool {
+	for _, child := range node.Children {
+		childStatus, err := wf.Status.Nodes.Get(child)
+		if err != nil {
+			log.Fatalf("Couldn't get child, panicking")
+			panic("Was not able to obtain child")
```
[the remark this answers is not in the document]
shuangkun: ok, I add a small pr to unify it.

shuangkun:
> just looked this over, small comment.
> 
> @shuangkun you might want to look at #12156. I feel like there's potentially a deeper root cause here with the retry logic being buggy -- see also [#12553 (comment)](https://github.com/argoproj/argo-workflows/pull/12553#discussion_r1493314678)

OK，I will have a look.

tczhao:
Thank you for the warm words @agilgur5 ,
I will comment in the other PR https://github.com/argoproj/argo-workflows/pull/12931

agilgur5:
Backported cleanly into `release-3.5` as 30f2e0d93

We may want to backport this to `release-3.4` as well, since #11395 was filed with 3.4.x

Joibel:
This has caused the regression in #13003, so I suggest not backporting until that is resolved because I feel that is a bigger regression than the thing this fixes.

agilgur5:
I really think the root cause I mentioned above https://github.com/argoproj/argo-workflows/pull/12817#pullrequestreview-1995821336 needs a deep dive. There's probably a refactor needed for the manual retry logic to correct all of the issues

JasonChen86899:
> I really think the root cause I mentioned above [#12817 (review)](https://github.com/argoproj/argo-workflows/pull/12817#pullrequestreview-1995821336) needs a deep dive. There's probably a refactor needed for the manual retry logic to correct all of the issues

There is an issue with this modification: the incorrect node was retained (with successful child nodes), causing the incorrect node to be unable to retry.

Issue #12565, opened 2024-01-23
Update outdated k8s go packages to 0.26+


Tomcli:
# Summary
There are list of go packages that are outdated and no longer provide security patches. We also want to make sure packages that shared with the KFP/KFP-Tekton deployment won't fall to behind as it will create a dependency issue.

List of Kubernetes client dependencies to update, ideally using 0.26+ as Kubernetes only provide security patches to the recent 4 major releases.
```
	k8s.io/api v0.24.3
	k8s.io/apimachinery v0.24.3
	k8s.io/cli-runtime v0.24.3
	k8s.io/client-go v0.24.3
        k8s.io/kubectl v0.24.3
```

List of dependencies that are nice to have, the grpc-gateway v1 no longer has security patches and openAPI v2 schema is out of support few years ago. 
```
	github.com/grpc-ecosystem/grpc-gateway v1.16.0
```

## Use Cases

When would you use this?

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritize the proposals with the most 👍.

agilgur5:
Note that for Argo 3.4 and 3.5, we do need to support some older k8s versions per the [support matrix](https://argo-workflows.readthedocs.io/en/release-3.5/releases/#kubernetes-compatibility-matrix), so any deps have to allow for some version skew as well

BabisK:
Would love to have the `grpc-gateway` updated to version 2. It breaks my build rigt now.

terrytangyuan:
@Tomcli Is this something you want to work on?

Tomcli:
Hi @terrytangyuan, I'm busy with grading KFP to the new golang and dependencies for the KF 1.9 release. I think for KF 1.9 our goal is first to upgrade Argo to 3.4. If anyone can take on this issue it will be great. Otherwise I can come back once I finished all the features/issues from KFP side.

terrytangyuan:
No worries. As long as this is not blocking KFP upgrade. We can leave this here for anyone who has time to take.

terrytangyuan:
Sounds good

blkperl:
I've opened https://github.com/argoproj/argo-workflows/pull/12847 to get us back to a version with security updates.

blkperl:
I've opened https://github.com/argoproj/argo-workflows/issues/12851 for the OpenAPI update as it seems like a large project

blkperl:
[...]

PR #6558, opened 2021-08-17
chore: Update k8s to v0.21.4

blkperl:
Signed-off-by: William Van Hevelingen <william.vanhevelingen@acquia.com>

Tips:

* Maybe add you organization to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md).
* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it does not need to pass
* Sign-off your commits to pass the DCO check: `git commit --signoff`.
* Run `make pre-commit -B` to fix codegen or lint problems. 
* Say how how you tested your changes. If you changed the UI, attach screenshots.
* If changes were requested, and you've made them, then dismis the review to get it looked at again.
* You can ask for help!

PR #12822, opened 2024-03-20
chore(deps): bump k8s dependencies to 1.25

blkperl:
### Motivation

This PR is part of series of PRs to get back to a supported version of K8s

### Modifications

* updated go.mod with the v0.25.16 version
* refreshed releases.md with the latest k8s versions 
* removed clusterName from tests as it's not valid in metadata
* the remaining changes were generated by make pre-commit 

### Verification

Existing tests should verify everything is working. We already have CI running on K8s 1.25


terrytangyuan (approved):
Thanks!

agilgur5:
Ah I had looked at this before, but hadn't finished reviewing. Terry beat me to it! Should probably assign myself next time around to reduce work duplication.

> This PR is part of series of PRs to get back to a supported version of K8s

Sounds like a partial fix for #12565 then. Would you like to work on that in general?

> Existing tests should verify everything is working. We already have CI running on K8s 1.25

For reference: https://github.com/argoproj/argo-workflows/blob/24f791b9f5d9c3d946670366e1c86d802ea48ce1/.github/workflows/ci-build.yaml#L207

> * refreshed releases.md with the latest k8s versions

Nice! now that we have versioned docs we can do so while still leaving prior version info in earlier docs versions

agilgur5 (commented):
@blkperl I just noticed that we have some codegen that apparently relies on the k8s version too: [`k8s.io/code-generator`](https://github.com/kubernetes/code-generator) on [these](https://github.com/argoproj/argo-workflows/blob/ac9cb3ae93bb441bb7ce09b74e99789adf599eae/Makefile#L627) [two](https://github.com/argoproj/argo-workflows/blob/ac9cb3ae93bb441bb7ce09b74e99789adf599eae/Makefile#L307) lines of the `Makefile`. Apparently the last time it was updated was in a 3 year old PR of yours #6558 😅 . So we may be missing some newer codegen in the Controller's Go CR clients (or maybe there are no significant changes in the codegen, 🤷 )

`controller-gen` might need updating too? although it's version is not the same as a k8s version afaict

PR #12825, opened 2024-03-20
fix: pass dnsconfig to agent pod. Fixes: #12824

### LINKED ISSUE #12824: Dnsconfig does not take effect in Agent pod
### Pre-requisites

I want to use a local dnsserver in Http template to report some info.
But dnsconfig does not take effect, the domain name cannot be resolved and the network is blocked.

## what i want

I want the agent pod also use workfow dnsConfig.

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: http-template-
spec:
  entrypoint: main
  dnsPolicy: "None"
  dnsConfig:
    nameservers:
      - 100.32.2.5
      - 100.33.3.5
  templates:
    - name: main
      steps:
        - - name: get-google-homepage
            template: http
            arguments:
              parameters: [{name: url, value: "https://www.google.com"}]
    - name: http
      inputs:
        parameters:
          - name: url
      http:
        timeoutSeconds: 20 # Default 30
        url: "{{inputs.parameters.url}}"
        method: "GET" # Default GET
        headers:
          - name: "x-header-name"
            value: "test-value"
        # Template will succeed if evaluated to true, otherwise will fail
        # Available variables:
        #  request.body: string, the request body
        #  request.headers: map[string][]string, the request headers
        #  response.url: string, the request url
        #  response.method: string, the request method
        #  response.statusCode: int, the response status code
        #  response.body: string, the response body
        #  response.headers: map[string][]string, the response headers
        successCondition: "response.body contains \"google\"" # available since v3.3
        body: "test body" # Change request body
```

```text
```

```text
```

shuangkun:
Fixes #12824

### Motivation

Pass dnsconfig to agent pod.

### Verification

Local Tests.


tczhao (changes requested):
make sense, since dnsconfig applies at the workflowspec level.

---

but mention this in the `type WorkflowSpec struct {`
`Set DNS policy for the pod and argo agent`

shuangkun:
> and argo agent

OK, Thanks.

tczhao (approved):

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -326,7 +326,7 @@ type WorkflowSpec struct {
 	// Host networking requested for this workflow pod. Default to false.
 	HostNetwork *bool `json:"hostNetwork,omitempty" protobuf:"bytes,14,opt,name=hostNetwork"`
 
-	// Set DNS policy for the pod.
+	// Set DNS policy for the pod and argo agent.
```
agilgur5, line 329:
```suggestion
	// Set DNS policy for workflow pods.
```

For consistency, the rest of the descriptions say either "workflow pods" or "all pods in the workflow", both of which would include the agent

shuangkun: Thanks！changed it

agilgur5 (approved):
LGTM, thanks for the fix!

PR #12830, opened 2024-03-21
fix: Enable realtime metric gc after its workflow is completed. Fixes #12790

### LINKED ISSUE #12790: Real-time metrics disappear while workflow is running when metricTTL is < workflow run time.
### Pre-requisites

Workflow controller configmap is configured as follows:

```yaml
  metricsConfig: |
    metricsTTL: "3m"
```

I created a workflow that is infinitely sleeping. The workflow template has a metric that just tracks the duration as a real-time gauge metric.

Expected:

The metric is continued to be reported on until the workflow terminates. Once the workflow terminates, the metric should live for another 3 minutes before being cleaned up.

This sounds like the expected behavior based on https://github.com/argoproj/argo-workflows/issues/10503#issuecomment-1481926351

What happened instead:

The metric stopped reporting shortly after the metric TTL while the workflow is still running.

Green = Duration Metric
Yellow = some prometheus metric for the workflow pod, indicating it is not deleted

image

I tested the ttl with various different timings (10m, 20m, 1h) and confirmed the behavior scales with the TTL, though it's not always exactly the ttl time setting (i.e. when ttl is 10m, it can take up to 20m).

We cannot just turn off the metric TTL as that would affect the non-realtime metrics.

I also confirmed that our Prometheus instance was up and running, and that the Argo controller did not shut down (I did observe separately that if the controller gets shut down and a new one comes up, the metrics would not be continued either, but that is a different scenario and seems like a different bug).

### Version

v3.4.10

```YAML
kind: WorkflowTemplate
metadata:
  name: sleep
  namespace: argo
spec:
  entrypoint: sleep
  templates:
    - name: sleep
      container:
        image: alpine:latest
        command: [sh, -c]
        args: ["while true; do sleep 100000; done"]
      metrics:
        prometheus:
        - name: duration_metric
          help: Duration of metric
          gauge:
            realtime: true
            value: '{{duration}}'
```

```text

{"level":"info","msg":"Processing workflow","namespace":"argo","time":"2024-03-12T21:53:59.861Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Updated phase  -\u003e Running","namespace":"argo","time":"2024-03-12T21:53:59.965Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Pod node sleep-8ftls initialized Pending","namespace":"argo","time":"2024-03-12T21:53:59.966Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Created pod: sleep-8ftls (sleep-8ftls)","namespace":"argo","time":"2024-03-12T21:54:00.067Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"TaskSet Reconciliation","namespace":"argo","time":"2024-03-12T21:54:00.067Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"reconcileAgentPod","namespace":"argo","time":"2024-03-12T21:54:00.067Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Workflow update successful","namespace":"argo","phase":"Running","resourceVersion":"398814","time":"2024-03-12T21:54:00.098Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Processing workflow","namespace":"argo","time":"2024-03-12T21:54:10.068Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Task-result reconciliation","namespace":"argo","numObjs":0,"time":"2024-03-12T21:54:10.068Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"node changed","namespace":"argo","new.message":"","new.phase":"Running","new.progress":"0/1","nodeID":"sleep-8ftls","old.message":"","old.phase":"Pending","old.progress":"0/1","time":"2024-03-12T21:54:10.069Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"TaskSet Reconciliation","namespace":"argo","time":"2024-03-12T21:54:10.069Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"reconcileAgentPod","namespace":"argo","time":"2024-03-12T21:54:10.069Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Workflow update successful","namespace":"argo","phase":"Running","resourceVersion":"399005","time":"2024-03-12T21:54:10.100Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Processing workflow","namespace":"argo","time":"2024-03-12T21:54:20.101Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"Task-result reconciliation","namespace":"argo","numObjs":0,"time":"2024-03-12T21:54:20.102Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"node unchanged","namespace":"argo","nodeID":"sleep-8ftls","time":"2024-03-12T21:54:20.102Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"TaskSet Reconciliation","namespace":"argo","time":"2024-03-12T21:54:20.102Z","workflow":"sleep-8ftls"}
{"level":"info","msg":"reconcileAgentPod","namespace":"argo","time":"2024-03-12T21:54:20.102Z","workflow":"sleep-8ftls"}
```

```text
{"level":"info","msg":"Starting Workflow Executor","time":"2024-03-12T21:54:02.642Z","version":"untagged"}
{"Duration":1000000000,"Factor":1.6,"Jitter":0.5,"Steps":5,"level":"info","msg":"Using executor retry strategy","time":"2024-03-12T21:54:02.645Z"}
{"deadline":"0001-01-01T00:00:00Z","includeScriptOutput":false,"level":"info","msg":"Executor initialized","namespace":"argo","podName":"sleep-8ftls","template":"{\"name\":\"sleep\",\"inputs\":{},\"outputs\":{},\"metadata\":{},\"container\":{\"name\":\"\",\"image\":\"alpine:latest\",\"command\":[\"sh\",\"-c\"],\"args\":[\"while true; do sleep 100000; done\"],\"resources\":{}},\"podSpecPatch\":\"{\\\"enableServiceLinks\\\": false}\\n\",\"metrics\":{\"prometheus\":[{\"name\":\"duration_metric\",\"help\":\"Duration of metric\",\"gauge\":{\"value\":\"{{duration}}\",\"realtime\":true}}]}}","time":"2024-03-12T21:54:02.645Z","version":"\u0026Version{Version:untagged,BuildDate:2024-03-01T06:42:25Z,GitCommit:unknown,GitTag:untagged,GitTreeState:clean,GoVersion:go1.21.7,Compiler:gc,Platform:linux/amd64,}"}
{"level":"info","msg":"Starting deadline monitor","time":"2024-03-12T21:54:02.645Z"}
{"level":"info","msg":"Alloc=6803 TotalAlloc=13592 Sys=20069 NumGC=6 Goroutines=7","time":"2024-03-12T21:59:02.646Z"}
```

jswxstw:
Fixes #12790 

### Motivation

Realtime metrics may be cleaned up, even if the workflow is still running.

### Modifications

Adding two boolean value: `realtime` and `complete` in metric to indicate whether the metric is realtime type and whether it is completed.
Enable realtime metric gc only after its workflow is completed.

### Verification

local test and UT

### DIFF
--- a/workflow/metrics/metrics.go
+++ b/workflow/metrics/metrics.go
@@ -42,6 +42,8 @@ func (s ServerConfig) SameServerAs(other ServerConfig) bool {
 type metric struct {
 	metric      prometheus.Metric
 	lastUpdated time.Time
+	realtime    bool
+	completed   bool
 }
 
 type Metrics struct {
@@ -155,6 +157,23 @@ func (m *Metrics) StopRealtimeMetricsForKey(key string) {
 
 	realtimeMetrics := m.workflows[key]
 	for _, metric := range realtimeMetrics {
+		if realtimeMetric, ok := m.customMetrics[metric]; ok {
+			realtimeMetric.completed = true
+			m.customMetrics[metric] = realtimeMetric
+		}
+	}
+}
+
+func (m *Metrics) DeleteRealtimeMetricsForKey(key string) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	if _, exists := m.workflows[key]; !exists {
+		return
+	}
+
+	realtimeMetrics := m.workflows[key]
+	for _, metric := range realtimeMetrics {
 		delete(m.customMetrics, metric)
 	}
 
@@ -190,7 +209,7 @@ func (m *Metrics) UpsertCustomMetric(key string, ownerKey string, newMetric prom
 	} else {
 		m.metricNameHelps[name] = help
 	}
-	m.customMetrics[key] = metric{metric: newMetric, lastUpdated: time.Now()}
+	m.customMetrics[key] = metric{metric: newMetric, lastUpdated: time.Now(), realtime: realtime}
 
 	// If this is a realtime metric, track it
 	if realtime {
--- a/workflow/metrics/metrics_test.go
+++ b/workflow/metrics/metrics_test.go
@@ -131,6 +131,53 @@ func TestMetricGC(t *testing.T) {
 	assert.Len(t, m.customMetrics, 0)
 }
 
+func TestRealtimeMetricGC(t *testing.T) {
+	config := ServerConfig{
+		Enabled: true,
+		Path:    DefaultMetricsServerPath,
+		Port:    DefaultMetricsServerPort,
+		TTL:     1 * time.Second,
+	}
+	m := New(config, config)
+	assert.Len(t, m.customMetrics, 0)
+
+	err := m.UpsertCustomMetric("realtime_metric", "workflow-uid", newCounter("test", "test", nil), true)
+	if assert.NoError(t, err) {
+		assert.Len(t, m.customMetrics, 1)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go m.garbageCollector(ctx)
+
+	// simulate workflow is still running.
+	timeoutTime := time.Now().Add(time.Second * 2)
+	// Ensure we get at least one TTL run
+	for time.Now().Before(timeoutTime) {
+		// Break if we know our test will pass.
+		if len(m.customMetrics) == 0 {
+			break
+		}
+		// Sleep to prevent overloading test worker CPU.
+		time.Sleep(100 * time.Millisecond)
+	}
+	assert.Len(t, m.customMetrics, 1)
+
+	// simulate workflow is completed.
+	m.StopRealtimeMetricsForKey("workflow-uid")
+	timeoutTime = time.Now().Add(time.Second * 2)
+	// Ensure we get at least one TTL run
+	for time.Now().Before(timeoutTime) {
+		// Break if we know our test will pass.
+		if len(m.customMetrics) == 0 {
+			break
+		}
+		// Sleep to prevent overloading test worker CPU.
+		time.Sleep(100 * time.Millisecond)
+	}
+	assert.Len(t, m.customMetrics, 0)
+}
+
 func TestWorkflowQueueMetrics(t *testing.T) {
 	config := ServerConfig{
 		Enabled: true,
@@ -171,7 +218,7 @@ func TestRealTimeMetricDeletion(t *testing.T) {
 	assert.NotEmpty(t, m.workflows["123"])
 	assert.Len(t, m.customMetrics, 1)
 
-	m.StopRealtimeMetricsForKey("123")
+	m.DeleteRealtimeMetricsForKey("123")
 	assert.Empty(t, m.workflows["123"])
 	assert.Len(t, m.customMetrics, 0)
 
@@ -183,7 +230,7 @@ func TestRealTimeMetricDeletion(t *testing.T) {
 	assert.Empty(t, m.workflows["456"])
 	assert.Len(t, m.customMetrics, 1)
 
-	m.StopRealtimeMetricsForKey("456")
+	m.DeleteRealtimeMetricsForKey("456")
 	assert.Empty(t, m.workflows["456"])
 	assert.Len(t, m.customMetrics, 1)
 }
--- a/workflow/metrics/server.go
+++ b/workflow/metrics/server.go
@@ -139,7 +139,12 @@ func (m *Metrics) garbageCollector(ctx context.Context) {
 		case <-ticker.C:
 			for key, metric := range m.customMetrics {
 				if time.Since(metric.lastUpdated) > m.metricsConfig.TTL {
-					delete(m.customMetrics, key)
+					switch {
+					case metric.realtime && metric.completed:
+						delete(m.customMetrics, key)
+					case !metric.realtime:
+						delete(m.customMetrics, key)
+					}
 				}
 			}
 		}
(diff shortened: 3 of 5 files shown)


workflow/metrics/metrics.go
```diff
@@ -42,6 +42,8 @@ func (s ServerConfig) SameServerAs(other ServerConfig) bool {
 type metric struct {
 	metric      prometheus.Metric
 	lastUpdated time.Time
+	realtime    bool
```
Joibel, line 45: metric.IsRealtime() does this already, I'm not sure we should be storing this.

workflow/metrics/metrics.go
```diff
@@ -153,6 +155,23 @@ func (m *Metrics) StopRealtimeMetricsForKey(key string) {
 		return
 	}
 
+	realtimeMetrics := m.workflows[key]
+	for _, metric := range realtimeMetrics {
+		if realtimeMetric, ok := m.customMetrics[metric]; ok {
+			realtimeMetric.completed = true
```
Joibel, line 161: Should this also update `lastUpdated` so that the metric lives TTL seconds beyond the end of the workflow rather than potentially ending immediately? It feels likely this would have been set recently, but this would ensure it.

workflow/metrics/metrics.go
```diff
@@ -153,6 +155,23 @@ func (m *Metrics) StopRealtimeMetricsForKey(key string) {
...
+			m.customMetrics[metric] = realtimeMetric
```
Joibel, line 162: What is this line doing, shouldn't it already be in this array?

workflow/metrics/server.go
```diff
@@ -138,7 +138,11 @@ func (m *Metrics) garbageCollector(ctx context.Context) {
 			return
 		case <-ticker.C:
 			for key, metric := range m.customMetrics {
-				if time.Since(metric.lastUpdated) > m.metricsConfig.TTL {
+				if metric.realtime {
```
Joibel, line 141: I'd prefer this as
```
if time.Since(metric.lastUpdated) > m.metricsConfig.TTL {
  switch {
    case metric.metric.IsRealtime() && metric.completed:
        delete(m.customMetrics, key)
    case !metric.metric.IsRealtime():
        delete(m.customMetrics, key)
  }
}
```

Or perhaps read IsRealtime() just once and cache, but test TTL because it's the primary test here. The result is identical, I'm aiming for understandability.

Joibel (changes requested):

workflow/metrics/metrics.go
jswxstw: `IsRealtime()` is in `wfv1.Prometheus`, but metric here is `prometheus.Metric` and it does not have this function.

workflow/metrics/metrics.go
jswxstw: `lastUpdated` of the metric has been updated in workflow operation, so it is no need to update it again in `persistUpdates`.

workflow/metrics/metrics.go
jswxstw: The value of `customMetrics` map is `struct` type, not a pointer type, so `realtimeMetric` is actually a copy and I need to update it to the map again.

workflow/metrics/server.go
jswxstw: I originally considered writing it like you did, but when it comes to realtime metrics that are not yet complete, there's no need to calculate whether they exceed the TTL, which can save some time.
How do you think about it?

workflow/metrics/metrics.go
Joibel: Ah, yep.

workflow/metrics/metrics.go
Joibel: Grr, I hate that pattern. Understood.

workflow/metrics/metrics.go
Joibel: Ok.

workflow/metrics/server.go
Joibel: I'd prefer readability and understandability over performance here. It's not performance critical nor happening very frequently.

jswxstw: Got it, done.

workflow/metrics/metrics_test.go
```diff
@@ -112,6 +112,10 @@ func TestMetricGC(t *testing.T) {
 	if assert.NoError(t, err) {
 		assert.Len(t, m.customMetrics, 1)
 	}
+	err = m.UpsertCustomMetric("realtime_metric", "workflow-uid", newCounter("test", "test", nil), true)
```
Joibel, line 115: I can see what you're doing with this test, but without this diff to read this test is slightly obscure. I know overlapping the two tests in time is good for speed of the run, but I'd still prefer to do the whole test isolated from the previous test for understandability.

Joibel (changes requested):

workflow/metrics/metrics_test.go
jswxstw: Ok, I add a new test: `TestRealtimeMetricGC`.

Joibel (approved):
Thanks for your patience. LGTM

PR #12803, opened 2024-03-14
feat: Add update command for cron,template,cluster-template. Fixes: #5464 #7344

shuangkun:
Fixes #5464 
Fixes #7344 
_edited by agilgur5:_ Supersedes / builds on top of #5465

### Motivation

Add update command supports easy updating of template, cron, cluster-template

### Verification

local test and e2e test.

PR #12839, opened 2024-03-24
refactor(cli): move common functions to `util.go`

shuangkun:
As https://github.com/argoproj/argo-workflows/pull/12803#issuecomment-2016698490  mentioned, I move "`unmarshal*` and `print*`" to `util.go`.


cmd/argo/commands/clustertemplate/list.go
```diff
@@ -57,14 +54,3 @@ func NewListCommand() *cobra.Command {
 	command.Flags().StringVarP(&listArgs.output, "output", "o", "", "Output format. One of: wide|name")
 	return command
 }
-
-func printTable(wfList []wfv1.ClusterWorkflowTemplate) {
```
agilgur5, line 61: `printTable` is actually _only_ used inside of the respective `list.go` files, so it shouldn't be moved. 

If you did a regex search on `print*`, that's my bad, I didn't realize this `printTable` function would be included in that search.
I only [meant](https://github.com/argoproj/argo-workflows/pull/12803#discussion_r1536717614) the `print<Resource>` functions, like `printClusterWorkflowTemplate*` (and the ones for `cron` and `template` commands, respectively).
`print<Resource>` and `unmarshal<Resource>` are used in multiple files, so they should really be defined in a common helper file like `util.go`. `printTable` is only used in one file, so it doesn't match them.

cmd/argo/commands/cron/list.go
```diff
@@ -60,32 +55,3 @@ func NewListCommand() *cobra.Command {
 	command.Flags().StringVarP(&listArgs.labelSelector, "selector", "l", "", "Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2). Matching objects must satisfy all of the specified label constraints.")
 	return command
 }
-
-func printTable(wfList []wfv1.CronWorkflow, listArgs *listFlags) {
```
agilgur5, line 64: not `printTable`, same as above

cmd/argo/commands/cron/util.go
agilgur5, line 1: you missed moving `unmarshalCronWorkflows` here

cmd/argo/commands/template/list.go
```diff
@@ -56,20 +53,3 @@ func NewListCommand() *cobra.Command {
 	command.Flags().StringVarP(&listArgs.output, "output", "o", "", "Output format. One of: wide|name")
 	return command
 }
-
-func printTable(wfList []wfv1.WorkflowTemplate, listArgs *listFlags) {
```
agilgur5, line 60: not `printTable`, same as above

agilgur5 (commented):
`printTable` doesn't need to be moved, sorry about the confusion -- see in-line comment.

Also `unmarshalCronWorkflows` was missed -- see in-line comment.

cmd/argo/commands/clustertemplate/list.go
shuangkun: OK, I modified it, thanks！

cmd/argo/commands/cron/util.go
shuangkun: Thanks, I forgot it before, added it now.

agilgur5 (approved):
Thanks for following up on this!

PR #12843, opened 2024-03-25
docs(sso): clarify Dex `staticClients` ID

DavidGamba:
Clarify that staticClients id is the OIDC client ID, not the secret name.

Fixes #TODO

### DIFF
--- a/docs/argo-server-sso-argocd.md
+++ b/docs/argo-server-sso-argocd.md
@@ -54,7 +54,8 @@ data:
   dex.config: |
     # Setting staticClients allows Argo Workflows to use Argo CD's Dex installation for authentication
     staticClients:
-      - id: argo-workflows-sso # this is the non base64 OIDC clientID 
+      # This is the OIDC client ID in plaintext
+      - id: argo-workflows-sso
         name: Argo Workflow
         redirectURIs:
           - https://argo-workflows.mydomain.com/oauth2/callback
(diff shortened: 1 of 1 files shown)


docs/argo-server-sso-argocd.md
```diff
@@ -54,7 +54,7 @@ data:
   dex.config: |
     # Setting staticClients allows Argo Workflows to use Argo CD's Dex installation for authentication
     staticClients:
-      - id: argo-workflows-sso
+      - id: argo-workflows-sso # this is the non base64 OIDC clientID
```
agilgur5, line 57:
```suggestion
      # This is OIDC client ID in plaintext
      - id: argo-workflows-sso
```

- match the style of comments in this doc
- "plaintext" is clearer than "non base64" 

also line 64 below does mention this, although not in the clearest way

DavidGamba: Thanks, I read this document like a hundred times and it wasn't until I saw a comment in an issue that I realized this was the actual client ID.

Unfortunately in this doc both the secret and the client ID have the same name so it was a bit confusing to me.

agilgur5: Yea I could see how that could be confusing. I have yet to do a full copy edit of this page, so it is likely lacking in clarity and [directness](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language) (most docs are, in Argo and elsewhere).

I don't think I had the same problem setting up SSO myself though, but I knew what Dex was looking for and knew to go to Dex docs as well, so it was perhaps easier for me to grok as such (and so goes docs prioritization -- they tend to get more attention when there are more obvious issues 😅 )

agilgur5 (approved):
Thanks for the clarification!

PR #12847, opened 2024-03-27
chore(deps): bump k8s dependencies to 1.26

blkperl:
### Motivation

Update k8s dependencies to a maintained version

### Modifications

* updated go.mod to k8s 1.26
* updated calls to AddEventHandler to handle errors
* fixed linter errors due to pointer function name changes
* changed hack/docs/check-env-doc.sh to use env to chose the bash version as mapfile is not available in the default macos bash version

### Verification

No manual verification, the existing test suite will catch issues.


terrytangyuan (commented):
Hmm https://github.com/argoproj/argo-workflows/pull/12822 was just merged.

agilgur5:
> Hmm #12822 was just merged.

That was 1.25, this is 1.26

blkperl:
@terrytangyuan I separated the version upgrades to make them easier to review since many generated files change on each version update.

terrytangyuan:
>That was 1.25, this is 1.26

Yeah I know. I was curious whether we cannot upgrade to 1.26 directly.

terrytangyuan:
>make them easier to review since many generated files change on each version update.

Well, now we have to review diffs on generated files again

terrytangyuan (approved):

workflow/common/util.go
```diff
@@ -100,7 +101,7 @@ func ExecPodContainer(restConfig *rest.Config, namespace string, pod string, con
 func GetExecutorOutput(exec remotecommand.Executor) (*bytes.Buffer, *bytes.Buffer, error) {
 	var stdOut bytes.Buffer
 	var stdErr bytes.Buffer
-	err := exec.Stream(remotecommand.StreamOptions{
+	err := exec.StreamWithContext(context.TODO(), remotecommand.StreamOptions{
```
agilgur5, line 104: this would need a bit of refactoring for the entire function chain to pass a `context` down, but this could inherit the `context` of [`processNextPodCleanupItem`](https://github.com/argoproj/argo-workflows/blob/748ae475d3400f2aa51800c2337607811550d989/workflow/controller/controller.go#L546)

agilgur5:
Thanks for all the tedious changes!

> * changed hack/docs/check-env-doc.sh to use env to chose the bash version as mapfile is not available in the default macos bash version

Yea I mentioned portability in #12652 😐 I believe newer Macs are on newer Bash by default. Hopefully that change helps with portability though

agilgur5:
> > make them easier to review since many generated files change on each version update.
> 
> Well, now we have to review diffs on generated files again

I took a rough look (not a fully detailed review), but the diff is pretty massive, so I do think this helps. Also, the 1.26 changes (this and the tests in #12854) do change our support matrix, so I also think that is a helpful separation.

Perhaps worth noting, there are tools for "Stacked PRs" like this, such as https://graphite.dev/ and [Sapling](https://github.com/facebook/sapling) but they tend to require fairly deep code review integration to function optimally.
GH's code review leaves a lot to desire when it comes to "Stacked PRs" and other rebase or iteration features (e.g. only requiring review of lines changed per commit and force push). Gerrit's had support for some of those features for a decade or more at this point 😕

blkperl:
This one should be safe for older versions of K8s or at least 1.25 but yes https://github.com/argoproj/argo-workflows/pull/12854 updates the docs to say 1.25 is no longer officially supported.

workflow/common/util.go
blkperl: Neat, I've opened https://github.com/argoproj/argo-workflows/pull/12858

terrytangyuan:
@blkperl Could you take a look at the failures in main branch?

blkperl:
@terrytangyuan I looked at the failure and didn't understand the failure. The past two commits to main passed though so it seems like an intermittent issue.

workflow/controller/controller.go
```diff
@@ -291,15 +291,38 @@ func (wfc *WorkflowController) Run(ctx context.Context, wfWorkers, workflowTTLWo
...
+	if err != nil {
+		log.Fatal(err)
+	}
 	wfc.updateEstimatorFactory()
 
-	wfc.configMapInformer = wfc.newConfigMapInformer()
+	wfc.configMapInformer, err = wfc.newConfigMapInformer()
+	if err != nil {
+		log.Fatal(err)
+	}
```
agilgur5, line 325: so except for the `add*InformerHandler` functions, all of the `new*Informer` functions will never error out, so we can actually skip the error checking.
they [only error](https://github.com/kubernetes/client-go/blob/46588f2726fa3e25b1704d6418190f424f95a990/tools/cache/shared_informer.go#L580) in the case of the informer having been stopped or started, and they get created within these functions (and started lower down in this function), so that will never be the case.


I'm going to write up a separate PR to fix that as I happened to work on that and notice it during the process of #11855.

agilgur5: Removed in #12917

PR #12632, opened 2024-02-06
fix: use optimistic concurrency strategy when updating pod status

sakai-ast:
Fixes https://github.com/argoproj/argo-workflows/pull/12596

### Motivation
https://github.com/argoproj/argo-workflows/pull/12596#issuecomment-1928042158

### Verification
ut and e2e test

PR #12862, opened 2024-03-30
fix: optimise pod finalizers with merge patch and resourceVersion

imliuda:
Improvement of #12413

### Motivation

The origin patch using a json patch, which may cause a http 422 error, and it is not idempotent. And it needs a cron task to remove finalizers of workflows has been deleted when workflow-controller is down, or there may be lot of orphaned pods.

But, using finalizer mechanism to prevent pod deletionn is still a dangerous method, there may be some potential risk, for example, when load is high, workflow-controller may be restart, so can't work appropriately, and pod creation speed may be faster than deletion (may need change clean pod-cleanup-workers count), etc.

### Modifications

1. change patch type from json patch to merge patch
2. add a cron task to remove finalizers due to workflow deleted, and after timeout
3. reuse pod clean up queue to delete pods
4. delete `getPodFromAPI()` method, which may heavy the load of apiserver

### Verification

I have tested in my local environment by creating, deleting workflows in case of workflow-controller running and not running.


workflow/controller/controller.go
```diff
@@ -542,6 +541,52 @@ func (wfc *WorkflowController) runPodCleanup(ctx context.Context) {
 	}
 }
 
+func (wfc *WorkflowController) getPodFinalizerRemovalPatch(namespace, podName string) (string, bool) {
+	if os.Getenv("ARGO_POD_STATUS_CAPTURE_FINALIZER") != "true" {
+		return "", false
+	}
+	found := false
+	keeps := make([]string, 0)
+	pod, err := wfc.getPod(namespace, podName)
```
toVersus, line 550: Is it safe to assume that the Informer cache here is up to date? If other controllers are adding or removing Pod finalizers at the same time, is there a chance that finalizers could be added or removed unintentionally?

workflow/controller/controller.go
```diff
@@ -542,6 +541,52 @@ func (wfc *WorkflowController) runPodCleanup(ctx context.Context) {
 	}
 }
 
+func (wfc *WorkflowController) getPodFinalizerRemovalPatch(namespace, podName string) (string, bool) {
```
toVersus, line 544: Wouldn't it be better to generate a JSON merge patch from the original and modified objects, rather than creating one manually? For example, something like this:

```diff
diff --git a/workflow/controller/controller.go b/workflow/controller/controller.go
index bb0b84d8a..f62649465 100644
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -10,6 +10,7 @@ import (
        "syscall"
        "time"

+       jsonpatch "github.com/evanphx/json-patch"
        "github.com/upper/db/v4"

        "github.com/argoproj/pkg/errors"
@@ -541,24 +542,38 @@ func (wfc *WorkflowController) runPodCleanup(ctx context.Context) {
        }
 }

-func (wfc *WorkflowController) getPodFinalizerRemovalPatch(namespace, podName string) (string, bool) {
+func (wfc *WorkflowController) getPodFinalizerRemovalPatch(namespace, podName string) ([]byte, error) {
        if os.Getenv("ARGO_POD_STATUS_CAPTURE_FINALIZER") != "true" {
-               return "", false
+               return nil, nil
        }
-       found := false
+       oldPod, err := wfc.getPod(namespace, podName)
+       if err != nil {
+               return nil, nil
+       }
+
+       newPod := oldPod.DeepCopy()
        keeps := make([]string, 0)
-       pod, err := wfc.getPod(namespace, podName)
-       if err == nil && pod != nil {
-               for _, f := range pod.ObjectMeta.Finalizers {
-                       if f == common.FinalizerPodStatus {
-                               found = true
-                       } else {
-                               keeps = append(keeps, f)
-                       }
+       for _, f := range newPod.ObjectMeta.Finalizers {
+               if f != common.FinalizerPodStatus {
+                       keeps = append(keeps, f)
                }
        }
-       finalizers, _ := json.Marshal(keeps)
-       return string(finalizers), found
+       newPod.Finalizers = keeps
+
+       oldData, err := json.Marshal(oldPod)
+       if err != nil {
+               return nil, err
+       }
+       newData, err := json.Marshal(newPod)
+       if err != nil {
+               return nil, err
+       }
+
+       mergePatch, err := jsonpatch.CreateMergePatch(oldData, newData)
+       if err != nil {
+               return nil, err
+       }
+       return mergePatch, nil
 }

 func (wfc *WorkflowController) podHasFinalizer(pod *apiv1.Pod) bool {
@@ -571,18 +586,23 @@ func (wfc *WorkflowController) podHasFinalizer(pod *apiv1.Pod) bool {
 }

 func (wfc *WorkflowController) removePodFinalizer(ctx context.Context, namespace, podName string) error {
-       if finalizers, found := wfc.getPodFinalizerRemovalPatch(namespace, podName); found {
-               // merge patch is idempotent
-               _, err := wfc.kubeclientset.CoreV1().Pods(namespace).Patch(
-                       ctx,
-                       podName,
-                       types.MergePatchType,
-                       []byte(fmt.Sprintf(`{"metadata": {"finalizers": %s}}`, finalizers)),
-                       metav1.PatchOptions{},
-               )
-               if err != nil {
-                       return err
-               }
+       mergePatch, err := wfc.getPodFinalizerRemovalPatch(namespace, podName)
+       if err != nil {
+               return err
+       }
+       if mergePatch == nil {
+               return nil
+       }
+
+       // merge patch is idempotent
+       if _, err := wfc.kubeclientset.CoreV1().Pods(namespace).Patch(
+               ctx,
+               podName,
+               types.MergePatchType,
+               mergePatch,
+               metav1.PatchOptions{},
+       ); err != nil {
+               return err
        }
        return nil
 }
@@ -619,18 +639,20 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
                                return err
                        }
                case labelPodCompleted:
-                       patch := `{"metadata": {"labels": {"workflows.argoproj.io/completed": "true"}}}`
-                       if finalizers, found := wfc.getPodFinalizerRemovalPatch(namespace, podName); found {
-                               patch = fmt.Sprintf(`{"metadata": {"finalizers": %s, "labels": {"workflows.argoproj.io/completed": "true"}}}`, finalizers)
+                       mergePath, err := wfc.getPodFinalizerRemovalPatch(namespace, podName)
+                       if err != nil {
+                               return err
                        }
-                       _, err := pods.Patch(
+                       if mergePath == nil {
+                               return nil
+                       }
+                       if _, err := pods.Patch(
                                ctx,
                                podName,
                                types.MergePatchType,
-                               []byte(patch),
+                               mergePath,
                                metav1.PatchOptions{},
-                       )
-                       if err != nil {
+                       ); err != nil {
                                return err
                        }
                case deletePod:
```

workflow/controller/controller.go
imliuda: This is argo workflow-controller managed pod, it seems that no other controller will add finalizers to this pod. If there is such a controller, it should have already added at pod creation time, if other controller is removing the controller, it can remove it one more time when recieving a Update event. But, what else controller wiil add a finalizer to a workflow pod...

workflow/controller/controller.go
imliuda: I can write a new function `getPodCleanupPatch(namespace, podName string, labelPodCompleted, removeFinalizer bool) ([]byte, erroir) {}` to generate the patch. About if using the json-patch, I think the original way is more straight-forward, in our environment, some pod's size can be greater than hundred KBs, which could cost extra cpu and memory if using json-patch.

docs/environment-variables.md
```diff
@@ -8,52 +8,53 @@ This document outlines environment variables that can be used to customize behav
...
+| `ARGO_AGENT_MEMORY_LIMIT`                | `resource.Quantity`  | `256m`                       | Memory resource limit for the agent.                                                                                                                                                                                                                                     |
+| `ARGO_POD_STATUS_CAPTURE_FINALIZER`      | `bool`               | `false`                      | The finalizer blocks the deletion of pods until the controller captures their status.                                                                                                                                                                                    | 
+| `ARGO_POD_STATUS_CAPTURE_TIMEOUT`        | `time.Duration`      | `10m`                        | After the timeout time, workflow controller will remove the finalizer anyway to avoid a potential risk of pods increased infinitely.                                                                                                                                     |
```
agilgur5, line 26: this is the one line that was added here, for reference. can set "Hide whitespace" in GH diff to properly see that.

agilgur5 (commented):
> Improvement of https://github.com/argoproj/argo-workflows/pull/12413

For reference, that was improved in #12596, where we also [suggested](https://github.com/argoproj/argo-workflows/pull/12596#issuecomment-1926257286) changing to an `update`, which was then completed in #12632

> The origin patch using a json patch, which may cause a http 422 error, and it is not idempotent.

After #12632, it uses an `update` instead of a `patch`, which has a `ResourceVersion` in it which allows k8s to calculate the diff if there are multiple changes, or otherwise return a 409 conflict. So an `update` should have the same effect.

Though a merge patch might be more efficient than an `update` if it can provide the same guarantees

<hr> 

Regarding removing Pod finalizers if a Workflow has been deleted, as you mentioned in https://github.com/argoproj/argo-workflows/pull/12413#issuecomment-1996618841, I believe the more appropriate way would be to also add a finalizer on the Workflow _itself_. That way a Workflow cannot be deleted until it has completed all reconciliation, which would include removing Pod finalizers.
If a user does a force delete or removes the finalizer manually, then it would be on them to remove corresponding Pod finalizers (e.g. with a script, as mentioned in https://github.com/argoproj/argo-workflows/pull/12413#issuecomment-1982840060)

The "cron task" could cause orphaned resources, which is what a finalizer is intended to prevent.

workflow/controller/controller.go
toVersus: > I think the original way is more straight-forward, in our environment, some pod's size can be greater than hundred KBs, which could cost extra cpu and memory if using json-patch.

I see. Your point makes perfect sense to me.

> I can write a new function getPodCleanupPatch(namespace, podName string, labelPodCompleted, removeFinalizer bool) ([]byte, erroir) {} to generate the patch.

It might be more efficient than handling strings.

workflow/controller/controller.go
toVersus: Ah, so generating an idempotent merge patch can avoid these cases.

imliuda:
@agilgur5 I have viewed #12596 and #12632, I think using Get() and Update() for each pod is not a good way, eg, if pod creation rate is 10/s, each pod's size is 200KB (usally it has very big evironment vars), then the additional bandwidh will be 4MB/s, this is a waste. @juliev0 how do you think about this? If there really is a controller will add finalizer to argo pods, then it should have been added when pod creation, if other controller has removed it's own finalizer, we may add it to pods again by Patch, the other controller could just remove it one more time.

> The "cron task" could cause orphaned resources, which is what a finalizer is intended to prevent.

@agilgur5 Add a finalizer to workflow is indeed an available option, but why "cron task" could cause orhpaned resosurces? Currently, the syncPodFinalizer() is a just simple way to ensure pod finalizers can be removed, it is more like a supervisor, it does't care about workflows statuses, controller's heathy status, etc.

agilgur5:
> then the additional bandwidth will be 4MB/s, this is a waste.

Yes I mentioned in my [comment](https://github.com/argoproj/argo-workflows/pull/12862#pullrequestreview-1970009577) that _if_ the same guarantees can be achieved, then this would indeed be an optimization.

> the other controller could just remove it one more time.

We should never be touching a finalizer that is not our own. This applies for all Controllers in k8s, not just Argo. We should also not assume how other Controllers react and definitely should not dictate how other Controllers react if Argo messes with their finalizers. Argo shouldn't touch their finalizers to begin with as they are not controlled by Argo.

Efficiency is not really relevant if the behavior is not correct.

> it does't care about workflows statuses, controller's heathy status, etc.

Which means it can be inconsistent.

> but why "cron task" could cause orphaned resources?

If I delete my Workflow before it has finished cleaning up, I now have orphaned resources. A finalizer is intended to prevent this scenario.

imliuda:
> If I delete my Workflow before it has finished cleaning up, I now have orphaned resources. A finalizer is intended to prevent this scenario.

Every pod have a owner referece point to a workflow, if workflow is deleted, kubernetes garbage collector will delete the pods automatically. It is just that pod have finalizers on it, so it will still exists until the finalizer is removed. The "cron task" run very 15s to remove finalizers of pods which owner workflow has been deleted (namely orphaned resource). 

The "cron task" will also remove finalizers of normal pods (that owner workflow still exists) when it is terminated for a while , just in case of there are have lot of workflows that it can't handle in time, or controller restart infinitely due to bugs or something like leader election timeout, etc.

Although, "cron task" might not be a appropriate native way, but it is a simple and reliable way.

About Patch or Get & Update, I think we can make if configurable, so that user can make a choice based on their real circumstances.

imliuda:
I have found merge patch can also follow the resourceVersion check if we pass it to request. we can verify it by:
```
kubectl run nginx --image=nginx
kubectl get pod nginx -oyaml | grep resourceVersion
# pass the resourceVersion, execute the command below twice or more times
kubectl -v 8 patch --type=merge pod  nginx -p '{"metadata":{"finalizers":[],"resourceVersion":"24472721"}}'
```
I have also increased the default pod gc workers count, lest the speed of deletion is slower than the speed of creation in case of api throttling.

juliev0:
Hey all, sorry I haven’t had a chance to look at this yet. I’ll try to take a look later this week, but also feel free to act independently according to what you think @agilgur5. Thanks!

juliev0:
Okay, just read through the thread. I'm okay with switching the `Update()` with a merge patch to reduce transmitted bytes.

The other thing you're saying to remove is the `Get()` which is made as part of the `getPodFromAPI()` call. Want to check with @sakai-ast (if he remembers), as I know there was some behavior seen that caused adding this.  

So, as far as the cron, it does seem like using a Workflow finalizer would be a more elegant solution, and I think would enable us to remove [this](https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/controller.go#L1068-L1080) code which occurs when the deletion timestamp gets added, right? Unlike the cron approach, it may cause some Pods that were supposed to be deleted when they completed to exist longer than they would if say a Workflow doesn't get deleted for a long time, but I think we are looking at a fairly rare case, right? and are just trying to ensure that these Pods get cleaned up eventually?

imliuda:
Yes, of course, using workflow finalizer is a more elegent way. The reason why I choose to use a cron is that I have encountered some situations like apiserver OOM, workflow leader election timeouts (due to api calls cost long time), leading to workflow-controller always restarts. And after every restart, may not all workflows can get reconciled, then it restarts again. So in a worst case, there may be some workflows can never get reconciled.

In the cron, DeleteFunc is not the only change to remove the finalizers, it will also check the pods that finished over a timeout duration, and then remove the fianlizer on it. This cron will be executed immediately after workflow started, and then every 15s. This is the currently situation.

If in a finalizer way, I think I need some more time to do a research to find a solution to avoid the above cases. Hoping more advice can be gave.

juliev0:
> Yes, of course, using workflow finalizer is a more elegent way. The reason why I choose to use a cron is that I have encountered some situations like apiserver OOM, workflow leader election timeouts (due to api calls cost long time), leading to workflow-controller always restarts. And after every restart, may not all workflows can get reconciled, then it restarts again. So in a worst case, there may be some workflows can never get reconciled.

Okay, let me try to understand the issues. So, you mention the workflow leader election timeouts. Is this due to rate limiting occurring either from the K8S side or the client side? And how is this affecting the apiserver memory? 

Are you essentially saying that you need the immediacy of the cron vs having to wait until the workflow's been deleted?

imliuda:
> > Yes, of course, using workflow finalizer is a more elegent way. The reason why I choose to use a cron is that I have encountered some situations like apiserver OOM, workflow leader election timeouts (due to api calls cost long time), leading to workflow-controller always restarts. And after every restart, may not all workflows can get reconciled, then it restarts again. So in a worst case, there may be some workflows can never get reconciled.
> 
> Okay, let me try to understand the issues. So, you mention the workflow leader election timeouts. Is this due to rate limiting occurring either from the K8S side or the client side? And how is this affecting the apiserver memory?
> 
> Are you essentially saying that you need the immediacy of the cron vs having to wait until the workflow's been deleted?

The apiserver is already in really high memory usage and low etcd performance before we upgrade, the workflow-controller restarts frequently (it is leader election timeout due to etcd bad performance), so after we upgrade, it still always restarts. There are hundreds of workflows running simultaneously in our clusters, through the monitoring, we observed each workflow operation time can take 3 seconds or more. So before all workflows get reconciled, workflow-controller may restart again, leading pod finalizer didn't get removed opportunely. And so with the number of pods increasing, apiserver got OOM.

So I treat the fianalizer removal as the first priority thing. And so far, in the finalizer way, I think the workflow queue's sort policy can be changed, making the workflow with non-zero DeletionTimestamp Popped and handled firstly. Even though, we still can't ensure the number of pods with finalizer will not increase continuously, so a util cheking the number of that pods may be introduced, once the number reachs some threshold, we can skip handling for a while, and wait the number dropped.

juliev0:
Sorry, can you clarify why it should be a cron vs just something done at start up? (since it sounded like it was an issue caused when the Workflow Controller is down)

imliuda:
> Sorry, can you clarify why it should be a cron vs just something done at start up? (since it sounded like it was an issue caused when the Workflow Controller is down)

 It's something should be done at start up.

juliev0:
> > Sorry, can you clarify why it should be a cron vs just something done at start up? (since it sounded like it was an issue caused when the Workflow Controller is down)
> 
> It's something should be done at start up.

Okay, if that's the case, then what do you and others think about just doing this at start up time, like maybe before the pod deletion workers have started?

imliuda:
> > > Sorry, can you clarify why it should be a cron vs just something done at start up? (since it sounded like it was an issue caused when the Workflow Controller is down)
> > 
> > 
> > It's something should be done at start up.
> 
> Okay, if that's the case, then what do you and others think about just doing this at start up time, like maybe before the pod deletion workers have started?

**The key point here is to ensure finalizers can be removed in any case** (eg, controlled restart frequently, apiserver oom, api throtling...). 

Do it at start time is a way of  handling controller restarts case. This may be not elegant, as I said above, if using workflow finalizers to do pod finalizer removal, we can sort workflows by DeletionTimestamp, so the associated pod's finalizer can also be removed as soon  as controller starts.

juliev0:
> **The key point here is to ensure finalizers can be removed in any case** (eg, controlled restart frequently, apiserver oom, api throtling...).
> 
So, my understanding is that if there are any issues that occur _during_ the Workflow operation (deletion queue), there will be retrying if there's any transient error [here](https://github.com/imliuda/argo-workflows/blob/master/workflow/controller/controller.go#L665). So, if that's the case, then we just need to handle the case of Workflows having been deleted while the Controller was down, right?

> Do it at start time is a way of handling controller restarts case. This may be not elegant, as I said above, if using workflow finalizers to do pod finalizer removal, we can sort workflows by DeletionTimestamp, so the associated pod's finalizer can also be removed as soon as controller starts.

Sorry, to clarify here, I wasn't referring to the "Workflow Finalizer" method. I just meant running your current cron function once at startup, rather than on repeat. Does it make sense or no?

agilgur5:
> About Patch or Get & Update, I think we can make if configurable, so that user can make a choice based on their real circumstances.

> Okay, just read through the thread. I'm okay with switching the `Update()` with a merge patch to reduce transmitted bytes.

Let's make sure we finish the `Patch` vs `Update` discussion.

It shouldn't be configurable -- it should always be correct. If the patch will never delete a _different_ finalizer, then we should use the patch, as it is more optimal -- there's no reason not to.
If it's _not_ correct, we shouldn't be doing it. Again, fundamentally, to make any assumptions and guarantees, you need to ensure correctness first.

I think that is more straightforward as well -- if it is correct and optimal, we can all agree and merge it in.

The "cron task" still has more discussion and iteration, and is altogether independent of `Patch` vs. `Update`, so it probably makes sense to split that off into a separate PR (and there's already a title with two subjects in the PR, which is often an indicator of mixed concerns).

> but I think we are looking at a fairly rare case, right?

> Are you essentially saying that you need the immediacy of the cron vs having to wait until the workflow's been deleted?

Same for the "cron task", _correctness_ first. Once that is guaranteed, we can optimize for these rarer high load scenarios.
As an example, the situation the finalizer was added for is a lot more common than this.

> So, if that's the case, then we just need to handle the case of Workflows having been deleted while the Controller was down, right?

Yea if I'm understanding this correctly, this is the unhandled edge case. "Down" including the Controller being too busy.
A workflow finalizer would prevent this scenario.

> So I treat the fianalizer removal as the first priority thing. And so far, in the finalizer way, I think the workflow queue's sort policy can be changed, making the workflow with non-zero DeletionTimestamp Popped and handled firstly. Even though, we still can't ensure the number of pods with finalizer will not increase continuously, so a util cheking the number of that pods may be introduced, once the number reachs some threshold, we can skip handling for a while, and wait the number dropped. [sic]

This makes sense to me and I agree generally speaking. This is a race condition, but yes, under high load, this makes the Controller effectively recover. I believe we should be able to get the number of Pods from the Informer cache (i.e. without another network request).

Implementing this is a little non-trivial though, this is basically user-space prioritization of running goroutines. If we can implement that coordination-free way, then I'm all for it. Basically the "priority" of clean up goroutines should be higher than Workflow processing -- then we wouldn't need to implement a "skip" mechanism either.
Goroutines don't have priorities though. The very naive way would be to just have more clean up workers. An approach that uses a limited amount of coordination would be for the workflow workers to wait when there's too many pods that need clean up. I actually can't think of a coordination free approach off the top of my head without priorities 🤔

workflow/controller/controller.go
```diff
@@ -542,6 +535,75 @@ func (wfc *WorkflowController) runPodCleanup(ctx context.Context) {
...
+	if finalizerEnabled {
+		if wfc.podHasStatusFinalizer(pod) {
+			finalizers := make([]string, 0)
+			for _, f := range pod.ObjectMeta.Finalizers {
+				if f != common.FinalizerPodStatus {
+					finalizers = append(finalizers, f)
+				}
+			}
+			metadata["finalizers"] = finalizers
+			metadata["resourceVersion"] = pod.ObjectMeta.ResourceVersion
```
agilgur5, line 566: > I have found merge patch can also follow the resourceVersion check if we pass it to request. [...]

Huh, if you can do that with a patch, then I guess this works. But we need to handle possible conflicts then; the previous `Update` code used `RetryOnConflict` for that.
The `Patch` request on line 593 does not handle conflicts

agilgur5: I was doing some investigation on this and popped up on some things:
1. https://github.com/kubernetes/kubernetes/issues/105146 -- deleting finalizers is the exact use-case for us.
   - More specifically that's a bug to watch out for if we were to use `$deleteFromPrimitiveList` (not sure which version of k8s it's fixed in).
   - More generically, we can use directives in a strategic merge patch to do this more cleanly. Stumbled upon this [SO answer](https://stackoverflow.com/a/64565206/3431180) answer which took me to these [SIG API Machinery docs](https://github.com/kubernetes/community/blob/94b696bef96aefba2ca9bf97029694565a84495e/contributors/devel/sig-api-machinery/strategic-merge-patch.md) that describe the directives you can use in k8s.
       - I think we should be able to use `$delete` or `$deleteFromPrimitiveList` for this.
1. We could maybe do an `Update` on the finalizer subresource? That would be less network since it's just finalizers

agilgur5 (commented):
Left some comments in-line below with regard to the merge patch -- I did not look at the "cron task" as we're not quite agreed on the approach of that to begin with

workflow/controller/controller.go
agilgur5: I followed up on this [below](https://github.com/argoproj/argo-workflows/pull/12862#discussion_r1555109592) (as did Julie in [a comment](https://github.com/argoproj/argo-workflows/pull/12862#issuecomment-2040762523) before that) -- let's continue there.

> But, what else controller wiil add a finalizer to a workflow pod...

There can be many, which is why I brought this part up [below](https://github.com/argoproj/argo-workflows/pull/12862#issuecomment-2028878367) as well. Pods are a generic building block in k8s and are not Argo-specific.

A really simple counter-example is a policy controller that does some clean up. It would run on _all_ Pods, regardless of what controller started them.

workflow/controller/controller.go
imliuda: 1. I can find that fix in release-1.23 and later branches. As https://github.com/kubernetes/kubernetes/issues/105146#issuecomment-925330928 mentioned, merge patch + resourverVersion can do a consistent patch. Wether to use $deleteFromPrimitiveList or merger patch + resourceVersion depends on the lowest version of kubernetes that argo wants to support. 
2. As we can see from https://github.com/kubernetes/kubernetes/blob/release-1.24/pkg/registry/core/rest/storage_core.go#L280, pod have no finalizer subresource.

imliuda:
So, this PR will only change the finalizer removal methods, now I use merge patch and resourceVersion to ensure resource consistence. When it has conflict, will retry it by this line `if errorsutil.IsTransientErr(err) || apierr.IsConflict(err) {`. @agilgur5 I have researched `$deleteFromPrimitiveList`'s bug was fixed from release-1.23, considering already existing cluster's version may be older than that version, I choosed using merge patch and resourceVersion method.

I have read all the replies above, and I will open a new issue to discuss the edge cases.

docs/environment-variables.md
```diff
@@ -3,57 +3,57 @@
 This document outlines environment variables that can be used to customize behavior.
 
 !!! Warning
-    Environment variables are typically added to test out experimental features and should not be used by most users.
-    Environment variables may be removed at any time.
+Environment variables are typically added to test out experimental features and should not be used by most users.
+Environment variables may be removed at any time.
```
agilgur5, line 7: this shouldn't be changed, it's supposed to be 4 space indented for the admonition (`!!! Warning`) [to work](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage)

imliuda: I's sorry, this may be caused by `make docs`, I have removed this.

agilgur5: It's not caused by `make docs`: the command only does linting and codegen, it does not do formatting. I suspect your editor may have tried to auto-format without understanding the `mkdocs-material` syntax.

workflow/controller/controller.go
```diff
@@ -1054,8 +1053,8 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
 						log.WithError(err).Error("Failed to list pods")
 					}
 					for _, p := range podList.Items {
-						if err := wfc.enablePodForDeletion(ctx, pods, p.Namespace, p.Name, false); err != nil {
-							log.WithError(err).Error("Failed to enable pod for deletion")
+						if slice.ContainsString(p.Finalizers, common.FinalizerPodStatus) {
+							wfc.queuePodForCleanup(p.Namespace, p.Name, deletePod)
```
agilgur5, line 1057: hmm this is a change from the previous behavior to only queue instead of patching... but the `List` already incurred an API call...

this logic seems to stem from https://github.com/argoproj/argo-workflows/pull/12413#issuecomment-1897899247

I suppose this is fine so long as they are eventually deleted and the queue is more stable than multiple immediate network requests for every delete

agilgur5 (commented):
The isolated merge patch fix logic looks mostly good to me. I left some small comments in-line below

workflow/controller/controller.go
```diff
@@ -522,6 +515,58 @@ func (wfc *WorkflowController) runPodCleanup(ctx context.Context) {
...
 
+func (wfc *WorkflowController) getPodCleanupPatch(pod *apiv1.Pod, labelPodCompleted bool) ([]byte, error) {
+	un := unstructured.Unstructured{}
+	if labelPodCompleted {
+		un.SetLabels(map[string]string{common.LabelKeyCompleted: "true"})
+	}
+
+	finalizerEnabled := os.Getenv("ARGO_POD_STATUS_CAPTURE_FINALIZER") == "true"
+	if finalizerEnabled && slice.ContainsString(pod.Finalizers, common.FinalizerPodStatus) {
+		un.SetFinalizers(slice.RemoveString(pod.Finalizers, common.FinalizerPodStatus))
```
agilgur5, line 526: this also does still loop over finalizers twice like I mentioned [before](https://github.com/argoproj/argo-workflows/pull/12862#discussion_r1555119124).

imliuda:
@agilgur5 Could you please review it again! I updated the codes as you suggested. After this, I will create a new issue to address how to ensure finalizers will be removed under kinds of edge cases .

workflow/controller/controller.go
```diff
@@ -558,9 +620,10 @@ func (wfc *WorkflowController) processNextPodCleanupItem(ctx context.Context) bo
 	logCtx := log.WithFields(log.Fields{"key": key, "action": action})
 	logCtx.Info("cleaning up pod")
 	err := func() error {
+		pods := wfc.kubeclientset.CoreV1().Pods(namespace)
```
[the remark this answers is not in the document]
agilgur5: > I see you also re-create `pods` in `patchPodForCleanup`

You optimized this part, but this is still not needed in every case -- I don't see a need to move into the top scope, and that seems like a de-opt. Why not just leave it as-is?

workflow/controller/controller.go
```diff
@@ -522,6 +516,63 @@ func (wfc *WorkflowController) runPodCleanup(ctx context.Context) {
...
+	if finalizerEnabled {
+		finalizers, found := make([]string, 0), false
+		for _, f := range pod.Finalizers {
+			if f == common.FinalizerPodStatus {
+				found = true
+			} else {
+				finalizers = append(finalizers, f)
+			}
+		}
+		if found {
```
agilgur5, line 535: this isn't quite what I meant [before](https://github.com/argoproj/argo-workflows/pull/12862#discussion_r1562025431) -- this is the same as two loops, and two loops are cleaner as you had before.

you can skip the second conditional if you actually use the `index` in `for index, f := range`

```suggestion
		finalizers := slice.RemoveString(pod.Finalizers, common.FinalizerPodStatus)
		if finalizers != pod.Finalizers {
			un.SetFinalizers(finalizers)
```

workflow/controller/controller_test.go
```diff
@@ -1181,6 +1182,45 @@ spec:
...
+		},
+	}
+
+	os.Setenv(common.EnvVarPodStatusCaptureFinalizer, "true")
+
+	// pod finalizer enabled, patch label
+	patch, err := wfc.getPodCleanupPatch(pod, true)
+	assert.Nil(t, err)
+	expected := `{"metadata":{"resourceVersion":"123456","finalizers":[],"labels":{"workflows.argoproj.io/completed":"true"}}}`
+	require.JSONEq(t, expected, string(patch))
```
agilgur5, line 1202: is there a reason you changed this to `require` from `assert`?

agilgur5 (commented):
a few more small follow-ups

agilgur5:
Also two E2Es [are](https://github.com/argoproj/argo-workflows/actions/runs/8662300377/job/23754122870?pr=12862) [failing](https://github.com/argoproj/argo-workflows/actions/runs/8662300377/job/23754125599?pr=12862). I haven't seen those flake before, although I did have a bit of a similar problem locally, not sure if it's related to your changes or not

workflow/controller/controller.go
imliuda: oh, I understand, but it seems we can not compare two slices directlly, so I used pointers, as I can see if no removal the return slice is the same as it passed in.

imliuda: It seems that we can not compare two slices directly, I used `len()` function to do that, I searched it is a field of SliceHeader, so it wouldn't iterate the elements.

imliuda:
> Also two E2Es [are](https://github.com/argoproj/argo-workflows/actions/runs/8662300377/job/23754122870?pr=12862) [failing](https://github.com/argoproj/argo-workflows/actions/runs/8662300377/job/23754125599?pr=12862). I haven't seen those flake before, although I did have a bit of a similar problem locally, not sure if it's related to your changes or not

I'm trying to find it in my local environment.

workflow/controller/controller.go
agilgur5: Ah I meant to compare their pointers (sorry I'm still getting used to Go nuances) since `RemoveString` returns the same slice when no changes. Although I see you tried that and it still didn't work? Odd 🤔

But `len()` is simpler anyway, so that works. Thanks for fixing and checking!

workflow/controller/controller.go
```diff
@@ -1054,8 +1057,8 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
 						log.WithError(err).Error("Failed to list pods")
 					}
 					for _, p := range podList.Items {
-						if err := wfc.enablePodForDeletion(ctx, pods, p.Namespace, p.Name, false); err != nil {
-							log.WithError(err).Error("Failed to enable pod for deletion")
+						if slice.ContainsString(p.Finalizers, common.FinalizerPodStatus) {
+							wfc.queuePodForCleanup(p.Namespace, p.Name, removeFinalizer)
```
agilgur5, line 1061: would `deletePod` not make sense here? as opposed to creating a new `removeFinalizer` action?

workflow/controller/controller_test.go
```diff
@@ -1181,6 +1181,51 @@ spec:
...
+
+	pod := &apiv1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:          map[string]string{common.LabelKeyCompleted: "false"},
+			Finalizers:      []string{common.FinalizerPodStatus},
+			ResourceVersion: "123456",
+		},
+	}
+
+	os.Setenv(common.EnvVarPodStatusCaptureFinalizer, "true")
```
agilgur5, line 1195:
```suggestion
	t.Setenv(common.EnvVarPodStatusCaptureFinalizer, "true")
```

`t.Setenv` has a cleanup at the end. 

You solved the E2E failures it seems like, but I was guessing at places that could cause it and lack of cleanup was one possibility I thought of.

workflow/controller/controller_test.go
```diff
@@ -1181,6 +1181,51 @@ spec:
...
+	expected = `{"metadata":{"resourceVersion":"123456","finalizers":[]}}`
+	assert.JSONEq(t, expected, string(patch))
+
+	// pod finalizer enabled, do not patch label, nil/empty finalizers
+	podWithNilFinalizers := &apiv1.Pod{}
+	patch, err = wfc.getPodCleanupPatch(podWithNilFinalizers, false)
+	assert.Nil(t, err)
+	assert.Nil(t, patch)
+
+	os.Setenv(common.EnvVarPodStatusCaptureFinalizer, "false")
```
agilgur5, line 1215:
```suggestion
	t.Setenv(common.EnvVarPodStatusCaptureFinalizer, "false")
```

same as above

workflow/controller/controller.go
imliuda: It is cased by this case: https://github.com/argoproj/argo-workflows/blob/f0a867b6325ec3f20369bf6c92518dc5bb2ebf59/test/e2e/pod_cleanup_test.go#L120
Workflow get deleted, so pod get deleted, too, so waiting pod completed wouldn't happen, but I have not found who deleted it and when. About the failure in Windows Tests, I am testing it in a virtual machine, but, i am struggling with network problems(see my location...). I need some help.

agilgur5: You can ignore the Windows tests, they're not required to pass as they're still unstable

workflow/controller/controller_test.go
```diff
@@ -2,15 +2,15 @@ package controller
 
 import (
 	"context"
+	"os"
```
agilgur5, line 5:
```suggestion
```

Lint is failing on this line as it is no longer needed after the `os.Setenv` -> `t.Setenv` change

Joibel (approved):

Joibel:
Merged in #13776

agilgur5:
> Merged in #13776

@Joibel, as an Approver, you can usually push to existing PRs (unless the contributor has unticked the box) and so don't necessarily need to create a superseding PR

PR #12866, opened 2024-03-31
chore(deps): bump open api generator in python sdk client to 7.4.0

YatinGupta777:
Fixes #TODO

### Motivation

Python SDK has been using open api generator 5.4.0 which is 2 years old. 

### Modifications

Upgraded to latest open api version 7.4.0 which was closed 3 weeks ago https://github.com/OpenAPITools/openapi-generator/milestones?state=closed


sdks/python/Makefile
```diff
@@ -30,7 +30,7 @@ else
 		sed 's/io.k8s.apimachinery.pkg.apis.meta.v1.//' \
 		> $(WD)/swagger.json
 	cp ../../LICENSE $(WD)/LICENSE
-	$(DOCKER) openapitools/openapi-generator-cli:v5.4.0 \
+	$(DOCKER) openapitools/openapi-generator-cli:v7.4.0 \
```
terrytangyuan, line 33: I think this was pinned to this version for a reason. Have you tested the generated SDK?

YatinGupta777: Havent tested it yet but the sdk unit test is failing so I think the interface has changed. Do you remember the reason for the specific version pinning? Or is there some specific testing that would need to be done?

terrytangyuan:
We should probably hold off on this and start migrating to Hera. cc @flaviuvadan

PR #12867, opened 2024-04-01
feat: allow custom http client and use ctx for http1 apiclient. Fixes #12827

### LINKED ISSUE #12827: HTTP1 client for Argo server should allow more flexible http client config and use context when creating new requests
# Summary

Currently the ArgoServerOpts expose a few config options but it does not allow setting http proxies. We have an API gateway that connects to multiple services across different network zones, so
- we need the http proxy for the connection to argo server
- we can't set the HTTP_PROXY env var because the API gateway is responsible for multiple services across network zones and we can't just set a proxy globally for all the services

What change needs making?

1. Add a httpClient option to [argo server options](https://github.com/argoproj/argo-workflows/blob/main/pkg/apiclient/argo-server-opts.go#L7C1-L17C2) for apiclient
```go
type ArgoServerOpts struct {
	// argo server host port, must be `host:port`, e.g. localhost:2746
	URL string
	// any base path needed (e.g. due to being behind an ingress)
	Path               string
	Secure             bool
	InsecureSkipVerify bool
	// whether or not to use HTTP1
	HTTP1   bool
	Headers []string
        // use this http client when supplied
        HTTPClient *http.Client
}
```
2. Use context when making a new request in [http1 facade](https://github.com/argoproj/argo-workflows/blob/main/pkg/apiclient/http1/facade.go#L100)
```go
func (h Facade) do(ctx context.Context, in interface{}, out interface{}, method string, path string) error {
        ...
	req, err := http.NewRequestWithContext(ctx, method, u.String(), bytes.NewReader(data))
        ...
}
```
3. update the facade interface to take context and all the http1 service clients to pass the context to facade, e.g.
```go
func (h WorkflowServiceClient) CreateWorkflow(ctx context.Context, in *workflowpkg.WorkflowCreateRequest, _ ...grpc.CallOption) (*wfv1.Workflow, error) {
	out := &wfv1.Workflow{}
	return out, h.Post(ctx, in, out, "/api/v1/workflows/{namespace}")
}
```

## Use Cases

The primary use case is that we would like to set a custom roundtripper for the argo server http1 client.

We have an api gateway (authenticated) that connects to multiple instances of argo servers across multiple network zones and when a request for argo server comes in for the api gateway, it needs to do the following:
- pass the bearer token from incoming request to the outgoing request to the target argo server
- set http proxy if the target argo server is in a different network zone
both can be done with a custom roundtripper, e.g.
```go
func RoundTrip(r *http.Request) (*http.Response, error) {
        // set auth header for the outgoing request to argo server
        // based on the request received by the api gateway
	token := r.Context().Value("token").(string)
	r.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
        // set proxy for the outgoing request based on network zones
	secureTransport := &http.Transport{Proxy: http.ProxyURL(s.proxy)}
	return secureTransport.RoundTrip(r)
}
```

When would you use this?

Not urgent, but I think this is nice to have :)

---

williamburgson:
the current implementation does not suite our usecase because the http Fascade interface does not take proxy url we have to use proxy for cross network calls for argo server

so we might as well just allow a fully configured http client to be passed in - this allows more flexible configurations of the HTTP1 client via custom roundtripper

Fixes #12827 

### Motivation

- Allow more flexible configuration for the http1 workflow apiclient so that caller can inject runtime headers (e.g. per-request tracing id) or use a http proxy via roundtripper
- Include the context for the request to argo server so that timeouts can be configured by the caller 

### Modifications

- Add configuration option for `HTTP1Client`
- Have `Facade` use the custom http client if provided
- Use `NewRequestWithContext` to include context from caller
- Update `do` to accept context from caller
- Update service clients to use context, `_ context.Context` -> `ctx context.Context`, and pass that do `Facade`

### DIFF
--- a/pkg/apiclient/http1/facade.go
+++ b/pkg/apiclient/http1/facade.go
@@ -3,6 +3,7 @@ package http1
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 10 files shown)


Joibel (approved):
LGTM. Thanks.

@isubasinghe, could you take a look please?

pkg/apiclient/http1/facade.go
```diff
@@ -65,12 +67,15 @@ func (h Facade) EventStreamReader(in interface{}, path string) (*bufio.Reader, e
...
+	client := h.httpClient
+	if h.httpClient == nil {
+		client = &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: h.insecureSkipVerify,
+				},
 			},
-		},
+		}
```
isubasinghe, line 78: Not sure why we need to do this, I am not entirely familiar with this part of the codebase, what are your opinions on ensuring that client is never null? 

This can be achieved by moving this check to NewFacade.

I am not sure why this wasn't done before.

isubasinghe (approved):
LGTM

PR #12860, opened 2024-03-29
chore(deps): bump github.com/argoproj/argo-events from 1.7.3 to 1.9.1

blkperl:
### Motivation

Add the necessary lint and generated files for https://github.com/argoproj/argo-workflows/pull/12505. Supersedes/Closes #12505

### Modifications

* updated `go.mod` to the latest `argo-events`
* fixed deprecated functions in `test/e2e/argo_server_test.go`
* ran `make pre-commit -B` for generated files

### Verification

None, existing tests should be sufficient

PR #12850, opened 2024-03-27
chore(deps): upgrade Cosign to v2.2.3

suzuki-shunsuke:
Fixes #12828

### Motivation

Recently, Sigstore has published a new TUF trust root.

https://sigstore.slack.com/archives/C01DGF0G8U9/p1710871645742299
https://blog.sigstore.dev/tuf-root-update/

A new TUF trust root doesn't support Cosign v1.13.2, so we should upgrade Cosign to v1.13.6 or v2.

### Modifications

- Upgrade Cosign to v2
- Set the command line option `-y`

```
    -y, --yes=false:
	skip confirmation prompts for non-destructive operations
```

Without this option, `cosign sign` and `cosign sign-blob` don't work in CI.

PR #12763, opened 2024-03-08
fix(deps): upgrade `undici` from 5.28.2 to 5.28.3 due to CVE

agilgur5:
Partial fix for #12031, "Vulnerabilities"

### Motivation

After the Snyk failure for #12753, I also checked OpenSSF Scorecard's current vuln report and found this vuln too: https://github.com/nodejs/undici/security/advisories/GHSA-3787-6prv-h9w3

### Modifications

- update `yarn.lock` to bump `undici`
  - `undici` is a transitive dep, used by `swagger-ui-react`
     - `swagger-ui-react` -> `swagger-client` -> `undici`

### Verification

1. Ran `yarn` after, no additional changes
1. Ran `npx yarn-audit-fix && yarn deduplicate` to check for more vulns and auto-update them (same as #12036), no others popped up
    - Same for plain `yarn audit`: `0 vulnerabilities found - Packages audited: 1488`
1. Ran `yarn` again, no additional changes

PR #12881, opened 2024-04-04
ci: use `chore(deps)` prefix for dependabot commits

agilgur5:
### Motivation

Dependabot's recent automated security updates (#12878, #12879, #12880)  (enabled after https://github.com/argoproj/argo-workflows/pull/12763#issuecomment-1986755883) used the `build(deps)` prefix instead of `chore(deps)`
  - note that I changed all their PR titles manually, but their commits still say `build(deps)` and most were auto-merged before I could change the title).

This matches the prefix we normally use for dependency updates (#12850, #12860, etc)
  - also this matches what dependabot was previously doing for non-security updates (#12512, #12507, etc), not sure why it used `build(deps)` for security updates (??)

### Modifications

- Add `commit-message.prefix: chore(deps)` to all `package-ecosystem`s in `dependabot.yml`
  - dependabot auto-adds `: ` after the prefix
  - Add `commit-message.prefix-development: chore(deps-dev)` to `package-ecosystem: npm` (not supported for `github-actions` or `gomod`)
    - Matches previous dependabot non-security updates, such as #12511, #12510, #12509, etc

### Verification

Matches the GH Docs in https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,8 @@ updates:
     labels:
       - type/dependencies
       - go
+    commit-message:
+      prefix: chore(deps)
 
   - package-ecosystem: "npm"
     directory: "/ui"
@@ -31,6 +33,9 @@ updates:
     labels:
       - type/dependencies
       - javascript
+    commit-message:
+      prefix: chore(deps)
+      prefix-development: chore(deps-dev)
 
   # build / CI dependencies
   - package-ecosystem: "pip"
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

agilgur5:
Confirmed this is working in #12891

PR #12894, opened 2024-04-05
chore(deps): upgrade `mkdocs-material` from 8.2.6 to 9.x

agilgur5:
Partial fix for #12031, "Vulnerabilities"

### Motivation

- [8.2.6](https://github.com/squidfunk/mkdocs-material/releases/tag/8.2.6) is 2 years old now
  - and there hasn't been an 8.x release since [9.0.0](https://github.com/squidfunk/mkdocs-material/releases/tag/9.0.0) was released 1.5 years ago
  - in particular, there are some CVEs in transitive deps of `mkdocs-material` now (such as `urllib3` CVEs: [GHSA-g4mx-q9vg-27p4](https://github.com/urllib3/urllib3/security/advisories/GHSA-g4mx-q9vg-27p4), [GHSA-hmv2-79q8-fv6g](https://github.com/advisories/GHSA-hmv2-79q8-fv6g), [GHSA-v845-jxx5-vc9f](https://github.com/advisories/GHSA-v845-jxx5-vc9f), etc)

### Modifications

- make the upgrade to 9.x in `docs/requirements.txt`
  - none of the removals in 9.0.0 affect our usage
  - main noticeable changes I can see are some CSS improvements

- also remove fixed dep pins in `docs/requirements.txt` and use a range
  - ideally we should use [`poetry`](https://github.com/python-poetry/poetry/) or similar and create a lockfile, but that's a separate topic with many more changes required -- this PR is focused on the upgrade

### Verification

1. `make docs`
1. `open site/index.html`
1. <details><summary>Took a look around, all seemed well to me. Mermaid plugin works, Field Reference with <code>md_in_html</code> works, admonitions work</summary>

    - Screenshot of how admonitions look like now (compare to https://github.com/argoproj/argo-workflows/pull/12809#issuecomment-2036237459): 
Screenshot 2024-04-05 at 1 02 05 AM

</details>

### Notes to Reviewers

- Dependabot didn't update this, I think that might have been because of either (or all of):
    - the fixed dep with no range 
    - that there were no patches within the `mkdocs-material` 8.x range
    - no lockfile to independently update `urllib3` etc without touching `mkdocs-material`

### Future Work

Add `poetry` or `Pipfile` config and a lockfile

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,2 @@
-mkdocs-material==8.2.6
-mkdocs-redirects==1.2.1
+mkdocs-material==9.*
+mkdocs-redirects==1.*


agilgur5:
> Add `poetry` or `Pipfile` config and a lockfile

Noting that both [are supported by dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem), as well as `pip-compile`

I was considering `Pipfile` just because it's default installed in many Python distributions, but `Pipfile` isn't officially supported by ReadTheDocs: https://github.com/readthedocs/readthedocs.org/issues/3181#issuecomment-975378486.
`Pipfile` seems to be dead? https://github.com/pypa/pipfile had a [last commit two years ago](https://github.com/pypa/pipfile/commit/4706d2cbd35e0b47a05a6421fa17f93827bc454f). That would explain the dependabot docs saying they only support versions "<= 2021-05-29".
Not sure if Hatch is the new spec-compliant replacement https://github.com/pypa/hatch (both under `pypa` GH org)?

I prefer `poetry` myself (and contributed to it years ago), and it has [officially documented instructions for ReadTheDocs](https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry), so I guess that makes that decision easy moving forward

agilgur5:
Oh I apparently haven't done a deep-dive on Python builds in [a bit](https://github.com/agilgur5/django-serializable-model/commit/0a5eeb2e0d7bd82cb6e083ce6588993f8c7f2696). With [PEP 517](https://www.python.org/dev/peps/pep-0517/) build backends were fully split (see also [`poetry-core`](https://github.com/python-poetry/poetry-core)), so now can install a `poetry` `pyproject.toml` even with `pip` (although it won't use the `poetry.lock`). This may be good enough for our use-cases -- or at least a first pass, so as to not have to change the build system too much (e.g. `Makefile` etc).

Need to do some more testing but working on that.

blkperl (approved):

terrytangyuan (approved):

agilgur5:
Backported cleanly to `release-3.5` as a20f69571

agilgur5:
> Partial fix for #12031, "Vulnerabilities"

> * in particular, there are some CVEs in transitive deps of `mkdocs-material` now (such as `urllib3` CVEs: [GHSA-g4mx-q9vg-27p4](https://github.com/urllib3/urllib3/security/advisories/GHSA-g4mx-q9vg-27p4), [GHSA-hmv2-79q8-fv6g](https://github.com/advisories/GHSA-hmv2-79q8-fv6g), [GHSA-v845-jxx5-vc9f](https://github.com/advisories/GHSA-v845-jxx5-vc9f), etc)

OpenSSF Scorecard is still detecting these Python CVEs, so I think it's actually finding them in the Python SDK, which does [use `urllib3` v1 directly](https://github.com/argoproj/argo-workflows/blob/13d553c45d4d9e7d3a27674e0e25f515e78629fe/sdks/python/client/requirements.txt#L3). Since it's generated though, we can't just upgrade it... that actually requires a new version of `openapi-generator-cli`, which requires a breaking change to the SDK 😕 . See also #12866

PR #12898, opened 2024-04-05
fix(ui): handle non-existent labels in `isWorkflowInCluster`

agilgur5:
Fixes [this Slack thread](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1712142240797399), no formal bug report was filed unfortunately

### Motivation

Fixes a bug where the Workflow Details page would error out with: `TypeError: Cannot read properties of undefined (reading 'workflows.argoproj.io/workflow-archiving-status')`

### Modifications

- use optional chaining to handle case of no labels

- similar to changes in #12170 for `isArchivedWorkflow`, which is just below `isWorkflowInCluster`

### Verification

No repro was provided unfortunately, but these are the only two places where this label is accessed and so could result in such an error.


tczhao (approved):

toyamagu-2021 (approved):
I'm not sure when the label becomes undefined, but this should solve the issue.

isubasinghe (approved):

PR #12904, opened 2024-04-06
test: Add test for failed workflow artifact garbage collection.

Garett-MacGowan:
Related discussion: #12845

### Motivation

#12845 suggested a possible issue with artifactgc for failed workflows. This test verifies that artifactgc works for failed workflows.

### Modifications

Add test for failed workflow artifact garbage collection.

### DIFF
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -224,6 +224,59 @@ func (s *ArtifactsSuite) TestStoppedWorkflow() {
 	}
 }
 
+func (s *ArtifactsSuite) TestFailedWorkflow() {
+	c, err := minio.New("localhost:9000", &minio.Options{
+		Creds: credentials.NewStaticV4("admin", "password", ""),
+	})
+	assert.NoError(s.T(), err)
+
+	// Ensure the artifacts aren't in the bucket.
+	_, err = c.StatObject(context.Background(), "my-bucket-3", "on-deletion-wf-failed", minio.StatObjectOptions{})
+	if err == nil {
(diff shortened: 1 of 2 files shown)


test/e2e/artifacts_test.go
```diff
@@ -224,6 +224,59 @@ func (s *ArtifactsSuite) TestStoppedWorkflow() {
 	}
 }
 
+func (s *ArtifactsSuite) TestFailedWorkflow() {
```
tczhao, line 227: The failed workflow is part of `fixtures.ToBeCompleted`
maybe we can include this in the https://github.com/argoproj/argo-workflows/blob/34967127bfe304da0f72dc1dba63558a1609b2d0/test/e2e/artifacts_test.go#L247 instead?

Joibel:
Thanks for putting this together, and sorry it sat for so long.

The problem behind #12845 turned out to be incomplete artifact status entries, fixed in #13066, which also landed a regression test for exactly this shape: a workflow that fails after an earlier step wrote an artifact, asserting the written artifact is still garbage collected on workflow deletion. #13678 later generalised that into the NonOptionalArtifactNotWritten / OptionalArtifactNotWritten cases in TestArtifactGC (workflowShouldSucceed: false, artifact-written/present expected deleted at workflow deletion), and main now also has ArtifactNotWrittenFailed cases. All of that shipped in v3.6.0.

Given the coverage already exists, and that the very similar TestStoppedWorkflow has since had to be skipped as flaky, I'm closing this rather than adding a second test of the same shape. If you think there's a distinct case still uncovered, please say so and we can reopen.

Issue #12739, opened 2024-03-05
Expr does not work as expected after upgrading to 3.5.5


vermaxik:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

We use JSON a lot in our project, and in some cases an object returns `null`. We have conditions or relations between tasks and would like to return `null` if there no nested object instead of non-parsed go template.

For this cases here is [optional chaining](https://github.com/expr-lang/expr/blob/master/docs/language-definition.md#optional-chaining) in expr-lang.

We were waiting for this in 3.5, but  after upgrading from 3.5.4 to 3.5.5 it does not work as expected. But in [playground](https://expr-lang.org/playground) it works as before.

### Version

3.5.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
Here is simple workflow:

apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: optinal-field
spec:
  arguments:
    parameters:
      - name: config
        value: '{"a": {"nested_a": 1}}'
      - name: config_null
        value: '{"a": null}'
  entrypoint: main
  templates:
    - name: main
      inputs:
        parameters:
          - name: a
            value: "{{= fromJSON(workflow.parameters.config).a?.nested_a }}"
          - name: a_null
            value: "{{= toJSON(fromJSON(workflow.parameters.config_null).a?.nested_a ?? null) }}"
      script:
        image: debian:9.4
        command: [bash]
        source: |
          echo "{{= inputs.parameters.a }}"
          echo "{{= inputs.parameters.a_null }}"
```

3.5.4 (empty object returns null)
<img width="552" alt="Screenshot 2024-03-05 at 11 08 37" src="https://github.com/argoproj/argo-workflows/assets/2705386/1f799588-f016-4aee-ae83-6ca417f36659">

3.5.5 
<img width="578" alt="Screenshot 2024-03-05 at 11 00 32" src="https://github.com/argoproj/argo-workflows/assets/2705386/46142a91-9eae-478a-a855-e73de9603ecb">
```


### Logs from the workflow controller

```text
[...]

PR #12909, opened 2024-04-08
fix(templating): return json, not go structs

b-erdem:
### Motivation

When you extract a value from json using jsonpath expressions it returns values properly, only for simple values. For instance when you pass the following JSON:
`{"employees": [{"name": "Baris", "age":43},{"name": "Mo", "age": 42}, {"name": "Jai", "age" :44}]}` and `$.employees[0].name` it'll return "Baris" correctly. But let's say if you want to return an item in the array, for instance with `$.employees[0]` query then you'll get `map[age:43 name:Baris]`. This is not correct, it returns internal representation of the value when returned value is not a simple value.

We're passing some complex values to each steps and some of the values returned from a jsonpath expression could be a complex object. So we face with issues and doing some workarounds to get this correctly. Instead I thought maybe it'd be useful to fix it directly here. With this PR we'll be able to get values correctly from jsonpath expressions. For instance `$.employees[0]` now would return: `{"age":43,"name":"Baris"}`  

### Modifications

The returned value is now passed to `json.Marshal` directly without formatting it. `fmt.Sprintf("%v", result))` this actually causes getting internal representation of the value.

The rest of the code is making sure we're returning a correct json value/object, mainly related to quotes. The changes I did in this PR is to cover when the returned object is a complex object and it contains strings, so we'll have quotes inside of the value. Then we need to properly encode them.

### Verification

* Added some unit tests.
* Added an example with a complex return value.
* Tested locally with some of our workflows.

Please let me know if there's anything missing or something I can improve more. Thanks!


util/template/expression_template.go
```diff
@@ -80,9 +81,20 @@ func expressionReplace(w io.Writer, expression string, env map[string]interface{
...
 	marshaledLength := len(resultMarshaled)
-	return w.Write(resultMarshaled[1 : marshaledLength-1])
+
+	// Trim leading and trailing quotes. The value is being inserted into something that's already a string.
+	if len(resultMarshaled) > 1 && resultMarshaled[0] == '"' && resultMarshaled[marshaledLength-1] == '"' {
+		return w.Write(resultMarshaled[1 : marshaledLength-1])
+	}
+
+	// Check if the value contains any quotes by adding double quotes at the beginning and end of the value.
+	// If so then escape quotes.
```
Joibel, line 92: Why are we escaping quotes here? Shouldn't the user should be free to choose whether they want to do this or not?

b-erdem: As you can see from the code line below, we're checking if we surround the result value with quotes is a valid JSON or not. The reason is, the value could be an array or object and if it's invalid(and you wouldn't want to return an invalid JSON value) when you surround it with quotes then it means it contains quotes inside of it already, so we escape it(and you wouldn't want to return an invalid JSON value).

Here is an example what it returns with these new changes and current code:

You can `Replace` function `util/template` mod with this input:
`{{=jsonpath('{"employees": [{"name": "Baris", "age":43}, {"name": "Mo", "age": 42}]}', '$.employees[*].name')}}`

With this new code, it returns: `{"value":"[\"Baris\",\"Mo\"]"}` which is a valid value.
And with current code, it returns: `{"value":""Baris","Mo""}` as you can see this is an invalid value.

You can experiment with it enabling/disabling my changes here and by modifying/tweaking test cases I added in `template_test.go` file.

I hope this answers your question.

Joibel: Yes, this does.
I was wondering if it would just make sense to simplify it down to always just quote.

```
    resultQuoted := []byte(strconv.Quote(string(resultMarshaled)))
	return w.Write(resultQuoted[1 : len(resultQuoted)-1])
```

The json.Valid test is probably more expensive than just quoting always.

As it is the code is hard to comprehend why anyone might need to check that json.Marshal returns valid json.

Joibel (changes requested):

util/template/expression_template.go
b-erdem: Actually, you're right. We don't need to check if the result is correct json or not, we can always quote. I did the change. Thanks!

Joibel (approved):
LGTM.

Thanks @b-erdem for your contribution.

b-erdem:
Hello @agilgur5 I think I need another approval to get this merged. :)

Joibel:
This is a breaking change from previous bad but usable behaviour. I am not sure whether this should be held until 3.6. If it goes in 3.5.6 it should be highlighted in the notes.

agilgur5:
@Joibel but did the previous behavior ever make sense / was it ever useful? From a quick glance, it sounds like it would always cause an error, so then we could straightforwardly merge this as a patch and wouldn't even necessarily need notes as the previous behavior did not work. 

I feel like there _might_ have been a use for this behavior before but I couldn't think of one off the top of my head

Joibel:
The previous behaviour didn't result in an error, was deterministic and predictable (you got the go printf `%v` representation).

This could have been used, it contains the values you want, just not in the representation expected. I wouldn't be surprised if someone somewhere does rely on this, and so we should point out that we are changing it.

b-erdem:
@Joibel @agilgur5 Hello! Do we have a plan when we want to merge this? 🤔

b-erdem:
> The previous behaviour didn't result in an error, was deterministic and predictable (you got the go printf `%v` representation).
> 
> This could have been used, it contains the values you want, just not in the representation expected. I wouldn't be surprised if someone somewhere does rely on this, and so we should point out that we are changing it.

Hi! As far as I tested, nothing changes for simple values. For complex values, since it returns go struct, it's not usable when you pass to the next steps(I'm not sure if there are cases where it works fine, though).

Even if it causes some issues for some people, it looks like an easy change to do. 🤔 

Lastly, do you have an idea to make sure we don't break anything? cc: @agilgur5

b-erdem:
> I am happy to merge this and put something into the release notes. It was wrong, but didn't error and was usable, so wanted to call that out.

I also think that way. Thanks! Who should I ask for additional approval that gives me the right to merge this PR then? 🤔

Joibel:
I'll merge it now, the checks weren't green, but they are now.

agilgur5:
Just re-stumbled upon #8930, which includes workarounds users have been using for some time that we would not want to break

Joibel:
Notes for maintainers: this will come out in 3.6 and should be noted in the release notes. Please don't backport to any other branches.

agilgur5:
Mmm if this is going to be a breaking change, the release note should really have been included in this PR too -- we already have some for 3.6 on `main` (e.g. #12653).
Also test cases would still be useful, as if there is some breakage for those cases, we can explicitly note them.

@Joibel @b-erdem can either of you work on that?

b-erdem:
@agilgur5 Hi, I’ll look into this next week.

agilgur5:
@Joibel @isubasinghe could one of you take over the release notes and test cases for this?
Since @b-erdem has not added that and it is pretty critical to do prior to a release as well as before we forget about this

The cases I mentioned [above](https://github.com/argoproj/argo-workflows/pull/12909#issuecomment-2167422961) of expressions on structs such as in #8930 and #12739 in particular need test cases and need checking for breakage.

agilgur5:
I wrote up a follow-up issue to track the above in #13541

PR #12912, opened 2024-04-08
fix: don't load entire archived workflow into memory in list APIs

jiachengxu:
Fixes https://github.com/argoproj/argo-workflows/pull/11121#discussion_r1469101802

This PR fixes the regression introduced in https://github.com/argoproj/argo-workflows/pull/11121#discussion_r1469101802. In this PR, we don't load the entire workflow to the memory since that causes high memory usage and slow query performance.


jessesuen (approved):
LGTM!

persist/sqldb/workflow_archive.go
```diff
@@ -174,16 +184,35 @@ func (r *workflowArchive) ListWorkflows(namespace string, name string, namePrefi
...
+				CreationTimestamp: v1.Time{Time: md.StartedAt},
+				Labels:            labels,
+				Annotations:       annotations,
+			},
+			Status: wfv1.WorkflowStatus{
+				Phase:      md.Phase,
+				StartedAt:  v1.Time{Time: md.StartedAt},
+				FinishedAt: v1.Time{Time: md.FinishedAt},
+				Progress:   wfv1.Progress(md.Progress),
+			},
```
agilgur5, line 215: What did you base the metadata requirements on? Based on [`workflows-service.ts`](https://github.com/argoproj/argo-workflows/blob/bbc1c08ce20e15b07de37f61f4a12889ec962623/ui/src/app/shared/services/workflows-service.ts#L39) and [`reports.tsx`](https://github.com/argoproj/argo-workflows/blob/bbc1c08ce20e15b07de37f61f4a12889ec962623/ui/src/app/reports/reports.tsx#L59), we may also want `status.message`, `status.estimatedDuration`, and `spec.suspend`

agilgur5:
Backported cleanly into `release-3.5` as 200f4d1e5

persist/sqldb/workflow_archive.go
jiachengxu: I was checking columns in the workflow UI and decided what metadata to include, but if the fields that you mentioned are required, I can open another PR to add them

agilgur5: Technically speaking, `fields` is a query parameter in the API, so ideally we'd dynamically set the fields.

If you didn't notice some of them it may be because you can expand the list?

jiachengxu: Oh, that could be, I will prepare a PR to add those fields.

agilgur5: This seems to be causing a regression now: #13098

agilgur5: Hmm looking at https://github.com/argoproj/argo-workflows/pull/11121#discussion_r1469101802 more, it may have also unintentionally introduced a feature where you can get all parts of an Archived Workflows in the `list` API, whereas you could not in 3.4 if I'm reading correctly.

I.e. fixing that bug also removed that feature. I don't necessarily mind that if it wasn't present in 3.4, but I can see how that could cause some back-and-forth turbulence for users (which, well, #11121 caused a lot of regressions unfortunately)

jiachengxu: PR to add the above missing fields: https://github.com/argoproj/argo-workflows/pull/13136

PR #12917, opened 2024-04-09
refactor: remove unnecessary `AddEventHandler` error handling

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/12847#discussion_r1556488443

### Motivation

- in k8s 1.26+, the `informer.AddEventHandler` function can return an error
  - however, this error will only happen when the informer [has been stopped already](https://github.com/kubernetes/client-go/blob/46588f2726fa3e25b1704d6418190f424f95a990/tools/cache/shared_informer.go#L580)
  - in the `new*Informer` functions, we literally just created them, with `AddEventHandler` on the very next line, so this error is impossible in these cases
    - so ignore the error and remove some error handling code around that

### Modifications

- add `//nolint:errcheck` directives with comments on errors that can be safely ignored
  - remove unnecessary error handling code
- remove some extra new lines

### Verification

lint passes, all E2Es pass. checked the informer source code myself.


blkperl (approved):
Thanks for cleaning this up!

isubasinghe (approved):

PR #12901, opened 2024-04-06
fix(deps): upgrade `http2` to v0.24. Fixes CVE-2023-45288

terrytangyuan:
Fixes CVE-2023-45288

This fixes the failing Snyk check on main branch.

PR #12921, opened 2024-04-10
fix(deps): upgrade `x/net` to v0.23.0. Fixes CVE-2023-45288

terrytangyuan:
Fixes CVE-2023-45288


go.mod
```diff
@@ -295,4 +295,4 @@ require (
 )
 
 // Avoid CVE-2023-45288
-replace golang.org/x/net/http2 => golang.org/x/net/http2 v0.24.0
+replace golang.org/x/net => golang.org/x/net v0.23.0
```
agilgur5, line 298: can't we change the indirect dep on line 275 instead?

agilgur5:
Follow-up to #12901, which didn't quite work

go.mod
agilgur5: Since this is a breaking change (and #12901 even more so if it worked, as it was to v0.24.0), I tried finding a changelog to see if this was safe to do. I couldn't find a changelog for `x/net` anywhere, so I diffed [`v0.22.0` with `v0.23.0`](https://github.com/golang/net/compare/v0.22.0...v0.23.0). Fortunately, there are few non-test and non-security changes, so I think this is safe to upgrade and low risk

agilgur5: From Terry [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1712776585174859?thread_ts=1712368909.910789&cid=C0510EUH90V):
> That would work too but I prefer this more explicit/common approach

sounds good to me then, I'm not a `go.mod` expert

agilgur5 (approved):
LGTM, confirmed that this seems to be a low risk upgrade [above](https://github.com/argoproj/argo-workflows/pull/12921#discussion_r1559948154) (in terms of breaking changes)

agilgur5:
Backported to `release-3.5` as 1c09db42ec69540ec64e5dd60a6daef3473c6783
- Still had a merge conflict in the `go.sum`, mostly in `x/` (standard library). I took this commit's versions and then ran `go mod tidy` to get it correct to the `go.mod`.

PR #12929, opened 2024-04-12
fix: Correct log level for agent containers

terrytangyuan:
The log level is currently not passed to main and init containers in agent pod.

### DIFF
--- a/workflow/controller/agent.go
+++ b/workflow/controller/agent.go
@@ -198,11 +198,11 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 	// the `init` container populates the shared empty-dir volume with tokens
 	agentInitCtr := agentCtrTemplate.DeepCopy()
 	agentInitCtr.Name = common.InitContainerName
-	agentInitCtr.Args = []string{"agent", "init"}
+	agentInitCtr.Args = []string{"agent", "init", "--loglevel", getExecutorLogLevel()}
 	// the `main` container runs the actual work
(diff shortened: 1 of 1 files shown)


shuangkun (approved):

sarabala1979 (approved):

workflow/controller/agent.go
```diff
@@ -198,11 +198,11 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 	// the `init` container populates the shared empty-dir volume with tokens
 	agentInitCtr := agentCtrTemplate.DeepCopy()
 	agentInitCtr.Name = common.InitContainerName
-	agentInitCtr.Args = []string{"agent", "init"}
+	agentInitCtr.Args = []string{"agent", "init", "--loglevel", getExecutorLogLevel()}
```
agilgur5, line 201: we probably also want to add `--log-format` and `--gloglevel` as it [accepts all 3](https://github.com/argoproj/argo-workflows/blob/13d553c45d4d9e7d3a27674e0e25f515e78629fe/cmd/argoexec/commands/root.go#L72-L74). `--log-format` is used in [a few places](https://github.com/argoproj/argo-workflows/blob/13d553c45d4d9e7d3a27674e0e25f515e78629fe/workflow/controller/workflowpod.go#L606) already as well.

I'll work on that as I'm working on a `--log-format` fix already

agilgur5:
Backported cleanly into `release-3.5` as 01f843828

workflow/controller/agent.go
agilgur5: Completed in #12979

PR #12930, opened 2024-04-12
fix(ui): remove unnecessary hard reload after delete

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/11121#discussion_r1561987732

### Motivation

- remove the hard `window.location.reload()` on deletion b/c the WorkflowsList page _does_ handle deletions
  - not sure what didn't work about it before, the code has improved but is largely semantically equivalent
    - I suspect it might have been the same root cause as #11711, i.e. caused by #11661

### Modifications

- remove the `window.location.reload()` and comments about it
  - `navigation.goto` is already there and already works properly

### Verification

<details><summary>Tested deletion locally without the <code>window.location.reload()</code> and it worked fine. </summary>

Screenshots:

Before deletion -- `coinflip-794pg` exists:
Screenshot 2024-02-17 at 1 37 37 PM

After deletion -- clicked into `coinflip-794pg` and clicked _Delete_  -- `coinflip-794pg` no longer exists:
Screenshot 2024-02-17 at 1 37 13 PM

</details>

### Notes to Reviewers

Might notice that my screenshots are nearly 2 months old 😅 I haven't gotten back to UI work in a bit and this was leftover code that I had in my `git stash`

### DIFF
--- a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
+++ b/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
@@ -214,9 +214,6 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
                                     }
 
                                     navigation.goto(uiUrl(`workflows/${workflow.metadata.namespace}`));
-                                    // TODO: This is a temporary workaround so that the list of workflows
-                                    //  is correctly displayed. Workflow list page needs to be more responsive.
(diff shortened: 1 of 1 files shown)


tczhao (approved):
make sense,
Verified server deletes the resource before returning a response

isubasinghe (approved):

PR #12931, opened 2024-04-12
fix: change fatal to panic. 

shuangkun:
comments here: https://github.com/argoproj/argo-workflows/pull/12817#discussion_r1561980703

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2824,7 +2824,7 @@ func (woc *wfOperationCtx) executeContainer(ctx context.Context, nodeName string
 func (woc *wfOperationCtx) getOutboundNodes(nodeID string) []string {
 	node, err := woc.wf.Status.Nodes.Get(nodeID)
 	if err != nil {
-		woc.log.Panic("was unable to obtain node for %s", nodeID)
+		woc.log.Panicf("was unable to obtain node for %s", nodeID)
 	}
 	switch node.Type {
 	case wfv1.NodeTypeSkipped, wfv1.NodeTypeSuspend, wfv1.NodeTypeHTTP, wfv1.NodeTypePlugin:
@@ -3266,7 +3266,7 @@ func (woc *wfOperationCtx) addChildNode(parent string, child string) {
 	childID := woc.wf.NodeID(child)
 	node, err := woc.wf.Status.Nodes.Get(parentID)
 	if err != nil {
-		woc.log.Panic("was unable to obtain node for %s", parentID)
+		woc.log.Panicf("was unable to obtain node for %s", parentID)
 	}
 	for _, nodeID := range node.Children {
 		if childID == nodeID {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2824,7 +2824,7 @@ func (woc *wfOperationCtx) executeContainer(ctx context.Context, nodeName string
 func (woc *wfOperationCtx) getOutboundNodes(nodeID string) []string {
 	node, err := woc.wf.Status.Nodes.Get(nodeID)
 	if err != nil {
-		woc.log.Fatalf("was unable to obtain node for %s", nodeID)
+		woc.log.Warnf("was unable to obtain node for %s", nodeID)
 		panic(fmt.Sprintf("Expected node for %s", nodeID))
 	}
 	switch node.Type {
@@ -3267,7 +3267,7 @@ func (woc *wfOperationCtx) addChildNode(parent string, child string) {
 	childID := woc.wf.NodeID(child)
 	node, err := woc.wf.Status.Nodes.Get(parentID)
 	if err != nil {
-		woc.log.Fatalf("was unable to obtain node for %s", parentID)
+		woc.log.Warnf("was unable to obtain node for %s", parentID)
 		panic(fmt.Sprintf("parent node %s not initialized", parent))
 	}
 	for _, nodeID := range node.Children {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2824,8 +2824,7 @@ func (woc *wfOperationCtx) executeContainer(ctx context.Context, nodeName string
 func (woc *wfOperationCtx) getOutboundNodes(nodeID string) []string {
 	node, err := woc.wf.Status.Nodes.Get(nodeID)
 	if err != nil {
-		woc.log.Warnf("was unable to obtain node for %s", nodeID)
-		panic(fmt.Sprintf("Expected node for %s", nodeID))
+		woc.log.Panic("was unable to obtain node for %s", nodeID)
 	}
 	switch node.Type {
 	case wfv1.NodeTypeSkipped, wfv1.NodeTypeSuspend, wfv1.NodeTypeHTTP, wfv1.NodeTypePlugin:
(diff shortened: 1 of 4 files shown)


shuangkun:
@agilgur5  could you

workflow/progress/updater.go
```diff
@@ -72,7 +72,7 @@ func sumProgress(wf *wfv1.Workflow, node wfv1.NodeStatus, visited map[string]boo
 		child, err := wf.Status.Nodes.Get(childNodeID)
 		if err != nil {
 			log.Warnf("Coudn't obtain child for %s, panicking", childNodeID)
-			continue
+			panic("was not able to get node")
```
agilgur5, line 75: I don't think we should change this one, it was intentionally `continue`d and _not_ `panic`'d

```suggestion
			continue
```

workflow/util/util.go
```diff
@@ -764,8 +764,8 @@ func getDescendantNodeIDs(wf *wfv1.Workflow, node wfv1.NodeStatus) []string {
 	for _, child := range node.Children {
 		childStatus, err := wf.Status.Nodes.Get(child)
 		if err != nil {
-			log.Fatalf("Couldn't get child, panicking")
-			panic("Was not able to obtain child")
+			log.Warnf("Coudn't obtain child for %s, panicking", child)
+			panic("was not able to obtain child")
```
agilgur5, line 768: I was gonna recommend outputting more details here too 👍 

also apparently there is a shorthand for this as well: [`log.Panic`](https://pkg.go.dev/log#Panic)

```suggestion
			log.Panicf("Coudn't obtain child for %s, panicking", child)
```

also `Error` is the appropriate log level here if we're panicking, vs. `Warn`

workflow/util/util.go
```diff
@@ -776,8 +776,8 @@ func isDescendantNodeSucceeded(wf *wfv1.Workflow, node wfv1.NodeStatus, nodeIDsT
...
```
agilgur5, line 780:
```suggestion
			log.Panicf("Coudn't obtain child for %s, panicking", child)
```

workflow/util/util.go
```diff
@@ -816,8 +816,8 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow,
 	for {
 		currentNode, err := oldWF.Status.Nodes.Get(currentNodeID)
 		if err != nil {
-			log.Fatalf("dying due to inability to obtain node for %s", currentNodeID)
-			panic("was not able to get node, panicking")
+			log.Warnf("dying due to inability to obtain node for %s, panicking", currentNodeID)
+			panic("was not able to get node")
```
agilgur5, line 820:
```suggestion
			log.Panicf("unable to obtain node for %s, panicking", currentNodeID)
```

same as above

workflow/util/util.go
```diff
@@ -827,7 +827,7 @@ func resetConnectedParentGroupNodes(oldWF *wfv1.Workflow, newWF *wfv1.Workflow,
 		if currentNode.BoundaryID != "" && currentNode.BoundaryID != oldWF.ObjectMeta.Name {
 			parentNode, err := oldWF.Status.Nodes.Get(currentNode.BoundaryID)
 			if err != nil {
-				log.Fatalf("panicking unable to obtain node for %s", currentNode.BoundaryID)
+				log.Warnf("unable to obtain node for %s, panicking", currentNode.BoundaryID)
 				panic("was not able to get node")
```
agilgur5, line 831:
```suggestion
				log.Panicf("unable to obtain node for %s, panicking", currentNode.BoundaryID)
```

same as above

agilgur5 (commented):
some small changes -- we can use `log.Panic` actually

tczhao (changes requested):
Up for a discussion,

This util `FormulateResubmitWorkflow` function
is only run by the argo servers,
I'm thinking maybe we should remove and replace all panic/fatal in this function with error/warn
so the server keeps alive and returns the appropriate status code,err

workflow/controller/operator.go
```diff
@@ -2824,8 +2824,7 @@ func (woc *wfOperationCtx) executeContainer(ctx context.Context, nodeName string
 func (woc *wfOperationCtx) getOutboundNodes(nodeID string) []string {
 	node, err := woc.wf.Status.Nodes.Get(nodeID)
 	if err != nil {
-		woc.log.Fatalf("was unable to obtain node for %s", nodeID)
-		panic(fmt.Sprintf("Expected node for %s", nodeID))
+		woc.log.Panic("was unable to obtain node for %s", nodeID)
```
agilgur5, line 2827:
```suggestion
		woc.log.Panicf("was unable to obtain node for %s", nodeID)
```
missing `f`

workflow/controller/operator.go
```diff
@@ -3267,8 +3266,7 @@ func (woc *wfOperationCtx) addChildNode(parent string, child string) {
 	childID := woc.wf.NodeID(child)
 	node, err := woc.wf.Status.Nodes.Get(parentID)
 	if err != nil {
-		woc.log.Fatalf("was unable to obtain node for %s", parentID)
-		panic(fmt.Sprintf("parent node %s not initialized", parent))
+		woc.log.Panic("was unable to obtain node for %s", parentID)
```
agilgur5, line 3269:
```suggestion
		woc.log.Panicf("was unable to obtain node for %s", parentID)
```

missing `f`

agilgur5:
> is only run by the argo servers,
> I'm thinking maybe we should remove and replace all panic/fatal in this function with error/warn
> so the server keeps alive and returns the appropriate status code,err

Yes, that's why I suggested the `panic` vs. `fatal`. Panics are recoverable and we have a few `recover` statements and functions.

Regarding the status code, these would generally all be 500s; they're indicative of a bug. I would think a `panic` would end up returning a 500, but I'm not sure

shuangkun:
> Up for a discussion,
> 
> This util `FormulateResubmitWorkflow` function is only run by the argo servers, I'm thinking maybe we should remove and replace all panic/fatal in this function with error/warn so the server keeps alive and returns the appropriate status code,err

Ok, changed fatal to error in `FormulateResubmitWorkflow`

workflow/util/util.go
```diff
@@ -699,7 +699,7 @@ func FormulateResubmitWorkflow(ctx context.Context, wf *wfv1.Workflow, memoized
 	onExitNodeName := wf.ObjectMeta.Name + ".onExit"
 	err := packer.DecompressWorkflow(wf)
 	if err != nil {
-		log.Fatal(err)
+		log.Error(err)
```
agilgur5, line 702: these aren't equivalent -- `log.Error` doesn't `exit`. so we should either return the error or panic. for equivalency with the previous logic, it would be `log.Panic`

shuangkun: OK, changed

tczhao:
> Yes, that's why I suggested the panic vs. fatal. Panics are recoverable and we have a few recover statements and functions.

right, found recover https://github.com/argoproj/argo-workflows/blob/main/util/grpc/interceptor.go#L21

agilgur5 (approved):
LGTM, thanks for fixing these cases

tczhao (approved):

agilgur5:
@tczhao for reference all these node not found errors were added in #11451 as there were some actual bugs where nodes didn't exist but silently returned fine.

in the cases where we _know for sure_ these errors will never happen, we should ignore the errors and not handle them at all, similar to what I did in #12917 .

the problem is that we don't know all the cases 😕 
eventually these should all really be removed or only exist in a few places where, for instance, the `status` got corrupted and is not actually in the Controller's ability to fix, like if a user literally re-wrote the `status` with `kubectl`

tczhao:
@agilgur5 Got it, thank you for the context

agilgur5:
Backported cleanly to `release-3.5` as 9c2581ad0

PR #12932, opened 2024-04-13
fix(ui): properly get archive logs when no node was chosen

agilgur5:
Fixes #12590, Fixes #12759
- Partially replaces #12591 with a root cause fix (that I wrote & pushed [myself](https://github.com/argoproj/argo-workflows/pull/12591#discussion_r1562071827) to the PR) and without an additional feature, so that this can be properly backported to 3.4.x and 3.5.x

Follow-up to #9464

### Motivation

- a node doesn't have to be clicked on in the UI to get logs, you can also click on the top-level logs button
  - in this case, there is no nodeId, but we can infer/derive one based on the Pod chosen in the drop-down

- a nodeId is required for archive logs, so this fixes archive logs when no node was chosen
  - which is a bug / regression that seems to stem from the `POD_NAMES=v2` change

### Modifications

- use the `podNamesToNodeIDs` map from #9464 to derive a `nodeId`

### Verification

Manually tested that clicking the top-level logs button and then choosing a pod name gets logs for both archive and non-archive. Screenshot below, note that the URL has no `nodeId` in it as a node was not clicked

Screenshot 2024-04-13 at 2 03 05 PM

PR #12933, opened 2024-04-13
refactor(ui): simplify `hasArtifactLogs` with optional chaining

agilgur5:
Similar to other PRs I've had that simplify code or fix bugs by using optional chaining

### Motivation

- remove 4 LoC with optional chaining short-hand

### Modifications

- replace conditionals with optional chaining
- `array.findIndex(...) !== -1` is also equivalent to `array.some(...)`

### Verification

These are semantically equivalent, but I also had this in my local codebase while writing & debugging #12932 et al (which is how I stumbled upon this) and it worked the same


tczhao (approved):
👍 `some` semantics is clearly more readable than `findIndex`

isubasinghe (approved):
LGTM

PR #12934, opened 2024-04-13
refactor(ui): simplify `getContainerLogs` by removing `switchMap`

agilgur5:
### Motivation

- this can just use plain conditionals instead of RxJS specific logic
  - cleaner, easier to understand, less LoC, etc

Aligns with generally simplifying the codebase and removing RxJS specific logic, that isn't necessary in many cases, especially with modern JS

### Modifications

- remove the RxJS specific logic in `getContainerLogs` and replace it with plain conditionals

### Verification

Semantically equivalent; in particular, type-check continues to pass

### DIFF
--- a/ui/src/app/shared/services/workflows-service.ts
+++ b/ui/src/app/shared/services/workflows-service.ts
@@ -261,14 +261,10 @@ export const WorkflowsService = {
         }
 
         // return archived log if main container is finished and has artifact
-        return from(this.isWorkflowNodePendingOrRunning(workflow, nodeId)).pipe(
-            switchMap(isPendingOrRunning => {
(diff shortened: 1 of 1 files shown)


ui/src/app/shared/services/workflows-service.ts
```diff
@@ -261,14 +261,10 @@ export const WorkflowsService = {
         }
 
         // return archived log if main container is finished and has artifact
-        return from(this.isWorkflowNodePendingOrRunning(workflow, nodeId)).pipe(
```
agilgur5, line 264: oh nevermind, this is actually an async function that can also error, so some of this logic is actually different.

I'm not sure that `isWorkflowNodePendingOrRunning` actually needs a `get` request in the first place though, so maybe will revisit both together to simplify

PR #12935, opened 2024-04-15
fix: correct order in artifactGC error log message

static-moonlight:
fix an error log message, which currently uses a wrong order of parameter, as mention in #12845 (point 5).


agilgur5:
> as mention in #12845 (point 5).

yea thanks for noticing and pointing that out!
for reference, I found the specific line number in the code in https://github.com/argoproj/argo-workflows/discussions/12845#discussioncomment-8988692

agilgur5 (approved):
Thanks for fixing the error log!

agilgur5:
Backported cleanly to `release-3.5` as 909fdaa

PR #12939, opened 2024-04-15
fix(ui): default to `main` container name in event source logs API call

aniro-s:
fixes argoproj#12938


agilgur5 (approved):
LGTM given that this is identical to #9438

agilgur5:
Backported cleanly into `release-3.5` as fe5c6128c

PR #12942, opened 2024-04-16
fix(build): close `pkg/apiclient/_.secondary.swagger.json`

testwill:
Fixes #TODO


Joibel (approved):

Joibel:
LGTM, thanks for this.

agilgur5 (approved):
Makes sense to me and is general good practice to `close` file descriptors. Did you experience some error with codgen around this? Since it isn't used after and should be GC'd shortly after as well

Looks like [`ReadFile`](https://pkg.go.dev/os#ReadFile) and `WriteFile` don't return file descriptors, so it would only be valid for `Create`

Issue #12868, opened 2024-04-02
Argo Workflows main workflows list page does not show workflow retry is running


amfage:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

If a workflow fails, when viewing the workflow and clicking "retry", when returning to the main workflows listing page, the workflow still shows failed, although the workflow is running. Sometimes the workflow listing page will, after a second or maybe 5 seconds, show the correct status, but sometimes it stays displaying the failed status.
![retry-workflow-argo-3-5-5](https://github.com/argoproj/argo-workflows/assets/21299036/fd58cafc-482f-49c0-a419-27bec396650a)
The expected behaviour is that the icon shows the blue circle running icon, and the started, finished, duration and message areas show that the workflow is running, as it would if the workflow was running for the first time. This is how it behaved in the previous version we were running which was 3.4.11.

### Version

v3.5.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
# This example sleeps for 30 seconds then fails, so it can be retried
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: test-simple-fail-for-retry-
spec:
  entrypoint: template-fail
  templates:
  - name: template-fail
    script:
      image: python:alpine3.6
      command: ["python"]
      # fail
      source: |
        import sys;
        import time;
        time.sleep(30); 
        sys.exit(1)
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep test-fail-for-retry-ml94n
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

agilgur5:
> after a second or maybe 5 seconds, show the correct status

That sounds correct. The Workflows List page runs a `ListWatch` on the visible k8s resources, so there can be a time delay between when the retry occurs and when it shows up in the `ListWatch`.

[...]

PR #12950, opened 2024-04-18
fix(ui): use router navigation instead of page load after submit

agilgur5:
I noticed this while investigating https://github.com/argoproj/argo-workflows/issues/12868#issuecomment-2033385525, although I don't think this would fix that issue.
Also related to / similar to #12930 

### Motivation

- `document.location.href` causes the browser to load a new page which is a full page load
  - `navigation.goto` should always be used when we're routing within the single-page app (SPA)
    - this only changes the internal route so only the next component needs to render, not the entire page
      - (and user-facing `history`, same as changing the `location`)

- fix `name` and `namespace` in Workflow Details to actually change when the URL changes
  - they were previously set as `state` despite not actually being `state`, meaning they only ever received the initial URL and no further changes
    - in particular, this is required for a resubmit to work, as it re-routes to the same component (Workflow Details), but with a different URL

### Modifications

- use `navigation.goto` instead of assigning `document.location.href` in the submit, resubmit, and retry panels
- remove `state` for `name` and `namespace` in Workflow Details and instead always get them from the URL
- some other tiny code style optimizations in surrounding code

### Verification

I made a few screencaptures to verify this:
1. <details><summary>Submit no longer does a full page reload</summary>

    https://github.com/argoproj/argo-workflows/assets/4970083/d3888986-c66d-4f55-a54d-579e0dd91d38

</details>

1. <details><summary>Resubmit no longer does a full page reload</summary>

    https://github.com/argoproj/argo-workflows/assets/4970083/c6059d05-3b60-48a0-8abc-2d7006d134ef

</details>

1. <details><summary>Retry no longer does a full page reload</summary>

    https://github.com/argoproj/argo-workflows/assets/4970083/c2aac826-81d2-4856-9302-792a44a16d24

    (I used a different software to record this initially, that's why it's a bit different from the other two)

</details>

### DIFF
--- a/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
+++ b/ui/src/app/workflows/components/resubmit-workflow-panel.tsx
@@ -1,6 +1,7 @@
 import {Checkbox} from 'argo-ui';
-import React, {useState} from 'react';
+import React, {useContext, useState} from 'react';
 import {Parameter, ResubmitOpts, Workflow} from '../../../models';
+import {Context} from '../../shared/context';
 import {uiUrl} from '../../shared/base';
 import {ErrorNotice} from '../../shared/components/error-notice';
 import {ParametersInput} from '../../shared/components/parameters-input';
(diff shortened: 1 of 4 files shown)


toyamagu-2021 (approved):
Thanks! LGTM

isubasinghe (approved):

Issue #12949, opened 2024-04-17
EKS Pod Identity S3 Artifact is not working on 3.5.5


rafilkmp3:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

I have the following configmap:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: workflows-artifact-repository
  namespace: workflows
data:
  v2-s3-artifact-repository: |
    s3:
      bucket: redacted-prod-artifacts
      endpoint: s3.amazonaws.com
      region: us-east-2
      useSDKCreds: true
```
The service account is already set up correctly and the minio client is assuming the correct role with more than enough permissions to S3. Also, the role already have the Trust Relationship as:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Service": "pods.eks.amazonaws.com"
            },
            "Action": [
                "sts:TagSession",
                "sts:AssumeRole"
            ]
        }
    ]
}
```



### Version

v3.5.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
spec:
  workflowSpec:
    templates:
      - name: main
        inputs: {}
        outputs: {}
        nodeSelector:
          kubernetes.io/arch: amd64
        metadata: {}
        steps:
          - - name: generate-data
              template: generate-data
              arguments:
                parameters:
                  - name: lookback_hours
                    value: '1'
```
and this is the template used:
```yaml
      - name: generate-data
        inputs:
          parameters:
            - name: lookback_hours
              value: '1'
        outputs:
          artifacts:
            - name: interactions-csv
              path: /app/interactions.csv
            - name: items-csv
              path: /app/items.csv
            - name: users-csv
              path: /app/users.csv
        nodeSelector:
          kubernetes.io/arch: amd64
        metadata: {}
        script:
          name: ''
          image: >-
            redacted.dkr.ecr.us-east-2.amazonaws.com/personalize-updater:latest
          command:
            - bash
          resources:
            limits:
              memory: 12Gi
            requests:
              cpu: '2'
[...]

Issue #12946, opened 2024-04-16
Offloading large workflows raises db syntax errors that prevent the offloading


Yarin-Shitrit:
### Pre-requisites

- [X] I have double-checked my configuration
- [ ] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

I have tried to enable `nodeStatusOffload` for the workflow-controller, and configured the postgresql under the persistence section in the workflow-controller-configuration.
The connection is successful, as there're migration logs in the controller.
But when it reaches the offloading part, it crashes with this error:
```
workflow is longer than maximum allowed size. compressed size 1053022 > maxSize 1048576. Tried to offload but encountered an error: pq: syntax error at or near '(' 
```
I also tried doing this with mysql instead of postgresql but got another syntax error: 
```
workflow is longer than maximum allowed size. compressed size 1053022 > maxSize 1048576. Tried to offload but encountered an error: ERROR 1064: You have an error in your SQL syntax; check the manual that corresponds to your MYSQL server version for the right syntax to use near '('clustername', 'namespace', 'uid', 'nodes', 'version') VALUES (?' at line 2.
```
I am trying to make my Argo Server run very large workflows, and to do so I generated a sample DAG with demo steps that simply print, and tried to run 700 of those without any dependency between the tasks.
 

### Version

v3.4.6

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
I simply ran a workflow that make some prints, not anything complex.
And loaded my DAG with 700 steps like this.
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

agilgur5:
> - [x] I can confirm the issue exists when I tested with `:latest`

> v3.4.6

You're on a pretty old version, so I would suggest trying with `:latest` or 3.4.16 at least.
Please fill out the issue template accurately, it asks for `:latest` usage very intentionally.

agilgur5:
[...]

Issue #12802, opened 2024-03-14
Etcd get full with ~500 workflows.


leryn1122:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

We run ~500 workflows and ~500 pods concurrently as offline tasks in prod env. Etcd got full rapidly at the size of 8G.
It resulted in that etcd and apiserver turned into unavailable and the argo workflow controller auto restarted frequently.
Our team concluded that etcd and apiserver may be unavailable if running and pending workflows flood into etcd according to monitoring and metrics.

For now, the team’s solutions are:
- Limiting workflows quotas
- Optimizing the size of workflow template rendored by biz
- Writing scripts to check and compress etcd if full as a schedule task
- Migrating biz argo into another cluster alone

It is expected that argo workflows do not flood into etcd or impact the stability of whole cluster.

### Version

v3.4.10

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
Limited by NDA.
```

### Logs from the workflow controller

```text
time="2024-03-06T01:59:48.872Z" level=info msg="Mark node slice-587-20231201141424-20240128125944-20240201143506-fen7hx62[0].xxxxxxx[1].subtask(20:raw/587/2023/12/6/1734100703048318977/ros2/20231206104000_20231206104046_5m/raw_587_20231206104000_20231206104046.db3)[2].xxxxxxx[0].xxxxxxx[0].xxxxxxx(0) as Pending, due to: admission webhook \"resourcesquotas.quota.kubesphere.io\" denied the request: pods \"slice-587-20231201141424-20240128125944-20240201143506-fen7hx62-3557973944\" is forbidden: exceeded quota: argo, requested: count/pods=1, used: count/pods=822, limited: count/pods=600" namespace=argo workflow=slice-587-20231201141424-20240128125944-20240201143506-fen7hx62
[...]

PR #12951, opened 2024-04-18
chore: clarify `:latest` in bug templates

agilgur5:
### Motivation

- there have been many cases of users checking the box despite not testing `:latest`, sometimes not even testing the latest patch version
  - this is an effort to clarify that / make it more explicit what that means to hopefully reduce that occurrence
  - some recent examples of this occurrence include: https://github.com/argoproj/argo-workflows/issues/12946#issuecomment-2060000421, https://github.com/argoproj/argo-workflows/issues/12802#issuecomment-2007756351, https://github.com/argoproj/argo-workflows/issues/12949#issuecomment-2062322286
  
@Joibel and I have also briefly talked about this during Contributor Meetings

### Modifications

- Clarify "tested with `:latest`" by adding "image tag" and giving an example

### Verification

the strings are valid, but otherwise n/a, this is GH specific. there is also no local preview for issue forms unfortunately: https://github.com/orgs/community/discussions/7039, https://github.com/orgs/community/discussions/4402, https://github.com/orgs/community/discussions/63402#discussioncomment-6697333, https://github.com/orgs/community/discussions/4265, etc

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -9,7 +9,7 @@ body:
       options:
         - label: I have double-checked my configuration
           required: true
-        - label: I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`
+        - label: I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
           required: true
         - label: I have searched existing issues and could not find a match for this bug
           required: true
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -9,7 +9,7 @@ body:
       options:
         - label: I have double-checked my configuration
           required: true
-        - label: I can confirm the issue exists when I tested with `:latest`
+        - label: I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`
           required: true
         - label: I have searched existing issues and could not find a match for this bug
           required: true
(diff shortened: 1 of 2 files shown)


agilgur5:
> I'll approve in the interests of "It's definitely an incremental improvement"

Yea that was my intent and I had a similar thought process (although not entirely the same, see below). I had been thinking of removing the `required` attribute of the checkbox as it's not quite accurate, but that almost certainly would get abused.
Wording it in a way to be accurate and have the intended implications & consequences is a bit challenging -- in the interim I figure we can improve the existing wording and intent at least, and when we have something better, we can use that.

> The burden of testing on :latest has always seemed like slightly too high a bar for some kinds of issues. In theory it implies (but doesn't mean) that we're not interested in bug reports against 3.4.16 (latest minor release at the time of writing)

I don't think it's a particularly high bar, but yes it's not quite correct. Especially for regressions, where it working in `:latest` may not actually matter. It can help with debugging and diagnosing though.

In general, for, say, more seasoned reporters, I would expect that they would still file an issue and explain their rationale. Or, in other words, this could help deter lower quality issues but not deter higher quality ones, or perhaps even push those even higher.

There is a chance it deters some who rigorously follow rules, but those tend to be the same people who would file an issue nonetheless, recognizing the rule is a bit off and that their issue is definitely a bug. At least such has been my experience in OSS and career (and uh I have some similar characteristics myself 😅).

> However, that's quite a lot to write in one line in a bug template!
> 
> I wonder if we just need a linked-to page describing our expectations for end user testing before filing a bug.

Yes this sentence is pretty long as-is. Unfortunately such a page would rarely get read, so I'm not sure it's worth the time to draft up, especially given how nuanced it would have to be (and therefore the iteration and effort required while not giving much benefit in return). It would only get read by high quality reporters in my experience, and I don't want to add more time to them than necessary either.

There is perhaps a compromise that is "good enough" for most demographics -- we just another sentence along the lines of "If not, I have explained why in my description". That would be more accurate and so mitigates misunderstandings by higher quality reporters, but could also be abused by lower quality reporters as "I'm not on `:latest`". Perhaps re-written as "[...] I have explained, **in detail**, [...]" could mitigate that.

agilgur5:
> we just add another sentence along the lines of "If not, I have explained why in my description".

> Perhaps re-written as "[...] I have explained why, **in detail**, [...]" could mitigate that.

I added this to the PR

Joibel (approved):
Seems like a good balance, thanks @agilgur5

PR #12958, opened 2024-04-21
chore: use `errors.New` to replace `fmt.Errorf` with no parameters

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


agilgur5:
> will much better

You seem to have not completed your sentence, why is it much better?

PR #6982, opened 2021-10-19
fix: Add pod name format annotation. Fixes #6962 and #6989

JPZ13:
Signed-off-by: J.P. Zivalich <jp@pipekit.io>

This PR:
- Adds the pod name version as an annotation to a workflow
- Displays ui logs for a given workflow based on the annotation
- Adds a function to get a template name from a template ref

Fixes #6962 and #6989

PR #12928, opened 2024-04-11
fix: correct pod names for inline templates. Fixes #12895

AlbeeSo:
Fixes #12895

### Motivation

https://github.com/argoproj/argo-workflows/blob/87a2041ba7507c15070a3855b924c1075c50c268/workflow/util/pod_name.go#L60
```
inline-cant-show-errors-9ptb9--3016472844   0/2     Error       0               25h
inline-cant-show-errors-9ptb9--3942528235   0/2     Completed   0               25h
```
Incorrect pod names of inline templates will lead to empty UI logs problem.

### Verification

Test with the workflow mentioned in issue, got result like image post below:
image

Issue #12895, opened 2024-04-05
Inline Workflows logs in UI always empty


artem-zherdiev-ingio:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

Inline Workflows always show an empty log in UI, it doesn't matter if the task succeeded or not. We are still able to read logs from the pod in the console.

Expectation: inline workflows should show correct logs in UI.

### Version

v3.5.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: inline-cant-show-errors
  namespace: workflows
spec:
  serviceAccountName: workflows-executor

  podGC:
    strategy: OnWorkflowSuccess

  entrypoint: pipeline
  templates:
    - name: pipeline
      dag:
        tasks:
          - name: broken-configmap
            inline:
              serviceAccountName: configmap-owner
              resource:
                action: get
                manifest: |
                  apiVersion: v1
                  kind: ConfigMap
                  metadata:
                    name: non-existing-configmap
                    namespace: workflows
              outputs:
                parameters:
                  - name: value
                    valueFrom:
                      default: "default-value"
                      jqFilter: '.data["value"]'
          - name: correct-shell
            inline:
              container:
                image: alpine:3.7
                command: [sh, -c]
                args: ["echo hello"]
          - name: broken-shell
            inline:
              container:
                image: alpine:3.7
                command: [sh, -c]
                args: ["cat /tmp/non-existing-file"]
```

### Logs from the workflow controller

```text
time="2024-04-05T12:04:45.437Z" level=info msg="Processing workflow" Phase= ResourceVersion=285687282 namespace=workflows workflow=inline-cant-show-errors-ppxjk
time="2024-04-05T12:04:45.459Z" level=info msg="Task-result reconciliation" namespace=workflows numObjs=0 workflow=inline-cant-show-errors-ppxjk
time="2024-04-05T12:04:45.459Z" level=info msg="Updated phase  -> Running" namespace=workflows workflow=inline-cant-show-errors-ppxjk
[...]

PR #12964, opened 2024-04-22
refactor: simplify `getPodName` and make consistent with back-end

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/6982#discussion_r1574359684, Fixes https://github.com/argoproj/argo-workflows/pull/9168#discussion_r1574264448

### Motivation

Simplify `getPodName` as it required lots of args that could be retrieved from the existing Workflow and Node. Simplify surrounding code as well

Make it more consistent to the back-end as well, particularly after #12928.
  - It seems to have diverged unintentionally within the same PR per https://github.com/argoproj/argo-workflows/pull/6982#discussion_r1574359684 

### Modifications

- simplify calls to `getPodName` to only require the Workflow and Node as args
   - the rest can be derived from them, so we can remove all the duplicate, extraneous logic from all callers
     - `workflow-details.tsx` was particularly simplified as a result, a whole function was replaced with a one-liner
   - also make `getPodName` in the UI more closely match `GeneratePodName` on the back-end

- adjust `pod-name.test.ts` to match the modified args
  - also heavily simplify one of the test suites to cast some types instead of creating an unnecessary structure

- <details><summary>simplify other parts of <code>pod-name.ts</code></summary>
  
    - simplify `getTemplateNameFromNode` to a one-liner that's easier to read and reads like the comment too
      - JS has [truthiness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), so `if (node.templateName)` checks for `undefined`, `''`, and others
      - `node?.` is not necessary per https://github.com/argoproj/argo-workflows/pull/9168#discussion_r1574264448, and we can invert the conditional to `if (node.templateRef?.template)`
      - put it all together and you get a one-liner
      
    - slightly optimize `ensurePodNamePrefixLength` by moving the static calculation of `maxPrefixLength` into the module scope and outside the function scope
      - it does not change per function invocation
      - do the same on the back-end too in `pod_name.go`
      
    - use named functions for better tracing compared to `const` variables assigned to anonymous functions, same as #12062 et al
    
</details>

### Verification

Existing tests pass

Also tested manually with the Workflow from #12895. <details><summary> Screenshot: </summary>

Screenshot 2024-04-22 at 12 17 24 PM

</details>

### Notes to Reviewers

It's easier to read each commit independently than it is to read the PR as a whole, that's why I split them that way

### Future Work

1. `pod-name.ts` should only be `export`ing `getPodName`. the rest is not used anywhere except for the tests
    - The tests need a good bit of adjustment for this though, so for now, I left them as-is to show they still pass
1. The back-end's `GeneratePodName` can similarly be simplified, but it requires a lot more refactoring of functions of up the chain
    - partial workflow information, partial node information, etc gets passed down through a few layers of the call stack

### DIFF
--- a/ui/src/app/shared/pod-name.test.ts
+++ b/ui/src/app/shared/pod-name.test.ts
@@ -1,4 +1,6 @@
-import {Inputs, MemoizationStatus, NodePhase, NodeStatus, NodeType, Outputs, RetryStrategy} from '../../models';
+import {NodeStatus, Workflow} from '../../models';
+import {ANNOTATION_KEY_POD_NAME_VERSION} from './annotations';
+
 import {createFNVHash, ensurePodNamePrefixLength, getPodName, getTemplateNameFromNode, k8sNamingHashLength, maxK8sResourceNameLength, POD_NAME_V1, POD_NAME_V2} from './pod-name';
 
 describe('pod names', () => {
@@ -9,9 +11,6 @@ describe('pod names', () => {
     });
 
     // note: the below is intended to be equivalent to the server-side Go code in workflow/util/pod_name_test.go
-    const nodeName = 'nodename';
-    const nodeID = '1';
-
     const shortWfName = 'wfname';
     const shortTemplateName = 'templatename';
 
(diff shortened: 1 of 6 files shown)


tczhao (approved):
> It's easier to read each commit independently than it is to read the PR as a whole, that's why I split them that way

This is gold

isubasinghe (approved):

agilgur5:
Backported this refactor to `release-3.5` in order to reduce (significant) merge conflicts with #13039. This itself had a few merge conflicts, but they were all style changes that didn't get backported (e.g. `const funcName = () =>` -> `function funcName()`, newlines in import order, and file renames in imports)

PR #6356, opened 2021-07-16
fix(controller): Mark workflows wait for semaphore as pending. Fixes #6351

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>
Fixes #6351

PR #12965, opened 2024-04-22
fix(controller): set pod name version annotation when no lock

### LINKED ISSUE #10178: No logs accessible in the UI on workflows that went from Pending to Running
### Pre-requisites

Argo Workflows controller is configured with `parallelism: 10`. 

Workflows that are created in pending, because too many workflows are already running, will not have the annotation `workflows.argoproj.io/pod-name-format: v2` associated to them, therefore logs will not be able to be viewed using the UI.

It can be reproduced using the attached workflow and the command:
```
for i in {1..20}; do kubectl create -f test.yaml; done
```

It seems to happen because the annotation is only set here:
https://github.com/argoproj/argo-workflows/blob/2eb871bf2f1bd9008ab09995b560600e5c594153/workflow/controller/operator.go#L271

but not here: https://github.com/argoproj/argo-workflows/blob/2eb871bf2f1bd9008ab09995b560600e5c594153/workflow/controller/operator.go#L251

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: test-logs-
spec:
  entrypoint: test-dag
  templates:
  - name: echo
    container:
      image: alpine:3.7
      command: [echo, "OK!"]
  - dag:
      tasks:
      - name: test
        template: test
    name: test-dag
  - dag:
      tasks:
      - name: 'echo-1'
        template: echo
      - depends: echo-1
        name: 'echo-2'
        template: echo
    name: test
```

```text
time="2022-12-05T09:14:17.554Z" level=info msg="Workflow processing has been postponed due to max parallelism limit" key=cloud-workflows/test-logs-54tz6
time="2022-12-05T09:14:17.554Z" level=info msg="Updated phase  -> Pending" namespace=cloud-workflows workflow=test-logs-54tz6
time="2022-12-05T09:14:17.554Z" level=info msg="Updated message  -> Workflow processing has been postponed because too many workflows are already running" namespace=cloud-workflows workflow=test-log>
time="2022-12-05T09:14:17.555Z" level=info msg="Workflow to be dehydrated" Workflow Size=652
time="2022-12-05T09:14:17.653Z" level=info msg="Update workflows 200"
time="2022-12-05T09:14:17.654Z" level=info msg="Workflow update successful" namespace=cloud-workflows phase=Pending resourceVersion=276945745 workflow=test-logs-54tz6
...
time="2022-12-05T09:14:58.332Z" level=info msg="Processing workflow" namespace=cloud-workflows workflow=test-logs-54tz6
time="2022-12-05T09:14:58.340Z" level=info msg="Task-result reconciliation" namespace=cloud-workflows numObjs=0 workflow=test-logs-54tz6
time="2022-12-05T09:14:58.340Z" level=info msg="Updated phase Pending -> Running" namespace=cloud-workflows workflow=test-logs-54tz6
```

```text
N/A
```

agilgur5:
Fixes #10178, Fixes https://github.com/argoproj/argo-workflows/pull/6982#discussion_r1574356857, Fixes https://github.com/argoproj/argo-workflows/pull/10735#discussion_r1575092505
Partly addresses https://github.com/argoproj/argo-workflows/pull/6356#discussion_r1574367137, but not entirely, as the PDB is a more complicated case. Will file a separate issue for that. **EDIT**: filed #12966 

### Motivation

- `setWfPodNamesAnnotation` was added to the `Unknown` phase check a few lines below this, but not to this line
  - these are the two places where `Unknown` is handled
  - without this, the annotation wouldn't be set if the workflow couldn't grab a lock, creating a mismatch when using synchronization

### Modifications

- add `setWfPodNamesAnnotation` to the other `Unknown` phase block
  - also simplify the conditional into one with an `&&` rather than nesting
    - reducing nesting reduces code complexity

- slightly simplify `setWfPodNamesAnnotation` as well

### Verification

Tested with the same Workflow from #12966, creating two in quick succession (one waits on the mutex lock from the other), `workflows.argoproj.io/pod-name-format: v2` annotations present in both.

Should probably add a case to `operator_concurrency_test.go`, TBD

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -259,16 +259,15 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 			return
 		}
 		woc.updated = wfUpdate
-		if !acquired {
-			if !woc.releaseLocksForPendingShuttingdownWfs(ctx) {
-				woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
-				phase := woc.wf.Status.Phase
-				if phase == wfv1.WorkflowUnknown {
-					phase = wfv1.WorkflowPending
-				}
-				woc.markWorkflowPhase(ctx, phase, msg)
-				return
+		if !acquired && !woc.releaseLocksForPendingShuttingdownWfs(ctx) {
+			woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
+			phase := woc.wf.Status.Phase
+			if phase == wfv1.WorkflowUnknown {
+				phase = wfv1.WorkflowPending
+				setWfPodNamesAnnotation(woc.wf)
 			}
+			woc.markWorkflowPhase(ctx, phase, msg)
+			return
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


tczhao (commented):
Looks good, but we better have the test case added before merge right?

workflow/controller/operator.go
```diff
@@ -259,16 +259,15 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
...
-				}
-				woc.markWorkflowPhase(ctx, phase, msg)
-				return
-			}
+		if !acquired && !woc.releaseLocksForPendingShuttingdownWfs(ctx) {
+			woc.log.Warn("Workflow processing has been postponed due to concurrency limit")
+			phase := woc.wf.Status.Phase
+			if phase == wfv1.WorkflowUnknown {
+				phase = wfv1.WorkflowPending
+				setWfPodNamesAnnotation(woc.wf)
```
juliev0, line 267: hate to be picky but wondering if we could just set this in one place at the top of the function? like maybe logic like "if annotation isn't set, then set it". Is there any reason it should ever not be set? That would handle any future cases where we preemptively exit for whatever reason.

Issue #12967, opened 2024-04-22
Incorrect Pod Name in Workflow Failures JSON When Using `templateRef`


ryancurrah:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### Slack thread

Issue was originally reported on Slack: https://cloud-native.slack.com/archives/C01QW9QSSSK/p1710875939718929

### What happened/what did you expect to happen?

There appears to be a bug in the `workflow.failures` JSON, specifically in the naming convention used for the pod name. The expected pod name format should include the task name between two hyphens (--) as observed in the UI and via Kubectl commands. However, the JSON output does not reflect this format, leading to potential issues when programmatically using this pod name to perform an action that requires it, like generating a link to the failed task logs.

In our Workflow here is the failure message with the incorrect `podName`.

```json
[
  {
    "displayName": "pr-go-test-cdx-ryaniscool-1179-rck45",
    "message": "",
    "templateName": "main",
    "phase": "Failed",
    "podName": "pr-go-test-cdx-ryaniscool-1179-rck45",
    "finishedAt": "2024-03-19T19:10:00Z"
  },
  {
    "displayName": "setup",
    "message": "Error (exit code 1)",
    "templateName": "",
    "phase": "Failed",
    "podName": "pr-go-test-cdx-ryaniscool-1179-rck45--1909009551",
    "finishedAt": "2024-03-19T19:09:54Z"
  }
]
```

**Details:**
- **Workflow Name:** pr-go-test-cdx-ryaniscool-1179-rck45
- **Expected Pod Name:** pr-go-test-cdx-ryaniscool-1179-rck45-setup-1909009551
- **Actual Pod Name in JSON:** pr-go-test-cdx-ryaniscool-1179-rck45--1909009551

This discrepancy was identified while attempting to provide developers with direct links to the failed workflow step in the UI. The incorrect pod name format prevents these links from functioning correctly.

**Relevant Code:**
The issue seems to originate from the pod name generation logic in the operator.go file:
```go
[...]

PR #12970, opened 2024-04-23
fix: use GetTemplateFromNode to determine template name

ryancurrah:
The way to get a template name in a `NodeStatus` is different if using template references. Not all places in the code base handle this small detail. I searched for all reads of `NodeStatus.TemplateName` field and replaced it to use the `GetTemplateFromNode` function. This will ensure that the template name is never empty and retrieved uniformly. 

There are three places in the code base that still manually handle these details: 
- https://github.com/argoproj/argo-workflows/blob/main/cmd/argo/commands/common/get.go#L488-L492
- https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/retry_tweak.go#L18-L33. 
- https://github.com/argoproj/argo-workflows/pull/12970#discussion_r1576605599

If you think we should change these to also use `GetTemplateFromNode` let me know.

Fixes https://github.com/argoproj/argo-workflows/issues/12967

### Motivation

I wanted to create a link to the failed task logs pane in the UI, it requires the pod name to be correct in the workflow failures json, but in the failures json the pod name was incorrect because the template name field was empty.

### Modifications

Changed almost all references of `NodeStatus.TemplateName` to use `GetTemplateFromNode(NodeStatus.TemplateName)`.

### Verification

Ran unit tests.

### DIFF
--- a/util/printer/workflow-printer.go
+++ b/util/printer/workflow-printer.go
@@ -137,7 +137,7 @@ func countPendingRunningCompletedNodes(wf *wfv1.Workflow) (int, int, int) {
 	running := 0
 	completed := 0
 	for _, node := range wf.Status.Nodes {
-		tmpl := wf.GetTemplateByName(node.TemplateName)
+		tmpl := wf.GetTemplateByName(util.GetTemplateFromNode(node))
 		if tmpl == nil || !tmpl.IsPodType() {
 			continue
 		}
--- a/workflow/controller/artifact_gc.go
+++ b/workflow/controller/artifact_gc.go
@@ -179,10 +179,7 @@ func (woc *wfOperationCtx) processArtifactGCStrategy(ctx context.Context, strate
 			woc.log.Errorf("Was unable to obtain node for %s", artifactSearchResult.NodeID)
 			return fmt.Errorf("can't process Artifact GC Strategy %s: node ID %q not found in Status??", strategy, artifactSearchResult.NodeID)
 		}
-		templateName := node.TemplateName
-		if templateName == "" && node.GetTemplateRef() != nil {
-			templateName = node.GetTemplateRef().Template
-		}
+		templateName := util.GetTemplateFromNode(*node)
 		if templateName == "" {
 			return fmt.Errorf("can't process Artifact GC Strategy %s: node %+v has an unnamed template", strategy, node)
 		}
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2437,7 +2437,7 @@ func (woc *wfOperationCtx) GetNodeTemplate(node *wfv1.NodeStatus) (*wfv1.Templat
 		}
 		return tmpl, nil
 	}
-	return woc.wf.GetTemplateByName(wfutil.GetTemplateFromNode(*node)), nil
+	return woc.wf.GetTemplateByName(node.TemplateName), nil
 }
 
 func (woc *wfOperationCtx) markWorkflowRunning(ctx context.Context) {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -405,9 +405,9 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 				failedNodeStatus{
 					DisplayName:  node.DisplayName,
 					Message:      node.Message,
-					TemplateName: node.TemplateName,
+					TemplateName: wfutil.GetTemplateFromNode(node),
 					Phase:        string(node.Phase),
-					PodName:      wfutil.GeneratePodName(woc.wf.Name, node.Name, node.TemplateName, node.ID, wfutil.GetPodNameVersion()),
+					PodName:      wfutil.GeneratePodName(woc.wf.Name, node.Name, wfutil.GetTemplateFromNode(node), node.ID, wfutil.GetPodNameVersion()),
 					FinishedAt:   node.FinishedAt,
 				})
 		}
@@ -1338,7 +1338,7 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 			new.Phase = wfv1.NodeSucceeded
 		} else {
 			new.Phase, new.Message = woc.inferFailedReason(pod, tmpl)
-			woc.log.WithField("displayName", old.DisplayName).WithField("templateName", old.TemplateName).
+			woc.log.WithField("displayName", old.DisplayName).WithField("templateName", wfutil.GetTemplateFromNode(*old)).
 				WithField("pod", pod.Name).Infof("Pod failed: %s", new.Message)
 		}
 		new.Daemoned = nil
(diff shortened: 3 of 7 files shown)


workflow/controller/operator.go
```diff
@@ -2437,7 +2437,7 @@ func (woc *wfOperationCtx) GetNodeTemplate(node *wfv1.NodeStatus) (*wfv1.Templat
 		}
 		return tmpl, nil
 	}
-	return woc.wf.GetTemplateByName(node.TemplateName), nil
+	return woc.wf.GetTemplateByName(wfutil.GetTemplateFromNode(*node)), nil
```
agilgur5, line 2440: the lines above this check for `node.TemplateRef`, so I think you might be able to simplify them. or the `GetTemplateFromNode` is redundant here

workflow/controller/operator.go
```diff
@@ -2628,7 +2628,7 @@ func (woc *wfOperationCtx) getPodByNode(node *wfv1.NodeStatus) (*apiv1.Pod, erro
 		return nil, fmt.Errorf("Expected node type %s, got %s", wfv1.NodeTypePod, node.Type)
 	}
 
-	podName := woc.getPodName(node.Name, node.TemplateName)
+	podName := woc.getPodName(node.Name, wfutil.GetTemplateFromNode(*node))
```
agilgur5, line 2631: we should eventually refactor this to just accept the `node` as an argument. it requires refactoring elsewhere though, similar to #12964 "Future Work"

workflow/controller/artifact_gc.go
```diff
@@ -179,10 +179,7 @@ func (woc *wfOperationCtx) processArtifactGCStrategy(ctx context.Context, strate
 			woc.log.Errorf("Was unable to obtain node for %s", artifactSearchResult.NodeID)
 			return fmt.Errorf("can't process Artifact GC Strategy %s: node ID %q not found in Status??", strategy, artifactSearchResult.NodeID)
 		}
-		templateName := node.TemplateName
-		if templateName == "" && node.GetTemplateRef() != nil {
-			templateName = node.GetTemplateRef().Template
-		}
+		templateName := util.GetTemplateFromNode(*node)
```
agilgur5, line 182: I believe `func (n *NodeStatus) GetTemplateRef() *TemplateRef {` in `workflow_types.go` is no longer needed and can be removed now

agilgur5: oh maybe not... `TemplateReferenceHolder` appears to be a supertype... although it's not explicitly used that way... it looks like it's being used as an implicit structural type... that's confusing... should probably be refactored to be a properly inherited nominal type

agilgur5: and `GetTemplateName` is technically correct since it's supposed to return the raw `TemplateName`... yea that type is really confusing... there's almost certainly some years old tech debt with the structural types 😕

util/printer/workflow-printer.go
```diff
@@ -137,7 +137,7 @@ func countPendingRunningCompletedNodes(wf *wfv1.Workflow) (int, int, int) {
 	running := 0
 	completed := 0
 	for _, node := range wf.Status.Nodes {
-		tmpl := wf.GetTemplateByName(node.TemplateName)
+		tmpl := wf.GetTemplateByName(util.GetTemplateFromNode(node))
 		if tmpl == nil || !tmpl.IsPodType() {
 			continue
 		}
```
agilgur5, line 143: I don't know if the math would work out entirely the same, but this could potentially be simplified to just check `node.Type` instead of getting the template only to get its type

agilgur5 (commented):
I think there's two places that can be further refactored here. Otherwise this looks like it probably fixes some other subtle bugs with `templateRef`s beyond yours as well

util/printer/workflow-printer.go
ryancurrah: Hrm which types should we be checking for?

```
NodeTypePod       NodeType = "Pod"
NodeTypeContainer NodeType = "Container"
NodeTypeSteps     NodeType = "Steps"
NodeTypeStepGroup NodeType = "StepGroup"
NodeTypeDAG       NodeType = "DAG"
NodeTypeTaskGroup NodeType = "TaskGroup"
NodeTypeRetry     NodeType = "Retry"
NodeTypeSkipped   NodeType = "Skipped"
NodeTypeSuspend   NodeType = "Suspend"
NodeTypeHTTP      NodeType = "HTTP"
NodeTypePlugin    NodeType = "Plugin"
```

workflow/controller/operator.go
ryancurrah: I opted to undo this change.

util/printer/workflow-printer.go
agilgur5: Correct. If the number is the same, then it shouldn't matter. I'm not sure if it's entirely semantically equivalent though

ryancurrah: Neither am I and there is no test coverage for this code. I'll try running this change locally with debugging and verify. It might take some time

agilgur5: Oh I didn't realize this didn't have a unit test. It does seem to have coverage via other CLI commands, just not its _own_ unit test.

In that case, that simplification can be done as a separate PR (if possible, or just tests if not would be great too!). The change you made here is still correct.

agilgur5 (approved):
LGTM. Thanks for reporting and fixing this!

We have some tech debt follow-ups too which would be good to resolve eventually per above comments (since uh, we do have a good amount of tech debt and that's what adds up to bugs like this)

agilgur5:
> There are three places in the code base that still manually handle these details:

FYI, I edited your comment to add https://github.com/argoproj/argo-workflows/pull/12970#discussion_r1576605599.
Also changed the links to the codebase to use permalinks with a SHA so that they don't become outdated (general best practice).

util/printer/workflow-printer.go
agilgur5: > The existing check is `!tmpl.IsPodType()`, so I imagine `node.Type != NodeTypePod` would be equivalent

ok well actually this did not need much verification -- the function just above `IsPodType`, `GetNodeType`, has this [exact conversion logic](https://github.com/argoproj/argo-workflows/blob/3f9b992cdddb09633ef329c13eab1823ca6e5f3e/pkg/apis/workflow/v1alpha1/workflow_types.go#L3030). 

It would be `node.Type != wfv1.NodeTypePod` in this file specifically.

> Neither am I and there is no test coverage for the lines I'm changing.

Also there is an existing test case that covers this, the [`--wide` test](https://github.com/argoproj/argo-workflows/blob/3f9b992cdddb09633ef329c13eab1823ca6e5f3e/util/printer/workflow-printer_test.go#L87) ("P/R/C" = "Pending/Running/Completed"). More unit tests or test cases wouldn't hurt though

agilgur5: Simplified in #13291

Issue #12814, opened 2024-03-16
Workflow Details page makes unnecessary Archived request


ryancurrah:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

Users of Argo Workflows version ~3.5.5~ 3.5 experience an issue where the Workflow Details page attempts to fetch workflows as if they were archived, even when the archived workflows feature is not utilized. This behavior results in an unnecessary API call, leading to the error message `getting archived workflows not supported` and potentially degrading the UI experience.

The issue manifests before the workflow details page loads, with the UI erroneously trying to retrieve an archived version of the workflow based on its UID.

~Initial troubleshooting suggested setting the `ARGO_ARTIFACT_SERVER` environment variable to `false` as a potential fix. However, the error persisted on the workflow details page, and archived logs ceased to function after applying this workaround.~ _edited by agilgur5: this is actually unrelated, see [below comment](https://github.com/argoproj/argo-workflows/issues/12814#issuecomment-2002524593)_.

Discussion on [Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1710510295856659) highlighted a section of the frontend code responsible for this behavior, indicating that the UI checks for an archived workflow status unnecessarily. This check occurs even when archived workflows are not enabled, which should indicate that the workflow is not considered archived.

The conversation concluded with the suggestion to file a GitHub issue to track this problem, aiming for a more concise and efficient handling of workflow status checks in the UI to prevent unnecessary API calls and improve the user experience.

### Potential Solutions:

1. Adjust the frontend code to ensure that archived workflow checks are only performed when necessary, based on the presence of appropriate configuration settings.

### Version

v3.5.5

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
metadata:
  name: delightful-bear
  namespace: default
  labels:
    example: 'true'
spec:
  arguments:
    parameters:
      - name: message
        value: hello argo
  entrypoint: argosay
  templates:
    - name: argosay
      inputs:
        parameters:
          - name: message
            value: '{{workflow.parameters.message}}'
      container:
        name: main
[...]

PR #12972, opened 2024-04-24
fix(ui): try retrieving live workflow first then archived

ryancurrah:
Retrieving an archived workflow is the fallback now.

Related Slack thread: https://cloud-native.slack.com/archives/C01QW9QSSSK/p1710600357875049\?thread_ts\=1710510295.856659\&cid\=C01QW9QSSSK

Fixes https://github.com/argoproj/argo-workflows/issues/12814

### Motivation

Currently an error appears when loading a workflow because archived workflows are not enabled and the ui tries to get the archived workflow first. 

### Modifications

Changed the order in which workflows are retrieved from the back end.

### Verification

None, not sure how to verify at the moment.

### DIFF
--- a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
+++ b/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
@@ -388,6 +388,8 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
                 const wf = await services.workflows.get(namespace, name);
                 setUid(wf.metadata.uid);
                 setWorkflow(wf);
+                setError(null);
+                return;
             } catch (err) {
                 if (err.status !== 404 && uid === '') {
                     setError(err);
@@ -397,6 +399,8 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
                 try {
                     const archivedWf = await services.workflows.getArchived(namespace, uid);
                     setWorkflow(archivedWf);
+                    setError(null);
+                    return;
                 } catch (archiveErr) {
                     if (archiveErr.status === 500 && archiveErr.response.body.message === 'getting archived workflows not supported') {
                         setError(err);
--- a/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
+++ b/ui/src/app/workflows/components/workflow-details/workflow-details.tsx
@@ -384,33 +384,26 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
     // Get workflow
     useEffect(() => {
         (async () => {
-            let archivedWf: Workflow;
-            if (uid !== '') {
-                try {
-                    archivedWf = await services.workflows.getArchived(namespace, uid);
-                    setError(null);
-                } catch (err) {
-                    if (err.status !== 404) {
-                        setError(err);
-                    }
-                }
-            }
-
             try {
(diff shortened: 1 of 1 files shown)


ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -384,31 +384,18 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
...
+                setWorkflow(wf);
             } catch (err) {
-                if (archivedWf) {
-                    setWorkflow(archivedWf);
+                if (err.status === 404 && uid !== '') {
+                    try {
+                        const archivedWf = await services.workflows.getArchived(namespace, uid);
+                        setWorkflow(archivedWf);
+                    } catch (archiveErr) {
+                        setError(archiveErr);
```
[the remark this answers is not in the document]
agilgur5: You can catch that specific error from `archiveErr` and then set it to the 404 `err` instead

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -384,31 +384,18 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
...
-                    setUid(wf.metadata.uid);
-                } else {
-                    setWorkflow(archivedWf);
-                }
+                setUid(wf.metadata.uid);
+                setWorkflow(wf);
             } catch (err) {
-                if (archivedWf) {
-                    setWorkflow(archivedWf);
+                if (err.status === 404 && uid !== '') {
```
agilgur5, line 392: you can invert this conditional and return early to reduce nesting

ryancurrah: Done.

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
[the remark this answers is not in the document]
ryancurrah: Ok I caught the error looking for the specific error message, if true error out with the live workflow error.

ui/src/app/workflows/components/workflow-details/workflow-details.tsx
```diff
@@ -384,33 +384,26 @@ export function WorkflowDetails({history, location, match}: RouteComponentProps<
...
-                } catch (err) {
-                    if (err.status !== 404) {
-                        setError(err);
-                    }
-                }
-            }
-
             try {
                 const wf = await services.workflows.get(namespace, name);
-                setError(null);
```
agilgur5, line 401: `setError(null);` should still be done whenever successful -- in order to reset the error if there was one.

so both in this case and with the archived case

ryancurrah: How does this look? https://github.com/argoproj/argo-workflows/pull/12972/commits/37a5c8ee848d81c810f231407735f48256dfe282

agilgur5: the returns aren't strictly necessary, but nbd

ryancurrah: Ack. I'm out of my comfort zone in TypeScript, thanks for putting up with my noobness.

PR #12977, opened 2024-04-24
build(cli): add riscv64 support

mengzhuo:
### Motivation

As riscv growing popularity and being adopted by more projects
https://github.com/carlosedp/riscv-bringup 

and commercial Cloud provider like Scaleway https://www.scaleway.com/en/news/scaleway-launches-its-risc-v-servers-in-the-cloud-a-world-first-and-a-firm-commitment-to-technological-independence/

It's possible to run argo workflow by riscv users.

### Modifications

Adds riscv64 support

### Verification

 ```
make dist/argo-linux-riscv64 && ./dist/argo-linux-riscv64
```
argo prints help verbose.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,7 @@ endif
 dist/argo-linux-amd64: GOARGS = GOOS=linux GOARCH=amd64
 dist/argo-linux-arm64: GOARGS = GOOS=linux GOARCH=arm64
 dist/argo-linux-ppc64le: GOARGS = GOOS=linux GOARCH=ppc64le
+dist/argo-linux-riscv64: GOARGS = GOOS=linux GOARCH=riscv64
 dist/argo-linux-s390x: GOARGS = GOOS=linux GOARCH=s390x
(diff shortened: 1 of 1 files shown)


Joibel:
This will only support the CLI on riscV, but that's the only component we support on non amd64/arm64 for other less popular architectures.
It will not allow the workflow's kubernetes components to run on riscV.
Anyway, LGTM

Joibel (approved):

Joibel:
@mengzhuo, you will need to sign your commits according to https://github.com/argoproj/argoproj/blob/main/community/CONTRIBUTING.md. You will need to rewrite this commit to make the DCO check pass.

agilgur5:
> This will only support the CLI on riscV, but that's the only component we support on non amd64/arm64 for other less popular architectures.

^This, see also #12067 for the rest

agilgur5 (approved):
LGTM. There is existing precedent for CLI binaries, as Alan noted

PR #11777, opened 2023-09-08
feat: Add support for retrieving ClientId, ClientSecret and SSO secret from env var as opposed to creating a kube Secret

ebourgeois:
Fixes #10395
Fixes #11772

### Motivation

This change adds support for retrieving ClientId, ClientSecret and SSO secret from env var as opposed to creating a kube Secret, allowing us to provide our own and not use Kube Secrets at all.

### Modifications

The changes made were:

In the SSO auth code and config, to solve the above problem statement, we believe the following
would be required:

- Allow setting clientId from environment variable
- Allow setting clientSecret from environment variable
- Allow setting of PKCS1 SSO secret from environment variable

### Verification

We successfully:

1. backed up the `sso` secret
2. deleted the `sso` resource in our Kube namespace.
3. created 3 environment variables from our secret store:
    1. `ARGO_CLIENT_ID`: our SSO client ID
    2. `ARGO_CLIENT_SECRET`: our client secret
    3. `ARGO_PRIVATE_KEY`: PKCS1 key, encoded in base64
4. Added the the new mapping into the `sso` configmap:
```yaml
      privateKeyEnvName: ARGO_PRIVATE_KEY
      clientIdEnvName: ARGO_CLIENT_ID
      clientSecretEnvName: ARGO_CLIENT_SECRET
```
5. started workflow server successfully
6. able to login with our SSO and use our custom groups 

Regarding fix #10395:

In this PR, we are fixing issue: https://github.com/argoproj/argo-workflows/issues/10395
    - this issue was caused by the creation of a new httpClient and not
      re-using the existing one in `sso.go`, where `InsecureSkipVerify`
      is being set and not propagated tot the groups claim call

Issue #10395, opened 2023-01-24
Argo Server SSO returns 401 even with RBAC disabled


JunaidChaudry:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

Based on the documentation at https://argoproj.github.io/argo-workflows/argo-server-sso/ , my understanding is that when using Argo server auth mode sso, if RBAC is disabled, then any user able to authenticate with the OIDC should be able to login. 
However, I keep getting a 401 HTTP error after the oauth2/callback redirect. I have tried Dex OIDC as well as Microsoft, both giving the same end result... A page saying "This page isn’t working right nowIf the problem continues, contact the site owner. HTTP ERROR 401".

All that I see in the server logs is the following without much information. 
``` 
time="2023-01-24T15:23:19.799Z" level=info duration="40.682µs" method=GET path=/oauth2/redirect size=257 status=302
time="2023-01-24T15:23:33.734Z" level=info duration="91.996µs" method=GET path=index.html size=473 status=0
time="2023-01-24T15:23:44.949Z" level=info duration=65.110083ms method=GET path=/oauth2/callback size=0 status=401
```

*Steps to Reproduce*
- Deploy argoworkflow with the following configmap 
- Try to login using the OIDC

Here's my configmap 

```yaml
data:
  config: |
    sso:
      issuer: https://<dex-url>/dex
      clientId:
        name: argo-workflow-dev-oidc
        key: client-id
      clientSecret:
        name: argo-workflow-dev-oidc
        key: client-secret
      redirectUrl: https://<argo-workflow-url>/oauth2/callback
      rbac:
        enabled: false
    nodeEvents:
      enabled: true
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: argo-workflows
    meta.helm.sh/release-namespace: argow
  creationTimestamp: "2023-01-23T22:19:50Z"
  labels:
    app.kubernetes.io/component: workflow-controller
    app.kubernetes.io/instance: argo-workflows
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: argo-workflows-cm
    app.kubernetes.io/part-of: argo-workflows
    helm.sh/chart: argo-workflows-0.22.8
[...]

PR #12982, opened 2024-04-25
fix: `insecureSkipVerify` for `GetUserInfoGroups`

sycured:
### Motivation

When running with `insecureSkipVerify: true` and a `userInfoPath` like `userInfoPath: /userinfo`, the call is falling because it's still checking the certificate.

### Modifications

I pass the `httpClient` built for the SSO as a parameter of the function `GetUserInfoGroups`.

### Verification

I ran it using Argo CD with Dex and custom (internal certificate).
Argo Workflows doesn't fail anymore about certificate verification.

### DIFF
--- a/server/auth/types/claims.go
+++ b/server/auth/types/claims.go
@@ -85,7 +85,7 @@ func (c *Claims) GetCustomGroup(customKeyName string) ([]string, error) {
 	return newSlice, nil
 }
 
-func (c *Claims) GetUserInfoGroups(accessToken, issuer, userInfoPath string) ([]string, error) {
+func (c *Claims) GetUserInfoGroups(httpClient HttpClient, accessToken, issuer, userInfoPath string) ([]string, error) {
 	url := fmt.Sprintf("%s%s", issuer, userInfoPath)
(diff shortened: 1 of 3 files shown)


agilgur5:
This would fix #10395 and properly split the fix from https://github.com/argoproj/argo-workflows/pull/11777#pullrequestreview-1674742405 as well

server/auth/types/claims.go
```diff
@@ -85,7 +85,7 @@ func (c *Claims) GetCustomGroup(customKeyName string) ([]string, error) {
 	return newSlice, nil
 }
 
-func (c *Claims) GetUserInfoGroups(accessToken, issuer, userInfoPath string) ([]string, error) {
+func (c *Claims) GetUserInfoGroups(httpClient HttpClient, accessToken, issuer, userInfoPath string) ([]string, error) {
```
agilgur5, line 88: great this resolves https://github.com/argoproj/argo-workflows/pull/11777#discussion_r1357553506 and related comments

agilgur5 (approved):
LGTM, thanks for neatly fixing this!

PR #12984, opened 2024-04-26
feat(executor): set seccomp profile to runtimedefault

lukashankeln:
### Motivation

We are using Kyverno Policies to enforce different security settings within our cluster. For example running containers as non Root.

But also to explicitly have set the SeccompProfile to RuntimeDefault. See the [official Documentation](https://kubernetes.io/docs/tutorials/security/seccomp/#create-a-pod-that-uses-the-container-runtime-default-seccomp-profile)

Currently the artifact GC is beeing blocked by our Kyverno policy. On my search for values to change the default (e.g. via Helm) i found the code I changed in this PR.

### Modifications

Explicitly Setting the Seccomp Profile to RuntimeDefault when creating the artifact GC Pod.

### Verification

The E2E tests should catch any issues, if you want me to test this locally let me know

### DIFF
--- a/workflow/common/security_context.go
+++ b/workflow/common/security_context.go
@@ -1,0 +1,26 @@
+package common
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
+)
+
+func MinimalCtrSC() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+		Privileged:               pointer.Bool(false),
+		RunAsNonRoot:             pointer.Bool(true),
+		RunAsUser:                pointer.Int64(8737),
+		ReadOnlyRootFilesystem:   pointer.Bool(true),
+		AllowPrivilegeEscalation: pointer.Bool(false),
+		SeccompProfile:           &corev1.SeccompProfile{Type: "RuntimeDefault"},
+	}
+}
+
+func MinimalPodSC() *corev1.PodSecurityContext {
+	return &corev1.PodSecurityContext{
+		RunAsNonRoot:   pointer.Bool(true),
+		RunAsUser:      pointer.Int64(8737),
+		SeccompProfile: &corev1.SeccompProfile{Type: "RuntimeDefault"},
+	}
+}
--- a/workflow/controller/artifact_gc.go
+++ b/workflow/controller/artifact_gc.go
@@ -431,7 +431,8 @@ func (woc *wfOperationCtx) createArtifactGCPod(ctx context.Context, strategy wfv
 			OwnerReferences: ownerReferences,
 		},
 		Spec: corev1.PodSpec{
-			Volumes: volumes,
+			Volumes:         volumes,
+			SecurityContext: common.MinimalPodSC(),
 			Containers: []corev1.Container{
 				{
 					Name:            common.MainContainerName,
@@ -444,14 +445,7 @@ func (woc *wfOperationCtx) createArtifactGCPod(ctx context.Context, strategy wfv
 					// if this pod is breached by an attacker we:
 					// * prevent installation of any new packages
 					// * modification of the file-system
-					SecurityContext: &corev1.SecurityContext{
-						Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
-						Privileged:               pointer.Bool(false),
-						RunAsNonRoot:             pointer.Bool(true),
-						RunAsUser:                pointer.Int64(8737),
-						ReadOnlyRootFilesystem:   pointer.Bool(true),
-						AllowPrivilegeEscalation: pointer.Bool(false),
-					},
+					SecurityContext: common.MinimalCtrSC(),
 					// if this pod is breached by an attacker these limits prevent excessive CPU and memory usage
 					Resources: corev1.ResourceRequirements{
 						Limits: map[corev1.ResourceName]resource.Quantity{
--- a/test/e2e/executor_plugins_test.go
+++ b/test/e2e/executor_plugins_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 	apiv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
 
@@ -39,8 +40,9 @@ func (s *ExecutorPluginsSuite) TestTemplateExecutor() {
 				spec := pod.Spec
 				assert.Equal(t, pointer.Bool(false), spec.AutomountServiceAccountToken)
 				assert.Equal(t, &apiv1.PodSecurityContext{
-					RunAsUser:    pointer.Int64(8737),
-					RunAsNonRoot: pointer.Bool(true),
+					RunAsUser:      pointer.Int64(8737),
+					RunAsNonRoot:   pointer.Bool(true),
+					SeccompProfile: &v1.SeccompProfile{Type: "RuntimeDefault"},
 				}, spec.SecurityContext)
 				if assert.Len(t, spec.Volumes, 4) {
 					assert.Contains(t, spec.Volumes[0].Name, "kube-api-access-")
@@ -71,7 +73,9 @@ func (s *ExecutorPluginsSuite) TestTemplateExecutor() {
 								RunAsNonRoot:             pointer.Bool(true),
 								AllowPrivilegeEscalation: pointer.Bool(false),
 								ReadOnlyRootFilesystem:   pointer.Bool(true),
+								Privileged:               pointer.Bool(false),
 								Capabilities:             &apiv1.Capabilities{Drop: []apiv1.Capability{"ALL"}},
+								SeccompProfile:           &v1.SeccompProfile{Type: "RuntimeDefault"},
 							}, agent.SecurityContext)
 						}
 					}
--- a/workflow/controller/agent.go
+++ b/workflow/controller/agent.go
@@ -174,15 +174,7 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
 		Image:           woc.controller.executorImage(),
 		ImagePullPolicy: woc.controller.executorImagePullPolicy(),
 		Env:             envVars,
-		SecurityContext: &apiv1.SecurityContext{
-			Capabilities: &apiv1.Capabilities{
-				Drop: []apiv1.Capability{"ALL"},
-			},
-			RunAsNonRoot:             pointer.Bool(true),
-			RunAsUser:                pointer.Int64(8737),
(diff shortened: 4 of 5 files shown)


agilgur5:
You might notice `podSpecPatch` a few lines below -- you can set that to change anything on the Pod

workflow/controller/artifact_gc.go
```diff
@@ -451,6 +451,7 @@ func (woc *wfOperationCtx) createArtifactGCPod(ctx context.Context, strategy wfv
 						RunAsUser:                pointer.Int64(8737),
 						ReadOnlyRootFilesystem:   pointer.Bool(true),
 						AllowPrivilegeEscalation: pointer.Bool(false),
+						SeccompProfile:           &corev1.SeccompProfile{Type: "RuntimeDefault"},
```
agilgur5, line 454: Is `seccomp` supported everywhere? AppArmor & SELinux aren't IIRC. Want to make sure we don't break compatibility

Also if we set this here, we should really set this on the `agent` and on `resource` containers as well (all 3 are locked down -- perhaps some of those could be refactored into a common config)

lukashankeln: Thank you for the hint regarding the `podSpecPatch` that did fix our initial problem. 

<details>
  <summary>For everyone finding this in the future, these are the changes I made to our workflows that produce artifacts</summary>
  
```
kind: CronWorkflow
metadata:
  name: some-workflow
  namespace: some-namespace
spec:
  workflowSpec:
    artifactGC:
      podSpecPatch: |
        securityContext:
          seccompProfile:
            type: RuntimeDefault
```
</details>

Setting this for every workflow might not be the prettiest solutions but it works for us. 

After a short little search  it seems like seccomp was first introduced in the linux kernel in version 2.6.12 in 2005 and should be present in every more or less recent linux distro.
But i cant be 100% certain and i dont know how k8s would handle this scenario.

How do you want to continue with this?
If  you want to continue with this, i would implement the changes you suggested above.

agilgur5: It seems like it should be supported in every k8s distro afaict. It seems like if a CRI doesn't support seccomp the setting will just be ignored.
 And yes all non-EoL Linux support it.

I can't find a possible compat issue from a handful of searches either, so it seems safe enough to include, particularly as a feature.
Based on that, let's go ahead and implement it on this and the other two containers too. We'll get feedback on the RCs if there are compat issues.

Do you also want to file a follow-up issue to create more specific seccomp profiles for each container? Since we should be able to determine all their possible syscalls ahead of time. And then for newer k8s versions could [distribute those profiles via annotations](https://kubernetes.io/blog/2024/03/07/cri-o-seccomp-oci-artifacts/)

agilgur5: > Setting this for every workflow might not be the prettiest solutions but it works for us.

You can use [`workflowDefaults`](https://argo-workflows.readthedocs.io/en/latest/default-workflow-specs/) to set this controller-wide

lukashankeln: Sorry, it took me a while to come back to this.

I refactored the SecurityContext for Pod and Container into a common function
(my thought to a function was because auf immutability)

Added this in the agent and the artifact_gc.

I couldn't find the `resource` you talked about, can you give me a hint on where to look?

workflow/common/security_context.go
```diff
@@ -0,0 +1,26 @@
+package common
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
+)
+
+func ContainerSecurityContext() *corev1.SecurityContext {
```
agilgur5, line 8:
```suggestion
func MinimalCtrSC() *corev1.SecurityContext {
```

I think we should be more specific here with "minimal".

The other part is just shortening the name with common abbreviations

workflow/common/security_context.go
```diff
@@ -0,0 +1,26 @@
...
+		Privileged:               pointer.Bool(false),
+		RunAsNonRoot:             pointer.Bool(true),
+		RunAsUser:                pointer.Int64(8737),
+		ReadOnlyRootFilesystem:   pointer.Bool(true),
+		AllowPrivilegeEscalation: pointer.Bool(false),
+		SeccompProfile:           &corev1.SeccompProfile{Type: "RuntimeDefault"},
+	}
+}
+
+func PodSecurityContext() *corev1.PodSecurityContext {
```
agilgur5, line 20:
```suggestion
func MinimalPodSC() *corev1.PodSecurityContext {
```

same as above

workflow/controller/artifact_gc.go
agilgur5: > that makes sense to me off-hand, but that's the one place I'll need to double-check if there's a better place to put it / name it / etc

small naming adjustment, but otherwise I think this is a good place to put it. we should probably split `common.go` into more files for it to be easier to find things. nice choice 🙂

agilgur5 (commented):
have some naming suggestions below

agilgur5 (commented):
1 change needed to the `resource` container, otherwise LGTM

agilgur5:
Oh and the other [remaining test failure](https://github.com/argoproj/argo-workflows/actions/runs/9028328717/job/24808748392?pr=12984) needs a change in `TestExecutorPluginsSuite/TestTemplateExecutor` to handle the now set SC fields (it was expecting them to be `nil` before):

```
executor_plugins_test.go:41: 
        	Error Trace:	/home/runner/work/argo-workflows/argo-workflows/test/e2e/executor_plugins_test.go:41
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/fixtures/then.go:262
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/executor_plugins_test.go:36
        	Error:      	Not equal: 
        	            	expected: &v1.PodSecurityContext{SELinuxOptions:(*v1.SELinuxOptions)(nil), WindowsOptions:(*v1.WindowsSecurityContextOptions)(nil), RunAsUser:(*int64)(0xc000107280), RunAsGroup:(*int64)(nil), RunAsNonRoot:(*bool)(0xc000107288), SupplementalGroups:[]int64(nil), FSGroup:(*int64)(nil), Sysctls:[]v1.Sysctl(nil), FSGroupChangePolicy:(*v1.PodFSGroupChangePolicy)(nil), SeccompProfile:(*v1.SeccompProfile)(nil)}
        	            	actual  : &v1.PodSecurityContext{SELinuxOptions:(*v1.SELinuxOptions)(nil), WindowsOptions:(*v1.WindowsSecurityContextOptions)(nil), RunAsUser:(*int64)(0xc000106f58), RunAsGroup:(*int64)(nil), RunAsNonRoot:(*bool)(0xc000106fe4), SupplementalGroups:[]int64(nil), FSGroup:(*int64)(nil), Sysctls:[]v1.Sysctl(nil), FSGroupChangePolicy:(*v1.PodFSGroupChangePolicy)(nil), SeccompProfile:(*v1.SeccompProfile)(0xc00050d290)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -10,3 +10,6 @@
        	            	  FSGroupChangePolicy: (*v1.PodFSGroupChangePolicy)(<nil>),
        	            	- SeccompProfile: (*v1.SeccompProfile)(<nil>)
        	            	+ SeccompProfile: (*v1.SeccompProfile)({
        	            	+  Type: (v1.SeccompProfileType) (len=14) "RuntimeDefault",
        	            	+  LocalhostProfile: (*string)(<nil>)
        	            	+ })
        	            	 })
        	Test:       	TestExecutorPluginsSuite/TestTemplateExecutor
    executor_plugins_test.go:69: 
        	Error Trace:	/home/runner/work/argo-workflows/argo-workflows/test/e2e/executor_plugins_test.go:69
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/fixtures/then.go:262
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/executor_plugins_test.go:[36](https://github.com/argoproj/argo-workflows/actions/runs/9028328717/job/24808748392?pr=12984#step:18:37)
        	Error:      	Not equal: 
        	            	expected: &v1.SecurityContext{Capabilities:(*v1.Capabilities)(0xc0005a6360), Privileged:(*bool)(nil), SELinuxOptions:(*v1.SELinuxOptions)(nil), WindowsOptions:(*v1.WindowsSecurityContextOptions)(nil), RunAsUser:(*int64)(0xc000728e18), RunAsGroup:(*int64)(nil), RunAsNonRoot:(*bool)(0xc000728e20), ReadOnlyRootFilesystem:(*bool)(0xc000728e22), AllowPrivilegeEscalation:(*bool)(0xc000728e21), ProcMount:(*v1.ProcMountType)(nil), SeccompProfile:(*v1.SeccompProfile)(nil)}
        	            	actual  : &v1.SecurityContext{Capabilities:(*v1.Capabilities)(0xc0001271a0), Privileged:(*bool)(0xc000106dbf), SELinuxOptions:(*v1.SELinuxOptions)(nil), WindowsOptions:(*v1.WindowsSecurityContextOptions)(nil), RunAsUser:(*int64)(0xc000106dc0), RunAsGroup:(*int64)(nil), RunAsNonRoot:(*bool)(0xc000106dcc), ReadOnlyRootFilesystem:(*bool)(0xc000106dcd), AllowPrivilegeEscalation:(*bool)(0xc000106dce), ProcMount:(*v1.ProcMountType)(nil), SeccompProfile:(*v1.SeccompProfile)(0xc000[50](https://github.com/argoproj/argo-workflows/actions/runs/9028328717/job/24808748392?pr=12984#step:18:51)d278)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -7,3 +7,3 @@
        	            	  }),
        	            	- Privileged: (*bool)(<nil>),
        	            	+ Privileged: (*bool)(false),
        	            	  SELinuxOptions: (*v1.SELinuxOptions)(<nil>),
        	            	@@ -16,3 +16,6 @@
        	            	  ProcMount: (*v1.ProcMountType)(<nil>),
        	            	- SeccompProfile: (*v1.SeccompProfile)(<nil>)
        	            	+ SeccompProfile: (*v1.SeccompProfile)({
        	            	+  Type: (v1.SeccompProfileType) (len=14) "RuntimeDefault",
        	            	+  LocalhostProfile: (*string)(<nil>)
        	            	+ })
        	            	 })
        	Test:       	TestExecutorPluginsSuite/TestTemplateExecutor
=== FAIL: ExecutorPluginsSuite/TestTemplateExecutor
```

agilgur5 (approved):
LGTM, thanks for iterating on this and improving the default security contexts!

lukashankeln:
> LGTM, thanks for iterating on this and improving the default security contexts!

Thank you for helping me out. One test was still failing, that should pass now.

workflow/controller/artifact_gc.go
lukashankeln: > Do you also want to file a follow-up issue to create more specific seccomp profiles for each container? Since we should be able to determine all their possible syscalls ahead of time. And then for newer k8s versions could [distribute those profiles via annotations](https://kubernetes.io/blog/2024/03/07/cri-o-seccomp-oci-artifacts/)


fyi: created the new issue: https://github.com/argoproj/argo-workflows/issues/13045

PR #12989, opened 2024-04-28
build: Remove misleading logs from crdgen.sh

terrytangyuan:
The current logs are misleading which may look like both the minimal and full CRDs are applied to the cluster twice:
```
Patching ...
Creating ...
```
The PR makes it less confusing.


agilgur5 (approved):
LGTM. I've never noticed this but yea the words "creating" and "patching" are k8s terms and so could definitely be misunderstood, so seems like a helpful change to me

PR #12149, opened 2023-11-04
build: rename `master` branch to `main` branch

agilgur5:
Fixes #12127

### Motivation

- Prefer more inclusive language

### Modifications

- change GH Actions to use `main` branch in several branch checks
  - and change GH issue templates to use `main` branch
- change scripts to use `main` branch

- change docs to use `main` branch
  - also in many places we had absolute links instead of relative links
    - docs convention is to use relative links where possible
    - changing these to relative links meant that we no longer need to change the branch
    - some docs also linked to the codebase when there's a page in the docs for it; relative link to the docs where appropriate
  - change a few docs to use a permalink instead of a link to `master`
    - Upgrading Notes and Proposals in particular linked to a specific line in a file, which is no longer the correct line
      - fix the broken links with permalinks, since those docs are time-specific
  - change a few links to `argoproj/argoproj/master` to use `main` as well, since per https://github.com/argoproj/argoproj/issues/35 that repo's branch was already renamed 
  - misc docs improvements while changing these lines
    - use code backticks when referring to specific resource names, directories, etc, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#code-style-inline-code)
    - prefer 1 sentence per line where the whole file didn't need changing
    - add a few links to other docs where appropriate (e.g. when referencing environment variables)
    - improve ambiguous link text, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#links)
    - etc

- change UI references (that linked to examples) to use the `main` branch
  - and change one UI reference to link to the rendered docs website instead of the docs on `main`

### Verification

1. `make docs` and `make codegen` passes

1. Used a PR in my own fork (https://github.com/agilgur5/argo-workflows/pull/2) for testing purposes [as suggested](https://github.com/argoproj/argo-workflows/pull/12149#pullrequestreview-1745563352) below.
    - All actions [ran correctly](https://github.com/agilgur5/argo-workflows/actions/runs/6990844946?pr=2) as intended on the `main` branch.

### Notes to Reviewers

We should rename the branch on GH _before_ this is merged.
The branch rename functionality in GH will ensure that everything is redirected to the new branch name as well.

PR #12999, opened 2024-04-30
fix(docs): Update buildkit example

highb:
### Motivation

There are two issues with the buildkit example: 1) The default branch is `master` but the argo-workflows repo only has `main` 2) The build task refers to a go project that no longer exists in this repo.

### Modifications

* Updated the default branch for the example to `main`
* Removed the `build` step entirely from the example

If preferred, I could modify the build step to use the example go project from that example workflow instead.

### Verification

Ran the workflow from my local Argo workflows instance after setting up the Docker credentials as noted by the example.

### DIFF
--- a/examples/buildkit-template.yaml
+++ b/examples/buildkit-template.yaml
@@ -35,7 +35,7 @@ spec:
       - name: repo
         value: https://github.com/argoproj/argo-workflows
       - name: branch
-        value: master
+        value: main
       - name: path
         value: test/e2e/images/argosay/v2
       - name: image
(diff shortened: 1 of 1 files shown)


agilgur5 (commented):
Thanks for updating the outdated contents of this example!

I definitely missed `main` here in #12149, although GH should redirect the request such that it still works, did it not?

> If preferred, I could modify the build step to use the example go project from that example workflow instead.

I think we do want to show a build step in this example for it to more closely resemble a standard build pipeline.

PR #12926, opened 2024-04-11
fix: setBucketLifecycleRule error in OSS Artifact Driver.  Fixes #12925

AlbeeSo:
Fixes #12925

### Motivation

OSS does not support to set different rules at a time for a same prefix and that's why I received a 400 error. And more details added in related issue.
```
time="2024-04-09T16:04:23.147Z" level=info msg="Save artifact" artifactName=out duration=1.270302632s error="oss: service returned error: StatusCode=400, ErrorCode=InvalidRequest, ErrorMessage=\"Found same prefix in different rules\", RequestId=6615670711BF5732361C11E2, Ec=0014-00000085" key=argo-workflows/test/lifecycletime="2024-04-09T16:04:23.147Z" level=error msg="executor error: oss: service returned error: StatusCode=400, ErrorCode=InvalidRequest, ErrorMessage=\"Found same prefix in different rules\", RequestId=6615670711BF5732361C11E2, Ec=0014-00000085"
```

### Modifications

LifecycleRule modified to:
- Set Transitions with MarkInfrequentAccessAfterDays (Objects will be converted to IA)
- Set Expiration with MarkDeletionAfterDays. (Objects will be deleted if Versioning is disabled, or changed to non-current version or else)
- Set AbortMultipartUpload with MarkDeletionAfterDays also. (Aborted Uploaded Parts will be deleted)

### Verification

Please test with workflow
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: set-lifecycle-
spec:
  entrypoint: main
  templates:
    - name: main
      metadata:
        annotations:
          k8s.aliyun.com/eci-extra-ephemeral-storage: "5Gi"
          k8s.aliyun.com/eci-use-specs : "ecs.g7.xlarge"
      container:
        image: alpine:latest
        command:
          - sh
          - -c
        args:
          - |
            mkdir -p /out
            dd if=/dev/random of=/out/lifecycle.txt bs=2M count=1024
            echo "created files!"
      outputs:
        artifacts:
          - name: out
            path: /out/lifecycle.txt
            oss:
              endpoint: http://oss-cn-zhangjiakou-internal.aliyuncs.com
              bucket: my-argo-workflow
              key: argo-workflows/test/lifecycle
              accessKeySecret:
                name: my-argo-workflow-credentials
                key: accessKey
              secretKeySecret:
                name: my-argo-workflow-credentials
                key: secretKey
              lifecycleRule:
[...]

PR #12741, opened 2024-03-05
fix: retry large archived wf. Fixes #12740

heidongxianhua:
Fixes #12740

PR #13002, opened 2024-05-02
fix: test if e2e tests are failing due to flakiness

isubasinghe:
Fixes #TODO

### DIFF
--- a/workflow/controller/dag.go
+++ b/workflow/controller/dag.go
@@ -95,6 +95,7 @@ func (d *dagContext) GetTaskDependsLogic(taskName string) string {
 }
 
 func (d *dagContext) resolveDependencies(taskName string) {
+	// TESTING CODE CHANGE TO SEE IF E2E tests fail
 	dependencies, resolvedDependsLogic := common.GetTaskDependencies(d.GetTask(taskName), d)
 	var dependencyTasks []string
 	for dep := range dependencies {


agilgur5:
Rebased the branch to an earlier version of `main` before #12736 was merged to see if that's the cause. 
See also [Slack thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1714810708644859)

agilgur5:
CI is still [failing](https://github.com/argoproj/argo-workflows/actions/runs/8951659032/job/24588254052?pr=13002) with the same errors, so my next guess would be #12741

agilgur5:
And it's still [failing](https://github.com/argoproj/argo-workflows/actions/runs/8951849772/job/24588624747?pr=13002) with the same error on a version of `main` prior to #12741 being merged. I'm even more confused now....

isubasinghe:
Hmm I thought it was https://github.com/argoproj/argo-workflows/pull/12736 as well due to a git blame I came upon. 

I suppose the quickest way to find out what the first offending commit is to git bisect.

agilgur5:
That uh assumes knowing a "good" commit, and I'm not sure which one is good 😬
Alan and I were thinking [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1714810708644859?thread_ts=1714810708.644859&cid=C0510EUH90V) that it could potentially be an unpinned dep somewhere, though I couldn't pinpoint one that added up in the manifests. I was looking at the debug logs and wondering if the tests might be pulling `argocli:latest` from the web instead of the locally built one

agilgur5:
The last rebase I tried was on top of / after #12972, and it's still failing with the same [errors](https://github.com/argoproj/argo-workflows/actions/runs/8952027939/job/24588994127?pr=13002)

agilgur5:
Ok rebased once more to after #12917 annnd [still same error](https://github.com/argoproj/argo-workflows/actions/runs/8955031896/job/24595351345?pr=13002). I chose that one as it was what #12926 was based on, which just merged (i.e. passed E2E on its own branch).

So pretty sure it's not due to a commit at this point

agilgur5:
Noting here that the failing API test does result in a nil pointer dereference error in the Server logs:

```
server: time="2024-05-05T02:31:39.703Z" level=info msg="Get artifact file" artifactName=local-artifact namespace=argo nodeId=wf-stopped-8j2gw-1075147760 uid=3b15ead1-612c-4820-8d62-e72018069605
server: 2024/05/05 02:31:39 http: panic serving [::1]:54866: runtime error: invalid memory address or nil pointer dereference
server: goroutine 295 [running]:
server: net/http.(*conn).serve.func1()
server: 	/opt/hostedtoolcache/go/1.21.9/x64/src/net/http/server.go:1868 +0xb9
server: panic({0x2994660?, 0x4b2cd40?})
server: 	/opt/hostedtoolcache/go/1.21.9/x64/src/runtime/panic.go:920 +0x270
server: github.com/argoproj/argo-workflows/v3/server/artifacts.(*ArtifactServer).GetArtifactFile(0xc000716960, {0x3303b00, 0xc0002d80e0}, 0xc000c3c000)
server: 	/home/runner/work/argo-workflows/argo-workflows/server/artifacts/artifact_server.go:141 +0x8b8
```

Joibel:
I've managed to produce `ErrImageNeverPull: Container image "argoproj/argocli:latest" is not present with pull policy of Never` locally by adding a Never image pull policy for `artgc-dag-workflow-stopper` which is used in `artgc-dag-wf-self-delete`, one of the failing tests.

I can't see where the argocli image gets built and pushed during e2e, so I'll see if I can find a satisfactory fix for this.

Joibel:
This is fixed in #13018

agilgur5:
> Alan and I were thinking [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1714810708644859?thread_ts=1714810708.644859&cid=C0510EUH90V) that it could potentially be an unpinned dep somewhere, though I couldn't pinpoint one that added up in the manifests. I was looking at the debug logs and wondering if the tests might be pulling `argocli:latest` from the web instead of the locally built one

> I've managed to produce `ErrImageNeverPull: Container image "argoproj/argocli:latest" is not present with pull policy of Never` locally by adding a Never image pull policy for `artgc-dag-workflow-stopper` which is used in `artgc-dag-wf-self-delete`, one of the failing tests.

> This is fixed in #13018

Thanks Alan for confirming my hypothesis and fixing the `imagePullPolicy` so that we don't run into that issue again!

We should still pin all the other images in the test manifests too, to prevent a similar type of bug (or supply chain attack via tests)

PR #13004, opened 2024-05-02
fix: ignore retry node when check succeeded descendant nodes. Fixes: #13003

### LINKED ISSUE #13003: Workflows that failed before upgrade to 3.5.6 fail to retry
### Pre-requisites

1. We were on 3.5.5
1. Some workflows failed
1. Upgraded to 3.5.6
1. Retried some of the workflows
Result:

    - The failed tasks within the workflows disappeared, keeping only the successful ones
    - The workflows got stuck in a running state
    - Still The workflows are shown in the failed list
    - Unable to retry again, because it's running
    - Unable to Stop 

Reproducible on any workflow

### Version

v3.5.6

```YAML
any workflow reproduces it
```

```text
```

```text
```

shuangkun:
ignore retry node when check descendant nodes.

Fixes #13003

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -971,7 +971,7 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 				log.Debugf("Reset %s node %s since it's a group node", node.Name, string(node.Phase))
 				continue
 			} else {
-				if isDescendantNodeSucceeded(wf, node, nodeIDsToReset) {
+				if node.Type != wfv1.NodeTypeRetry && isDescendantNodeSucceeded(wf, node, nodeIDsToReset) {
 					log.Debugf("Node %s remains as is since it has succeed child nodes.", node.Name)
 					newWF.Status.Nodes.Set(node.ID, node)
 					continue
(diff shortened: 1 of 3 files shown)


workflow/util/util.go
```diff
@@ -971,7 +971,7 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 				log.Debugf("Reset %s node %s since it's a group node", node.Name, string(node.Phase))
 				continue
 			} else {
-				if isDescendantNodeSucceeded(wf, node, nodeIDsToReset) {
+				if node.Type == wfv1.NodeTypePod && isDescendantNodeSucceeded(wf, node, nodeIDsToReset) {
```
Joibel, line 974: It feels like this isn't the whole list of NodeTypes that should be considered here.
HTTP, Plugin and Container come to mind, but I haven't thought it through completely.

Joibel (changes requested):

workflow/util/util.go
shuangkun: Oh, I originally wrote it as not equal to retry, but there was a test error, and now I changed it back.

terrytangyuan (approved):

PR #12858, opened 2024-03-28
refactor: use context from `processNextPodCleanupItem`

blkperl:
### Motivation

Addresses https://github.com/argoproj/argo-workflows/pull/12847#discussion_r1543760541

### Modifications

* Use the context from processNextPodCleanupItem in GetExecutorOutput

### Verification

Existing test coverage will be sufficient

PR #13012, opened 2024-05-06
fix: add timeout for executor signal

### LINKED ISSUE #13011: add timeout for executor signal
### Pre-requisites

The method ExecPodContainerAndGetOutput in signal.go will stuck, ince this method  (GetExecutorOutput) blocks while waiting for the remote stdout output, causing the goroutine to block for a long time, or even for a long time, we should add a timeout to prevent this problem from happening again.

This causes the cleanup to block all four goroutines by default, thus growing indefinitely in the pod_clean_up queue depth

### Version

main

```YAML
none
```

```text
none
```

```text
none
```

fyp711:
Fixes #13011

### Modifications

I added a context timeout to prevent spdy from blocking for a long time

### DIFF
--- a/workflow/signal/signal.go
+++ b/workflow/signal/signal.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -49,6 +50,8 @@ func ExecPodContainerAndGetOutput(ctx context.Context, restConfig *rest.Config, 
 	if err != nil {
 		return err
 	}
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
 	stdout, stderr, err := common.GetExecutorOutput(ctx, x)
 	log.
 		WithField("namespace", namespace).
--- a/workflow/signal/signal.go
+++ b/workflow/signal/signal.go
@@ -50,6 +50,7 @@ func ExecPodContainerAndGetOutput(ctx context.Context, restConfig *rest.Config, 
 	if err != nil {
 		return err
 	}
+	// workaround for when exec does not properly return: https://github.com/kubernetes/kubernetes/pull/103177
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
 	defer cancel()
 	stdout, stderr, err := common.GetExecutorOutput(ctx, x)


fyp711:
Could someone take a discussion for this issue ? It will block all pod cleanup workers. It's very serious.

workflow/signal/signal.go
```diff
@@ -15,6 +17,8 @@ import (
 	"github.com/argoproj/argo-workflows/v3/workflow/common"
 )
 
+var spdyTimeout = env.LookupEnvDurationOr("SPDY_TIMEOUT", 10*time.Minute)
```
Joibel, line 20: The name of the environment variable to control this is difficult to 'discover'.

I'd suggest `POD_OUTPUT_TIMEOUT` or something along those lines. No user cares about the protocol in use.

This will also need adding to the documentation.

workflow/signal/signal.go
```diff
@@ -49,6 +53,8 @@ func ExecPodContainerAndGetOutput(ctx context.Context, restConfig *rest.Config,
 	if err != nil {
 		return err
 	}
+	ctx, cancel := context.WithTimeout(ctx, spdyTimeout)
```
Joibel, line 56: Make spdyTimeout a local variable to this function, rather than a global please.

Joibel (changes requested):

workflow/signal/signal.go
fyp711: Thanks for your review, i will fix it. Thanks

workflow/signal/signal.go
fyp711: > Make spdyTimeout a local variable to this function, rather than a global please.

Thanks, i will fix it.

workflow/signal/signal.go
agilgur5: `EXECUTOR_SIGNAL_TIMEOUT` was what I was thinking.

Although I'm not sure that this needs to be configurable in the first place, or as long as 10min.

It isn't entirely clear what we should do after a timeout error either though

fyp711: > It isn't entirely clear what we should do after a timeout error either though, requeueing may not fix it

Thanks for your advices,  in our environment, we observe that the pod has actually been removed, but this goroutine will still block here.

fyp711: So, in my opinion, we should set a timeout that can never happen. The original purpose of this method change in client-go was also to support timeout cancelable
image

fyp711: > So, in my opinion, we should set a timeout that can never happen. The original purpose of this method change in client-go was also to support timeout cancelable

see https://github.com/kubernetes/kubernetes/pull/103177

fyp711:
@Joibel @agilgur5 Hi, could you help me to review again? I have already fixed the issues mentioned earlier. Thanks!

Joibel:
> @Joibel @agilgur5 Hi, could you help me to review again? I have already fixed the issues mentioned earlier. Thanks!

Could you answer my question in #13011 on when you think this might be useful please.

fyp711:
> > @Joibel @agilgur5 Hi, could you help me to review again? I have already fixed the issues mentioned earlier. Thanks!
> 
> Could you answer my question in #13011 on when you think this might be useful please.

Okay, I missed that message. I'm sorry. Now I'll reply

docs/environment-variables.md
```diff
@@ -31,6 +31,7 @@ This document outlines environment variables that can be used to customize behav
...
 | `EVENT_AGGREGATION_WITH_ANNOTATIONS`     | `bool`              | `false`                                                                                     | Whether event annotations will be used when aggregating events.                                                                                                                                                                                                          |
+| `EXECUTOR_SIGNAL_TIMEOUT`                | `time.Duration`     | `10m`                                                                                       | When the executor does not finish after this time, it will times out and simply returns an err                                                                                                                                                                           |
```
Joibel, line 34: Suggest rewording the help:
`How long to wait after the executor has been terminated before returning an error`

Joibel (changes requested):

docs/environment-variables.md
fyp711: Default I think 10 minutes will do, which is a time that will never happen. It's also a very safe time. Thanks

fyp711: I think that's a good idea, Thanks!

docs/environment-variables.md
```diff
@@ -31,6 +31,7 @@ This document outlines environment variables that can be used to customize behav
...
 | `EVENT_AGGREGATION_WITH_ANNOTATIONS`     | `bool`              | `false`                                                                                     | Whether event annotations will be used when aggregating events.                                                                                                                                                                                                          |
+| `EXECUTOR_SIGNAL_TIMEOUT`                | `time.Duration`     | `10m`                                                                                       | When the executor does not finish after this time, it will times out and simply returns an err. I'm happy with 10 minutes.                                                                                                                                               |
```
agilgur5, line 34: looks this was a copy+paste -- that would not be an appropriate documentation line
```suggestion
| `EXECUTOR_SIGNAL_TIMEOUT`                | `time.Duration`     | `10m`                                                                                       | Timeout for a signal to the executor.                                                                                                                                               |
```

What Alan suggested [above](https://github.com/argoproj/argo-workflows/pull/13012#discussion_r1593644855) was:
```suggestion
| `EXECUTOR_SIGNAL_TIMEOUT`                | `time.Duration`     | `10m`                                                                                       | How long to wait after the executor has been terminated before timing out.                                                                                                                                               |
```

I made a small modification to that as well

workflow/signal/signal.go
agilgur5: > Thanks for your advices, in our environment, we observe that the pod has actually been deleted, but this goroutine will still block here.

Interesting, so a `kubectl exec` is sent to an already terminated Pod and the k8s API Server does not return?
Or the signal via `exec` works but the request does not complete for some reason?

docs/environment-variables.md
agilgur5: Follow-up [below](https://github.com/argoproj/argo-workflows/pull/13012#discussion_r1594592937)

> I'm happy with 10 minutes - although @agilgur5 also had [an opinion](https://github.com/argoproj/argo-workflows/pull/13012#discussion_r1592666100) on that.

But yes, if this is solely to prevent blocking, we can set this lower and don't need it to be configurable. If it were to be configurable it might make sense as a duration _after_ `terminationGracePeriodSeconds`.

I imagine 2min or lower would be safe as a default

fyp711: @agilgur5 @Joibel In my opinion, i think the best is 2min too. I think it needs to be configurable, it should add some flexibility.

docs/environment-variables.md
fyp711: Thanks for helping me revise it.

fyp711:
I updated the code, thanks for help me.

fyp711:
All CI's passed, very successful

workflow/signal/signal.go
```diff
@@ -49,6 +52,9 @@ func ExecPodContainerAndGetOutput(ctx context.Context, restConfig *rest.Config,
 	if err != nil {
 		return err
 	}
+	var executorSignalTimeout = env.LookupEnvDurationOr("EXECUTOR_SIGNAL_TIMEOUT", 2*time.Minute)
```
Joibel, line 55: nit: I'd write this as `executorSignalTimeout := env.LookupEnvDurationOr("EXECUTOR_SIGNAL_TIMEOUT", 2*time.Minute)`, it's more idiomatic golang to do it that way.

Joibel (changes requested):
Thanks, looking good to me, just a minor coding style thing.

I'll let @agilgur5 and you decide on whether this should be configurable. I'm happy with it with the 2minute configurable timeout but don't have strong opinions either way.

docs/environment-variables.md
Joibel: Thanks for fixing this.

workflow/signal/signal.go
fyp711: ok

fyp711: fixed

Joibel (approved):

workflow/signal/signal.go
Joibel: Thanks. Thank you for your contribution and thank you for your patience with our requests.

fyp711: My pleasure. I want the code to be perfect. Thanks.

fyp711: All CI's passed

workflow/signal/signal.go
agilgur5: @fyp711 you didn't answer the above; do you know why it works that way?

docs/environment-variables.md
agilgur5: > Because it is often difficult to update code in production environments, it is relatively easy to update configuration

For sure, but you wrote [above](https://github.com/argoproj/argo-workflows/pull/13012#discussion_r1592743019) that this should never be hit, so in that case, it shouldn't need configuring.

workflow/signal/signal.go
```diff
@@ -49,6 +50,8 @@ func ExecPodContainerAndGetOutput(ctx context.Context, restConfig *rest.Config,
 	if err != nil {
 		return err
 	}
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
```
fyp711, line 54: @agilgur5 I modified it in this way. Thanks.

fyp711: I thought about it and it really doesn't need to be configurable.

fyp711: >  you didn't answer the above; do you know why it works that way?

hi @agilgur5 , Colud i ask which question are you talking about ? That message is too lang, i don't understand which one

agilgur5: [This one](https://github.com/argoproj/argo-workflows/pull/13012#discussion_r1594598882)

workflow/signal/signal.go
fyp711: > Interesting, so a `kubectl exec` is sent to an already terminated Pod and the k8s API Server does not return? Or the signal via `exec` works but the request does not complete for some reason?
@agilgur5 I think `exec` has been executed successfully, because that pod is already deleted. I think the reason is apiserver does not return. But actually, I don't know why apiserver didn't return, I tried to analyze it, but it was very difficult. Further k8s community related submissions are this link below. You can have a look. Thank you.
https://github.com/kubernetes/kubernetes/pull/103177

agilgur5: Ok, gotcha. Alan and I were trying to understand _why_ this was happening here and in https://github.com/argoproj/argo-workflows/issues/13011#issuecomment-2100244803.

I read the k8s PR, however it similarly doesn't explain _why_. It seems like possibly a k8s API Server bug or `client-go` bug and there should really be an upstream issue for this that we can point to while we use this workaround.

In the interim, I'll leave a comment here noting that it's a workaround and referencing the PR

workflow/signal/signal.go
```diff
@@ -49,6 +50,8 @@ func ExecPodContainerAndGetOutput(ctx context.Context, restConfig *rest.Config,
 	if err != nil {
 		return err
 	}
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
```
agilgur5, line 53:
```suggestion
    // workaround for when exec does not properly return: https://github.com/kubernetes/kubernetes/pull/103177
	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
```

Follow-up to https://github.com/argoproj/argo-workflows/pull/13012#discussion_r1602409330

agilgur5 (approved):
LGTM. Thanks for reporting this odd bug with k8s and providing the workaround

agilgur5:
I wasn't able to backport this into v3.5.x as it has a merge conflict b/c it depends on #12858, which itself has a merge conflict b/c it depends on #12847 (which itself depends on other things that weren't backported to v3.5.x)

fyp711:
> I wasn't able to backport this into v3.5.x as it has a merge conflict b/c it depends on #12858, which itself has a merge conflict b/c it depends on #12847 (which itself depends on other things that weren't backported to v3.5.x)

Why the v3.5.x branch is different from main branch.

fyp711:
> > I wasn't able to backport this into v3.5.x as it has a merge conflict b/c it depends on #12858, which itself has a merge conflict b/c it depends on #12847 (which itself depends on other things that weren't backported to v3.5.x)
> 
> Why the v3.5.x branch is different from main branch.

Oh, i see you want to add https://github.com/argoproj/argo-workflows/pull/12847 into v3.6.0 Release.

agilgur5:
Correct, the `main` branch tracks _development_, which includes new features and breaking changes of any subsequent minor, and #12847 is a breaking change (as it may not work on older k8s)

Issue #13015, opened 2024-05-07
executor plugins example from documentation does not work


dgolja:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

[Executor_plugins](https://argo-workflows.readthedocs.io/en/stable/executor_plugins/) example provided in the documentation does not work even after adjusting the default service account permissions.

To setup the initial environment I followed the [quick guide](https://argo-workflows.readthedocs.io/en/stable/quick-start/) steps and updated `default` service account  RBAC permissions.

```bash
kubectl -n argo apply -f - <<EOF
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: agent
rules:
- apiGroups:
  - argoproj.io
  resources:
  - workflowtasksets
  verbs:
  - list
  - watch
- apiGroups:
  - argoproj.io
  resources:
  - workflowtasksets/status
  verbs:
  - patch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: argo-binding-agent
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: agent
subjects:
- kind: ServiceAccount
  name: default
  namespace: argo
---
apiVersion: v1
kind: Secret
metadata:
  name: default.service-account-token
  annotations:
    kubernetes.io/service-account.name: default
type: kubernetes.io/service-account-token
EOF
```

Also it's odd that the logs are complaining about `system:serviceaccount:argo:argo` permissions, even If I am not setting the service account to `argo` in the Workflow. Based on the [documentation](https://argo-workflows.readthedocs.io/en/stable/workflow-rbac/) it should use `default`.

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: hello-executor-plugin-example-
spec:
  entrypoint: main
  templates:
[...]

PR #7688, opened 2022-01-28
docs: Improve plugin template docs. Fixes #7683

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>
Fixes #7683

Don't bother creating a PR until you've done this:

* [ ] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it
  does not need to pass.
* Once required tests have passed, you can make it "Ready for review".
* Say how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* If changes were requested, and you've made them, then dismiss the review to get it looked at again.
* Add you organization to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md) if you like.
* You can ask for help!

PR #13019, opened 2024-05-08
fix(doc): Update executor plugins example

dgolja:
Fixes https://github.com/argoproj/argo-workflows/issues/13015

### Motivation

Having a working example from the documentation.

### Modifications

The main practical modification involves updating the user ID. Other changes, especially the RBAC role is to provide some guidance what permissions you need for the example to work ([see](https://github.com/argoproj/argo-workflows/blob/main/manifests/quick-start-minimal.yaml#L952-L974)). 

### Verification

Example provided in the documentation now works as expected.

### DIFF
--- a/docs/executor_plugins.md
+++ b/docs/executor_plugins.md
@@ -30,6 +30,15 @@ controller:
       value: "true"
 ```
 
+In case you want to modify a running `workflow-controller`:
+
+```bash
+kubectl patch deployment \
+  workflow-controller \
+  --namespace argo \
+  -p '{"spec":{"template":{"spec":{"containers":[{"name":"workflow-controller","env":[{"name":"ARGO_EXECUTOR_PLUGINS","value":"true"}]}]}}}}'
+```
+
 ## Template Executor
 
 This is a plugin that runs custom "plugin" templates, e.g. for non-pod tasks such as Tekton builds, Spark jobs, sending
@@ -162,7 +171,7 @@ spec:
         - containerPort: 4355
       securityContext:
         runAsNonRoot: true
-        runAsUser: 65534 # nobody
+        runAsUser: 1000
       resources:
         requests:
           memory: "64Mi"
@@ -201,6 +210,30 @@ spec:
 ```
 
 You'll see the workflow complete successfully.
+
+**Note**: The service account running the workflow needs at least the following permissions. If <= v3.2 you must replace `workflowtasksets/status` with `workflowtasksets`.
+
+```yaml
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: agent
+rules:
+- apiGroups:
+  - argoproj.io
+  resources:
+  - workflowtasksets
+  verbs:
+  - list
+  - watch
+- apiGroups:
+  - argoproj.io
+  resources:
+  - workflowtasksets/status
+  verbs:
+  - patch
+```
 
 ### Discovery
 


docs/executor_plugins.md
```diff
@@ -162,7 +171,7 @@ spec:
         - containerPort: 4355
       securityContext:
         runAsNonRoot: true
-        runAsUser: 65534 # nobody
+        runAsUser: 1000
```
agilgur5, line 174: can you elaborate why this is necessary? I'm not sure if this is necessary for all users, or just your specific environment

docs/executor_plugins.md
```diff
@@ -202,6 +211,30 @@ spec:
 
 You'll see the workflow complete successfully.
 
+**Note**: The service account running the workflow needs at least the following permissions. If <= v3.2 you must replace `workflowtasksets/status` with `workflowtasksets`.
```
agilgur5, line 214: I think this block should go in the `## Configuration` section above a subsection `### Permissions`. It's more important than a "Note" -- this is required

agilgur5: The Argo "agent" is mentioned in the [HTTP template docs](https://argo-workflows.readthedocs.io/en/release-3.4/http-template/#argo-agent) as well. Ideally we'd unify them and have a page on the Agent that both link to with regard to permissions / more information. The Agent is poorly documented; I didn't even know much about it until a few months ago

Potentially a section in the existing [Workflow RBAC page](https://argo-workflows.readthedocs.io/en/release-3.4/workflow-rbac/) might make sense, which is what the [quick-start manifest already links to](https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/manifests/quick-start/base/agent-role.yaml#L1)

agilgur5: I would leave the "If <= v3.2" out of this and leave it in the Workflow description annotation like [the quick start manifest](https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/manifests/quick-start/base/agent-role.yaml). Let's keep the two identical.

Even better would be to make that an in-line comment on the `workflowtasksets/status` line in both

docs/executor_plugins.md
```diff
@@ -30,6 +30,15 @@ controller:
       value: "true"
 ```
 
+In case you want to modify a running `workflow-controller`:
```
agilgur5, line 33:
```suggestion
If you want to modify a running Controller:
```

- simplify, "in case" -> "if", per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- "controller" is how it is referred to in this doc and elsewhere, be consistent

agilgur5 (commented):
A small copy-edit below and also a suggestion to unify Agent permissions docs

docs/executor_plugins.md
dgolja: Not sure if it's a local environmental problem, but leaving that to `65534` the examples never works. I provided more info in the original issue. I'll be happy to be proven wrong.

I use OS X (13.6 (22G120)) and Docker Desktop 4.14.1.

agilgur5: Yes you mentioned this in [the issue](https://github.com/argoproj/argo-workflows/issues/13015#issuecomment-2098378193) but you didn't say _why_ there either. That's why I asked here

agilgur5: This doesn't explain why it needs `runAsUser: 1000`, but apparently the tests do exactly that: https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/test/e2e/manifests/plugins/hello-executor-plugin-configmap.yaml#L71 ... not sure why the docs were different 🤔

agilgur5: `git blame` shows it was actually added in #7855, not during the original feature #7256

agilgur5: I see it was added in [the docs too](https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/docs/executor_plugins.md?plain=1#L250) in #7688.

I couldn't [find a place in the code](https://github.com/search?q=repo%3Aargoproj%2Fargo-workflows%20runAsUser&type=code) where this was actually enforced though.
The only relevant part seemed to be that the [default Executor `runAsUser` is `1000`](https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/docs/workflow-controller-configmap.yaml#L189) (partly from #8062).

dgolja: Correct you can not enforce the `runAsUser` value for the `ExecutorPlugin` and you can customise that. I thought that was the whole point of documenting that, comparing to the other init/main agent where the `runAsUser` is implicitly defined. For this particular docker image used in the example that UID does not work, so if someone will copy/paste the example and try to run it would fail. That's the reason I wanted to address it.

docs/executor_plugins.md
dgolja: Let me know what you think about the latest change. I added the `### Permissions` section with some relevant links and removed the `Role` definition. Not every user may experience permission issues, because it depends on how they installed WF. I think this new section letting them know they may need to adjust the permissions are adequate for now.

I agree that the documentation for the agent and permissions, in general, could be better. Let me think about it a bit more, and for now, I'll provide this as a stopgap. I may create a follow-up PR to improve the permissions documentation, but first, I need to dig more into the code to fully understand it.

docs/executor_plugins.md
agilgur5: > I thought that was the whole point of documenting that, comparing to the other init/main agent where the `runAsUser` is implicitly defined.

Honestly I wasn't sure myself 😅 Was trying to hunt that down as well as why it was failing



> For this particular docker image used in the example that UID does not work, so if someone will copy/paste the example and try to run it would fail. That's the reason I wanted to address it.

Ah so it's this particular image that fails with that UID. I was thinking it was a deeper problem 😅 . Thanks for clarifying!

docs/executor_plugins.md
agilgur5: > I think this new section letting them know they may need to adjust the permissions are adequate for now.

Yea we'd still leave this section or something similar if we move the actual Agent docs to its own page; this section would just link to that, something like "Plugins use the [Argo Agent](workflow-rbac.md#agent), ensure you have appropriate RBAC for your Workflow".



> Let me know what you think about the latest change.

It looks like the current text is mostly just a copy of the HTTP Template doc? That section causes confusion as-is the way it is written so if we're going to add it, I'd prefer to unify in the Workflow RBAC page. We can just have a short description there and show the RBAC. Can improve that in later PRs

I can help guide that if you're ok relying on my knowledge.

dgolja:
This is outdated so closing it.

PR #13023, opened 2024-05-08
fix: don't rebuild `ui/dist/app/index.html` in `argocli-build` stage

Joibel:
The timestamps in CI are incorrect, and may result in an attempt to rebuild this file when we don't have yarn installed. So don't.

This is a problem introduced in #13018. _edit by agilgur5:_ See also [Slack thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1715172548595289?thread_ts=1715163699.813329&cid=C0510EUH90V) for some debugging details on this

The timestamps don't always line up like expected when we come to check if we need to rebuild ui/dist/app/index.heml in the argocli-build step so force them fixed.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,8 @@ ARG GIT_TREE_STATE
 RUN mkdir -p ui/dist
 COPY --from=argo-ui ui/dist/app ui/dist/app
 
+# Ensure in github CI that we don't attempt to rebuild this file
+RUN touch ui/dist/app/index.html
 RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build STATIC_FILES=true make dist/argo GIT_COMMIT=${GIT_COMMIT} GIT_TAG=${GIT_TAG} GIT_TREE_STATE=${GIT_TREE_STATE}
 
 ####################################################################################################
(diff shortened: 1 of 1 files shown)


Joibel:
This should have run e2e tests really, but it didn't.

Dockerfile
```diff
@@ -72,6 +72,8 @@ ARG GIT_TREE_STATE
 RUN mkdir -p ui/dist
 COPY --from=argo-ui ui/dist/app ui/dist/app
 
+# Ensure in github CI that we don't attempt to rebuild this file
+RUN touch ui/dist/app/index.html
```
agilgur5, line 76:
```suggestion
# update timestamp so that `make` doesn't try to rebuild this -- it was already built in the previous stage
RUN touch ui/dist/app/index.html

```
- move this up next to the `argo-ui` `COPY` statement
- be more specific in the comment

Joibel:
> > This should have run e2e tests really, but it didn't.
> 
> We may want to add `Dockerfile` to [`common`](https://github.com/argoproj/argo-workflows/blob/2f15709952075ba208cf45ec459e66a14fa37d02/.github/workflows/ci-build.yaml#L39) in the `changed-files` list

See #13024

agilgur5:
Would like to rebase this on top of #13024 to make sure E2Es run and pass here

agilgur5 (approved):
thanks for debugging and fixing this Alan!

Joibel:
Thank you for testing and getting them through Anton.

PR #13024, opened 2024-05-08
ci: run E2E tests if `Dockerfile` changes

Joibel:
When the Dockerfile changes we should re-test most of the world

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -42,6 +42,7 @@ jobs:
               - tasks.yaml
             tests: &tests
               - *common
+              - Dockerfile
               - cmd/**
               - config/**
               - errors/**
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -42,7 +42,6 @@ jobs:
               - tasks.yaml
             tests: &tests
               - *common
-              - Dockerfile
               - cmd/**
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
               - tasks.yaml
             tests: &tests
               - *common
+              - Dockerfile
```
agilgur5, line 45: if we're not adding it to `common`, I think it would make more sense in `e2e-tests`, since those require the image builds. `tests` are units that just run Go tests directly, no Docker involved

agilgur5: Yea looks like `e2e-tests` are the only ones that use the images (well and the image builds, but those currently run unconditionally, per #12006 "Future Work")

agilgur5: updated this [myself](https://github.com/argoproj/argo-workflows/pull/13024/commits/c0d0e4e015a3bf8ad2eac6337ef0b9f0ea4a4a60) in order to get CI passing sooner rather than later, since Alan's left for the day in his timezone

agilgur5 (approved):
LGTM, per above, `Dockerfile` is used to build images that are used in the E2E tests.

We may want to further add files that change the `docker build` context, same as #12006 "Future Work"

Issue #12455, opened 2024-01-04
Upgrade actions/upload-artifact and actions/download-artifact to V4


Garett-MacGowan:
# Summary

Once https://github.com/actions/download-artifact/issues/249 is resolved, we can upgrade actions/[upload/download]-artifact to V4. The project is currently using V3 since downgrading from V4 due to the issue above.

Previously upgraded to V4 in https://github.com/argoproj/argo-workflows/pull/12384
Downgraded to V3 in https://github.com/argoproj/argo-workflows/pull/12454

## Use Cases

V4 has some performance improvments which would be nice to integrate, but only once it is stable.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

Garett-MacGowan:
@agilgur5 I've created this ticket to keep track of the eventua desire to upgrade after https://github.com/argoproj/argo-workflows/pull/12454 is merged.

Garett-MacGowan:
https://github.com/actions/download-artifact/issues/249 has been closed but users are still reporting [issues](https://github.com/actions/download-artifact/issues/249#issuecomment-1955241332).

Garett-MacGowan:
@agilgur5 Based on this[](https://github.com/actions/download-artifact/issues/249#issuecomment-2040750798), it may be safe to upgrade to V4 now.

agilgur5:
🤔 The previous two comments have a ton of upvotes, so I'm not so sure about that...

Garett-MacGowan:
> 🤔 The previous two comments have a ton of upvotes, so I'm not so sure about that...

I'm definitely not sure, but based on the example failed action, the failure mode is different.

Garett-MacGowan:
@agilgur5 You're probably right. Not stable enough to warrant the upgrade. Lets watch [this](https://github.com/actions/download-artifact/issues/294)

Joibel:
This has failed for me on my fork - https://github.com/Joibel/argo-workflows/actions/runs/12695097633
We have to fix this before the 30th of Jan https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ although the individual repos say this was deprecated on the 30th of November 2024 https://github.com/actions/upload-artifact/blob/65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08/README.md?plain=1#L4

I'll put together an upgrade for this

Joibel:
[...]

PR #13027, opened 2024-05-09
ci: optimise image building

Joibel:
Build the argoexec and argocli images in a matrix github action.

Download and install the images into k3d in one action each.

This commit will require changes to the Required steps in github which I cannot do myself. I'm not sure why that one `argoexec-image` is listed as required at all, it's a dependency of testing rather than a proper part of the test or CI.

Also updates upload and download artifact actions to latest v4, which seem to be working OK now. v3 will deprecate by November. *edit by agilgur5*: this was to support the `merge-multiple` param per [below](https://github.com/argoproj/argo-workflows/pull/13027#issuecomment-2117549171)


.github/workflows/ci-build.yaml
```diff
@@ -151,57 +151,39 @@ jobs:
...
+          target: ${{matrix.image}}
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Upload
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 #v4.3.3
         with:
-          name: argocli
-          path: /tmp/argocli_image.tar
+          name: ${{matrix.image}}_image.tar
```
agilgur5, line 180:
```suggestion
          name: ${{matrix.image}}
```

From the previous code, this should just be the name of the image. Not sure if this was a copy+paste typo?

agilgur5:
Follow-up to #13018 / https://github.com/argoproj/argo-workflows/pull/13018#discussion_r1594188099

> This commit will require changes to the Required steps in github which I cannot do myself.

Neither can I. Only @terrytangyuan and @sarabala1979 can change the required steps in the branch protection rules.

> I'm not sure why that one `argoexec-image` is listed as required at all, it's a dependency of testing rather than a proper part of the test or CI.

All of them were required prior to #12006. Technically speaking, the image build could have a different set of changed files than E2Es although it is currently unconditional. Making sure it passes is also safer just in case -- PRs shouldn't break image builds.

> Also updates upload and download artifact actions to latest v4, which seem to be working OK now. v3 will deprecate by November.

We have a follow-up issue for this #12455 where there were some lingering problems last we checked. There's also one more place that uses `upload-artifact` v3, the [docs build](https://github.com/argoproj/argo-workflows/blob/e3b0bb6b289eafcbe5ecea9c1e3b920ca2bc31a8/.github/workflows/docs.yaml#L40). And if working we should remove the [dependabot ignore](https://github.com/argoproj/argo-workflows/blob/e3b0bb6b289eafcbe5ecea9c1e3b920ca2bc31a8/.github/dependabot.yml#L68) as well.
So I'd probably suggest splitting that into a separate PR more generally

.github/workflows/ci-build.yaml
Joibel: It's deliberate. We're pattern matching this with `pattern: '*_image.tar'` on line 164. This pattern would be `'*'` without the extra parts to the name.

Joibel:
I've reverted upload+download artifact actions again.

Joibel:
> I've reverted upload+download artifact actions again.

`merge-multiple` doesn't work in v3. The docs are not versioned, so it's hard to discover how stuff should work. So that's part of why I'd updated, I remember now.

Also, I'm talking to myself again.

.github/workflows/ci-build.yaml
agilgur5: Yes, I thought that would suffice, no? Or were you trying to be extra careful in case at some point other artifacts were added or something?

terrytangyuan (approved):

terrytangyuan:
Merged

.github/workflows/ci-build.yaml
```diff
@@ -276,20 +258,16 @@ jobs:
...
-          name: argoexec
-          path: /tmp
-      - name: Load argoexec image
-        run: docker load < /tmp/argoexec_image.tar
-      - name: Download argocli image
+      - name: Download images
         uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
         with:
-          name: argocli
+          pattern: '*_image.tar'
```
agilgur5, line 264: > It seems like ["download all" is available](https://github.com/actions/download-artifact/tree/v3.0.2?tab=readme-ov-file#download-all-artifacts) if you don't specify `name` and use a directory.

Annnd that appears to be what we're using here since `pattern` is invalid in `actions/download-artifact@v3`.

This is getting a warning [in CI](https://github.com/argoproj/argo-workflows/actions/runs/9249013492) for every E2E test in the matrix:
```
[E2E Tests (test-executor, minimal)](https://github.com/argoproj/argo-workflows/actions/runs/9249013492/job/25442227788#step:8:1)
Unexpected input(s) 'pattern', valid inputs are ['name', 'path']
```
Screenshot 2024-05-27 at 1 44 33 AM

PR #13039, opened 2024-05-12
fix(ui): Use proper podname for containersets. Fixes #13038

### LINKED ISSUE #13038: UI: Containerset nodes receive the wrong podname
### Pre-requisites

When trying to get the logs for a (any) containerset container in the UI, the user is faced with a blank page. 
image
This is due to the podname being incorrectly calculated for these nodes, which makes the request to the backend return an empty response.

Expected behavior would be to be able to view the logs in similar fashion as for pod nodes.

### Version

latest

```YAML
kind: Workflow
metadata:
  generateName: graph-
  labels:
  annotations:
      This workflow demonstrates running a graph of tasks within containers in a single pod.
    workflows.argoproj.io/version: ">= 3.1.0"
spec:
  entrypoint: main
  templates:
    - name: main
      containerSet:
        containers:
          - name: a
            image: argoproj/argosay:v2
          - name: b
            image: argoproj/argosay:v2
            dependencies: ["a"]
          - name: c
            image: argoproj/argosay:v2
            dependencies: ["a"]
          - name: d
            image: argoproj/argosay:v2
            dependencies: ["b", "c"]
```

```text
N/A
```

```text
N/A
```

instauro:
Fixes #13038 
### Motivation

Containerset nodes currently do not retrieve their logs in the UI. This is due to the podname calculation failing, which makes the request to the backend incorrect. This fix will hopefully become obsolete when https://github.com/argoproj/argo-workflows/issues/12528 is fixed, but until then this should do the trick.

### Modifications

Uses a better podname when the node type is `Container`.
image
image

### Verification

Tried a few different containerset templates to verify the fix, and a few none-containetset templates to verify that there is no regression.

### DIFF
--- a/ui/src/app/shared/pod-name.ts
+++ b/ui/src/app/shared/pod-name.ts
@@ -11,7 +11,7 @@ const maxPrefixLength = maxK8sResourceNameLength - k8sNamingHashLength;
 // getPodName returns a deterministic pod name
 // In case templateName is not defined or that version is explicitly set to  POD_NAME_V1, it will return the nodeID (v1)
 // In other cases it will return a combination of workflow name, template name, and a hash (v2)
-// note: this is intended to be equivalent to the server-side Go code in workflow/util/pod_name.go
+// note: this is intended to be equivalent to the server-side Go code in workflow/util/pod_name.go#GeneratePodName
 export function getPodName(workflow: Workflow, node: NodeStatus): string {
     const version = workflow.metadata?.annotations?.[ANNOTATION_KEY_POD_NAME_VERSION];
     if (version === POD_NAME_V1) {
@@ -19,7 +19,10 @@ export function getPodName(workflow: Workflow, node: NodeStatus): string {
     }
 
     const workflowName = workflow.metadata.name;
-    if (workflowName === node.name) {
+    // convert containerSet node name to its corresponding pod node name by removing the ".<containerName>" postfix
+    // this part is from workflow/controller/container_set_template.go#executeContainerSet; the inverse never happens in the back-end, so is unique to the front-end
+    const podNodeName = node.type == 'Container' ? node.name.replace(/\.[^/.]+$/, '') : node.name;
+    if (workflowName === podNodeName) {
         return workflowName;
     }
 
@@ -30,7 +33,7 @@ export function getPodName(workflow: Workflow, node: NodeStatus): string {
     }
     prefix = ensurePodNamePrefixLength(prefix);
 
-    const hash = createFNVHash(node.name);
+    const hash = createFNVHash(podNodeName);
     return `${prefix}-${hash}`;
 }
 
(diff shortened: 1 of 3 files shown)


ui/src/app/shared/pod-name.ts
```diff
@@ -19,7 +19,8 @@ export function getPodName(workflow: Workflow, node: NodeStatus): string {
     }
 
     const workflowName = workflow.metadata.name;
-    if (workflowName === node.name) {
+    const podNodeName = node.type == 'Container' ? node.name.replace(/\.[^/.]+$/, '') : node.name;
```
agilgur5, line 22: hmm this doesn't seem to match [the back-end](https://github.com/argoproj/argo-workflows/blob/0c1398ea6fa66cc77156f15a6d69c260c6b524da/workflow/util/pod_name.go#L51). these two functions should be equivalent per the comment on line 14 in this file

instauro: Sure, reverted the change and moved it outside the function call. Maybe there are still better ways to do this

agilgur5: It would be great if you could locate where in the back-end this gets set, that way we can keep them equivalent, whether in this function or elsewhere.

instauro: Looks like the pod is created based on the pod node name.
https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/container_set_template.go#L20-L24

Then the containerset nodes are created
https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/container_set_template.go#L31-L46
Where the container node names are just their pod node name concatenated with the container name.
```go
ctxNodeName := fmt.Sprintf("%s.%s", nodeName, c.Name)
```
So removing this postfix should map from containerset node name to pod node name, which can be used to calculate the pod name using the existing function.

agilgur5: Yea it looks like ContainerSet node -> Pod node name never happens directly in the back-end, which explains why they're not equivalent.

In that case, I'd prefer your earlier variant, since I intentionally consolidated `getPodName` in #12964.
Let's just leave a comment on why this block differs from the back-end then for explicitness

agilgur5: Reverted it back to the old variant and added some comments in https://github.com/argoproj/argo-workflows/pull/13039/commits/39c44cb3e7c78a43fa47413f0f94a910dfa7d935

GH PRs still don't record _all_ versions of amended commits (unlike Gerrit which has had this for at least a decade if not more 😕), but I remembered the new GH activity feed existed and was able to [find it there](https://github.com/instauro/argo-workflows/activity?ref=fix/containerset-podname).

agilgur5 (approved):
Made the revert myself plus added comments per https://github.com/argoproj/argo-workflows/pull/13039#discussion_r1650211514, so LGTM now

Thanks for root causing and fixing this!

PR #13046, opened 2024-05-13
refactor(ui): optimize state in logs viewer 

agilgur5:
### Motivation

simplify & optimize some of the state in the `WorkflowLogsViewer` component

### Modifications

- only retrieve values from `localStorage` once, on the initial state set
  - the callback only happens once, while previously it was retrieved on every render but only used on the first
  - also improve the typings a bit by using generics instead of a cast

- replace the secondary log filter state with a debounce
  - it was giving a 1000ms delay, which is exactly what you can do with a debounce
  - and we already have a debounce func in the utils, so just re-use that

### Verification

<details><summary>Screenshot of the filter still working: </summary>

Screenshot 2024-05-13 at 3 15 49 PM

</details>


isubasinghe (approved):

isubasinghe:
> only retrieve values from localStorage once, on the initial state set

Thanks for this, I was just about to ask about what was wrong with the previous. I haven't used react in quite a while.

PR #13047, opened 2024-05-14
fix(docs): Clarify quick start installation. Fixes #13032 

### LINKED ISSUE #13032: Docs bug: following the quick-start guide _too_ explicitly can get users in trouble
### Pre-requisites

Today I followed the quick start guide, and had trouble specifically with this part:

>To install Argo Workflows, go to the [releases page](https://github.com/argoproj/argo-workflows/releases/latest). Scroll down to the Controller and Server section and execute the kubectl commands.

I did this, which led to me applying `install.yaml`, not `quick-start-minimal.yaml`, which I think is what is intended for the quick-start guide. Therefore, when I did the hello-world step, I encountered:

```
pods "hello-world-xmsc9" is forbidden: User "system:serviceaccount:argo:default" cannot patch resource "pods" in API group "" in the namespace "argo"
```

This is because the hello-world step depends on resources defined in `quick-start-minimal.yaml`.

The example command immediately following this instruction does use the correct filename `quick-start-minimal.yaml`, but I ignored that example command for a few reasons:

* I already copied the command from the GitHub releases page as instructed in 1st paragraph
* Example contains a slug `v<<ARGO_WORKFLOWS_VERSION>>`, so it's not directly copyable.
* The difference between this command and the command in GitHub release page is hidden because the page isn't wide enough
* The text referred to it as an example, so I did not consider it an "instruction" to follow

image

#### Proposed resolution

Replace the language  "execute the kubectl commands" with "navigate to `quick-start-minimal.yaml`, right click, and select 'copy link' to get the correct YAML URL. Craft a `kubectl` command to apply that yaml following this example: ..."

Or:

Replace the language "execute the kubectl commands" with "to find example kubectl commands."

Add a mkdocs callout of type "danger": "Do not directly execute those commands, as we **do not** want to apply `install.yaml`. Instead, we want to use `quick-start-minimal.yaml`. Here's an example command: ..."

### Version

v3.5.6

```YAML
Follow quick start guide following instructions too explicitly
```

```text
n/a
```

```text
n/a
```

mfisher87:
I got hung up on these instructions because I copied the commands from the releases page instead of from the quick start page, so I applied `install.yaml` instead of `quick-start-minimal.yaml`. I hope this change will help others not get confused like me :)

I tried `make pre-commit -B` (had to install `go` first) and it failed with `protoc` not found. Do I need to keep working my way through these, or is it OK to ignore for a docs change like this?

Fixes #13032

### Motivation

I followed the quick start instructions explicitly, but ended up with an install that couldn't execute Hello World step.

### Modifications

Alter docs to explicitly instruct the user to install the quick start manifest.

I also moved a warning admonition because I felt it belonged after prose I added in this PR.

### Verification

Rendered the docs, looked good

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -20,7 +20,7 @@ For quick testing, you can use a local cluster with:
 
 ## Install Argo Workflows
 
-First, specify the version you'd like to install in an environment variable.
+First, specify the version you want to install in an environment variable.
 Modify the command below:
 
 ```bash
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -14,20 +14,27 @@ For quick testing, you can use a local cluster with:
 * [k3s](https://k3s.io/) or [k3d](https://k3d.io/)
 * [Docker Desktop](https://www.docker.com/products/docker-desktop/)
 
+## Install Argo Workflows
+
+To install Argo Workflows for development or evaluation, use the
+`quick-start-minimal.yaml` manifest provided with each release.
+
 !!! Warning "Development vs. Production"
     These instructions are intended to help you get started quickly. They are not suitable for production.
     For production installs, please refer to [the installation documentation](installation.md).
 
-## Install Argo Workflows
+First, specify the version you'd like to install in an environment variable. Modify the
+command below:
 
-To install Argo Workflows, go to the [releases page](https://github.com/argoproj/argo-workflows/releases/latest).
-Scroll down to the `Controller and Server` section and execute the `kubectl` commands.
+```bash
+ARGO_WORKFLOWS_VERSION="vX.Y.Z"
+```
 
-Below is an example of the install commands (substitute `<<ARGO_WORKFLOWS_VERSION>>` with the version number):
+Next, copy the command below to apply the quick-start manifest:
 
 ```bash
 kubectl create namespace argo
-kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v<<ARGO_WORKFLOWS_VERSION>>/quick-start-minimal.yaml
+kubectl apply -n argo -f "https://github.com/argoproj/argo-workflows/releases/download/${ARGO_WORKFLOWS_VERSION}/quick-start-minimal.yaml"
 ```
 
 ## Install the Argo Workflows CLI
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -20,8 +20,8 @@ For quick testing, you can use a local cluster with:
 
 ## Install Argo Workflows
 
-First, specify the version you'd like to install in an environment variable. Modify the
-command below:
+First, specify the version you'd like to install in an environment variable.
+Modify the command below:
 
 ```bash
 ARGO_WORKFLOWS_VERSION="vX.Y.Z"
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -31,7 +31,7 @@ command below:
 ARGO_WORKFLOWS_VERSION="vX.Y.Z"
 ```
 
-Next, copy the command below to apply the quick-start manifest:
+Then, copy the commands below to apply the quick-start manifest:
 
 ```bash
 kubectl create namespace argo
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -14,10 +14,6 @@ For quick testing, you can use a local cluster with:
 * [k3s](https://k3s.io/) or [k3d](https://k3d.io/)
 * [Docker Desktop](https://www.docker.com/products/docker-desktop/)
 
-
-To install Argo Workflows for development or evaluation, use the
-`quick-start-minimal.yaml` manifest provided with each release.
-
 !!! Warning "Development vs. Production"
     These instructions are intended to help you get started quickly. They are not suitable for production.
     For production installs, please refer to [the installation documentation](installation.md).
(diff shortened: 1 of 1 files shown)


mfisher87:
Apologies, knew I should have opened as a draft but clicked too hastily and missed the little arrow on the button.

docs/quick-start.md
```diff
@@ -14,20 +14,27 @@ For quick testing, you can use a local cluster with:
...
     For production installs, please refer to [the installation documentation](installation.md).
 
-## Install Argo Workflows
+First, specify the version you'd like to install in an environment variable. Modify the
+command below:
 
-To install Argo Workflows, go to the [releases page](https://github.com/argoproj/argo-workflows/releases/latest).
-Scroll down to the `Controller and Server` section and execute the `kubectl` commands.
+```bash
+ARGO_WORKFLOWS_VERSION="vX.Y.Z"
```
mfisher87, line 30: I was really hoping there was a way for Mkdocs to use the version number it "knows" about (from the upper-right corner), but failed to. I'm a newer Mkdocs user so it could have just been not knowing the right thing to search for, but I've done this before in Sphinx and couldn't figure it out in the time I had :upside_down_face:

agilgur5:
> I tried `make pre-commit -B` (had to install `go` first) and it failed with `protoc` not found. Do I need to keep working my way through these, or is it OK to ignore for a docs change like this?

It's not necessary for most of the docs, only the codegen'd ones. CI passes so you're good.
Also, the `Makefile` [defines `protoc`](https://github.com/argoproj/argo-workflows/blob/cd52436b238980e64f12f65bbd097fe87d3c98b7/Makefile#L140) 🤔

docs/quick-start.md
agilgur5: I'm pretty sure this requires a different plugin; `mkdocs-macros` [can do it](https://mkdocs-macros-plugin.readthedocs.io/en/latest/git_info/#tag-and-short_tag).

mfisher87:
Great, thanks for confirming! Happy to make any changes or rebase or whatever :)

mfisher87:
Thanks for your time, Anton :heart: 

Just want to say I'm really impressed with the community y'all have built here :star_struck:

docs/quick-start.md
```diff
@@ -14,20 +14,27 @@ For quick testing, you can use a local cluster with:
...
+```
 
-Below is an example of the install commands (substitute `<<ARGO_WORKFLOWS_VERSION>>` with the version number):
+Next, copy the command below to apply the quick-start manifest:
```
agilgur5, line 33:
```suggestion
Then, copy the commands below to apply the quick-start manifest:
```

- grammar: "command" -> "commands" as there are two commands (one to create the NS as well)
- style: "Next" -> "Then" -- "next" is often not used for a _final_ instruction. There is still more after this, but not in this section. "then" is used more ambiguously

docs/quick-start.md
```diff
@@ -14,20 +14,27 @@ For quick testing, you can use a local cluster with:
 * [k3s](https://k3s.io/) or [k3d](https://k3d.io/)
 * [Docker Desktop](https://www.docker.com/products/docker-desktop/)
 
+## Install Argo Workflows
+
+To install Argo Workflows for development or evaluation, use the
```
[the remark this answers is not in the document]
agilgur5: (also generally prefer at least 1 sentence per line of markdown, [per style guide](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/). it makes diffs easier. short follow-up sentences are usually ok, but prefer to one-line long sentences)

agilgur5 (commented):
A few small changes below

docs/quick-start.md
```diff
@@ -14,20 +14,27 @@ For quick testing, you can use a local cluster with:
 * [k3s](https://k3s.io/) or [k3d](https://k3d.io/)
 * [Docker Desktop](https://www.docker.com/products/docker-desktop/)
 
+## Install Argo Workflows
```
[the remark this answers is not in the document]
mfisher87: 81580bc9a

docs/quick-start.md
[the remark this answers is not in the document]
mfisher87: 39a95131e, b5abdac46

docs/quick-start.md
```diff
@@ -20,14 +20,18 @@ For quick testing, you can use a local cluster with:
 
 ## Install Argo Workflows
 
-To install Argo Workflows, go to the [releases page](https://github.com/argoproj/argo-workflows/releases/latest).
-Scroll down to the `Controller and Server` section and execute the `kubectl` commands.
+First, specify the version you'd like to install in an environment variable.
```
mfisher87, line 23: @agilgur5 Do you have strong feelings about use of contractions?

docs/quick-start.md
mfisher87: Oo thank you :)

docs/quick-start.md
agilgur5: not too strongly. some style guides do suggest against them. They can also be harder to understand for those without strong English (though in my personal experience, people tend to memorize what it means without knowing how the contraction works).

That being said, now that you pointed it out, usually contractions aren't using [present tense](https://kubernetes.io/docs/contribute/style/style-guide/#use-present-tense) and so [can be simplified](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language):

```suggestion
First, specify the version you want to install in an environment variable.
```

agilgur5 (approved):
LGTM. Thanks for pointing this out and fixing it!

PR #13064, opened 2024-05-18
feat: argo_template env variable without inputs

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/8790

### Motivation

reduce etcd growth

### Modifications

will likely need env variable and else block to maintain existing functionality for some users


tooptoop4:
what do u think @jswxstw ?

jswxstw:
> what do u think @jswxstw ?

I think this is a better way to decrease the size of pod envs: https://github.com/argoproj/argo-workflows/issues/8790#issuecomment-2381964013.

tooptoop4:
superseded by https://github.com/argoproj/argo-workflows/pull/13742

agilgur5:
You can force push over your old commits to fix DCO instead of opening a new PR (which creates a bit of pollution)

Issue #12993, opened 2024-04-29
Workflow processing fails to complete due to incomplete WorkflowTaskResult from interrupted pod


jesseanttila-cai:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

As part of #12402 (included from v3.5.3 onwards), workflow pod `wait`-container behavior was changed to create a placeholder (incomplete) WorkflowTaskResult before waiting for the `main`-container to complete. 

https://github.com/argoproj/argo-workflows/blob/0fdf74511d4671cf0c8c334aa2d90ecd61c5acce/cmd/argoexec/commands/wait.go#L38-L42

The WorkflowTaskResult is finalized after output artifacts, logs etc. have been handled: 

https://github.com/argoproj/argo-workflows/blob/0fdf74511d4671cf0c8c334aa2d90ecd61c5acce/cmd/argoexec/commands/wait.go#L34

If the `wait`-container is interrupted in a way that prevents `FinalizeOutput` from being called (e.g. pod deletion without sufficient grace period), an incomplete WorkflowTaskResult remains with the `workflows.argoproj.io/report-outputs-completed` label set to `false`. Retries of the same task will produce additional WorkflowTaskResults and will not mark the previous one complete. This leaves the workflow stuck in `Processing` state until the WorkflowTaskResult is manually edited to mark it complete.

The reproduction example workflow simulates forced pod deletion using a pod that deletes itself, leaving behind an incomplete WorkflowTaskResult. The included workflow controller log snippet shows the resulting processing loop.

This issue may be one of the causes of #12103.

### Version

v3.5.3

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: workflow-hang-example-
spec:
  entrypoint: self-delete
  templates:
    - name: self-delete
      retryStrategy:
        limit: "1"
        retryPolicy: Always
      podSpecPatch: |
[...]

PR #13066, opened 2024-05-19
fix: don't necessarily include all artifacts from templates in node outputs

juliev0:
The fact that it's being listed in the NodeStatus is causing ArtifactGC to attempt to delete it (and fail doing so).

Fixes: #12845 

### Motivation

If an Artifact file didn't get written, it shouldn't be in the outputs for the TaskResult (and thus the Outputs for the Node) just because it was listed as an Artifact in the Workflow.

### Modifications

Only if the Artifact file is successfully written does the wait container include it as one of the Output Artifacts in its `WorkflowTaskResult`. The Workflow Controller uses that `WorkflowTaskResult` information to populate the `NodeStatus` of the Workflow. This is the information that ArtifactGC uses to determine what needs to be deleted. 

### Verification

The person who submitted the Issue [tested](https://github.com/argoproj/argo-workflows/discussions/12845#discussioncomment-9616341) the image I made, and it repeatedly worked to solve his issue of ArtifactGC failing. (Note it also prevented his other artifacts from getting deleted, because the failed deletion preempted the other deletions from being attempted (as @agilgur5 [pointed out](https://github.com/argoproj/argo-workflows/discussions/12845#discussioncomment-9598353), this could be another improvement, to continue the deletion process independent of failure)).

I did start to create an e2e test for this issue. But it seems that with minio, I can't repeat the same failure using the "main" branch. It seems that attempting to delete an artifact which doesn't exist doesn't result in a failure for some reason.

### DIFF
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -148,7 +148,7 @@ type artifactDerivedKey struct {
 
 func (al *s3Location) getS3Key(wf *wfv1.Workflow) (string, error) {
 	if al.specifiedKey == "" && al.derivedKey == nil {
-		panic(fmt.Sprintf("invalid artifactLocation: %+v, must have knownKey or derivedKey set", al))
+		panic(fmt.Sprintf("invalid artifactLocation: %+v, must have specifiedKey or derivedKey set", al))
 	}
 
 	if al.specifiedKey != "" {
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -407,8 +407,8 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		when.
 			DeleteWorkflow().
 			WaitForWorkflowDeletion()
-
-		when.Then().ExpectWorkflowDeleted()
+			Then().
+			ExpectWorkflowDeleted()
 
 		when = when.RemoveFinalizers(false) // just in case - if the above test failed we need to forcibly remove the finalizer for Artifact GC
 
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -257,12 +257,14 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 	for _, tt := range []struct {
 		workflowFile                 string
 		hasGC                        bool
+		workflowShouldSucceed        bool
 		expectedArtifacts            []artifactState
 		expectedGCPodsOnWFCompletion int
 	}{
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-multi-strategy-multi-anno.yaml",
 			hasGC:                        true,
+			workflowShouldSucceed:        true,
 			expectedGCPodsOnWFCompletion: 2,
 			expectedArtifacts: []artifactState{
 				artifactState{"first-on-completion-1", "my-bucket-2", true, false},
@@ -276,6 +278,7 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-from-template.yaml",
 			hasGC:                        true,
+			workflowShouldSucceed:        true,
 			expectedGCPodsOnWFCompletion: 1,
 			expectedArtifacts: []artifactState{
 				artifactState{"on-completion", "my-bucket-2", true, false},
@@ -286,6 +289,7 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-from-template-2.yaml",
 			hasGC:                        true,
+			workflowShouldSucceed:        true,
 			expectedGCPodsOnWFCompletion: 1,
 			expectedArtifacts: []artifactState{
 				artifactState{"on-completion", "my-bucket-2", true, false},
@@ -296,6 +300,7 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-step-wf-tmpl.yaml",
 			hasGC:                        true,
+			workflowShouldSucceed:        true,
 			expectedGCPodsOnWFCompletion: 1,
 			expectedArtifacts: []artifactState{
 				artifactState{"on-completion", "my-bucket-2", true, false},
@@ -306,6 +311,7 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-step-wf-tmpl-2.yaml",
 			hasGC:                        true,
+			workflowShouldSucceed:        true,
 			expectedGCPodsOnWFCompletion: 1,
 			expectedArtifacts: []artifactState{
 				artifactState{"on-completion", "my-bucket-2", true, false},
@@ -316,6 +322,7 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		{
 			workflowFile:                 "@testdata/artifactgc/artgc-from-ref-template.yaml",
 			hasGC:                        true,
+			workflowShouldSucceed:        true,
 			expectedGCPodsOnWFCompletion: 1,
 			expectedArtifacts: []artifactState{
 				artifactState{"on-completion", "my-bucket-2", true, false},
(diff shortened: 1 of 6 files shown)


workflow/executor/executor.go
```diff
@@ -291,26 +291,27 @@ func (we *WorkflowExecutor) StageFiles() error {
...
+		return artifacts, argoerrs.InternalWrapError(err)
 	}
 
 	for i, art := range we.Template.Outputs.Artifacts {
 		err := we.saveArtifact(ctx, common.MainContainerName, &art)
 		if err != nil {
-			return err
+			return artifacts, err
 		}
-		we.Template.Outputs.Artifacts[i] = art
```
juliev0, line 311: instead of updating this directly, create a new list of the artifacts that we actually saved and return that

Garett-MacGowan: LGTM. So before we were using the we.Template.Outputs.Artifacts to pass to ReportOutputs but now we are creating a fresh `wfv1.Artifacts` object that will only include the successfully saved artifacts. This makes sense.

Garett-MacGowan (approved):
LGTM

agilgur5:
> (as @agilgur5 [pointed out](https://github.com/argoproj/argo-workflows/discussions/12845#discussioncomment-9598353), this could be another improvement, to continue the deletion process independent of failure)

I was thinking about this and it will probably be fixed by parallel artifact GC (#11768), as that inherently has to process each artifact independently.

test/e2e/artifacts_test.go
```diff
@@ -352,7 +368,7 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 				}
 			})
 
-		if when.WorkflowCondition(func(wf *wfv1.Workflow) bool {
+		if tt.workflowShouldSucceed && when.WorkflowCondition(func(wf *wfv1.Workflow) bool {
```
juliev0, line 371: was previously trying to avoid intermittent test failures due to the workflow itself failing when it shouldn't - but now there's a Workflow which fails in here so this is how I'm allowing that one to be failed

test/e2e/testdata/artifactgc/artgc-artifact-not-written.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: testing-artifact-not-present-
```
agilgur5, line 4:
```suggestion
  generateName: artgc-artifact-not-written-
```

these normally match the filename by convention, no?

test/e2e/testdata/artifactgc/artgc-artifact-not-written.yaml
```diff
@@ -0,0 +1,28 @@
...
+  artifactGC:
+    strategy: OnWorkflowDeletion
+  podGC:
+    strategy: ""
+  templates:
+    - name: entrypoint
+      steps:
+      - - name: on-deletion
+          template: on-deletion
+    - name: on-deletion
```
agilgur5, line 16:
```suggestion
      - - name: artifact-not-written
          template: artifact-not-written
    - name: artifact-not-written
```

clearer naming

test/e2e/testdata/artifactgc/artgc-artifact-not-written.yaml
```diff
@@ -0,0 +1,28 @@
...
+      container:
+        image: argoproj/argosay:v2
+        command:
+          - sh
+          - -c
+        args:
+          - |
+            echo "not writing anything"
```
agilgur5, line 24:
```suggestion
            echo "intentionally not writing anything to disk"
```
to be super explicit 😅

test/e2e/artifacts_test.go
```diff
@@ -392,6 +408,8 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 			DeleteWorkflow().
 			WaitForWorkflowDeletion()
 
+		when.Then().ExpectWorkflowDeleted()
+
```
agilgur5, line 412:
```suggestion
			Then().
			ExpectWorkflowDeleted()

```

can this be put into the chain?

agilgur5 (commented):
Mostly LGTM. I left a few style comments on the tests, one suggestion for the test case, and a question on the saving + reporting. Still learning my way around the artifacts part of the codebase

workflow/executor/executor.go
juliev0: > If the wait container is interrupted during this loop, some artifacts may save to S3 (etc) but not be reported

How do you mean it wouldn't be reported? Since the background context is passed into both [SaveArtifacts](https://github.com/argoproj/argo-workflows/blob/main/cmd/argoexec/commands/wait.go#L60)() and [ReportOutputs](https://github.com/argoproj/argo-workflows/blob/main/cmd/argoexec/commands/wait.go#L69)() below, I believe all of that logic is still supposed to be executed. Is there something different you're referring to besides the context being cancelled?

agilgur5: If the loop is long or the container is otherwise non-gracefully terminated, the request context won't matter.

Also realized that saving artifacts can be parallelized here to reduce that chance and improve throughput

test/e2e/artifacts_test.go
```diff
@@ -391,6 +407,8 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 		when.
 			DeleteWorkflow().
 			WaitForWorkflowDeletion()
```
agilgur5, line 409:
```suggestion
			WaitForWorkflowDeletion().
```

Lint is [failing](https://github.com/argoproj/argo-workflows/actions/runs/9455186352/job/26044675796?pr=13066) as the previous line needs a trailing dot now for the chain

test/e2e/artifacts_test.go
```diff
@@ -327,6 +334,15 @@ func (s *ArtifactsSuite) TestArtifactGC() {
...
+			expectedGCPodsOnWFCompletion: 0,
+			expectedArtifacts:            []artifactState{},
+		},
+		// Workflow fails to write an artifact that's been defined as an Output
+		{
+			workflowFile:                 "@testdata/artifactgc/artgc-artifact-not-written.yaml",
+			hasGC:                        true,
+			workflowShouldSucceed:        false, // artifact not being present causes Workflow to fail
 			expectedGCPodsOnWFCompletion: 0,
 			expectedArtifacts:            []artifactState{},
```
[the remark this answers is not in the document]
juliev0: Okay, I've modified the test Workflow so that it also successfully saves an artifact file and garbage collects it. Had to update the test to look for artifact keys names that are auto-generated.

juliev0:
Thanks for fixing the lint issue @agilgur5 !

workflow/executor/executor.go
juliev0: So, this is basically if this Pod receives a SIGKILL? I don't think there's anything that can be done in that case, is there? (at least from the Workflow Pod side)

test/e2e/artifacts_test.go
```diff
@@ -128,12 +128,53 @@ func (s *ArtifactsSuite) TestGlobalArtifactPassing() {
...
+	derivedKey   *artifactDerivedKey // exact key needs to be derived
+}
+
+type artifactDerivedKey struct {
+	templateName string
+	artifactName string
+}
+
+func (al *s3Location) getS3Key(wf *wfv1.Workflow) (string, error) {
+
```
agilgur5, line 150: nit: unnecessary new line
```suggestion
```

test/e2e/artifacts_test.go
```diff
@@ -128,12 +128,53 @@ func (s *ArtifactsSuite) TestGlobalArtifactPassing() {
...
+	if al.specifiedKey == "" && al.derivedKey == nil {
+		panic(fmt.Sprintf("invalid artifactLocation: %+v, must have knownKey or derivedKey set", al))
```
agilgur5, line 152:
```suggestion
		panic(fmt.Sprintf("invalid artifactLocation: %+v, must have specifiedKey or derivedKey set", al))
```

fix name

test/e2e/artifacts_test.go
```diff
@@ -390,25 +454,32 @@ func (s *ArtifactsSuite) TestArtifactGC() {
...
 		when = when.RemoveFinalizers(false) // just in case - if the above test failed we need to forcibly remove the finalizer for Artifact GC
 
 		then = when.Then()
 
 		for _, expectedArtifact := range tt.expectedArtifacts {
+			artifactKey, err := expectedArtifact.artifactLocation.getS3Key(when.GetWorkflow())
+			fmt.Printf("artifact key: %q\n", artifactKey)
+			if err != nil {
+				panic(err)
+			}
```
agilgur5, line 470: should this be below the `continue` conditional? Since it's not used by it, so can be skipped in that case. Unless you want to print each key?

agilgur5 (approved):
Some very tiny comments below, otherwise LGTM. Marking as approved so you can merge once you fix or answer those

workflow/executor/executor.go
agilgur5: More or less yes, but there are things we can do to mitigate that or detect that scenario, such as parallel uploads and writing pieces of the report as each artifact completes uploading (although k8s does not have streaming requests afaik, so that would put more strain on the api server when there are many artifacts).

I also recently remembered we had #12413 merged for 3.6 that does help prevent that too.

(To be clear, this is pre-existing logic / an incidental finding, so not going to block review on that, but there are perhaps actions we should take around this case. Perhaps better discussed in #12993 though)

juliev0: sounds good

PR #12470, opened 2024-01-05
fix: Resolve vulnerabilities in axios

terrytangyuan:
CI failed during Snyk scan:

```
Issues with no direct upgrade or patch:
  ✗ Prototype Pollution [High Severity][https://security.snyk.io/vuln/SNYK-JS-AXIOS-6144788] in axios@1.6.2
    introduced by swagger-ui-react@4.19.1 > swagger-client@3.24.6 > @swagger-api/apidom-reference@0.89.0 > axios@1.6.2
  This issue was fixed in versions: 1.6.4
  ✗ Improper Input Validation [High Severity][https://security.snyk.io/vuln/SNYK-JS-FOLLOWREDIRECTS-6141137] in follow-redirects@1.15.3
    introduced by swagger-ui-react@4.19.1 > swagger-client@3.24.6 > @swagger-api/apidom-reference@0.89.0 > axios@1.6.2 > follow-redirects@1.15.3
  This issue was fixed in versions: 1.15.4
```

Issue #11851, opened 2023-09-20
 Release v3.4 patch releases discussion


terrytangyuan:
This issue tracks commits for v3.4 patch releases.

agilgur5:
Do we want to cherry-pick in my UI date filter fixes? i.e. #11742, #11792, #11840

If so, I think I'd want to make another PR to remove the date filter default from the upgrade notes modal (as #11840 removes the default)

terrytangyuan:
Sure I think we'll include all fixes if there's no conflict.

terrytangyuan:
We need to include https://github.com/argoproj/argo-workflows/pull/11971 to unblock the build.

terrytangyuan:
@juliev0 and @sarabala1979 suggests including https://github.com/argoproj/argo-workflows/pull/11774

agilgur5:
> Sure I think we'll include all fixes if there's no conflict.

Ok, I added #11982 on top of those since they'll be included in a 3.4.x release then (the upgrade note in the modal is only for 3.4)

terrytangyuan:
From @juliev0:
> Without it, users in a cluster with too many sso groups can't log in to Argo Server.

So we could consider that a fix.

terrytangyuan:
Security fix https://github.com/argoproj/argo-workflows/pull/11986

terrytangyuan:
https://github.com/argoproj/argo-workflows/pull/11993

terrytangyuan:
Working branch https://github.com/argoproj/argo-workflows/commits/release-3.4.12

terrytangyuan:
Tag created: https://github.com/argoproj/argo-workflows/releases/tag/v3.4.12
Images will be published once CI succeeds: https://github.com/argoproj/argo-workflows/actions/runs/6587281270/job/17897237879

terrytangyuan:
Released

terrytangyuan:
Reusing this issue to track any potential bug fixes for future v3.4 patch releases.

terrytangyuan:
From @juliev0, https://github.com/argoproj/argo-workflows/commit/75bd0b83a479997da1940e048d5161b11cecb303

terrytangyuan:
Security fix https://github.com/argoproj/argo-workflows/pull/12111

terrytangyuan:
https://github.com/argoproj/argo-workflows/releases/tag/v3.4.13

terrytangyuan:
TODOs:

[...]

PR #13069, opened 2024-05-20
fix(deps): upgrade swagger-ui-react v5.17.3, react-dom v18.3.1, and react v18.3.1. Fixes CVEs

terrytangyuan:
This fixes CVEs: https://security.snyk.io/vuln/SNYK-JS-MICROMATCH-6838728, https://security.snyk.io/vuln/SNYK-JS-BRACES-6838727

```
   Upgrade swagger-ui-react@5.10.0 to swagger-ui-react@5.17.3 to fix
  ✗ Uncontrolled resource consumption (new) [High Severity][https://security.snyk.io/vuln/SNYK-JS-BRACES-6838727] in braces@3.0.2
    introduced by swagger-ui-react@5.10.0 > patch-package@8.0.0 > find-yarn-workspace-root@2.0.0 > micromatch@4.0.5 > braces@3.0.2
  ✗ Inefficient Regular Expression Complexity (new) [High Severity][https://security.snyk.io/vuln/SNYK-JS-MICROMATCH-6838728] in micromatch@4.0.5
    introduced by swagger-ui-react@5.10.0 > patch-package@8.0.0 > find-yarn-workspace-root@2.0.0 > micromatch@4.0.5
```


ui/package.json
```diff
@@ -28,16 +28,16 @@
         "moment": "^2.30.1",
         "monaco-editor": "^0.45.0",
         "prop-types": "^15.8.1",
-        "react": "^16.14.0",
+        "react": "^18.3.1",
         "react-chartjs-2": "^2.11.2",
         "react-datepicker": "^4.25.0",
-        "react-dom": "^16.14.0",
+        "react-dom": "^18.3.1",
```
terrytangyuan, line 34: Followed https://react.dev/blog/2022/03/08/react-18-upgrade-guide

```
yarn add react react-dom
```

ui/src/app/index.tsx
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react';
-import * as ReactDOM from 'react-dom';
 import {App} from './app';
+import {createRoot} from 'react-dom/client';
 
-ReactDOM.render(<App />, document.getElementById('app'));
+const container = document.getElementById('app');
+const root = createRoot(container!);
+root.render(<App />);
```
terrytangyuan, line 7: Reference: https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis

juliev0 (approved):

agilgur5 (commented):
### dep `overrides` aka `resolutions`

So I mentioned [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1716231247108939?thread_ts=1716215647.577759&cid=C0510EUH90V) that this needs a few _overrides_... Upgrading React by _two major versions_ is not what I meant... 😕 

Notably, I did not update `swagger-ui-react` past v5.10.x before in #12540 very intentionally due to its compatibility issues.
Since then, the issue I linked there, https://github.com/swagger-api/swagger-ui/issues/9495, was updated with a docs page https://github.com/swagger-api/swagger-ui/wiki/Using-older-version-of-React that we could follow for compatibility. That doc lists what dep overrides are necessary, and that was what I meant.

I mentioned that I hadn't fully tested this locally yet in Slack either -- I was getting this error https://github.com/swagger-api/swagger-ui/issues/9929 at the time, which may require another override.

### patch isolation, breaking major updates

We should try and isolate patches, especially security patches, as much as possible, _in general_.
Bumping another dep by two major versions, especially the foundational UI framework, React, is the opposite of that.

Such a foundational, breaking upgrade should also really have a thorough review by a UI SME as well, which this did not.
The "Verification" section, as is required in the PR template, and which is in general important for the UI (where we lack many tests and visual nuances are relevant), and especially for such a bump, is missing here as well.

We had not upgraded before due to several incompatibilities. We already had deprecated functionality and off the top of my head, two notable ones I can think of include:
1. #11891 and a few prior refactors of mine removed `BasePage`, which used the [legacy, undocumented context API](https://legacy.reactjs.org/docs/legacy-context.html), which was only promised to work with React 16 and would be removed in future breaking changes.
1. Argo UI still has many class components. In particular, several were using [unsafe lifecycle methods](https://legacy.reactjs.org/docs/react-component.html#unsafe_componentwillmount) which I only recently updated to use the non-deprecated `UNSAFE_` prefix in https://github.com/argoproj/argo-ui/pull/509. The deprecated, unprefixed variants were similarly only promised to work with React 16 and were removed in React 17.
  - I also only recently updated `argo-ui` to a version with those changes in #12516. That was 2 weeks ago.
  
So if upgrading React by two major versions works, it is almost entirely coincidence due to refactoring work I had been doing months prior to make such an upgrade possible.

Notably, such coincidences mean that there is no way to backport this cleanly.

You also wanted to backport this to Argo 3.4 in https://github.com/argoproj/argo-workflows/issues/11851#issuecomment-2120589137; outside of the issues above, Argo 3.4 is itself only [on `swagger-ui-react` v4](https://github.com/argoproj/argo-workflows/blob/89cbdb53361cbe59fbe81b887ee82722cce5de54/ui/package.json#L41) as #12540 was never backported.

I hope all the careful refactoring and months-long changes I had been making to prepare for such an upgrade, as well as all the incompatibilities that may result from attempting to backport this PR, illustrate why such breaking upgrades to foundational frameworks need significant diligence & review around them.

### CVE has minimal impact, would prefer to remove `swagger-ui-react` entirely tbh

`swagger-ui-react` is only used on a single page in the UI, to load the API docs. As of https://github.com/argoproj/argo-workflows/pull/12061, it isn't even loaded until a user navigates to that page.

As such, a CVE in `swagger-ui-react` is of minimal effect to Argo users _in general_.
These _particular_ CVEs are in deps of deps of `swagger-ui-react` and are not really exploitable as such (maybe with a malicious swagger doc, which would require a few other exploits to be able to do, and even then is only a maybe).

Since `swagger-ui-react` is such a big dep with complex deps of its own (that have also had CVEs more than once, e.g. #12470, #12058) and has various efficiency issues per #6645 et al, I've actually been thinking of entirely removing the API Docs page to reduce our complexity & deps as well as improve the speed of installation and build. 
Users can still use the [API reference in the docs](https://argo-workflows.readthedocs.io/en/latest/swagger/) (which is aided now by versioned docs #11390) and if they want the specific version of their Server, could download it from the API and load it locally in their own editor etc.

ui/src/app/index.tsx
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react';
-import * as ReactDOM from 'react-dom';
 import {App} from './app';
+import {createRoot} from 'react-dom/client';
```
agilgur5, line 3: we should leave the import in the same place; the `react` deps and the external deps were grouped and alphabetized previously.

```suggestion
import {createRoot} from 'react-dom/client';

import {App} from './app';
```

I've been meaning to add an ESLint or Prettier plugin to auto sort imports for a bit now

ui/src/app/index.tsx
```diff
@@ -1,5 +1,7 @@
...
 
-ReactDOM.render(<App />, document.getElementById('app'));
+const container = document.getElementById('app');
+const root = createRoot(container!);
```
agilgur5, line 6:
```suggestion
const root = createRoot(document.getElementById('app')!);
```

we can probably simplify this, the `container` var is otherwise unused

agilgur5:
The React upgrade also produced a bunch of new install warnings too:

```
warning " > argo-ui@1.0.0" has incorrect peer dependency "react@^16.9.3".
warning " > argo-ui@1.0.0" has incorrect peer dependency "react-dom@^16.9.3".
warning "argo-ui > react-autocomplete@1.8.1" has incorrect peer dependency "react@^0.14.7 || ^15.0.0-0 || ^16.0.0-0".
warning "argo-ui > react-autocomplete@1.8.1" has incorrect peer dependency "react-dom@^0.14.7 || ^15.0.0-0 || ^16.0.0-0".
warning "argo-ui > react-form@2.16.3" has incorrect peer dependency "react@^15.6.1 || 16".
warning "argo-ui > react-form@2.16.3" has incorrect peer dependency "react-dom@^15.6.1 || 16".
warning "argo-ui > react-form > react-redux@5.1.2" has incorrect peer dependency "react@^0.14.0 || ^15.0.0-0 || ^16.0.0-0".
warning " > react-chartjs-2@2.11.2" has incorrect peer dependency "react@^0.14.0 || ^15.0.0 || ^16.0.0 || ^17.0.0".
warning " > react-chartjs-2@2.11.2" has incorrect peer dependency "react-dom@^0.14.0 || ^15.0.0 || ^16.0.0 || ^17.0.0".
```

I'm not sure if all of those are safely ignoreable. _Some_ of them are certainly fixable though (`react-autocomplete` is not, since the library was archived years ago)

ui/src/app/index.tsx
agilgur5: Imports sorted in #13813

agilgur5:
`swagger-ui-react` removed in #13818
React 18 upgrade reverted in #13821

Issue #12589, opened 2024-01-30
proposal: metrics overhaul including opentelemetry


Joibel:
# Proposal for improving workflows metrics

This proposal is an overhaul of workflow metrics to bring them up to date, correct them, and make sure they are useful. It is intended to be a breaking change requiring users to update their use of metrics.

## [Opentelemetery](https://opentelemetry.io/)

I'd like to move to using opentelemetry (otel) as the libraries for collecting metrics. They are mature and the design of the interfaces encourages good practices.

We'd still be able to expose the resulting metrics as a prometheus compatible `/metrics` scrapable endpoint using any prometheus compatible scraper.

It would allow us to also emit the same metrics using the opentelemetry protocol with the benefits this provides. This would also allow us to emit tracing using the same protocol and libraries. I do not intend to implement tracing as part of this proposal.

Each transmission protocol would be separately configured. Data collection would be identical in the code no matter which protocols are in use.

# Changes to the metrics

Some words have specific meanings in the world of observability. [Opentelemetry metrics](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments) explains these words - make particular note of counter and gauge which are currently used incorrectly within workflows. Counts MUST only ever go up.

A counter is something that ever-increases (unless the controller is restarted, in which case it will reset to 0, which the querying tool will understand and do the correct thing). You can use this to monitor changes using "increase" functions which allow you to monitor changes over your choice of time period. Gauges can be constructed from counters given both a count of incoming and outgoing events.

This is being done with the mindset of in future running multiple active controllers (horizontal controller scaling [Horizontally Scalable Controller](https://github.com/argoproj/argo-workflows/issues/9990) ). The existing metrics are hard to use in this context.

## Existing metrics

| Existing name | Proposed name | User reported problems |
| --- | --- | --- |
| argo_pod_missing | argo_workflows_pod_missing_count | Match the prefix and change this to a counter of incidents. As a gauge you cannot easily tell what happened on workflow controller restart |
| `argo_workflows_count` | `argo_workflows_gauge` | This is emitting the number workflow resources that are in the cluster. If the user makes use of garbage collection and/or archiving, the workflows will no longer be counted (and the 'count' will go down). |
| new | `argo_workflows_total_count` | count workflows as they enter each phase (`pending`, `running`, `errored`, `succeeded`, `failed`, `removed` ) as a true counter |
| `argo_workflows_error_count` | no change |  |
[...]

PR #13077, opened 2024-05-23
feat: new Prometheus metrics build_info

nlamirault:
Fixes #TODO

### DIFF
--- a/workflow/metrics/server.go
+++ b/workflow/metrics/server.go
@@ -53,6 +53,8 @@ func (m *Metrics) RunServer(ctx context.Context, isDummy bool) {
 		go runServer(m.telemetryConfig, telemetryRegistry, ctx, isDummy)
 	}
 
+	recordBuildInfo()
+
 	// Run the metrics server
 	go runServer(m.metricsConfig, metricsRegistry, ctx, isDummy)
 
--- a/workflow/metrics/server.go
+++ b/workflow/metrics/server.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"runtime"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


Joibel:
I had proposed adding this in #12589. I am hoping to get that in for the next release, so this may be redundant work.

If you'd like to carry on with it please fill out the PR description, and add this to the documentation.

I have also included all of the values from GetVersion(), and not derived through other means. For compatibility with this I'd like it to match something like this.

```
	labelBuildVersion      string = `version`
	labelBuildPlatform     string = `platform`
	labelBuildGoVer        string = `goversion`
	labelBuildDate         string = `build`
	labelBuildCompiler     string = `compiler`
	labelBuildGitCommit    string = `commit`
	labelBuildGitTreeState string = `treestate`
	labelBuildGitTag       string = `tag`
```
...
```
	{name: labelBuildVersion, value: version.Version},
	{name: labelBuildPlatform, value: version.Platform},
	{name: labelBuildGoVer, value: version.GoVersion},
	{name: labelBuildDate, value: version.BuildDate},
	{name: labelBuildCompiler, value: version.Compiler},
	{name: labelBuildGitCommit, value: version.GitCommit},
	{name: labelBuildGitTreeState, value: version.GitTreeState},
	{name: labelBuildGitTag, value: version.GitTag},
```

workflow/metrics/server.go
```diff
@@ -150,3 +165,9 @@ func (m *Metrics) garbageCollector(ctx context.Context) {
 		}
 	}
 }
+
+// recordBuildInfo publishes information about Argo-Rollouts version and runtime info through an info metric (gauge).
```
agilgur5, line 169: this isn't Argo Rollouts
```suggestion
// recordBuildInfo publishes metadata about Argo's version and runtime through a metric (gauge).
```

workflow/metrics/server.go
```diff
@@ -14,9 +15,20 @@ import (
...
 
+	"github.com/argoproj/argo-workflows/v3"
 	tlsutils "github.com/argoproj/argo-workflows/v3/util/tls"
 )
 
+var (
+	buildInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "build_info",
+			Help: "A metric with a constant '1' value labeled by version from which Argo-Workflows was built.",
```
agilgur5, line 26: and is not referred to as "Argo-Workflows" within the codebase or docs (generally speaking). 

this should also be consistent with the comment

```suggestion
			Help: "A metric with a constant '1' value labeled with metadata about Argo's version and runtime.",
```

agilgur5 (commented):
some minor modifications below to the current code, otherwise Alan can do follow-on reviews

agilgur5:
Superseded by #13269

PR #13087, opened 2024-05-24
docs: clarify `Skipped` and `Omitted` in enhanced `depends`

oliverdain:
docs: update the enhanced depends documentation

It wasn't clear to me what `Omitted` and `Skipped` meant. I'm still not sure I have this correct but I'm going off [advice I got](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1716512517981539) in the Argo Slack.

### DIFF
--- a/docs/enhanced-depends-logic.md
+++ b/docs/enhanced-depends-logic.md
@@ -25,7 +25,10 @@ available task results is as follows:
 | `.Omitted` | Task Omitted | Task was omitted |
 | `.Daemoned` | Task is Daemoned and is not Pending | |
 
-For convenience, if an omitted task result is equivalent to `(task.Succeeded || task.Skipped || task.Daemoned)`.
+A tasks is considered `Skipped` if its `when` condition evaluates to false. On the other hand, if a task doesn't run
+because its `depends` evaluated to false it is `Omitted`.
+
+For convenience, an omitted task result is equivalent to `(task.Succeeded || task.Skipped || task.Daemoned)`.
 
 For example:
 
--- a/docs/enhanced-depends-logic.md
+++ b/docs/enhanced-depends-logic.md
@@ -28,7 +28,6 @@ available task results is as follows:
 A tasks is considered `Skipped` if its `when` condition evaluates to false. On the other hand, if a task doesn't run
 because its `depends` evaluated to false it is `Omitted`.
 
-
 For convenience, an omitted task result is equivalent to `(task.Succeeded || task.Skipped || task.Daemoned)`.
 
 For example:
--- a/docs/enhanced-depends-logic.md
+++ b/docs/enhanced-depends-logic.md
@@ -28,6 +28,7 @@ available task results is as follows:
 A tasks is considered `Skipped` if its `when` condition evaluates to false. On the other hand, if a task doesn't run
 because its `depends` evaluated to false it is `Omitted`.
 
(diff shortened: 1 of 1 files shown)


Joibel (approved):

Joibel:
Thanks @oliverdain for your contribution.

docs/enhanced-depends-logic.md
```diff
@@ -25,7 +25,10 @@ available task results is as follows:
 | `.Omitted` | Task Omitted | Task was omitted |
 | `.Daemoned` | Task is Daemoned and is not Pending | |
 
-For convenience, if an omitted task result is equivalent to `(task.Succeeded || task.Skipped || task.Daemoned)`.
+A tasks is considered `Skipped` if its `when` condition evaluates to false. On the other hand, if a task doesn't run
+because its `depends` evaluated to false it is `Omitted`.
```
agilgur5, line 29:
```suggestion
A task is `Skipped` if its `when` condition evaluated to false. A task is `Omitted` if its `depends` condition evaluated to false.
```

- grammar: singular "A task", not "A tasks"
- grammar: use consistent tense -- "evaluates" + "evaluated" -> "evaluated"
  - past tense is also correct in this case, the task result is after the task.
- simplify, [per k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - also "On the other hand" is an idiom that should be avoided, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-jargon-and-idioms)
    - and it is not quite accurate either; it typically refers to opposites or contrasts, whereas `depends` and `when` can be used together and are independent.
    
Also, rather than having this as two separate sentences below the table, these should just be directly in the table.

docs/enhanced-depends-logic.md
```diff
@@ -25,7 +25,10 @@ available task results is as follows:
...
+
+For convenience, an omitted task result is equivalent to `(task.Succeeded || task.Skipped || task.Daemoned)`.
 
 For example:
```
agilgur5, line 33: You didn't change this line much (just removed the grammatically incorrect "if"), but the word "omitted" here is very confusing given that it means the literal "omitted" and not the term `.Omitted` defined above.

We should use a different word:
```suggestion
An unspecified result is equivalent to `(task.Succeeded || task.Skipped || task.Daemoned)`. For example:
```

And should have less space between it and the example, which also alleviates confusion

agilgur5 (commented):
I left some grammatical corrections below, but this can also be improved and simplified, per [our style guide](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/).

I think this whole page could use a copy-edit though (like some of my previous docs changes), so I can make those changes as part of a larger rewrite. **EDIT**: See #13110

docs/enhanced-depends-logic.md
agilgur5: all of the above completed in #13110

docs/enhanced-depends-logic.md
agilgur5: Completed in #13110

docs/enhanced-depends-logic.md
oliverdain: Nice. Yes, #13110 is a better fix than mine.

Issue #13089, opened 2024-05-24
reduce pod definition size


tooptoop4:
# Summary

The pod spec for steps run from argo is quite large and can fill etcd. Wanting to trim this down, I imagine there will be 2 parts:
1. code changes
2. docs for configration of init vs main vs wait containers

thoughts/Qs:
a) is ARGO_TEMPLATE env variable needed on all 3 containers?
b) are volumeMounts needed on both main and wait containers?
c) do all 3 containers need environment variables for communicating with s3 ie via artifactRepository?
d) ARGO_TEMPLATE  is huge, is the ARGO_TEMPLATE env variable containing things it doesn't need? perhaps some containers need things within that other containers don't?
e) are ARGO_PROGRESS_PATCH_TICK_DURATION/ARGO_PROGRESS_FILE_TICK_DURATION/ARGO_INCLUDE_SCRIPT_OUTPUT/ARGO_PROGRESS_FILE env variables needed?
f) do the commands on all the containers need --loglevel/--log-format?

## Use Cases

ensure etcd does not fill up

jswxstw:
> * is ARGO_TEMPLATE env variable needed on all 3 containers?

ARGO_TEMPLATE may be huge, #12325 provides an optimization solution for EnvVarTemplate offload, perhaps it can be made the default logic. However, its lifecycle is aligned with the workflow, not the pod, considering delete it when pod gc?

> are ARGO_PROGRESS_PATCH_TICK_DURATION/ARGO_PROGRESS_FILE_TICK_DURATION/ARGO_INCLUDE_SCRIPT_OUTPUT/ARGO_PROGRESS_FILE env variables needed?

ARGO_PROGRESS_PATCH_TICK_DURATION/ARGO_PROGRESS_FILE_TICK_DURATION/ARGO_PROGRESS_FILE are used to implement [self reporting progress](https://argo-workflows.readthedocs.io/en/stable/progress/#self-reporting-progress).
ARGO_INCLUDE_SCRIPT_OUTPUT is used to determine whether stdout needs to be saved.

I think optimizing the reuse of ARGO_TEMPLATE would be sufficient. The other aspects have minimal impact, there's no need to be overly demanding.

tooptoop4:
@jswxstw do u know what ARGO_TEMPLATE is for?

jswxstw:
> @jswxstw do u know what ARGO_TEMPLATE is for? and do all 3 containers need it?

[...]

PR #12979, opened 2024-04-24
fix: consistently set executor log options 

agilgur5:
Fixes #12788, Fixes https://github.com/argoproj/argo-workflows/pull/12929#discussion_r1572707894

### Motivation

- loglevel, gloglevel, and log format were inconsistently set around the codebase for executor containers
  - meaning that some had partial settings and some had no settings, depending on the command run etc

### Modifications

- for all commands, including `emissary`, `agent`, `resource`, and `data`
  - set loglevel, gloglevel, and log format

- replace `getExecutorLogLevel` helper with `getExecutorLogOpts` helper
- add a helper in `util/cmd/glog.go` to get the gloglevel

### Verification

Existing tests pass

PR #13093, opened 2024-05-25
feat: don't add args with same value as default(#13089)

tooptoop4:
Fixes point 6 of #13089


agilgur5 (commented):
Unit tests and lint are failing and need adjusting.

This should also be built on top of #12979 which refactors things as-is.

Though this is such a tiny optimization that I'm not sure it's worth complicating the code and tests for to begin with

Joibel:
I agree with @agilgur5 here. This is a lot of extra code for a minor shrinkage of the resulting pods. I'd prefer we didn't merge this as is.

PR #4470, opened 2020-11-05
feat(ui): Argo Events API and UI. Fixes #888

alexec:
Checklist:

* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).

Features

* New API endpoints for Argo Events
* New event-flow page
* Create, edit, view, log event sources and sensors in the UI
* Embeddable widgets
* New workflow log viewer
* Configurable "Get help" button
* Refactored code to use more robust React functional components 
* Seemless reconnection on network errors
* Many small tweaks 
* Lots of bugs fixed

Fixes #888

PR #13094, opened 2024-05-27
refactor(ui): heavily simplify `WorkflowCreator` effect

agilgur5:
### Motivation

Reduce tech debt, code complexity, etc

### Modifications

- reduce nesting in this effect quite significantly with early returns
  - and remove an unnecessary `switch` statement
    - (not sure if it had more than one `case` before? it `git blame`s back to #4470 which moved the UI from `argo-ui` to this repo)

### Verification

No semantic changes, functionally equivalent.
Tested creating a Workflow in different ways and it worked fine.

### Notes to Reviewers

The diff is much easier to read with [the ignore whitespace setting](https://github.com/argoproj/argo-workflows/pull/13094/files?diff=unified&w=1)


tczhao (approved):
lgtm

PR #13095, opened 2024-05-27
chore: remove Jason Meridth from Sustainability Effort

juliev0:
### Motivation

Jason's switched projects at work and no longer has time for Workflows, unfortunately.


jmeridth (approved):

agilgur5:
JM works for GH now, so different employer too 😅 

Sadge to hear though 😿

agilgur5 (approved):

juliev0:
> JM works for GH now, so different employer too 😅
> 
> Sadge to hear though 😿

Ahh, got it

PR #12976, opened 2024-04-24
chore!: remove legacy `patch` `pods` fallback

agilgur5:
### Motivation

- the fallback is old and insecure, and the error confuses users as it's not mentioned in the docs (as it's legacy and a fallback)
  - e.g. #12783 and also frequently on Slack. See #13041 for more such examples of confusion
- it's also tech debt that we have to write code around specifically right now
- it's no longer needed and hasn't been the main RBAC in a few versions, so remove it in the next breaking minor release

### Modifications

- remove the Executor code that patches pods
- remove the operator code that reads the patched annotations

### Verification

Tests pass

Issue #12783, opened 2024-03-11
`User "system:serviceaccount:argo:default" cannot patch resource "pods" in API group "" in the namespace "argo"`


DenysSkrytskyi:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

WF template: https://argo-workflows.readthedocs.io/en/latest/walk-through/hello-world/
Version: v3.5.1 - no error
Version: v3.5.2 and higher- with error

Error:
```
hello-world-hnd8w   Error    9m31s   Error (exit code 1): pods "hello-world-hnd8w" is forbidden: User "system:serviceaccount:argo:default" cannot patch resource "pods" in API group "" in the namespace "argo"
```

### Version

v3.5.2

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow                  # new type of k8s spec
metadata:
  generateName: hello-world-    # name of the workflow spec
spec:
  entrypoint: whalesay          # invoke the whalesay template
  templates:
    - name: whalesay              # name of the template
      container:
        image: docker/whalesay
        command: [ cowsay ]
        args: [ "hello world" ]
        resources: # limit the resources
          limits:
            memory: 32Mi
            cpu: 100m
```

### Logs from the workflow controller

```text
time="2024-03-11T21:25:52.335Z" level=info msg="Processing workflow" namespace=argo workflow=hello-world-nnkdk
time="2024-03-11T21:25:52.339Z" level=info msg="Task-result reconciliation" namespace=argo numObjs=0 workflow=hello-world-nnkdk
time="2024-03-11T21:25:52.339Z" level=info msg="Updated phase  -> Running" namespace=argo workflow=hello-world-nnkdk
time="2024-03-11T21:25:52.339Z" level=warning msg="Node was nil, will be initialized as type Skipped" namespace=argo workflow=hello-world-nnkdk
time="2024-03-11T21:25:52.340Z" level=info msg="was unable to obtain node for , letting display name to be nodeName" namespace=argo workflow=hello-world-nnkdk
[...]

PR #7999, opened 2022-02-24
docs: Split out executor/agent roles in quick-start RBAC.

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>

The aim of this PR is to discourage excessive permissioning of workflows, as well as make it very clear what permissions each executor needs.

TODO:

- [x] fix spelling
- [x] search for workflow-role.yaml links and update them

PR #13100, opened 2024-05-27
chore!: remove legacy patch pods fallback

Garett-MacGowan:
PR Takeover https://github.com/argoproj/argo-workflows/pull/12976. Description is mostly verbatim

### Motivation

- the fallback is old and insecure, and the error confuses users as it's not mentioned in the docs (as it's legacy and a fallback)
    - e.g. https://github.com/argoproj/argo-workflows/issues/12783 and also frequently on Slack. See #13041 for more such examples of confusion
- it's also tech debt that we have to write code around specifically right now
- it's no longer needed and hasn't been the main RBAC in a few versions, so remove it in the next breaking minor release

### Modifications

- remove the Executor code that patches pods
- remove the operator code that reads the patched annotations

### Verification

Tests pass


Garett-MacGowan:
@agilgur5 my WIP takeover. I'll finish this at some point this week.

Garett-MacGowan:
@agilgur5 Some of these tests are failing because some of the test workflows use the argo service account and the argo service account doesn't have `create` and `patch` permissions for `workflowtaskresults`. Would there be an issue with adding those permissions to the `argo-role`?

agilgur5:
> some of the test workflows use the argo service account and the argo service account doesn't have `create` and `patch` permissions for `workflowtaskresults`.

Oh. That's surprising, that means these tests have been using the wrong RBAC for some time and relying on the fallback 😬

> Would there be an issue with adding those permissions to the `argo-role`?

Yes, especially in the official install manifests as you have them now. The Controller does not need those privileges and so shouldn't have them.
The Executor should really have it's own SA & Role. We can add those to the test overlays. I don't remember which install the tests use, but if they already use the quick start, I believe there is an Executor Role in those manifests already (although I might be mistaken, it may just have excess privileges in general since it's a non-prod quick start).

Garett-MacGowan:
Yes, the tests use the various quick start manifests depending on the profile you choose. There is an executor role (same as in the docs) that is bound to the default service account. Are you suggesting that I bind the executor role to the argo service account in the e2e manifest mixins?

Below is one of the test cases where the argo service account needs the executor permissions. If i'm not mistaken the template service account (not just the pod service account) needs the executor permissions. Since it is creating a resource, it also needs the argo permissions.

```
kind: Workflow
metadata:
  generateName: k8s-resource-tmpl-with-pod-
spec:
  serviceAccount: argo
  entrypoint: main
  templates:
    - name: main
      serviceAccountName: argo
      resource:
        action: create
        setOwnerReference: true
        successCondition: status.phase == Succeeded
        failureCondition: status.phase == Failed
        manifest: |
          apiVersion: v1
          kind: Pod
          metadata:
            generateName: k8s-pod-resource-
          spec:
            serviceAccountName: argo
            containers:
            - name: argosay-container
              image: argoproj/argosay:v2
              command: ["/argosay"]
            restartPolicy: Never
```

agilgur5:
> that is bound to the default service account.

Ah right of course since quick start.

> Are you suggesting that I bind the executor role to the argo service account in the e2e manifest mixins?

> I could also create a dedicated service account for these tests with the required permissions, I suppose.

The latter would be more correct and least privilege (i.e. less possible future bugs with regard to the tests)

> If I'm not mistaken, the template service account (not just the pod service account) needs the executor permissions

The `Pod` _within_ the `resource` doesn't need an SA at all since it just runs `argosay`. Only the template needs the SA, which runs the `create` action and ofc needs Executor permissions.

workflow/controller/controller_test.go
```diff
@@ -513,6 +484,43 @@ func syncPodsInformer(ctx context.Context, woc *wfOperationCtx, podObjs ...apiv1
 	}
 }
 
+func withOutputs(outputs wfv1.Outputs) with {
```
agilgur5, line 487: nit: could you put these in-line with `with` and the old functions / above `createRunningPods`? i.e. so that the diff lines up 😅

Garett-MacGowan:
> > that is bound to the default service account.
> 
> Ah right of course since quick start.
> 
> > Are you suggesting that I bind the executor role to the argo service account in the e2e manifest mixins?
> 
> > I could also create a dedicated service account for these tests with the required permissions, I suppose.
> 
> The latter would be more correct and least privilege (i.e. less possible future bugs with regard to the tests)
> 
> > If I'm not mistaken, the template service account (not just the pod service account) needs the executor permissions
> 
> The `Pod` _within_ the `resource` doesn't need an SA at all since it just runs `argosay`. Only the template needs the SA, which runs the `create` action and ofc needs Executor permissions.

I just realized the pod create permissions are bound to the default service account with the pod-manager role. Switching the tests that used the argo service account to use the default service account instead.

Garett-MacGowan:
@agilgur5 I think this test is/was wrong? The container doesn't inherit permissions from the executor, right?

I changed
```
  executor:
    serviceAccountName: argo
```
to
```
  executor:
    serviceAccountName: default
```

The full test
```
func (s *WorkflowSuite) TestContainerTemplateAutomountServiceAccountTokenDisabled() {
	s.Given().Workflow(`
kind: Workflow
metadata:
  generateName: get-resources-via-container-template-
  namespace: argo
spec:
  serviceAccountName: argo
  automountServiceAccountToken: false
  executor:
    serviceAccountName: default
  entrypoint: main
  templates:
    - name: main
      container:
        name: main
        image: bitnami/kubectl
        command:
          - sh
        args:
          - -c
          - |
           kubectl get cm 
`).
		When().
		SubmitWorkflow().
		WaitForWorkflow(fixtures.ToBeSucceeded, time.Minute*11).
		Then().
		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
			assert.Equal(t, wfv1.WorkflowSucceeded, status.Phase)
		})
}
```

The failure
```
    workflow_test.go:53: condition never and cannot be met because the workflow is done
Checking expectation get-resources-via-container-template-slp6q
get-resources-via-container-template-slp6q : Failed Error (exit code 1)
    workflow_test.go:56: 
        	Error Trace:	/home/runner/work/argo-workflows/argo-workflows/test/e2e/workflow_test.go:56
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/fixtures/then.go:69
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/fixtures/then.go:44
        	            				/home/runner/work/argo-workflows/argo-workflows/test/e2e/workflow_test.go:55
        	Error:      	Not equal: 
        	            	expected: "Succeeded"
        	            	actual  : "Failed"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(v1alpha1.WorkflowPhase) (len=9) "Succeeded"
        	            	+(v1alpha1.WorkflowPhase) (len=6) "Failed"
        	            	 
        	Test:       	TestWorkflowSuite/TestContainerTemplateAutomountServiceAccountTokenDisabled
```

Shouldn't we be expecting the failure?

Garett-MacGowan:
@agilgur5 Ready for review

agilgur5 (commented):
Oh I totally forgot to add a release note in my initial variant of the PR -- we should add a small one to the 3.6 `upgrading.md` and just link to the Workflow RBAC page (where I removed the old RBAC in #12975)

Garett-MacGowan:
> Oh I totally forgot to add a release note in my initial variant of the PR -- we should add a small one to the 3.6 `upgrading.md` and just link to the Workflow RBAC page (where I removed the old RBAC in #12975)

Added.

Joibel:
This looks good to me, thanks for the work on it @Garett-MacGowan.

I was wondering if we could modify the error message you see in the GUI to include a link to the documentation for fixing it as it's really common to see people with incorrect RBAC still. Currently you see
```
Error (exit code 1): workflowtaskresults.argoproj.io is forbidden: User "system:serviceaccount:argo:default" cannot create resource "workflowtaskresults" in API group "argoproj.io" in the namespace "argo"
```
which is probably obvious to you and I, but for most people who hope it'd just work, they'll be confused.

It doesn't have to be a part of this PR though.

I'll leave it to @agilgur5 to have a look over it.

Garett-MacGowan:
> This looks good to me, thanks for the work on it @Garett-MacGowan.
> 
> I was wondering if we could modify the error message you see in the GUI to include a link to the documentation for fixing it as it's really common to see people with incorrect RBAC still. Currently you see
> 
> ```
> Error (exit code 1): workflowtaskresults.argoproj.io is forbidden: User "system:serviceaccount:argo:default" cannot create resource "workflowtaskresults" in API group "argoproj.io" in the namespace "argo"
> ```
> 
> which is probably obvious to you and I, but for most people who hope it'd just work, they'll be confused.
> 
> It doesn't have to be a part of this PR though.
> 
> I'll leave it to @agilgur5 to have a look over it.

@Joibel Thanks. I could add a link to the GUI. Will put it in another PR.

Joibel:
@Garett-MacGowan, any chance you could bring this up to date so it can be merged?

Garett-MacGowan:
> @Garett-MacGowan, any chance you could bring this up to date so it can be merged?

Sure, I'll work on it today

Garett-MacGowan:
> `AnnotationKeyProgress`

Thanks @jswxstw.

@Joibel sorry, life stuff happened. I'll get to this today, hopefully.

Garett-MacGowan:
> This looks good to me, thanks for the work on it @Garett-MacGowan.
> 
> I was wondering if we could modify the error message you see in the GUI to include a link to the documentation for fixing it as it's really common to see people with incorrect RBAC still. Currently you see
> 
> ```
> Error (exit code 1): workflowtaskresults.argoproj.io is forbidden: User "system:serviceaccount:argo:default" cannot create resource "workflowtaskresults" in API group "argoproj.io" in the namespace "argo"
> ```
> 
> which is probably obvious to you and I, but for most people who hope it'd just work, they'll be confused.
> 
> It doesn't have to be a part of this PR though.
> 
> I'll leave it to @agilgur5 to have a look over it.

Could you point me to where this is displayed in the UI? It has been a while. Thanks.

Garett-MacGowan:
@Joibel @agilgur5 Ready for review.

@jswxstw Do you mind taking a look to verify that what I've done is what you had in mind?

Thanks.

jswxstw:
> @jswxstw Do you mind taking a look to verify that what I've done is what you had in mind?

It looks good to me now. 

By the way, does [this document](https://argo-workflows.readthedocs.io/en/stable/progress/#self-reporting-progress) need to be updated since its content is outdated?
> The executor will read this file every 3s and if there was an update, **patch the pod annotations** with `workflows.argoproj.io/progress: N/M`.

Garett-MacGowan:
> > @jswxstw Do you mind taking a look to verify that what I've done is what you had in mind?
> 
> It looks good to me now.
> 
> By the way, does [this document](https://argo-workflows.readthedocs.io/en/stable/progress/#self-reporting-progress) need to be updated since its content is outdated?
> 
> > The executor will read this file every 3s and if there was an update, **patch the pod annotations** with `workflows.argoproj.io/progress: N/M`.

What in particular is out of date? The pod patches still happen, they just need the RBAC now. See `monitorProgress` of `executor.go`

jswxstw:
> What in particular is out of date? The pod patches still happen, they just need the RBAC now. See monitorProgress of executor.go

Aren’t pod patches removed by the current PR and progress can only be reported to `WorkflowTaskResult` now?

Garett-MacGowan:
> Aren’t pod patches removed by the current PR and progress can only be reported to `WorkflowTaskResult` now?

Right, sorry. The updates to progress are done via `WorkflowTaskResult` and are reflected by
```
newNode.Progress = result.Progress
...
woc.wf.Status.Nodes.Set(nodeID, *newNode)
```
in `taskResultReconciliation`. The pod annotations are not updated anymore.

How about the following change?

Instead of:
> The executor will read this file every 3s and if there was an update, patch the pod annotations with workflows.argoproj.io/progress: N/M. The controller picks this up and writes the progress to the appropriate Status properties.

We can say:

> The executor will read this file every 3s, and if there was an update, create a WorkflowTaskResult containing the progress update. The controller picks this up and writes the progress to the appropriate Status properties.

Garett-MacGowan:
Wait... I just checked the file on my branch. It has already been updated...

```
The executor will read this file every 3s and if there was an update,
patch the matching `WorkflowTaskResults` resource with the value `N/M`.
The controller picks this up and writes the progress to the appropriate Status properties
every 1m.
```

Garett-MacGowan:
@agilgur5 I think this is ready for review.

workflow/controller/controller_test.go
```diff
@@ -447,20 +448,42 @@ func listPods(woc *wfOperationCtx) (*apiv1.PodList, error) {
...
+		if err != nil {
+			panic(err)
+		}
 	}
 }
-func withProgress(v string) with { return withAnnotation(common.AnnotationKeyProgress, v) }
 
 func withExitCode(v int32) with {
-	return func(pod *apiv1.Pod) {
+	return func(pod *apiv1.Pod, woc *wfOperationCtx) {
```
Joibel, line 486:
```suggestion
	return func(pod *apiv1.Pod, _ *wfOperationCtx) {
```

Joibel (approved):
One suggestion, but LGTM

Garett-MacGowan:
> One suggestion, but LGTM

Thanks @Joibel

agilgur5:
> > I was wondering if we could modify the error message you see in the GUI to include a link to the documentation for fixing it as it's really common to see people with incorrect RBAC still. Currently you see
> > ```
> > Error (exit code 1): workflowtaskresults.argoproj.io is forbidden: User "system:serviceaccount:argo:default" cannot create resource "workflowtaskresults" in API group "argoproj.io" in the namespace "argo"
> > ```
> 
> Could you point me to where this is displayed in the UI? It has been a while. Thanks.

Sorry I forgot to respond to this at the time Alan wrote it and then had not been feeling well for much of the past week.

This does not appear in the UI per se, it appears in logs, which can appear in the UI. I did specifically address this with an FAQ entry in https://github.com/argoproj/argo-workflows/pull/13041 (as listed in the PR description).
That should be changed for 3.6 then. 

Alan specifically meant to catch that error and instead return a link to the docs. [This line](https://github.com/argoproj/argo-workflows/pull/13100/files#diff-791eed50c295312394166c66addd3b676cc50ba3126730304c1fab2d5cac7a23R805) logs a link to the docs, but does not change the end error

manifests/quick-start/base/workflow-default-rolebinding.yaml
```diff
@@ -1,11 +0,0 @@
-apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
-metadata:
-  name: workflow-default-binding
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: workflow
```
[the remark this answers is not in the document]
agilgur5: Looks like it was supposed to be removed / renamed in https://github.com/argoproj/argo-workflows/pull/7999 (see also [this removed file](https://github.com/argoproj/argo-workflows/pull/7999/files#diff-5eecd917194d6b7bcbd2091b2e2b562bdd715e56bae37813d9dc73226e2ce6cd)) but got replaced by other files instead if I'm reading correctly

workflow/controller/controller_test.go
```diff
@@ -447,20 +448,42 @@ func listPods(woc *wfOperationCtx) (*apiv1.PodList, error) {
...
+				context.Background(),
+				taskResult,
+				metav1.CreateOptions{},
+			)
+		if err != nil {
+			panic(err)
+		}
 	}
 }
-func withProgress(v string) with { return withAnnotation(common.AnnotationKeyProgress, v) }
```
agilgur5, line 460: as in this line

workflow/controller/operator_test.go
```diff
@@ -291,25 +291,17 @@ spec:
...
-	woc.operate(ctx)
-
-	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
-	assert.Equal(t, wfv1.Progress("50/100"), woc.wf.Status.Progress)
-	assert.Equal(t, wfv1.Progress("50/100"), woc.wf.Status.Nodes[woc.wf.Name].Progress)
-	pod = woc.wf.Status.Nodes.FindByDisplayName("pod")
-	assert.Equal(t, wfv1.Progress("50/100"), pod.Progress)
-
-	// mock workflow uses legacy/insecure pod patch
-	makePodsPhase(ctx, woc, apiv1.PodSucceeded, withAnnotation(common.AnnotationKeyReportOutputsCompleted, "true"), withProgress("100/100"))
```
agilgur5, line 312: and these assertions above

agilgur5 (commented):
> Wait... I just checked the file on my branch. It has already been updated...

Re: progress, https://github.com/argoproj/argo-workflows/pull/12089 modified this and was recently merged (the PR is a year old otherwise). I actually mentioned @jswxstw's https://github.com/argoproj/argo-workflows/pull/13260 there too; I'm honestly not entirely sure how self-reported progress is supposed to work now and which versions it changed in -- I'm guessing in 3.4, same as `WorkflowTaskResults` in general. We should probably backport some of the documentation changes based on that, for instance.

I'm also not sure if [`withProgress`](https://github.com/argoproj/argo-workflows/pull/13100/files#diff-086dbc62d3bcf73bbfb8c164e22ff70454137962f1f0e50ce0323c5cd4ab2ae4L460) and its usage should have been removed here then; it sounds like it perhaps should be modified similarly to use `WorkflowTaskResults`

dlemfh:
Hi team, I wanted to ask if you would be willing to consider re-allowing (opting-in to) the `patch pods` method for our edge-case?

We use https://admiralty.io/ along with argo-workflows, to dynamically allocate workflow pods across multiple kubernetes clusters.

image
More specifically we use the "cloud bursting" scenario, where
- Argo-workflows is installed only to the primary cluster. (Thus, workflows are submitted to the primary cluster.)
- Any pod created as part of a workflow may either run on the primary cluster or may be "bursted" to the secondary cluster, in which case
  - a "proxy pod" is stationed to a virtual node in the primary cluster
  - while the actual pod is cloned and scheduled to the secondary cluster to be run for real.
- Admiralty syncs the proxy & actual pods, ensuring that info like the pod status outcome is propagated back to the primary cluster.

This works for us before argo-workflows v3.6.0, because the outcome of a pod is self-contained within the pod as annotations/labels (added via pod patch), and this info is always synced back to the primary cluster for argo-workflows controller to be able to pick up.

Would it be possible for users like us to accept the security concern regarding `pod patch` and be allowed to opt-in to it?

Issue #12966, opened 2024-04-22
PDB not created when using mutex or semaphore and Workflow waiting on lock


agilgur5:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

This is a follow-up to my findings in https://github.com/argoproj/argo-workflows/pull/6356#discussion_r1574367137 / https://github.com/argoproj/argo-workflows/issues/10178#issuecomment-2068596407 / #12965 . This is technically a regression from 3.2.

When using a semaphore or mutex, if your Workflow cannot start due to waiting for a lock, any PDB on it will not be created.

PDB should be created regardless of usage of semaphore or mutex

### Version

v3.5.6, latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
TBD need to make a smaller repro than #10178, but that one will repro with the `parallelism` setting, multiple runs with the script, and by adding a PDB.
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

agilgur5:
Per https://github.com/argoproj/argo-workflows/pull/6356#discussion_r1574367137 and https://github.com/argoproj/argo-workflows/issues/10178#issuecomment-2068596407, the solution here is actually not entirely straightforward unfortunately (otherwise, I would have submitted a PR directly). A Workflow without a lock shouldn't be creating resources like a PDB.

That's not the worst thing if we were to do that, but it's not quite correct and affects latency when using synchronization as well. So ideally this needs a larger refactor

PR #13102, opened 2024-05-28
fix: don't fail workflow if PDB creation fails

Joibel:
### Motivation

It is likely that a busy kubernetes API will fail to create a PDB occasionally. This results in the whole workflow being marked as failed, when this is most likely something that will succeed on retry.

### Modifications

PDB creation only happens to workflows in state Unknown, so delay marking the workflow is running until after we know if PDB creation has succeeded.

If PDB creation fails then requeue the workflow and log the failure.

### Verification

Modified `createPDBResource()` to mostly fail (1 in 20 succeeded). Submitted multiple simultaneous copies of [examples/default-pdb-support.yaml](../blob/main/examples/default-pdb-support.yaml) modified for a longer sleep, and ensured they:
* All started eventually
* All started with their PDB having been created

This isn't suitable for a regression test that I can think of without hooking createPDBResource.

### Design

I don't like having to reorder code to make this work.
Ideally the workflow controller would act like an operator, formulating a list of k8s objects that should exist, and then reconcile that with what actually exists in an independent reconciler later. This is too big a change for this bug fix.

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -283,15 +283,15 @@ func (woc *wfOperationCtx) operate(ctx context.Context) {
 	}
 
 	if woc.wf.Status.Phase == wfv1.WorkflowUnknown {
-		woc.markWorkflowRunning(ctx)
-		setWfPodNamesAnnotation(woc.wf)
-
 		err := woc.createPDBResource(ctx)
 		if err != nil {
-			msg := fmt.Sprintf("Unable to create PDB resource for workflow, %s error: %s", woc.wf.Name, err)
-			woc.markWorkflowFailed(ctx, msg)
+			woc.log.WithError(err).WithField("workflow", woc.wf.Name).Error("PDB creation failed")
+			woc.requeue()
 			return
 		}
+
+		woc.markWorkflowRunning(ctx)
+		setWfPodNamesAnnotation(woc.wf)
 
 		woc.workflowDeadline = woc.getWorkflowDeadline()
 


agilgur5:
Related: #12965, #12966. Particularly with regard to reordering code; those are both similarly `Unknown` state only and very early in the `operate` function.

> and then reconcile that with what actually exists in an independent reconciler later.

This is potentially less efficient though, see also #12974 

> This isn't suitable for a regression test that I can think of without hooking createPDBResource.

We could potentially add a similar test that I mentioned in #12965 to `operator_concurrency_test.go`?

Joibel:
> Related: #12965, #12966. Particularly with regard to reordering code; those are both similarly `Unknown` state only and very early in the `operate` function.
> 
> > and then reconcile that with what actually exists in an independent reconciler later.
> 
> This is potentially less efficient though, see also #12974

I'd be expecting to reconcile against an Informer, not directly against the cluster. Slightly less efficient, yes, but more provably doing the right thing.

It'd also open up individual WorkflowTemplates owning resources like PVCs more easily than could currently be implemented.

Anyway, we're quite off topic for this review!

> > This isn't suitable for a regression test that I can think of without hooking createPDBResource.
> 
> We could potentially add a similar test that I mentioned in #12965 to `operator_concurrency_test.go`?

This fix isn't about concurrency, it's about an isolated workflow failing due to PDB `create` failing when there is pressure on etcd.

I tested it by hooking into `createPDBResource()` and making that deterministically fail, but that isn't a suitable strategy for a unit or e2e test I don't feel.

Have I missed something about how it could be tested?

agilgur5:
> I'd be expecting to reconcile against an Informer, not directly against the cluster. Slightly less efficient, yes, but more provably doing the right thing.

Ah, we'd need optimized Informers for these resources then. We may want to label Argo-specific resources as well for optimization, similar to https://github.com/argoproj/argo-workflows/pull/11855#discussion_r1556460508

> This fix isn't about concurrency, it's about an isolated workflow failing due to PDB `create` failing when there is pressure on etcd.

Oh, sorry I think I misinterpreted your PR description since you had:

> Submitted multiple simultaneous copies of [examples/default-pdb-support.yaml](../blob/main/examples/default-pdb-support.yaml) modified for a longer sleep, and ensured they:

So I thought that would induce failure with multiple, concurrent Workflows

> but that isn't a suitable strategy for a unit or e2e test I don't feel.

I would really like to figure out a way to test these load or race dependent bugs, since we're getting increasingly more of them (or, they may be some of few bugs left). I had been thinking of making a new test suite for them in the matrix, but still not sure how to go about implementation.

agilgur5 (approved):
LGTM, will have to figure out testing in the future

PR #10911, opened 2023-04-14
docs: Improve documentation on lifecycle hooks

Joibel:
In response to #8954, improve documentation

Update the LifecycleHook documentation to improve clarity on what should and should not work, and how they differ from exit handlers.

Improve the links to the examples by making them the `master` versions instead of an historic snapshot.

Fix the lifecycle hook example by naming the hooks not `exit`.

This is a documentation and examples only change.

PR #13106, opened 2024-05-28
docs: outputs from previous steps are _not_ available in hooks

oliverdain:
The documentation said you could use outputs from previous steps in a hook but not outputs from the current step/task. Experimentally the opposite seems to be true. See this Slack conversation for some details: https://cloud-native.slack.com/archives/C01QW9QSSSK/p1716585083642879

I'm not sure this new documentation is correct. Maybe there are situations where you can use other outputs? Also, note that the docs I updated were based on steps and in my use-case I was using a DAG.

### DIFF
--- a/docs/lifecyclehook.md
+++ b/docs/lifecyclehook.md
@@ -25,7 +25,7 @@ In other words, a `LifecycleHook` functions like an [exit handler](https://githu
 
 ## Unsupported conditions
 
-- [`outputs`](fields.md#outputs) are not usable since `LifecycleHook` executes during execution time and `outputs` are not produced until the step is completed. You can use outputs from previous steps, just not the one you're hooking into. If you'd like to use outputs create an exit handler instead - all the status variable are available there so you can still conditionally decide what to do.
+- [`outputs`](fields.md#outputs) from the step you are hooking into are available but no other outputs can be used.
 
 ## Notification use case
 


agilgur5:
This is possibly a bug, not a documentation mistake. I haven't looked at your issue in detail yet, but IIRC it should be possible.

agilgur5:
Also you may have noticed I changed your PR titles twice -- like commit subject best practices, a PR title should summarize _what_ it does. Further details would be in the body/description. Remember that these changes are visible in the repo history and the changelog.
Concretely, "what I think is a documentation bug" does not tell a reader _what_ the PR/commit does. That is information for the body/description. 
Also minor note, specifying "documentation" twice in a `docs:` prefixed commit is redundant -- conciseness is important for readers too, and PR titles/commit subjects are typically cut off after a certain point as well.

agilgur5:
> This is possibly a bug, not a documentation mistake. I haven't looked at your issue in detail yet, but IIRC it should be possible.

cc @Joibel who wrote part of this section of the docs in #10911

agilgur5:
See #13619 for a bug report on this. And that indeed has some indicators of a bug like _sometimes_ working /!non-determinism

agilgur5:
Closing this out as this is a bug bug (not a documentation mistake) and should be resolved by #13639

PR #13115, opened 2024-05-29
docs: link to a more up-to-date Grafana dashboard

agilgur5:
### Motivation

Link to a more up-to-date Grafana dashboard.
Previous: https://grafana.com/grafana/dashboards/13927-argoworkflow-metrics/, has not received an update in 2+ years (not since early 2022)
New: https://grafana.com/grafana/dashboards/20348-argo-workflows-metrics/, has received updates in 2024

Pipekit was working on this and I was wondering where their dashboard was and then I saw it in two [recent](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1715607073926439) [Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1716303716690499?thread_ts=1716303570.506539&cid=C01QW9QSSSK) threads.

### Modifications

- just update the link, nothing else

### Verification

`make docs` passes

### Notes to Reviewers / Future Work

- disclaimer that this is Pipekit's up-to-date fork of the community one previously linked
- ideally, we should maintain an official open-source dashboard somewhere, but in the interim, this is (much?) better than the old link
  - (also yes, source code for dashboards isn't the most readable thing, but we can work around that)
  - I imagine keeping it in this repo would be most straightforward, as changes to the metrics and dashboards can happen simultaneously, but we could also host in a separate community repo similar to the Helm charts

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -149,7 +149,7 @@ labels wherever it is emitted. This is how these metrics are "linked" as belongi
 
 ### Grafana Dashboard for Argo Controller Metrics
 
-Please see the [Argo Workflows metrics](https://grafana.com/grafana/dashboards/13927) Grafana dashboard.
+Please see the [Argo Workflows metrics](https://grafana.com/grafana/dashboards/20348-argo-workflows-metrics/) Grafana dashboard.
 
 ## Defining metrics
 


Joibel (approved):
LGTM, thanks

PR #13116, opened 2024-05-30
feat: skip sending workflow events (#13042)

### LINKED ISSUE #13042: Option to disable writing k8s events at workflow level
# Summary

Environment variable/config to stop writing k8s events

## Use Cases

When would you use this? When i want to stop etcd filling up

These parts of the code are writing k8s events which end up being stored in etcd:
https://github.com/argoproj/argo-workflows/blob/0bfaea64b88a4b4f0b79c389f63ea2ce326a747e/workflow/controller/operator.go#L2339-L2344
https://github.com/argoproj/argo-workflows/blob/0bfaea64b88a4b4f0b79c389f63ea2ce326a747e/workflow/controller/operator.go#L783

With a lot of workflows being churned through these events add a lot of storage pressure to etcd. At the same time i have not found these events useful as i can always look at logs in the controller pod

tooptoop4:
Fixes #13042


agilgur5:
Replaced by #13746

PR #13135, opened 2024-06-02
fix: only set ARGO_PROGRESS_FILE if self reporting progress (#13089)

tooptoop4:
Fixes point 5 of #13089


agilgur5:
Replaced by #13743

PR #13138, opened 2024-06-02
docs: remove Giscus integration

agilgur5:
Follow-up to https://github.com/argoproj/argo-workflows/pull/12360#issuecomment-1970762484 and a [previous Contributor Meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.9ogkxyklinsn) about that

### Motivation

Remove Giscus as it does not work well with versioned documentation and creates inefficient, hard-to-search and hard-to-answer questions
  - Giscus creates a separate page for each version of the docs, including `release-3.4`, `release-3.5`, `latest` and `stable`  <details>
     - each of those has its own URL, and that is what Giscus uses. Translations would amplify this as well.
     - that also means the old GH Pages docs still have their own Giscus pages, but are no longer used 😕
     - this makes for lots of duplicative discussion pages, which makes it harder to search and find information as well
     - it also sometimes gets the link **wrong**, linking to a different existing discussion to a page
       - this has only been a problem I noticed since versioned docs -- Giscus attempts a fuzzy search, which can be incorrect, especially for higher level pages

</details>
       
  - Giscus doesn't exactly build up a knowledge base of questions and answers, unlike issues and regular discussions and even Slack to an extent
    - Giscus creates GH Discussions that are not "Q&A", but "General". There can also be multiple questions within a single Giscus page (which is why changing to "Q&A" would not necessarily help).
      - <details><summary>This is difficult if not impossible to organize as such. You could have 12 separate threads within one page.</summary>
      
        - this also makes search unusable -- you'll get a result for a page with 12 threads in it and have to attempt to find which thread addresses your question
          - that page also has a generic, non-specific title (the URL). In general, search and non-search pages get clogged up by these non-specific titles, which makes search far less helpful
          - you can't just scroll to the bottom and expect to see an answer there as with an issue
          - and you can't expect a highlighted answer as with "Q&A"
          - you also can't close or lock the thread for spam or once it's been sufficiently answered, because it is always ongoing for a page
        - answering can be difficult for the same reasons too, so questions just get lost into the abyss
        - some users end up seeing all these and asking questions via Giscus instead of a more efficient & searchable method as well, making this an ever compounding problem
        - (tbh, I think all of this is a larger problem with GH Discussions _in general_, but that's a bit of a separate topic)
        
      </details>
      
    - It doesn't encourage searching your questions first either, so we get duplicative questions on Giscus with some frequency. For example: https://github.com/argoproj/argo-workflows/discussions/10320#discussioncomment-6272672, https://github.com/argoproj/argo-workflows/discussions/10320#discussioncomment-7837204, https://github.com/argoproj/argo-workflows/discussions/12451#discussioncomment-8006065. We can't necessarily even mark as duplicates because they are sub-threads.
    
  - Overall, Giscus is greatly inefficient, if not actively counter-productive, and causes duplicates on duplicates
    - As user support is one of the largest uses of time in the project, that is pretty important to optimize

### Modifications

- As such, remove the Giscus integration and replace with a footer suggestion to search GH Discussions and Slack, with links to each
  - Note: the GH Discussion link I used is slightly different from the default; this one searches open _and_ closed discussions - the default only searches open, but questions that have already been answered may very well be closed
  - This also roughly matches [the UI's help page](https://github.com/argoproj/argo-workflows/blob/85c88320c2f6678d26a1d468cd0ad84517ab4148/ui/src/app/help/help.tsx#L9) as well, so this improves consistency too

### Verification

`make docs` passes and the new footer looks as expected. Screenshots:

<details> <summary> Before: </summary>

Screenshot 2024-06-02 at 3 34 18 PM

</details>

<details> <summary> After: </summary>

Screenshot 2024-06-02 at 3 33 57 PM

</details>

### Future Work

Once this is done and backported to all versioned docs, I can slowly start to close out, organize, and lock old Giscus threads
1. Close out each discussion
1. Try to edit the initial/top comment of each Discussion with a summary listing each thread. This attempts to make them more searchable but is still not great and is _very_ tedious / requires significant manual effort (i.e. is unsustainable for maintainers if kept around)
1. Change the title to add prefix "[archived] "
  1. Attempt to make a more specific title if possible (if all threads are related to a certain topic)
1. Then try to answer any leftover questions inside of these that got lost to the abyss
1. Then lock the Discussion

Per my previous efforts in https://github.com/argoproj/argo-workflows/pull/12360#issuecomment-1970762484, there are like [~100+ of these](https://github.com/argoproj/argo-workflows/discussions?discussions_q=author%3Agiscus%5Bbot%5D+), so it will take time.

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -2,23 +2,11 @@
 {% block content %}
 {{ super() }}
 
-<!-- Giscus -->
-<h2 id="__comments">{{ lang.t("meta.comments") }}</h2>
+{# Footer #}
+<hr />
 
-<script src="https://giscus.app/client.js"
-        data-repo="argoproj/argo-workflows"
-        data-repo-id="MDEwOlJlcG9zaXRvcnkxMDA5ODI0NDk="
-        data-category="General"
-        data-category-id="MDE4OkRpc2N1c3Npb25DYXRlZ29yeTMyMTA3OTY1"
-        data-mapping="pathname"
-        data-reactions-enabled="1"
-        data-emit-metadata="0"
-        data-input-position="bottom"
-        data-theme="light"
-        data-lang="en"
-        data-loading="lazy"
-        crossorigin="anonymous"
-        async>
-</script>
+<h2>Have a question?</h2>
+
+Search on <a href="https://github.com/argoproj/argo-workflows/discussions?discussions_q=">GitHub Discussions</a> and <a href="https://argoproj.github.io/community/join-slack/">Slack</a>.
 
 {% endblock %}


jmeridth (approved):

terrytangyuan (approved):

terrytangyuan:
Agreed

agilgur5:
> backported to all versioned docs

Completed for `release-3.5` and `release-3.4`. So just `stable` (unfortunately the most important) is leftover now, as that's only updated when there's a new GH release.

> I can slowly start to close out, organize, and lock old Giscus threads

Can gradually work on all non-`stable` threads at least now

agilgur5:
> Can gradually work on all non-`stable` threads at least now

There's only a handful of these right now.

I've gotten through [around half](https://github.com/argoproj/argo-workflows/discussions?discussions_q=author%3Agiscus%5Bbot%5D+is%3Aclosed) of non-`stable` threads now, particularly the ones with fewer comments.
Note that for a few I closed them but left them [unlocked](https://github.com/argoproj/argo-workflows/discussions?discussions_q=author%3Agiscus%5Bbot%5D+is%3Aclosed+is%3Aunlocked) when there were some unanswered questions, but not very important ones. If I thought some questions were important, I might leave the discussion as open for now.

PR #13145, opened 2024-06-05
fix(build): correct `make` target deps for Go binaries

Joibel:
Fixes the makefile when editing files under certain circumstances.

### Motivation

@Adrien-D was editing the files using vscode on a Mac and finding that although kit was attempting a rebuild when his file was saved the result was not running.

This was tracked down to `make dist/argo` (he was working on the server component) reporting ` 'dist/argo' is up to date.`

The binary targets only depend upon the timestamps of the package subdirectories, not the timestamps of the `.go` files within.

### Modifications

This file changes the dependencies for the three golang binaries in the makefile to be the .go files under all the packages, rather than just the package directories

libc-dev doesn't make it into the final image, just the builder image. Because we now depend on the .go files we need to be able to rebuild `pkg/apis/workflow/v1alpha1/openapi_generated.go`, who's make target depends upon `/go/bin/openapi-gen` which needs libc-dev to build.

### Verification

See the last make in this for evidence of the problem, and the second for proof it works.
```
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ make dist/workflow-controller
GIT_COMMIT=01bf1cfc82834a5fc50bcba42cdacc0eb1fbcde1 GIT_BRANCH=HEAD GIT_TAG=untagged GIT_TREE_STATE=dirty RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=k3d-k3s-default DOCKER_DESKTOP=false K3D=true DOCKER_PUSH=false
RUN_MODE=local PROFILE=minimal AUTH_MODE=hybrid SECURE=false STATIC_FILES=false ALWAYS_OFFLOAD_NODE_STATUS=false UPPERIO_DB_DEBUG=0 LOG_LEVEL=debug NAMESPACED=true
make: 'dist/workflow-controller' is up to date.
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ touch workflow/controller/controller.go
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ make dist/workflow-controller
GIT_COMMIT=01bf1cfc82834a5fc50bcba42cdacc0eb1fbcde1 GIT_BRANCH=HEAD GIT_TAG=untagged GIT_TREE_STATE=dirty RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=k3d-k3s-default DOCKER_DESKTOP=false K3D=true DOCKER_PUSH=false
RUN_MODE=local PROFILE=minimal AUTH_MODE=hybrid SECURE=false STATIC_FILES=false ALWAYS_OFFLOAD_NODE_STATUS=false UPPERIO_DB_DEBUG=0 LOG_LEVEL=debug NAMESPACED=true
CGO_ENABLED=0 go build -gcflags '' -v -ldflags '-X github.com/argoproj/argo-workflows/v3.version=latest -X github.com/argoproj/argo-workflows/v3.buildDate=2024-06-05T13:10:23Z -X github.com/argoproj/argo-workflows/v3.gitCommit=01bf1cfc82834a5fc50bcba42cdacc0eb1fbcde1 -X github.com/argoproj/argo-workflows/v3.gitTreeState=dirty -X github.com/argoproj/argo-workflows/v3.gitTag=untagged -extldflags -static' -o dist/workflow-controller ./cmd/workflow-controller
github.com/argoproj/argo-workflows/v3/cmd/workflow-controller
gitvscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ git checkout HEAD~ -- Makefile
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ make dist/workflow-controller
GIT_COMMIT=01bf1cfc82834a5fc50bcba42cdacc0eb1fbcde1 GIT_BRANCH=HEAD GIT_TAG=untagged GIT_TREE_STATE=dirty RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=k3d-k3s-default DOCKER_DESKTOP=false K3D=true DOCKER_PUSH=false
RUN_MODE=local PROFILE=minimal AUTH_MODE=hybrid SECURE=false STATIC_FILES=false ALWAYS_OFFLOAD_NODE_STATUS=false UPPERIO_DB_DEBUG=0 LOG_LEVEL=debug NAMESPACED=true
make: 'dist/workflow-controller' is up to date.
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ touch workflow/controller/controller.go
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (01bf1cfc8) $ make dist/workflow-controller
GIT_COMMIT=01bf1cfc82834a5fc50bcba42cdacc0eb1fbcde1 GIT_BRANCH=HEAD GIT_TAG=untagged GIT_TREE_STATE=dirty RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=k3d-k3s-default DOCKER_DESKTOP=false K3D=true DOCKER_PUSH=false
RUN_MODE=local PROFILE=minimal AUTH_MODE=hybrid SECURE=false STATIC_FILES=false ALWAYS_OFFLOAD_NODE_STATUS=false UPPERIO_DB_DEBUG=0 LOG_LEVEL=debug NAMESPACED=true
make: 'dist/workflow-controller' is up to date.
```

This works well under native linux and under the .devcontainer docker.

If someone could verify the `xargs+find` invocation I use actually works under native Mac, that would be good. I know MacOS ships with some horrifically outdated unix tools.


terrytangyuan (approved):
Thanks! Great write-up in the description.

Makefile
```diff
@@ -118,9 +118,9 @@ ifndef $(GOPATH)
...
 endif
 
-ARGOEXEC_PKGS    := $(shell echo cmd/argoexec            && go list -f '{{ join .Deps "\n" }}' ./cmd/argoexec/            | grep 'argoproj/argo-workflows/v3/' | cut -c 39-)
-CLI_PKGS         := $(shell echo cmd/argo                && go list -f '{{ join .Deps "\n" }}' ./cmd/argo/                | grep 'argoproj/argo-workflows/v3/' | cut -c 39-)
-CONTROLLER_PKGS  := $(shell echo cmd/workflow-controller && go list -f '{{ join .Deps "\n" }}' ./cmd/workflow-controller/ | grep 'argoproj/argo-workflows/v3/' | cut -c 39-)
+ARGOEXEC_PKG_FILES    := $(shell echo cmd/argoexec            && go list -f '{{ join .Deps "\n" }}' ./cmd/argoexec/            | grep 'argoproj/argo-workflows/v3/' | cut -c 39- | xargs -I {} find {} -type f -name "*.go")
+CLI_PKG_FILES         := $(shell echo cmd/argo                && go list -f '{{ join .Deps "\n" }}' ./cmd/argo/                | grep 'argoproj/argo-workflows/v3/' | cut -c 39- | xargs -I {} find {} -type f -name "*.go")
+CONTROLLER_PKG_FILES  := $(shell echo cmd/workflow-controller && go list -f '{{ join .Deps "\n" }}' ./cmd/workflow-controller/ | grep 'argoproj/argo-workflows/v3/' | cut -c 39- | xargs -I {} find {} -type f -name "*.go")
```
terrytangyuan, line 123: Can you paste the outputs of these? We need to confirm this change is correct

terrytangyuan (changes requested):

Makefile
Joibel: Here we go. I hope you enjoy! 
[cli_pkg_files.txt](https://github.com/user-attachments/files/15737780/cli_pkg_files.txt)
[controller_pkg_files.txt](https://github.com/user-attachments/files/15737781/controller_pkg_files.txt)
[argoexec_pkg_files.txt](https://github.com/user-attachments/files/15737783/argoexec_pkg_files.txt)

Joibel:
During CI the generate go files (specifically `pkg/apis/workflow/v1alpha1/openapi_generated.go` gets generated without a "newer than bin/openapi-gen" timestamp. This I cannot reproduce locally, but causes it to get continuously regenerated as out of date. So I've added a `touch` to it's generation, which fixes the CI.

This isn't the first time github's CI has had odd timestamp behaviour (#13023)

Makefile
```diff
@@ -118,9 +118,9 @@ ifndef $(GOPATH)
 	export GOPATH
 endif
 
-ARGOEXEC_PKGS    := $(shell echo cmd/argoexec            && go list -f '{{ join .Deps "\n" }}' ./cmd/argoexec/            | grep 'argoproj/argo-workflows/v3/' | cut -c 39-)
-CLI_PKGS         := $(shell echo cmd/argo                && go list -f '{{ join .Deps "\n" }}' ./cmd/argo/                | grep 'argoproj/argo-workflows/v3/' | cut -c 39-)
-CONTROLLER_PKGS  := $(shell echo cmd/workflow-controller && go list -f '{{ join .Deps "\n" }}' ./cmd/workflow-controller/ | grep 'argoproj/argo-workflows/v3/' | cut -c 39-)
+ARGOEXEC_PKG_FILES    := $(shell echo cmd/argoexec            && go list -f '{{ join .Deps "\n" }}' ./cmd/argoexec/            | grep 'argoproj/argo-workflows/v3/' | cut -c 39- | xargs -I {} find {} -type f -name "*.go")
```
agilgur5, line 121: a solution that uses more `go list` and less of other commands would be preferable since it would be more intelligent as well as more portable cross-platform, but this works on my Mac

for instance, I got this working:
```sh
go list -f '{{ join .Deps "\n" }}' ./cmd/argoexec/ |  grep 'argoproj/argo-workflows/v3/' | xargs go list -f '{{ range $file := .GoFiles }}{{ print $.ImportPath "/" $file "\n" }}{{ end }}' | cut -c 39-
```

this produces a shorter list -- it ignores test files and only uses the files of the package, not all files in the dir. I think it's more precise, but I might be wrong on that one and maybe I'm missing something

agilgur5 (commented):
Ah I have had this issue before of the binaries failing to update, but hadn't looked into it as it was fairly simple to workaround. Thanks for figuring out the issue! That it was only using directories is hard to tell from a quick glance

I left some small suggestions below that might optimize this

Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apk update && apk add --no-cache \
     wget \
     curl \
     gcc \
+    libc-dev \
```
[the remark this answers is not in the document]
Joibel: The `openapi_generated.go` file depends upon `$(GOPATH)/bin/openapi-gen`. In the builder stage of image building `$(GOPATH)/bin/openapi-gen` does not exist on startup. `$(GOPATH)/bin/openapi-gen` is a buildable make target, but it relies on `libc-dev` to build.

Previously the built binaries did not depend upon the `.go` files, which is what this PR is mainly fixing. Now it does, and `openapi_generated.go` is seen, correctly, as out of date, because `$(GOPATH)/bin/openapi-gen` is not present, and it depends upon it. So the Makefile builds `$(GOPATH)/bin/openapi-gen` and then rebuilds `openapi_generated.go`. This all seems reasonable and correct. `$(GOPATH)/bin/openapi-gen` will always install the same version, it's pinned.

The `touch` is here because in github actions (only), the timestamp for `openapi_generated.go` isn't updated when it is built. I suspect github have some magical filesystem shenannigans going on, I have linked a previous bonkers touch required in githubland above (#13023). If this timestamp doesn't get updated then make (correctly) considers that .go file to be continuously out of date, and for reasons I didn't look into kit's invocation from `make start` which is running the controller + server for e2e testing in CI is continuously deciding these two components need rebuilding and restarting. This, understandably, makes CI horrifically unreliable. Running them through kit in CI is questionable, but in the interests of small commits, I'm not removing it here.

Makefile
Joibel: Yes, I agree.

Joibel: The outputs are [above](https://github.com/argoproj/argo-workflows/pull/13145#discussion_r1632865882) for diffing. The original list output was not `uniq` so the number of lines of diff was higher than expected.

Makefile
Joibel: @terrytangyuan please could you take another look here, you're blocking this one.

terrytangyuan (approved):

Dockerfile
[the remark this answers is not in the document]
agilgur5: Thanks for clarifying this.

> In the builder stage of image building `$(GOPATH)/bin/openapi-gen` does not exist on startup

Ah I see, and the `builder` stage is also used during the `argoexec-build` below which has the `make dist/argoexec`

> The `touch` is here because in github actions (only), the timestamp for `openapi_generated.go` isn't updated when it is built. I suspect github have some magical filesystem shenannigans going on, I have linked a previous bonkers touch required in githubland above (#13023)

I reviewed that one so I'm aware of the weird filesystem issue. I was thinking that, in the optimistic case, the OpenAPI generated files already exist and so it wouldn't need to rebuild

> because `$(GOPATH)/bin/openapi-gen` is not present, and it depends upon it

This was the part I missed. There's potentially a way of avoiding this in the optimistic case (which is the one that the release build works on since all code is already committed then and passed CI with generated files), but in the other cases, it might actually need to build it, so I suppose it's probably not worth working around as such and that could be more buggy.

PR #13155, opened 2024-06-08
fix(release): set `$DOCKER_CONFIG` if unset

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/12791#issuecomment-2155929515

### Motivation

Per the linked comment above, it seems like `Azure/docker-login` may have set this env var and so when that was changed to `docker/login-action`, this line errored out

### Modifications

- set a default for `$DOCKER_CONFIG` in case it is unset
    - Per [Docker's docs](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables), the default is `~/.docker/config.json` (can also confirm locally)

### Verification

This new line runs locally for me:
```
❯  : ${DOCKER_CONFIG:=~/.docker}; echo "$DOCKER_CONFIG"
/<redacted>/.docker
```
and `$DOCKER_CONFIG` is unset without it

But testing the entire GHA workflow is unfortunately very non-trivial 😕

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -172,6 +172,7 @@ jobs:
           COSIGN_PRIVATE_KEY: ${{secrets.COSIGN_PRIVATE_KEY}}
           COSIGN_PASSWORD: ${{secrets.COSIGN_PASSWORD}}
         run: |
+          : ${DOCKER_CONFIG:=~/.docker}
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

agilgur5:
[All green](https://github.com/argoproj/argo-workflows/actions/runs/9428611708/job/25974609373) on `main` now

PR #13156, opened 2024-06-09
docs: example workflow - force-delete-pods.yaml

jrbe228:
### Motivation

Workaround for stuck pods described [here](https://github.com/argoproj/argo-workflows/discussions/13146). May be generally useful for non-responsive pod clean up. 

### Modifications

Force deletes all pods in the argo namespace except the argo-server, workflow-controller, and the pod running this workflow.

### Verification

Tested against pods in various Running, Terminating, Completed, Error states.


jrbe228:
Any help on the Codegen linting step would be appreciated

PR #2670, opened 2020-04-13
fix: Make ClusterWorkflowTemplate optional for namespaced Installation

sarabala1979:
Checklist:

* [ ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [ ] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  #2664 
* [ ] I've signed the CLA.
* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ ] My builds are green. Try syncing with master if they are not. 
* [ ] My organization is added to USERS.md.

PR #13160, opened 2024-06-10
feat: Add name filters for workflow list (exact/prefix/pattern)

### LINKED ISSUE #12161: UI: 3.5 missing name + name prefix filters for archived workflows

Adrien-D:
Fixes #12161

### Motivation

We used to have name/prefix filtering for archived workflows before unifying APIs. We need to implement it back.

### Modifications

API :
- `Name` filtering was already implemented
- I plugged back the `NamePrefix` to make it work
- I added `NamePattern` the same way prefix is implemented

UI :
- As its three way of filtering, and no reason to have those three at the same time, I made UX choice with a picker on the label. This is inspired from grafana's UI.

Screenshot 2024-06-10 at 14 40 15
Screenshot 2024-06-10 at 14 40 22

### Verification

I tested the feature locally and I added tests for the API.


ui/src/app/shared/utils.ts
```diff
@@ -152,12 +152,12 @@ export const Utils = {
...
         }
-        if (filter.namePrefix) {
-            queryParams.push(`namePrefix=${filter.namePrefix}`);
-        }
         if (filter.namePattern) {
             queryParams.push(`namePattern=${filter.namePattern}`);
         }
+        if (filter.namePrefix) {
+            queryParams.push(`namePrefix=${filter.namePrefix}`);
+        }
```
agilgur5, line 160: let's not change the ordering unnecessarily

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
```diff
@@ -11,7 +11,10 @@ import {NamespaceFilter} from '../../../shared/components/namespace-filter';
 import {TagsInput} from '../../../shared/components/tags-input/tags-input';
 import {services} from '../../../shared/services';
 
+import {InputFilter} from '../../../shared/components/input-filter';
 import './workflow-filters.scss';
+import {DropDown} from '../../../shared/components/dropdown/dropdown';
+import classNames from 'classnames';
```
agilgur5, line 17: please follow the existing import order conventions: external imports at the top, then external CSS, then internal imports, then internal CSS (also yes an ESLint or `prettier` plugin would be great)

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
```diff
@@ -57,13 +90,45 @@ export function WorkflowFilters(props: WorkflowFilterProps) {
         return results;
     }, [props.workflows, props.phaseItems]);
 
+    const handleNameFilterChange = (item: {title: string; id: string}) => {
```
agilgur5, line 93: please prefer named functions as they are better for debugging/tracing, as per #12062 et al (and the non-memoized functions in this same component)

```suggestion
    function handleNameFilterChange(item: {title: string, id: string}) {
```

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
```diff
@@ -57,13 +90,45 @@ export function WorkflowFilters(props: WorkflowFilterProps) {
...
+                                    <span>{item.title}</span>
+                                </li>
+                            ))}
+                        </ul>
+                    </DropDown>
+                    {nameFilter.id === 'namePrefix' ? <InputFilter value={props.namePrefix} name='wfNamePrefix' onChange={props.setNamePrefix} placeholder='Search...' /> : null}
+                    {nameFilter.id === 'namePattern' ? (
+                        <InputFilter value={props.namePattern} name='wfNamePattern' onChange={props.setNamePattern} placeholder='Search...' />
+                    ) : null}
+                    {nameFilter.id === 'name' ? <InputFilter value={props.name} name='wfName' onChange={props.setName} placeholder='Search...' /> : null}
```
agilgur5, line 130: this can be simplified and made neater by having an `if` statement above and changing the `value` and `onChange` based on that

also placeholders [should show example values](https://ux.stackexchange.com/a/131591/86443)

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
```diff
@@ -57,13 +90,45 @@ export function WorkflowFilters(props: WorkflowFilterProps) {
         return results;
     }, [props.workflows, props.phaseItems]);
 
+    const handleNameFilterChange = (item: {title: string; id: string}) => {
+        props.setNamePrefix('');
+        props.setNamePattern('');
+        props.setName('');
```
agilgur5, line 96: I actually wouldn't suggest clearing these out necessarily; I don't think that's necessarily ideal UX.
If I did an exact search and didn't find anything, I might want to change to a prefix or contains search using the same query

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -212,6 +244,12 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
                                 setFinishedBefore(date);
                                 clearSelectedWorkflows(); // date filters are client-side, but clear similar to the server-side ones for consistency
                             }}
+                            name={name}
+                            setName={setName}
+                            namePrefix={namePrefix}
+                            setNamePrefix={setNamePrefix}
+                            namePattern={namePattern}
+                            setNamePattern={setNamePattern}
```
agilgur5, line 252: I don't think this is actually necessary -- as you wrote, only 1 will be set at a time and they are all strings. We can just keep state for a single variable and query differently depending on what is chosen. E.g. `nameValue` or `nameInput`

pkg/apiclient/workflow/workflow.proto
```diff
@@ -30,9 +30,11 @@ message WorkflowGetRequest {
 
 message WorkflowListRequest {
   string namespace = 1;
-  k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 2;
+  string namePrefix = 2;
+  string namePattern = 3;
```
agilgur5, line 34: let's add some comments to describe these. the comments should make it to the Swagger and the API docs are sorely lacking descriptions

ui/src/app/shared/utils.ts
```diff
@@ -136,7 +136,7 @@ export const Utils = {
         resourceVersion?: string;
     }) {
         const queryParams: string[] = [];
-        const fieldSelector = this.fieldSelectorParams(filter.namespace, filter.name, filter.createdAfter, filter.finishedBefore);
+        const fieldSelector = this.fieldSelectorParams(filter.namespace, filter.name, filter.createdAfter, filter.finishedBefore, filter.namePrefix, filter.namePattern);
```
agilgur5, line 139: hmm the last two are not exactly field selectors (they are not valid k8s field selectors)

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
agilgur5: actually the below comment will simplify this even more

server/artifacts/artifact_server.go
```diff
@@ -504,7 +504,7 @@ func (a *ArtifactServer) getWorkflowAndValidate(ctx context.Context, namespace s
 }
 
 func (a *ArtifactServer) validateAccess(ctx context.Context, wf *wfv1.Workflow) error {
-	allowed, err := auth.CanI(ctx, "get", "workflows", wf.Namespace, wf.Name)
+	allowed, err := auth.CanI(ctx, "get", "workflows", wf.Namespace)
```
agilgur5, line 507: Maybe you misinterpreted my [previous comment on the issue](https://github.com/argoproj/argo-workflows/issues/12161#issuecomment-2155862968) -- I was saying that `name` should actually be used in `auth.CanI` as it seemed to have been _unintentionally_ removed in [this line](https://github.com/argoproj/argo-workflows/pull/2670/files#diff-10789fdfe9fcd47b01b89c63b756fa92802e56ce0cd545b4df4bce1b2f2e8a07L21) of #2670.

If a `name` is specified, it should be part of the query, which I believe will work with `resourceNames`

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
agilgur5: this helper will also no longer be needed after below comment

agilgur5 (changes requested):
Left several suggestions, simplifications, etc in-line below.

Might have a few more after that, particularly on the UI component and CSS changes that I did not check closely on this pass

pkg/apiclient/workflow/workflow.proto
```diff
@@ -30,9 +30,11 @@ message WorkflowGetRequest {
...
+  k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 4;
   // Fields to be included or excluded in the response. e.g. "items.spec,items.status.phase", "-items.status.nodes"
-  string fields = 3;
+  string fields = 5;
```
jiachengxu, line 37: Changing the tag number of the existing fields might lead to some issues of decoding of current existing client, as mentioned in the [proto 3 language guide](https://protobuf.dev/programming-guides/proto3/#updating):
> “Changing” the field number is equivalent to deleting the field and adding a new field with the same type.

and here are [consequences of reusing field numbers](https://protobuf.dev/programming-guides/proto3/#consequences), so I think we might just simply add the new fields without changing the existing ones:

```
message WorkflowListRequest {
  string namespace = 1;
  k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 2;
  string fields = 3;
  string namePrefix = 4;
  string namePattern = 5;
```

jiachengxu (changes requested):

persist/sqldb/selector.go
```diff
@@ -44,6 +44,9 @@ func BuildWorkflowSelector(in string, inArgs []any, tableName, labelTableName st
 	if options.NamePrefix != "" {
 		clauses = append(clauses, db.Raw("name like ?", options.NamePrefix+"%"))
 	}
+	if options.NamePattern != "" {
+		clauses = append(clauses, db.Raw("name like ?", "%"+options.NamePattern+"%"))
```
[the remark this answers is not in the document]
Adrien-D: @agilgur5 yes then I saw that this is how its called for workflows templates, from this PR https://github.com/argoproj/argo-workflows/pull/11684. That's why I changed to `contains` to `pattern` for consistency. But I agree with you, I prefer `contains` as well
Should I change ?

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
Adrien-D: Yes we could work on that, I was assuming that Eslint was already reordering for me 🙃

server/artifacts/artifact_server.go
Adrien-D: Oh yes sorry, I misinterpreted what you were saying, I'm reverting those changes !

Adrien-D:
I fixed my PR with your suggestions. Please note that I updated `argo-ui` package to last merged commit to be able to remove `autoHighlight` ([one of my other fixed issue](https://github.com/argoproj/argo-ui/issues/559)) on this particular autocomplete component

agilgur5:
> Please note that I updated `argo-ui` package to last merged commit to be able to remove `autoHighlight` ([one of my other fixed issue](https://github.com/argoproj/argo-ui/issues/559)) on this particular autocomplete component

Mmm I wouldn't include that as part of this PR since it is independent. Like we'd probably want to cherry-pick the `argo-ui` update/fixes into 3.4, which this PR will ofc not make it into.

Also there isn't an autocomplete component used in this PR?

Adrien-D:
> Also there isn't an autocomplete component used in this PR?
There is, it's the component behind `InputFilter` used for filtering 

But I get your point, we can merge this PR without it and I'll make two other PR to update the `argo-ui` package, and fix the behavior here

pkg/apiclient/workflow/workflow.proto
```diff
@@ -33,6 +33,10 @@ message WorkflowListRequest {
   k8s.io.apimachinery.pkg.apis.meta.v1.ListOptions listOptions = 2;
   // Fields to be included or excluded in the response. e.g. "items.spec,items.status.phase", "-items.status.nodes"
   string fields = 3;
+  // Return workflows having this name prefix
+  string namePrefix = 4;
+  // Return workflows with name containing this value
+  string namePattern = 5;
```
agilgur5, line 39: I'm wondering if we should make the API match the UI actually; `name` as a string and `nameFilter` as an email that defaults to `"Exact"`

Right now, it's not clear that these are mutually exclusive. We could add that in the comments, but having it only allow one at a time feels much more explicit

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
```diff
@@ -57,13 +83,44 @@ export function WorkflowFilters(props: WorkflowFilterProps) {
...
+                                    <span>{item.title}</span>
+                                </li>
+                            ))}
+                        </ul>
+                    </DropDown>
+                    <InputFilter
+                        value={props.nameFilter.value}
+                        name='wfNameFilter'
+                        onChange={value => props.setNameFilter({value, type: props.nameFilter.type})}
+                        placeholder='Search...'
```
agilgur5, line 120: See my [previous comment](https://github.com/argoproj/argo-workflows/pull/13160#discussion_r1633500429) about placeholders [being examples](https://ux.stackexchange.com/a/131591/86443)

agilgur5 (commented):
A few more comments on the filters below.

I haven't had time to take a look at the "contains" wording or CSS since I'm on vacation for the rest of the week and not at my computer

pkg/apiclient/workflow/workflow.proto
Adrien-D: Yeah I like that idea, so we could have `nameFilter` as `Exact` | `Contains` | `Prefix`, and then `nameValue` as `string`

I'm not familiar with the use case, but does it make more sense to have `Exact` as default instead of `Contains` ?

agilgur5: `Exact` as default would be backward compatible, since previously it was the only usage of `name`

persist/sqldb/selector.go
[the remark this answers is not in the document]
agilgur5: I think this is no longer relevant now due to https://github.com/argoproj/argo-workflows/pull/13160#discussion_r1638758194.

But yes, I'd do `contains` and then modify the code from #11684 to match. The PR has a pretty significant performance issue as-is (https://github.com/argoproj/argo-workflows/pull/11684#discussion_r1469090044) so it needs a re-write anyway, potentially to use the new SQLite DB for WorkflowTemplates as well.
I linked to here in [a comment there](https://github.com/argoproj/argo-workflows/pull/11684#discussion_r1649789718) about `contains` for future reference

tooptoop4 (approved):

Joibel (approved):

ui/src/app/workflows/components/workflow-filters/workflow-filters.tsx
agilgur5: I had to manually fix this lots of these issues in https://github.com/argoproj/argo-workflows/pull/13813 as they are partly unsortable by tooling since they were placed after a side effect import

PR #13163, opened 2024-06-10
fix: process metrics later in `executeTemplate`. Fixes #13162

### LINKED ISSUE #13162: Metrics Error: failure metric for workflow status not reporting correctly if in WorkflowTemplate
### Pre-requisites

Included failure metric with condition `when: {{status}} == Failed` in WorkflowTemplate. Expected metric to appear when workflow referencing this template failed however this is not the case. Works when metric is for success. Failure metric works if it is on workflow level instead of template.

### Version

latest

```YAML
kind: WorkflowTemplate
metadata:
  name: dag-task-test
spec:
  entrypoint: dag-task
  templates:
  - name: dag-task
    metrics:
        prometheus:
        - counter:
            value: "1"
          # help: DAG Task Workflow succeeded
          # name: dagtask_workflow_success
          # when: '{{status}} == Succeeded'
          help: DAG Task Workflow failed
          name: dagtask_workflow_failure
          when: '{{status}} == Failed'
    dag:
      tasks:
      - name: TEST-ONE
        template: echo
        arguments:
          parameters: 
            - name: message
              value: "console output-->TEST-{{item.command}}"
            - name: tag
              value: "{{item.tag}}"
        withItems:
          - { tag: TEST-ONE-A, command: ONE-A }
          - { tag: TEST-ONE-B, command: ONE-B }

      - name: TEST-TWO
        template: echo
        arguments:
          parameters: 
            - name: message
              value: "console output-->TEST-{{item.command}}"
            - name: tag
              value: "{{item.tag}}"
        withItems:
          - { tag: TEST-TWO-A, command: TWO-A }
          - { tag: TEST-TWO-B, command: TWO-B }

  - name: echo
    inputs:
      parameters:
      - name: message
      - name: tag
    container:
      image: alpine:3.7
      # command: [echo, "{{inputs.parameters.message}}"]
      # to fail it
      command: ["{{inputs.parameters.message}}"]
----
kind: Workflow
metadata:
  generateName: dag-task
spec:
  entrypoint: whalesay
  templates:
    - name: whalesay
      steps:
        - - name: hello-world
            templateRef:
              name: dag-task-test
              template: dag-task
```

```text
```

```text
```

dpadhiar:
Fixes #13162 

### Motivation

Failure metric when defined in WorkflowTemplate wasn't being reported. 

### Modifications

Changed where the `processedMetrics` are checked during reconciliation such that the current `Node` shows as Failed and the metric is processed correctly.

### Verification

Ran example workflow from initial issue to test that metric is now showing.


agilgur5 (changes requested):
> Ran example workflow from initial issue to test that metric is now showing.

Please do add a regression test for this

dpadhiar:
> > Ran example workflow from initial issue to test that metric is now showing.
> 
> Please do add a regression test for this. Otherwise, I'm not an expert in the templating part of the codebase so I'll let others review

Will do, can add a test case for using the metrics in a template.

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
...
+  entrypoint: dag-task
+  templates:
+  - name: dag-task
+    metrics:
+        prometheus:
+        - counter:
+            value: "1"
+          # help: DAG Task Workflow succeeded
+          # name: dagtask_workflow_success
+          # when: '{{status}} == Succeeded'
```
agilgur5, line 15: remove test comments
```suggestion
```

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: dag-task-test
```
agilgur5, line 4:
```suggestion
  name: template-failed-metric
```

by convention, these have the same name as their filename

test/e2e/metrics_test.go
```diff
@@ -77,6 +77,25 @@ func (s *MetricsSuite) TestRetryMetrics() {
 		})
 }
 
+func (s *MetricsSuite) TestFailedMetric() {
+	s.Given().
+		WorkflowTemplate(`@testdata/template-failed-metric.yaml`).
+		Workflow(`@testdata/wf-failed-metric.yaml`).
```
agilgur5, line 83: is a `WorkflowTemplate` necessary to reproduce this? Or could you just run it as a plain Workflow?

```suggestion
		Workflow(`@testdata/template-failed-metric.yaml`).
```

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
...
+  - name: echo
+    inputs:
+      parameters:
+      - name: message
+      - name: tag
+    container:
+      image: alpine:3.7
+      # command: [echo, "{{inputs.parameters.message}}"]
+      # to fail it
+      command: ["{{inputs.parameters.message}}"]
```
agilgur5, line 54: let's not leave in dead code / unnecessary test comments. 

also did you want to have this?
```suggestion
      command: [echo]
      args: ["{{inputs.parameters.message}}"]
```

because otherwise none of the messages are actual commands?

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
...
+        template: echo
+        arguments:
+          parameters: 
+            - name: message
+              value: "console output-->TEST-{{item.command}}"
+            - name: tag
+              value: "{{item.tag}}"
+        withItems:
+          - { tag: TEST-TWO-A, command: TWO-A }
+          - { tag: TEST-TWO-B, command: TWO-B }
```
agilgur5, line 43:
```suggestion
```

I don't this is necessary, is it? It seems unnecessarily duplicative

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
...
+  - name: dag-task
+    metrics:
+        prometheus:
+        - counter:
+            value: "1"
+          # help: DAG Task Workflow succeeded
+          # name: dagtask_workflow_success
+          # when: '{{status}} == Succeeded'
+          help: DAG Task Workflow failed
+          name: dagtask_workflow_failure
```
agilgur5, line 17:
```suggestion
          help: Task failed
          name: task_failure
```

use less confusing and simpler naming/help text

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
...
+        template: echo
+        arguments:
+          parameters: 
+            - name: message
+              value: "console output-->TEST-{{item.command}}"
+            - name: tag
+              value: "{{item.tag}}"
+        withItems:
+          - { tag: TEST-ONE-A, command: ONE-A }
+          - { tag: TEST-ONE-B, command: ONE-B }
```
agilgur5, line 31:
```suggestion
      - name: test
        template: echo
        arguments:
          parameters: 
            - name: message
              value: "{{item.message}}"
        withItems:
          - { message: testA }
          - { message: testB }
```

I don't think the "tag" is necessary, and the "command" looks like a misnomer to me.

In fact I'm not sure you really need a loop at all:
```suggestion
      - name: test
        template: echo
        arguments:
          parameters: 
            - name: message
              value: "test"
```

test/e2e/testdata/template-failed-metric.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: dag-task-test
+spec:
+  entrypoint: dag-task
+  templates:
+  - name: dag-task
```
agilgur5, line 8:
```suggestion
  entrypoint: dag
  templates:
  - name: dag
```

it's a DAG, not a task, so make the naming clearer

test/e2e/testdata/template-failed-metric.yaml
agilgur5: Although I'd recommend a more specific name for both the file and the `name`:

```suggestion
  name: template-status-failed-conditional-metric
```

agilgur5 (commented):
Correct me if I'm wrong, but it looks to me like the test case can be quite heavily simplified. We should keep tests as concise and isolated as possible, using only the features we are trying to test

There are also leftover test comments in it that shouldn't make it into the codebase.
And some style issues

dpadhiar:
> Correct me if I'm wrong, but it looks to me like the test case can be quite heavily simplified. We should keep tests as concise and isolated as possible, using only the features we are trying to test
> 
> There are also leftover test comments in it that shouldn't make it into the codebase. And some style issues

You are right, I did rush a bit to push the PR in (customer issue at work) so let me go ahead and clean it up / address your comments today.

test/e2e/metrics_test.go
dpadhiar: The initial issue was found when using a `WorkflowTemplate`. If you try a metric using `{{status}} == Failed` on the Workflow level, it will still work. Thus this fix is for when the metric is found at the template level.

test/e2e/testdata/template-failed-metric.yaml
dpadhiar: Can remove, this was needed when I was testing changes out manually.

test/e2e/testdata/template-failed-metric.yaml
dpadhiar: This workflow just comes from the repo itself, it's just a DAG with a metric. It would be better to simplify it like this, will do.

test/e2e/metrics_test.go
dpadhiar: I see what you mean now. I could make that change if you'd want (or make it another case)

agilgur5: The tests conventionally use plain Workflows unless a `WorkflowTemplate` is specifically necessary, so I would change to match that

workflow/controller/operator.go
```diff
@@ -2270,6 +2247,29 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
...
+		// In this check, a completed node may or may not have existed prior to this execution. If it did exist, ensure that it wasn't
+		// completed before this execution. If it did not exist prior, then we can infer that it was completed during this execution.
+		// The statement "(!ok || !prevNodeStatus.Fulfilled())" checks for this behavior and represents the material conditional
+		// "ok -> !prevNodeStatus.Fulfilled()" (https://en.wikipedia.org/wiki/Material_conditional)
+		if prevNodeStatus, ok := woc.preExecutionNodePhases[node.ID]; (!ok || !prevNodeStatus.Fulfilled()) && node.Fulfilled() {
+			localScope, realTimeScope := woc.prepareMetricScope(node)
+			woc.computeMetrics(processedTmpl.Metrics.Prometheus, localScope, realTimeScope, false)
+		}
+	}
+
```
juliev0, line 2272: Can you clarify how moving the code helped? What is it about the lines above this code that they needed to be executed first?

test/e2e/metrics_test.go
dpadhiar: Now updated, please re-review when you can, thanks!

test/e2e/testdata/wf-template-status-failed-conditional-metric.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: wf-failed-conditional-metric-
```
agilgur5, line 4:
```suggestion
  generateName: template-status-failed-conditional-metric-
```

again this should match the filename

test/e2e/metrics_test.go
```diff
@@ -94,6 +94,23 @@ func (s *MetricsSuite) TestDAGMetrics() {
 		})
 }
 
+func (s *MetricsSuite) TestFailedMetric() {
+	s.Given().
+		Workflow(`@testdata/wf-template-status-failed-conditional-metric.yaml`).
```
agilgur5, line 99:
```suggestion
		Workflow(`@testdata/template-status-failed-conditional-metric.yaml`).
```

the `wf-` in the name is not needed anymore

test/e2e/metrics_test.go
agilgur5: Small follow-up [below](https://github.com/argoproj/argo-workflows/pull/13163#discussion_r1643076506)

agilgur5 (approved):
Tests LGTM. Will let others review the rest

sarabala1979 (approved):
LGTM Thanks @dpadhiar. It is a good catch.
Thanks @agilgur5 and @juliev0 to review that.

PR #13165, opened 2024-06-11
fix: only evaluate retry expression for fail/error node. Fixes #13058

### LINKED ISSUE #13058: When workflow level retry expression evaluated False, workflow message is always `retryStrategy.expression evaluated to false` regardless of actual failure
### Pre-requisites

When workflow level retryExpression is used.
regardless of workflow succeed or failure,
the message at workflow level is always `retryStrategy.expression evaluated to false`

when no expression is used
- workflow succeed -> no workflow message
- workflow fail -> `no more retry left`

when expression is used
- workflow succeed -> `retryStrategy.expression evaluated to false`
- workflow fail (expression evaluated False) -> `retryStrategy.expression evaluated to false`
- workflow fail (expression evaluated True) -> `no more retry left`

This is problematic if one relies on the workflow message for metrc, analytics or debug.
We expect when expression is used, the workflow level message follows the same pattern as normal retry

### Version

v3.5.2, latest (https://github.com/argoproj/argo-workflows/commit/71f1d860b2c665ad87e8c313c4f865b829d07626)

```YAML
kind: Workflow
metadata:
  generateName: dag-nested-
  namespace: argo
spec:
  templates:
    - name: echo
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: alpine:3.7
        command:
          - exit
          - '1'
        resources: {}
    - name: nested-diamond
      outputs: {}
      metadata: {}
      dag:
        tasks:
          - name: A
            template: echo
          - name: B
            template: echo
            depends: A
          - name: C
            template: echo
            depends: A
          - name: D
            template: echo
            depends: B && C
  entrypoint: nested-diamond
  arguments: {}
  retryStrategy:
    retryPolicy: OnError
    expression: >-
      lastRetry.status == 'Errored'
```

```text
```

```text
```

tczhao:
Fixes #13058 

### Motivation

Evaluate retry expression after evaluating lastchildstatus,
- so that successful workflow/node doesn't receive `retryStrategy.expression evaluated to false` in the message field

This also aligns with
https://github.com/argoproj/argo-workflows/blob/6c8393fa392557637e9a23faa9a38696acaba207/workflow/controller/operator.go#L1091
- retry only intended to work with nodes that are failed or errored

### Verification

- add unit test, fail
- add operator change, succeed
- workflow run verify `retryStrategy.expression evaluated to false` doesn't show up on successful workflow

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1049,18 +1049,6 @@ func (woc *wfOperationCtx) processNodeRetries(node *wfv1.NodeStatus, retryStrate
 		node = woc.markNodePhase(node.Name, node.Phase, "")
 	}
 
-	if retryStrategy.Expression != "" && len(childNodeIds) > 0 {
-		localScope := buildRetryStrategyLocalScope(node, woc.wf.Status.Nodes)
-		scope := env.GetFuncMap(localScope)
-		shouldContinue, err := argoexpr.EvalBool(retryStrategy.Expression, scope)
-		if err != nil {
-			return nil, false, err
-		}
-		if !shouldContinue && lastChildNode.Fulfilled() {
-			return woc.markNodePhase(node.Name, lastChildNode.Phase, "retryStrategy.expression evaluated to false"), true, nil
-		}
-	}
-
 	var retryOnFailed bool
 	var retryOnError bool
 	switch retryStrategy.RetryPolicyActual() {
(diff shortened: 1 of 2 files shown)


workflow/controller/operator.go
```diff
@@ -1001,6 +989,18 @@ func (woc *wfOperationCtx) processNodeRetries(node *wfv1.NodeStatus, retryStrate
 		return woc.markNodePhase(node.Name, lastChildNode.Phase, message), true, nil
 	}
 
+	if retryStrategy.Expression != "" && len(childNodeIds) > 0 {
```
Joibel, line 992: I was expecting this to move even further down the code or be refactored in some other way, so that we checked `retryStrategy` first.

For example - in the event you have a retryStrategy of `OnFailure`, but the node is in error, we'll still evaluate the expression.

What do you think?

tczhao: So instead of retry on [`Supplied Policy AND Expression`](https://argo-workflows.readthedocs.io/en/latest/retries/#retry-policies),
We want to refactor this into `Supplied Policy OR Expression`, is this right? I think this is the way to go

Joibel: No, I don't want to change it to `OR`. This is would be a non-backwards compatible change, and require version 4 of workflows.

Your original problem is that the expression is evaluated even if the policy wouldn't have allowed a retry. This is still the case, as far as I can tell, but in a more limited scenario.

I think the expression should only be evaluated if the policy would allow a retry.

tczhao: Make sense, updated

Joibel (changes requested):
Can we add an extra test to show we don't get a `retryStrategy.expression` evaluation when the retryPolicy will not retry.

Otherwise, I've had a play with it, and it does what I'd hoped it would do, thank you.

Joibel (approved):

Issue #12572, opened 2024-01-25
Pod was rejected: The node had condition: [DiskPressure]. 


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issue exists when I tested with `:latest`
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

i got this error "Pod was rejected: The node had condition: [DiskPressure]. " twice out of 100000 different workflow runs

can this be treated as a transient error and autoretried by the controller?

### Version

3.4.11

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}

level=info msg=\"node changed\" namespace=auth new.message=\"Pod was rejected: The node had condition: [DiskPressure]. \" new.phase=Failed new.progress=0/1 nodeID=redactwf-815878683 old.message= old.phase=Pending old.progress=0/1 workflow=redactwf"
level=info msg=\"Pod failed: Pod was rejected: The node had condition: [DiskPressure]. \" displayName=\"redact(0)\" namespace=auth pod=redactwf-redact-815878683 templateName=redact workflow=redactwf"
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

agilgur5:
> can this be treated as a transient error and autoretried by the controller?

You can set the env var `TRANSIENT_ERROR_PATTERN` to add additional patterns to be treated as transient.

Not sure if there would be built-in support for this per https://github.com/argoproj/argo-workflows/pull/12567#pullrequestreview-1841581831, cc @terrytangyuan . It is a node error, so maybe

Joibel:
I am unsure if this can be covered under transient error handling.

My understanding of this:
Transient error handling is where we attempt a network call, and the result of the call is one that if we retry it later on it may succeed.
What we have in this case is a successful call to create a pod in the cluster. The pod has been created, in that the kubernetes object is there. Then the cluster has either failed to make a running pod on a node, or has created it and then it has been evicted.
[...]

PR #13180, opened 2024-06-13
feat: parse exit code when Outputs is not populated

tooptoop4:
Relates to https://github.com/argoproj/argo-workflows/issues/12572#issuecomment-2162093707
and https://github.com/argoproj/argo-workflows/blame/465c7b6d6abd06a36165955d7fd01d9db2b6a2d4/workflow/controller/operator.go#L1621
Seems more of a workaround than addressing real root cause

cc: @tczhao @shuangkun @toyamagu-2021


agilgur5:
Seems like this was superseded by #13228

PR #13016, opened 2024-05-07
fix: Only apply execution control to nodes that are not part of exit handler.

jswxstw:
Fixes #13060 
Fixes #13252

### Motivation

1. This PR https://github.com/argoproj/argo-workflows/pull/12441 has support plugin nodes shutdown, but HTTP node still not.
2. ~~Fixes _ #12659~~  Split to PR: #13294

3. Fixes #13060 Fixes #13252

#### Modification 1
Execution control has been applied to the nodes with created pods after pod reconciliation. However, pending and suspended nodes do not have created pods, and taskset nodes use the agent pod. Mark these nodes failed when shutting down or exceeding deadline since pod reconciliation does not take effect on them.

#### Modification 2
~~Agent pod will only be created when taskset nodes exist, so do not delete it if it does not exist.~~
Split to PR: #13294

#### Modification 3
Only apply execution control for nodes that are not part of exit handler:
- `failNodesWithoutCreatedPodsAfterDeadlineOrShutdown`: `Pending` nodes will not be failed if they are part of exit handler.
- `createWorkflowPod`: Do not set `ARGO_DEADLINE` if it is onExitPod. 

### Verification

local test and e2e tests

PR #13185, opened 2024-06-14
fix: OnExit node can retry when retry a workflow (#13184)

### LINKED ISSUE #13184: When a workflow is retried, the OnExit node does not retry, and its parent node does not retry in some scenarios
### Pre-requisites

1. OnExit node can retry when a workflow is retried
2. If a parent node has only a succeed child and this child is only onExit node, this parent root can also retry

### Version

latest

```YAML
kind: WorkflowTemplate
metadata:
  name: retry-workflow-with-continueon
spec:
  entrypoint: dag
  templates:
    - name: dag
      retryStrategy:
        limit: 2
        retryPolicy: OnError
      dag:
        failFast: false
        tasks:
          - name: success
            template: node-to-exit
            arguments:
              parameters:
                - name: exitCode
                  value: 0
          - name: failure
            template: node-to-exit
            dependencies: [success]
            onExit: on-exit
            arguments:
              parameters:
                - name: exitCode
                  value: 1
          - name: task-after-failure
            template: node-to-exit
            dependencies: [failure]
            arguments:
              parameters:
                - name: exitCode
                  value: 0
          - name: continue
            template: node-to-exit
            continueOn:
              failed: true
            dependencies: [success]
            arguments:
              parameters:
                - name: exitCode
                  value: 2
          - name: task-after-continue
            template: node-to-exit
            dependencies: [continue]
            arguments:
              parameters:
                - name: exitCode
                  value: 0

    - name: node-to-exit
      inputs:
        parameters:
          - name: exitCode
      container:
        image: alpine:3.7
        command: [ sh, "-c", "exit {{inputs.parameters.exitCode}}" ]

    - name: on-exit
      container:
        image: alpine:3.7
        command: [ sh, "-c", "exit 0" ]
```

```text
```

```text
```

JasonChen86899:
Fixes #13184 

### Motivation

fix retry logic

### Modifications

1.  function `FormulateRetryWorkflow` in `workflow/util/util.go` file
2.  function `isDescendantNodeSucceeded` in `workflow/util/util.go` file

### Verification
OnExit node and its parent can be retried when a workflow is retried
yaml: `test/e2e/testdata/retry-workflow-with-continueon.yaml`
1. do retrying
image

2. done
image

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -54,6 +54,10 @@ import (
 	"github.com/argoproj/argo-workflows/v3/workflow/validate"
 )
 
+const (
+	onExitNodeNameSuffix = ".onExit"
+)
+
 // NewWorkflowInformer returns the workflow informer used by the controller. This is actually
 // a custom built UnstructuredInformer which is in actuality returning unstructured.Unstructured
 // objects. We no longer return WorkflowInformer due to:
@@ -780,6 +784,10 @@ func isDescendantNodeSucceeded(wf *wfv1.Workflow, node wfv1.NodeStatus, nodeIDsT
 		if err != nil {
 			log.Panicf("Coudn't obtain child for %s, panicking", child)
 		}
+		// skip onExit child
+		if strings.HasSuffix(childStatus.Name, onExitNodeNameSuffix) {
+			continue
+		}
 		_, present := nodeIDsToReset[child]
 		if (!present && childStatus.Phase == wfv1.NodeSucceeded) || isDescendantNodeSucceeded(wf, *childStatus, nodeIDsToReset) {
 			return true
@@ -885,7 +893,6 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 		}
 	}
 
-	onExitNodeName := wf.ObjectMeta.Name + ".onExit"
 	// Get all children of nodes that match filter
 	nodeIDsToReset, err := getNodeIDsToReset(restartSuccessful, nodeFieldSelector, wf.Status.Nodes)
 	if err != nil {
@@ -905,7 +912,7 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 		}
 		switch node.Phase {
 		case wfv1.NodeSucceeded, wfv1.NodeSkipped:
-			if strings.HasPrefix(node.Name, onExitNodeName) || doForceResetNode {
+			if strings.HasSuffix(node.Name, onExitNodeNameSuffix) || doForceResetNode {
 				log.Debugf("Force reset for node: %s", node.Name)
 				// Reset parent node if this node is a step/task group or DAG.
 				if isGroupNode(node) && node.BoundaryID != "" {
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -54,10 +54,6 @@ import (
 	"github.com/argoproj/argo-workflows/v3/workflow/validate"
 )
 
-const (
-	onExitNodeNameSuffix = ".onExit"
-)
-
 // NewWorkflowInformer returns the workflow informer used by the controller. This is actually
 // a custom built UnstructuredInformer which is in actuality returning unstructured.Unstructured
 // objects. We no longer return WorkflowInformer due to:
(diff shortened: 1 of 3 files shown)


tooptoop4:
what if there are multiple onExit nodes?

JasonChen86899:
> what if there are multiple onExit nodes?

The logic of multiple onExit nodes and one onExit node is the same, and both depend on the execution status of the parent node of onExit
1. If the parent node needs to perform a retry, the onExit node will be retried
2. If the parent node does not need to perform a retry, onExit will not be retried

There are two types of parent nodes that can be retried:
1. There are no successful child nodes other than the onExit node
2. In addition to the onExit node, there are other successful child nodes, and all of these nodes are required to retry at the same time

workflow/util/util.go
```diff
@@ -905,7 +912,7 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 		}
 		switch node.Phase {
 		case wfv1.NodeSucceeded, wfv1.NodeSkipped:
-			if strings.HasPrefix(node.Name, onExitNodeName) || doForceResetNode {
+			if strings.HasSuffix(node.Name, onExitNodeNameSuffix) || doForceResetNode {
```
jswxstw, line 915: Succeeded child nodes of onExitNode will not be retried?

jswxstw:
Can you try this workflow:
```yaml
kind: Workflow
metadata:
  name: workflow-exit-handler-demo
spec:
  entrypoint: main
  onExit: exit-handler
  templates:
  - name: main
    steps:
    - - name: fail
        template: fail
  - name: exit-handler
    steps:
    - - name: alarm
        template: alarm
  - name: alarm
    container:
      image: alpine:3.18
      command: [sh, -c]
      args: ["echo alarm"]
  - name: fail
    container:
      image: alpine:3.18
      command: [sh, -c]
      args: ["exit 1"]
```
Can step `alarm` in the exit handler be retried after manual retry?

JasonChen86899:
@jswxstw Thank you very much for the example, I found that the alarm was not re-executed after I performed a retry, based on this example I re-modified the code logic, see the latest code submission for details

jswxstw:
> @jswxstw Thank you very much for the example, I found that the alarm was not re-executed after I performed a retry, based on this example I re-modified the code logic, see the latest code submission for details

There is already a function to determine whether or not node run as exit handler. You should consider using it directly.
https://github.com/argoproj/argo-workflows/blob/ac9cb3ae93bb441bb7ce09b74e99789adf599eae/pkg/apis/workflow/v1alpha1/workflow_types.go#L2384-L2387

However, this function has a bug now: the `DisplayName` of child nodes in exit handler(if it runs as `Steps`) misses the `.onExit` suffix.
I fixed it in #13016 as below, but this requires confirmation from other reviewers.
image

JasonChen86899:
@jswxstw The function you provided is great, I will modify the code based on it, thanks

Review your change: All children of onExit are also added with the `.OnExit` suffix. The current onExit child node does not have this suffix, which is why running your example makes it impossible to retry with my pervious code.
I have a question for your changes: how about DAG, does DAG have such a bug as well?

So `isExitNode` function that I think can do this: recursively determine whether the parent group node contains the .onExit suffix. Then the children node name no need be added this suffix (related pr:https://github.com/argoproj/argo-workflows/pull/13016)

jswxstw:
> I have a question for your changes: how about DAG, does DAG have such a bug as well?

Good question. DAG do have the same bug.

> So isExitNode function that I think can do this: recursively determine whether the parent group node contains the .onExit suffix. Then the children node name no need be added this suffix (related pr:https://github.com/argoproj/argo-workflows/pull/13016)

Good idea! But `isExitNode` is just the function of `NodeStatus`, it can not access its root node with `boundaryID`.

JasonChen86899:
> Good idea! But `isExitNode` is just the function of `NodeStatus`, it can not access its root node with `boundaryID`.

boundaryID is parent group node just like this function `resetConnectedParentGroupNodes` https://github.com/argoproj/argo-workflows/blob/ac9cb3ae93bb441bb7ce09b74e99789adf599eae/workflow/util/util.go#L815

jswxstw:
> But isExitNode is just the function of NodeStatus, it can not access its root node with boundaryID.

#13016 add a new function `IsPartOfExitHandler` as below, which applies to exit handler of types such as Steps, DAG, Container, and ContainerSet.
```golang
// IsPartOfExitHandler returns whether node is part of exit handler.
func (woc *wfOperationCtx) IsPartOfExitHandler(node wfv1.NodeStatus) bool {
	if node.BoundaryID == "" {
		return node.IsExitNode()
	}
	if boundaryNode, err := woc.wf.Status.Nodes.Get(node.BoundaryID); err == nil {
		return boundaryNode.IsExitNode()
	}
	return false
}
```

JasonChen86899:
@jswxstw the code has been updated and can be reviewed

Issue #9989, opened 2022-11-07
SSO authentication not working


simox-83:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I upgraded Argo Workflow from 3.1.13 to 3.4.3. SSO Authentication was working fine with 3.1.13; however, the 3.4.3 doesn't seem to work.
The SSO configuration (Okta) has not changed.

When I tried to open the UI and click on login on the SSO, I get a red banner on the down right corner saying `Failed to load version/info Error: Unauthorized`. After that the web page just tries to load and after sometime it replies with `test-ce-argo-server-integration.k8s.cnqr.tech didn't send any data.ERR_EMPTY_RESPONSE`

### Version

3.4.3

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
N/A - The issue happens at login time, so I can't run any workflow.
```

### Logs from the workflow controller

I am attaching the logs of the workflow server, because the error happens during authentication:

```
time="2022-11-07T13:37:32.115Z" level=info duration=2.752873ms method=GET path=/main.2430295409b8b54e52ad.js size=1471060 status=0
time="2022-11-07T13:37:32.117Z" level=info duration="23.156µs" method=GET path=index.html size=0 status=304
time="2022-11-07T13:37:33.935Z" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = token not valid for running mode" grpc.code=Unauthenticated grpc.method=GetUserInfo grpc.service=info.InfoService grpc.start_time="2022-11-07T13:37:33Z" grpc.time_ms=0.047 span.kind=server system=grpc
time="2022-11-07T13:37:33.935Z" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = token not valid for running mode" grpc.code=Unauthenticated grpc.method=GetInfo grpc.service=info.InfoService grpc.start_time="2022-11-07T13:37:33Z" grpc.time_ms=0.028 span.kind=server system=grpc
time="2022-11-07T13:37:33.935Z" level=info duration=1.628208ms method=GET path=/api/v1/userinfo size=56 status=401
[...]

PR #13189, opened 2024-06-14
fix(server): don't use cluster scope list + watch in namespaced mode. Fixes #13177

### LINKED ISSUE #13177: v3.5.7: namespaced install causes the server to complain about list+watch on cluster scoped workflows
### Pre-requisites

To reproduce install 3.5.7 namespaced 
`kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.5.7/namespace-install.yaml`

You will see
```
W0613 08:46:47.382693       1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.3/tools/cache/reflector.go:167: failed to list *v1alpha1.Workflow: workflows.argoproj.io is forbidden: User "system:serviceaccount:argo:argo-server" cannot list
 resource "workflows" in API group "argoproj.io" at the cluster scope                                                                                                                                                                          
E0613 08:46:47.382720       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.3/tools/cache/reflector.go:167: Failed to watch *v1alpha1.Workflow: failed to list *v1alpha1.Workflow: workflows.argoproj.io is forbidden: User "system:servicea
ccount:argo:argo-server" cannot list resource "workflows" in API group "argoproj.io" at the cluster scope    
```

This does not happen with 3.5.6 namespaced installs.

I believe this is coming from #12736 / #13021 on lines https://github.com/argoproj/argo-workflows/blob/b28486cdc0a2174d68a0d0d04ba788847552b826/server/workflow/workflow_server.go#L73 and https://github.com/argoproj/argo-workflows/blob/b28486cdc0a2174d68a0d0d04ba788847552b826/server/workflow/workflow_server.go#L76

It feels like these should be namespaced when the server is running namespaced.

### Version

v3.5.7

```YAML
Not relevant.
```

```text
Not relevant. Server logs above.
```

```text
Not relevant
```

jiachengxu:
Fixes https://github.com/argoproj/argo-workflows/issues/13177

### Verification
Try to apply the following manifests into a k8s cluster:
```
kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.5.7/namespace-install.yaml
```
With the image built from this PR, don't see the following permission errors anymore:

> 14.612814       1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.3/tools/cache/reflector.go:167: failed to list *v1alpha1.Workflow: workflows.argoproj.io is forbidden: User "system:serviceaccount:argo:argo-server" cannot list resource "workflows" in API group "argoproj.io" at the cluster scope
E0614 16:39:14.612865       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.3/tools/cache/reflector.go:167: Failed to watch *v1alpha1.Workflow: failed to list *v1alpha1.Workflow: workflows.argoproj.io is forbidden: User "system:serviceaccount:argo:argo-server" cannot list resource "workflows" in API group "argoproj.io" at the cluster scope

### DIFF
--- a/pkg/apiclient/argo-kube-client.go
+++ b/pkg/apiclient/argo-kube-client.go
@@ -6,6 +6,7 @@ import (
 
 	eventsource "github.com/argoproj/argo-events/pkg/client/eventsource/clientset/versioned"
 	sensor "github.com/argoproj/argo-events/pkg/client/sensor/clientset/versioned"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
@@ -92,7 +93,7 @@ func newArgoKubeClient(ctx context.Context, clientConfig clientcmd.ClientConfig,
 func (a *argoKubeClient) NewWorkflowServiceClient() workflowpkg.WorkflowServiceClient {
 	wfArchive := sqldb.NullWorkflowArchive
 	wfLister := store.NewKubeLister(a.wfClient)
-	return &errorTranslatingWorkflowServiceClient{&argoKubeWorkflowServiceClient{workflowserver.NewWorkflowServer(a.instanceIDService, argoKubeOffloadNodeStatusRepo, wfArchive, a.wfClient, wfLister, nil)}}
+	return &errorTranslatingWorkflowServiceClient{&argoKubeWorkflowServiceClient{workflowserver.NewWorkflowServer(a.instanceIDService, argoKubeOffloadNodeStatusRepo, wfArchive, a.wfClient, wfLister, nil, v1.NamespaceAll)}}
 }
 
 func (a *argoKubeClient) NewCronWorkflowServiceClient() (cronworkflow.CronWorkflowServiceClient, error) {
--- a/pkg/apiclient/argo-kube-client.go
+++ b/pkg/apiclient/argo-kube-client.go
@@ -6,7 +6,6 @@ import (
 
 	eventsource "github.com/argoproj/argo-events/pkg/client/eventsource/clientset/versioned"
 	sensor "github.com/argoproj/argo-events/pkg/client/sensor/clientset/versioned"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
@@ -93,7 +92,7 @@ func newArgoKubeClient(ctx context.Context, clientConfig clientcmd.ClientConfig,
 func (a *argoKubeClient) NewWorkflowServiceClient() workflowpkg.WorkflowServiceClient {
 	wfArchive := sqldb.NullWorkflowArchive
 	wfLister := store.NewKubeLister(a.wfClient)
-	return &errorTranslatingWorkflowServiceClient{&argoKubeWorkflowServiceClient{workflowserver.NewWorkflowServer(a.instanceIDService, argoKubeOffloadNodeStatusRepo, wfArchive, a.wfClient, wfLister, nil, v1.NamespaceAll)}}
+	return &errorTranslatingWorkflowServiceClient{&argoKubeWorkflowServiceClient{workflowserver.NewWorkflowServer(a.instanceIDService, argoKubeOffloadNodeStatusRepo, wfArchive, a.wfClient, wfLister, nil, nil)}}
 }
 
 func (a *argoKubeClient) NewCronWorkflowServiceClient() (cronworkflow.CronWorkflowServiceClient, error) {
--- a/server/apiserver/argoserver.go
+++ b/server/apiserver/argoserver.go
@@ -124,9 +124,9 @@ func init() {
 	}
 }
 
-func getResourceCacheNamespace(opts ArgoServerOpts) string {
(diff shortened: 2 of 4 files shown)


jiachengxu:
Hi @Joibel @agilgur5, could you take a look at the PR?

server/apiserver/argoserver.go
```diff
@@ -124,9 +124,9 @@ func init() {
 	}
 }
 
-func getResourceCacheNamespace(opts ArgoServerOpts) string {
-	if opts.ManagedNamespace != "" {
-		return opts.ManagedNamespace
+func getResourceCacheNamespace(managedNamespace string) string {
+	if managedNamespace != "" {
+		return managedNamespace
```
agilgur5, line 129: this logic is not quite correct (potentially due to #9989?) -- if not `managedNamespace`, it should check if `namespaced`, and, if so, use `namespace`, otherwise use cluster scope. Same as I wrote in https://github.com/argoproj/argo-workflows/issues/13177#issuecomment-2167280162.

Right now this is missing namespaced mode without managed namespace

agilgur5: Nevermind, apparently the CLI init handles this with [`managedNamespace = namespace`](https://github.com/argoproj/argo-workflows/blob/0ca0c0f721c9bf05c63dea98168cfa87441656c7/cmd/argo/commands/server.go#L98) in that case.

pkg/apiclient/argo-kube-client.go
```diff
@@ -92,7 +93,7 @@ func newArgoKubeClient(ctx context.Context, clientConfig clientcmd.ClientConfig,
 func (a *argoKubeClient) NewWorkflowServiceClient() workflowpkg.WorkflowServiceClient {
 	wfArchive := sqldb.NullWorkflowArchive
 	wfLister := store.NewKubeLister(a.wfClient)
-	return &errorTranslatingWorkflowServiceClient{&argoKubeWorkflowServiceClient{workflowserver.NewWorkflowServer(a.instanceIDService, argoKubeOffloadNodeStatusRepo, wfArchive, a.wfClient, wfLister, nil)}}
+	return &errorTranslatingWorkflowServiceClient{&argoKubeWorkflowServiceClient{workflowserver.NewWorkflowServer(a.instanceIDService, argoKubeOffloadNodeStatusRepo, wfArchive, a.wfClient, wfLister, nil, v1.NamespaceAll)}}
```
agilgur5, line 96: The SDK doesn't need a namespace since it doesn't use the Reflector/Store, right?

agilgur5 (commented):
Thanks for fixing this, just one small comment

pkg/apiclient/argo-kube-client.go
jiachengxu: yes, you are right. Although it doesn't matter what namespace we pass here, I made it a nil pointer to make it more clear as you suggested.

server/apiserver/argoserver.go
jiachengxu: Thank you for the review, yes, CLI init handles that case.

pkg/apiclient/argo-kube-client.go
agilgur5: Yea the explicitness is helpful.
Although I forgot that `nil` only works for pointers in Go, and that the string default `""` is, perhaps confusingly, equivalent to cluster-scope, so you had to refactor a bunch here. The other option would be to use an invalid NS name like `"-1"`, but I think `nil` is clearer than that.

server/workflow/workflow_server_test.go
```diff
@@ -638,7 +638,7 @@ func getWorkflowServer() (workflowpkg.WorkflowServiceServer, context.Context) {
 	if err = wfStore.Add(&wfObj5); err != nil {
 		panic(err)
 	}
-	server := NewWorkflowServer(instanceIdSvc, offloadNodeStatusRepo, archivedRepo, wfClientset, wfStore, wfStore)
+	server := NewWorkflowServer(instanceIdSvc, offloadNodeStatusRepo, archivedRepo, wfClientset, wfStore, wfStore, nil)
```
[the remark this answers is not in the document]
jiachengxu: oh, I think I changed it by accident, and I changed it back the `metav1.NamespaceAll`

> With the new && namespace != nil check, this now takes a different path (but still passes apparently?)

This doesn't really matter regarding the test because we don't [start the reflector to perform list and watch](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server.go#L86-L88) in the test and we just [manually add objects to the reflector/store](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server_test.go#L631-L640).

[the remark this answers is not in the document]
agilgur5: But the reflector is [set](https://github.com/argoproj/argo-workflows/blob/dc6a18dc545bbf35103ffddb81a014eefcbb403d/server/workflow/workflow_server.go#L79) since the store is not `nil` in this case, so it would start wouldn't it? It just isn't used actively in this test

agilgur5 (approved):
LGTM, thanks for fixing this! Will have a patch out soon!

server/workflow/workflow_server_test.go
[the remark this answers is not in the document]
jiachengxu: the [`Run` method](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server.go#L85-L89) is not called in the test, so the reflector is not started.

PR #13198, opened 2024-06-17
fix: Do not reset the root node by default. Fixes #13196

### LINKED ISSUE #13196: Single-node workflow with failed exit handler cannot be retried correctly
### Pre-requisites

The demo workflow failed because of exit handler failure.
image
The succeeded node `workflow-exit-handler-fail` turn to `Error` phase with message: "pod deleted" after manual retry.
image
https://github.com/argoproj/argo-workflows/blob/8f3860d7683068ab4f93bc725eb24dbbaee78e11/workflow/util/util.go#L993-L998
The root node is reset to `Running`, I think this is the root cause.

### Version

latest

```YAML
kind: Workflow
metadata:
  name: workflow-exit-handler-fail
spec:
  entrypoint: echo
  onExit: exit-handler
  templates:
  - name: echo
    container:
      image: alpine:3.18
      command: [sh, -c]
      args: ["echo hello-world"]
  - name: fail
    container:
      image: alpine:3.18
      command: [sh, -c]
      args: ["exit 1"]
  - name: exit-handler
    steps:
      - - name: exit-handler-task
          template: fail
```

```text
```

```text
```

jswxstw:
Fixes #13196 

### Motivation

Single-node workflow with failed exit handler can not be retried correctly.

### Modifications

Only reset root node when it is `FailedOrError`, remove the default reset logic.

### Verification

local test and e2e test

### DIFF
--- a/test/e2e/cli_test.go
+++ b/test/e2e/cli_test.go
@@ -932,6 +932,47 @@ func (s *CLISuite) TestRetryWorkflowWithContinueOn() {
 		})
 }
 
+func (s *CLISuite) TestRetryWorkflowWithFailedExitHandler() {
+	var workflowName string
+	s.Given().
+		Workflow(`@testdata/retry-workflow-with-failed-exit-handler.yaml`).
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeFailed).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			workflowName = metadata.Name
+			assert.Equal(t, 2, len(status.Nodes))
+		}).
+		RunCli([]string{"retry", workflowName}, func(t *testing.T, output string, err error) {
+			if assert.NoError(t, err, output) {
+				assert.Contains(t, output, "Name:")
+				assert.Contains(t, output, "Namespace:")
+			}
+		})
+
+	s.Given().
+		When().
+		WaitForWorkflow(fixtures.ToBeCompleted).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			workflowName = metadata.Name
+			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
+			assert.Equal(t, 2, len(status.Nodes))
+		}).
+		ExpectWorkflowNode(func(status wfv1.NodeStatus) bool {
+			return status.Name == workflowName
+		}, func(t *testing.T, status *wfv1.NodeStatus, pod *corev1.Pod) {
+			assert.Equal(t, wfv1.NodeSucceeded, status.Phase)
+		}).
+		ExpectWorkflowNode(func(status wfv1.NodeStatus) bool {
+			return strings.Contains(status.Name, ".onExit")
+		}, func(t *testing.T, status *wfv1.NodeStatus, pod *corev1.Pod) {
+			assert.Equal(t, wfv1.NodeFailed, status.Phase)
+			assert.Contains(t, status.Message, "exit code 1")
+		})
+}
+
 func (s *CLISuite) TestWorkflowStop() {
 	s.Given().
 		Workflow("@smoke/basic.yaml").
--- a/test/e2e/cli_test.go
+++ b/test/e2e/cli_test.go
@@ -942,7 +942,7 @@ func (s *CLISuite) TestRetryWorkflowWithFailedExitHandler() {
 		Then().
 		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
 			workflowName = metadata.Name
-			assert.Equal(t, 2, len(status.Nodes))
+			assert.Len(t, status.Nodes, 2)
 		}).
 		RunCli([]string{"retry", workflowName}, func(t *testing.T, output string, err error) {
 			if assert.NoError(t, err, output) {
@@ -958,7 +958,7 @@ func (s *CLISuite) TestRetryWorkflowWithFailedExitHandler() {
 		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
 			workflowName = metadata.Name
 			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
-			assert.Equal(t, 2, len(status.Nodes))
+			assert.Len(t, status.Nodes, 2)
 		}).
 		ExpectWorkflowNode(func(status wfv1.NodeStatus) bool {
 			return status.Name == workflowName
--- a/test/e2e/testdata/retry-workflow-with-failed-exit-handler.yaml
+++ b/test/e2e/testdata/retry-workflow-with-failed-exit-handler.yaml
@@ -1,0 +1,18 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  name: retry-workflow-with-failed-exit-handler
+spec:
+  entrypoint: hello
+  onExit: exit-handler
+  templates:
+  - name: hello
+    container:
+      image: alpine:3.18
+      command: [sh, -c]
+      args: ["echo hello"]
+  - name: exit-handler
+    container:
+      image: alpine:3.18
+      command: [sh, -c]
+      args: ["exit 1"]
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -989,13 +989,6 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 			// Do not allow retry of workflows with pods in Running/Pending phase
 			return nil, nil, errors.InternalErrorf("Workflow cannot be retried with node %s in %s phase", node.Name, node.Phase)
 		}
-
-		if node.Name == wf.ObjectMeta.Name {
-			log.Debugf("Reset root node: %s", node.Name)
-			newNode := node.DeepCopy()
-			newWF.Status.Nodes.Set(newNode.ID, resetNode(*newNode))
-			continue
-		}
 	}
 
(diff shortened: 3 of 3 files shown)


terrytangyuan (commented):
> Single-node workflow with failed exit handler can not be retried correctly.

Can you add a test case?

jswxstw:
> > Single-node workflow with failed exit handler can not be retried correctly.
> 
> Can you add a test case?

Sure.

juliev0:
Thanks for fixing. Would you mind taking me through what happened to help me understand? I don't know how the exit handler is related.

First, the root node got reset to "Running" as you mentioned. So, how did the node being in "Running" cause the retry to fail? (any links into the code if possible will help)

Thanks!

jswxstw:
> First, the root node got reset to "Running" as you mentioned. So, how did the node being in "Running" cause the retry to fail? (any links into the code if possible will help)

The root node got reset to `Running` from `Succeeded` state, which is not as expected:
- Firstly, it should not re-run, because it is already `Succeeded`.
- Secondly, its pod is completed, so this node can not run correctly: new pod will not be created because it already exists and then this node turns to `Error` with message `pod deleted`.

jswxstw:
> > First, the root node got reset to "Running" as you mentioned. So, how did the node being in "Running" cause the retry to fail? (any links into the code if possible will help)
> 
> The root node of `Pod` type got reset to `Running` from `Succeeded` state, which is not as expected:
> 
> * Firstly, it should not re-run, because it is already `Succeeded`.
> * Secondly, its pod is completed, so this node can not run correctly: new pod will not be created because it already exists and then this node turns to `Error` with message `pod deleted`.

juliev0:
> > First, the root node got reset to "Running" as you mentioned. So, how did the node being in "Running" cause the retry to fail? (any links into the code if possible will help)
> 
> The root node of `Pod` type got reset to `Running` from `Succeeded` state, which is not as expected:
> 
> * Firstly, it should not re-run, because it is already `Succeeded`.
> * Secondly, its pod is completed, so this node can not run correctly: new pod will not be created because it already exists and then this node turns to `Error` with message `pod deleted`.

Got it, thanks. I see the difference is the Group node would be a DAG, Steps, etc, which doesn't run. 

And regarding the Description, does it matter that there was an exit handler, or could it just have easily been a single node workflow that failed, and would've had the same issue?

jswxstw:
> And regarding the Description, does it matter that there was an exit handler, or could it just have easily been a single node workflow that failed, and would've had the same issue?

If there is only one single node without exit handler, the workflow will be `Succeeded` which is not retryable, so it will not cause any problems.

juliev0:
> > And regarding the Description, does it matter that there was an exit handler, or could it just have easily been a single node workflow that failed, and would've had the same issue?
> 
> If there is only one single node without exit handler, the workflow will be `Succeeded` which is not retryable, so it will not cause any problems.

I'm confused. Is this not a single node Workflow?:
```
iVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: entrypoint-fail
spec:
  entrypoint: failing-entrypoint
  templates:
  - name: failing-entrypoint
    container:
      image: alpine:3.18
      command: [sh, -c]
      args: ["exit 1"]
```

jswxstw:
> I'm confused. Is this not a single node Workflow?:

Yes, this is a single node worklfow. But the root node is of pod type which is no need to be reset, since it will be deleted after manuly retry.
```shell
# argo retry entrypoint-fail
INFO[2024-08-25T00:25:22.288Z] Deleting pod                                  podDeleted=entrypoint-fail
INFO[2024-08-25T00:25:22.295Z] Workflow to be dehydrated                     Workflow Size=902
Name:                entrypoint-fail
Namespace:           argo
Status:              Running
Conditions:          
 PodRunning          False
 Completed           False
Created:             Sun Aug 25 00:24:35 +0800 (47 seconds ago)
Started:             Sun Aug 25 00:25:22 +0800 (now)
Duration:            0 seconds
Progress:            0/1
ResourcesDuration:   0s*(1 cpu),2s*(100Mi memory)
```

juliev0:
So, just trying to understand - this statement isn't necessarily true in all cases, right?:

```
If there is only one single node without exit handler, the workflow will be Succeeded which is not retryable, so it will not cause any problems.
```
since the Workflow I have above is an example of an unsucceeded Workflow, right? But you're saying that in my example, the Root node is of type `Pod`? 

And in your example what type is the Root node? I imagine there are 2 nodes in yours, right? Can you help me understand what the tree looks like for that?

Thank you for bearing with me. I'm guessing your change is a good one. I just want to make sure I fully understand.

juliev0:
Okay, I just decided to try running yours. This is the result:
```
  nodes:
    workflow-exit-handler-fail:
      displayName: workflow-exit-handler-fail
      finishedAt: "2024-08-25T18:42:37Z"
      hostNodeName: lima-rancher-desktop
      id: workflow-exit-handler-fail
      name: workflow-exit-handler-fail
      outputs:
        artifacts:
        - name: main-logs
          s3:
            key: workflow-exit-handler-fail/workflow-exit-handler-fail/main.log
        exitCode: "0"
      phase: Succeeded
      progress: 1/1
      resourcesDuration:
        cpu: 0
        memory: 5
      startedAt: "2024-08-25T18:42:28Z"
      templateName: echo
      templateScope: local/workflow-exit-handler-fail
      type: Pod
    workflow-exit-handler-fail-925896592:
      children:
      - workflow-exit-handler-fail-1000917866
      displayName: workflow-exit-handler-fail.onExit
      finishedAt: "2024-08-25T18:42:47Z"
      id: workflow-exit-handler-fail-925896592
      message: child 'workflow-exit-handler-fail-2100426797' failed
      name: workflow-exit-handler-fail.onExit
      nodeFlag:
        hooked: true
      outboundNodes:
      - workflow-exit-handler-fail-2100426797
      phase: Failed
      progress: 0/1
      resourcesDuration:
        cpu: 0
        memory: 2
      startedAt: "2024-08-25T18:42:40Z"
      templateName: exit-handler
      templateScope: local/workflow-exit-handler-fail
      type: Steps
    workflow-exit-handler-fail-1000917866:
      boundaryID: workflow-exit-handler-fail-925896592
      children:
      - workflow-exit-handler-fail-2100426797
      displayName: '[0]'
      finishedAt: "2024-08-25T18:42:47Z"
      id: workflow-exit-handler-fail-1000917866
      message: child 'workflow-exit-handler-fail-2100426797' failed
      name: workflow-exit-handler-fail.onExit[0]
      nodeFlag: {}
      phase: Failed
      progress: 0/1
      resourcesDuration:
        cpu: 0
        memory: 2
      startedAt: "2024-08-25T18:42:40Z"
      templateScope: local/workflow-exit-handler-fail
      type: StepGroup
    workflow-exit-handler-fail-2100426797:
      boundaryID: workflow-exit-handler-fail-925896592
      displayName: exit-handler-task
      finishedAt: "2024-08-25T18:42:44Z"
      hostNodeName: lima-rancher-desktop
      id: workflow-exit-handler-fail-2100426797
      message: Error (exit code 1)
      name: workflow-exit-handler-fail.onExit[0].exit-handler-task
      outputs:
        artifacts:
        - name: main-logs
          s3:
            key: workflow-exit-handler-fail/workflow-exit-handler-fail-fail-2100426797/main.log
        exitCode: "1"
      phase: Failed
      progress: 0/1
      resourcesDuration:
        cpu: 0
        memory: 2
      startedAt: "2024-08-25T18:42:40Z"
      templateName: fail
      templateScope: local/workflow-exit-handler-fail
      type: Pod

```

So, root is a `Pod` type. And in addition there's a `Steps` with child `StepGroup` with child `Pod` representing the ExitHandler.

juliev0:
The thing I'm not clear on is why [this](https://github.com/argoproj/argo-workflows/blob/8f3860d7683068ab4f93bc725eb24dbbaee78e11/workflow/util/util.go#L993-L998) special logic for resetting root node even exists at all. Do we know why it's not sufficient to just catch all errored GroupNodes [here](https://github.com/argoproj/argo-workflows/blob/8f3860d7683068ab4f93bc725eb24dbbaee78e11/workflow/util/util.go#L972)? (including the root node) Is it important to all reset Succeeded Root node, not just Errored?

jswxstw:
> And in your example what type is the Root node? I imagine there are many nodes in yours, right? Can you help me understand what the tree looks like for that?

The phase of the root node and the workflow is usually the same if there is no exit handler.
The type of root node is usually `Steps` or `DAG` since a workflow usually have many steps or tasks.
My example is very special which breaks the above two rules: 
- root node is of `Pod` type because there is only one node.
- root node is `Succeeded` but workflow is `Failed` because exit handler runs failed.

Workflow is `Failed`, so I can retry it, then the root node of `Pod` type is reset to `Running` from `Succeeded` which is unacceptable.

jswxstw:
> The thing I'm not clear on is why [this](https://github.com/argoproj/argo-workflows/blob/8f3860d7683068ab4f93bc725eb24dbbaee78e11/workflow/util/util.go#L993-L998) special logic for resetting root node even exists at all. Do we know why it's not sufficient to just catch all errored GroupNodes [here](https://github.com/argoproj/argo-workflows/blob/8f3860d7683068ab4f93bc725eb24dbbaee78e11/workflow/util/util.go#L972)? (including the root node) Is it important to also reset Succeeded Root node, not just Errored?

I didn't notice this before. I agree with you that [this](https://github.com/argoproj/argo-workflows/blob/8f3860d7683068ab4f93bc725eb24dbbaee78e11/workflow/util/util.go#L993-L998) special logic for resetting root node seems useless now since `Steps` is belong to `GroupNode`.

> Is it important to also reset Succeeded Root node, not just Errored?

Maybe it is useful in situation of `restartSuccessful`.

juliev0:
Hey @terrytangyuan - I know this code was written a couple years back but do you have any input for [this](https://github.com/argoproj/argo-workflows/pull/13198#issuecomment-2308971897) question?

terrytangyuan:
It should be fine to remove that special handling.

juliev0:
> It should be fine to remove that special handling.

Thanks for the quick response @terrytangyuan !

juliev0 (approved):

Issue #11354, opened 2023-07-13
"Pod failed: Error (exit code 255)" but no retry


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

i have workflow with 2 steps, first step succeeded, but 2nd step shows no logs at all and has "Error (exit code 255)" in the UI.

if pod couldn't start i would expect it to retry (since i have retryPolicy: "OnError")

here are kubectl events output:
```
{"time":"2023-07-13T22:00:07.575403594Z","stream":"stdout","_p":"F","log":"auth                18m         Normal    TaintManagerEviction           pod/failedpod                                            taint-controller                                                   Cancelling deletion of Pod auth/failedpod                                                                                                                                                                      18m          2       failedpod.17718b4a8ba756ef
{"time":"2023-07-13T22:00:07.57589893Z","stream":"stdout","_p":"F","log":"auth                18m         Warning   FailedCreatePodSandBox         pod/failedpod                                            kubelet, redactnode          Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox \"redactid\": plugin type=\"aws-cni\" name=\"aws-cni\" failed (add): add cmd: failed to assign an IP address to container   18m          1       failedpod.17718b4a8d4a3809
{"time":"2023-07-13T22:00:07.575931441Z","stream":"stdout","_p":"F","log":"auth                18m         Warning   FailedScheduling               pod/earlierstepsuccesspod                                                  default-scheduler                                                  0/3 nodes are available: 1 Insufficient memory, 3 Insufficient cpu. preemption: 0/3 nodes are available: 3 No preemption victims found for incoming pod.                                                                                                                         20m          8       earlierstepsuccesspod.17718b374ee51bf4
{"time":"2023-07-13T22:00:07.577291549Z","stream":"stdout","_p":"F","log":"auth                19m         Normal    TriggeredScaleUp               pod/earlierstepsuccesspod                                                  cluster-autoscaler                                                 pod triggered scale-up: [{redactnodegroup 1->2 (max: 5)}]                                                                                                                                                        19m          1       earlierstepsuccesspod.17718b4330a0da1e
[...]

PR #13201, opened 2024-06-17
fix: skip clear message when node transition from pending to fail. Fixes #13200

### LINKED ISSUE #13200: Pod doesn't retry when using `"PodInitializing"` in `TRANSIENT_ERROR_PATTERN`
### Pre-requisites

When a spot node get evicted, pod has reason/message `"PodInitializing"`
However, specify `"PodInitializing"` in TRANSIENT_ERROR_PATTERN doesn't retry the pod

### Version

latest (https://github.com/argoproj/argo-workflows/commit/669d4a4e986b359dde3cd07e5616be8f713fe23d)

```YAML
Any workflow, but node needs to be evicted during podinitializing
```

```text
timestamp0,                    status: pending, reason: ""
timestamp1,                    status: pending, reason: "PodInitializing"
timestamp2(immediately after), status: fail,    reason: ""
```

```text
```

tczhao:
Fixes #13200 

### Motivation

Allow retry to use pod message when pod transition from pending to fail.

normally we have
```
timestamp0, status: pending, reason: ""
timestamp1, status: fail,    reason: "e.g. containerd issue"
```

but for podinitializing, the transition are the following
```
timestamp0,                    status: pending, reason: ""
timestamp1,                    status: pending, reason: "PodInitializing"
timestamp2(immediately after), status: fail,    reason: ""
```

### Modifications

This PR fixes the issue, we don't overwrite message with "" when pod transition from pending to fail phase
```
timestamp0,                    status: pending, reason: ""
timestamp1,                    status: pending, reason: "PodInitializing"
timestamp2(immediately after), status: fail,    reason: "PodInitializing"
```

### Verification

Add unit test, test failed
Add changes, test succeed
Also release to our production env for a week and pods able to retry on all `PodInitializing` message when configured in `TRANSIENT_ERROR_PATTERN` without seeing any other issue


workflow/controller/operator.go
```diff
@@ -1471,8 +1471,8 @@ func (woc *wfOperationCtx) assessNodeStatus(pod *apiv1.Pod, old *wfv1.NodeStatus
 		}
 	}
 
-	// if we are transitioning from Pending to a different state, clear out unchanged message
-	if old.Phase == wfv1.NodePending && new.Phase != wfv1.NodePending && old.Message == new.Message {
+	// if we are transitioning from Pending to a different state (except Fail), clear out unchanged message
+	if old.Phase == wfv1.NodePending && new.Phase != wfv1.NodePending && new.Phase != wfv1.NodeFailed && old.Message == new.Message {
```
agilgur5, line 1475: should this also applied to the `Errored` phase?

agilgur5:
I'm wondering if this might partially help with #11354 and #12572 too 🤔

workflow/controller/operator.go
tczhao: Make sense

tczhao:
> I'm wondering if this might partially help with #11354 and #12572 too 🤔

I think so, as a workaround solution, use `Message` instead of `ExitCode`

agilgur5 (approved):
makes sense to me

juliev0:
I see this has been approved. @tczhao do you want to fix the merge conflict, and then it can be merged?

PR #13203, opened 2024-06-17
fix(server): switch to `JSON_EXTRACT` and `JSON_UNQUOTE` for MySQL/MariaDB. Fixes #13202

### LINKED ISSUE #13202: 3.5.6: Archive workflow sql syntax error with mariadb
### Pre-requisites

since `v3.5.6` using mariadb can not list workflows because of [200f4d1e5](https://github.com/argoproj/argo-workflows/commit/200f4d1e5ffee0a57a9e7a9995b95da15230eb97) / #12912

exception
```
{"error":"rpc error: code = Internal desc = Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '\u003e\u003e'$.metadata.labels', '{}') as labels,coalesce(workflow-\u003e\u003e'$.metadata.annota...' at line 5","grpc.code":"Internal","grpc.method":"ListWorkflows","grpc.service":"workflow.WorkflowService","grpc.start_time":"2024-06-17T14:57:19Z","grpc.time_ms":6.307,"level":"error","msg":"finished unary call with code Internal","span.kind":"server","system":"grpc","time":"2024-06-17T14:57:19.890Z"}
{"duration":6836621,"level":"info","method":"GET","msg":"","path":"/api/v1/workflows","size":292,"status":500,"time":"2024-06-17T14:57:19.890Z"}
```
because json extract with `->` is not supported by mariadb (see also https://jira.mariadb.org/browse/MDEV-13594)
we should use `JSON_EXTRACT(field,path)`

### Version

v3.5.6

```YAML
nope
```

```text
```

```text
```

spaced:
fix: use standard `JSON_EXTRACT` and `JSON_UNQUOTE` function for list workflow using mariadb. Fixes #13202

`-->` is only a alias to `JSON_EXTRACT()` function
> (The [->](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#operator_json-column-path) operator serves as an alias for the [JSON_EXTRACT()](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#function_json-extract)

for some fields it was even wrong (for example it makes no sense to unquote the `$.metadata.labels`.
mariadb does not support this (see feature request in mariadb: https://jira.mariadb.org/browse/MDEV-13594)

Fixes #13202
Follow-up to #12912

### Motivation

be compatible with mariadb

### Modifications

- replace `->` with `JSON_EXTRACT`
- replace `-->` with `JSON_UNQUOTE(JSON_EXTRACT(..,..)`

### Verification

locally


agilgur5:
@jiachengxu could you review/double-check this? Thanks again for all your follow-ups ❤️

terrytangyuan (commented):
The changes make sense but I wonder if we should introduce additional e2e test suits for other DBs for compatibility

jiachengxu:
The change looks good to me, and to Terry's point, I think it makes sense to add e2e tests for DB compatibility

spaced:
@terrytangyuan @jiachengxu shouldt the e2e test with mariab part of this PR or anything else is needed to get merged? It would be also nice if this can be back ported to  version `3.5.x`.

agilgur5:
> It would be also nice if this can be back ported to version `3.5.x`.

We'd do that automatically as it's a fix 👍 
I'm also release managing `3.5.x` right now and so also know to backport it when merged.

> should the e2e test with mariab part of this PR or anything else is needed to get merged?

IMO, I think this is fine as-is since it fixes an unintended compatibility regression (although really should be on MariaDB to be compatible, no?). Tests would be great to add, but we pretty rarely change DB queries, I'd be fine if they're added later. @terrytangyuan are you fine with that?

agilgur5 (approved):

terrytangyuan (approved):

PR #13000, opened 2024-05-01
feat(ci): Add retest workflow. Fixes #12864

miltalex:
Fixes #12864 

### Motivation

Due to flaky tests, it would be nice to allow `Members` to re-run failing workflows.

### Modifications

Extend github workflows with `retest.yaml`. It adds a new workflow run which is triggered on `issue_comment:[created,edited]`. It requires the comment to contain exactly the `/retest`. It will execute a sequence of API calls to fetch the latest SHA, the latest run ID and trigger the rerun of the failed jobs.

Note: If the workflow doesn't contain any failed jobs it will fail.

### Verification

Used a forked repository https://github.com/miltalex/argo-workflows . Merged the new workflow on `main` branch and opened PRs (https://github.com/miltalex/argo-workflows/pull/1 , https://github.com/miltalex/argo-workflows/pull/2 , https://github.com/miltalex/argo-workflows/pull/3) in order to:

- [x] Test new comment `/retest` &rarr; Trigger failed jobs
- [x] Test edit comment  to `retest` &rarr;  Retest workflow is skipped
- [x] Test edit comment to `/retest` &rarr; Trigger failed jobs
- [x] Test new comment without `/retest` &rarr;  Retest workflow is skipped
- [x] Test new comment with `/retest` and no failing jobs &rarr;  Retest workflow failed due to no failing jobs.

PR #13208, opened 2024-06-18
ci(retest): filter more comments to use GHA less

agilgur5:
Follow-up to #13000

### Motivation

- they get triggerred [quite often](https://github.com/argoproj/argo-workflows/actions/workflows/retest.yaml) (for every comment), and though they only run for 2s, it would be good to filter out more of them and use less CI time

### Modifications

- only run on `created` events, skip all `edited` events
  - the vast majority of `edited` events will be for unrelated comments
  - you can delete and comment again to trigger a retest instead if you had a typo

- only run on PR comments, following [GH docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only)
  - it would fail otherwise, but this prevents that scenario entirely

### DIFF
--- a/.github/workflows/retest.yaml
+++ b/.github/workflows/retest.yaml
@@ -1,14 +1,15 @@
 name: Detect and Trigger Retest
 on:
   issue_comment:
-    types: [created, edited]
+    types: [created]
 
 permissions:
   contents: read
 
 jobs:
   retest:
-    if: github.event.comment.author_association == 'MEMBER' && github.event.comment.body == '/retest'
+    # PR comments where a Member types "/retest" exactly
+    if: github.event.issue.pull_request && github.event.comment.author_association == 'MEMBER' && github.event.comment.body == '/retest'
     permissions:
       actions: write # for re-running failed jobs: https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#re-run-a-job-from-a-workflow-run
     runs-on: ubuntu-latest


juliev0 (approved):

.github/workflows/retest.yaml
```diff
@@ -1,14 +1,15 @@
...
+    types: [created]
 
 permissions:
   contents: read
 
 jobs:
   retest:
-    if: github.event.comment.author_association == 'MEMBER' && github.event.comment.body == '/retest'
+    # PR comments where a Member types "/retest" exactly
+    if: github.event.issue.pull_request && github.event.comment.author_association == 'MEMBER' && github.event.comment.body == '/retest'
```
juliev0, line 12: So, this only works for Members? Can I assume that Reviewers, Approvers, and Leads are all in the 'MEMBER' author_association?

agilgur5: Yes, it's anyone in the Argoproj org. Note that we all also have the "Member" next to our name in comments.

Also that part hasn't changed in the PR 😅

juliev0: > Note that we all also have the "Member" next to our name in comments.
Got it!

> Also that part hasn't changed in the PR
Hey, I can still ask questions :)

agilgur5: For sure! I think I missed that you approved first and then commented

agilgur5:
> Verification

Can confirm this worked in https://github.com/argoproj/argo-workflows/pull/13317#issuecomment-2211986079

PR #13228, opened 2024-06-20
fix: parse exit code when Outputs is not populated

tooptoop4:
Relates to https://github.com/argoproj/argo-workflows/issues/12572#issuecomment-2162093707
and https://github.com/argoproj/argo-workflows/blame/465c7b6d6abd06a36165955d7fd01d9db2b6a2d4/workflow/controller/operator.go#L1621
Seems more of a workaround than addressing real root cause

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1097,8 +1097,10 @@ func (woc *wfOperationCtx) processNodeRetries(node *wfv1.NodeStatus, retryStrate
 		if err != nil {
 			return nil, false, err
 		}
-		if !shouldContinue && lastChildNode.Fulfilled() {
+		if !shouldContinue {
 			return woc.markNodePhase(node.Name, lastChildNode.Phase, "retryStrategy.expression evaluated to false"), true, nil
+		} else {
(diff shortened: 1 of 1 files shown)


agilgur5:
Not sure why you filed a duplicate of #13180... It is once again missing DCO as well

tooptoop4:
so after running this for a while it seems to never extract an exit code, but the extra debug logs are handy. whenever exit is -1 which is most of the time, message is empty 98% of the time or something like "lastRetry.duration:57.383139809 lastRetry.exitCode:-1 lastRetry.message:child 'redact-1335299960' failed lastRetry.status:Failed retries:1"          seems there is some broader issue of why it is not finding the right message/exitcode

jswxstw:
I don't understand why we need to extract the exit code from the message.
Under what circumstances would the exit code be nil but appear in the message?

tooptoop4:
from above tests it is useless, but these debug logs did help uncover steps referencing a template defined in the same yaml are a reliable reproduction of exitcode not being set, maybe i should make issue for that

tooptoop4:
this PR did not solve any issue but it helped to investigate exactly what the root cause is, which is now being tracked in https://github.com/argoproj/argo-workflows/issues/13297

PR #13232, opened 2024-06-21
feat!: OpenTelemetry metrics

Joibel:
Note to reviewers, this is a draft to check CI at the moment.
---
Led by #12589.

### Motivation

As discussed in #12589 OpenTelemetry is the future of observability. This PR changes the underlying codebase to collect metrics using the otel libraries, whilst attempting to retain compatibility with prometheus scraping where this makes sense.

It helps lay the groundwork for adding workflow tracing using otel.

This PR amends and extends the built in metrics to be more useful and correctly named.

This PR does not attempt do do anything else with otel, nor does it attempt to change custom metrics.

### Modifications

Mostly removed prometheus libraries, and replaced with otel libraries.

Allows use of both prometheus `/metrics` scraping and opentelemetry protocol transmission.

See [the migration guide](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics-3.6.md) for a quick summary of the changes.

See [the metrics documentation](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics.md) for more details on new metrics.

Extends the workqueue metrics to include all of them exposed by [client-go/workqueue](https://github.com/kubernetes/client-go/blob/babfe96631905312244f34500d2fd1a8a10b186c/util/workqueue/metrics.go#L173)

Allows removal of metrics, removal of attributes (labels) and reconfiguration of histogram buckets for all metrics using `options` in the [workflow-controller-configmap](https://github.com/Joibel/argo-workflows/blob/otel/docs/workflow-controller-configmap.yaml#L233)

Allows [temporality configuration](https://github.com/Joibel/argo-workflows/blob/otel/docs/workflow-controller-configmap.yaml#L231) for opentelemetry counters.

Adds a number of other new metrics:
* [build info](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics.md#build_info)
* [cronworkflow triggers](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics.md#cronworkflows_triggered_total)
* [k8s request duration](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics.md#k8s_request_duration)
* [workflow template runtimes](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics.md#workflowtemplate_runtime)
* [workflow template counts](https://github.com/Joibel/argo-workflows/blob/otel/docs/metrics.md#workflowtemplate_triggered_total)

Removed `argo_workflows_workflows_processed_count` as a non-functioning metric.

`originName` enhancements proposed in #12589 were not added.


docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+    workflowtemplate_runtime:
+      disable: true
+    k8s_request_duration:
+      disable: true
+    queue_duration:
+      disable: true
+    queue_longest_running:
+      disable: true
+    queue_retries:
+      disable: true}
```
agilgur5, line 59: typo
```suggestion
      disable: true
```

docs/metrics.md
```diff
@@ -17,6 +20,99 @@ for many cases; some examples:
...
+        - name
+    k8s_request_duration:
+      histogramBuckets: [ 1.0, 2.0, 10.0 ]
+```
+
+### Options
+
+Using options you can manipulate the metrics created by the workflow controller. These options apply to the built-in metrics and any custom metrics you create. Each option applies to the named metric only, and applies to all output methods.
+
+`disable: true` will disable the emission of the metric from the system.
```
agilgur5, line 97: Typically the convention in the ecosystem, such as Helm, is `enabled: false` -- note that this includes `metricsConfig.enabled` just above this.

The tense and convention is important to keep consistent and predictable IMO

docs/metrics.md
```diff
@@ -170,7 +441,7 @@ When defining a `histogram`, `buckets` must also be provided (see below).
 
 [Argo variables](variables.md) can be included anywhere in the metric spec, such as in `labels`, `name`, `help`, `when`, etc.
 
-Metric names can only contain alphanumeric characters, `_`, and `:`.
+Metric names can only contain alpha-numeric characters, `_`, and `:`.
```
agilgur5, line 444:
```suggestion
Metric names can only contain alphanumeric characters, `_`, and `:`.
```

"[alphanumeric](https://www.merriam-webster.com/dictionary/alphanumeric)" is a word with no hyphen

.spelling
```diff
@@ -120,6 +122,7 @@ Welch
 WorkflowTemplate
 WorkflowTemplates
 a.m.
+alpha-numerics
```
agilgur5, line 125:
```suggestion
```

per above

.spelling
```diff
@@ -193,6 +196,7 @@ parameterizing
 params
 pprof
 pre-commit
+prometheus
```
agilgur5, line 199: it seems like this isn't needed if this is properly capitalized everywhere? It should be capitalized, as it is a proper noun

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
+# Metrics upgrade notes
+
+Metrics have changed between 3.5 and 3.6.
```
agilgur5, line 3:
```suggestion
Metrics have changed in 3.6.
```

"between 3.5 and 3.6" is implicit, and it also affects versions prior to 3.5 as well, so remove that.
We also conventionally only state the version number things changed.

docs/metrics-3.6.md
agilgur5, line 1: Conventionally, this is placed in the Upgrading Guide (`docs/upgrading.md`)

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+* `queue_duration`
+* `queue_longest_running`
+* `queue_retries`
+* `queue_unfinished_work`
+* `pod_pending`
+
+and can be disabled with
+
+```yaml
+metricsConfig:
```
agilgur5, line 36:
```suggestion
metricsConfig: |
```

must be a string since this is in a ConfigMap

docs/metrics.md
```diff
@@ -58,67 +155,241 @@ at the current time, they should not be used to report historical data such as:
...
+### Counter, gauge and histogram
+
+The following terms have specific meanings for the metrics that we collect:
+
+- `counter`: A value that accumulates over time – you can think of this like an odometer on a car; it only ever goes up.
+    - The rate of change of counters can be a very powerful tool in understanding these metrics.
+- `gauge`: Measures a current value at the time it is read. An example would be the fuel gauge in a vehicle.
+- `histogram`: A client-side aggregation of values, such as request latencies. A histogram is a good choice if you are interested in value statistics. For example: How many requests take fewer than 1 second?
+
+Thanks to [OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments) for these explanations.
```
agilgur5, line 167:
```suggestion
These terms are [defined by OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments):

> - `counter`: A value that accumulates over time – you can think of this like an odometer on a car; it only ever goes up.
    - The rate of change of counters can be a very powerful tool in understanding these metrics.
> - `gauge`: Measures a current value at the time it is read. An example would be the fuel gauge in a vehicle.
> - `histogram`: A client-side aggregation of values, such as request latencies. A histogram is a good choice if you are interested in value statistics. For example: How many requests take fewer than 1 second?

```

This looks to be almost entirely a direct quote, so we should quote it. Though I would prefer to just link to it if we have no notable differences

docs/metrics.md
```diff
@@ -58,67 +155,241 @@ at the current time, they should not be used to report historical data such as:
...
+
 ### Default Controller Metrics
 
 Metrics for the Four Golden Signals are:
```
agilgur5, line 171:
```suggestion
Metrics for the [Four Golden Signals](https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals) are:
```

let's add a link while we're here

docs/metrics.md
```diff
@@ -58,67 +155,241 @@ at the current time, they should not be used to report historical data such as:
...
-- Traffic: `argo_workflows_count` and `argo_workflows_queue_depth_count`
-- Errors: `argo_workflows_count` and `argo_workflows_error_count`
-- Saturation: `argo_workflows_workers_busy` and `argo_workflows_workflow_condition`
+- Latency: `queue_latency`
+- Traffic: `gauge` and `queue_depth_gauge`
+- Errors: `count` and `error_count`
+- Saturation: `workers_busy` and `workflow_condition`
 
+!!! High cardinality
+    Some metric attributes may have high cardinality and are marked with ⚠️ to warn you. You may need to disable this metric or disable the attribute.
```
agilgur5, line 179:
```suggestion
    Some metric attributes may have high cardinality and are marked with ⚠️.
    You may want to disable the metric or attribute.
    
```

- We should use proper [`mkdocs-material` icons](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/) for this icon
- we should explain _why_ they may want to disable it, such as high storage usage or similar. Idk if there's a link we can provide for that, but that would be ideal
- Missing new line after note

docs/metrics.md
```diff
@@ -58,67 +155,241 @@ at the current time, they should not be used to report historical data such as:
...
+| `treestate` | Whether the git tree was `dirty` or `clean` when this was built        |
+| `tag`       | The tag on the git commit or `untagged` if it was not tagged           |
+
+#### `cronworkflows_triggered_total`
+
+A counter of the number of times a CronWorkflow has been
+
+| attribute   | explanation                               |
+|-------------|-------------------------------------------|
+| `name`     | ⚠️ The name of the CronWorkflow. |
```
agilgur5, line 203: I think it would make more sense to have the icon next to the attribute rather than its explanation -- much easier to see that way:

```suggestion
| ⚠️ `name`     | The name of the CronWorkflow. |
```

Also we should consider automatically generating all these from the go types as we do with the Fields Reference and others

docs/metrics.md
```diff
@@ -17,6 +20,99 @@ for many cases; some examples:
...
+
+## Metrics configuration
+
+It is possible to collect metrics via the OpenTelemetry protocol or via prometheus compatible scraping. Both of these mechanisms can be enabled at the same time, which could be useful if you'd like to migrate from one system to the other. Using multiple protocols at the same time is not intended for long term use however.
+
+OpenTelemetry is the recommended way of collecting metrics, and the OpenTelemetry collector can export metrics to prometheus via [the prometheus remote write exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/README.md).
+
+### OpenTelemetry protocol
+
+To enable the OpenTelemetry protocol you must set the environment variable `OTEL_EXPORTER_OTLP_ENDPOINT` or `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`. It will not default to the default value if left blank, instead no Open
```
agilgur5, line 33:
```suggestion
To enable the OpenTelemetry protocol you must set the environment variable `OTEL_EXPORTER_OTLP_ENDPOINT` or `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`.
It will not use the default value if left blank, instead no Open
```


- simplify, per k8s style guide
- prefer 1 sentence per line, per style guide
- incomplete sentence?

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+
+Metrics have changed between 3.5 and 3.6.
+
+You can now retrieve metrics using the OpenTelemetry Protocol using the [OpenTelemetry collector](https://opentelemetry.io/docs/collector/), and this is the recommended mechanism.
+
+These notes explain the differences in using the prometheus `/metrics` endpoint to scrape metrics for a minimal effort upgrade. It is not recommended you follow this guide blindly, the new metrics have been introduced because they add value, and so they should be worth collecting and using.
+
+## TLS
+
+The prometheus `/metrics` endpoint defaults to TLS on.
```
agilgur5, line 11:
```suggestion
The Prometheus `/metrics` endpoint now has TLS enabled by default.
```

- add "now" to specify that this has changed -- it sounded like it but wasn't explicit
- "enabled by default" matches naming conventions

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+
+To disable this set `.metricsConfig.secure` to `false`.
```
agilgur5, line 13:
```suggestion
To disable this set `metricsConfig.secure` to `false`.
```

The `.` prefix is not used anywhere else

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+      disable: true}
+    queue_unfinished_work:
+      disable: true
+    pod_pending:
+      disable: true
+```
+
+## Renamed metrics
+
+If you are using these metrics in your recording rules, dashboards or alerts you will need to use their new name after the upgrade:
```
agilgur5, line 68:
```suggestion
If you are using these metrics in your recording rules, dashboards, or alerts, you will need to update their names after the upgrade:
```

missing commas

Also I really think this section and the next one should be above "New Metrics" -- they are breaking changes and so are most notable, while the rest are additions. It should be in order of importance / highest impact to existing users

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+    queue_duration:
+      disable: true
+    queue_longest_running:
+      disable: true
+    queue_retries:
+      disable: true}
+    queue_unfinished_work:
+      disable: true
+    pod_pending:
+      disable: true
```
agilgur5, line 63:
```suggestion
```

we really only need to show one of these as an example, the rest are redundant. let's keep things simple and concise, per the k8s style guide

docs/metrics-3.6.md
```diff
@@ -0,0 +1,81 @@
...
+
+These notes explain the differences in using the prometheus `/metrics` endpoint to scrape metrics for a minimal effort upgrade. It is not recommended you follow this guide blindly, the new metrics have been introduced because they add value, and so they should be worth collecting and using.
+
+## TLS
+
+The prometheus `/metrics` endpoint defaults to TLS on.
+
+To disable this set `.metricsConfig.secure` to `false`.
+
+## New metrics
```
agilgur5, line 15: hmm, this isn't a breaking change... we typically don't describe additions in the upgrading notes

agilgur5 (changes requested):
I only briefly went through this, with particular focus on the docs and a bit on the config -- just an initial quick review. There's a good amount of docs changes needed.

As I've mentioned in other docs PRs you've had, please follow the docs style guides. In this quick look through, I particularly noticed many misses on the k8s style guide regarding ["Simple and direct language"](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language) and our style guide regarding ["Prefer 1 sentence per line of markdown"](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/).
The latter is explicit, while the former contains examples -- if you're struggling with that, a rough mindset you can use while reading/writing/editing is "can I say this with less words?" (while ofc still following the rest of the style guide, existing conventions, and without making things very confusing). The answer is very often "yes" as this style of writing can be quite different from the way we speak (you and I in particular use more "level of confidence" or "hedging" words as well as other descriptors). For instance, in the very preceding sentence, can remove "very", "quite", etc to simplify as: "Often this writing style is different from the way we speak".

docs/metrics.md
```diff
@@ -30,8 +126,9 @@ Default controller metrics can be scraped from service ```workflow-controller-me
 
 Metrics that inform on the state of a Workflow, or a series of Workflows. These custom metrics are defined by the user in the Workflow spec.
 
-Emitting custom metrics is the responsibility of the emitter owner. Since the user defines Workflows in Argo, the user is responsible
-for emitting metrics correctly.
+Emitting custom metrics is the responsibility of the emitter owner. Since the user defines Workflows in Argo, the user is responsible for emitting metrics correctly.
+
+Currently custom metrics and their labels must be valid prometheus and OpenTelemetry metric names, which limits them to alpha-numerics, `_`. This applies even if you're only using OpenTelemetry for metrics.
```
agilgur5, line 131:
```suggestion
Currently custom metrics and their labels must be valid Prometheus and OpenTelemetry metric names, which limits them to alphanumeric characters and `_`.
This applies even if you're only using OpenTelemetry for metrics.
```

- use consistent wording for "alphanumeric characters"
- fix missing "and"
- prefer 1 sentence per line of markdown, per style guide

This can also likely be simplified, per k8s style guide

docs/workflow-controller-configmap.yaml
```diff
@@ -225,8 +225,19 @@ data:
...
     # IgnoreErrors is a flag that instructs prometheus to ignore metric emission errors. Default is "false"
     ignoreErrors: false
-    # Use a self-signed cert for TLS, default false
-    secure: false
+    # Use a self-signed cert for TLS, default true
+    secure: true
+    # Which temporality to use for opentelemetry, defaults to Cumulative
+    temporality: Delta
+    # Options for configuring individual metrics
+    options:
```
agilgur5, line 233: I feel like there's a better name for this, but it is evading me right now

agilgur5 (commented):
Given how large this is, I wonder if we should split this into multiple PRs for readability & reviewability.

For instance:
- Prometheus -> OTel
- New metrics
- New configurations

docs/metrics-3.6.md
Joibel: It is a breaking change for anyone who doesn't explicitly set `secure` in their configmap.

docs/metrics.md
Joibel: Yeah, this and the 'options' naming were just where I'd ended up and wanted to see if my changes passed in proper CI.

docs/workflow-controller-configmap.yaml
Joibel: I've changed it to "modifiers" locally, which is the best I have for now.

docs/metrics.md
Joibel: Automatically generating would be lovely, but more scope creep, so not doing it now.

Joibel:
Thanks for all of the review comments.

I wasn't really ready for a review yet (as noted by this being in draft and the big comment at the top of the description), but I've tried to accommodate your comments.

Splitting this would be pretty non-fun given everything except for the base commit would want rebasing every time anything in the base commit changes. I'd tried to give fair warning in #12589 that I'd be doing what I've done.

The config changes are pretty small.
The new metrics are nearly all self contained.

docs/metrics.md
Joibel: I don't see an precedent for usage of non-emoji based icons. I prefer the emoji based ones because all text editors render them as icons.
An explanation of high cardinality is mentioned at least twice elsewhere including a link to an explanation. I felt it was redundant to explain this more.

.spelling
Joibel: Done.

docs/metrics.md
Joibel: We have no notable differences, but I felt it was worth including directly as it's obviously been an area of confusion and I want to ensure users fully grasp the concepts.

docs/metrics.md
Joibel: I disagree on the icon positioning, and prefer it with the description. It is in effect a concise part of the description, and I don't feel it is important enough to warrant moving left.

docs/metrics-3.6.md
Joibel: Sorted, thanks.

docs/metrics-3.6.md
Joibel: I've moved it. It's quite big, which is why I'd split it, but I don't mind either way.

docs/metrics-3.6.md
Joibel: Yep, sorted.

docs/metrics-3.6.md
agilgur5: I was commenting on the "New Metrics" section, not the TLS bit. It's also the longest section, so removing it would make the upgrade notes a lot shorter.

I think conventionally we just add new features to the existing docs (no specific upgrade note for them in the docs, though maybe in the release blog post) and add the "v3.6 and after" modifier where needed.

docs/metrics.md
agilgur5: Since you did write a "High Cardinality" warning already, it did sound important enough 😅

docs/workflow-controller-configmap.yaml
agilgur5: That does sound better, although also feels like it could be better 😅

agilgur5:
> I wasn't really ready for a review yet (as noted by this being in draft and the big comment at the top of the description), but I've tried to accommodate your comments.

Yes I noticed that of course. I didn't go through the code at all, so kept it high-level and not CI impacting. I figure you might want early feedback if you put up a draft and that early feedback is always better than later feedback. You don't have to respond to the comments immediately, we can just keep a running list of feedback until it's more ready and then can address them -- usually, in my experience, drafts have a slower review cycle on both ends.

agilgur5:
> Splitting this would be pretty non-fun given everything except for the base commit would want rebasing every time anything in the base commit changes.

PR stacks a la Graphite or Sapling (etc) can help automate that (both are overlays on top of Git; they're not as useful without their respective review tools though, and dedicated review tools are better than GitHub code review 9 times out of 10). I still do it all manually myself though, but it's exactly what PR stacks are designed for.
But the PRs sequential and largely contained, as you wrote, so there wouldn't be too many conflicts and they'd still be reviewed sequentially given that GH does not support PR stacks (so a subsequent diff will contain everything until the prior one is merged).

> I'd tried to give fair warning in #12589 that I'd be doing what I've done.

As the main reviewer there, I'm well aware of that. That doesn't change any readability or reviewability concerns.
The config changes in particular are new and expose a new API surface that we'd want to try to keep more stable.
There are also more new metrics than you originally wrote about.

> The config changes are pretty small.
> The new metrics are nearly all self contained.

At ~5k LoC changed across 64 files -- with almost none of it being generated code either -- anything can help. The new metrics in particular I would think might not need much iteration, but the OTel, config, and docs changes definitely need a careful eye given how fundamental they are as well as to reduce chance of future breakage and make as understandable to the user as possible.

docs/metrics-3.6.md
Joibel: New metrics will get consumed into your data collection tool by default, which may be costly. It feels important to warn people that they could be in for an increased bill by upgrading if they don't do something about it. I'd only expect people to read the upgrading guide.

Joibel:
Replaced by #13265

PR #13234, opened 2024-06-22
ci(changelog): list the tag in the PR title

agilgur5:
### Motivation

Make the automated changelog PRs have distinct titles for easy at-a-glance readability
- This is what I've manually done for #12720, #12956, #13096, and #13206
  - I also retroactively updated many [older changelog PRs](https://github.com/argoproj/argo-workflows/pulls?q=is%3Apr+changelog+is%3Aclosed+author%3Agithub-actions%5Bbot%5D) to include those as well

### Modifications

- add the tag in the PR title and commit subject for the `changelog.yaml` GHA Workflow
  - use GHA [`github` context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) for the ref name

- also "updated" -> "update" for active voice/present tense

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -28,7 +28,7 @@ jobs:
       - run: ./hack/changelog.sh > CHANGELOG.md
       - uses: peter-evans/create-pull-request@a4f52f8033a6168103c2538976c07b467e8163bc # v6.0.1
         with:
-          title: 'docs: updated CHANGELOG.md'
-          commit-message: 'docs: updated CHANGELOG.md'
+          title: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
+          commit-message: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
           branch: create-pull-request/changelog
           signoff: true


terrytangyuan (changes requested):
Could you merge multiple PRs into one?

agilgur5:
This and #13235 are independent of each other, so that would not be best practice, as said several times previously.

.github/workflows/changelog.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       - run: ./hack/changelog.sh > CHANGELOG.md
       - uses: peter-evans/create-pull-request@a4f52f8033a6168103c2538976c07b467e8163bc # v6.0.1
         with:
-          title: 'docs: updated CHANGELOG.md'
-          commit-message: 'docs: updated CHANGELOG.md'
+          title: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
+          commit-message: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
```
juliev0, line 32: I guess there's no good way to test this, huh? In that case, I just wanted to make sure there was no risk of `${{ github.ref_name }}` not getting evaluated or not getting evaluated correctly, but I do see it listed [here](https://docs.github.com/en/actions/learn-github-actions/contexts) as "The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub." which in this case should be the tag. So, I think I'm good.

juliev0 (approved):

.github/workflows/changelog.yaml
agilgur5: Other than making a fork and triggering actions (which is _very_ tedious) unfortunately not. 

[`act`](https://github.com/nektos/act) is the closest equivalent for local running, and it doesn't quite support everything either; not sure if it would work for this one.

agilgur5:
@terrytangyuan this is blocked on your request changes

terrytangyuan (approved):

agilgur5:
Can confirm this worked for #13423 . Only thing is that, per https://github.com/argoproj/argo-workflows/pull/13414#issuecomment-2257543521, this PR must apparently be cherry-picked into the release branch for it to take effect from a push of a tag off that branch.

PR #13235, opened 2024-06-22
build(docs): include full changelog and all contributors

agilgur5:
Partial fix for https://github.com/argoproj/argo-workflows/pull/13206#discussion_r1643723732

### Motivation

- there's a filter on the commits in the changelog generation script, but the CHANGELOG does not mention that

### Modifications

- say "Selected Changes" to explicitly note that it is not full
  - and put a link to the Full Changelog/Full Diff via GH compare view above that

- include all Contributors to the release, not just the filtered ones
  - also collapse the list of contributors by default with the HTML `<details>` tag

### Verification

I ran the `changelog.sh` script locally and committed the resulting CHANGELOG.md in this PR.
- See the [rendered commit](https://github.com/argoproj/argo-workflows/blob/6ee1565702071b919de7e273742afc8811e9f656/CHANGELOG.md)
- I also double-checked the entirety of the diff myself

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v3.5.8 (2024-06-17)
 
+Full Changelog: [v3.5.7...v3.5.8](https://github.com/argoproj/argo-workflows/compare/v3.5.7...v3.5.8)
+
+### Selected Changes
+
 * [d13891154](https://github.com/argoproj/argo-workflows/commit/d1389115484f52d22d1cdcae29139518cbf2fc03) fix(deps): bump `github.com/Azure/azure-sdk-for-go/sdk/azidentity` from 1.5.1 to 1.6.0 to fix CVE (#13197)
 * [10488d655](https://github.com/argoproj/argo-workflows/commit/10488d655a78c28bb6e3e6bca490a5496addd605) fix: don't necessarily include all artifacts from templates in node outputs (#13066)
 * [c2204ae03](https://github.com/argoproj/argo-workflows/commit/c2204ae03de97acf1c589c898180bdb9942f1524) fix(server): don't use cluster scope list + watch in namespaced mode. Fixes #13177 (#13189)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1100,15 +1291,22 @@
 * Ruin09
 * Son Bui
 * Suraj Banakar(बानकर) | スラジ
+* Yuan (Terry) Tang
```
terrytangyuan, line 1294: Hmm I wonder why this shows up in addition to "Yuan Tang" below

agilgur5: It's probably because you had a different name in your commit author previously. Since the script uses pure `git` (no GitHub to do, for instance, email -> GH user), it pulls entirely from the commit metadata in the repo

terrytangyuan: Is there a way to hack/merge these different ones?

agilgur5: Not without making the script even hackier than it already is 😬 
I would prefer to use GH IDs where possible, but that would require some non-trivial work to do

You'd effectively have to make a dictionary of author names to reference and choose one by some preference (I guess latest one), which means an associative array in Bash, which requires Bash 4+. So another loop and a dictionary lookup. 

I did notice this in the diff, but thought it was nbd, especially since this PR collapses the Contributors section anyway. Your name here though, so a bit different I get it

agilgur5: > You'd effectively have to make a dictionary of author names to reference and choose one by some preference (I guess latest one), which means an associative array in Bash, which requires Bash 4+. So another loop and a dictionary lookup.

Oh wait, the author name and email are the (non-unique) identifier, so you'd actually need an associative array for old name to new name (which there could be multiple matches for and which would be difficult to tell). In other words, hacky 🙃

agilgur5: IMO, I'd punt this until we can do GH ID lookups, but that's not perfect either with regard to historical emails/IDs. Which, to be explicit, probably wouldn't happen anytime soon given the effort to impact ratio

juliev0 (approved):

PR #13240, opened 2024-06-24
docs(sso): clarify Dex `staticClients` ID for `argo-cd/values.yaml` block

sahil-sharma:
### Motivation

Fixing wordings for SSO with ArgoCD for client_id

### Modifications

Add a comment in argo-cd/values.yaml in documentation to reflect that `client_id` should be the actual OIDC client-id received from OIDC provider and it should be in plain-text format.
This info was missing earlier.

### DIFF
--- a/docs/argo-server-sso-argocd.md
+++ b/docs/argo-server-sso-argocd.md
@@ -126,6 +126,7 @@ data:
        config:
          dex.config: |
            staticClients:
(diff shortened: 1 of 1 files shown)


agilgur5:
Basically a copy of #12843 but for the Helm Chart

agilgur5 (approved):
Thanks for adding this is another location

PR #13250, opened 2024-06-26
fix(cli): `argo lint` with strict should report case-sensitive errors. Fixes #13006

### LINKED ISSUE #13006: `argo lint` is not case sensitive
### Pre-requisites

When attempting to lint a `CronWorkflow` definition file, `argo lint` does not detect keys with incorrect capitalisation. If these are applied using `kubectl`, then the cron workflow does not work as expected. Using `argo cron create` works, as this tool appears to correct the capitalisation.

See example below, with upper case `U` in `schedUle`. I would expect `argo lint` to produce an error. Instead it produces the following:

```
➜  argo lint cronWorkflow.yaml
✔ no linting errors found!
```

### Version

v3.5.6

```YAML
kind: CronWorkflow
metadata:
  name: test-cronworkflow
  namespace: argo
spec:
  schedUle: "30 13 * * *"
  timezone: "Europe/London"
  workflowspec:
    entrypoint: main
    templates:
      - name: main
        steps:
          - - name: 1-test-echo
              template: test-echo
      - name: test-echo
        container:
          image: alpine:3.7
          command: [echo, "Hello world"]
```

```text
N/A
```

```text
N/A
```

miltalex:
Fixes #13006

## Motivation

Currently, when linting a `CronWorkflow` definition file, `argo lint` fails to detect keys with incorrect capitalization. This oversight can lead to unexpected behavior when applying these definitions using `kubectl`. Prior to this fix, `argo cron create` would ignore capitalization issues and proceed with the provided values, creating an inconsistency between `kubectl` and `argo cron create` behaviors.

By addressing this issue, we aim to:

1. Enhance the linting process to catch capitalization errors early in the development cycle.
2. Align the behavior of `argo cron create/argo submit` with `kubectl`, ensuring both tools report errors for incorrectly capitalized fields.
3. Streamline the workflow creation process by providing consistent feedback across different Argo and Kubernetes tools.

This change will help prevent potential errors by catching and reporting capitalization issues uniformly across all relevant tools.

## Modifications

To address this issue, we've made the following change:

Modified the strict path of the linting process to utilize the `sigs.k8s.io/json` library for unmarshaling.
   - This library provides case-sensitive unmarshaling in strict mode.
   - Fields with incorrect capitalization (e.g., "schedUle" instead of "schedule") will now be unmarshaled as-is.
   - The strict mode will then report these incorrectly capitalized fields as errors using the `k8s.io/apimachinery/pkg/runtime` library.

This change ensures that `argo lint` will now flag incorrectly capitalized fields when strict mode is enabled, providing developers with immediate feedback on potential issues in their workflow definitions. It also aligns the behavior of `argo lint` and `argo cron create/argo submit` with `kubectl`, creating a consistent experience across these tools.

## Verification

To validate the effectiveness of this fix, I have extended the test suite with two new unit tests:

1. A test case for strict mode enabled, verifying that capitalization errors are correctly reported.
2. A test case for strict mode disabled, ensuring that the existing behavior remains unchanged.

### DIFF
--- a/cmd/argo/commands/lint_test.go
+++ b/cmd/argo/commands/lint_test.go
@@ -172,4 +172,45 @@ spec:
 
 		assert.False(t, fatal, "should not have exited")
 	})
+
+	workflowCaseSensitivePath := filepath.Join(subdir, "workflowCaseSensitive.yaml")
+	err = os.WriteFile(workflowCaseSensitivePath, []byte(`
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: hello-world-
+spec:
+  entrypoInt: whalesay
+  templates:
+    - name: whalesay
+      container:
+        image: docker/whalesay
+        command: [ cowsay ]
+        args: [ "hello world" ]
+        resources:
+          limits:
+            memory: 32Mi
+            cpu: 100m
+`), 0644)
+	require.NoError(t, err)
+
+	t.Run("linting a workflow with case sensitive fields and strict enabled", func(t *testing.T) {
+		defer func() { logrus.StandardLogger().ExitFunc = nil }()
+		var fatal bool
+		logrus.StandardLogger().ExitFunc = func(int) { fatal = true }
+
+		runLint(context.Background(), []string{workflowCaseSensitivePath}, true, nil, "pretty", true)
+
+		assert.True(t, fatal, "should have exited")
+	})
+
+	t.Run("linting a workflow with case sensitive fields and strict disabled", func(t *testing.T) {
+		defer func() { logrus.StandardLogger().ExitFunc = nil }()
+		var fatal bool
+		logrus.StandardLogger().ExitFunc = func(int) { fatal = true }
+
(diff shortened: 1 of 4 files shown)


Joibel (approved):
I am happy with this.

@agilgur5, I notice you were discussing making `pkg` do this strict check instead of doing it in here. That feels slightly better as an approach, but I don't want to block this on that. Would you rather we went that way?

agilgur5:
> @agilgur5, I notice you were discussing making `pkg` do this strict check instead of doing it in here. That feels slightly better as an approach, but I don't want to block this on that. Would you rather we went that way?

In https://github.com/argoproj/argo-workflows/issues/13006#issuecomment-2156786547 I was primarily concerned about "strict + case-sensitive". If the k8s variant handles this, then perhaps the functions in `pkg` are not even necessary and can be replaced and removed.

But if we wanted to, we could add this as an _additional_ function in `pkg` and then load from the same place. For this case, I don't mind either way; would in fact prefer replacing and removing if possible.

PR #13263, opened 2024-06-28
fix: oss internal error should retry. Fixes #13262

### LINKED ISSUE #13262: OSS InternalError also should retry
### Pre-requisites

Sometimes the OSS pressure is too great and an internal error may occur. In this case, retrying can solve the problem.
```
time="2024-06-20T04:04:59 UTC" level=info msg="Save artifact" artifactName=abcd 
duration=12.543267547s error=
"oss: service returned error: StatusCode=500, ErrorCode=InternalError, ErrorMessage=\"Please contact the server administrator, oss@service.
aliyun.com\", RequestId=6673AA5F82C3D334359A2202, Ec=0001-00000000" 
```

### Version

latest

```YAML
any workflow
```

```text
```

```text
```

shuangkun:
Fixes #13262

### DIFF
--- a/workflow/artifacts/oss/oss.go
+++ b/workflow/artifacts/oss/oss.go
@@ -40,7 +40,7 @@ var (
 	defaultRetry                       = wait.Backoff{Duration: time.Second * 2, Factor: 2.0, Steps: 5, Jitter: 0.1}
 
 	// OSS error code reference: https://error-center.alibabacloud.com/status/product/Oss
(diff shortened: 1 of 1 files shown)


agilgur5 (changes requested):
See my comment in https://github.com/argoproj/argo-workflows/issues/13262#issuecomment-2198281099. I think this should be fixed upstream to return a different error, not by all the various clients

agilgur5 (approved):
LGTM as apparently the docs say to retry, see the issue comment https://github.com/argoproj/argo-workflows/issues/13262#issuecomment-2199023381 and further for some links. Potentially still needs further work upstream, but we'll need to support old versions in any case

Issue #7052, opened 2021-10-25
Enable TLS by default on Metrics/Telemetry


simster7:
# Summary

What change needs making?

Enable TLS by default on Metrics/Telemetry

# Use Cases

When would you use this?

Following ToB recommendation and decision in: https://github.com/argoproj/argo-workflows/pull/7041

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

PR #13266, opened 2024-06-28
feat!: enable Prometheus TLS by default

Joibel:
This appears to [have been slated for 3.5](https://github.com/argoproj/argo-workflows/blob/0322579ac813f345562f384ed6203a7f45faa484/workflow/controller/controller.go#L1361) but it didn't happen.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -68,8 +68,8 @@ metricsConfig: |
   # IgnoreErrors is a flag that instructs prometheus to ignore metric emission errors. Default is "false"
   ignoreErrors: false
 
-  # Use a self-signed cert for TLS
-  secure: false
+  # Use a self-signed cert for TLS, default true
+  secure: true
 ```
 
 The metric names emitted by this mechanism are prefixed with `argo_workflows_`.
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -68,7 +68,8 @@ metricsConfig: |
   # IgnoreErrors is a flag that instructs prometheus to ignore metric emission errors. Default is "false"
   ignoreErrors: false
 
-  # Use a self-signed cert for TLS, default true
+  # Use a self-signed cert for TLS
+  # >= 3.6: default true
   secure: true
 ```
 
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -52,7 +52,6 @@ Custom metrics, as defined by a workflow, could be defined as one type (say coun
 #### TLS
 
 The Prometheus `/metrics` endpoint now has TLS enabled by default.
-
 To disable this set `metricsConfig.secure` to `false`.
 
 ## Upgrading to v3.5
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -49,6 +49,12 @@ Custom metrics, as defined by a workflow, could be defined as one type (say coun
 
 `metricsTTL` for histogram metrics is not functional as opentelemetry doesn't allow deletion of metrics. This is faked via asynchronous meters for the other metric types.
 
+#### TLS
+
+The Prometheus `/metrics` endpoint now has TLS enabled by default.
+
+To disable this set `metricsConfig.secure` to `false`.
+
 ## Upgrading to v3.5
 
 There are no known breaking changes in this release.
--- a/docs/workflow-controller-configmap.yaml
+++ b/docs/workflow-controller-configmap.yaml
@@ -225,8 +225,8 @@ data:
     metricsTTL: "10m"
     # IgnoreErrors is a flag that instructs prometheus to ignore metric emission errors. Default is "false"
     ignoreErrors: false
-    # Use a self-signed cert for TLS, default false
-    secure: false
+    # Use a self-signed cert for TLS, default true
+    secure: true
 
     # DEPRECATED: Legacy metrics are now removed, this field is ignored
     disableLegacy: false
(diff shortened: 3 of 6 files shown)


Joibel:
Ah, fun. We don't run CI on things going to branches.

workflow/controller/controller.go
```diff
@@ -1359,8 +1359,8 @@ func (wfc *WorkflowController) getMetricsServerConfig() *metrics.Config {
 		Port:         wfc.Config.MetricsConfig.Port,
 		TTL:          time.Duration(wfc.Config.MetricsConfig.MetricsTTL),
 		IgnoreErrors: wfc.Config.MetricsConfig.IgnoreErrors,
-		// Default to false until v3.5
-		Secure: wfc.Config.MetricsConfig.GetSecure(false),
+		// Enable by default in 3.6
```
agilgur5, line 1362:
```suggestion
```

The previous comment was basically a TODO, right? So we don't need it anymore

docs/workflow-controller-configmap.yaml
```diff
@@ -225,8 +225,8 @@ data:
     metricsTTL: "10m"
     # IgnoreErrors is a flag that instructs prometheus to ignore metric emission errors. Default is "false"
     ignoreErrors: false
-    # Use a self-signed cert for TLS, default false
-    secure: false
+    # Use a self-signed cert for TLS, default true
```
agilgur5, line 228: this doc also has versions listed, so we should continue to do that:
```suggestion
    # Use a self-signed cert for TLS
    # >= 3.6: default true
```

docs/workflow-controller-configmap.yaml
```diff
@@ -225,8 +225,8 @@ data:
...
+    secure: true
 
     # DEPRECATED: Legacy metrics are now removed, this field is ignored
     disableLegacy: false
```
agilgur5, line 232: probably makes sense to remove this one of your PRs as well

docs/metrics.md
```diff
@@ -68,8 +68,8 @@ metricsConfig: |
   # IgnoreErrors is a flag that instructs prometheus to ignore metric emission errors. Default is "false"
   ignoreErrors: false
 
-  # Use a self-signed cert for TLS
-  secure: false
+  # Use a self-signed cert for TLS, default true
```
agilgur5, line 71:
```suggestion
  # Use a self-signed cert for TLS
  # >= 3.6: default true
```

same as above, since this is a copy+paste of the doc

docs/upgrading.md
```diff
@@ -47,6 +47,12 @@ Custom metric names and labels must be valid Prometheus and OpenTelemetry names
 
 Custom metrics, as defined by a workflow, could be defined as one type (say counter) in one workflow, and then as a histogram of the same name in a different workflow. This would work in 3.5 if the first usage of the metric had reached TTL and been deleted. This will no-longer work in 3.6, and custom metrics may not be redefined. It doesn't really make sense to change a metric in this way, and the OpenTelemetry SDK prevents you from doing so.
 
+#### TLS
+
+The Prometheus `/metrics` endpoint now has TLS enabled by default.
+
+To disable this set `metricsConfig.secure` to `false`.
```
agilgur5, line 54:
```suggestion
To disable this set `metricsConfig.secure` to `false`.
```
nit: Following the prior upgrading notes from `--base-href`, no new line when listing how to use prior behavior. Saves some space.

Also TIL [no comma if <4 words in introductory preposition](https://www.grammarly.com/blog/commas-after-introductory-phrases/#:~:text=When%20an%20introductory%20prepositional%20phrase,four%20words%2C%20use%20a%20comma.)

agilgur5 (commented):
Some small comments, otherwise LGTM

docs/workflow-controller-configmap.yaml
Joibel: This could be separated from this stack.

We'd break compatibility if we remove it from the golang struct too, but I'm inclined to do that anyway. Thoughts?

Joibel:
> Ah, fun. We don't run CI on things going to branches.

My intention is to fold all these down into the base #13265 which will be the last to merge. It'll then CI the stack together.

agilgur5:
I would actually do the opposite -- once the `opentelemetry` branch is merged, the base branch here will automatically change to `main`.

The opposite also has the benefits of decoupling the first PR; it would be merged first, not last, and there is never a large combined diff to approve either

docs/workflow-controller-configmap.yaml
agilgur5: > This could be separated from this stack.

Yea can be, if it breaks compat then it'd be more of a timing thing to release together.

> We'd break compatibility if we remove it from the golang struct too, but I'm inclined to do that anyway. Thoughts?

It is? I thought it was just dated docs at this point 😅
But yes still worth removing, though I guess we might need an upgrade note?

agilgur5:
> This appears to [have been slated for 3.5](https://github.com/argoproj/argo-workflows/blob/0322579ac813f345562f384ed6203a7f45faa484/workflow/controller/controller.go#L1361) but it didn't happen.

Fixes #7052 apparently. Alex mentioned that the manifests might need changing there? It's the same port. And I actually couldn't find any metrics Service in the manifests; only one is [in the metrics docs page](https://github.com/argoproj/argo-workflows/blob/0322579ac813f345562f384ed6203a7f45faa484/docs/metrics.md?plain=1#L305) and I don't think it needs updating since port and protocol (TCP) have not changed

agilgur5 (approved):
LGTM

PR #13267, opened 2024-06-28
feat: enable configuration of temporality in opentelemetry metrics

Joibel:
Implemented because of https://github.com/argoproj/argo-workflows/issues/12589#issuecomment-2174216603

Some backends (e.g. prometheus) only support cumulative temporality while others (e.g. dynatrace) only support delta metrics.

It is possible to configure an opentelemetry collector between argo workflow controller and the backend to convert cumulative metrics into delta metrics but it has some limits (e.g. when controller is restarted or when it is scaled up).

This commit enables the choice of temporality for OpenTelemetry

Temporality is always cumulative by definition in Prometheus

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -269,9 +269,9 @@ type MetricsConfig struct {
 	IgnoreErrors bool `json:"ignoreErrors,omitempty"`
 	// Secure is a flag that starts the metrics servers using TLS, defaults to true
 	Secure *bool `json:"secure,omitempty"`
-	// Temporality configures the temporality of the opentelemetry metrics.
-	// Valid values are Cumulative and Delta, defaulting to cumulative.
-	// This has no effect on prometheus metrics, which are always cumulative
+	// Temporality of the OpenTelemetry metrics.
+	// Enum of Cumulative or Delta, defaulting to Cumulative.
+	// No effect on Prometheus metrics, which are always Cumulative.
 	Temporality MetricsTemporality `json:"temporality,omitempty"`
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -245,6 +245,13 @@ type MySQLConfig struct {
 	Options map[string]string `json:"options,omitempty"`
 }
 
+type MetricsTemporality string
+
+const (
+	MetricsTemporalityCumulative MetricsTemporality = "Cumulative"
+	MetricsTemporalityDelta      MetricsTemporality = "Delta"
+)
+
 // MetricsConfig defines a config for a metrics server
 type MetricsConfig struct {
 	// Enabled controls metric emission. Default is true, set "enabled: false" to turn off
@@ -262,6 +269,10 @@ type MetricsConfig struct {
 	IgnoreErrors bool `json:"ignoreErrors,omitempty"`
 	// Secure is a flag that starts the metrics servers using TLS, defaults to true
 	Secure *bool `json:"secure,omitempty"`
+	// Temporality configures the temporality of the opentelemetry metrics.
+	// Valid values are Cumulative and Delta, defaulting to cumulative.
+	// This has no effect on prometheus metrics, which are always cumulative
+	Temporality MetricsTemporality `json:"temporality,omitempty"`
 }
 
 func (mc MetricsConfig) GetSecure(defaultValue bool) bool {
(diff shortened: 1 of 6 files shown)


docs/workflow-controller-configmap.yaml
```diff
@@ -228,6 +228,8 @@ data:
     # Use a self-signed cert for TLS
     # >= 3.6: default true
     secure: true
+    # Which temporality to use for opentelemetry, defaults to Cumulative
```
agilgur5, line 231:
```suggestion
    # >= 3.6. Which temporality to use for OpenTelemetry. Default is "Cumulative"
```

- add version number
- use "Default is 'x'" like the other comments here

agilgur5: An OTel link describing temporality would probably be helpful here too

docs/metrics.md
```diff
@@ -50,6 +50,14 @@ receivers:
 
 You can use the [OpenTelemetry operator](https://opentelemetry.io/docs/kubernetes/operator/) to setup the collector and instrument the workflow-controller.
 
+You can adjust temporality of the OpenTelemetry metrics configuration by changing values in the [Workflow Controller Config Map](workflow-controller-configmap.md).
```
agilgur5, line 53:
```suggestion
You can configure the temporality of OpenTelemetry metrics in the [Workflow Controller ConfigMap](workflow-controller-configmap.md).
```

- ConfigMap is one word (also UpperCamelCase, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#api-resource-names))
- simplify -- lot of redundant words here

agilgur5: OTel link for temporality would be good here too

config/config.go
```diff
@@ -262,6 +269,10 @@ type MetricsConfig struct {
 	IgnoreErrors bool `json:"ignoreErrors,omitempty"`
 	// Secure is a flag that starts the metrics servers using TLS, defaults to true
 	Secure *bool `json:"secure,omitempty"`
+	// Temporality configures the temporality of the opentelemetry metrics.
+	// Valid values are Cumulative and Delta, defaulting to cumulative.
+	// This has no effect on prometheus metrics, which are always cumulative
```
agilgur5, line 274:
```suggestion
	// Temporality of the OpenTelemetry metrics.
	// Enum of Cumulative or Delta, defaulting to Cumulative.
	// No effect on Prometheus metrics, which are always Cumulative.
```

- simplify
- consistent capitalization

config/config.go
```diff
@@ -245,6 +245,13 @@ type MySQLConfig struct {
 	Options map[string]string `json:"options,omitempty"`
 }
 
+type MetricsTemporality string
+
+const (
+	MetricsTemporalityCumulative MetricsTemporality = "Cumulative"
+	MetricsTemporalityDelta      MetricsTemporality = "Delta"
+)
```
agilgur5, line 253: could we import these enums from the OTel SDK?

workflow/metrics/metrics.go
```diff
@@ -117,3 +120,18 @@ func (m *Metrics) populate(ctx context.Context, adders ...addMetric) error {
 	}
 	return nil
 }
+
+func getTemporality(config *Config) metricsdk.TemporalitySelector {
+	switch config.Temporality {
+	case wfconfig.MetricsTemporalityCumulative:
+		return func(metricsdk.InstrumentKind) metricdata.Temporality {
+			return metricdata.CumulativeTemporality
+		}
```
agilgur5, line 129: is there no `metricdata.CumulativeTemporalitySelector`? since `Default` exists I would think the other two would too

config/config.go
Joibel: You might have thought, but they're not exported as textual enums.

workflow/metrics/metrics.go
Joibel: I'd also have thought this, but no.

docs/metrics.md
Joibel: Link added

agilgur5 (approved):
LGTM.

A bit odd that the OTel SDK doesn't have all the exports, maybe worth making a PR upstream to improve that as I would think we wouldn't be the only ones to want that.

PR #13268, opened 2024-06-28
feat: enable configuration of individual metrics

Joibel:
Allows removal of metrics, removal of attributes (labels), and reconfiguration of histogram buckets for all metrics using `modifiers` in the workflow-controller-configmap.

Implemented using opentelemetry views, these modifiers work for Prometheus metrics also


isubasinghe (approved):
LGTM

PR #13270, opened 2024-06-28
feat: new leader metric

Joibel:
Adds a leader metric which is 1 when this workflow controller is actively running workflows.

With Prometheus output the non-leader doesn't emit metrics. This is preserved in the new implementation, but OpenTelemetry does emit metrics.

Using this metric you can suppress or otherwise manipulate the metrics from a non-leader controller if you would like to.


agilgur5:
> With Prometheus output the non-leader doesn't emit metrics. This is preserved in the new implementation, but OpenTelemetry does emit metrics.

Is this not possible in OTel? IIRC that logic was entirely in our code for Prometheus, so I would think we could apply the same logic

Joibel:
> > With Prometheus output the non-leader doesn't emit metrics. This is preserved in the new implementation, but OpenTelemetry does emit metrics.
> 
> Is this not possible in OTel? IIRC that logic was entirely in our code for Prometheus, so I would think we could apply the same logic

Oh, we possibly could.

Having unmonitorable pods isn't something I consider a good idea though.

agilgur5:
> Having unmonitorable pods isn't something I consider a good idea though.

Very true. I would say we could disable "workflow" metrics (counts and such) on followers, but retain the standard Golang telemetry info and things like build info etc that are not impacted by leader election

Joibel:
> > Having unmonitorable pods isn't something I consider a good idea though.
> 
> Very true. I would say we could disable emitting "workflow" metrics (counts and such) on followers, but retain the standard Golang telemetry info and things like build info etc that are not impacted by leader election

We could, but we shouldn't.

What does it even mean for a counter to disappear and then reappear? What should the value be when it reappears? OpenTelemetry doesn't allow for deleting metrics, so we'd have to game the system to make this happen.

If someone wants to create this behaviour they can pop an otel-collector on as a sidecar and make it happen, but the reverse isn't true if we implement disappearing metrics.

Why do you think that it's a good idea to do this?

agilgur5:
> What does it even mean for a counter to disappear and then reappear? What should the value be when it reappears? OpenTelemetry doesn't allow for deleting metrics, so we'd have to game the system to make this happen.

We're not deleting the metrics, they're just not present / not emitted. This does already happen for temporary metrics. In a dashboard, they just disappear in the time range they're not present, which, in this case, is totally fine.

> Why do you think that it's a good idea to do this?

Primarily backward-compat to not change how users may already expect the follower metrics to work. It is simpler UX-wise as well; the user wouldn't have to account for it in their dashboards as the followers won't show up given they have no metrics

workflow/controller/controller.go
```diff
@@ -225,6 +225,7 @@ func NewWorkflowController(ctx context.Context, restConfig *rest.Config, kubecli
 			PodPhase:          wfc.getPodPhaseMetrics,
 			WorkflowPhase:     wfc.getWorkflowPhaseMetrics,
 			WorkflowCondition: wfc.getWorkflowConditionMetrics,
+			LeaderState:       wfc.IsLeader,
```
agilgur5, line 228: why not name `LeaderState` as `IsLeader` for consistency? Because of the `0`/`1` gauge? the type should make that inferable. Also already have `addIsLeader` as well

workflow/metrics/leader.go
```diff
@@ -0,0 +1,44 @@
...
+type leaderGauge struct {
+	callback LeaderStateCallback
+	gauge    *instrument
+}
+
+func addIsLeader(ctx context.Context, m *Metrics) error {
+	const nameLeader = `leader`
+	err := m.createInstrument(int64ObservableGauge,
+		nameLeader,
+		"Emits 1 if this is the leader when leader elections are enabled, or 0 otherwise. Always 1 when leader elections are disabled.",
```
agilgur5, line 20:
```suggestion
		"Emits 1 if leader, 0 otherwise. Always 1 if leader election is disabled.",
```

same as above

workflow/metrics/leader.go
```diff
@@ -0,0 +1,44 @@
...
+		return err
+	}
+	if m.callbacks.LeaderState != nil {
+		lGauge := leaderGauge{
+			callback: m.callbacks.LeaderState,
+			gauge:    m.allInstruments[nameLeader],
+		}
+		return m.allInstruments[nameLeader].registerCallback(m, lGauge.update)
+	}
+	return nil
```
agilgur5, line 34: would probably make sense to invert this conditional (and all other equivalent instances of it), no? Since it would only be `nil` for the test cases, right?

```suggestion
	if m.callbacks.LeaderState == nil {
		return nil
	}    

	lGauge := leaderGauge{
		callback: m.callbacks.LeaderState,
		gauge:    m.allInstruments[nameLeader],
	}
	return m.allInstruments[nameLeader].registerCallback(m, lGauge.update)
```

workflow/metrics/leader_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestIsLeader(t *testing.T) {
+	_, te, err := createTestMetrics(
+		&Config{},
+		Callbacks{
+			LeaderState: func() bool {
+				return true
+			},
+		})
+	if assert.NoError(t, err) {
```
agilgur5, line 18: we do this in a lot of places in the tests, but I feel like these conditionals should be inverted too.

won't block on that though since it's pre-existing style.

workflow/metrics/test_exporter.go
```diff
@@ -200,6 +217,16 @@ func (t *TestExporter) GetFloat64GaugeValue(name string, attribs *attribute.Set)
 	return gauge.Value, err
 }
 
+// GetFloat64GaugeValue returns an otel int64 gauge value
+// for reading in a test
```
agilgur5, line 221:
```suggestion
// GetInt64GaugeValue returns an otel int64 gauge value for test reads
```

- `Float` -> `Int`
- why not one line it?
- simplify a bit -- I also got a little confused by "reading in a test" as I interpreted that initially as "I read the test into the record"

workflow/metrics/test_exporter.go
```diff
@@ -139,6 +139,23 @@ func (t *TestExporter) getNamedFloat64GaugeData(name string, attribs *attribute.
 	return nil, fmt.Errorf("%s type gauge[float64] not found in %v", name, mtc)
 }
 
+func (t *TestExporter) getNamedInt64GaugeData(name string, attribs *attribute.Set) (*metricdata.DataPoint[int64], error) {
+	mtc, err := t.getNamedMetric(name)
```
agilgur5, line 143: isn't this usually `m`? vs. `mtc` which feels like a new abbreviation

workflow/metrics/test_exporter.go
```diff
@@ -139,6 +139,23 @@ func (t *TestExporter) getNamedFloat64GaugeData(name string, attribs *attribute.
...
+
+	if gauge, ok := mtc.Data.(metricdata.Gauge[int64]); ok {
+		for _, dataPoint := range gauge.DataPoints {
+			if dataPoint.Attributes.Equals(attribs) {
+				return &dataPoint, nil
+			}
+		}
+		return nil, fmt.Errorf("%s named gauge[float64] with attribs %v not found in %v", name, attribs, mtc)
+	}
+	return nil, fmt.Errorf("%s type gauge[float64] not found in %v", name, mtc)
```
agilgur5, line 156: similarly could invert the first conditional (which I believe is how it's usually done in Go):

```suggestion
	gauge, ok := mtc.Data.(metricdata.Gauge[int64])
	if !ok {
		return nil, fmt.Errorf("%s type gauge[float64] not found in %v", name, mtc)
	}

	for _, dataPoint := range gauge.DataPoints {
		if dataPoint.Attributes.Equals(attribs) {
			return &dataPoint, nil
		}
	}
	return nil, fmt.Errorf("%s named gauge[float64] with attribs %v not found in %v", name, attribs, mtc)
```

agilgur5 (commented):
primarily docs and style comments; content of the code otherwise LGTM

workflow/metrics/leader_test.go
agilgur5: or rather, these could all be replaced with `require.` statements instead of conditionals for the same effect but without nesting

Joibel: I prefer require

docs/metrics.md
```diff
@@ -248,6 +248,14 @@ A counter of the number of API requests sent to the Kubernetes API.
 | `verb`        | The verb of the request, such as `Get` or `List`                   |
 | `status_code` | The HTTP status code of the response                               |
 
+#### `leader`
```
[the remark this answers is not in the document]
Joibel: Yeah, I like this.

workflow/controller/controller.go
Joibel: Yeah, that's a decent call

workflow/metrics/test_exporter.go
Joibel: No, this isn't the same thing as the usual `m`, which is usually the "global metric object". This is a specific, opentelemetry internal, obejct type.

The difference in naming appears to have worked successfully here then.

Joibel: This kind of thing is better reviewed in #13265 really

agilgur5: Yea I noticed that some style originates from the first PR, but it is easier to understand as smaller bits 😅 
I'm getting to that one, I just need a lot larger chunk of time for it given how large it is (it also still doesn't adhere to the docs style guide per https://github.com/argoproj/argo-workflows/pull/13232#pullrequestreview-2133572454).

agilgur5 (commented):
Not sure if you missed them, but there's still [3](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1661311642) [unresolved](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1661330486) [suggestions](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1667248031) from my previous review

Joibel:
> Not sure if you missed them, but there's still [3](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1661311642) [unresolved](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1661330486) [suggestions](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1667248031) from my previous review

I missed one.

I'll resolve things myself with you in future - different people have different expectations and some places/people want to resolve things themselves when they are just comments!

agilgur5:
> > Not sure if you missed them, but there's still [3](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1661311642) [unresolved](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1661330486) [suggestions](https://github.com/argoproj/argo-workflows/pull/13270#discussion_r1667248031) from my previous review
> 
> I missed one.

I'm a bit confused, you agreed with the other two but didn't make the changes? The metric is still `leader` vs. `is_leader` and there's still a few `if assert.NoError` that can be replaced with `require.NoError`

> I'll resolve things myself with you in future - different people have different expectations and some places/people want to resolve things themselves when they are just comments!

I can also resolve them myself if you're fine with that. I try not to be too prescriptive as there can be a similar but different way of achieving a suggestion. So I generally don't do that unless someone's fine with that (or it's a new contributor and would be faster for me to make the changes myself etc)

Joibel:
Ah, I see. I have fixed those locally, but somehow it didn't end up in the stack correctly. I'll sort that, apologies for the confusion.

agilgur5 (approved):
LGTM

agilgur5:
> I can also resolve them myself if you're fine with that. I try not to be too prescriptive as there can be a similar but different way of achieving a suggestion. So I generally don't do that unless someone's fine with that (or it's a new contributor and would be faster for me to make the changes myself etc)

For future reference, wanted to clarify if this was fine?

PR #13271, opened 2024-06-28
feat: new k8s request duration histogram metric

Joibel:
This is a new metric which is a superset of the k8s_request_total
metric. The old metric is retained for backwards compatibility.

A common issue with workflows is that it stresses the kubernetes
API. Using this metric you can get insights into how long each type of
request is taking which should enable optimizations to help alleviate
problems in this area.

The new metric has the same attributes/labels of `kind`, `verb` and
`status_code` but now emits them with a time duration for each type of
response.

Note to reviewers: this is now a standlone commit

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -233,6 +233,20 @@ A counter of the number of API requests sent to the Kubernetes API.
 | `verb`        | The verb of the request, such as `Get` or `List`                   |
 | `status_code` | The HTTP status code of the response                               |
 
+This metric is calculable from `k8s_request_duration`, and it is suggested you just collect that metric instead.
+
+#### `k8s_request_duration`
+
+A histogram recording how long each type of request took.
+
+| attribute     | explanation                                                        |
+|---------------|--------------------------------------------------------------------|
+| `kind`        | The kubernetes `kind` involved in the request such as `configmaps` |
+| `verb`        | The verb of the request, such as `Get` or `List`                   |
+| `status_code` | The HTTP status code of the response                               |
+
+This is contains all the information contained in `k8s_request_total` along with timings.
+
 #### `is_leader`
 
 A gauge indicating if this Controller is the [leader](high-availability.md#workflow-controller).
--- a/workflow/metrics/metrics_k8s_request.go
+++ b/workflow/metrics/metrics_k8s_request.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"context"
 	"net/http"
+	"time"
 
 	"k8s.io/client-go/rest"
 
@@ -10,7 +11,8 @@ import (
 )
 
 const (
-	nameK8sRequestTotal = `k8s_request_total`
+	nameK8sRequestTotal    = `k8s_request_total`
+	nameK8sRequestDuration = `k8s_request_duration`
 )
 
 func addK8sRequests(_ context.Context, m *Metrics) error {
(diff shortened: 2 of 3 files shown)


isubasinghe (approved):
LGTM, just made a comment, I will leave the choice here up to you.

workflow/metrics/metrics_k8s_request.go
```diff
@@ -39,7 +48,9 @@ type metricsRoundTripper struct {
 var k8sMetrics metricsRoundTripper
 
 func (m metricsRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	startTime := time.Now()
 	x, err := m.roundTripper.RoundTrip(r)
+	duration := time.Since(startTime)
```
isubasinghe, line 53: it is not safe to assume monotonic time here, the duration value could be negative.
I would suggest dropping a negative `duration`, but will leave the decision up to you.

Joibel: My understanding is that this should be robust (on linux where a monotonic clock will exist).
https://pkg.go.dev/time#hdr-Monotonic_Clocks

Joibel: My problem with adding the "don't record below zero" is that it:
* Silently breaks this as a correct counter
* Acknowledges that the result is sometimes a lie. If it's sometimes a lie we can't tell when.

I don't think this happens, so I'm going to go with it as is.

docs/metrics.md
```diff
@@ -233,6 +233,20 @@ A counter of the number of API requests sent to the Kubernetes API.
...
+
+A histogram recording how long each type of request took.
+
+| attribute     | explanation                                                        |
+|---------------|--------------------------------------------------------------------|
+| `kind`        | The kubernetes `kind` involved in the request such as `configmaps` |
+| `verb`        | The verb of the request, such as `Get` or `List`                   |
+| `status_code` | The HTTP status code of the response                               |
+
+This is contains all the information contained in `k8s_request_total` along with timings.
```
agilgur5, line 248:
```suggestion
This metrics contains all data in `k8s_request_total` along with timings
```

There's a typo here. Also simplify.

agilgur5: Also is there a reason you put this below the table instead of above?

docs/metrics.md
```diff
@@ -233,6 +233,20 @@ A counter of the number of API requests sent to the Kubernetes API.
 | `verb`        | The verb of the request, such as `Get` or `List`                   |
 | `status_code` | The HTTP status code of the response                               |
 
+This metric is calculable from `k8s_request_duration`, and it is suggested you just collect that metric instead.
+
+#### `k8s_request_duration`
+
```
agilgur5, line 239:
```suggestion

> v3.6 and after

```

Version reference

docs/metrics.md
```diff
@@ -233,6 +233,20 @@ A counter of the number of API requests sent to the Kubernetes API.
 | `verb`        | The verb of the request, such as `Get` or `List`                   |
 | `status_code` | The HTTP status code of the response                               |
 
+This metric is calculable from `k8s_request_duration`, and it is suggested you just collect that metric instead.
```
agilgur5, line 236: Should this say that it is deprecated? I feel like that would be very relevant info to be above the table instead of below

```suggestion
_Deprecated_: Use `k8s_request_duration` instead.
```

agilgur5 (commented):
Docs comments

PR #13273, opened 2024-06-28
feat: new pod pending counter metric

Joibel:
The workflow controller is a kubernetes controller creating
pods. Sometimes those pods do not start, and will remain in pending.

This metric counts the number of pods that may have been observed as
pending, by namespace and truncated reason. The reason is the first
part of the kubernetes pod pending `Reason` up to the first `:` if
present.

It ignores all pods in the `PodInitializing` state as this I consider
unremarkable and temporary.

This is intended for users to create alerts on particular `reasons` or
if this metric is climbing unusually rapidly.

Note to reviewers: this is now a standalone commit

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -302,6 +302,19 @@ You should only see this under high load.
 
 `recently_started` is controlled by the [environment variable](environment-variables.md) `RECENTLY_STARTED_POD_DURATION` and defaults to 10 seconds.
 
+#### `pod_pending_count`
+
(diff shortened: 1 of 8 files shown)


isubasinghe (approved):
LGTM

agilgur5:
> It ignores all pods in the `PodInitializing` state as this I consider unremarkable and temporary.

IIRC there are some cases where users had Pods stuck in this state, so it might be useful for detecting those.

PR #13275, opened 2024-06-28
feat: new template counting and phase counting metrics

Joibel:
From #12589.

New metric `total_count` which is like the old `count` metric and the new `gauge` metric, but a counter, not a gauge. The gauge shows a snapshot of what is happening right now in the cluster, the counter can answer questions like how many `Failed` workflows have there been in the last 24 hours.

Two further metrics for counting uses of WorkflowTemplates via workflowTemplateRef only. These store the name of the WorkflowTemplate or ClusterWorkflowTemplate if the `cluster_scope` label is true, and the namespace where it is used. `workflowtemplate_triggered_total` counts the number of uses. `workflowtemplate_runtime` records how long
each phase the workflow running the template spent in seconds.

Note to reviewers: this is now a standlone commit


agilgur5:
> Note to reviewers: this is part of a stack of reviews for metrics changes. Please don't merge until the rest of the stack is also ready.

@Joibel since these ones aren't draft, does that mean they're ready to review but not ready to merge?

What we can do to indicate that is to add a "[DO NOT MERGE] " prefix to the title of each of these -- since that will fail the PR title check, it won't be mergeable and is a pretty big indicator too. Or "[Please Review but DO NOT MERGE] " or something

Joibel:
These are ready for review.

Merging isn't a disaster, it just makes any merged into PRs slightly harder to review. I was hoping anyone reviewing would do enough reading not to merge with the existing words I've written.

The whole thing is an experiment in highly stacked PRs from my point of view.

workflow/metrics/counter_template.go
```diff
@@ -0,0 +1,33 @@
...
+
+func templateLabels(name, namespace string, cluster bool) instAttribs {
+	return instAttribs{
+		{name: labelTemplateName, value: name},
+		{name: labelTemplateNamespace, value: namespace},
+		{name: labelTemplateCluster, value: cluster},
+	}
+}
+
+func (m *Metrics) CountWorkflowTemplate(ctx context.Context, phase, name, namespace string, cluster bool) {
```
isubasinghe, line 28: The argument name `phase` is somewhat misleading here in my opinion. 
A comment and an enum for the `phase` argument would make sense here.

Because this is a different type effectively, one that inherits all possible values of the `Phase` field along with an additional `New` field`

isubasinghe (changes requested):
Just a small comment

workflow/metrics/counter_template.go
Joibel: I've done a new type for this. Still called it `phase` in here, I hope that makes sense.

isubasinghe (approved):

Issue #13105, opened 2024-05-28
Possible clear-text logging of sensitive information in CLI `token.go`


makashu:
# Summary

What change needs making?
CodeQL flagged cmd/argo/commands/auth/token.go as printing Sensitive data returned by HTTP request headers flows to a logging call on line 21.

## Use Cases

When would you use this?

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

makashu:
I am working on this issue. Thanks!

agilgur5:
> CodeQL flagged cmd/argo/commands/auth/token.go as printing Sensitive data returned by HTTP request headers flows to a logging call on line 21.

That's this line:
https://github.com/argoproj/argo-workflows/blob/f492ea3609297617a533f1905777bffa57731102/cmd/argo/commands/auth/token.go#L21

It does exactly what the command's help text says it does:
https://github.com/argoproj/argo-workflows/blob/f492ea3609297617a533f1905777bffa57731102/cmd/argo/commands/auth/token.go#L15

That's a false positive, it's literally intended to print out the token

PR #13277, opened 2024-06-28
Remove token log in `argo auth token` command

makashu:
Fix a potential issue if the authentication string contains any secrets or tokens by replacing with a log statement

Fixes #TODO

### DIFF
--- a/cmd/argo/commands/auth/token.go
+++ b/cmd/argo/commands/auth/token.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"fmt"
 	"os"
+	"log"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


agilgur5:
> Fixes #TODO

Fixes #13105 , that you had authored.

For the same reason I [closed the issue](https://github.com/argoproj/argo-workflows/issues/13105#issuecomment-2136251343), closing this PR as invalid. The logging of a Secret is intentional here, the command is intended to print the Secret

makashu:
> > Fixes #TODO
> 
> Fixes #13105 , that you had opened.
> 
> For the same reason I [closed the issue](https://github.com/argoproj/argo-workflows/issues/13105#issuecomment-2136251343), closing this PR as invalid. As I wrote there, the logging of a Secret is intentional in this command and does exactly what the command says it will do. Please do not ignore maintainer comments.

Apology for the late response and sorry if you felt that the comments were ignored. 
Regarding your comment that logging of secret is intentional, does that mean it needs to be printed for the end-user consumption? Even though it is intentional printing secret is that not a security concern and what if someone gets hold of that token? I agree that the code says so but would that not make sense to just provide informational message rather than the actual token itself?

PR #13280, opened 2024-06-29
fix(docs): clarify CronWorkflow `startingDeadlineSeconds`. Fixes #12971

### LINKED ISSUE #12971: Clarify docs for CronWorkflow `startingDeadlineSeconds`
### Pre-requisites

The [CronWorkflow docs](https://argo-workflows.readthedocs.io/en/latest/cron-workflows/#crash-recovery) describe `startingDeadlineSeconds` as:
> Mainly startingDeadlineSeconds can be set to specify the maximum number of seconds past the last successful run of a CronWorkflow during which a missed run will still be executed.
>
> For example, if a CronWorkflow that runs every minute is last run at 12:05:00, and the controller crashes between 12:05:55 and 12:06:05, then the expected execution time of 12:06:00 would be missed. However, if startingDeadlineSeconds is set to a value greater than 65 (the amount of time passing between the last scheduled run time of 12:05:00 and the current controller restart time of 12:06:05), then a single instance of the CronWorkflow will be executed exactly at 12:06:05.

This does not match the logic in the [implementation](https://github.com/argoproj/argo-workflows/blob/v3.5.5/workflow/cron/operator.go#L271-L278). Based on the code, the docs should be reworded to something like the following:
> Mainly startingDeadlineSeconds can be set to specify the maximum number of seconds past the last scheduled run of a CronWorkflow during which a missed run will still be executed.
>
> For example, if a CronWorkflow that runs every minute is last run at 12:05:00, and the controller crashes between 12:05:55 and 12:06:05, then the expected execution time of 12:06:00 would be missed. However, if startingDeadlineSeconds is set to a value greater than 5 (the amount of time passing between the last scheduled run time of 12:06:00 and the current time of 12:06:05), then a single instance of the CronWorkflow will be executed exactly at 12:06:05.

Similar changes should be made elsewhere, as the comparison is not against the last successful run, but rather the last desired/expected/scheduled run time.

tl;dr `startingDeadlineSeconds` should be clearly documented as the "grace period" beyond the scheduled run during which a CronWorkflow may still be executed.

### Version

v3.5.5

```YAML
n/a
```

```text
n/a
```

```text
n/a
```

qingfeng777:
Fixes #12971 

### Motivation

fix doc

### Modifications

Clarify docs for CronWorkflow  field startingDeadlineSeconds

### Verification

Run locally and check.


docs/cron-workflows.md
```diff
@@ -58,9 +58,9 @@ More detailed documentation for the specific library used is [documented here](h
 
 ### Crash Recovery
 
-If the `workflow-controller` crashes (and hence the `CronWorkflow` controller), there are some options you can set to ensure that `CronWorkflows` that would have been scheduled while the controller was down can still run. Mainly `startingDeadlineSeconds` can be set to specify the maximum number of seconds past the last successful run of a `CronWorkflow` during which a missed run will still be executed.
+If the `workflow-controller` crashes (and hence the `CronWorkflow` controller), there are some options you can set to ensure that `CronWorkflows` that would have been scheduled while the controller was down can still run. Mainly `startingDeadlineSeconds` can be set to specify the maximum number of seconds past the last scheduled run of a `CronWorkflow` during which a missed run will still be executed.
```
agilgur5, line 61:
```suggestion
If the Controller crashes, you can ensure that any missed schedules still run.

With `startingDeadlineSeconds` you can specify a maximum grace period past the last scheduled time during which it will still run.
```

- simplify, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - `CronWorkflows` is redundant, since this the Cron Workflow page
- use the phrase "grace period", as suggested in the issue

These style guides weren't used back when this doc was originally written, but we should update it to the latest style guides whenever we change it

docs/cron-workflows.md
```diff
@@ -58,9 +58,9 @@ More detailed documentation for the specific library used is [documented here](h
...
 
-For example, if a `CronWorkflow` that runs every minute is last run at 12:05:00, and the controller crashes between 12:05:55 and 12:06:05, then the expected execution time of 12:06:00 would be missed. However, if `startingDeadlineSeconds` is set to a value greater than 65 (the amount of time passing between the last scheduled run time of 12:05:00 and the current controller restart time of 12:06:05), then a single instance of the `CronWorkflow` will be executed exactly at 12:06:05.
+For example, if a `CronWorkflow` that runs every minute is last run at 12:05:00, and the controller crashes between 12:05:55 and 12:06:05, then the expected execution time of 12:06:00 would be missed. However, if `startingDeadlineSeconds` is set to a value greater than 5 (the amount of time passing between the last scheduled run time of 12:06:00 and the current time of 12:06:05), then a single instance of the `CronWorkflow` will be executed exactly at 12:06:05.
```
agilgur5, line 63:
```suggestion
For example, if a `CronWorkflow` that runs every minute is last run at 12:05:00, and the controller crashes between 12:05:55 and 12:06:05, then the expected execution time of 12:06:00 would be missed.
However, if `startingDeadlineSeconds` is set to a value greater than 5 (the time passed between the last scheduled time of 12:06:00 and the current time of 12:06:05), then a single instance of the `CronWorkflow` will be executed exactly at 12:06:05.
```

same as above

docs/cron-workflows.md
```diff
@@ -58,9 +58,9 @@ More detailed documentation for the specific library used is [documented here](h
```
agilgur5, line 58: I can't comment on unchanged lines but the description of `activeDeadlineSeconds` above should be updated as well

agilgur5: Also ideally the [Fields Reference](https://argo-workflows.readthedocs.io/en/latest/fields/) should be updated as well; who's source is in `cron_workflow_types.go`. Although that one isn't incorrect at least, it's just different

agilgur5 (commented):
Further simplifications & style changes below

agilgur5 (approved):
Thanks for adding the clarification to the docs

PR #13286, opened 2024-07-01
fix: don't clean up old offloaded records during save. Fixes: #13220

### LINKED ISSUE #13220: Stop a workflow during offload causes `upper: no more rows in this result set`
### Pre-requisites

When the workflow updates the nodestatus to the database and deletes the old records, if someone stops the workflow during the update, a conflict will occur. However, reapply cannot parse woc.orig at this time, resulting in upper: no more rows in this result set

### Version

latest

```YAML
Any large workflow
```

```text
```

```text
```

shuangkun:
don't clean up old records when save

Fixes #13220

### DIFF
--- a/persist/sqldb/offload_node_status_repo.go
+++ b/persist/sqldb/offload_node_status_repo.go
@@ -95,27 +95,8 @@ func (wdc *nodeOffloadRepo) Save(uid, namespace string, nodes wfv1.Nodes) (strin
 		}
 		logCtx.WithField("err", err).Info("Ignoring duplicate key error")
 	}
-
-	logCtx.Debug("Nodes offloaded, cleaning up old offloads")
-
-	// This might fail, which kind of fine (maybe a bug).
-	// It might not delete all records, which is also fine, as we always key on resource version.
-	// We also want to keep enough around so that we can service watches.
-	rs, err := wdc.session.SQL().
-		DeleteFrom(wdc.tableName).
-		Where(db.Cond{"clustername": wdc.clusterName}).
-		And(db.Cond{"uid": uid}).
-		And(db.Cond{"version <>": version}).
-		And(wdc.oldOffload()).
-		Exec()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


imliuda (approved):
We have encoutered the same issue, and the delete operation will cause many dead locks when there are many slow queries. I agree this this change.

persist/sqldb/offload_node_status_repo.go
```diff
@@ -95,27 +95,8 @@ func (wdc *nodeOffloadRepo) Save(uid, namespace string, nodes wfv1.Nodes) (strin
...
-		And(wdc.oldOffload()).
-		Exec()
-	if err != nil {
-		return "", err
-	}
-	rowsAffected, err := rs.RowsAffected()
-	if err != nil {
-		return "", err
-	}
-	logCtx.WithField("rowsAffected", rowsAffected).Debug("Deleted offloaded nodes")
```
juliev0, line 118: Makes sense what you're saying. So, I notice that `Save()` can be called in either of [these cases](https://github.com/argoproj/argo-workflows/blob/v3.5.8/workflow/hydrator/hydrator.go#L106). Just want to confirm that the normal deletion method will take care of deleting the old Archived Workflows in either of those cases?

shuangkun: Don‘t take care of. 
We already have a periodic cleanup mechanism, so I think we can remove the cleanup during the save phase, because it does sometimes delete things that should not be deleted.

juliev0: I understand. But I'm just wondering if you tested both of these cases: `if packer.IsTooLargeError(err)` as well as `alwaysOffloadNodeStatus`

shuangkun: Yes, both cases have passed the test.

juliev0 (approved):

PR #13294, opened 2024-07-02
fix: Only cleanup agent pod if exists. Fixes #12659

### LINKED ISSUE #12659: controller sending many pod delete requests that result in 404 response

jswxstw:
### Motivation

Fixes #12659 

### Modifications

Agent pod will only be created when taskset nodes exist, so do not delete it if it does not exist.

### Verification

local test and updated UT

### DIFF
--- a/workflow/controller/controller_test.go
+++ b/workflow/controller/controller_test.go
@@ -1141,7 +1141,6 @@ spec:
 
 	woc.operate(ctx)
 	assert.True(t, controller.processNextPodCleanupItem(ctx))
-	assert.True(t, controller.processNextPodCleanupItem(ctx))
 	assert.Equal(t, wfv1.WorkflowSucceeded, woc.wf.Status.Phase)
 	podCleanupKey := "test/my-wf/labelPodCompleted"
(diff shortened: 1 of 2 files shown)


agilgur5 (approved):
Follow-up to https://github.com/argoproj/argo-workflows/pull/12679#discussion_r1507237560, so LGTM.

Thanks for fixing the issue and finding the right function to check!

PR #13308, opened 2024-07-05
fix(ui): hide `Workflow gone` message when workflow is archived

### LINKED ISSUE #13305: Workflow Archiving: `Workflow gone` error shortly after Workflow completion
### Pre-requisites

**Issue Summary**:
Since enabling Workflow Archiving, in the UI we have been observing an error popup saying "Workflow gone" which appears shortly after a workflow completes in the UI. This issue does not seem to leave any relevant information in the argo-server logs.

**Steps to Reproduce**:
1. Enable Workflow Archiving.
2. Execute a workflow.
3. Observe the UI shortly after the workflow completes.

**Observed Behavior**:
- The error popup "Workflow gone" appears immediately after a workflow completes.
- This error seems to occur as soon as the workflow is added to Postgres.
- Refreshing the page makes the error message disappear. 
- The error has no negative effect except for a bad user experience.

**Expected Behavior**:
- The UI should not show a "Workflow gone" popup immediately after the workflow completes.
- The workflow should transition smoothly to the archived state without error popups in the UI.

**Additional Information**:
- Workflow TTL is configured to 1 hour, but the popup appears before the 1-hour TTL, right after the workflow completes.
- This issue seems to be a legacy bug from the 3.5 changeover to a single view of in-cluster and archived workflows.
- The issue is likely an annoyance for users and should be handled better in the UI.

**Screenshots**:
Screenshot 2024-07-04 at 5 38 12 PM

**Workaround**:
- Refreshing the page makes the UI error popup disappear.

**References**:
- Slack conversation reference with Alan Clucas:
    - Alan Clucas mentioned that the error message made sense when workflows were moving between tabs in the UI, but now it isn't relevant and should be handled better.

### Version

v3.5.8

```YAML
Any workflow that uses Workflow Archiving will work.
```

```text
N/A
```

```text
N/A
```

Joibel:
Fixes #13305

### Motivation

In workflows version 3.5 the workflow and archived views were merged.

When a workflow is deleted from the cluster the `Workflow gone` message is shown in the UI to a user who is viewing the workflow. It is still possible to interact with the workflow if it is archived as though it was in cluster. This message is confusing to users.

### Modifications

Moved the notification to be an alternative to `archivalStatus` on "DELETED" instead of always shown.

### Verification

With archiving on, delete a workflow which is archived. See that the message is not shown.
With archiving on, allow the controller to delete a workflow which is archived. See that the message is not shown
With archiving off, delete a workflow. See that the message is shown.


agilgur5 (commented):
> Fixes #13305

It's a little unclear to me that this solves the root cause of the issue, since per https://github.com/argoproj/argo-workflows/issues/13305#issuecomment-2211055496, it shouldn't be getting a `DELETED` event at all, as it has yet to be deleted (it's only been archived, meaning it exists _both_ in etcd and the archive DB).

I'm not really sure how a `DELETED` event is occurring if it's not deleted though 🤔

PR #13310, opened 2024-07-05
feat: add `status.lastRunAt` field to WorkflowTemplates. Fixes #1915

### LINKED ISSUE #1915: Record last time when a WorkflowTemplate was used in a status field
# Summary

Record last time when a WorkflowTemplate was used in a status field

# Motivation

It would be useful to know when the last time a WorkflowTemplate was run

# Proposal

Create a status field for the template and modify a "LastRun" field whenever it is run

qingfeng777:
Fixes #1915 

### Motivation

It would be useful to know when the last time a WorkflowTemplate was run

### Modifications

- add WorkflowTemplateStatus.LastRunAt filed to workflow template filed
- enable workflowTemplate support record last run time

img_v3_02ch_489267f4-e9f9-4204-967a-ed83b98d1c2g

### Verification

run locally, submit workflow from template and check


Joibel:
The implementation of the feature doesn't do what I expected.

Currently, it is only recording uses of WorkflowTemplates when invoked through the user interface or CLI, and will not record uses in many other circumstances:
* Uses from CronWorkflows
* Uses when a workflow is injected via the kubernetes API, or other system which uses this mechanism such as argo-events
* Uses of a template when it is not used through `workflowTemplateRef`.

I am happy, if we document it as so, that this only records `workflowTemplateRef` uses of a WorkflowTemplate. It must  record all uses via `workflowTemplateRef` though. Take a look at [`fetchWorkflowSpec`](https://github.com/argoproj/argo-workflows/blob/1a81dfb305a8576bea276a8cb2d0ad30c03c2cfb/workflow/controller/operator.go#L3821).

I was also expecting this new value to be visible in the user interface, although I might be persuaded that that should be in a separate PR if you have a good reason.

Joibel (changes requested):
This should be performed by the controller to catch all uses, not by the server.

qingfeng777:
> The implementation of the feature doesn't do what I expected.
> 
> Currently, it is only recording uses of WorkflowTemplates when invoked through the user interface or CLI, and will not record uses in many other circumstances:
> 
> * Uses from CronWorkflows
> * Uses when a workflow is injected via the kubernetes API, or other system which uses this mechanism such as argo-events
> * Uses of a template when it is not used through `workflowTemplateRef`.
> 
> I am happy, if we document it as so, that this only records `workflowTemplateRef` uses of a WorkflowTemplate. It must record all uses via `workflowTemplateRef` though. Take a look at [`fetchWorkflowSpec`](https://github.com/argoproj/argo-workflows/blob/1a81dfb305a8576bea276a8cb2d0ad30c03c2cfb/workflow/controller/operator.go#L3821).
> 
> I was also expecting this new value to be visible in the user interface, although I might be persuaded that that should be in a separate PR if you have a good reason.

Thank you for being so helpful. You were absolutely right. I will make a separate PR to add the Status field and then see how to perform this task with the controller to catch all uses.

Joibel (changes requested):
By itself, adding the field but not setting it isn't useful, and is possibly misleading as someone might reasonably expect it to get set.

Could you add the code to set this value from the controller to this same PR please.

qingfeng777:
> By itself, adding the field but not setting it isn't useful, and is possibly misleading as someone might reasonably expect it to get set.
> 
> Could you add the code to set this value from the controller to this same PR please.

Thank you for your advice, I will add the code to set this value from the controller to the same PR later.

qingfeng777:
> By itself, adding the field but not setting it isn't useful, and is possibly misleading as someone might reasonably expect it to get set.
> 
> Could you add the code to set this value from the controller to this same PR please.

Hi,

In my first attempt, I updated the Getter in context.go to an Operator (supporting both get and update operations), using the clientSet as a parameter. However, the requeue and get workflow did not function as expected.

In my second attempt, I used NewContextFromClientSet instead of NewContext, to hold both the informer getter and clientSet. Unfortunately, I was unable to retrieve clientSet in some callers of the ValidateWorkflow function.

Ultimately, I implemented the feature like this. I am unsure if there is a better approach. Could you please review it again?

Thank you.

workflow/templateresolution/context.go
```diff
@@ -135,6 +162,12 @@ func (ctx *Context) GetTemplateFromRef(tmplRef *wfv1.TemplateRef) (*wfv1.Templat
 		}
 		return nil, err
 	}
+	if !tmplRef.ClusterScope && wftmpl != nil {
```
Joibel, line 165: Here we don't update ClusterWorkflowTemplates (`!tmplRef.ClusterScope`). They should be updated too. It means cwftmplClient which you've added doesn't do any work ever.

workflow/templateresolution/context.go
```diff
@@ -166,6 +199,17 @@ func (ctx *Context) GetTemplateScope() string {
 	return string(ctx.tmplBase.GetResourceScope()) + "/" + ctx.tmplBase.GetName()
 }
 
+// UpdateTemplateStatus update the WorkflowTemplate.Status of a given WorkflowTemplate.
+func (ctx *Context) UpdateTemplateStatus(wftmple *wfv1.WorkflowTemplate) (*wfv1.WorkflowTemplate, error) {
```
Joibel, line 203: UpdateTemplateStatus should not be capitalised, as it is not necessary to [export](https://go.dev/tour/basics/3) it. Rename to `updateTemplateStatus`.

workflow/templateresolution/context.go
```diff
@@ -166,6 +199,17 @@ func (ctx *Context) GetTemplateScope() string {
...
+	if ctx.wftmplClient == nil {
+		ctx.log.Warnln("while try update template status, wftmplClient not set")
+		return wftmple, nil
+	}
+
+	wftmple.Status.LastRunAt = metav1.Now()
+	return ctx.wftmplClient.Update(context.TODO(), wftmple, metav1.UpdateOptions{})
```
Joibel, line 210: This is going to be very hard on the kubernetes API, updating for every time we look at the template, and doing it during execution. I think this needs to go into a deduplicating queue, which is doing rate limited writes to the actual API.
One run of the [dag diamond](https://github.com/argoproj/argo-workflows/blob/main/examples/workflow-template/dag.yaml) called this 33 times.

workflow/templateresolution/context.go
```diff
@@ -166,6 +199,17 @@ func (ctx *Context) GetTemplateScope() string {
 	return string(ctx.tmplBase.GetResourceScope()) + "/" + ctx.tmplBase.GetName()
 }
 
+// UpdateTemplateStatus update the WorkflowTemplate.Status of a given WorkflowTemplate.
+func (ctx *Context) UpdateTemplateStatus(wftmple *wfv1.WorkflowTemplate) (*wfv1.WorkflowTemplate, error) {
+	if ctx.wftmplClient == nil {
+		ctx.log.Warnln("while try update template status, wftmplClient not set")
```
Joibel, line 205: This error message is coming out when running the [dag diamond](https://github.com/argoproj/argo-workflows/blob/main/examples/workflow-template/dag.yaml), which probably shouldn't happen. Could you investigate.

Joibel (changes requested):
This is a hard problem to solve correctly. The main objection here is how hard this is on the kubernetes API, and how it will slow down execution.

workflow/templateresolution/context.go
qingfeng777: How about setting a ticker to run every ten minutes? The ticker can maintain a map that stores the updated workflows and their corresponding update times, ensuring updates are performed once every ten minutes.

workflow/templateresolution/context.go
qingfeng777: How about adding the feature of updating ClusterWorkflowTemplates' status in a separate PR?
It will be easier for me to manage

workflow/templateresolution/context.go
Joibel: I believe you run inside a devcontainer. So I assume you're testing with `make start`.

Assuming you are in the argo-workflows checkout directory - the logs will be in `logs/controller.log`. You can run `kubectl create -f examples/dag_diamond.yaml` and I got this error.

workflow/templateresolution/context.go
Joibel: I'd want the two merged at the same time, so they'd need to be stacked PRs which is hard to manage too. Get it right for non-cluster and the clusterwftemplates should be pretty simple to add at the end.

qingfeng777: OK, I got it.

workflow/templateresolution/context.go
qingfeng777: Thank you, I will try it later.

workflow/templateresolution/context.go
qingfeng777: Thank you, I ran in my machine directly, and got the logs in `logs/controller.log` too.

qingfeng777: But, I can't reproduce it. Those are what I did.
`
git checkout <target-branch>
make clean
make start UI=true

 kubectl create -f examples/dag-diamond.yaml
    + kubectl create -f examples/dag-diamond.yaml
    workflow.argoproj.io/dag-diamond-67fvs created
    
➜  logs git:(support-record-preset) cat controller.log|grep while
➜  logs git:(support-record-preset)
`
And I checked other files in the `logs`, not found  the error log too.
Did I miss something? or could you recheck it?

qingfeng777: I tried to reproduce by running dag workflow with a tmpl-ref, with the following commands:
```
kubectl create -f examples/workflow-template/templates.yaml
kubectl create -f examples/workflow-template/dag.yaml
```
but I still couldn't find the error log, so I mark this conversation resolved. If I missed something, feel free to reopen it.
Thank you a lot.

Joibel:
I am away for 2 weeks, I will take a look when I can.

qingfeng777:
> I am away for 2 weeks, I will take a look when I can.

Happy holidays!

qingfeng777:
Hi，@Joibel, could you take a look at this PR? thank you.

terrytangyuan (commented):
I wonder instead of just showing last run, we can keep a history of runs.

PR #13319, opened 2024-07-07
chore(manifests): consistently put k8s "header" at the top

agilgur5:
Seen this before, but was more striking while searching for YAML files without `kind`s (for test purposes) in #13317 -- all other files were very quickly discernible 

### Motivation

- several manifests, particularly Secrets and ConfigMaps, had the "header" with `kind` and `metadata` at the bottom
  - but still had `apiVersion` at the top
  - this is pretty confusing to look at and doesn't match the rest of the manifests, so make it consistent

### Modifications

- move `kind` and `metadata` to the top of the manifests, under `apiVersion`, as is common k8s convention

### Verification

n/a, pure style change


Joibel (approved):
LGTM

PR #13320, opened 2024-07-07
docs: only display `main` status for Snyk badge

agilgur5:
### Motivation

- the link went to `main`, but the query for the badge did not
- this should only show `main`, and then for release branches show the release branch
  - this will need manual backporting as such (np, I can do that)

### Modifications

- add a branch query to the SVG

### Verification

- Copied from the [Actions page](https://github.com/argoproj/argo-workflows/actions/workflows/snyk.yml?query=branch%3Amain)'s "Create status badge" button (under the 3 dots)
- Can see the [rendered manifest](https://github.com/argoproj/argo-workflows/blob/5a81c4c1eeba1da355d5ccea392ec5d4441b4a0c/README.md)


Joibel (approved):
Ah, this makes sense. Thanks.

agilgur5:
> this will need manual backporting as such (np, I can do that)

Backported with modifications to `release-3.5` as 9f55687e38fc89176ad1ce99d43b70f8f47d584b.

I actually couldn't cherry-pick it into `release-3.4` as it had a merge conflict because the docs are older. Might eventually be able to get the docs up-to-date if there aren't too many conflicts (I made some progress on that, but it's pretty far behind)

Issue #13164, opened 2024-06-11
outbound node of itself


tooptoop4:
### Pre-requisites

- [X] I have double-checked my configuration
- [ ] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

looking at controller logs i see:

```
time="2024-06-11T03:46:00.232Z" level=info msg="Outbound nodes of mywf-375343128 is [mywf-375343128]" namespace=redact workflow=mywf
time="2024-06-11T03:46:00.232Z" level=info msg="Outbound nodes of mywf-2338343363 is [mywf-375343128]" namespace=redact workflow=mywf
```

so it seems to consider itself node as an outbound node

https://github.com/argoproj/argo-workflows/blob/v3.4.17/workflow/controller/steps.go#L212-L223

### Version

3.4.11

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
n/a
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

PR #13325, opened 2024-07-09
chore(gh): add link to SO MVCE in bug templates

agilgur5:
### Motivation

- further highlighting what minimal means, since we do get some reports that are _far_ from minimal
  - and also some reports without a repro what-so-ever 😕
  - Recent examples: #13314, #13238, #13164, etc etc

- will this help?
  - a small bit maybe. higher quality reports do this already as do higher quality SO questions, but lower quality reports & questions still miss this and so this page still gets linked often
    - I figure we should at least _try_ to educate lower quality reporters on how to do this
      - hence do this automatically in the issue template, over manually doing it in responses (which may still be necessary unfortunately, but hey then it's a quick copy+paste at least)
  - I also added a description mentioning use in prioritization, to hopefully incentivize the behavior as well (and a non-reproducible issue is effectively not prioritized what-so-ever, so this is accurate)

### Modifications

- Add a link to SO's "minimal, reproducible" page (fka MVCE): https://stackoverflow.com/help/minimal-reproducible-example 
- Add a description explaining that this is "essential" and necessary for "prioritization" to attempt to incentivize the behavior

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -30,7 +30,8 @@ body:
   - type: textarea
     id: failing-workflow
     attributes:
-      label: Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.
+      label: Paste a [minimal](https://stackoverflow.com/help/minimal-reproducible-example) workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.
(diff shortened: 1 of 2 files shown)


Joibel (approved):
It seems a shame you can't linkify in the label where it might get more attention, but at least we have an easy link to point people at now. I like the page you've linked to a lot.

agilgur5:
Yea I link to it manually with some frequency. It has gotten a lot of iteration and effort on it since on SO it is very common to link to as well. The page might have existed for over a decade (idr)

Issue #13103, opened 2024-05-28
DOM text reinterpreted as HTML in `workflow-logs-viewer.tsx`


jessicacunha:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

DOM text is reinterpreted as HTML without escaping meta-characters.

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

jessicacunha:
I`m working in this issue

agilgur5:
I assume this is related to the other two CodeQL issues filed within the same few hours, #13105 and #13104. 

You didn't cite your sources here though nor list the specific line number. Please cite your sources and directly quote your sources. Also I'm not sure who organized these issues, but it would have made sense to notify organizations like Argo before filing issues, as well as list that they are part of an organized effort in the issues.
At least one of these has also been a very obvious false positive, which is a poor use of limited maintainer time.

I also don't see any line number in [`workflow-logs-viewer.tsx`](https://github.com/argoproj/argo-workflows/blob/f492ea3609297617a533f1905777bffa57731102/ui/src/app/workflows/components/workflow-logs-viewer/workflow-logs-viewer.tsx) that would correspond to "DOM text reinterpreted as HTML" either, and with regard to "escaping meta-characters", React itself does this to prevent XSS, unless one uses [`dangerouslySetInnerHTML`](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html), which Argo does not.

PR #13326, opened 2024-07-09
fix: #13104 Addressing incorrect-integer-conversion warning flagged by CodeQL

### LINKED ISSUE #13104: Fix incorrect conversion between integer types in `parametrizable.go`
# Summary
CodeQL flagged util/intstr/parametrizable.go as having an incorrect conversion of an integer with architecture-dependent bit size from strconv.Atoi to a lower bit size type int32 without an upper bound check on line 32.

StephenDunne97:
### Related issue

Fixes #13104

### Motivation

This issue was identified by CodeQL as part of an opensource minijam. util/intstr/parametrizable.go  was flagged as having an incorrect conversion of an integer with architecture-dependent bit size from strconv.Atoi to a lower bit size type int32 without an upper bound check on line 32.

### Modifications

Added code that checks if the value of a given integer v is within the range of a 32-bit signed integer.

### Verification
Code was tested locally.

### DIFF
--- a/util/intstr/parametrizable.go
+++ b/util/intstr/parametrizable.go
@@ -29,6 +29,11 @@ func Int32(is *intstr.IntOrString) (*int32, error) {
 	if v == nil || err != nil {
 		return nil, err
 	}
+	
+	if *v > math.MaxInt32 || *v < math.MinInt32 {
+		return nil, fmt.Errorf("integer out of range for int32: %d", *v)
+	}
+
 	i := int32(*v)
 	return &i, err
 }


agilgur5:
> as part of an opensource minijam

It looks like all of these issues (#13103, #13104, #13105) and PRs (this one and #13277) originated from Fidelity? Was there some internal hackathon?

Please see https://github.com/argoproj/argo-workflows/issues/13103#issuecomment-2136262574; these issues and PRs really did not follow any best practices (which is even more surprising given they were corporate), did not follow templates or pass tests often, had no responses from authors, took over a month and a half before a PR was made, and some were entirely (and clearly) invalid false positives.

The organizers and authors should really follow proper open source conventions and best practices _at minimum_. Citing sources, directly quoting them, avoiding false positives, etc should also really be done. 
Without those, these are actively counter-productive as maintainers spend more time than the authors to read and validate, and sometimes even mark as invalid false positive etc (in that case, _entirely_ unproductive).
OSS is a community effort, often run by many passionate, unpaid volunteers (like myself), so spending maintainer time wisely and properly communicating with maintainers is important, less problematic, much more effective and efficient, and honestly just more & properly respectful of maintainer time.

PR #13332, opened 2024-07-10
fix: allow nodes without workflowTaskResults

Joibel:
Fixes #TODO


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1976,7 +1976,7 @@ func (ws *WorkflowStatus) IsTaskResultIncomplete(name string) bool {
 	if found {
 		return !value
 	}
-	return true
+	return false
```
agilgur5, line 1979: could you add a comment here to clarify this behavior? from an out-of-context reading, this would be unexpected / unintuitive
```suggestion
	return false // workflows from older versions do not have this status, so assume completed if this is missing
```

agilgur5 (approved):
LGTM, thanks for analyzing and fixing this!

PR #13333, opened 2024-07-11
chore(gh): fix grammar error in regression template

agilgur5:
### Motivation

Just a small grammar and consistency issue I noticed while writing up #13324  etc

### Modifications

- "what you expected" -> "what did you expect"

- also make the bug templates use two full sentences instead

### Verification

Pure string change. See also [rendered commit](https://github.com/argoproj/argo-workflows/blob/2d1443ed7c36c0ab17e7f135e67498c70fe5643c/.github/ISSUE_TEMPLATE/regression.yaml)


Joibel (approved):
LGTM

PR #13340, opened 2024-07-12
refactor: reorganize some `Makefile` variables

agilgur5:
Just some reorganization I've been brewing for a while

### Motivation

- some of these were mixed in random places
  - instead put them together with related chunks

### Modifications

- categories:
  - `VERSION` with other build metadata
  - test options together
  - docker image publishing options together
  - pure env vars all together

- other misc changes:
  - remove unused `DEV_IMAGE`
  - remove some unnecessary new lines between related variables
  - in-line some comments

### Verification

pure style change, no semantic changes

### Notes for Reviewers

This is ofc subjective and there's potentially many other things to reorganize too (like splitting up the `Makefile` tasks into more files), but I think I got something somewhat small here that is definitively better than the previous version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,45 +9,35 @@ USE_NIX := false
 MAKEFLAGS += --no-builtin-rules
 .SUFFIXES:
 
+# -- build metadata
 BUILD_DATE            := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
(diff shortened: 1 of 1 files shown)


Joibel (approved):
Thank you, much tidier and easier to find the things that are related to each other.

PR #13345, opened 2024-07-15
refactor: use templateScope var instead of repeatedly retrieving it from function

qingfeng777:
### Motivation

For better code readability

### Modifications

Use templateScope instead of repeatedly retrieving it from the function.

### Verification

run locally


agilgur5 (approved):
LGTM to me, I don't think these will ever be different. cc @Joibel who wrote the recursion detection code

Joibel (approved):
LGTM. The recursion code was mostly a resurrected stale PR, so not really mine either.

PR #13347, opened 2024-07-16
fix(server): don't return `undefined` SA NS

### LINKED ISSUE #12658: UI defaults to `undefined` namespace
### Pre-requisites

I've installed latest Argo Workflows through the Helm chart and created a role, rolebinding, serviceaccount and secret (in the argo namespace) for accessing the UI with the token. The UI always defaults to an Undefined namespace and I always have to switch it to argo. I thought this should have been fixed with: https://github.com/argoproj/argo-workflows/pull/9008  or do I need to configure the default namespace somewhere? I couldn't find any option.

Screenshot from 2024-02-11 10-47-43

### Version

3.5.4

```YAML
Not related to workflows
```

```text
Not related to workflows
```

```text
Not related to workflows
```

agilgur5:
Fixes #12658, for which the root cause also resulted in confusion in #13316
Follow-up to #9008

### Motivation

- previously, in client auth mode, no SA name or ns was listed, causing `userNamespace` to be `undefined`
  - and that would cause `currentNamespace` to be set as the string literal `"undefined"`

### Modifications

- derive the SA name and namespace from the `subject` at the end of `claims.go`
  - invert the conditional as well for less nesting and complexity, similar to other refactors of mine

- add some defensive code to the UI as well, so that in case of `undefined` (which should be substantially rarer now), don't set the namespace in the URL to literally "undefined" <details>

  - invert `fixLocalStorageString` conditional because it very confusingly would _implicitly_ return `undefined` before
    - and catch the case of this returning `undefined` in `currentNamespace`
      - use the `??` nullish coaelscing operator to handle null or undefined but not empty string etc (unlike truthy)
        - (this operator may not have existed at the time the original code was written)
      - also remove the incorrect comment here

</details>

### Verification

- Added unit tests
- <details><summary>Manually tested and got this correct result in the UI:</summary>

    Screenshot 2024-07-13 at 7 09 05 PM

    - As well as this correct `localStorage` item: Screenshot 2024-07-13 at 7 09 39 PM
    
    **EDIT**: woops I accidentally flipped these, fixed per [below comment](https://github.com/argoproj/argo-workflows/pull/13347#issuecomment-2229726424)
    
</details>

### Notes to Reviewers

- ["Hide whitespace" view](https://github.com/argoproj/argo-workflows/pull/13347/files?diff=unified&w=1) shows the `claims.go` diff as much smaller.

- This works for SA-based and token-based auth for `client` auth mode as well as `server` auth mode. During local dev though, `server` auth mode runs with local kubeconfig, which in the default k3s set-up, actually uses a client cert for auth. This has no `bearerToken` or similar to pull out. I could not figure out a way to pull out claims from that via `client-go` and there may not be a way of doing so.
  - This case is very rare for users though, so this PR should still fix 99% of the issue

### DIFF
--- a/server/auth/serviceaccount/claims.go
+++ b/server/auth/serviceaccount/claims.go
@@ -48,15 +48,15 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
 		return nil, fmt.Errorf("failed to unmarshal bearer token's JWT payload: %w", err)
 	}
 
-	// attempt to derive ServiceAccount and ServiceAccountNamespace from Subject
+	// attempt to derive SA name and namespace from Subject
 	// "system:serviceaccount:argo:jenkins" -> "argo", "jenkins"
 	// note that the SA name can have a colon in it, although the rest cannot
 	parts = strings.SplitN(claims.Subject, ":", 4)
 	if len(parts) < 4 {
 		return claims, nil
 	}
-	claims.ServiceAccountName = parts[2]
-	claims.ServiceAccountNamespace = parts[3]
+	claims.ServiceAccountNamespace = parts[2]
+	claims.ServiceAccountName = parts[3]
 
 	return claims, nil
 }
--- a/server/auth/serviceaccount/claims.go
+++ b/server/auth/serviceaccount/claims.go
@@ -17,32 +17,46 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
 	username := restConfig.Username
 	if username != "" {
 		return &types.Claims{Claims: jwt.Claims{Subject: username}}, nil
-	} else if restConfig.BearerToken != "" || restConfig.BearerTokenFile != "" {
-		bearerToken := restConfig.BearerToken
-		if bearerToken == "" {
-			// should only ever be used for service accounts
-			data, err := os.ReadFile(restConfig.BearerTokenFile)
-			if err != nil {
-				return nil, fmt.Errorf("failed to read bearer token file: %w", err)
-			}
-			bearerToken = string(data)
-		}
-		parts := strings.SplitN(bearerToken, ".", 3)
-		if len(parts) != 3 {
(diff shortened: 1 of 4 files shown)


agilgur5:
> * Manually tested and got this correct result in the UI:

Somehow I missed that these were flipped 😅 ; fixed that in the [most recent commit](https://github.com/argoproj/argo-workflows/pull/13347/commits/bfc7e1d1e6b1d2122bc3f3286638a63a598d5fca). Wrote the unit test correctly but the actual code flipped 😅

server/auth/serviceaccount/claims.go
```diff
@@ -17,32 +17,46 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
...
+
+	// attempt to derive SA name and namespace from Subject
+	// "system:serviceaccount:argo:jenkins" -> "argo", "jenkins"
+	// note that the SA name can have a colon in it, although the rest cannot
+	parts = strings.SplitN(claims.Subject, ":", 4)
+	if len(parts) < 4 {
 		return claims, nil
-	} else {
-		return nil, nil
 	}
```
juliev0, line 57: So, is `claims.Subject` always set to that? I'm not really familiar, so f you have any link to share on that, it would be helpful, thanks.

agilgur5: For a k8s SA token, yes. In the issue I experimented a bit with different types of k8s auth. An SA token from a `Secret` or from `kubectl create token` always has this type of subject. So primarily that was done via experimentation on actual tokens.

In this section of the k8s docs there's an example of a decoded token (this one is equivalent to one generated by `kubectl create token`): https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection

IIRC, all other forms of auth (e.g. kubeconfig) are not token auth, so they wouldn't reach here.

juliev0 (approved):

PR #13111, opened 2024-05-29
fix: silence devcontainer `Permission denied` error. Fixes #13109

Adrien-D:
Fixes #13109

### Motivation

`chown` command fails for some reason on `.git` folder on my mac M1

### Modifications

I silenced the error by adding `|| true` on the failing command on `.devcontainer/pre-build.sh`

-> The error is probably due to a mac-specific thing. It'll be to do with how the OS has mounted that volume to the dev docker container
-> There is no gitignore-style file for devcontainers, to ignore `.git` directory

### Verification

The impact of the change are quite small, but the process reached the end

Screenshot 2024-05-29 at 15 16 01

PR #13348, opened 2024-07-16
fix(devcontainer): remove `-R` in `chown`

agilgur5:
Follow-up to #13111, might make the `|| true` no longer necessary as well

### Motivation

- it seems like the `-R` isn't necessary??
  - that would take very long for me to run and without it I seemed to get a fresh build working fine

### Modifications

- remove the recursive `-R` from the `chown` command in `pre-build.sh`

### Verification

Did a local rebuild
~Will also confirm with a fresh Codespace of this branch (it's building right now)~
- **EDIT**: Made a fresh Codespace, which built and ran `make start UI=true` successfully (and the UI, Controller, Server et al were running correctly as well)

### DIFF
--- a/.devcontainer/pre-build.sh
+++ b/.devcontainer/pre-build.sh
@@ -20,7 +20,7 @@ sudo apt update
 sudo apt install -y protobuf-compiler
 
 # Make sure go path is owned by vscode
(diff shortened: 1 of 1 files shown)


Joibel (approved):
I can't work out if something has changed to mean this is fine now, but it's also fine for me.

PR #13350, opened 2024-07-16
fix(build): slightly optimize UI tasks in `kit` `tasks.yaml`

agilgur5:
### Motivation

- while the UI needs the server to connect to the API, it can start and run in parallel
  - it also has some small functionality independently of the Server; it'll just error on API requests

- in particular, when I was working on the Server and UI simultaneously (e.g. #13347), any Server change would cause the whole UI task to restart, even if I did not change the UI
    - this was pretty slow and inefficient as such, so remove the dep

### Modifications

- remove `ui` task's dependency on `server` task
- remove unnecessary `mutex: download`
  - the two dep installations can run concurrently, they're in totally separate directories and use different tools (go vs yarn)
  - something I did in https://github.com/argoproj/argo-workflows/pull/11928#discussion_r1343064901 as well

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -18,7 +18,6 @@ spec:
   tasks:
     - name: go-deps
       command: go mod download
-      mutex: downloads
     - name: install
       command: sh -c "make install PROFILE=$PROFILE"
       env:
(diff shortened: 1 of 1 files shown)


Joibel (approved):
This works for me. I'd never really understood why the mutex was needed at all either. Thanks.

PR #13352, opened 2024-07-16
docs: split out <3.x.x `CHANGELOG.md` for faster loading

agilgur5:
Fixes "Future Work" section of #13236

### Motivation

- the `CHANGELOG.md` is quite large and so can take some time to load as it's a giant file (>10k LoC)

### Modifications

- split the <3.x.x versions of the changelog into a separate `CHANGELOG-2-x-x.md` file
  - this is about a 50% decrease in size, 5k+ LoC split off
  - it is exceedingly rare that folks are still on 2.x.x
  - link the two changelogs together as well at the bottom and top, respectively
  - modify changelog script to skip <3.x.x
    - we could also modify it to make two files, but I don't think we need to at this time

### Verification

Ran `./hack/changelog.sh > CHANGELOG.md` and got no diff


Joibel (approved):
LGTM

Issue #11688, opened 2023-08-28
UI for semaphores/mutex state


tooptoop4:
for example https://github.com/linz/topo-workflows/blob/f52b71b39767fe14e3da5eefd352e62028c23143/config/semaphores.yml#L1C1-L10C104 defines many different semaphores

would be good to have a UI page that lists how many of each semaphore is currently being consumed by workflows ie 3 out of 8.
likewise for mutex 0 out of 1 or 1 out of 1

airflow has a similar page (see image in this link) --> https://stackoverflow.com/questions/53193282/apache-airflow-pools-used-slots-available-slots

PR #13358, opened 2024-07-17
feat: multiple mutexes and semaphores

### LINKED ISSUE #5022: Support for multi lock or semaphores
# Summary

Support for multi lock or semaphores

# Use Cases

We have a case where we have multiple data pipelines pulling from  different data sources and a pipeline compiling data models at the end.
Some of these pipelines can run different strategies (full at 1 am, incremental pulls every 10 mins or they can be triggered on-demand). Let's call them `A` and `B`.
Let's call the other pipeline `C`.

To avoid full and incremental to run a the same time, we could have a semaphore `A` with value 1 and avoid pipelines of type A to run simultaneously. The same can be applied for `B`.
However, to avoid running `C` while `A` or `B` are running, we need support for two concurrent semaphores: `A` and `B` with value 1.

The above could also be achieved with support for multiple locks.

---

### LINKED ISSUE #11859: mutex ignored if using semaphore at same top level
### Pre-requisites

logs show nothing about mutex and i verified 2 workflows were able to run at same time

logs show semaphore is working.

then i tried removing semaphore and mutex worked. so seems they don't work together?

https://github.com/argoproj/argo-workflows/blob/master/workflow/sync/sync_manager.go#L280-L283 seems like it doesn't loop through both types of synchronization cc @sarabala1979 

### Version

3.4.11

```YAML
kind: Workflow
metadata:
  generateName: redact-
spec:
  serviceAccountName: argo-workflows-server
  onExit: exit-handler
  entrypoint: flow
  synchronization:
    semaphore:
      configMapKeyRef:
        name: argowf-semaphores
        key: redact
    mutex:
      name: "{{= sprig.coalesce('something','catch') }}"
...can't share the rest
```
```


```text
```

```text
```

Joibel:
Fixes #5022
Fixes #11859

This is mostly backwards compatible. It is minorly breaking only in as much as semaphore + mutex at the same time will now actually take both, whereas before it ignored the mutex.

## Motivation

As a workflows users I would like to be able to use more than one sync option in a workflow or template.

## Modifications

Added `mutexes` and `semaphores` to the `synchronization` block in workflow types as lists. Legacy `mutex` and `semaphore` kept in and continue to work (will be added to the mutexes/semaphores list) but marked for deprecation.

Logic changed to acquire a full list of all synchronization items (`syncItem`) when `TryAcquire` and `Release` from sync manager.

`TryAcquire` from `sync_manager` calls a new `checkAcquire` method, whilst holding the sync_manager internal golang `sync.Mutex` to check if they are all acquirable before acquiring them. This honors priority.

`TryAcquire` also returns the failed lock name in the case of failure to acquire to reduce the number of exported `package sync` methods and complexity outside of `sync`.

The interface to the workflow status block has not changed.

Reduced the exported types and methods from the sync module as far as possible (lower case instead of upper case).

Removed the internal golang sync mutex from `PrioritySemaphore` as all methods are called from `sync_manager.go` when under it's own lock. The alternative was unnecessarily complicating the calls inside `semaphore.go`. This is faster, but less clean in case someone wants to reuse the PrioritySemaphore type.

Renamed the manager in `sync_manager.go` where used: `cm`->`sm` and `concurrencyManager`->`syncManager`.

~Synchronization documentation updated~ -- **EDIT**: split to #13393 

## Verification

New and updated unit tests.
New and updated e2e smoke tests.

## Stacked PR

This is stacked on top of #13393 now.


terrytangyuan:
@jessesuen Would you like to review this as well since you might be relying on the existing behavior of the synchronization feature?

docs/synchronization.md
```diff
@@ -142,13 +144,64 @@ spec:
 
 Examples:
 
-1. [Workflow level semaphore](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-wf-level.yaml)
+1. [Workflow level semaphore](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-wf-level.yamlxb)
```
tooptoop4, line 147: is yamlxb typo?

docs/synchronization.md
```diff
@@ -142,13 +144,64 @@ spec:
...
+The workflow will block until all of these mutexes and semaphores are available.
+
+### Priority
+
+Workflows can have a `priority` set in their specification.
+Workflows with a higher priority value will be queued to take a semaphore or mutex before a lower priority workflow, even if they have been waiting for less time.
+
+!!! Warning
+    A high priority workflow waiting on multiple mutexes or semaphore will make all other workflows which want to acquire those mutexes wait for it to acquire and release all the mutexes or semaphores it is waiting on.
+    This applies even if the lower priority workflows only wish to acquire a subset of those mutexes or semaphores.
```
tooptoop4, line 179: low priority wf won't start running even if the 1 mutex it needs is free? whereas the high priority wf can't even run because it still needs another few mutexes?

docs/synchronization.md
Joibel: Yes, thanks, I'll fix that.

docs/synchronization.md
Joibel: Correct.

You've made me realise I should document the time ordered queue as well.

tooptoop4: this seems too restrictive, if the high priority one is unable to run due to multiple locks might as well let the low priority one that just needs 1 lock run

Joibel:
> @jessesuen Would you like to review this as well since you might be relying on the existing behavior of the synchronization feature?

It shouldn't be a breaking change for existing users.

docs/synchronization.md
tooptoop4: i think priority should only apply if all locks are satisfied at the same time, or maybe this could be some kind of new config flag

Joibel: I'm not going to implement that here. It would be a new feature to have a different scheduling system. I'm not even convinced trying to shoehorn complex schedulers into argo-workflows is a good idea. It needs some proper design.

I've updated the documentation to explain more how the queue works.

None of the queuing is new, I've just documented how it worked and how it works in the presence of multiple locks.

agilgur5:
@Joibel could you please re-format your PR descriptions? I think they're line-breaking at 80 chars or something, which depending on the width of the GH description, can sometimes cause a double line-break or otherwise look weird. I manually re-formatted a few of your other PRs, but this one is quite long

I can review this as I had worked briefly on these, but I'm not promising any timeline as I already have too many things on my plate (including all the metrics PRs, and docs/UI/Server and some others are often falling on my plate too 😕  most features have docs portions too...)

agilgur5:
> This is mostly backwards compatible. It is minorly breaking only in as much as semaphore + mutex at the same time will now actually take both, whereas before it ignored the mutex.

Are we considering this breaking? The silent ignore is a pure bug to me, where the user _intent_ was to do both, but you'd find out after that didn't happen.

I was actually thinking that would be its own fix PR that would be backported to older versions, whereas the feature to have multiple ofc would not be backported

agilgur5 (commented):
> This is mostly backwards compatible. It is minorly breaking only in as much as semaphore + mutex at the same time will now actually take both, whereas before it ignored the mutex.

Are we considering this breaking? The silent ignore is a pure bug to me, where the user _intent_ was to do both, but you'd find out after that didn't happen.

I was actually thinking that would be its own fix PR that would be backported to older versions, whereas the feature to have multiple ofc would not be backported

docs/synchronization.md
```diff
@@ -147,8 +150,64 @@ Examples:
 1. [Step level semaphore](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-tmpl-level.yaml)
 1. [Step level mutex](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level.yaml)
 
-### Other Parallelism support
+### Multiple synchronization
```
agilgur5, line 153: This needs a version specifier

examples/synchronization-mutex-tmpl-level.yaml
```diff
@@ -28,18 +28,18 @@ spec:
 
   - name: acquire-lock
     synchronization:
-      mutex:
-        name: welcome
+      mutexes:
+        - name: welcome
```
agilgur5, line 32: It might be good to leave the other one in commented out with version specifiers for both. Since people will look at the examples and not know that they have future syntax etc

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1628,10 +1628,14 @@ type TemplateRef struct {
...
 type Synchronization struct {
-	// Semaphore holds the Semaphore configuration
+	// Semaphore holds the Semaphore configuration - deprecated, use semaphores instead
 	Semaphore *SemaphoreRef `json:"semaphore,omitempty" protobuf:"bytes,1,opt,name=semaphore"`
-	// Mutex holds the Mutex lock details
+	// Mutex holds the Mutex lock details - deprecated, use mutexes instead
 	Mutex *Mutex `json:"mutex,omitempty" protobuf:"bytes,2,opt,name=mutex"`
+	// Semaphores holds the list of Semaphores configuration
+	Semaphores []*SemaphoreRef `json:"semaphores,omitempty" protobuf:"bytes,3,opt,name=semaphores"`
+	// Mutexes holds the list of Mutex lock details
```
agilgur5, line 1637: These also need version specifiers since they appear in the Fields reference, API docs, etc
```suggestion
	// v3.6 and after: Semaphores holds the list of Semaphores configuration
	Semaphores []*SemaphoreRef `json:"semaphores,omitempty" protobuf:"bytes,3,opt,name=semaphores"`
	// v3.6 and after: Mutexes holds the list of Mutex lock details
```

docs/parallelism.md
agilgur5, line 1: two comments:
1. This should be split out into its own PR as this will be backported. It's been on my todo list for a while, so thanks for adding it.

1. Do we want this to be a separate file? They are closely related and the back-links are fine, but this adds a whole new page to the docs. We could have them in the same file if we rename the page to something like "Concurrency Control" or something, thereby being more generic than the `synchronization` field

agilgur5 (dismissed):
Version specifiers are needed everywhere. Also the parallelism doc should be split into its own PR as it is very backport worthy.

Joibel:
> > This is mostly backwards compatible. It is minorly breaking only in as much as semaphore + mutex at the same time will now actually take both, whereas before it ignored the mutex.
> 
> Are we considering this breaking? The silent ignore is a pure bug to me, where the user _intent_ was to do both, but you'd find out after that didn't happen.
> 
> I was actually thinking that would be its own fix PR that would be backported to older versions, whereas the feature to have multiple ofc would not be backported

These are intertwined features in my implementation. One semaphore and one mutex is just a special case of multiple synchronization items. The bulk of this PR is necessary to make that case work, and I'd consider it too invasive to backport.

I consider the *Alternatively* wording sufficient to say it was not supposed to work before, hence should not be backported. https://github.com/argoproj/argo-workflows/blob/c16c5e4adfbc1e321f9e9293d8364904c6db4bf9/docs/synchronization.md?plain=1#L11

I'm not fussed whether we call this breaking though.

docs/parallelism.md
Joibel: Separated, along with backportable synchronization.md updates to #13393.
This is now stacked on that.

I do want it to be a separate file - see #13393

agilgur5:
> The bulk of this PR is necessary to make that case work, and I'd consider it too invasive to backport.

Gotcha. Yea when I examined the issue it could not be done easily either (whereas multi-semaphore was more straightforward). 

In that case, I am wondering whether we should have a separate "fix" to backport that errors out when both are provided. Since this feature will only land in 3.6 and the silent ignore in earlier versions is very confusing/surprising as it stands

Joibel:
> In that case, I am wondering whether we should have a separate "fix" to backport that errors out when both are provided. Since this feature will only land in 3.6 and the silent ignore in earlier versions is very confusing/surprising as it stands

Yes, this would be nice if it got done.

juliev0:
@sarabala1979 would you be able to review this one? (I think it's up your alley)

sarabala1979:
I will take a look

workflow/sync/lock_name.go
```diff
@@ -8,91 +8,99 @@ import (
...
+func getMutexLockName(mtx *v1alpha1.Mutex, wfNamespace string) *lockName {
+	namespace := mtx.Namespace
+	if namespace == "" {
+		namespace = wfNamespace
+	}
+	return newLockName(namespace, mtx.Name, "", lockKindMutex)
+}
+
+func getLockName(item *syncItem, wfNamespace string) (*lockName, error) {
+	switch {
```
sarabala1979, line 54: SyncItem should have a semaphore or mutex. It needs to be validated in validate.go

sarabala1979:
@Joibel Thanks for contributing. It is a good feature and a more complex feature.  Can we set some time to change the walkthrough? Please let me know your availability

workflow/sync/sync_manager.go
```diff
@@ -113,102 +111,146 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
...
+	for i, syncItem := range syncItems {
+		syncLockName, err := getLockName(syncItem, wf.Namespace)
 		if err != nil {
-			return false, false, "", err
+			return false, false, "", failedLockName, fmt.Errorf("requested configuration is invalid: %w", err)
 		}
-		cm.syncLockMap[lockKey] = lock
+		lockKeys[i] = syncLockName.encodeName()
 	}
+	for i, lockKey := range lockKeys {
```
sarabala1979, line 140: I think Code will maintain the order of locks.
How will the code handle the below deadlock scenario?
WF1 configured 3 mutex locks (l1,l2,l3)
WF2 configured 2 locks (l3,l2)
WF2 acquired l3 and waiting for L2
WF1 acquired L1, L2 and waiting for L3

workflow/sync/lock_name.go
Joibel: syncItem is always accessed from syncManager, and always under a scenario where syncManager holds `Manager.lock`. Rather than add more overhead here, I'm relying on that for mutual exclusion.

syncItem is an internal type for sync_manger to manage all synchronization types (mutex and semaphore) as a single type. It's never user exposed and so I can't see how it needs validating.

workflow/sync/sync_manager.go
Joibel: The old code could deadlock, in that you could nest templates which performed a sequence that caused two workflows to hold and request locks that they could never escape from.
For example
* WF1 template `tp1` acquires mutex `l1`, and later `tp2` which wants mutex `l2`
* WF2 template `tp3` acquires mutex `l2` and then calls `tp1`  then it will then be deadlocked.

I don't think we should attempt to mitigate deadlocks - but providing a good UI around how locks are being held in a more accessible way would be something I'd like to have.

In your scenario I am assuming the locks are being acquired at the same time. There is not a problem:
* When WF2 runs it will only be able to either acquire **both** locks or neither.
* [`checkAcquire`](https://github.com/argoproj/argo-workflows/blob/22b64f98998754216e56bbf768fc66f92ecb0f5e/workflow/sync/sync_manager.go#L175) is called to ensure we can acquire **all** the locks we want, and then we acquire them.
* (This is done under the protection of the go mutex for sync manager).
* So either WF2 will have the two locks it needs, or WF1 will have the 3 locks it needs. Neither can hold partial locks under the multiple lock code. (They can with nested templates, but they could before).

workflow/sync/sync_manager.go
```diff
@@ -113,102 +111,146 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
...
+		creationTime := wf.CreationTimestamp
+		lock.addToQueue(holderKey, priority, creationTime.Time)
+
+		ensureInit(wf, syncItems[i].getType())
+		acquired, already, newMsg := lock.checkAcquire(holderKey)
+		if !acquired && !already {
+			allAcquirable = false
+			if msg == "" {
+				msg = newMsg
+			}
```
isubasinghe, line 180: Is it important that we keep initialising locks when we know we cannot proceed?
Why not `break` instead?

workflow/sync/sync_manager.go
```diff
@@ -113,102 +111,146 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
...
+	default: // Not all acquirable
+		updated := false
+		for i, lockKey := range lockKeys {
+			currentHolders := sm.getCurrentLockHolders(lockKey)
+			if wf.Status.Synchronization.GetStatus(syncItems[i].getType()).LockWaiting(holderKey, lockKey, currentHolders) {
+				updated = true
+			}
+		}
+		return false, updated, msg, failedLockName, nil
+	}
```
isubasinghe, line 211: This would be more readable in a if statement with early return imo.

workflow/sync/sync_manager.go
isubasinghe: Yup agree with @Joibel on this one

workflow/sync/semaphore.go
```diff
@@ -3,69 +3,63 @@ package sync
...
 	"strings"
-	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
 	sema "golang.org/x/sync/semaphore"
 )
 
-type PrioritySemaphore struct {
+type prioritySemaphore struct {
```
isubasinghe, line 12: Not relevant on this PR but just a note: 
We should get rid of the semaphore interface and use this directly.

There is no need for indirection anymore given we just have a semaphore now.

workflow/sync/lock_name.go
isubasinghe: Imo SyncItem does not need anything else here, agree with @Joibel on this one.

isubasinghe (commented):
Just had a few questions

workflow/sync/sync_manager.go
agilgur5: > providing a good UI around how locks are being held in a more accessible way would be something I'd like to have.

There is a feature request for this in #11688

> I don't think we should attempt to mitigate deadlocks

I generally agree on this, as this is basically impossible to do well and becomes a general programming bug/problem to deal with (including Turing completeness and what not).

There is a simple case that is out of scope of this PR that I believe is currently allowed but could be mitigated:
When a Workflow's own templates cause a deadlock.

workflow/sync/sync_manager.go
Joibel: Each lock has a separate queue. We want to call addToQueue for all the locks we'd like to acquire.

Pop a break in here and the multiple semaphore tests [TestPriority/Priority wf1 will be able to grab the mutex here](https://github.com/argoproj/argo-workflows/blob/22b64f98998754216e56bbf768fc66f92ecb0f5e/workflow/sync/multiple_test.go#L378) before `wfhigh` violating priority, and [TestMultipleMutexLock/MultipleMutexOrdering wf2 will be able to grab mutex two despite wfall asking for it first](https://github.com/argoproj/argo-workflows/blob/22b64f98998754216e56bbf768fc66f92ecb0f5e/workflow/sync/multiple_test.go#L158) violating ordering.

workflow/sync/sync_manager.go
Joibel: I was under the impression `switch` was considered idiomatic go, so I've been trying to wean myself off ifs even for short ones.
Do you require it changing?

workflow/sync/syncitems.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (i *syncItem) getType() v1alpha1.SynchronizationType {
+	switch {
+	case i.semaphore != nil:
+		return v1alpha1.SynchronizationTypeSemaphore
+	case i.mutex != nil:
+		return v1alpha1.SynchronizationTypeMutex
+	default:
+		return v1alpha1.SynchronizationTypeUnknown
```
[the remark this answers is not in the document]
Joibel: I did consider it. I didn't introduce `SynchronizationTypeUnknown`, and it would mean a change to [GetStatus](https://github.com/argoproj/argo-workflows/blob/22b64f98998754216e56bbf768fc66f92ecb0f5e/pkg/apis/workflow/v1alpha1/workflow_types.go#L3878) which I'm often calling as `GetStatus(getType())` so I tried not to refactor too far.
Can we do this in another PR instead to avoid scope creep.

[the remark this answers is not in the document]
isubasinghe: Yeah okay in that case, this makes sense for a later PR

workflow/sync/sync_manager.go
isubasinghe: Right okay fair enough in that case.

workflow/sync/sync_manager.go
isubasinghe: No requirement, I will leave it up to you, because now we are effectively talking about subjective preferences.
I just think this pattern will read nicer:
```go
if cond {
  return true
}
return false
```

isubasinghe (approved):
Thanks for clarifying my questions, I've put in a suggestion feel free to make the change or just merge this in.

docs/synchronization.md
Joibel: I've moved this down from however it ended up in the top of the file

examples/synchronization-mutex-tmpl-level.yaml
Joibel: Added as separate examples so that they're:
* Runnable
* Tested

JPZ13:
Hey @agilgur5 - have @Joibel's additional commits addressed your change request? We need to get this one moving for a customer

sarabala1979:
I had discuss with @Joibel. I liked the Approach and approved. @agilgur5 is aready reviewing the code indetail. Please approved it once comments are address

agilgur5:
I haven't been feeling great the past week, so dismissed my review to unblock this. Can iterate more on docs later as I do have some thoughts on them

docs/synchronization.md
agilgur5: @tooptoop4 filed a follow-up feature request for this: https://github.com/argoproj/argo-workflows/issues/13630

examples/synchronization-mutex-tmpl-level.yaml
agilgur5: I could have been more specific here -- it was a quick initial glance -- but I meant _all_ the examples, including the in-line ones in the docs

> * Tested

This is good to have though.

> Since people will look at the examples and not know that they have future syntax etc

Indeed it looks like this is happening almost as soon as this was merged 😕 https://github.com/argoproj/argo-workflows/issues/13644 plus [two](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1726994955560259) [Slack threads](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1726835793647219?thread_ts=1726835793.647219&cid=C01QW9QSSSK). The issue and former Slack thread are also from a [long](https://github.com/argoproj/argo-workflows/discussions?discussions_q=author%3Ani-todo-spot) [time](https://github.com/argoproj/argo-workflows/issues?q=author%3Ani-todo-spot) user, i.e. not a newbie.

I'm writing up a PR to try to mitigate that

agilgur5: Addressed in #13645

agilgur5:
> I haven't been feeling great the past week, so dismissed my review to unblock this. Can iterate more on docs later as I do have some thoughts on them, but haven't quite had the brainpower to review properly

I wrote up a follow-up docs PR in #13645 which also addresses the immediate user confusion with versions per #13644 et al

workflow/sync/lock_name.go
agilgur5: similar concept was discussed below in https://github.com/argoproj/argo-workflows/pull/13358#discussion_r1714498505. `Unknown` already existed before, but would be good to prevent it entirely during validation as a separate PR

agilgur5:
> > In that case, I am wondering whether we should have a separate "fix" to backport that errors out when both are provided. Since this feature will only land in 3.6 and the silent ignore in earlier versions is very confusing/surprising as it stands
> 
> Yes, this would be nice if it got done.

I have a WIP PR for this in #13669. Might want someone here to take that over as some of the logic would be desirable to transfer to 3.6+ as well. It would also be a good place to add deprecation warnings in the logs in 3.6+

PR #13360, opened 2024-07-18
feat: Support authentication via Shared Access Signatures (SAS) for Azure artifacts (Fixes #10297) 

### LINKED ISSUE #10297: Support authentication via Shared Access Signatures (SAS) for Azure artifacts
# Summary

Support authentication via Shared Access Signatures (SAS) for Azure artifacts

# Use Cases

With argo v3.4 support for Azure artifacts has been implemented. As far as I understand, the current implementation only supports authentication via [Storage Account access keys](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage?toc=%2Fazure%2Fstorage%2Fblobs%2Ftoc.json&bc=%2Fazure%2Fstorage%2Fblobs%2Fbreadcrumb%2Ftoc.json&tabs=azure-portal), but not via [Shared Access Signatures (SAS)](https://learn.microsoft.com/en-us/azure/storage/blobs/sas-service-create?tabs=dotnet). 
I would like to be able to use SAS  for authentication, too. 

@brianloss Since you worked on this feature, could you confirm that I am right why my assumption? I could not find a way to make it work with SAS tokens. (and thanks btw for implementing it!)

Me/my team would potentially be willing do this contribution.

---

kavishdahekar-sap:
Fixes #10297

### Motivation

Currently, argo-workflows support Azure blob storage artifacts by authentication only via Access Keys. Azure, however, also supports auth via [Shared Access Signatures (SAS)](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview?toc=%2Fazure%2Fstorage%2Fblobs%2Ftoc.json&bc=%2Fazure%2Fstorage%2Fblobs%2Fbreadcrumb%2Ftoc.json).
This PR adds support for authentication via SAS.

### Modifications

To avoid spec changes, we re-use `spec.name.inputs.name.azure.accountKeySecret` to hold either an access key or SAS token. A new `isSASAccountKey` method checks whether the provided secret contains an access key or a SAS token. Since SAS tokens are nothing but URL query parameters, a simple regex can identify a SAS token. Once identified, we use the SAS token to initialize a `NewClientWithNoCredential` client instead of the `NewClientWithSharedKeyCredential` which will continue to be used for regular access keys.

### Verification

Tests have been added to `workflow/artifacts/azure/azure_test.go` to check both the `isSASAccountKey` as well as the correct functioning of `NewClientWithNoCredential` via azurite.

### DIFF
--- a/docs/configure-artifact-repository.md
+++ b/docs/configure-artifact-repository.md
@@ -393,6 +393,52 @@ used and authentication with Azure will be attempted using a
 [`DefaultAzureCredential`](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication)
 instead.
 
+## Configuring Azure Blob Storage with SAS 
+
+If you do not wish to use an access key, you may also use a [shared access signature (SAS)](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview?toc=%2Fazure%2Fstorage%2Fblobs%2Ftoc.json&bc=%2Fazure%2Fstorage%2Fblobs%2Fbreadcrumb%2Ftoc.json).
+Create an Azure Storage account and a container within that account. There are a number of
+ways to accomplish this, including the [Azure Portal](https://portal.azure.com) or the
+[CLI](https://docs.microsoft.com/en-us/cli/azure/).
+
+1. Retrieve the blob service endpoint for the storage account. For example:
+
+   ```bash
+   az storage account show -n mystorageaccountname --query 'primaryEndpoints.blob' -otsv
+   ```
+
+2. Retrieve the shares access signature for the storage account. For example:
+
+   ```bash
+   az storage container generate-sas --account-name <storage-account> --name <container> --permissions acdlrw --expiry <date-time> --auth-mode key
+   ```
+
+3. Create a kubernetes secret to hold the storage account key. For example:
+
+   ```bash
+   kubectl create secret generic my-azure-storage-credentials \
+     --from-literal "sas=$(az storage container generate-sas --account-name <storage-account> --name <container> --permissions acdlrw --expiry <date-time> --auth-mode key)"
+   ```
+
+4. Configure `azure` artifact as following in the yaml.
+
+```yaml
+artifacts:
+  - name: message
+    path: /tmp/message
+    azure:
+      endpoint: https://mystorageaccountname.blob.core.windows.net
+      container: my-container-name
+      blob: path/in/container
+      # accountKeySecret is a secret selector.
+      # It references the k8s secret named 'my-azure-storage-credentials'.
+      # This secret is expected to have have the key 'sas',
+      # containing the base64 encoded shared access signature to the storage account.
+
+      accountKeySecret:
+        name: my-azure-storage-credentials
+        key: sas
+```
+
 ## Configure the Default Artifact Repository
 
 In order for Argo to use your artifact repository, you can configure it as the
--- a/docs/configure-artifact-repository.md
+++ b/docs/configure-artifact-repository.md
@@ -166,7 +166,7 @@ artifacts:
       key: path/in/bucket
       # serviceAccountKeySecret is a secret selector.
       # It references the k8s secret named 'my-gcs-credentials'.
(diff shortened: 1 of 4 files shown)


docs/configure-artifact-repository.md
```diff
@@ -344,6 +344,10 @@ Create an Azure Storage account and a container within that account. There are a
 ways to accomplish this, including the [Azure Portal](https://portal.azure.com) or the
 [CLI](https://docs.microsoft.com/en-us/cli/azure/).
 
+There are two ways to allow argo to authenticate it's access to your Azure storage account.
+
+### Using Azure access keys
```
brianloss, line 349: This makes it seem like it's either access key or SAS. However, there's also the `useSDKCreds` option where you don't need a secret set and it will use a `DefaultAzureCredential` which supports a number of methods of authentication. I would call that option out first, since it's probably the most secure and preferred auth method. Then access key and SAS are the other two options.

kavishdahekar-sap: Yes that makes sense! Let me push a change to the doc!

kavishdahekar-sap:
Sorry I am not sure why the code-owners were not added as reviewers to this PR.
So mentioning the reviewers personally here, could any one of you please review this PR? Thanks!
(@agilgur5, @sarabala1979, @terrytangyuan)

brianloss (approved):
LGTM

kavishdahekar-sap:
Hi @agilgur5, could you please be able to review this PR or add relevant reviewers who could help with reviewing?
Thank you so much!

terrytangyuan (approved):
Thanks for the PR @kavishdahekar-sap and @brianloss for reviewing!

docs/configure-artifact-repository.md
```diff
@@ -373,25 +359,98 @@ artifacts:
...
+      # then useSDKCreds should be set to true. The accountKeySecret is not required
+      # and will not be used in this case.
+      useSDKCreds: true  
 ```
 
-If `useSDKCreds` is set to `true`, then the `accountKeySecret` value is not
-used and authentication with Azure will be attempted using a
-[`DefaultAzureCredential`](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication)
-instead.
+In addition to managed identities, Argo workflows also support authentication using access keys and SAS tokens.
```
agilgur5, line 369:
```suggestion
You can also authenticate with [Access Keys](#using-azure-access-keys) and [Shared Access Signatures](#using-azure-shared-access-signatures].
```

- simplify, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- use active voice, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice)
- Access Keys is a proper noun, no?

- also link to the two sections and use the full name for SAS as it has not been introduced yet.

docs/configure-artifact-repository.md
```diff
@@ -373,25 +359,98 @@ artifacts:
...
+          # It references the k8s secret named 'my-azure-storage-credentials'.
+          # This secret is expected to have the key 'account-access-key',
+          # containing the base64 encoded credentials to the storage account.
+          accountKeySecret:
+            name: my-azure-storage-credentials
+            key: account-access-key     
+    ```
+
+### Using Azure Shared Access Signatures (SAS)
+
```
agilgur5, line 412: This needs a version specifier a la #12581 et al

```suggestion

> v3.6 and after

```

agilgur5 (commented):
This will need a more thorough docs review, but off hand at the very least it needs version specifiers

docs/configure-artifact-repository.md
kavishdahekar-sap: Thanks @agilgur5 ! I'll raise a separate PR for this!

agilgur5: Addressed in #13376

docs/configure-artifact-repository.md
agilgur5: Addressed in #13376

PR #13363, opened 2024-07-18
fix(cron): `schedules` with `timezone`

Joibel:
### Motivation

Using `schedules` with a timezone specified does not work. The cron schedule given to the cron backend includes the CRON_TZ=zone string and should not.

### Modifications

The actual fix is one line to correctly honor the timezone flag on `getSchedules` so that it doesn't double append when that flag is set, nor single append when it is not. GetSchedules() would return an invalid cron string resulting in the error:

```
could not schedule CronWorkflow" cronWorkflow=argo/go-forward error="expected exactly 5 fields, found 6: [CRON_TZ=America/Los_Angeles 30 2 * * *]
```

### Verification

Added some more unit tests to check this behaves as expected, and a smoke e2e test.

I have to say, the original set of tests for #12616 looked like it really should cover this, so I was surprised when nothing broke when I made the change and it took a while to work out what was going wrong.

### Target branches

Just `main`, do not try to backport.

### DIFF
--- a/pkg/apis/workflow/v1alpha1/cron_workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/cron_workflow_types.go
@@ -174,7 +174,7 @@ func (c *CronWorkflowSpec) getSchedules(withTimezone bool) []string {
 			if withTimezone {
 				schedule = c.withTimezone(schedule)
 			}
-			schedules[i] = c.withTimezone(schedule)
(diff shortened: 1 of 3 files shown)


eduardodbr:
Looks good, thanks for the fix!

agilgur5 (approved):
LGTM, thanks for discovering, fixing, and testing this!

PR #13292, opened 2024-07-02
docs: full copy-edit of CronWorkflows page

agilgur5:
### Motivation

Made a lot of modifications to this page during reviews of #13280 and #12696, so wanted to make the rest match the current style guides.
Also aligns with all my other full copy-edits of docs pages

### Modifications

- <details><summary>follow the style guides</summary>

  - simplify a bunch, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - use active voice and present tense where possible, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-present-tense)
  - avoid Latin phrases, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases)
  - avoid leveling words, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding)
  - prefer 1 sentence per line of markdown, per [style guide](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/)
  
</details>

- <details><summary>reformat the table of options so that it is readable in markdown </summary>

  - don't center the columns, left-align them
    - there's no reason to center them and this is inconsistent with the rest of the docs; unnecessarily complicated
      - note that later rows did not do this either
  - put the ending `|` one space after the row
    - the very long amount of whitespace would wrap around every row even in my wide screen monitors, despite that only a few options have long descriptions

</details>

- add Wikipedia links to IANA tzs and DST similar to the existing cron link
- specify that the same cron library is used by k8s `CronJobs`

### Verification

`make docs` passes

Issue #12694, opened 2024-02-23
RFC/Proposal: DAG Transforms for extensibility, composability, and maintainability


agilgur5:
# Summary

Allow for pluggable DAG transforms to implement various features. Closely related to #6943, but in a way that could potentally substantially reduce maintenance of the core by reducing its feature set by converting some features to transforms. It would also make Workflows much more composable, allowing for some of those features to be decoupled and have independent versions, meaning their release cycle would not be dependent on the core.

## Use Cases

For example:
1. `steps` could be rewritten as a DAG transform -- `steps` would just be a different syntax that converts into DAGs
1. `hooks` could be rewritten as a DAG transform -- `hooks` syntax could also convert into regular DAGs
1. `loops` could be a DAG transform
1. Executor Plugins could be rewritten as DAG transforms -- `container` tasks with custom `image`s are very similar
1. Artifact Plugins #5862 and potentially all of artifacts in general could be rewritten as DAG transforms -- the `init` and `wait` containers could be replaced with custom `image`s similar to Executor Plugins above
1. `http`, `data`, `resource`, `inline` etc templates could all be DAG transforms
1. Probably more

#### Pure compute transforms - can be implemented with secure WASM sandbox

As many DAG transforms could be pure computation with no requirement of network access or I/O, WASM could be a great way to implement these types of "plugins" (as mentioned in https://github.com/argoproj/argo-workflows/issues/6943#issuecomment-1901951588) as WASM is a secure sandbox and supports many languages.
By no network access or I/O, I mean that a DAG transform "plugin" could just take the existing DAG as input and output a new DAG. That new DAG could use custom images etc (that the user would have to trust), but the transform itself could be pure compute.

Ultimately, WASM is more an implementation detail, but as a secure sandbox, it allows us to run such code on the Controller directly. Similar to how Envoy and many other tools allow WASM extensions.
A custom image could do something similar, but in a less secure fashion, with a lifecycle requirement (like current executor plugins), and some deployment complexity.

#### Registering transforms

Transform WASM binaries could be mounted to the Controller (some [CD plugins](https://argo-cd.readthedocs.io/en/stable/developer-guide/extensions/ui-extensions/) work similarly as files on the Server or Controller). The Controller would look in a certain directory and (attempt to) register all those binaries as transforms.

[...]

Issue #8348, opened 2022-04-09
CronWorkflows: support more schedule expressions like "last day of the month"


tooptoop4:
ability to specify schedule like 11am on '2nd weekday of the month', 'last 5 days of the month', 'last business day of the month'

see examples at the end of:
https://confluence.atlassian.com/servicedeskserver030/constructing-cron-expressions-for-a-filter-subscription-761769111.html

https://en.wikipedia.org/wiki/Cron#Non-standard_characters
https://github.com/robfig/cron/pull/325#issuecomment-1046319369
https://github.com/robfig/cron/pull/397
https://github.com/robfig/cron/pull/321
cc @robfig

navi1024:
To be precise following 3 expressions are supported by quartz scheduler and there plentiful usecases where these are useful

> * The ‘L’ character is allowed for the day-of-month and day-of-week fields.
> * The ‘W’ is used to specify the weekday (Monday-Friday) nearest the given day.
> * The ‘#’ is used to specify “the nth” XXX weekday of the month.

Also, there is no year literal supported in cron expression as on today. It is mainly limitation of the upstream library. 

Kuberentes also uses same cron library and following is bug reported 

https://github.com/kubernetes/kubernetes/issues/121088

Issue #8347, opened 2022-04-09
CronWorkflow referencing custom calendars (ie with public holidays defined)


tooptoop4:
define a custom calendar that has some public holidays excluded (maybe in some configmap)
then set a cronwf to run on the days defined in the custom calendar

dqwork:
Would just like to express interest in this too

PR #13365, opened 2024-07-19
feat: minimum interval for CronWorkflows

Joibel:
A new field in CronWorkflows `minimumInterval` which specifies a minimum monotonic time between workflow creations.

### Motivation

This has two main uses I can think of
* allow CronWorkflows which on first glance will run once a day to do so even in a daylight saving observing timezone.
* allow "run every interval" workflows which run rapidly on first injection into a cluster and as soon as possible again if interval has passed after downtime of the controller.

This is also possible a generically useful change now that we have `schedules` which allow for much more complex scheduling.

### Modifications

If we attempt to schedule inside minimum interval do not do so.

### Verification

New e2e test

Tested both of the next DST changes forward and backward against the CronWorkflows in the documentation.


agilgur5 (commented):
The usefulness, esp in the DST example, makes sense. However, I'm wondering if there is precedent for this in other cron implementations? Since I imagine this has been handled before

agilgur5:
Maybe a different question then, how do people currently work around this issue in existing schedulers? (other than, well, not using a TZ or time affected by DST)

Joibel:
> Maybe a different question then, how do people currently work around this issue in existing schedulers? (other than, well, not using a TZ or time affected by DST)

Apart from those, I have no evidence of anyone doing it. It's a hard thing to research outside of `cron`, but no solutions for cron based scheduling anyway.

docs/cron-workflows.md
```diff
@@ -41,16 +41,17 @@ You can use `CronWorkflow.spec.workflowMetadata` to add `labels` and `annotation
...
-| `startingDeadlineSeconds`    | `0`                    | Seconds after [the last scheduled time](#crash-recovery) during which a missed `Workflow` will still be run. |
-| `successfulJobsHistoryLimit` | `3`                    | Number of successful `Workflows` to persist |
-| `failedJobsHistoryLimit`     | `1`                    | Number of failed `Workflows` to persist |
-| `stopStrategy`               | `nil`                  | v3.6 and after: defines if the CronWorkflow should stop scheduling based on a condition |
+| Option Name                  | Default Value          | Description                                                                                                                                                                                |
+|:----------------------------:|:----------------------:|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
agilgur5, line 45: I'd prefer if we didn't change the formatting for every row. In #13292 I specifically addressed that and tried to use a formatting that wouldn't require changing every other row if only one description changed

docs/cron-workflows.md
```diff
@@ -108,6 +109,36 @@ For example, with `timezone: America/Los_Angeles`:
...
+An example 01:30:00 schedule
+
+```yaml
+schedule: 30 1 * * *
+minimumInterval: 120m
+```
+
+This will schedule at the first 01:30 on a skip backwards change.
+The second will not run because of `minimumInterval`.
+
```
agilgur5, line 141:
```suggestion
#### `minimumInterval`

You can set a minimum interval between runs.

For example, if you want to ensure you only run once a day during a DST leap forward:

```yaml
schedules:
  - 30 2 * * *
  - 0 3 * * *
minimumInterval: 60m
\```

The 3:00 schedule will never run except on the day the clock leaps forward over 2:30.
In that case, only the 3:00 schedule will run and the 2:30 will not.

Another example, if you want to ensure you only run once a day during a DST leap backward:

```yaml
schedule: 30 1 * * *
minimumInterval: 120m
\```

The first run will be at 01:30.
On the day the clock leaps backward, the second will not run because of `minimumInterval`.

```

- I think we can simplify this a bit and combine the two sections as two examples. We can also reduce the verbal descriptions as the examples are a lot clearer than them (I found the initial descriptions to be a bit confusing and verbose. The descriptions afterward are good, though I simplified those as well)
- Use consistent language with "runs" vs. "schedule", and "leap" vs "skip"
  - "leap" has a less technical terminology and is used often for DST (and infrequently for non-DST), so I think works a bit better as specific to DST

docs/cron-workflows.md
```diff
@@ -41,16 +41,17 @@ You can use `CronWorkflow.spec.workflowMetadata` to add `labels` and `annotation
...
+| `timezone`                   | Machine timezone       | [IANA Timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) to run `Workflows`. Example: `America/Los_Angeles`                                                           |
+| `suspend`                    | `false`                | If `true` Workflow scheduling will not occur. Can be set from the CLI, GitOps, or directly                                                                                                 |
+| `concurrencyPolicy`          | `Allow`                | What to do if multiple `Workflows` are scheduled at the same time. `Allow`: allow all, `Replace`: remove all old before scheduling new, `Forbid`: do not allow any new while there are old |
+| `minimumInterval`            | None                   | The minimum interval between executions of this workflow. Examples: `90s`, `10m`, `2h`                                                                                                     |
```
agilgur5, line 50:
```suggestion
| `minimumInterval`            | None                   | The minimum time between runs. Examples: `90s`, `10m`, `2h` |
```

simplify the description and consistently use "runs". Also use "time" (vs "interval") like the field description, which also indicates the units.

agilgur5 (commented):
A few docs/spec suggestions

docs/cron-workflows.md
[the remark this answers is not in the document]
Joibel: Ah, TIL. Nice to know that.

agilgur5:
I'm also wondering if we perhaps want to add some expression "conditional" of some sort. Since `minimumInterval` and other such features could open a lot of flood gates to supporting things outside of the cron spec/scheduler. We already use expressions in a lot of places and they can be quite flexible so that would open up more possibilities into user-land instead of requiring core features

Joibel:
> I'm also wondering if we perhaps want to add some expression "conditional" of some sort. Since `minimumInterval` and other such features could open a lot of flood gates to supporting things outside of the cron spec/scheduler. We already use expressions in a lot of places and they can be quite flexible so that would open up more possibilities into user-land instead of requiring core features.
> 
> Off the top of my head and pure pseudo-code, `minimumInterval` could be expressed as something like `when: schedule[1].nextScheduledTime - schedule[0].nextScheduledTime > 120m`.
> 
> Oh, and now I'm realizing that that could potentially be done on the Workflow itself already as a workaround is similar to existing ones like #8348, #8347 etc. Hmmm... maybe leaving that with user-land workaround is better than adding this to core...

I think we should support this without starting a workflow. It pushes the decision left, and reduces load considerably. I'm aware of users with tens of thousands of cron workflows, where the overhead of an in workflow decision would be unusable.

I'm also in favour of having well tested patterns for this stuff. It's hard work to test a scheduled workflow, and the cost of having your schedule go wrong can be high. This is why I'd quite like to [support gronx](https://github.com/argoproj/argo-workflows/pull/13365#issuecomment-2239042254) as an alternative cron implementation, it would be simpler to get some of the other behaviours you've linked to with that, and more likely to work consistently and be easier to provide support.

I like the idea of this being an expression, I'm sure we can come up with other uses for it. I would instead go for `now - cronworkflow.lastScheduledTime >120m` as a way of implementing `minimumInterval` as it has the other properties I was after.

Exposing the schedule would be possible I guess.

Then it comes down to whether this is something we'd consider reliable. I'll experiment with implementing it this way.

isubasinghe:
I generally like @agilgur5 's suggestion here. 
A `when` clause or something similar would be excellent.

agilgur5:
> I think we should support this without starting a workflow. It pushes the decision left, and reduces load considerably. I'm aware of users with tens of thousands of cron workflows, where the overhead of an in workflow decision would be unusable.

I don't necessarily disagree with this, but it raises the question of how much we should put into the core if it can be done in user-land. Same reason I've suggested DAG transforms (#12694) and would like to get more plugins supported. Expressions are broad enough that they can support quite a lot of use-cases, so they are somewhat similar to plugins in that respect though.

> I would instead go for `now - cronworkflow.lastScheduledTime >120m` as a way of implementing `minimumInterval` as it has the other properties I was after.

Yea that makes more sense, my suggestion was very back of the napkin.

> I'm also in favour of having well tested patterns for this stuff. It's hard work to test a scheduled workflow, and the cost of having your schedule go wrong can be high.

Agree

> This is why I'd quite like to [support gronx](https://github.com/argoproj/argo-workflows/pull/13365#issuecomment-2239042254) as an alternative cron implementation, it would be simpler to get some of the other behaviours you've linked to with that, and more likely to work consistently and be easier to provide support.

> I thought about adopting [gronx](https://github.com/adhocore/gronx) as an alternative cron scheduler. The user would have a choice per workflow. It would be relatively clean to do. But doesn't solve either problem this does.

I hadn't thought about giving an option and running multiple schedulers, but I suppose we could do that too.
Though I do wonder at what point that should be a plugin too -- i.e. "plug-in your favorite cron scheduler". To be fair, most of them are very similar, but have some small but important differences 😕
And I do think defaulting to k8s's cron scheduler helps for consistency too, as you mentioned as well.

Joibel:
#13474 is the replacement PR

docs/cron-workflows.md
Joibel: I didn't like making this a single document. In the skim reading case you MUST only care about one or the other per cronworkflow, so have kept the two items separate.

I have also kept the description verbose as they are complex to understand I believe it's helpful for users to understand how these work rather than just use them as magic words.

Joibel:
#13723 contains the DST skip forward/backwards documentation from this. Comments should be made there from now on please.

docs/cron-workflows.md
agilgur5: > In the skim reading case

> I have also kept the description verbose as they are complex to understand I believe it's helpful for users to understand how these work

These two are mutually exclusive scenarios. Either you can skim read it or it's verbose, and therefore you cannot skim read it. To understand how it works, you also would not be skim reading.

> In the skim reading case

In the skim reading case, you'd just search "DST" rather than _which_ transition. 
But I don't think we should be tailoring docs to skim readers in any case, that's not a good way of building depth, and as per above, you can really only choose one or the other.

> I have also kept the description verbose as they are complex to understand I believe it's helpful for users to understand how these work 

I don't think they're complex to understand IMO, and the verbosity actually makes it _harder_ to understand. More words is not necessarily a good thing, especially when the words are inconsistent. The style guide says to use [simplicity and directness](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language) for a reason, after all. And exceptions to the style guide lead to bikeshedding ad infinitum.

> rather than just use them as magic words.

There's a Wikipedia link provided. We don't define the term "leap", it already exists. The same way we don't define the term "Pod" and rather link to the k8s docs.

agilgur5: Followed up with https://github.com/argoproj/argo-workflows/pull/13723#discussion_r1805617431

PR #13037, opened 2024-05-12
fix(test): add a `sleep` to `TestLoadToStream`

agilgur5:
### Motivation

- it very infrequently has a test flake in it
  - have had this in my `git stash` since at least September, and only hit this a few times in those ~9 months
    - I thought it had been fixed by some other changes (#11753 in particular) for a while, but [just experienced it again](https://github.com/argoproj/argo-workflows/actions/runs/9046011165/job/24856369833) (and [twice](https://github.com/argoproj/argo-workflows/actions/runs/9248749132/job/25439502972) [more](https://github.com/argoproj/argo-workflows/actions/runs/9249013492/job/25441555518)), so nope
    
<details> <summary>Error log:</summary>

```
--- FAIL: TestLoadToStream (0.01s)
    --- FAIL: TestLoadToStream/Success (0.01s)
        load_to_stream_test.go:99: 
            	Error Trace:	/home/runner/work/argo-workflows/argo-workflows/workflow/artifacts/common/load_to_stream_test.go:99
            	Error:      	Not equal: 
            	            	expected: 25
            	            	actual  : 26
            	Test:       	TestLoadToStream/Success
```

</details>

### Modifications

- also fix incorrect comment in `when.go` -- [`defaultTimeout` defaults to 60s](https://github.com/argoproj/argo-workflows/blob/0c1398ea6fa66cc77156f15a6d69c260c6b524da/test/e2e/fixtures/e2e_suite.go#L48), not 30s

### Verification

Test passes more often

PR #13366, opened 2024-07-19
fix(test): load to stream tmp file counting

Joibel:
Replaces #13037 

### Motivation

This test has an infrequent flake, see #13037 

### Modifications

The current test counts files in `/tmp` and hopes there are the same number before and after the test, which is fragile

Add a prefix to the stream temporary files.
Only count files in `/tmp` which have this prefix.

### Verification

Tested locally. Tested with a file called `/tmp/wfstream-asdfasd` and extra logging in the test to show the numbers are as expected.
Tested with the `os.Remove()` removed from `selfDestructingFile.Close()` to show failure.

### DIFF
--- a/workflow/artifacts/common/load_to_stream.go
+++ b/workflow/artifacts/common/load_to_stream.go
@@ -12,6 +12,8 @@ import (
 	wfv1 "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 )
 
+const loadToStreamPrefix = `wfstream-`
+
 // wrapper around os.File enables us to remove the file when it gets closed
 type selfDestructingFile struct {
 	os.File
@@ -28,7 +30,7 @@ func (w selfDestructingFile) Close() error {
 func LoadToStream(a *wfv1.Artifact, g ArtifactDriver) (io.ReadCloser, error) {
 	log.Infof("Efficient artifact streaming is not supported for type %v: see https://github.com/argoproj/argo-workflows/issues/8489",
 		reflect.TypeOf(g))
-	filename := "/tmp/" + rand.String(32)
+	filename := "/tmp/" + loadToStreamPrefix + rand.String(32)
 	if err := g.Load(a, filename); err != nil {
 		return nil, err
 	}
--- a/workflow/artifacts/common/load_to_stream_test.go
+++ b/workflow/artifacts/common/load_to_stream_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -55,6 +56,22 @@ func (a *fakeArtifactDriver) ListObjects(artifact *wfv1.Artifact) ([]string, err
 	return nil, fmt.Errorf("not implemented")
 }
 
+func filteredFiles(t *testing.T) ([]os.DirEntry, error) {
+	t.Helper()
+
+	filtered := make([]os.DirEntry, 0)
+	entries, err := os.ReadDir("/tmp/")
+	if err != nil {
+		return filtered, err
+	}
+	for _, entry := range entries {
+		if strings.HasPrefix(entry.Name(), loadToStreamPrefix) {
+			filtered = append(filtered, entry)
+		}
+	}
+	return filtered, err
+}
+
 func TestLoadToStream(t *testing.T) {
 	tests := map[string]struct {
 		artifactDriver ArtifactDriver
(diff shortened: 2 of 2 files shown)


workflow/artifacts/common/load_to_stream.go
```diff
@@ -28,7 +28,7 @@ func (w selfDestructingFile) Close() error {
 func LoadToStream(a *wfv1.Artifact, g ArtifactDriver) (io.ReadCloser, error) {
 	log.Infof("Efficient artifact streaming is not supported for type %v: see https://github.com/argoproj/argo-workflows/issues/8489",
 		reflect.TypeOf(g))
-	filename := "/tmp/" + rand.String(32)
+	filename := "/tmp/wfstream-" + rand.String(32)
```
agilgur5, line 31: Hmm while this source change seems innocuous enough, this does suggest that maybe the test shouldn't be counting files here at all?

Like if the filename should be passed in or returned instead and tested directly, rather than indirectly counting files

agilgur5 (commented):
Thanks for root causing this!

workflow/artifacts/common/load_to_stream.go
Joibel: Perhaps. I considered doing the right thing with [os.CreateTemp](https://pkg.go.dev/os#CreateTemp) or MkdirTemp

But this is a test of a fix to hack around a lack of stream loading in various artifact drivers. The function must conform to the interface there.

The test's primary purpose isn't to check deletion, and mangling the code to make just this portion of the test workable in this case appeared more likely to introduce more bugs than be beneficial. The better choice than doing that would be to remove this test IMHO.

agilgur5: > I considered doing the right thing with [os.CreateTemp](https://pkg.go.dev/os#CreateTemp) or MkdirTemp

Yea that would be more correct too



> The function must conform to the interface there.

It actually doesn't seem to need to as far as I can tell. [`OpenStream`](https://github.com/argoproj/argo-workflows/blob/1f0adca09374362d47cf5fd9f19b4b4b2e104d03/workflow/artifacts/common/common.go#L21) is part of our internal interface (which is changeable too), but not `LoadToStream`.

So returning an additional `filename` or similar should be fine, we just may need to change the `OpenStream` calls that use this function to ignore that return arg.



> The test's primary purpose isn't to check deletion, and mangling the code to make just this portion of the test workable in this case appeared more likely to introduce more bugs than be beneficial. The better choice than doing that would be to remove this part of the test IMHO.

IMO I thought changing the filename was actually more intrusive than the return arg proposed above, since that actually changes user-visible outputs (not that a random temp filename should be relied on, but it does cause a change nonetheless), whereas the return arg would be purely internal.

Joibel: A function/method is implying a contract in it's parameters.

Returning the filename from `LoadToStream` would imply that this filename was usable by the caller. It is not, and under no circumstances should be used. I hadn't even considered changing this interface for this reason, and am still happy with this choice.

agilgur5: Ah yes, the classic conundrum of "this is only done for testability" 😅.
We do have code that does this in several places: in the UI, exported variables only used in tests, in the Server, `ClientForAuthorization` is passed in purely so it can be mocked in the tests (and plenty more). The latter option would work here too without violating the contract, instead it breaks encapsulation 😕

In general, I agree, we shouldn't change source code for the purpose of testability; that feels backwards. I suppose the problem is that you did do so here already; in functional terms you made the side-effect easier to observe, which the tests do. So the question is "how to correctly test a side-effect?".
I think it's a good safety test, so I'd rather not remove it, although potentially that behavior could be tested differently, e.g. a unit on the `Close` function. Maybe that's the best option?

If we go with your route, I would still prefer to change it to `os.CreateTemp` and make `wfstream-` a constant that the tests use.

Joibel: Sometimes testability improves the code, so then I'm happy to do it.

`os.CreateTemp` returns you a filehandle. `ArtifactDriver.Load()` takes a filename - so can't do that without more invasive surgery.

I could make `wfstream-` a constant, but otherwise this is getting out of hand for a PR to just fix a test flake.

agilgur5: > `ArtifactDriver.Load()` takes a filename - so can't do that without more invasive surgery.

oh dang, I missed that, sorry. maybe worth a separate "best practice" change in the future to refactor that, but yea def not worth right now.



> I could make `wfstream-` a constant

I'd be fine with that or remove this part of the test and replacing it with a unit on the `Close` function or something similar that is a more direct test. Let me know what you think. I think either is better than the current

> but otherwise this is getting out of hand for a PR to just fix a test flake.

we tend to discuss more "philosophical" matters of how to write the code properly now and in the future in general. I personally consider that a worthwhile investment, as it is not purely about the flake and may in fact impact how we handle other flakes and other side-effect testing code, i.e. setting precedent and best practices for the future.

Joibel: I've gone with the constant.

agilgur5 (approved):
LGTM, replacement with `os.CreateTemp` [deferred](https://github.com/argoproj/argo-workflows/pull/13366#discussion_r1686353185) as it requires a more substantial refactor

PR #13367, opened 2024-07-19
chore: remove legacy go build tags

Joibel:
Since [go 1.18](https://go.dev/doc/go1.18) `go fix` will remove the `// +build` lines from code as it is superseded by `//go:build`.

Two autogenerated files still have these in and I cannot see a way to prevent the generators recreating them, but otherwise this is just a tidy up. I considered automating a `go fix` run in CI, but that's hindered by this.

### DIFF
--- a/hack/docs/fields.go
+++ b/hack/docs/fields.go
@@ -1,5 +1,4 @@
 //go:build !fields
-// +build !fields
 
 package main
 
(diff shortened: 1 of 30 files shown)


agilgur5 (approved):
Ah I had been wondering why they seemed duplicative; legacy tags apparently.

LGTM, just removals and CI still runs smoothly.

Issue #1780, opened 2019-11-19
Pass `parallelism` via CLI? Error on string vs int


sebinsua:
Is it possible to pass in parallelism via the command line?

I tried to setup a parameter to do this, but got an error about the value not being a number.

```
Failed to parse workflow:
error unmarshaling JSON: while decoding JSON: json:
cannot unmarshal string into Go struct field WorkflowSpec.parallelism of type int64
```

If it's not possible by default to set this for the workflow from the command line, is there a way of coercing a string containing a number into a number?

**Edit**: Do you have [anything like an `int` filter](https://stackoverflow.com/a/28234670)? I heard you were using Jinja but I can't find it within the codebase.

simster7:
> Is it possible to pass in parallelism via the command line?

Currently not possible.

> Do you have anything like an int filter? I heard you were using Jinja but I can't find it within the codebase.

Not as far as I know.

> If it's not possible by default to set this for the workflow from the command line, is there a way of coercing a string containing a number into a number?

Could you elaborate a bit on this? Are you trying to run dynamically generated Workflows?

sebinsua:
> > If it's not possible by default to set this for the workflow from the command line, is there a way of coercing a string containing a number into a number?
> 
> Could you elaborate a bit on this? Are you trying to run dynamically generated Workflows?

I tried this at one point, but it doesn't work:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: global-parallelism-parameter-
spec:
  entrypoint: entrypoint-name
  arguments:
    parameters:
    - name: parallelism
      value: 10
  parallelism: "{{workflow.parameters.parallelism}}"
```

I'm not trying to run dynamicly generated workflows. I just wanted to be able to configure the level of parallelism a workflow should be run by at the point of submission.

simster7:
I see. I don't think we currently support a way to pass in parameters to the `Workflow` object... only to `Template` objects within a workflow.

A workaround suggestion I could give you would be to use `sed` to replace the placeholder value before submitting. Something like this:

```sh
$ cat workflow.yaml | sed 's/"{{workflow.parameters.parallelism}}"/10/g' | argo submit --watch -
```

dan-katz:
I'm also running into this issue when trying to have parallelism be a parameter of a workflow template

alexec:
I don't think any information is needed? The label was removed not added.

I think the solution executed in #3610 could work here too. 

[...]

PR #13368, opened 2024-07-19
feat: Support variables for workflow parallelism

andrzejressel:
Fixes partially #1780

### Motivation

This MR allows string (and by extension variables) to be used as Workflow parallelism

### Modifications

Workflow parallelism type as changed to IntOrString

### Verification

```
kind: Workflow
metadata:
  generateName: loop-sequence-
spec:
  parallelism: "{{workflow.parameters.parallelism}}"
  entrypoint: loop-sequence-example
  arguments:
    parameters:
    - name: parallelism
  templates:
  - name: loop-sequence-example
    steps:
    - - name: hello-world-x5
        template: hello-world
        withSequence:
          count: "5"

  - name: hello-world
    container:
      image: docker/whalesay:latest
      command: [cowsay]
      args: ["hello world!"]
```

I've tried also allowing this on Template level (I even have branch here https://github.com/andrzejressel/argo-workflows/tree/parallelism-string), but in these 2 places template is not resolved:

- https://github.com/andrzejressel/argo-workflows/blob/parallelism-string/workflow/controller/operator.go#L2772
- https://github.com/andrzejressel/argo-workflows/blob/parallelism-string/workflow/controller/operator.go#L2806


test/e2e/functional_test.go
```diff
@@ -783,6 +783,35 @@ spec:
...
+    container:
+      image: argoproj/argosay:v2
+      args: [sleep, 1s]
+`).
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow().
+		Then().
+		ExpectWorkflow(func(t *testing.T, md *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			assert.Equal(t, wfv1.WorkflowSucceeded, status.Phase)
```
agilgur5, line 811: mm I feel like we should probably check that this the parameter was correctly evaluated as well.
My brain seems to not be remembering how to do that off the top of my head though 😅

agilgur5:
> I've tried also allowing this on Template level (I even have branch here https://github.com/andrzejressel/argo-workflows/tree/parallelism-string), but in these 2 places template is not resolved:
> 
> * https://github.com/andrzejressel/argo-workflows/blob/parallelism-string/workflow/controller/operator.go#L2772
> * https://github.com/andrzejressel/argo-workflows/blob/parallelism-string/workflow/controller/operator.go#L2806

You probably need to make a change to the templating code or validation code. Maybe @eduardodbr could help as he's done some templating work before (it's one of few areas of the codebase that I haven't worked on too much yet)

PR #13369, opened 2024-07-19
fix(cron): various follow-ups for multiple schedules

eduardodbr:
### Motivation

Currently `argo cron list` and `argo cron get` are showing the schedules with timezone, when available. This PR removes it as it doesn't bring any value because it can be seen in the `timezone` field.

### Modifications

The schedule string can now be requested with and without the timezone. 
With timezone, it is used for the annotation `last-used-schedule`, so changing the timezone while keeping the schedule will be seen as a new schedule. 
Without timezone, it is used in `argo cron` CLI.

Also:
- update every instance of `schedule` to be `schedules` as suggested by @agilgur5 https://github.com/argoproj/argo-workflows/pull/12616#discussion_r1501850996
- update fields in CronWorkflows table for multiple schedules per https://github.com/argoproj/argo-workflows/pull/12616#pullrequestreview-1889629343
- add in-line version references and deprecation per https://github.com/argoproj/argo-workflows/pull/12616#discussion_r1495330587
- add example

### Verification

Current: 

`argo cron list` 

image

`argo cron get`

image

With this PR changes:

`argo cron list`

image

`argo cron get`

image

The `CronWorkflow` label `last-used-schedule` still uses the schedule with timezone

image


agilgur5 (commented):
It might make sense to split out the `schedule` -> `schedules` change https://github.com/argoproj/argo-workflows/pull/12616#discussion_r1501850996 into its own PR. That and the `v3.6 and after:` and other #12616 review follow-ups can be merged pretty much immediately while you're still working out the test failures etc here. And the `v3.6 and after:` portion is particularly useful for people who are looking at the `latest` docs.

Also there was one miss from my previous review https://github.com/argoproj/argo-workflows/pull/12616#pullrequestreview-1889629343, the CronWorkflow docs page's fields table

eduardodbr:
whille implementing this, I also found that `schedule` was still marked as required in the open api spec/json schema/CRD spec so I removed it. Do you think it should have a dedicated PR or is it ok to have it here? @agilgur5

agilgur5:
Is the CLI fix only relevant to 3.6? Or does it also impact 3.5/3.4?

If these are all just follow-up fixes to the multiple `schedules` PR, we can leave it all as one I suppose -- just would rename the PR.

If the CLI fix is relevant to 3.5/3.4, then I think we should split the rest out, so that it can be backported

eduardodbr:
> Is the CLI fix only relevant to 3.6? Or does it also impact 3.5/3.4?
> 
> If these are all just follow-up fixes to the multiple `schedules` PR, we can leave it all as one I suppose -- just would rename the PR.
> 
> If the CLI fix is relevant to 3.5/3.4, then I think we should split the rest out, so that it can be backported

this only applies to 3.6 as it was introduced in #12616

agilgur5:
I know `schedules` was, but I wasn't sure if this TZ duplication happened in earlier versions with single `schedule` as well

eduardodbr:
> I know `schedules` was, but I wasn't sure if this TZ duplication happened in earlier versions with single `schedule` as well

the timezone in `argo cron lilst` and `argo cron get` was also introduced in that PR

agilgur5:
Ok gotcha, wanted to make sure since I hadn't reviewed that whole PR. Then let's just fix-up the title and description to include all follow-ups as one

Also btw, no need to quote reply if you're replying to the whole thing in order. Quoting is useful for replying to parts at a time or out of order comments. In this case they read fluidly down and so quoting actually reads duplicatively

ui/src/app/cron-workflows/cron-workflow-list.tsx
```diff
@@ -156,7 +156,7 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
                                         <div className='columns small-2'>{w.metadata.namespace}</div>
                                         <div className='columns small-1'>{w.spec.timezone}</div>
                                         <div className='columns small-1'>
-                                            {w.spec.schedule != ''
+                                            {(w.spec.schedule ?? '') != ''
```
agilgur5, line 159:
```suggestion
                                            {w.spec.schedule
```

you can just use [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) for all these

ui/src/app/cron-workflows/cron-workflow-list.tsx
```diff
@@ -156,7 +156,7 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
...
                                                 ? w.spec.schedule
                                                 : w.spec.schedules.map(schedule => (
```
agilgur5, line 161:
```suggestion
                                            {w.spec.schedule || w.spec.schedules.map(schedule => (
```

this specific one can be further simplified

agilgur5: although tbh normalizing `schedule` to a single element array of `schedules` would be simpler to do everywhere

Joibel:
@eduardodbr, any chance that you'd have a chance to fix/update this so it can go into 3.6?

We're currently missing the documentation on multiple `schedules`, which is in this PR.

eduardodbr:
Hi @Joibel will try to finish this ASAP, sorry for the delay

docs/cron-workflows.md
```diff
@@ -43,7 +43,8 @@ You can use `CronWorkflow.spec.workflowMetadata` to add `labels` and `annotation
 
 | Option Name                  | Default Value          | Description |
 |:----------------------------:|:----------------------:|-------------|
-| `schedule`                   | None, must be provided | [Cron schedule](#cron-schedule-syntax) to run `Workflows`. Example: `5 4 * * *` |
+| `schedule`                   | None | [Cron schedule](#cron-schedule-syntax) to run `Workflows`. Example: `5 4 * * *`. Either `schedule` or `schedules` must be provided. |
```
agilgur5, line 46:
```suggestion
| `schedule`                   | None | [Cron schedule](#cron-schedule-syntax) to run `Workflows`. Example: `5 4 * * *`. Deprecated, see `schedules`. |
```

We can probably modify this to just say that `schedule` is deprecated. Similar to #13358.

@Joibel does that make sense to you for this feature?

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -45,7 +45,7 @@ type CronWorkflowSpec struct {
 	// WorkflowSpec is the spec of the workflow to be run
 	WorkflowSpec WorkflowSpec `json:"workflowSpec" protobuf:"bytes,1,opt,name=workflowSpec,casttype=WorkflowSpec"`
 	// Schedule is a schedule to run the Workflow in Cron format
```
agilgur5, line 47:
```suggestion
	// Schedule is a schedule to run the Workflow in Cron format. Deprecated, see Schedules
```

same as above

examples/cron-workflow-multiple-schedules.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: hello-world-multiple-schedules
+spec:
+  schedules: 
+  - "*/3 * * * *"
+  - "*/2 * * * *"
```
agilgur5, line 8:
```suggestion
  schedules: # v3.6 and after
    - "*/3 * * * *"
    - "*/2 * * * *"
```

- add in-line version reference here
- consistent indentation

agilgur5 (commented):
some notes on in-line deprecation and versions

docs/cron-workflows.md
Joibel: Yes, I agree.
It is a trivial change to upgrade and totally automatable. We can then remove the old implementation sooner.

ui/src/app/cron-workflows/cron-workflow-list.tsx
```diff
@@ -222,9 +222,10 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
     );
 }
 
-function getCronNextScheduledTime(spec: CronWorkflowSpec): Date {
-    if (spec.schedule != '') {
-        return getNextScheduledTime(spec.schedule, spec.timezone);
+function getSpecNextScheduledTime(spec: CronWorkflowSpec): Date {
+    if (spec.schedule && spec.schedule.trim() !== '') {
+        const nextTime = getNextScheduledTime(spec.schedule, spec.timezone);
+        return nextTime;
```
agilgur5, line 228:
```suggestion
        return getNextScheduledTime(spec.schedule, spec.timezone);
```

unnecessary variable

agilgur5: I'm also not sure we should have the `trim()` check here and nowhere else. I don't even think the UI should handle that tbh, and I'm not sure that would be valid YAML either

agilgur5 (commented):
Thanks for updating this! Just a small comment below

ui/src/app/cron-workflows/cron-workflow-list.tsx
eduardodbr: sorry, I found an issue with next run time when using multiple schedules and I changed the code to debug. I forgot to revert. Should be good now

ui/src/app/cron-workflows/cron-workflow-list.tsx
```diff
@@ -222,8 +222,8 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
     );
 }
 
-function getCronNextScheduledTime(spec: CronWorkflowSpec): Date {
-    if (spec.schedule != '') {
+function getSpecNextScheduledTime(spec: CronWorkflowSpec): Date {
+    if (spec.schedule && spec.schedule !== '') {
```
agilgur5, line 226:
```suggestion
    if (spec.schedule) {
```

truthy covers the case of empty string already, so the `&& ..` is redundant

agilgur5 (approved):
LGTM, thanks for iterating on this!

agilgur5:
Looks like there's some test failures in CI

Issue #13356, opened 2024-07-17
3.5: Cannot list archived workflow in UI with `/archived-workflows/`


zhucan:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

expect to list archived workflow like version 3.4.9:
![image](https://github.com/user-attachments/assets/05917864-75ef-4648-b01d-f417936d9889)

but under version v3.5.8, can not list archived workflow:
![image](https://github.com/user-attachments/assets/edd366c2-baaf-482e-810e-ba9f7221f65b)

### Version(s)

v3.5.8

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
only create a normal wf, and waiting the wf was archived.
```

### Logs from the workflow controller

```text
no need to check the logs of controller.
```

### Logs from in your workflow's wait container

```text
no need to check the logs of wait container.
```

zhucan:
other question is, user curl cmd to get wf, can reponse successfully, but list wf with limit, hangs
Get on wf:
```
curl -H "Authorization: $ARGO_TOKEN" 10.245.10.132:2746/api/v1/workflows/argo-data-closeloop/prod--prod--perception-autolabel-detection3d-taging-1-1-36zv6b6
[...]

PR #13371, opened 2024-07-21
docs: add Archived Workflows merge to the Upgrading Notes

agilgur5:
Addresses #12545, https://github.com/argoproj/argo-workflows/discussions/13126#discussioncomment-9643345, #13356, etc

### Motivation

- The Archived Workflows UI merge (#11121) been asked about several times on issues, discussions, and Slack, so address it directly in the docs
  - previously, it was mentioned only in [the 3.5 blog post](https://blog.argoproj.io/whats-new-in-argo-workflows-v3-5-f260e8603ca6) and nowhere else, so I can understand some of the confusion as the blog is a bit of a secondary resource that is not the easiest to search either

### Modifications

- add a section in the 3.5 upgrading notes about this
  - explicitly say that it is not breaking as the API was not removed
    - (although one could say it is "UI breaking")
    
- also add a small mention about it in the UI's new version modal

### Verification

Docs and UI builds pass

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -17,7 +17,16 @@ Use `ARGO_ALLOWED_LINK_PROTOCOL` and `ARGO_BASE_HREF` instead.
 
 ## Upgrading to v3.5
 
-There are no known breaking changes in this release. Please file an issue if you encounter any unexpected problems after upgrading.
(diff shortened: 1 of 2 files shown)


Joibel (approved):
Seems like a sensible thing to call out.

PR #13372, opened 2024-07-22
fix(server): improve confusing login errors

### LINKED ISSUE #7921: Unnecessary error prior to logging in with SSO
## Summary

Once sso is configured an error will appear on the landing page, before login has occurred. This is not a valid error as the user just needs to login. However it effects the aesthetic of the landing page and can give off the impression that something is wrong with the Argo server.

Have a general landing page for login. Similar to ArgoCD

image

### LINKED ISSUE #12168: Failed to load version/info Error: {"code":16,"message":"token not valid.. with SSO setup
### Pre-requisites

Open a website for argo-workflows and not see

```
Failed to load version/info Error: {"code":16,"message":"token not valid. see https://argoproj.github.io/argo-workflows/faq/"}
```

before doing anything.

To be clear, SSO login works, but the error message appearing every time appears troublesome.

### Version

3.5.1 and latest

```YAML
1. Install argo-workflows
2. Use sso via sso argument using the community helm chart (i dont think it being a chart is relevant here)
3. Open the website

The steps above are what got me here, I suspect there is something else going on that is causing this though?
```

```text
time="2023-11-08T16:57:41.079Z" level=debug msg="received config map argo-events/argo-events-controller update"
time="2023-11-08T16:57:43.114Z" level=debug msg="received config map argo-events/argo-events-controller update"
time="2023-11-08T16:57:45.151Z" level=debug msg="received config map argo-events/argo-events-controller update"
time="2023-11-08T16:57:45.620Z" level=debug msg="Get leases 200"
time="2023-11-08T16:57:45.632Z" level=debug msg="Update leases 200"
time="2023-11-08T16:57:47.201Z" level=debug msg="received config map argo-events/argo-events-controller update"
time="2023-11-08T16:57:49.263Z" level=debug msg="received config map argo-events/argo-events-controller update"

adding them for completeness, i don't think anything there is applicable here sadly
```

```text
n/a
```

agilgur5:
Fixes #7921, Fixes #12070, Fixes #12168

### Motivation

`token not valid` errors are very confusing for non-operator end users. they also appear before you've even logged in or after you just logged out.

#### fix(ui): don't show auth error when no cookie

- per the in-line comment, this generally means you haven't logged in yet
  - unless you're in SSO auth mode, in which case, you would've had an earlier error
  - or in Server auth mode, in which case, you wouldn't get an error in the first place
  
- if it's coming from a failed login, instead of saying "Failed to load version/info", say "Login Failed", which is a lot more concrete and understandable at a glance (that doesn't require any knowledge of API endpoints or the code)

- <details><summary>also refactor some code</summary>

  - use `getCookie` func from the previously unused `cookie` util file
  - standardize the `login` page to use the `setCookie` func as well, which effectively duplicated the logic previously
    - name the `login` func as `setUser`, which is a lot less confusing than `user`
  - also make the "authorization" cookie name a shared constant
  - use async/await for all the auth per #11740 
  </details>

#### fix(server): be more specific about validation errors

- "token not valid" is not the most helpful thing, we can say exactly what's wrong with it
  - which we do for SSO auth errors already

- **NOTE**: this technically removes the fallback to `server` auth when using it in conjunction with another one
  - you generally don't want to use `server` with anything, so that can often be a misconfiguration
    - at my last job, we didn't realize our token was incorrectly formatted until we finally set-up SSO instead of `server` for the UI (previously we had `oauth2-proxy` in front)
      - because it was silently falling back to `server` when `client` had failed
  - I might change this as it's technically breaking as such, though I think in a good way -- may make that a separate breaking change potentially

### DIFF
--- a/server/auth/gatekeeper.go
+++ b/server/auth/gatekeeper.go
@@ -180,9 +180,10 @@ func (s gatekeeper) getClients(ctx context.Context, req interface{}) (*servertyp
 	valid := false
 	var mode Mode
 	var authorization string
+	var err error
 
 	for _, token := range authorizations {
-		mode, valid = s.Modes.GetMode(token)
+		mode, valid, err = s.Modes.GetMode(token)
 		// Stop checking after the first valid token
 		if valid {
 			authorization = token
@@ -190,7 +191,9 @@ func (s gatekeeper) getClients(ctx context.Context, req interface{}) (*servertyp
 		}
 	}
 	if !valid {
-		return nil, nil, status.Error(codes.Unauthenticated, "token not valid. see https://argo-workflows.readthedocs.io/en/latest/faq/")
+		// this could be just an improperly formatted token, or it could be indicative of a misconfiguration, which is common when setting up auth
+		log.WithError(err).Warn("token not valid. if you're setting up auth, you might have a misconfiguration: https://argo-workflows.readthedocs.io/en/latest/faq/")
+		return nil, nil, status.Error(codes.Unauthenticated, err.Error())
 	}
 	switch mode {
 	case Client:
--- a/server/auth/mode.go
+++ b/server/auth/mode.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/argoproj/argo-workflows/v3/server/auth/sso"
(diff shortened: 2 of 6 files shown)


Joibel:
@vlad-ivanov-name - please feel free to pick this one up and finish it. Please ensure you keep @agilgur5's patches as they are and add on top of them if you're using any of his work, and then the PR get's co-author credits correctly.

PR #13375, opened 2024-07-22
fix(devcontainer): `chown` regression for `make codegen`

Joibel:
https://github.com/argoproj/argo-workflows/pull/13348 removed a recursive flag from chown. I believe was done because this is slow on macintosh filesystems.

This [breaks codegen](https://cloud-native.slack.com/archives/C0510EUH90V/p1721637621193269) with
```
fatal: could not create leading directories of '/home/vscode/go/src/github.com/gogo/protobuf': Permission denied
```

This is a smaller recursive chown which fixes the issue for me.

### DIFF
--- a/.devcontainer/pre-build.sh
+++ b/.devcontainer/pre-build.sh
@@ -21,7 +21,8 @@ sudo apt install -y protobuf-compiler
 
 # Make sure go path is owned by vscode
 sudo chown vscode:vscode /home/vscode/go || true
-sudo chown -R vscode:vscode /home/vscode/go/src || true
+sudo chown vscode:vscode /home/vscode/go/src || true
+sudo chown vscode:vscode /home/vscode/go/src/github.com || true
 
 # download dependencies and do first-pass compile
 CI=1 kit pre-up
--- a/.devcontainer/pre-build.sh
+++ b/.devcontainer/pre-build.sh
@@ -21,6 +21,7 @@ sudo apt install -y protobuf-compiler
 
 # Make sure go path is owned by vscode
 sudo chown vscode:vscode /home/vscode/go || true
+sudo chown -R vscode:vscode /home/vscode/go/src || true
 
 # download dependencies and do first-pass compile
 CI=1 kit pre-up


qingfeng777:
It worked in my development environment.

agilgur5:
> It worked in my development environment.

The [fix works](https://cloud-native.slack.com/archives/C0510EUH90V/p1721662786271789?thread_ts=1721637621.193269&cid=C0510EUH90V), to clarify

agilgur5:
> This [breaks codegen](https://cloud-native.slack.com/archives/C0510EUH90V/p1721637621193269) with

Dang, can confirm it does this in my devcontainer too. This might be due to similar reasons as #13145 that codegen didn't run.

Thank you both for catching this!

> I believe was done because this is slow on macintosh filesystems.

Atrociously slow -- many minutes, I don't think I've ever once waited till the end, I have no idea how long it could take total. Since it worked fine even with me cancelling part-way, something about it isn't entirely necessary.
Although I'm not sure it's Mac FS specific per se, since the command is running _inside_ the container on files that only exist in the container, and has no affect on the host FS. But it clearly doesn't happen everywhere, so idk what the root cause is.

From a quick trace, it does seem like a lot of time is spent on the `.git` directory; even if the error is silenced from #13111, it still recurses through it. I don't think we need to recurse through, especially not the entire source dir, we just need to figure out which directories need to have `chown` applied

.devcontainer/pre-build.sh
```diff
@@ -21,6 +21,7 @@ sudo apt install -y protobuf-compiler
 
 # Make sure go path is owned by vscode
 sudo chown vscode:vscode /home/vscode/go || true
+sudo chown -R vscode:vscode /home/vscode/go/src || true
```
agilgur5, line 24:
```suggestion
sudo chown vscode:vscode /home/vscode/go/src || true
```

let me spin up a Codespace to make sure, but I think removing the `-R` worked fine for `make codegen`

agilgur5:
```suggestion
sudo chown vscode:vscode /home/vscode/go/src || true
sudo chown vscode:vscode /home/vscode/go/src/github.com || true
```

Had to add one more directory down in the Codespace. I'm not sure if only the last one is required or if all 3 are required. Might check that one more time

agilgur5: ok [looks like](https://github.com/agilgur5/argo-workflows/commit/2e771049af3a1921ded47234eae3e28e9b918ba4) all 3 dirs are necessary but no more than those 😅

agilgur5: I'm gonna make this change to unblock new devcontainers, seemed like you wouldn't mind per previous comments, but I [wasn't sure](https://github.com/argoproj/argo-workflows/pull/13270#issuecomment-2243999188)

agilgur5 (approved):

Joibel:
> > I believe was done because this is slow on macintosh filesystems.
> 
> Atrociously slow -- many minutes, I don't think I've ever once waited till the end, I have no idea how long it could take total. Since it worked fine even with me cancelling part-way, something about it isn't entirely necessary. Although I'm not sure it's Mac FS specific per se, since the command is running _inside_ the container on files that only exist in the container, and has no affect on the host FS. But it clearly doesn't happen everywhere, so idk what the root cause is.

I'm not saying it's mac specific, but either APFS or the docker desktop virtualisation through to APFS is causing slowdown. The writes are happening in the docker overlay, but they presumably have to thunk through to understand the underlying files.

On Linux with the source files on ext4 and using btrfs as the docker storage driver this takes less than 2 seconds. When I wrote this chown originally I wasn't using btrfs, just overlay2, and it was a much slower computer, and it was still unnoticable as a part of startup.

agilgur5:
> The writes are happening in the docker overlay, but they presumably have to thunk through to understand the underlying files.

Presumably, although something feels off that other FS commands aren't similarly slow

I'm curious if someone else has the same issue on Mac? Did Adrien have it too?

> From a quick trace, it does seem like a lot of time is spent on the `.git` directory

The other outlier is that my `.git` directory is probably larger than most, since I have lots of branches, stashes, a big reflog, a few remotes to other contributors etc.
Potentially an FS inefficiency combined with that causes the slowdown

Issue #13055, opened 2024-05-15
Allow skipping a workflow if mutex cannot be acquired


nettrino:
# Summary

If a mutex is at the workflow level, other workflows with the same mutex will wait till the mutex is acquired. However, there would be cases where we would want to skip a workflow entirely if another instance of the workflow has the mutex. This could be doable via some configuration option like the following:

```yaml
                synchronization:
                    mutex:
                        name: test
                  acquire-failure-policy: skip-workflow
``` 

We could expose different types of policies like `skip-workflow`, `wait`, etc.

## Use Cases

There are cases were a workflow runs to ingest / update some data set, for instance fetch new rows from a database for a given entity. Running two of the same workflows will have no meaning as they will fetch the exact same data, thus a mutex prevents this duplication, however currently it does not prevent duplicating work back-to-back without custom logic within the workflow executable. A policy that would determine what to do if a lock is not acquired would allow skipping the secondary (redundant) workflow entirely.

Beyond this issue, one could allow more complex blocking policies as well, where a mutex is considered to be "held" for longer periods of time than a workflow is running for. For instance, say I ingest data that get updated every 12h from some remote source. If a single one of the workflows has been run within that time window, we could force the mutex to be held for 12h, even if the workflow has been completed, so as no other workflows can run and do redundant work within that time period. This would be useful because, even if skip-workflow is accepted as a feature request, one could schedule redundant workflows to be running one after the other without lock contention.

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

Joibel:
As semaphores and mutexes can be used at a finer grained level than a whole Workflow, I'd suggest the policy should be just called `skip`, but with the implementation implied by `skipTemplate`.

To implement #12757 (which is mine, so I do have an interest here) you'd need a `steal` policy, which is problematic to implement for semaphores, as which one do you steal - I guess you could have `stealOldest` or `stealNewest`. We could just disallow steal on semaphores, and only allow it on mutex. Do we implement `stealWithStop` separately from `stealWithTerminate`?

It's somewhat preferable to implement this mechanism over #12757's proposal because it's more powerful. It's also much less intuitive for the simple case that that issue is aiming for and requires more typing.

nettrino:
> I'd suggest the policy should be just called `skip`

+1 on the above definitely cleaner. 

Re:  
[...]

PR #13393, opened 2024-07-25
docs: synchronization and paralellism docs improvements

### LINKED ISSUE #12437: Add workflow prioritization docs
# Summary

Add workflow priority docs

## Use Cases

It seems from the workflow spec that you can assign a priority to a workflow, but there are no docs about it or an example.

---

Joibel:
Split from #13358
Fixes #12437 

This is a rewrite of the docs for synchronization and parallelism, split from the multiple mutexes and semaphores code. This PR is suitable for backporting to 3.5 and 3.4.

* New markdown document on parallelism for the operator guide:
  - locks and parallelism will generally have different use cases. I think it is better not to have them on the same page so you don't think you could use one instead of the other
  - as far as I know more pages doesn't cost us anything.
  - priority for parallelism documented

* Synchronisation adds:
  - information on queuing
  - better explain what a mutex is
  - explain behaviour in case of specifying both semaphore and mutex
  - paralellism in workflow and template spec better explained
  - add example links

Conform to style guide and 1 sentence per line


docs/parallelism.md
```diff
@@ -0,0 +1,44 @@
+# Limiting parallelism
+
+You can restrict the number of workflows being executed at any time using these mechanisms.
```
agilgur5, line 3:
```suggestion
You can restrict the number of workflow executions.
```

- simplify

docs/parallelism.md
```diff
@@ -0,0 +1,44 @@
...
+You can restrict parallelism within a workflow using `parallelism` within a workflow or template.
+This only restricts total concurrent executions of steps or tasks within the same workflow.
+
+Examples
+
+1 [`parallelism-limit.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-limit.yaml) restricts the parallelism of a [loop](./walk-through/loops.md)
+1 [`parallelism-nested.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-nested.yaml) restricts the parallelism of a nested loop
+1 [`parallelism-nested-dag.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-nested-dag.yaml) restricts the number of dag tasks that can be run at any one time
+1 [`parallelism-nested-workflow.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-nested-workflow.yaml) shows how parallelism is inherited by children
+1 [`parallelism-template-limit.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-template-limit.yaml) shows how parallelism of looped templates is also restricted
```
agilgur5, line 40:
```suggestion
1. [`parallelism-limit.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-limit.yaml) restricts the parallelism of a [loop](./walk-through/loops.md)
1. [`parallelism-nested.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-nested.yaml) restricts the parallelism of a nested loop
1. [`parallelism-nested-dag.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-nested-dag.yaml) restricts the number of dag tasks that can be run at any one time
1. [`parallelism-nested-workflow.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-nested-workflow.yaml) shows how parallelism is inherited by children
1. [`parallelism-template-limit.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/parallelism-template-limit.yaml) shows how parallelism of looped templates is also restricted
```

Dots in front of the `1` were missing, which caused invalid formatting

docs/parallelism.md
```diff
@@ -0,0 +1,44 @@
...
+Workflows can have a `priority` set in their specification.
+
+Workflows with a higher priority number that have not started due to controller level parallelism will be started before lower priority workflows.
+
+## Workflow level
+
+You can restrict parallelism within a workflow using `parallelism` within a workflow or template.
+This only restricts total concurrent executions of steps or tasks within the same workflow.
+
+Examples
```
agilgur5, line 34:
```suggestion
Examples:
```

missing colon

docs/parallelism.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+You can also limit the number of workflows that can execute in a single namespace.
+
+```yaml
+data:
+  namespaceParallelism: "4"
+```
+
+Workflows that are executing but restricted from running more nodes due to the other mechanisms will still count towards the parallelism limits.
```
agilgur5, line 21: hmmm there is a WIP PR to change this: #12799. if we consider that as "buggy" behavior (as the issue and PR do), then we probably shouldn't document it as it is unintended

.spelling
```diff
@@ -183,6 +183,7 @@ memoizing
 metadata
 minikube
 mutex
+mutexes
```
agilgur5, line 186: For reference, "mutexes" appears to be preferred over "mutices": https://english.stackexchange.com/q/37455/527544

docs/synchronization.md
```diff
@@ -4,13 +4,12 @@
 
 ## Introduction
```
agilgur5, line 6:
```suggestion
```

The "Introduction" heading is redundant as well; I've removed it in some places already (#13110 and I think some others too)

docs/synchronization.md
```diff
@@ -147,8 +152,16 @@ Examples:
...
+The queue is then ordered by `CreationTimestamp` of the Workflow; older Workflows will be ordered before newer workflows.
+
+Workflows are only be allowed to take a lock if they are at the front of the queue for that lock.
+
+## Parallelism
 
-In addition to this synchronization, the workflow controller supports a parallelism setting that applies to all workflows
-in the system (it is not granular to a class of workflows, or tasks withing them). Furthermore, there is a parallelism setting
-at the workflow and template level, but this only restricts total concurrent executions of tasks within the same workflow.
+See also [how you can restrict parallelism](./parallelism.md) in other ways.
```
agilgur5, line 167:
```suggestion
You can also [restrict parallelism in other ways](parallelism.md).
```

- simplify
- remove `./` as it's normally not used in the docs (the few places it is are outliers)

docs/synchronization.md
```diff
@@ -147,8 +152,16 @@ Examples:
...
+
+When a Workflow cannot take a lock it will be placed into a ordered queue.
+
+Workflows can have a `priority` set in their specification.
+The queue is first ordered by priority, with a higher priority number being placed before a lower priority number.
+The queue is then ordered by `CreationTimestamp` of the Workflow; older Workflows will be ordered before newer workflows.
+
+Workflows are only be allowed to take a lock if they are at the front of the queue for that lock.
+
+## Parallelism
```
agilgur5, line 165:
```suggestion
## Other Parallelism support
```

I'd actually keep the heading as-is in case it was linked to previously -- that way we don't break old links. It's not a substantially different heading, so I'd prefer non-breakage over simplification in this case

agilgur5 (commented):
Quick first-pass, will need to spend a bit more time on reviewing it later

agilgur5:
> * as far as I know more pages doesn't cost us anything.

it doesn't per se. it's more about which is better UX. Too many pages can be hard to navigate. Equally too much on one page can be hard to navigate. If we have a good reason to split it, I think that's mostly what matters

docs/synchronization.md
Joibel: Ok, good call

docs/parallelism.md
Joibel: I don't only mean that mechanism. Suspended workflows will also do this.

I think the original intention would have been not to do what #12799 is trying to fix, because it's simpler. (The current fix there violates parallelism. An aware user could deliberately violate parallelism using it to get their work done.)

I'd like the current documentation to document current behaviour, so I'd like to keep this in.

docs/parallelism.md
Joibel: I think the extra words here are important.
"Restrict the number of workflow executions." is not enough detail. Total number of executions ever?

I'll go with "You can restrict the number of simultaneous workflow executions."

agilgur5: That was the exact word I was thinking to add 🙂👍

agilgur5:
> * locks and parallelism will generally have different use cases.

Is this entirely accurate? You might have seen more use-cases than me, but a lot of users do in fact swap parallelism and semaphores. If I'm not mistaken, I believe semaphores are a strict superset of parallelism in terms of functionality (i.e. semaphores can do everything parallelism can do and more).

docs/parallelism.md
agilgur5: Yea I think there's a few other cases that could all be considered "buggy". The primary difference though, is if it's documented, people won't file bug reports or fix it necessarily -- it's expected behavior. If we link to an existing bug report, or otherwise state that this behavior may be optimized in the future, I think that could be ok as a compromise.

Regarding #12799, please do review there if you have an opinion on the approach! The initial approach was to requeue, which was a lot simpler, the current one is a bit different and I honestly have yet to think through correctness on it

docs/synchronization.md
```diff
@@ -52,7 +53,7 @@ spec:
       args: ["hello world"]
 ```
 
-Using a mutex:
+Using a mutex achieves the same thing as a count `"1"` semaphore:
```
blkperl, line 56: Can we add an example where the `mutex` name is dynamic?

For example, a workflow or multiple workflows that interact with the same external resource such as a database would want to prevent copy/backup/restore workflows from occurring at the same time.

```
mutex:
   name: "database-lock-{{inputs.parameters.dbName}}"
```

- database-backup workflow A  - locks database A
- database-copy workflow B - waits for database-backup workflow A to release the mutex

blkperl (commented):
Thanks for improving these docs! A few thoughts below:

Is there any impact from a controller restart (voluntary or involuntary) that impacts parallelism or synchronization that we should document?

I assume synchronization is not included in activeDeadlineSeconds and that a workflow will wait forever for a lock. Have we considered a mutex timeout that will cause the workflow to fail if they fail to lock within the time period? Edit: Found https://github.com/argoproj/argo-workflows/issues/13055 which seems to be similar. 

Are there any other pitfalls to document?

docs/synchronization.md
Joibel: We could. It's outside the scope of what I was trying to achieve in this PR, and would prefer someone did that in a separate PR.

Joibel:
> Thanks for improving these docs! A few thoughts below:
> 
> Is there any impact from a controller restart (voluntary or involuntary) that impacts parallelism or synchronization that we should document?

There shouldn't be. #8684 documents a case where things don't behave as desired. The controller should be indexing everything and rebuilding the in-memory state before running.

> I assume synchronization is not included in activeDeadlineSeconds and that a workflow will wait forever for a lock. Have we considered a mutex timeout that will cause the workflow to fail if they fail to lock within the time period? Edit: Found #13055 which seems to be similar.

I'm not sure why you'd assume synchronization is not included in `activeDeadlineSeconds`? I'm not sure what you mean by this exactly, but I would expect a workflow that hits `activeDeadlineSeconds` to be stopped. This would free any held locks and return any that it held.

#13055 and timeouts on mutexes are out of scope of documenting what we already have so won't be included in this PR.

> Are there any other pitfalls to document?

I've documented those I'm aware of.

agilgur5:
Yea I generally agree on cluster-level parallelism. Workflow/template-level parallelism can blur the lines a lot, especially as that is user controllable just like semaphores.

I'm wondering if maybe we split out cluster-level parallelism into the Operator Guide instead of an adjacent page in the User Guide. That would definitely keep the intents aligned. We can still leave the links as-is as well.

Workflow/template-level parallelism could maybe remain in the Synchronization page? Then it segues nicely into the link to the Parallelism in the Operator Guide too

Joibel:
> I'm wondering if maybe we split out cluster-level parallelism into the Operator Guide instead of an adjacent page in the User Guide. That would definitely keep the intents aligned. We can still leave the links as-is as well.
> 
> Workflow/template-level parallelism could maybe remain in the Synchronization page? Then it segues nicely into the link to the Parallelism page in the Operator Guide too

I like this, and I've done it.

Joibel:
@agilgur5, any chance of a re-review on this one please?

agilgur5:
Was working on it last night, should be done soon

docs/parallelism.md
```diff
@@ -0,0 +1,31 @@
...
+## Controller level
+
+You can limit the total number of workflows that can execute at any one time in the [workflow controller ConfigMap](./workflow-controller-configmap.yaml).
+
+```yaml
+data:
+  parallelism: "10"
+```
+
+You can also limit the number of workflows that can execute in a single namespace.
```
agilgur5, line 14:
```suggestion
You can also limit the total number of parallel workflow executions in a single namespace:
```
same as above

docs/parallelism.md
```diff
@@ -0,0 +1,31 @@
...
+```yaml
+data:
+  namespaceParallelism: "4"
+```
+
+Workflows that are executing but restricted from running more nodes due to the other mechanisms will still count towards the parallelism limits.
+
+### Priority
+
+Workflows can have a `priority` set in their specification.
```
agilgur5, line 25:
```suggestion
You can set a `priority` on Workflows.
```

active voice, same as above

docs/parallelism.md
```diff
@@ -0,0 +1,31 @@
...
+You can also limit the number of workflows that can execute in a single namespace.
+
+```yaml
+data:
+  namespaceParallelism: "4"
+```
+
+Workflows that are executing but restricted from running more nodes due to the other mechanisms will still count towards the parallelism limits.
+
+### Priority
```
agilgur5, line 23: Should this be called "Queueing" to match the "Synchronization" page?

```suggestion
### Queueing
```

agilgur5: I also wonder if we should just link to the "Synchronization" page, as, if I'm not mistaken, it's effectively identical here

docs/parallelism.md
```diff
@@ -0,0 +1,31 @@
...
+
+Workflows can have a `priority` set in their specification.
+
+Workflows with a higher priority number that have not started due to controller level parallelism will be started before lower priority workflows.
```
agilgur5, line 27:
```suggestion
Workflows that have not started yet will be queued: Workflows with higher priority numbers will start before lower priority ones.
```

simplify.

although this feels identical to the "Queuing" section in the "Synchornization" page as I mentioned above

docs/parallelism.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Synchronization
+
+You can use [mutexes, semaphores and parallelism](./synchronization.md) to control the parallel execution of sections of a workflow.
```
agilgur5, line 31:
```suggestion
You can also use [mutexes, semaphores, and parallelism](synchronization.md) to control the parallel execution of workflows and templates.
```

- Oxford comma pl0x
- remove `./` in link, same as above
- it's not just "sections of a workflow", since it is a superset, so I added "also" and used "workflows or templates" as that page does

agilgur5 (commented):
Left a pretty sizeable number of changes below, mostly on internal consistency and some simplifications

docs/synchronization.md
```diff
@@ -147,8 +151,29 @@ Examples:
 1. [Step level semaphore](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-tmpl-level.yaml)
 1. [Step level mutex](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level.yaml)
 
-### Other Parallelism support
+## Queuing
+
+When a Workflow cannot take a lock it will be placed into a ordered queue.
+
+Workflows can have a `priority` set in their specification.
```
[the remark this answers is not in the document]
Joibel: I don't like this as is because it doesn't lead a user as to how to do this. I'll work out some new wording.

docs/parallelism.md
Joibel: Same problem as above.

docs/parallelism.md
Joibel: The queue is an implementation detail from the users perspective. All they care about is that priority exists and they can use it to prioritise. They don't need to know this is implemented by queueing.

docs/parallelism.md
Joibel: This is a different queue, unrelated to any synchronization based queues.

You need to pass through this queue before you can enter any synchronization based queue, and this is important to understand.

In the Synchronization section the queue is about "What happens when a mutex isn't available". At the moment you get no option, you enter a queue. We may in future have alternative options to skip this section of the workflow, or wait in the queue for some limited time.

It's probable that people can guess they would enter a queue, and here the details they cannot guess at are the ordering of the queue. It's not a FIFO as you may guess as a simple implementation, it's priority then workflow creation time.

docs/synchronization.md
```diff
@@ -24,11 +22,15 @@ data:
...
+Each synchronization block may only refer to either a semaphore or a mutex.
+If you specify both only the semaphore will be locked.
+
+## Workflow-level Synchronization
 
-Workflow-level synchronization limits parallel execution of the workflow if workflows have the same synchronization reference.
-In this example, Workflow refers to `workflow` synchronization key which is configured as limit 1,
-so only one workflow instance will be executed at given time even multiple workflows created.
+You can limit parallel execution of a workflow by using Workflow-level synchronization.
+If multiple workflows have the same synchronization reference they will be limited by that synchronization reference.
```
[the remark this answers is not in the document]
Joibel: These were deliberately separated to allow the user to understand better. Your rewrite loses the first sentence's content as a separately understandable item.

docs/synchronization.md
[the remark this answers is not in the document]
Joibel: Linked to priority in parallelism.md, which now describes this, mostly using your wording.

docs/parallelism.md
Joibel: I've added a description of how to use this, and the missing information of default priority.

docs/parallelism.md
Joibel: Your rewrite here doesn't explain why workflows may not be started, so I've adopted some of your rewrite.

docs/synchronization.md
```diff
@@ -147,8 +151,29 @@ Examples:
...
+Workflows can have a `priority` set in their specification.
+The queue is first ordered by priority, with a higher priority number being placed before a lower priority number.
+The queue is then ordered by `CreationTimestamp` of the Workflow; older Workflows will be ordered before newer workflows.
+
+Workflows are only be allowed to take a lock if they are at the front of the queue for that lock.
+
+## Workflow level parallelism
+
+You can restrict parallelism within a workflow using `parallelism` within a workflow or template.
+This only restricts total concurrent executions of steps or tasks within the same workflow.
```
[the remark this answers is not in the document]
agilgur5: Yea I was thinking about that too, but I'm not sure that's something a user would assume changes in a Workflow vs WorkflowTemplate -- generally, it shouldn't, as that's how templates work in Argo.
Beginners may think that initially, but once you write enough workflows, you realize that's an overcomplication and there's no special case logic like that and nowhere do the docs suggest that either. Saying this in a separate sentence may also make you think there are other cases that are different (there aren't).

My alternative suggestion in that case was:
```suggestion
You can use `parallelism` within a workflow or template to restrict the total concurrent executions of steps or tasks within the same workflow.
```
But that reads a bit repetitively.

There's also no mention of WorkflowTemplate in any of the variations, so if you substitute the word "workflow" with "workflowtemplate", you may very well _still_ think it applies to all child workflows.
So I actually don't like this one or the original. The first suggestion I made makes no implication whatsoever as such.

If anything, it might be better to use the first suggestion plus a `!!! Note`, but that kind of note seems better in the WorkflowTemplate docs actually

agilgur5:
Oh looks like you made some changes while I was mid-review. No IRL synchronization 😅

Joibel:
> Oh looks like you made some changes while I was mid-review. No IRL synchronization 😅

Sorry, was providing updates to the bits I'd had left undone here - in between real work.

docs/synchronization.md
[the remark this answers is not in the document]
Joibel: I'm still not preferring any of this, I think my version may be pendantic but it is unambiguous.
* I cannot find a way to misinterpret it as "there being other cases".
* It would be clear to a beginner.
* I want the documentation for a field to be all in one place, not needing to discover this field has more documentation in the WFT page.

docs/parallelism.md
agilgur5: I made that comment before your change to link them, so let me re-read the new variant

docs/synchronization.md
```diff
@@ -147,8 +152,29 @@ Examples:
...
 1. [Step level mutex](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level.yaml)
 
-### Other Parallelism support
+## Queuing
+
+When a Workflow cannot acquire a lock it will be placed into a ordered queue.
+
+You can set a [`priority`](parallelism.md#priority) on Workflows.
+The queue is first ordered by priority: a higher priority number is placed before a lower priority number.
+The queue is then ordered by `creationTimestamp`: older Workflows are placed before newer workflows.
```
agilgur5, line 161:
```suggestion
When a workflow cannot acquire a lock it will be placed into a ordered queue.

You can set a [`priority`](parallelism.md#priority) on workflows.
The queue is first ordered by priority: a higher priority number is placed before a lower priority number.
The queue is then ordered by `creationTimestamp`: older workflows are placed before newer workflows.
```

lower-case "workflows" for consistency

agilgur5 (commented):
I like this iteration, only some small comments mostly on style/consistency

agilgur5:
I made the [two](https://github.com/argoproj/argo-workflows/pull/13393#discussion_r1739804849) [remaining](https://github.com/argoproj/argo-workflows/pull/13393#discussion_r1739805932) changes myself as one is broken syntax and the other is purely formatting. Since you set auto-merge I assume that's fine

docs/parallelism.md
```diff
@@ -0,0 +1,43 @@
...
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: priority-
+spec:
+  priority: 3
+  ...
+````
```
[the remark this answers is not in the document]
agilgur5: Fixed that myself in [this commit](https://github.com/argoproj/argo-workflows/pull/13393/commits/6e78e889344df95bdb9b1bdc79e753df9ecb7df1)

agilgur5 (approved):
LGTM, thanks again for substantially improving this and iterating on it!

PR #13404, opened 2024-07-26
build: resolve error for toolchain not available

terrytangyuan:
This fixes the following issue in my local setup:
```
go: downloading go1.22 (darwin/arm64)
go: download go1.22 for darwin/arm64: toolchain not available
```

See https://github.com/golang/go/issues/65568

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/argoproj/argo-workflows/v3
 
-go 1.22.0
+go 1.22.5
 
 require (
 	cloud.google.com/go/storage v1.36.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/argoproj/argo-workflows/v3
 
-go 1.22
+go 1.22.0
 
 require (
 	cloud.google.com/go/storage v1.36.0


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/argoproj/argo-workflows/v3
 
-go 1.22
+go 1.22.0
```
agilgur5, line 3:
```suggestion
go 1.22.5
```

1.22.5 looks like it's latest: https://github.com/golang/go/releases/tag/go1.22.5

Does that work?

terrytangyuan: updated

agilgur5 (approved):

isubasinghe:
This broke `gopls` for me, is anyone having the same issue?

Joibel:
`gopls` is working fine for me, when using the version I have installed via nixos's home-manager from nixos 24.05.

isubasinghe:
Yeah I was using the version on the snap store.

Using PopOS at the moment, wifi drivers on NixOS have been not so fun.

Not sure why but some weird DNS resolving issue on NixOS when on wifi, which is quite bizarre.

isubasinghe:
Solution is just to install it via `go install` to grab the latest one.

Issue #13405, opened 2024-07-26
High severity vulnerability found in `jackc/pgx/v4`


terrytangyuan:
# Summary

```
✗ High severity vulnerability found in github.com/jackc/pgx/v4
  Description: Uncaught Exception
  Info: https://security.snyk.io/vuln/SNYK-GOLANG-GITHUBCOMJACKCPGXV4-7416900
  Introduced through: github.com/upper/db/v4/adapter/postgresql@4.7.0
  From: github.com/upper/db/v4/adapter/postgresql@4.7.0 > github.com/jackc/pgx/v4/stdlib@4.18.2 > github.com/jackc/pgx/v4@4.18.2

```
---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
@terrytangyuan this looks like a false positive, I mentioned it [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1721421729011109). Reported it to Snyk per the thread, but they have yet to respond 😕

terrytangyuan:
Yep I saw that thread. We should still track it on GitHub.

terrytangyuan:
Disabling this one in https://github.com/argoproj/argo-workflows/pull/13406

agilgur5:
Snyk still hasn't responded to my ticket/email (almost 2 weeks now), but I double-checked and it appears they have revoked the advisory in their DB now: https://security.snyk.io/vuln/SNYK-GOLANG-GITHUBCOMJACKCPGXV4-7416900

agilgur5:
Got a response from Snyk on Aug 2nd saying they opened an investigation with their Security Team. Got a further response today saying it was actually a false positive upstream too, c.f. https://github.com/golang/vulndb/issues/2966.
Not clear why Snyk was behind that long though, I did ask for further details on that.

Should be all good to remove the ignore now though with multiple confirmations now

terrytangyuan:
Cool I'll submit a PR to move the ignore

PR #13406, opened 2024-07-27
ci: temporarily ignore false positive Snyk vulnerability for `jackc/pgx/v4`

terrytangyuan:
Fixes https://github.com/argoproj/argo-workflows/issues/13405.

Reference doc: https://docs.snyk.io/manage-risk/policies/the-.snyk-file#use-the-.snyk-file-with-snyk-open-source

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,7 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities
+version: v1.25.0
+ignore:
+  SNYK-GOLANG-GITHUBCOMJACKCPGXV4-7416900:
+    - '*':
+        reason: False Positive
+        expires: 2024-09-01T11:11:11.001Z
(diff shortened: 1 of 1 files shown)


.snyk
```diff
@@ -0,0 +1,7 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities
+version: v1.25.0
+ignore:
+  SNYK-GOLANG-GITHUBCOMJACKCPGXV4-7416900:
+    - '*':
+        reason: False Positive
```
agilgur5, line 6: we should definitely detail this more, it's good practice to do so.

```suggestion
        reason: "False Positive: vuln is for pgx/v5, not pgx/v4: https://pkg.go.dev/vuln/GO-2024-2567"
```

agilgur5 (approved):

agilgur5:
So in this case, auto-merged was enabled prior to review... That ended up making the merged commit use the original title and description; notably missing me as a co-author from my more detailed `reason` suggestion [above](https://github.com/argoproj/argo-workflows/pull/13406#discussion_r1693974509)

I'm thinking it makes more sense for a reviewer to enable auto-merge and not the author. Generally the intent for auto-merge is to not need to come back when tests complete, so I think it makes sense to only enable post-review/post-approval

PR #13408, opened 2024-07-27
docs: add example of CI with ephemeral VMs with Azure CLI + Ansible

jrbe228:
docs: Azure Infra Example

### Motivation

Example for automating software builds / tests on ephemeral Windows VMs. Very helpful for me, could be for others also. Feel free to edit or close PR, as appropriate. 

Use Azure CLI to 
- Create VM 
- Perform Installs via Ansible 
- Other tests / builds, if applicable
- Delete VM

### Modifications

No changes to Argo engine or UI

### Verification

Works for me :) 
image


agilgur5 (commented):
Thanks for providing an example for others!

The examples here tend to be more generic in nature (although not all). I'm wondering if this might be better placed in https://github.com/argoproj-labs/argo-workflows-catalog? @terrytangyuan what do you think?

jrbe228:
This PR was intended to show ephemeral VMs could be used in a workflow, even if still very far from a "first-class citizen". I chose the Azure CLI for convenience. Now I'm wondering if https://kubevirt.io/ is a better concept. No testing yet on my end. Just reading docs.

terrytangyuan (commented):
Thanks for the PR. We probably have similar comments as in https://github.com/argoproj/argo-workflows/pull/13895

PR #13419, opened 2024-07-31
feat(controller): optimize memory with queue when archiving is rate-limited. Fixes #13418

### LINKED ISSUE #13418: Optimize workflow-controller memory usage when archiving workflow operations are rate-limited
# Summary

When archiving workflows operations are severely rate-limited，there is a big memory increase in workflow controller, as shown is below grafana metrics:
9cea4862af9774db0b852e4c66106dd2

The following is a comparison between when there is no rate limit and when there is rate limited:
89d94bd9df2c520c8857ce5181bb6959

Found by analyzing client-go code，there is a growing ring buffer to store all the archiving workflow when the archiving operation is blocked by rate-limiter:
- shared_informer.go:
image
image
- ring_growing.go:
image
image

To solve this problem, we can add a archived_wf_queue to archive the workflow asynchronously when listening the Add and Update events instead of archiving the workflow synchronously, like this:
image

Here is the comparison before I make the change and after I make the change:
a230bda903e593942ecad272784bc444

## Use Cases

When archiving workflows operations are rate-limited

ChenRussell:
Fix #13418 

### Motivation

When archiving workflows operations are severely rate-limited，there is a big memory usage increase in workflow controller

### Modifications

To solve this problem, we can add a archived_wf_queue to archive the workflow asynchronously when listening the Add and Update events instead of archiving the workflow synchronously

### Verification

run locally and tested


cmd/workflow-controller/main.go
```diff
@@ -58,6 +58,7 @@ func NewRootCommand() *cobra.Command {
 		workflowTTLWorkers      int    // --workflow-ttl-workers
 		podCleanupWorkers       int    // --pod-cleanup-workers
 		cronWorkflowWorkers     int    // --cron-workflow-workers
+		archivedWorkflowWorkers int    // --archived-workflow-workers
```
agilgur5, line 61: Mmm adding a new flag certainly sounds like a feature to me, not a fix

workflow/controller/controller.go
```diff
@@ -789,6 +795,32 @@ func (wfc *WorkflowController) runWorker() {
 	}
 }
 
+func (wfc *WorkflowController) runArchiveWorker() {
+	ctx := context.TODO()
```
agilgur5, line 799: This seems like it should either pass through the existing context, as the other functions do, or be identical to `runWorker` above, which also has a panic/crash handler

workflow/controller/controller.go
```diff
@@ -789,6 +795,32 @@ func (wfc *WorkflowController) runWorker() {
...
+	for wfc.processNextArchiveItem(ctx) {
+	}
+}
+
+func (wfc *WorkflowController) processNextArchiveItem(ctx context.Context) bool {
```
agilgur5, line 804: This should be below `processNextItem`, matching the ordering of `runWorker` and `runArchiveWorker`

agilgur5 (commented):
Thanks for root causing the memory increase and optimizing it!

I think we'd want to have a few people review this since it adds a new queue (with its own metrics, for instance) and a new flag.
There's also a consideration around naming those two, especially since there are potentially multiple different archive workers (e.g. GC'ing archived workflows is another one)

cmd/workflow-controller/main.go
ChenRussell: I change the commit message to a feature

agilgur5: The PR title is the biggest part, so I updated that as well. Thanks

agilgur5 (approved):
Makes sense to me. @Joibel could you double-check this when you're back from vacation? Specifically as this adds a new queue with metrics. Also if you have any thoughts on the naming

workflow/controller/controller.go
```diff
@@ -1081,10 +1115,16 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context)
 		},
 		Handler: cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
-				wfc.archiveWorkflow(ctx, obj)
+				key, err := cache.MetaNamespaceKeyFunc(obj)
+				if err == nil {
+					wfc.archiveWfQueue.Add(key)
```
fyp711, line 1120: LGTM! But why call the `wfc.archiveWfQueue.Add` method instead of the `wfc.archiveWfQueue.AddRateLimited` method?

ChenRussell: I think it's not necessary to call AddRateLimited，the workflow_ttl_queue doesn't call AddRateLimited either.

fyp711 (approved):

ChenRussell:
> Makes sense to me. @Joibel could you double-check this when you're back from vacation? Specifically as this adds a new queue with metrics. Also if you have any thoughts on the naming

@Joibel  could you review this PR when you are free, thanks

Joibel:
> @Joibel could you review this PR when you are free, thanks

Yes, I'll take a look.

workflow/controller/controller.go
```diff
@@ -225,6 +226,7 @@ func NewWorkflowController(ctx context.Context, restConfig *rest.Config, kubecli
 	wfc.wfQueue = wfc.metrics.RateLimiterWithBusyWorkers(&fixedItemIntervalRateLimiter{}, "workflow_queue")
 	wfc.throttler = wfc.newThrottler()
 	wfc.podCleanupQueue = wfc.metrics.RateLimiterWithBusyWorkers(workqueue.DefaultControllerRateLimiter(), "pod_cleanup_queue")
+	wfc.archiveWfQueue = wfc.metrics.RateLimiterWithBusyWorkers(workqueue.DefaultControllerRateLimiter(), "archived_wf_queue")
```
Joibel, line 229: I would prefer to call this the `workflow_archive_queue`. Certainly not `archived` with a d, as these are not yet archived.

workflow/controller/controller.go
Joibel: It is not necessary, but it does mean that this is not rate limited. It feels like this is a secondary function of the controller and would be better if it was rate limited. Is there a reason you choose not to rate limit it?

workflow/controller/controller.go
```diff
@@ -354,6 +357,9 @@ func (wfc *WorkflowController) Run(ctx context.Context, wfWorkers, workflowTTLWo
 	for i := 0; i < wfWorkers; i++ {
 		go wait.Until(wfc.runWorker, time.Second, ctx.Done())
 	}
+	for i := 0; i < archivedWorkflowWorkers; i++ {
```
Joibel, line 360: Overall I would prefer it if this was separated out into it's own controller, modelled on the [garbage collector](https://github.com/argoproj/argo-workflows/blob/2d7e2b55914a89653613f75d2d8b19ccb9c01c39/workflow/controller/controller.go#L347). I know podGC is still wrapped up into this controller, but that too feels like it could be separated.

They could still share an informer, but the concerns would be better separated.

I realise this isn't all the fault of this specific PR, but I'd like it to be done if possible.

Joibel (changes requested):

workflow/controller/controller.go
agilgur5: agree, separated controllers/files are a lot easier to read. some of them being here and others being separated is confusing

workflow/controller/controller.go
agilgur5: > Is there a reason you choose not to rate limit it?

my read was that it was primarily a copy per above, and not for a particular reason

> the workflow_ttl_queue doesn't call AddRateLimited either.

ChenRussell: > It is not necessary, but it does mean that this is not rate limited. It feels like this is a secondary function of the controller and would be better if it was rate limited. Is there a reason you choose not to rate limit it?

the main reason is I want to keep the same with workflow_ttl_queue, because that two queues are similar, both of them are dealing with completed workflow, and that is not very frequent action IMO.

workflow/controller/controller.go
ChenRussell: agree!

workflow/controller/controller.go
ChenRussell: I agree to submit a separate refactor PR, since there is a lot to change in workflow controller, and that is not specific to this PR

Joibel: Thank you

Joibel (approved):

ChenRussell:
@Joibel  all checks have passed now, could you merge this PR? thanks

Joibel:
Thank you for working through this.

agilgur5:
@ChenRussell small reminder on the refactor [discussed above](https://github.com/argoproj/argo-workflows/pull/13419#discussion_r1717846706)

ChenRussell:
> @ChenRussell small reminder on the refactor [discussed above](https://github.com/argoproj/argo-workflows/pull/13419#discussion_r1717846706)

Yes, I will finish it

workflow/controller/controller.go
agilgur5: For reference, see #13536 for this separated change

PR #13425, opened 2024-08-01
docs(artifacts): add note about server access

reilly3000:
### Motivation

When using Workload Identity the calling Kubernetes service account assumes GKE workload identity if it is annotated to do so, but per current docs it isn't clear that there are at least two service accounts that will need to be able to call the bucket. I am raising this change to help others avoid finding 500 errors when trying to retrieve archived logs for Workflows.

### Modifications

Added docs content to explain that argo-server needs Workload identity enabled for use with Archive Logs

### Verification

n/a

### DIFF
--- a/docs/configure-artifact-repository.md
+++ b/docs/configure-artifact-repository.md
@@ -182,6 +182,9 @@ create the service account key and store it as a Kubernetes secret,
 `serviceAccountKeySecret` is also not needed in this case. Please follow the
 link to configure Workload Identity
 (<https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity>).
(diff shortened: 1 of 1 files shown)


agilgur5 (commented):
This doesn't seem specific to archive logs, rather artifacts in general. If you want to view them in the UI, the Server needs to be able to read them, yes

agilgur5:
The [S3 IRSA section](https://github.com/argoproj/argo-workflows/blob/b59d7c1cd3c0191cb9a2c6ec91af47da2f21520e/docs/configure-artifact-repository.md?plain=1#L135) mentions the Server too. I would say to either follow that for now or make a new section about it properly. This page definitely needs a bit of a rewrite though (I started one many months ago actually, but the page is quite long so I never quite finished).

PR #12527, opened 2024-01-16
feat: Add proxy support to Argo CLI API client

shimako55:
Fixes #10794

### Motivation

**Issue**:
In the current version of Argo Workflows, the API client does not function properly in environments that use a proxy server.

**Purpose**:
This PR aims to modify the Argo Workflows API client to ensure it functions correctly when routing through a proxy server.

### Modifications

**Summary of Changes**:
Made modifications in `cmd/argo/commands/client/conn.go`, `pkg/apiclient/apiclient.go`, `pkg/apiclient/http1-client.go`, and `pkg/apiclient/http1/facade.go`.

**Specific Changes**:
Added proxy functionality using `http.ProxyURL`.
Integrated new proxy settings into the API client construction process.
Added proxy settings to the configuration of `http.Client` and `http.Transport`.

**Reason for Changes**:
To support the operation of the API client in proxy environments.

### Verification

**Testing Method**:
Conducted manual tests in an environment with a configured proxy server.

**Test Results**:
Confirmed that the API client operates correctly when routing through a proxy.

**Reproduction Steps**:
Replace http://example.com:8080 with your actual proxy server URL in the following commands:
```
$ ./argo template list --proxy-url=http://example.com:8080

NAME
awesome-tiger
```
or
```
$ export HTTPS_PROXY=http://example.com:8080
$ ./argo template list

NAME
awesome-tiger
```

PR #13447, opened 2024-08-09
feat(cli): Extend gRPC and HTTP clients to support client certificate. Fixes #13437

### LINKED ISSUE #13437: CLI ignores TLS client certificate with `--argo-http1` and `--argo-server`
### Pre-requisites

We have a corporate proxy that requires client certificate authentication to access services.

Therefore, I specify `--client-certificate` and `--client-key`

This does work:
`argo --server https://kubernetes-endpoint -n argo --token='XXX' --client-certificate ~/.secrets/client.pem --client-key ~/.secrets/key.pem list`

This does not work:
`argo --argo-http1 --client-certificate ~/.secrets/client.pem --client-key ~/.secrets/key.pem --loglevel DEBUG  --argo-server argo-server-instance:443 -n argo --token=XXX list`

Output:
```
DEBU[2024-08-05T22:10:42.749Z] CLI version                                   version="{v3.5.10 2024-08-01T05:52:04Z 25829927431d9a0f46d17b72ae74aedb8d700884 v3.5.10 clean go1.21.12 gc darwin/amd64}"
DEBU[2024-08-05T22:10:42.749Z] Client options                                opts="(argoServerOpts=(url=argo-server-instance:443,path=,secure=true,insecureSkipVerify=false,http=true),instanceID=)"
DEBU[2024-08-05T22:10:42.751Z]                                               listOpts="&ListOptions{LabelSelector:,FieldSelector:,Watch:false,ResourceVersion:,TimeoutSeconds:nil,Limit:0,Continue:,AllowWatchBookmarks:false,ResourceVersionMatch:,}"
DEBU[2024-08-05T22:10:42.753Z] curl -X GET -H 'Authorization: ******' -d '' 'https://argo-server-instance:443/api/v1/workflows/argo?fields=metadata%2Citems.metadata%2Citems.spec%2Citems.status.phase%2Citems.status.message%2Citems.status.finishedAt%2Citems.status.startedAt%2Citems.status.estimatedDuration%2Citems.status.progress' 
FATA[2024-08-05T22:10:42.830Z] Get "https://argo-server-instance:443/api/v1/workflows/argo?fields=metadata%2Citems.metadata%2Citems.spec%2Citems.status.phase%2Citems.status.message%2Citems.status.finishedAt%2Citems.status.startedAt%2Citems.status.estimatedDuration%2Citems.status.progress": EOF
```

Whereas when I copy the curl command from the output, fill in the token and add the certs, it works again:
`curl --cert ~/.secrets/client.pem --key ~/.secrets/key.pem -X GET -H 'Authorization: Bearer XXX' -d '' 'https://argo-server-instance:443/api/v1/workflows/argo?fields=metadata%2Citems.metadata%2Citems.spec%2Citems.status.phase%2Citems.status.message%2Citems.status.finishedAt%2Citems.status.startedAt%2Citems.status.estimatedDuration%2Citems.status.progress'`

I suspect that the client certs are ignored when using `--argo-http1` and `--argo-server`.

### Version(s)

v3.5.10

```YAML
Has nothing to do with workflows
```

```text
```

```text
```

miltalex:
Fixes #13437

### Motivation

This PR addresses a feature request to extend the client TLS certificate support functionality in server mode. Currently, the `--client-certificate` and `--client-key` flags are inherited from `kubectl` and are only used when connecting to Kubernetes directly via `client-go` (the Kubernetes Go SDK). The CLI doesn't use these flags itself and passes them through to `client-go`. However, in server mode, `client-go` is not used.

This enhancement aims to reuse these flags in server mode by passing them to the gRPC and HTTP clients as well. This will provide consistent client certificate authentication capabilities across different connection modes, improving the overall flexibility and security of the system.

### Modifications

The following changes have been implemented to extend the gRPC and HTTP1 client with client TLS certificate support:

1. In `cmd/argo/commands/client/conn.go`:
     - Added `AddKubectlFlagsToArgoServerOpts()` function to include client certificate and key in `ArgoServerOpts`.
     - Called `AddKubectlFlagsToArgoServerOpts()` to apply the changes.
2. In `pkg/apiclient/argo-server-opts.go`:
     - Added `ClientCert` and `ClientKey` fields to `ArgoServerOpts` struct.
     - Updated `String()` method to include client certificate and key information in the output. + unit tests
3. In `pkg/apiclient/apiclient.go`:
     - Updated `newHTTP1Client()` call to include client certificate and key parameters.
4. In `pkg/apiclient/argo-server-client.go`:
     - Implemented TLS configuration logic to support client certificates in gRPC connections.
     - Added conditional logic to create TLS config with or without client certificates based on provided options.
5. In `pkg/apiclient/http1/facade.go`:
     - Added `ClientCert` and `ClientKey` fields to `Facade` struct.
     - Implemented TLS configuration logic to support client certificates in http1 connections.
6. Adapted path and parameters.

These changes enable the use of client TLS certificates in both gRPC and HTTP1 clients when connecting to the Argo server in server mode. This enhancement provides a more consistent and secure authentication mechanism across different connection types.

### Verification

The changes implemented in this PR were verified through a local testing environment designed to replicate the reported issue. The verification process consisted of the following steps:

1. Environment Setup:
       - Deployed a K3D cluster on macOS.
       - Installed Argo Workflows in the cluster.
       - Set up an Nginx server configured as a proxy, expecting client certificates.
       - Generated self-signed certificates for testing purposes.
2. Issue Replication:
       - Used the commands provided in the original issue report to connect to `localhost:63748`.
       - Successfully replicated the "bug" where client certificates were not being used in server mode.
3. Solution Testing:
       - Applied the changes from this PR to the local environment.
       - Reran the same commands used to replicate the issue.
       - Observed logs to confirm that the client certificate was now being used correctly in the connection process.
4. Limitations of Testing:
       - Due to the complexity of the setup, creation of end-to-end (e2e) tests was challenging and not implemented in this verification process.
       - Testing was limited to the local environment with self-signed certificates. Additional testing in staging/production environments with properly signed certificates was not performed.

## Summary by CodeRabbit

* **New Features**
  * Enabled client TLS certificate authentication for both gRPC and HTTP/1.1, including server-mode usage.
  * Client certificate/key settings are now applied consistently across connection paths.

* **Bug Fixes**
  * Added stricter validation: both client certificate and key must be provided together.
  * Improved TLS handling by enforcing TLS 1.2+ and returning clearer certificate/key configuration errors.

* **Tests**
  * Added end-to-end coverage verifying the client certificate is presented for both gRPC and HTTP/1.1.

* **Documentation**
  * Updated notes clarifying client TLS flags behavior in server mode.


pkg/apiclient/http1/facade.go
```diff
@@ -115,11 +126,20 @@ func (h Facade) do(ctx context.Context, in interface{}, out interface{}, method
...
+			InsecureSkipVerify: h.insecureSkipVerify,
+		}
+		// Load client certificate if paths are provided
+		if h.clientCert != "" && h.clientKey != "" {
+			cert, err := tls.LoadX509KeyPair(h.clientCert, h.clientKey)
+			if err != nil {
+				return err
+			}
+			tlsConfig.Certificates = []tls.Certificate{cert}
+		}
```
agilgur5, line 139: this is identical to the block added above, so a helper function probably makes sense here to DRY -- `getTLSConfig(h)` or something

pkg/apiclient/argo-server-client.go
```diff
@@ -63,7 +63,28 @@ func (a *argoServerClient) NewInfoServiceClient() (infopkg.InfoServiceClient, er
...
+			tlsConfig = &tls.Config{
+				Certificates:       []tls.Certificate{cert},
+				InsecureSkipVerify: opts.InsecureSkipVerify,
+			}
+		} else {
+			// Create TLS configuration without client certificates
+			tlsConfig = &tls.Config{
+				InsecureSkipVerify: opts.InsecureSkipVerify,
+			}
+		}
```
agilgur5, line 85: same helper as above could be used here too if I'm not mistaken

cmd/argo/commands/client/conn.go
```diff
@@ -54,6 +54,11 @@ func AddAPIClientFlagsToCmd(cmd *cobra.Command) {
 	cmd.PersistentFlags().BoolVarP(&ArgoServerOpts.InsecureSkipVerify, "insecure-skip-verify", "k", os.Getenv("ARGO_INSECURE_SKIP_VERIFY") == "true", "If true, the Argo Server's certificate will not be checked for validity. This will make your HTTPS connections insecure. Defaults to the ARGO_INSECURE_SKIP_VERIFY environment variable.")
 }
 
+func AddKubectlFlagsToArgoServerOpts() {
+	ArgoServerOpts.ClientCert = overrides.AuthInfo.ClientCertificate
+	ArgoServerOpts.ClientKey = overrides.AuthInfo.ClientKey
+}
```
agilgur5, line 60: I think this would make more sense to put right below `AddKubectlFlagsToCmd` to keep the `kubectl` inheritance together

we probably also want to add a comment here, something like `// some kubectl flags apply to server mode as well`

cmd/argo/commands/root.go
```diff
@@ -121,6 +121,7 @@ If your server is behind an ingress with a path (running "argo server --base-hre
 
 	client.AddKubectlFlagsToCmd(command)
 	client.AddAPIClientFlagsToCmd(command)
+	client.AddKubectlFlagsToArgoServerOpts()
```
agilgur5, line 124:
```suggestion
	client.AddKubectlFlagsToCmd(command)
	client.AddKubectlFlagsToArgoServerOpts()
	client.AddAPIClientFlagsToCmd(command)
```

same as above re: ordering

pkg/apiclient/apiclient.go
```diff
@@ -73,7 +73,7 @@ func NewClientFromOpts(opts Opts) (context.Context, Client, error) {
 		if opts.AuthSupplier == nil {
 			return nil, nil, fmt.Errorf("AuthSupplier cannot be empty when connecting to Argo Server")
 		}
-		return newHTTP1Client(opts.ArgoServerOpts.GetURL(), opts.AuthSupplier(), opts.ArgoServerOpts.InsecureSkipVerify, opts.ArgoServerOpts.Headers, opts.ArgoServerOpts.HTTP1Client)
+		return newHTTP1Client(opts.ArgoServerOpts.GetURL(), opts.AuthSupplier(), opts.ArgoServerOpts.InsecureSkipVerify, opts.ArgoServerOpts.Headers, opts.ArgoServerOpts.HTTP1Client, opts.ArgoServerOpts.ClientCert, opts.ArgoServerOpts.ClientKey)
```
agilgur5, line 76: this is fine for now, but given how many options we're passing through, we may want to consider passing through `ArgoServerOpts` entirely (or `opts` entirely)

agilgur5 (commented):
Nice work! I left some mostly stylistic comments below

> - Due to the complexity of the setup, creation of end-to-end (e2e) tests was challenging and not implemented in this verification process.

Yea that makes sense to me. Unit tests may be fine for now.

Also we could probably have a solid e2e without the set-up, that just verifies that certs were passed in the request from the CLI. Doesn't actually have to auth them or run a proxy etc

cmd/argo/commands/client/conn.go
miltalex: Indeed, that is much better I was also unsure if it is a good idea to create a new `func`. I will adjust it

pkg/apiclient/apiclient.go
miltalex: Nice catch, I can refactor this in a commit here will improve readability by a lot.

cmd/argo/commands/client/conn.go
agilgur5: The separate function is fine, especially as it is a side-effect, but putting the two functions together in the same file helps keep the functionality together

pkg/apiclient/http1/facade.go
miltalex: Create a helper function in `utils/tls`

cmd/argo/commands/client/conn.go
miltalex: So for the above, I noticed the following, I had to use pointers to string cause we call this during the `root` `cmd` the `overrides.AuthInfo` is still empty. If we do not want to use pointers here I can move the `AddKubectlFlagsToArgoServerOpts()` call inside the ``NewApiClient` func.

pkg/apiclient/apiclient.go
miltalex: I am passing opts to the `newHTTP1Client`

util/tls/tls_test.go
```diff
@@ -38,3 +38,62 @@ func TestGeneratePEM(t *testing.T) {
...
+
+			if tt.wantErr {
+				require.Error(t, err)
+				require.Nil(t, config)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, config)
+			require.Equal(t, tt.insecureSkipVerify, config.InsecureSkipVerify)
```
agilgur5, line 90: so you shouldn't use `require` _everywhere_. only in places where you'd have an `if assert`. same as https://github.com/argoproj/argo-workflows/pull/13474#discussion_r1718397249

pkg/apiclient/argo-server-opts.go
```diff
@@ -17,6 +17,10 @@ type ArgoServerOpts struct {
 	// use custom http client
 	HTTP1Client *http.Client
 	Headers     []string
+
+	// client Certificates
```
agilgur5, line 21:
```suggestion
	// client certificate
```
singular, plus use consistent spacing and capitalization as the other opts

util/tls/testdata/valid_tls_cert.pem
agilgur5, line 1: can you use the existing cert and key in this directory?

cmd/argo/commands/client/conn.go
agilgur5: did you retest to make sure this works? my only concern is the ordering such that `overrides` is no longer empty by the time it's called

agilgur5 (commented):
some small comments below mostly on the testing code

also not sure if you had caught my above [comment](https://github.com/argoproj/argo-workflows/pull/13447#pullrequestreview-2231629228) on e2es:
> Also we could probably have a solid e2e without the set-up, that just verifies that certs were passed in the request from the CLI. Doesn't actually have to auth them or run a proxy etc

util/tls/testdata/valid_tls_cert.pem
miltalex: When I tried it was complaining that it wasn't a valid certificate. Tested also with OpenSSL, I wasn't sure if I should replace it or create a new one.

cmd/argo/commands/client/conn.go
miltalex: Yeah I did a couple of retests and I could see the certificates, also I could see them on the debug logs for argoserver opts, when before it wasn't there. Could also see the handshake at the nginx logs.

agilgur5: Great! Just wanted to make sure since the ordering in the CLI can be a bit hard to trace

util/tls/testdata/valid_tls_cert.pem
agilgur5: The `D` at the [end of the cert](https://github.com/argoproj/argo-workflows/blob/e26eb0e6b56d9f683c76a6d2353ab3c0591e4d9c/util/tls/testdata/valid_tls.crt#L29) did look broken/unintentional.

It seems like #7621 added this as test data but was missing tests? 🤔 In any case, it's test data, so so long as tests still pass, replacing it is fine. I'd check if removing the `D` fixes it as a smaller diff is always good.

> I wasn't sure 

Feel free to ask in your own PR comments, in-line or otherwise! Can also leave a note in a "Notes to Reviewers" section in the PR as I sometimes do

miltalex: > The D at the [end of the cert](https://github.com/argoproj/argo-workflows/blob/e26eb0e6b56d9f683c76a6d2353ab3c0591e4d9c/util/tls/testdata/valid_tls.crt#L29) did look broken/unintentional.

yeah even without it still it is not valid. I will replace it since it is not used. 

> Feel free to ask in your own PR comments, in-line or otherwise! Can also leave a note in a "Notes to Reviewers" section in the PR as I sometimes do

Cool tip, thanks, I will keep it in mind for next time

pkg/apiclient/argo-server-opts.go
```diff
@@ -17,6 +17,10 @@ type ArgoServerOpts struct {
 	// use custom http client
 	HTTP1Client *http.Client
 	Headers     []string
+
```
agilgur5, line 20:
```suggestion
```

inconsistent new line, as mentioned [previously](https://github.com/argoproj/argo-workflows/pull/13447#discussion_r1721102469)

agilgur5 (commented):
A simplified E2E test as mentioned [before](https://github.com/argoproj/argo-workflows/pull/13447#pullrequestreview-2244454763) would still be good to have

util/tls/tls.go
```diff
@@ -151,3 +151,18 @@ func GetServerTLSConfigFromSecret(ctx context.Context, kubectlConfig kubernetes.
 		NextProtos:   []string{"h2"},
 	}, nil
 }
+
```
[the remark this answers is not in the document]
miltalex: updated docs with suggestion

isubasinghe:
Thanks a lot for rebasing this!

Joibel (commented):
### Code review

Reviewed with 11 parallel review agents, with each candidate finding then verified by an independent scoring pass. 12 findings survived (7 high-confidence, 5 moderate) and are posted as inline comments on the relevant lines — the most significant is that the e2e “gRPC” subtest actually runs over HTTP/1, leaving the new gRPC client-cert path without e2e coverage.

One out-of-scope observation, not raised inline: `argo cp` builds its own TLS config in `cmd/argo/commands/cp.go` and so still ignores `--client-certificate`/`--client-key`; since this PR doesn’t touch those lines it may deserve a follow-up issue rather than a change here.

test/e2e/cli_test.go
```diff
@@ -2114,6 +2123,215 @@ func (s *CLISuite) TestWorkflowConvert() {
 	})
 }
 
+func (s *CLISuite) TestClientCerts() {
+	s.Run("gRPC", func() {
+		tlsServerAddr, clientCertPath, clientKeyPath, receivedCertCh := s.startTLSServerWithClientAuth()
+		s.runCliWithClientCerts(tlsServerAddr, clientCertPath, clientKeyPath, []string{"version"}, receivedCertCh)
+	})
```
Joibel, line 2135: **This “gRPC” subtest doesn’t actually exercise gRPC.** `CLISuite.BeforeTest` calls `setMode(HTTP1)`, which sets `ARGO_HTTP1=true`, and `--argo-http1` defaults to that env var (see `cmd/argo/commands/client/conn.go`). This subtest never passes `--argo-http1=false` or unsets the env, so both subtests go through the HTTP/1 facade and the new gRPC client-cert path (`newClientConn` → `tls.GetTLSConfig` in `pkg/apiclient/argo-server-client.go`) gets no e2e coverage. The mock server is a plain `http.Server` and can’t answer gRPC anyway — `version` succeeding here is itself evidence the request went over HTTP/1. Genuinely covering the gRPC path needs a gRPC-capable test server; otherwise this subtest should be renamed/removed so it doesn’t claim coverage it doesn’t provide.

pkg/apiclient/http1/facade.go
```diff
@@ -159,6 +138,27 @@ func (h Facade) do(ctx context.Context, in any, out any, method string, path str
...
+		return nil, err
+	}
+	if h.httpClient != nil {
+		return h.httpClient, nil
+	}
+	tlsConfig, err := tls.GetTLSConfig(h.clientCert, h.clientKey, h.insecureSkipVerify)
+	if err != nil {
+		return nil, err
+	}
+	return &http.Client{
```
Joibel: `client()` runs per request, so this now re-reads and re-parses the client cert/key files from disk on **every** HTTP request (`GetTLSConfig` → `tls.LoadX509KeyPair`). Consider loading the keypair once (cache the `tls.Certificate` on the `Facade`, or use `tls.Config.GetClientCertificate`).

util/tls/testdata/valid_tls.crt
```diff
@@ -1,29 +1,21 @@
 -----BEGIN CERTIFICATE-----
```
Joibel: This regenerated fixture is valid 2024-08-11 → 2025-08-11, i.e. already expired. `TestGetTLSConfig` passes only because `tls.LoadX509KeyPair` doesn’t check NotBefore/NotAfter — any future test doing a real handshake against it will fail. The previous fixture had exactly the same flaw (expired 2022), and this PR’s own e2e test already shows the better pattern: generate certs at runtime (or use a very long-dated cert).

pkg/apiclient/http1/facade.go
```diff
@@ -159,6 +138,27 @@ func (h Facade) do(ctx context.Context, in any, out any, method string, path str
 	return nil
 }
 
+func (h Facade) client(req *http.Request, disableKeepAlives bool) (*http.Client, error) {
+	proxyURL, err := h.proxyFunc()(req)
+	if err != nil {
+		return nil, err
+	}
+	if h.httpClient != nil {
```
Joibel, line 172: The new `client()` method has two error paths (proxy resolution and cert loading) but no coverage in `facade_test.go` — the feature is only tested via the `util/tls` unit tests and the cli-build-tagged e2e test. Precedent from #12527 (same files, same option-threading pattern): the reviewer requested tests in `apiclient_test.go` and `facade_test.go` before acceptance.

.features/pending/13437-client-tls-server-mode.md
```diff
@@ -0,0 +1,13 @@
...
+Issues: 13437
+
+The client TLS certificate support functionality has been extended in server mode.
+Previously, the `--client-certificate` and `--client-key` flags were inherited from kubectl
+and were only used when connecting to Kubernetes directly via client-go (the Kubernetes Go SDK).
+The CLI did not use these flags itself and passed them through to client-go. However, in server mode, client-go is not used.
+
+These flags are now reused in server mode by passing them to the gRPC and HTTP clients as well.
+This provides consistent client certificate authentication capabilities across different connection modes,
+improving the overall flexibility and security of the system.
```
Joibel: This reads as internal history rather than the user-aimed guidance `.features/TEMPLATE.md` asks for: no usage example, no mention that **both** flags must be provided together (the code errors on one without the other), and it ends on generic filler (“improving the overall flexibility and security of the system”). Also a couple of one-sentence-per-line violations (one sentence split across lines 7–8; two sentences packed on line 9).

pkg/apiclient/http1-client.go
```diff
@@ -47,6 +45,15 @@ func (h httpClient) NewSyncServiceClient(_ context.Context) (syncpkg.SyncService
 	return http1.SyncServiceClient(h), nil
 }
 
-func newHTTP1Client(ctx context.Context, baseURL string, auth string, insecureSkipVerify bool, headers []string, customHTTPClient *http.Client, proxy func(*http.Request) (*url.URL, error)) (context.Context, Client, error) {
-	return ctx, httpClient(http1.NewFacade(baseURL, auth, insecureSkipVerify, headers, customHTTPClient, proxy)), nil
+func newHTTP1Client(ctx context.Context, opts Opts) (context.Context, Client, error) {
```
Joibel: `newHTTP1Client` now takes the whole `Opts` while its sibling `newArgoServerClient` still takes narrow `(ArgoServerOpts, authString)` — the two parallel Argo-Server client constructors end up with inconsistent contracts, and this one’s dependency surface exceeds what it uses. Suggest aligning both on the same shape (either both take `ArgoServerOpts` + what they need, or both take `Opts`).

pkg/apiclient/http1/facade.go
```diff
@@ -28,10 +28,12 @@ type Facade struct {
 	headers            []string
 	httpClient         *http.Client
 	proxy              func(*http.Request) (*url.URL, error)
+	clientCert         string
+	clientKey          string
 }
 
-func NewFacade(baseURL, authorization string, insecureSkipVerify bool, headers []string, httpClient *http.Client, proxy func(*http.Request) (*url.URL, error)) Facade {
-	return Facade{baseURL, authorization, insecureSkipVerify, headers, httpClient, proxy}
+func NewFacade(baseURL, authorization string, insecureSkipVerify bool, headers []string, httpClient *http.Client, proxy func(*http.Request) (*url.URL, error), clientCert, clientKey string) Facade {
```
Joibel: `NewFacade` is now 8 positional parameters ending in two adjacent `string`s (`clientCert, clientKey`) — a transposition at a call site compiles fine and only fails at runtime. Worth converting to an options/config struct (or passing `ArgoServerOpts` through) before the list grows again.

util/tls/tls.go
```diff
@@ -150,3 +150,25 @@ func GetServerTLSConfigFromSecret(ctx context.Context, kubectlConfig kubernetes.
...
+// GetTLSConfig creates a TLS configuration for client connections.
+// If clientCert and clientKey are provided, they will be loaded and included in the configuration.
+// The insecureSkipVerify parameter controls whether the server's certificate will be verified.
```
Joibel: Nit: the doc comment misses two behaviors callers can observe: (1) supplying exactly one of clientCert/clientKey returns an error; (2) `MinVersion` is unconditionally TLS 1.2, unlike the sibling constructors in this file which take `tlsMinVersion` as a parameter. Also, siblings are role-named (`GetServerTLSConfigFromSecret`…) — `GetClientTLSConfig` would make the client role explicit.

cmd/argo/commands/client/conn.go
```diff
@@ -36,6 +36,12 @@ func AddKubectlFlagsToCmd(cmd *cobra.Command) {
 	clientcmd.BindOverrideFlags(&overrides, cmd.PersistentFlags(), kflags)
 }
 
+// AddKubectlFlagsToArgoServerOpts applies kubectl flags to server mode as well.
+func AddKubectlFlagsToArgoServerOpts() {
```
Joibel: Nit: the name mirrors `AddKubectlFlagsToCmd`/`AddAPIClientFlagsToCmd`, which register cobra flags, but this registers nothing — it copies two already-parsed values between package globals — and the doc comment (“applies kubectl flags”) overstates that scope. It’s also exported with a single same-package caller; could be unexported or inlined into `NewAPIClient`.

cmd/argo/commands/client/conn.go
```diff
@@ -36,6 +36,12 @@ func AddKubectlFlagsToCmd(cmd *cobra.Command) {
...
+	ArgoServerOpts.ClientCert = overrides.AuthInfo.ClientCertificate
+	ArgoServerOpts.ClientKey = overrides.AuthInfo.ClientKey
```
Joibel: This reads only the `--client-certificate`/`--client-key` **flag** overrides, so client certs embedded in the user’s kubeconfig context are still silently ignored in server mode — while they *do* work in direct-Kubernetes mode via client-go. That asymmetry is a reasonable scope decision, but it will surprise kubeconfig-based users; worth a doc note or an explicit mention in the feature file.

.features/pending/13437-client-tls-server-mode.md
```diff
@@ -0,0 +1,13 @@
+Description: Extend client TLS certificate support in server mode
```
Joibel, line 1: `docs/tls.md` (the canonical TLS guide) isn’t updated for this: nothing tells users the flags now work in server mode, that both must be supplied together, or that — unlike `--secure`/`--insecure-skip-verify` — there are no `ARGO_*` env-var equivalents for them.

util/tls/tls.go
```diff
@@ -150,3 +150,24 @@ func GetServerTLSConfigFromSecret(ctx context.Context, kubectlConfig kubernetes.
...
+
+// GetClientTLSConfig creates a TLS 1.2 or newer configuration for client connections.
+// Client certificate authentication requires both clientCert and clientKey.
+// The insecureSkipVerify parameter controls whether the server's certificate is verified.
+func GetClientTLSConfig(clientCert, clientKey string, insecureSkipVerify bool) (*tls.Config, error) {
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: insecureSkipVerify,
+		MinVersion:         tls.VersionTLS12,
+	}
+	if (clientCert == "") != (clientKey == "") {
```
Joibel, line 163: This is only on the http1 path - so suggest moving it to `conn.go`.

cmd/argo/commands/client/conn.go
```diff
@@ -57,6 +57,10 @@ func AddAPIClientFlagsToCmd(cmd *cobra.Command) {
 }
 
 func NewAPIClient(ctx context.Context) (context.Context, apiclient.Client, error) {
+	// Reuse the explicit kubectl client certificate flags in server mode.
+	ArgoServerOpts.ClientCert = overrides.AuthInfo.ClientCertificate
+	ArgoServerOpts.ClientKey = overrides.AuthInfo.ClientKey
+
```
Joibel, line 68: I think we should do the xor check here instead of (or as well as) `GetClientTLSConfig`

```suggestion
  if (opts.ClientCert == "") != (opts.ClientKey == "") {
      return nil, errors.New("--client-certificate and --client-key must be provided together")
  }
```

Joibel (changes requested):
A more major thing, at the moment there's no way to pass in a Certificate Authority Cert, so we can do full mTLS - we don't have a hook to provide CACert so we can validate the server. I think we should add it, otherwise this is only the client side. It looks like a relatively small add, push back if you're sure you only want client cert support.

Joibel (approved):
Nice, one, thank you. An epic journey!

PR #13429, opened 2024-08-02
fix(docs): replace outdated `whalesay` image with `busybox`

agilgur5:
Fixes #11858, Fixes #13388, Fixes https://github.com/argoproj/argo-workflows/discussions/12767#discussioncomment-9210927

### Motivation

The `whalesay` image was [last updated 9 years ago](https://hub.docker.com/r/docker/whalesay) and does not support Docker Manifest v2 nor `arm64` architectures.
Per https://github.com/docker/whalesay/issues/6, where I tweeted at Docker and emailed support twice, Docker has no intention to update the image (it just requires a rebuild, that's it... 🤷 ) nor even mark it as deprecated or archived... 😕 

### Modifications

- replace `whalesay` image with `busybox` image
  - also instead of `cowsay`, use plain `echo`
  - allow `busybox` in tested examples

<details><summary><h4> secondary modifications </h4></summary>

- rename several templates that were named `whalesay` unnecessarily as well
  - they often just printed a message, and so are now named `print-message`
  - others always printed "hello world", and so are now named `hello-world`
  - some didn't even use the `whalesay` image, and so are now named more accordingly

- also clarify a few other template names that would do more than print a message
  - now named `print-message-from-file` or `hello-world-to-file`

- plus make some small style adjustments per the docs style guide while at it
  - and remove some trailing newlines in codeblocks
  
</details>

### Verification

- No more results for `whalesay` or `cowsay` in `docs` and `examples`
  - well, there are a few leftover, from the upgrading guide and the configmap, where `whalesay` is very intentionally used as a manifest v1 example. also a few `cowsay` ones leftover are all from the `argosay` image, see "Future Work" below
- Some of these, like the "hello world" doc, I manually tested
- The examples have an E2E suite that tests all of them

### Future Work

1. Replace `argosay` as well and remove `argosay` entirely from the codebase
    - this does need a bit more testing just to make sure it's no longer needed with Emissary
1. Replace both `argosay` and `whalesay` in all tests 
   - there seems to be 1000+ references in the tests 😕
   - they don't seem to cause problems yet and aren't user-facing

Issue #13061, opened 2024-05-16
v3.5.5 `unknown name groups` during SSO


Freddybob4244:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### Summary

All non-admin users assume a read-only role.  After v3.5.5 is applied, non-admin users can no longer use argo-workflows and are prompted with a generic error in the UI:

![image](https://github.com/argoproj/argo-workflows/assets/22138929/06543b97-1d9a-48c9-92c9-07cf228c08bc)
![image](https://github.com/argoproj/argo-workflows/assets/22138929/edfa53cc-5de0-45b2-ab94-c6b621651540)

Admin users can navigate normally.

Argo-server pod logs provide a bit more useful insight:

```
time="2024-05-15T17:49:33.746Z" level=error msg="failed to perform RBAC authorization" error="failed to evaluate rule: unknown name groups (1:43)\n | 'ac1ef805-ac6a-4ce9-854a-1cb406aa7121' in groups\n | ..........................................^"

time="2024-05-15T17:49:33.746Z" level=warning msg="finished unary call with code PermissionDenied" error="rpc error: code = PermissionDenied desc = not allowed" grpc.code=PermissionDenied grpc.method=ListWorkflows grpc.service=workflow.WorkflowService grpc.start_

time="2024-05-15T17:49:33Z" grpc.time_ms=1.682 span.kind=server system=grpc`

time="2024-05-15T17:49:33.746Z" level=info duration=2.267779ms method=GET path=/api/v1/workflows/argo size=34 status=403
```

There aren't any additional logs that are related to this error that I can find.

### Testing Performed

To confirm 3.5.5 introduced the issue I tested a few different versions with a read-only test user in a sandbox installation.

- Installed 3.5.3 - No issues
- Installed 3.5.4 - No issues
- Installed 3.5.5 - Issue introduced
- Installed 3.5.6 - Issue persists
- Installed Lastest - Issue persists

### Configurations

<details> <summary> <h4>Argo-Server </h4> </summary>

```yaml
[...]

PR #13448, opened 2024-08-09
fix: avoid exit handler nil pointer when missing outputs. Fixes #13445

### LINKED ISSUE #13445: Exit handler nil pointer crash when failed step misses output parameter
### Pre-requisites

When an exit handler is called on a failed step that has an output parameter which is null/empty (because the step crashed before creating the output file used in `valueFrom`), the exit handler fails with the message: `runtime error: invalid memory address or nil pointer dereference`. Seems similar to https://github.com/argoproj/argo-workflows/pull/12288 but in this case, the output params assumed to be always present.

Originally tested on v3.5.5 but also present in 3.5.10

### Version(s)

v3.5.5, v3.5.10

```YAML
metadata:
  name: exit-handler-bug
spec:
  entrypoint: failure-workflow
  templates:
    - name: failure-workflow
      steps:
        - - name: step1
            template: intentional-fail
            hooks:
              exit:
                template: lifecycle-hook
                arguments:
                  parameters:
                    - name: step-status
                      value: '{{steps.step1.status}}'
    - name: intentional-fail
      outputs:
        parameters:
          - name: hello-param
            valueFrom:
              path: /tmp/hello_world.txt
      container:
        image: alpine:latest
        command: ["sh", "-c"]
    - name: lifecycle-hook
      inputs:
        parameters:
          - name: step-status
      http:
        url: 'http://localhost:80/{{inputs.parameters.step-status}}'
```

```text
time="2024-08-09T08:45:01.126Z" level=info msg="Processing workflow" Phase= ResourceVersion=258701967 namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.131Z" level=info msg="resolved artifact repository" artifactRepositoryRef=default-artifact-repository
time="2024-08-09T08:45:01.131Z" level=info msg="Task-result reconciliation" namespace=workflows numObjs=0 workflow=exit-handler-bug
time="2024-08-09T08:45:01.131Z" level=info msg="Updated phase  -> Running" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.131Z" level=warning msg="Node was nil, will be initialized as type Skipped" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.132Z" level=info msg="was unable to obtain node for , letting display name to be nodeName" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.132Z" level=info msg="Steps node exit-handler-bug initialized Running" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.132Z" level=info msg="StepGroup node exit-handler-bug-3373743507 initialized Running" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.132Z" level=warning msg="Node was nil, will be initialized as type Skipped" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.132Z" level=info msg="Pod node exit-handler-bug-2010843852 initialized Pending" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.158Z" level=info msg="Created pod: exit-handler-bug[0].step1 (exit-handler-bug-intentional-fail-2010843852)" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.158Z" level=info msg="Workflow step group node exit-handler-bug-3373743507 not yet completed" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.158Z" level=info msg="TaskSet Reconciliation" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.158Z" level=info msg=reconcileAgentPod namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:01.159Z" level=info msg="Workflow to be dehydrated" Workflow Size=1816
time="2024-08-09T08:45:01.170Z" level=info msg="Workflow update successful" namespace=workflows phase=Running resourceVersion=258701971 workflow=exit-handler-bug
time="2024-08-09T08:45:11.160Z" level=info msg="Processing workflow" Phase=Running ResourceVersion=258701971 namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:11.160Z" level=info msg="Task-result reconciliation" namespace=workflows numObjs=1 workflow=exit-handler-bug
time="2024-08-09T08:45:11.160Z" level=info msg="task-result changed" namespace=workflows nodeID=exit-handler-bug-2010843852 workflow=exit-handler-bug
time="2024-08-09T08:45:11.160Z" level=info msg="Pod failed: Error (exit code 1)" displayName=step1 namespace=workflows pod=exit-handler-bug-intentional-fail-2010843852 templateName=intentional-fail workflow=exit-handler-bug
time="2024-08-09T08:45:11.161Z" level=info msg="node changed" namespace=workflows new.message="Error (exit code 1)" new.phase=Failed new.progress=0/1 nodeID=exit-handler-bug-2010843852 old.message= old.phase=Pending old.progress=0/1 workflow=exit-handler-bug
time="2024-08-09T08:45:11.161Z" level=info msg="Running OnExit handler" lifeCycleHook="&LifecycleHook{Template:lifecycle-hook,Arguments:Arguments{Parameters:[]Parameter{Parameter{Name:step-status,Default:nil,Value:*{{steps.step1.status}},ValueFrom:nil,GlobalName:,Enum:[],Description:nil,},},Artifacts:[]Artifact{},},TemplateRef:nil,Expression:,}" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:11.161Z" level=error msg="Recovered from panic" namespace=workflows r="runtime error: invalid memory address or nil pointer dereference" stack="goroutine 283 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:24 +0x5e\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).operate.func2()\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:200 +0xb4\npanic({0x20c5ea0?, 0x39ef3f0?})\n\t/usr/local/go/src/runtime/panic.go:920 +0x270\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).resolveExitTmplArgument(0xc0007e6300, {{0xc00739ec00, 0x1, 0x1}, {0x0, 0x0, 0x0}}, {0xc017475ad0, 0xb}, 0xc016ff58c0, ...)\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/exit_handler.go:60 +0xa89\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).runOnExitNode(0xc0007e6300, {0x27be158, 0x3a5b2c0}, 0xc00739eba0, 0xc00082cf00, {0xc0174751d0, 0x10}, 0x10?, {0xc017475ad0, 0xb}, ...)\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/exit_handler.go:36 +0x445\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).executeStepGroup(0xc0007e6300, {0x27be158, 0x3a5b2c0}, {0xc0007e6d80, 0x1, 0x1}, {0xc010ce35c0, 0x13}, 0xc0008678e8)\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/steps.go:330 +0x1445\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).executeSteps(0xc0007e6300, {0x27be158, 0x3a5b2c0}, {0xc017474f00, 0x10}, 0xc016ff5b40, {0xc010ce34a0, 0x16}, 0xc003e59680, {0x27c1a60, ...}, ...)\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/steps.go:108 +0x7ea\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).executeTemplate(0xc0007e6300, {0x27be158, 0x3a5b2c0}, {0xc017474f00, 0x10}, {0x27c1a60, 0xc0007e6a80?}, 0x454c49465f4e454b?, {{0x0, 0x0, ...}, ...}, ...)\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:2142 +0x2e36\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*wfOperationCtx).operate(0xc0007e6300, {0x27be158?, 0x3a5b2c0})\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:370 +0x18cb\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*WorkflowController).processNextItem(0xc000222000, {0x27be158, 0x3a5b2c0})\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/controller.go:811 +0x627\ngithub.com/argoproj/argo-workflows/v3/workflow/controller.(*WorkflowController).runWorker(0xc00ee24ea0?)\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/controller.go:734 +0x88\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0x30?)\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:155 +0x33\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil(0x0?, {0x2796540, 0xc00f2cce40}, 0x1, 0xc0006fede0)\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:156 +0xaf\nk8s.io/apimachinery/pkg/util/wait.JitterUntil(0x0?, 0x3b9aca00, 0x0, 0x0?, 0x0?)\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:133 +0x7f\nk8s.io/apimachinery/pkg/util/wait.Until(0x0?, 0x0?, 0x0?)\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:90 +0x1e\ncreated by github.com/argoproj/argo-workflows/v3/workflow/controller.(*WorkflowController).Run in goroutine 28\n\t/go/src/github.com/argoproj/argo-workflows/workflow/controller/controller.go:344 +0x1885\n" workflow=exit-handler-bug
time="2024-08-09T08:45:11.161Z" level=info msg="Updated phase Running -> Error" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:11.161Z" level=info msg="Updated message  -> runtime error: invalid memory address or nil pointer dereference" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:11.161Z" level=info msg="Marking workflow completed" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:11.162Z" level=info msg="Marking workflow as pending archiving" namespace=workflows workflow=exit-handler-bug
time="2024-08-09T08:45:11.162Z" level=info msg="Workflow to be dehydrated" Workflow Size=2622
time="2024-08-09T08:45:11.167Z" level=info msg="cleaning up pod" action=deletePod key=workflows/exit-handler-bug-1340600742-agent/deletePod
time="2024-08-09T08:45:11.172Z" level=info msg="Workflow update successful" namespace=workflows phase=Error resourceVersion=258702051 workflow=exit-handler-bug
time="2024-08-09T08:45:11.181Z" level=info msg="archiving workflow" namespace=workflows uid=bf8ba3a8-9c7a-4a3a-83b5-a02ced40f10e workflow=exit-handler-bug
time="2024-08-09T08:45:11.205Z" level=info msg="Queueing Error workflow workflows/exit-handler-bug for delete in 1h0m0s due to TTL"
time="2024-08-09T08:45:16.181Z" level=info msg="cleaning up pod" action=deletePod key=workflows/exit-handler-bug-intentional-fail-2010843852/deletePod
```

```text
-
```

miltalex:
Fixes #13445

### Motivation

This PR addresses a critical issue where exit handlers fail when called on a failed step with a `null` or empty output parameter. Specifically, when a step crashes before creating the output file referenced in `valueFrom`, the exit handler encounters a runtime error: `"invalid memory address or nil pointer dereference".` The problem has been observed in Argo Workflow versions 3.5.5 through 3.5.10.
### Modifications

To resolve this issue:
1. Added a `nil` check before calling `params.Value.String()` in the relevant code section.
2. This check ensures that we only attempt to access the parameter value when it exists, preventing the `nil` pointer dereference.

### Verification

To ensure the effectiveness of this fix and prevent regression, I've added a new end-to-end (e2e) test case. This test case verifies:

1. The presence of an exit handler node in the workflow graph after a step failure.
2. Proper population of the exit handler's input based on the output from the failed step.

Previously, when the issue occurred, the exit node was never created, and the workflow terminated immediately after the first step. The new test confirms that:
1. The workflow correctly transitions to the exit handler node instead of prematurely completing.
2. The exit handler receives and can process the output parameter from the failed step, which during the next `operate` is populated and we can see the `failed` status.
image

### DIFF
--- a/test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
+++ b/test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
@@ -34,6 +34,6 @@ spec:
         parameters:
           - name: hello-param
       container:
-        image: argoproj/argosay:v1
-        command: [cowsay]
+        image: busybox
+        command: [echo]
         args: ["Hello param: {{inputs.parameters.hello-param}}"]
--- a/test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
+++ b/test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
@@ -1,0 +1,39 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: exit-handler-fail-missing-output-
+  labels:
+    test: test
+spec:
+  entrypoint: failure-workflow
+  templates:
+    - name: failure-workflow
+      steps:
+        - - name: step1
+            template: intentional-fail
+            hooks:
+              exit:
+                template: lifecycle-hook
+                arguments:
+                  parameters:
+                    - name: hello-param
+                      value: '{{steps.step1.outputs.parameters.hello-param}}'
+    - name: intentional-fail
+      outputs:
+        parameters:
+          - name: hello-param
+            valueFrom:
+              path: /tmp/hello_world.txt
+      container:
+        image: alpine:latest
+        command: ["sh", "-c"]
+        args: ["echo intentional failure; exit 1"]
+    - name: lifecycle-hook
+      inputs:
+        parameters:
+          - name: hello-param
+      container:
+        image: argoproj/argosay:v1
+        command: [cowsay]
+        args: ["Hello param: {{inputs.parameters.hello-param}}"]
--- a/workflow/controller/exit_handler.go
+++ b/workflow/controller/exit_handler.go
@@ -57,9 +57,11 @@ func (woc *wfOperationCtx) resolveExitTmplArgument(args wfv1.Arguments, prefix s
 	}
 	if outputs != nil {
 		for _, param := range outputs.Parameters {
+			value := ""
 			if param.Value != nil {
-				scope.addParamToScope(fmt.Sprintf("%s.outputs.parameters.%s", prefix, param.Name), param.Value.String())
+				value = param.Value.String()
 			}
+			scope.addParamToScope(fmt.Sprintf("%s.outputs.parameters.%s", prefix, param.Name), value)
 		}
 		for _, arts := range outputs.Artifacts {
 			scope.addArtifactToScope(fmt.Sprintf("%s.outputs.artifacts.%s", prefix, arts.Name), arts)
--- a/workflow/controller/exit_handler.go
+++ b/workflow/controller/exit_handler.go
@@ -57,7 +57,9 @@ func (woc *wfOperationCtx) resolveExitTmplArgument(args wfv1.Arguments, prefix s
 	}
 	if outputs != nil {
 		for _, param := range outputs.Parameters {
-			scope.addParamToScope(fmt.Sprintf("%s.outputs.parameters.%s", prefix, param.Name), param.Value.String())
+			if param.Value != nil {
+				scope.addParamToScope(fmt.Sprintf("%s.outputs.parameters.%s", prefix, param.Name), param.Value.String())
+			}
 		}
 		for _, arts := range outputs.Artifacts {
 			scope.addArtifactToScope(fmt.Sprintf("%s.outputs.artifacts.%s", prefix, arts.Name), arts)
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -118,7 +118,6 @@ func (s *FunctionalSuite) TestWhenExpressions() {
 }
 
 func (s *FunctionalSuite) TestJSONVariables() {
-
 	s.Given().
 		Workflow("@testdata/json-variables.yaml").
 		When().
@@ -1061,9 +1060,7 @@ spec:
 				}
 			}
 			assert.True(t, memoHit)
-
 		})
-
 }
 
 func (s *FunctionalSuite) TestStepLevelMemoizeNoOutput() {
@@ -1121,9 +1118,7 @@ spec:
 				}
 			}
 			assert.True(t, memoHit)
-
 		})
-
 }
 
 func (s *FunctionalSuite) TestDAGLevelMemoize() {
@@ -1187,9 +1182,7 @@ spec:
 				}
 			}
 			assert.True(t, memoHit)
-
 		})
-
 }
 
 func (s *FunctionalSuite) TestDAGLevelMemoizeNoOutput() {
@@ -1248,9 +1241,7 @@ spec:
 				}
 			}
 			assert.True(t, memoHit)
-
 		})
-
 }
 
 func (s *FunctionalSuite) TestContainerSetRetryFail() {
@@ -1331,7 +1322,6 @@ func (s *FunctionalSuite) TestEntrypointName() {
 			assert.Equal(t, wfv1.NodeSucceeded, n.Phase)
 			assert.Equal(t, "bar", n.Inputs.Parameters[0].Value.String())
 		})
-
 }
 
 func (s *FunctionalSuite) TestMissingStepsInUI() {
@@ -1378,5 +1368,37 @@ func (s *FunctionalSuite) TestTerminateWorkflowWhileOnExitHandlerRunning() {
 				}
 			}
 			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
+		})
+}
+
+// Exit handler ensure when failed steps ensure no crash and output parameter
+func (s *FunctionalSuite) TestWorkflowExitHandlerCrashEnsureNodeIsPresent() {
+	s.Given().
+		Workflow("@expectedfailures/exit-handler-fail-missing-output.yaml").
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeRunning).
+		WaitForWorkflow(fixtures.ToBeFailed).
+		Then().
+		ExpectWorkflow(func(t *testing.T, metadata *v1.ObjectMeta, status *wfv1.WorkflowStatus) {
+			var hasExitNode bool
+			var exitNodeName string
+
+			for _, node := range status.Nodes {
+				if !node.IsExitNode() {
+					continue
+				}
+				hasExitNode = true
+				exitNodeName = node.DisplayName
+			}
+			assert.True(t, hasExitNode)
+			assert.NotEmpty(t, exitNodeName)
+
+			hookNode := status.Nodes.FindByDisplayName(exitNodeName)
+
+			require.NotNil(t, hookNode)
+			assert.NotNil(t, hookNode.Inputs)
+			require.Len(t, hookNode.Inputs.Parameters, 1)
+			assert.NotNil(t, hookNode.Inputs.Parameters[0].Value)
 		})
 }


workflow/controller/exit_handler.go
```diff
@@ -57,7 +57,9 @@ func (woc *wfOperationCtx) resolveExitTmplArgument(args wfv1.Arguments, prefix s
 	}
 	if outputs != nil {
 		for _, param := range outputs.Parameters {
-			scope.addParamToScope(fmt.Sprintf("%s.outputs.parameters.%s", prefix, param.Name), param.Value.String())
+			if param.Value != nil {
+				scope.addParamToScope(fmt.Sprintf("%s.outputs.parameters.%s", prefix, param.Name), param.Value.String())
```
agilgur5, line 61: what will the error be if the parameter is `nil`? because an `unknown name` error like #13061 may be confusing. I wonder if empty string might be better? not entirely sure

agilgur5 (changes requested):
Thanks for tackling this! Left some simplifications in-line as well a question of whether we want to not add to scope or add to scope as empty string

test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
```diff
@@ -0,0 +1,37 @@
...
+      container:
+        image: alpine:latest
+        command: ["sh", "-c"]
+        args: ["echo intentional failure; exit 1"]
+    - name: lifecycle-hook
+      inputs:
+        parameters:
+          - name: step-status
+      http:
+        url: 'http://localhost:80/{{inputs.parameters.step-status}}'
```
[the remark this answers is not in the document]
miltalex: of course, I just reused the workflow from the issue.

test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
```diff
@@ -0,0 +1,37 @@
...
+      steps:
+        - - name: step1
+            template: intentional-fail
+            hooks:
+              exit:
+                template: lifecycle-hook
+                arguments:
+                  parameters:
+                    - name: step-status
+                      value: '{{steps.step1.status}}'
```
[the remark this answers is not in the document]
miltalex: I reused the workflow from the example that was triggering the nil pointer issue. I will have a look at the `hello-param` as well.

workflow/controller/exit_handler.go
miltalex: While debugging, I noticed that the first crash was caused by the `step-status` not being populated when the `resolveExitHandlerTmpl` was called. But once the status is available, if I recall correctly on the next iteration of operate the parameter is filled, and at the end, we can see it as an input (screenshot at verification with `Failed`)

Basically, I noticed before the fix:

1.  Step is failing.
2.  Exit handler node is not created due to `nil` pointer when resolving the template.
3.  Workflow is marked as done (2 nodes).

After the fix:

1.  Step is failing.
2.  Exit node handler is created and we get the `step-status` value.
3.  Workflow is marked as done only after exit handler node is finished.

agilgur5: Yes, that's because you have `step-status` here, I'm wondering if a user actually attempts to use `hello-param` as mention [above](https://github.com/argoproj/argo-workflows/pull/13448#discussion_r1712689839), what kind of error they'll get

miltalex: After testing it will get the same panic. The only difference is that will not resolve the template 
```
level=debug msg="unresolved is allowed " error=unresolved
```
and it will continue with the execution and input as shown below.
image

I am not quite sure which case is better empty string or does not add to the scope cause both will require digging into what happened with the previous step. Maybe an empty string is "safer".

agilgur5: Ah the dreaded unresolved check... That's meant to be used to defer a resolution when not possible at the current time (e.g. no output _yet_). But in this case, it will _never_ be possible.

The question is if we can consistently determine that is the case without breaking anything.

If so, yes an empty string would make more sense as the zero/null value of string (since nil itself wouldn't be valid here). A user could also detect that in an expression and handle the empty case then too.

Cc @isubasinghe might be worth a check from you as well given that you've dealt with the unresolved check before too

agilgur5: Also some closely related issues on non-existence checks: #11549, #12739, #12317

We'd want to make sure #12739 still works. And this might actually properly fix #12317?

miltalex: Thank you very much for the detailed information. I will wait for the valuable input from @isubasinghe and in the meantime I will try to test/debug the above issues.

agilgur5: The first two I linked already have root causes for reference. The last one may not. We do need better testing of expressions in general though as they can be somewhat easy to accidentally introduce breaking changes to. (and have some bugs, as well as inconsistencies in templating languages that I'd planned to eventually unify per #9529)

test/e2e/expectedfailures/exit-handler-fail-missing-output.yaml
```diff
@@ -0,0 +1,39 @@
...
+        image: alpine:latest
+        command: ["sh", "-c"]
+        args: ["echo intentional failure; exit 1"]
+    - name: lifecycle-hook
+      inputs:
+        parameters:
+          - name: hello-param
+      container:
+        image: argoproj/argosay:v1
+        command: [cowsay]
```
agilgur5, line 38:
```suggestion
        image: busybox
        command: [echo]
```

`argosay` is slated to be removed per https://github.com/argoproj/argo-workflows/pull/11613#issuecomment-1957786530 and #13429, "Future Work" part 1, so let's not add any more usage of it. can use `busybox` with plain `echo` as I did in #13429

miltalex: Thanks for the hint. Updated 🚀

agilgur5 (approved):
LGTM, will await a second review, preferably by @isubasinghe who's aware of the pitfalls of the unresolved check

isubasinghe (approved):
LGTM, I don't really like the empty string.

But imo fixing that is beyond the scope of this PR/issue.

We need to properly formalise expressions/scope.

bmjhversteeg:
Many thanks for the fix @miltalex & @agilgur5 !

agilgur5:
> But imo fixing that is beyond the scope of this PR/issue

True, it is. But I think empty string/zero value makes more sense than erroring or not resolving, since people want to detect non-existence in expressions.

PR #13454, opened 2024-08-12
fix: mark node failed if pod absent

isubasinghe:
Fixes #TODO


workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+				// pod created but informed not yet updated
+				woc.log.Errorf("couldn't obtain node for %s, but found pod, this is not expected, doing nothing", result.Name)
+			}
+			continue
+		}
+
+		if foundPod {
+			woc.log.Debugf("Got pod %s with phase %s for task result %s and node id %s with label %s", pod.Name, pod.Status.Phase, resultName, result.Name, label)
+			woc.log.Debugf("The node phase was %s for node named %s", node.Phase, node.Name)
+		} else if !foundPod && !node.Completed() {
```
isubasinghe, line 113: We could technically hit this case where a pod exists but hasn't made it to the cache yet.

isubasinghe:
Also @Joibel note that I haven't been able to reproduce any of the open source issues.
So this hasn't been tested against https://github.com/argoproj/argo-workflows/issues/12993

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+			if foundPod {
+				// how does this path make any sense?
+				// pod created but informed not yet updated
+				woc.log.Errorf("couldn't obtain node for %s, but found pod, this is not expected, doing nothing", result.Name)
+			}
+			continue
+		}
+
+		if foundPod {
+			woc.log.Debugf("Got pod %s with phase %s for task result %s and node id %s with label %s", pod.Name, pod.Status.Phase, resultName, result.Name, label)
```
Joibel, line 111: Do we need to leave in these debug lines?

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+}
+
+func (woc *wfOperationCtx) taskResultReconciliation() error {
+
 	objs, _ := woc.controller.taskResultInformer.GetIndexer().ByIndex(indexes.WorkflowIndex, woc.wf.Namespace+"/"+woc.wf.Name)
 	woc.log.WithField("numObjs", len(objs)).Info("Task-result reconciliation")
+
+	taskresultCli := woc.controller.wfclientset.ArgoprojV1alpha1().WorkflowTaskResults(woc.wf.Namespace)
+	// We will be hitting the cache here, should we bypass the cache?
+	podList, err := woc.getAllWorkflowPods()
```
Joibel, line 76: Suggest extracting this down to L88 as a separate method (something like `getWorkflowPodMap`) to help readability

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+	return time.Since(node.StartedAt.Time) <= envutil.LookupEnvDurationOr("POD_ABSENT_TIMEOUT", 2*time.Minute)
+}
+
+func (woc *wfOperationCtx) taskResultReconciliation() error {
+
 	objs, _ := woc.controller.taskResultInformer.GetIndexer().ByIndex(indexes.WorkflowIndex, woc.wf.Namespace+"/"+woc.wf.Name)
 	woc.log.WithField("numObjs", len(objs)).Info("Task-result reconciliation")
+
+	taskresultCli := woc.controller.wfclientset.ArgoprojV1alpha1().WorkflowTaskResults(woc.wf.Namespace)
+	// We will be hitting the cache here, should we bypass the cache?
```
Joibel, line 75: Hitting the cache feels fine, we shouldn't bypass it or this would be horrific for performance.

Suggest tidying the comment.

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+				woc.log.Errorf("couldn't obtain node for %s, but found pod, this is not expected, doing nothing", result.Name)
+			}
+			continue
+		}
+
+		if foundPod {
+			woc.log.Debugf("Got pod %s with phase %s for task result %s and node id %s with label %s", pod.Name, pod.Status.Phase, resultName, result.Name, label)
+			woc.log.Debugf("The node phase was %s for node named %s", node.Phase, node.Name)
+		} else if !foundPod && !node.Completed() {
+			woc.log.Debugf("couldn't find pod")
```
Joibel, line 114: Suggest removing this or tidying.

Joibel (changes requested):

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+func shouldErrorPodTimeout(node *wfv1.NodeStatus) bool {
+	return time.Since(node.StartedAt.Time) <= envutil.LookupEnvDurationOr("POD_ABSENT_TIMEOUT", 2*time.Minute)
+}
+
+func (woc *wfOperationCtx) taskResultReconciliation() error {
+
 	objs, _ := woc.controller.taskResultInformer.GetIndexer().ByIndex(indexes.WorkflowIndex, woc.wf.Namespace+"/"+woc.wf.Name)
 	woc.log.WithField("numObjs", len(objs)).Info("Task-result reconciliation")
+
+	taskresultCli := woc.controller.wfclientset.ArgoprojV1alpha1().WorkflowTaskResults(woc.wf.Namespace)
```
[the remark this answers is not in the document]
isubasinghe: Done

workflow/controller/taskresult.go
isubasinghe: I thought they might be useful when debugging this function.
I can remove them if needed, but I don't see any harm in leaving it in.

workflow/controller/taskresult.go
isubasinghe: Sounds good.

workflow/controller/taskresult.go
isubasinghe: Sure, can do

workflow/controller/taskresult.go
isubasinghe: Sure

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+			woc.log.Debugf("Got pod %s with phase %s for task result %s and node id %s with label %s", pod.Name, pod.Status.Phase, resultName, result.Name, label)
+			woc.log.Debugf("The node phase was %s for node named %s", node.Phase, node.Name)
+		} else if !foundPod && !node.Completed() {
+			woc.log.Debugf("couldn't find pod")
+			timeout := shouldErrorPodTimeout(node)
+			if timeout {
+				woc.log.Infof("Determined controller should timeout for %s", result.Name)
+				result.Labels[common.LabelKeyReportOutputsCompleted] = "true"
+
+				// patch the label
```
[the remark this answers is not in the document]
isubasinghe: I just thought it was nicer if the wfTaskResult was also updated the moment we decided to treat this wfTaskResult as completed.

Currently this would still work, but the completed value would be false. This is fine because it would get cleaned up by the controller anyway.

We can ensure we hit the api exactly once though, by checking if the label doesn't exist or isnt' true before patching it.

workflow/controller/taskresult.go
```diff
@@ -53,21 +62,100 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+				woc.log.Debugf("Determined controller shouldn't timeout %s", result.Name)
+			}
+		}
 		// If the task result is completed, set the state to true.
-		if result.Labels[common.LabelKeyReportOutputsCompleted] == "true" {
+		if label == "true" {
 			woc.log.Debugf("Marking task result complete %s", resultName)
 			woc.wf.Status.MarkTaskResultComplete(resultName)
-		} else {
+		} else if label != "true" && foundLabel {
```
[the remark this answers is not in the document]
isubasinghe: I didn't think it made sense to mark the task result as incomplete when the label used to determine status is missing.

This only ever should happen when an upgrade happens from 3.4.x to 3.5.y.

I've changed the else if branch to be more obvious, semantically it is equivalent.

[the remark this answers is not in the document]
Joibel: I feel it does make sense.

The workflow is waiting on this WorkflowTaskResult, and will not progress without it (unless this PR's main purpose kicks in). Storing that information in the status seems relevant and useful - we are waiting on it.

I don't look at it as an echo of the thing in the WorkflowTaskResults (why duplicate a piece of information) but rather "This workflow's current view of the world". This is a change from what the original author of this code wrote.

workflow/controller/taskresult.go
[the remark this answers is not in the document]
Joibel: This is still an expensive operation to perform, and for no-one's benefit that I can see.

This workflow will stop caring about this WorkflowTaskResult after this moment, and no-one else does. I'd rather it was deleted than patched. I'm pretty sure it'll get deleted fairly shortly after this anyway.

workflow/controller/taskresult.go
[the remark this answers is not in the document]
Joibel: We've discussed this and I understand @isubasinghe's reasoning now. The alternative is worse, so going with this.

Joibel (approved):

PR #12462, opened 2024-01-04
docs: update doc links to readthedocs/release-3.5

jmeridth:
Formerly #12460

Relates to #12414

### Modifications

- [x] updated docs links to https://argo-workflows.readthedocs.io/en/release-3.5
- [x] updated github issues config (like we did in main) to point to latest

PR #13455, opened 2024-08-12
fix(docs): Provide versioned links to documentation

Joibel:
Raised [in response to this](https://github.com/argoproj/argo-workflows/pull/13265#discussion_r1699400713)

## Motivation

Instead of always linking to latest, link to the versioned documentation where the version of the built software is known

## Modifications

Added a method to split the software version into components.

Changed help.`root` to provide a versioned link to documentation and modified all previous `const` values to be functions as this is now runtime determined. Making the whole thing `const` may be possible but the current encapsulation is nice and none of this is important for performance.

Removed the unused WorkflowTemplates doc links

## Verification

`make codegen` still generates the same link for `argo server`. Forcing the version to 3.5 generates expected 3.5 links.

### DIFF
--- a/pkg/apis/workflow/v1alpha1/version_types.go
+++ b/pkg/apis/workflow/v1alpha1/version_types.go
@@ -1,5 +1,10 @@
 package v1alpha1
 
+import (
+	"errors"
+	"regexp"
+)
+
 type Version struct {
 	Version      string `json:"version" protobuf:"bytes,1,opt,name=version"`
 	BuildDate    string `json:"buildDate" protobuf:"bytes,2,opt,name=buildDate"`
@@ -9,4 +14,17 @@ type Version struct {
 	GoVersion    string `json:"goVersion" protobuf:"bytes,6,opt,name=goVersion"`
 	Compiler     string `json:"compiler" protobuf:"bytes,7,opt,name=compiler"`
 	Platform     string `json:"platform" protobuf:"bytes,8,opt,name=platform"`
+}
+
+var verRe = regexp.MustCompile(`^v(\d+)\.(\d+)\.(\d+)`)
+
+// BrokenDown returns the major, minor and release components
+// of the version number, or error if this is not a release
+// The error path is considered "normal" in a non-release build.
+func (v Version) Components() (string, string, string, error) {
+	matches := verRe.FindStringSubmatch(v.Version)
+	if matches == nil || matches[1] == "0" {
+		return ``, ``, ``, errors.New("Not a formal release")
+	}
+	return matches[1], matches[2], matches[3], nil
 }
--- a/util/help/topics.go
+++ b/util/help/topics.go
@@ -1,13 +1,35 @@
 package help
 
-const (
-	root       = "https://argo-workflows.readthedocs.io/en/latest"
-	ArgoServer = root + "/argo-server/"
-	CLI        = root + "/cli/argo"
+import (
+	"fmt"
 
-	WorkflowTemplates                          = root + "/workflow-templates/"
-	WorkflowTemplatesReferencingOtherTemplates = WorkflowTemplates + "#referencing-other-workflowtemplates"
-
-	Scaling                        = root + "/scaling/"
-	ConfigureMaximumRecursionDepth = Scaling + "#maximum-recursion-depth"
+	"github.com/argoproj/argo-workflows/v3"
 )
+
+func root() string {
+	version := `latest`
+	if major, minor, _, err := argo.GetVersion().Components(); err == nil {
+		version = fmt.Sprintf("release-%s.%s", major, minor)
+	}
+	return fmt.Sprintf("https://argo-workflows.readthedocs.io/en/%s", version)
+}
+
+// ArgoServer returns a URL to the argo-server documentation
+func ArgoServer() string {
+	return root() + "/argo-server/"
+}
+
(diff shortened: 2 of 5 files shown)


pkg/apis/workflow/v1alpha1/version_types.go
```diff
@@ -10,3 +15,16 @@ type Version struct {
...
+var verRe = regexp.MustCompile(`^v(\d+)\.(\d+)\.(\d+)`)
+
+// BrokenDown returns the major, minor and release components
+// of the version number, or error if this is not a release
+// The error path is considered "normal" in a non-release build.
+func (v Version) Components() (string, string, string, error) {
+	matches := verRe.FindStringSubmatch(v.Version)
+	if matches == nil || matches[1] == "0" {
+		return ``, ``, ``, errors.New("Not a formal release")
+	}
```
isubasinghe, line 28: Sanity check this maybe? check length before access

isubasinghe (approved):

pkg/apis/workflow/v1alpha1/version_types.go
isubasinghe: Not too concerned, won't occur on an official release.

Joibel: We won't get a match (as in `matches` will be `nil`) unless there are exactly 3 matches.
I'm relying on go to short-circuit evaluation of the RHS of the `||` in this case.

isubasinghe: Ah right, okay

pkg/apis/workflow/v1alpha1/version_types.go
```diff
@@ -10,3 +15,16 @@ type Version struct {
 	Compiler     string `json:"compiler" protobuf:"bytes,7,opt,name=compiler"`
 	Platform     string `json:"platform" protobuf:"bytes,8,opt,name=platform"`
 }
+
+var verRe = regexp.MustCompile(`^v(\d+)\.(\d+)\.(\d+)`)
+
+// BrokenDown returns the major, minor and release components
```
agilgur5, line 21:
```suggestion
Components returns the major, minor, and patch components
```

the name is incorrect in the comment; also missing "patch". 
tbh though, I think having names in comments as this codebase does is an anti-pattern since it is redundant.

I might also suggest naming this `MajorMinorPatch` which would be more specific than `Components`

util/help/topics.go
```diff
@@ -1,13 +1,35 @@
 package help
 
-const (
-	root       = "https://argo-workflows.readthedocs.io/en/latest"
```
agilgur5, line 4: > Raised [in response to this](https://github.com/argoproj/argo-workflows/pull/13265#discussion_r1699400713)

For reference, this was versioned before as well -- each release branch changes the tail of its `root`. 
See also #12446, #12462, #12473.

That's why I changed the title to say "dynamically get", as before it was statically referenced. The scope is also for the CLI.

Note that per the above PRs, there are still several other places that use static version links that could be updated to use this dynamic variant instead.
Some won't be updateable as they're in static documentation (e.g. example YAML, swagger, direct README and other direct markdown as opposed to docs website rendered, etc). UI would also need similar code to pull out the version numbers to make its references dynamic

pkg/apis/workflow/v1alpha1/version_types.go
Joibel: This is a [godoc func description](https://tip.golang.org/doc/comment#func). Or it would be if I hadn't messed it up. I write other code that linter enforces these on exports and I think they are good.

I like `MajorMinorPatch`, it's better.

util/help/topics.go
Joibel: Yeah, this is this story: I forgot this existed at all when writing #13265, despite me having used it before in #11646 - so #13265 wasn't using this, and I wrote this function and pulled it out into this PR.

Anyway, I think we're in agreement this is a good change, so I'll fix up some of it in #13265.

pkg/apis/workflow/v1alpha1/version_types.go
agilgur5: Is it convention to start with the function name?
Odd, that feels like an anti-pattern to me since you usually read the function signature first, including in rendered godoc. But if that's how it is 🤷

Joibel: See the [link above](https://tip.golang.org/doc/comment#func), that's the rule.

Revive's [exported](https://github.com/mgechev/revive/blob/05c4801f9be1c3727e640c5c314cdf0694fce956/RULES_DESCRIPTIONS.md#exported) linter would complain if it didn't.

Joibel: I'm not misunderstanding you. 

On [that page](https://tip.golang.org/doc/comment) it specifies (can't link to the sentence)

> As with packages (above) and funcs (below), doc comments for types start with complete sentences naming the declared symbol.

Also of note is the first section on packages where it notes

> As can be seen in this example, Go doc comments use complete sentences.

agilgur5: Oh, you had linked to the "Functions" section before but this is actually outside of the functions section so I hadn't read that part 😅

Thanks for clarifying and direct quoting it so I could find and understand it!

PR #13459, opened 2024-08-12
refactor(tests): testifylint don't use `Equal` for floating point tests

Joibel:
Part of [the refactor](https://github.com/argoproj/argo-workflows/pulls?q=is%3Apr+testifylint+) started by #13365 to enable us to use testifylint, shift to using InEpsilon or InDelta for floating point comparisons. Mostly use `InEpsilon` as it is more resilient, but using `InDelta` for testing against zero.

Reviewers: As with the others in this series, please feel free to apply patches as you see fit.

### DIFF
--- a/util/env/env_test.go
+++ b/util/env/env_test.go
@@ -28,13 +28,13 @@ func TestLookupEnvIntOr(t *testing.T) {
 }
 
 func TestLookupEnvFloatOr(t *testing.T) {
-	assert.Equal(t, 1., LookupEnvFloatOr("", 1.), "default value")
+	assert.InEpsilon(t, 1., LookupEnvFloatOr("", 1.), 0.001, "default value")
 	t.Setenv("FOO", "not-float")
 	assert.Panics(t, func() { LookupEnvFloatOr("FOO", 1.) }, "bad value")
 	t.Setenv("FOO", "2.0")
-	assert.Equal(t, 2., LookupEnvFloatOr("FOO", 1.), "env var value")
+	assert.InEpsilon(t, 2., LookupEnvFloatOr("FOO", 1.), 0.001, "env var value")
 	t.Setenv("FOO", "")
-	assert.Equal(t, 1., LookupEnvFloatOr("FOO", 1.), "empty var value; default value")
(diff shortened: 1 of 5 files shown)


agilgur5 (approved):
Code LGTM, but one question: Does `testifylint` require this? For general floating point arithmetic, this makes sense, but it seems like all of the cases here do not involve large or long floats and so don't have the same chance of being inaccurate (and the tests passed before too)

agilgur5:
Mmm I'm going to suggest the same as I did in https://github.com/argoproj/argo-workflows/pull/13406#issuecomment-2254597304 that auto-merge before approval is an anti-pattern. I didn't even notice it was enabled here

Joibel:
> Code LGTM, but one question: Does `testifylint` require this? For general floating point arithmetic, this makes sense, but it seems like all of the cases here do not involve large or long floats and so don't have the same chance of being inaccurate (and the tests passed before too)

[Yes](https://github.com/Antonboom/testifylint?tab=readme-ov-file#float-compare) it does. Well, it is a default enabled linter which I agree with. It seems unlikely that rounding will affect the tests in question, but it felt better to leave the linter on.

> Mmm I'm going to suggest the same as I did in https://github.com/argoproj/argo-workflows/pull/13406#issuecomment-2254597304 that auto-merge before approval is an anti-pattern. I didn't even notice it was enabled here

Sorry, I missed that comment. I'll try to remember we're not using that here in future.

Joibel:
Also worth noting that most of these are removed by #13265 where I've moved metrics which make most sense as integers to be integers.

agilgur5:
> Well, it is a default enabled linter which I agree with. It seems unlikely that rounding will affect the tests in question, but it felt better to leave the linter on.

Yea I prefer to stick with the defaults generally as well (for the same reasons that `go fmt`, `prettier`, and `standard` exist), just wanted to make sure.

> Sorry, I missed that comment. I'll try to remember we're not using that here in future.

It's main use-case is to auto-merge once CI/tests are complete, so I think it only makes sense after a PR has been approved such that all changes (including to title and co-authors) are already in.

PR #13466, opened 2024-08-14
fix(controller): set `WatchListPageSize` when `WatchError`

### LINKED ISSUE #13461: Always set ResourceVersion="" may make etcd overload
### Pre-requisites

https://github.com/argoproj/argo-workflows/pull/11343
I noticed that in this pr, ResourceVersion= "" is permanently set in wfInformer, and I think this change will cause the etcd load to be too high.
see this part of code below, if ` ResourceVersion= ""` K8s apiserver will always request etcd directly.
`resourceVersion := opts.ResourceVersion
	pred := opts.Predicate
	pagingEnabled := utilfeature.DefaultFeatureGate.Enabled(features.APIListChunking)
	hasContinuation := pagingEnabled && len(pred.Continue) > 0
	hasLimit := pagingEnabled && pred.Limit > 0 && resourceVersion != "0"
	if resourceVersion == "" || hasContinuation || hasLimit || opts.ResourceVersionMatch == metav1.ResourceVersionMatchExact {
		// If resourceVersion is not specified, serve it from underlying
		// storage (for backward compatibility). If a continuation is
		// requested, serve it from the underlying storage as well.
		// Limits are only sent to storage when resourceVersion is non-zero
		// since the watch cache isn't able to perform continuations, and
		// limits are ignored when resource version is zero
		return c.storage.GetToList(ctx, key, opts, listObj)
	}
`

### Version(s)

3.4.9

```YAML
None
```

```text
None
```

```text
None
```

fyp711:
Fixes #13461 

### Motivation

If relist options , `case options.ResourceVersion != "" && options.ResourceVersion != "0": pager.PageSize = 0`. It's will set `options.Limit=0`, and than you will set   `ResourceVersion=''`. So when do relist, it lists all objects from etcd. 

Below code is a part of `reflector.go`.
```go
go func() {
defer func() {
		if r := recover(); r != nil {
			panicCh <- r
		}
	}()
	// Attempt to gather list in chunks, if supported by listerWatcher, if not, the first
	// list request will return the full response.
	pager := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
		return r.listerWatcher.List(opts)
	}))
	switch {
	case r.WatchListPageSize != 0:
		pager.PageSize = r.WatchListPageSize
	case r.paginatedResult:
		// We got a paginated result initially. Assume this resource and server honor
		// paging requests (i.e. watch cache is probably disabled) and leave the default
		// pager size set.
	case options.ResourceVersion != "" && options.ResourceVersion != "0":
		// User didn't explicitly request pagination.
		//
		// With ResourceVersion != "", we have a possibility to list from watch cache,
		// but we do that (for ResourceVersion != "0") only if Limit is unset.
		// To avoid thundering herd on etcd (e.g. on master upgrades), we explicitly
		// switch off pagination to force listing from watch cache (if enabled).
		// With the existing semantic of RV (result is at least as fresh as provided RV),
		// this is correct and doesn't lead to going back in time.
		//
		// We also don't turn off pagination for ResourceVersion="0", since watch cache
		// is ignoring Limit in that case anyway, and if watch cache is not enabled
		// we don't introduce regression.
		pager.PageSize = 0
	}

	list, paginatedResult, err = pager.List(context.Background(), options)
	if isExpiredError(err) || isTooLargeResourceVersionError(err) {
		r.setIsLastSyncResourceVersionUnavailable(true)
		// Retry immediately if the resource version used to list is unavailable.
		// The pager already falls back to full list if paginated list calls fail due to an "Expired" error on
		// continuation pages, but the pager might not be enabled, the full list might fail because the
		// resource version it is listing at is expired or the cache may not yet be synced to the provided
		// resource version. So we need to fallback to resourceVersion="" in all to recover and ensure
		// the reflector makes forward progress.
		list, paginatedResult, err = pager.List(context.Background(), metav1.ListOptions{ResourceVersion: r.relistResourceVersion()})
	}
	close(listCh)
		}()
```

see this part of code below, if ` ResourceVersion= ""` K8s apiserver will always request etcd directly.

```go
func (c *Cacher) GetToList(ctx context.Context, key string, opts storage.ListOptions, listObj runtime.Object) error {
        resourceVersion := opts.ResourceVersion
	pred := opts.Predicate
	pagingEnabled := utilfeature.DefaultFeatureGate.Enabled(features.APIListChunking)
	hasContinuation := pagingEnabled && len(pred.Continue) > 0
	hasLimit := pagingEnabled && pred.Limit > 0 && resourceVersion != "0"
	if resourceVersion == "" || hasContinuation || hasLimit || opts.ResourceVersionMatch == metav1.ResourceVersionMatchExact {
		// If resourceVersion is not specified, serve it from underlying
		// storage (for backward compatibility). If a continuation is
		// requested, serve it from the underlying storage as well.
		// Limits are only sent to storage when resourceVersion is non-zero
		// since the watch cache isn't able to perform continuations, and
		// limits are ignored when resource version is zero
		return c.storage.GetToList(ctx, key, opts, listObj)
	}
```

### Modifications

Add `Limit=500` when set `options.ResourceVersion = ""`
```go
       if options.Limit == 0 {
		options.Limit = 500
	}
```

### Verification
Wait informer watch failed，and then relist called. You can see that `options.Limit=0` in debug mode.


fyp711:
About pageSize is aligned with the default pageSize of pager in client-go. see https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/pager/pager.go#L31

workflow/common/common.go
```diff
@@ -260,6 +260,9 @@ const (
 	ArgoProgressPath = VarRunArgoPath + "/progress"
 
 	ConfigMapName = "workflow-controller-configmap"
+
+	// DefaultPageSize is the limit value of the request k8s apiserver
```
agilgur5, line 264:
```suggestion
	// DefaultPageSize is the limit value of the request to the k8s apiserver. Aligned with client-go default: https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/pager/pager.go#L31
```

Let's put your PR comment into the in-line comment

workflow/controller/informer/tolerant_cluster_workflow_template_informer.go
```diff
@@ -3,6 +3,8 @@ package informer
 import (
 	"time"
 
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
```
agilgur5, line 6: this should be in the same block as the other internal imports below (starting at line 15)

workflow/controller/informer/tolerant_workflow_template_informer.go
```diff
@@ -3,6 +3,8 @@ package informer
 import (
 	"time"
 
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
```
agilgur5, line 6: same as above

agilgur5 (commented):
Some stylistic comments below

agilgur5:
> Wait informer watch failed，and then relist called. You can see that `options.Limit=0` in debug mode.

this makes sense to me, although I'm not sure if this could have other ramifications/consequences. both during relist, which makes sense to be a full list, and during other operations as this code block is not necessarily specific to relist

Joibel:
@terrytangyuan - could you look at this. Given how #12133 fixed quite a subtle problem, it'd be good to have your input on this one.

terrytangyuan (commented):
Interesting discovery. cc @jessesuen to take a look as well

fyp711:
> Some stylistic comments below

Fixed, Thanks

workflow/controller/controller.go
```diff
@@ -915,7 +915,12 @@ func (wfc *WorkflowController) tweakListRequestListOptions(options *metav1.ListO
 	options.LabelSelector = labelSelector.String()
 	// `ResourceVersion=0` does not honor the `limit` in API calls, which results in making significant List calls
 	// without `limit`. For details, see https://github.com/argoproj/argo-workflows/pull/11343
-	options.ResourceVersion = ""
+	if options.ResourceVersion == "0" {
+		options.ResourceVersion = ""
+	}
```
fyp711, line 920: The recorded LastResourceVersion cannot be overwritten.  see https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/cache/reflector.go#L574

jessesuen (approved):
Thank you for the thorough explanation as well as the links to the client-go code to support the need for this change. The rationale for this PR makes sense to me.

Here is my summary of the original issue, the history, and this PR (if only for my own edification):
* The Kubernetes cache informer will call list with `resourceVersion=0`. `resourceVersion=0` tells the K8s API server that it is allowed to use its API server cache. However, it has the consequence of pagination (a.k.a limits/chunking) not working and returns **_all_** workflows in the response. This can lead to OOM issues in the controller in systems with a lot of workflows.
* To solve this, #11343 had blindly set `resourceVersion=""` in all list calls, in order to ensure that pagination would always work. This would prevent OOM situations in the controller by avoiding non-paginated huge LIST responses.
* However, as @fyp711 discovered, unconditionally setting `resourceVersion=""` is not correct, since the informer might also perform a "relist", which is a list call containing an actual numeric version, e.g. `resourceVersion=1234`. The fact that our controller overrides `1234` with `""` is incorrect behavior because ignoring `resourceVersion=1234` could cause consistency issues, but also causes the request to reach etcd.
* The fix here is to achieve both concerns:
    1. We will still ignore the cache informer's request to list with `resourceVersion=0`, replacing it with `resourceVersion=""`, since we know `0` will cause memory issues
    2. But we will now *honor* the cache informer's request to list with `resourceVersion=1234`. List with `resourceVersion=1234` *does* honor page sizes (I verified this), and we avoid potential consistency issues as well as strain on etcd (since  `resourceVersion=1234` allows the server to use its API server cache).

jessesuen:
The subtle differences between the list options that are worth repeating:

* `resourceVersion=&limit=500` - result is directly from etcd. limit is honored
* `resourceVersion=0&limit=500` - result from API server cache. limit is **_not_** honored (all objects will be returned)
* `resourceVersion=1234&limit=500` - result from API server cache. limit is honored

agilgur5:
Thanks for the great summary Jesse, that's very helpful!

My only remaining question is the effect/ramifications of the `limit=0` -> `limit=500` change here

terrytangyuan (commented):
Thank you! LGTM. Can we add the link to the PR to the code comments and record the rationale of the changes?

fyp711:
> Thank you! LGTM. Can we add the link to the PR to the code comments and record the rationale of the changes?

@terrytangyuan fixed, thanks.

terrytangyuan (approved):

workflow/controller/controller.go
```diff
@@ -915,7 +915,17 @@ func (wfc *WorkflowController) tweakListRequestListOptions(options *metav1.ListO
...
 	// without `limit`. For details, see https://github.com/argoproj/argo-workflows/pull/11343
-	options.ResourceVersion = ""
+	// The reflector will record `lastSyncResourceVersion`. If `ResourceVersion != "0"`, we should use the value
+	// recorded by the reflector. see https://github.com/argoproj/argo-workflows/pull/13466
+	if options.ResourceVersion == "0" {
+		options.ResourceVersion = ""
+	}
+	// The reflector will set the Limit to `0` when `ResourceVersion != "" && ResourceVersion != "0"`, which will fail
+	// to limit the number of workflow returns. Timeouts and other errors may occur when there are a lots of workflows.
+	// see https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/cache/reflector.go#L286
```
agilgur5, line 925: So in https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/cache/reflector.go#L289-L384 it says that no limit retrieves from watch cache and is intentionally done to avoid a thundering herd on etcd; so that sounds like it could have negative consequences?

workflow/controller/taskresult.go
```diff
@@ -33,7 +33,17 @@ func (wfc *WorkflowController) newWorkflowTaskResultInformer() cache.SharedIndex
...
+			// recorded by the reflector. see https://github.com/argoproj/argo-workflows/pull/13466
+			if options.ResourceVersion == "0" {
+				options.ResourceVersion = ""
+			}
+			// The reflector will set the Limit to `0` when `ResourceVersion != "" && ResourceVersion != "0"`, which will fail
+			// to limit the number of workflow returns. Timeouts and other errors may occur when there are a lots of workflows.
+			// see https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/cache/reflector.go#L286
+			if options.Limit == 0 {
+				options.Limit = common.DefaultPageSize
+			}
```
agilgur5, line 46: I'm thinking we may want to split this whole block into a common helper function that we use for all these informers so that it's consolidated and consistent and harder to accidentally change one and not another

workflow/controller/controller.go
fyp711: @agilgur5 Thank you for reading, let me share my opinion:
First of all, the purpose of this submission is to make the Limit take effect, in order to solve a series of problems caused by returning a large amount of workflow at once.
Next, the `ListAndWatch` method will only execute List once each time, and then start continuous Watch, so I think this operation will not cause too much load on etcd.
Summary: The ListWatch is not a frequent List request, Setting `Limit = xx` can ensure pagination return.

imliuda:
> 2\. List with `resourceVersion=1234` _does_ honor page sizes (I verified this), and we avoid potential consistency issues as well as strain on etcd (since  `resourceVersion=1234` allows the server to use its API server cache).


I checked kubernetes source, 1.30 and former versions may not honor this, but, set `resourceVersion=1234`  is still needed to ensure no events will be missed.

workflow/controller/taskresult.go
fyp711: Resolved, I added two new util methods.

fyp711: @agilgur5 Could you take a view again, thanks!

jessesuen:
> effect/ramifications of the limit=0 -> limit=500 change here

This check makes sure that we never perform a list request that wants all objects in the cluster, instead always ensuring that responses are paginated. Since the caller did not specify a limit, we set this to the client-go default of 500.

imliuda:
> > > 2. List with resourceVersion=1234 does honor page sizes (I verified this), and we avoid potential consistency issues as well as strain on etcd (since resourceVersion=1234 allows the server to use its API server cache).
> 
> > I checked kubernetes source, 1.30 and former versions may not honor this, but, set resourceVersion=1234 is still needed to ensure no events will be missed.
> 
> I just tested with v1.30, v1.29, v1.26. In all cases, when using a `resourceVersion` > 0 the `limit` parameter is honored.

Sorry, maybe I didn't express clearly, when `resourceVersion` and `limit` both set, the request will go through etcd, not apiserver cache, if kubernetes is 1.30 or former versions. Following code is from release-1.30.

image

fyp711:
@jessesuen @Joibel @agilgur5 @terrytangyuan @imliuda Unfortunately, I did a stress test and found some issues. The current way of setting the Limit is incorrect, which may result in only 500 records of informer data in certain situations, as shown below.
https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/pager/pager.go#L93C3-L110C2

The correct way is to set the WatchListPageSize of the reflector. Later I will submit a new fix. Thanks.
https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/cache/reflector.go#L99

workflow/controller/controller.go
agilgur5: Yea I get that the List portion is not as frequent (it will also happen during informer rebuilds), though part of the purpose of this PR is to reduce load on etcd, and this actually _causes_ a relist from etcd instead of the watch cache

agilgur5:
Thanks for stress testing this!

> The current way of setting the Limit is incorrect, which may result in only 500 records of informer data in certain situations, as shown below.
> https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/pager/pager.go#L93C3-L110C2

Yea this is what I was concerned about, that setting it unconditionally could break cache rebuilds or other times the informer decides to do a full relist, i.e. it can break the expected behavior of an informer. In higher load scenarios, the informer is likely to keep track of more than 500 records

agilgur5:
> informer rebuilds

> cache rebuilds

Apparently the informer framework's resync does not do a full re-list anymore per https://github.com/argoproj/gitops-engine/pull/617#discussion_r1698660165?

@wojtek-t if you have a minute to check this PR, your expertise could be very valuable here as well! 🙏

imliuda:
if got a `errors.IsResourceExpired()` error, the reflector will set `resourceVersion=0`, try to do a full list, but it will be updated to 500 by `util.CheckLimit(options)`,  then `pager.List()` only return 500 items.

```
// len(list) = 500, paginatedResult=false, err=nil
list, paginatedResult, err = pager.List(context.Background(), options)
```

workflow/controller/controller.go
```diff
@@ -1525,3 +1538,14 @@ func (wfc *WorkflowController) IsLeader() bool {
...
+func (wfc *WorkflowController) WatchErrorHandler(r *cache.Reflector, err error) {
+	cache.DefaultWatchErrorHandler(r, err)
+	if err != io.EOF {
+		// The reflector will set the Limit to `0` when `ResourceVersion != "" && ResourceVersion != "0"`, which will fail
+		// to limit the number of workflow returns. Timeouts and other errors may occur when there are a lots of workflows.
+		// see https://github.com/kubernetes/client-go/blob/ee1a5aaf793a9ace9c433f5fb26a19058ed5f37c/tools/cache/reflector.go#L286
+		r.WatchListPageSize = common.DefaultPageSize
+		wfc.WatchListPageSize = r.WatchListPageSize
+	}
+}
```
fyp711, line 1551: I added an error handler to set the `WatchListPageSize` of the `reflector` to ensure that it can meet the Limit restriction when an error occurs in `ListAndWatch`. 
Currently, I have only found this way to set `WatchListPageSize`. If anyone has a better solution, I am happy to discuss it with you.

fyp711:
@wojtek-t Could you help us review this pr ? We do believe that your experience can make this submission better, if you have free time. Thanks !

fyp711:
Hi @agilgur5 , I have updated a new fix solution, what do you think about this fix?

tooptoop4:
who gon merge dis

tooptoop4:
@fyp711 do u have a screenshot showing metrics graph of before vs after impact of this change?

fyp711:
> @fyp711 do u have a screenshot showing metrics graph of before vs after impact of this change?

Currently no, This submission is an optimization based on this https://github.com/argoproj/argo-workflows/pull/11343 
Optimize two key points: 
1. Only set to empty when resourceVersion = 0 
2. Set the `WatchListPageSize` property to ensure pagination

tooptoop4:
would be good to know how much this helps etcd

fyp711:
> would be good to know how much this helps etcd

I modified the description of the issue, the previous description was not very accurate, the previous discussion should be about another issue.

Issue #13471, opened 2024-08-15
proposal: change expression syntax to not overlap with Helm


isubasinghe:
# Summary

We currently support multiple templating/expression engines without any strong reason.
This issue attempts to formalize what the new expression syntax should look like.

# Suggestion

We support expr-lang as is, we feed an environment mapping containing all variables needed when compiling/running said expression.
The values in this map can be accessed by using the `$` prefix.

This will obviously be a breaking change, but hopefully will simplify a giant part of the expression/templating logic that exists in argo workflows and hopefully reduce user confusion in the process.

We can deprecate existing solutions in the next release, then  remove these dependencies entirely in the `next release + 1`.

## Before
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: hello-world-parameters-
spec:
  entrypoint: print-message
  arguments:
    parameters:
      - name: message
        value: hello world
  templates:
    - name: print-message
      inputs:
        parameters:
          - name: message
      container:
        image: busybox
        command: [ echo ]
        args: [ "{{inputs.parameters.message}}" ]
```
## After
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: hello-world-parameters-
spec:
  entrypoint: print-message
  arguments:
    parameters:
      - name: message
        value: hello world
  templates:
    - name: print-message
      inputs:
        parameters:
          - name: message
      container:
        image: busybox
        command: [ echo ]
        args: [ "$inputs.parameters.message" ]
```

isubasinghe:
Copying @agilgur5 's comment [here](https://github.com/argoproj/argo-workflows/issues/9529#issuecomment-1980011539)

> I've written about my plans to standardize on `expr` in a few other places ([#7576 (comment)](https://github.com/argoproj/argo-workflows/issues/7576#issuecomment-1728725234), #12694, [#12693 (comment)](https://github.com/argoproj/argo-workflows/issues/12693#issuecomment-1974873029), etc), figured this issue should really be the central source of truth for that and so will document here.
> 
>     1. Close all related/duplicate issues (did this some months ago)
> 
>     2. "Soft deprecate" `govaluate` by removing it from the docs and examples and replacing it with `expr` everywhere, per [Confusingly, `when` _kind of_ supports expr expressions #7576 (comment)](https://github.com/argoproj/argo-workflows/issues/7576#issuecomment-1728725234)
> 
[...]

Issue #9529, opened 2022-09-06
Fix `{{..}}` templating -- standardize on `expr`


alexec:
The current templating system appears to be the primary cause of confusion and bugs. These are caused by the following:

1. Template evaluation is unpredicable. By which I mean, you don't know when it'll evaluate and what the output will be. This can result in a workflow that passes just some of the time
2. We use several evaluation technologies: Govaluate, fasttemplate, Expr+Sprig.

Minor issues:

1. `{{..}}` does not play nicely with Helm templating. 
2. `podSpecPatch` is arguably a work-around rather than a solution.

It would be great to remove this entires class of issue.

To fix this we should:

* Standardize on `expr` as the templating technology (easy).
* Make evaluation time predicable - at field access time (hard).

agilgur5:
I've written about my plans to standardize on `expr` in a few other places (https://github.com/argoproj/argo-workflows/issues/7576#issuecomment-1728725234, https://github.com/argoproj/argo-workflows/issues/12694, https://github.com/argoproj/argo-workflows/issues/12693#issuecomment-1974873029, etc), figured this issue should really be the central source of truth for that and so will document here.

1. Close all related/duplicate issues (did this some months ago)
1. "Soft deprecate" `govaluate` by removing it from the docs and examples and replacing it with `expr` everywhere, per https://github.com/argoproj/argo-workflows/issues/7576#issuecomment-1728725234
1. "Hard" deprecate `govaluate` in 3.7 (as an example, 3.6 is the next minor, so if the docs are updated in 3.5's time, give it one more minor to settle in) by detecting its usage (or non-`expr` usage) and logging out a deprecation warning saying to switch to `expr`
1. Fully remove `govaluate` in 3.8 and error out when its usage is detected, saying to switch to `expr`

Per #7576, as `when` already supports `expr` (albeit with some confusing error messages), we can do this without a breaking spec change to the CRs

agilgur5:
> 1. `{{..}}` does not play nicely with Helm templating.

Syntax replacement has been spilt to #13471.

To be specific, this issue focuses on the `expr` standardization.

PR #13474, opened 2024-08-15
feat(cron): cronworkflows `when` clause

isubasinghe:
### Modifications

This supersedes https://github.com/argoproj/argo-workflows/pull/13365 allowing for the evaluation of expr-lang expressions before running a CronWorkflow. 

The scope of variables in this PR is quite limited, but can be  further extended in later PRs when variable scoping is properly defined/formalised.

### Verification

Added some tests for the evaluation function.


Joibel (changes requested):
This needs a documentation page.

I'd also like to bring across the documentation from #13365 regarding how to do the various things mentioned there but using this new technique. I'm happy to contribute that, but the basic docs need to be in here first I think.

agilgur5:
> This needs documenting, with examples and the variables we can use.

See also the ["Variables" page](https://argo-workflows.readthedocs.io/en/release-3.5/variables/)

workflow/cron/operator.go
```diff
@@ -178,6 +184,67 @@ func (woc *cronWfOperationCtx) patch(ctx context.Context, patch map[string]inter
...
+	}
+	m := make(map[string]interface{})
+	tm := time.Date(0, 1,
+		1, 0, 0, 0, 0, time.UTC)
+
+	m["lastScheduledTimeNull"] = false
+
+	if cron.Status.LastScheduledTime != nil {
+		tm = cron.Status.LastScheduledTime.Time
+		m["lastScheduledTimeNull"] = true
```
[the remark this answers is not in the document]
isubasinghe: Yup, thanks for catching this.

workflow/cron/operator.go
```diff
@@ -178,6 +184,67 @@ func (woc *cronWfOperationCtx) patch(ctx context.Context, patch map[string]inter
...
+	}
+
+	m["lastScheduledTime"] = tm
```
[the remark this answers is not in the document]
isubasinghe: Ack, I can do this

workflow/controller/operator.go
```diff
@@ -3526,14 +3526,18 @@ func processItem(tmpl template.Template, name string, index int, item wfv1.Item,
 		return "", errors.Errorf(errors.CodeBadRequest, "withItems[%d] expected string, number, list, or map. received: %v", index, item)
 	}
 	var newStepStr string
+	interReplaceMap := make(map[string]interface{})
```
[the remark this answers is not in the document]
isubasinghe: Yeah this should definitely happen

util/template/replace.go
```diff
@@ -15,8 +15,11 @@ func Replace(s string, replaceMap map[string]string, allowUnresolved bool) (stri
 	if err != nil {
 		return "", err
 	}
-
-	replacedString, err := t.Replace(replaceMap, allowUnresolved)
+	interReplaceMap := make(map[string]interface{})
```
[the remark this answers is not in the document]
isubasinghe: I did try this at first, it requires changing the code in quite a lot of places, with arguably no benefit for the generality given the current expressions only operate on strings everywhere else.

Given that we plan to change expressions/scoping in the future, I didn't want to pollute this PR with unnecessary code.

docs/variables.md
```diff
@@ -184,6 +184,18 @@ Only available for `successCondition`
 | `response.body` | Response body (`string`) |
 | `response.headers` | Response headers (`map[string][]string`) |
 
+### CronWorkflow Templates
+
+> v3.6 and after, note that the the variables here can be used with the full power of expr-lang expression.
```
agilgur5, line 189:
```suggestion
> v3.6 and after
```

This is not necessary, and all variables can be used with expressions, which the top of this page covers

docs/variables.md
```diff
@@ -184,6 +184,18 @@ Only available for `successCondition`
...
+
+| Variable | Description|
+|----------|------------|
+| `cronworkflows.name` | Name of the CronWorkflow method (`string`) |
```
agilgur5, line 193:
```suggestion
| `cronworkflow.name` | Name of the CronWorkflow (`string`) |
```

This should be singular, it only references a single CronWorkflow. That would also match the `workflow` variables

Also there isn't a "method" here?

docs/variables.md
```diff
@@ -184,6 +184,18 @@ Only available for `successCondition`
...
+| `cronworkflows.numActive` | Number of running CronWorkflows (`int`) |
```
agilgur5, line 194: I think you meant child Workflows
```suggestion
| `cronworkflow.numActive` | Number of running child Workflows (`int`) |
```

agilgur5: I'm also not really sure we should expose this, as it's accuracy will fluctuate and so it's not that reliable

docs/variables.md
```diff
@@ -184,6 +184,18 @@ Only available for `successCondition`
...
+| `cronworkflows.lastScheduledTimeNull` | When `cronworkflows.lastScheduledTime` is a sane value this is set to `false` else `true` (`bool`) |
+| `cronworkflows.lastScheduledTime` | The time since this workflow was last scheduled (`time.Time`) |
+| `cronworkflows.phase` | Phase is an enum of `Active` or `Stopped`. It changes to `Stopped` when `stopStrategy.condition` is true (`string`) |
```
agilgur5, line 197: Similarly I'm not sure what the value of exposing this is. It will always be "Active". It won't be evaluated when it is "Stopped".

docs/variables.md
```diff
@@ -184,6 +184,18 @@ Only available for `successCondition`
...
 
+### CronWorkflow Templates
+
+> v3.6 and after, note that the the variables here can be used with the full power of expr-lang expression.
+
+| Variable | Description|
+|----------|------------|
+| `cronworkflows.name` | Name of the CronWorkflow method (`string`) |
+| `cronworkflows.numActive` | Number of running CronWorkflows (`int`) |
+| `cronworkflows.lastScheduledTimeNull` | When `cronworkflows.lastScheduledTime` is a sane value this is set to `false` else `true` (`bool`) |
```
agilgur5, line 195: Can't we just make `lastScheduledTime` as `nil`?

This should also be simplified and avoid [assumption words](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding)

I would also recommend the substantially easier to understand and more conventional `hasLastScheduledTime` if truly necessary 

```suggestion
| `cronworkflows.hasLastScheduledTime` | `true` if `cronworkflows.lastScheduledTime` exists else `false` (`bool`) |
```

examples/cron-when-limit-concurrency.yaml
```diff
@@ -0,0 +1,21 @@
...
+# to control concurrency
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: sleep-when-limit-concurrency
+spec:
+  schedule: "* * * * *"
+  timezone: "America/Los_Angeles"   # Default to local machine timezone
+  startingDeadlineSeconds: 0
+  suspend: false                    # Set to "true" to suspend scheduling
```
agilgur5, line 11: the above 3 fields are not relevant to this CronWorkflow, so should be removed. Otherwise they show up in other places in the Fields Reference and can be confusing as an example

examples/cron-when.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: sleep-when
+spec:
+  schedule: "* * * * *"
+  timezone: "America/Los_Angeles"   # Default to local machine timezone
+  startingDeadlineSeconds: 0
+  suspend: false                    # Set to "true" to suspend scheduling
```
agilgur5, line 9: same as above

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -65,6 +65,9 @@ type CronWorkflowSpec struct {
 	StopStrategy *StopStrategy `json:"stopStrategy,omitempty" protobuf:"bytes,10,opt,name=stopStrategy"`
 	// Schedules is a list of schedules to run the Workflow in Cron format
 	Schedules []string `json:"schedules,omitempty" protobuf:"bytes,11,opt,name=schedules"`
+	// v3.6 and after: When clause can be used to determine a run should or shouldn't be scheduled.
+	// This new When clause allows for the full expressivity of expr-lang.
```
agilgur5, line 69:
```suggestion
```

This is similarly not necessary. You should just say it uses an expression, which is defined in the "Variables" page and linked to from various places as well

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -65,6 +65,9 @@ type CronWorkflowSpec struct {
 	StopStrategy *StopStrategy `json:"stopStrategy,omitempty" protobuf:"bytes,10,opt,name=stopStrategy"`
 	// Schedules is a list of schedules to run the Workflow in Cron format
 	Schedules []string `json:"schedules,omitempty" protobuf:"bytes,11,opt,name=schedules"`
+	// v3.6 and after: When clause can be used to determine a run should or shouldn't be scheduled.
```
agilgur5, line 68:
```suggestion
	// v3.6 and after: When is an expression that determines if a run should be scheduled.
```

- per above
- and simplify

docs/variables.md
```diff
@@ -184,6 +184,18 @@ Only available for `successCondition`
 | `response.body` | Response body (`string`) |
 | `response.headers` | Response headers (`map[string][]string`) |
 
+### CronWorkflow Templates
```
agilgur5, line 187:
```suggestion
### CronWorkflows 
```

This applies to all CronWorkflows.
Also "Templates" in this context refers to "template types", i.e. steps, DAGs, HTTP, etc

docs/variables.md
agilgur5: Also while the `status` does call it "Active", that's not a state we use externally (it also might be a legacy naming thing). "Running" is the term used, and I would recommend that here too

agilgur5 (changes requested):
Mainly reviewed the docs and examples below, but that also lends itself to some usage concerns I have about some of the variables

docs/variables.md
isubasinghe: > Can't we just make lastScheduledTime as nil in this case?

No we cannot really do this, but changed to `hasLastScheduledTime`

docs/variables.md
isubasinghe: Thanks, I fixed this

docs/variables.md
isubasinghe: Alright, I am happy to remove this.

docs/variables.md
isubasinghe: removed

docs/variables.md
```diff
@@ -184,6 +184,16 @@ Only available for `successCondition`
...
 | `response.headers` | Response headers (`map[string][]string`) |
 
+### CronWorkflows
+
+> v3.6 and after
+
+| Variable | Description|
+|----------|------------|
+| `cronworkflow.name` | Name of the CronWorkflow (`string`)
+| `cronworkflow.hasLastScheduledTime` | When `cronworkflows.lastScheduledTime` is a sane value this is set to `false` else `true` (`bool`) |
```
Joibel, line 194: This description looks like it is inverted from what I'd hope this variable contains.

docs/variables.md
```diff
@@ -184,6 +184,16 @@ Only available for `successCondition`
...
+| `cronworkflow.lastScheduledTime` | The time since this workflow was last scheduled (`time.Time`) |
```
Joibel, line 195: Document value when `hasLastScheduledTime` is false - because you can skip that check for some deltas with the implementation you've chosen - and I know you did this deliberately.

docs/variables.md
```diff
@@ -184,6 +184,16 @@ Only available for `successCondition`
 | `response.body` | Response body (`string`) |
 | `response.headers` | Response headers (`map[string][]string`) |
 
+### CronWorkflows
```
Joibel, line 187: I was expecting to be able to access namespace, annotations and labels like you can from `workflow`.

I don't know what for in this scenario, but just as general variables which would be in scope to mirror workflows

Joibel (changes requested):

docs/variables.md
agilgur5: Can you elaborate on why not?

docs/variables.md
agilgur5: Yes this missed the other changes I suggested [above](https://github.com/argoproj/argo-workflows/pull/13474#discussion_r1721277481) too

```suggestion
| `cronworkflow.hasLastScheduledTime` | `true` if `cronworkflows.lastScheduledTime` exists else `false` (`bool`) |
```

docs/variables.md
agilgur5: I might suggest leaving those out for now until there's more of a need from users (as I did with some of the previous ones Isitha now removed). Especially since the variables are mostly useful for `when` and we recommend using `workflowTemplateRef`, so there isn't much of a place to use them either

Joibel: I know I wrote "I don't know what for" - but actually now my use case would be the "minimum interval" one - I'd write an identical expression for `when` on all workflows, which pulled minimum interval from an annotation.

docs/variables.md
agilgur5: We discussed this in [last week's Contributor Meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.jh0wp5re8e6), and it seemed like this should be doable to set as `nil` as `expr` can handle that and there are existing Argo variables that can be `nil` or a zero value in certain scenarios.

My suggestion was to therefore leave `nil`, but in docs and examples, always show how to handle this scenario

docs/variables.md
agilgur5: That is a clever implementation, though I'd hazard a guess that it's not one that a majority of users would use 😅 
So "more of a need from users" would still apply

isubasinghe: @agilgur5 I personally see no harm in including labels and annotations, I would argue that I would actually expect them there since workflows has them.

docs/variables.md
isubasinghe: I've changed this such that this variable is no more.

docs/variables.md
agilgur5: > I would argue that I would actually expect them there since workflows has them.

I'm not sure that the `govaluate` version of `when` actually has this (though you can throw an `expr` expression anyway)

> without any negative consequences

Every feature is more things to maintain, have backward compat for, etc

util/template/replace.go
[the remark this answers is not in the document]
agilgur5: Expressions will always be strings -- since we're operating on YAML, they're only valid in strings. That's a hard limitation unless I'm missing something

We do have a few hacks to accept `IntOrString` in places that are normally integers (and which can be very confusing). There isn't really a way to workaround that in the schema otherwise though, since at submission time it is indeed a YAML string for an expression.

workflow/cron/operator.go
```diff
@@ -178,6 +184,89 @@ func (woc *cronWfOperationCtx) patch(ctx context.Context, patch map[string]inter
 	}
 }
 
+// TODO: refactor shouldExecute in steps.go
+func shouldExecute(when string) (bool, error) {
+	if when == "" {
+		return true, nil
+	}
+	expression, err := govaluate.NewEvaluableExpression(when)
```
agilgur5, line 192: oh? I thought we were using `expr` here given `govaluate`'s deprecation??

workflow/cron/operator.go
```diff
@@ -178,6 +184,89 @@ func (woc *cronWfOperationCtx) patch(ctx context.Context, patch map[string]inter
...
+	}
+
+	addSetField("name", cron.Name)
+	addSetField("namespace", cron.Namespace)
+	addSetField("labels", cron.Labels)
+	addSetField("annotations", cron.Labels)
+
+	labelsStr, err := json.Marshal(&cron.Labels)
+	if err != nil {
+		// We shouldn't hit this
```
agilgur5, line 230: I'd replace this with a lint ignore, similar to https://github.com/argoproj/argo-workflows/pull/12917, as that simplifies the code

docs/variables.md
Joibel: > That is a clever implementation, though I'd hazard a guess that it's not one that a majority of users would use 😅 So "more of a need from users" would still apply

This PR is addressing a user need, and we'd be suggesting this mechanism for them to implement their requirements if it were available :shrug:. So we'd like it to be available.

The alternative for another use of this when clause would be datetime literals in the when clause (more than once due to wanting to match day+month+year, which is also icky.

agilgur5: > It feels like a weird asymmetry, I'm not sure why the variable scopes should look different for the two types where they don't need to.

There will be asymmetry already; there are [several other variables available on Workflows](https://argo-workflows.readthedocs.io/en/release-3.5/variables/#global), so really the question is where to draw the line, not whether to make them entirely symmetrical.

> The alternative for another use of this when clause would be datetime literals in the when clause (more than once due to wanting to match day+month+year, which is also icky).

Doesn't the annotation approach also require a literal? I'm confused. You're only replacing the `interval` variable.
Or do you mean a different use-case entirely?
Also `expr` fortunately does have [`date` manipulation functions](https://expr-lang.org/docs/language-definition#date)

> and we'd be suggesting this mechanism for them to implement their requirements if it were available 🤷. So we'd like it to be available.

You'd be suggesting annotations for them? I guess that depends on the experience level of the user 🤔 ; annotations and template substitution can be pretty complicated for beginners to grasp, but for more advanced users that would make sense I suppose

But yea, if you'd be suggesting that, I suppose that's enough need. The point on symmetry still stands though

Joibel (approved):

docs/variables.md
Joibel: This would be something we'd be designing, and the user would just fill out the annotations.

agilgur5: Oh. Like a pre-filled template?
Huh I hadn't even thought of that tbh. That makes a lot more sense 😅 Thanks for elaborating that!

I see, CronWorkflows themselves don't have parameter input so you'd need to meta-template it. I'd probably personally end up doing that with Helm rather than raw (actually talked about this with Isitha last week, I did a lot of Helm -> WorkflowTemplate -> submit Workflow through API -> create CD Application -> Helm -> WorkflowTemplate at my last job, so it was _very_ meta. Understandably, that's not the route for everyone 😅)

workflow/cron/operator.go
```diff
@@ -7,20 +7,25 @@ import (
 	"sort"
 	"time"
 
+	"github.com/Knetic/govaluate"
```
terrytangyuan, line 10: Is this a new dependency?

workflow/cron/operator_test.go
```diff
@@ -27,7 +27,7 @@ var scheduledWf = `
     name: hello-world
     namespace: argo
     resourceVersion: "53389"
-    selfLink: /apis/argoproj.io/v1alpha1/namespaces/argo/cronworkflows/hello-world
+    selfLink: /apis/argoproj.io/v1alpha1/namespaces/argo/cronworkflow.hello-world
```
terrytangyuan, line 30: Is this intentional?

workflow/cron/operator_test.go
```diff
@@ -386,8 +386,8 @@ func TestMissedScheduleAfterCronScheduleWithForbid(t *testing.T) {
 	var cronWf v1alpha1.CronWorkflow
 	v1alpha1.MustUnmarshal([]byte(forbidMissedSchedule), &cronWf)
 	// StartingDeadlineSeconds is after the current second, so cron should be run
-	//startingDeadlineSeconds := int64(35)
-	//cronWf.Spec.StartingDeadlineSeconds = &startingDeadlineSeconds
+	// startingDeadlineSeconds := int64(35)
+	// cronWf.Spec.StartingDeadlineSeconds = &startingDeadlineSeconds
 	t.Run("ForbiddenWithMissedScheduleAfterCron", func(t *testing.T) {
```
terrytangyuan, line 391: Do we still need these?

workflow/cron/operator_test.go
```diff
@@ -339,7 +339,7 @@ var forbidMissedSchedule = `apiVersion: argoproj.io/v1alpha1
 kind: CronWorkflow
 metadata:
   annotations:
-    cronworkflows.argoproj.io/last-used-schedule: CRON_TZ=America/Los_Angeles 0-36/1
+    cronworkflow.argoproj.io/last-used-schedule: CRON_TZ=America/Los_Angeles 0-36/1
```
terrytangyuan, line 342: Would this be a breaking change if users are already using it?

workflow/cron/operator_test.go
isubasinghe: This just formatted this way, removing now, good catch.

workflow/cron/operator_test.go
isubasinghe: Formatting issue I think

workflow/cron/operator.go
isubasinghe: Not yet, we are using the exact logic for how When is evaluated in hooks as well.

workflow/cron/operator.go
isubasinghe: No, its how When is evaluated in hooks as well

agilgur5: No, it's an existing dep. But it is a deprecated dep as I wrote [above](https://github.com/argoproj/argo-workflows/pull/13474#discussion_r1732030191)

workflow/cron/operator.go
isubasinghe: Sure I can do this.

isubasinghe: Actually we shouldn't do this, everywhere we Marshal we still error check.

JPZ13:
Hey @agilgur5 - Hope you had a good holiday weekend. I believe all of the changes you requested have been updated. I might have missed one, so please verify that. Could you remove the change request or request more changes so that we can keep moving forward on this? We'd like to roll out the `when` clause soon and recommend it to some of our customers

isubasinghe:
Note: This uses the deprecated `govaluate` library. This is very intentional and done to keep the logic similiar to the other When fields. Initially I implemented this using purely `expr` but after a chat with @Joibel we decided it was best to keep the logic as similiar as possible.

This field is intended to be deprecated and replaced with a `whenExpression` field in the future.

I will create another issue to unify the logic used in the `When` fields and to also use `expr` there instead.

workflow/cron/operator.go
agilgur5: Then the comment should be removed -- it's really one or the other

workflow/cron/operator.go
agilgur5: See follow-ups [below](https://github.com/argoproj/argo-workflows/pull/13474#issuecomment-2327932625)

agilgur5:
For context, Isitha followed up with me on Slack and I repeated my concern about using deprecated `govaluate` [above](https://github.com/argoproj/argo-workflows/pull/13474#discussion_r1732030191). The examples here do use embedded `expr` though, which is part of the "soft" deprecation I outlined in https://github.com/argoproj/argo-workflows/issues/9529#issuecomment-1980011539 / https://github.com/argoproj/argo-workflows/issues/7576#issuecomment-1728725234. However, it's still concerning to add a new feature using a deprecated feature / library.
I said that such usage actually makes me more strongly consider `whenExpression` of https://github.com/argoproj/argo-workflows/issues/7831, which I left open as an option, but isn't part of the same plan linked above.

Isitha explained that he talked to Alan and decided to use `whenExpression` instead, to which I said there was no context for in the PR, which Isitha then added above.
With no rationale, adding a field that uses deprecated `govaluate` goes directly against Isitha, mine, and Alex C's stated intents in #13471 and #9529

> This field is intended to be deprecated and replaced with a `whenExpression` field in the future.

This is an option I left open, but is not something that had been previously discussed -- it should be.

Also I would release a new field that uses a deprecated library and itself will be shortly deprecated.
If anything, if we want to keep the logic identical -- we should just release this as `whenExpression` and skip creating a new field that will be deprecated for CronWorkflows

JPZ13:
Thank you for the feedback @agilgur5. It's a bit hard for me to parse through the entire comment given I haven't been following all of the discussions and am strapped for time. Could you concisely give us a couple bullet points of:
- where you feel there has been an inappropriate closed discussion, and
- examples of a technical solution you would consider to be acceptable?

Our goal is to maintain consensus, keep our customers on mainline Argo Workflows expressions, and move quickly but thoroughly. I think working off of more concrete examples of what a `when` or `whenExpression` will look like will help move us forward

agilgur5:
If Terry and others would like to tiebreak and allow adding a deprecated feature, I'm fine with that, but I think that should be properly discussed. In particular, I think everyone in the discussion needs to correctly understand the context of `govaluate` -- that it is _only_ used in `when` which is itself the confusing outlier; everything else uses `expr` (including embedded in `when` itself), which is better documented and understood. We shouldn't be making decisions with incorrect data.

tooptoop4:
the way i see it perfection should not be enemy of good enough. if someone has good enough working 'now' it should be released, after all we are volunteers and blocking progress unless its perfect will lead to less contributions

agilgur5:
> if someone has good enough working 'now' it should be released

If we were to go with that principle, Isitha's first version was in fact written with `expr`

Also I wouldn't agree with this principle in general; case in point being Boeing, or all the bugs with 3.5, which you had argued to revert

There are different bars for different problems and oversimplification of nuances does not often result in good decisions.

workflow/cron/operator.go
isubasinghe: Will remove the comment in that case, we should be consistent in logic even if we expect marshalling to always  succeed.

agilgur5: If it has a comment, I don't think we need consistency, given that's what the comment explains. The lint ignores I wrote in #12917 are followed by comments (as is best practice for any lint ignore)

terrytangyuan:
>It's already widely adopted in the code base and the syntax is familiar to existing users.

What I meant is that, `when` uses `govaluate` and it's widely used by existing users before more advanced expression became available. I didn't mean that it's used more frequently in the code base.

terrytangyuan:
>the way i see it perfection should not be enemy of good enough. if someone has good enough working 'now' it should be released, after all we are volunteers and blocking progress unless its perfect will lead to less contributions

@tooptoop4 Love what you said here. I don't know who you are since you don't have any information in your profile but I really appreciate your involvement in the project and helpful feedback you've provided so far.

agilgur5:
> it's widely used by existing users before more advanced expression became available

Just so we're on the same page, that is before 3.1: https://github.com/argoproj/argo-workflows/pull/5115
Possibly longer for `expression` fields. Like it's pretty old (and could have been deprecated years ago as such, it just never was).

Modern users are much more familiar with `expr` given its docs and how frequently it's used compared to `govaluate`'s one place.

agilgur5:
Should the templating syntax and docs here be unified with `stopStrategy` of #12305 ?

For instance, the variables there may make sense to prefix with `cronworkflow.`. and the docs could be moved to a subheading under this PR's docs

terrytangyuan (approved):

terrytangyuan:
Thanks @isubasinghe! This is good to go. We can continue discuss migration path and work on consistency in a follow-up issue. @isubasinghe @Joibel Could any of you create a follow-up issue?

agilgur5:
> @isubasinghe @Joibel Could any of you create a follow-up issue?

Regarding `when` -> `whenExpression`, I would propose we add to the existing #7831.



> Should the templating syntax and docs here be unified with `stopStrategy` of #12305 ?
> 
> For instance, the variables there may make sense to prefix with `cronworkflow.`. and the docs could be moved to a subheading under this PR's docs

This might require a separate issue -- we should get any of these variable name consistency changes in before 3.6.0

tooptoop4:
@isubasinghe could this be used to run first monday of the month or last friday of the month?

agilgur5:
You can use arbitrary expressions, so you might be able to workaround it for https://github.com/argoproj/argo-workflows/issues/8348 with pure date/time manipulation.

For instance, if you use [`now().Day()`](https://expr-lang.org/docs/language-definition#date-functions), you can get the day of the month (as in today is "the 18th").
So if you have a `schedule` weekly on Fridays and a `when` with `now().Day() > 23`, that might get you close (day of the month is a bit coarse to use ofc)

isubasinghe:
You can also get the day it seems in expr-lang: https://expr-lang.org/docs/language-definition#date-functions
So you can probably use @agilgur5 's solution to get this behaviour @tooptoop4 .

Probably a little bit hacky, but dates are always at least a little bit hacky :)

isubasinghe:
Perhaps another feature could be  stored functions to help with more complex expressions, that way you could reference these stored functions instead of having to write your expressions inline.

agilgur5:
> > Should the templating syntax and docs here be unified with `stopStrategy` of #12305 ?
> > For instance, the variables there may make sense to prefix with `cronworkflow.`. and the docs could be moved to a subheading under this PR's docs
> 
> This might require a separate issue -- we should get any of these variable name consistency changes in before 3.6.0

Alan and I discussed this in the [Oct 1st Contributor Meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.gfymrsa8jhi7) and agreed on this. See #13702 as a follow-up

agilgur5:
> I'd also like to bring across the documentation from #13365 regarding how to do the various things mentioned there but using this new technique. I'm happy to contribute that, but the basic docs need to be in here first I think.

This has been added in #13723

PR #13476, opened 2024-08-16
feat: support kubernetes equality-based ops for field selector metadata.name

### LINKED ISSUE #13468: Field selectors should support all native k8s operators
# Summary

What change needs making?

Currently in the `field-selector` function there are a lot of limitations compared to native `kubectl`.

https://github.com/argoproj/argo-workflows/blob/89f2ed1d041d2f555d646ee02b77c7e1dc2bda10/server/utils/list_options.go#L77

For instance, `metadata.name` or `metadata.namespace` only work with `=` and are converted into a simple name lookup.

The k8s package as far as I know already allows for a similar approach as `labels.Requirements`, where you can use the full range of kubectl `field-selectors`.

https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

The function should be updated to allow for instance the usage of operators like `!=`

## Use Cases

When would you use this?

Currently we want to use the argo cli to stop a list of ongoing workflows using a combination of criteria such as:

`argo stop --field-selector "metadata.name!=workflow_name" -l workflows.argoproj.io/workflow-template!=some_template"`

If we omit the `--field-selector` flag this works flawlessly, with it it returns the following error:

`FATA[2024-08-14T15:36:43.802Z] rpc error: code = InvalidArgument desc = unsupported requirement metadata.name!=workflow_name`

---

miltalex:
Partially Fixes #13468

### DIFF
--- a/persist/sqldb/selector.go
+++ b/persist/sqldb/selector.go
@@ -9,7 +9,7 @@ import (
 func BuildArchivedWorkflowSelector(selector db.Selector, tableName, labelTableName string, t dbType, options utils.ListOptions, count bool) (db.Selector, error) {
 	selector = selector.
 		And(namespaceEqual(options.Namespace)).
-		And(nameEqual(options.Name)).
+		And(nameOperator(options.Name, options.NameOperator)).
 		And(namePrefixClause(options.NamePrefix)).
 		And(startedAtFromClause(options.MinStartedAt)).
 		And(startedAtToClause(options.MaxStartedAt))
--- a/persist/sqldb/workflow_archive.go
+++ b/persist/sqldb/workflow_archive.go
@@ -284,6 +284,25 @@ func nameEqual(name string) db.Cond {
 	return db.Cond{}
 }
 
+func nameNotEqual(name string) db.Cond {
+	if name != "" {
+		return db.Cond{"name !=": name}
+	}
+	return db.Cond{}
+}
+
+func nameOperator(name, op string) db.Cond {
+	if name != "" {
+		switch op {
+		case "!=":
+			return nameNotEqual(name)
+		case "=", "==", "":
+			return nameEqual(name)
+		}
+	}
+	return db.Cond{}
+}
+
 func namePrefixClause(namePrefix string) db.Cond {
 	if namePrefix != "" {
 		return db.Cond{"name LIKE": namePrefix + "%"}
@@ -389,7 +408,6 @@ func (r *workflowArchive) GetWorkflowForEstimator(namespace string, requirements
 			FinishedAt: v1.Time{Time: awf.FinishedAt},
 		},
 	}, nil
-
 }
 
 func (r *workflowArchive) DeleteWorkflow(uid string) error {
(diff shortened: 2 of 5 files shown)


Joibel:
Is there any reason this couldn't also be implemented for namespaces, it feels nearly identical?

miltalex:
> Is there any reason this couldn't also be implemented for namespaces, it feels nearly identical?

I noticed that for the namespace we need to adapt more codepaths. Mostly right now, we use the to validate access to the namespace. For the `!=` we need to adapt it in a way that we query all the namespaces and if there is an error we return it (I have a stash as a `WIP`)
```
// verify if we have permission to list Workflows
allowed, err := auth.CanI(ctx, "list", workflow.WorkflowPlural, options.Namespace, "")
```
Apart from that there are the `stores` that need adapting for example which I didn't dive into cause it affects the clients as well.
```
wfList, err := k.wfClient.ArgoprojV1alpha1().Workflows(namespace).List(ctx, listOptions)
``` 
For the `sqlite_store` I need to adapt the `BuildWorkflowSelector` .  As well the `clusterManagedNamespaceAndInstanceID` func for archived workflows. Those are the things I have noted until now, there might be cases that I have missed.

Since the namespace is touching a lot more components I thought I could split the MRs to make it easier for testing review etc

Please let me know @Joibel if I am mistaken or confused smth 🚀

Joibel:
Thanks for your comprehensive answer, I hadn't realised `namespace` would be so far reaching. The separate PRs make sense.

tooptoop4:
@miltalex i guess this needs some changes since https://github.com/argoproj/argo-workflows/pull/13160 has been merged. From what I can tell https://github.com/argoproj/argo-workflows/pull/13476 mentions != operator which other PR does not have

miltalex:
@tooptoop4 yes I will spend some time this weekend. I will rebase and check what needs to be adjusted

.features/pending/operators-field-selector.md
```diff
@@ -0,0 +1,17 @@
...
+Additional details about the feature written in markdown, aimed at users who want to learn about it
+* Explain when you would want to use the feature
+* Include code examples if applicable
+  * Provide working examples
+  * Format code using back-ticks
+* Use Kubernetes style
+* One sentence per line of markdown
+-->
+
+Currently, our system only supports the '=' operator for the field selector metadata.name when interacting with Kubernetes resources. This limitation restricts the flexibility and functionality of our queries. The goal of this PR is to expand support to include the '==' and '!=' operators, aligning our capabilities with native Kubernetes functionality. This enhancement will provide users with more granular control over resource selection and filtering, improving overall system usability and compatibility with standard Kubernetes practices.
```
Joibel: This is a note for people upgrading to the new feature, so write it from that point of view.

Don't tell me what it used to be like, nor mention `PR` because this is for people reading the release notes. Just tell me what the new feature provides.

miltalex: Thanks for the suggestion, I did a rewrite on the feature description

Joibel (approved):
Thanks for persevering with this

PR #13480, opened 2024-08-16
feat(ui): allow relative `links`. Fixes #13479

### LINKED ISSUE #13479: Allow Argo UI links to be relative
> Would this be possible (so that domain does not need to be configured in each link)?
> 
> ```yaml
> customLinks:
>   - url: /workflows?label=my.custom.label%foo
>     name: My Custom Label
> ```

@jessesuen For the custom links, we extended the existing links feature: (https://argoproj.github.io/argo-workflows/links/) by adding another scope `workflow-list`, and for now `links` feature doesn't support that, but I think it should be possible to implement it.

_Originally posted by @jiachengxu in https://github.com/argoproj/argo-workflows/issues/10447#issuecomment-1478854972_

cardoe:
Added the ability for users to supply relative links within the UI for the links feature so that buttons and links do not have to have the absolute URL baked into each link. This will allow for easily adding buttons for different filters on the workflow list page for example. fixes #13479 ref #10447

## Summary by CodeRabbit

- **New Features**
  - UI links can now use relative URLs, including namespace-relative paths and query-only URLs.
  - Relative links are automatically resolved against the base UI URL, while absolute links continue to work unchanged.

- **Documentation**
  - Added configuration guidance and examples for using relative UI links.

### DIFF
--- a/docs/links.md
+++ b/docs/links.md
@@ -36,3 +36,7 @@ For example, one may find it useful to define a custom label in the workflow and
 
 We can also access workflow fields in a pod link. For example, `${workflow.metadata.name}` returns
 the name of the workflow instead of the name of the pod. If the field doesn't exist on the workflow then the value will be an empty string.
+
+> v3.6 and after
+
+Links can be relative when configured and the UI will prefix them with the base UI URL.
--- a/docs/links.md
+++ b/docs/links.md
@@ -36,3 +36,16 @@ For example, one may find it useful to define a custom label in the workflow and
 
 We can also access workflow fields in a pod link. For example, `${workflow.metadata.name}` returns
 the name of the workflow instead of the name of the pod. If the field doesn't exist on the workflow then the value will be an empty string.
+
+> v4.2 and after
+
+You can configure a link with a relative URL instead of an absolute one.
+The UI prefixes a relative link with the base UI URL, so you do not have to hard-code the server's address in each link.
+This is convenient for links that point back into the UI, such as a button that opens the workflow list with a preset filter:
+
+```yaml
+links:
+  - name: Failed workflows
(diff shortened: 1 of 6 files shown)


cardoe:
@jessesuen this should address your ask in #10447

docs/links.md
```diff
@@ -36,3 +36,7 @@ For example, one may find it useful to define a custom label in the workflow and
 
 We can also access workflow fields in a pod link. For example, `${workflow.metadata.name}` returns
 the name of the workflow instead of the name of the pod. If the field doesn't exist on the workflow then the value will be an empty string.
+
+> v3.6 and after
```
tico24: v3.6 shipped a while back. Wouldn't this land in 4.1 now?

docs/links.md
```diff
@@ -36,3 +36,7 @@ For example, one may find it useful to define a custom label in the workflow and
...
+
+Links can be relative when configured and the UI will prefix them with the base UI URL.
```
tico24: `uiUrl()` puts the link straight after `<base href>`, so the `/my-namespace` in the test comes out as `//my-namespace`, unless I'm misreading it. Why would I use a relative link over an absolute one, and could you show one? Both asked for in the [docs guide](https://argo-workflows.readthedocs.io/en/latest/doc-changes/).

tico24: Small one: "will prefix" is future tense and "when configured" is passive, both in the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#patterns-to-avoid).

tico24 (commented):
Docs review only, the Go and UI files are for @argoproj/argo-workflows-approvers.

Issue #5862, opened 2021-05-07
Custom artifact repository plugins


jalberti:
# Summary

There are several Github issues that discuss volumes and how they can and do relate to artifact management, e.g. https://github.com/argoproj/argo-workflows/issues/1024, https://github.com/argoproj/argo-workflows/issues/1349. Today the supported artifact repository types are fixed, and extensions are only feasible in-tree, if I understand this correctly? I’m wondering if there were any specific thoughts on how CSI could be used, or a CSI like concept, to allow artifact repositories to be provided as plugins?

The problem at the moment is, unless I overlooked something, that while I can get volumes into the pod, into my workflow, and I can also write arbitrary code, and can even have side cars as demonstrated here https://github.com/argoproj/argo-workflows/issues/4988, ... if my input or output does not fit into the standard supported artifact repository set, I'm not able to use the concept of artifact declaration. Essential the "declaration" of "input/output" is lost. If one would be able to bring their own repository, then we can use the concept of artifacts together with custom code, i.e. we would have extensibility and declarative artifacts.

E.g. we could define a stable JSON-RPC interface, and allow community provided artifact repositories in form of a side car container? The artifact repository provider would create the container image with the JSON-RPC server, Argo would act as the JSON-RPC client. The Argo workflow would define an artifact repository of type jsonrpc, the artifact repository driver container and the workflow step container would share a common volumeMount such that artifact content does not need to pass the json-rpc pipe.

# Use Cases

This would be useful to add support for custom community supported artifact repositories.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

alexec:
This would fix the #1540 support.

Joibel:
Here is my proposal to implement this.

# Core Technical Constraints

The plugin architecture must operate within Argo Workflows’ execution model where short-lived pods handle workflow steps, requiring instantaneous startup without runtime dependency resolution.
Init containers must complete artifact downloads before main container execution begins.
Plugins should not utilize shared volumes for artifact transfer due to potential multi-gigabyte file sizes and lack of guaranteed shared storage availability for this.

A single workflow pod can access and store multiple artifacts which can use different artifact drivers for each artifact for both inputs and outputs.

# Where is the artifact code used:

- argo-server (argocli image): ArtifactServer (server/artifacts/artifact_server.go)
- argoexec: WorkflowExecutor (workflow/executor/executor.go)
[...]

PR #13483, opened 2024-08-19
feat: support data transfer protection in HDFS artifacts (#13482)

### LINKED ISSUE #13482: Support data transfer protection in HDFS artifacts
# Summary

What change needs making?

**AS-IS:**
An error occurs when connecting to an HDFS cluster with `dfs.data.transfer.protection` enabled. The error message is `read: connection reset by peer`, similar to the issue described in [this issue](https://github.com/colinmarc/hdfs/issues/184).

**TO-BE:**
The HDFS client supports the data transfer protection option, which can be configured to `authentication`, `integrity`, `privacy`, or left empty, as detailed in [this pull request](https://github.com/colinmarc/hdfs/pull/236). With this configuration, the issue described above is resolved.

This feature is available in versions v2.4.0, v2.3.0, v2.2.1, and v2.2.0 (see [commit](https://github.com/colinmarc/hdfs/commit/1596ee1307eb423a2ed092713f1368157d67ce86)). Since we are using version v2.4.0 of the HDFS client (as specified [here](https://github.com/argoproj/argo-workflows/blob/main/go.mod#L18)), we can utilize this feature.

## Use Cases

When would you use this?

As shown in https://github.com/argoproj/argo-workflows/blob/main/examples/hdfs-artifact.yaml, I am working on storing artifacts in HDFS. In my environment, the HDFS cluster has `dfs.data.transfer.protection` set to `authentication`. When I run the example, I encounter a `read: connection reset by peer` error.

However, with the enhancement, the operation would succeed by properly configuring the `dfs.data.transfer.protection` setting.

---

Looka149:
Fixes #13482

### Motivation

As shown in https://github.com/argoproj/argo-workflows/blob/main/examples/hdfs-artifact.yaml, I am working on storing artifacts in HDFS. In my environment, the HDFS cluster has `dfs.data.transfer.protection` set to `authentication`. When I run the example, I encounter a `read: connection reset by peer` error.

### Modifications

The HDFS client supports the data transfer protection option, which can be configured to `authentication`, `integrity`, `privacy`, or left empty, as detailed in [this pull request](https://github.com/colinmarc/hdfs/pull/236). With this configuration, the issue described above is resolved.

This feature is available in versions v2.4.0, v2.3.0, v2.2.1, and v2.2.0 (see [commit](https://github.com/colinmarc/hdfs/commit/1596ee1307eb423a2ed092713f1368157d67ce86)). Since we are using version v2.4.0 of the HDFS client (as specified [here](https://github.com/argoproj/argo-workflows/blob/main/go.mod#L18)), we can utilize this feature.

### Verification

Referring to the [hdfs-artifact.yaml example](https://github.com/argoproj/argo-workflows/blob/main/examples/hdfs-artifact.yaml), you can specify the value of `artifacts.hdfs.dataTransferProtection` as 'authentication', 'integrity', 'privacy', or leave it empty.

```yaml
  - name: print-message
    inputs:
      artifacts:
      - name: message
        path: /tmp/message
        hdfs:
          dataTransferProtection: authentication
```

If an incorrect value is provided, the following error will occur.

```bash
$ kubectl get workflow
NAME                  STATUS   AGE     MESSAGE
hdfs-artifact-td82h   Failed   33m     invalid spec: templates.artifact-example.steps[1].consume-artifact templates.print-message.inputs.artifacts.message.hdfs.dataTransferProtection must be one of `authentication`, `integrity`, `privacy`, or empty
```


Looka149:
@terrytangyuan @sarabala1979 
Hi, is there anything else needed to proceed with the approval? Thanks.

workflow/artifacts/hdfs/hdfs.go
```diff
@@ -73,6 +80,14 @@ func ValidateArtifact(errPrefix string, art *wfv1.HDFSArtifact) error {
 	if hasKrbCCache && (art.KrbServicePrincipalName == "" || art.KrbConfigConfigMap == nil) {
 		return errors.Errorf(errors.CodeBadRequest, "%s.krbServicePrincipalName and %s.krbConfigConfigMap are required with %s.krbCCacheSecret", errPrefix, errPrefix, errPrefix)
 	}
+
+	switch art.DataTransferProtection {
+	case "", DataTransferProtectionAuthentication, DataTransferProtectionIntegrity, DataTransferProtectionPrivacy:
+		// Do nothing
+	default:
+		return errors.Errorf(errors.CodeBadRequest, "%s.dataTransferProtection must be one of 'authentication', 'integrity', 'privacy', or empty", errPrefix)
+	}
```
juliev0, line 89: Thanks for adding the validation here, although I'm torn on whether we should validate this or not. On one hand, if the `ClientOptions` ever change for `DataTransferProtection` then we'd have to update this when the version in go.mod gets updated.

juliev0:
@Looka149 Your PR looks good. I can merge it. Only thing I'm thinking is that it may be better not to validate on these specific strings which could theoretically change, per my [comment](https://github.com/argoproj/argo-workflows/pull/13483#discussion_r1742878461), but I may be convinced otherwise. Thanks.

workflow/artifacts/hdfs/hdfs.go
Looka149: I agree that your concern is valid. I initially added validation based on the three modes specified in https://github.com/colinmarc/hdfs/blob/master/client.go#L24-L28. While I don't expect these options to change frequently, **I agree that if they do change in the future, this validation might be easily overlooked.**

Looka149:
@juliev0 Thank you for reviewing my PR, I agreed with your feedback and have removed the validation. Thank you.

juliev0:
@Looka149 assuming the CI failures aren't related to your change, would you mind adding [empty commits ](https://www.freecodecamp.org/news/how-to-push-an-empty-commit-with-git/) until the CI passes?

Looka149:
@juliev0 I have merged the main branch, and the CI tests have completed successfully.

juliev0 (approved):

workflow/artifacts/hdfs/util.go
```diff
@@ -8,9 +8,10 @@ import (
 	"github.com/jcmturner/gokrb5/v8/config"
 )
 
-func createHDFSClient(addresses []string, user string, krbOptions *KrbOptions) (*hdfs.Client, error) {
+func createHDFSClient(addresses []string, user string, dataTransferProtection string, krbOptions *KrbOptions) (*hdfs.Client, error) {
 	options := hdfs.ClientOptions{
-		Addresses: addresses,
+		Addresses:              addresses,
+		DataTransferProtection: dataTransferProtection,
```
agilgur5, line 14: To make artifacts more sustainable (other than plugins #5862), I've been arguing we should perhaps accept arbitrary YAML string output. E.g. in this case:
```yaml
clientOptions: |
  dataTransferProtection: true
```

That way we don't have to add a new feature every time a user wants to configure something else

Issue #13148, opened 2024-06-06
Links from workflows' containers in UI stopped being clickable after update


kzzalews:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

Links from workflows' containers in UI stopped being clickable after switch from Argo Workflows version 3.4.10 to 3.5.7.

Below you can see screenshots from two Argo Workflows versions. These show logs from single step (pod) running in our workflows and these logs could be viewed from Argo Workflows web UI. Please, notice that in old version link is underlined and clickable. In newer version it isn't clickable. It isn't interpreted as such. We've tested that on different web browsers (Safari, Firefox and Chrome).

Anonymized screenshot from version 3.4.10:
![argo_3 4 10](https://github.com/argoproj/argo-workflows/assets/21981908/8cde69dd-d9fa-4b61-add8-757ebb5c771b)

Anonymized screenshot from version 3.5.7:
![argo_3 5 7](https://github.com/argoproj/argo-workflows/assets/21981908/101fe08b-8353-44f6-a4a7-fc70c1eac68d)

### Version

3.5.7

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
N/A
```

### Logs from the workflow controller

```text
N/A
```

### Logs from in your workflow's wait container

```text
N/A
```

agilgur5:
> I suspect that there were some changes in UI modules related to logs rendering (e.g. framework or library update) that lead to such effect.

Correct, specifically an update of `argo-ui` in #11585 almost a year ago, which was one of my very first UI contributions. It did also get backported to 3.4.17 in #12998.
More specifically, [this line](https://github.com/argoproj/argo-workflows/pull/11585/files#diff-3a968206d6de2fecfc5dacd7d94bab7744c9f5d5c999a816164d95cbc135c316L2292) upgraded `xterm` from `2.4.0` to `^4.19.0`, which is used as a log viewer.
[...]

PR #13494, opened 2024-08-22
feat(ui): Make URLs clickable in workflow node info

Adrien-D:
No linked issue as it's a very small quick win, not sure it needs discussions

### Motivation

In workflow node details we may have templates that output links to other systems or other workflows. So that would be great to make navigation easier by showing links instead of text.

### Modifications

I implemented `linkify-it` library to find url matches and wrap those in a link tag.

| Before | After |
|---|---|
| Screenshot 2024-08-22 at 11 41 20 | Screenshot 2024-08-22 at 11 41 01 |
| Screenshot 2024-08-22 at 11 41 24 | Screenshot 2024-08-22 at 11 41 05 |

### Verification

I tested of several workflow to make sure everything is fine


ui/src/app/workflows/components/workflow-node-info/workflow-node-info.tsx
```diff
@@ -68,10 +69,40 @@ interface Props {
     onResume?: () => void;
 }
 
+const linkify = new LinkifyIt();
+
+function linkifyText(text: string) {
```
agilgur5, line 74: This should probably be a util function given that it's likely to be used in other places.

That also raises the question: is this the only place we'd add this? Does this open too many floodgates?

ui/package.json
```diff
@@ -25,6 +25,7 @@
         "dagre": "^0.8.5",
         "history": "^4.10.1",
         "js-yaml": "^4.1.0",
+        "linkify-it": "^5.0.0",
```
agilgur5, line 28: We don't seem to use all the logic from this library: https://github.com/markdown-it/linkify-it. It also doesn't seem tree-shakeable given that it relies on prototype methods

I might suggest something smaller, in the spirit of #12059

agilgur5 (commented):
Oo I like the table! A few considerations below

tooptoop4:
do u think same approach would work on https://github.com/argoproj/argo-workflows/issues/13148 ?

agilgur5:
No, since that uses a terminal with `xterm`. If we simplified it as I [mentioned there](https://github.com/argoproj/argo-workflows/issues/13148#issuecomment-2153129909), then it could work depending on the approach

ui/src/app/workflows/components/workflow-node-info/workflow-node-info.tsx
Adrien-D: No you're right we could definitely use this in other places, can't see where atm but I'll make it easily reusable

ui/package.json
Adrien-D: You're right about prototype but seems to me that the impact of this library is quite negligible compared to others we could use

---- 

### Linkify-it (current) 
Linkify-it

### Linkifyjs (other I could find)
Linkifyjs

---- 
Also a note about lazy and the project.. (I wrote it here first but let's discuss it [on the issue here](https://github.com/argoproj/argo-workflows/issues/12059#issuecomment-2310473252))

Adrien-D: I agree that the library is only partly tree shakable, but it looks quite small at the end. Should I keep looking for another lib ?

isubasinghe (approved):
LGTM

PR #13497, opened 2024-08-23
feat: concurrency policy triggered counter metric

Joibel:
### Motivation

The (new for 3.6) `cronworkflows_triggered_total` metric can be misleading with `concurrencyPolicy: Forbid`. In fixing that I believed it would be much more useful to also count these policies being triggered for cases where this is unexpected.

So this PR adds a new metric called `cronworkflows_concurrencypolicy_triggered` which counts the number of times a CronWorkflows `concurrencyPolicy` has changed something.

The `concurency_policy` attribute can be:
* `Forbid`: Counts the number of times a workflow was not started because one was already running
* `Replace`: Counts the number of times a workflow was terminated because it was being replaced

### Modifications

This PR moves the call to increment `cronworkflows_triggered_total` so that it doesn't count workflow runs suppressed by `Forbid`. This is documented and the test updated to test it.

This PR also tidies some of the metrics documentation and language around CronWorkflows.

### Verification

Existing test case modified to become 2 test cases checking both of the mentioned metrics, once for Forbid and once for Replace.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -201,14 +201,25 @@ Metrics for the [Four Golden Signals](https://sre.google/sre-book/monitoring-dis
     Some metric attributes may have high cardinality and are marked with ⚠️ to warn you. You may need to disable this metric or disable the attribute.
 <!-- titles should be the exact metric name for deep-linking, alphabetical ordered -->
 <!-- titles are without argo_workflows prefix -->
-#### `cronworkflows_triggered_total`
+#### `cronworkflows_concurrencypolicy_triggered`
 
-A counter of the number of times a CronWorkflow has been
+A counter of the number of times a CronWorkflow has triggered it's `concurrencyPolicy` to limit the number of workflows running.
 
 | attribute   | explanation                               |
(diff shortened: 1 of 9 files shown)


docs/metrics.md
```diff
@@ -201,14 +201,25 @@ Metrics for the [Four Golden Signals](https://sre.google/sre-book/monitoring-dis
     Some metric attributes may have high cardinality and are marked with ⚠️ to warn you. You may need to disable this metric or disable the attribute.
 <!-- titles should be the exact metric name for deep-linking, alphabetical ordered -->
 <!-- titles are without argo_workflows prefix -->
+#### `cronworkflows_concurrencypolicy_triggered`
+
+A counter of the number of times a CronWorkflow has triggered it's `concurrencyPolicy` to limit the number of workflows running.
```
juliev0, line 206:
```suggestion
A counter of the number of times a CronWorkflow has triggered its `concurrencyPolicy` to limit the number of workflows running.
```

isubasinghe (approved):
LGTM

PR #13500, opened 2024-08-24
fix(controller): remove ArtifactGC `finalizer` when no artifacts. Fixes #13499

### LINKED ISSUE #13499: Cannot delete failed Workflow due to ArtifactGC finalizer
### Pre-requisites

run example `examples/artifact-gc-workflow.yaml`

1. `kubectl create -f examples/artifact-gc-workflow.yaml`
2. because I am using a Linux/arm64 system, the workflow failed.
```
kubectl get workflow
NAME                STATUS   AGE   MESSAGE
artifact-gc-qzcrv   Failed   10s   Error (exit code 64): failed to start command: fork/exec /bin/sh: exec format error
```
3. so i decide to delete the workflow, but at this time, deleting the workflow is stuck and cannot be successful
get workflow:
```yaml
kind: Workflow
metadata:
  annotations:
    workflows.argoproj.io/pod-name-format: v2
  creationTimestamp: "2024-08-24T10:08:29Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2024-08-24T10:13:21Z"
  finalizers:
  - workflows.argoproj.io/artifact-gc
  generateName: artifact-gc-
  generation: 5
  labels:
    workflows.argoproj.io/completed: "true"
    workflows.argoproj.io/phase: Failed
  name: artifact-gc-qzcrv
  namespace: default
  resourceVersion: "8957"
  uid: 4e159dab-2888-4097-a4c3-9ff67758d28d
spec:
...
```

i found that `finalizers` still existed, it should be removed

### Version(s)

v3.5.10

run `examples/artifact-gc-workflow.yaml` on arm mac, and then delete the workflow

```text
time="2024-08-24T10:08:40.025Z" level=info msg="node changed" namespace=default new.message="Error (exit code 64): failed to start command: fork/exec /bin/sh: exec format error" new.phase=Failed new.progress=0/1 nodeID=artifact-gc-qzcrv old.message= old.phase=Pending old.progress=0/1 workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.026Z" level=info msg="TaskSet Reconciliation" namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.026Z" level=info msg=reconcileAgentPod namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.026Z" level=info msg="Updated phase Running -> Failed" namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.026Z" level=info msg="Updated message  -> Error (exit code 64): failed to start command: fork/exec /bin/sh: exec format error" namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.026Z" level=info msg="Marking workflow completed" namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.026Z" level=info msg="Workflow to be dehydrated" Workflow Size=1689
time="2024-08-24T10:08:40.048Z" level=info msg="Workflow update successful" namespace=default phase=Failed resourceVersion=8669 workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:40.069Z" level=info msg="cleaning up pod" action=labelPodCompleted key=default/artifact-gc-qzcrv/labelPodCompleted
time="2024-08-24T10:08:45.068Z" level=info msg="Processing workflow" Phase=Failed ResourceVersion=8669 namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:45.069Z" level=info msg="Task-result reconciliation" namespace=default numObjs=0 workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:45.069Z" level=info msg="Workflow to be dehydrated" Workflow Size=1715
time="2024-08-24T10:08:45.088Z" level=info msg="Workflow update successful" namespace=default phase=Failed resourceVersion=8680 workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:50.093Z" level=info msg="Processing workflow" Phase=Failed ResourceVersion=8680 namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:08:50.093Z" level=info msg="Task-result reconciliation" namespace=default numObjs=0 workflow=artifact-gc-qzcrv
time="2024-08-24T10:12:52.475Z" level=info msg="Alloc=5451 TotalAlloc=18959 Sys=24165 NumGC=8 Goroutines=175"
time="2024-08-24T10:13:31.589Z" level=info msg="Processing workflow" Phase=Failed ResourceVersion=8957 namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:13:31.589Z" level=info msg="Task-result reconciliation" namespace=default numObjs=0 workflow=artifact-gc-qzcrv
time="2024-08-24T10:13:31.590Z" level=info msg="Workflow to be dehydrated" Workflow Size=1751
time="2024-08-24T10:13:31.604Z" level=info msg="Workflow update successful" namespace=default phase=Failed resourceVersion=8967 workflow=artifact-gc-qzcrv
time="2024-08-24T10:13:36.610Z" level=info msg="Processing workflow" Phase=Failed ResourceVersion=8967 namespace=default workflow=artifact-gc-qzcrv
time="2024-08-24T10:13:36.610Z" level=info msg="Task-result reconciliation" namespace=default numObjs=0 workflow=artifact-gc-qzcrv
```

```text
time="2024-08-24T10:08:31.403Z" level=info msg="Starting Workflow Executor" version=v3.5.10
time="2024-08-24T10:08:31.404Z" level=info msg="Using executor retry strategy" Duration=1s Factor=1.6 Jitter=0.5 Steps=5
time="2024-08-24T10:08:31.404Z" level=info msg="Executor initialized" deadline="0001-01-01 00:00:00 +0000 UTC" includeScriptOutput=false namespace=default podName=artifact-gc-qzcrv templateName=main version="&Version{Version:v3.5.10,BuildDate:2024-08-01T05:12:26Z,GitCommit:25829927431d9a0f46d17b72ae74aedb8d700884,GitTag:v3.5.10,GitTreeState:clean,GoVersion:go1.21.12,Compiler:gc,Platform:linux/arm64,}"
time="2024-08-24T10:08:31.414Z" level=info msg="Starting deadline monitor"
time="2024-08-24T10:08:32.414Z" level=info msg="Main container completed" error="<nil>"
time="2024-08-24T10:08:32.414Z" level=info msg="No Script output reference in workflow. Capturing script output ignored"
time="2024-08-24T10:08:32.414Z" level=info msg="No output parameters"
time="2024-08-24T10:08:32.414Z" level=info msg="Saving output artifacts"
time="2024-08-24T10:08:32.415Z" level=info msg="Staging artifact: on-completion"
time="2024-08-24T10:08:32.415Z" level=info msg="Copying /tmp/on-completion.txt from container base image layer to /tmp/argo/outputs/artifacts/on-completion.tgz"
time="2024-08-24T10:08:32.415Z" level=info msg="/var/run/argo/outputs/artifacts/tmp/on-completion.txt.tgz -> /tmp/argo/outputs/artifacts/on-completion.tgz"
time="2024-08-24T10:08:32.415Z" level=error msg="executor error: open /var/run/argo/outputs/artifacts/tmp/on-completion.txt.tgz: no such file or directory"
time="2024-08-24T10:08:32.437Z" level=info msg="Alloc=8902 TotalAlloc=13535 Sys=24165 NumGC=3 Goroutines=8"
time="2024-08-24T10:08:32.456Z" level=fatal msg="open /var/run/argo/outputs/artifacts/tmp/on-completion.txt.tgz: no such file or directory"
```

chengjoey:
Fixes #13499 

### Motivation

if workflow pods length is zero, `anyPodSuccess` should be true, then the `removeFinalizer` can be executed without force

### Modifications

`anyPodSuccess := len(pods) == 0`

### Verification

run `examples/artifact-gc-workflow.yaml` on mac m1 os, and then delete the workflow

### DIFF
--- a/test/e2e/artifacts_test.go
+++ b/test/e2e/artifacts_test.go
@@ -388,6 +388,14 @@ func (s *ArtifactsSuite) TestArtifactGC() {
 				artifactState{s3Location{bucketName: "my-bucket", derivedKey: &artifactDerivedKey{templateName: "artifact-written", artifactName: "present"}}, false, true},
 			},
 		},
+		// Workflow defined output artifact but execution failed, no artifacts to be gced
+		{
+			workflowFile:                 "@testdata/artifactgc/artgc-artifact-not-written-failed.yaml",
+			hasGC:                        true,
+			workflowShouldSucceed:        false,
+			expectedGCPodsOnWFCompletion: 0,
+			expectedArtifacts:            []artifactState{},
+		},
 	} {
 		// for each test make sure that:
 		// 1. the finalizer gets added
--- a/test/e2e/testdata/artifactgc/artgc-artifact-not-written-failed.yaml
+++ b/test/e2e/testdata/artifactgc/artgc-artifact-not-written-failed.yaml
@@ -1,0 +1,27 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: artgc-artifact-not-written-failed-
+spec:
+  entrypoint: main
+  artifactGC:
+    strategy: OnWorkflowDeletion
+  templates:
+    - name: main
+      container:
+        image: argoproj/argosay:v2
+        command:
+          - sh
+          - -c
+        args:
+          - |
+            echo "intentionally not writing anything to disk and intentional failure"
+            exit 1
+      outputs:
+        artifacts:
+          - name: on-completion
+            path: /tmp/on-completion.txt
+            s3:
+              key: on-completion.txt
+            artifactGC:
+              strategy: OnWorkflowCompletion
--- a/workflow/controller/artifact_gc.go
+++ b/workflow/controller/artifact_gc.go
@@ -508,7 +508,6 @@ func (woc *wfOperationCtx) processArtifactGCCompletion(ctx context.Context) erro
 		return fmt.Errorf("failed to get pods from informer: %w", err)
 	}
 
-	anyPodSuccess := false
 	for _, obj := range pods {
 		pod := obj.(*corev1.Pod)
 		if pod.Labels[common.LabelKeyComponent] != artifactGCComponent { // make sure it's an Artifact GC Pod
@@ -534,10 +533,8 @@ func (woc *wfOperationCtx) processArtifactGCCompletion(ctx context.Context) erro
 			if err != nil {
 				return err
 			}
+
 			woc.wf.Status.ArtifactGCStatus.SetArtifactGCPodRecouped(pod.Name, true)
-			if phase == corev1.PodSucceeded {
-				anyPodSuccess = true
-			}
 			woc.updated = true
 		}
 	}
@@ -548,7 +545,7 @@ func (woc *wfOperationCtx) processArtifactGCCompletion(ctx context.Context) erro
 		removeFinalizer = woc.wf.Status.ArtifactGCStatus.AllArtifactGCPodsRecouped()
 	} else {
 		// check if all artifacts have been deleted and if so remove Finalizer
-		removeFinalizer = anyPodSuccess && woc.allArtifactsDeleted()
+		removeFinalizer = woc.allArtifactsDeleted()
 	}
 	if removeFinalizer {
 		woc.log.Infof("no remaining artifacts to GC, removing artifact GC finalizer (forceFinalizerRemoval=%v)", forceFinalizerRemoval)


agilgur5 (commented):
I'm not sure this is quite correct. A length of `0` could also mean that an ArtifactGC Pod has yet to run.

In your issue, the ArtifactGC Pod failed, and this is exactly what [the `forceFinalizerRemoval` field is for](https://argo-workflows.readthedocs.io/en/release-3.5/walk-through/artifacts/#release-versions-35)

chengjoey:
> I'm not sure this is quite correct. A length of `0` could also mean that an ArtifactGC Pod has yet to run.
> 
> In your issue, the ArtifactGC Pod failed, and this is exactly what [the `forceFinalizerRemoval` field is for](https://argo-workflows.readthedocs.io/en/release-3.5/walk-through/artifacts/#release-versions-35)

https://github.com/argoproj/argo-workflows/blob/ddbb3c7ad5b498d50514b3c1158ded56e333d75b/workflow/controller/operator.go#L242-L250

hi @agilgur5 , artifacts gc will only be running when task reconciliation is completed. I think the pod has finished running at this time. and i think this is a normal operation to delete the workflow, no need to `force`, increase the cost of the deletion operation, because this is not a gc problem that requires forced deletion

agilgur5:
TaskResult reconciliation should already be completed. In your issue, I'm pretty sure the ArtifactGC Pod _itself_ fails. As it failed, the Controller cannot be sure that the artifact was deleted or not

agilgur5:
> I'm pretty sure the ArtifactGC Pod _itself_ fails (due to the architecture)

Hmm, no, ArtifactGC can run on `arm64`; your Pod that fails due to arch is the `argosay` image's arch -- see https://github.com/argoproj/argo-workflows/pull/11613#issuecomment-1920685874 that I was going to fix recently as I may have permissions to the image repositories.

> artifacts gc will only be running when task reconciliation is completed

> TaskResult reconciliation should already be completed

Or you're experiencing a different race condition from v3.5.3+ / v3.5.5+. This actually sounds like the root cause is #12993, which was recently fixed in #13454

chengjoey:
> > I'm pretty sure the ArtifactGC Pod _itself_ fails (due to the architecture)
> 
> Hmm, no, ArtifactGC can run on `arm64`; your Pod that fails due to arch is the `argosay` image's arch -- see [#11613 (comment)](https://github.com/argoproj/argo-workflows/pull/11613#issuecomment-1920685874) that I was going to fix recently as I may have permissions to the image repositories.
> 
> > artifacts gc will only be running when task reconciliation is completed
> 
> > TaskResult reconciliation should already be completed
> 
> Or you're experiencing a different race condition from v3.5.3+ / v3.5.5+. This actually sounds like the root cause is #12993, which was recently fixed in #13454

Yes, I know that the reason why I failed to run is because the image `argoproj/argosay:v2` in the examples does not support the linux/arm64 architecture, not because artifact-gc cannot run in arm64. I raised this PR because my workflow failed and no artifact was generated, so there is no need to do GC. I hope to be able to delete the workflow directly without adding additional operations `force`

agilgur5:
> I raised this PR because my workflow failed and no artifact was generated, so there is no need to do GC.

Yes, the second half of [my previous comment](https://github.com/argoproj/argo-workflows/pull/13500#issuecomment-2308383109) is relevant with regard to TaskResult reconciliation. This looks like it is caused by a bug in TaskResult reconciliation and not specific to ArtifactGC

chengjoey:
> Yes, the second half of [my previous comment](https://github.com/argoproj/argo-workflows/pull/13500#issuecomment-2308383109) is relevant with regard to TaskResult reconciliation. This looks like it is caused by a bug in TaskResult reconciliation and not specific to ArtifactGC

thanks @agilgur5 , i got it

agilgur5:
Ok this makes more sense now given your [last comment on the issue](https://github.com/argoproj/argo-workflows/issues/13499#issuecomment-2308451871), thanks again for elaborating! 

From my [first comment above](https://github.com/argoproj/argo-workflows/pull/13500#pullrequestreview-2258675551) though:

> A length of `0` could also mean that an ArtifactGC Pod has yet to run

I think we could potentially remove this variable entirely and just replace with a check that all ArtifactGC Pods were recouped?

Otherwise it looks like `anyPodSuccess` is only used for the one check you linked to, so perhaps want to invert that or rename the variable if it's still needed.

cc @juliev0 here on the PR as well

chengjoey:
thanks for your patience @agilgur5 @juliev0 , In this PR I removed `anyPodSuccess`,  `allArtifactsDeleted` is enough to ensure that artifacts can be GCed

juliev0:
> thanks for your patience @agilgur5 @juliev0 , In this PR I removed `anyPodSuccess`, `allArtifactsDeleted` is enough to ensure that artifacts can be GCed

thanks for the modification! The e2e tests are sometimes flakey, so if the e2e test failures are unrelated to your changes, feel free to push [empty commits](https://www.freecodecamp.org/news/how-to-push-an-empty-commit-with-git/) until it passes

agilgur5:
> so I'd be okay to remove the `anyPodSuccess` altogether if it makes the code simpler.

@juliev0 did you see my comment [above](https://github.com/argoproj/argo-workflows/pull/13500#issuecomment-2308636365)? I thought the check might be to make sure that there were no leftover ArtifactGC Pods that were not yet recouped

agilgur5 (commented):
@chengjoey do you think you could add a regression test for this scenario? If I'm not mistaken, you should be able to use any `main` container that intentionally fails while having `output.artifacts` specified

chengjoey:
Ok, I will add an e2e test

juliev0:
>  I thought the check might be to make sure that there were no leftover ArtifactGC Pods that were not yet recouped. 

Hmm, does it accomplish that? I'm having trouble following if so. I think it was just to eliminate work.

Thanks for adding the test @chengjoey. If I'm not mistaken, I think you should be able to add it as another case to [this](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/artifacts_test.go#L287) test. This test verifies that the finalizer is removed by [checking](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/artifacts_test.go#L458) that the workflow actually gets deleted.

test/e2e/artifacts_test.go
```diff
@@ -818,6 +818,52 @@ spec:
...
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeFailed).
+		Then().
+		ExpectWorkflow(func(t *testing.T, meta *metav1.ObjectMeta, _ *wfv1.WorkflowStatus) {
+			assert.Equal(t, 0, len(meta.Finalizers))
+		}).
+		When().
+		DeleteWorkflow().
+		WaitForWorkflowDeletion()
+}
```
juliev0, line 865: did it work better to add the separate test as opposed to incorporating it as a case in `TestArtifactGC()`?

test/e2e/testdata/artifactgc/artgc-has-gc-but-failed-no-artifacts.yaml
```diff
@@ -0,0 +1,30 @@
...
+          - name: on-completion
+            path: /tmp/on-completion.txt
+            s3:
+              key: on-completion.txt
+            artifactGC:
+              strategy: OnWorkflowCompletion
+          - name: on-deletion
+            path: /tmp/on-deletion.txt
+            s3:
+              key: on-deletion.txt
```
agilgur5, line 30:
```suggestion

```

only one artifact is necessary, no? we should keep the test minimal

test/e2e/testdata/artifactgc/artgc-has-gc-but-failed-no-artifacts.yaml
```diff
@@ -0,0 +1,30 @@
...
+  templates:
+    - name: main
+      container:
+        image: argoproj/argosay:v2
+        command:
+          - sh
+          - -c
+        args:
+          - |
+            exit 1
```
agilgur5, line 18: I'd probably also `echo` a debug message for clarity and consistency with [other](https://github.com/argoproj/argo-workflows/blob/62e33abded6765568781c519ee3e78b823a99acc/test/e2e/testdata/artifactgc/artgc-artifact-not-written.yaml#L39) [tests](https://github.com/argoproj/argo-workflows/blob/ddbb3c7ad5b498d50514b3c1158ded56e333d75b/examples/continue-on-fail.yaml#L34)
```suggestion
            echo "intentionally not writing anything to disk and intentional failure"
            exit 1
```

test/e2e/testdata/artifactgc/artgc-has-gc-but-failed-no-artifacts.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: artifact-gc-but-failed-no-artifacts-
```
agilgur5, line 4: the filename should match this by convention

also I think we should have this named similarly to `artgc-artifact-not-written.yaml` from #13066; the only difference is that that one has another successful artifact. (also `has-gc` is redundant here)

I might name this one `artgc-artifact-not-written-failed`? Or maybe this one as `artgc-artifact-not-written` and rename that one to `artgc-artifact-not-written-and-written` or something

agilgur5 (commented):
mostly consistency comments on the test below

chengjoey:
> mostly consistency comments on the test below

thanks @agilgur5 , all has done

agilgur5 (approved):
LGTM, thanks for the iteration!

Will let Julie take another look at the tests before merge

juliev0 (approved):

juliev0:
Yes, thank you @chengjoey !

PR #11839, opened 2023-09-18
fix: retry only proper node (#11589)

toyamagu-2021:
Signed-off-by: toyamagu-2021 <toyamagu2021@gmail.com>

Fixes #11589 

### Motivation

* Retry only retried nodes.

### Modifications

* When retring, exclude hooked nodes. 

### Verification

* retryStrategy works correctly
  * image
* This PR also works fine for WorkflowTemplateLevelHook
  * image
* When running, the hook node is DAG
  * image
* When running the hook node is Step
  * image
* Works file for a most complicated case
  * image

<details>
<summary>manifest</summary>

```yaml
spec:
  templates:
    - name: argosay
      inputs: {}
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: argoproj/argosay:v2
        command:
          - /bin/sh
          - '-c'
        args:
          - /bin/sleep 2; exit 100; /argosay
        resources: {}
      retryStrategy:
        limit: 1
    - name: hook
      inputs: {}
      outputs: {}
      metadata: {}
      steps:
        - - name: A
            template: argosay
            arguments: {}
            hooks:
              failed:
                template: argosay2
                arguments: {}
                expression: steps.A.status == "Failed"
              running:
                template: argosay2
                arguments: {}
                expression: steps.A.status == "Running"
              succeed:
                template: argosay2
                arguments: {}
                expression: steps.A.status == "Succeeded"
        - - name: B
            template: argosay
            arguments: {}
    - name: argosay2
      inputs: {}
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: argoproj/argosay:v2
        command:
          - /bin/sh
          - '-c'
        args:
          - exit 0
        resources: {}
  entrypoint: argosay
  arguments: {}
  hooks:
    running:
      template: hook
      arguments: {}
      expression: workflow.status == "Running"
```

</details>

PR #13504, opened 2024-08-26
fix: provide fallback for 3.4 to 3.5 transition with absent `NodeFlag`. Fixes #12162

### LINKED ISSUE #12162: During upgrade: `Maximum recursion depth exceeded` error when not using recursion
### Pre-requisites

never saw this error in 3.4.11

but after upgrading to 3.5.1 
got `error in entry template execution: Maximum recursion depth exceeded. See https://argoproj.github.io/argo-workflows/scaling/#maximum-recursion-depth` on 2 different small (1 step) workflows

### Version

3.5.1

```YAML
n/a
```

```text
```

```text
```

isubasinghe:
Fixes #12162

### DIFF
--- a/workflow/common/util.go
+++ b/workflow/common/util.go
@@ -323,6 +323,32 @@ func IsDone(un *unstructured.Unstructured) bool {
 		un.GetLabels()[LabelKeyWorkflowArchivingStatus] != "Pending"
 }
 
+func CheckHookNode(nodeName string) bool {
+	names := strings.Split(nodeName, ".")
+	if len(names) == 2 {
+		return names[1] == "onExit"
+	}
+
+	if len(names) <= 2 {
+		return false
+	}
+
+	if names[len(names)-1] == "onExit" || names[len(names)-2] == "hooks" {
+		return true
+	}
+	return false
+}
+
+// Used to check if a child node is a retry type
+func CheckRetryNode(parentsMap map[string]*wfv1.NodeStatus, nodeID string) bool {
+	parent, found := parentsMap[nodeID]
+	if !found {
+		return false
+	}
+
+	return parent.Type == wfv1.NodeTypeRetry
+}
+
 // Check whether child hooked nodes Fulfilled
 func CheckAllHooksFullfilled(node *wfv1.NodeStatus, nodes wfv1.Nodes) bool {
 	childs := node.Children
--- a/workflow/common/util.go
+++ b/workflow/common/util.go
@@ -323,6 +323,8 @@ func IsDone(un *unstructured.Unstructured) bool {
 		un.GetLabels()[LabelKeyWorkflowArchivingStatus] != "Pending"
 }
 
+// CheckHookNode is used to determine if
+// a node was a hook node via its name.
 func CheckHookNode(nodeName string) bool {
 	names := strings.Split(nodeName, ".")
 	if len(names) == 2 {
@@ -339,8 +341,9 @@ func CheckHookNode(nodeName string) bool {
 	return false
 }
 
-// Used to check if a child node is a retry type
-func CheckRetryNode(parentsMap map[string]*wfv1.NodeStatus, nodeID string) bool {
+// CheckRetryNodeParent determines if a node is a retriable node,
+// that is if the node is the child of a retry node.
+func CheckRetryNodeParent(parentsMap map[string]*wfv1.NodeStatus, nodeID string) bool {
 	parent, found := parentsMap[nodeID]
 	if !found {
 		return false
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -205,10 +205,6 @@ func (woc *wfOperationCtx) maybeUpgradeWithNodeFlags() {
 		}
 	}
 
-	// Can be merged in with the previous loop.
-	// Merging wouldn't impact performance in a significant enough way.
-	// In fact this might be more performant thanks to better branch prediction
-	// in the previous loop.
 	for key, node := range woc.wf.Status.Nodes {
 		newNode := node
 		if common.CheckHookNode(node.Name) {
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -181,7 +181,7 @@ func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOper
 	return &woc
 }
 
-// Determines if this workflow is definitely after the introduction of
+// definitePostNodeFlagsWf determines if this workflow is definitely after the introduction of
 // the NodeFlag field.
 // used to avoid the O(n*m) loop in maybeUpgradeWithNodeFlags
 func (woc *wfOperationCtx) definitePostNodeFlagsWf() bool {
@@ -193,8 +193,9 @@ func (woc *wfOperationCtx) definitePostNodeFlagsWf() bool {
 	return false
 }
 
+// maybeUpgradeWithNodeFlags upgrades old workflows without
+// NodeFlag fields into ones with.
 func (woc *wfOperationCtx) maybeUpgradeWithNodeFlags() {
-
 	parentsMap := make(map[string]*wfv1.NodeStatus)
 
 	for _, node := range woc.wf.Status.Nodes {
@@ -214,7 +215,7 @@ func (woc *wfOperationCtx) maybeUpgradeWithNodeFlags() {
 			newNode.NodeFlag = &wfv1.NodeFlag{Hooked: true}
 		}
 
-		if common.CheckRetryNode(parentsMap, node.ID) {
+		if common.CheckRetryNodeParent(parentsMap, node.ID) {
 			if newNode.NodeFlag == nil {
 				newNode.NodeFlag = &wfv1.NodeFlag{}
 			}
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -181,6 +181,49 @@ func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOper
 	return &woc
 }
 
+// Determines if this workflow is definitely after the introduction of
+// the NodeFlag field.
+// used to avoid the O(n*m) loop in maybeUpgradeWithNodeFlags
+func (woc *wfOperationCtx) definitePostNodeFlagsWf() bool {
+	for _, node := range woc.wf.Status.Nodes {
+		if node.NodeFlag != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (woc *wfOperationCtx) maybeUpgradeWithNodeFlags() {
+
+	parentsMap := make(map[string]*wfv1.NodeStatus)
+
+	for _, node := range woc.wf.Status.Nodes {
+		node := node
+		for _, childNodeID := range node.Children {
+			parentsMap[childNodeID] = &node
+		}
+	}
+
+	// Can be merged in with the previous loop.
+	// Merging wouldn't impact performance in a significant enough way.
+	// In fact this might be more performant thanks to better branch prediction
+	// in the previous loop.
+	for key, node := range woc.wf.Status.Nodes {
+		newNode := node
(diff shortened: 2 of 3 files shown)


workflow/common/util.go
```diff
@@ -324,6 +324,23 @@ func IsDone(un *unstructured.Unstructured) bool {
 		un.GetLabels()[LabelKeyWorkflowArchivingStatus] != "Pending"
 }
 
+// Remove ASAP in 3.6, used **only** for backward compatability.
```
agilgur5, line 327: It would probably make more sense to PR this to the `release-3.5` branch then? Or is part of this meant to stay on `main`?

isubasinghe: Depends on how long we want to support  this fallback fix for, was hoping for agreement that we do not expect workflows initially run with 3.4 to finish to completion when the controller is upgraded to 3.6 while workflows are running.

workflow/controller/operator.go
```diff
@@ -1822,6 +1822,18 @@ func (woc *wfOperationCtx) possiblyGetRetryChildNode(node *wfv1.NodeStatus) *wfv
...
+				if !common.CheckHookNode(childNode.Name) {
+					return childNode
+				} else {
+					continue
+				}
+			}
+
 			if childNode.NodeFlag == nil || !childNode.NodeFlag.Hooked {
 				return childNode
 			}
```
isubasinghe, line 1839: Can remove the nil check here, but this is very intentionally kept because the above code isn't meant to be around for long.

workflow/controller/operator.go
```diff
@@ -1822,6 +1822,18 @@ func (woc *wfOperationCtx) possiblyGetRetryChildNode(node *wfv1.NodeStatus) *wfv
...
+
+			// childNode.nodeFlag was missing, this can only happen if NodeStatus is old
+			if childNode.NodeFlag == nil {
+				// fallback where nodeFlags is absent
+				// ensure that we do not return
+				if !common.CheckHookNode(childNode.Name) {
+					return childNode
+				} else {
+					continue
```
isubasinghe, line 1833: breaks 3.5.x code now, since any version before this PR might have NodeFlag be nil

workflow/controller/operator.go
```diff
@@ -4134,9 +4159,19 @@ func getChildNodeIdsRetried(node *wfv1.NodeStatus, nodes wfv1.Nodes) []string {
 	childrenIds := []string{}
 	for i := 0; i < len(node.Children); i++ {
 		n := getChildNodeIndex(node, nodes, i)
-		if n == nil || n.NodeFlag == nil {
+		if n == nil {
+			continue
+		}
+
+		// fallback code here
+		if n.NodeFlag == nil {
```
isubasinghe, line 4167: This behaviour is now different: 
Is it safe that we could be returning StepGroup etc now?

We can potentially get the same behaviour back by checking node type and only allowing for retri-able node types.

workflow/common/util.go
agilgur5: Yea I can agree to that

isubasinghe: Done!

workflow/controller/operator.go
isubasinghe: Double checked this, this is perfectly safe given our usage of the `getChildNodeIdsAndLastRetriedNode` function.
The name of the function is however misleading and should be changed.

isubasinghe:
> I'm not sure if this would simplify the code, but if we detect a 3.4 Workflow, could we just skip the recursion check on it entirely? Since it did not exist in 3.4

Not really this isn't a recursion check, the PR title is incorrect. 
3.4 workflows missed a NodeFlag, 3.5 introduced one.

This fix mandates a NodeFlag to be created now, it takes the lack of a NodeFlag to imply older controller to have created the workflow, then it tries to mimic 3.4 logic.

agilgur5:
> Not really this isn't a recursion check, the PR title is incorrect.
> 3.4 workflows missed a NodeFlag, 3.5 introduced one.

I changed the title again. I'm not entirely sure I follow the root cause -- your comments in #12162 don't really outline it either.
That issue is about the recursion check added in #11646. `NodeFlag` was added in #11839 (i'm not entirely certain it's necessary either, since it seems like it could be derived). Are you saying that #11839 caused an infinite loop somewhere that #11646 then detected?

workflow/controller/operator.go
```diff
@@ -1864,15 +1886,16 @@ type executeTemplateOpts struct {
 	// activeDeadlineSeconds is a deadline to set to any pods executed. This is necessary for pods to inherit backoff.maxDuration
 	executionDeadline time.Time
 	// nodeFlag tracks node information such as hook or retry
-	nodeFlag *wfv1.NodeFlag
+	// MUST create since we rely on lack of nodeFlag for backwards compatability.
+	nodeFlag wfv1.NodeFlag
```
Joibel, line 1890: I don't like this as a difference between 3.5 and 3.6.

Joibel (changes requested):
I'm worried about 3.5 and 3.6 diverging in behaviour with this.

As a less invasive change did you consider scanning the workflow at the beginning of operate to upgrade it?
* If you find a nodeFlags field, abort scan (early dropout for all updated workflows to avoid the cost
* Scan each node and calling CheckHookNode generate and attach a nodeFlags to the nodes as appropriate

This scan would then not be needed in 3.6, but it would be the only difference.

isubasinghe:
> I changed the title again. I'm not entirely sure I follow the root cause -- your comments in #12162 don't really outline it either. That issue is about the recursion check added in #11646. `NodeFlag` was added in #11839 (i'm not entirely certain it's necessary either, since it seems like it could be derived). Are you saying that #11839 caused an infinite loop somewhere that #11646 then detected?


No worries. 

Basically the `NodeFlag` was added in #11839 along with a function to get all retry nodes by using the `NodeFlag` field.
The problem is that 3.4 workflows do not have this. This leads to the controller thinking that the retry node  was never created. It then throws the controller into an infinite recursion where it tries to executeTemplate recursively.

isubasinghe:
> As a less invasive change did you consider scanning the workflow at the beginning of operate to upgrade it?

No, mostly because the mandating of the NodeFlag effectively provides a blast radius. But I don't think this is needed.
I've re-read the retry logic. 

I think I can effectively create a map containing Nodes mapped into their NodeFlags.

Thanks for the suggestion, that is definitely less invasive than this fix.

workflow/controller/operator.go
```diff
@@ -2521,15 +2541,15 @@ func (woc *wfOperationCtx) initializeCacheNode(nodeName string, resolvedTmpl *wf
...
+func (woc *wfOperationCtx) initializeCacheHitNode(nodeName string, resolvedTmpl *wfv1.Template, templateScope string, orgTmpl wfv1.TemplateReferenceHolder, boundaryID string, outputs *wfv1.Outputs, memStat *wfv1.MemoizationStatus, nodeFlag wfv1.NodeFlag, messages ...string) *wfv1.NodeStatus {
 	node := woc.initializeCacheNode(nodeName, resolvedTmpl, templateScope, orgTmpl, boundaryID, memStat, nodeFlag, messages...)
 	node.Phase = wfv1.NodeSucceeded
 	node.Outputs = outputs
 	node.FinishedAt = metav1.Time{Time: time.Now().UTC()}
 	return node
 }
 
-func (woc *wfOperationCtx) initializeNode(nodeName string, nodeType wfv1.NodeType, templateScope string, orgTmpl wfv1.TemplateReferenceHolder, boundaryID string, phase wfv1.NodePhase, nodeFlag *wfv1.NodeFlag, messages ...string) *wfv1.NodeStatus {
+func (woc *wfOperationCtx) initializeNode(nodeName string, nodeType wfv1.NodeType, templateScope string, orgTmpl wfv1.TemplateReferenceHolder, boundaryID string, phase wfv1.NodePhase, nodeFlag wfv1.NodeFlag, messages ...string) *wfv1.NodeStatus {
```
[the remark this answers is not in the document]
isubasinghe: I will revert this in that case, I have an idea what a simpler fix for this could look like now.

isubasinghe:
I am pretty happy with this now actually :)

workflow/controller/operator.go
```diff
@@ -181,6 +181,49 @@ func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOper
...
+	for _, node := range woc.wf.Status.Nodes {
+		if node.NodeFlag != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (woc *wfOperationCtx) maybeUpgradeWithNodeFlags() {
+
```
Joibel, line 197: nit: blank line

workflow/common/util.go
```diff
@@ -323,6 +323,32 @@ func IsDone(un *unstructured.Unstructured) bool {
 		un.GetLabels()[LabelKeyWorkflowArchivingStatus] != "Pending"
 }
 
+func CheckHookNode(nodeName string) bool {
```
Joibel, line 326: Maybe add a [godoc](https://go.dev/doc/comment) comment

workflow/common/util.go
```diff
@@ -323,6 +323,32 @@ func IsDone(un *unstructured.Unstructured) bool {
...
+		return false
+	}
+
+	if names[len(names)-1] == "onExit" || names[len(names)-2] == "hooks" {
+		return true
+	}
+	return false
+}
+
+// Used to check if a child node is a retry type
```
Joibel, line 342: Can you make this a [godoc](https://go.dev/doc/comment) comment (e.g. start it with the function name) please.

workflow/controller/operator.go
```diff
@@ -181,6 +181,49 @@ func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOper
 	return &woc
 }
 
+// Determines if this workflow is definitely after the introduction of
```
Joibel, line 184: Can you make this a [godoc](https://go.dev/doc/comment) comment (e.g. start it with the function name) please.

workflow/controller/operator.go
```diff
@@ -181,6 +181,49 @@ func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOper
...
+func (woc *wfOperationCtx) definitePostNodeFlagsWf() bool {
+	for _, node := range woc.wf.Status.Nodes {
+		if node.NodeFlag != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (woc *wfOperationCtx) maybeUpgradeWithNodeFlags() {
```
Joibel, line 196: Should we force add a non nil nodeFlag in this to the root node to ensure `definitePostNodeFlagsWf` returns true as fast as possible?

Should new workflows created by this controller also do that at creation?

Joibel (changes requested):

workflow/controller/operator.go
isubasinghe: I don't think we need to, even for extremely large workflows (100,000 nodes) , a simple for loop over their nodes shouldn't take much time. I think there are other parts of the codebase that will suffer before these two functions become noticeably slow.

Joibel: Ok, I'll accept that.

workflow/controller/operator.go
```diff
@@ -181,6 +181,50 @@ func newWorkflowOperationCtx(wf *wfv1.Workflow, wfc *WorkflowController) *wfOper
...
+	parentsMap := make(map[string]*wfv1.NodeStatus)
+
+	for _, node := range woc.wf.Status.Nodes {
+		node := node
+		for _, childNodeID := range node.Children {
+			parentsMap[childNodeID] = &node
+		}
+	}
+
+	// Can be merged in with the previous loop.
```
Joibel, line 208: Make a decision and merge the loops or don't. Do remove this comment.

Joibel (changes requested):

Joibel (approved):

PR #13508, opened 2024-08-27
fix(docs): use `sh` instead of `bash` with `busybox`

chengjoey:
create `examples/work-avoidance.yaml` 
echo pod occur error:
```
Error: failed to find name in PATH: exec: "bash": executable file not found in $PATH
```

### Motivation

use `sh` instead of `bash`

### DIFF
--- a/examples/work-avoidance.yaml
+++ b/examples/work-avoidance.yaml
@@ -65,7 +65,7 @@ spec:
       script:
         image: busybox
         command:
-          - bash
+          - sh
           - -eux
         source: |
           marker=/work/markers/$(date +%Y-%m-%d)-echo-{{inputs.parameters.num}}


agilgur5 (commented):
Ah, follow-up to #13429, nice catch!
Were there any other uses of `bash` you could find?

agilgur5:
Answering my own question, did [a search](https://github.com/search?q=repo%3Aargoproj%2Fargo-workflows+bash&type=code) and the rest uses other images. Some of them do use `argosay` though, which is slated for removal as well, so those will need to be careful around

chengjoey:
> Were there any other uses of `bash` you could find?

I checked all the `busybox` examples, and only found this one using `bash`. others all use `sh` or directly use command (echo, true, mkdir etc)

agilgur5 (approved):
Nice catch and thanks for double-checking others!

PR #13509, opened 2024-08-27
feat(ui): Adjust the node display name to include which template is invoking it. Fixes #11106

### LINKED ISSUE #11106: Group UI nodes based on `templateRef`
# Summary

I was chatting with @mshatkhin23, and he mentioned that he'd like to have the workflow graph grouped based on the workflow template ref. Given this workflow snippet (using templates instead of refs here for simplicity):
```yaml
- name: dag-tier-3
  dag:
    tasks:
      - name: task-1
        template: flakey-container
      - name: task-2
        depends: 'task-1'
        template: flakey-container
- name: dag-tier-2
  dag:
    tasks:
      - name: task-1
        template: dag-tier-3
      - name: task-2
        depends: 'task-1'
        template: dag-tier-3
- name: user-workflow
  dag:
    tasks:
      - name: task-1
        template: dag-tier-2
      - name: task-2
        depends: 'task-1'
        template: dag-tier-2
```
It produces the following graph:
image

We'd like to find a way to group or collapse things based on a given template. In the graph above, it's hard to know which `task-1` is in user-workflow or `dag-tier-2` or `dag-tier-3`. Ideally we:
- collapse templateRefs from the start
- adjust the display name so that the task name also includes which template is invoking it

# Use Cases

This should help eliminate confusion when looking at the Argo Workflows graph. As workflows scale and orgs manage lots of templates, there will naturally be naming collisions that arise

---

Adrien-D:
Fixes #11106. This is 1/2 PR, the 2/2 will be the grouping/collapse PR.

### Motivation

It's confusing to look at the Workflows graph with templateRef and tasks having the same names

### Modifications

I added the ability to show the invoking template name on nodes. To keep the actual display, I added the switch button on the toolbar.
Also updated the title to include the full id and name of the node as the test is truncated.

Demo

https://github.com/user-attachments/assets/5f6ad3c0-a43f-40d0-818a-b65724974b9d

### Verification

See screen recording

### DIFF
--- a/ui/src/app/shared/components/graph/graph-panel.tsx
+++ b/ui/src/app/shared/components/graph/graph-panel.tsx
@@ -247,7 +247,9 @@ export function GraphPanel(props: Props) {
                                 .filter(([n, label]) => label.x !== null && visible(n))
                                 .map(([n, label]) => (
                                     <g key={`node/${n}`} transform={`translate(${label.x},${label.y})`}>
-                                        <title>{n}</title>
+                                        <title>
+                                            {n} ({label.label})
+                                        </title>
                                         <g
                                             className={`node ${label.classNames || ''} ${props.selectedNode === n ? ' selected' : ''}`}
                                             style={{strokeWidth: nodeSize / 15}}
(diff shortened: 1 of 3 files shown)


agilgur5:
Looks like this was combined and superseded by #13511.

When closing an issue or PR, please make sure to state the reason for it and add any links if relevant.

Adrien-D:
Sure I will do that next time. It was indeed combined and superseded by https://github.com/argoproj/argo-workflows/pull/13511. The reason was that the modified code amount was small and was very similar so it can be reviewed in one time.

PR #13511, opened 2024-08-27
feat(ui): Group nodes based on `templateRef` and show invoking template name. Fixes #11106

### LINKED ISSUE #11106: Group UI nodes based on `templateRef`

Adrien-D:
Fixes #11106.

### Motivation

It's confusing to look at the Workflows graph with templateRef and tasks having the same names. Creates linear graph without template hierarchy.

### Modifications

- Added the ability to show the invoking template name on nodes. To keep the actual display, I added the switch button on the toolbar. Also updated the title to include the full id and name of the node as the test is truncated.
- Added the grouping by templateRef using `boundaryId`. This seems to have the desired results. 

Demo

https://github.com/user-attachments/assets/763ea3b5-7349-405f-8fa9-420d9a8a091e

### Verification

I tested on simple workflows, especially the one listed on the original issue, see screen recording

### DIFF
--- a/ui/src/app/workflows/components/workflow-dag/workflow-dag.tsx
+++ b/ui/src/app/workflows/components/workflow-dag/workflow-dag.tsx
@@ -13,6 +13,7 @@ import {WorkflowDagRenderOptionsPanel} from './workflow-dag-render-options-panel
 export interface WorkflowDagRenderOptions {
     expandNodes: Set<string>;
     showArtifacts: boolean;
+    showInvokingTemplateName: boolean;
 }
 
 interface WorkflowDagProps {
@@ -36,18 +37,6 @@ function getNodeLabelTemplateName(n: NodeStatus): string {
     return n.templateName || (n.templateRef && n.templateRef.template + '/' + n.templateRef.name) || 'no template';
 }
 
-function nodeLabel(n: NodeStatus) {
-    const phase = n.type === 'Suspend' && n.phase === 'Running' ? 'Suspended' : n.phase;
-    return {
-        label: shortNodeName(n),
-        genre: n.type,
-        icon: icons[phase] || icons.Pending,
-        progress: phase === 'Running' && progress(n),
-        classNames: phase,
-        tags: new Set([getNodeLabelTemplateName(n)])
-    };
-}
-
 const classNames = (() => {
     const v: {[label: string]: boolean} = {
         Artifact: true,
@@ -65,7 +54,8 @@ export class WorkflowDag extends React.Component<WorkflowDagProps, WorkflowDagRe
         super(props);
         this.state = {
             expandNodes: new Set(),
-            showArtifacts: localStorage.getItem('showArtifacts') !== 'false'
+            showArtifacts: localStorage.getItem('showArtifacts') !== 'false',
+            showInvokingTemplateName: localStorage.getItem('showInvokingTemplateName') !== 'false'
         };
     }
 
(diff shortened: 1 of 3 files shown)


mshatkhin23:
@Adrien-D thanks for working on this! looks great

Clarifying question - do I understand correctly that collapsing is only happening for templateRefs (I.e. templates from other WorkflowTemplates)? 

If we have a task/step that references another template defined in our Workflow / Workflow Template, would that also be collapsible?

Adrien-D:
@mshatkhin23 yes collapsing will only happen on `templateRefs`

> If we have a task/step that references another template defined in our Workflow / Workflow Template, would that also be collapsible?

How do you reference it ? Can you give a very small example ?

djanjic:
Tested locally multiple use-cases and everything went well.

Adrien-D:
@mshatkhin23 It will only be collapsible by `templateRefs`, so 
in this example that won't collapse 
```
- name: user-workflow
  dag:
    tasks:
      - name: task-1
        template: dag-tier-2
      - name: task-2
        depends: 'task-1'
        template: dag-tier-2
```

whereas is this case it will collapse
```
- name: user-workflow
  dag:
    tasks:
      - name: task-1
        templateRef:
          clusterScope: true/false
          name: external-workflow-template
          template: external-dag-tier-2
          ```

**Comment by Adrien-D** [comment]
Happy to merge this one as it is and discuss about collapsing on `templates` in next issue !

File: ui/src/app/workflows/components/workflow-dag/workflow-dag.tsx [hunk 1]
```diff
@@ -100,8 +92,28 @@ export class WorkflowDag extends React.Component<WorkflowDagProps, WorkflowDagRe
         );
     }
 
+    private nodeLabel(n: NodeStatus, parent?: NodeStatus) {
+        const phase = n.type === 'Suspend' && n.phase === 'Running' ? 'Suspended' : n.phase;
+        let label = shortNodeName(n);
+
+        if (this.state.showInvokingTemplateName) {
+            label = `${parent?.templateRef?.name ?? parent?.templateName ?? n.templateName}:${label}`;
+        }
```
isubasinghe, line 101: I thought this would break for inline templates but it works, nice work!

isubasinghe (approved):
LGTM -> I tested this out and worked.

This PR has however raised some questions about the logic used in the traverse function. (You didn't author this).

isubasinghe:
TODO: Investigate why `allNodes` in workflow-dag.tsx is indexed by nodeID but then later on indexed by nodeName.

PR #13514, opened 2024-08-27
refactor: separate metrics code for reuse

Joibel:
This PR is intended as a pure refactor to move code.

It separates the reusable metrics code into something that could be reused in a separate module.

This intention is proved in a separate PR which adds metrics to the server. #13515 should be used as a reference for how this would work.

The `version` metric is moved to also be reusable as it is simple and obvious to be reused in the server.

Labels are renamed to [Attributes](https://opentelemetry.io/docs/specs/otel/common/#attribute) as per OpenTelemetry.


util/telemetry/instrument.go
```diff
@@ -155,3 +155,23 @@ func (m *Metrics) buckets(name string, defaultBuckets []float64) []float64 {
...
+	return i.otel
+}
+
+func (i *Instrument) SetUserdata(data interface{}) {
+	i.userdata = data
+}
+
+func (i *Instrument) GetUserdata() interface{} {
+	return i.userdata
+}
```
isubasinghe, line 177: Why not export these fields? 
Sure someone can mutate them, but userdata can be set, otel is an interface.

No big issues though, just a suggestion, feel free to ignore.

util/telemetry/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+	res := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceName(serviceName),
+	)
+
+	options := make([]metricsdk.Option, 0)
+	options = append(options, metricsdk.WithResource(res))
+	_, otlpEnabled := os.LookupEnv(`OTEL_EXPORTER_OTLP_ENDPOINT`)
+	_, otlpMetricsEnabled := os.LookupEnv(`OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`)
+	if otlpEnabled || otlpMetricsEnabled {
```
isubasinghe, line 56: These variables need to be added to the environment-variables.md file

isubasinghe (changes requested):
I think the environment-variables.md file needs to be updated as well.

util/telemetry/instrument.go
Joibel: Mutating the fields which are read-only would cause pretty bad damage internally. Previously these were all internal to the workflow/`metrics` package, but Userdata needs mangling by custom metrics, so I had to expose it as RW. At least it's explicit.

isubasinghe (approved):
LGTM

PR #13516, opened 2024-08-27
fix: remove non-transient logs on missing `artifact-repositories` configmap

### LINKED ISSUE #9085: trivial - reduce log spamming about artifact-repositories
I see this message printed in my argo-workflows-workflow-controller pod logs about 15 times every hour:
```time="2022-06-30T11:03:40.245Z" level=warning msg="Non-transient error: configmaps \"artifact-repositories\" not found"```

perhaps controller can change to only look for the configmap on startup of the pod?

argo version is 3.3.8

### PR DESCRIPTION
**Pull request by thor**
Fixes #9085.

### Motivation

Remove high levels of error messages that aren't errors. Reduce additional observability costs associated with running Argo Workflows, although negligble.

Clarify the difference between side-effect and non-side-effect functions in utility module.

### Modifications

Renames `IsTransientErr` to `LogIsTransientErr`, which is easily undoable if it's more noise than signal.

Introduces a `LogOnlyTransientErr`, which only logs errors if they're transient. Suitable for a use-case such as trying to fetch an object that doesn't exist when it's optional.

### Verification

No verification performed due to the (deceptive) simplicity of the change.

### DIFF
--- a/util/errors/errors.go
+++ b/util/errors/errors.go
@@ -23,12 +23,31 @@ func IgnoreContainerNotFoundErr(err error) error {
 	return err
 }
 
+// IsTransientErr reports whether the error is transient and logs it.
 func IsTransientErr(err error) bool {
+	isTransient := IsTransientErrQuiet(err)
+	if !isTransient {
+		log.Warnf("Non-transient error: %v", err)
+	}
+	return isTransient
+}
+
+// IsTransientErrQuiet reports whether the error is transient and logs only if it is.
+func IsTransientErrQuiet(err error) bool {
+	isTransient := isTransientErr(err)
+	if isTransient {
+		log.Infof("Transient error: %v", err)
+	}
+	return isTransient
+}
+
+// isTransientErr reports whether the error is transient.
+func isTransientErr(err error) bool {
 	if err == nil {
 		return false
 	}
 	err = argoerrs.Cause(err)
-	isTransient := isExceededQuotaErr(err) ||
+	return isExceededQuotaErr(err) ||
 		apierr.IsTooManyRequests(err) ||
 		isResourceQuotaConflictErr(err) ||
 		isResourceQuotaTimeoutErr(err) ||
@@ -39,12 +58,6 @@ func IsTransientErr(err error) bool {
 		matchTransientErrPattern(err) ||
 		errors.Is(err, NewErrTransient("")) ||
 		isTransientSqbErr(err)
-	if isTransient {
-		log.Infof("Transient error: %v", err)
-	} else {
-		log.Warnf("Non-transient error: %v", err)
-	}
-	return isTransient
 }
 
 func matchTransientErrPattern(err error) bool {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: util/errors/errors.go [hunk 1]
```diff
@@ -23,12 +23,14 @@ func IgnoreContainerNotFoundErr(err error) error {
 	return err
 }
 
+// As close to side-effect free of a check of the transient or non-transient state of an error.
```
Joibel, line 26: Could you format this and the other comments in here as a [godoc](https://tip.golang.org/doc/comment) compabile comment please?

util/errors/errors.go
```diff
@@ -23,12 +23,14 @@ func IgnoreContainerNotFoundErr(err error) error {
...
+// See LogIsTransientErr().
 func IsTransientErr(err error) bool {
```
Joibel, line 28: If this is now not used outside this module you can rename it to `isTransientError` to stop it being exported.

Joibel (changes requested):
The two new function names `LogIsTransientError` and `LogOnlyIsTransient` error imply things about their purpose which aren't really their primary use (logging, not a test).

They also don't immediately make sense -  two functions called LogDogName and LogOnlyDogName I wouldn't know which to call to log the dogs name. The function names you've created make sense if you understand the history. The `Is` part of their name helps, but...

I'd suggest either calling them `IsTransientErr` and `IsTransientErrVerbose`, or my slight preference would be `IsTransientErrQuiet` and `IsTransientErr`. 

Naming things [is hard](https://martinfowler.com/bliki/TwoHardThings.html).

thor:
> The two new function names `LogIsTransientError` and `LogOnlyIsTransient` error imply things about their purpose which aren't really their primary use (logging, not a test).
> 
> They also don't immediately make sense - two functions called LogDogName and LogOnlyDogName I wouldn't know which to call to log the dogs name. The function names you've created make sense if you understand the history. The `Is` part of their name helps, but... 
>[...]
> Naming things [is hard](https://martinfowler.com/bliki/TwoHardThings.html).

Thanks for the review, @Joibel, much appreciated. I solidly agree, naming things is hard.

> I'd suggest either calling them `IsTransientErr` and `IsTransientErrVerbose`, or my slight preference would be `IsTransientErrQuiet` and `IsTransientErr`.

Considering that the specific scenario wants to ignore *non-transient* errors only, do I understand you correctly that `IsTransientErrQuiet` seems appropriate although it won't be quiet, just less verbose?

util/errors/errors.go
thor: I think I've complied with the godoc style in the new commit, but let me know. 🙌

Joibel:
> EDIT: I've assumed that you indeed meant `IsTransientErrQuiet` was appropriate although it's not fully quiet, re: naming things, and simplified the PR to only switch the lookup of the configuration map `artifact-repositories` to `IsTransientErrQuiet`.

It's `Quieter`, but in the interests of sane shorter names it felt better to just call it Quiet.

Thanks, for doing this.

Joibel (approved):

thor:
@Joibel I don't think the [minimal E2E test that's failing here](https://github.com/argoproj/argo-workflows/actions/runs/10848644231/job/30106303388?pr=13516) isrelevant to the code change. Is that something you could confirm, and if so, merge the PR?

Joibel:
> @Joibel I don't think the [minimal E2E test that's failing here](https://github.com/argoproj/argo-workflows/actions/runs/10848644231/job/30106303388?pr=13516) isrelevant to the code change. Is that something you could confirm, and if so, merge the PR?

I'll re-run them and see if they pass, they look like a flakey test but I'd like to be sure.

PR #13518, opened 2024-08-28
Leo/add full name in details

### LINKED ISSUE #11949: UI: Show full (long) name of workflow in list
# Summary
So let say that you want to search (CNTRL+F) for workflow with generated id like `-zxy123` in `long-long-long-long-long-long-workflow-name-zxy123` 

But cannot pin point which is the worklow that we looking for

## Use Cases

When I go to the UI what i see is `long-long-long-lon...` (i dont know what the minimum character to make it have `...`) when search `-zxy123` it found 1 but i cannot point which is which (imagine 20-30 long-long-long-lon...) 

image

And I also expect if I opened `Show Details` it will show full name of the workflow. But it doesnt
image

---

polarbear567:
Fixes #11949

### Motivation

Issue 11949 want to add full name in details.

### Modifications

Added full name in details.

### Verification

I built argo in my local, and watched the change in ui.

### DIFF
--- a/ui/src/app/workflows/components/workflow-drawer/workflow-drawer.tsx
+++ b/ui/src/app/workflows/components/workflow-drawer/workflow-drawer.tsx
@@ -41,6 +41,10 @@ export function WorkflowDrawer(props: WorkflowDrawerProps) {
                     <div className='workflow-drawer__message--content'>{wf.status.message}</div>
                 </div>
             )}
+            <div className='workflow-drawer__section'>
+                <div className='workflow-drawer__title'>NAME</div>
+                <div className='workflow-drawer__labels'>{wf.metadata.name}</div>
+            </div>
             {!wf.status || !wf.status.conditions ? null : (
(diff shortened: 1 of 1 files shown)


agilgur5:
Looks like this was superseded by #13519.
For future reference, please use the same branch/PR when making modifications instead of opening a new one

polarbear567:
> Looks like this was superseded by #13519. For future reference, please use the same branch/PR when making modifications instead of opening a new one

Sure, I use wrong branch prefix, the check can not pass. So I close this PR.

agilgur5:
CI has no checks for branch prefixes, so I'm not sure what check you're referring to.
It looks like you used the default _title_ for the PR, which is based on branch, but you can freely edit any title of your own PRs. I also have permissions to edit other people's PR titles, which you may note I did here.

polarbear567:
Got it, as you said, my title used the default title, so the title check did not pass. I now understand, thank you.

PR #13519, opened 2024-08-28
feat(ui): add full workflow name in list details. Fixes #11949

### LINKED ISSUE #11949: UI: Show full (long) name of workflow in list

polarbear567:
Fixes #11949

### Motivation

Issue 11949 want to add full name in details.

### Modifications

Added full name in details.

### Verification

I built argo in my local, and watched the change in ui.

### DIFF
--- a/ui/src/app/workflows/components/workflow-drawer/workflow-drawer.tsx
+++ b/ui/src/app/workflows/components/workflow-drawer/workflow-drawer.tsx
@@ -41,6 +41,10 @@ export function WorkflowDrawer(props: WorkflowDrawerProps) {
                     <div className='workflow-drawer__message--content'>{wf.status.message}</div>
                 </div>
             )}
+            <div className='workflow-drawer__section'>
+                <div className='workflow-drawer__title'>NAME</div>
+                <div className='workflow-drawer__labels'>{wf.metadata.name}</div>
+            </div>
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
Can you test the responsiveness and various lengths of names?

polarbear567:
long
short
Hi @terrytangyuan 
This is screenshot. What else do I need to provide?

agilgur5:
A screenshot on a small window width -- the browser devtools has options for this: https://developer.chrome.com/docs/devtools/device-mode

In this case, it's a full row, so not sure if it's possible for it to even overflow given the name size limit in k8s

agilgur5 (approved):
Code LGTM, this is exactly what I suggested in the issue

terrytangyuan (approved):

PR #13523, opened 2024-08-28
docs: add `workflow_archive_queue` to metrics docs

Joibel:
Add workflow_archive_queue to metrics documentation and sort the list alphabetically.

This was introduced in #13418.


agilgur5 (approved):
good catch

PR #13536, opened 2024-08-30
refactor: separate workflow archive queue into own controller

ChenRussell:
### Motivation

seperate out workflow archive queue into it's own controller, to make it clear and easy to read

### Modifications

build a new controller for workflow archive queue

### Verification

run locally and tested

### DIFF
--- a/workflow/archive/archive_controller.go
+++ b/workflow/archive/archive_controller.go
@@ -1,0 +1,187 @@
+package archive
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	syncpkg "github.com/argoproj/pkg/sync"
+	log "github.com/sirupsen/logrus"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	runtimeutil "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/argoproj/argo-workflows/v3/persist/sqldb"
+	wfclientset "github.com/argoproj/argo-workflows/v3/pkg/client/clientset/versioned"
+	"github.com/argoproj/argo-workflows/v3/workflow/common"
+	"github.com/argoproj/argo-workflows/v3/workflow/hydrator"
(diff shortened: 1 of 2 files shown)


ChenRussell:
@agilgur5 @Joibel  could you review this PR? we mentioned this refactor before in https://github.com/argoproj/argo-workflows/pull/13419

workflow/archive/archive_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/argoproj/argo-workflows/v3/workflow/util"
+)
+
+type Controller struct {
+	wfclientset     wfclientset.Interface
+	wfInformer      cache.SharedIndexInformer
+	wfArchiveQueue  workqueue.RateLimitingInterface
+	hydrator        hydrator.Interface
+	wfArchive       sqldb.WorkflowArchive
+	workflowKeyLock syncpkg.KeyLock
```
Joibel, line 34: This is no longer the same lock as the one used in the main controller. It feels like this might matter, does it?

ChenRussell: Yes, I think pass the lock by pointer can fix it

workflow/archive/archive_controller.go
```diff
@@ -0,0 +1,187 @@
...
+// NewController returns a new workflow archive controller
+func NewController(ctx context.Context, wfClientset wfclientset.Interface, wfInformer cache.SharedIndexInformer, metrics *metrics.Metrics, hydrator hydrator.Interface, wfArchive sqldb.WorkflowArchive, workflowKeyLock *syncpkg.KeyLock) *Controller {
+	controller := &Controller{
+		wfclientset:     wfClientset,
+		wfInformer:      wfInformer,
+		wfArchiveQueue:  metrics.RateLimiterWithBusyWorkers(ctx, workqueue.DefaultControllerRateLimiter(), "workflow_archive_queue"),
+		metrics:         metrics,
+		hydrator:        hydrator,
+		wfArchive:       wfArchive,
+		workflowKeyLock: *workflowKeyLock,
```
Joibel, line 47: This will cause a copy operation as you're assigning to a non-pointer, so the one in use in this package will still be a separate one.

workflow/archive/archive_controller.go
Joibel: This isn't a pointer, it needs to be for this to work.

Joibel (changes requested):

workflow/archive/archive_controller.go
ChenRussell: Updated!

Joibel:
Thank you for doing this. Code inspection looks good but I'd like to find a bit of time for a local test before merging.

ChenRussell:
> Thank you for doing this. Code inspection looks good but I'd like to find a bit of time for a local test before merging.

Thanks, take your time.

Issue #13530, opened 2024-08-29
PDB JSON Schema Only Allows Number


sstaley-hioscar:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

I recently deployed argo workflows in my organization's production cluster. I also wrote a wrapper cdk8s chart that engineers and data scientists use to deploy argo workflow charts. In developing the chart, we decided to add json schema validation in order to prevent errant workflow configurations being applied to the cluster.

In doing so we noticed that the pod disruption budget json schema is incorrect. "minAvailable" and "maxAvailable" cannot be set as a percentage for workflows or k8s jobs. They must be set as a numerical value. However, the json schema requires a  string. This allows for "100%", etc, which will not work.
https://raw.githubusercontent.com/argoproj/argo-workflows/v3.5.8/api/jsonschema/schema.json
The offending definition:
```
    "io.k8s.apimachinery.pkg.util.intstr.IntOrString": {
      "type": "string"
    },
```
We changed this to "integer" in our forked schema.

related issues:
https://github.com/argoproj/argo-workflows/issues/6349
https://github.com/argoproj/argo-workflows/issues/6060
https://github.com/argoproj/argo-workflows/issues/4817#issuecomment-754819948

### Version(s)

v3.5.8

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
N/A
```

### Logs from the workflow controller

```text
N/A
```

### Logs from in your workflow's wait container

```text
N/A
```

agilgur5:
This is a follow-up [from Slack](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1722972355571829?thread_ts=1722972355.571829&cid=C01QW9QSSSK). 

> related issues:

I specifically mentioned these 3 that I had stumbled upon that detail dealing with the `IntOrString` variability in OpenAPI v2.
[...]

PR #13543, opened 2024-09-01
fix(api): `deleteDelayDuration` should be a string

### LINKED ISSUE #12631: `deleteDelayDuration` JSON schema and validation mismatch
### Pre-requisites

I'm really not understanding the syntax for `deleteDelayDuration` implemented in 3.5 https://github.com/argoproj/argo-workflows/pull/11325.
The examples use
```yaml
deleteDelayDuration: 1h
```
But the schema (https://github.com/argoproj/argo-workflows/blob/v3.5.4/api/openapi-spec/swagger.json#L14403) says it should be
```yaml
deleteDelayDuration:
  duration: 1h
```
The schema-correct option fails with a validation error in the UI.

Is the schema wrong, the docs wrong or the code wrong?  I cannot determine what is correct, although only the schema-incorrect version is submittable.

### Version

3.5.4

```YAML
kind: Workflow
metadata:
  generateName: pod-gc-strategy-
spec:
  entrypoint: main
  podGC:
    strategy: OnPodCompletion
    deleteDelayDuration:
      duration: 30s
  templates:
  - name: main
    container:
      image: alpine:3.7
      command: [sh, -c]
      args: ["exit 0"]
```

### Logs from the workflow ~controller~ server

No logs from the controller, the workflow cannot be submitted from the UI as a 400 error is returned from the Server.

```
time="2024-02-06T13:37:24.226Z" level=info duration="145.147µs" method=POST path=/api/v1/workflows/argo size=134 status=400
```

No logs, no workflow is created.

agilgur5:
Fixes #12631 , which is due to a tooling bug that I noticed in one of my first ever PRs to Argo: https://github.com/argoproj/argo-workflows/pull/11325#discussion_r1259005153

### Motivation

- the CRD tooling [gets this right](https://github.com/argoproj/argo-workflows/pull/11325/files#diff-8d25cbf1d36d3757ca1352d836cf87610c72c2ebdbadae87ec160dee7eb39223) because the k8s `apimachinery` package has code that `kube-openapi` generator understands: https://github.com/kubernetes/apimachinery/blob/2465dc5239ab8827a637148a78b380c278b4a5f4/pkg/apis/meta/v1/duration.go#L65
- meanwhile, the API tooling is different and not k8s-specific, so it doesn't know how to properly handle this type
  - this results in the API spec showing an incorrect schema, the one of the unmarshaled struct, instead of a plain string

### Modifications

- workaround the tooling issues by declaring it as a string and parsing the string to a `time.Duration` ourselves
  - i.e. instead of the k8s `metav1.Duration` struct's unmarshaling

- re-use `ParseStringToDuration` util func
  - rename its internal variables to not be specific to suspend as it's used by other fields as well
  - remove the duplicate func in `operator.go`
- add warning if an error occurs during parsing

### Notes to Reviewers

This fix is backward-compatible as it does not affect the CRDs (which were correct already) and only fixes the broken API piece.

### Future Work

This should be changed to `deleteDelaySeconds` per [k8s convention](https://github.com/kubernetes/community/blob/fb55d44/contributors/devel/sig-architecture/api-conventions.md#units) as mentioned in https://github.com/argoproj/argo-workflows/issues/12631#issuecomment-2239327945. That would also obviate the need for this entirely as well. Changing the other durations in the spec to seconds would also remove the need for duration parsing code as well


agilgur5:
Codegen is [failing](https://github.com/argoproj/argo-workflows/actions/runs/10649817321/job/29520461523?pr=13543) and seemingly resulting in the rest of the errors.

The diff makes sense, but I oddly can't seem to get my local env to reproduce it. might be time to rebuild the devcontainer from scratch 🤷

agilgur5:
> might be time to rebuild the devcontainer from scratch 🤷

yea that fixed it. and sped up my devcontainer a bit for whatever reason

terrytangyuan (commented):
FYI you can make the PR as draft when you create the PR.

agilgur5:
I know, my `codegen` was wacky locally (it did a partial generation somehow), so I didn't know it was incorrect until CI showed that

juliev0 (approved):

juliev0:
Approved. Out of curiosity, what are all the ways that we're using the json schema? Is it just being used in the Swagger documentation or more than that?

agilgur5:
The SDKs are generated off of it as well (and apparently the field docs too??), as the generated changes here show. 

It's also used for [IDE schema support](https://argo-workflows.readthedocs.io/en/release-3.5/ide-setup/) and the schema support in the UI editor. And people use it with other generators and tooling too, like Jsonnet in the [Slack thread](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1723635558441409) I linked in the issue, and like `cdk8s` in https://github.com/argoproj/argo-workflows/issues/13530

Joibel:
I agree. I did the release of 3.5.11 but this shouldn't have been in it. It should have been marked with `!` in the title to prevent it being backported as it is breaking.

@philBrown we don't follow (semver)[https://argo-workflows.readthedocs.io/en/latest/releases/#versioning] but this should have waited for 3.6.

@agilgur5, @juliev0, @terrytangyuan - I'm happy to revert and release a 3.5.12 undoing this. Opinions?

philBrown:
FYI I had to update our code with something like

```go
if duration, err := wf.Spec.PodGC.GetDeleteDelayDuration(); err == nil {
    //...
}
```
Thanks for the utility function

agilgur5:
Per the PR description, it's not breaking:

> ### Notes to Reviewers
> This fix is backward-compatible as it does not affect the CRDs (which were correct already) and only fixes the broken API piece.

The types did not change as it was always a string, except in the API, where it did not work.

philBrown:
> it's not breaking

@agilgur5 I beg to differ...

- I have consumer code using the Golang module
- I updated the patch release
- My code is now broken

agilgur5:
> using the Golang module

Internal Go types are internal. Golang doesn't have a way of defining "published" or "not published" given you can use any Git repo as a source. There are no boundaries defined there: if the entire module were considered "public", then many many internal changes and refactors would be considered breaking.

The [Public API](https://argo-workflows.readthedocs.io/en/release-3.5/public-api/) is defined by schemas, such as the CRD and OpenAPI schema.
There is a [Golang SDK](https://argo-workflows.readthedocs.io/en/release-3.5/client-libraries/#auto-generated-client-libraries) which connects to the API and therefore follows its schema.

agilgur5:
> Thanks for the utility function

That is an internal helper, for reference. If we were to remove that, that should not be a breaking change, as an example. If you are using it, that would break your usage though, as we don't really have a way of regulating that to disallow usage

Using the types from the schema is a stable surface that can be relied on instead. And SDKs can be/are versioned separately. But the internal types are not versioned since they are internal. It would be a pretty big effort to make them able to be versioned too.

Joibel:
> > using the Golang module
> 
> Internal Go types are internal. Golang doesn't have a way of defining "published" or "not published" given you can use any Git repo as a source. There are no boundaries defined there: if the entire module were considered "public", then many many internal changes and refactors would be considered breaking.

For future reference, it is possible to make un-importable modules in go: https://pkg.go.dev/cmd/go#hdr-Internal_Directories

Argo Workflows hasn't done this, for better or for worse.

PR #13546, opened 2024-09-02
feat(sso): add client cert and root CA options. Fixes #13545. Fixes #7198.

### LINKED ISSUE #7198: custom root CA for argo-server SSO
# Summary

Continuing on from https://github.com/argoproj/argo-workflows/pull/6978, we should also add a config to provide a custom root CA, (rather than encouraging users to blindly trust all certs).

Here is an example config:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: workflow-controller-configmap
  namespace: argo
data:
  sso: |
    ...
    ## `rootCaFile` is just an example name
    rootCaFile: "/etc/oauth2-proxy/certs/ca.crt"
```

---

### LINKED ISSUE #13545: SSO Provider mTLS support
# Summary

Client certificate configuration option to support mutual TLS with SSO provider. Requests to an SSO provider's endpoint that enforces client certificate authentication will fail with 401 errors unless a certificate is present in the request. 

## Use Cases

* When attempting to ensure mTLS between services
* When the workflow's SSO provider's endpoint enforces client certificate authentication

---

Simple-Analysis:
Fixes #13545 and fixes #7198.

### Motivation

To enable SSO features by ensuring that communication can occur with SSO provider's endpoint that enforces client certificate authentication.

### Modifications

* Added client certificate options for httpClient used in SSO provider communications
* Added custom root CA option for httpClient used in SSO provider communications

### Verification

Unit tests to ensure certificates are consumed properly.


Simple-Analysis:
This pull request addresses a specific issue that is currently preventing us from using Argo Workflows across multiple teams, and I’d be happy to make any necessary changes or provide additional context if required.

If you have any feedback or guidance on how I can help move this forward, I’d be more than happy to assist. Thanks so much for your time and consideration!

@agilgur5 and @Joibel I noticed this PR was moved between a few milestones, so I was curious how the milestone assignment works with regard to which PRs make the cut?

server/auth/sso/sso.go
```diff
@@ -113,8 +112,20 @@ func newSso(
...
+		ClientCert:         c.ClientCert,
+		ClientKey:          c.ClientKey,
+		InsecureSkipVerify: c.InsecureSkipVerify,
+		CACert:             c.CACert,
+	}
+
+	// Create http client
+	httpClient, err := createHttpClient(httpClientConfig)
+	if err != nil {
+		log.Fatalf("failed to create HTTP client: %v", err)
```
MasonM: `log.Fatalf()` will call `os.Exit(1)`, which means the caller won't be able to handle this error properly, and it'll be difficult to write unit tests for this. Since `newSso()` can return errors, I think we should propagate the error, like we're doing elsewhere in this function:
```suggestion
		return nil, fmt.Errorf("failed to create HTTP client: %w", err)
```

config/sso.go
```diff
@@ -29,10 +29,15 @@ type SSOConfig struct {
...
-	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
 	// FilterGroupsRegex filters groups using regular expressions
 	FilterGroupsRegex []string `json:"filterGroupsRegex,omitempty"`
+	// client certificates used for mTLS with the provider
+	ClientCert         string `json:"clientCert,omitempty"`
+	ClientKey          string `json:"clientKey,omitempty"`
+	// InsecureSkipVerify skips TLS certificate verification
+	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+	// custom CA certificate file
+	CACert string `json:"caCert,omitempty"`
```
MasonM: In ArgoCD, the corresponding option is named `rootCA`, and instead of a file path, it contains the cert as a string: https://github.com/argoproj/argo-cd/blob/c377101491406203f479fb74a573269eb5c8e50c/docs/operator-manual/user-management/index.md#configuring-a-custom-root-ca-certificate-for-communicating-with-the-oidc-provider
Here's the code it uses to load the cert: https://github.com/argoproj/argo-cd/blob/c377101491406203f479fb74a573269eb5c8e50c/util/settings/settings.go#L1820-L1828

I think we should copy what ArgoCD is doing, partly for consistently, and partly because it makes it easier to configure the local dev environment to facilitate end-to-end testing using `make start UI=true PROFILE=sso`. Being able to test this end-to-end is pretty important, since there's a lot of things that can go wrong that can't be caught with unit tests. We use [Dex](https://dexidp.io/) for authentication in the dev environment, which supports HTTPS ([config reference](https://github.com/dexidp/dex/blob/9a9a90080b61d65b83deb85c10c5a261ea8ed1d5/config.yaml.dist#L54-L59)), but we don't have that configured right now: https://github.com/argoproj/argo-workflows/blob/e6fe72a7561bb6e086d6cad33aef314f01591665/manifests/components/sso/dex/dex-cm.yaml#L14-L15

It should be possible to set those config options using [configMapGenerator](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/configmapgenerator/#overriding-base-configmap-values) in `test/e2e/manifests/sso/kustomization.yaml`.

MasonM (commented):
Thanks for your contribution! I can't think of any problems with this off-hand, but without the ability to test it end-to-end, it's really hard to tell. There's a ton of edge cases with the SSO logic, partly due to the accumulation of so many config options that lead to different code paths.

For the root certification functionality, it shouldn't be difficult to add support to the local dev environment so this can be tested via `make start UI=true PROFILE=sso`, but I don't know about the client certificate functionality. I think splitting up the PR into two (one for root certificate support and one for client certificate support) would be best.

config/sso.go
Simple-Analysis: @MasonM I truly appreciate you looking at this PR and for the feedback; thank you! 

I kept both features in one PR because they share common TLS configuration patterns and are often needed together (especially in our environment).

Good catch with the rootCA! I agree with keeping it consistent. I also added rootCAFile to allow passing the rootCAs in a file, which is sometimes easier if teams follow a certain pattern for their deployments. So, to extend your suggestion, I was hoping this could also resemble how both options are supported by the x509 package. What are your thoughts?

I agree that the unit tests might not capture all edge cases, and that additional integration tests might be useful but also difficult to establish. In our case, the client certificate support is a basic requirement because all traffic must go through proxies that enforce mTLS. That said, we can't setup Workflows with OIDC support unless all requests have client certs, even for requests to the .well-known endpoint. I'd worry that integration tests for those requirements might be hard to achieve and would also be somewhat fragile. 

You're right about the testing challenges. While full mTLS proxy simulation might be complex, I'd be open to implementing the Dex HTTPS configuration you suggested for testing the rootCA functionality. Would that be a good starting point?

Let me know your thoughts – I'm more than happy to support whatever the community recommends here. I'm just glad there is some traction here so that we can get SSO enabled :)

MasonM: Thanks @Simple-Analysis! I'm nervous about adding support for both `rootCA` and `rootCAFile` if there's no way of testing either locally. I agree adding integration tests is probably not feasible, but adding support for manual testing using the local development environment is, at least for the `rootCA` option. 

As an example, I made changes to this code in https://github.com/argoproj/argo-workflows/pull/13747, which wasn't testable at that time. So I entered https://github.com/argoproj/argo-workflows/pull/13674 to update the local development environment to support a new `UI_SECURE` option to facilitate testing, then updated the former PR to add step-by-step testing instructions, along with a video recording showing the fix being tested end-to-end.

Simple-Analysis: No problem @MasonM, I understand your concern and I will start looking at support for testing in a local dev environment.

MasonM: Here's another example of similar changes to the dev environment to support testing SSO: https://github.com/argoproj/argo-workflows/pull/14894

MasonM:
FYI: @bradfordwagner entered https://github.com/argoproj/argo-workflows/pull/14989 to only add support for the `root CA` option. I confirmed it's possible to test that in the local development environment: https://github.com/argoproj/argo-workflows/pull/14989#pullrequestreview-3386387456

Simple-Analysis:
> FYI: @bradfordwagner entered #14989 to only add support for the `root CA` option. I confirmed it's possible to test that in the local development environment: [#14989 (review)](https://github.com/argoproj/argo-workflows/pull/14989#pullrequestreview-3386387456)

Thats's awesome! Looking through  #14989, it seems like I can leverage the tests that were setup to satisfy your requirements for this PR as well. There's some overlap in our implementation and it looks like  #14989 is further along than this one with regard to the approval process, so I'll standby and keep an eye on it. Once merged, I'll adjust this PR accordingly.

PR #13549, opened 2024-09-02
feat(script): executable perms. Fixes #3496

### LINKED ISSUE #3496: Error: `exec: "/argo/staging/script": permission denied`

andrzejressel:
Fixes #3496

### Motivation

I can't find a reason why script files should not be executable

### Modifications

Script files are now executable

### Verification

```yaml
kind: Workflow
metadata:
  generateName: scripts-bash-
spec:
  entrypoint: bash-script-example
  templates:
  - name: bash-script-example
    steps:
    - - name: generate
        template: gen-random-int-bash

  - name: gen-random-int-bash
    script:
      image: debian:9.4
      command: [bash, -c]
      source: |                                         # Contents of the here-script
        cat /dev/urandom | od -N2 -An -i | awk -v f=1 -v r=100 '{printf "%i\n", f + r * $1 / 65536}'
```

### DIFF
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -950,7 +950,7 @@ spec:
     - name: whalesay
       script:
         image: argoproj/argosay:v2
-        command: ["bash"]
+        command: ["bash", "-c"]
         source: |
           ls -l /argo/staging
           cat /argo/stahing/script
--- a/workflow/executor/executor.go
+++ b/workflow/executor/executor.go
@@ -267,11 +267,13 @@ func (we *WorkflowExecutor) LoadArtifacts(ctx context.Context) error {
 func (we *WorkflowExecutor) StageFiles() error {
 	var filePath string
 	var body []byte
+	var perm fs.FileMode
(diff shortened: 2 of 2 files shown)


agilgur5:
Oh, this actually duplicates #12787. That tries to solve a different problem with this same solution. The root cause there may be fixable differently, but indeed I was considering merging that for this purpose alone.

Let's go with that one then. Thanks for finding an old issue about it (I didn't know that existed!) and raising a PR for it

Issue #13496, opened 2024-08-23
v3.5.5+: Workflow stuck in `Running` but all nodes completed -- incorrect RBAC


alexpeelman:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

I have a workflow template that recursively calls a DAG and uses some conditional logic to skip/omit certain tasks. It also slaves on the built-in suspend template.

What I notice is that all nodes and pods run to completion and are in either a Succeeded, Skipped or Omitted state but the workflow status is still `Running` 

```
Name:                workflow-keeps-running
Namespace:           argo-events
ServiceAccount:      argo-workflows
Status:              Running
Conditions:          
 PodRunning          False
Created:             Fri Aug 23 15:42:49 +0200 (3 minutes ago)
Started:             Fri Aug 23 15:42:49 +0200 (3 minutes ago)
Duration:            3 minutes 26 seconds
Progress:            8/8
ResourcesDuration:   0s*(1 cpu),17s*(100Mi memory)

STEP                          TEMPLATE       PODNAME                                          DURATION  MESSAGE
 ✔ workflow-keeps-running     entrypoint                                                                                              
 ├─✔ task                     task-template  workflow-keeps-running-task-template-2223580216  3s                                      
 ├─✔ await-task               delay                                                                                                   
 ├─○ task-finished            finishing                                                                 when 'false' evaluated false  
 └─✔ task-next-iteration      entrypoint                                                                                              
   ├─✔ task                   task-template  workflow-keeps-running-task-template-2686283671  3s                                      
   ├─✔ await-task             delay                                                                                                   
   ├─○ task-finished          finishing                                                                 when 'false' evaluated false  
   └─✔ task-next-iteration    entrypoint                                                                                              
     ├─✔ task                 task-template  workflow-keeps-running-task-template-2162925756  3s                                      
     ├─✔ await-task           delay                                                                                                   
     ├─○ task-finished        finishing                                                                 when 'false' evaluated false  
[...]

PR #13551, opened 2024-09-03
fix(cli): check for completion if `finishedAt` unset during `wait`

### LINKED ISSUE #13550: `argo wait` hangs when `finishedAt` is not set on workflow
### Pre-requisites

When running `argo wait` on a workflow that was terminated or finished successfully, but did not have the `finishedAt` status set `argo wait <workflow>` hangs without response. The expected behavior is for the command to return immediately as the workflow is in a terminal state.

### Version(s)

8a67009c80e6c842836281872ab9ebdc1c2fb8a3

To my understanding a workflow can have `finishedAt` field null due to various reasons. I was able to reproduce it when the issue from https://github.com/argoproj/argo-workflows/issues/13496 was manifested.

To reproduce, create the following RBAC first:

```YAML
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test-workflow
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: test-workflow
rules:
  - apiGroups: [""]
    resources: ["pods", "configmaps", "events"]
    verbs: ["get", "create", "update", "list", "watch", "patch"]
  - apiGroups: [""]
    resources: ["secrets"]
    verbs: ["get", "create", "update", "list", "watch", "patch", "delete"]
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    verbs: ["get", "create", "update", "list", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: test-workflow
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: test-workflow
subjects:
  - kind: ServiceAccount
    name: test-workflow
```
Then submit the following workflow:
```yaml
kind: Workflow
metadata:
  generateName: test-wf-
spec:
  archiveLogs: false
  entrypoint: main-dag
  templates:
  - name: sleep
    serviceAccountName: test-workflow
    script:
      workingDir: '/tmp'
      image: 'alpine'
      command: [sh]
      source: |
        sleep 120
  - name: main-dag
    dag:
      tasks:
      - name: sleep
        template: sleep

```
Then terminate the workflow with `argo terminate` (once the sleep pod starts)
Now when `argo wait` is run on that workflow it will hang indefinitely.

We can see that in the status field for the workflow the `taskResultsCompletionStatus` for the single task of this workflow is set to false, `finishedAt` is set to `null`. 

This is the complete workflow object with the status:
```yaml
kind: Workflow
metadata:
  annotations:
      This is a simple hello world example.
    workflows.argoproj.io/pod-name-format: v2
  creationTimestamp: "2024-09-03T06:28:32Z"
  generateName: test-wf-
  generation: 6
  labels:
    workflows.argoproj.io/completed: "false"
    workflows.argoproj.io/phase: Failed
  name: test-wf-t4fhk
  namespace: argo-workflows
  resourceVersion: "535170640"
  uid: 6f114300-7b83-453c-bd08-c18e2fbe325c
spec:
  archiveLogs: false
  arguments: {}
  entrypoint: main-dag
  nodeSelector:
    node-type: workflows
  podGC:
  serviceAccountName: workflows-default
  shutdown: Terminate
  templates:
  - inputs: {}
    metadata: {}
    name: sleep
    outputs: {}
    script:
      command:
      - sh
      image: alpine
      name: ""
      resources: {}
      source: |
        sleep 120
      workingDir: /tmp
    serviceAccountName: test-workflow
  - dag:
      tasks:
      - arguments: {}
        name: sleep
        template: sleep
    inputs: {}
    metadata: {}
    name: main-dag
    outputs: {}
  tolerations:
  - effect: NoSchedule
    key: codefresh.io
    operator: Equal
    value: workflows
  ttlStrategy:
    secondsAfterCompletion: 86400
    secondsAfterFailure: 86400
    secondsAfterSuccess: 86400
status:
  artifactGCStatus:
    notSpecified: true
  artifactRepositoryRef:
    artifactRepository:
      archiveLogs: true
      s3:
        bucket: runtime-test-1-workflows-artifacts
        endpoint: s3.amazonaws.com
        region: us-east-1
        useSDKCreds: true
    configMap: artifact-repositories
    key: default-v1
    namespace: argo-workflows
  conditions:
  - status: "False"
    type: PodRunning
  finishedAt: null
  message: Stopped with strategy 'Terminate'
  nodes:
    test-wf-t4fhk:
      children:
      - test-wf-t4fhk-1469664994
      displayName: test-wf-t4fhk
      finishedAt: "2024-09-03T06:30:08Z"
      id: test-wf-t4fhk
      name: test-wf-t4fhk
      outboundNodes:
      - test-wf-t4fhk-1469664994
      phase: Failed
      progress: 0/1
      startedAt: "2024-09-03T06:28:32Z"
      templateName: main-dag
      templateScope: local/test-wf-t4fhk
      type: DAG
    test-wf-t4fhk-1469664994:
      boundaryID: test-wf-t4fhk
      displayName: sleep
      finishedAt: "2024-09-03T06:30:08Z"
      hostNodeName: ip-10-146-65-184.ec2.internal
      id: test-wf-t4fhk-1469664994
      message: 'workflow shutdown with strategy:  Terminate'
      name: test-wf-t4fhk.sleep
      phase: Failed
      progress: 0/1
      startedAt: "2024-09-03T06:28:32Z"
      templateName: sleep
      templateScope: local/test-wf-t4fhk
      type: Pod
  phase: Failed
  progress: 0/1
  startedAt: "2024-09-03T06:28:32Z"
  taskResultsCompletionStatus:
    test-wf-t4fhk-sleep-1469664994: false
```

I realize that the task completion is a separate issue (mentioned above) - but there is also faulty logic in wait that relies only on the `finishedAt` status - when there are edge cases where the workflow has a terminal phase but `finishedAt` is not set.

```text
Nothing relevant appears in controller logs
```

```text
Not relevant for this issue
```

ilia-medvedev:
Fixes #13550 

### Motivation

Described in the issue #13550

### Modifications

Changed wait to rely on workflow phase, in addition to finishedAt field

### Verification

Tested locally on various worklfows

### DIFF
--- a/cmd/argo/commands/common/wait.go
+++ b/cmd/argo/commands/common/wait.go
@@ -69,7 +69,16 @@ func waitOnOne(serviceClient workflowpkg.WorkflowServiceClient, ctx context.Cont
 			continue
 		}
 		wf := event.Object
-		if wf != nil && !wf.Status.FinishedAt.IsZero() {
+
+		isWorkflowInTerminalState := func(wf *wfv1.Workflow) bool {
+			if !wf.Status.FinishedAt.IsZero() {
+				return true
+			} else {
+				return wf.Status.Phase == wfv1.WorkflowFailed || wf.Status.Phase == wfv1.WorkflowError || wf.Status.Phase == wfv1.WorkflowSucceeded
+			}
+		}
+
+		if wf != nil && isWorkflowInTerminalState(wf) {
 			if !quiet {
 				fmt.Printf("%s %s at %v\n", wfName, wf.Status.Phase, wf.Status.FinishedAt)
 			}
--- a/cmd/argo/commands/common/wait.go
+++ b/cmd/argo/commands/common/wait.go
@@ -70,15 +70,7 @@ func waitOnOne(serviceClient workflowpkg.WorkflowServiceClient, ctx context.Cont
 		}
 		wf := event.Object
 
-		isWorkflowInTerminalState := func(wf *wfv1.Workflow) bool {
-			if !wf.Status.FinishedAt.IsZero() {
-				return true
-			} else {
-				return wf.Status.Phase == wfv1.WorkflowFailed || wf.Status.Phase == wfv1.WorkflowError || wf.Status.Phase == wfv1.WorkflowSucceeded
-			}
-		}
-
-		if wf != nil && isWorkflowInTerminalState(wf) {
+		if wf != nil && wf.Status.Phase.Completed() {
 			if !quiet {
 				fmt.Printf("%s %s at %v\n", wfName, wf.Status.Phase, wf.Status.FinishedAt)
 			}


cmd/argo/commands/common/wait.go
```diff
@@ -69,7 +69,16 @@ func waitOnOne(serviceClient workflowpkg.WorkflowServiceClient, ctx context.Cont
 			continue
 		}
 		wf := event.Object
-		if wf != nil && !wf.Status.FinishedAt.IsZero() {
+
+		isWorkflowInTerminalState := func(wf *wfv1.Workflow) bool {
+			if !wf.Status.FinishedAt.IsZero() {
+				return true
+			} else {
+				return wf.Status.Phase == wfv1.WorkflowFailed || wf.Status.Phase == wfv1.WorkflowError || wf.Status.Phase == wfv1.WorkflowSucceeded
```
Joibel, line 77: This is available as `wf..Status.Phase.Completed()`

cmd/argo/commands/common/wait.go
```diff
@@ -69,7 +69,16 @@ func waitOnOne(serviceClient workflowpkg.WorkflowServiceClient, ctx context.Cont
 			continue
 		}
 		wf := event.Object
-		if wf != nil && !wf.Status.FinishedAt.IsZero() {
+
+		isWorkflowInTerminalState := func(wf *wfv1.Workflow) bool {
```
Joibel, line 73: Can `FinishedAt` ever be non-zero when the Phase isn't completed? Is that a redundant check now?

Joibel (dismissed):

agilgur5 (commented):
Per the analysis in your issue, it sounds like this is a bug in the Controller not setting `finishedAt`, no?

I suppose adding defensive code is not harmful, but it doesn't solve the root cause

cmd/argo/commands/common/wait.go
ilia-medvedev-codefresh: Thanks for the suggestion! I switched the condition to use it

cmd/argo/commands/common/wait.go
ilia-medvedev-codefresh: I believe you are right and it is redundant, removed this check. Now only checking for the phase

agilgur5 (changes requested):
Looks like @jswxstw root-caused this in https://github.com/argoproj/argo-workflows/issues/13550#issuecomment-2328180340, and the Workflow is actually stuck in `Running` on an older version

Since it is stuck in `Running`, this fix would not help. 


We could re-word this as a plain refactor I suppose -- it's not a fix.
I am a bit concerned that this could potentially _conceal_ a bug that'd we'd want to be very loud, but the print statement does still display `finishedAt`, which would be `nil` in such a case and therefore still look erroneous. Although a hanging `wait` is potentially more likely to cause someone to file an issue, which is a good thing in this case (error loudly, not silently)

jswxstw:
> Looks like @jswxstw root-caused this in https://github.com/argoproj/argo-workflows/issues/13550#issuecomment-2328180340, and the Workflow is actually stuck in Running on an older version

#13550 and #13496 are both introduced by legacy patch pods fallback, and they are not fixed now, may be fixed by #13100.

Joibel:
As you found in #13550, the reproduction turned out to be a Workflow stuck in `Running` on 3.5.4 rather than a terminal Workflow with `finishedAt` unset, and the underlying cause (the legacy patch pods fallback) was removed in
#13100 for v3.6.0. On main, `markWorkflowPhase` sets `Status.FinishedAt` in
the same branch that sets a terminal phase, so the two cannot diverge.

PR #13553, opened 2024-09-03
fix: ensure Holding keys carry all information needed

isubasinghe:
Fixes #TODO


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3766,13 +3773,26 @@ func (ss *SemaphoreStatus) LockWaiting(holderKey, lockKey string, currentHolders
...
+}
+func GetHoldingNameV2OrV1(holderKey string) string {
+	switch GetHolderKeyVersion() {
+	case HoldingNameV2:
+		return GetHoldingNameV2(holderKey)
+	default:
+		return GetHoldingNameV1(holderKey)
+	}
+}
+
```
isubasinghe, line 3792: These could move into sync_manager. That way they do not need to be  exported.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3853,21 +3870,36 @@ func (ms *MutexStatus) LockWaiting(holderKey, lockKey string, currentHolders []s
 	if i < 0 {
 		ms.Waiting = append(ms.Waiting, MutexHolding{Mutex: lockKey, Holder: currentHolders[0]})
 		return true
-	} else if mutexWaiting.Holder != currentHolders[0] {
+	} else if !(mutexWaiting.Holder == currentHolders[0]) {
```
isubasinghe, line 3873: What is this mate?

workflow/controller/operator_concurrency_test.go
```diff
@@ -1097,6 +1097,7 @@ spec:
 
 	ctx := context.Background()
 
+	os.Setenv("HOLDER_KEY_VERSION", "v1")
```
isubasinghe, line 1100: Document this maybe?

workflow/sync/sync_manager.go
```diff
@@ -240,10 +272,18 @@ func (cm *Manager) ReleaseAll(wf *wfv1.Workflow) bool {
...
-			resourceKey := getResourceKey(wf.Namespace, wf.Name, holding.Holder)
-			syncLockHolder.release(resourceKey)
+			logKey := ""
+			if v1alpha1.CheckHolderKeyVersion(holding.Holder) == v1alpha1.HoldingNameV1 {
+				resourceKey := getResourceKey(wf.Namespace, wf.Name, holding.Holder)
+				logKey = resourceKey
+				syncLockHolder.release(resourceKey)
+			} else {
+				logKey = holding.Holder
+				syncLockHolder.release(holding.Holder)
```
isubasinghe, line 282: You meant to put the syncLockHolder.release after the else statement

isubasinghe (commented):
Generally looks alright, a few minor issues

agilgur5 (commented):
In https://github.com/argoproj/argo-workflows/issues/8684#issuecomment-2292929055, you said this would be a breaking change? Then this should have a `!` in the title and the appropriate milestone.

Or are you making this backward compatible with the env var?

> NOTE: E2E tests **HAVE NOT BEEN RUN WITH THIS ENV VAR**.
> Largely due to being unable to run the E2E tests in a reliable way on my machine.

If this is intended to be required in the future, we should probably enable it during testing similar to https://github.com/argoproj/argo-workflows/pull/12413#discussion_r1468099849

isubasinghe:
> In [#8684 (comment)](https://github.com/argoproj/argo-workflows/issues/8684#issuecomment-2292929055), you said this would be a breaking change? Then this should have a `!` in the title and the appropriate milestone.
> 
> Or are you making this backward compatible with the env var?
Yeah this should mean that you can  upgrade workflows to use the new holder key name without any issue, no longer a breaking change. 

> 
> > NOTE: E2E tests **HAVE NOT BEEN RUN WITH THIS ENV VAR**.
> > Largely due to being unable to run the E2E tests in a reliable way on my machine.
> 
> If this is intended to be required in the future, we should probably enable it during testing similar to [#12413 (comment)](https://github.com/argoproj/argo-workflows/pull/12413#discussion_r1468099849)
Yeah that sounds good to me.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3766,13 +3773,26 @@ func (ss *SemaphoreStatus) LockWaiting(holderKey, lockKey string, currentHolders
...
 
+func GetHolderKeyVersion() HoldingNameVersion {
+	switch os.Getenv("HOLDER_KEY_VERSION") {
+	case "v2":
+		return HoldingNameV2
+	default:
+		return HoldingNameV1
+	}
+}
+func GetHoldingNameV2OrV1(holderKey string) string {
```
agilgur5, line 3784:
```suggestion
func GetHoldingName(holderKey string) string {
```

We can simplify the function name, no?

workflow/sync/mutex_test.go
```diff
@@ -126,6 +127,7 @@ func TestMutexLock(t *testing.T) {
 		assert.Len(t, concurrenyMgr.syncLockMap, 1)
 	})
 	t.Run("WfLevelMutexAcquireAndRelease", func(t *testing.T) {
+		os.Setenv("HOLDER_KEY_VERSION", "v1")
```
agilgur5, line 130:
```suggestion
		t.Setenv("HOLDER_KEY_VERSION", "v1")
```

Per #13036

workflow/controller/operator_concurrency_test.go
```diff
@@ -1050,6 +1050,11 @@ func TestSynchronizationForPendingShuttingdownWfs(t *testing.T) {
 }
 
 func TestWorkflowMemoizationWithMutex(t *testing.T) {
+	// This is needed because this test explicitly checks the behaviour
+	// of Holding
+	oldVersion := os.Getenv("HOLDER_KEY_VERSION")
+	os.Setenv("HOLDER_KEY_VERSION", "v1")
+	defer os.Setenv("HOLDER_KEY_VERSION", oldVersion)
```
agilgur5, line 1057: same as above

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3766,13 +3773,26 @@ func (ss *SemaphoreStatus) LockWaiting(holderKey, lockKey string, currentHolders
...
+	switch GetHolderKeyVersion() {
```
agilgur5, line 3785: Is the env var actually necessary? Since you have `CheckHolderKeyVersion`, this seems like you should always be able to use V2 unless V1 was already used

workflow/sync/mutex_test.go
agilgur5: Also the env var needs to be documented on the "Environment Variables" page of the docs

pkg/apis/workflow/v1alpha1/workflow_types.go
sarabala1979: +1

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: yes it is, it still is needed in LockAcquired calls, you are right about it not being needed for the Released calls.

workflow/controller/operator_concurrency_test.go
isubasinghe: Thanks for the suggestion, done that!

pkg/apis/workflow/v1alpha1/workflow_types.go
agilgur5: For `LockAcquired`, you can just always use V2, no?

docs/environment-variables.md
```diff
@@ -55,6 +55,7 @@ This document outlines environment variables that can be used to customize behav
...
 | `SEMAPHORE_NOTIFY_DELAY`                 | `time.Duration`     | `1s`                                                                                        | Tuning Delay when notifying semaphore waiters about availability in the semaphore                                                                                                                                                                                        |
 | `WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS` | `bool` | `true` | Whether to watch the Controller's ConfigMap and semaphore ConfigMaps for run-time changes. When disabled, the Controller will only read these ConfigMaps once and will have to be manually restarted to pick up new changes. |
+| `HOLDER_KEY_VERSION`                     | `string`            | `v1`                                                                                        | Whether to have the Holding field be populated using a reconstructed key (v1) or persistent key (v2)                       |
```
agilgur5, line 58: These are in alphabetical order.
Also I don't think a user would have any idea how to interpret this 😕 "Holding field" is not really user-facing and "reconstructed" and "persistent" are not terms a user would understand either

Per above comment, I think you can just default to V2 always. The V1 detection is just for backward-compat with older workflows

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: Yeah you can if you use `CheckHolderKeyVersion` but why? It's much cleaner to just use the env var (which you do need for `LockReleased`). Why infer information when you have it?

docs/environment-variables.md
isubasinghe: Yeah that is a valid point. Do you have any idea of how to phrase this?

I think we should always use V1, this changes how the workflow status is populated, it should be opt in at least for some time.

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: No we shouldn't do that, this will change how the workflow status  is populated. This should be opt in at least for sometime imo.

agilgur5: It does, but it is a fix and it is objectively better than the previous option.

Unlike `POD_NAMES`, there is also little to no client-side usage of this that would incur breakage

isubasinghe: Alright lets wait on a tie breaking vote here, I think @Joibel is with you on this one.

pkg/apis/workflow/v1alpha1/workflow_types.go
Joibel: Yes, do this.

Fewer exports is almost always the right answer.

pkg/apis/workflow/v1alpha1/workflow_types.go
Joibel: I think we should drop support for legacy holding keys, and just run with the new code.

Reasoning:
* Just V2 is simple
* There is a lot of code here supporting both key types
* Removing support for v1 (except on upgrade) reduces the amount of code even further.

Joibel (changes requested):
At minimum we need duplicates of the tests that are V1 only as V2 tests.

My preference is to drop V1 completely, agreeing with @agilgur5

workflow/sync/sync_manager.go
```diff
@@ -101,8 +112,17 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
 				if mutex == nil {
 					mutex := cm.initializeMutex(holding.Mutex)
 					if holding.Holder != "" {
-						resourceKey := getResourceKey(wf.Namespace, wf.Name, holding.Holder)
-						mutex.acquire(resourceKey)
+						var key string
```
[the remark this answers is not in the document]
isubasinghe: Done

isubasinghe:
legacy holding keys have been eliminated, now only v2 style keys exist. Meaning existing keys now test v2 directly.
We upgrade v1 keys to v2. There is a test for this upgrade process in place.

workflow/sync/sync_manager_test.go
```diff
@@ -865,3 +872,124 @@ status:
...
+
+func TestMutexMigration(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	kube := fake.NewSimpleClientset()
+
+	syncLimitFunc := GetSyncLimitFunc(kube)
+
+	concurrenyMgr := NewLockManager(syncLimitFunc, func(key string) {
+		// nextKey = key
```
isubasinghe, line 920: Add another test for migration with workflow level mutexes maybe ?

isubasinghe (commented):
@Joibel opinions?

workflow/sync/sync_manager_test.go
isubasinghe: Yeah wait on review

isubasinghe:
This PR is now ready, I believe I have addressed the edge cases. Unfortunately there is some ambiguity (as expected really) if a sync exist at both the template level and workflow level.

We just choose to go with workflow level in this case of ambiguity.

workflow/sync/sync_manager.go
```diff
@@ -101,8 +178,13 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
 				if mutex == nil {
 					mutex := cm.initializeMutex(holding.Mutex)
 					if holding.Holder != "" {
-						resourceKey := getResourceKey(wf.Namespace, wf.Name, holding.Holder)
-						mutex.acquire(resourceKey)
+						level, err := getWorkflowSyncLevelByName(&wf, holding.Mutex)
+						if err != nil {
+							log.Warnf("cannot obtain lock level for %s", holding.Mutex)
+							continue
+						}
```
isubasinghe, line 185: Since this could error out, it might be more resilient to actually check if we need to care about the level before calling this function.

workflow/sync/sync_manager.go
```diff
@@ -115,6 +197,7 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
 // TryAcquire tries to acquire the lock from semaphore.
 // It returns status of acquiring a lock , status of Workflow status updated, waiting message if lock is not available and any error encountered
 func (cm *Manager) TryAcquire(wf *wfv1.Workflow, nodeName string, syncLockRef *wfv1.Synchronization) (bool, bool, string, error) {
+	log.Infof("[DEBUG] called TryAcquire  with holderKey %s", getHolderKey(wf, nodeName))
```
isubasinghe, line 200: Remove this

workflow/sync/sync_manager.go
```diff
@@ -66,6 +66,77 @@ func (cm *Manager) CheckWorkflowExistence() {
...
+
+type SyncLevelType int
+
+const (
+	WorkflowLevel SyncLevelType = 1
+	TemplateLevel SyncLevelType = 2
+	ErrorLevel    SyncLevelType = 3
+)
+
+// legacy keys can be of the form
```
Joibel, line 87: Suggest `HoldingNameV1` instead of `legacy`

workflow/sync/sync_manager.go
```diff
@@ -66,6 +66,77 @@ func (cm *Manager) CheckWorkflowExistence() {
...
+			return WorkflowLevel, nil
+		}
+	}
+
+	var firstErr error
+	for _, template := range wf.Spec.Templates {
+		if template.Synchronization != nil {
+			syncLockName, err := GetLockName(template.Synchronization, wf.Namespace)
+			if err != nil {
+				firstErr = err
```
Joibel, line 125: firstErr looks to actually be lastErr

workflow/sync/sync_manager.go
```diff
@@ -66,6 +66,77 @@ func (cm *Manager) CheckWorkflowExistence() {
...
+				continue
+			}
+			checkName := syncLockName.EncodeName()
+			if lockName == checkName {
+				return TemplateLevel, nil
+			}
+		}
+	}
+	if firstErr == nil {
```
Joibel, line 134: Why store firstErr and then not use it in the error message?

Could be a `bool` instead, but prefer to spit out the error text here.

workflow/sync/sync_manager.go
```diff
@@ -86,11 +157,17 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
 				}
 
 				for _, holders := range holding.Holders {
-					resourceKey := getResourceKey(wf.Namespace, wf.Name, holders)
-					if semaphore != nil && semaphore.acquire(resourceKey) {
-						log.Infof("Lock acquired by %s from %s", resourceKey, holding.Semaphore)
+					level, err := getWorkflowSyncLevelByName(&wf, holding.Semaphore)
+					if err != nil {
+						log.Warnf("cannot obtain level for %s", holding.Semaphore)
```
Joibel, line 162: Output `err` here, embelish if needed, but don't use `err` as a boolean.

workflow/sync/sync_manager.go
```diff
@@ -101,8 +178,13 @@ func (cm *Manager) Initialize(wfs []wfv1.Workflow) {
 				if mutex == nil {
 					mutex := cm.initializeMutex(holding.Mutex)
 					if holding.Holder != "" {
-						resourceKey := getResourceKey(wf.Namespace, wf.Name, holding.Holder)
-						mutex.acquire(resourceKey)
+						level, err := getWorkflowSyncLevelByName(&wf, holding.Mutex)
+						if err != nil {
+							log.Warnf("cannot obtain lock level for %s", holding.Mutex)
```
Joibel, line 183: Same as above, use `err`

workflow/sync/sync_manager.go
Joibel: In the mutex case it cannot be held more than once, so level should always be accurate.

An error indicates a bug here doesn't it?

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3875,13 +3875,23 @@ func (ms *MutexStatus) LockWaiting(holderKey, lockKey string, currentHolders []s
 	return false
 }
 
-func (ms *MutexStatus) LockAcquired(holderKey, lockKey string, currentHolders []string) bool {
-	i, mutexHolding := ms.GetHolding(lockKey)
+// GetHoldingName returns holding names without
+// any splitting of strings.
+func GetHoldingName(holderKey string) string {
```
Joibel, line 3880: This function doesn't really do anything, shouldn't we drop it to aid readability?

workflow/sync/sync_manager_test.go
```diff
@@ -865,3 +872,360 @@ status:
...
```
Joibel, line 1113: Drop commented out code here and further down?

Joibel (changes requested):

workflow/sync/sync_manager.go
isubasinghe: Yup that is true, I just wanted any err really, shall change the name

workflow/sync/sync_manager.go
isubasinghe: Yeah we shouldn't really hit this error, it indicates a bug in `getWorkflowSyncLevel` or in the functions it calls or in the mutex logic itself.

workflow/sync/sync_manager.go
isubasinghe: Sure will do.

workflow/sync/sync_manager.go
isubasinghe: Done

workflow/sync/sync_manager.go
isubasinghe: Left it as it is.

isubasinghe:
Irrelevant tests seem to be broken since a rebase.

Joibel (approved):

Issue #12449, opened 2024-01-03
Add support for ppc64le


valen-mascarenhas14:
# Summary

Requesting the argo team create official images of argo-workflows for ppc64le?

## Use Cases

Our use case involves building Argo-workflow  on a Kubernetes cluster with ppc64le architecture. The Argo image serves as a crucial dependency for facilitating seamless integration and testing of Kubeflow pipelines on ppc64le

##  Proposal

Change build infrastructure to build ppc64le variants of argo-workflows.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

valen-mascarenhas14:
Our goal is to make workflow-controller, argocli, and argoexec images multi-arched. 

 We propose to add this 
  **platform: [ linux/amd64, linux/arm64, linux/ppc64le ]** 
  to the following line in the YAML file at [link](https://github.com/argoproj/argo-workflows/blob/main/.github/workflows/release.yaml#L29). for enhancing the image release pipeline to include ppc64le architecture support

terrytangyuan:
> The Argo image serves as a crucial dependency for facilitating seamless integration and testing of Kubeflow pipelines on ppc64le

Can you point me to where KFP relies on it?

terrytangyuan:
I meant documentation around the requirement of ppc64le.

ghatwala:
hey @terrytangyuan  we are trying to enable ppc64le pipeline in our local prow cluster , details are here in this issue- https://github.com/GoogleCloudPlatform/oss-test-infra/issues/1972#issuecomment-1842798928

valen-mascarenhas14:
We need the Argo workflow to run the e2e tests for the kubeflow pipelines . Here's a high level overview on how the test workflow would look like [link](https://github.com/kubeflow/pipelines/tree/master/test#:~:text=At%20a%20high%20level%2C%20a%20typical%20test%20workflow%20will)

terrytangyuan:
I see. If it's just as simple as adding one more platform in the workflow, then I don't see why not supporting it.

terrytangyuan:
Would you like to make the change and test if it works in your fork before submitting the PR?

ghatwala:
yes we could , requesting @valen-mascarenhas14 to try it once via fork and then submit PR here.

agilgur5:
> I see. If it's just as simple as adding one more platform in the workflow, then I don't see why not supporting it.

[...]

PR #13568, opened 2024-09-06
fix(ci): fix windows argoexec build

### LINKED ISSUE #13562: Image Releases are partially failing on `main`
### Pre-requisites

The Windows build has been failing on CI since https://github.com/argoproj/argo-workflows/pull/13145.

While the Windows build is not used often, this means that the combined, multi-arch manifest for `latest` is not released either, per https://github.com/argoproj/argo-workflows/issues/13556#issuecomment-2331780589.
This means that `latest` is outdated, and the arch-specific tags like `latest-linux-amd64` need to be used instead for an up-to-date image

### Version(s)

`main`, does not affect existing versions (yet)

```YAML
n/a
```

```text
n/a
```

```text
n/a
```

Joibel:
Fixes #13562 

### Motivation

The windows build in the release.yaml CI was broken by #13145, we need releases to work so this hack fixes it.

(CI should probably prove this is buildable rather than waiting for release to run.)

### Modifications

In #13145 the dependencies for the binaries were fixed to be technically correct - as in the binaries depended upon the .go files that they were built from.

This had the side effect of causing `pkg/apis/workflow/v1alpha1/openapi_generated.go` to get rebuilt when building `dist/argoexec`. This is technically correct behaviour.

This fails when building on windows because windows has [stupid limitations](https://github.com/microsoft/Windows-Containers/issues/361). This means that the [Makefile step](https://github.com/argoproj/argo-workflows/blob/7173a271bb9c59ca67df7a06965eb80afd37c0cb/Makefile#L351) which trys to create a symlink fails with `ln: failed to create symbolic link 'v3' -> '.': File name too long`.

Attempting to enable `LongPathsEnabled` (What year is it?) doesn't help either [inside](https://github.com/Joibel/argo-workflows/actions/runs/10735396476/job/29772583465) or [outside](https://github.com/Joibel/argo-workflows/actions/runs/10735565566/job/29773137712) the container run.

So this PR runs the windows build with a hack reverting the dependencies when building the windows `dist/argoexec` only via a Makefile variable `HACK_PKG_FILES_AS_PKGS`.

I consider this "fine" as we don't directly support development on Windows, and the reverted behaviour there if anyone really wants to use it isn't that terrible (as in people lived with it for years before #13145 fixed it)

### Verification

Verified [here](https://github.com/Joibel/argo-workflows/actions/runs/10735975366/job/29774399350)


agilgur5:
> This fails when building on windows because windows has [stupid limitations](https://github.com/microsoft/Windows-Containers/issues/361). This means that the [Makefile step](https://github.com/argoproj/argo-workflows/blob/7173a271bb9c59ca67df7a06965eb80afd37c0cb/Makefile#L351) which trys to create a symlink fails with `ln: failed to create symbolic link 'v3' -> '.': File name too long`.
> 
> Attempting to enable `LongPathsEnabled` (What year is it?) doesn't help either [inside](https://github.com/Joibel/argo-workflows/actions/runs/10735396476/job/29772583465) or [outside](https://github.com/Joibel/argo-workflows/actions/runs/10735565566/job/29773137712) the container run.

Hmm my read of the error message was that `.` is problematic or something, since `v3` and `.` are both short and so it's just giving an incorrect error message on some fallthrough case. That would explain why `LongPathsEnabled` has no effect.
Unless it's one of the files inside the directory being too long and Windows recreates or links them all (or something) during a symlink?

To be fair, the symlinking part of the build is potentially one of the weirdest pieces and can be problematic in other ways (e.g. parallelized operations per https://github.com/argoproj/argo-workflows/pull/12583#issuecomment-1913944498). I've been meaning to sit down with it and see if it can be refactored out.

Joibel:
`.` resolves to `/Users/<something>/go/src/github.com/argoproj/argo-workflows`, and `ln -s` appears when run from git bash under windows to effectively be akin to a `cp -a`. The result doesn't even look like a symlink, and it doesn't seem to behave like `mklink`.

I chose this way of doing it as expedient rather than perhaps the best choice. Potentially windows could be detected and mklink could be invoked. Docker under windows in a VM doesn't work for me (nested VMs) right now to play with this in a sane way.

Makefile
```diff
@@ -107,9 +107,19 @@ ifndef $(GOPATH)
 endif
 
 # -- file lists
-ARGOEXEC_PKG_FILES        := $(shell go list -f '{{ join .Deps "\n" }}' ./cmd/argoexec/ |  grep 'argoproj/argo-workflows/v3/' | xargs go list -f '{{ range $$file := .GoFiles }}{{ print $$.ImportPath "/" $$file "\n" }}{{ end }}' | cut -c 39-)
-CLI_PKG_FILES             := $(shell go list -f '{{ join .Deps "\n" }}' ./cmd/argo/ |  grep 'argoproj/argo-workflows/v3/' | xargs go list -f '{{ range $$file := .GoFiles }}{{ print $$.ImportPath "/" $$file "\n" }}{{ end }}' | cut -c 39-)
-CONTROLLER_PKG_FILES      := $(shell go list -f '{{ join .Deps "\n" }}' ./cmd/workflow-controller/ |  grep 'argoproj/argo-workflows/v3/' | xargs go list -f '{{ range $$file := .GoFiles }}{{ print $$.ImportPath "/" $$file "\n" }}{{ end }}' | cut -c 39-)
+HACK_PKG_FILES_AS_PKGS ?= false
+ifeq ($(HACK_PKG_FILES_AS_PKGS),false)
```
agilgur5, line 111: Would it make more sense to detect Windows instead of having the Windows Dockerfile use an env var?

agilgur5:
> we don't directly support development on Windows

We do have contributors on Windows also, including Tianchu, per #12603. That issue also involves a symlink _file_ in the codebase.

But yes I agree that reverting to the old behavior isn't that harmful

Makefile
agilgur5: We should also add a comment here, since out of context this would be very confusing to a reader

agilgur5 (approved):
Approving to fix releases in the interim (as it is Friday and after hours in Alan's time). Can make more changes in further PRs

Thanks for rigorously testing this Alan!

agilgur5:
> (CI should probably prove this is buildable rather than waiting for release to run.)

I was thinking about this too, this would require us to build the images for multiple architectures in CI, which will take more time (substantially so, the cross-compilation is quite slow). We could do that if we make the other steps not wait on it I suppose

Joibel:
This one is an easy parallel step, which will only slightly damage the number of server hamsters each build uses, so shouldn't impact too much.

agilgur5:
The E2E tests depend on the image build, so the longest build of those will cause the longest delay. I pulled the numbers in https://github.com/argoproj/argo-workflows/issues/12449#issuecomment-2257604118 and it's an order of magnitude longer (~2min -> ~20min in best case).

That's why I said that if we wanted to do this, it would have to be a separate, non-dependent step.

But also there's a limit to the amount of parallelism GH allows per account/org, so the more we add, the more likely we are to hit that limit and cause things to become serial as a result. We do seem to already hit this occasionally if we have many simultaneous PRs or if CD is doing a release (the GH jobs will be pending for a while before starting).

Also not sure if you missed it, but I did leave [one comment](https://github.com/argoproj/argo-workflows/pull/13568#discussion_r1747398484) on this PR specifically

Makefile
Joibel: If we can detect Windows and do this the long way around we could attempt a mklink there too.

If I get access to some kind of windows that let's me play I might have a go, but I'd rather right now it was awkward and people who can fix it found this and tried to fix it.

agilgur5: You can definitely detect Windows, in the same way you can detect a Linux OS, e.g. [with `uname` or `$OSTYPE`](https://stackoverflow.com/q/394230/3431180). There are some variations on that (i.e. prior to WSL etc when Cygwin and MinGW were more used), but technically the main one we need to handle is the GH + Docker one, the rest a Windows contributor could figure out

PR #13569, opened 2024-09-06
refactor(ci): move api knowledge to the matrix

Joibel:
### Motivation

Small refactor of the CI to expose differences between the matrix entries as matrix variables.

Also reduces duplication of the conditional. 

### Modifications

Move whether we build and start the API to being a matrix value rather than evaluated in 3 places

### Verification

Let's see if CI passes...?
Will check the API isn't built and started in a false case.


agilgur5 (approved):
Makes sense to me, this definitely reads better and feels more correct

PR #13570, opened 2024-09-06
feat: add support for multi user approval in suspend. fixes #12108 

### LINKED ISSUE #12108: Second-person approval of a workflow
# Summary

Record the user who creates a workflow, and the user who approves a step in the workflow using intermediate parameters, and make sure the user who approves the workflow is not the one who creates the workflow.

Currently it seems like we can record who approves a step using intermediate parameters in https://github.com/argoproj/argo-workflows/blob/master/workflow/util/util.go#L383-L386, and there's no way I can find it in the UI with the example in https://argoproj.github.io/argo-workflows/intermediate-inputs/#intermediate-parameters-approval-example and with SSO auth enabled.

## Use Cases

This is needed for compliance use cases where we need a second person to approve some operations.

---

rvandernoort:
Fixes #12108 

### Motivation

This feature adds support for multi approval in the suspend state instead of using the self-declared single-point enum strategy. This allows for better interactivity and fine-grained access to promotions for instance. 

I'm publishing this already now as I am looking to catch some early feedback as this is my first time contributing to this project. This is also one of my first larger PRs (in Go) so I appreciate any feedback that you can provide me and I will try to address any comments as quickly and thoroughly as possible.

### Modifications

Currently, the state of this PR is as follows:
- [x] introduce core including `ApproverStatus` field, an approve function, and two core test cases.
- [x] add ui changes to handle approvals
- [x] handle approvals
  - Currently functionality only accepts approval of your logged in account (in blue) (however this is hardcoded and untested) 
- [ ] use accounts instead of email string 
  - (how do I do this correctly and how do I test it? I tried to use any auth method using the devcontainer but didn't manage to get it working.)
- [ ] change timestamp to approved timestamp once approval is complete
- [ ] cli cmds?

##### Single approval
single_approve
single_approval
single_approved
single_approved_pop

##### Multi approval
double_approve
double_approval
double_approved
double_approved_pop

### Verification

- [ ] unit tests
  - [x] single approval
  - [x] multi approval
  - [ ] missing approval (not working, could use help if even possible)
- [ ] e2e tests
  - [ ] tbd... (suggestions welcome)


rvandernoort:
Managed to get the core functionality working and looking pretty decent I think. However, I'm a bit stuck on authentication and linking the approvals to users or groups. Would appreciate some help if anyone is able.

agilgur5 (commented):
> However, I'm a bit stuck on authentication and linking the approvals to users or groups.

This can be done in the API, but the current approach is insecure in a few ways by relying on `status`. I detailed this more in the issue: https://github.com/argoproj/argo-workflows/issues/12108#issuecomment-2347984091

If you just want to implement a UI flow with auth only done there, I would suggest making a UI plugin for this #6945

PR #13574, opened 2024-09-07
feat: Allow non flat addressing for withItems/withParams. Fixes #11192

### LINKED ISSUE #11192: Allow for more complex data structures used with `withParam`

weafscast:
Fixes #11192 

### Motivation

[Original PR](https://github.com/argoproj/argo-workflows/pull/11215) for the issue went stale and needed more test cases.

### Modifications

In workflow/controller/operator.go
- Created function fillReplaceMap to go through an item's value and recursively create an index/address for each nested value type. In effect implementing https://github.com/tidwall/sjson#path-syntax, but with as little impact to the structure as I could muster.

- Flattened a conditional.

In workflow/controller/operator_test.go
 - Updated processItem test to test more scenarios.

### Verification

Added tests, verified functionality locally.

Co-Author: [awilcots](https://github.com/awilcots) (Owns the stale pr, on top of which this pr has been built).

### DIFF
--- a/workflow/controller/operator_test.go
+++ b/workflow/controller/operator_test.go
@@ -6813,52 +6813,96 @@ func TestFailSuspendedAndPendingNodesAfterShutdown(t *testing.T) {
 
 func Test_processItem(t *testing.T) {
 	testCases := []struct {
+		name       string
 		taskName   string
-		task       wfv1.DAGTask
+		withParam  string
+		withItems  []wfv1.Item
 		wantedName string
 	}{
 		{
-			taskName: "task-name",
-			task: wfv1.DAGTask{
-				WithParam: `[{"number": 2, "string": "foo", "list": [0, "1"], "json": {"number": 2, "string": "foo", "list": [0, "1"]}}]`,
-			},
-			wantedName: `task-name(0:json.list.0:0,json.list.1:1,json.list:[0,"1"],json.number:2,json.string:foo,json:{"list":[0,"1"],"number":2,"string":"foo"},list.0:0,list.1:1,list:[0,"1"],number:2,string:foo)`,
+			name:       "WithParam: Complex JSON object",
+			taskName:   "task-name",
+			withParam:  `[{"number": 2, "string": "foo", "list": [0, "1"], "json": {"number": 2, "string": "foo", "list": [0, "1"]}}]`,
+			wantedName: `task-name(0:json.list.0:0,json.list.1:1,json.list:[0,"1"],json.number:2,json.string:foo,json:{"number":2,"string":"foo","list":[0,"1"]},list.0:0,list.1:1,list:[0,"1"],number:2,string:foo)`,
 		},
 		{
+			name:       "WithParam: Array of strings",
+			taskName:   "task-name",
+			withParam:  `[["arr1", "arr2"]]`,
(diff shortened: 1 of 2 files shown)


weafscast:
cc/ @juliev0 @jaen

weafscast:
@agilgur5 what's the general review cycle, can't find it in docs.

juliev0:
Hey @weafscast - sorry for delays in reviewing - this has been an issue of too few reviewers, with pretty much no reviewer dedicated to this repo. That's why I started [this](https://github.com/argoproj/argo-workflows/blob/main/community/sustainability_effort.md). 

Hopefully somebody will be able to look at your PR sometime soon, though. Thanks for contributing!

Consider the possibility of forking this repo if needed.

workflow/controller/operator_test.go
```diff
@@ -6812,19 +6812,98 @@ func TestFailSuspendedAndPendingNodesAfterShutdown(t *testing.T) {
...
+
+			tmpl, err := template.NewTemplate(string(taskBytes))
+			require.NoError(t, err, "Failed to create template")
+
+			var newTask wfv1.DAGTask
+			newTaskName, err := processItem(tmpl, tc.taskName, 0, items[0], &newTask, "")
+			require.NoError(t, err, "processItem failed")
+			assert.Equal(t, tc.wantedName, newTaskName)
+		})
+	}
```
juliev0, line 6906: Hey @weafscast - trying to understand the purpose of `processItem()` function, as I'm not really familiar with it. I'm gathering based on the test that it's the **name** of the Task that's really being set and that's it? Is it the name that makes it so that a user can reference `"{{item.tag}}"`? And is it downstream code that then processes that name to help make sense of "{{item.tag}}"`? 

As for your PR, would an example reference be: `"{{item.tag.subtag}}"`?

weafscast:
> Consider the possibility of forking this repo if needed.

Hey, I am doing this out of interest, so I am fine contributing to this one only, just wanted to understand it, since it was undocumented.

juliev0:
> > Consider the possibility of forking this repo if needed.
> 
> Hey, I am doing this out of interest, so I am fine contributing to this one only, just wanted to understand it, since it was undocumented.

A reasonable question for sure. There's no official cycle, or FIFO process unfortunately. It's all pretty much "best effort".

PR #5879, opened 2021-05-11
[WIP] feat: Add NodeName support

JohannesOlsson:
Adding support for NodeName as per #5821

Issue #13542, opened 2024-09-01
Remove gRPC + protobufs from API


agilgur5:
# Summary

The API's gRPC support is not often used and not worth the maintenance according to various maintainers. Some currently used gRPC libraries are also deprecated need upgrading, like `grpc-gateway` v1: https://github.com/argoproj/argo-workflows/issues/12565#issuecomment-2026688627.

The HTTP support is also based off the gRPC support (via `grpc-gateway`) and this can cause some inconsistencies as well, such as #12721

The protobufs only exist for the purpose of gRPC and their OpenAPI definitions can sometimes differ from the k8s CRD due to tooling differences as well, such as with #13530 and #12631. Some protobuf libraries are also deprecated and need updating, like #11053

The build complexity to support protobufs and gRPC is also non-trivial. A decent number of tools are installed in the `Makefile` solely for protobuf + gRPC codegen.

## Details

Per https://github.com/argoproj/argo-workflows/issues/11053#issuecomment-2185664503, https://github.com/argoproj/argo-workflows/issues/13530#issuecomment-2322026951, https://github.com/argoproj/argo-workflows/issues/12565#issuecomment-2026688627, gRPC is only used in the API and is a source of many maintenance headaches as well as build complexity. Given that many of the libraries are deprecated and in need of non-trivial breaking upgrades, it may be wise to just remove gRPC support entirely.

gRPC also predates Alex C, who also thought it and protobufs were too much of a maintenance hassle.

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

Joibel:
In this proposal how would we continue to support HTTP?

agilgur5:
I've seen folks suggest plain `mux` before, which might work. Alex C was deferential toward using Gin.
I don't have a particularly strong opinion here and I'm not a Go expert to have a concrete preference either.

Issue #12393, opened 2023-12-20
ArtifactGC: Inherit / Allow Specification of Scheduling Constraints


Garett-MacGowan:
# Summary

Currently, the ArtifactGC pod has no scheduling constraints inherited or specified in the artifactGC config. I would like to add scheduling constraints to the ArtifactGC pod.

I am willing to implement this myself but I would like approval of the idea before hand. TLDR proposal: inherit scheduling constraints from the workflow pod and/or allow the specification of entirely different constraints.

This could be broken up into two tickets, one for inheriting constraints, and another for explicitly defining affinity and tolerations.

## Problem

A lack of scheduling constraints is a problem for setups where all node pools in a cluster have taints that must be tolerated, making the ArtifactGC pod unschedulable.

## Use Cases

Aside from the aforementioned problem, having fine-grained control over ArtifactGC pod placement could be used for cost optimization. E.g. garbage collecting on a very cheap node pool.

## Proposed Solution

1) Default to no constraints (same as existing behavior)
```
workflowSpec:
  # Clean up artifacts https://argoproj.github.io/argo-workflows/walk-through/artifacts/#artifact-garbage-collection
  artifactGC:
    strategy: OnWorkflowDeletion
    serviceAccountName: argo-workflows-sa
```

2) Inherit constraints flag
```
workflowSpec:
  # Clean up artifacts https://argoproj.github.io/argo-workflows/walk-through/artifacts/#artifact-garbage-collection
  artifactGC:
    strategy: OnWorkflowDeletion
    serviceAccountName: argo-workflows-sa
    inheritWorkflowSchedulingConstraints: true
```

3) Allow definition of affinity & tolerations
```
workflowSpec:
  # Clean up artifacts https://argoproj.github.io/argo-workflows/walk-through/artifacts/#artifact-garbage-collection
  artifactGC:
    strategy: OnWorkflowDeletion
    serviceAccountName: argo-workflows-sa
    affinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
            - matchExpressions:
                - key: node_pool
                  operator: In
                  values:
                    - my-cheap-node-pool
    tolerations:
      - key: node_pool
        value: my-cheap-node-pool
        effect: NoSchedule
        operator: Equal
```

@juliev0 tagging you here since I know you deal with / have dealt with ArtifactGC.

---

**Message from the maintainers**:

[...]

Issue #11485, opened 2023-07-31
resources for the artgc pod


roelarents:
# Summary

Ability to set resources for the artifact garbage collection (artgc) pod.

Currently, the resources are [hard coded](https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/artifact_gc.go#L450) (to 64Mi mem and 0.1 cpu).

I could imagine being able to provide a `podSpecPatch` for the artifactgc pod with each workflow-template. And falling back to the defaults in [`mainContainer.resources`](https://argoproj.github.io/argo-workflows/workflow-controller-configmap.yaml). But perhaps there are other ways that make more sense.

@juliev0 , do you have any ideas/opinion about this as the author of the artifact gc code?

# Use Cases

We have workflows with many artifacts (magnitude of millions). The artifact garbage collection pods (`*-artgc-*`) for these get OOMKilled. Leaving the workflow in an undeletable (unless forced) state and leaving behind uncollected garbage artifacts.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

juliev0:
That's a good idea. Perhaps `podSpecPatch` could be added to [ArtifactGC](https://github.com/argoproj/argo-workflows/blob/master/pkg/apis/workflow/v1alpha1/workflow_types.go#L1047) or [WorkflowLevelArtifactGC](https://github.com/argoproj/argo-workflows/blob/master/pkg/apis/workflow/v1alpha1/workflow_types.go#L1039C6-L1039C6) struct? (the latter may be sufficient and less complicated)

PR #13575, opened 2024-09-07
feat: add support for k8s `nodeName` for scheduling

### LINKED ISSUE #5821: support k8s `nodeName` for scheduling
# Summary

argo-workflow at present only support node-selector. But If we also support `nodeName`, the k8s's scheduler will just ignore this pod, will accelerate the pod's creation.

# Use Cases

Run large scale pods on virtual kubelet. In that case, we know the virtual-kubelet's node name. And we don't need the scheduler. We wanna reduce the pressure of scheduler.

---

weafscast:
Fixes #5821 

### Motivation

Issue has been open since a while with a [stale pr](https://github.com/argoproj/argo-workflows/pull/5879).

### Modifications

Add a New property named `NodeName`

### Verification

Updated existing operator_test.go to confirm that the nodename is set.

Co-Author: [JohannesOlsson](https://github.com/JohannesOlsson) (Owns the stale pr, on top of which this pr has been built).


weafscast:
Need to fix code gen for python.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -429,6 +429,11 @@ type WorkflowSpec struct {
 	// ArtifactGC describes the strategy to use when deleting artifacts from completed or deleted workflows (applies to all output Artifacts
 	// unless Artifact.ArtifactGC is specified, which overrides this)
 	ArtifactGC *WorkflowLevelArtifactGC `json:"artifactGC,omitempty" protobuf:"bytes,43,opt,name=artifactGC"`
+
+	// NodeName is a selector which will result in all pods of the workflow
+	// to be scheduled on the selected node(s). This is able to be overridden by
+	// a nodeName specified in the template.
+	NodeName string `json:"nodeName,omitempty" protobuf:"bytes,44,opt,name=nodeName"`
```
agilgur5, line 436: this would make sense to put next to `nodeSelector` so that related fields are grouped together (also makes sense alphabetically)

agilgur5 (commented):
Does this borrow any code from the older PRs? If so, those folks should be listed as co-authors for proper credit. Same with your other PR out #13574

Also this is probably able to be worked around with `podSpecPatch` these days, so I wonder if this is worthwhile to add to the spec

pkg/apis/workflow/v1alpha1/workflow_types.go
weafscast: Done. Initially, I wanted to keep them sorted based on the proto tag, so adding them would be convenient in the future, but I also see an issue that aims to drop support for proto, so this doesn't hold value anymore.

agilgur5: Oh, the proto number is technically relevant, though I don't think this is the first time we've added a field in the middle.
And yes #13542 is likely an upcoming change

agilgur5:
> Co-Author: [JohannesOlsson](https://github.com/JohannesOlsson) (Owns the stale pr, on top of which this pr has been built).

Thanks for adding proper credit.
By co-author I literally meant the `Co-authored-by:` trailer in the commit that [GitHub recognizes](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors). In the PR description itself, I think linking to the old PR is enough

weafscast:
> > Co-Author: [JohannesOlsson](https://github.com/JohannesOlsson) (Owns the stale pr, on top of which this pr has been built).
> 
> Thanks for adding proper credit. By co-author I literally meant the `Co-authored-by:` trailer in the commit that [GitHub recognizes](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors). In the PR description itself, I think linking to the old PR is enough

Added to the commit: https://github.com/argoproj/argo-workflows/pull/13575/commits/e52303278ddc024f9a834abd34bf2f802cf08c23

server/static/files.go
agilgur5, line 1: Codegen is failing on this file; I don't think it should change for this PR either, as it's for UI files

workflow/controller/workflowpod.go
```diff
@@ -786,6 +786,13 @@ func (woc *wfOperationCtx) addSchedulingConstraints(pod *apiv1.Pod, wfSpec *wfv1
 	} else if len(wfSpec.NodeSelector) > 0 {
 		pod.Spec.NodeSelector = wfSpec.NodeSelector
 	}
+	// Set nodeName (if specified)
+	if tmpl.NodeName != "" {
+		pod.Spec.NodeName = tmpl.NodeName
+	} else if wfSpec.NodeName != "" {
```
agilgur5, line 792: Is the `boundaryTemplate` check relevant here too?

workflow/controller/operator_test.go
```diff
@@ -2929,6 +2931,7 @@ func TestWorkflowSpecParam(t *testing.T) {
 	assert.True(t, found)
 
 	assert.Equal(t, "my-host", pod.Spec.NodeSelector["kubernetes.io/hostname"])
+	assert.Equal(t, "my-node", pod.Spec.NodeName)
```
agilgur5, line 2934: A template-level test would be good as well to cover both conditionals

agilgur5:
> Also this is probably able to be worked around with `podSpecPatch` these days, so I wonder if this is worthwhile to add to the spec

Two examples of this in recent times are for artifact GC in #12393 and #11485.

`podSpecPatch` is not the most ideal though (schema-less), but is more sustainable. I'd want to hear opinions from other maintainers

weafscast:
Got it, I would wait for comments here before further fixing this pr.

Joibel:
> `podSpecPatch` is not the most ideal though (schema-less), but is more sustainable. I'd want to hear opinions from other maintainers

My preference would be to **not** adopt this change.

`podSpecPatch` covers the requirement adequately, and despite being schema-less could technically still be validated as a good option.

`nodeName` specifically is poor practice to use, and we should not encourage it. `nodeSelector` and affinity rules and tolerations all being better choices.

weafscast:
Closing this one.

PR #13589, opened 2024-09-12
chore(test): increase executor memory

isubasinghe:
### Motivation
The executor constantly OOMs locally for me when I run with `make start UI=true`. This increases the memory limits to reduce this from happening.

### Modifications
Increase the memory limits in the controller's configmap.

### Verification

Tested with `kubectl get configmaps` to make sure executor memory limits increased.

### DIFF
--- a/test/e2e/manifests/mixins/workflow-controller-configmap.yaml
+++ b/test/e2e/manifests/mixins/workflow-controller-configmap.yaml
@@ -13,7 +13,7 @@ data:
     resources:
       requests:
         cpu: 0.1
-        memory: 128Mi
+        memory: 64Mi
       limits:
         cpu: 0.5
         memory: 256Mi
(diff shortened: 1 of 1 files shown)


test/e2e/manifests/mixins/workflow-controller-configmap.yaml
```diff
@@ -13,10 +13,10 @@ data:
     resources:
       requests:
         cpu: 0.1
-        memory: 64Mi
+        memory: 128Mi
```
agilgur5, line 16: why not just increase the limit but leave the request as is?

isubasinghe: Sure, can do that.

agilgur5 (approved):

Issue #11120, opened 2023-05-23
Prune unmaintained dependencies


JPZ13:
# Summary

Similar to #11053 that @isubasinghe reported, we have many dependencies in our go.mod file (and I assume more in the package.json) that are no longer being actively maintained. We should start culling these dependencies. It gets us the following advantages:
- Fewer attack vectors
- Less code to maintain
- Easier to modify behavior
- Less need to maintain forks of dependencies

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

JPZ13:
Running list of unmaintained go packages in Argo Workflows dependencies:
- github.com/Knetic/govaluate
- github.com/blushft/go-diagrams
- github.com/doublerebel/bellows
- github.com/gorilla/handlers
- github.com/gorilla/websocket
- github.com/robfig/cron
- github.com/sirupsen/logrus
- github.com/skratchdot/open-golang
- github.com/tidwall/gjson
- github.com/valyala/fasttemplate
- github.com/xeipuuv/gojsonschema

ryancurrah:
For logging, in Go 1.21 there will be structured logger in the standard library.

Proposal: https://www.reddit.com/r/golang/comments/11sdqia/slog_proposal_accepted_for_go_121.

You can even use it now via [golang.org/x/exp/slog](https://pkg.go.dev/golang.org/x/exp/slog).

tidwall:
@JPZ13 Just out of curiosity what metric did you use to determine which packages are unmaintained?

JPZ13:
@tidwall - eyeballed a few and then also used this: https://isitmaintained.com/

gjson was showing up as [unmaintained](https://isitmaintained.com/project/tidwall/gjson), so I included it. Happy to scratch from the list if it's a false positive

tidwall:
@JPZ13 Thanks for the info. I consider gjson a maintained project.

tico24:
You can find EOL packages by using https://github.com/xeol-io/xeol

terrytangyuan:
Gorilla projects should be back alive. https://github.com/gorilla/mux/issues/707#issuecomment-1629838706

jamietanna:
That's cool @tico24 - I've built a similar thing at https://dmd.tanna.dev using endoflife.date as well as community-sourced data!

weafscast:
[...]

PR #13591, opened 2024-09-12
refactor: remove `bellows` dep

weafscast:
Partial fix for #11120

### Motivation

The library used for this function was unmaintained, and since it's a simple util it's easier to write it on our own.

SideNote: This is a fresh set of code, doesn't import library so we are good with license here.

### Modifications

- Added our own expansion and flattening util functions along with the tests 
- removed bellow library from dependency.

### Verification

Added unit test cases

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	github.com/blushft/go-diagrams v0.0.0-20201006005127-c78c821223d9
 	github.com/colinmarc/hdfs/v2 v2.4.0
 	github.com/coreos/go-oidc/v3 v3.9.0
-	github.com/doublerebel/bellows v0.0.0-20160303004610-f177d92a03d3
 	github.com/evanphx/json-patch v5.8.0+incompatible
 	github.com/expr-lang/expr v1.16.9
 	github.com/gavv/httpexpect/v2 v2.16.0
--- a/go.sum
+++ b/go.sum
@@ -228,8 +228,6 @@ github.com/docker/docker-credential-helpers v0.7.0 h1:xtCHsjxogADNZcdv1pKUHXryef
 github.com/docker/docker-credential-helpers v0.7.0/go.mod h1:rETQfLdHNT3foU5kuNkFR1R1V12OJRRO5lzt2D1b5X0=
 github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96/go.mod h1:Qh8CwZgvJUkLughtfhJv5dyTYa91l1fOUCrgjqmcifM=
 github.com/docopt/docopt-go v0.0.0-20180111231733-ee0de3bc6815/go.mod h1:WwZ+bS3ebgob9U8Nd0kOddGdZWjyMGR8Wziv+TBNwSE=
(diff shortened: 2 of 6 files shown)


weafscast:
I am not sure if this needs a new issue all-together?

cc/ @agilgur5

ryancurrah:
+1 for little bit of copying.

terrytangyuan (commented):
Could you add a link to the reference implementation/original source?

agilgur5 (commented):
> +1 for a little bit of copying.

Personally, I'd rather not us maintain more, especially when we already lack contributors. Also the tool is tiny, so even if it is deprecated, we may not need any updates to it anyway.

It looks like Alex C opened a few issues in the `bellow` repo a few years ago, most notably https://github.com/doublerebel/bellows/issues/9, pointing to this fork: https://github.com/integration-system/bellows

> SideNote: This is a fresh set of code, doesn't import library so we are good with license here.

If you copied any code, then it would be under their license, but if not, then it'd be under Argo's

> I am not sure if this needs a new issue all-together?

This is fine, I edited your description to reference the other issue

weafscast:
> Could you add a link to the reference implementation/original source?

There is no reference that i have used. I have made sure to run the test against this as well to make sure the behaviour matches. https://github.com/doublerebel/bellows

weafscast:
> Personally, I'd rather not us maintain more, especially when we already lack contributors. Also the tool is tiny, so even if it is deprecated, we may not need any updates to it anyway.

I would be slightly biased towards writing simple utils instead of using a library for the same as long as comprehensive test are there for util.

isubasinghe:
With @agilgur5 on this one, lets keep bellows for now until expressions as a whole are tackled.

weafscast:
I would slightly disagree here. I would rather have tested unmaintained code (what this PR would become in the worst case) than non-tested non-maintained code(the library that we use). 

Independently we should merge this -> https://github.com/argoproj/argo-workflows/pull/13591/files#diff-193ec69866aa16c7ccb83764540afe3ddeea0b19cd2c268a17fb14b65bce5239

agilgur5:
> I would slightly disagree here. I would rather have tested unmaintained code (what this PR would become in the worst case) than non-tested non-maintained code(the library that we use).

> It looks like Alex C opened a few issues in the `bellow` repo a few years ago, most notably [doublerebel/bellows#9](https://github.com/doublerebel/bellows/issues/9), pointing to this fork: https://github.com/integration-system/bellows

This fork has some tests. I didn't realize it didn't have tests, although that can be remedied in a fork

> Independently we should merge this ->

Feel free to split it off into a separate PR

isubasinghe:
@weafscast yeah I think adding the tests as an independent PR is something I would be okay with. 

I do appreciate your contribution, its just that expressions need some rethinking and it doesn't make sense to make new changes (especially ones that may add more maintenance effort) to expressions related libraries when we don't know  what the state of expressions would look like in the next 6 months or so.

weafscast:
Splitted tests pr -> https://github.com/argoproj/argo-workflows/pull/13664/files

Closing this pr for the meanwhile.

Issue #13410, opened 2024-07-29
Invalid javascript depedencies


xnox:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

Building v3.5.0-576-g1239fd022 (main branch)

Invoking `make ui/dist/app/index.html` prints many javascript dependency warnings:

```console
JOBS=max yarn --cwd ui install
yarn install v1.22.22
[1/5] Validating package.json...
[2/5] Resolving packages...
warning Resolution field "types-ramda@0.29.4" is incompatible with requested version "types-ramda@^0.30.0"
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning " > argo-ui@1.0.0" has incorrect peer dependency "react@^16.9.3".
warning " > argo-ui@1.0.0" has incorrect peer dependency "react-dom@^16.9.3".
warning "argo-ui > foundation-sites@6.8.1" has unmet peer dependency "jquery@>=3.6.0".
warning "argo-ui > foundation-sites@6.8.1" has unmet peer dependency "motion-ui@latest".
warning "argo-ui > foundation-sites@6.8.1" has unmet peer dependency "what-input@>=5.2.10".
warning "argo-ui > react-autocomplete@1.8.1" has incorrect peer dependency "react@^0.14.7 || ^15.0.0-0 || ^16.0.0-0".
warning "argo-ui > react-autocomplete@1.8.1" has incorrect peer dependency "react-dom@^0.14.7 || ^15.0.0-0 || ^16.0.0-0".
warning "argo-ui > react-form@2.16.3" has incorrect peer dependency "react@^15.6.1 || 16".
warning "argo-ui > react-form@2.16.3" has incorrect peer dependency "react-dom@^15.6.1 || 16".
warning "argo-ui > react-form > react-redux@5.1.2" has incorrect peer dependency "react@^0.14.0 || ^15.0.0-0 || ^16.0.0-0".
warning "argo-ui > react-form > redux-thunk@2.4.2" has incorrect peer dependency "redux@^4".
[...]

PR #13593, opened 2024-09-13
fix(ui): handle React 18 batching for "Submit" button on details page. Fixes #13453

### LINKED ISSUE #13453: 3.6.0-rc1 UI: Unable to click Submit button on WorkflowTemplate, ClusterWorkflowTemplate, and CronWorkflow detail.
### Pre-requisites

When I create a new WorkflowTemplate from the UI, the Submit button remains Disabled and cannot be pressed.
The button does not become Enabled even if I perform operations such as Modify->Update.
image

### Version(s)

16f0a8ea3cc0690d32994d309ac6b6e26a9409ca

```YAML
metadata:
  name: delightful-tiger
  namespace: argo
  labels:
    example: 'true'
spec:
  workflowMetadata:
    labels:
      example: 'true'
  entrypoint: argosay
  arguments:
    parameters:
      - name: message
        value: hello argo
  templates:
    - name: argosay
      inputs:
        parameters:
          - name: message
            value: '{{workflow.parameters.message}}'
      container:
        name: main
        image: argoproj/argosay:v2
        command:
          - /argosay
        args:
          - echo
          - '{{inputs.parameters.message}}'
  ttlStrategy:
    secondsAfterCompletion: 300
  podGC:
    strategy: OnPodCompletion
```

```text
N/A
```

```text
N/A
```

MasonM:
Fixes #13453 

### Motivation

The React 18 upgrade in https://github.com/argoproj/argo-workflows/pull/13069 introduced a bug on the details page where the "Submit" button is disabled immediately on page load. I believe this is happening due to [batching changes in React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching) that affect the following two `useEffect()` calls, which are common to all the components modified in this PR:
https://github.com/argoproj/argo-workflows/blob/729ac174618696e35324e16a739088a2f06aa0ec/ui/src/app/event-sources/event-source-details.tsx#L68-L81

The first `useEffect()` call runs immediately and invokes `setEventSource(newEventSource)`, which causes the second `useEffect()` call to be fired, since the `eventSource` dependency has changed. Since both are invoking `setEdited()`, this is basically a race condition, since the state of `edited` is going to depend on whether these calls are batched together (the behavior in React 18) or fired separately (the old behavior).

### Modifications

This PR turns the `edited` variable into derived state that dynamically compares objects using `JSON.stringify()`, which eliminates the need for `useEffect(() => setEdited(true), [template])` call, and therefore fixes the  race condition. 

This will have a slight performance hit, which we could solve using `useMemo()`, but I wasn't sure if that was worth it.

### Verification

Tested locally by editing a template at http://localhost:8081/workflow-templates?namespace=argo:

https://github.com/user-attachments/assets/537e3d5b-ec67-4d5a-8cb3-0e551e87ae54


ui/src/app/shared/components/object-parser.ts
```diff
@@ -10,3 +10,7 @@ export function parse<T>(value: string): T {
 export function stringify<T>(value: T, type: string) {
     return type === 'yaml' ? jsyaml.dump(value, {noRefs: true}) : JSON.stringify(value, null, '  ');
 }
+
+export function isEqual<T>(value1: T, value2: T) {
```
MasonM, line 14: I could replace this with [fast-deep-equal](https://www.npmjs.com/package/fast-deep-equal), which is already a transitive dependency: https://github.com/argoproj/argo-workflows/blob/8065edb34df0e188fd88181b4660c705997e6dcf/ui/yarn.lock#L4654

but `docs/CONTRIBUTING.md` says there has to be a "strong use case" to add a new dependency, and I'm not sure this qualifies.

agilgur5:
Hmm, so per https://github.com/argoproj/argo-workflows/issues/13410#issuecomment-2257365014 and https://github.com/argoproj/argo-workflows/pull/13069#issuecomment-2159579279, I was planning to revert the React upgrade on `main` (and did not cherry-pick it). 

That being said, the difference in behavior should be handled regardless, but this may be lower priority as a result

ui/src/app/shared/components/object-parser.ts
agilgur5: Nice work cross-checking the dependencies policy! `fast-deep-equal` hasn't had a release in 4 years and seems abandoned per https://github.com/epoberezkin/fast-deep-equal/issues/145 so I think there's other reasons to not rely on it directly as well.

A one-liner util is probably good enough as-is, though I'd add a JSDoc comment mentioning it uses `JSON.stringify` as a caveat so that it is visible in IDEs during usage.

Also I don't think this specifically belongs in `object-parser.ts`, since this file is all YAML helper functions that are primarily used in `object-editor.tsx`

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
```diff
@@ -47,7 +48,11 @@ export function ClusterWorkflowTemplateDetails({history, location, match}: Route
         [history]
     );
 
-    useEffect(() => setEdited(true), [template]);
+    const resetTemplate = (template: ClusterWorkflowTemplate) => {
+        setTemplate(template);
+        setInitialTemplate(template);
+    };
```
agilgur5, line 54:
```suggestion
    function resetTemplate(template: ClusterWorkflowTemplate) {
        setTemplate(template);
        setInitialTemplate(template);
    }
```

use named functions instead of consts assigned to anonymous functions for better tracing per https://github.com/argoproj/argo-workflows/pull/12062

also helper functions and derivations are conventionally placed before all the `useEffect` calls

ui/src/app/shared/components/object-parser.ts
agilgur5: Oh, each details page does transitively use `object-editor`, so maybe this makes sense as a location actually 🤔

ui/src/app/sensors/sensor-details.tsx
```diff
@@ -59,12 +60,16 @@ export function SensorDetails({match, location, history}: RouteComponentProps<an
         services.sensor
             .get(name, namespace)
             .then(setSensor)
```
agilgur5, line 62:
```suggestion
            .then(resetSensor)
```

shouldn't this also use `resetSensor`? this is the initial retrieval, so it would make sense to set the initial one here.
also I think this is the odd one out of all the other changed components

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
agilgur5: (same ofc applies to the other changed files as well)

agilgur5 (commented):
Nice root cause analysis! We might want to double-check some other places for batching bugs too 🤔 seems like React didn't have any auto-detection for this unfortunately


I left some mostly stylistic comments below, with one behavior correction.

> This will have a slight performance hit, which we could solve using `useMemo()`, but I wasn't sure if that was worth it.

Regarding `useMemo`, in this case I think it would be worth it, since these are all editor components, so they are latency sensitive to user input

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
MasonM: Makes sense. Done in 2747d1755c11b49749fb6780bcb2a46f2f4b81e4

ui/src/app/sensors/sensor-details.tsx
MasonM: You're right. Thanks for the catch! Fixed in 1ef4958e4469e8ebd41719a81a45201385e7e476

ui/src/app/shared/components/object-parser.ts
MasonM: That makes sense. I added a comment in 2747d1755c11b49749fb6780bcb2a46f2f4b81e4. 

You're right that this file is transitively included by each component, and I couldn't think of a better place to put this.

MasonM:
@agilgur5 Thanks for the quick and thorough review! I pushed a couple commits to fix the stylistic issues, switch to `useMemo()`, and fix the bug on `sensor-details.tsx`

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
```diff
@@ -47,7 +55,6 @@ export function ClusterWorkflowTemplateDetails({history, location, match}: Route
         [history]
     );
 
-    useEffect(() => setEdited(true), [template]);
```
agilgur5, line 50: Was thinking about the `useMemo`'s dependencies list, and I'm thinking that we could simplify it even further if we only wanted to match this logic: given that the dependencies are object refs, this is pure ref equality, so we could technically simplify `edited` to `template !== initialTemplate`.

The current memoization logic is almost that, just that when the refs are equal, it will do the JSON comparison. But because we use `setTemplate` on any edit, I think it is effectively exactly the same. So I'm pretty sure the simplification would be equivalent.

What do you think?

(something something "react compiler" something something "dependencies list are too easy to make mistakes with" something something)

MasonM: I tried that initially, but there's a couple problems with relying on object equality:
1. If you make an edit, then undo that edit, `edited` will remain true.
2. If you click "Update", `edited` will remain true. I think this is happening because of how the `<ObjectEditor>` serializes/deserializes the value. When "Update" is fired, the `<ObjectEditor>`  component will be re-rendered with the new value, which gets serialized here: https://github.com/argoproj/argo-workflows/blob/6cac182ff247ff37a13b8efc6cb4ff11fcdd20c3/ui/src/app/shared/components/object-editor.tsx#L25
 Then, the `onChange` handler is fired (which is bound to `setTemplate()`), which parses it back to an object: https://github.com/argoproj/argo-workflows/blob/6cac182ff247ff37a13b8efc6cb4ff11fcdd20c3/ui/src/app/shared/components/object-editor.tsx#L132
  Even though the object is going to have the same properties, it's still considered distinct.

ui/src/app/cron-workflows/cron-workflow-details.tsx
```diff
@@ -36,9 +37,16 @@ export function CronWorkflowDetails({match, location, history}: RouteComponentPr
     const [columns, setColumns] = useState<models.Column[]>([]);
 
     const [cronWorkflow, setCronWorkflow] = useState<CronWorkflow>();
-    const [edited, setEdited] = useState(false);
+    const [initialCronWorkflow, setInitialCronWorkflow] = useState<CronWorkflow>();
```
Adrien-D, line 40: Nice analysis ! You could use a Ref for `initialCronWorkflow` as you don't expect a rerendering in your case.

Also you could implement a more reusable logic in a separate testable hook as you reuse it in multiple places

I was thinking about something like (not tested) : 

```
const useStateIsEdited = function <T>(initialStateValue?: T) {
  const [initialValue, setInitialValue] = useState<T | undefined>(initialStateValue)
  const [value, setValue] = useState<T>()

  const edited = useMemo(() => !isEqual(value, initialValue), [value, initialValue])

  return [value, setValue, setInitialValue, edited] as const
}
```

That we reuse everywhere by calling
```
const [cronWorkflow, setCronWorkflow, setInitialCronWorkflow,, edited] = useStateIsEdited<T>()
```

Otherwise LGTM

agilgur5: Yea I was thinking we could refactor a hook here too. It's a pretty time one, but perhaps worth the encapsulation. I don't really have a strong opinion on it

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
agilgur5: Yes, that's correct; this would've been the previous behavior (prior to this bug) as well if I'm not mistaken.
So while the JSON equality is more accurate than the previous behavior, it does have a performance trade-off, which is perhaps not worthwhile solely for the sake of UI button accuracy.

> just that when the refs are equal, it will do the JSON comparison. But because we use `setTemplate` on any edit,

I did make a mistake in my previous statement^ though, "equal" -> "unequal". Meaning the JSON comparison will pretty much happen on every change, so the `useMemo` is effectively not used as the refs are never equal.

Since this is an input box that is latency sensitive, I'm thinking that simplifying it to the previous behavior, even if it's not entirely accurate, would be worthwhile compared to a JSON comparison on each change, as the `useMemo` is effectively ignored.

MasonM: I checked, and on the `main` branch, the `stringify()` function is called 4 times for every change made in `<ObjectEditor>`:
1. https://github.com/argoproj/argo-workflows/blob/5a2fa4ee98b5c427ef7475df9c8c3146ed1afb54/ui/src/app/shared/components/object-editor.tsx#L25
2. https://github.com/argoproj/argo-workflows/blob/5a2fa4ee98b5c427ef7475df9c8c3146ed1afb54/ui/src/app/shared/components/object-editor.tsx#L29
3. https://github.com/argoproj/argo-workflows/blob/5a2fa4ee98b5c427ef7475df9c8c3146ed1afb54/ui/src/app/shared/components/object-editor.tsx#L29 (again)
4. https://github.com/argoproj/argo-workflows/blob/5a2fa4ee98b5c427ef7475df9c8c3146ed1afb54/ui/src/app/shared/components/object-editor.tsx#L38

It'd surprise me if adding a couple more calls would have a significant performance impact, but you're right that making performance worse might not be worth fixing this bug. I pushed 24c0206b9071503467034cb67716ef0b27158b0f to delete the `isEqual()` logic and go back to direct object comparisons.

To fix this without affecting performance, I think we need to refactor `<ObjectEditor>` to only work with strings, not parsed objects. Parsing will have to be lifted up to the details component (`<ClusterWorkflowTemplateDetails>`  in this case), which can then easily compare strings to see what's changed. That's not hard, but it does involve quite a lot of changes, so it's probably best done in a separate PR.

ui/src/app/cron-workflows/cron-workflow-details.tsx
MasonM: @Adrien-D That's a good idea.  I pushed 24c0206b9071503467034cb67716ef0b27158b0f to create a `useEditableResource()` hook. I didn't write tests since I think we'd need to add [@testing-library/react](https://github.com/testing-library/react-testing-library) as a dependency first.

agilgur5: Yea I'd like to add tests with `testing-library` eventually to the UI as it is sorely lacking them and some places have fairly complex logic

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
agilgur5: Yea that first one could be made into a callback function instead very simply. The last one has that comment around it regarding whitespace https://github.com/argoproj/argo-workflows/blob/5a2fa4ee98b5c427ef7475df9c8c3146ed1afb54/ui/src/app/shared/components/object-editor.tsx#L36-L37

I've optimized the rest of the UI code-splitting this component before in #12150 so this component is a bit of a hassle 😕 

> but you're right that making performance worse might not be worth fixing this bug.

yea the buttons working even though there has been no semantic change is pretty tiny and fairly harmless. I'd argue it's maybe even worth further simplifying if possible just to remove more complex code.

Otherwise if we wanted to keep a bunch of features while being latency sensitive, a separate thread / web worker might be the way to go to keep the UI thread light.

ui/src/app/shared/use-editable-resource.ts
```diff
@@ -0,0 +1,22 @@
+import {useState} from 'react';
+
+/**
+ * useEditableResource is a hook to manage the state of a resource that be edited and updated.
+ */
```
agilgur5, line 5: Since this is JSDoc and not GoDoc, we can simplify this to be less repetitive / verbose. There's also a typo "that be edited"

We should also mention the ref comparison same as we did the JSON comparison in the `isEqual` function before. And can one-line it
```suggestion
/** Hook that manages the state of a resource that can be edited. Uses ref comparison */
```

agilgur5: It also could make sense to rename this function/file `use-editable-object` to match `object-editor`.

Maybe even makes sense to have as part of the `ObjectEditor` component? 🤔

ui/src/app/shared/use-editable-resource.ts
```diff
@@ -0,0 +1,22 @@
...
+export function useEditableResource<T>(initial?: T): [T, boolean, React.Dispatch<T>, (value: T) => void] {
+    const [value, setValue] = useState<T>(initial);
+    const [initialValue, setInitialValue] = useState<T>(initial);
+
+    // TODO: Fix this so that it handles object comparison properly.
+    // Currently, this returns true if you make a change and immediately undo it, or if you save your changes.
+    // This could be solved using "const edited = JSON.stringify(value) !== JSON.stringify(initialValue)",
+    // but that has a performance penalty.
```
agilgur5, line 13: I don't think this is a TODO if we've intentionally decided _not_ to do it. We can just say notate we decided not to do it for performance purposes of a latency sensitive input

```suggestion
    // this is pure ref equality instead of deep object comparison etc as that has performance implications and inputs are latency sensitive
```

agilgur5 (commented):
Just some stylistic comments below

ui/src/app/shared/use-editable-resource.ts
MasonM: The other hooks that have documentation follow the `<x> is a React hook` format:
https://github.com/argoproj/argo-workflows/blob/5f70eee137344f98786f61168cb32cbe7e493ec0/ui/src/app/shared/use-resizable-width.ts#L27
https://github.com/argoproj/argo-workflows/blob/5f70eee137344f98786f61168cb32cbe7e493ec0/ui/src/app/shared/use-resizable-width.ts#L27

I updated the comment in cf9a3f383d51dc74fc6e8edd0f86b9f413e14d52 to match, and renamed it to `use-editable-object`.

I don't see how this could be part of `<ObjectEditor>` because the `edited` variable is needed by the details components.

ui/src/app/shared/use-editable-resource.ts
MasonM: IMO, this is worth fixing, but I went ahead and updated the comment in cf9a3f383d51dc74fc6e8edd0f86b9f413e14d52

ui/src/app/shared/use-editable-resource.ts
agilgur5: > The other hooks that have documentation follow the `<x> is a React hook` format:

Ah that works for consistency then I suppose. Some of the original ones written were almost certainly bringing GoDoc conventions into JSDoc (there's a handful of non-JS idioms in the UI here and there).

> I don't see how this could be part of `<ObjectEditor>` because the `edited` variable is needed by the details components.

I meant as another exported function inside the same file, not directly integrated into the component.
As in, there are some conventions in the ecosystem of `import {useComponent, Component}` that we could follow. Although the usage is a tad different here with regard to the tracked state and props

ui/src/app/shared/use-editable-resource.ts
agilgur5: Can continue that discussion in the [above thread](https://github.com/argoproj/argo-workflows/pull/13593#discussion_r1780324389) then, which is more specifically perf related

ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx
MasonM: Here's a POC of what I was suggesting: https://github.com/argoproj/argo-workflows/compare/main...MasonM:argo-workflows:poc-fix-13453

The `<ObjectEditorPlain>` component is a copy of `<ObjectEditor>` for POC purposes. The parsing and language handling code has been lifted up to the details component, where it's encapsulated in the `useEditableObject()` hook. 

This completely eliminates all the unnecessary calls to `stringify()` when there's a change, and also fixes the bugs with the "Update" button. It does call `parse()` once on a change, but that's an improvement from the `main` branch, where it's called twice.

If that looks good to you, I can clean that up and enter a PR.

ui/src/app/shared/use-editable-resource.ts
agilgur5: @MasonM any thoughts on the above? Otherwise this LGTM and I can merge

MasonM: I don't think it makes sense to couple this to `<ObjectEditor>`, because that will make it more difficult to move forward with the approach I posted above in https://github.com/argoproj/argo-workflows/pull/13593#discussion_r1781972661. That POC fully decouples `<ObjectEditorPlain>` from all serialization/deseralization and moves that logic into `useEditableComponent()`, which significantly improves performance and is a bit cleaner (IMO). Of course, that's a fairly significant change, and probably should come after this PR, but I think we'll want to do this eventually.

But if you feel strongly, I can go ahead and make that change.

agilgur5: I don't feel strongly.
I didn't read that change too closely yet so from a quick look I thought putting the imports/files together might actually help move in that direction, but I might've been mistaken. Since the performance change may make it even more likely that any consumer of the component would want to also use the hook (and it is currently used like that) and so combining the imports makes it more obvious how to use.

> `useEditableComponent()`

I guess your intent with this naming is that the hook could in the future be extended to other editable components as well? As in that it's not specific to `ObjectEditor` per se

agilgur5 (approved):
LGTM, thanks for iterating on this and ironing out the performance too!

ui/src/app/shared/use-editable-resource.ts
MasonM: > I guess your intent with this naming is that the hook could in the future be extended to other editable components as well?

Honestly, that was a typo. I meant `useEditableObject()`. It could theoretically be extended to other editable components, but I can't think of anything that would benefit.

agilgur5: Oh ok, then I think it would make sense to combine the imports / exports so that it's more obvious the hook is supposed to be used together with the component. 

Technically doesn't have to be in the same files, could have an import/export stub instead

Issue #12041, opened 2023-10-20
UI: Add `namespace` input box to `UserInfo` page for SSO RBAC NS delegation


agilgur5:
# Summary

Add an input box to the `UserInfo` page where a user can type in a namespace. It would default to the Argo installation namespace, as it currently does today.

## Use Cases

Right now, when using SSO RBAC with Namespace delegation, there's no way to see in the UI what SA will map to your user in a specific namespace.
You have to try a request in that namespace and then check the Server logs. Users who don't have access to the Server logs have no way to debug or even see what SA they got assigned in an NS.

Moreover, as the `UserInfo` page only shows you the default SA in the Argo installation namespace (see also #10968), that can be very confusing to a user who only has access to a specific namespace; they will effectively never be able to see their permissions for that namespace.

#### Server-side Details

This does require a new query parameter to be able to be passed to the backend in `/api/v1/userinfo`, i.e. `?namespace=my-namespace`.

In `server/auth/gatekeeper.go`, [this line](https://github.com/argoproj/argo-workflows/blob/7b2c4aa3e7290b186029d1b53eea1cb69166f7a0/server/auth/gatekeeper.go#L307) in the `rbacAuthorization` function gets the namespace via the [`getNamespace`](https://github.com/argoproj/argo-workflows/blob/7b2c4aa3e7290b186029d1b53eea1cb69166f7a0/server/auth/gatekeeper.go#L228) function. That function merely takes a request as an argument, so we just have to make sure the request is properly formatted as it expects.

I'm not entirely sure how it does the conversion from gRPC to Go struct, as this is one of the places where almost everything is auto-generated that keeps confusing me (and other contributors). IIRC the source of truth here is [`info.proto`](https://github.com/argoproj/argo-workflows/blob/7b2c4aa3e7290b186029d1b53eea1cb69166f7a0/pkg/apiclient/info/info.proto#L24), but I'm not entirely sure how to ensure that adding a `namespace` property there corresponds to `GetNamespace()` returning that property specifically.

#### UI Details

I think an input box for `Namespace` at the top of the page would suffice. Default to the Argo installation namespace as mentioned above. Should use the same input box and behavior as the `Namespace` input on the Workflows List page.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

gabrielfsousa:
i also would love to have the UserInfo and ns correct in UI

agilgur5:
[...]

PR #13596, opened 2024-09-13
feat(ui): Add namespace input to ClusterWorkflowTemplate submit. Fixes #10398

### LINKED ISSUE #10398: Allow setting namespace when submitting a ClusteWorkflowTemplate from the UI
# Summary

Currently, there's no way (or I haven't found one) to set the workflow namespace when submitting a `ClusterWorkflowTemplate` from the UI.

# Use Cases

A template may be submitted in different namespaces depending on which team/access/whatever is needed for that template. This is equivalent of doing 

```
argo submit --from clusterworkflowtemplate/my-template -n my-choosen-namespace
```
Current workaround is creating a WorkflowTemplate in each namespace, but that's really clunky.

---

MasonM:
Fixes #10398

### Motivation

This adds support for specifying the namespace when submitting a workflow from a ClusterWorkflowTemplate using an input field on the "Submit Workflow" panel.

### Modifications

Updates the `<SubmitWorkflowPanel>` to allow specifying the namespace via `<TextInput>`. Initially, I used the `<InputFilter>` component for this so it could take advantage of autocomplete, but then I realized that's not appropriate here because this isn't a filter and we don't want the "x" button to clear the field.

### Verification

Tested locally at http://localhost:8081/cluster-workflow-templates:

https://github.com/user-attachments/assets/2f2d6312-2806-4efa-ad79-10017bd4fd12


ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -10,8 +11,10 @@ import {TagsInput} from '../../shared/components/tags-input/tags-input';
 import {services} from '../../shared/services';
 import {Utils} from '../../shared/utils';
 
+type WorkflowTemplateType = 'ClusterWorkflowTemplate' | 'WorkflowTemplate';
+
 interface Props {
-    kind: string;
+    kind: WorkflowTemplateType;
```
agilgur5, line 17: it would probably make sense to just inline this if you're not using the type anywhere else

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -31,6 +34,7 @@ export function SubmitWorkflowPanel(props: Props) {
     const {navigation} = useContext(Context);
     const [entrypoint, setEntrypoint] = useState(workflowEntrypoint);
     const [parameters, setParameters] = useState<Parameter[]>([]);
+    const [namespace, setNamespace] = useState(props.namespace);
```
agilgur5, line 37: So this is not entirely correct to have `props.namespace` be the initial state. For a `WorkflowTemplate`, it will _always_ be `props.namespace`, and for a `ClusterWorkflowTemplate`, it will always be the `namespace` state.

This would be particularly relevant in cases where this panel is off-screen but still rendered; the next time it's on-screen, the prop would not be used.
That's normally not how this component is used, but that is why the correctness matters if it were to be used that way.

Also this should be ordered after the `useContext` instead of in between the two parameter state variables for better organization

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -85,6 +89,12 @@ export function SubmitWorkflowPanel(props: Props) {
                         }}
                     />
                 </div>
+                {props.kind === 'ClusterWorkflowTemplate' && (
+                    <div key='namespace' style={{marginBottom: 25}}>
+                        <label>Namespace</label>
+                        <TextInput value={namespace} onChange={setNamespace} />
+                    </div>
+                )}
```
agilgur5, line 97: This would make more sense at the top of the panel, before the entrypoint, rather than after. Since the first thing you think about is the namespace and then you can get into parameters etc

agilgur5 (commented):
Thanks for the the contribution!

I've left some comments in-line below.

Also, per the issue, did the submit panel just not work before this? If so, that issue should be labeled a bug and this a fix

agilgur5:
Also if you want to work on another namespace input in the UI, https://github.com/argoproj/argo-workflows/issues/12041 is highly requested 🙂

ui/src/app/workflows/components/submit-workflow-panel.tsx
MasonM: Sounds good. Done in 59dbea3690b1334d8f9b4cfc1e04ef60cc679141

ui/src/app/workflows/components/submit-workflow-panel.tsx
MasonM: You're right. I pushed 59dbea3690b1334d8f9b4cfc1e04ef60cc679141 to lift the state up to `<ClusterWorkflowTemplateDetails>`, which is simpler than handling it here

ui/src/app/workflows/components/submit-workflow-panel.tsx
MasonM: That makes sense. Done in 59dbea3690b1334d8f9b4cfc1e04ef60cc679141. I also updated the screen recording in the PR description

MasonM:
@agilgur5 Thanks for the review! 

> Also, per the issue, did the submit panel just not work before this? If so, that issue should be labeled a bug and this a fix

Due to the bug in https://github.com/argoproj/argo-workflows/pull/13593, it wasn't possible to click the "Submit" button, so I tested this by rebasing my branch with that fix. I confirmed it's possible to submit it with the fix, so this isn't a bug.

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -1,4 +1,5 @@
 import {Select} from 'argo-ui/src/components/select/select';
+import {TextInput} from '../../shared/components/text-input';
```
agilgur5, line 2: This should be ordered below with the other `shared/components` imports

MasonM: Okay, done in fccd30813004c5564e7dd82b67639df8e0f0ade6

timscottbell:
@MasonM is there anything blocking this feature from being delivered? Without it you can't use clusterworkflowtemplates easily in the UI across different namespaces, which defeats the purpose.

We are using argo-workflows and our product team wants us to only give the UI to customers, so this is a blocker for using clusterworkflowtemplates.

MasonM:
@timscottbell Nope, just waiting on a review. cc @isubasinghe @Joibel

Joibel (approved):
Thanks for doing this.

PR #13597, opened 2024-09-14
feat: use container name as archive log key

### LINKED ISSUE #13590: logs file name
# Summary

Use the container name within a step to generate the log file name. Currently, it is main.log for every step by default. We have run into the issue that the artifact log file is always called main.log thus it is getting overwritten everytime in azure artifacts and we lose all logs besides the logs from the last step.

## Use Cases

To have the ability to gather the logs from all steps easily

## Useful information

I went through the code to find a solution and managed to find argo-workflows/workflow/executor/executor.go:600 (template.getmaincontainernames function) and argo-workflows/pkg/apis/workflow/v1alpha1/workflow_types.go:3105 (savelogs function)
I am now currently using a containerset with a single container to use the name of the container for the log file. It's a workaround for a feature that, I feel, should not be too hard to implement.

---

chengjoey:
Fixes #13590

### Motivation

if tmpl type is `Container` and container.Name not empty,  use container name as artifact log file name

### Verification

e2e test

### DIFF
--- a/workflow/executor/executor.go
+++ b/workflow/executor/executor.go
@@ -49,6 +49,10 @@ import (
 const (
 	// This directory temporarily stores the tarballs of the artifacts before uploading
 	tempOutArtDir = "/tmp/argo/outputs/artifacts"
+	// The suffix of container log file
+	containerLogFileSuffix = ".log"
+	// The suffix of container log artifact name
+	containerLogArtifactSuffix = "-logs"
 )
 
 // WorkflowExecutor is program which runs as the init/wait container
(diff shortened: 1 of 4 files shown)


workflow/executor/executor.go
```diff
@@ -627,13 +627,19 @@ func (we *WorkflowExecutor) SaveLogs(ctx context.Context) []wfv1.Artifact {
 // saveContainerLogs saves a single container's log into a file
 func (we *WorkflowExecutor) saveContainerLogs(ctx context.Context, tempLogsDir, containerName string) (*wfv1.Artifact, error) {
 	fileName := containerName + ".log"
+	artName := containerName + "-logs"
+	if containerName == common.MainContainerName && we.Template.GetType() == wfv1.TemplateTypeContainer &&
+		len(we.Template.Container.Name) > 0 {
+		fileName = we.Template.Container.Name + ".log"
+		artName = we.Template.Container.Name + "-logs"
+	}
```
terrytangyuan, line 635: Refactor these into constants?

chengjoey: done, refactor `.log` and `-logs` into constants

shuangkun (approved):

shuangkun (commented):
https://github.com/argoproj/argo-workflows/blob/4742e9dd6ab2b797d32cb0953849fdcfe82ea325/ui/src/shared/services/workflows-service.ts#L265

Remove the restriction of this main container?  And fix this problem at the same time. #13785

chengjoey:
> https://github.com/argoproj/argo-workflows/blob/4742e9dd6ab2b797d32cb0953849fdcfe82ea325/ui/src/shared/services/workflows-service.ts#L265
> 
> Remove the restriction of this main container? And fix this problem at the same time. #13785

done

gaurang9991:
@chengjoey can you progress on this topic or this is abandoned ?

PR #13602, opened 2024-09-14
chore: update golangci-lint to v1.61.0

blkperl:
### Motivation

* A newer version of golangci-lint is requried to support golang 1.23

### Modifications

Changed `exportloopref` to `copyloopvar`

Fixed the following lint errors
* `printf: non-constant format string in call to github.com/sirupsen/logrus.Infof`
* `printf: non-constant format string in call to (*testing.common).Fatalf (govet)`
* `printf: non-constant format string in call to fmt.Sprintf (govet)`
* `printf: non-constant format string in call to google.golang.org/grpc/status.Errorf (govet)`
* `suite-broken-parallel: testify v1 does not support suite's parallel tests and subtests`
* `suite-subtest-run: use s.Run to run subtest (testifylint)`
* `The copy of the 'for' variable "limit" can be deleted (Go 1.22+) (copyloopvar)`

### Verification

Lint passes and existing test coverage


isubasinghe:
I believe this fixes memory usage issues, currently occasionally hit 84gb of memory usage before OOM.

Joibel:
> I believe this fixes memory usage issues, currently occasionally hit 84gb of memory usage before OOM.

Which part fixes that?

isubasinghe:
> Which part fixes that?

The newer version of golangci-lint

Joibel:
> > Which part fixes that?
> 
> The newer version of golangci-lint

Oh, during linting, not at runtime. More coffee needed. Ok - I remember you saying it did mad things for you.

test/e2e/cron_test.go
```diff
@@ -43,7 +43,6 @@ func (s *CronSuite) TearDownSuite() {
 
 func (s *CronSuite) TestBasic() {
 	s.Run("TestBasic", func() {
-		s.T().Parallel()
```
Joibel, line 46: These don't do anything, and as this was the only attempt in the code to parallelise tests, we could remove the `-parallel` flag on the test suite.

I'd like to find a way to reintroduce parallel testing in the future though, so let's not do that.

Joibel (approved):

PR #13604, opened 2024-09-15
docs(contrib): remove outdated section on using multiple terminals

MasonM:
### Motivation

I noticed the `CTRL` variable was deleted in 05e13182bb6be97ba6f91a24ea0876cc349d9ab6 and now does nothing, so I figured it's best to delete this section to avoid confusion. This has confused one other contributor before: https://github.com/argoproj/argo-workflows/issues/11879
>I don't think using [one terminal for the controller + server and another for the UI works...](https://argoproj.github.io/argo-workflows/running-locally/#using-multiple-terminals). I'm not seeing any reference to CTRL in the make file. When I try to run the command for starting the UI alone (make start CTRL=false) fails since it tries to boot up the controller again (it is already running).
### Modifications

Delete outdated section

### Verification

N/A

### DIFF
--- a/docs/running-locally.md
+++ b/docs/running-locally.md
@@ -245,15 +245,3 @@ go tool pprof http://localhost:6060/debug/pprof/profile   # 30-second CPU profil
 go tool pprof http://localhost:6060/debug/pprof/heap      # heap profile
 go tool pprof http://localhost:6060/debug/pprof/block     # goroutine blocking profile
 ```
-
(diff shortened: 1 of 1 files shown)


Joibel:
I don't think CTRL was ever implemented for kit.

For anyone who comes across this PR and feels like re-implementing CTRL, I'm not against having it. As we've lived without it for a while I'll merge this

Joibel (approved):

PR #13607, opened 2024-09-16
test: match on multi-node clusters as well

Joibel:
Running this test on a multi-node cluster may fail this test as originally written.

There are two conceived success modes for the test:
1. Workflow successfully schedules a second pod on the retry, and then the workflow fails as it is designed to fail. That is the `wfv1.WorkflowFailed` branch.
2. Workflow fails to schedule a second pod so it still `wfv1.WorkflowRunning` and this is because there is only one node in the cluster (e.g. running under k3d).

It may actually be that there are multiple nodes in the cluster but only one for test running, or there is an autoscaler available which may fulfil the node but not in 5 seconds.

So this changes the test match text to allow for more available nodes where the retry pod is still at least briefly unschedulable.


terrytangyuan (approved):
LGTM but should the title be "test: xxx"?

agilgur5:
Some of these are probably good to backport. But the title should state what the commit/PR does, not the problem

PR #13609, opened 2024-09-16
fix(sso): respect `X-Forwarded-*` headers for redirect URL. Fixes #13031

### LINKED ISSUE #13031: Incorrect redirect after Login when terminating TLS with reverse proxy
### Pre-requisites

**Description:**

When clicking an Argo Workflows link, if I'm not logged in, I'm redirected to the login page. After logging in, I'm redirected to the workflows page instead of the original page I was trying to access.

**Steps to reproduce:**

1. Click an Argo Workflows link to a specific page
3. If not logged in, click the "Login" button
4. After logging in, observe that you are redirected to the workflows page (not the page you intended to visit)

**Debugging information:**

* The issue is located in [sso.go#L351-L353](https://github.com/argoproj/argo-workflows/blob/v3.5.5/server/auth/sso/sso.go#L351-L353)
* The condition `strings.HasPrefix(cookie.Value, prefix)` always evaluates to false due to a mismatch between the HTTP and HTTPS schemes.
* The `secure` variable is set to false because `opts.TLSConfig` is nil, indicating that Argo Server is not terminating TLS. Instead, a reverse proxy is used to terminate TLS. So https://github.com/argoproj/argo-workflows/blob/v3.5.5/server/auth/sso/sso.go#L345-L348 ends up being `http` when infact we are using `https`.

**Expected behavior:**

After logging in, the user should be redirected to the original project they were trying to access, not the last viewed project.

**Version:** Argo Workflows v3.5.5
### Version

v3.5.5

```YAML
n/a
```

```text
```

```text
```

MasonM:
Fixes #13031

### Motivation
When running Argo in plain text mode behind a TLS termination proxy with SSO enabled, you will always be redirected to the workflows page after logging in, regardless of which page you came from. This is happening because `server/auth/sso/sso.go` is using the protocol and host of the request to determine the redirect URL, which can be different than the original request when a proxy is involved.

### Modifications

The fix involves reading from the [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) and [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) headers when present. This is almost the same as the fix that `oauth2-proxy` made when they faced a similar issue in https://github.com/oauth2-proxy/oauth2-proxy/pull/957. The only major difference is that `oauth2-proxy` only reads those headers when the `--reverse-proxy` option is set, but I didn't add a corresponding option here. The only reason I can think of for doing so is to preserve backwards-compatibility, since it's theoretically possible this could break someone's setup. I can't think of any real-world situations where that can happen, but let me know if I should put this behind a option and I'm happy to do so.

### Verification

Testing instructions:
1. Run `make start UI_SECURE=true PROFILE=sso`
2. Visit https://localhost:8080/cron-workflows
3. Click the "Login" button
4. Click "Log in with Example"
5. Click "Grant Access"
6. Confirm you're returned to https://localhost:8080/cron-workflows

https://github.com/user-attachments/assets/854d54cf-91e6-41e9-8e31-3bae20fb2722


ui/src/app/webpack.config.js
```diff
@@ -107,15 +109,14 @@ const config = {
...
-        proxy: {
-            '/api/v1': proxyConf,
-            '/artifact-files': proxyConf,
-            '/artifacts': proxyConf,
-            '/input-artifacts': proxyConf,
-            '/artifacts-by-uid': proxyConf,
-            '/input-artifacts-by-uid': proxyConf,
-            '/oauth2': proxyConf
-        }
+        proxy: [
```
MasonM, line 112: FYI: The old object syntax is deprecated and was removed in https://github.com/webpack/webpack-dev-server/pull/4694

jmeridth (approved):
We are having the exact same issue.  Thank you for coding this up.

MasonM:
FYI: I split off the dev environment improvements to https://github.com/argoproj/argo-workflows/pull/13674

server/auth/sso/sso.go
```diff
@@ -376,14 +372,26 @@ func (s *sso) getRedirectUrl(r *http.Request) string {
...
 	}
+	return fmt.Sprintf("%soauth2/callback", s.getRedirectUrlPrefix(r))
+}
 
+func (s *sso) getRedirectUrlPrefix(r *http.Request) string {
 	proto := "http"
 
-	if r.URL.Scheme != "" {
+	// Get protocol from the X-Forwarded-Proto header if set to support running Argo behind a TLS termination proxy
+	if r.Header.Get("X-Forwarded-Proto") != "" {
```
agilgur5, line 382: my main concern with changing configuration based on proxy is that it becomes susceptible to MITMs or similar -- anyone can send such a header.

Whereas `redirectURL` is configured by an admin

agilgur5:
> Let me know if I should split this off to a separate PR.

> FYI: I split off the dev environment improvements to #13674

Thanks Mason. I requested this [over Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1727541871558059?thread_ts=1727525422.773999&cid=C0510EUH90V) as it is would make it easier to review and backport, and is a general best practice for smaller PRs.

server/auth/sso/sso.go
agilgur5: Hmm this was previously using the request URL (`r.URL.Scheme`), which I believe could be similarly spoofed via proxy 🤔

MasonM: Yes, but that was already an inherent problem with this setup, because if you're running a TLS termination in front of Argo, you have to trust everything in between you and the proxy (including the proxy). The only way of protecting against a malicious proxy is through E2E encryption. 

Also, if someone does MITM the connection and has the ability to inject arbitrary headers, then they almost certainly have the ability to completely rewrite the response body too, in which case there's no need for them to redirect the user to a malicious page. They can just directly serve up whatever content they want,.

agilgur5: With the specific setup of the issue, yes, but this potentially extends it further too. Even with E2EE, a maliciously crafted request could potentially send someone to an uncontrolled/malicious URL, similar to an open redirect.

MasonM: > Even with E2EE, a maliciously crafted request could potentially send someone to an uncontrolled/malicious URL, similar to an open redirect.

I can't think of any attack vector where that can happen that wasn't already a viable attack vector. If E2E encryption is used, the only means for an attacker to inject arbitrary headers is to either compromise the Argo server or the end user's device, and in both cases the attacker isn't going to bother with open redirect vulnerabilities because they can just rewrite the page to be whatever they want.

Regardless, I entered

agilgur5: > If E2E encryption is used, the only means for an attacker to inject arbitrary headers is to either compromise the Argo server or the end user's device

The simplest and one of the most common variants is to, for instance, write an informative blog post that sends requests to the API, and one of those requests is maliciously crafted. The user then copy+pastes it unknowingly.

In CVSS, this kind of exploit requires "User Interaction", but does not require compromising the user's device

MasonM: Sorry, I'm still not seeing how an attacker could inject arbitrary headers. I know an attacker can control the redirect URL via something like https://localhost:8080/oauth2/redirect?redirect=https://evil.com, but that's useless if they can't control the request headers too. Even if the attacker can execute Javascript and send AJAX requests, the CORS policy won't allow it. Try running this in a DevTools console at https://example.com :
```javascript
fetch('https://localhost:8080/oauth2/redirect?redirect=https://evil.com', {headers: {"X-Forwarded-Host": "evil.com"}})
``` 

You should see an error like this:
```
Access to fetch at 'https://localhost:8080/oauth2/redirect?redirect=https://evil.com' from origin 'https://example.com' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: Redirect is not allowed for a preflight request.
```

MasonM:
Superseded by https://github.com/argoproj/argo-workflows/pull/13747

PR #13615, opened 2024-09-18
feat: Set `steps.*.ip` to a json list if the node is an aggregate

### LINKED ISSUE #13614: Set `steps.*.ip` for `withSequence`
# Summary

Presently, if a node is marked as `daemon: true`, its IP is available as `steps.*.ip` or `tasks.*.ip`. Unlike outputs, this is not set if the step is an aggregate like  `withSequence`. If this were consistent, this field would be set to a JSON list of IPs.

## Use Cases

The motivating use case is where a pool of independent simulation workers are spawned at the start of the workflow, and then issued tasks by other steps in the workflow. Presently, this is possible, but requires the use of the k8s api to find the pod IP's.

---

meln5674:
Fixes https://github.com/argoproj/argo-workflows/issues/13614

### Motivation

See https://github.com/argoproj/argo-workflows/issues/13614

### Modifications

Aggregate nodes of demonized steps/tasks now have the `ip` field set to a JSON list of the pod IPs.

Docs have been updated to note this fact.

### Verification

An e2e test has been added in which a set of nginx damon pods are created and then curl'ed, one each for withSequence/Items/Param. I noticed there was no test for the original single pod ip functionality, so I added a test for it as well to confirm this did not break it.

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -139,8 +139,8 @@ sprig.trim(inputs.parameters['my-string-param'])
 | Variable | Description|
 |----------|------------|
 | `steps.name` | Name of the step |
-| `steps.<STEPNAME>.id` | unique id of container step |
-| `steps.<STEPNAME>.ip` | IP address of a previous daemon container step |
+| `steps.<STEPNAME>.id` | unique id of container step. |
+| `steps.<STEPNAME>.ip` | IP address of a previous daemon container step. When the previous step uses `withItems` or `withParams`, this contains a JSON array of the IP's of each invocation |
 | `steps.<STEPNAME>.status` | Phase status of any previous step |
 | `steps.<STEPNAME>.exitCode` | Exit code of any previous script or container step |
 | `steps.<STEPNAME>.startedAt` | Time-stamp when the step started |
--- a/docs/walk-through/daemon-containers.md
+++ b/docs/walk-through/daemon-containers.md
@@ -76,3 +76,5 @@ spec:
 ```
 
 Step templates use the `steps` prefix to refer to another step: for example `{{steps.influx.ip}}`. In DAG templates, the `tasks` prefix is used instead: for example `{{tasks.influx.ip}}`.
+
+If the step or task uses `withSequence`, `withItems`, or `withParam`, then this will instead be set to a JSON list of IP addresses.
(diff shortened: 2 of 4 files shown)


meln5674:
Unit tests seem to have failed as a result from a 500 from dockerhub, could someone please restart them? Might be worthwhile to run https://github.com/distribution/distribution in-memory using https://pkg.go.dev/net/http/httptest  to make this test hermetic.

docs/variables.md
```diff
@@ -139,8 +139,8 @@ sprig.trim(inputs.parameters['my-string-param'])
 | Variable | Description|
 |----------|------------|
 | `steps.name` | Name of the step |
-| `steps.<STEPNAME>.id` | unique id of container step |
-| `steps.<STEPNAME>.ip` | IP address of a previous daemon container step |
+| `steps.<STEPNAME>.id` | unique id of container step. |
+| `steps.<STEPNAME>.ip` | IP address of a previous daemon container step. When the previous step uses `withItems` or `withParams`, this contains a JSON array of the IP's of each invocation |
```
tico24, line 143: Doesn't `tasks.<TASKNAME>.ip` further down need the same edit? All four of your e2e tests are DAGs. `withSequence` is missing here too, though you do list it on the other page.

tico24 (commented):
Docs review only, the Go and the e2e test are for @argoproj/argo-workflows-approvers. Couple of bits below.

docs/walk-through/daemon-containers.md
```diff
@@ -76,3 +76,5 @@ spec:
 ```
 
 Step templates use the `steps` prefix to refer to another step: for example `{{steps.influx.ip}}`. In DAG templates, the `tasks` prefix is used instead: for example `{{tasks.influx.ip}}`.
+
+If the step or task uses `withSequence`, `withItems`, or `withParam`, then this will instead be set to a JSON list of IP addresses.
```
tico24, line 80: "will instead be set to" is future tense and passive. Something like "this is instead a JSON list of IP addresses"? Per the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#patterns-to-avoid).

tico24, line 80: Could you show what you'd do with the list? The `withParam: '{{tasks.nginxen.ip}}'` from your e2e test would do it. On its own this sentence leaves you guessing, and the [docs guide](https://argo-workflows.readthedocs.io/en/latest/doc-changes/) wants working examples.

meln5674:
I'm going to be blunt, I entirely forgot I made this, and my team started a full migration away of argo workflows after the third month, and long since completed in the intervening years, of radio silence on this.

PR #13413, opened 2024-07-30
chore(deps)!: bump k8s dependencies to 1.30

blkperl:
### Motivation

* K8s 1.26 is no longer receiving security patches

### Modifications

* updated `go.mod` to k8s 1.30
* updated `dist/kubernetes.swagger.json` version to 1.30
* updated `workflow/executor/resource.go` to non deprecated functions to appease the linter
* updated from controller-gen@v0.14.0 to controller-gen@v0.15.0
* updated various files to the new ptr library to appease the linter
* changed the metrics test contains assertion from status_code=200 to status_code=400 to pass the test
* changed `getEvents` test helper to truncate annotations from events so that `TestEventNodeEvents` passes

### Verification

Existing test coverage

PR #13623, opened 2024-09-19
build: ensure Makefile driven dependencies are up to date

Joibel:
# Motivation

As a developer when the Makefile changes the version of a program that it uses or another dependency I would like to use the new version of that program or dependency

Created in response to https://cloud-native.slack.com/archives/C0510EUH90V/p1726666701563199

# Modifications

For all make targets that specify their version in the Makefile specify the Makefile as a dependency so they will get refetched if the Makefile changes. This is crude as it will refetch all dependencies regardless of whether they have changed or not, so touching the Makefile will cause a slower rebuild. Correctness trumps speed here in my opinion.

Add `-y` to the `apt get` call installing clang-format so that it becomes non-interactive. It's annoying to have to type `y` when this runs anyway.

Use the special Makefile variable `$@` instead of repeating the target name explicitly in the rules I was touching where it was relevant - following the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle.

Most annoyingly `recurl.sh` doesn't do any work if the target file exists, so remove the file before fetching in the case where that is used so we get an updated version. This foxed me for a while as this was the file causing the original issue linked above in slack.

# Verification

Locally `make codegen` successfully on a dirty checkout which had run codegen prior to https://github.com/argoproj/argo-workflows/pull/13413. 

# Notes to approvers

This is pushed to argoproj so you can make direct edits if you feel that's appropriate for collective ownership of fixing this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ ui/dist/app/index.html: $(shell find ui/src -type f && find ui -maxdepth 1 -type
 	# `yarn build` is slow, so we guard it with a up-to-date check.
 	JOBS=max yarn --cwd ui build
 
-$(GOPATH)/bin/staticfiles:
+$(GOPATH)/bin/staticfiles: Makefile
 # update this in Nix when updating it here
 ifneq ($(USE_NIX), true)
 	go install bou.ke/staticfiles@dd04075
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):
I changed the title. Otherwise LGTM

Joibel:
> I changed the title. Otherwise LGTM

Thanks, I forget about these other titles. I'll try to remember.

I think the `test-functional` failures are genuine. I'm investigating.

agilgur5:
> > I changed the title. Otherwise LGTM
> 
> Thanks, I forget about these other titles. I'll try to remember.

As I mentioned in https://github.com/argoproj/argo-workflows/pull/13607#issuecomment-2354454694, the `fix` prefix could be fine if we plan to backport the PR, since the `cherry-pick.sh` script [filters on prefix](https://github.com/argoproj/argo-workflows/blob/ce7f9bfb9b45f009b3e85fabe5e6410de23c7c5f/hack/cherry-pick.sh#L6)

PR #13624, opened 2024-09-19
test: fix HttpHandler test

Joibel:
`http://dummy.restapiexample.com/api/v1/employees` is called as a test endpoint in this test and must respond 2xx for this test to pass. It is offline, so switch to `https://github.com`.

If github is offline we're probably not running CI anyway.

### DIFF
--- a/test/e2e/hooks_test.go
+++ b/test/e2e/hooks_test.go
@@ -851,7 +851,7 @@ spec:
           template: http
     - name: http
       http:
-        url: http://dummy.restapiexample.com/api/v1/employees
+        url: https://github.com/
 `).When().
 		SubmitWorkflow().
 		WaitForWorkflow(fixtures.ToBeCompleted).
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

test/e2e/hooks_test.go
```diff
@@ -851,7 +851,7 @@ spec:
           template: http
     - name: http
       http:
-        url: http://dummy.restapiexample.com/api/v1/employees
+        url: https://github.com/
```
terrytangyuan, line 854: Maybe we should call GitHub API endpoint?

terrytangyuan: Non-blocking comment

Joibel: I felt that was likely to get rate limited, and we're not calling it with a token so it may randomly fail for that.

I considered just calling google instead.

terrytangyuan: Sounds good

Joibel: I've remembered we've got our own http endpoint to test against already, `httpbin`, so switched to that.

PR #13635, opened 2024-09-20
feat(cli): add version header + warning on client-server mismatch. Fixes #9212

### LINKED ISSUE #9212: CLI should warn when CLI version is outdated

MasonM:
Fixes #9212

### Motivation

This adds a warning message to the CLI when it detects a mismatch between the client and server versions. 

### Modifications

There was another PR (https://github.com/argoproj/argo-workflows/pull/11909) for this issue that implemented it by making a blocking API call to `/api/v1/version` in a `PersistentPreRun` hook. This PR takes a different approach: have the server send the version in a new header called `argo-version`, which the client will extract. There's several advantages to this approach:
1. Negligible performance impact, since no additional requests are needed.
2. Warning is only shown when the command would normally send an API request
3. Can be useful for bug triaging, since the header can be seen in `curl` output:
    ```
    $ curl -v http://localhost:2746/api/v1/info
    * Host localhost:2746 was resolved.
    * IPv6: ::1
    * IPv4: 127.0.0.1
    *   Trying [::1]:2746...
    * connect to ::1 port 2746 from ::1 port 62973 failed: Connection refused
    *   Trying 127.0.0.1:2746...
    * Connected to localhost (127.0.0.1) port 2746
    > GET /api/v1/info HTTP/1.1
    > Host: localhost:2746
    > User-Agent: curl/8.8.0
    > Accept: */*
    >
    * Request completely sent off
    < HTTP/1.1 200 OK
    < Content-Type: application/json
    < Grpc-Metadata-Argo-Version: latest+303bcce.dirty
    < Grpc-Metadata-Content-Type: application/grpc
    < X-Ratelimit-Limit: 1000
    < X-Ratelimit-Remaining: 999
    < X-Ratelimit-Reset: Fri, 20 Sep 2024 17:44:03 UTC
    < Date: Fri, 20 Sep 2024 17:44:02 GMT
    < Content-Length: 1392
    <
    * Connection #0 to host localhost left intact
    {"managedNamespace":"argo","links":[{"name":"Workflow Link","scope":"workflow","url":"http://logging-facility?namespace=${metadata.namespace}\u0026workflowName=${metadata.name}\u0026startedAt=${status.startedAt}\u0026finishedAt=${status.finishedAt}"},{"name":"Pod Link",
    "scope":"pod","url":"http://logging-facility?namespace=${metadata.namespace}\u0026podName=${metadata.name}\u0026startedAt=${status.startedAt}\u0026finishedAt=${status.finishedAt}"},{"name":"Pod Logs Link","scope":"pod-logs","url":"http://logging-facility?namespace=${met
    adata.namespace}\u0026podName=${metadata.name}\u0026startedAt=${status.startedAt}\u0026finishedAt=${status.finishedAt}"},{"name":"Event Source Logs Link","scope":"event-source-logs","url":"http://logging-facility?namespace=${metadata.namespace}\u0026podName=${metadata.n
    ame}\u0026startedAt=${status.startedAt}\u0026finishedAt=${status.finishedAt}"},{"name":"Sensor Logs Link","scope":"sensor-logs","url":"http://logging-facility?namespace=${metadata.namespace}\u0026podName=${metadata.name}\u0026startedAt=${status.startedAt}\u0026finishedA
    t=${status.finishedAt}"},{"name":"Completed Workflows","scope":"workflow-list","url":"http://workflows?label=workflows.argoproj.io/completed=true"}],"modals":{"feedback":true,"firstTimeUser":true,"newVersion":true},"columns":[{"name":"Workflow Completed","type":"label",
    "key":"workflows.argoproj.io/completed"}]}%
    ```

Exposing the version information has security implications, since it could be used by attackers to identify vulnerable Argo servers. To mitigate that, the header is not sent on 401 errors. Of course, if a user is exposing their Argo server to the internet without authentication, then an attacker could see this header, but then they've got bigger problems (and an attacker could just call `/api/v1/version`).

This is implemented on the client and server side using [grpc-go interceptors](https://github.com/grpc/grpc-go/blob/master/examples/features/interceptor/README.md). On the server side, there's interceptors to set the version header in the response. On the client side, there's an interceptor to check the response for the header and stash it in a global variable (which is obviously not ideal, but I couldn't think of a better way to handle that).

### Verification

Testing process:
1. Manually changed the version to `v9.99` here: https://github.com/argoproj/argo-workflows/blob/ce7f9bfb9b45f009b3e85fabe5e6410de23c7c5f/Makefile#L95
2. Ran `make cli && cp dist/argo argo2`
3. Ran `git checkout Makefile && make start API=true`
5. Ran `ARGO_SECURE=false ARGO_TOKEN="Bearer $(kubectl get secret argo-server.service-account-token -o=jsonpath='{.data.token}' | base64 --decode)" ARGO_SERVER=localhost:2746 ./dist/argo2 list`

Output:
```
No workflows found
WARN[2024-09-20T18:03:26.116Z] CLI version (v9.99+303bcce.dirty) does not match server version (latest+303bcce.dirty). This can lead to unexpected behavior.
```


MasonM:
The one test failure is pretty clearly unrelated, and seems to be due to resource issues:
```
 ErrImageNeverPull: Container image "quay.io/argoproj/argocli:latest" is not present with pull policy of Never

    artifacts_test.go:218: timeout after 1m30s waiting for condition
```

@agilgur5 do you know how to selectively re-run that failed test?

util/cmd/cmd_test.go
```diff
@@ -3,6 +3,12 @@ package cmd
 import (
 	"reflect"
 	"testing"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
```
agilgur5, line 8:
```suggestion
	logtest "github.com/sirupsen/logrus/hooks/test"
```

we don't have another `test` here, but I feel like being specific is  useful here

util/grpc/interceptor.go
agilgur5, line 1: gRPC is slated to be removed per #13542

agilgur5 (commented):
I'm not necessarily opposed to this given that I mentioned using a header in https://github.com/argoproj/argo-workflows/pull/11909#issuecomment-1749103813, but modifying gRPC a decent bit here seems suboptimal given #13542

agilgur5:
> @agilgur5 do you know how to selectively re-run that failed test?

In CI, you can't. You can re-run an individual CI job with a `/retest` comment after #13000 et al, but you need to be an Argoproj Member to do so. Other Members can add that comment for you and Approvers can manually re-run CI jobs as well.

Also relevant, would you be interested in becoming [a Member](https://github.com/argoproj/argoproj/blob/224c2e6/community/membership.md)? You have enough contributions to qualify now and enough would merge by the next quarterly Membership Meeting (November) as well

util/cmd/cmd_test.go
MasonM: Okay, done in 5cff94056bd25bb9d961342de614ac8abaccd0aa

util/grpc/interceptor.go
MasonM: Thanks for letting me know. These interceptors will need to be ported to whatever framework is chosen for https://github.com/argoproj/argo-workflows/issues/13542. Both https://github.com/gin-gonic/gin and https://github.com/gorilla/mux have middleware support and should work, but I haven't verified.

I guess this PR should either closed or left open and re-evaluated after https://github.com/argoproj/argo-workflows/issues/13542 is done. I'm fine with either one.

MasonM:
@agilgur5 Thanks!

Yes, I'm interested in becoming a member, but I need to get sign off from my employer first. Are you willing to sponsor me?

agilgur5:
Yes I mentioned it, so ofc I can be one sponsor 🙂
Alan up-voted my comment and is potentially another sponsor, but we'll definitely be able to get you a second or more too.

You mentioned you were contributing in your personal capacity, so you don't have to list your employer as your affiliation, but it is usually good to cross-check with them anyway. Pretty sure Adobe would be happy to have more contributors affiliated too 😉

util/grpc/interceptor.go
agilgur5: Oh, I think I missed that this file was mostly middleware logic and not _that_ gRPC specific. I think you're right in that case, most of this file would have to be ported. My mistake, sorry, we rarely touch the gRPC logic, so I probably glazed over that!

Also if you're interested in helping with that migration, that would be welcome! You might be more familiar with the Go API ecosystem than me 😅

util/grpc/interceptor.go
```diff
@@ -50,6 +58,38 @@ var (
 	}
 )
 
+// SetVersionHeaderUnaryServerInterceptor returns a new unary server interceptor that sets the argo-version header
+func SetVersionHeaderUnaryServerInterceptor(version wfv1.Version) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		err := grpc.SetHeader(ctx, metadata.Pairs(ArgoVersionHeader, version.Version))
```
agilgur5, line 64: another major difference with a header is its security properties; the version from `/info` is only exposed to authenticated clients. we'd have to set this only if the request is authenticated

MasonM: I mentioned this in the PR description, but the header won't be added for 401 errors, because the `as.gatekeeper.UnaryServerInterceptor()` is added before this one:
https://github.com/MasonM/argo-workflows/blob/5cff94056bd25bb9d961342de614ac8abaccd0aa/server/apiserver/argoserver.go#L306

Since `as.gatekeeper.UnaryServerInterceptor()` returns an error on an auth failure, that should prevent this interceptor from running, unless I'm misunderstanding how this works. I tested this locally:
```
$ curl -v http://localhost:2746/api/v1/info
* Host localhost:2746 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:2746...
* connect to ::1 port 2746 from ::1 port 60801 failed: Connection refused
*   Trying 127.0.0.1:2746...
* Connected to localhost (127.0.0.1) port 2746
> GET /api/v1/info HTTP/1.1
> Host: localhost:2746
> User-Agent: curl/8.8.0
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< Content-Type: application/json
< Trailer: Grpc-Trailer-Content-Type
< X-Ratelimit-Limit: 1000
< X-Ratelimit-Remaining: 999
< X-Ratelimit-Reset: Sat, 21 Sep 2024 23:29:45 UTC
< Date: Sat, 21 Sep 2024 23:29:44 GMT
< Transfer-Encoding: chunked
<
* Connection #0 to host localhost left intact
{"code":16,"message":"token not valid. see https://argo-workflows.readthedocs.io/en/latest/faq/"}
```

agilgur5: Oh nice, thanks for confirming and testing that!

And sorry you are catching all my mistakes in review on this PR 😅 I was having a smoldering migraine all last week, so my brain is probably still adjusting (that's also why I didn't approve much last week)

agilgur5: This is actually an excellently detailed PR description that answers all my questions 👏  that's also coming from the person who tends to write the most detailed descriptions in this repo 🙂

agilgur5: I'm wondering if might be worth adding an E2E test to ensure there's no security regressions with this in the future -- what do you think?

util/cmd/cmd.go
```diff
@@ -45,6 +45,13 @@ func PrintVersion(cliName string, version wfv1.Version, short bool) {
 	fmt.Printf("  Platform: %s\n", version.Platform)
 }
 
+// PrintVersionMismatchWarning detects if there's a mismatch between the client and server versions and prints a warning if so
+func PrintVersionMismatchWarning(clientVersion wfv1.Version, serverVersion string) {
+	if serverVersion != "" && clientVersion.GitTag != "" && serverVersion != clientVersion.Version {
```
agilgur5, line 50: this is identical to the previous PR, although I wonder if we want to have some sort of behavior for unset versions. 🤔
that would normally only happen with development versions and a contributor would be aware of this, but I'm thinking in the case of a untagged internal build or something, a warning could be useful. although that's pretty niche of a case, so probably not worthwhile

agilgur5 (approved):
Awesome work with all the `grpc-go` interceptor nuances here, I learned some things from your code 🙂 

I just have two non-blocking comments that we may or may not want to act further on

util/grpc/interceptor.go
MasonM: It's good you asked for that, because in the process of writing that test I realized that the header actually is being returned on 401s when `--auth-mode=client` is used. Earlier, I had only tested with `--auth-mode=sso`, which causes `as.gatekeeper.UnaryServerInterceptor()` to immediately error out on auth failures: https://github.com/argoproj/argo-workflows/blob/6cac182ff247ff37a13b8efc6cb4ff11fcdd20c3/server/auth/gatekeeper.go#L207-L210

But with `--auth-mode=client`. it doesn't actually verify the token is valid, just that it's well-formed. It's only later in the request lifecycle (when the token is used) that you can get the 401.

I pushed f590dcb0b754e95f57c698f3c3d7dcf7d43e68ab to change the interceptors to be post-processors, and only set the header if there isn't an error. Of course, I also updated the E2E tests.

util/cmd/cmd.go
MasonM: That could be useful, but I'm worried printing a warning every time an untagged build of the Argo CLI is run will become more annoying than helpful. We could add a CLI flag to suppress the warning, but that seems overkill.

util/grpc/interceptor.go
agilgur5: Yes that's correct regarding `--auth-mode=client`. And yes I was trying to make sure it was a post-processor with my limited knowledge of `grpc-go`. I didn't realize there was a timing difference to the code though (just commented on the general area); fortunately you figured it out eventually so at least now that's properly covered 😅

util/grpc/interceptor.go
```diff
@@ -50,6 +58,46 @@ var (
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		m, err := handler(ctx, req)
+		if err == nil {
+			// Don't set header if there was an error because attackers could use it to find vulnerable Argo servers
+			err := grpc.SetHeader(ctx, metadata.Pairs(ArgoVersionHeader, version.Version))
+			if err != nil {
+				log.Warnf("Failed to set header '%s': %s", ArgoVersionHeader, err)
+			}
+		}
+		return m, err
```
agilgur5, line 72: will this return the original error, or the second error from `SetHeader`? we should return the original one, as the warning is sufficient for the header

in most languages, I would think it's the second right now, but you are using `:=` instead of `=` in the second one which would also normally have a lint error in most languages for shadowed variables with the same name.

MasonM: It'll return the original header, because Go is [lexically scoped](https://go.dev/ref/spec#Declarations_and_scope), and the `if err == nil {` block establishes a new scope. I pushed eb08e7665c092a11cbb24f49953d6fdf02d3b1d8 to add tests cases for this: https://github.com/argoproj/argo-workflows/blob/b958ca759f930eefedb46c584fdaad752a0b4dc6/util/grpc/interceptor_test.go#L126-L135

agilgur5: Right that's what I thought; normally linters disallow that as variable "shadowing". For clarity and better readability, I would suggest rewriting to `origErr` and `err` like in https://github.com/argoproj/argo-workflows/pull/11674

agilgur5: Since the equivalent code is used in the function below this, this could be split into an encapsulated helper.

MasonM: Okay, I pushed a8a283fd3ecbab8c00fa3f421e41419966212d74 to rename `err` to `origErr`.

Also, this can't be refactored into a helper because `grpc.SetHeader()` doesn't work for stream servers, because there's no context available to the inerceptor.

agilgur5: Ah I thought there was a diff I noticed earlier that I missed this time around, thanks for pointing that out again!

agilgur5 (approved):
Thanks for iterating on this and being careful of the possible security ramifications!

PR #13640, opened 2024-09-21
test: properly detect + optimize example suite

agilgur5:
### Motivation

Used this script directly while testing #13429 and it was problematic when there were errors; this isolates the errors and optimizes the script

Also the example test suite did not run in CI per https://github.com/argoproj/argo-workflows/pull/13429#issuecomment-2267597431; add this to the diff detection

#### script changes
- create the workflow and retrieve its name in one request
- delete workflow after running it by name
  - the label delete on every iteration caused some wacky race conditions locally when testing, i.e. sometimes with broken workflows

- also add more logging
  - already had "Running ", now has "Waiting for completion ", and "Deleting "

#### ci changes

- detect diff in `examples/**` dir and with the script to run E2Es
- `useApi` -> `use-api` for stylistic consistency, follow-up to #13569

### Verification

`test-examples` suite passes

### Future Work

1. [ ] Per https://github.com/argoproj/argo-workflows/pull/13429#issuecomment-2267597431, the diff detection could be further optimized to only run the example suite if only examples changed. Right now, the entire E2E suite will run instead. That is safer/more correct than not running any tests, but suboptimal.
    - This would require a decent sized refactor of the CI logic, and it is not often that examples _alone_ change, so punting that for now


Joibel (approved):
LGTM, thank you

Issue #13637, opened 2024-09-21
3.6.0-rc1: Controller panics due to metrics round-tripper data race


meln5674:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

[These](https://github.com/argoproj/argo-workflows/blob/main/workflow/metrics/metrics_k8s_request.go) [two](https://github.com/argoproj/argo-workflows/blob/main/util/logs/log-k8s-requests.go) roundtripper wrappers, instantiated [here](https://github.com/argoproj/argo-workflows/blob/main/cmd/workflow-controller/main.go#L95-L96) introduce a data race.

To demonstrate, apply this modification to the Makefile

```diff
--- a/Makefile
+++ b/Makefile
@@ -223,9 +223,9 @@ controller: dist/workflow-controller
 dist/workflow-controller: $(CONTROLLER_PKG_FILES) go.sum
 ifeq ($(shell uname -s),Darwin)
        # if local, then build fast: use CGO and dynamic-linking
-       go build -gcflags '${GCFLAGS}' -v -ldflags '${LDFLAGS}' -o $@ ./cmd/workflow-controller
+       go build -race -gcflags '${GCFLAGS}' -v -ldflags '${LDFLAGS}' -o $@ ./cmd/workflow-controller
 else
-       CGO_ENABLED=0 go build -gcflags '${GCFLAGS}' -v -ldflags '${LDFLAGS} -extldflags -static' -o $@ ./cmd/workflow-controller
+       go build -race -gcflags '${GCFLAGS}' -v -ldflags '${LDFLAGS} -extldflags -static' -o $@ ./cmd/workflow-controller
 endif
```

Then execute `make test-functional E2E_ENV_FACTOR=2` (make sure to show all logs from `make start`)

Stack trace of the data race follow below.

[...]

PR #13641, opened 2024-09-21
fix: Prevent data race from global metrics round-tripper

meln5674:
Fixes https://github.com/argoproj/argo-workflows/issues/13637

### Motivation

[This line](https://github.com/argoproj/argo-workflows/blob/ce7f9bfb9b45f009b3e85fabe5e6410de23c7c5f/workflow/metrics/metrics_k8s_request.go#L75) introduces a data race by globally storing a round-tripper used by the kubernetes client. If a new request starts before the first one completes, and the first request attempts to use the same round-tripper it originally created before the second finishes upgrading its connection, it will result in a panic due to a nil `net.Conn` in the underlying SPDY implementation.

While it is sensible to store a global handle to the metrics that this round-tripper records, storing the round-tripper itself is not.

### Modifications

This patch retains the "context" of the metrics (i.e. the actual `ctx` value, plus the handle to the metrics themselves) as global, but scopes the round-tripper to each connection by refactoring the context to its own type and global variable, and using an embedded pointer to it in the round-tripper implementation.

### Verification

E2E functional tests were run locally. Additionally, the PR tests identified in the issue as failing as a result of this race now pass, and the controller logs from running without the race detector were visually inspected to confirm that no panics occurred after multiple re-runs of the originally failing test.

### DIFF
--- a/workflow/metrics/metrics_k8s_request.go
+++ b/workflow/metrics/metrics_k8s_request.go
@@ -38,15 +38,19 @@ func addK8sRequests(_ context.Context, m *Metrics) error {
 	return err
 }
 
+type metricsRoundTripperContext struct {
+	ctx     context.Context
+	metrics *Metrics
+}
+
 type metricsRoundTripper struct {
-	ctx          context.Context
+	*metricsRoundTripperContext
 	roundTripper http.RoundTripper
-	metrics      *Metrics
 }
 
 // This is a messy global as we need to register as a roundtripper before
 // we can instantiate metrics
-var k8sMetrics metricsRoundTripper
+var k8sMetrics metricsRoundTripperContext
 
 func (m metricsRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	startTime := time.Now()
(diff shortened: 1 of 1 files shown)


meln5674:
Can anyone comment on [this test failure](https://github.com/argoproj/argo-workflows/actions/runs/10974775444/job/30473708609?pr=13641#step:16:12084)? Why is it expecting 404? That metric seems to be reporting healthily [here](https://github.com/argoproj/argo-workflows/actions/runs/10974775444/job/30473708609?pr=13641#step:16:5502). Was this maybe a hack added to work around something related to this same issue?

Joibel:
I believe this is a flakey test that does just sometimes fail. I don't remember introducing it when the roundtripper was added. I've kicked it to run again.

Anyway, that's for finding this and testing the RC. I will take a proper look when I have access to a proper computer.

meln5674:
Regarding the `test-executor` failure, it looks like the executor image is going missing between sometime after  its loaded at the start of the test, and it looks like the same thing is happening on a lot of recent runs of it. I've seen similar issues before my own actions when the runner is low on storage and kubelet tries to garbage collect images I need later in the test. You can mitigate this with the `imageGC{High,Low}ThresholdPercent` [kubelet config fields](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/), which are easy enough to set in KinD, but I've never done it in K3s.

agilgur5 (approved):
Thanks for investigating and your detailed root cause analysis!
If you want to improve on anything else in the repo, like other data races, flakey tests (which are themselves sometimes due to races), race detecting tests, or anything else, that'd be great 🙂

This LGTM but I'll defer to Alan for final approval and merge since he wrote this code recently

Joibel (approved):
Nice find, thank you for fixing it.

Issue #13644, opened 2024-09-22
`semaphores` error: `cannot get LockName for a Sync of Unknown type`


ni-todo-spot:
### Pre-requisites

- [X] I have double-checked my configuration
- [ ] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

Tried following the example of using a semaphore in WF but once I submit the WF using the semaphore - I get an error.

I've copied and pasted the example used in the docs (& I did mind the namespace name).

### Version(s)

v3.5.8

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

Followed the steps described in here: https://argo-workflows.readthedocs.io/en/latest/synchronization/#synchronization

created the CM in my argo namespace:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
 name: my-config
data:
  workflow: "1"  # Only one workflow can run at given time in particular namespace
  template: "2"  # Two instances of template can run at a given time in particular namespace
```

Next created the dummy WF using the lock:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: synchronization-wf-level-
spec:
  entrypoint: hello-world
  synchronization:
    semaphores:
      - configMapKeyRef:
          name: my-config
          key: workflow
  templates:
  - name: hello-world
    container:
      image: busybox
      command: [echo]
      args: ["hello world"]
```

Once I submit the WF via UI - I see in the in the execution window this error message:
```
 Failed: Failed to acquire the synchronization lock. requested configuration is invalid: cannot get LockName for a Sync of Unknown type
```

### Logs from the workflow controller

I have 2 controllers deployed. These are the logs of controller A:
```
{"Phase":"","ResourceVersion":"1279411754","level":"info","msg":"Processing workflow","namespace":"argo","time":"2024-09-22T07:54:17.594Z","workflow":"synchronization-wf-level-ww7cf"}
[...]

PR #13645, opened 2024-09-22
docs: in-line versioning + deprecation for multi-semaphore

agilgur5:
Addresses #13644, https://github.com/argoproj/argo-workflows/pull/13358#discussion_r1770585773 etc: <details>
> > Since people will look at the examples and not know that they have future syntax etc
> 
> Indeed it looks like this is happening almost as soon as this was merged 😕 #13644 plus [two](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1726994955560259) [Slack threads](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1726835793647219?thread_ts=1726835793.647219&cid=C01QW9QSSSK). The issue and former Slack thread are also from a [long](https://github.com/argoproj/argo-workflows/discussions?discussions_q=author%3Ani-todo-spot) [time](https://github.com/argoproj/argo-workflows/issues?q=author%3Ani-todo-spot) user, i.e. not a newbie.

And ofc [this contributor thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1721656672337809) and [previous Contributor Meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.5a2asxhc64ss) with Alan and I butting heads mostly over in-line version specifiers 😅 

</details>

### Motivation

Mitigate users not realizing they're on the `latest` version of the docs and attempting to use features that don't exist yet, getting errors, and then filing issues or support requests about it. 
To some extent, it is inevitable, but in-line versioning is a mitigation in an effort to help users self-service.

### Modifications

- remove warning about using mutexes and semaphores together as that is possible in 3.6 after #13358 -- the warning should only be in earlier versions

- use [snippet embeds](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files) to directly use examples in the docs without duplication
  - single source-of-truth that is testable, consistent, and more discoverable. also avoids potential mistakes/differences, etc
    - Mentioned this in https://github.com/argoproj/argo-workflows/pull/13429#issuecomment-2267597431 and elsewhere, thought this was a good time to start using them
  - no longer need a list of examples on this page as such

- add in-line comments with version specifiers on `mutexes` and `semaphores` syntax
- add deprecation comments with version specifiers on `mutex` and `semaphore` syntax
- add alternative syntax in comments, as suggested in https://github.com/argoproj/argo-workflows/pull/13358#discussion_r1685896125
  - this should make it very hard for users to get wrong, which is the intent
  
- remove "Legacy" section as this does not match conventions
  - we conventionally don't explicitly describe prior state in the docs and leave that to the older versions of the docs (#11390)
    - (if we did, we could have "legacy" sections ad infinitum. In [the Slack thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1721898722960129?thread_ts=1721656672.337809&cid=C0510EUH90V), this is also 3., that we do _not_ do and is one of the purposes of versioned docs)
  - users also don't read separate sections like this, they get an error and file an issue before they even reach this section, _literally_ per #13644 et al
  - the in-line comments cover this with more direct examples and as you're reading the example instead of after. this follows "show, don't tell" and style guide on directness.
    - the in-line is also more implicit of a reference to older syntax as opposed to an entire explicit section on it
  - we also conventionally don't describe bugs in the docs; combining `mutex` with `semaphore` and it not erroring out but silently ignoring the `mutex` is a confusing bug.
    - (if we did describe bugs, we could have those ad infinitum in the docs as well) 

### Verification

`make docs` passes.

<details><summary>Here's a screenshot of the example snippet embeds:</summary>

Screenshot 2024-09-22 at 1 06 26 PM

</details>

### Future Work

1. [ ] Use embeds for examples in more places in the docs

### DIFF
--- a/docs/fields.md
+++ b/docs/fields.md
@@ -2399,7 +2399,11 @@ Mutex holds Mutex configuration
 
 - [`synchronization-mutex-tmpl-level-legacy.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level-legacy.yaml)
 
+- [`synchronization-mutex-tmpl-level.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level.yaml)
+
 - [`synchronization-mutex-wf-level-legacy.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-wf-level-legacy.yaml)
+
+- [`synchronization-mutex-wf-level.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-wf-level.yaml)
 </details>
 
 ### Fields
@@ -3291,7 +3295,11 @@ MutexStatus contains which objects hold mutex locks, and which objects this work
 
 - [`synchronization-mutex-tmpl-level-legacy.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level-legacy.yaml)
 
+- [`synchronization-mutex-tmpl-level.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level.yaml)
+
 - [`synchronization-mutex-wf-level-legacy.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-wf-level-legacy.yaml)
+
+- [`synchronization-mutex-wf-level.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-wf-level.yaml)
 </details>
 
 ### Fields
--- a/docs/synchronization.md
+++ b/docs/synchronization.md
@@ -208,30 +208,6 @@ Examples:
 !!! Warning
     If a Workflow is at the front of the queue and it needs to acquire multiple locks, all other Workflows that also need those same locks will wait. This applies even if the other Workflows only wish to acquire a subset of those locks.
 
-## Legacy
-
-In workflows prior to 3.6 you can only specify one lock in any one workflow or template using either a mutex:
-
-```yaml
-    synchronization:
-      mutex:
-     ...
-```
-
-or a semaphore:
-
-```yaml
-    synchronizaion:
-   semamphore:
-     ...
-```
-
-Specifying both would not work in <3.6, only the semaphore would be used.
-
-The single `mutex` and `semaphore` syntax still works in version 3.6 but is considered deprecated.
-Both the `mutex` and the `semaphore` will be taken in version 3.6 with this syntax.
-This syntax can be mixed with `mutexes` and `semaphores`, all locks will be required.
-
 ## Other Parallelism support
 
 You can also [restrict parallelism at the Controller-level](parallelism.md).
--- a/docs/synchronization.md
+++ b/docs/synchronization.md
@@ -24,10 +24,6 @@ data:
   template: "2"  # Two instances of template can run at a given time in particular namespace
 ```
 
-!!! Warning
-    Each synchronization block may only refer to either a semaphore or a mutex.
-    If you specify both only the semaphore will be locked.
-
 ## Workflow-level Synchronization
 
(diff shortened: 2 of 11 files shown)


docs/synchronization.md
```diff
@@ -24,10 +24,6 @@ data:
   template: "2"  # Two instances of template can run at a given time in particular namespace
 ```
 
-!!! Warning
-    Each synchronization block may only refer to either a semaphore or a mutex.
-    If you specify both only the semaphore will be locked.
```
agilgur5, line 29: No longer relevant in 3.6

docs/synchronization.md
```diff
@@ -204,30 +109,6 @@ Examples:
...
-
-or a semaphore:
-
-```yaml
-    synchronizaion:
-   semamphore:
-     ...
-```
-
-Specifying both would not work in <3.6, only the semaphore would be used.
```
agilgur5, line 225: We just need to fix this bug per https://github.com/argoproj/argo-workflows/pull/13358#issuecomment-2250795586 and https://github.com/argoproj/argo-workflows/issues/11859 -- i.e. give a validation error instead of silently ignoring it in 3.5 and prior versions

docs/synchronization.md
```diff
@@ -204,30 +109,6 @@ Examples:
...
-
-The single `mutex` and `semaphore` syntax still works in version 3.6 but is considered deprecated.
-Both the `mutex` and the `semaphore` will be taken in version 3.6 with this syntax.
-This syntax can be mixed with `mutexes` and `semaphores`, all locks will be required.
```
agilgur5, line 229: if we want to describe this, this would be better described with another example

examples/synchronization-mutex-tmpl-level-legacy.yaml
```diff
@@ -28,18 +27,22 @@ spec:
 
   - name: acquire-lock
     synchronization:
-      mutex:
+      mutex: # deprecated: v3.5 and before
         name: welcome
+      # mutexes: # v3.6 and after
+      #   - name: welcome
```
agilgur5, line 33: I'm inclined to just move these to internal tests and otherwise remove the legacy examples entirely from the user-facing `examples/` dir

and flipped syntax + in-line comments here is a stopgap, especially as the in-line deprecation is hard to understand without the future syntax right next to it (and users are likely to miss `-legacy` name too)

docs/fields.md
```diff
@@ -2399,7 +2399,11 @@ Mutex holds Mutex configuration
 
 - [`synchronization-mutex-tmpl-level-legacy.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level-legacy.yaml)
 
+- [`synchronization-mutex-tmpl-level.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-tmpl-level.yaml)
+
 - [`synchronization-mutex-wf-level-legacy.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-wf-level-legacy.yaml)
+
+- [`synchronization-mutex-wf-level.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/synchronization-mutex-wf-level.yaml)
```
agilgur5, line 2406: this is suboptimal as the field checker is plain text search, but nbd in context as it does show it and say it's deprecated

docs/synchronization.md
agilgur5: See #13669

docs/synchronization.md
```diff
@@ -36,44 +32,14 @@ In this example the synchronization key `workflow` is configured as limit `"1"`,
...
-          name: my-config
-          key: workflow
-  templates:
-  - name: hello-world
-    container:
-      image: busybox
-      command: [echo]
-      args: ["hello world"]
+```yaml title="examples/synchronization-wf-level.yaml"
+--8<-- "examples/synchronization-wf-level.yaml:12"
```
juliev0, line 36: neat

juliev0 (approved):

PR #13648, opened 2024-09-23
fix(controller): node message should be updated when lock msg changed

### LINKED ISSUE #13629: Templated mutex name not resolved
### Pre-requisites

The mutex name is not resolved correctly when it is based on a workflow parameter, which is also based on another workflow parameter.

See the workflow below.
1. mutex name is defined as `{{ workflow.parameters.derived-mutex-name }}`
2. `derived-mutex-name` is defined as `"{{ workflow.parameters.base-mutex-name }}"`
3. `base-mutex-name` is defined as a parameter of the argo submit command line, let's say `mycustommutex`

In this workflow dag is used (and might be the reason of the defect).

With this workflow, if I run the following command:
```
argo submit my-workflow.yml --parameter base-mutex-name=mycustommutex
```
 I get the following message indicating the name of the mutex at runtime:
```
Waiting for argo-wf/Mutex{{ workflow.parameters.base-mutex-name }} lock. Lock status: 0/1
```
However the command `echo {{ workflow.parameters.derived-mutex-name }}` outputs correctly the resolved name:
```
mycustommutex
```

I have not tested this scenario with the latest version, because I'm not responsible of the deployment of argo in my company.

### Version(s)

v3.5.8

```YAML
kind: Workflow
metadata:
  generateName: mutex-name-
spec:
  entrypoint: main
  arguments:
    parameters:
    - name: base-mutex-name
    - name: derived-mutex-name
      value: "{{ workflow.parameters.base-mutex-name }}"
  templates:
  - name: main
    dag:
      tasks:
      - name: first
        template: print-mutex-name
      - name: second
        template: print-mutex-name
  - name: print-mutex-name
    synchronization:
      mutex:
        name: "{{ workflow.parameters.derived-mutex-name }}"
    container:
      image: alpine
      command: [ "/bin/sh", "-c" ]
      args:
      - |
        #!/bin/bash
        echo {{ workflow.parameters.derived-mutex-name }}
        sleep 30
```

```text
I'm not granted the permission to get these logs
```

```text
I'm not granted the permission to get these logs
```

chengjoey:
Fixes #13629

### Motivation
When the sync lock is not released, the latest lock-msg is always used as the node message

### DIFF
--- a/workflow/controller/operator_test.go
+++ b/workflow/controller/operator_test.go
@@ -8491,6 +8491,10 @@ metadata:
   namespace: argo
 spec:
   entrypoint: whalesay
+  arguments:
+    parameters:
+    - name: derived-mutex-name
+      value: welcome
   templates:
   - container:
       args:
@@ -8501,7 +8505,7 @@ spec:
     name: whalesay
     synchronization:
       mutex:
-        name: welcome
+        name: "{{ workflow.parameters.derived-mutex-name }}"
   ttlStrategy:
     secondsAfterCompletion: 600
 status:
@@ -8510,7 +8514,6 @@ status:
       displayName: hello-world-mpdht
       finishedAt: null
       id: hello-world-mpdht
-      message: 'Waiting for argo/Mutex/welcome lock. Lock status: 0/1 '
       name: hello-world-mpdht
       phase: Pending
       progress: 0/1
@@ -8537,6 +8540,8 @@ func TestMutexWfPendingWithNoPod(t *testing.T) {
 	ctx := context.Background()
 	controller.syncManager = sync.NewLockManager(GetSyncLimitFunc(ctx, controller.kubeclientset), func(key string) {
 	}, workflowExistenceFunc)
+
+	// preempt lock
 	_, _, _, _, err := controller.syncManager.TryAcquire(wf, "test", &wfv1.Synchronization{Mutex: &wfv1.Mutex{Name: "welcome"}})
 	require.NoError(t, err)
 	woc := newWorkflowOperationCtx(wf, controller)
@@ -8544,6 +8549,11 @@ func TestMutexWfPendingWithNoPod(t *testing.T) {
 	woc.operate(ctx)
 	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
 	assert.Equal(t, wfv1.NodePending, woc.wf.Status.Nodes.FindByDisplayName("hello-world-mpdht").Phase)
+	assert.Equal(t, "Waiting for argo/Mutex/welcome lock. Lock status: 0/1", woc.wf.Status.Nodes.FindByDisplayName("hello-world-mpdht").Message)
+
+	woc.controller.syncManager.Release(wf, "test", &wfv1.Synchronization{Mutex: &wfv1.Mutex{Name: "welcome"}})
+	woc.operate(ctx)
+	assert.Equal(t, "", woc.wf.Status.Nodes.FindByDisplayName("hello-world-mpdht").Message)
 }
 
 var wfGlobalArtifactNil = `apiVersion: argoproj.io/v1alpha1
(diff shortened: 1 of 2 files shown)


chengjoey:
/hold
add test

MasonM:
@chengjoey FYI: I entered https://github.com/argoproj/argo-workflows/pull/13660 to fix the flaky tests. After that's merged, click the "Update branch" button at the bottom of the PR

Joibel:
> /hold cancel

@chengjoey is this ready for review, I'm unsure of what you meant by this comment?

chengjoey:
> > /hold cancel
> 
> @chengjoey is this ready for review, I'm unsure of what you meant by this comment?

yeah, it is ready for review, since I didn't add the test at first, I used `/hold cancel` to get ready after adding the test

workflow/controller/operator_test.go
```diff
@@ -8512,13 +8517,18 @@ func TestMutexWfPendingWithNoPod(t *testing.T) {
 	ctx := context.Background()
 	controller.syncManager = sync.NewLockManager(GetSyncLimitFunc(ctx, controller.kubeclientset), func(key string) {
 	}, workflowExistenceFunc)
-	_, _, _, _, err := controller.syncManager.TryAcquire(wf, "test", &wfv1.Synchronization{Mutex: &wfv1.Mutex{Name: "welcome"}})
+	lockAcquired, _, msg, _, err := controller.syncManager.TryAcquire(wf, "test", &wfv1.Synchronization{Mutex: &wfv1.Mutex{Name: "welcome"}})
 	require.NoError(t, err)
 	woc := newWorkflowOperationCtx(wf, controller)
+	if !lockAcquired {
```
Joibel, line 8523: This feels wrong for a test.

Surely the outcome of `TryAquire` should always be the same for this test, and therefore we could assert that `!lockAcquired`? If lockAquired is true here we're not testing the new functionality.

Joibel (changes requested):
As a follow up to this it would be possible to list all of the failing locks rather than just the first, so if you feel like doing that it would be useful.

workflow/controller/operator_test.go
chengjoey: done~, I modified the test case and handed over the modification of msg and status to `operate`

workflow/controller/operator_test.go
```diff
@@ -8510,15 +8512,20 @@ func TestMutexWfPendingWithNoPod(t *testing.T) {
 	cancel, controller := newController(wf)
 	defer cancel()
 	ctx := context.Background()
-	controller.syncManager = sync.NewLockManager(GetSyncLimitFunc(ctx, controller.kubeclientset), func(key string) {
```
Joibel, line 8513: Why has this been removed in this iteration?

workflow/controller/operator_test.go
```diff
@@ -8510,15 +8512,20 @@ func TestMutexWfPendingWithNoPod(t *testing.T) {
...
+
+	// preempt lock
 	_, _, _, _, err := controller.syncManager.TryAcquire(wf, "test", &wfv1.Synchronization{Mutex: &wfv1.Mutex{Name: "welcome"}})
 	require.NoError(t, err)
 	woc := newWorkflowOperationCtx(wf, controller)
 
 	woc.operate(ctx)
 	assert.Equal(t, wfv1.WorkflowRunning, woc.wf.Status.Phase)
 	assert.Equal(t, wfv1.NodePending, woc.wf.Status.Nodes.FindByDisplayName("hello-world-mpdht").Phase)
+	assert.Equal(t, "Waiting for argo/Mutex/welcome lock. Lock status: 0/1", woc.wf.Status.Nodes.FindByDisplayName("hello-world-mpdht").Message)
```
Joibel, line 8524: :+1:

Joibel (changes requested):

workflow/controller/operator_test.go
Joibel: This part looks good.

workflow/controller/operator_test.go
chengjoey: syncManager will be created by `newController`, so maybe this is unnecessary

https://github.com/argoproj/argo-workflows/blob/5310c396ee1e9f4e5575f62ef5340a86540d8679/workflow/controller/controller_test.go#L331-L333

chengjoey: thanks @Joibel , i got it. 
done

Joibel:
This needs the linter issue fixing @chengjoey

chengjoey:
> This needs the linter issue fixing @chengjoey

thanks @Joibel 

done

Joibel (approved):

PR #13660, opened 2024-09-25
ci: fix flaky image gc and mysql connect + add k3s debug log

MasonM:
### Motivation

A bunch of tests have been failing intermittently for awhile now, which is blocking PRs. For example, the `E2E Tests (test-executor, v1.28.13+k3s1, minimal, false)` test fails >50% of the time with the error `ErrImageNeverPull: Container image "quay.io/argoproj/argocli:latest" is not present with pull policy of Never.` ([example run]( https://github.com/argoproj/argo-workflows/actions/runs/11022848664/job/30613101811)). Many of these issues can't be diagnosed without access to the k3s logs.

### Modifications

1. Print the k3s logs using `journalctl` on a build failure. You can see k3s is being run with systemd from the output of the `Install and start K3S` step:
```
[INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
[INFO]  systemd: Enabling k3s unit
Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service → /etc/systemd/system/k3s.service.
[INFO]  systemd: Starting k3s
```
2. Add a step to clean up disk space by deleting unused files (e.g. Haskell), since the k3s logs clearly indicate pods are being evicted due to DiskPressure
3. Update `make wait PROFILE=mysql` to wait for mysql to be available, which should fix intermittent failures like `persistence.go:29: test panicked: dial tcp [::1]:3306: connect: connection refused`
4. Add a [Kubelet configuration file](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) to tell k3s not to garbage collect images, since that's what's causing the `ErrImageNeverPull` errors

### Verification

Will watch action output

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -319,7 +319,7 @@ jobs:
             UI=false \
             POD_STATUS_CAPTURE_FINALIZER=true > /tmp/argo.log 2>&1 &
       - name: Wait for controller to be up
-        run: make wait API=${{matrix.use-api}}
+        run: make wait PROFILE=${{matrix.profile}} API=${{matrix.use-api}}
         timeout-minutes: 5
       - name: Run tests ${{matrix.test}}
         run: make ${{matrix.test}} E2E_SUITE_TIMEOUT=20m STATIC_FILES=false
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -240,6 +240,13 @@ jobs:
             profile: minimal
             use-api: false
     steps:
+      - name: Maximize build space
+        uses: AdityaGarg8/remove-unwanted-software@8831c82abf29b34eb2caac48d5f999ecfc0d8eef
+        with:
+          remove-dotnet: 'true'
+          remove-android: 'true'
+          remove-haskell: 'true'
+          remove-codeql: 'true'
       - name: Install socat (needed by Kubernetes)
         run: sudo apt-get -y install socat
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -280,10 +280,6 @@ jobs:
           kind: KubeletConfiguration
           imageMinimumGCAge: 1h
           imageGCHighThresholdPercent: 100
-          evictionHard:
-            imagefs.available: 0%
-            nodefs.available: 0%
-            nodefs.inodesFree: 0%
           EOT
 
           curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=stable INSTALL_K3S_EXEC='--docker --kubelet-arg=config=/tmp/kubelet.config' K3S_KUBECONFIG_MODE=644 sh -
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -313,6 +313,9 @@ jobs:
         run: make ${{matrix.test}} E2E_SUITE_TIMEOUT=20m STATIC_FILES=false
 
       # failure debugging below
+      - name: Failure debug - k3s logs
+        if: ${{ failure() }}
+        run: journalctl -u k3s
       - name: Failure debug - describe MinIO/MySQL deployment
         if: ${{ failure() }}
         run: |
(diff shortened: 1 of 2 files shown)


agilgur5:
This is a follow-up to [this Slack thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1727215958212459)

@meln5674 mentioned this might be due to image GC in https://github.com/argoproj/argo-workflows/pull/13641#issuecomment-2365328985. Seems like you saw that in the logs too per the thread:

```
2024-09-25T02:54:13.4531309Z Sep 25 02:47:55 fv-az1240-855 k3s[2120]: E0925 02:47:55.910731    2120 kubelet.go:1427] "Image garbage collection failed multiple times in a row" err="Failed to garbage collect required amount of images. Attempted to free 9859177676 bytes, but only found 4090859086 bytes eligible to free."
```

Makefile
```diff
@@ -574,6 +574,10 @@ ifeq ($(API),true)
 	# Wait for Argo Server
 	until lsof -i :2746 > /dev/null ; do sleep 10s ; done
 endif
+ifeq ($(PROFILE),mysql)
+	# Wait for MySQL
+	until (: < /dev/tcp/localhost/3306) ; do sleep 10s ; done
```
MasonM, line 579: Note: I'm intentionally not using `lsof -i :3306` like the lines above because that isn't sufficient to prevent `connection refused` errors, because `lsof -i :3306` only tells you if something is listening on port 3306. I believe those errors happening in between the time `mysqld` binds to port 3306 and when it's ready to accept connections. Accessing `/dev/tcp/localhost/3306` actually opens a connection, and will fail if the connection is refused.

.github/workflows/ci-build.yaml
```diff
@@ -240,6 +240,13 @@ jobs:
             profile: minimal
             use-api: false
     steps:
+      - name: Free up disk space
```
agilgur5, line 243:
```suggestion
      - name: Free up unused disk space
```

clarify a bit

.github/workflows/ci-build.yaml
```diff
@@ -240,6 +240,13 @@ jobs:
...
+        run: |
+          printf "==> Available space before cleanup\n"
+          df -h
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /usr/local/.ghcup /opt/hostedtoolcache/CodeQL
```
agilgur5, line 247:
```suggestion
          # these directories are not used by E2Es
          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /usr/local/.ghcup /opt/hostedtoolcache/CodeQL
```

add a clarifying comment

.github/workflows/ci-build.yaml
```diff
@@ -267,7 +274,15 @@ jobs:
...
             export INSTALL_K3S_VERSION=${{ matrix.install_k3s_version }}
           fi
-          curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=stable INSTALL_K3S_EXEC=--docker K3S_KUBECONFIG_MODE=644 sh -
+
+          cat >/tmp/kubelet.config <<EOT
+          apiVersion: kubelet.config.k8s.io/v1beta1
+          kind: KubeletConfiguration
+          imageMinimumGCAge: 1h
+          imageGCHighThresholdPercent: 100
+          EOT
```
agilgur5, line 283: I'd probably put this into an actual file in the repo with comments as well

agilgur5 (approved):
Thanks for tracking this down Mason!

Approving this so we can unblock CI, but I do have some small comments below that we can resolve in a follow-up PR

Joibel:
@MasonM thanks for finding and fixing this.

MasonM:
Thanks @agilgur5! I entered https://github.com/argoproj/argo-workflows/pull/13670 to address your comments.

.github/workflows/ci-build.yaml
agilgur5: Added in #13670

.github/workflows/ci-build.yaml
agilgur5: Added in #13670

.github/workflows/ci-build.yaml
agilgur5: Completed in #13670

PR #13664, opened 2024-09-25
test: add test for expression template `hasVarInEnv`

weafscast:
Fixes #TODO

### Motivation

Split from https://github.com/argoproj/argo-workflows/pull/13591 so that this can be independently merged

### Modifications

Add test cases

### Verification

Add test cases

### DIFF
--- a/util/template/expression_template_test.go
+++ b/util/template/expression_template_test.go
@@ -34,3 +34,50 @@ func Test_hasWorkflowParameters(t *testing.T) {
 		assert.False(t, hasWorkflowFailures("workflow .failures"))
 	})
 }
+
+func Test_hasVarInEnv(t *testing.T) {
+	t.Run("parameterExistsInEnv", func(t *testing.T) {
+		env := map[string]interface{}{
+			"workflow": map[string]interface{}{
+				"status": "Succeeded",
+			},
+		}
+		assert.True(t, hasVarInEnv(env, "workflow.status"))
+	})
+
+	t.Run("parameterDoesNotExistInEnv", func(t *testing.T) {
+		env := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


util/template/expression_template_test.go
```diff
@@ -34,3 +34,50 @@ func Test_hasWorkflowParameters(t *testing.T) {
 		assert.False(t, hasWorkflowFailures("workflow .failures"))
 	})
 }
+
+func Test_hasVarInEnv(t *testing.T) {
```
isubasinghe, line 38: Please use the existing function naming convention of `TestHasVarInEnv`

isubasinghe (changes requested):

util/template/expression_template_test.go
weafscast: will change, I just kept it consistent with this file.

weafscast, line 38:
```suggestion
func TestHasVarInEnv(t *testing.T) {
```

weafscast:
cc/ @agilgur5

Joibel:
Thanks for splitting this out, and sorry it sat for so long.

The coverage you were adding here has since landed independently: #15442 rewrote `hasVarInEnv` (longest-prefix matching plus reflection over maps, structs, embedded fields and pointers) and added a `Test_hasVarInEnv` that already covers present/absent keys, nested maps, deeper nesting, struct and pointer fields, and empty/nil envs - a superset of the cases in this PR. Subsequent PRs (#16223, #16274) added tests for `anyVarNotInEnv`, `missingVarsInEnv` and `getIdentifiers` alongside it.

Closing as already covered on main. Thanks again for the contribution, and please do open new PRs if you spot gaps in the current tests.

PR #13666, opened 2024-09-26
docs: add sidecar kill command for linkerd-proxy

rowieg:
Addresses https://github.com/argoproj/argo-workflows/issues/12103#issuecomment-2376057950

### Motivation
This adds documentation for an annotation to shut down linkerd-proxy containers that are added as a sidecar to workflow pods through the linkerd service mesh.
If these containers are not closed the pod keeps running and is not completed even though the workflow is done


agilgur5 (approved):
Thanks for adding this!

I've modified the first sentence of your description as this does not resolve that issue, but does address some [unrelated](https://github.com/argoproj/argo-workflows/issues/12103#issuecomment-2364101333) comments

PR #13667, opened 2024-09-26
docs: Clarify headless service for promethus scraping

tico24:
### Motivation

The docs now separate prometheus and OTEL metrics information, but there was some prometheus stuff hanging towards the bottom of the page. It makes sense to move this up.

Also people struggled to read the last line about using a headless service, so made it more obvious and updated the example to match.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -116,11 +116,15 @@ metricsConfig: |
   secure: true
 ```
 
-The metric names emitted by this mechanism are prefixed with `argo_workflows_`.
-`Attributes` are exposed as Prometheus `labels` of the same name.
+The metric names emitted by this mechanism are prefixed with `argo_workflows_`. `Attributes` are exposed as Prometheus `labels` of the same name.
 
 Prometheus metrics will return empty metrics on a workflow controller which is not the leader.
-All metrics emitted over Prometheus will have `argo_workflows_` prefixed to their name.
+
+By port-forwarding to the leader controller Pod you can view the metrics in your browser at `https://localhost:9090/metrics`. Assuming you only have one controller replica, you can port-forward with:
+
+```bash
+kubectl -n argo port-forward deploy/workflow-controller 9090:9090
+```
 
 ### Common
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -60,7 +60,8 @@ metricsConfig: |
 
 ### Prometheus scraping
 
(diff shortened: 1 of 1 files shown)


Joibel:
Did you deliberately drop the port-forwarding docs?

Joibel:
> Yes, but this will give you the impression I really thought about it. My vague reasoning is that you now define the port, path and security of the URL, and we already define the service. If you can't work out that the service points to the metrics and that you could port forward that, then maybe this isn't for you?
> 
> Happy to add it back though.

We document this elsewhere, and for beginners this is a useful way to understand or remember they can see the numbers there to verify where in this kubernetes world of duct tape and string things aren't going well for them today.

docs/metrics.md
```diff
@@ -81,11 +116,15 @@ metricsConfig: |
   secure: true
 ```
 
-The metric names emitted by this mechanism are prefixed with `argo_workflows_`.
-`Attributes` are exposed as Prometheus `labels` of the same name.
+The metric names emitted by this mechanism are prefixed with `argo_workflows_`. `Attributes` are exposed as Prometheus `labels` of the same name.
```
Joibel, line 119: Also here.

docs/metrics.md
```diff
@@ -81,11 +116,15 @@ metricsConfig: |
...
 
 Prometheus metrics will return empty metrics on a workflow controller which is not the leader.
-All metrics emitted over Prometheus will have `argo_workflows_` prefixed to their name.
+
+By port-forwarding to the leader controller Pod you can view the metrics in your browser at `https://localhost:9090/metrics`. Assuming you only have one controller replica, you can port-forward with:
```
Joibel, line 123: Also here.

docs/metrics.md
```diff
@@ -60,6 +60,41 @@ metricsConfig: |
 
 ### Prometheus scraping
 
+A metrics service is not installed as part of [the default installation](quick-start.md) so you will need to add one if you wish to use a Prometheus Service Monitor. If you have more than one controller pod, using one as a [hot-standby](high-availability.md), you should use [a headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) to ensure that each pod is being scraped so that no metrics are missed.
```
Joibel, line 63: Please convert to [one sentence per line](https://github.com/argoproj/argo-workflows/blob/bf64aeb92ae7d250a677545dae377585513089c9/docs/doc-changes.md?plain=1#L12) for differencing and patching purposes.

Joibel (changes requested):

Joibel (approved):

PR #13669, opened 2024-09-26
fix(sync): error if mutex and semaphore are used together in <=v3.5

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/issues/11859
Addresses https://github.com/argoproj/argo-workflows/pull/13358#issuecomment-2250795586, https://github.com/argoproj/argo-workflows/pull/13645#discussion_r1770613962

### Motivation

- this is an invalid spec/syntax, so instead of silently ignoring the mutex, throw a validation error

### Modifications

- add `validateSync` function to `validate.go` and use during `validateWorkflow` and `validateTemplate`
  - check for using mutex and semaphore together and also using neither
    - the latter check means that we could simplify later code around "unknown" sync types, i.e. https://github.com/argoproj/argo-workflows/pull/13358#discussion_r1705600604, https://github.com/argoproj/argo-workflows/pull/13358#discussion_r1714498505 

- remove doc sentence on mutex being ignored, as it will error now instead

### Verification

add unit tests for both invalid cases and one valid case for Workflow-level and template-level synchronization in `validate_test.go`

### Future Work

1. [ ] make similar validation for 3.6 -- neither case in particular
1. [ ] simplify later logic around "unknown" sync types per above

### DIFF
--- a/docs/synchronization.md
+++ b/docs/synchronization.md
@@ -26,7 +26,6 @@ data:
 
 !!! Warning
     Each synchronization block may only refer to either a semaphore or a mutex.
-    If you specify both only the semaphore will be locked.
 
 ## Workflow-level Synchronization
 
--- a/workflow/validate/testdata/synchronization-invalid-both.yaml
+++ b/workflow/validate/testdata/synchronization-invalid-both.yaml
@@ -6,7 +6,8 @@ spec:
   entrypoint: echo-template
   synchronization:
     # invalid: should only have 1 at a time per block
-    mutex: hello-world
+    mutex:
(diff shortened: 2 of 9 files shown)


agilgur5:
Ah, right, `ci-build.yaml` isn't running for PRs to release branches

agilgur5:
> [ ] make similar validation for 3.6 -- neither case in particular

it might make sense to do 3.6 validation first with this PR as a backportable variant of that?

Joibel:
This would only apply to 3.5, which is now out of support

PR #7440, opened 2021-12-17
fix: pkg/errors is no longer maintained

stephpalis:
Signed-off-by: Stephanie Palis <stephanie.palis@acquia.com>
Fixes #7402

Don't bother creating a PR until you've done this:

* [x] Run `make pre-commit -B` to fix codegen, lint, and commit message problems.

Create your PR as a draft.

* Your PR needs to pass the required checks before it can be approved. If the check is not required (e.g. E2E tests) it
  does not need to pass.
* Once required tests have passed, you can make it "Ready for review".
* Say how how you tested your changes. If you changed the UI, attach screenshots.

Tips:

* If changes were requested, and you've made them, then dismiss the review to get it looked at again.
* Add you organization to [USERS.md](https://github.com/argoproj/argo-workflows/blob/master/USERS.md) if you like.
* You can ask for help!
___________________________________________________________________________________
This PR removes the functionality of StackTracer as a result of removing pkg/errors

PR #9026, opened 2022-06-22
feat: Add support for Azure Blob Storage artifacts Fixes #1540

brianloss:
Add support for artifacts using Azure Blob Storage native APIs rather
than minio. Testing has been performed against a local instance.

Signed-off-by: Brian Loss <brianloss@gmail.com>

Fixes #1540

PR #13673, opened 2024-09-27
refactor(deps): drop `pkg/errors` as a direct dependency

### LINKED ISSUE #7402: pkg/errors is no longer maintained
# Summary

pkg/errors is no longer maintained. See https://github.com/pkg/errors/issues/245

# Use Cases

We can replaces all calls to errors.Wrap/f with fmt.Errorf and removes the unneeded dependency on
github.com/pkg/errors since we don't support go 1.12 or lower.  See https://github.com/microsoft/go-winio/pull/227 for an example.

---

blkperl:
Fixes #7402

This a follow up to https://github.com/argoproj/argo-workflows/pull/7440 where we missed one file.

### Motivation

* pkgs/errors is no longer maintained

### Modifications

* changed workflow/artifacts/azure/azure.go to use fmt instead of errors

### Verification

E2E tests will verify

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,6 @@ require (
 	github.com/jcmturner/gokrb5/v8 v8.4.4
 	github.com/klauspost/pgzip v1.2.6
 	github.com/minio/minio-go/v7 v7.0.66
-	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.19.0
 	github.com/prometheus/common v0.48.0
 	github.com/robfig/cron/v3 v3.0.1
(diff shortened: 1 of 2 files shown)


agilgur5:
> This a follow up to #7440 where we missed one file.

Huh it looks like the `go.mod` [changed there too](https://github.com/argoproj/argo-workflows/pull/7440/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6), so I'm guessing this line was added after or there was a merge conflict or something that restored it

agilgur5:
Looks like #9026 added it back, potentially because it used older code as a base

go.mod
```diff
@@ -118,6 +117,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/ncruces/go-strftime v0.1.9 // indirect
 	github.com/pjbgf/sha1cd v0.3.0 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
```
agilgur5, line 120: Can you check which transitive dep uses this? If it's `argo/pkg` for instance, we could refactor it out of there as well

agilgur5 (approved):
LGTM, thanks for fixing the one leftover line

go.mod
blkperl: Yes it's argo/pkg which looks like it hasn't been updated in a long time.

agilgur5: We do infrequently update code there when we need to support features on Workflows (e.g. some expression templating code and some S3 code).

Do you think you could take a look and do a refactor there too? That way all Argo projects can stop depending on it indirectly

agilgur5: Completed in https://github.com/argoproj/pkg/pull/747

PR #13674, opened 2024-09-28
build: support TLS for UI and Server with `make start`

MasonM:
### Motivation

This was split off from https://github.com/argoproj/argo-workflows/pull/13609 to make it easier to review. This  adds a new `make start UI_SECURE=true` flag that can be used to test Argo with a TLS termination proxy in front, which is needed for testing bugs like https://github.com/argoproj/argo-workflows/issues/13031. Also, `make start SECURE=true UI=true` was broken, so I fixed that too.

### Modifications

* Update the `proxy` option in `ui/src/app/webpack.config.js` to use the new array syntax ([docs](https://webpack.js.org/configuration/dev-server/#devserverproxy)). The old object syntax is deprecated and was removed in https://github.com/webpack/webpack-dev-server/pull/4694
* Fixed `ui/src/app/webpack.config.js` to respect the `ARGO_SECURE` variable when determining the proxy endpoint
* Added `UI_SECURE` variable to `Makefile` and have `ui/src/app/webpack.config.js` use it to set the `server` option ([docs](https://webpack.js.org/configuration/dev-server/#devserverserver)).
* Updated `dev/nix/conf.nix` to pass the same variables as `Makefile`. I didn't test this.
* Updated `docs/running-locally.md` with information on the `SECURE`/`UI_SECURE` flags. Also, I added a section on debugging, since it took me some time to figure this out.

### Verification

Tested the following combinations and ensured the workflow list rendered properly each time:
1. `make start UI=true`
2. `make start UI_SECURE=true`
3. `make start UI_SECURE=true SECURE=true`
4. `make start UI=true SECURE=true`

Screenshot with  `make start UI_SECURE=true`:
Screenshot 2024-09-28 at 11 27 03 AM


isubasinghe (commented):
Tested and UI works as expected.

I am personally okay with the manual changing of the `redirectUrl`, the idiomatic way in this codebase would be to add something to the hacks folder to change the url to `https` instead of `http`.

I don't really like that solution, but this PR as a result represent a decision to diverge from that thinking so I will leave it up for another maintainer to decide.

MasonM:
@isubasinghe Okay, I pushed d507aa638dd351c8b802b175defe85502c306c52 to add a `hack/update-sso-redirect-url.sh` script that rewrites the `redirectUrl` automatically. It's messy, but I can't think of a better solution.

docs/running-locally.md
```diff
@@ -206,6 +221,34 @@ Tests often fail: that's good. To diagnose failure:
 
 If tests run slowly or time out, factory reset your Kubernetes cluster.
 
+### Debugging using Visual Studio Code
+
+When using the Dev Container with VSCode, add the following launch configuration to `.vscode/launch.json`:
```
agilgur5, line 226: We could just add this to `.vscode/launch.json` directly instead of having a snippet here. Especially as we already have devcontainer config and Nix config, so there is precent for that. I had been thinking of doing that myself as well at some point

docs/running-locally.md
```diff
@@ -154,6 +154,21 @@ To test SSO integration, use `PROFILE=sso`:
 make start UI=true PROFILE=sso
 ```
 
+## TLS
+
+By default, `make start` will start Argo in [plain text mode](./tls.md#plain-text).
```
agilgur5, line 159:
```suggestion
By default, `make start` will start Argo in [plain text mode](tls.md#plain-text).
```

`./` isn't needed and isn't used in docs links

docs/running-locally.md
```diff
@@ -154,6 +154,21 @@ To test SSO integration, use `PROFILE=sso`:
...
+To simulate a TLS proxy in front of Argo, use `UI_SECURE=true` (which implies `UI=true`):
+
+```bash
+make start UI_SECURE=true
+```
+
+To start Argo in [encrypted mode](./tls.md#encrypted), use `SECURE=true`, which can be optionally combined with `UI_SECURE=true`:
```
agilgur5, line 166:
```suggestion
To start Argo in [encrypted mode](tls.md#encrypted), use `SECURE=true`, which can be optionally combined with `UI_SECURE=true`:
```

same as above

agilgur5 (commented):
Nice work on this Mason! I just have a few small comments below

> * Update the `proxy` option in `ui/src/app/webpack.config.js` to use the new array syntax ([docs](https://webpack.js.org/configuration/dev-server/#devserverproxy)). The old object syntax is deprecated and was removed in [refactor!: allow only Array for the proxy option webpack/webpack-dev-server#4694](https://github.com/webpack/webpack-dev-server/pull/4694)

Makes sense though I don't think I've seen a deprecation warning on this? Are we on an older version?

docs/running-locally.md
MasonM: There were two reasons I didn't directly commit `.vscode/launch.json`:
1. Anyone that already has a `.vscode/launch.json` will have it overwritten by Git the next time they do a `git pull`
2. It's common to add ad-hoc launch configurations for debugging specific things, which we don't want to commit

But I found an alternative that should avoid both problems: specifying default launch configurations inside `devcontainers.json`. I went ahead and did that in c430b3c7e0a3a6159270d939f945c06ea1c7594a. Docs: https://containers.dev/supporting#visual-studio

docs/running-locally.md
MasonM: Okay, done in c430b3c7e0a3a6159270d939f945c06ea1c7594a

docs/running-locally.md
MasonM: Okay, done in c430b3c7e0a3a6159270d939f945c06ea1c7594a

MasonM:
@agilgur5 Thanks for the review!

> Makes sense though I don't think I've seen a deprecation warning on this? Are we on an older version?

Yes, we're on webpack-dev-server v4.15.1:
https://github.com/argoproj/argo-workflows/blob/bf7760a46e29bf2571bd3e9d852220b44339bb60/ui/yarn.lock#L10063-L10064

and https://github.com/webpack/webpack-dev-server/commit/d7643251db183a00f277e76a640bad3ef77c2a8e was released in webpack-dev-server v5.0.0

docs/running-locally.md
agilgur5: > 2\. It's common to add ad-hoc launch configurations for debugging specific things, which we don't want to commit

I don't necessarily disagree with this and I have done this before. Though if they're temporary, I just don't commit those changes. Or sometimes I chuck them into my user config instead.

> But I found an alternative that should avoid both problems:

This seems clean enough to me though

agilgur5 (approved):
LGTM

PR #13675, opened 2024-09-29
docs(cli): add example for `argo archive delete`

leehosu:
Fixes #11898

### Motivation

Modify the argo archive delete command for more information

### Modifications

Added example for the argo archive delete command.

### Verification

run `argo archive delete -h`


agilgur5 (approved):
I modified the title of the PR as there is only a singular example and no options here. Please modify any similar PRs to match.

Otherwise LGTM, the comment does add more than the short description has

PR #13680, opened 2024-09-29
fix(cron): allow unresolved variables outside of `when`

isubasinghe:
This PR allows for unresolved variables to be present in cronworkflows during the evaluation of the `when` clause.
This is needed because workflow variables are not accessible/unresolved when the clause is evaluated.

### DIFF
--- a/workflow/cron/operator_test.go
+++ b/workflow/cron/operator_test.go
@@ -714,7 +714,7 @@ func TestEvaluateWhen(t *testing.T) {
 	assert.True(t, result)
 }
 
-func TestEvaluateWhenUnresolved(t *testing.T) {
+func TestEvaluateWhenUnresolvedOutside(t *testing.T) {
 	var cronWf v1alpha1.CronWorkflow
 	v1alpha1.MustUnmarshal([]byte(scheduledWf), &cronWf)
 	param := v1alpha1.Parameter{Name: "scheduled-time", Value: v1alpha1.AnyStringPtr("{{workflow.scheduledTime}}")}
--- a/workflow/cron/operator_test.go
+++ b/workflow/cron/operator_test.go
@@ -713,3 +713,18 @@ func TestEvaluateWhen(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, result)
 }
+
+func TestEvaluateWhenUnresolved(t *testing.T) {
+	var cronWf v1alpha1.CronWorkflow
+	v1alpha1.MustUnmarshal([]byte(scheduledWf), &cronWf)
+	param := v1alpha1.Parameter{Name: "scheduled-time", Value: v1alpha1.AnyStringPtr("{{workflow.scheduledTime}}")}
+	params := []v1alpha1.Parameter{param}
+	argument := v1alpha1.Arguments{Parameters: params}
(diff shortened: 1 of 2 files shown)


agilgur5 (commented):
> This is needed because workflow variables are not accessible/unresolved when the clause is evaluated.

Can you provide an example of this? It's not clear what you're referencing from such a short description.

There should also be a test for the same example

isubasinghe:
> > This is needed because workflow variables are not accessible/unresolved when the clause is evaluated.

I've provided an example in the test. The variables that a workflow may have, for example: {{workflow.name}} will not be available during the call to `evalWhen`, this means we must allow for unresolved variables when calling the templating engine.

agilgur5:
Oh you mean inside the `workflowSpec`? Can you limit the eval to just the `when` field instead?

isubasinghe:
> Oh you mean inside the `workflowSpec`? Can you limit the eval to just the `when` field instead?

Yup, unfortunately, this isn't really possible. Otherwise, that would have been my preferred solution.

agilgur5:
Can you elaborate why?

If I'm not mistaken, this would also allow invalid variables in `when`, which is a source of constant confusion in Argo templating.

isubasinghe:
> Can you elaborate why?
> 
> If I'm not mistaken, this would also end up allowing invalid variables in `when`, which is a source of constant confusion in Argo templating.

The current templating engine doesn't care, it looks for any string with the "{{" pattern and attempts to replace, and it does this at the workflow level. 

We absolutely need a refactor of how all this works.

agilgur5:
The functions used here, [`NewTemplate`](https://github.com/argoproj/argo-workflows/blob/e7883c4fe849020b3a0503dc6dd2a7a9c911c386/util/template/template.go#L21) and `Replace`, take an arbitrary string. So if I'm not mistaken, you can make that work on a single field without any refactoring of the templating logic

isubasinghe:
> The functions used here, [`NewTemplate`](https://github.com/argoproj/argo-workflows/blob/e7883c4fe849020b3a0503dc6dd2a7a9c911c386/util/template/template.go#L21) and `Replace`, take an arbitrary string. So if I'm not mistaken, you can make that work on a single field without any refactoring of the templating logic internals.


Yeah you are correct, I just wanted to keep the logic consistent across wherever we use templating but I suppose this is better.

isubasinghe:
@agilgur5 alright, just did that.

workflow/cron/operator_test.go
```diff
@@ -713,3 +713,18 @@ func TestEvaluateWhen(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, result)
 }
+
+func TestEvaluateWhenUnresolved(t *testing.T) {
```
agilgur5, line 717:
```suggestion
func TestEvaluateWhenUnresolvedOutside(t *testing.T) {
```

since this is testing "outside of `when`", might be good to clarify

agilgur5:
LGTM now, just one stylistic comment

workflow/cron/operator_test.go
isubasinghe: Done!

agilgur5 (approved):

PR #13702, opened 2024-10-04
fix!: consolidate cronworkflow variables

Joibel:
## Context

Issues #13474 and #12305 both introduced expressions into CronWorkflows as new features for 3.6.
There are some inconsistencies between the two, per https://github.com/argoproj/argo-workflows/pull/13474#issuecomment-2347998579 and https://github.com/argoproj/argo-workflows/pull/12305#issuecomment-2381625468

This PR was discussed at the [October 1st 2024 Contributor meeting](https://docs.google.com/document/d/1tznN5LB-KrNkC6dmeIzpKfuyvZWgCJpRuS-E84zseC8/edit#heading=h.gfymrsa8jhi7).

## Motivation

The expressions for `when` include the prefix `cronworkflow.`. Those for `stopStrategy.condition` do not.

Unify both of these under `cronworkflow.` and allow both sets of variables to be used in both expressions. This simplifies the documentation and is less surprising. It also allows some use cases where you could check for previous run failures and take decisions based on that and time of day - so perhaps run every hour in the `schedule` but then use `when` to only actually take action if the previous run this day failed.

`stopStrategy`'s `condition` is mostly not a field name used in workflows, so in agreement with @agilgur5 I've renamed it to `expression`.

## Modifications

Renamed `stopStrategy.condition` to `stopStrategy.expression`

Created a single function to generate all the variables for both expressions and use that instead of the two disparate mechanisms.

Modified some code comments to be godoc compliant.

## Verification

Existing tests still pass with appropriate amendments to the test cases

## Note

This is a breaking change vs 3.6-rc1 and rc2 for users of `stopStrategy` only. This is not considered properly breaking as those are release candidates and not full releases.


Joibel:
Summoning
- @eduardodbr, the effect of this change is mostly on your implementation, so I'd appreciate feedback
- @isubasinghe, this doesn't much affect `when`, but I'd also appreciate feedback
- @agilgur5, as you're the one who pointed this out and I'd like to know if it matches your expectations

agilgur5 (dismissed):
Could you remove the `@tag` of me in the commit message? I'm getting notified for every single commit right now and uh would prefer not to, including when commits are backported or forked 😅

docs/fields.md
```diff
@@ -1994,12 +1994,12 @@ SynchronizationStatus stores the status of semaphore and mutex.
...
 ## StopStrategy
 
-v3.6 and after: StopStrategy defines if the CronWorkflow should stop scheduling based on a condition
+StopStrategy defines if the CronWorkflow should stop scheduling based on an expression. v3.6 and after
 
 ### Fields
 | Field Name | Field Type | Description   |
 |:----------:|:----------:|---------------|
-|`condition`|`string`|v3.6 and after: Condition is an expression that stops scheduling workflows when true. Use the variables `failed` or `succeeded` to access the number of failed or successful child workflows.|
+|`expression`|`string`|Expression is an expression that stops scheduling workflows when true. Use the variables `io.argoproj.workflow.v1alpha1.failed` or `cronio.argoproj.REPLACEME.v1alpha1.succeeded` to access the number of failed or successful child workflows. v3.6 and after|
```
tooptoop4, line 2002: REPLACEME ?

tooptoop4: cronio?

tooptoop4: same typos in other files

agilgur5: Oh nice catch. This is actually being generated like this from `cron_workflow_types.go` 👀

isubasinghe (commented):
I do apologize but I think this:
`hack/api/swagger/swaggify.sh`
is getting in your way.

Joibel:
> I do apoligise but I think this: hack/api/swagger/swaggify.sh is getting in your way.

This is annoying. I'm not going to attempt to address this hack at the moment, so I've reworded it. Probably the correct fix is to understand the json and only replace in the two (I think) places in the structure which actually want it to happen, or at least not in `title` and `description` fields.

docs/fields.md
Joibel: @tooptoop4 - nice find - I've addressed it to stop the erroneous substitution for now.

docs/cron-workflows.md
```diff
@@ -120,14 +121,14 @@ For example, if you want to stop scheduling new workflows after one success:
```
agilgur5, line 121: Line 118 above needs to be changed now as well since you removed the `stopStrategy` section in the Fields Reference

docs/cron-workflows.md
```diff
@@ -51,7 +51,8 @@ You can use `CronWorkflow.spec.workflowMetadata` to add `labels` and `annotation
 | `startingDeadlineSeconds`    | `0`                    | Seconds after [the last scheduled time](#crash-recovery) during which a missed `Workflow` will still be run. |
 | `successfulJobsHistoryLimit` | `3`                    | Number of successful `Workflows` to persist |
 | `failedJobsHistoryLimit`     | `1`                    | Number of failed `Workflows` to persist |
-| `stopStrategy`               | `nil`                  | v3.6 and after: defines if the CronWorkflow should stop scheduling based on a condition |
+| `stopStrategy`               | `nil`                  | v3.6 and after: defines if the CronWorkflow should stop scheduling based on an expression |
+| `stopStrategy.expression`    | None                   | v3.6 and after: an expression which if present must evaluate to false for the workflow to be created |
```
agilgur5, line 55: I don't think it's necessary to explicitly explain both in the table as they are identical right now. You can instead link to the usage section below

```suggestion
| `stopStrategy`               | `nil`                  | v3.6 and after: defines if the CronWorkflow [should stop scheduling](#automatically-stopping-a-cronworkflow) based on an expression |
```

docs/cron-workflows.md
```diff
@@ -120,14 +121,14 @@ For example, if you want to stop scheduling new workflows after one success:
...
 
 You can also stop scheduling new workflows after three failures with:
 
 ```yaml
 stopStrategy:
-  condition: "failed >= 3"
+  expression: "cronworkflow.failed >= 3"
 ```
 
 <!-- markdownlint-disable MD046 -- this is indented due to the admonition, not a code block -->
```
agilgur5, line 134: the admonition below also needs modification to use the `cronworkflow.` prefix

docs/cron-workflows.md
```diff
@@ -39,7 +39,7 @@ In the above example it would be similar to `test-cron-wf-tj6fe`.
 
 You can use `CronWorkflow.spec.workflowMetadata` to add `labels` and `annotations`.
 
-### `CronWorkflow` Options
+### `CronWorkflow` Spec
```
agilgur5, line 42: I'm surprised this doesn't fail `markdownlint`, but this is the existing `h2` already. With this change, this section is now "## `CronWorkflow` Spec" -> "### `CronWorkflow` Spec"

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -69,11 +69,12 @@ type CronWorkflowSpec struct {
 	When string `json:"when,omitempty" protobuf:"bytes,12,opt,name=when"`
 }
 
-// v3.6 and after: StopStrategy defines if the CronWorkflow should stop scheduling based on a condition
+// StopStrategy defines if the CronWorkflow should stop scheduling based on an expression. v3.6 and after
```
agilgur5, line 72: The version annotation being consistently the first thing is intentional so that it is the first thing you see. This change makes this inconsistent with every other version annotation in the codebase, like lines 64-68 above.

I believe you made this change for GoDoc purposes, which does raise a conundrum since that apparently _requires_ it to start the same way? If it's just a convention, the consistency and visibility should superseded that.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
```diff
@@ -69,11 +69,12 @@ type CronWorkflowSpec struct {
...
 type StopStrategy struct {
-	// v3.6 and after: Condition is an expression that stops scheduling workflows when true. Use the
+	// Expression is an expression that stops scheduling workflows when true. Use the `cronworkflow`
 	// variables `failed` or `succeeded` to access the number of failed or successful child workflows.
```
agilgur5, line 75:
```suggestion
	// v3.6 and after: Expression is an expression that stops scheduling workflows when true. Use the
	// variables `cronworkflow`.`failed` or `cronworkflow`.`succeeded` to access the number of failed or successful child workflows.
```

I'd suggest a different workaround, leave the dot without backticks. That way it'll still copy+paste and read correctly, just have awkward formatting that is not that noticeable

agilgur5 (changes requested):
There are [some](https://github.com/argoproj/argo-workflows/pull/13702#discussion_r1790606359) [misses](https://github.com/argoproj/argo-workflows/pull/13702#discussion_r1790610091) here and a [semantic error](https://github.com/argoproj/argo-workflows/pull/13702#discussion_r1790612501) in the docs. Do keep in mind the surrounding context when making docs changes, as they do reference and build upon each other.
The version annotations are also intentionally at the beginning.
See in-line comments below

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
Joibel: Dirty hack, I like it.

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
Joibel: I did that to make it GoDoc compliant.

docs/cron-workflows.md
Joibel: Yes. I've reverted this but it's still a bad title.

docs/cron-workflows.md
Joibel: I've left in the .expression part as I feel it should be possible to use this table to understand what to do, and `expression` is unguessable.

docs/cron-workflows.md
Joibel: Good catch, fixed.

docs/cron-workflows.md
agilgur5: Calling it "Spec" wouldn't quite be correct either bc it doesn't include all the fields, such as the ones above and others. The closest thing might be to call it "Fields", as in to match the "Field Reference". Though I honestly feel this duplicates the existing section there that we could just link to instead: https://argo-workflows.readthedocs.io/en/latest/fields/#cronworkflowspec

isubasinghe (approved):
LGTM

docs/cron-workflows.md
```diff
@@ -51,7 +51,7 @@ You can use `CronWorkflow.spec.workflowMetadata` to add `labels` and `annotation
 | `startingDeadlineSeconds`    | `0`                    | Seconds after [the last scheduled time](#crash-recovery) during which a missed `Workflow` will still be run. |
 | `successfulJobsHistoryLimit` | `3`                    | Number of successful `Workflows` to persist |
 | `failedJobsHistoryLimit`     | `1`                    | Number of failed `Workflows` to persist |
-| `stopStrategy`               | `nil`                  | v3.6 and after: defines if the CronWorkflow should stop scheduling based on a condition |
+| `stopStrategy.expression`    | `nil`                  | v3.6 and after: defines if the CronWorkflow should stop scheduling based on an expression, which if present must evaluate to false for the workflow to be created |
```
agilgur5, line 54:
```suggestion
| `stopStrategy.expression`    | `nil`                  | v3.6 and after: defines if the CronWorkflow [should stop scheduling](#automatically-stopping-a-cronworkflow) based on an expression |
```

- As suggested before
  - and still following existing language and [k8s style guide on simplicity and directness](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

pkg/apis/workflow/v1alpha1/cron_workflow_types.go
agilgur5: That does not address my comment and, as I wrote above, does not match the entirety of rest of the code however...

docs/cron-workflows.md
agilgur5: That works, however the new description does not match what I wrote above, adding unnecessary and inconsistent verbosity, and missing a link. See follow-up thread

agilgur5 (commented):
> [Joibel](https://github.com/Joibel) dismissed [agilgur5](https://github.com/agilgur5)’s [stale review](https://github.com/argoproj/argo-workflows/pull/13702#pullrequestreview-2352533749) [5 days ago](https://github.com/argoproj/argo-workflows/pull/13702#event-14696214968):
Requested changes have been addressed

One of them was not addressed and the other was only partially addressed

PR #13710, opened 2024-10-05
fix: ignore pod started time when incorrectly set to epoch. Fixes #13709

### LINKED ISSUE #13709: Incorrect resources duration when `containerd` sets incorrect `startedAt`
### Pre-requisites

When running Argo Workflows we occasionally see that our workflows report _surprisingly high_ CPU and Memory resource durations.

image

We have traced this back to an issue with Containerd which will sometimes fails a Pod and for some reason sets the `startedAt` date to "the epoch" (ie: `1970-01-01T00:00:00Z`).

```json
...
"initContainerStatuses": [
{
  "containerID": "containerd://b55ff9eb8687235ae0fecd4216522b75b6f7a0fab5eca18de8c06365bd2636fc",
  "image": "quay.io/argoproject/argoexec:v3.5.11",
  "imageID": "quay.io/argoproject/argoexec@sha256:bb3938480cbe7a7c0f053eb77a5d3edb22c868bf57407bf066fd105961f26c72",
  "lastState": {},
  "name": "init",
  "ready": false,
  "restartCount": 0,
  "started": false,
  "state": {
	"terminated": {
	  "containerID": "containerd://b55ff9eb8687235ae0fecd4216522b75b6f7a0fab5eca18de8c06365bd2636fc",
	  "exitCode": 128,
	  "finishedAt": "2024-09-13T13:28:03Z",
	  "message": "failed to create containerd task: failed to create shim task: context deadline exceeded: unknown",
	  "reason": "StartError",
	  "startedAt": "1970-01-01T00:00:00Z" <---- here
	}
  }
}
],
...
```

This results in the incorrect calculation of a duration due to the function below which assumes that a startedAt data of 1970 is expected.

https://github.com/argoproj/argo-workflows/blob/25bbb71cced32b671f9ad35f0ffd1f0ddb8226ee/util/resource/summary.go#L16-L22

 I appreciate this _feels_ like more of a containerd bug that an argo-workflows bug, however I think it is worth adding some logic to handle this case more gracefully.

For additional context, we are running on top of managed GKE when we see this bug. I am not attaching a reproducible workflow because... well... this bug is challenging to reproduce!

I will propose a fix very soon and will let the maintainers decide whether it is worth patching. From our perspective it's worth it because we use the Argo Workflows reported CPU/Memory usage to bill downstream customers... 😨 

### Version(s)

v3.5.11

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

AntoineDao:
Fixes #13709

### Motivation

We have observed that containerd somtimes fails a pod on GKE and sets the `startedAt` value to "1970-01-01T00:00:00Z" which causes argo workflows to calculate invalid resources duration.

### Modifications

Added a check for resources duration function that will set the cpu and memory values to 0 if the container `startedAt` time being equal to "1970-01-01T00:00:00Z".

### Verification

I have added a unit test for the existing functionality (calculating resources durations) and a unit test for the added functionality (setting value to 0 if started at is 1970).


agilgur5 (commented):
See issue comments regarding this being an upstream `containerd` issue

AntoineDao:
See here for reasoning behind closing this PR -> https://github.com/argoproj/argo-workflows/issues/13709#issuecomment-2513801368

Issue #13056, opened 2024-05-16
Broken redirect to archived workflows


apiwoni:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

The following URL https://{{serverUrl}}/workflows/{namespace}/{workflowName} no longer redirects to  https://{{serverUrl}}/archived-workflows/{namespace}/{metadata.uid} starting with v3.4.14 when [fix(ui): missing uiUrl in ArchivedWorkflowsList #12172](https://github.com/argoproj/argo-workflows/pull/12172) has been committed.

Argo UI shows the following:

`
Error: Failed to execute 'pushState' on 'History': A history state object with URL 'https://archived-workflows/{{namepsace}}/87e895f1-50df-443b-971d-218beae9d25f' cannot be created in a document with origin 'https://xxx' and URL 'https://xxx/archived-workflows?namespace={namespace}&name=workflowName&deep=true'.
    at https://xxx/main.fa82dae05c4e68e1ec09.js:446:1047769
    at Object.confirmTransitionTo (https://xxx/main.fa82dae05c4e68e1ec09.js:446:1045618)
    at Object.push (https://xxx/main.fa82dae05c4e68e1ec09.js:446:1047680)
    at t.componentDidUpdate (https://xxx/main.fa82dae05c4e68e1ec09.js:433:843312)
    at is (https://xxx/main.fa82dae05c4e68e1ec09.js:398:83471)
    at hc (https://xxx/main.fa82dae05c4e68e1ec09.js:398:101246)
    at t.unstable_runWithPriority (https://xxx/main.fa82dae05c4e68e1ec09.js:406:3844)
    at Ho (https://xxx/main.fa82dae05c4e68e1ec09.js:398:45024)
    at pc (https://xxx/main.fa82dae05c4e68e1ec09.js:398:97718)
    at Qs (https://xxx/main.fa82dae05c4e68e1ec09.js:398:93872)
`

[...]

PR #13712, opened 2024-10-06
fix(ui): uiUrl in ArchivedWorkflowsList already has leading slash

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/13056


agilgur5:
Replaced by #13713

Issue #13601, opened 2024-09-14
3.5: Further optimize Archive List API call / DB query


agilgur5:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

This is a follow-up to https://github.com/argoproj/argo-workflows/issues/13295 and https://github.com/argoproj/argo-workflows/pull/13566. We can further improve the query performance, seemingly by another order of magnitude, by adding a union index:

> > ```sql
> > CREATE INDEX argo_archived_workflows_i5 ON argo_archived_workflows (clustername, startedat);
> > ```

This can be added as a migration after [this line in `migration.go`](https://github.com/argoproj/argo-workflows/blob/8065edb34df0e188fd88181b4660c705997e6dcf/persist/sqldb/migrate.go#L260).

We should also test the performance of this index with and without the subquery. If the subquery is no longer necessary, then we can revert #13566 along with this change

#### More context

Quoting myself in https://github.com/argoproj/argo-workflows/pull/13566#pullrequestreview-2301010281:
> Note that this query seems to be further optimizable per https://github.com/argoproj/argo-workflows/issues/13295#issuecomment-2340964248, so we may want to open a new issue to add further optimizations, but this at least fixes the performance issues and gets us to a much more reasonable baseline

Quoting @Danny5487401 from https://github.com/argoproj/argo-workflows/issues/13563#issuecomment-2339660938
> > @agilgur5 three methods i tested:
> > 
> > 1. nice implemention by sub,according to https://github.com/argoproj/argo-workflows/pull/13566
> > 
> > [...]
> >   
> > it drops from 37s 768ms to 1 s 981 ms
> > 
> > 2. force index (argo_archived_workflows_i4)
> > 
> > ```sql
> > SELECT
> >     [...]
> > FROM `argo_archived_workflows` force index (argo_archived_workflows_i4)
> > WHERE [...]
> > ORDER BY `startedat` DESC
> > LIMIT 20;
> > ```
> > 
> > it drops from 37s 768ms to 138 ms
[...]

PR #13715, opened 2024-10-06
test: simple DB CLI for local development

MasonM:
### Motivation

This adds a small CLI tool for developers to use when working with the DB locally. It provides two functions:
1. Explicitly migrate the DB (which is normally only done when the workflow controller starts) to help test migrations:

     ```
    $ go run ./hack/db migrate
    INFO[0000] Migrating database schema                     clusterName=default dbType=postgres
    INFO[0000] applying database change                      change="CREATE INDEX argo_archived_workflows_i5 ON argo_archived_workflows (clustername, startedat)" changeSchemaVersion=60
     ```
2. Insert randomly-generated archived workflows into the DB, which is intended to help test query performance (e.g.  https://github.com/argoproj/argo-workflows/issues/13601)

    ```
    $ time go run ./hack/db/main.go fake-archived-workflows --rows 100000
    Using seed 2600415997737113470
    Clusters: [cx5j6rrzm5vnkfqqr4d]
    Namespaces: [bs98pxsfhs9z v95848 hq76xbj4kq7vrdp49vm ghzj6vcrth 262wb8w2b8v2wd2p2p9]
    Inserted 100000 rows

    real    17m56.554s
    user    1m53.833s
    sys     0m43.581s
    ```

    This is obviously not as efficient as it could be, but it's only intended to be run on an ad-hoc basis when manually testing query performance. To make it fast, we'd probably have to switch to direct SQL inserts, which would couple this script to the DB schema logic in `persist/sqldb/workflow_archive.go`.

### Modifications
All logic is centralized in `hack/db/main.go`. I could've easily split that up, but that doesn't seem worth it, since this is only ever going to be used by developers.

### Verification

Manually ran `go run ./hack/db`. See above for the output when using PostgreSQL. Here's the corresponding output for MySQL:
```
$ go run ./hack/db migrate --dsn mysql:password@tcp/argo
INFO[0000] Migrating database schema                     clusterName=default dbType=mysql

$ time go run ./hack/db/main.go fake-archived-workflows --rows 100 --dsn mysql:password@tcp/argo
Using seed 5504222375807483325
Clusters: [jckpg]
Namespaces: [7qr6rmgz5d8hz8zskt4 zst56t4wv9pkh8 jxlrvhtx4 g7bjn crk8qhxdtnddr2f]
Inserted 100 rows

real    0m3.261s
user    0m1.605s
sys     0m0.500s
```


blkperl (approved):

Joibel (approved):
Thanks for putting this together

PR #13720, opened 2024-10-07
docs: add 3.6 json templating upgrading notes

### LINKED ISSUE #13541: 3.6: JSON templating change release notes & test cases
# Summary

#12909 made some fairly fundamental changes to JSON/struct usage in templates, so this needs to be documented in the release notes, per https://github.com/argoproj/argo-workflows/pull/12909#issuecomment-2285300935

There were also a few cases in particular that need testing for breaking changes.

## Use Cases

- [x] Release Notes
- [ ] Test cases
  - [x] #8930 -- `toJSON` + `jsonpath` on struct
  - [x] #12739 -- `toJSON` + `fromJSON` on struct
  - [ ] Other test cases on expressions on structs?
  - [ ] Other test cases for this change in general?

---

Joibel:
Fixes #13541
Follow up to add upgrading notes to #12909 

### Motivation

#12909 is a breaking change and wasn't listed in the upgrading notes

### Modifications

Add upgrading notes for this change. The change is subtle and probably affects few users. Some may have worked around it as per #8930 and can optionally remove `toJson` from expressions, but this is not necessary.

### Verification

`make docs` passes

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -5,6 +5,11 @@ the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summar
 
 ## Upgrading to v3.6
 
+### JSON templating fix
+
+When accessing a structure or array from json structure in a `jsonpath` expression you could end up getting a golang representation of the structure in error.
+This will now return the json representation as hoped - see #12909.
+
 ### Fixed Server `--basehref` inconsistency
 
 For consistency, the Server now uses `--base-href` and `ARGO_BASE_HREF`.
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -5,11 +5,6 @@ the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summar
 
 ## Upgrading to v3.6
 
-### JSON templating fix
-
-When accessing a structure or array from json structure in a `jsonpath` expression you could end up getting a golang representation of the structure in error.
-This will now return the json representation as hoped - see #12909.
-
 ### Fixed Server `--basehref` inconsistency
 
 For consistency, the Server now uses `--base-href` and `ARGO_BASE_HREF`.
@@ -83,6 +78,11 @@ Custom metrics, as defined by a workflow, could be defined as one type (say coun
 
 The Prometheus `/metrics` endpoint now has TLS enabled by default.
 To disable this set `metricsConfig.secure` to `false`.
+
+### JSON templating fix
+
+When returning a map or array in an expression, you would get a Golang representation.
+This now returns plain JSON.
 
 ## Upgrading to v3.5
 


docs/upgrading.md
```diff
@@ -5,6 +5,11 @@ the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summar
 
 ## Upgrading to v3.6
 
+### JSON templating fix
+
+When accessing a structure or array from json structure in a `jsonpath` expression you could end up getting a golang representation of the structure in error.
```
agilgur5, line 10: While the test cases in #12909 are `jsonpath` specific, the code is not, as far as I understand, otherwise it could have been applied to the `jsonpath` helper specifically.

The wording is also duplicative and inconsistent:
```suggestion
When accessing a map or array from JSON in a `jsonpath` expression, you would get a Golang representation.
```
- consistently use "JSON", not both "json" and "JSON" (esp in the same section)
  - "Golang" should similarly be capitalized as a proper noun
- simplify, per k8s style guide
  - "structure" is not needed (was previously used 3 times in one sentence) and is ambiguous, whereas "map" is a data structure consistent with "array"
  - "in error" is confusing/ambiguous as there is no error message. It's not needed either, we can just say what was returned before and what it returned now.

docs/upgrading.md
```diff
@@ -5,6 +5,11 @@ the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summar
...
+This will now return the json representation as hoped - see #12909.
```
agilgur5, line 11:
```suggestion
This now returns plain JSON.
```

- simplify, per k8s style guide

I'm also not sure we need to reference the PR and I'm not sure that referencing it would be helpful. They sometimes are and aren't referenced in the upgrading notes, but when they are it is part of the section heading and not in the contents. A header reference makes more sense to me as a way to correlate to the changelog items.

Also the PR number is not automatically linked

docs/upgrading.md
```diff
@@ -5,6 +5,11 @@ the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summar
 
 ## Upgrading to v3.6
 
+### JSON templating fix
```
agilgur5, line 8: I'd probably move this near the legacy `patch` `pod` removal if the thought is that it shouldn't have much impact. I.e. order of importance (then all the metrics changes as things after it could be missed accidentally since it's a large section)

Joibel: I've moved it to the bottom as least important.

docs/upgrading.md
Joibel: Done

docs/upgrading.md
Joibel: Done, and slightly simplified again

tico24 (approved):

isubasinghe (approved):
LGTM

PR #13723, opened 2024-10-07
docs(cron): add `when` to table + examples for DST leap handling

### LINKED ISSUE #13700: Add docs for CronWorkflows `when`
# Summary

The 3.6 feature `when` for CronWorkflows has no documentation of the actual when clause.

The new variables are documented in #13474 but the field is not.

Joibel:
### Motivation

This PR brings in the "once a day" documentation from #13365

Fixes #13700 whilst I'm there

### Modifications

Add `when` clause to cron workflows "options" list

Document skip forward and skip backward correctors for DST as previously documented in #13365

### Verification

`make docs` passes


agilgur5:
> This PR brings in the "once a day" documentation from #13365

As requested in https://github.com/argoproj/argo-workflows/pull/13474#pullrequestreview-2242246764

isubasinghe (approved):
LGTM

docs/cron-workflows.md
```diff
@@ -109,6 +110,41 @@ For example, with `timezone: America/Los_Angeles`:
...
+An example 01:30:00 schedule
+
+```yaml
+schedules:
+  - 30 1 * * *
+when: "{{= cronworkflow.lastScheduledTime == nil || (now() - cronworkflow.lastScheduledTime).Seconds() > 7200 }}"
+```
+
+This will schedule at the first 01:30 on a skip backwards change.
+The second will not run because of the `when` expression, which prevents this workflow running more often than once every 2 hours..
```
agilgur5, line 146:
```suggestion
The second will not run because of the `when` expression, which prevents this workflow running more often than once every 2 hours.
```
typo

docs/cron-workflows.md
```diff
@@ -109,6 +110,41 @@ For example, with `timezone: America/Los_Angeles`:
...
+  - 0 3 * * *
+when: "{{= cronworkflow.lastScheduledTime == nil || (now() - cronworkflow.lastScheduledTime).Seconds() > 3600 }}"
+```
+
+The 3:00 run of the schedule will not be scheduled every day of the year except on the day when the clock leaps forward over 2:30.
+The `when` expression prevents this workflow from running more than once an hour.
+In that case the 3:00 run will run, as 2:30 was skipped over.
+
+!!! Warning "Can run at 3:00"
+ If you create this CronWorkflow between the desired time and 3:00 it will run at 3:00 as it has never run before.
```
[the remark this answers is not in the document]
agilgur5: Since this was merged before I had a chance to review, can see what this syntax error currently looks like [live](https://argo-workflows.readthedocs.io/en/latest/cron-workflows/#skip-forward-missing-schedule):
Screenshot 2024-10-17 at 8 02 07 PM

docs/cron-workflows.md
```diff
@@ -109,6 +110,41 @@ For example, with `timezone: America/Los_Angeles`:
     |            | 2        | 2020-11-02 02:01:00 -0800 PST |
     |            | 3        | 2020-11-03 02:01:00 -0800 PST |
 
+#### Skip forward (missing schedule)
```
agilgur5, line 113: Per https://github.com/argoproj/argo-workflows/pull/13365#discussion_r1790563038, even if you left these split, I would still simplify these, per the style guide, and still use the word "leap" as that is the common term for this and not "skip" (which has other connotations)

agilgur5 (commented):
I didn't have a chance to review this but did have comments WIP, so I'll leave them here

PR #13724, opened 2024-10-07
chore(docs): fix artifact repository ref example name typo

Joibel:
This example doesn't have anything to do with artifactory.


agilgur5 (approved):
nice catch

PR #11295, opened 2023-07-05
fix(controller): Enable dummy metrics server on non-leader workflow controller

sakai-ast:
Fixes #10037

### Motivation

Our datadog agents on GKE collect prometheus metrics from multiple workflow-controllers through the k8s Service. However, the metrics server only starts on the leader workflow-controller, causing some datadog agents to frequently output "Connection refused" error logs.
Therefore, I am aiming to resolve this issue.

### Modifications

I have tried to implement the comment: https://github.com/argoproj/argo-workflows/issues/8283#issuecomment-1379988006

### Verification

I have added tests and verified the functionality locally.

PR #13731, opened 2024-10-09
fix(controller): handle race when starting metrics server. Fixes #10807

### LINKED ISSUE #10807: Flakey tests

MasonM:
Partial fix for #10807 
### Motivation
While investigating the flaky `MetricsSuite/TestMetricsEndpoint` test, I noticed `listen tcp :9090: bind: address already in use` errors in the controller logs ([example](https://github.com/argoproj/argo-workflows/actions/runs/11221357877/job/31191811077)):

```
controller: time="2024-10-07T18:22:14.793Z" level=info msg="Starting dummy metrics server at localhost:9090/metrics"
server: time="2024-10-07T18:22:14.793Z" level=info msg="Creating event controller" asyncDispatch=false operationQueueSize=16 workerCount=4
server: time="2024-10-07T18:22:14.800Z" level=info msg="GRPC Server Max Message Size, MaxGRPCMessageSize, is set" GRPC_MESSAGE_SIZE=104857600
server: time="2024-10-07T18:22:14.800Z" level=info msg="Argo Server started successfully on http://localhost:2746" url="http://localhost:2746"
controller: I1007 18:22:14.800947   25045 leaderelection.go:260] successfully acquired lease argo/workflow-controller
controller: time="2024-10-07T18:22:14.801Z" level=info msg="new leader" leader=local
controller: time="2024-10-07T18:22:14.801Z" level=info msg="Generating Self Signed TLS Certificates for Telemetry Servers"
controller: time="2024-10-07T18:22:14.802Z" level=info msg="Starting prometheus metrics server at localhost:9090/metrics"
controller: panic: listen tcp :9090: bind: address already in use
controller:
controller: goroutine 37 [running]:
controller: github.com/argoproj/argo-workflows/v3/util/telemetry.(*Metrics).RunPrometheusServer.func2()
controller: 	/home/runner/work/argo-workflows/argo-workflows/util/telemetry/exporter_prometheus.go:94 +0x16a
controller: created by github.com/argoproj/argo-workflows/v3/util/telemetry.(*Metrics).RunPrometheusServer in goroutine 36
controller: 	/home/runner/work/argo-workflows/argo-workflows/util/telemetry/exporter_prometheus.go:91 +0x53c
2024/10/07 18:22:14 controller: process exited 25045: exit status 2
controller: exit status 2
2024/10/07 18:22:14 controller: backing off 4s
```

The controller gracefully handles this by automatically restarting, but it definitely shouldn't be happening. I believe this is a race condition introduced in https://github.com/argoproj/argo-workflows/pull/11295. Here's the sequence of events that trigger this:
1. Controller starts
2. Dummy metrics server started on port 9090: https://github.com/argoproj/argo-workflows/blob/07703ab1e5e61f1735008bf79847af49f01af817/cmd/workflow-controller/main.go#L139
3. Leader election takes place and controller starts leading: https://github.com/argoproj/argo-workflows/blob/07703ab1e5e61f1735008bf79847af49f01af817/cmd/workflow-controller/main.go#L141-L151
4. Context for dummy metrics server cancelled: https://github.com/argoproj/argo-workflows/blob/07703ab1e5e61f1735008bf79847af49f01af817/cmd/workflow-controller/main.go#L152
5. Dummy metrics server shuts down: https://github.com/argoproj/argo-workflows/blob/07703ab1e5e61f1735008bf79847af49f01af817/util/telemetry/exporter_prometheus.go#L106-L112
6. Prometheus metrics server started on 9090: https://github.com/argoproj/argo-workflows/blob/07703ab1e5e61f1735008bf79847af49f01af817/cmd/workflow-controller/main.go#L154

The problems is steps 5-6 can happen out-of-order, because the shutdown happens after the context is cancelled. Per [the docs](https://pkg.go.dev/context#CancelFunc), "a CancelFunc does not wait for the work to stop".

### Modifications

The controller needs to explicitly wait for the dummy metrics server to shut down properly before starting the Prometheus metrics server. There's many ways of doing that, and this uses a `WaitGroup`, as that's the simplest approach I could think of.

### Verification
Ran tests locally

### DIFF
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -159,7 +159,11 @@ func NewRootCommand() *cobra.Command {
 							dummyCancel()
 							wg.Wait()
 							go wfController.Run(ctx, workflowWorkers, workflowTTLWorkers, podCleanupWorkers, cronWorkflowWorkers, workflowArchiveWorkers)
-							go wfController.RunPrometheusServer(ctx, false)
+							wg.Add(1)
+							go func() {
+								wfController.RunPrometheusServer(ctx, false)
+								wg.Done()
+							}()
 						},
 						OnStoppedLeading: func() {
 							log.WithField("id", nodeID).Info("stopped leading")
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -121,7 +121,7 @@ func NewRootCommand() *cobra.Command {
 				log.WithField("id", "single-instance").Info("starting leading")
 
 				go wfController.Run(ctx, workflowWorkers, workflowTTLWorkers, podCleanupWorkers, cronWorkflowWorkers, workflowArchiveWorkers)
-				go wfController.RunPrometheusServer(ctx, false)
+				wfController.RunPrometheusServer(ctx, false)
 			} else {
 				nodeID, ok := os.LookupEnv("LEADER_ELECTION_IDENTITY")
 				if !ok {
@@ -136,7 +136,7 @@ func NewRootCommand() *cobra.Command {
 				// for controlling the dummy metrics server
 				dummyCtx, dummyCancel := context.WithCancel(context.Background())
 				defer dummyCancel()
-				go wfController.RunPrometheusServer(dummyCtx, true)
+				wg := wfController.RunPrometheusServer(dummyCtx, true)
 
 				go leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
 					Lock: &resourcelock.LeaseLock{
@@ -150,12 +150,14 @@ func NewRootCommand() *cobra.Command {
 					Callbacks: leaderelection.LeaderCallbacks{
 						OnStartedLeading: func(ctx context.Context) {
 							dummyCancel()
+							wg.Wait()
 							go wfController.Run(ctx, workflowWorkers, workflowTTLWorkers, podCleanupWorkers, cronWorkflowWorkers, workflowArchiveWorkers)
 							go wfController.RunPrometheusServer(ctx, false)
 						},
 						OnStoppedLeading: func() {
 							log.WithField("id", nodeID).Info("stopped leading")
 							cancel()
+							wg.Wait()
 							go wfController.RunPrometheusServer(dummyCtx, true)
 						},
 						OnNewLeader: func(identity string) {
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/argoproj/pkg/cli"
@@ -121,7 +122,7 @@ func NewRootCommand() *cobra.Command {
 				log.WithField("id", "single-instance").Info("starting leading")
 
 				go wfController.Run(ctx, workflowWorkers, workflowTTLWorkers, podCleanupWorkers, cronWorkflowWorkers, workflowArchiveWorkers)
-				wfController.RunPrometheusServer(ctx, false)
+				go wfController.RunPrometheusServer(ctx, false)
 			} else {
 				nodeID, ok := os.LookupEnv("LEADER_ELECTION_IDENTITY")
 				if !ok {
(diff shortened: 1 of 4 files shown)


agilgur5:
> I believe this is a race condition introduced in #11295.

I guess some changes from https://github.com/argoproj/argo-workflows/pull/13265 or related made this happen more frequently? I was trying to figure out why it recently started 🤔

util/telemetry/exporter_prometheus.go
```diff
@@ -53,7 +54,8 @@ func (config *Config) port() int {
 
 // RunPrometheusServer starts a prometheus metrics server
 // If 'isDummy' is set to true, the dummy metrics server will be started. If it's false, the prometheus metrics server will be started
-func (m *Metrics) RunPrometheusServer(ctx context.Context, isDummy bool) {
+func (m *Metrics) RunPrometheusServer(ctx context.Context, isDummy bool, wg *sync.WaitGroup) {
+	defer wg.Done()
```
agilgur5, line 58: instead of adding `Done()` here, could you wrap the call of `RunPrometheusServer` with a `wg.Done()`? that would keep the encapsulation. See below comment for concrete suggestion

cmd/workflow-controller/main.go
```diff
@@ -136,7 +136,7 @@ func NewRootCommand() *cobra.Command {
 				// for controlling the dummy metrics server
 				dummyCtx, dummyCancel := context.WithCancel(context.Background())
 				defer dummyCancel()
-				go wfController.RunPrometheusServer(dummyCtx, true)
+				wg := wfController.RunPrometheusServer(dummyCtx, true)
```
agilgur5, line 139: This would be simpler than breaking encapsulation, no?
```suggestion
				var wg gosync.WaitGroup
				wg.Add(1)
				go func() {
					wfController.RunPrometheusServer(dummyCtx, true)
					wg.Done()
				}
```

workflow/controller/controller.go
```diff
@@ -379,8 +379,11 @@ func (wfc *WorkflowController) Run(ctx context.Context, wfWorkers, workflowTTLWo
 	<-ctx.Done()
 }
 
-func (wfc *WorkflowController) RunPrometheusServer(ctx context.Context, isDummy bool) {
-	go wfc.metrics.RunPrometheusServer(ctx, isDummy)
+func (wfc *WorkflowController) RunPrometheusServer(ctx context.Context, isDummy bool) *gosync.WaitGroup {
+	var wg gosync.WaitGroup
+	wg.Add(1)
+	go wfc.metrics.RunPrometheusServer(ctx, isDummy, &wg)
```
agilgur5, line 385: or make this synchronous instead?

agilgur5 (commented):
some ideas below

util/telemetry/exporter_prometheus.go
MasonM: You're right, that' a bit simpler. Done in 5dbe063d15a3dc6bf66ccd5151a55eff042c0db1

cmd/workflow-controller/main.go
MasonM: Yep, that looks nicer. Done in 5dbe063d15a3dc6bf66ccd5151a55eff042c0db1. Also, fixed a bug in `OnStoppedLeading` in e3f666c59c55f3ea6f1d50eeeaa4b8e89ebfa8bc

workflow/controller/controller.go
MasonM: Done in https://github.com/argoproj/argo-workflows/commit/5dbe063d15a3dc6bf66ccd5151a55eff042c0db1

agilgur5 (approved):
LGTM, thanks for root causing the test flake and adding all the details here!

I am still curious about why it started happening more frequently though if you uncovered that already. If not, then I wouldn't spend more time on it

Joibel:
Thank you for fixing this.

I am not sure why #13265 made this worse, but I suspect it did. Whatever is happening under the hood between otel and prometheus go library will be radically different.

I would probably like a fix where the waitgroup was encapsulated on the metrics side so the tests could test the same thing as the actual running code - a redesign of the interface to switch prometheus metrics on and off. This is good for now though, so thank you.

I was going to revisit shutdown of telemetry at some point because it becomes more fun when you're trying to ensure traces have all been dispatched before you terminate - so it might come out of that.

PR #5035, opened 2021-02-05
chore!: Remove deprecated fields

simster7:
Closes: https://github.com/argoproj/argo/issues/4997

Checklist:

* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo/blob/master/USERS.md).

PR #13733, opened 2024-10-10
docs: delete deprecated & removed immediate template docs

Joibel:
This documentation on a deprecated feature refers to code that was removed in f31e0c6f92ec5e383d2f32f57a822a518cbbef86 / #5035 in February 2021.

It is no longer adding value to keep around, so remove it.

### DIFF
--- a/docs/workflow-templates.md
+++ b/docs/workflow-templates.md
@@ -249,37 +249,6 @@ spec:
               value: "hello world"
 ```
 
-You should **never** reference another template directly on a `template` object (outside of a `steps` or `dag` template).
-This includes both using `template` and `templateRef`.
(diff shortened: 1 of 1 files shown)


agilgur5 (approved):
Thanks for tracking down the removal! I was wondering if this had been removed already and if it throws a validation error now. Do you know if it does the latter?

Joibel:
> Thanks for tracking down the removal! I was wondering if this had been removed already and if it throws a validation error now. Do you know if it does the latter?

The error is somewhat unhelpful in how you've messed up exactly, but yes, the validator rejects it: `template '<name>' type is unknown`.

PR #13734, opened 2024-10-10
fix: correct retry logic

### LINKED ISSUE #12543: Retrying specific failed node does not work

isubasinghe:
Fixes #12543 and other retry related issues.

### Motivation
The current retry logic is too simplistic, it relies only on resetting nodes by traversing boundary nodes.
This approach does not work, it needs to be a combination of manual traversal and boundary node traversal.

In addition to this the current logic does not take care of various edge cases, for example container sets.

It was clear some rethinking needed to be done such that:
a) edge cases were taken care of.
b) retry logic was simpler to understand and hopefully fix if any further bug were to arise.

### Modifications

This is a complete rewrite of the `FormulateRetryWorkflow` function.

### Verification

Extensive testing performed manually to verify that behaviour is as expected.

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -972,7 +972,7 @@ func consumeDAG(n *node, resetFunc resetFn) (*node, error) {
 }
 
 func consumePod(n *node, resetFunc resetFn, addToDelete deleteFn) (*node, error) {
-	// this sets to reset but resets are overriden by deletes in the final FormulateRetryWorkflow logic.
+	// this sets to reset but resets are overridden by deletes in the final FormulateRetryWorkflow logic.
 	curr, err := consumeTill(n, wfv1.NodeTypePod, resetFunc)
 	if err != nil {
 		return nil, err
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -940,40 +940,85 @@ func getChildren(n *node) map[string]bool {
 
 type resetFn func(string, bool)
 type deleteFn func(string, bool)
+type tillFn func(*node) (bool, bool)
 
-func consumeTill(n *node, nodeType wfv1.NodeType, resetFunc resetFn) (*node, error) {
+func getTillFnNodeType(nodeType wfv1.NodeType) tillFn {
+	return func(n *node) (bool, bool) {
+		return n.n.Type == nodeType, true
+	}
+}
+
+func consumeTill(n *node, should tillFn, resetFunc resetFn) (*node, error) {
 	curr := n
 	for {
 		if curr == nil {
-			return nil, fmt.Errorf("expected type %s but ran out of nodes to explore", nodeType)
+			return nil, fmt.Errorf("was seeking node but ran out of nodes to explore")
 		}
 
-		if curr.n.Type == nodeType {
-			resetFunc(curr.n.ID, true)
+		if foundNode, shouldReset := should(curr); foundNode {
+			if shouldReset {
+				resetFunc(curr.n.ID, true)
+			}
 			return curr, nil
 		}
 		curr = curr.parent
 	}
 }
 
+func getTillBoundaryFn(boundaryID string) tillFn {
+	return func(n *node) (bool, bool) {
+		return n.n.ID == boundaryID, n.n.BoundaryID != ""
+	}
+}
+
+func consumeBoundaries(n *node, resetFunc resetFn) (*node, error) {
+	curr := n
+	for {
+		if curr == nil {
+			return curr, nil
+		}
+		if curr.parent != nil && curr.parent.n.Type == wfv1.NodeTypeStepGroup {
+			resetFunc(curr.parent.n.ID, true)
+		}
+		seekingBoundaryID := curr.n.BoundaryID
+		if seekingBoundaryID == "" {
+			return curr.parent, nil
+		}
+		var err error
+		curr, err = consumeTill(curr, getTillBoundaryFn(seekingBoundaryID), resetFunc)
+		if err != nil {
+			return nil, err
+		}
+	}
+}
+
 func consumeStepGroup(n *node, resetFunc resetFn) (*node, error) {
-	return consumeTill(n, wfv1.NodeTypeStepGroup, resetFunc)
+	return consumeTill(n, getTillFnNodeType(wfv1.NodeTypeStepGroup), resetFunc)
 }
 
 func consumeSteps(n *node, resetFunc resetFn) (*node, error) {
-	return consumeTill(n, wfv1.NodeTypeSteps, resetFunc)
+	n, err := consumeTill(n, getTillFnNodeType(wfv1.NodeTypeSteps), resetFunc)
+	if err != nil {
+		return nil, err
+	}
+	return consumeBoundaries(n, resetFunc)
 }
+
 func consumeTaskGroup(n *node, resetFunc resetFn) (*node, error) {
-	return consumeTill(n, wfv1.NodeTypeTaskGroup, resetFunc)
+	return consumeTill(n, getTillFnNodeType(wfv1.NodeTypeTaskGroup), resetFunc)
 }
 
 func consumeDAG(n *node, resetFunc resetFn) (*node, error) {
-	return consumeTill(n, wfv1.NodeTypeDAG, resetFunc)
+	n, err := consumeTill(n, getTillFnNodeType(wfv1.NodeTypeDAG), resetFunc)
+	if err != nil {
+		return nil, err
+	}
+	return consumeBoundaries(n, resetFunc)
 }
 
 func consumePod(n *node, resetFunc resetFn, addToDelete deleteFn) (*node, error) {
 	// this sets to reset but resets are overridden by deletes in the final FormulateRetryWorkflow logic.
-	curr, err := consumeTill(n, wfv1.NodeTypePod, resetFunc)
+	curr, err := consumeTill(n, getTillFnNodeType(wfv1.NodeTypePod), resetFunc)
 	if err != nil {
 		return nil, err
 	}
@@ -1029,6 +1074,7 @@ func resetPath(allNodes []*node, toNode string) (map[string]bool, map[string]boo
 		mustFind = wfv1.NodeTypePod
 	}
 
+	findBoundaries := false
 	for {
 
 		if curr == nil {
@@ -1042,11 +1088,13 @@ func resetPath(allNodes []*node, toNode string) (map[string]bool, map[string]boo
 			//ignore
 		case wfv1.NodeTypeSteps:
 			addToReset(curr.n.ID, false)
+			findBoundaries = true
 		case wfv1.NodeTypeStepGroup:
 			addToReset(curr.n.ID, false)
 			mustFind = wfv1.NodeTypeSteps
 		case wfv1.NodeTypeDAG:
 			addToReset(curr.n.ID, false)
+			findBoundaries = true
 		case wfv1.NodeTypeTaskGroup:
 			addToReset(curr.n.ID, false)
 			mustFind = wfv1.NodeTypeDAG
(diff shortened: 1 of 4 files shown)


Joibel:
This might superceed #12105

workflow/util/util_test.go
```diff
@@ -2092,18 +2014,1794 @@ func TestRetryWorkflowWithNestedDAGsWithSuspendNodes(t *testing.T) {
...
+func TestNestedDAG(t *testing.T) {
+	require := require.New(t)
+	wf := wfv1.MustUnmarshalWorkflow(nestedDAG)
+
+	newWf, podsToDelete, err := FormulateRetryWorkflow(context.Background(), wf, true, "id=dag-nested-zxlc2-744943701", []string{})
+	require.NoError(err)
+	_ = newWf
+	_ = podsToDelete
+
+}
```
isubasinghe, line 3807: These should actually test for individual node status as well.

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+func consumePod(n *node, resetFunc resetFn, addToDelete deleteFn) (*node, error) {
+	// this sets to reset but resets are overridden by deletes in the final FormulateRetryWorkflow logic.
+	curr, err := consumeTill(n, getTillFnNodeType(wfv1.NodeTypePod), resetFunc)
+	if err != nil {
+		return nil, err
+	}
+	addToDelete(curr.n.ID, true)
+	children := getChildren(curr)
+	for childID := range children {
+		addToDelete(childID, true)
```
Joibel, line 1028: This recurses one child only, which is surprising... one child might be correct today but it might be a fun source of orphans in the future. This is not a sentence I expected to write.

Ah, I suspect that new line 1346 prevents us keeping children of deleted nodes anyway, so this whole child deletion loop could be removed instead.

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+	}
+
+	addToReset := func(nodeID string, addToNode bool) {
+		if nodeID == toNode && !addToNode {
+			return
+		}
+		nodesToReset[nodeID] = true
+	}
+
+	addToDelete := func(nodeID string, addToNode bool) {
```
Joibel, line 1062: `addToNode` is always `true`, so simplify

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+
+	if err != nil {
+		return nil, nil, err
+	}
+	l := len(nodes)
+	if l <= 0 {
+		return nodesToReset, nodesToDelete, nil
+	}
+
+	addToReset := func(nodeID string, addToNode bool) {
```
Joibel, line 1055: `addToNode` is always `false`, so simplify

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+	return res
+}
+func shouldRetryFailedType(nodeTyp wfv1.NodeType) bool {
+	if nodeTyp == wfv1.NodeTypePod || nodeTyp == wfv1.NodeTypeContainer {
+		return true
+	}
+	return false
+}
+
+// dagSortedNodes sorts the nodes based on the order they were created, omits onExitNode
```
Joibel, line 1171: Technically not in creation order as children are all appended batched without regard to order.

If `task-1` finishes later than `task-2` then task-1's children may  get created later than task-2's but that won't be reflected here.

workflow/util/util.go
```diff
@@ -853,178 +1218,152 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
...
+		toReset = setUnion(toReset, pathToReset)
+		toDelete = setUnion(toDelete, pathToDelete)
 	}
 
-	onExitNodeName := wf.ObjectMeta.Name + ".onExit"
-	// Get all children of nodes that match filter
-	nodeIDsToReset, err := getNodeIDsToReset(restartSuccessful, nodeFieldSelector, wf.Status.Nodes)
-	if err != nil {
-		return nil, nil, err
+	for k := range toReset {
```
Joibel, line 1288: Swap `k` for `nodeName`. Short variable names for array indices are OK, but here it's a map key and more confusing if not named for what it is.

workflow/util/util.go
```diff
@@ -853,178 +1218,152 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
...
-				deletedNodes[node.ID] = true
+				queue.Remove(currNode)
 			}
-			// do not add this status to the node. pretend as if this node never existed.
-		default:
-			// Do not allow retry of workflows with pods in Running/Pending phase
-			return nil, nil, errors.InternalErrorf("Workflow cannot be retried with node %s in %s phase", node.Name, node.Phase)
+		}
+		if n.Name == wf.Name && !shouldRetryFailedType(n.Type) {
+			newWf.Status.Nodes.Set(id, resetNode(*n.DeepCopy()))
```
Joibel, line 1335: Do we need this `DeepCopy`, it seems to me that `resetNode` will cause a copy?

workflow/util/util_test.go
```diff
@@ -2092,18 +2014,1794 @@ func TestRetryWorkflowWithNestedDAGsWithSuspendNodes(t *testing.T) {
...
+  startedAt: "2024-09-19T02:43:30Z"
+  taskResultsCompletionStatus:
+    steps-4k5vn-1044844302: true
+    steps-4k5vn-2627784879: true
+    steps-4k5vn-2644562498: true
+
+`
+
+func TestStepsRetryWorkflow(t *testing.T) {
+	assert := assert.New(t)
```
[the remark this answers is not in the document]
isubasinghe: Ah I was waiting for an opinion on https://github.com/argoproj/argo-workflows/pull/13734#discussion_r1808224838. I plan to use these when I fill in the tests. 

I was unsure if it was worth the effort of these tests. I wrote these to test edge conditions and make sure they worked as expected. They will prevent regressions though, so I will add them.

workflow/util/util_test.go
```diff
@@ -1026,51 +1025,6 @@ func TestRetryExitHandler(t *testing.T) {
 func TestFormulateRetryWorkflow(t *testing.T) {
 	ctx := context.Background()
 	wfClient := argofake.NewSimpleClientset().ArgoprojV1alpha1().Workflows("my-ns")
-	createdTime := metav1.Time{Time: time.Now().Add(-1 * time.Second).UTC()}
-	finishedTime := metav1.Time{Time: createdTime.Add(time.Second * 2)}
-	t.Run("Steps", func(t *testing.T) {
```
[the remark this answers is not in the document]
isubasinghe: I will check those now, the logic differed quite a lot. This sets nodes to Running I believe, whereas I delete them. 
This is needed to actually get a pod allocated, this why sometimes when  you hit retry, a workflow can be stuck on Running.

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+type parentMap map[string]*wfv1.NodeStatus
+
+type node struct {
+	n        *wfv1.NodeStatus
+	parent   *node
+	children []*node
+}
+
+func newWorkflowsDag(wf *wfv1.Workflow) ([]*node, error) {
+
```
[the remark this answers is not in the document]
isubasinghe: I love my blank lines :)

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+	}
+
+	slices.Reverse(reverseNodes)
+	return reverseNodes, nil
+}
+
+func getChildren(n *node) map[string]bool {
+	children := make(map[string]bool)
+	queue := list.New()
+	queue.PushBack(n)
```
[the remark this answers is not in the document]
isubasinghe: No real reason, I prefer non-recursive functions in non-functional languages, mostly owing to eliminating a source of stack overflows. 

I don't think we would ever stack overflow in practice here,.
I think Go's tail call optimisation isn't great as well so didn't want to rely upon that.

workflow/util/util.go
isubasinghe: Yeah its more, topological sorted order. Will change that.

workflow/util/util.go
isubasinghe: It will cause a copy but only a shallow one. This isn't great for nodes that will mutate over time. 
A reset node will mutate so it does need a deep copy.  There probably isn't pointers shared across multiple nodes, but its better to be safe than introduce some weird aliasing bug. 

I was actually missing one so this comment helped me catch it, so thanks.

workflow/util/util_test.go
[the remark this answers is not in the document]
isubasinghe: It's not really testable easily, we cannot establish a dag for these status fields.  The root node is not present. 
This will error out complaining that the root node couldn't be found.

workflow/util/util.go
isubasinghe: > This recurses one child only 

I don't think so ? We get all children of curr. The getChildren does a BFS to obtain all the children 

> Ah, I suspect that new line 1346 prevents us keeping children of deleted nodes anyway, so this whole child deletion loop could be removed instead.

Yes but we shouldn't rely on that kind of inference in my opinion. For all other executable node types, we need to track the deletion id so that we can delete the pods. We shouldn't diverge from this logic just because it happens not  to be needed for containersets.

workflow/util/util.go
isubasinghe: It is actually safe to get rid  of the `addToNode` arg here, the reason being the deletion overrides the resetting. 
Maybe a comment  explaining so and removing this arg is best.

workflow/util/util.go
```diff
@@ -799,50 +788,426 @@ func deletePodNodeDuringRetryWorkflow(wf *wfv1.Workflow, node wfv1.NodeStatus, d
...
+			if shouldReset {
+				resetFunc(curr.n.ID, true)
+			}
+			return curr, nil
+		}
+		curr = curr.parent
+	}
+}
+
+func getTillBoundaryFn(boundaryID string) tillFn {
```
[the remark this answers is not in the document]
isubasinghe: done!

workflow/util/util.go
```diff
@@ -880,151 +858,437 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
...
-				log.Debugf("Deleted %s node %s since it's not a group node", node.Name, string(node.Phase))
-				deletedPods, podsToDelete = deletePodNodeDuringRetryWorkflow(wf, node, deletedPods, podsToDelete)
-				log.Debugf("Deleted pod node: %s", node.Name)
-				deletedNodes[node.ID] = true
-			}
-			// do not add this status to the node. pretend as if this node never existed.
-		default:
-			// Do not allow retry of workflows with pods in Running/Pending phase
-			return nil, nil, errors.InternalErrorf("Workflow cannot be retried with node %s in %s phase", node.Name, node.Phase)
+	deleteNodes, err := getNodeIDsToResetNoChildren(restartSuccessful, nodeFieldSelector, wf.Status.Nodes)
```
gzbvas: Would I have to set the --restart-successful in order to identify the nodes that need to be deleted, even if I only wish to retry a specific failed node?

isubasinghe: No this wouldn't be the case

workflow/util/util.go
Joibel: Ok, I follow and agree now. Thanks.

Joibel (approved):
LGTM.

shuangkun:
Could this PR be cherry-picked into the 3.6.5 branch? Some customers have inquired about it, but I haven't found any prior discussions related to the 3.6.x patches. Thanks @Joibel

Joibel:
I've created a patch releases discussion #14259.

I haven't backported this yet, it feels like #14124 should be addressed before this is backported otherwise we're knowingly introducing a regression.

PR #13735, opened 2024-10-10
feat: deprecations metric

Joibel:
A metric to measure the use of deprecated features in the code base

### Motivation

We have a number of deprecated features.

As a user it will be hard to know if it is safe to upgrade beyond the point when the deprecated feature becomes removed.

In preparation for removing the features this metric can be monitored to ensure you're good to upgrade.

### Modifications

There is a deliberate design decision to ensure the counting was simple. It would be worse for us to have a bug in the counting code which meant we didn't flag up a use when there was one. So each deprecated field is flagged at point of use by the code base.

This has problems:
* We may not care in this case - for example the validator called from the CLI doesn't have telemetry, so this doesn't work here.
* We will look at some fields multiple times in a single run of the reconciliation loop, so the number will go up faster than once per workflow run. I considered that attempting to get 'once per' behaviour would introduce more bugs, and this provides the necessary information

There are two new modules:

#### context

This is a general purpose module for reading and writing object metadata into [context](https://pkg.go.dev/context) which is the canonical way of passing data around. This avoids us having to expose too much information in the wrong places.

#### deprecation

This is a separate module with no imports except `context` above which can be called from anywhere without anything other than a context. The context does not need to contain the desired information.

It holds an evil global variable specifying what to do when a deprecation is flagged. When this has been initialised by the controller to point to the metrics function which records deprecations we get the metric number going up. Otherwise it's silent.

If the namespace is available in `context` we report it. This will be the namespace of the running workflow or cronworkflow.

#### Overall

This means this modifies a lot of code paths to pass context, but I consider this a reasonable thing that we should want in general anyway.

### NOT Implemented here

The validator does not flag up deprecated features. It should so I've raised #13736 

We could log these too, including `name` which might help in hunting down stragglers. (Some of these come from WorkflowTemplates, so it's not immediately going to point the finger).

### Verification

Unit tests have been added. e2e test for each deprecation added.


util/context/context.go
```diff
@@ -0,0 +1,35 @@
...
+	ctx = context.WithValue(ctx, name, meta.Name)
+	return context.WithValue(ctx, namespace, meta.Namespace)
+}
+
+func ObjectName(ctx context.Context) string {
+	if n, ok := ctx.Value(name).(string); ok {
+		return n
+	}
+	return ""
+}
```
isubasinghe, line 28: I'd personally prefer returning a boolean as well.
Exactly like how maps work

isubasinghe, line 28: I'd personally prefer returning a boolean as well.
Exactly like how maps work.

I know this is guaranteed to have this key though. 

But we do check the ok in the cast, continuing on with this lack of assuming things, we should also return an `ok` bool.

util/context/context.go
```diff
@@ -0,0 +1,35 @@
...
+
+func ObjectNamespace(ctx context.Context) string {
+	if n, ok := ctx.Value(namespace).(string); ok {
+		return n
+	}
```
isubasinghe, line 33: same here

isubasinghe (approved):
Minor comment, will leave up to you to decide.

util/context/context.go
Joibel: When someone needs it we can add that, right now it wouldn't be used, so not adding in this commit.

mkdocs.yml
```diff
@@ -248,6 +248,7 @@ nav:
           - offloading-large-workflows.md
           - workflow-archive.md
           - metrics.md
+          - deprecations.md
```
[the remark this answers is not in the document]
agilgur5: > Although arguably deprecations should be part of the upgrading guide.

I see this has been done in https://github.com/argoproj/argo-workflows/pull/13757 now

docs/deprecations.md
agilgur5, line 1: Similarly I would put all the H2 headings here a under a 3.6 version heading so that the upgrading guide could reference it per version. With no versions on this page, it's hard to understand when something was deprecated, and some of these also reference features that are only available in newer versions

As we have already had users attempt to use 3.5 and 3.6 features in older versions when there were no version annotations (https://github.com/argoproj/argo-workflows/pull/13645 as one example), I could definitely see users reading this, seeing the deprecation, trying the new syntax, and then getting a syntax error and raising an issue.

docs/deprecations.md
```diff
@@ -0,0 +1,75 @@
+# Deprecations
+
+Sometimes a feature of Argo Workflows is deprecated.
+You should stop using a deprecated feature as it may be removed in a future minor or major release of Argo Workflows.
+
+To determine if you are using a deprecated feature the [`deprecated_feature`](metrics.md#deprecated_feature) metric can help.
```
agilgur5, line 6: It would definitely be good to have a version annotation for this metric too, as people will see this page, try to use it, and not realize that it's not available in older versions, and perhaps even think they're in the clear because of that.

docs/deprecations.md
```diff
@@ -0,0 +1,75 @@
...
+This metric will go up for each use of a deprecated feature by the workflow controller.
+This means it may go up once or many times for a single event.
+If the number is going up the feature is still in use by your system.
+If the metric is not present or no longer increasing are no longer using the monitored deprecated features.
+
+## `cronworkflow schedule`
```
agilgur5, line 12: I would name these for their respective features so that they are easily correlated, and then perhaps have a "> `cronworkflow schedule` attribute" to indicate that this is what the metric displays. I think it's difficult to tell that this is referring to those specific names unless you have specifically came from the metric's page and read it in detail.

This page isn't specific to the metric though, it's just named "Deprecations" in general  as well, so it can reference it but should be its own independent page as well

docs/deprecations.md
```diff
@@ -0,0 +1,75 @@
...
+spec:
+  schedule: "30 1 * * *"
+```
+
+is replaced with
+
+```yaml
+spec:
+  schedules:
+    - "30 1 * * *"
```
agilgur5, line 27: These examples would likely be easier to understand quickly with a ```diff, i.e.:

```diff
-   schedule: "30 1 * * *"
+   schedules:
+     - "30 1 * * *"
```

docs/deprecations.md
```diff
@@ -0,0 +1,75 @@
...
+To determine if you are using a deprecated feature the [`deprecated_feature`](metrics.md#deprecated_feature) metric can help.
+This metric will go up for each use of a deprecated feature by the workflow controller.
+This means it may go up once or many times for a single event.
+If the number is going up the feature is still in use by your system.
+If the metric is not present or no longer increasing are no longer using the monitored deprecated features.
+
+## `cronworkflow schedule`
+
+The spec field `schedule` which takes a single value is replaced by `schedules` which takes a list.
+To update this replace the `schedule` with `schedules` as in the following example
```
agilgur5, line 15:
```suggestion
To update this replace the `schedule` with `schedules` as in the following example:
```

Colons are also missing in prior to all the examples, which are conventionally (and gramatically) used when describing an example in the docs

docs/deprecations.md
agilgur5: It would also be good to link to the feature page for each of these which will have more details and examples

agilgur5 (commented):
Had some WIP comments here too

docs/deprecations.md
```diff
@@ -0,0 +1,75 @@
...
+  semaphores:
+    - configMapKeyRef:
+        name: my-config
+        key: workflow
+```
+
+## `workflow podpriority`
+
+The Workflow spec field `podPriority` which takes a numeric value is deprecated and `podPriorityClassName` should be used instead.
+To update this you will need a [PriorityClass](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass) in your cluster and refer to that using `podPriorityClassName`.
```
agilgur5, line 75: unrelated, can use backticks here to avoid the `.spelling` addition
```suggestion
To update this you will need a [`PriorityClass`](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass) in your cluster and refer to that using `podPriorityClassName`.
```

docs/metrics.md
```diff
@@ -265,6 +265,23 @@ Suppressed runs due to `concurrencyPolicy: Forbid` will not be counted.
...
+|-------------|---------------------------------------------|
+| `feature`   | The name of the feature used                |
+| `namespace` | The namespace of the item using the feature |
+
+`feature` will be one of:
+
+- [`cronworkflow schedule`](deprecations.md#cronworkflow_schedule)
+- [`synchronization mutex`](deprecations.md#synchronization_mutex)
+- [`synchronization semaphore`](deprecations.md#synchronization_semaphore)
+- [`workflow podpriority`](deprecations.md#workflow_podpriority)
```
agilgur5, line 283:
```suggestion
- [`cronworkflow schedule`](deprecations.md#cronworkflow-schedule)
- [`synchronization mutex`](deprecations.md#synchronization-mutex)
- [`synchronization semaphore`](deprecations.md#synchronization-semaphore)
- [`workflow podpriority`](deprecations.md#workflow-podpriority)
```

- c.f. https://argo-workflows.readthedocs.io/en/release-3.6/deprecations/#cronworkflow-schedule
- hyphens are used, not underscores, same as https://github.com/argoproj/argo-workflows/pull/13739#discussion_r1815494815

agilgur5 (commented):
Similar to https://github.com/argoproj/argo-workflows/pull/13739#pullrequestreview-2393377242, `mkdocs build` is giving some warnings about links from this PR:

```
INFO    -  Doc file 'metrics.md' contains a link 'deprecations.md#cronworkflow_schedule', but the doc 'deprecations.md' does not contain an anchor '#cronworkflow_schedule'.
INFO    -  Doc file 'metrics.md' contains a link 'deprecations.md#synchronization_mutex', but the doc 'deprecations.md' does not contain an anchor '#synchronization_mutex'.
INFO    -  Doc file 'metrics.md' contains a link 'deprecations.md#synchronization_semaphore', but the doc 'deprecations.md' does not contain an anchor '#synchronization_semaphore'.
```

PR #13741, opened 2024-10-11
feat: don't add configmap volumemounts in wait container. Partial fix for #13089

tooptoop4:
Fixes point 2 of https://github.com/argoproj/argo-workflows/issues/13089

### Motivation
Reduce etcd usage

### Modifications
Imagine your workflow mounts 10 different .py related configmaps. Wait container has no use for these configmaps. Init container never had them, only main container needs them

### Verification
Been running successfully in my org for weeks across many different types of workflows

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -1080,6 +1080,16 @@ func addOutputArtifactsVolumes(pod *apiv1.Pod, tmpl *wfv1.Template) {
 			if util.IsWindowsUNCPath(mnt.MountPath, tmpl) {
 				continue
 			}
+			isConfigMapVolume := false
+			for _, vol := range pod.Spec.Volumes {
+				if vol.Name == mnt.Name && vol.ConfigMap != nil {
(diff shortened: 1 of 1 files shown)


jswxstw:
Do you mean the ConfigMap named `argo-env-config`?

tooptoop4:
> Do you mean the ConfigMap named `argo-env-config`?
> 
> https://github.com/argoproj/argo-workflows/blob/5be8cbfe612ef49cdccb86068df79098d2198b8c/workflow/controller/workflowpod.go#L488-L497

@jswxstw i mean to exclude user configmaps like somecode.py they choose to mount, of which there could be several different ones mounted

jswxstw:
> @jswxstw i mean to exclude user configmaps like somecode.py they choose to mount, of which there could be several different ones mounted

You should write the PR description more clearly, otherwise we won't know why and how you actually do.

tooptoop4:
who gon merge dis

tooptoop4:
can u merge dis @tczhao

PR #13743, opened 2024-10-11
fix: only set `ARGO_PROGRESS_FILE` when needed. Partial fix for #13089

tooptoop4:
Fixes point 5 of https://github.com/argoproj/argo-workflows/issues/13089

`ARGO_PROGRESS_FILE` is only used when self reporting progress

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -304,13 +304,16 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 		{Name: common.EnvVarNodeID, Value: nodeID},
 		{Name: common.EnvVarIncludeScriptOutput, Value: strconv.FormatBool(opts.includeScriptOutput)},
 		{Name: common.EnvVarDeadline, Value: woc.getDeadline(opts).Format(time.RFC3339)},
-		{Name: common.EnvVarProgressFile, Value: common.ArgoProgressPath},
 	}
 
-	// only set tick durations if progress is enabled. The EnvVarProgressFile is always set (user convenience) but the
-	// progress is only monitored if the tick durations are >0.
+	// only set tick durations/EnvVarProgressFile if progress is enabled.
+	// The progress is only monitored if the tick durations are >0.
 	if woc.controller.progressPatchTickDuration != 0 && woc.controller.progressFileTickDuration != 0 {
 		envVars = append(envVars,
+			apiv1.EnvVar{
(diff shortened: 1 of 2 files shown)


workflow/controller/workflowpod.go
```diff
@@ -304,13 +304,16 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
...
-	// only set tick durations if progress is enabled. The EnvVarProgressFile is always set (user convenience) but the
-	// progress is only monitored if the tick durations are >0.
+	// only set tick durations/EnvVarProgressFile if progress is enabled.
+	// The progress is only monitored if the tick durations are >0.
 	if woc.controller.progressPatchTickDuration != 0 && woc.controller.progressFileTickDuration != 0 {
 		envVars = append(envVars,
+			apiv1.EnvVar{
+				Name:  common.EnvVarProgressFile,
+				Value: common.ArgoProgressPath,
+			},
```
juliev0, line 316: I guess this change is okay. Judging from the fact that it does look like wait container will only [write](https://github.com/argoproj/argo-workflows/blob/v3.5.11/workflow/executor/executor.go#L1125-L1126) to Progress File defined by this env var if `we.annotationPatchTickDuration != 0 && we.readProgressFileTickDuration != 0`, I guess it won't break anything. 

Only thing is that this change will only reduce the Pod size a tiny bit. But that doesn't mean we can't merge it I suppose.

I saw you mentioned in the original [issue](https://github.com/argoproj/argo-workflows/issues/13089) the idea of limiting certain Environment variables to certain containers. It does seem like maybe only "wait" container needs these?

@jswxstw any thoughts?

jswxstw: > It does seem like maybe only "wait" container needs these?

"main" container also need `EnvVarProgressFile` to determine where to write the progress, but `init` container do not need it.

In my opinion, there are currently some redundant environment variables, but most of them are quite small, except for `ARGO_TEMPLATE` which I think can be optimized like https://github.com/argoproj/argo-workflows/issues/8790#issuecomment-2381964013.

juliev0: Right. Were you interested in specifying container-specific environment variables for this PR @tooptoop4 or were you wanting to keep the change for this small?

tooptoop4: wanted to keep this PR small

juliev0 (approved):

PR #13746, opened 2024-10-11
feat: add config to skip sending workflow events. Fixes #13042

### LINKED ISSUE #13042: Option to disable writing k8s events at workflow level

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/13042

### DIFF
--- a/config/workflow_events.go
+++ b/config/workflow_events.go
@@ -1,0 +1,9 @@
+package config
+
+type WorkflowEvents struct {
+	Enabled   *bool `json:"enabled,omitempty"`
+}
+
+func (e WorkflowEvents) IsEnabled() bool {
+	return e.Enabled == nil || *e.Enabled
+}
--- a/config/workflow_events.go
+++ b/config/workflow_events.go
@@ -1,7 +1,7 @@
 package config
 
 type WorkflowEvents struct {
-	Enabled   *bool `json:"enabled,omitempty"`
+	Enabled *bool `json:"enabled,omitempty"`
 }
 
 func (e WorkflowEvents) IsEnabled() bool {
--- a/docs/workflow-controller-configmap.yaml
+++ b/docs/workflow-controller-configmap.yaml
@@ -41,6 +41,13 @@ data:
   nodeEvents: |
     enabled: true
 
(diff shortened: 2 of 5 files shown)


docs/workflow-controller-configmap.yaml
```diff
@@ -41,6 +41,13 @@ data:
   nodeEvents: |
     enabled: true
 
+  # Whether or not to emit events on workflow completion. These can take a up a lot of space in
+  # k8s (typically etcd) resulting in errors when trying to create new events:
+  # "Unable to create audit event: etcdserver: mvcc: database space exceeded"
+  # This config item allows you to disable this.
+  workflowEvents: |
```
agilgur5, line 48: This needs a version reference like `nodeEvents` and other configs have

docs/workflow-controller-configmap.yaml
```diff
@@ -41,6 +41,13 @@ data:
   nodeEvents: |
     enabled: true
 
+  # Whether or not to emit events on workflow completion. These can take a up a lot of space in
```
agilgur5, line 44: It seems to not be on completion per the code you modified, but actually on any status change, if I'm not mistaken.

Also we can shorten / replace the following sentence by saying "See nodeEvents above"

config/workflow_events.go
agilgur5, line 1: If you're copying `node_events.go`, should also copy the unit tests of `node_events_test.go`

agilgur5 (approved):
LGTM, thanks for adding this

PR #13748, opened 2024-10-13
build: fix commit-msg Git hook

MasonM:
### Motivation

The Git hook that verifies the commit messages consistently fails in both the DevContainer and my Ubuntu 24 desktop if the commit title has a scope:
```
$ printf 'feat(sso): test\n\nSigned-off-by: test' > msg && ./hack/git/hooks/commit-msg msg
grep: warning: ? at start of expression
grep: warning: ? at start of expression
❌ Commit message must be semantic: https://github.com/zeke/semantic-pull-requests

$ grep --version
grep (GNU grep) 3.11
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others; see
<https://git.savannah.gnu.org/cgit/grep.git/tree/AUTHORS>.

grep -P uses PCRE2 10.42 2022-12-11
```

This is because `(?:` isn't valid extended regular expression syntax.

### Modifications

 Since there aren't any backreferences in this regexp, there's no need for a non-capturing group, so I replaced both instances of `(?:` with `(`.

### Verification
```
$ printf 'feat(sso): test\n\nSigned-off-by: test' > msg && ./hack/git/hooks/commit-msg msg
✅ Commit looks good
```

### DIFF
--- a/hack/git/hooks/commit-msg
+++ b/hack/git/hooks/commit-msg
@@ -6,7 +6,7 @@ grep -q 'Signed-off-by: ' "$1" || {
   exit 1
 }
 
-grep -qE '^(?:build|feat|fix|docs|style|refactor|perf|test|ci|chore|revert)\(?(?:\w+|\s|\-|_)?\)?!?:\s\w+' "$1" || grep -q 'Merge' "$1" || {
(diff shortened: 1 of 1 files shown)


isubasinghe:
This works, thanks for fixing this.

isubasinghe (approved):
LGTM

Issue #8626, opened 2022-05-05
Support Expression templating for the workflow Links


loristo:
# Summary

The templating system for the Links could be improved in a few ways to support:
- keys containing special characters
- usage of functions

The templating system for the links would work the same way as the Expression templating system for the Workflows (https://argoproj.github.io/argo-workflows/variables/#expression).

# Use Cases

Usage of keys containing special characters would be useful if I want to use a workflow metadata like `workflows.argoproj.io/workflow-template` in a Link.

The usage of functions would allow the modification of the fields. For example, I would like to add a few seconds to `status.finishedAtEpoch` to take into account the Promtail scrap delay. This could also be useful for formatting, trimming, etc.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
Note that there is a bit of a problem with supporting this in the API because `links` are retrieved from the `info` endpoint and therefore don't have access to the particular page's context in order to be parametrized with variables

PR #13752, opened 2024-10-14
fix(ui): allow `links` to metadata with dots. Fixes #11741

### LINKED ISSUE #11741: UI: `links` for label containing dot not working

Joibel:
Fixes #11741

### Motivation

You cannot create UI links to our own annotations or labels.

Argo Workflows uses the format `workflows.argoproj.io/something` for annotations and labels. These annotations and labels have dots (`.`) in them.

The [links](https://argo-workflows.readthedocs.io/en/stable/links/) documentation says you can access all workflow fields, using a dot to go down into the child entry.

These two conflict making it impossible to access our own labels and annotations, so fix it for annotations and labels.

This was discovered whilst adding an annotation to the workflows that was useful to link from in a separate feature.

### Modifications

When splitting the variable name up use a specialised splitting function that understands that things under `workflow.metadata.labels` and `workflow.metadata.annotations` are always names and not to split further. It is crude

### Verification

Added a unit test for the hit the special code case. There are already unit tests which test the other case.

Manually tested my new link works.

### DIFF
--- a/ui/src/app/shared/components/links.test.ts
+++ b/ui/src/app/shared/components/links.test.ts
@@ -56,4 +56,19 @@ describe('process URL', () => {
 
         expect(processURL('https://logging?${workflow.annotations.logQuery}${workflow.annotations.additionalLogParams}', object)).toBe('https://logging?query=env:qa');
     });
+
+    test('workflows annotation.', () => {
+        const object = {
+            status: {},
+            workflow: {
+                metadata: {
+                    annotations: {
+                        'workflows.argoproj.io/pod-name-format': 'v2'
+                    }
+                }
+            }
+        };
+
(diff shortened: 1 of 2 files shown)


agilgur5:
For reference, `links` are also an outlier in using their own custom, simplified templating system and with their own syntax; see also: https://github.com/argoproj/argo-workflows/issues/8626

Issue #12166, opened 2023-11-07
`WorkflowEventBinding` CRD has no documentation


metasim:
# Summary

It is customary for K8S CRDs to be self documenting, enabling one to run `kubectl explain <CRD path>` (or use your IDE's features to do so) to assist in defining K8S resources. This is missing from `WorkflowEventBinding`, making it unnecessarily burdensome to figure out proper usage.

Here's the current output when running `kubectl explain WorkflowEventBinding.spec`.

```text
GROUP:      argoproj.io
KIND:       WorkflowEventBinding
VERSION:    v1alpha1

FIELD: spec <Object>

DESCRIPTION:
    <empty>

```

Note: This may apply to other Argo CRDs; this is just the one I'm attempting to use right now.

## Use Cases

When implementing Argo event bindings.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

metasim:
`WorkflowEventBinding` also appears to be missing from https://argoproj.github.io/argo-workflows/fields/

Issue #8190, opened 2022-03-18
Add descriptions to `kubectl explain`


kevin-hanselman:
# Summary

`kubectl explain` is an invaluable tool for quickly exploring Kubernetes resources. CRDs can provide descriptions that `kubectl explain` can pick up. For example, see [Knative](https://github.com/knative/eventing/issues/3054).

# Use Cases 

Instead of going to the Argo Workflows webpage, I'd like to run, for example, `kubectl explain workflow` for guidance on building a Workflow resource.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

kevin-hanselman:
> This is like `man` for CRDs?

@alexec it's like `man` for all Kubernetes resources -- built-ins and CRDs alike.

danielhelfand:
This may be a bit challenging for resources like Workflows, which will require a schema for this feature to work. Due to the size of the CRDs with the schema, this will cause errors like below when deploying the CRD with descriptions: 

```
for: "manifests/base/crds/full/argoproj.io_workflows.yaml": Request entity too large: limit is 3145728
```

It seems schemas were intentionally left out due to this issue. 

However, it still may be possible for smaller CRDs to support kubectl explain.

drew-viles:
Even if a description can't be put in, it would be nice to know what lies under `spec`.

For example, if something is a `map[string]string` or a `bool` etc. Having nothing isn't really useful for development as it means people need to know the fields they want already before they go to the docs for an example of it rather than being able to deduce how to lay something out from the explain command

An example of this would be when I was creating a CronWorkflow. I presumed base on [the example here](https://argoproj.github.io/argo-workflows/workflow-concepts/#dag) that the `.spec.workflowSpec.templates.dag.tasks.template.name` would be the name of a WorkflowTemplate, however this isn't really clear. I only know by hunting through examples in the GitHub repo that you would have to use `spec.workflowSpec.workflowTemplateName.name` to reference a workflow template.

I feel there is room for improvement here and if I can find the time to help with this I will. Either getting the explain command to work on a more descriptive level or improving any docs where possible.

drew-viles:
I've located this - it's something for now I guess.
https://argoproj.github.io/argo-workflows/fields/#workflow

Just not ideal when working on a cli 😉

PR #13753, opened 2024-10-14
build(docs): include `WorkflowEventBinding` in field reference

MasonM:
Partial fix for #12166

### Motivation

The [WorkflowEventBinding](https://argo-workflows.readthedocs.io/en/latest/events/#submitting-a-workflow-from-a-workflow-template) CRD is not listed at https://argo-workflows.readthedocs.io/en/latest/fields/, nor is the spec available via `kubectl explain WorkflowEventBinding.spec` due to https://github.com/argoproj/argo-workflows/issues/8190#issuecomment-1101955858. This fixes the former problem, and https://github.com/argoproj/argo-workflows/pull/13754 fixes the latter.
### Modifications

Add `io.argoproj.workflow.v1alpha1.WorkflowEventBinding` to the field list in `hack/docs/fields.go`
### Verification
Ran `make docs-serve` and visited http://localhost:8000/fields/#workfloweventbinding:

Screenshot 2024-10-14 at 2 02 25 PM


docs/fields.md
```diff
@@ -1295,6 +1317,392 @@ CronWorkflowStatus is the status of a CronWorkflow
...
 |`succeeded`|`integer`|v3.6 and after: Succeeded counts how many times child workflows succeeded|
 
+## WorkflowEventBindingSpec
+
+_No description available_
+
+<details markdown>
+<summary>Examples with this field (click to open)</summary>
+
+- [`archive-location.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/archive-location.yaml)
```
MasonM, line 1327: FYI: These examples shouldn't be here, since they don't reference `WorkflowEventBinding` at all. This is also present in every other CRD with `spec` field because the code that generates this page assumes `spec` means the same for all CRDs: https://github.com/argoproj/argo-workflows/blob/948127c7938b32a9a56ff74e31c2367d694d876d/hack/docs/fields.go#L302-L306

agilgur5: The other 4 CRDs are the same though (more or less), so it sounds like there needs to be a change to the script to handle this CRD properly

MasonM: Okay, I pushed a hack in cf20a755b54aa41b643fdfbf9ec357f2338aace8 to exclude examples containing `"spec"` from anything except `WorkflowSpec` and `CronWorkflowSpec`

docs/fields.md
```diff
@@ -5963,380 +6039,6 @@ WindowsSecurityContextOptions contain Windows-specific options and credentials.
 
 PersistentVolumeClaimSpec describes the common attributes of storage devices and allows a Source for provider-specific attributes
 
-<details markdown>
```
Joibel, line 5966: :+1:

docs/fields.md
```diff
@@ -2001,6 +2037,38 @@ v3.6 and after: StopStrategy defines if the CronWorkflow should stop scheduling
 |:----------:|:----------:|---------------|
 |`condition`|`string`|v3.6 and after: Condition is an expression that stops scheduling workflows when true. Use the variables `failed` or `succeeded` to access the number of failed or successful child workflows.|
 
+## Event
+
+_No description available_
```
Joibel, line 2042: It would be good to go through and add descriptions at some point, but this isn't the only place where we haven't got them.

Joibel (approved):

PR #13756, opened 2024-10-15
fix(controller): Add configmap already exists judgment on offload Env…

### LINKED ISSUE #13755: When offloading env, no error judgment was made for configmap.
### Pre-requisites

## What happened?
if workflow pod used offload envVarTemplate, when pod create failed, and a transient error occurred. The pod will retry on moment, then this node will failed with `IsAlreadyExists` error.
image
## What did you expect to happen?
we should process the `AlreadyExists` error

### Version(s)

main

```YAML
None
```

```text
None
```

```text
None
```

fyp711:
…VarTemplate

Fixes #13755 

### Motivation

if workflow pod used offload envVarTemplate, when pod create failed, and a transient error occurred. The pod will retry on moment, then this node will failed with IsAlreadyExists error.

### Modifications

We should process `Already Exists` error

### Verification

The best way to verifiy is constructing a too lang template, and than simulate a transient error

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -460,9 +460,13 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 		}
 		created, err := woc.controller.kubeclientset.CoreV1().ConfigMaps(woc.wf.ObjectMeta.Namespace).Create(ctx, cm, metav1.CreateOptions{})
 		if err != nil {
-			return nil, err
+			if !apierr.IsAlreadyExists(err) {
+				return nil, err
+			}
+			woc.log.Infof("Configmap already exists: %s", cm.Name)
+		} else {
+			woc.log.Infof("Created configmap: %s", created.Name)
 		}
-		woc.log.Infof("Created configmap: %s", created.Name)
 
 		volumeConfig := apiv1.Volume{
 			Name: "argo-env-config",


workflow/controller/workflowpod.go
```diff
@@ -460,7 +460,9 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 		}
 		created, err := woc.controller.kubeclientset.CoreV1().ConfigMaps(woc.wf.ObjectMeta.Namespace).Create(ctx, cm, metav1.CreateOptions{})
 		if err != nil {
-			return nil, err
+			if !apierr.IsAlreadyExists(err) {
+				return nil, err
+			}
 		}
 		woc.log.Infof("Created configmap: %s", created.Name)
```
Joibel, line 467: This log message is now misleading, could you fix it (I would suggest two possible logs, one for created and one for already exists)..

workflow/controller/workflowpod.go
```diff
@@ -460,7 +460,9 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
 		}
 		created, err := woc.controller.kubeclientset.CoreV1().ConfigMaps(woc.wf.ObjectMeta.Namespace).Create(ctx, cm, metav1.CreateOptions{})
 		if err != nil {
-			return nil, err
+			if !apierr.IsAlreadyExists(err) {
```
Joibel, line 463: nit: probably neater as a single line
```go
if err != nil && !apierr.IsAlreadyExists(err)
```

Joibel (changes requested):

workflow/controller/workflowpod.go
fyp711: fixed, thanks.

fyp711: ok

fyp711: fixed

workflow/controller/workflowpod.go
```diff
@@ -460,9 +460,13 @@ func (woc *wfOperationCtx) createWorkflowPod(ctx context.Context, nodeName strin
...
+				return nil, err
+			}
+			woc.log.Infof("Configmap is already exists: %s", cm.Name)
```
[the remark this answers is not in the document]
fyp711: Done

Joibel:
Thanks for this, and for testing 3.6

PR #13760, opened 2024-10-15
fix: better error message for multiple workflow controllers running

gregsheremeta:
### Motivation

Per various GitHub issues and personal experience, the "expected < 2 pods, got 2 - this is a bug" error message mostly happens these days when users accidentally have multiple workflow controllers running. This is usually caused by a bad install, e.g. installing a cluster-wide Argo Workflows alongside a namespace-scoped one, and having both controllers reconciling Workflows.

### Modifications

Improve the error message to advise users to check for this.

### Verification

I didn't verify this, since it's just a text change.

### DIFF
--- a/workflow/controller/workflowpod.go
+++ b/workflow/controller/workflowpod.go
@@ -566,7 +566,9 @@ func (woc *wfOperationCtx) podExists(nodeID string) (existing *apiv1.Pod, exists
 	}
 
 	if objectCount > 1 {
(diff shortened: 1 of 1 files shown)


tooptoop4 (approved):

tczhao (approved):
The issue could be due to problem in leader election, since we typically have multiple workflow-controller running but only 1 is doing the actual work.

gregsheremeta:
who can merge? This is my first contribution :)

terrytangyuan (approved):

terrytangyuan:
Thanks Greg!

PR #13762, opened 2024-10-15
docs: update CONTRIBUTORS.md including guidance on PR change request timeliness and PR takeover

Joibel:
### Motivation

Document something that I think we already kind of do with regard to getting PRs reviewed.

This is somewhat to show we accept PR takeovers, in case someone is willing to pick up an PR but doesn't think they're allowed to.

### Modifications

Add a section on Timeliness to the contributing guidelines suggesting the process for taking over a PR and ensuring credits go to the appropriate person.

One sentence per line

Removed section on not assigning to non-members, this feels unwelcoming.

### Verification

docs only: Checked in github markdown.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -10,13 +10,14 @@ See [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/master/code-o
 
 ## Community Meetings (monthly)
 
-A monthly opportunity for users and maintainers of Workflows and Events to share their current work and
-hear about what’s coming on the roadmap. Please join us! For Community Meeting information, minutes and recordings
-please [see here](http://bit.ly/argo-wf-cmty-mtng).
+A monthly opportunity for users and maintainers of Workflows and Events to share their current work and hear about what's coming on the roadmap.
+Please join us!
+For Community Meeting information, minutes and recordings please [see here](http://bit.ly/argo-wf-cmty-mtng).
 
 ## Contributor Meetings (twice monthly)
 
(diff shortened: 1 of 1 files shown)


agilgur5 (commented):
> This is somewhat to show we accept PR takeovers, in case someone is willing to pick up an PR but doesn't think they're allowed to.

> suggesting the process for taking over a PR and ensuring credits go to the appropriate person.

I think this makes sense to document since we do get these questions and make these suggestions, although the tone currently reads more "rule" than "suggested process". I think the credit part is the only thing that is a firm "rule" (and also proper copyright/attribution/courtesy/respect/OSS norm/etc).

Also would be good to mention the `problem/more information needed` label here as well, since it does apply to PRs and the specific case of authors not responding timely.

I can make more specific language suggestions later

Joibel:
I have 

> > This is somewhat to show we accept PR takeovers, in case someone is willing to pick up an PR but doesn't think they're allowed to.
> 
> > suggesting the process for taking over a PR and ensuring credits go to the appropriate person.
> 
> I think this makes sense to document since we do get these questions and make these suggestions, although the tone currently reads more "rule" than "suggested process". I think the credit part is the only thing that should be a firm "rule" (and also proper copyright/attribution/courtesy/respect/OSS norm/etc).
> 
> Also would be good to mention the `problem/more information needed` label here as well, since it does apply to PRs and the specific case of authors not responding timely.
> 
> I can make more specific language suggestions later

I have softened the wording in general, but made the co-author section more strong.

isubasinghe (approved):

PR #13766, opened 2024-10-15
feat(ui): prefill parameters for workflow submit form. Fixes #12124

### LINKED ISSUE #12124: Prefill parameters for the workflow submit form
# Summary

When submitting workflows through the web ui, it's not possible (or just not documented) to prefill the parameters through the URL (without triggering the Workflow right away). Deeplinking from external systems with the parameters prefilled allows to integrate with external systems for easier runbook automation.

## Use Cases

Coming from Jenkins users might be familiar with "buildWithParameters" jobs. These were useful e.g. when linking from Alerts/Incidents to Troubleshooting Workflows passing the correct context information into the parameters.

An alert could e.g. easily pass the Pod and Namespace information into the workflow as a parameter so the user doesn't have to copy/paste it. 

---

sairam91:
Fixes #12124

### Motivation
This PR  adds support for developers to prefil submit workflow form by passing in the values through query parameters.

To prefil a submit workflow developers can pass in the following query parameters

```
sidePanel=submit-new-workflow
template=<template that the workflow should follow>
parameters[key]=value // Here the key would be the parameter for the job and value would be the value of the parameter.
You should provide the keys that are supported by the specified template
```

### Modifications
- Updated `workflow-list` component
- Updated `submit-workflow-panel.tsx` component to default entry point if provided.
- Updated `workflow-creator.tsx` to read default values from query parameters.
 - The workflow will be prefilled with the parameters that the template supports
- Added new shared helper function to parse query parameter of the format `?parameters[key]=value`.


ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -330,7 +343,23 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
                     )}
                 </div>
             </div>
-            <SlidingPanel isShown={!!getSidePanel()} onClose={() => navigation.goto('.', {sidePanel: null})}>
+            <SlidingPanel
+                isShown={!!getSidePanel()}
+                onClose={() => {
+                    // Remove any lingering query params
+                    const qParams: {[key: string]: string | null} = {
```
sairam91, line 350: This is duplicate can be removed.

ui/src/app/shared/get_workflow_params.ts
```diff
@@ -0,0 +1,33 @@
...
+    // If a match is found, return the captured key
+    if (match) {
+        return match[1];
+    }
+
+    // If no match is found, return null or an empty string
+    return null; // Or return '';
+}
+
+export function getWorkflowParametersFromQuery(): {[key: string]: string} {
```
sairam91, line 21: Add Doc to function

MasonM:
Good start! A couple initial comments:
1. This project uses the [history](https://github.com/remix-run/history/) library, so the `useEffect()` hooks need to incorporate it as a dependency. That means passing through the `history` object to the affected components. Ideally, we'd use `useQueryParams()` too, but when I tried that, the `history.listen()` callback wouldn't fire. I think that's due to the interdependency with the `workflowTemplates`/`workflowParameters` props.
2.  It'd be simpler to push the logic for handling workflow parameters to `ui/src/app/workflows/components/submit-workflow-panel.tsx`

Here's a commit that does both things: https://github.com/MasonM/argo-workflows/commit/d2c11c87503c5054c9ce86d5a19918ed79f8cc24

agilgur5 (commented):
> 1. [...] Ideally, we'd use `useQueryParams()` too, but when I tried that, the `history.listen()` callback wouldn't fire. I think that's due to the interdependency with the `workflowTemplates`/`workflowParameters` props

In this case, I don't think it's necessary, because the URL should not change after the page loads, since the purpose is deeplinking. `history.listen()` is to make sure the state and query params match.
It could still be used, but may be unnecessary complexity in this case.

> 2. It'd be simpler to push the logic for handling workflow parameters to `ui/src/app/workflows/components/submit-workflow-panel.tsx`

Agree, that was what I thought on first glance of this yesterday too, but hadn't commented yet since it's still draft and wanted to see if you or Sairam would pick that up in iteration 🙂

ui/src/app/shared/get_workflow_params.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Method to extract a key from query parameter. The query parameter will be of the following format
+ * ?parameters[key]=value.
+ * This method will extract the key from the query parameter.
+ * @param inputString
+ * @returns
+ */
+function extractKey(inputString: string): string | null {
+    // Use regular expression to match the key within square brackets
+    const match = inputString.match(/parameters\[(.*?)\]/);
```
MasonM, line 11: Probably should anchor this so it doesn't match something like `redirectparameters[foo]`
```suggestion
    const match = inputString.match(/^parameters\[(.*?)\]$/);
```

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
MasonM: Could you explain why this is necessary? The `navigation.goto()` function already has logic to automatically strip out the query parameters if there's been a path change: https://github.com/argoproj/argo-ui/blob/54f36c723d9a0e5d3386689298cbb9c27767fa00/src/components/navigation.ts#L20-L21

If there hasn't been a path change, I think it's fine to leave these query parameters in.

ui/src/app/workflows/components/submit-workflow-panel.tsx
```diff
@@ -28,13 +31,27 @@ const defaultTemplate: Template = {
...
+        const templatePropertiesInQuery = getWorkflowParametersFromQuery(props.history);
+        // Get the user arguments from the query params
+        const updatedParams = workflowParameters.map(param => {
+            const queryValue = templatePropertiesInQuery[param.name];
+            const p: Parameter = {
+                name: param.name,
+                value: queryValue || param.value
+            };
+            return p;
+        });
```
MasonM, line 51: You can simplify this a bit by getting rid of the explicit type, because TypeScript is smart enough to know this matches `Parameter[]`:
```suggestion
        const updatedParams = workflowParameters.map(param => ({
            name: param.name,
            value: templatePropertiesInQuery[param.name] || param.value
        }));
```

ui/src/app/shared/get_workflow_params.ts
```diff
@@ -0,0 +1,37 @@
+import {History} from 'history';
+/**
+ * Method to extract a key from query parameter. The query parameter will be of the following format
+ * ?parameters[key]=value.
+ * This method will extract the key from the query parameter.
+ * @param inputString
+ * @returns
```
MasonM, line 7: This repository usually doesn't have JSDoc annotations since it uses TypeScript:
```suggestion
```

ui/src/app/shared/get_workflow_params.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (match) {
+        return match[1];
+    }
+
+    // If no match is found, return null or an empty string
+    return null; // Or return '';
+}
+/**
+ * Returns the workflow parameters from the query parameters.
+ * @returns {}
```
MasonM, line 23: Same as above
```suggestion
```

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -135,6 +136,18 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
...
+        // Add any workflow parameters to the query
+        const workflowProperties = getWorkflowParametersFromQuery(history);
+
+        Object.keys(workflowProperties).forEach(key => {
+            params.append(`parameters[${key}]`, workflowProperties[key]);
+        });
+
+        // Add the sidePanel query parameter if it exists
+        params.append('sidePanel', getSidePanel());
+        params.append('template', queryParams.get('template'));
```
MasonM, line 149: I'm not sure why this component is taking an allow-list approach to query parameters, but you could get rid of this extra code by changing `const params = new URLSearchParams()` to `const params = new URLSearchParams(history.location.search);`

I'm not sure what side-effects that'll have, though.

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
sairam91: So the workflow this would fix is
 - Open the submit page prefilled through the URL
 - Close the submit page
 - Click on Submit again, 
 
 In this case, if the parameters are preserved, we would open the prefilled page. Removing the parameters will ensure the developer can start the page fresh without pre filled data.

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
sairam91: We could preserve all the parameters. We might have to try different scenarios to check for side-effects.

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
MasonM: Okay, thanks. IMO, it'd be better to remove this, because if a developer wants to start fresh, they can click on the "Workflows" icon on the left bar (or navigate to another page and back again)

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -126,7 +126,7 @@ export function WorkflowsList({match, location, history}: RouteComponentProps<an
         }
         storage.setItem('options', options, {} as WorkflowListRenderOptions);
 
-        const params = new URLSearchParams();
+        const params = new URLSearchParams(history.location.search);
```
MasonM, line 129: I found one side-effect of this change: if you visit http://localhost:8080/workflows?namespace=argo and refresh the page, it'll append `&namespace=argo` each time you refresh. That's happening because `historyUrl()` isn't deduplicating query parameters, which it should IMO. Here's a commit that fixes that: https://github.com/MasonM/argo-workflows/commit/8b48d5b627a17155c1a5a4b2778623d59aaa0e70

I haven't extensively tested that, so it's possible there's other side-effects

ui/src/app/shared/get_workflow_params.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    it('should return an empty object when there are no query parameters', () => {
+        const history = createBrowserHistory();
+        const result = getWorkflowParametersFromQuery(history);
+        expect(result).toEqual({});
+    });
+
+    it('should return the parameters provided in the URL', () => {
+        const history = createBrowserHistory();
+        const path = createPath({pathname: '/workflows', search: '?parameters[key1]=value1&parameters[key2]=value2'});
+        history.location.search = path;
```
MasonM, line 14: This test is failing because `window.location.search` should only contain the query string, not the path:
```suggestion
        history.location.search = '?parameters[key1]=value1&parameters[key2]=value2';
```
Also, if you haven't already, you should install the Jest extension in the DevContainer so you can debug the tests and set breakpoints:
Screenshot 2024-10-17 at 11 25 15 AM

ui/src/app/shared/get_workflow_params.test.ts
```diff
@@ -0,0 +1,32 @@
...
+        expect(result).toEqual({
+            key1: 'value1',
+            key2: 'value2'
+        });
+    });
+
+    it('should only return the parameters provided in the URL', () => {
+        const history = createBrowserHistory();
+        const path = createPath({pathname: '/workflows', search: '?parameters[key1]=value1&parameters[key2]=value2&test=123'});
+        history.location.search = path;
```
MasonM, line 25: Same as above:
```suggestion
        history.location.search = '?parameters[key1]=value1&parameters[key2]=value2&test=123';
```

ui/src/app/workflows/components/workflows-list/workflows-list.tsx
MasonM: Okay, that makes sense. I checked and this is the only page using a`<SlidingPanel>` that calls `navigation.goto()` on close. Every other page just sets a parameter on close, e.g. https://github.com/argoproj/argo-workflows/blob/e09d2f1121c142bc47cc12f543b1790a92baf8d1/ui/src/app/cluster-workflow-templates/cluster-workflow-template-details.tsx#L145

I'm not sure why this page is special in that respect, but it could be a developer experience thing

MasonM:
@sairam91 Sorry for the delay. You mentioned you were going to fill out the "Verification" section of the PR, so I was waiting for that. If you're unsure of what to put there, I think this would work (you can copy-and-paste this):
### Verification
1. Run `make start UI=true`
2. Visit http://127.0.0.1:8080/workflow-templates?namespace=argo and create a new `WorkflowTemplate` named with the default example
image
3. Visit http://127.0.0.1:8080/workflows?namespace=argo&sidePanel=submit-new-workflow&template=test&parameters%5Bmessage%5D=test123 and verify fields are populated correctly
image


Also, it looks like there's conflicts. If you can add me as a collaborator at https://github.com/sairam91/argo-workflows/settings, I can resolve them for you, if you want.

sairam91:
Closed in favor of https://github.com/argoproj/argo-workflows/pull/13922

PR #13769, opened 2024-10-16
ci: increase timeout for e2e-tests

MasonM:
Partial fix for #10807

### Motivation

Occasionally, tests are being canceled due to exceeding the 30 minute timeout. Example: https://github.com/argoproj/argo-workflows/actions/runs/11358993638/job/31594490093

While there are optimizations we can make to the test suite to mitigate this, sometimes resource contention causes things to slow to a crawl for reasons outside our control. In the example above, it spent 9m15s on "Free up unused disk space", which normally takes ~12s. The only thing that step is doing is running `rm -rf`, so that indicates extremely high I/O contention on the runner.

### Modifications

This removes the timeout so it uses the default of 360 minutes ([docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)). This is what ArgoCD does ([source](
https://github.com/argoproj/argo-cd/blob/049ba0aab3965f3fa73447ae2db08f3a2afd95a0/.github/workflows/ci-build.yaml#L398)), and E2E tests for ArgoCD PRs often take ~50 minutes ([example](https://github.com/argoproj/argo-cd/actions/runs/11352918460/job/31576914061)).

### Verification
Wait for actions to run

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -189,6 +189,9 @@ jobs:
     needs: [ changed-files, argo-images ]
     if: ${{ needs.changed-files.outputs.e2e-tests == 'true' }}
     runs-on: ubuntu-latest
+    # These tests usually finish in ~25m, but occasionally they take much longer due to resource
+    # contention on the runner, which we have no control over.
+    timeout-minutes: 60
     env:
       KUBECONFIG: /home/runner/.kubeconfig
       E2E_ENV_FACTOR: 2
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -189,7 +189,6 @@ jobs:
     needs: [ changed-files, argo-images ]
     if: ${{ needs.changed-files.outputs.e2e-tests == 'true' }}
     runs-on: ubuntu-latest
-    timeout-minutes: 30
     env:
       KUBECONFIG: /home/runner/.kubeconfig
       E2E_ENV_FACTOR: 2


.github/workflows/ci-build.yaml
```diff
@@ -189,7 +189,6 @@ jobs:
     needs: [ changed-files, argo-images ]
     if: ${{ needs.changed-files.outputs.e2e-tests == 'true' }}
     runs-on: ubuntu-latest
-    timeout-minutes: 30
```
Joibel, line 192: My understanding is that argoproj shares the compute resources amongst all running github actions.

I do agree with increasing the timeout, but I'd rather limit it to a number smaller than 6 hours, say 60 minutes. This would prevent bad tests which run on for much longer than expected from just wasting cycles that could then be given to other github actions that are waiting to happen.

Joibel (changes requested):

.github/workflows/ci-build.yaml
MasonM: > My understanding is that argoproj shares the compute resources amongst all running github actions.

I don't think that's true, because this is a public repository. Private repositories are limited to 2,000 minutes under the free plan ([source](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions)), but this limit doesn't apply to public repositories.

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the account's plan. Any usage beyond the included amounts is controlled by spending limits.

Anyway, I pushed 67722f146d15736e0b54471efa49833bb4a25c6c to reduce the timeout to 60 minutes

Joibel: I was assuming the [maximum runners limit of 20](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits) would kick in. When a lot of PRs are in flight at once then we do see actions not starting, so something is causing that. I would like to prevent rogue CI processes hogging the runners.

Thank you for the change.

Joibel (approved):

.github/workflows/ci-build.yaml
Joibel: If we see 60 is genuinely too small we can bump it again.

agilgur5: I've mentioned the GH concurrency limits in https://github.com/argoproj/argo-workflows/issues/10807#issuecomment-2094289752 and elsewhere and can confirm we do see some queuing when there are many PRs or when releases are being run.

This might be worth noting somewhere in the contributor docs as a point of reference

PR #13775, opened 2024-10-17
refactor: remove `util/slice` and use standard `slices` library

Joibel:
### Motivation

`/util/slices` can be replaced by the official [slices library](https://pkg.go.dev/slices) added in go 1.21. This library is also the official version of `golang.org/x/exp/slices`.

Replace all uses of these libraries with slices

### Modifications

Replace `slice.ContainsString()` with `slices.Contains()`. This is very simple.

Replace `slice.RemoveString()` with `slices.DeleteFunc()`, which involves a small inline boolean comparison function. `RemoveString` returned a copy of the slice, `DeleteFunc` modifies the slice in place. In the one case where we're using RemoveString to clone the slice (`getPodCleanupPatch`), `slices.Clone()` it before modifying.

Also fix a test name with a spelling mistake.

### Verification

Test suite to catch the problems


Joibel:
Draft as it collides with #13776/#12862

Joibel:
> Draft as it collides with #13776/#12862

This is now merged and this rebased on top.

util/slice/slice.go
```diff
@@ -1,21 +0,0 @@
...
-
-func RemoveString(slice []string, element string) []string {
-	for i, v := range slice {
-		if element == v {
-			ret := make([]string, 0, len(slice)-1)
-			ret = append(ret, slice[:i]...)
-			return append(ret, slice[i+1:]...)
-		}
-	}
-	return slice
```
agilgur5, line 11: I assume that you decided that it wasn't worth keeping this file for a simplified version of this helper that uses `DeleteFunc`? i.e. `return slices.DeleteFunc(slice, func (x string) bool { return x == element }`

agilgur5 (approved):
LGTM. Thanks for simplifying this. I left one question here around having a helper vs an anonymous in-line func for `DeleteFunc`

util/slice/slice.go
Joibel: I considered it, but didn't do it as:
* It behaves differently from the original (modifying the input slice in place). This is sort of surprising, as it also returns the resulting slice. The `slices` version documents this for us.
* `RemoveString()` only operated on arrays of `string`s, not generic arrays, so overall encouraging it to stay around and potentially get duplicated to add `RemoveMyOtherType()` was worse long term. Possibly genericing it would have been useful, but probably worse as a long term option.

agilgur5: Yea I considered making it a generic `RemoveElement` as well. 

Thanks for noting this; I don't have a strong opinion on this but wanted to make sure the rationale was documented since the decision seemed intentional

PR #13777, opened 2024-10-17
fix: remove JSON cast when querying archived workflows

MasonM:
### Motivation

With PostgreSQL, the `argo_archived_workflows.workflow` column has been of type `json` ever since 8a1e611a03da8374567c9654f8baf29b66c83c6e, which was released as v2.5.0. Therefore, the `::json` casts do nothing, and prevent users from improving performance by migrating to JSONB using the following query:
```sql
alter table argo_archived_workflows alter column workflow set data type jsonb using workflow::jsonb
```

Without the changes in this PR, running the above will massively slow down the queries, because casting JSONB to JSON is expensive. With the changes, I'm measuring a ~80% performance boost (see below). [Other users have reported](https://github.com/argoproj/argo-workflows/issues/13601#issuecomment-2376579589) similar improvements.

The only downside to migrating to JSONB is it can take a long time if you've got a ton of workflows (~72s on my local DB with 100,000 workflows). I entered https://github.com/argoproj/argo-workflows/pull/13779 with the migration, but I'm entering this separately so it can hopefully go out in 3.6.0. That way, we can tell users to run the above query if they run into performance issues, without having to do a patch release with the migration.

### Modifications

Replace `(workflow::json)` with just `workflow`

### Verification
See https://github.com/argoproj/argo-workflows/pull/13779


terrytangyuan (commented):
@jiachengxu Would you mind reviewing this?

MasonM:
Here's the PR for the migration, which builds on top of this PR: https://github.com/argoproj/argo-workflows/pull/13779

jiachengxu (approved):
LGTM

jessesuen (approved):

agilgur5:
> @jiachengxu Would you mind reviewing this?

@terrytangyuan Jiacheng is also apparently not in the "members" team a la https://github.com/argoproj/argo-workflows/issues/8790#issuecomment-2412016304, so that seems to be why we can't directly "request a review" via GH

terrytangyuan:
Added @jiachengxu to members

agilgur5:
Thanks Terry 👍

PR #13781, opened 2024-10-18
docs: remove incorrect statement about DIND not working

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/discussions/13010#discussioncomment-9372559, https://github.com/argoproj/argo-workflows/pull/10778#issuecomment-2105499778

### Motivation

As users have mentioned, DIND does still work and the specific DIND image used comes with Docker Engine and so does not rely on the k8s container engine

### Modifications

- replace the `dockershim` note with one about "Alternatives", since DIND is still not as secure
  - use a proper admonition for the note as well, instead of "Note:" with a horizontal rule after

- also follow docs style guide for the rest of the page while at it
  - prefer 1 sentence per line, per [style guide](https://argo-workflows.readthedocs.io/en/release-3.5/doc-changes/)
  - use active voice, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice)
  - simple and direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
  - consistently address the user as "you" (instead of using both "you" and "we"), per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you)
  - not in the style guide, but link to GH instead of `raw.github` as GH is much nicer to read (e.g. syntax highlighting, line numbers, etc etc) and won't auto-download a file on certain browsers (both especially for mobile)

- add a link to the sidecar docs when it is mentioned as well

### Verification

`make docs` passes


joebowbeer:
Thanks for doing this.

Joibel (approved):
Thanks.

PR #13790, opened 2024-10-19
feat: include container name in error message. Fixes #10007

### LINKED ISSUE #10007: OOMKilled error for gitclone does not say which container was killed? init, main, wait ?
### Pre-requisites

image

 i have 5g memory set for this step, sometimes works, sometimes doesn't

not sure if i need to increase init or wait container memory too?

is it possible to make the error message say which container was OOMkilled?

### Version

3.4.3

```YAML
https://github.com/argoproj/argo-workflows/blob/9f5759b5bd2a01d0f2930faa20ad5a769395eb99/examples/input-artifact-git.yaml like this one
```

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/10007


shuangkun (approved):

workflow/controller/operator.go
```diff
@@ -1629,6 +1629,7 @@ func (woc *wfOperationCtx) inferFailedReason(pod *apiv1.Pod, tmpl *wfv1.Template
 		if t.Message != "" {
 			msg = fmt.Sprintf("%s: %s", msg, t.Message)
 		}
+		msg = fmt.Sprintf("%s in %s", msg, ctr.Name)
```
agilgur5, line 1632: I feel like `<name>: <message>` is easier to interpret than `<message> in <name>` for all of these

workflow/controller/operator.go
```diff
@@ -1406,7 +1406,7 @@ func (woc *wfOperationCtx) assessNodeStatus(ctx context.Context, pod *apiv1.Pod,
 			woc.markNodePhase(ctrNodeName, wfv1.NodeRunning)
 		case c.State.Terminated != nil:
 			exitCode := int(c.State.Terminated.ExitCode)
-			message := fmt.Sprintf("%s (exit code %d): %s", c.State.Terminated.Reason, exitCode, c.State.Terminated.Message)
+			message := fmt.Sprintf("%s (exit code %d): %s in %s", c.State.Terminated.Reason, exitCode, c.State.Terminated.Message, c.Name)
```
agilgur5, line 1409:
```suggestion
			message := fmt.Sprintf("%s: %s (exit code %d): %s", c.Name, c.State.Terminated.Reason, exitCode, c.State.Terminated.Message)
```

same as above

test/e2e/cli_test.go
```diff
@@ -894,7 +894,7 @@ func (s *CLISuite) TestWorkflowRetryWithRecreatedPVC() {
 			assert.Equal(t, wfv1.NodeFailed, status.Nodes.FindByDisplayName("print").Phase)
 			// This step is failed intentionally to allow retry. The error message is not related to PVC that is deleted
 			// previously since it is re-created during retry.
-			assert.Equal(t, "Error (exit code 1)", status.Nodes.FindByDisplayName("print").Message)
+			assert.Equal(t, "Error (exit code 1) in main", status.Nodes.FindByDisplayName("print").Message)
```
agilgur5, line 897:
```suggestion
			assert.Equal(t, "main: Error (exit code 1)", status.Nodes.FindByDisplayName("print").Message)
```

same as above

tooptoop4:
docs test seems unrelated

agilgur5:
> /retest

Oh that only runs for `ci-build.yaml`, right. Manually retrigerred the Docs CI. Never seen that flake before though 🤔

agilgur5 (approved):
This makes sense to me, but would like more folks to take a look since it will change a lot of messages

tczhao (commented):
I think this a breaking change, we better have [a documentation](https://argo-workflows.readthedocs.io/en/release-3.6/upgrading/) for this.
I can imagine there exist users using [conditional retry](https://argo-workflows.readthedocs.io/en/release-3.5/retries/#conditional-retries) with exact matching on message

agilgur5:
There are some libraries that consider errors as part of the API boundary, but we currently [don't](https://argo-workflows.readthedocs.io/en/release-3.5/public-api/). This also isn't the first time we've changed an error.



> with exact matching

Some libraries, like `protobuf`, even intentionally insert randomness to prevent you from doing exact matching like with https://github.com/argoproj/argo-workflows/pull/12737#issuecomment-1977948270 / https://github.com/prometheus/client_model/issues/83

tooptoop4:
> I think this a breaking change, we better have [a documentation](https://argo-workflows.readthedocs.io/en/release-3.6/upgrading/) for this. I can imagine there exist users using [conditional retry](https://argo-workflows.readthedocs.io/en/release-3.5/retries/#conditional-retries) with exact matching on message

added a note

docs/upgrading.md
```diff
@@ -104,6 +104,10 @@ To disable this set `metricsConfig.secure` to `false`.
 When returning a map or array in an expression, you would get a Golang representation.
 This now returns plain JSON.
 
+### Container name in error messages
```
agilgur5, line 107:
```suggestion
### Added container name to workflow node error messages
```

- this could be more specific
- it should also be in active voice of what was changed

docs/upgrading.md
```diff
@@ -104,6 +104,10 @@ To disable this set `metricsConfig.secure` to `false`.
...
+
+Error messages are prefixed with container name, you may need to adjust your `lastRetry.message` expressions or `TRANSIENT_ERROR_PATTERN` variable. See [Conditional retries](retries.md)
```
agilgur5, line 109:
```suggestion
Workflow node error messages are now prefixed with the container name.
If you are using [Conditional Retries](retries.md), you may need to adjust your usage of `lastRetry.message` expressions or the `TRANSIENT_ERROR_PATTERN` environment variable.
```

agilgur5 (commented):
Some copy-edits below

tczhao (approved):

agilgur5 (approved):

PR #13794, opened 2024-10-20
feat(cli): `archive list` option to filter by name prefix. Fixes #10524

### LINKED ISSUE #10524: argo archive list needs a filter on wfname
argo archive list doesn't seem to accept filter on wfname

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/10524

### DIFF
--- a/cmd/argo/commands/archive/list.go
+++ b/cmd/argo/commands/archive/list.go
@@ -21,6 +21,7 @@ func NewListCommand() *cobra.Command {
 		selector  string
 		output    = common.NewPrintWorkflowOutputValue("wide")
 		chunkSize int64
+		prefix  string
 	)
 	command := &cobra.Command{
 		Use:   "list",
@@ -47,7 +48,7 @@ func NewListCommand() *cobra.Command {
 				return err
 			}
 			namespace := client.Namespace()
-			workflows, err := listArchivedWorkflows(ctx, serviceClient, namespace, selector, chunkSize)
+			workflows, err := listArchivedWorkflows(ctx, serviceClient, namespace, selector, chunkSize, prefix)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


cmd/argo/commands/archive/list.go
```diff
@@ -47,28 +48,29 @@ func NewListCommand() *cobra.Command {
...
-			workflows, err := listArchivedWorkflows(ctx, serviceClient, namespace, selector, chunkSize)
+			workflows, err := listArchivedWorkflows(ctx, serviceClient, namespace, selector, chunkSize, prefix)
 			if err != nil {
 				return err
 			}
 			return printer.PrintWorkflows(workflows, os.Stdout, printer.PrintOpts{Output: output.String(), Namespace: true, UID: true})
 		},
 	}
 	command.Flags().VarP(&output, "output", "o", "Output format. "+output.Usage())
+	command.Flags().StringVar(&prefix, "prefix", "", "Filter workflows by prefix")
```
agilgur5, line 59: I feel like this should be more specific like "name prefix"; the API parameter is `NamePrefix` too, so it would be good to match
```suggestion
	command.Flags().StringVar(&prefix, "name-prefix", "", "Filter workflows by name prefix")
```

agilgur5: It would also make sense to consider the style of #13160, where `--name` can be modified by `--name-filter`, which is one of "Exact" / "Prefix" / "Pattern".

Although it didn't implement that for the Archived API specifically

agilgur5: I see `--prefix` and the description verbatim [match the `list` CLI](https://github.com/argoproj/argo-workflows/blob/bb5130e841ab02cfe5b65f19c214fe309ca88bce/cmd/argo/commands/list.go#L117), so perhaps we leave that in for backward-compat but also add `--name` and `--name-filter` to match #13160

PR #13795, opened 2024-10-21
fix(cron): correctly run when `startingDeadlineSeconds` and `timezone` are set

### LINKED ISSUE #13786: v3.6-rc: CronWorkflows with `timezone` and `startingDeadlineSeconds` may run at the wrong time if relisted at the wrong time
### Pre-requisites

This bug is only in 3.6, not in 3.5.

I'm intending to fix this myself, but filing this in case anyone else discovers it before I've managed to write the tests.

In #12616 `shouldOutstandingWorkflowsBeRun` stopped taking account for timezone in the schedules it was analyzing. More specifically https://github.com/argoproj/argo-workflows/blob/c9b1477fd575bf06bed43ca2139f74aa3af4285c/workflow/cron/operator.go#L323 should be calling `GetSchedulesWithTimezone()` instead of plain `GetSchedules()` as we're attempting to compare with a timezone compensated value in `now`.

The fix is just that switch I believe, but I'll get my head around a regression test before submitting a PR.

If your Informer relist happens during the startingDeadlineSeconds in of your schedule when thought of without Timezone (in UTC or controller local?) you'll get an incorrect run of the CronWorkflow.

### Version(s)

3.6.0-rc2

```YAML
kind: CronWorkflow
metadata:
  name: hello-world-multiple-schedules
spec:
  schedules: # v3.6 and after
    - "0 3 * * *"
    - "0 2 * * *"
  timezone: "Pacific/Auckland"   # Default to local machine timezone
  startingDeadlineSeconds: 120
  concurrencyPolicy: "Replace"      # Default to "Allow"
  successfulJobsHistoryLimit: 4     # Default 3
  failedJobsHistoryLimit: 4         # Default 1
  suspend: false                    # Set to "true" to suspend scheduling
  workflowSpec:
    entrypoint: whalesay
    templates:
      - name: whalesay
        container:
          image: docker/whalesay:latest
          command: [cowsay]
          args: ["🕓 hello world. Scheduled on: {{workflow.scheduledTime}}"]
```

```text
missed an execution at <Sometime> and is within StartingDeadline
```

```text
Irrelevant
```

Joibel:
Fixes #13786 

### Motivation

If a CronWorkflow is tested for whether it should run within `startingDeadlineSeconds` it tests it in the wrong timezone.

This doesn't matter for hourly or minutely CWFs.
This doesn't matter if a CWF doesn't have a timezone associated with it.

This test happens on Informer relist, so if the relist occurs at the wrong time the CWF can be erroneously fired.

This was introduced in #12616 (see linked issue) in version 3.6.

### Modifications

The main change is to use `GetSchedulesWithTimezone()` rather than `GetSchedules()` when iterating over the schedules to check startingDeadlineSeconds.

The code was confusing (and confusing before #12616) in that it attempted to modify `timeNow()` to be in a local time. This was irrelevant and made spotting the bug harder. I have removed it. All `time.Time` comparisons are timezone aware, so modifying `time.Now()` using `In()` has no effect on any comparisons made.

### Verification

Manual testing with a 20second relist interval.

Two new unit tests (in a single function). If you reverse the main modification above both tests will then not pass and give inverse results for the `missedExecutionTime` assertions.

These tests will assert if run in Auckland timezone. This could be improved by making the test more complex (pick one of two timezones), but I didn't think this was worthwhile at the moment.

### Notes

This PR should not be backported to v3.5 or earlier

### DIFF
--- a/workflow/cron/operator.go
+++ b/workflow/cron/operator.go
@@ -320,27 +320,13 @@ func (woc *cronWfOperationCtx) shouldOutstandingWorkflowsBeRun(ctx context.Conte
 	}
 	// If this CronWorkflow has been run before, check if we have missed any scheduled executions
 	if woc.cronWf.Status.LastScheduledTime != nil {
-		for _, schedule := range woc.cronWf.Spec.GetSchedules(ctx) {
+		for _, schedule := range woc.cronWf.Spec.GetSchedulesWithTimezone(ctx) {
 			var now time.Time
 			var cronSchedule cron.Schedule
-			if woc.cronWf.Spec.Timezone != "" {
-				loc, err := time.LoadLocation(woc.cronWf.Spec.Timezone)
-				if err != nil {
-					return time.Time{}, fmt.Errorf("invalid timezone '%s': %s", woc.cronWf.Spec.Timezone, err)
-				}
-				now = time.Now().In(loc)
-
-				cronSchedule, err = cron.ParseStandard(schedule)
-				if err != nil {
-					return time.Time{}, fmt.Errorf("unable to form timezone schedule '%s': %s", schedule, err)
-				}
-			} else {
-				var err error
-				now = time.Now()
-				cronSchedule, err = cron.ParseStandard(schedule)
-				if err != nil {
-					return time.Time{}, err
-				}
+			now = time.Now()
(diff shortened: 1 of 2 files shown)


isubasinghe (approved):
LGTM

PR #13812, opened 2024-10-24
docs: update `examples/README.md` to its current usage

agilgur5:
Noticed this during https://github.com/argoproj/argo-workflows/pull/13645 et al when setting up embedded examples

### Motivation

- it previously said things were moved to the walk through, but the whole directory is still used
  - perhaps in the past this README in the past was used like a walk through, but this is no longer accurate and is confusing / misleading

### Modifications

- update it to mention its usage in the docs, such as in embedded examples like in https://github.com/argoproj/argo-workflows/pull/13645 and the [Field Reference](https://argo-workflows.readthedocs.io/en/release-3.5/fields/)
  - could also remove the file entirely, but the directory is used, so I figure it could have a short summary of how/where it is used

### Verification

markdown renders correctly (see [rich diff](https://github.com/argoproj/argo-workflows/pull/13812/files?short_path=49aaa28#diff-49aaa2819e35a856818ecec8c9fa7e1c79ad028d3f44bd749736353cfb51bac9))

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,3 +1,3 @@
 # Documentation by Example
 
-This has been moved to [the docs](https://argo-workflows.readthedocs.io/en/latest/walk-through/).
+This directory contains various examples and is referenced by [the docs site](https://argo-workflows.readthedocs.io).


Joibel (approved):
:+1:

PR #13343, opened 2024-07-15
feat(ui): Retry a single workflow step manually

Adrien-D:
Fixes #10139

### Motivation

Add the ability to manually retry a single step on workflows.

### Modifications

I based my changes on the retry workflow, but abstracting the `Node Field Selector` with the current selected node, less flexible but very user friendly.

Screenshot 2024-07-15 at 10 57 09
Screenshot 2024-07-15 at 10 57 18
Screenshot 2024-07-15 at 10 57 25
Screenshot 2024-07-15 at 10 57 38

### Verification

See screenshots

PR #13284, opened 2024-06-30
feat(ui): Add start/end workflows ISO display switch

Adrien-D:
Fixes #12943

### Motivation

To give the user the ability to display workflows start/stop time in ISO format.

### Modifications

I added the ability to switch time format with a toggle clock icon. This settings is persisted in localStorage.

Screenshot 2024-06-30 at 14 05 43
Screenshot 2024-06-30 at 14 05 49
Screenshot 2024-06-30 at 14 05 53

### Verification

I tested on multiple workflows to make sure this settings is shared across workflows.

PR #11811, opened 2023-09-13
feat: show history about completed runs in each cron workflow

juijeong8324:
Fixes #11706 

### Motivation
I want to check history about completed runs in each cron workflows.

### Modifications
image
You can check the history in cron workflow details page when you click specific cron-workflow.

image
Also, If the cron workflow doesn't have completed workflows, you can see explanation as above.

### Verification
Local running

PR #13813, opened 2024-10-24
build(ui): automatically sort imports

agilgur5:
Follow-up to https://github.com/argoproj/argo-workflows/pull/12539 "Future Work" part 3, https://github.com/argoproj/argo-workflows/pull/13160#discussion_r1633495520, https://github.com/argoproj/argo-workflows/pull/11811#discussion_r1331813108, and other places I've mentioned it

### Motivation

Automatically sort imports for consistent style, like Python's `isort` but for JS/TS

### Modifications

- using https://github.com/IanVS/prettier-plugin-sort-imports
  - with [default config](https://github.com/IanVS/prettier-plugin-sort-imports/blob/e28fa0e92832a011a923fc5e3bc5a20705102406/README.md?plain=1#L198) plus [a newline](https://github.com/IanVS/prettier-plugin-sort-imports/blob/e28fa0e92832a011a923fc5e3bc5a20705102406/README.md?plain=1#L250) before relative imports and [CSS handling](https://github.com/IanVS/prettier-plugin-sort-imports/blob/e28fa0e92832a011a923fc5e3bc5a20705102406/README.md?plain=1#L234)
    - tried to get it as close to the existing style as possible with the smallest diff and least config
  - plus [specify TS version](https://github.com/IanVS/prettier-plugin-sort-imports/blob/e28fa0e92832a011a923fc5e3bc5a20705102406/README.md?plain=1#L358) [according to `yarn.lock`](https://github.com/argoproj/argo-workflows/blob/49ff7a44ba307416282a1f5cd3b844d19bce7f88/ui/yarn.lock#L9723)
  
- ran `yarn lint` and committed changes

- manually fix https://github.com/argoproj/argo-workflows/pull/13343#discussion_r1792535276 and https://github.com/argoproj/argo-workflows/pull/13284#discussion_r1792538253 as those are [unsortable since they were placed after side-effect imports](https://github.com/IanVS/prettier-plugin-sort-imports/blob/e28fa0e92832a011a923fc5e3bc5a20705102406/README.md?plain=1#L153)
  - could only auto-fix some simpler grouping / alphabetization issues of https://github.com/argoproj/argo-workflows/pull/13284#discussion_r1792540976 et al

### Verification

`yarn lint` has no more changes, otherwise this is purely stylistic and has no semantic changes

### Notes to Reviewers

- decided to use a `prettier` plugin instead of an `eslint` plugin as there was less configuration
  - and I generally jive with letting a tool decide consistent style, although `prettier` is not quite zero config, but "low" config
  - https://github.com/import-js/eslint-plugin-import / https://github.com/un-ts/eslint-plugin-import-x would be the alternative
    - ESLint plugin could still potentially be useful in the future for its static analysis (not just formatting)


tczhao (approved):

terrytangyuan:
Added @Adrien-D

PR #13814, opened 2024-10-25
refactor(ui): simplify retry node by parametrizing retry workflow panel

agilgur5:
Fixes https://github.com/argoproj/argo-workflows/pull/13343#pullrequestreview-2359030878

### Motivation

Simplifies and consolidates UI manual retry code -- 100+ LoC consolidated

### Modifications

- parametrize based on the optional `props.nodeId`
  - slightly different heading and no field selector input, but otherwise the same
  - `props.onRetrySuccess` can actually be used in both variants
  - also ensure tooltip is in both
  - and improve wording a bit

- use the same retry panel element for both as (not just component)
  - which has nicer styling and animation, unlike the panel within a panel that was previously used
    - <details><summary>Screenshots:</summary>
    
        before:
        Screenshot 2024-10-24 at 9 02 25 PM

       after:
        Screenshot 2024-10-24 at 9 23 17 PM

   </details>

- fix types to not be optional in `workflow-node-info.tsx` (fixes https://github.com/argoproj/argo-workflows/pull/13343#discussion_r1794714750) and simplify code as such
- plus a few small other simplifications as mentioned in my above linked review

### Verification

Manually tested, works same as before, but with nicer styling + animation. See screenshots above


agilgur5:
I couldn't request a review from @Adrien-D due to not being in the "members" team -- same reason as https://github.com/argoproj/argo-workflows/issues/8790#issuecomment-2412016304 -- so tagging in this comment instead

PR #13818, opened 2024-10-25
refactor(deps)!: remove `swagger-ui-react`

agilgur5:
Partial fix for #13410 -- see #13821 for the other half to downgrade `react` that is a pure fix (vs this is a breaking change that is sizeable in its own right)
Replaces #13069 to resolve CVEs by just removing `swagger-ui-react` entirely
- c.f. https://github.com/argoproj/argo-workflows/pull/13069#pullrequestreview-2079815858, [Slack thread](https://cloud-native.slack.com/archives/C0510EUH90V/p1718130524678479), https://github.com/argoproj/argo-workflows/issues/11997#issuecomment-2363922918, etc

### Motivation

Removes the Swagger UI entirely from the UI as it is a gigantic dep (2nd largest only behind Monaco Editor) and often has vulnerable dependencies that are difficult to resolve without affecting other deps. According to #12061, this would shave off ~2MB / 826KB minified / 250KB gzipped from the UI bundle and will also speed up builds and installs as such.

It also no longer necessary to have in the UI as there is a Swagger UI in the docs after https://github.com/argoproj/argo-workflows/pull/10923 and the docs are versioned after https://github.com/argoproj/argo-workflows/issues/11390

### Modifications

- remove `swagger-ui-react` dep
  - and `@types/swagger-ui-react` devDep
- replace `/apidocs` UI page with a few links: download the OpenAPI / Swagger spec, download the JSON schema, and go to the Swagger UI in the versioned documentation
  - uses `ZeroState` component same as other pages, namely [the Plugins page](https://github.com/argoproj/argo-workflows/blob/1017c1dbe2ccf3f0a59ae68f04d954dd92e69673/ui/src/app/plugins/plugin-list.tsx#L32)
- replace other UI links to the API Docs with versioned docs links as well
  - could leave them as-is, but they seem to be intended for use with the Swagger UI
  - ensure `target='blank'` and `rel='noreferrer'` consistent with existing links in the UI 

#### Docs update 

update and simplify REST API page
- client auth and access tokens are already described in other pages, can remove these and just link
- remove bullet about interactive usage in the UI
- update bullet about the swagger reference as it is always up-to-date (it pulls from `main`) and can be versioned now as well

### Verification

Manually tested and confirmed the page works as intended. Screenshot of new API docs page:
Screenshot 2024-10-25 at 4 03 36 PM

### Future Work

1. [x] Downgrade React to fix a bunch of compat warnings (and unexpected errors like https://github.com/argoproj/argo-workflows/pull/13593) etc per above linked #13410, https://github.com/argoproj/argo-workflows/pull/13069#issuecomment-2159579279, etc
    - **EDIT**: Completed in #13821 
1. [x] Have the versioned API reference page pull its reference from the same version GH, instead of `main`
    - **EDIT**: Completed in #13824


tooptoop4:
if swagger ui is removed from each operator's server should there be an entry in breaking changes/upgrading guide?

agilgur5:
It is breaking, so we could, although the UI itself tells you what to do and there is no upgrade instruction otherwise. The impacted UI page having a message for a deprecation window for the Archived List UI could have made sense too, i.e. a better option than https://github.com/argoproj/argo-workflows/pull/13371

I'm inclined to agree and note it in this case, although we do want to be careful about making the UI a "breakable" surface area and where to draw boundaries as it's not quite the same as the schema / API / etc ["Public API"](https://argo-workflows.readthedocs.io/en/release-3.5/public-api/) (also I noticed metrics are missing there) but rather a consumer of those.

Joibel:
Could this be added to `docs/upgrading.md` to note the removal of the feature?

agilgur5:
Added an upgrading note in https://github.com/argoproj/argo-workflows/pull/13818/commits/d7062396bf6536da49558aea0ed689dd0ab7673c

Joibel (approved):
LGTM, thanks for sorting this one

MasonM:
@Joibel I can take this over if you'd like, though I don't have access to push to https://github.com/agilgur5/argo-workflows/tree/refactor-remove-swagger-ui (which I'm assuming I'll get when [my membership request](https://github.com/argoproj/argoproj/issues/327) goes through). 

I'm assuming the best way forward is to create a new PR with a branch forked from https://github.com/agilgur5/argo-workflows/tree/refactor-remove-swagger-ui. Does that sound reasonable?

MasonM:
@agilgur5 Thanks! It works, and I resolved the conflicts

terrytangyuan (approved):
Thanks!

PR #10945, opened 2023-04-19
fix: unable to connect cluster when AutomountServiceAccountToken is disabled. Fixes #10937

maxsxu:
Fixes #10937

Issue #10937, opened 2023-04-18
Unable to connect cluster when specified executor serviceAccountName


maxsxu:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

**What happened**

Occurred below errors while creating resource via container/script template

```
create-resources-via-container-template: error: failed to create serviceaccount: Post "http://localhost:8080/api/v1/namespaces/default/serviceaccounts?fieldManager=kubectl-create&fieldValidation=Strict": dia
l tcp 127.0.0.1:8080: connect: connection refused
```

**what you expected to happen?**

Able to create resource via container/script template successfully.

### Version

latest

### Paste a small workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
metadata:
  name: create-resources-via-container-template
  namespace: argo
spec:
  executor:
    serviceAccountName: sa-writer
  serviceAccountName: sa-writer
  automountServiceAccountToken: false
  entrypoint: main
  templates:
    - name: main
      container:
        name: main
        image: bitnami/kubectl
        command:
          - sh
        args:
          - -c
          - |
           kubectl create sa test
```

### Logs from the workflow controller

```text
healthy
```

### Logs from in your workflow's wait container

```text
healthy
```

maxsxu:
> Would you like to submit a PR to fix it?

@terrytangyuan 👍 Please assign this issue to me.

agilgur5:
Per https://github.com/argoproj/argo-workflows/issues/12848#issuecomment-2026546244, this seems to have been incorrectly analyzed; an SA token shouldn't be mounted if `automountServiceAccountToken: false` -- you have literally instructed it not to. An error trying to reach the k8s API with no SA token would be correct behavior.

PR #13820, opened 2024-10-26
fix: don't mount SA token when `automountServiceAccountToken: false`. Fixes #12848

### LINKED ISSUE #12848: `resource` template getting duplicate `exec-sa-token` volume mounts with `automountServiceAccountToken: false`
### Pre-requisites

Running a [workflow of workflows](https://argo-workflows.readthedocs.io/en/latest/workflow-of-workflows/) with `automountServiceAccountToken: false` results in a repeated `exec-sa-token` volume mount in the child pod spec.

When trying to run the child workflow, the pod spec validation fails with 

```
Pod "my-workflow-id-sub-id" is invalid: spec.containers[1].volumeMounts[1].mountPath: Invalid value: "/var/run/secrets/kubernetes.io/serviceaccount": must be unique
```

It appears that the [code in `workflowpod.go`](https://github.com/argoproj/argo-workflows/blob/v3.5.2/workflow/controller/workflowpod.go#L225-L238) is running multiple times for the same container in these cases.

### Version

v3.5.2

```YAML
---
kind: WorkflowTemplate
metadata:
  name: workflow-template-submittable
spec:
  automountServiceAccountToken: false # 👈
  executor:
    serviceAccountName: workflow-executor
  entrypoint: whalesay-template
  arguments:
    parameters:
      - name: message
        value: hello world
  templates:
    - name: whalesay-template
      inputs:
        parameters:
          - name: message
      container:
        image: docker/whalesay
        command: [cowsay]

---
kind: Workflow
metadata:
  generateName: workflow-of-workflows-
spec:
  automountServiceAccountToken: false # 👈
  executor:
    serviceAccountName: workflow-executor
  entrypoint: main
  templates:
    - name: main
      steps:
        - - name: workflow1
            template: resource-without-argument
            arguments:
              parameters:
              - name: workflowtemplate
                value: "workflow-template-submittable"

    - name: resource-without-argument
      inputs:
        parameters:
          - name: workflowtemplate
      resource:
        action: create
        manifest: |
          kind: Workflow
          metadata:
            generateName: workflow-of-workflows-1-
          spec:
            workflowTemplateRef:
              name: {{inputs.parameters.workflowtemplate}}
        successCondition: status.phase == Succeeded
        failureCondition: status.phase in (Failed, Error)
```

```text
{
  "level": "warning",
  "msg": "Non-transient error: Pod \"my-workflow-id-sub-id\" is invalid: spec.containers[1].volumeMounts[1].mountPath: Invalid value: \"/var/run/secrets/kubernetes.io/serviceaccount\": must be unique",
  "time": "..."
}
```

```text
n/a
```

MinyiZ:
Fixes #12848

### Motivation

https://github.com/argoproj/argo-workflows/pull/10945 introduced a regression as it is mounting a serviceaccount token on the main container when `automountServiceAccountToken: false`. 

The [security docs](https://argo-workflows.readthedocs.io/en/release-3.5/security/#workflow-pod-permissions) say 

> The main container will have the service account token mounted, allowing the main container to patch pods (among other permissions). Set `automountServiceAccountToken` to false to prevent this.

https://github.com/argoproj/argo-workflows/issues/10937 was not a bug. It is possible to explicitly mount a token to container and script template workflows even if `automountServiceAccountToken: false` so the code change was not needed.

### Modifications

- Reverts https://github.com/argoproj/argo-workflows/pull/10945
- Add test for `resource` template with `automountServiceAccountToken: false` and `executor.serviceAccountName` set

### Verification

Tests passed


agilgur5 (commented):
Regarding the tests here and from #10945, a `resource` template should correctly fail if `automountServiceAccountToken: false` since, well, there is no SA token for it to use

MinyiZ:
> Regarding the tests here and from #10945, a `resource` template should correctly fail if `automountServiceAccountToken: false` since, well, there is no SA token for it to use

I've fixed the tests from https://github.com/argoproj/argo-workflows/pull/10945 to demonstrate that it is possible to explicitly mount a token to container and script template workflows even if `automountServiceAccountToken: false` so the code change was not needed.

I believe `resource` templates shouldn't fail if `automountServiceAccountToken: false` because the exec token is mounted at [mainCtr := woc.newExecContainer(common.MainContainerName, tmpl)](https://github.com/argoproj/argo-workflows/blob/25bbb71cced32b671f9ad35f0ffd1f0ddb8226ee/workflow/controller/operator.go#L3453C13-L3453C33) and I've added an e2e test to demonstrate that.

MinyiZ:
All e2e tests are passing now. I'm not sure why the unit test `TestConfigMapCacheSaveOperate` in `operator.go` is failing. I will take a look later.

agilgur5:
> I believe `resource` templates shouldn't fail if `automountServiceAccountToken: false` because the exec token is mounted at

Ah nice, I wasn't sure if `resource` templates handled `executor.serviceAccountName` since they use a `main` container.

> I've fixed the tests from #10945 to demonstrate that it is possible to explicitly mount a token to container and script template workflows even if `automountServiceAccountToken: false` so the code change was not needed.

I saw that, but I don't think that test is very functional / necessary. Basically you're manually mounting the token instead of `automount`ing it. A user shouldn't really ever do that manually though, they should just `automount`. I think both those tests should just be removed

> [MinyiZ](https://github.com/MinyiZ) changed the title ~fix: don't mount SA token when automountServiceAccountToken: false. Fixes #12848~ fix: resource template getting duplicate exec-sa-token volume mounts with automountServiceAccountToken: false. Fixes #12848 [5 hours ago](https://github.com/argoproj/argo-workflows/pull/13820#event-14913607990)

Regarding the PR title, my change was intentional. The PR title should summarize what the PR does, not what the issue is. The PR title and issue title should not be equivalent.

MinyiZ:
> I think both those tests should just be removed.
Yep agreed done.

> Regarding the PR title, my change was intentional.
Ooops sorry about that. I unintentionally reverted your change when I was fixing up a typo. I've changed it back now.

agilgur5 (approved):
LGTM, thanks for taking a look and fixing this!

PR #13836, opened 2024-10-29
fix: use templateName where  possible else nothing

isubasinghe:
### Motivation

This change ensures that we directly use the templateName first. In the case that we do not have this field, we then omit the template name all together. 

### Verification
Not verified at the moment. Testing is a TODO.


Adrien-D (approved):
Looks good to me

Joibel (approved):

Issue #11391, opened 2023-07-18
UI: Env variable to hide yaml submission


tooptoop4:
Would be great to hide this button
https://github.com/argoproj/argo-workflows/blob/v3.4.8/ui/src/app/workflows/components/workflow-creator.tsx#L79-L84

Then can allow SSO users to just submit workflows from existing templates but not submit arbitrary new code!

agilgur5:
You can use the [Workflow Restrictions](https://argoproj.github.io/argo-workflows/workflow-restrictions/) feature to only allow users to submit via existing templates.

Removing the UI button would not necessarily suffice, as the API can still be accessed (e.g. someone could make a request from within their browser console). API permissions are based off RBAC, which is not specific enough to only allow certain kinds of Workflows.
If not using Workflow Restrictions, another way to workaround that would be to use policies / admission controllers to create more granular logic.

tooptoop4:
would workflow restrictions prevent an adhoc run of a cronworkflow that doesn't use a template?

agilgur5:
I believe it _should_ apply to `CronWorkflow`s as well, yes.

I've never tried submitting a `CronWorkflow` with no template though, so that's a good question and as such I'm not _100%_ confident of that.
But if Workflow Restrictions didn't apply to `CronWorkflow`s, I would consider that to be a bug. Can tag me if you find such a bug and I'd look into fixing it 🙂

agilgur5:
Ah right, I forgot, a `CronWorkflow` just [creates `Workflow`s](https://argoproj.github.io/argo-workflows/cron-workflows/#workflowspec-and-workflowmetadata) (see also [the code here](https://github.com/argoproj/argo-workflows/blob/6e209cdaa6d01180239551d6c13bf320d380b887/workflow/cron/operator.go#L94)), so yes, it will go through the same validation as a `Workflow`.

agilgur5:
This seems to have been a duplicate of #7326

agilgur5:
> This seems to have been a duplicate of #7326

Wrong issue, sorry

PR #13840, opened 2024-10-30
feat(ui): option to disable workflow editor (#11391)

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/11391

### DIFF
--- a/ui/src/workflows/components/workflow-creator.tsx
+++ b/ui/src/workflows/components/workflow-creator.tsx
@@ -21,6 +21,7 @@ export function WorkflowCreator({namespace, onCreate}: {namespace: string; onCre
     const [stage, setStage] = useState<Stage>('choose-method');
     const [workflow, setWorkflow] = useState<Workflow>();
     const [error, setError] = useState<Error>();
+    const isFullEditorDisabled = process.env.UI_DISABLE_FULL_EDITOR === 'true';
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


Joibel:
As discussed by agilgur5 in #11391 this doesn't prevent submission of arbitrary workflows and there is already a mechanism that does.

tooptoop4:
templateReferencing is too basic, doesn't allow certain superusers to not use template ie via cli

ui/src/workflows/components/workflow-creator.tsx
```diff
@@ -21,6 +21,7 @@ export function WorkflowCreator({namespace, onCreate}: {namespace: string; onCre
     const [stage, setStage] = useState<Stage>('choose-method');
     const [workflow, setWorkflow] = useState<Workflow>();
     const [error, setError] = useState<Error>();
+    const isFullEditorDisabled = process.env.UI_DISABLE_FULL_EDITOR === 'true';
```
MasonM, line 24: Unfortunately, this isn't going to work, because this file is executed by the browser, and `process` is only valid in a NodeJS context. There's no way for the browser to know what environment variables are defined at the server level without an API call.

How did you test this? I ask because you should have seen a "process is not defined" error immediately on loading the workflow creator. If you didn't see an error, then something is probably wrong with your local environment.

PR #13842, opened 2024-10-30
fix: make "Pod was rejected:" errors transient

### LINKED ISSUE #12572: Pod was rejected: The node had condition: [DiskPressure].

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/12572

### DIFF
--- a/util/errors/errors.go
+++ b/util/errors/errors.go
@@ -55,6 +55,7 @@ func isTransientErr(err error) bool {
 		apierr.IsServerTimeout(err) ||
 		apierr.IsServiceUnavailable(err) ||
 		isTransientEtcdErr(err) ||
+		isTransientPodRejectedErr(err) ||
 		matchTransientErrPattern(err) ||
 		errors.Is(err, NewErrTransient("")) ||
 		isTransientSqbErr(err)
@@ -88,6 +89,14 @@ func isTransientEtcdErr(err error) bool {
 	if strings.Contains(err.Error(), "etcdserver: leader changed") {
 		return true
 	} else if strings.Contains(err.Error(), "etcdserver: request timed out") {
+		return true
+	}
+	return false
+}
+
+func isTransientPodRejectedErr(err error) bool {
+	// This type of eviction happens before Pod could ever start
+	if strings.Contains(err.Error(), "Pod was rejected:") {
 		return true
 	}
 	return false
--- a/util/errors/errors.go
+++ b/util/errors/errors.go
@@ -96,10 +96,7 @@ func isTransientEtcdErr(err error) bool {
 
 func isTransientPodRejectedErr(err error) bool {
 	// This type of eviction happens before Pod could ever start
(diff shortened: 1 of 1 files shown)


tczhao:
This error happens when
- a node in your Kubernetes cluster runs out of disk space
- timing/logic between scheduler and kubelet, usage by other pods in the same node could cause issue when the pod got scheduled but failed to start

I don't believe this is a transient pattern as the cluster node could be struggling for real reason.
Many other error messages could match the "Pod was rejected" pattern, and should be handled properly at the kubelet end.
Argo wf user can use retry: Always to retry such pod

tooptoop4:
i don't want to retry for every type of error

akloss-terion (approved):

jholt96:
Any timeline on this getting merged? My company is impacted and would love to see this merged. Anything I can do to assist?

RobinDinkel:
What is blocking this PR from being merged? Setting TRANSIENT_ERROR_PATTERN for pods being rejected due to disk pressure (when there is no disk pressure) seems like a workaround.

Joibel:
The underlying problem here - #12572, pods rejected with "Pod was rejected: The node had condition: [DiskPressure]." never being retried - was fixed in v4.0.0 by #15086. The controller can now detect a pod that failed before its main container ever started and recreate it, rather than inferring intent from the error string. Enable it with:
```
    failedPodRestart:
      enabled: true
      maxRestarts: 3
```

See docs/pod-restarts.md. Restartable reasons are Evicted (which is what the kubelet sets for the DiskPressure admission rejection), NodeShutdown, NodeAffinity and UnexpectedAdmissionError.

That mechanism proves the pod never ran before restarting it, which addresses the objection raised here and on the issue that a substring match on "Pod was rejected:" is too broad - it also catches OutOfcpu, OutOfmemory and other admission failures where retrying blindly is not obviously safe.

Closing as superseded. If you are still seeing rejection reasons that aren't restarted (OutOfcpu / OutOfmemory in particular are not currently in the restartable set), please open an issue against isRestartableReason in workflow/controller/pod_restart.go and we can extend it there.

PR #13855, opened 2024-11-02
fix(operator): handle Unauthorized error. Partial fix for #12103

tooptoop4:
Partial fix for https://github.com/argoproj/argo-workflows/issues/12103


Joibel (changes requested):
The description of this change doesn't tell me why it's being done or how it was verified.

This change appears to make in some cases only, authorization errors be considered retryable.

Authorization errors being transient seems to be a bug in the kubernetes platform, and so not something that should be fixed here. Is there an upstream issue?

PR #13873, opened 2024-11-06
fix(ui): fallback to cluster logs when artifact logs not found. Fixes #12948 

### LINKED ISSUE #12948: " no artifact logs are available " when workflow is achived.
### Pre-requisites

We just upgrade the argo workflow from 3.4.4 to 3.5.5. We enabled archive 
```
persistence:
    connectionPool:
      maxIdleConns: 100
      maxOpenConns: 0
    # save the entire workflow into etcd and DB
    nodeStatusOffLoad: true
    # enable archiving of old workflows
    archive: true
    postgresql:
```
but didn't enable archive logs.
```
artifactRepository:
  # -- Archive the main container logs as an artifact
  archiveLogs: false

```
Before upgrade, we can see logs of the finished workflow (either success or fail) from UI(the server gets the log from pod I guess), but after upgrade, the UI will complain " no artifact logs are available "  and no logs returned.

Is it an expected result ? or is any configuration item controlling this behavior ?

### Version

v3.5.5

```YAML
any workflows
```

```text
```

```text
```

tooptoop4:
Fixes https://github.com/argoproj/argo-workflows/issues/12948

### DIFF
--- a/ui/src/shared/services/workflows-service.ts
+++ b/ui/src/shared/services/workflows-service.ts
@@ -253,21 +253,8 @@ export const WorkflowsService = {
         const getLogsFromArtifact = () => this.getContainerLogsFromArtifact(workflow, nodeId, container, grep, archived);
         const getLogsFromCluster = () => this.getContainerLogsFromCluster(workflow, podName, container, grep);
 
-        // If our workflow is archived, don't even bother inspecting the cluster for logs since it's likely
-        // that the Workflow and associated pods have been deleted
-        if (archived) {
-            return getLogsFromArtifact();
-        }
-
-        // return archived log if main container is finished and has artifact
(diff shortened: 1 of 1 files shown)


Joibel:
The underlying problem in #12948 - an archived-but-still-live workflow short-circuiting to artifact logs and reporting `no artifact logs are available` - was fixed by #14087 (released in v3.6.3), which replaced the `if (archived)` early return with `if (!isWorkflowInCluster(workflow))`. Archived workflows whose pods still exist now go down the cluster-log path with the usual artifact/cluster fallback, and #12948 has been closed as completed.

The lines this PR deletes no longer exist on main, and unconditionally preferring artifact logs feels like the wrong call. Closing as fixed

PR #13879, opened 2024-11-08
feat: add validating webhook for workflows

### LINKED ISSUE #13503: Validating Admission Webhook
# Summary

We currently have various scenarios where an invalid `Workflow` will be allowed in the cluster, causing bugs to be found much later instead of upfront. The CLI and API handle (some) validation, but direct to k8s will miss some of this.

## Use Cases

The lack of any validation when submitting directly to k8s can cause some unexpected scenarios like #6781 , #12693, #10630. In all those issues and some others, I've mentioned the concept of a [Validating Admission Webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) to prevent these scenarios and give clearer error messages: https://github.com/argoproj/argo-workflows/issues/12693#issuecomment-1980000601, https://github.com/argoproj/argo-workflows/issues/10630#issuecomment-2038271560, https://github.com/argoproj/argo-workflows/issues/6781#issuecomment-1984910847

Note that even if we do have a validating admission webhook, it may be optional, and it may not handle resources that were created before it existed, so the Controller and Informers would still have to be "tolerant" (the word the codebase uses) of invalid resources. This is primarily a feature to improve UX

## Implementation Details

There are 2 primary ways of doing Validating Admission in k8s these days:
1. [Admission Webhooks](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook)
   - This is simple in terms of _logic_, in that we'd just re-use the existing validation logic, either with the Controller or the Server (typically admission is a Controller responsibility)
   - Problematically, k8s webhooks require TLS certs that the k8s control plane trusts, which complicates Deployment quite a bit as it requires `cert-manager` etc.
     - See also #9737 that occurred after `cert-manager` was unintentionally required in the manifests for external-facing TLS certs (not internal ones).
     
1. [Admission Policies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) (stable as of k8s 1.30)
    - This is basically a policy engine built-in to k8s (instead of external like Kyverno) that supports CEL
    - This would require rewriting some validation in CEL, but is otherwise quite straightforward to deploy:
       1. Can be used [directly in-line with CRDs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules)
           - Problematically, our CRDs are already too large (see also #11266), which is why we have "minimal" CRDs that have no schema. Otherwise the JSON schema itself would already do some form of schema validation
       1. Or can be separate resources. Given the problems with CRD size, this might be the best possible solution 

I would probably suggest an optional of version of 1 for consistency, and then 2.b. as something deployed with all manifests as a baseline. We could even start making simple policies already and add them to the manifests and then gradually build them up

---

williamburgson:
Fixes #13503

### Motivation

validate workflow definition before they reach controller

### Modifications

TODO


Joibel:
@williamburgson I was actually looking at how we could adopt this last week - would it be OK to continue to use your code and credit you if we do?

Issue #13848, opened 2024-10-31
Mark memoized node as cached


MenD32:
# Summary

Memoized nodes' status should reflect that they are memoized in the UI. 

## Use Cases

When would you use this?

when using memoization, its much easier to debug workflows when you can see that they have been cached. currently in large workflows you'd have to go over all the nodes one by one to understand if they are cached, or you could go timeline tab.
both of those options offer a bad UX, that becomes worse as the workflows get larger. 

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

Joibel:
Design ideas or PR welcome

shuangkun:
Looking forward to seeing this feat

PR #13883, opened 2024-11-08
feat: mark memoized node as cached

MenD32:
Fixes https://github.com/argoproj/argo-workflows/issues/13848

### Motivation

when using memoization, its much easier to debug workflows when you can see that nodes have been cached. currently in large workflows you'd have to go over all the nodes one by one to understand if they are cached, or you could go to the timeline tab.
both of those options offer a bad user experience, that becomes worse as the workflows get larger.
   

### Modifications

changed the icon of successful cached nodes to a database icon (from font awesome) to mark them as cached.

### Verification

tested in the UI that it works using the examples/memoized-simple.yaml example

#### screenshots:
first workflow (no cache hit)
Screenshot 2024-11-08 at 14 53 33

second workflow (cache hit)
Screenshot 2024-11-08 at 14 53 50


elliotgunton:
I like the choice of icon - for the abstract concept of "cached" it seems to align with the rest of the world https://www.google.com/search?q=icon+to+represent+cache

But replacing the check mark within the node itself might be confusing for existing and new users alike - it looks like it's a step related to database access, and two workflows running the same thing will look quite different. I would suggest an additive approach e.g. an overlaid database symbol - quick sketch from a google slide

image

Joibel (changes requested):
I agree with https://github.com/argoproj/argo-workflows/pull/13883#issuecomment-2501010877, that the complete change of icon isn't good UX, and would prefer something similar to @elliotgunton's suggestion.

MenD32:
I've added the requested change, I'm not much of a UI UX designer so if you have any changes you'd like to have to the colors and what not, please share

MenD32:
Screenshot 2024-11-26 at 18 21 49

this is the new icon

Adrien-D (approved):
Approving because the phase will still always be `Succeeded`

Joibel (approved):
Awesome, thanks for iterating on this.

PR #13885, opened 2024-11-08
chore: update kit reference in Makefile

tooptoop4:
Without this on mac the wrong kit project was being installed (https://github.com/johnlindquist/kit instead of https://github.com/kitproj/kit)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -533,12 +533,7 @@ dist/argosay:
 .PHONY: kit
 kit: Makefile
 ifeq ($(shell command -v kit),)
-ifeq ($(shell uname),Darwin)
-	brew tap kitproj/kit --custom-remote https://github.com/kitproj/kit
-	brew install kit
-else
 	curl -q https://raw.githubusercontent.com/kitproj/kit/main/install.sh | tag=v0.1.8 sh
-endif
 endif
 
 


Makefile
```diff
@@ -533,13 +533,8 @@ dist/argosay:
 .PHONY: kit
 kit: Makefile
 ifeq ($(shell command -v kit),)
-ifeq ($(shell uname),Darwin)
-	brew tap kitproj/kit --custom-remote https://github.com/kitproj/kit
-	brew install kit
```
agilgur5, line 538: You can instead do `brew install kitproj/kit/kit` per the [Brew docs on duplicate names](https://docs.brew.sh/Taps#formula-with-duplicate-names)

Although personally I'm not particularly a fan of the `Makefile` installing system deps (such as via Brew or `pip` or `npm -g`) in the first place

radusora: It seems the Homebrew formula is currently missing from the main branch (see: [commit 7508171b](https://github.com/kitproj/kit/commit/7508171b6ef2c2ea8e8e67f8a2b17a2822dea94c)). If the plan is to avoid installing prerequisites through the CLI, we should consider removing the corresponding make target, any associated references, and updating the contributor documentation accordingly. Alternatively, if we intend to keep it, we should update it to use the fully qualified formula name and include the appropriate tag to align with the non-Darwin section.

Joibel:
We're using tilt now. Closing

PR #13891, opened 2024-11-11
chore(log): suppress sql warnings (such as upper: slow query)

tooptoop4:
Reduces excess logging (which can $)

Related to https://github.com/upper/db/issues/688


MasonM:
I find the slow query logs very useful, so I don't think we should be suppressing it. You already have the ability to suppress it by setting the `UPPER_DB_LOG` environment variable; no code changes needed.

tooptoop4:
UPPER_DB_LOG=ERROR not having any effect

PR #11446, opened 2023-07-25
Draft for a new feature to support Secrets as a type of parameters

umi0410:
Related to #5506

### Disclaimer

This is just a draft, and the purpose of this PR is to look for a mentor who could help me with detailed implementations. As the test code has not been modified, the tests will fail, and this PR lacks test code at moment.

### Motivation

Currently, we can use parameters referring to ConfigMaps but cannot use parameters referring to Secrets.

### Modifications

I added a new type of parameters that refer to Secrets based on the code for the case where a parameter refers to ConfigMaps.

### Verification

1. Create a Secret which will be referred to by a Workflow.
```
kubectl create secret generic workflow-parameter --from-literal credential=THIS_IS_SAMPLE_CREDENTIAL
```
2. Create a Workflow whose parameter refers to the Secret.
```
❯ cat << EOF | argo submit --log -
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: secret-parameter-
spec:
  entrypoint: echo
  arguments:
    parameters:
    - name: credential
      valueFrom:
         secretKeyRef:
           name: workflow-parameter
           key: credential
  templates:
  - name: echo
    inputs:
      parameters:
      - name: credential
    container:
      image: alpine:3
      command: [echo]
      args: ["Parameter: {{inputs.parameters.credential}}"]
EOF
```
3. In the logs, you can see the parameter has been passed to the Workflow and its Pod. In this example, the value of the parameter is `THIS_IS_SAMPLE_CREDENTIAL`.
```
Name:                secret-parameter-vhbln
Namespace:           argo
ServiceAccount:      unset (will run with the default ServiceAccount)
Status:              Pending
Created:             Sun Jul 23 21:36:50 +0900 (now)
Progress:
Parameters:
secret-parameter-vhbln: time="2023-07-23T12:36:52.292Z" level=info msg="capturing logs" argo=true
secret-parameter-vhbln: Parameter: THIS_IS_SAMPLE_CREDENTIAL
secret-parameter-vhbln: time="2023-07-23T12:36:53.304Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

#### 1. Credentials are not redacted properly.
For the current implementation, the plain value of secrets are shown in logs and Pod manifests as follows:
```
❯ kubectl logs secret-parameter-vhbln -c init | grep THIS_IS_SAMPLE_CREDENTIAL
[...]

PR #13899, opened 2024-11-14
feat: Support loading parameter values from secrets. Fixes: #5506

### LINKED ISSUE #5506: Passing k8s secret to a workflow as input
# Summary

I know it is possible to pass a secret to a container via ENV or FILE. but it is not possible to pass it as an argument of a workflow.

# Use Cases

I have a few tools which accept values as command line args. These args are values which I'm not convenient in putting in YAMLs as plaintext and would like to store as k8s secrets.

---

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

shuangkun:
Fixes #5506  Reference: https://github.com/argoproj/argo-workflows/pull/11446

Unsolved issues：

Hide secret value in argo template, log, UI

### Verification

Loccal Test and UT


terrytangyuan (commented):
Can you paste some example outputs and screenshots?

shuangkun:
> Can you paste some example outputs and screenshots?

Sure. 

1) deploy a secret
```
apiVersion: v1
kind: Secret
metadata:
  name: simple-parameters
  labels:
    # Note that this label is required for the informer to detect this Secret.
    workflows.argoproj.io/secret-type: Parameter
data:
  msg: "aGVsbG8gd29ybGQK"
```
```
opensource % echo "aGVsbG8gd29ybGQK" | base64 -d
hello world
```
2) Then we can submit a workflow to reference this secret use new cli.
```
kind: Workflow
metadata:
  generateName: global-parameter-values-from-secret-
  labels:
  annotations:
      This example demonstrates loading global parameter values from a Secret.
      Note that the "simple-parameters" Secret (defined in `examples/secrets/simple-parameters-secret.yaml`) needs to be created first before submitting this workflow.
spec:
  entrypoint: print-message
  # Parameters can also be passed via secret reference.
  arguments:
    parameters:
      - name: message
        valueFrom:
          secretKeyRef:
            name: simple-parameters
            key: msg

  templates:
    - name: print-message
      container:
        image: busybox
        command: ["echo"]
        args: ["{{workflow.parameters.message}}"]
```
```
tianshuangkun@U-4YKHFNR6-2229 opensource % /Users/tianshuangkun/go/src/github.com/argoproj/argo-workflows/dist/argo submit test.yaml
Name:                global-parameter-values-from-secret-68jb8
Namespace:           argo
Status:              Pending
Created:             Sun Jan 05 12:31:10 +0800 (now)
Progress:
Parameters:
```
3) View the results and successfully reference the secret
```
tianshuangkun@U-4YKHFNR6-2229 opensource % /Users/tianshuangkun/go/src/github.com/argoproj/argo-workflows/dist/argo get @latest
Name:                global-parameter-values-from-secret-rtfml
Namespace:           argo
Conditions:
 PodRunning          False
 Completed           True
Created:             Sun Jan 05 12:33:22 +0800 (26 seconds ago)
Started:             Sun Jan 05 12:33:22 +0800 (26 seconds ago)
Finished:            Sun Jan 05 12:33:30 +0800 (18 seconds ago)
Duration:            8 seconds
Progress:            1/1
ResourcesDuration:   2s*(100Mi memory),0s*(1 cpu)
Parameters:

STEP                                          TEMPLATE       PODNAME                                    DURATION  MESSAGE
 ✔ global-parameter-values-from-secret-rtfml  print-message  global-parameter-values-from-secret-rtfml  5s
tianshuangkun@U-4YKHFNR6-2229 opensource % /Users/tianshuangkun/go/src/github.com/argoproj/argo-workflows/dist/argo logs @latest
global-parameter-values-from-secret-rtfml: time="2025-01-05T04:33:26.162Z" level=info msg="capturing logs" argo=true
global-parameter-values-from-secret-rtfml: time="2025-01-05T04:33:26.163Z" level=debug msg="ignore signal child exited" argo=true
global-parameter-values-from-secret-rtfml: hello world
global-parameter-values-from-secret-rtfml:
global-parameter-values-from-secret-rtfml: time="2025-01-05T04:33:26.171Z" level=debug msg="ignore signal child exited" argo=true
global-parameter-values-from-secret-rtfml: time="2025-01-05T04:33:27.164Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

manifests/cluster-install/workflow-controller-rbac/workflow-controller-role.yaml
```diff
@@ -17,4 +17,5 @@ rules:
       - secrets
     verbs:
       - get
+      - watch
```
terrytangyuan, line 20: This introduces a new requirement. @jessesuen Would you be okay with this addition?

jessesuen: This would be equivalent to adding `list`, which is expanding the scope such that the controller see all secrets in the namespace. (Actually, I'm now wondering why `list` wasn't added).

Is this Role used only for the controller's namespace (e.g. `argo`)?

Joibel:
I am uncomfortable with this change without quite a lot more safety.

Things I dislike:
* It has no documentation.
  - We should document you can do this, and the risks associated with allowing it
* @shuangkun suggests it needs work on redacting, and I don't want to merge this portion without that work being done
  - I'm happy/would encourage that to be in a separate PR for readability
* It should be possible to run the controller without the new permissions
  - this needs a test
* The default permissions and manifests should not change to allow this

I think this is a problematic enough change that some kind of proposal should have been written prior to any implementation so we could have discussed if we even wanted to support this. I'm unsure we do as I can't think of use cases where this isn't better supported by the alternatives or workarounds.

PR #13750, opened 2024-10-13
fix(ui): switch from js-yaml to yaml. Fixes #12205

MasonM:
Fixes #12205

### Motivation

[js-yaml](https://www.npmjs.com/package/js-yaml) has had no releases in over 3 years. Also, the fact it only supports YAML 1.2 is leading to bugs, since Kubernetes uses YAML 1.1 (see https://github.com/kubernetes/kubernetes/issues/34146).

### Modifications

This switches over to [yaml](https://www.npmjs.com/package/yaml), which had a release last month, is [licensed under ISC](https://github.com/eemeli/yaml/blob/main/LICENSE), and supports YAML 1.1. I also added test cases for the parsing/stringifying code.

### Verification
Copied the example from https://github.com/argoproj/argo-workflows/issues/12205 over to http://localhost:8080/workflow-templates and verified the `mode: 0755` was changed to `mode: 493`:

https://github.com/user-attachments/assets/54690403-4189-499a-a2c1-65293026ea77

PR #13913, opened 2024-11-17
fix(ui): fix broken workflowtemplate submit button. Fixes #13892

### LINKED ISSUE #13892: UI (when 2 tabs open) 3.6.0rc4 - submit workflowtemplate greyed out
### Pre-requisites

in the ui it lets me create a new workflowtemplate and update it, but the button to submit it is greyed out. i'm using the quickstart

### Version(s)

3.6.0rc4

```YAML
default
```

```text

n/a
```

```text

n/a
```

instauro:
Fixes #13892 

### Motivation

Currently the submit button for workflowtemplates is broken both in main and the new 3.6.0 release. The reason is that the UI thinks the content of the workflowtemplate has been updated in the editor, even when it has not.

The process looks like this:
We fetch the workflow from the backend and keep it as a javascript object, convert it to a yaml string, update the editor content with this string, and then when the editor content is updated we convert it back to a javascript object. We then compare to original object with the parsed object to determine if it has been updated or not. This requires parse and stringify to be inverses, but when we downgraded yaml version from 1.2 to 1.1 in https://github.com/argoproj/argo-workflows/pull/13750 we accidently broke this assumption, because we are now parsing as 1.1 but stringifying as 1.2. This fails when we are dealing with some fields like creationTimestamp, where these versions seem to differ.

### Modifications

This PR sets the yaml version as 1.1 for both parse and stringify.

### Verification

I ran the UI locally in the devcontainer with the fix and submitted workflowtemplates and clusterworkflowtemplates. I also added a test which fails before the fix but passes after the fix.

### DIFF
--- a/ui/src/shared/components/object-parser.test.ts
+++ b/ui/src/shared/components/object-parser.test.ts
@@ -1,6 +1,15 @@
 import {exampleWorkflowTemplate} from '../examples';
 import {parse, stringify} from './object-parser';
 
+describe('parse and stringify are inverses', () => {
+    it('handles creationTimestamp', () => {
+        const testWorkflowTemplate = exampleWorkflowTemplate('test');
+        testWorkflowTemplate.metadata.creationTimestamp = '2024-11-17T06:56:52Z';
+        expect(parse(stringify(testWorkflowTemplate, 'yaml'))).toEqual(testWorkflowTemplate);
+        expect(parse(stringify(testWorkflowTemplate, 'json'))).toEqual(testWorkflowTemplate);
+    });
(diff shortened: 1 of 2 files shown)


MasonM (approved):
Good catch! I entered https://github.com/argoproj/argo-workflows/pull/13915 that fixes this by eliminating the unnecessary `stringify()` calls, but that's orthogonal to this fix, and I think both should be merged.

isubasinghe (approved):
Thanks, this makes sense, I am glad this isn't to do with React.

PR #13916, opened 2024-11-19
feat: working on migratting off logrus in controller

isubasinghe:


MasonM:
You're probably already aware, but one feature of logrus we need is spying on the logging statements in unit tests using https://pkg.go.dev/github.com/sirupsen/logrus/hooks/test. Example: https://github.com/argoproj/argo-workflows/blob/2fd54884844bb76d760466027afa023c5bfd6b64/util/cmd/cmd_test.go#L145-L147

To replicate that, I think the `Logger` interface needs a notion of hooks

isubasinghe:
Thanks @MasonM, I do plan to tackle this hook issue eventually, I just want to think about the hook thing a bit first before I change the interface, mainly I don't want to add this feature if its only used for tests and I will try to seek a workaround first.

Joibel:
Replaced by #14527

PR #13922, opened 2024-11-20
feat(ui): prefill parameters for workflow submit form. Fixes #12124

### LINKED ISSUE #12124: Prefill parameters for the workflow submit form

sairam91:
Fixes #12124

### Motivation
This PR  adds support for developers to prefil submit workflow form by passing in the values through query parameters.

To prefil a submit workflow developers can pass in the following query parameters

```
sidePanel=submit-new-workflow
template=<template that the workflow should follow>
parameters[key]=value // Here the key would be the parameter for the job and value would be the value of the parameter.
You should provide the keys that are supported by the specified template
```

### Modifications
- Updated `workflow-list` component
- Updated `submit-workflow-panel.tsx` component to default entry point if provided.
- Updated `workflow-creator.tsx` to read default values from query parameters.
 - The workflow will be prefilled with the parameters that the template supports
- Added new shared helper function to parse query parameter of the format `?parameters[key]=value`.

### Verification

✅ Visit http://localhost:8080/workflows?namespace=argo&limit=50 and create a new workflow and ensure the default values are populated.
Screenshot 2024-11-19 at 9 56 16 PM

✅  Visit the [prefilled URL](http://localhost:8080/workflows?namespace=argo&limit=50&sidePanel=submit-new-workflow&template=delightful-dragon&parameters%5Bmessage%5D=test-hello&parameters%5Blimit%5D=2000) and ensure that the values from url query are populated.
Screenshot 2024-11-19 at 9 56 33 PM

### DIFF
--- a/ui/src/cluster-workflow-templates/cluster-workflow-template-details.tsx
+++ b/ui/src/cluster-workflow-templates/cluster-workflow-template-details.tsx
@@ -150,6 +150,7 @@ export function ClusterWorkflowTemplateDetails({history, location, match}: Route
                         entrypoint={template.spec.entrypoint}
                         templates={template.spec.templates || []}
                         workflowParameters={template.spec.arguments.parameters || []}
+                        history={history}
                     />
                 </SlidingPanel>
             )}
--- a/ui/src/shared/get_workflow_params.test.ts
+++ b/ui/src/shared/get_workflow_params.test.ts
@@ -1,0 +1,38 @@
+import {createBrowserHistory} from 'history';
+
+import {getWorkflowParametersFromQuery} from './get_workflow_params';
+
+describe('get_workflow_params', () => {
+    it('should return an empty object when there are no query parameters', () => {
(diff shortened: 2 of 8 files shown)


MasonM (approved):
Good job!

terrytangyuan (approved):
Great feature! Thank you for your contribution!

PR #13923, opened 2024-11-21
fix(ui): Clickable URLs are messing up formatting in the UI

Adrien-D:
### Motivation

Since [this change](https://github.com/argoproj/argo-workflows/pull/13494/files), URLs are now clickable but splitted in columns which is not the desired behavior

Screenshot 2024-11-21 at 16 28 43
Screenshot 2024-11-21 at 16 28 36

### Modifications

I wrapped the `LinkifiedText` component with a simple span to leave the text on the same line

### Verification

Screenshot 2024-11-21 at 16 28 55


MasonM (approved):
Verified locally

Joibel (approved):
LGTM, thanks Adrien

Issue #6781, opened 2021-09-21
`argo cron` should report when CronWorkflow names exceed the length limit


crenshaw-dev:
# Summary

The `argo cron` command faithfully lists CronWorkflows, even ones that are rendered invalid due to the name exceeding the 52-character limit.

The workflow controller does log the issue. But `argo cron` gives the impression that nothing is wrong.

# Use Cases

My team would benefit from this when a dev (who is unfamiliar with the workflow controller and its logs) deploys a CronWorkflow and then uses `argo cron` to maintain that CronWorkflow.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

tyrken:
No error seen in Argo Web UI either, and it's only in the workflow controller logs when first created, not when you expect it to be scheduled.

terrytangyuan:
Would anyone want to submit a PR for this?

tyrken:
Sure, but as a non-golang, non-JS dev it might be a little clunky.  Guessing I'd need to add a generic `message` field to the status like Workflow has, or maybe better a `NameInvalid` condition given you already seem to show that in the Web UI at least.

agilgur5:
Related: #10264 for the UI.

It would be nice if the Controller could _statically_ reject a `CronWorkflow` if it can tell that its generated name would be too long. That would prevent issues like this from happening in the first place.

bkanuka:
My team just spent far too long diagnosing why a cron workflow was not running.  Everything in the UI looked good, including saying that the workflow would run in X minutes and then it just _wouldn't_. 

ArgoCD (which deploys the CronWorkflows) also showed that CronWorkflow was properly deployed.

I believe this is actually a pretty high priority issue because it (almost) silently prevents a CronWorkflow from running, while simultaneously indicating that all is okay and that it _will_ run.  I completely agree that the CronWorkflow manifest should have been rejected when submitted to the cluster.

@alexec @terrytangyuan could you take another look at this issue and adjust labels if you see fit?  I'm happy to provide more details if it helps.

agilgur5:
> could you take another look at this issue and adjust labels if you see fit?

the labels are correct. we do not currently have a priority label for features.

> ArgoCD (which deploys the CronWorkflows) also showed that CronWorkflow was properly deployed.

Did you use Server-Side Apply for this? Because there is also https://github.com/argoproj/argo-cd/issues/16817 when using client-side apply

[...]

PR #13926, opened 2024-11-22
feat(controller): show invalid spec on cronwf res

carlosrodfern:
Fixes: #6781

### Motivation

Help the user see the validation errors on the cronwf resources early, before they find out later that the workflow didn't trigger.

### Modifications

Add the event to the description of the cronwf, e.g.

```
Events:
  Type     Reason     Age   From                 Message
  ----     ------     ----  ----                 -------
  Warning  Invalid  14s   workflow-controller  cron workflow name "test-cron-wf-very-very-very-very-very-very-very-very-very-very-very-very-long" must not be more than 52 characters long (currently 77)
```

### Verification

I created my own image of `workflow-controller` and deployed it in a local k3s kubernetes with argo-workflow.

I used this artifact as a test:

```yaml
kind: CronWorkflow
metadata:
  name: test-cron-wf-very-very-very-very-very-very-very-very-very-very-very-very-long
spec:
  schedule: "5 * * * *"
  concurrencyPolicy: "Replace"
  startingDeadlineSeconds: 0
  workflowSpec:
    entrypoint: date
    templates:
    - name: date
      container:
        image: alpine:3.6
        command: [sh, -c]
        args: ["date; sleep 90"]
```

### DIFF
--- a/workflow/cron/controller.go
+++ b/workflow/cron/controller.go
@@ -179,6 +179,7 @@ func (cc *Controller) processNextCronItem(ctx context.Context) bool {
 
 	err = cronWorkflowOperationCtx.validateCronWorkflow(ctx)
 	if err != nil {
+		cc.eventRecorderManager.Get(un.GetNamespace()).Event(un, apiv1.EventTypeWarning, "Invalid", err.Error())
 		logCtx.WithError(err).Error("invalid cron workflow")
 		return true
 	}


carlosrodfern:
I'm still investigating the possibiilty to actually make it a status, e.g. `Ready` if the cronworkflow is good, or something else otherwise. The goal is that if I deploy cronworkflows with ArgoCD, for example, the health status of the cronworkflow would be failed because of invalid spec, and ArgoCD would mark the app syncing as failed.

tooptoop4 (changes requested):
if woc.controller.Config.WorkflowEvents.IsEnabled() { like https://github.com/argoproj/argo-workflows/pull/13746/files

Joibel:
There is already [a metric](https://argo-workflows.readthedocs.io/en/release-3.6/metrics/#error_count) for this.

The best solution is an admission controller which will reject them. This is started in #13879

carlosrodfern:
> There is already [a metric](https://argo-workflows.readthedocs.io/en/release-3.6/metrics/#error_count) for this.
> 
> The best solution is an admission controller which will reject them. This is started in #13879

That looks like the best solution. That specific PR is for the `Workflow` artifact. We would need another validating webhook for `CronWorkflow`, and even for the `WorkflowTemplate`s as well, correct?

Joibel:
> That looks like the best solution. That specific PR is for the `Workflow` artifact. We would need another validating webhook for `CronWorkflow`, and even for the `WorkflowTemplate`s as well, correct?

Yes. Initially getting it in with the framework is the hard part, but then it should be extended to cover those two,  ClusterWFTemplates, and EventBindings to cover all the "user facing" custom resources.

carlosrodfern:
> > That looks like the best solution. That specific PR is for the `Workflow` artifact. We would need another validating webhook for `CronWorkflow`, and even for the `WorkflowTemplate`s as well, correct?
> 
> Yes. Initially getting it in with the framework is the hard part, but then it should be extended to cover those two, ClusterWFTemplates, and EventBindings to cover all the "user facing" custom resources.

Once merged, I'll work on a PR for at least the cronworkflow one.

> > That looks like the best solution. That specific PR is for the `Workflow` artifact. We would need another validating webhook for `CronWorkflow`, and even for the `WorkflowTemplate`s as well, correct?
> 
> Yes. Initially getting it in with the framework is the hard part, but then it should be extended to cover those two, ClusterWFTemplates, and EventBindings to cover all the "user facing" custom resources.

Awesome :). I'll close this PR then, and I'll just watch the one being worked on, to then bring a new PR extending the new webhook :+1:

PR #13915, opened 2024-11-18
fix(ui): improve editor performance and fix Submit button. Fixes #13892

MasonM:
Fixes #13892

### Motivation
The `<ObjectEditor>` component used by all the edit pages encapsulates the logic for serializing/deserializing the object being edited, but it's extremely inefficient. For example, if you're editing a `WorkflowTemplate`, every single keystroke will cause [parse()](https://github.com/argoproj/argo-workflows/blob/6da0ff825cf318f518f640b5415db4df6b26656d/ui/src/shared/components/object-parser.ts#L3) to be called twice and [stringify()](https://github.com/argoproj/argo-workflows/blob/6da0ff825cf318f518f640b5415db4df6b26656d/ui/src/shared/components/object-parser.ts#L15) to be called 4 times. For large documents, this serialization/deserialization overhead adds up quickly.

Additionally, the "Submit" button is sometimes incorrectly grayed out, e.g. after clicking the "Update" button. This is happening because `useEditableObject()` is using reference comparisons on the object being edited, since it doesn't have access to the serialized string version encapsulated by `<ObjectEditor>`.

### Modifications

This decouples `<ObjectEditor>` from the serialization/deseralization logic and lifts the state involved up to the `useEditableObject()` hook, as discussed in https://github.com/argoproj/argo-workflows/pull/13593#discussion_r1781972661. Doing so eliminates unnecessary serialization/deseralization: now, if you're editing a `WorkflowTemplate`, each keystroke will only call `parse()` once and won't call `stringify()` at all. This also fixes issues with the Submit button not reflecting the current page state, since `useEditableObject()` now uses string comparisons instead of ref comparisons.

Also, I did some minor cleanup:
* Delete `ui/src/cluster-workflow-templates/cluster-workflow-template-editor.tsx` because it was identical to `ui/src/workflow-templates/workflow-template-editor.tsx`
* Delete `ui/src/shared/components/resource-editor/resource.scss` because it stopped being used in #4470 
* Delete `ui/src/shared/components/resource-editor/resource-editor.tsx`, which was a thin wrapper around `<ObjectEditor>`, because the only files using it weren't actually passing any of the props (e.g. `onSubmit`) that enabled the extra functionality.
### Verification
Followed these steps to verify on each of the following pages: http://localhost:8080/workflow-templates, http://localhost:8080/cluster-workflow-templates, http://localhost:8080/cron-workflows, http://localhost:8080/event-sources, and http://localhost:8080/sensors

[...]

PR #13931, opened 2024-11-23
fix(ui): handle parsing errors properly in object editor

MasonM:
### Motivation
https://github.com/argoproj/argo-workflows/pull/13915 introduced a regression where edits made in an object editor that introduce syntax errors cause a full screen error. 
image

This is happening because parsing was lifted out of `<ObjectEditor>` and into a reducer. Before https://github.com/argoproj/argo-workflows/pull/13915, the object editor used the following to catch parse errors: https://github.com/argoproj/argo-workflows/blob/eb4f2456e077d89324ea71ef5f5e92cdccd8157a/ui/src/shared/components/object-editor.tsx#L131-L136

After  https://github.com/argoproj/argo-workflows/pull/13915, the parsing logic was moved to `reducer()`, which is executed async, and therefore errors aren't propagated to the `catch (e) {}` block:
https://github.com/argoproj/argo-workflows/blob/1392ef516fd69a2cae875d746c1155f940e4948c/ui/src/shared/components/object-editor.tsx#L117-L122

### Modifications

This moves the parsing logic to the `setObject()` function (which gets passed as `onChange` to `<ObjectEditor>`), which means errors will now be caught by the `catch (e) {}` block.

### Verification

Tested by going http://localhost:8080/workflow-templates, creating a new `WorkflowTemplate`, and making edits.

https://github.com/user-attachments/assets/df1b774b-6614-4096-884d-534333fc4f46

### DIFF
--- a/ui/src/shared/use-editable-object.test.ts
+++ b/ui/src/shared/use-editable-object.test.ts
@@ -1,4 +1,4 @@
-import {createInitialState, reducer} from './use-editable-object';
+import {createInitialState, reducer, setObjectActionCreator} from './use-editable-object';
 
 describe('createInitialState', () => {
     test('without object', () => {
@@ -59,8 +59,8 @@ describe('reducer', () => {
         });
     });
 
-    test('setObject with string', () => {
-        const newState = reducer(testState, {type: 'setObject', payload: 'a: 2'});
+    test('setObject with object and string', () => {
+        const newState = reducer(testState, {type: 'setObject', payload: {object: {a: 2}, serialization: 'a: 2'}});
         expect(newState).toEqual({
             object: {a: 2},
             serialization: 'a: 2',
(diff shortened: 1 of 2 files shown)


Adrien-D:
Tested on local env and it works well

Do you think its also possible to have the `try/catch` logic in the reducer and return a `parsingError` state ? So the hook would be 'safe' to use by itself without throwing an error ?

MasonM:
@Adrien-D The problem is the parsing error needs to be propagated to `<ObjectEditor>` at the time `onChange()` is called, and I don't see any way of doing that when the parsing logic is in the reducer without lifting up the `error` state, which would involve a lot more changes.

Adrien-D:
@MasonM yes that would require more work and a bit of refactoring, its more of a design thinking, kind of weird to me to have a hook that can throw an error. Your changes LGTM though

Adrien-D (approved):

Joibel (approved):

PR #13943, opened 2024-11-26
refactor: autogenerate metrics docs

Joibel:
### Motivation

As mentioned at some point during the opentelemetry metrics changes, it would be nice if documentation could be autogenerated.

This PR does this so that there is a single source of truth for metrics and attributes.

This makes attribute documentation consistent, and the docs match the implementation in more places.

It will help when we do something about server metrics, and also with tracing (as tracing will want managed attributes and auto-generated docs).

### Modifications

Create a new yaml file `util/telemetry/builder/values.yaml` which contains the attributes and metrics.

`util/telemetry/builder/builder.go` converts this into go structs and documentation which is inserted into the metrics.md docs. The go structs implement a new interface for builtin metrics.

This code change has the downside of making the shared utils directory aware of every metric again.

This fixes a few things:
* The `HELP` for prometheus metrics will now exactly match the first line in the documentation.
* The documentation for metrics are guaranteed to be in alphabetical order.
* `workers_busy_count` has the correct attribute name (this is a legacy metric - the implementation matched 3.5 which is correct, but the docs did not match the implementation of having a different attribute name from the other queue metrics).

This can be further improved by enforcing the attributes which may mismatch at the moment, and removing all of the calls to `Name()` by instead allowing the operators to operate on the builtin metrics. This is left for a separate PR. This will also optimise the calls to the operators by pre-creating the attributes.

### Verification

Unit and integration tests.

Some inspection of prometheus metrics from before and after.


docs/metrics.md
```diff
@@ -421,101 +428,188 @@ This and associated metrics are all directly sourced from the [client-go workque
...
 
 A counter of workflows that have entered each phase for tracking them through their life-cycle, by namespace.
 
-| attribute   | explanation                                    |
-|-------------|------------------------------------------------|
-| `phase`     | The phase that the workflow has entered        |
+| attribute | explanation |
+|-----------|-------------|
+| `phase` | The phase that the workflow has entered |
 | `namespace` | The namespace in which the workflow is running |
```
tico24, line 544: ... was run?

docs/metrics.md
```diff
@@ -242,38 +242,38 @@ Metrics for the [Four Golden Signals](https://sre.google/sre-book/monitoring-dis
...
 🚨 This counter may go up much more than once for a single use of the feature.
 
-| attribute   | explanation                                 |
-|-------------|---------------------------------------------|
-| `feature`   | The name of the feature used                |
-| `namespace` | The namespace of the item using the feature |
+| attribute | explanation |
+|-----------|-------------|
+| `feature` | The name of the feature used |
+| `namespace` | The namespace in which the workflow is running |
```
tico24, line 276: was run?

docs/metrics.md
```diff
@@ -282,129 +282,136 @@ A counter which goes up when a feature which is [deprecated](deprecations.md) is
...
+
+| attribute | explanation |
+|-----------|-------------|
+| `phase` | The phase that the pod is in |
 
 #### `pods_total_count`
 
-A gauge of the number of pods which have entered each phase and then observed by the controller.
-This is not directly controlled by the workflow controller, so it is possible for some pod phases to be missed.
+Total number of Pods that have entered each phase.
```
tico24, line 396: we doing uppercase P for pods or not?

tico24 (changes requested):
nits commented.

I'm slightly surprised that the new markdown tables are passing CI. I thought there was a table linter? They aren't nice and pretty like they were before.

Makefile
```diff
@@ -632,8 +632,21 @@ clean:
 	go clean
 	rm -Rf test-results node_modules vendor v2 v3 argoexec-linux-amd64 dist/* ui/dist
 
-# swagger
+# Built telemetry files
```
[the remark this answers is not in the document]
Joibel: Fixed

docs/metrics.md
Joibel: Chosen a past and present variant

docs/metrics.md
Joibel: No, no upper case Ps for you

docs/metrics.md
Joibel: Chosen a past and present variant

Joibel:
> I'm slightly surprised that the new markdown tables are passing CI. I thought there was a table linter? They aren't nice and pretty like they were before.

They aren't. The linter doesn't mind it, and some have been modified to look like this elsewhere by hand.

To make them pretty would need a prettifier, and this way is also easier for hand editing a bit like one sentence per line is.

Not remedying.

tico24 (approved):

isubasinghe (approved):
This is fine, I think I'd personally prefer if we used a markdown AST like https://github.com/yuin/goldmark. 

This makes it possible to render the markdown nicely, it also makes parsing and doing inline modifications much easier. 

No strong preferences though, this is not critical code I guess, as long as its generating valid markdown, its fine.

Joibel:
> This is fine, I think I'd personally prefer if we used a markdown AST like https://github.com/yuin/goldmark.
> 
> This makes it possible to render the markdown nicely, it also makes parsing and doing inline modifications much easier.

Goldmark doesn't do nice rendering, it has an emphasis on reading markdown. I've implemented nao1215/markdown which brings in more dependencies, but does also make the tables prettier so that @tico24 can pretend he's happy too.

terrytangyuan (approved):
This is great. Thanks!

isubasinghe (approved):
Thanks  for the changes.

PR #13945, opened 2024-11-27
ci: pin workflows to Ubuntu 24

MasonM:
### Motivation
On December 5th, the `ubuntu-latest` runner will start pointing to Ubuntu 24: https://github.com/actions/runner-images/issues/10636

This updates all workflows to use Ubuntu 24 so we don't have any unexpected surprises.

### Modifications
`s/ubuntu-latest/ubuntu-24.04`
### Verification
Wait for workflows to run


Joibel (approved):
Thanks for being proactive on this one. I definitely prefer pinned things like this.

PR #13953, opened 2024-11-28
docs: correctly document `result` output parameter

MasonM:
### Motivation
HTTP templates set `output.result` to the response body, but this wasn't documented: https://github.com/argoproj/argo-workflows/blob/1f304ba6780ee043ea5ecd74e06787bcba298f8f/workflow/executor/agent.go#L260

This has confused users (e.g. https://github.com/argoproj/argo-workflows/discussions/13950#discussioncomment-11404302). Additionally, the API docs for `result` only mention script templates, but container templates also set it.
### Modifications

Update relevant docs  

### Verification

Ran `make docs-serve`

### DIFF
--- a/api/jsonschema/schema.json
+++ b/api/jsonschema/schema.json
@@ -6211,7 +6211,7 @@
           "x-kubernetes-patch-strategy": "merge"
         },
         "result": {
-          "description": "Result holds the result (stdout) of a script template",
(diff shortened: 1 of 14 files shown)


Joibel (approved):

Joibel:
Thanks for updating this

PR #13957, opened 2024-12-02
fix(api): properly authorize GET workflow fallback to archive

Joibel:
Cherry pick of patch for https://github.com/argoproj/argo-workflows/security/advisories/GHSA-h36c-m3rf-34h9

### DIFF
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -629,7 +629,7 @@ func (s *ArgoServerSuite) TestPermission() {
 	// Test get wf w/ archive fallback with good token
 	s.bearerToken = goodToken
 	s.Run("GetWFsFallbackArchivedGoodToken", func() {
-		s.e().GET("/api/v1/workflows/"+uid).
+		s.e().GET("/api/v1/workflows/"+nsName).
 			WithQuery("listOptions.labelSelector", "workflows.argoproj.io/test").
 			Expect().
 			Status(200)
@@ -638,7 +638,7 @@ func (s *ArgoServerSuite) TestPermission() {
 	// Test get wf w/ archive fallback with bad token
 	s.bearerToken = badToken
 	s.Run("GetWFsFallbackArchivedBadToken", func() {
-		s.e().GET("/api/v1/workflows/" + uid).
+		s.e().GET("/api/v1/workflows/" + nsName).
 			Expect().
 			Status(403)
 	})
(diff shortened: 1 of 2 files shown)


test/e2e/argo_server_test.go
```diff
@@ -623,7 +626,33 @@ func (s *ArgoServerSuite) TestPermission() {
 			Status(403)
 	})
 
+	// Test get wf w/ archive fallback with good token
+	s.bearerToken = goodToken
+	s.Run("GetWFsFallbackArchivedGoodToken", func() {
+		s.e().GET("/api/v1/workflows/"+nsName).
```
agilgur5, line 632: nit/incidental finding: so the spacing around the `+` is different on this line than on line 641 and 649 below. I would think `gofmt` would make these consistent though -- it didn't make changes in https://github.com/argoproj/argo-workflows/pull/12979#discussion_r1578646195 either so I was thinking it's possibly not running on test files? Afaik `gofmt` usually does `x + y` with a space on both sides of the `+`

See also https://github.com/golang/go/issues/12105 -- only one argument here

Joibel: Yeah, this is pretty bonkers, and I don't understand why. I haven't analysed it any way because I was concentrating on getting the release out.

CI does pick it up, as you can see [here](https://github.com/argoproj/argo-workflows/actions/runs/12121071524/job/33791245905) when I pushed it up initially with the fixed formatting. The [original test](https://github.com/argoproj/argo-workflows/blob/06c761b8cc993aa6ab60f8c35c3c95bb334f3da0/test/e2e/argo_server_test.go#L611) has the same thing, so the linter had 'formatted' it there too when you wrote it.

I'm most thrown by the fact the other two lines you pointed out are formatted correctly, it's just this one that's "wrong", yet they are all so similar.

agilgur5: > I haven't analysed it in any way because I was concentrating on getting the release out.

For sure 👍 
I just noticed it incidentally

> CI does pick it up, as you can see [here](https://github.com/argoproj/argo-workflows/actions/runs/12121071524/job/33791245905) when I pushed it up initially with the fixed formatting. The [original test](https://github.com/argoproj/argo-workflows/blob/06c761b8cc993aa6ab60f8c35c3c95bb334f3da0/test/e2e/argo_server_test.go#L611) has the same thing, so the linter had 'formatted' it there too when you wrote it.

Well that's more bizarre than I thought 👀 Thanks for checking that!
In this case I'd hazard a guess this might be a `gofmt` bug or some fairly odd quirk explained by some other line widths; potentially worth filing a bug report in either scenario 😅

agilgur5:
@Joibel could you update https://github.com/argoproj/argo-workflows/security/advisories/GHSA-h36c-m3rf-34h9 to mention that it is resolved by this PR? Usually we've had it written at the bottom of the "Summary" section (for example, see https://github.com/argoproj/argo-workflows/security/advisories/GHSA-ghjw-32xw-ffwr)

Just to be able to follow the fix accordingly and posterity/historical lineage.
(Right now you have to dissect the version numbers and diff to find this as there is no direct link from the advisory itself. The GH reviewed version includes the PR and commit if it's in the original as well.)

PR #13971, opened 2024-12-05
fix: cronOperator/serverResubmitWf retry create workflow on transient error. Fixes #13970

### LINKED ISSUE #13970: Retry on transient error when cron creating workflow
### Pre-requisites

Workflow controller checks transient error and retry on create pod, and many other processes that interact with k8s api,

This transient retry do not apply to Create workflow.

We have seen intermittent errors like `connection timed out`, `connection reset by peer` occasionally when k8s API is underpressure.
This causes cron to skipthe  schedule.

We should add a similar retry on transient error logic to create a workflow

### Version(s)

217b598684c6d0cb9384e8c649f8e73659c5f9e5

```YAML
-
```

```text
```

```text
```

tczhao:
Fixes #13970 

### Motivation

To address issue where a cron intiated create workflow could failed due network issue

### Modifications

Add retry to util.SubmitWorkflow

The `SubmitWorkflow` is used by `argo server.resubmit workflow` and `controller.cron operator`
The original issue was on cron operator, but having retry in the util function allows server to retry resubmit when there's an intermittent issue communicating with the controller, enabling a better user experience.

### Verification

The code style is identical to how other transient patterns are written.
We currently don't have a process to force k8s network error, thus hard to write additional test.
We do have a test for transient utility and a successful e2e test is an indication that no issue has been introduced

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -199,7 +199,7 @@ func SubmitWorkflow(ctx context.Context, wfIf v1alpha1.WorkflowInterface, wfClie
 		err = waitutil.Backoff(retry.DefaultRetry, func() (bool, error) {
 			var err error
 			runWf, err = wfIf.Create(ctx, wf, metav1.CreateOptions{})
-			return !errorsutil.IsTransientErr(err), nil
+			return !errorsutil.IsTransientErr(err), err
 		})
 		return runWf, err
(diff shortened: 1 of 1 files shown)


Joibel (approved):
Nice, thank you

PR #13981, opened 2024-12-10
docs: Fixed broken bash script when installing on macos

htquanq:
This is how I fixed bash script installing using cURL on my macbook.

### DIFF
--- a/hack/release-notes.md
+++ b/hack/release-notes.md
@@ -19,7 +19,7 @@ Available via `curl`
 ```bash
 # Detect OS
 ARGO_OS="darwin"
-if [[ uname -s != "Darwin" ]]; then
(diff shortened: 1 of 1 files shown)


MasonM (approved):
Thanks! The failing `DCO` check is because the commit wasn't signed off. You should be able to fix that with these commands:
```
git commit --signoff --amend
git push --force
```

terrytangyuan (approved):
LGTM once DCO check is fixed

htquanq:
Hello guys, I have signed off commits

saerdnaer:
The recent release note still have the same issue:

https://github.com/argoproj/argo-workflows/releases

image

htquanq:
@saerdnaer could you open another issue for releases page? I updated the doc but look like the releases page, it need someone who has permission to edit

PR #13995, opened 2024-12-13
fix(ui): skip history.push on initial page load. Fixes #13940

### LINKED ISSUE #13940: UI: Many steps are needed to get back to the detail screen from the list.
### Pre-requisites

After transitioning from the list to the detail screen, a browser back is required twice to return to the list.
This can be observed not only in Workflow, but also in CronWorkflow, WorkflowTemplate, and ClusterWorkflowTemplate.

### Version(s)

v3.6.0

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

tczhao:
Fixes #13940 

### Motivation

To resolve issue where user need to click backbutton twice to go back to previous page

### Modifications

useEffect always executes after page render, causing duplicate entry in history.
skip useEffect for initial page load

### Verification

Clicking on UI make sure below issue doesn't occur
> After transitioning from the list to the detail screen, a browser back is required twice to return to the list.
> This can be observed not only in Workflow, but also in CronWorkflow, WorkflowTemplate, and ClusterWorkflowTemplate.

Check other UI parts perform as usual prior to the change

### DIFF
--- a/ui/src/cron-workflows/cron-workflow-details.tsx
+++ b/ui/src/cron-workflows/cron-workflow-details.tsx
@@ -2,7 +2,7 @@ import {NotificationType} from 'argo-ui/src/components/notifications/notificatio
 import {Page} from 'argo-ui/src/components/page/page';
 import {SlidingPanel} from 'argo-ui/src/components/sliding-panel/sliding-panel';
 import * as React from 'react';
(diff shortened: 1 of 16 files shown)


terrytangyuan (approved):
Thank you!

PR #14004, opened 2024-12-16
fix: ensure no nulls in postgres. Fixes  #13711

### LINKED ISSUE #13711: API query error: `unsupported Unicode escape sequence (SQLSTATE 22P05)`
### Pre-requisites

After upgrading from v3.5.6 to v3.5.11, we see the following error in the Argo UI whenever trying to list `Succeeded` or `Failed` workflows:

```
{"code":13,"message":"ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)"}: ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)
```

argo-server logs show the following:

```
ERROR 2024-10-06T09:08:57.605638209Z [resource.labels.containerName: argo-server] time="2024-10-06T09:08:57.605Z" level=info duration="135.318µs" method=GET path=index.html size=487 status=0
INFO 2024-10-06T09:08:57.993832698Z [resource.labels.containerName: argo-server] 2024/10/06 09:08:57 Session ID: 00001
INFO 2024-10-06T09:08:57.993891852Z [resource.labels.containerName: argo-server] Query: SELECT count(*) as total FROM "argo_archived_workflows" WHERE (("clustername" = $1 AND "instanceid" = $2) AND "namespace" = $3 AND not exists (select 1 from argo_archived_workflows_labels where clustername = argo_archived_workflows.clustername and uid = argo_archived_workflows.uid and name = 'workflows.argoproj.io/controller-instanceid') AND exists (select 1 from argo_archived_workflows_labels where clustername = argo_archived_workflows.clustername and uid = argo_archived_workflows.uid and name = 'workflows.argoproj.io/phase' and value in ('Succeeded')))
INFO 2024-10-06T09:08:57.993898457Z [resource.labels.containerName: argo-server] Arguments: []interface {}{"default", "", "default"}
INFO 2024-10-06T09:08:57.993901898Z [resource.labels.containerName: argo-server] Stack:
INFO 2024-10-06T09:08:57.993904741Z [resource.labels.containerName: argo-server] fmt.(*pp).handleMethods@/usr/local/go/src/fmt/print.go:673
INFO 2024-10-06T09:08:57.993911097Z [resource.labels.containerName: argo-server] fmt.(*pp).printArg@/usr/local/go/src/fmt/print.go:756
INFO 2024-10-06T09:08:57.993913723Z [resource.labels.containerName: argo-server] fmt.(*pp).doPrint@/usr/local/go/src/fmt/print.go:1211
INFO 2024-10-06T09:08:57.993916162Z [resource.labels.containerName: argo-server] fmt.Append@/usr/local/go/src/fmt/print.go:289
INFO 2024-10-06T09:08:57.993918601Z [resource.labels.containerName: argo-server] log.(*Logger).Print.func1@/usr/local/go/src/log/log.go:261
INFO 2024-10-06T09:08:57.993921542Z [resource.labels.containerName: argo-server] log.(*Logger).output@/usr/local/go/src/log/log.go:238
INFO 2024-10-06T09:08:57.993924791Z [resource.labels.containerName: argo-server] log.(*Logger).Print@/usr/local/go/src/log/log.go:260
INFO 2024-10-06T09:08:57.993928471Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/persist/sqldb.(*workflowArchive).CountWorkflows@/go/src/github.com/argoproj/argo-workflows/persist/sqldb/workflow_archive.go:264
INFO 2024-10-06T09:08:57.993931447Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/workflow.(*workflowServer).ListWorkflows@/go/src/github.com/argoproj/argo-workflows/server/workflow/workflow_server.go:193
INFO 2024-10-06T09:08:57.993934447Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/pkg/apiclient/workflow._WorkflowService_ListWorkflows_Handler.func1@/go/src/github.com/argoproj/argo-workflows/pkg/apiclient/workflow/workflow.pb.go:1826
INFO 2024-10-06T09:08:57.993936732Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.RatelimitUnaryServerInterceptor.func5@/go/src/github.com/argoproj/argo-workflows/util/grpc/interceptor.go:65
INFO 2024-10-06T09:08:57.993942795Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.ChainUnaryServer.func6.1.1@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25
INFO 2024-10-06T09:08:57.993945407Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/auth.(*gatekeeper).UnaryServerInterceptor.func1@/go/src/github.com/argoproj/argo-workflows/server/auth/gatekeeper.go:98
INFO 2024-10-06T09:08:57.993948144Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.ChainUnaryServer.func6.1.1@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25
INFO 2024-10-06T09:08:57.993951345Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/util/grpc.glob..func1@/go/src/github.com/argoproj/argo-workflows/util/grpc/interceptor.go:45
INFO 2024-10-06T09:08:57.993953807Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.ChainUnaryServer.func6.1.1@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25
INFO 2024-10-06T09:08:57.993974087Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.PanicLoggerUnaryServerInterceptor.func4@/go/src/github.com/argoproj/argo-workflows/util/grpc/interceptor.go:26
INFO 2024-10-06T09:08:57.993976959Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.ChainUnaryServer.func6.1.1@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25
INFO 2024-10-06T09:08:57.993985441Z [resource.labels.containerName: argo-server] github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus.UnaryServerInterceptor.func1@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/logging/logrus/server_interceptors.go:31
INFO 2024-10-06T09:08:57.993988269Z [resource.labels.containerName: argo-server] github.com/argoproj/argo-workflows/v3/server/apiserver.(*argoServer).newGRPCServer.ChainUnaryServer.func6.1.1@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25
INFO 2024-10-06T09:08:57.993990595Z [resource.labels.containerName: argo-server] github.com/grpc-ecosystem/go-grpc-prometheus.init.(*ServerMetrics).UnaryServerInterceptor.func3@/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-prometheus@v1.2.0/server_metrics.go:107
INFO 2024-10-06T09:08:57.993992913Z [resource.labels.containerName: argo-server] Error: upper: slow query
INFO 2024-10-06T09:08:57.993995607Z [resource.labels.containerName: argo-server] Time taken: 0.65064s
INFO 2024-10-06T09:08:57.993997870Z [resource.labels.containerName: argo-server] Context: context.Background
INFO 2024-10-06T09:08:57.994000109Z [resource.labels.containerName: argo-server] {}
ERROR 2024-10-06T09:08:58.003874930Z [resource.labels.containerName: argo-server] time="2024-10-06T09:08:58.003Z" level=error msg="finished unary call with code Internal" error="rpc error: code = Internal desc = ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)" grpc.code=Internal grpc.method=ListWorkflows grpc.service=workflow.WorkflowService grpc.start_time="2024-10-06T09:08:57Z" grpc.time_ms=665.575 span.kind=server system=grpc
ERROR 2024-10-06T09:08:58.003922336Z [resource.labels.containerName: argo-server] time="2024-10-06T09:08:58.003Z" level=info duration=666.353369ms method=GET path=/api/v1/workflows/default size=83 status=500
```

Note also the following:

```sql
argo-production=> select count(*) from argo_archived_workflows where workflow::text LIKE '%\u0000%';
 count
-------
  1459
(1 row)

argo-production=> 
```

Our DB for archived workflows is Postgres 9.6.

We have the following defined in `workflow-controller-configmap`

```yaml
  persistence: |
    archive: true
    archiveTTL: 180d
    connectionPool:
      maxIdleConns: 100
      maxOpenConns: 0
      connMaxLifetime: 0s
    nodeStatusOffLoad: true
    postgresql:
      database: argo-production
      host: <Our DB host>
      passwordSecret:
        key: password
        name: argo-postgres-config
      port: 5432
      tableName: argo_workflows
      userNameSecret:
        key: username
        name: argo-postgres-config
 ```

### Version(s)

v3.5.11

This is in our production environment immediately after updating to 3.5.11

```text
```

```text
```

isubasinghe:
# Motivation
Unfortunately postgres does not handle NULL bytes in the json, this is as a result of a non-compliant implementation of JSON that exists in postgres. See  more details about this limitation [here](https://github.com/argoproj/argo-workflows/issues/13711#issuecomment-2507193576). 

# Modifications
This is arguably hacky solution, but this fix attempts to resolve this by mutating the workflow. 

# Verification
This has tests and additionally been verified by checking postgres.

### DIFF
--- a/persist/sqldb/raw_marshaller.go
+++ b/persist/sqldb/raw_marshaller.go
@@ -1,0 +1,79 @@
+// Package sqldb implements workflow archiving
+package sqldb
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+type pair struct {
+	k string
+	v interface{}
+}
+
+func convertStrings(m map[string]interface{}) (map[string]interface{}, error) {
+	for k, v := range m {
+		switch v := v.(type) {
+		case string:
+			tmp := strconv.QuoteToASCII(v)
+			tmp = tmp[1:]
+			tmp = tmp[:len(tmp)-1]
+			m[k] = tmp
+		case map[string]interface{}:
+			convertedInner, err := convertStrings(v)
+			if err != nil {
+				return nil, err
+			}
+			m[k] = convertedInner
+		case []any:
+			newList := []any{}
+			for _, val := range v {
+				// bit of a hack to reuse code
+				valM := map[string]interface{}{"entry": val}
+				retVal, err := convertStrings(valM)
+				if err != nil {
+					return nil, err
+				}
+				newList = append(newList, retVal["entry"])
+			}
+			m[k] = newList
+		default:
+			m[k] = v
+		}
+	}
+	return m, nil
+}
+
+func convertMap(jsonObject any) (map[string]interface{}, error) {
+
+	bytes, err := json.Marshal(jsonObject)
+	if err != nil {
+		return nil, err
+	}
+	// we do this to handle json tags without explicitly programming this
+	// ourselves.
+	oldMap := make(map[string]interface{})
+	err = json.Unmarshal(bytes, &oldMap)
+	if err != nil {
+		return nil, err
+	}
+
+	newMap, err := convertStrings(oldMap)
+	if err != nil {
+		return nil, err
+	}
+
+	return newMap, nil
+}
+
+func jsonMarshallRawStrings(jsonObject any) ([]byte, error) {
+	m, err := convertMap(jsonObject)
+	if err != nil {
+		return nil, err
+	}
+	b, err := json.Marshal(m)
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
--- a/persist/sqldb/raw_marshaller.go
+++ b/persist/sqldb/raw_marshaller.go
@@ -3,22 +3,20 @@ package sqldb
 
 import (
 	"encoding/json"
-	"strconv"
 )
-
-type pair struct {
-	k string
-	v interface{}
-}
 
 func convertStrings(m map[string]interface{}) (map[string]interface{}, error) {
 	for k, v := range m {
 		switch v := v.(type) {
 		case string:
-			tmp := strconv.QuoteToASCII(v)
+			ms, err := json.Marshal(v)
+			if err != nil {
+				return nil, err
+			}
+			tmp := string(ms)
 			tmp = tmp[1:]
 			tmp = tmp[:len(tmp)-1]
-			m[k] = tmp
+			m[k] = string(tmp)
 		case map[string]interface{}:
 			convertedInner, err := convertStrings(v)
 			if err != nil {
--- a/persist/sqldb/workflow_archive.go
+++ b/persist/sqldb/workflow_archive.go
@@ -99,9 +99,19 @@ func (r *workflowArchive) ArchiveWorkflow(wf *wfv1.Workflow) error {
 	logCtx := log.WithFields(log.Fields{"uid": wf.UID, "labels": wf.GetLabels()})
 	logCtx.Debug("Archiving workflow")
 	wf.ObjectMeta.Labels[common.LabelKeyWorkflowArchivingStatus] = "Persisted"
-	workflow, err := json.Marshal(wf)
-	if err != nil {
-		return err
+	var workflow []byte
+	var err error
+	if r.dbType == Postgres {
+		workflow, err = jsonMarshallRawStrings(wf)
+		if err != nil {
+			log.Errorf("was unable to marshal to json with raw strings: %s", err)
+			return err
+		}
+	} else {
+		workflow, err = json.Marshal(wf)
+		if err != nil {
+			return err
+		}
 	}
 	return r.session.Tx(func(sess db.Session) error {
 		_, err := sess.SQL().
--- a/persist/sqldb/raw_marshaller_test.go
+++ b/persist/sqldb/raw_marshaller_test.go
@@ -1,0 +1,52 @@
+package sqldb
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type simpleStruct struct {
+	Val         string          `json:"val"`
+	Val2        *string         `json:"val2"`
+	Val3        *string         `json:"val3,omitempty"`
+	InnerSimple *simpleStruct   `json:"sstruct,omitempty"`
+	List        []string        `json:"list"`
+	List2       *[]simpleStruct `json:"list2,omitempty"`
+}
+
+func TestStrings(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+
+	val2 := "val2 \uC582"
+	val3 := "val3 \uC583"
+	inner := simpleStruct{
+		Val: "inner val \uc589",
+	}
+	l := []simpleStruct{inner}
+	s := simpleStruct{
+		Val:         "hello \x00",
+		Val2:        &val2,
+		Val3:        &val3,
+		InnerSimple: &inner,
+		List2:       &l,
+	}
+
+	newMap, err := convertMap(s)
+	require.NoError(err)
+	assert.Contains(string(newMap["val"].(string)), "\\x00")
+	assert.Contains(string(newMap["val3"].(string)), "\\uc583")
+
+	innerMapI, ok := newMap["sstruct"]
+	require.True(ok)
+	innerMap, ok := innerMapI.(map[string]interface{})
+	require.True(ok)
+	assert.Contains(string((innerMap["val"]).(string)), "inner val \\uc589")
+
+	_, err = json.Marshal(newMap)
+	require.NoError(err)
+
+}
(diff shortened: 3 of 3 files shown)


persist/sqldb/raw_marshaller.go
```diff
@@ -0,0 +1,77 @@
...
+		switch v := v.(type) {
+		case string:
+			ms, err := json.Marshal(v)
+			if err != nil {
+				return nil, err
+			}
+			tmp := string(ms)
+			tmp = tmp[1:]
+			tmp = tmp[:len(tmp)-1]
+			m[k] = string(tmp)
```
isubasinghe, line 19: Horrible horrible hackiness

isubasinghe:
I have a couple of questions before we merge this.

What are opinions on a serialize to []byte and then remove nulls approach instead? It seems saner. 
Should this be opt in via an environment variable, we are after all mutating the workflow.

tooptoop4:
please don't remove the nulls, we use them in our bash script

isubasinghe:
> please don't remove the nulls, we use them in our bash script

This doesn't remove nulls from workflows, it escapes to \\u0000 when persisting to postgres. 

Your workflow wouldn't persist to postgres anyway if it had nulls.

We can make this optional if you want ?

isubasinghe:
> just want to be able to run a workflow containing null character

This fix doesn't break this so you got nothing to worry about there.

MasonM:
AFAICT we don't have any automated tests for the PostgreSQL-specific logic in `persist/sqldb/workflow_archive.go`, and it seems like we should considering how tricky these edge cases can get. I entered https://github.com/argoproj/argo-workflows/pull/14011 to have the CI run the API tests (which exercise the archiving logic) on PostgreSQL

persist/sqldb/workflow_archive.go
```diff
@@ -99,9 +99,19 @@ func (r *workflowArchive) ArchiveWorkflow(wf *wfv1.Workflow) error {
 	logCtx := log.WithFields(log.Fields{"uid": wf.UID, "labels": wf.GetLabels()})
 	logCtx.Debug("Archiving workflow")
 	wf.ObjectMeta.Labels[common.LabelKeyWorkflowArchivingStatus] = "Persisted"
-	workflow, err := json.Marshal(wf)
-	if err != nil {
-		return err
+	var workflow []byte
+	var err error
+	if r.dbType == Postgres {
+		workflow, err = jsonMarshallRawStrings(wf)
```
tooptoop4, line 105: would be good to ensure the new marshalled json is the same as the old way (with exception of the null char), could there be some if condition that compares them? otherwise there is risk that null character gets solved but other characters start breaking

MasonM: My PR has a test for this: https://github.com/argoproj/argo-workflows/pull/14011

tooptoop4: @MasonM that is only for 1 tiny workflow, in the wild there are all kinds of workflows with various funky characters. i think a runtime check to ensure correctness is better than risk of a different/incorrect workflow being archived/retried

isubasinghe: @tooptoop4 @MasonM sounds like maybe its better to serialize first and then remove null bytes instead of working at this layer? I was hoping to avoid that but perhaps that is indeed better.

isubasinghe:
I have another fix that works on the byte level and is able to restore the original workflow. I think I want that to go in instead.

isubasinghe:
I abandoned the byte level fix in favour of this: https://github.com/argoproj/argo-workflows/pull/14030


Why? 
Well json encoding doesn't preserve order, this makes it a bit more difficult to track the changed bytes. 
This means that we need to really track byte changed for each recursive field and store them in postgres. 


To properly do this byte level fix would be a ton of effort, but to add to this, it would be error prone as it relies on a lot of moving parts in the form of these field/offset pairs that track modifications.

PR #14012, opened 2024-12-18
ci: optimize tests by only building binaries once

MasonM:
### Motivation
All the E2E tests job run `make controller` to build the workflow controller, and several also run `make cli` to build the CLI, which takes ~2 minutes. It'll be faster to have a central "build binaries" job that does that and uploads them as an artifact, which each E2E job can then download.

### Modifications
Add `build-binaries` job to build the controller and CLI and upload them as an artifact, then download those artifacts in each E2E job.
### Verification

Wait for E2E tests to run.


.github/workflows/ci-build.yaml
```diff
@@ -154,6 +154,33 @@ jobs:
...
+          - target: controller
+          - target: cli
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version: "1.23"
+          cache: true
+      - name: Build
+        run: make ${{matrix.target}}
```
tczhao, line 176: Is `STATIC_FILES=false` being left out on purpose?
Having it should make the build faster

MasonM: It's actually set to false here, so no need to explicitly set it: https://github.com/argoproj/argo-workflows/blob/abd14b7801647984859fd33e738a060199b40d9a/Makefile#L44-L49

blkperl (approved):

PR #14013, opened 2024-12-18
ci: don't need to install socat

MasonM:
### Motivation
socat used to be required for `kubectl port-forward` to work, but this is no longer true: https://github.com/kubernetes/kubernetes/pull/100679

This should reduce the time for each E2E test job by ~10 seconds.

### Modifications
Remove `sudo apt-get -y install socat` line

### Verification
Wait for E2E tests to pass


MasonM:
My mistake: containerd no longer uses socat, but we're running k3s with cri-dockerd, which does: https://github.com/k3s-io/cri-dockerd/blob/4995f339edcffdf890406b3f1477d34e38477f18/streaming/streaming_others.go#L46

PR #14016, opened 2024-12-18
docs: explicitly reference mkdocs.yml in readthedocs config

tico24:
Fixes an impending deprecation in readthedocs: https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

I believe this may need to be cherrypicked to all branches that we produce docs for.


terrytangyuan (approved):
I was just about to open a PR for this as well. Thank you!

PR #14024, opened 2024-12-22
test: fix and optimize flaky cron tests. Partial fix for #10807

### LINKED ISSUE #10807: Flakey tests

MasonM:
Fixes #10807

### Motivation

The cron tests have been some of the slowest and flakiest for awhile now, mostly due to using `time.Sleep()` to wait a fixed amount of time. The `TestCronSuite` function tries to workaround the flakiness by running the suite at the 30 second mark, but this is only effective for the first test in the suite.

### Modifications

This changes nearly all the cron tests to remove the `time.Sleep()` calls and instead use the API to poll for changes. The two tests I couldn't fix to stop using `time.Sleep()` are `TestBasicForbid` and `TestCronCountersForbid`, since I don't know of any way of verifying the `concurrencyPolicy` is respected without sleeping.

Also, most of the cron fixture data wasn't specifying workflow metadata properly using [workflowMetadata](https://argo-workflows.readthedocs.io/en/latest/fields/#cronworkflowspec), which meant they weren't being cleaned up automatically, so I fixed that too.

### Verification

Ran `make test-cron` locally

### DIFF
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -1006,6 +1006,9 @@ spec:
   startingDeadlineSeconds: 0
   successfulJobsHistoryLimit: 4
   failedJobsHistoryLimit: 2
+  workflowMetadata:
+    labels:
+      workflows.argoproj.io/test: "true"
   workflowSpec:
     podGC:
       strategy: OnPodCompletion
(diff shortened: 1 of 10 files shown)


terrytangyuan (approved):
This is great. Thanks!

PR #14026, opened 2024-12-23
fix: controller: ensures workflow reconciling task result properly when failing to received timely updates from api server

bom-d-van:
There seems to be no related open issue, so I just submit a pr. but it's related to https://github.com/argoproj/argo-workflows/pull/12537

### Motivation

error scenario: a pod for a step in a workflow has completed, and its task result are properly created and finalized by its wait container (judging from the exit status of the wait container), however, the task result informer in the controller leader has not received any updates about it (due to overloaded api server or etcd).

currently, the argo workflow controller doesn't handle the above scenario properly. it would mark the workflow node succeeded and shows no artifact outputs (even though they are already uploaded to the repository).

we did run into this situation in our production instance (it's v3.5.8). 

it's not easy to reproduce this problem, but we can have a manual fault injection in `workflow/controller/taskresult.go:func (woc *wfOperationCtx) taskResultReconciliation()` to simulate the situation and I did reproduce the issue on release v3.6.2:

```diff
+++ workflow/controller/taskresult.go
@@ -1,7 +1,9 @@
 package controller

 import (
+	"os"
 	"reflect"
+	"strings"
 	"time"

 	log "github.com/sirupsen/logrus"
@@ -62,6 +64,12 @@ func (woc *wfOperationCtx) taskResultReconciliation() {
 	objs, _ := woc.controller.taskResultInformer.GetIndexer().ByIndex(indexes.WorkflowIndex, woc.wf.Namespace+"/"+woc.wf.Name)
 	woc.log.WithField("numObjs", len(objs)).Info("Task-result reconciliation")

+	if strings.Contains(woc.wf.Name, "-xhu-debug-") {
+		if _, err := os.Stat("/tmp/xhu-debug-control"); err != nil {
+			return
+		}
+	}
```

### Modifications

the change is to forcefully mark the workflow having incomplete TaskResult in assessNodeStatus.

this fix doesn't handle the case when a pod failed, there are too many potentially failure scenarios (like the wait container might not be able to insert a task result). plus, a retry is probably needed when there are failures. the loss is probably not as great as a successful one.

### Verification

it's covered by the updated test case and my manual verification details are included bellow (using the fault injection above).

<details>

#### test workflow to verify the fix:

```yaml
kind: Workflow
metadata:
  generateName: test-xhu-debug-
spec:
  entrypoint: bash-script-example
  activeDeadlineSeconds: 3600
  templates:
  - name: bash-script-example
    steps:
    - - name: generate
        template: gen-random-int-bash
    - - name: print
        template: print-message
        arguments:
          parameters:
          - name: message
            value: "{{steps.generate.outputs.result}}"  # The result of the here-script

  - name: gen-random-int-bash
    outputs:
      artifacts:
      # generate hello-art artifact from /tmp/hello_world.txt
      # artifacts can be directories as well as files
      - name: hello-art
        path: /tmp/hello_world.txt
    script:
      image: reg.deeproute.ai/deeproute-public/nicolaka/netshoot:v0.13
      command: [bash]
      source: |                                         # Contents of the here-script
        touch /tmp/hello_world.txt
        cat /dev/urandom | od -N2 -An -i | awk -v f=1 -v r=100 '{printf "%i\n", f + r * $1 / 65536}'

  - name: print-message
    inputs:
      parameters:
      - name: message
    outputs:
      artifacts:
      # generate hello-art artifact from /tmp/hello_world.txt
      # artifacts can be directories as well as files
      - name: hello-art2
        path: /tmp/hello_world.txt
    container:
      image: reg.deeproute.ai/deeproute-public/nicolaka/netshoot:v0.13
      command: [sh, -c]
      args: ["touch /tmp/hello_world.txt; echo result was: {{inputs.parameters.message}}"]
```

#### with the manual fault injection, without the fix:

image

#### with the manual fault injection, with the fix:

image

</details>

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1455,6 +1455,24 @@ func (woc *wfOperationCtx) assessNodeStatus(ctx context.Context, pod *apiv1.Pod,
 		new.PodIP = pod.Status.PodIP
 	}
 
+	if resultName := woc.nodeID(pod); new.Phase == wfv1.NodeSucceeded && tmpl != nil &&
+		(tmpl.HasOutputs() || tmpl.HasOutput()) && !woc.wf.Status.IsTaskResultInited(resultName) {
+		// error scenario: a pod for a step in a workflow has completed, and its task
+		// result are properly created and finalized by its wait container (judging from
+		// the exit status of the wait container), however, the task result informer in
+		// the controller leader has not received any updates about it (due to overloaded
+		// api server or etcd).
+		//
+		// the change is to forcefully mark the workflow having incomplete TaskResult in
+		// assessNodeStatus.
+		//
+		// this fix doesn't handle the case when a pod failed, there are too many
+		// potentially failure scenarios (like the wait container might not be able to
+		// insert a task result). plus, a retry is probably needed when there are
+		// failures. the loss is probably not as great as a successful one.
+		woc.wf.Status.MarkTaskResultIncomplete(resultName)
+	}
+
 	new.HostNodeName = pod.Spec.NodeName
 
 	if !new.Progress.IsValid() {
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -2026,6 +2026,15 @@ func (ws *WorkflowStatus) IsTaskResultIncomplete(name string) bool {
 	return false // workflows from older versions do not have this status, so assume completed if this is missing
 }
 
+func (ws *WorkflowStatus) IsTaskResultInited(name string) bool {
+	if ws.TaskResultsCompletionStatus == nil {
+		return false
+	}
+
+	_, found := ws.TaskResultsCompletionStatus[name]
+	return found
+}
+
 func (ws *WorkflowStatus) IsOffloadNodeStatus() bool {
 	return ws.OffloadNodeStatusVersion != ""
 }
--- a/workflow/controller/operator_test.go
+++ b/workflow/controller/operator_test.go
@@ -10898,8 +10898,20 @@ spec:
         i = random.randint(1, 100)
         print(i)`
 
-// TestWorkflowNeedReconcile test whether a workflow need reconcile taskresults.
+// testWorkflowNeedReconcile test whether a workflow need reconcile taskresults.
 func TestWorkflowNeedReconcile(t *testing.T) {
+	t.Run("hasIncompleteTaskResult", func(t *testing.T) {
+		hasIncompleteTaskResult := true
+		testWorkflowNeedReconcileHelper(t, hasIncompleteTaskResult)
+	})
+
+	t.Run("hasNoTaskResult", func(t *testing.T) {
+		hasIncompleteTaskResult := false
+		testWorkflowNeedReconcileHelper(t, hasIncompleteTaskResult)
+	})
+}
+
+func testWorkflowNeedReconcileHelper(t *testing.T, hasIncompleteTaskResult bool) {
 	cancel, controller := newController()
 	defer cancel()
 	ctx := context.Background()
@@ -10920,9 +10932,13 @@ func TestWorkflowNeedReconcile(t *testing.T) {
 	wf, err = wfcset.Get(ctx, wf.ObjectMeta.Name, metav1.GetOptions{})
 	require.NoError(t, err)
 	woc = newWorkflowOperationCtx(wf, controller)
-	for _, node := range woc.wf.Status.Nodes {
-		woc.wf.Status.MarkTaskResultIncomplete(node.ID)
+
+	if hasIncompleteTaskResult {
+		for _, node := range woc.wf.Status.Nodes {
+			woc.wf.Status.MarkTaskResultIncomplete(node.ID)
+		}
 	}
+
 	err, podReconciliationCompleted := woc.podReconciliation(ctx)
 	require.NoError(t, err)
 	assert.False(t, podReconciliationCompleted)


workflow/controller/operator.go
```diff
@@ -1438,6 +1438,24 @@ func (woc *wfOperationCtx) assessNodeStatus(ctx context.Context, pod *apiv1.Pod,
...
+		// the controller leader has not received any updates about it (due to overloaded
+		// api server or etcd).
+		//
+		// the change is to forcefully mark the workflow having incomplete TaskResult in
+		// assessNodeStatus.
+		//
+		// this fix doesn't handle the case when a pod failed, there are too many
+		// potentially failure scenarios (like the wait container might not be able to
+		// insert a task result). plus, a retry is probably needed when there are
+		// failures. the loss is probably not as great as a successful one.
```
tczhao, line 1455: we have retry for transient errors in wait container to handle this
https://github.com/argoproj/argo-workflows/blob/27a59385f62de98b7bddc0087b03df90af29e87e/workflow/executor/executor.go#L852-L858

tczhao (commented):
Great finding.
A delay in taskresultInformer will result in the issue you are seeing.
However, if you experience k8sapi pressure and the retry inside informer fails, you will see a much bigger problem. e.g. argo uses the cache from informer, the k8sapi pressure may result in this cache out of sync, and they will never sync until you restarts controller

@shuangkun has more context related to this change, will ask him to review instead

bom-d-van:
> Great finding. A delay in taskresultInformer will result in the issue you are seeing. However, if you experience k8sapi pressure and the retry inside informer fails, you will see a much bigger problem. e.g. argo uses the cache from informer, the k8sapi pressure may result in this cache out of sync, and they will never sync until you restarts controller
> 
> @shuangkun has more context related to this change, will ask him to review instead

yep, but in a way, a controller restart could help us resolving the issue. arguably, at least for us, it's still better than having partially succeeded workflows.

shuangkun:
It seems that the problem of output not being displayed cannot be solved

bom-d-van:
> It seems that the problem of output not being displayed cannot be solved

That's strange. I don't have this problem on my simulation. Once the controller was able to receive task result from api-server, everything just continue to work like normal.

Could you share me your argo-workflow setup and test workflow configs? @shuangkun

shuangkun:
If the pod has been successfully created, but the taskresult has not been updated by watch, and mark TaskResultStatus is true, will the previously fixed problem（https://github.com/argoproj/argo-workflows/pull/12537） occur again? When the next task references the output of the previous step, the node information has not been updated by taskresults at this time, and the reference will report an error.

shuangkun:
Here is a judgment to avoid starting the next new pod when taskresult is not updated
https://github.com/argoproj/argo-workflows/blob/b1a65e79fe583134117c33814b54dcaba40c4238/workflow/controller/operator.go#L1144

isubasinghe:
I think we need a rethink of how we handle taskresults/pod statuses together.


We (as open source contributors, not assigning any blame to anyone here) keep putting out these fires, but we should solve this through some other means. 

Solving this properly is probably only achievable in a larger time frame and as a result it probably makes sense to go ahead with this change.

I'm mostly saying this because we hit another a bug relating to a similar problem.

I'm currently working on a state machine in TLA+ (well PlusCal)  to model this problem, after this is done, we should extract an algorithm out of this state machine and implement it in the controller.

isubasinghe:
If anyone wants to work on the TLA+ spec with me, let me know.

bom-d-van:
> If the pod has been successfully created, but the taskresult has not been updated by watch, and mark TaskResultStatus is true, will the previously fixed problem（https://github.com/argoproj/argo-workflows/pull/12537 ） occur again?

Should be good. This is actually fixing one more edge case when checking task result.

bom-d-van:
> Solving this properly is probably only achievable in a larger time frame and as a result it probably makes sense to go ahead with this change.

@isubasinghe In a way, maybe not using TaskResult CRD as the middle man for task result reconciliation might work better. Having argoexec/wait container talking directly to the controller would be better?

shuangkun:
> > I think we need a rethink of how we handle taskresults/pod statuses together.
> > We (as open source contributors, not assigning any blame to anyone here) keep putting out these fires, but we should solve this through some other means.
> > Solving this properly is probably only achievable in a larger time frame and as a result it probably makes sense to go ahead with this change.
> > I'm mostly saying this because we hit another a bug relating to a similar problem.
> > I'm currently working on a state machine in TLA+ (well PlusCal) to model this problem, after this is done, we should extract an algorithm out of this state machine and implement it in the controller.
> 
> @isubasinghe In a way, maybe not using TaskResult CRD as the middle man for task result reconciliation might work better. Having argoexec/wait container talking directly to the controller would be better?

TaskResults was probably introduced for security reasons (cancelling patch pod permissions) and complex outputs transmission. Is there a better cloud-native approach?

carolkao:
We also encounter this issue in our production environment with v3.6.2.

Lots of workflows encounter `unable to resolve outputs from scope: xxx` error, and after some investigation, we found it's high possible due to the delay of workflowtaskresult.

Actually, we notice this situation in previous version of Argo workflow, but workaround by removing `workflowtaskresults` permission, so it'll fallback to old strategy - update output status in the pod.

Does this fallback strategy still exist? We're looking forward to a stable strategy.

cyyangnb:
We encounter this issue a few times every day and found that the provided changes successfully helped us eliminate the `unable to resolve outputs from scope: xxx` errors.  

However, we still occasionally experience workflow failures due to the inability to resolve the previous step's output—specifically, the `outputs.result` of the previous step. I suspect this might be related to the `tmpl.HasOutputs()` check in @bom-d-van's fix. If the `workflowTemplate` does not specify `outputs.arguments` or `outputs.artifacts`, the fix to mark `taskResult` as incomplete does not seem to be applied as expected. Would it make sense to remove the tmpl.HasOutputs() check? (I'm new to the codebase, so I’d appreciate any insights on this.)

Here’s an example of a workflowTemplate that occasionally fails:  
```yaml
namespaced/my.test/main:
  steps:
    - - name: run-prepare-arguments
        template: prepare-arguments
        arguments:
          parameters:
            - name: src
              value: "{{inputs.parameters.src}}"
    - - name: run-content
        arguments:
          parameters:
            - name: content
              value: "{{steps.run-prepare-arguments.outputs.result}}"     # <-- not resolved successfully
        templateRef:
          name: my.other
          template: from-parameter

namespaced/my.test/prepare-arguments:
  name: prepare-arguments
  inputs:
    parameters:
      - name: src
  outputs: {}                             # <-- no outputs
  metadata: {}
  container:
    ...
```

Additionally, what would be the next step for this issue? Would you consider merging the fix first before the more comprehensive solution is provided?

bom-d-van:
> Would it make sense to remove the tmpl.HasOutputs() check? (I'm new to the codebase, so I’d appreciate any insights on this.)

hi @cyyangnb, thanks for the feedback. I was being fairly conservative on my fix patch. I think it's probably ok to drop the `tmpl.HasOutputs()` check. Maybe you can give it a shot and let us know if it truly works for your deployment.

cyyangnb:
> > Would it make sense to remove the tmpl.HasOutputs() check? (I'm new to the codebase, so I’d appreciate any insights on this.)
> 
> hi @cyyangnb, thanks for the feedback. I was being fairly conservative on my fix patch. I think it's probably ok to drop the `tmpl.HasOutputs()` check. Maybe you can give it a shot and let us know if it truly works for your deployment.

Hi @bom-d-van, thanks for your response! I found another method, `tmpl.HasOutput()` (without the "s"), which "returns true if the template can and will have outputs (i.e., exit code and result)."

So, I added it to the if condition:
```
if resultName := woc.nodeID(pod); new.Phase == wfv1.NodeSucceeded &&
    (tmpl.HasOutputs() || tmpl.HasOutput()) && !woc.wf.Status.IsTaskResultInited(resultName) {
```
This fix resolves the issue of unresolved `outputs.result` based on my tests.

bom-d-van:
Thanks @cyyangnb. I have incorporated your fix in my patch.

@tczhao @shuangkun Should we consider merging this? Given the feedbacks, it's a real problem and this fix seems reasonable for the situations that we are in.

workflow/controller/operator.go
```diff
@@ -1455,6 +1455,24 @@ func (woc *wfOperationCtx) assessNodeStatus(ctx context.Context, pod *apiv1.Pod,
 		new.PodIP = pod.Status.PodIP
 	}
 
+	if resultName := woc.nodeID(pod); new.Phase == wfv1.NodeSucceeded && tmpl != nil &&
+		(tmpl.HasOutputs() || tmpl.HasOutput()) && !woc.wf.Status.IsTaskResultInited(resultName) {
```
cyyangnb, line 1459: We found that there are cases where tmpl could be nil when reaching this line. A nil check should be added. e.g.,
```
func (tmpl *Template) HasOutput() bool {
	return tmpl != nil && (tmpl.Container != nil || tmpl.ContainerSet.HasContainerNamed("main") || tmpl.Script != nil || tmpl.Data != nil || tmpl.HTTP != nil || tmpl.Plugin != nil)
}
```

bom-d-van: It's already there in the patch, isn't it?

cyyangnb: You're absolutely right. Thanks for the clarification! 
Looking forward to seeing the patch merged!

Elentary:
Can we merge this, guys ?

Elentary:
@tczhao @shuangkun are we good to merge it ? That's a huge pain now to deal with Failed workflows when your k8s api is a little bit slow and updates are not very well timed

tooptoop4:
@coderabbitai merge

isubasinghe:
Isn't the need for this PR nullified by the TaskResultsSynced field I introduced in the node status? 
We do not consider a node to be fulfilled until we have received this.

isubasinghe:
As a result, I am going to close this PR @bom-d-van . I appreciate your contribution, regardless.

PR #14028, opened 2024-12-23
ci: improve performance by eliminating redundancies

MasonM:
### Motivation
There's three sources of redundancy that are slowing down CI builds:
1. `make install` is being redundantly run three times for each E2E test:
    1. In the workflow: https://github.com/argoproj/argo-workflows/blob/6699ab396f830210f6dcac4f00a9328a629c142f/.github/workflows/ci-build.yaml#L349-L350
    2. As a prerequisite of `make start`: https://github.com/argoproj/argo-workflows/blob/6699ab396f830210f6dcac4f00a9328a629c142f/Makefile#L547
    3. As a dependency by Kit: https://github.com/argoproj/argo-workflows/blob/6699ab396f830210f6dcac4f00a9328a629c142f/tasks.yaml#L39
2.  Similarly, the changes in https://github.com/argoproj/argo-workflows/pull/14012 to centralize binary building aren't actually preventing the binaries from being rebuilt in the E2E tests because `controller` and `cli` are prerequisites of `make start`, and listed as prerequisites in Kit.
3. Go dependencies are being re-downloaded in every job. The [actions/setup-go](https://github.com/actions/setup-go) is supposed to prevent that by caching the contents of `GOMODCACHE`. I think the problem is the cache is usually populated by the `docs` workflow, which doesn't fetch all the Go dependencies ([example](https://github.com/argoproj/argo-workflows/actions/runs/12460536734/job/34778897761)), so they aren't included when `actions/setup-go` saves the cache.
image

### Modifications

This eliminates the first two redundancies by removing the unnecessary prerequisites from the `Makefile` and the redundant `make install` step in `ci-build.yaml`. Also, I added `make --touch dist/*` to mark everything as up-to-date so the binaries aren't rebuilt ([docs](https://www.gnu.org/software/make/manual/html_node/Instead-of-Execution.html)). For the issue with the go dependencies not being in the cache, I updated `docs.yaml` to run `go mod download` to ensure they're included.

Also, I consolidated `Start controller/API` and `Wait for controller to be up` steps so you can see everything Kit is doing in the logs for the former, which will help prevent these kind of oversights in the future. The reason I didn't catch the issue with binaries being rebuilt in https://github.com/argoproj/argo-workflows/pull/14012 is because the logs are discarded unless the build failed (in which case they were visible in the `Failure debug - Controller/API logs` step).
### Verification

Wait for E2E tests

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -355,9 +355,8 @@ jobs:
             LOG_LEVEL=info \
             API=${{matrix.use-api}} \
             UI=false \
-            POD_STATUS_CAPTURE_FINALIZER=true > /tmp/argo.log 2>&1 &
-      - name: Wait for controller to be up
-        run: make wait PROFILE=${{matrix.profile}} API=${{matrix.use-api}}
+            POD_STATUS_CAPTURE_FINALIZER=true 2>&1 | tee /tmp/argo.log &
+          make wait PROFILE=${{matrix.profile}} API=${{matrix.use-api}}
         timeout-minutes: 5
       - name: Run tests ${{matrix.test}}
         run: make ${{matrix.test}} E2E_SUITE_TIMEOUT=20m STATIC_FILES=false
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -280,8 +280,6 @@ jobs:
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /usr/local/.ghcup /opt/hostedtoolcache/CodeQL
           printf "==> Available space after cleanup\n"
           df -h
-      - name: Install socat (needed by Kubernetes)
-        run: sudo apt-get -y install socat
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -300,6 +298,8 @@ jobs:
         with:
           python-version: '3.x'
           cache: pip
+      - name: Install socat (needed by Kubernetes) and kit
+        run: sudo apt-get -y install socat && make kit
```
MasonM, line 302: Adding `make kit` isn't necessary, since `kit` is a prerequisite for `make start`, but it's noisy, and I figure it's better to avoid polluting the logs for `Start controller/API`

tczhao:
👍Would you have any metrics on how much performance this improves?

MasonM:
@tczhao The first two redundancies impact the time for the "Start controller/API" step in the E2E test jobs. The new "Performance Metrics" page can show [metrics for the E2E jobs as a whole](https://github.com/argoproj/argo-workflows/actions/metrics/performance?tab=jobs&filters=workflow_file_name%3Aci-build.yaml), but it doesn't look at has step-level granularity. From a very quick look, it appears it reduces the time for that step by 2-3 minutes.

The third redundancy impacts everything Go-related, and the fix in this PR to avoid re-downloading Go dependencies isn't going to help with that until the cache is updated. I think it'll be 10-20 seconds.

tczhao (approved):
👍

PR #14032, opened 2024-12-26
fix(controller): step group stuck on running when exit hook has illegal expression

### LINKED ISSUE #14031: Workflow got stuck due to condition expression evaluation error
### Pre-requisites

Due to #13986 , we are trying to upgrade our argo to v3.6.2
After upgrading, our workflows which would be successful in `v3.4.8` are stuck in `v3.6.2`
And there is no any error logs either in `workflow-controller` and `argoexec`
截圖 2024-12-25 晚上8 10 58

after some investigations, I found the issue is occurred in the condition expression
```
hooks:
  exit:
    template: on-exit
    expression: steps["prepare"].outputs != null
```

and the lib used for expression validation has been changed in both v3.6.2 and v3.5.13
```
# v3.4.8
"github.com/antonmedv/expr"

# v3.5.13, v3.6.2
"github.com/expr-lang/expr"
```

the new exp validator cannot recognize null and return an error. 
(null and nil can be recognized in old validator.)
(the error didn't logged to workflow-controller)
```
err: unknown name null (1:29)\n | steps[\"prepare\"].outputs != null\n | ............................^
```

and the step-template handler just marked the node as `completed = false`, not error
```
# argo-workflows/workflow/controller/exit_handler.go
...
if exitHook.Expression != "" {
	execute, err = argoexpr.EvalBool(exitHook.Expression, env.GetFuncMap(template.EnvMap(woc.globalParams.Merge(scope.getParameters()))))
	if err != nil {
		return true, nil, err
	}
}

---

# argo-workflows/workflow/controller/steps.go
...
if childNode.Completed() {
    hasOnExitNode, onExitNode, err := woc.runOnExitNode(ctx, step.GetExitHook(woc.execWf.Spec.Arguments), childNode, stepsCtx.boundaryID, stepsCtx.tmplCtx, "steps."+step.Name, stepsCtx.scope)
    if hasOnExitNode && (onExitNode == nil || !onExitNode.Fulfilled() || err != nil) {
        // The onExit node is either not complete or has errored out, return.
        completed = false
    }
}

if !completed {
	return node, nil
}
```
I think that's what makes the workflow stuck
Can anyone help with this issue?

### Version(s)

v3.5.13, v3.6.2

```YAML
kind: Workflow
metadata:
  name: invalid-expression
spec:
  arguments:
    parameters:
    - name: message
      value: "prepare message"
  entrypoint: main
  templates:
    - name: main
      inputs:
        parameters:
          - name: message
      steps:
        - - name: prepare
            when: '{{=inputs.parameters.message != ""}}'
            template: prepare
            arguments:
              parameters:
                - name: message
                  value: '{{inputs.parameters.message}}'
        - - name: execute-script
            template: execute-target-script
            hooks:
              exit:
                template: on-exit
                expression: steps["prepare"].outputs != null
                arguments:
                  parameters:
                    - name: message
                      value: '{{steps.prepare.outputs.parameters.message}}'
    - name: execute-target-script
      script:
        image: 'python:alpine3.6'
        command: [ 'python' ]
        source: 'print("hello argo")'
    - name: prepare
      inputs:
        parameters:
          - name: message
      outputs:
        parameters:
          - name: message
            valueFrom:
              path: /tmp/message
      script:
        image: debian:9.4
        command: [ bash ]
        source: |
          echo '{{inputs.parameters.message}}'
          echo '{{inputs.parameters.message}}' > /tmp/message
    - name: on-exit
      inputs:
        parameters:
          - name: message
      script:
        image: debian:9.4
        command: [ bash ]
        source: |
          echo '{{inputs.parameters.message}}'
```

```text
```

```text
```

chengjoey:
Fixes #14031

### Motivation
The  error handling logic for `stepGroup` on `exithook` with illegal expressions should be consistent with `Dag`.

https://github.com/argoproj/argo-workflows/blob/27a59385f62de98b7bddc0087b03df90af29e87e/workflow/controller/dag.go#L289-L300

### DIFF
--- a/workflow/controller/steps.go
+++ b/workflow/controller/steps.go
@@ -109,7 +109,7 @@ func (woc *wfOperationCtx) executeSteps(ctx context.Context, nodeName string, tm
 
 		sgNode, err := woc.executeStepGroup(ctx, stepGroup.Steps, sgNodeName, &stepsCtx)
 		if err != nil {
-			return nil, err
+			return woc.markNodeError(sgNodeName, err), nil
 		}
 		if !sgNode.Fulfilled() {
 			woc.log.Infof("Workflow step group node %s not yet completed", sgNode.ID)
@@ -330,8 +330,12 @@ func (woc *wfOperationCtx) executeStepGroup(ctx context.Context, stepGroup []wfv
 			completed = false
 		} else if childNode.Completed() {
 			hasOnExitNode, onExitNode, err := woc.runOnExitNode(ctx, step.GetExitHook(woc.execWf.Spec.Arguments), childNode, stepsCtx.boundaryID, stepsCtx.tmplCtx, "steps."+step.Name, stepsCtx.scope)
-			if hasOnExitNode && (onExitNode == nil || !onExitNode.Fulfilled() || err != nil) {
-				// The onExit node is either not complete or has errored out, return.
+			// see https://github.com/argoproj/argo-workflows/issues/14031,
+			// we should return error otherwise the node will get stuck
+			if err != nil {
+				return node, err
+			}
+			if hasOnExitNode && (onExitNode == nil || !onExitNode.Fulfilled()) {
 				completed = false
 			}
 		}
--- a/util/expr/argoexpr/eval_test.go
+++ b/util/expr/argoexpr/eval_test.go
@@ -58,6 +58,17 @@ func TestEvalBool(t *testing.T) {
 			want:    true,
 			wantErr: false,
 		},
+		{
+			name: "test null expression",
+			args: args{
+				input: "steps[\"prepare\"].outputs != null",
+				env: map[string]interface{}{"steps": map[string]interface{}{
+					"prepare": map[string]interface{}{"outputs": "msg"},
+				}},
+			},
+			want:    false,
+			wantErr: true,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


workflow/controller/steps.go
```diff
@@ -330,7 +330,10 @@ func (woc *wfOperationCtx) executeStepGroup(ctx context.Context, stepGroup []wfv
 			completed = false
 		} else if childNode.Completed() {
 			hasOnExitNode, onExitNode, err := woc.runOnExitNode(ctx, step.GetExitHook(woc.execWf.Spec.Arguments), childNode, stepsCtx.boundaryID, stepsCtx.tmplCtx, "steps."+step.Name, stepsCtx.scope)
-			if hasOnExitNode && (onExitNode == nil || !onExitNode.Fulfilled() || err != nil) {
+			if err != nil {
+				return node, err
+			}
+			if hasOnExitNode && (onExitNode == nil || !onExitNode.Fulfilled()) {
 				// The onExit node is either not complete or has errored out, return.
```
jswxstw, line 337: It would be best to update the explanation in the comments as well.

jswxstw:
I ran the example from #14031 and found that node `StepGroup[1]` still remains `Running`. This issue seems to be quite common when workflow is marked as `Error` directly.

workflow/controller/steps.go
chengjoey: done

jswxstw (commented):
https://github.com/argoproj/argo-workflows/blob/27a59385f62de98b7bddc0087b03df90af29e87e/workflow/controller/steps.go#L110-L113

I think there is also need to mark `sgNode` as `Error` as I commented before: https://github.com/argoproj/argo-workflows/pull/14032#issuecomment-2562737735.
```golang
if err != nil {
    return woc.markNodeError(sgNodeName, err), nil
}
```

chengjoey:
> I think there is also need to mark `sgNode` as `Error` as I commented before: [#14032 (comment)](https://github.com/argoproj/argo-workflows/pull/14032#issuecomment-2562737735).

thanks @jswxstw , done

jswxstw (approved):

tczhao (approved):

PR #14034, opened 2024-12-27
feat: Fixes #8646 visualize wf before submitting

### LINKED ISSUE #8646: UI: Visualize a wf template before submitting (steps it will run, arguments it will take, etc..)
Currently, we do not have a way to visualise the spec of a workflow. A very long time ago I tried to produce a low-code way to specify workflows that used the existing graph component to visualise the specification. Essentially, you should each template as a node, and DAGs and steps as graphs. It all links together nicely.

Unperceivable:
Fixes #8646 

### Motivation

Stumbled across the open [issue](https://github.com/argoproj/argo-workflows/issues/8646) while developing a complex workflow, thought I might as well give it a go myself.

#### Disclaimer
First time doing any front-end development.

### Modifications

Added a "graph view" tab to all workflow/template creation/submission pages to visualize the workflow as a graph before submission.

- Preview graph from submit-new-workflow
Workflow
- Preview graph from submit-new-workflow using workflowTemplateRef
Workflow Ref
- Preview graph from workflow-templates:
Workflow Template
- Preview graph from cluster-workflow-templates:
Cluster Workflow
- Preview graph from cron-workflow:
Cron Workflow

### Verification

Ran all workflows/workflow-templates under ui/test/ui/ and visually compared the predicted graph view to the resultant output.
- Preview vs Run, Steps example:
Preview vs Run Steps example
- Preview vs Run, Dag example:
Preview vs Run Dag example
- Preview vs Run, Dag+Steps example:
Preview vs Dag+Steps

### Possible Improvements
- Add node detail view/panel
- Make it a Graph Editor/Allow workflow modifications through graph view
- Improve general UX
  - Ability to hide node type and/or edge labels
  - Dedicated Color/Icon? Currently using Skipped + Clock becuase Pending(Yellow) was harder to read due to low contrast.
  - Node names/labels are truncated, this may no longer be best option since "executionStrategy" is included in the label.
  - Edge labels are small and hard to read, sometimes overlapping with node names
  - Long node types such as StepGroup do not fit to the node.

### DIFF
--- a/ui/src/shared/components/editors/graph-viewer.tsx
+++ b/ui/src/shared/components/editors/graph-viewer.tsx
@@ -2,11 +2,64 @@ import * as React from 'react';
 
 import {genres} from '../../../workflows/components/workflow-dag/genres';
 import {WorkflowDagRenderOptionsPanel} from '../../../workflows/components/workflow-dag/workflow-dag-render-options-panel';
-import {DAGTask, Template, Workflow, WorkflowTemplate} from '../../models';
+import {DAGTask, Template, Workflow, WorkflowTemplate, CronWorkflow, ClusterWorkflowTemplate} from '../../models';
 import {GraphPanel} from '../graph/graph-panel';
 import {Graph} from '../graph/types';
+import {services} from '../../services';
+import {useEffect, useState} from 'react';
 
-export function GraphViewer({workflow}: {workflow: Workflow | WorkflowTemplate}) {
+export function GraphViewer({workflowDefinition}: {workflowDefinition: Workflow | WorkflowTemplate | ClusterWorkflowTemplate | CronWorkflow}) {
+    const [workflow, setWorkflow] = useState<Workflow | WorkflowTemplate | ClusterWorkflowTemplate>(workflowDefinition);
+    const [isLoading, setIsLoading] = useState(true); // To track if data is still loading
+
+    useEffect(() => {
+        if (workflow.spec.workflowTemplateRef) {
+            setWorkflowFromRefrence(workflow.spec.workflowTemplateRef.name).then(() => {
+                setIsLoading(false);
+            });
+        } else {
+            setIsLoading(false);
+        }
+    }, [workflow]);
+
+    if (isLoading) {
+        return <div>Loading...</div>;
+    }
+
+    const state = {
+        expandNodes: new Set(''),
+        showArtifacts: false,
+        showInvokingTemplateName: false,
+        showTemplateRefsGrouping: false,
+    };
+
+    const name = workflow.metadata.name ?? `${workflow.metadata.generateName}${generateNamePostfix(5)}`;
+    const graph = populateGraphFromWorkflow(workflow, name);
+
+    return (
+        <GraphPanel
+            storageScope='workflow-dag'
+            graph={graph}
+            nodeGenresTitle={'Node Type'}
+            nodeGenres={genres}
+            nodeClassNamesTitle={'Node Phase'}
+            nodeClassNames={{Pending: true}}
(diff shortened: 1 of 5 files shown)


ui/src/shared/components/editors/graph-viewer.tsx
```diff
@@ -0,0 +1,356 @@
...
+        classNames: 'Skipped',
+        progress: 0,
+        icon: 'clock'
+    };
+
+    useEffect(() => {
+        if ('workflowTemplateRef' in workflowDefinition.spec && isLoading) {
+            setWorkflowFromRefrence(workflowDefinition.spec.workflowTemplateRef.name).then(() => {
+                setIsLoading(false);
+            });
```
tczhao, line 36: cluster workflow template use case is missing
https://github.com/argoproj/argo-workflows/tree/main/examples/cluster-workflow-template

ui/src/shared/components/editors/graph-viewer.tsx
```diff
@@ -0,0 +1,356 @@
...
+            hideNodeTypes={false}
+            hideOptions={false}
+            options={<WorkflowDagRenderOptionsPanel {...state} onChange={workflowDagRenderOptions => saveOptions(workflowDagRenderOptions)} />}
+        />
+    );
+
+    function setWorkflowFromRefrence(name: string): Promise<void> {
+        return services.workflowTemplate.get(name, workflowDefinition.metadata.namespace).then(workflowTemplate => {
+            setWorkflow(workflowTemplate);
+        });
```
tczhao, line 74: need to handle error if template is not found

tczhao (changes requested):
Looks incredible!

ui/src/shared/components/editors/graph-viewer.tsx
Unperceivable: Just to clarify, the usecase is for when you manually submit a workflow using workflowTemplateRef for a clusterWorkflowTemplate?  Because currently there is no way to select a clusterWorkflowTemplate directly via the UI right?  
image
I assumed clusterWorkflowTemplates could not be submitted via reference for some reason...

ui/src/shared/components/editors/graph-viewer.tsx
Unperceivable: Added error handling: 180a14222
image

ui/src/shared/components/editors/graph-viewer.tsx
Unperceivable: Added clusterWorkflowTemplate reference support: 180a14222
image

Unperceivable: kind is optional so I have to cascade the calls either way.

tczhao: Thanks for all the screenshots and commit ref, they are really helpful

---

For `workflowTemplateRef` to use clusterworkflow template, the `clusterScope: true` needs to be there, perhaps we should use this to determine if we should do a clusterworkflow template lookup instead of the current fallback method

Unperceivable: ah ok, clusterScope is also optional but has to be available when it's a cluster template. That makes sense, thanks for explaining.   
Updated the logic: 0fdb470c8

tczhao (approved):

mohdmsl:
I see that this feature has been merged but has this been not released?

amit-disc:
Any Update on this? I also don't see this feature in the UI

tadeboro:
As far as I can tell, this was released as part of the v3.7 version that was released cca. 1 month ago.

PR #14037, opened 2024-12-29
refactor(ui):  artifacts table Fixes #8306

### LINKED ISSUE #8306: UI: Adjusting Artifacts table
# Summary

In would be great to have workflow Artifacts table on UI more readable.

This can be reached by:
-  adding column names
- adding column separators to clearly define where one ends and other starts
- adjusting fields width. It seems to me that the time column is excessively wide and at the same time the artifact name and the task name columns are too limited.
- making table scrollable horizontally for dealing with long names. Thus, they will always occupy one line.

# Use Cases

 It takes time to read and understand artifact info now, especially on first columns whose values that may merge with each other.
And I'm enclosing a screenshot to show how Artifacts looks from my side at the moment.

image

---

Unperceivable:
Fixes #8306

### Motivation

From original [issue](https://github.com/argoproj/argo-workflows/issues/8306)
> It takes time to read and understand artifact info now, especially on first columns whose values that may merge with each other.
(https://github.com/argoproj/argo-workflows/issues/8306)
previous implementation

### Modifications

Refactored workflow-details artifacts white-box to include table columns and expand a cell with text too long for it's cell on hover.
image

Added same functionality to workflow-node-info -> inputs/outputs
image

### DIFF
--- a/ui/src/workflows/components/workflow-artifacts.tsx
+++ b/ui/src/workflows/components/workflow-artifacts.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
 
-import {Timestamp} from '../../shared/components/timestamp';
+import {Timestamp, TimestampSwitch} from '../../shared/components/timestamp';
 import * as models from '../../shared/models';
 import {services} from '../../shared/services';
-import {TIMESTAMP_KEYS} from '../../shared/use-timestamp';
+import useTimestamp, {TIMESTAMP_KEYS} from '../../shared/use-timestamp';
 
 interface Props {
     workflow: models.Workflow;
@@ -12,6 +12,7 @@ interface Props {
 
 export function WorkflowArtifacts(props: Props) {
     const workflowStatusNodes = (props.workflow.status && props.workflow.status.nodes) || {};
+    const [storedDisplayISOFormat, setStoredDisplayISOFormat] = useTimestamp(TIMESTAMP_KEYS.WORKFLOW_NODE_ARTIFACT_CREATED);
     const artifacts =
         Object.keys(workflowStatusNodes)
             .map(nodeName => {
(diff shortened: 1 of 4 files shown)


ui/src/workflows/components/workflow-artifacts.tsx
```diff
@@ -40,21 +40,34 @@ export function WorkflowArtifacts(props: Props) {
     return (
         <div className='white-box'>
             <div className='white-box__details'>
+                <div className='row header'>
+                    <div className='columns download'>Download</div>
```
tczhao, line 44: The download symbol in each row is self-explanatory enough.
we probably don't need this download column, or download column name, what do you think?

tczhao (changes requested):
The table seems to resize when hovered at the cell, could you fix that

Unperceivable:
> The table seems to resize when hovered at the cell, could you fix that  
That was the intended change, from Modifications:  
> expand a cell with text too long for it's cell on hover.  
The alternative suggested in the ticket was a horizontal scroll feature. I figured there would be edge cases where some columns would not appear until you scroll.  
I am fine implementing whatever solution though.

tczhao (changes requested):
Thank you for the fix, last few things:
1. Could you make sure the table values align with the column name, e.g., left align? People will definitely raise an issue about this later on.
image
2. we are already on artifact panel, rather than "Artifact Name", using "Name" is probably sufficient enough
image

4. I think we better have the timeformat switch button at the column title instead of individual artifact table value, just like the workflow list page. (we could tackle this in a separate PR, create an issue for this and a separate PR)
image

Unperceivable:
> 1. Could you make sure the table values align with the column name, e.g., left align? People will definitely raise an issue  

Fixed padding issue ef7e1a11b  
> 2. we are already on artifact panel, rather than "Artifact Name", using "Name" is probably sufficient enough

Renamed column. but not in the node overview since there is no label to define artifacts in input/output 2f3b9da87   
> 3. I think we better have the timeformat switch button at the column title instead of individual artifact table value, just 
 like the workflow list page. (we could tackle this in a separate PR, create an issue for this and a separate PR)   

Moved timestamp switch to column fe5110658  
image  
image

tczhao (approved):
looks great! Thanks for the contribution and happy new year

PR #14039, opened 2024-12-30
fix: ensure namespace parallelism and parallelism work together. Fixes  #10985

### LINKED ISSUE #10985: Cannot set both parallelism and namespaceParallelism
### Pre-requisites

In the `workflow-controller-configmap`, setting:
``` 
  parallelism: "10"
  namespaceParallelism: "5"
```

**Expectation**
The global number of running workflows to be limited at 10, and the running workflows per namespace limited to 5.

**Actual**
However, in testing, argo-workflows limits the global number of workflows to 5.

Having each setting individually works as expected, but not when they are together. This is a cluster installation.

### Version

v3.4.5

Simple workflow that sleeps 5 mins.
```YAML
kind: Workflow
metadata:
  generateName: parallelism-test-
spec:
  entrypoint: test
  podGC:
  ttlStrategy:
    secondsAfterCompletion: 60
  templates:
  - name: test
    steps:
    - - name: test-1
        template: sleep
  - name: sleep
    container:
      image: alpine:latest
      command: [sleep, "300"]
```

Script that submits the above workflow 10 times to a namespace
```sh
#!/bin/bash

for i in {1..10}
do
    argo submit workflow.yaml -n $1 &
done
```

Submit to `default` and `devops` namespace
```
./submit.sh default
./submit.sh devops
```

### Screenshots from UI, notice the number of running workflows for each namespace, and stuck waiting.
Screenshot 2023-04-26 at 4 28 52 PM

Screenshot 2023-04-26 at 4 27 26 PM

[controller_logs.txt](https://github.com/argoproj/argo-workflows/files/11337003/controller_logs.txt)

I don't think this is relevant.

isubasinghe:
Fixes  #10985

### Motivation

The current throttler code does not work when used in unison with namespace parallelism and parallelism. This is because the concept of the chain throttler unfortunately doesn't work. 

### Modifications

Corrected a bug where workflows weren't added into the running list/map when the parallelism was set to 0.
Added tracking for namespace counts into a new throttler.

### Verification

Relied on tests to pass


MasonM:
@isubasinghe The E2E tests are broken due to a bug in Kit. I have a fix for it in https://github.com/argoproj/argo-workflows/pull/14048. Could you please review that PR?

isubasinghe:
> @isubasinghe The E2E tests are broken due to a bug in Kit. I have a fix for it in #14048. Could you please review that PR?

Thanks, I will do

workflow/sync/multi_throttler_test.go
```diff
@@ -0,0 +1,171 @@
...
+	assert.False(t, throttler.Admit("a/1"))
+	assert.False(t, throttler.Admit("b/0"))
+	assert.False(t, throttler.Admit("b/1"))
+	throttler.Remove("a/0")
+	assert.True(t, throttler.Admit("b/1"))
+}
+
+func TestMultiWithParallelismLimitAndPriority(t *testing.T) {
+	queuedKey := ""
+	throttler := NewMultiThrottler(2, make(map[string]int), 2, func(key string) { queuedKey = key })
```
Joibel, line 45: The original test had no namespace parallelism, so the second 2 here perhaps should be 0 to better mimic that test.

workflow/sync/multi_throttler.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (m *multiThrottler) queueThrottled() {
+	if m.totalParallelism != 0 && len(m.running) >= m.totalParallelism {
+		return
+	}
+
+	minPq := &priorityQueue{itemByKey: make(map[string]*item)}
+
+	cnts := make(map[string]int)
```
Joibel, line 142: This dubiously named variable appears to be write only, remove?

workflow/sync/multi_throttler.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+type Key = string
+type QueueFunc func(Key)
+
+// NewMultiThrottler creates a new multi throttler for throttling both namespace and global parallelism, a parallelism value of zero disables throttling
+func NewMultiThrottler(parallelism int, namespaceParallelism map[string]int, namespaceParallelismDefault int, queue QueueFunc) Throttler {
+	return &multiThrottler{
+		queue:                       queue,
+		namespaceParallelism:        namespaceParallelism,
```
Joibel, line 35: I understand why you have this, but as it is unusable by workflow users and I had imagined it might be a callback (to copy the pattern set by `queueFunc`) I wonder if we should keep in in this PR?

I'm OK with leaving it in, it's just a bit odd in this PR.

workflow/sync/multi_throttler.go
```diff
@@ -0,0 +1,243 @@
...
+	return nil
+}
+
+func (m *multiThrottler) namespaceCount(namespace string) (int, int) {
+	setLimit, has := m.namespaceParallelism[namespace]
+	if !has {
+		m.namespaceParallelism[namespace] = m.namespaceParallelismDefault
+		setLimit = m.namespaceParallelismDefault
+	}
+	count := 0
```
Joibel, line 82: Suggest early exit if setLimit==0 to save counting all the semaphores.

workflow/sync/multi_throttler_test.go
```diff
@@ -0,0 +1,171 @@
...
+	assert.True(t, throttler.Admit("a/1"))
+
+	assert.False(t, throttler.Admit("a/2"))
+	assert.False(t, throttler.Admit("a/3"))
+	assert.False(t, throttler.Admit("a/4"))
+	assert.False(t, throttler.Admit("b/1"))
+
+	throttler.Add("c/0", 8, time.Now())
+	assert.True(t, throttler.Admit("c/0"))
+}
```
Joibel, line 171: This test proves the priority between namespaces behaves as I suggest we document it:

```
func TestPriorityAcrossNamespaces(t *testing.T) {
	throttler := NewMultiThrottler(3, make(map[string]int), 1, func(Key) {})
	throttler.Add("a/0", 0, time.Now())
	throttler.Add("a/1", 0, time.Now())
	throttler.Add("a/2", 0, time.Now())
	throttler.Add("b/0", 1, time.Now())
	throttler.Add("b/1", 1, time.Now())
	throttler.Add("b/2", 1, time.Now())

	assert.True(t, throttler.Admit("a/0"))
	assert.True(t, throttler.Admit("b/0"))
	assert.False(t, throttler.Admit("a/1"))
	assert.False(t, throttler.Admit("a/2"))
	assert.True(t, throttler.Admit("b/0"))
	assert.False(t, throttler.Admit("b/1"))
	assert.False(t, throttler.Admit("b/2"))
	throttler.Remove("a/0")
	assert.False(t, throttler.Admit("b/1"))
	assert.True(t, throttler.Admit("a/1"))
	throttler.Remove("b/0")
	assert.True(t, throttler.Admit("b/1"))
	assert.False(t, throttler.Admit("a/2"))
}
```

Joibel (changes requested):
Also needs some documentation in `docs/parallelism.md` to document that when using namespaceParallelism a lower priority workflow in a namespace that has capacity will run before a high priority workflow in a namespace with no capacity.

I assume this is the intentional behaviour - it feels the sane option.

workflow/controller/controller_test.go
```diff
@@ -953,6 +953,7 @@ func TestNotifySemaphoreConfigUpdate(t *testing.T) {
 	for i := 0; i < 3; i++ {
 		key, _ := controller.wfQueue.Get()
 		controller.wfQueue.Done(key)
+		controller.wfQueue.Forget(key)
```
Joibel, line 956: Why this change?

workflow/sync/multi_throttler.go
isubasinghe: I could hide the implementation detail (map as an argument) from this PR in that case. I mean to immediately follow up with a PR that passes in a config containing limits for each namespace.

workflow/sync/multi_throttler.go
isubasinghe: hahahaha, yes I meant to remove this, thanks.

workflow/controller/controller_test.go
isubasinghe: It was a bug in the previous test I believe, if the rate limiter has any entries, it will be added later. This is makes the test flakey as a result because it now relies on timings.

isubasinghe:
The test have gotten flakier, unsure if this contributed or something else did.

workflow/controller/controller_test.go
Joibel: :+1:

workflow/sync/multi_throttler.go
Joibel: This will presumably not be a constant map (as in parallelism for a namespace will be modifyable on the fly without restarting the controller). The expectation was an annotation on the namespace or a configmap in the namespace.

workflow/sync/multi_throttler_test.go
Joibel: Did you intend to add this test case - I assume that's what a :+1: might have meant

workflow/sync/multi_throttler.go
isubasinghe: No it will not be, I will follow up with a change to make it dynamic.

workflow/sync/multi_throttler_test.go
isubasinghe: Ah I meant to add this, sorry

Joibel:
This PR seems to be the cause of the SignalsSuite test failures - they almost always fail with this change, and reverting makes them much better.

Issue #11266, opened 2023-06-26
Full CRD installation fails


ramanNarasimhan77:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I can confirm the issues exists when I tested with `:latest`
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/master/docs/CONTRIBUTING.md))

### What happened/what you expected to happen?

I tried installing the *FULL* CRD by updating the kustomization.yaml

```shell
$ git diff manifests/base/crds/kustomization.yaml
diff --git a/manifests/base/crds/kustomization.yaml b/manifests/base/crds/kustomization.yaml
index 3ccdade18..4bc23b6ee 100644
--- a/manifests/base/crds/kustomization.yaml
+++ b/manifests/base/crds/kustomization.yaml

@@ -2,4 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization

 resources:
-  - minimal
+  - full

$ kubectl create ns argo
$ kubectl apply -k manifests/cluster-install

```

Errors:
```text
Error from server (Invalid): error when creating "manifests/cluster-install": CustomResourceDefinition.apiextensions.k8s.io "clusterworkflowtemplates.argoproj.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[templateDefaults].properties[dag].properties[tasks].items.properties[inline].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[templateDefaults].properties[steps].items.items: Required value: must be specified, spec.validation.openAPIV3Schema.properties[spec].properties[templates].items.properties[dag].properties[tasks].items.properties[inline].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[templates].items.properties[steps].items.items: Required value: must be specified]


[...]

Issue #7393, opened 2021-12-13
Plugin template OpenAPI validation


alexec:
# Summary

Currently, the structure of a plugin template is free-from, it can even be a string:

```yaml
plugin:
  hello: "yes"
```

Thats fine. But we could allow plugin developers to define the OpenAPI schema and validate it when the template is created.

# Use Cases

When would you use this?

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

alexec:
The reason I did not include this is v1 was because, validation will also happen by the plugin itself. So this would duplicate that and might therefore not be needed.

PR #14044, opened 2025-01-02
fix(manifests): fix full CRDs and update tests to use them. Fixes #8532

### LINKED ISSUE #8532: Full CRDs are incomplete and causes json2jsii to fail
## Summary
The generated CustomResourceDefinitions seem to be incomplete because of the `steps` property.

https://github.com/argoproj/argo-workflows/blob/ecd91b1c4215a2ab8742f7c43eaade98a1d47eba/manifests/base/crds/full/argoproj.io_workflowtemplates.yaml#L7673-L7676

The `items` property has a `type` property but does not have an accompanying `items` property. 

 This causes `json2jsii` to fail. 
https://github.com/cdklabs/json2jsii/blob/60e891447e35a8bb1a9dac3496d6adce87d73292/src/type-generator.ts#L534-L540
This is an issue for tools that depend on `json2jsii`. https://github.com/cdk8s-team/cdk8s/issues/2157

However, the json schema has these missing properties.

https://github.com/argoproj/argo-workflows/blob/ecd91b1c4215a2ab8742f7c43eaade98a1d47eba/api/jsonschema/schema.json#L7315-L7321

https://github.com/argoproj/argo-workflows/blob/ecd91b1c4215a2ab8742f7c43eaade98a1d47eba/api/jsonschema/schema.json#L6449-L6454

It looks like this problem was fixed at one point with #4828, but was later reverted by @JPZ13 here #4810 because of the CRDs becoming too long. 

Is there way we can get the full CRDs for use with `json2jsii` as I do not believe it has any limit on CRD size like Kubernetes does. I'd like to have all the types for `steps` so that I can get strong typing with cdk8s.

---

MasonM:
Fixes #8532. Partial fix for #8190

### Motivation

The full CRDs at `manifests/base/crds/full` have many errors that make them only usable in editors (https://github.com/argoproj/argo-workflows/issues/11266#issuecomment-1637198777). They used to be installable in a cluster, but now that's no longer possible:
```bash
$ kubectl apply --server-side -k manifests/base/crds/full/
customresourcedefinition.apiextensions.k8s.io/workflowartifactgctasks.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workfloweventbindings.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workflowtaskresults.argoproj.io serverside-applied
Error from server (Invalid): CustomResourceDefinition.apiextensions.k8s.io "clusterworkflowtemplates.argoproj.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[templateDefaults].properties[dag].properties[tasks].items.properties[inline].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[templateDefaults].properties[steps].items.items: Required value: must be specified, spec.validation.openAPIV3Schema.properties[spec].properties[templates].items.properties[dag].properties[tasks].items.properties[inline].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[templates].items.properties[steps].items.items: Required value: must be specified]
Error from server (Invalid): CustomResourceDefinition.apiextensions.k8s.io "cronworkflows.argoproj.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[workflowSpec].properties[templateDefaults].properties[dag].properties[tasks].items.properties[inline].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[workflowSpec].properties[templateDefaults].properties[steps].items.items: Required value: must be specified, spec.validation.openAPIV3Schema.properties[spec].properties[workflowSpec].properties[templates].items.properties[dag].properties[tasks].items.properties[inline].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[workflowSpec].properties[templates].items.properties[steps].items.items: Required value: must be specified]
...
```

Having working full CRDs would enable many useful features, such as `kubectl explain` output (https://github.com/argoproj/argo-workflows/issues/8190), [validating admission policies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) (https://github.com/argoproj/argo-workflows/pull/14045), and integration with tools like cdk8s (https://github.com/argoproj/argo-workflows/issues/8532).

### Modifications

This PR looks intimidating, but 99% of the changes are in `manifests/base/crds/full`, which is not used in any of the release manifests. I tried to avoid changing the release manifests or API as much as possible, so there shouldn't be any backwards-compatibility concerns.

As explained at https://github.com/argoproj/argo-workflows/pull/13754#issuecomment-2412529946, there were two issues that prevented us from using the full CRDs everywhere:
1. [Kubernetes size limits](https://github.com/kubernetes/kubernetes/issues/82292) when using client-side apply. 
2. Wrong validation information due to kubebuilder limitations.

For the first issue, I verified that this is no longer an issue when using [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/), so in https://github.com/argoproj/argo-workflows/pull/14048, I updated the `make install` target to use it. For the second issue, I went through and added workarounds to `hack/manifests/crds.go` for all the kubebuilder issues I could find. Since the E2E test suite now uses the full CRDs, it should tell us if there's anything I missed.

I didn't update the release manifests to use the full CRDs, since that's a big change and we probably should wait awhile to make sure there aren't any unexpected issues. In the meantime, users can easily opt into the full CRDs if they want.

### Verification
Ran tests locally. Also, I verified the documentation changes using `make docs-serve`.


test/e2e/manifests/components/base/kustomization.yaml
```diff
@@ -10,8 +12,3 @@ patches:
 - path: workflow-controller-deployment.yaml
 - path: workflow-controller-cluster-workflow-template-rbac.yaml
 - path: minio-deployment.yaml
-
-labels:
-- includeSelectors: true
-  pairs:
-    app.kubernetes.io/part-of: argo
```
MasonM, line 17: FYI: This label was only needed when using the old `--prune` method. When using [apply sets](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#alternative-kubectl-apply-f-directory-prune), Kubernetes automatically generates the appropriate label.

test/e2e/malformed_resources_test.go
```diff
@@ -19,22 +18,12 @@ type MalformedResourcesSuite struct {
...
-		WorkflowName("malformed").
-		When().
-		// it is not possible to wait for this to finish, because it is malformed
-		Wait(3 * time.Second).
-		Then().
-		ExpectWorkflow(func(t *testing.T, metadata *metav1.ObjectMeta, status *wfv1.WorkflowStatus) {
-			assert.Equal(t, "malformed", metadata.Name)
-			assert.Equal(t, wfv1.WorkflowFailed, status.Phase)
-		})
+	s.Given().Exec("kubectl", []string{"apply", "-f", "testdata/malformed/malformed-workflow.yaml"}, fixtures.ErrorOutput("unknown field \"spec.arguments.parameters.someParam\""))
```
MasonM, line 21: Had to rewrite the tests in this file because malformed workflows will now be rejected at the API level instead of at runtime. We could test the runtime behavior of malformed workflows by adding a new profile that uses the minimal CRDs and having `test-executor` use that, but that seems overkill.

docs/installation.md
```diff
@@ -25,6 +25,16 @@ You can use Kustomize to patch your preferred [configurations](managed-namespace
 
 You can install Argo Workflows using the community maintained [Helm charts](https://github.com/argoproj/argo-helm).
 
+### Full CRDs
+
+The official release manifests come with stripped-down CRDs that omit validation information.
+This is a workaround for [Kubernetes size limitations](https://github.com/kubernetes/kubernetes/issues/82292) when using client-side apply.
+As of version 3.7, the full CRDs can be installed using [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) via the following command:
```
MasonM, line 32: If this doesn't go out as part of 3.7, I'll remove this before it's merged

MasonM:
I split off the minor build refactoring (e.g. switching to server-side apply) to a separate PR to make this one easier to review: https://github.com/argoproj/argo-workflows/pull/14048

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -663,6 +666,7 @@ type Template struct {
 	HTTP *HTTP `json:"http,omitempty" protobuf:"bytes,42,opt,name=http"`
 
 	// Plugin is a plugin template
+	// +kubebuilder:pruning:PreserveUnknownFields
```
Joibel, line 669: Why do we need this?

test/e2e/cron_test.go
```diff
@@ -458,25 +456,16 @@ spec:
 
 func (s *CronSuite) TestMalformedCronWorkflow() {
 	s.Given().
-		Exec("kubectl", []string{"apply", "-f", "testdata/malformed/malformed-cronworkflow.yaml"}, fixtures.NoError).
+		Exec("kubectl", []string{"apply", "-f", "testdata/malformed/malformed-cronworkflow.yaml"}, fixtures.ErrorOutput("unknown field \"spec.workflowSpec.arguments.parameters.someParam\"")).
```
Joibel, line 459: This test feels odd. Why did it apparently test a wellformed-cronworkflow..., who knows? Just as a delay for the Events to get emitted?

Anyway, can we drop the wellformed workflow from this test, or if it isn't tested elsewhere it should be moved to be in it's own test.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -1502,6 +1506,7 @@ type WorkflowStep struct {
 	Template string `json:"template,omitempty" protobuf:"bytes,2,opt,name=template"`
 
 	// Inline is the template. Template must be empty if this is declared (and vice-versa).
+	// +kubebuilder:pruning:PreserveUnknownFields
```
Joibel, line 1509: Same question, why do we allow unknown fields in a template, that feels very odd?

hack/manifests/crds.go
```diff
@@ -1,108 +1,107 @@
...
+}
 
-	shouldMinimize := false
-	if len(data) > 1024*1024 {
-		fmt.Printf("Minimizing %s due to CRD size (%d) exceeding 1MB\n", filename, len(data))
-		shouldMinimize = true
-	}
+// Workaround for missing generateName (see https://github.com/kubernetes-sigs/controller-tools/pull/640)
+func patchMetadata(schema *obj, baseFields ...string) {
+	schema.SetNestedField(map[string]string{"type": "string"}, append(baseFields, "generateName")...)
```
Joibel, line 42: Oh, yuck. Nice fix.

docs/installation.md
Joibel: :+1:

Joibel (commented):
Mostly just questions to see how you made some decisions here.

pkg/apis/workflow/v1alpha1/workflow_types.go
MasonM: Because plugins are defined as free-form objects. See https://github.com/argoproj/argo-workflows/issues/7393

Without this, the `TestExecutorPluginsSuite/TestTemplateExecutor` test fails:
```
Submitting workflow  template-executor-
W0111 01:58:53.394874   40529 warnings.go:70] unknown field "spec.templates[0].plugin.hello"
    executor_plugins_test.go:29: expected exactly one key, got 0
=== FAIL: ExecutorPluginsSuite/TestTemplateExecutor
```

pkg/apis/workflow/v1alpha1/workflow_types.go
MasonM: Because this is a recursive object (since the nested template can potentially contain steps/dags that also define `inline`), which aren't allowed in CRD, and causes kubebuilder to generate an empty object, e.g. https://github.com/argoproj/argo-workflows/blob/a7a72e7ca3c2bbf127b65fdcf1b5b85fbd2aad19/manifests/base/crds/full/argoproj.io_workflowtasksets.yaml#L3676

That empty object is the cause of the `Required value: must not be empty for specified object fields` errors you get when trying to import the full CRDs on the `main` branch (see the PR description).

I can't think of an easy fix for this. There's talk about this case in https://github.com/kubernetes/kubernetes/issues/62872, but I doubt `$ref` support is going to be added anytime soon (if ever). I guess I could update `hack/manifests/crds.go` to extract the normal definition for `Template`, walk through it to remove any nested `inline` fields, then splice that back in, but that would significantly increase the size of the schema and be a pain.

test/e2e/cron_test.go
MasonM: I'm not sure either, so I removed it in d3f2ee7377a1bd03ec470d44ac5cb491cfa55fb5

pkg/apis/workflow/v1alpha1/workflow_types.go
MasonM: I left some comments in 987bb2e71fccdbe9b58de91ac02ba64911b6c466 explaining this

MasonM:
The failing test (`SignalsSuite/TestStopBehavior`) is clearly unrelated and also failing in the last build on the `main` branch: https://github.com/argoproj/argo-workflows/actions/runs/12707760387/job/35432247793

test/e2e/manifests/components/base/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 
 resources:
-- ../../../../../manifests/base/crds/minimal
+- ../../../../../manifests/base/crds/full
```
terrytangyuan, line 5: This then won't test the minimal CRDs?

MasonM: Nope. There isn't much to test with the minimal CRDs, besides maybe ensuring they validate against the schema.  If you think that's worth testing, I could add a step to each E2E test job that runs `kubectl apply --server-side --validate=strict --dry-run=server --kustomize manifests/base/crds/minimal/`

MasonM: @terrytangyuan Would you like me to enter a PR that adds `kubectl apply --server-side --validate=strict --dry-run=server -f 'manifests/*.yaml'` to validate the release manfiests?

MasonM:
@Joibel Do you have any other questions? Is there anything else you'd like to see in this PR?

pkg/apis/workflow/v1alpha1/workflow_types.go
MasonM: I left some comments in https://github.com/argoproj/argo-workflows/commit/987bb2e71fccdbe9b58de91ac02ba64911b6c466 explaining this

test/e2e/manifests/components/base/kustomization.yaml
terrytangyuan: Yes please add that step. Thanks!

MasonM: Done! https://github.com/argoproj/argo-workflows/pull/14089

terrytangyuan (approved):
LGTM. Thanks!

PR #14001, opened 2024-12-16
test: deduplicate test manifests and fix "stress" profile

MasonM:
### Motivation
The duplication of the Kustomization files under `test/e2e/manifests/` makes them hard to follow. Also, the "stress" profile is broken, which I'm guessing was introduced in 87cb1559107ec88dd418229b38113d70ba2a8580:
```
$ make install PROFILE=stress
...
error: accumulating resources: accumulation err='merging resources from 'workflow-controller-podpriorityclass.yaml': may not add resource with an already registered id: PriorityClass.v1.scheduling.k8s.io/workflow-controller.[noNs]': must build at directory: '/home/vscode/go/src/github.com/argoproj/argo-workflows/test/e2e/manifests/stress/workflow-controller-podpriorityclass.yaml': file is not directory
```

### Modifications
I used [Kustomize Components](https://kubectl.docs.kubernetes.io/guides/config_management/components/) to clean up the manifests by extracting common configuration to a `base` component.

I also added a `local-argo` component to handle scaling `workflow-controller` and `argo-server` so we can remove the corresponding lines from `Makefile`, since it's cleaner to let `kustomize` handle that.

### Verification

I used the following script to verify nothing unexpected changed:
```
set -e
for PROFILE in events minimal mysql plugins postgres prometheus sso; do
  echo "diffing $PROFILE"
  git checkout --quiet main
  kubectl kustomize --load-restrictor=LoadRestrictionsNone test/e2e/manifests/$PROFILE > "old_$PROFILE.yaml"
  git checkout --quiet deduplicate-manifest
  kubectl kustomize --load-restrictor=LoadRestrictionsNone test/e2e/manifests/$PROFILE > "new_$PROFILE.yaml"
  diff -U3 "old_$PROFILE.yaml" "new_$PROFILE.yaml" || true
done
```

Output:
```
diffing events
--- old_events.yaml	2024-12-16 04:33:35.012533594 +0000
+++ new_events.yaml	2024-12-16 04:33:35.242545139 +0000
@@ -3448,11 +3448,12 @@
   - clusterworkflowtemplates
   - clusterworkflowtemplates/finalizers
   verbs:
+  - create
+  - delete
+  - watch
   - get
   - list
   - watch
-  - create
-  - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
@@ -3759,6 +3760,17 @@
 ---
 apiVersion: v1
 kind: Secret
+metadata:
+  annotations:
[...]

PR #14048, opened 2025-01-03
build: minor upgrades/refactoring to support full CRDs

MasonM:
### Motivation

This was split off from https://github.com/argoproj/argo-workflows/pull/14044 to make it easier to review. This makes a few minor improvements to the build to support those changes.

### Modifications

1. Update devcontainer to install k3s 1.29.10, since that's what we use in CI: https://github.com/argoproj/argo-workflows/blob/ef41f83f801a6ff48c87f882a6b75d0e37529134/.github/workflows/ci-build.yaml#L263 1.27 is EOL and doesn't support things like validation rules in CRDs.
2. Update `make install` to use [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/), which is meant to replace client-side apply and isn't affected by size limitations for the CRDs. Since `kubectl apply --server-side` isn't compatible with `kubectl apply --prune`, I had to switch to [apply sets](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#alternative-kubectl-apply-f-directory-prune), which is intended to replace allow list pruning, and seems to work just as well.
3. Minor refactoring of the manifests under `manifests/` to use [Kustomize Components](https://kubectl.docs.kubernetes.io/guides/config_management/components/) so that we can share code with the the manifests under `test/e2e/manifests` without duplication. See #14001 for more details on this approach.

### Verification

Built devcontainer locally and ran tests


manifests/components/sso/dex/kustomization.yaml
```diff
@@ -1,11 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-labels:
-  - includeSelectors: true
-    pairs:
-      app.kubernetes.io/part-of: dex
```
MasonM, line 7: We don't need to add this anymore since apply-set pruning automatically adds the appropriate label ([source](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#alternative-kubectl-apply-f-directory-prune)):
> With ApplySet-based pruning, kubectl adds the applyset.kubernetes.io/part-of=<parentID> label to each object in the set before they are sent to the server.

Makefile
```diff
@@ -530,7 +533,8 @@ ifeq ($(shell uname),Darwin)
 	brew tap kitproj/kit --custom-remote https://github.com/kitproj/kit
 	brew install kit
 else
-	curl -q https://raw.githubusercontent.com/kitproj/kit/main/install.sh | tag=v0.1.8 sh
+	@echo "Downloading Kit"
+	curl -fsL --retry 99 "https://github.com/kitproj/kit/releases/download/v0.1.8/kit_0.1.8_$$(uname)_$$(uname -m | sed 's/aarch64/arm64/').tar.gz" | sudo tar -C /usr/local/bin -xzf - kit
```
MasonM, line 537: This is a workaround for CI failures due to a bug introduced in https://raw.githubusercontent.com/kitproj/kit/main/install.sh recently: https://github.com/kitproj/kit/commit/46fe614511f819b2d2fa4cbab733a3a4397c998c

isubasinghe (approved):

Issue #7444, opened 2021-12-19
Mount volumes defined in a WorkflowTemplate when it's called by another workflow


cscetbon:
# Summary

### What change needs making?

Volumes set on a WorkflowTemplate should be used when called by another workflow

# Use Cases

### When would you use this?

Here is what doesn't work today

```
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: template1
spec:
  entrypoint: template1
  volumeClaimTemplates:
  - metadata:
      name: my-pvc
    spec:
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 5M
  arguments: &parameters
    parameters: 
    - name: run
  templates:
    - name: template1
      inputs: *parameters
      steps:
        - - name: template10
            template: template10
            arguments:
              parameters:
                - name: run
                  value: "{{inputs.parameters.run}}"
        - - name: template11
            templateRef:
              name: template2
              template: template2

    - name: template10
      inputs: *parameters
      script:
        image: debian:9.4
        command: [bash]
        source: |
          echo "{{inputs.parameters.run}}" > /mnt/pvc/output
        volumeMounts:
          - name: my-pvc
            mountPath: /mnt/pvc

---
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: template2
spec:
  entrypoint: template2
  templates:
    - name: template2
      script:
        image: debian:9.4
        command: [bash]
        source: |
          cat /mnt/pvc/output
        volumeMounts:
          - name: my-pvc
            mountPath: /mnt/pvc
---
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  name: cronwf-ex1
spec:
  schedule: 0 11 * * *
  concurrencyPolicy: Forbid
  workflowSpec:
    entrypoint: jobs
    # volumeClaimTemplates:
    # - metadata:
    #     name: my-pvc
    #   spec:
    #     accessModes: [ "ReadWriteOnce" ]
    #     resources:
    #       requests:
    #         storage: 5M
    templates:
    - name: jobs
      steps:
      - - name: template1
          templateRef:
            name: template1
            template: template1
          arguments:
            parameters:
            - name: run
              value: "{{item}}"
          withItems:
            - run1
            # -... 
            - run23
```

**works well**
_argo submit --from wftmpl/template1 --watch_ 

[...]

PR #14056, opened 2025-01-07
feat(controller): support volumeClaimTemplates can be referenced by templateRef. Fixes #13977 #7444

### LINKED ISSUE #13977: volumeClaimTemplates in WorkflowTemplate works only with workflowTemplateRef and not templateRef
### Pre-requisites

`volumeClaimTemplates` defined in WorkflowTemplate does not work when workflow template referred to via `templateRef`. 
It errors with the following message `
volume 'workdir' not found in workflow spec`

However, it works when WorkflowTemplate is referred to via `workflowTemplateRef`.
Documentation does not mention anything about this limitation.

Use case:
Multiple pods running in parallel  in the same instance of workflow can only use `ReadWriteOnce` access mode but this does not work when pods are scheduled on different nodes. Since there's no requirement for pods to share the same volume we would like to create volume per pod and the number of pods varies so number of volumes cannot be determined upfront.

### Version(s)

v3.5.12

```YAML
#Converted https://github.com/argoproj/argo-workflows/blob/main/examples/volumes-pvc.yaml to WorkflowTemplate as follows:

# This example demonstrates the ability for a workflow to create a
# temporary, ephemeral volume used by the workflow, and delete it
# when the workflow completes. It uses the same volumeClaimTemplates
# syntax as statefulsets.
kind: WorkflowTemplate
metadata:
  name: volumes-pvc-wftmpl
spec:
  entrypoint: volumes-pvc-example
  volumeClaimTemplates:
  - metadata:
      name: workdir
    spec:
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 64Mi

  templates:
  - name: volumes-pvc-example
    steps:
    - - name: generate
        template: whalesay
    - - name: print
        template: print-message

  - name: whalesay
    container:
      image: docker/whalesay:latest
      command: [sh, -c]
      args: ["echo generating message in volume; cowsay hello world | tee /mnt/vol/hello_world.txt"]
      volumeMounts:
      - name: workdir
        mountPath: /mnt/vol

  - name: print-message
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo getting message from volume; find /mnt/vol; cat /mnt/vol/hello_world.txt"]
      volumeMounts:
      - name: workdir
        mountPath: /mnt/vol
--- 
# This works
kind: Workflow
metadata:
  generateName: volumes-pvc-
spec:
  entrypoint: volumes-pvc-example
  workflowTemplateRef:
    name: volumes-pvc-wftmpl
---
# This does not work
kind: Workflow
metadata:
  generateName: volumes-pvc2-
spec:
  entrypoint: main
  templates:
  - name: main
    dag:
      tasks:
      - name: failed-volume
        templateRef:
          name: volumes-pvc-wftmpl
          template: volumes-pvc-example
```

```text
N/A
```

```text
N/A
```

chengjoey:
Fixes #13977 #7444

### Motivation

In order for volumeClaimTemplates to be referenced by the `templateRef`

The purpose is to automatically bring the corresponding volumeClaimTemplates into the current workflow when using `templateRef`.

### Modifications
1. add volumeClaimTemplates field in `template`
2. workflow operator create the pvcs referred by template

### Verification

e2e test


shuangkun:
looks good. please resolve conflicts!

workflow/controller/operator.go
```diff
@@ -1934,6 +1934,27 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
...
+				existed = true
+				break
+			}
+		}
+		if !existed {
+			vols = append(vols, vol)
+		}
+	}
+	if len(vols) > 0 {
+		woc.execWf.Spec.VolumeClaimTemplates = append(woc.execWf.Spec.VolumeClaimTemplates, vols...)
```
Joibel, line 1951: execWf.Spec should be considered immutable (constant), this isn't good practice.

Joibel (changes requested):
Is there any limit on how many times `createPVCs()` will be called, it looks like it will be called repeatedly.

workflow/controller/operator.go
chengjoey: okay, I might be able to put this part of the logic inside `setExecWorkflow(ctx context.Context)`, and reserve all the PVCs when initializing execWf.

b0ch3nski:
Hey :slightly_smiling_face:

Anything still blocking this feature? We are patiently waiting especially for this with the use case of starting multiple Workflow Templates (each having it's own volume) using `withParam` based on output of previous steps.

Currently we are forced to declare single big volume at the Workflow level which is quite inconvenient then to share it with multiple concurrent child templates.

Thanks!

workflow/controller/operator.go
chengjoey: @Joibel https://github.com/argoproj/argo-workflows/blob/581950f7ae9f039a10d392d40a111f8f33e06eb3/workflow/controller/operator.go#L148-L150

`execWf` is a copy, modifying `execWf` is not good practice, but is this a feasible solution?

kiblik:
Hello. Any update on this PR? We would like to use it.

PR #14072, opened 2025-01-10
docs: Pod Security Context: Volumes aren't required in >= v3.3 and runAsNonRoot

paulcwatts:
### Motivation

A note regarding volumes being required for `runAsNonRoot` took me down an incorrect path before I saw that was unnecessary in newer versions. I'm hoping this will save other developers a couple of hours of going down the same path.

### Modifications

Based on the PR discussion, I removed the note since v3.3 is no longer supported.

### DIFF
--- a/docs/workflow-pod-security-context.md
+++ b/docs/workflow-pod-security-context.md
@@ -24,5 +24,7 @@ You can configure this globally using [workflow defaults](default-workflow-specs
 !!! Warning "It is easy to make a workflow need root unintentionally"
     You may find that user's workflows have been written to require root with seemingly innocuous code. E.g. `mkdir /my-dir` would require root.
 
-!!! Note "You must use volumes for output artifacts"
-    If you use `runAsNonRoot` - you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
+!!! Note "You must use volumes for output artifacts (v3.3 or earlier)"
+    If you use `runAsNonRoot` in versions v3.3 or earlier, you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
+    In versions later than v3.3, the [Emissary executor](https://argo-workflows.readthedocs.io/en/latest/workflow-executors/#emissary-emissary)
+    allows artifacts on the base layer with `runAsNonRoot`.
--- a/docs/workflow-pod-security-context.md
+++ b/docs/workflow-pod-security-context.md
@@ -23,8 +23,3 @@ You can configure this globally using [workflow defaults](default-workflow-specs
 
 !!! Warning "It is easy to make a workflow need root unintentionally"
     You may find that user's workflows have been written to require root with seemingly innocuous code. E.g. `mkdir /my-dir` would require root.
-
-!!! Note "You must use volumes for output artifacts (v3.3 or earlier)"
-    If you use `runAsNonRoot` in versions v3.3 or earlier, you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
-    In versions later than v3.3, the [Emissary executor](workflow-executors.md#emissary-emissary)
-    allows artifacts on the base layer with `runAsNonRoot`.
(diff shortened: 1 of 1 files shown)


docs/workflow-pod-security-context.md
```diff
@@ -24,5 +24,7 @@ You can configure this globally using [workflow defaults](default-workflow-specs
 !!! Warning "It is easy to make a workflow need root unintentionally"
     You may find that user's workflows have been written to require root with seemingly innocuous code. E.g. `mkdir /my-dir` would require root.
 
-!!! Note "You must use volumes for output artifacts"
-    If you use `runAsNonRoot` - you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
+!!! Note "You must use volumes for output artifacts (v3.3 or earlier)"
+    If you use `runAsNonRoot` in versions v3.3 or earlier, you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
+    In versions later than v3.3, the [Emissary executor](https://argo-workflows.readthedocs.io/en/latest/workflow-executors/#emissary-emissary)
```
MasonM, line 29: Since the docs are versioned, we should use relative links here to ensure it goes to the right version:
```suggestion
    In versions later than v3.3, the [Emissary executor](workflow-executors.md#emissary-emissary)
```

MasonM (commented):
Thanks for your contribution! One small comment, but otherwise LGTM

MasonM (approved):

docs/workflow-pod-security-context.md
paulcwatts: Thanks, I applied the change!

docs/workflow-pod-security-context.md
```diff
@@ -24,5 +24,7 @@ You can configure this globally using [workflow defaults](default-workflow-specs
 !!! Warning "It is easy to make a workflow need root unintentionally"
     You may find that user's workflows have been written to require root with seemingly innocuous code. E.g. `mkdir /my-dir` would require root.
 
-!!! Note "You must use volumes for output artifacts"
-    If you use `runAsNonRoot` - you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
+!!! Note "You must use volumes for output artifacts (v3.3 or earlier)"
+    If you use `runAsNonRoot` in versions v3.3 or earlier, you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
+    In versions later than v3.3, the [Emissary executor](workflow-executors.md#emissary-emissary)
+    allows artifacts on the base layer with `runAsNonRoot`.
```
tczhao, line 30:
```suggestion
!!! Note "You must use volumes for output artifacts (v3.3 or earlier with non-emissary executor)"
    If you use `runAsNonRoot` in versions v3.3 or earlier with non-emissary executor, you cannot have output artifacts on base layer (e.g. `/tmp`). You must use a volume (e.g. [empty dir](empty-dir.md)).
    The [Emissary executor](workflow-executors.md#emissary-emissary) (available from v3.0, default since v3.3)
    allows artifacts on the base layer with `runAsNonRoot`.
```

Let's be more precise with the wording.
Otherwise, we could remove this line since <= v3.4 is no longer maintained

tczhao (dismissed):

docs/workflow-pod-security-context.md
Joibel: For all supported versions you don't need volumes it seems, so the whole note could just be removed?

paulcwatts: When I originally submitted the PR, I didn't know the docs were versioned or that 3.3 was no longer supported.

With that knowledge, if it were my project, I would also lean toward removing the note completely. 

The purpose for submitting the PR in the first place was hopefully to save people time from going down an wrong path.

paulcwatts: Should I just apply @tczhao 's suggestion, or should I remove the note completely?

tczhao: @paulcwatts Let's remove it then

paulcwatts: OK, removed!

MasonM (approved):

Joibel (approved):

PR #13491, opened 2024-08-21
fix: Mark taskResult completed if wait container terminated not gracefully. Fixes #13373

jswxstw:
Fixes #13373 

### Motivation

https://github.com/argoproj/argo-workflows/issues/12103#issuecomment-2296174496

### Modifications

Mark its taskResult as completed if pod has been terminated not gracefully.

### Verification

This issue only occurs when `wait` container is terminated not gracefully, it is hard to reproduce it.
1. submit a workflow with a long running pod.
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: sleep-example-
spec:
  entrypoint: main
  templates:
  - name: main
    steps:
    - - name: sleep
        template: sleep

  - name: sleep
    container:
      image: alpine:latest
      command: [sh, -c, sleep 3600]
```

2. `kill -9` wait/main container
```shell
ctr task kill --signal 9 8c66238d4723b9237c4e71d3fd16710f5f985d2ce8f7218ff987e143b3e8140b # wait container
ctr task kill --signal 9 5da93afce108bc59f746e8c8df791ae6c808deaedaa87abc8f6ea5f4697da2df # main container
```

3. Pod `Error` and wait container does not write `LabelKeyReportOutputsCompleted` to `true`
```shell
# sleep-example-tdnk7-sleep-1943014577   0/2     Error     0            9m21s
# k logs -f sleep-example-tdnk7-sleep-1943014577 -c wait
time="2024-08-21T03:55:14.519Z" level=info msg="Starting Workflow Executor" version=untagged
time="2024-08-21T03:55:14.522Z" level=info msg="Using executor retry strategy" Duration=1s Factor=1.6 Jitter=0.5 Steps=5
time="2024-08-21T03:55:14.522Z" level=info msg="Executor initialized" deadline="2024-08-21 04:00:13 +0000 UTC" includeScriptOutput=false namespace=argo podName=sleep-example-tdnk7-sleep-1943014577 templateName=sleep version="&Version{Version:untagged,BuildDate:2024-08-15T12:58:34Z,GitCommit:c143e3e78c1b6d585d64a2d9363bf0c9c64caf6c,GitTag:untagged,GitTreeState:clean,GoVersion:go1.22.6,Compiler:gc,Platform:linux/amd64,}"
time="2024-08-21T03:55:14.529Z" level=debug msg="Create workflowtaskresults 201"
time="2024-08-21T03:55:14.530Z" level=info msg="Starting deadline monitor"
[...]

PR #14088, opened 2025-01-15
fix: pod finalizer removal and odd pod status

Joibel:
WIP

#### Part 1

If pod finalizers are in use they should not prevent pod deletion after the pod is complete.

For example: If you have a podGC.strategy of `OnPodSuccess` with a `deleteDelayDuration` set and you delete the owning Workflow during the deleteDelayDuration then the pod will remain until `deleteDelayDuration` expires. If the workflow-controller is restarted during this window the pod is orphaned, with the finalizer still in place.

`blockOwnerDeletion` in the `ownerReference` of a pod does not prevent the owner (Workflow) being deleted in all circumstances

#### Part 2

It is possible for a node to disappear from a cluster as a surprise. In this case the Pod ContainerStatus could remain in running (because the container never went into any further state), whilst the Pod's own Status is in Error.

This PR attempts to recognise this case and set the Workflow Node status accordingly.

#### Split this PR

This PR should probably be two PRs for each of these, but for now I'm making this to allow full CI to check it over.

### Verification

Updated unit test for Part 1

### DIFF
--- a/workflow/controller/pod_cleanup.go
+++ b/workflow/controller/pod_cleanup.go
@@ -1,6 +1,8 @@
 package controller
 
 import (
+	"slices"
+
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
@@ -37,9 +39,10 @@ func (woc *wfOperationCtx) queuePodsForCleanup() {
 		if !nodePhase.Fulfilled() {
 			continue
 		}
-		switch determinePodCleanupAction(selector, pod.Labels, strategy, workflowPhase, pod.Status.Phase) {
+		switch determinePodCleanupAction(selector, pod.Labels, strategy, workflowPhase, pod.Status.Phase, pod.Finalizers) {
 		case deletePod:
 			woc.controller.queuePodForCleanupAfter(pod.Namespace, pod.Name, deletePod, delay)
+		case removeFinalizer:
 		case labelPodCompleted:
 			woc.controller.queuePodForCleanup(pod.Namespace, pod.Name, labelPodCompleted)
 		}
@@ -52,6 +55,7 @@ func determinePodCleanupAction(
 	strategy wfv1.PodGCStrategy,
 	workflowPhase wfv1.WorkflowPhase,
 	podPhase apiv1.PodPhase,
+	finalizers []string,
 ) podCleanupAction {
 	switch {
 	case !selector.Matches(labels.Set(podLabels)): // if the pod will never be deleted, label it now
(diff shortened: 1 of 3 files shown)


isubasinghe (approved):
This makes sense, I'm not entirely confident every edge case is accounted for but lets ship it for now.

Joibel:
@shuangkun and @jswxstw, I'd like to hear your thoughts on these changes if you have time.

Joibel:
I realise now that #13491 does something similar (I originally wrote this patch against the 3.5 codebase).

In the scenario we're seeing with a non-evicted pod, the workflow node still gets stuck with #13491 in place.

workflow/controller/pod_cleanup.go
```diff
@@ -37,9 +39,10 @@ func (woc *wfOperationCtx) queuePodsForCleanup() {
 		if !nodePhase.Fulfilled() {
 			continue
 		}
-		switch determinePodCleanupAction(selector, pod.Labels, strategy, workflowPhase, pod.Status.Phase) {
+		switch determinePodCleanupAction(selector, pod.Labels, strategy, workflowPhase, pod.Status.Phase, pod.Finalizers) {
 		case deletePod:
 			woc.controller.queuePodForCleanupAfter(pod.Namespace, pod.Name, deletePod, delay)
+		case removeFinalizer:
```
shuangkun, line 45: Reasonable, is it need to add 
```
wfc.queuePodForCleanup(p.Namespace, p.Name, removeFinalizer) 
```
here？

Joibel: Yeah, oops. I had thought the final parameter was `action`. Will fix.

isubasinghe: Nice catch @shuangkun, I missed this in the review.

isubasinghe: Maybe change this section to this: 
```go
		action := determinePodCleanupAction(selector, pod.Labels, strategy, workflowPhase, pod.Status.Phase, pod.Finalizers)
		if action == deletePod {
			woc.controller.queuePodForCleanupAfter(pod.Namespace, pod.Name, action, delay)
		} else {
			woc.controller.queuePodForCleanup(pod.Namespace, pod.Name, action)
		}
```

jswxstw:
> It is possible for a node to disappear from a cluster as a surprise. In this case the Pod ContainerStatus could remain in running (because the container never went into any further state), whilst the Pod's own Status is in Error. We have seen this in real clusters, but it is rare.

Indeed, this issue https://github.com/kubernetes/kubernetes/issues/98718 in older versions of Kubernetes can also cause inconsistencies between pod status and container status, which can lead to the workflow getting stuck.

PR #10784, opened 2023-03-30
fix: Allow script and container image to be set in templateDefault. Fixes #9633

RoryDoherty:
Made changes to validation to check if the script/container Image has been set in templateDefaults

**Testing**
Added unit tests to verify these situations

Also created the following via the locally deployed UI and then submitted both via the UI and they were successful:
```
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  generateName: demo-
spec:
  entrypoint: demo
  templateDefaults:
    script:
      image: alpine:latest
      command: [sh]
      workingDir: "/src"
  templates:
    - name: demo
      script:
        source: |
          pwd
```
and a cron workflow
```
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  generateName: demo-cron
spec:
  schedule: "00 4 * * 2"
  timezone: "UTC"
  workflowMetadata:
    generateName: demo-cron-test-
  workflowSpec:
    entrypoint: demo
    templateDefaults:
      script:
        image: alpine:latest
        command: [sh]
        workingDir: "/src"
    templates:
      - name: demo
        script:
          source: |
            pwd
```

Fixes #9633

PR #14094, opened 2025-01-17
test: refactor E2E examples tests and fix `Makefile`

MasonM:
### Motivation

This rewrites `hack/test-examples.sh` in Go to make those tests easier to debug and facilitate testing more complex scenarios, like the VAP added in https://github.com/argoproj/argo-workflows/pull/14045

This also fixes a `Makefile` bug that @panicboat reported in https://github.com/argoproj/argo-workflows/discussions/14090#discussioncomment-11862439. The expression `$(shell [ $PROFILE = plugins ] && echo false || echo true)` isn't escaping the `$` sign, which causes `make` to substitute the value of the `P` variable, if any. The [official docs](https://www.gnu.org/software/make/manual/html_node/Reference.html) recommend using parentheses:
> A dollar sign followed by a character other than a dollar sign, open-parenthesis or open-brace treats that single character as the variable name. Thus, you could reference the variable x with ‘$x’. However, this practice can lead to confusion (e.g., ‘$foo’ refers to the variable f followed by the string oo) so we recommend using parentheses or braces around all variables, even single-letter variables, unless omitting them gives significant readability improvements

### Modifications
* Added `KubectlApply` helper to reduce duplication
* Replaced `hack/test-example.sh` with `test/e2e/examples_test.go`. The new tests do strict validation, which caught a bug in `examples/webhdfs-input-output-artifacts.yaml`:
```
$ make test-examples
<SNIP>
examples_test.go:28: Error parsing ../../examples/webhdfs-input-output-artifacts.yaml: strict decoding error: unknown field "spec.templates[0].outputs.artifacts[0].overwrite"
=== FAIL: ExamplesSuite/TestExampleWorkflows
FAIL    github.com/argoproj/argo-workflows/v3/test/e2e  1.580s
FAIL
make: *** [Makefile:609: test-examples] Error 1
```
* Fixed `Makefile` bug
### Verification

Ran E2E tests.


hack/test-examples.sh
```diff
@@ -1,27 +0,0 @@
-#!/usr/bin/env bash
-set -eu -o pipefail
-
-# Load the configmaps that contains the parameter values used for certain examples.
-kubectl apply -f examples/configmaps/simple-parameters-configmap.yaml
-
-echo "Checking for banned images..."
-grep -lR 'workflows.argoproj.io/test' examples/*  | while read f ; do
-  echo " - $f"
-  test 0 == $(grep -o 'image: .*' $f | grep -cv 'argoproj/argosay:v2\|python:alpine3.6\|busybox')
```
tczhao, line 10: I'm wondering what are you thoughts on not including this?

Makefile
```diff
@@ -38,7 +38,7 @@ E2E_WAIT_TIMEOUT      ?= 90s # timeout for wait conditions
 E2E_PARALLEL          ?= 20
 E2E_SUITE_TIMEOUT     ?= 15m
 GOTEST                ?= go test -v -p 20
-ALL_BUILD_TAGS        ?= api,cli,cron,executor,examples,corefunctional,functional,plugins
+ALL_BUILD_TAGS        ?= api,cli,cron,executor,examples,corefunctional,functional,plugins,examples
```
tczhao, line 41: duplicate examples

tczhao (changes requested):

hack/test-examples.sh
MasonM: Because this is redundant, since we're already checking the image when `Given.Workflow()` is called: https://github.com/argoproj/argo-workflows/blob/b2a2c8f473483d92813e9752eb699e9a308f3e7f/test/e2e/fixtures/given.go#L86-L139

but now that I look again, that code is actually broken: it's only checking the images if the workflow defines `TemplateDefaults`, since otherwise, `templates` will be an empty array.  I think this was a bug introduced 2 years ago in https://github.com/argoproj/argo-workflows/pull/10784/. 

I fixed the bug so that it's properly verifying images again, but that caused a ton of tests to fail, since there were a lot of tests introduced since then that used a banned image. I went through and fixed everything. While I was at it, I changed everything using the deprecated `python:alpine3.6` image to use `argoproj/argosay:v2`, which removes the need for that check.

Makefile
MasonM: Thanks. I removed it.

test/e2e/testdata/complex-global-artifact-passing.yaml
```diff
@@ -155,15 +155,15 @@ spec:
 
 
   - container:
+      image: argoproj/argosay:v2
       command: [sh, -c]
-      args: ["sleep 1; cat /upload/testUpload; cat /upload/testUpload.txt > /upload/testUpload.txt"]
-      image: alpine:3.18.4
+      args: ["sleep 1; cat /upload/testInput.txt > /upload/testUpload.txt"]
```
MasonM, line 160: This one was a little tricky, because the behavior of `sh` in alpine is actually different than `sh` in `argoproj/argosay:v2`, because the former allows `cat /upload/testUpload.txt > /upload/testUpload.txt` and the latter does not:
```
 docker run --entrypoint=/bin/sh --rm alpine:3.18.4 -xc 'touch /tmp/foo; cat /tmp/foo > /tmp/foo'; echo $?                                               
+ touch /tmp/foo
+ cat /tmp/foo
0

$ docker run --entrypoint=/bin/sh --rm argoproj/argosay:v2 -xc 'touch /tmp/foo; cat /tmp/foo > /tmp/foo'; echo $?                                         
+ touch /tmp/foo
+ cat /tmp/foo
cat: /tmp/foo: input file is output file
1
```

I think the intention of this was to copy the input artifact to the output, so I updated it to do so

tczhao (approved):
Looks good!

PR #14103, opened 2025-01-20
feat: add support for databases enforcing strict data integrity through PKs. Fixes #13611

### LINKED ISSUE #13611: Add primary key to schema_history
# Summary

It would be good to have a primary key in the creation statement for the schema_history table.

## Use Cases

DB replication mechanisms and integrity-checking mechanisms sometimes rely on the ability to identify a single row. Primary keys are made for this.

I bumped into this while using a percona-xtradb database deployed with the Percona Operator (which isn't too uncommon, I had it running in the cluster for another db already, so re-used it for the argo database), which will deny the creation of this table with the default configuration:

```
Failed to init db: Error 1105 (HY000): Percona-XtraDB-Cluster prohibits use of DML command on a table (argo.schema_history) without an explicit primary key with pxc_strict_mode = ENFORCING or MASTER
```
While it might feel like it adds no value to the argo-workflows project, this is a minor change, won't affect anything else, but will allow users to deploy argo easier. I might add a PR if this would be allowed.

The changes would go in migrate.go, where the creation would get a `primary key(schema_version)` and an additional migration line in the big set which would do an alter-add-primary-key statement to try and add the primary key to an already existing table.

---

radusora:
Fixes #13611 

### Motivation

As outlined in #13611, certain database systems depend on the presence of primary keys to support replication and ensure data integrity. For instance, Percona XtraDB’s `pxc_strict_mode` enforces this requirement.

### Modifications

The `schema_history` table lacked a primary key, and adding one through existing migrations was not feasible because earlier migrations would fail in the environments described above. To address this, the table creation statement for `schema_history` was updated to include a primary key. Additionally, code was added to check for the presence of the primary key in existing installations and add it if it is missing (making sure the changes are idempotent).

### Verification

The changes were tested in a local development environment using both MySQL and Postgres, with scenarios including deploying a new cluster and modifying an existing one. Furthermore, a similar test was conducted with a Percona XtraDB cluster deployed via the Percona Operator.

Additionally, I verified that the query against `information_schema.table_constraints` is supported by all the databases we support, and that default permissions should suffice for accessing it.

### DIFF
--- a/persist/sqldb/migrate.go
+++ b/persist/sqldb/migrate.go
@@ -40,12 +40,43 @@ func ternary(condition bool, left, right change) change {
 }
 
 func (m migrate) Exec(ctx context.Context) (err error) {
+
+	dbType := dbTypeFor(m.session)
+
 	{
 		// poor mans SQL migration
-		_, err = m.session.SQL().Exec("create table if not exists schema_history(schema_version int not null)")
+		_, err = m.session.SQL().Exec("create table if not exists schema_history(schema_version int not null, primary key(schema_version))")
 		if err != nil {
 			return err
 		}
+
+		// Ensure the schema_history table has a primary key, creating it if necessary
+		// This logic is implemented separately from regular migrations to improve compatibility with databases running in strict or HA modes
+		dbIdentifierColumn := "table_schema"
+		if dbType == Postgres {
+			dbIdentifierColumn = "table_catalog"
+		}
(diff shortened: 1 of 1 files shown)


MasonM (approved):
Thanks! I tested this locally with both MySQL 8.4.4 and PostgreSQL 12.22

Joibel (approved):
Thanks for making this work

PR #14109, opened 2025-01-21
chore: remove leftover code

jswxstw:
remove leftover code in #5232


tczhao (approved):
Thanks for the fix
looks like everywhere else uses `http.RoundTripper` instead of `WebsocketRoundTripper`

PR #14110, opened 2025-01-21
docs: Add all the template types to the core concepts page

tico24:
### Motivation

I noticed that not all the template types are documented in the template definitions overview so this PR addresses that.
Also formats the page in accordance with the docs guidelines


tczhao (approved):

tczhao:
Thanks for the fix

PR #14116, opened 2025-01-23
docs: fix the table display issue in metrics.md

jswxstw:
### Motivation

Fix the table display issue in [metrics.md](https://argo-workflows.readthedocs.io/en/latest/metrics/)

Clipboard_Screenshot_1737622204

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -250,7 +250,7 @@ A counter of the number of times a CronWorkflow has triggered its `concurrencyPo
 
 |      attribute       |                                   explanation                                    |
 |----------------------|----------------------------------------------------------------------------------|
-| `name`               | ⚠️ The name of the CronWorkflow                                                  |
+| `name`               | ⚠️ The name of the CronWorkflow                                                   |
 | `namespace`          | The namespace that the CronWorkflow is in                                        |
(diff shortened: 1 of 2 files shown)


jswxstw:
https://github.com/argoproj/argo-workflows/blob/514ea3741bd1ce151ce51eba8686ae8e7ef91105/Makefile#L632

@Joibel Please take a look at this when you have time. This will cause the tables in the documentation to not display correctly.

terrytangyuan (commented):
I wonder if we can catch this kind of issues with markdown linter

jswxstw:
> I wonder if we can catch this kind of issues with markdown linter

[Markdownlint](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md) does not have a rule to recognize this type of case.
MkDocs uses a Markdown parser that strictly adheres to the CommonMark specification by default, while other parsers like GitHub Flavored Markdown (GFM) do not.

Joibel:
Thanks for noticing this @jswxstw. I've fixed the generator to do the correct thing with table breaks, and also made it double space the "No attributes" lines for consistency.

I can't trivially make the generator layout the lines with ⚠️ correctly, it should really done with a fix to https://github.com/nao1215/markdown, but I don't think it's worth the effort. Let me know if you disagree.

Joibel (approved):

PR #14127, opened 2025-01-27
chore: revise examples with `onExit` fields

tzing:
### Motivation

Find some examples has the deprecated `onExit` field. This PR revises those examples and follows the latest practice.

### Modifications

2 example manifest got changed.
Removed the `onExit` from step and changed to `hooks.exit.template`.
                

### Verification

Submitted the workflows to a private Argo workflow 3.6.2 service and it passed.


Joibel (approved):
Thanks, nice catch.

PR #14129, opened 2025-01-27
fix: split pod controller from workflow controller

Joibel:
Superceeds #14088 

### Motivation

Split the pod (formerly pod GC) controller from the main workflow controller to make it easier to reason about.

Attempt to handle various conditions whereby pods can become orphaned when pod finalizers are in use.

### Modifications

Split the pod controller components and its queue into the workflows/controller/pod subdirectory.

Callbacks into the workflow controller are provided so that pod events can enqueue workflows for reconciliation.

The pod controller can also now make decisions on it's own. This is especially important for orphaned pods where the workflow has undergone deletion without the pod finalizers being removed. The most obvious case here is a PodGC strategy with delete delay duration. In this case the only place the existence of the pod enqueued for deletion is in the pod GC delayed queue, which is in memory and subject to a workflow-controller restart losing track of it.

Added a number of testing hooks so that the workflow/controller tests didn't need rewriting where they need to inspect the internals of the pod-controller.

### Verification

Existing E2E and unit testing, and a few extensions.

### Documentation

This is mostly a refactor and amendment of code to accommodate more configurations.


workflow/controller/controller_test.go
```diff
@@ -326,14 +326,15 @@ func newController(options ...interface{}) (context.CancelFunc, *WorkflowControl
...
+		wfc.PodController = pod.NewController(ctx, &wfc.Config, wfc.restConfig, "", wfc.kubeclientset, wfc.wfInformer, wfc.metrics, wfc.enqueueWfFromPodLabel)
+
 		wfc.configMapInformer = wfc.newConfigMapInformer()
 		wfc.createSynchronizationManager(ctx)
 		_ = wfc.initManagers(ctx)
 
 		go wfc.wfInformer.Run(ctx.Done())
 		go wfc.wftmplInformer.Informer().Run(ctx.Done())
-		go wfc.podInformer.Run(ctx.Done())
+		go wfc.PodController.Run(ctx, 0)
```
isubasinghe, line 337: I presume this was done to ensure that the pods don't get processed by the workers, so that you may reliably inspect the queue?

workflow/controller/pod/controller.go
```diff
@@ -0,0 +1,322 @@
...
+		if !lastTransition.IsZero() && delayDuration > 0 {
+			delay = time.Until(lastTransition.Add(delayDuration))
+		}
+		c.log.WithFields(logrus.Fields{"action": action, "namespace": pod.Namespace, "podName": pod.Name, "podGC": podGC}).Info("queuing pod", "delay", delay)
+		switch {
+		case delay > 0:
+			c.queuePodForCleanupAfter(pod.Namespace, pod.Name, action, delay)
+		default:
+			c.queuePodForCleanup(pod.Namespace, pod.Name, action)
+		}
```
isubasinghe, line 236: I don't know how I feel about this code and `queuePodsForCleanup` being relatively similar. 
More in the sense that they both drive the reconciliation of pods but live in different packages.

workflow/controller/pod/controller.go
```diff
@@ -0,0 +1,322 @@
...
+	c.commonPodEvent(pod, true)
+}
+
+func newWorkflowPodWatch(ctx context.Context, clientSet kubernetes.Interface, instanceID, namespace *string) *cache.ListWatch {
+	c := clientSet.CoreV1().Pods(*namespace)
+	// completed=false
+	labelSelector := labels.NewSelector().
+		Add(*workflowReq).
+		// not sure if we should do this
+		Add(*incompleteReq).
```
isubasinghe, line 290: I think we should remove this personally, the controllers job is to manage all pods created by argo.
People upgrading into this version might have trouble with pods not being handled with this remaining the same.

isubasinghe (commented):
Looks good but I have some questions/issues

workflow/controller/pod/controller.go
```diff
@@ -0,0 +1,322 @@
...
+					return
+				}
+				podController.updatePodEvent(oldPod, newPod)
+			},
+			DeleteFunc: func(obj interface{}) {
+				podController.deletePodEvent(obj)
+			},
+		},
+	)
+	go podController.podInformer.Run(ctx.Done())
```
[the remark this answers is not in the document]
Joibel: Good point, I'll move it.

workflow/controller/pod/controller.go
Joibel: There is no path whereby the pods have the finalizer and are not `label`led as completed. In this state they drop out of view of the informer, yes. Monitoring pods that are marked as completed is a waste of resources, owner reference to the workflow does all the necessary work.

workflow/controller/pod/controller.go
Joibel: We discussed this on a call.

`queuePodsForCleanup` is the call into the pod-controller to reduce the kubernetes load and make life saner.
Thoughts:
* We must await the node status update to be successfully written to the workflow. This eliminates pure pod observation to know when we can do anything to them, unless we add a lot of extra load to k8s by writing something to the pods so the pod-controller can see the change.
* An option: The workflow informer is also observed by the pod-controller and it then unwraps the workflows to observe node statuses. This is costly in terms of CPU, and may also involve getting this information from the status offload database. Avoided for the double dip into workflow update events as we already know at this point we need to do a thing, and it's unclear that this is much less evil.
* `queuePodsForCleanup` know which pods are in need of cleanup (by doing the naughty list operation) because they are still in the informer (e.g. not marked complete or deleted) thus minimizing the number needing checking. If we sent this event via a queue into the pod-controller we would have to send the entire set of node statuses and teach the pod-informer to understand node to pod mapping.

isubasinghe (approved):
Just a small comment, good to go after thats been addressed.

PR #14132, opened 2025-01-27
docs: delete duplicate entry in example Role

ralphbean:
This PR does not fix any issue on file.

### Motivation

I just happened to notice the problem when reading the docs.

### Modifications

Simply deleted a duplicate entry that looks like a typo.

### Verification

None.


MasonM (approved):

Joibel (approved):
Thanks for taking the effort to make a PR to fix this.

PR #14142, opened 2025-02-01
ci: test k8s version 1.32

blkperl:
### Motivation

Verifying the test suite passes on the new K8s release

### Modifications

Changes CI to run against K8s 1.32

### Verification

E2E tests will suffice

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -259,15 +259,15 @@ jobs:
             profile: minimal
             use-api: true
           - test: test-executor
-            install_k3s_version: v1.29.10+k3s1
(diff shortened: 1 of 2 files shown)


MasonM (approved):
Thanks! I wonder if we should go ahead and drop support for 1.29. That would allow us to start including ValidatingAdmissionPolicies in the release manifests (see https://github.com/argoproj/argo-workflows/pull/14045), since that became stable in 1.30. cc @Joibel

terrytangyuan (approved):

terrytangyuan:
Yeah we can probably start dropping it now. Ref: https://endoflife.date/kubernetes

PR #14145, opened 2025-02-03
fix: Use force checkouts in git artifact. Fixes #14141

### LINKED ISSUE #14141: artifact build failed to load: failed to checkout "xxx": worktree contains unstaged changes
### Pre-requisites

When using the following git artifact configuration:

```yaml
      inputs:
        artifacts:
          - name: build
            path: /build
            git:
              repo: '{{inputs.parameters.repo}}'
              branch: '{{= sprig.trimPrefix(''refs/heads/'', inputs.parameters.branch) }}'
              revision: '{{inputs.parameters.commit}}'
              depth: 0
              usernameSecret:
                name: github-secrets
                key: username
              passwordSecret:
                name: github-secrets
                key: password
              singleBranch: true
```

I get the following error: `artifact build failed to load: failed to checkout "xxx": worktree contains unstaged changes`

After following the logic to this line: https://github.com/argoproj/argo-workflows/blob/release-3.6/workflow/artifacts/git/git.go#L167

I found that enabling a force checkout fixes my problem:

```go
import (
        ...
	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
)

	auth, err := ssh.NewSSHAgentAuth("git")
	if err != nil {
		return fmt.Errorf("failed auth: %w", err)
	}

	repoName := "git@github.com:DavidGamba/dgtools.git"
	commit := "84cdf63980b49f1cbe73e2e62fac6e2feba9b033"
	branch := "master"

	cloneOptions := &git.CloneOptions{
		URL:             repoName,
		Auth:            auth,
		Depth:           0,
		ReferenceName:   plumbing.NewBranchReferenceName(branch),
		SingleBranch:    true,
		InsecureSkipTLS: true,
		Progress:        os.Stdout,
	}
	repo, err := git.PlainClone("/tmp/gittest", false, cloneOptions)
	if err != nil {
		return fmt.Errorf("failed clone: %w", err)
	}

	// checkout
	wt, err := repo.Worktree()
	if err != nil {
		return fmt.Errorf("failed worktree: %w", err)
	}
	err = wt.Checkout(&git.CheckoutOptions{
		Hash:  plumbing.NewHash(commit),
		Force: true,
	})
	if err != nil {
		return fmt.Errorf("failed checkout: %w", err)
	}


```

And removing the force checkout reproduces the problem.

I don't see any reason why Force checkout can't be the default and I am willing to submit the PR if you are OK with it.

Logs from init container:

```
0ef7b6432925e49882cb4294adccf5912ec,GitTag:v3.6.2,GitTreeState:clean,GoVersion:go1.23.3,Compiler:gc,Platform:linux/amd64,}"
time="2025-01-31T19:38:29 UTC" level=info msg="Loading script source to /argo/staging/script"
time="2025-01-31T19:38:29 UTC" level=info msg="Start loading input artifacts..."
time="2025-01-31T19:38:29 UTC" level=info msg="Downloading artifact: build"
time="2025-01-31T19:38:31 UTC" level=info msg="Load artifact" artifactName=build duration=2.396653351s error="failed to checkout \"0ed01a498b9be55d63b769b077003a4e60b5d6f0\": worktree contains unstaged changes" key=
time="2025-01-31T19:38:31 UTC" level=error msg="executor error: artifact build failed to load: failed to checkout \"0ed01a498b9be55d63b769b077003a4e60b5d6f0\": worktree contains unstaged changes"
time="2025-01-31T19:38:31 UTC" level=info msg="Alloc=60052 TotalAlloc=386522 Sys=148987 NumGC=15 Goroutines=5"
Error: artifact build failed to load: failed to checkout "0ed01a498b9be55d63b769b077003a4e60b5d6f0": worktree contains unstaged changes
Usage:
  argoexec init [flags]
Flags:
  -h, --help   help for init
Global Flags:
      --as string                      Username to impersonate for the operation
      --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --as-uid string                  UID to impersonate for the operation
      --certificate-authority string   Path to a cert file for the certificate authority
      --client-certificate string      Path to a client certificate file for TLS
      --client-key string              Path to a client key file for TLS
      --cluster string                 The name of the kubeconfig cluster to use
      --context string                 The name of the kubeconfig context to use
      --disable-compression            If true, opt-out of response compression for all requests to the server
      --gloglevel int                  Set the glog logging level
      --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kubeconfig string              Path to a kube config. Only required if out-of-cluster
      --log-format string              The formatter to use for logs. One of: text|json (default "text")
      --loglevel string                Set the logging level. One of: debug|info|warn|error (default "info")
  -n, --namespace string               If present, the namespace scope for this CLI request
      --password string                Password for basic authentication to the API server
      --proxy-url string               If provided, this URL will be used to connect via proxy
      --request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
      --server string                  The address and port of the Kubernetes API server
      --tls-server-name string         If provided, this name will be used to validate server certificate. If this is not provided, hostname used to contact the server is used.
      --token string                   Bearer token for authentication to the API server
      --user string                    The name of the kubeconfig user to use
      --username string                Username for basic authentication to the API server
artifact build failed to load: failed to checkout "0ed01a498b9be55d63b769b077003a4e60b5d6f0": worktree contains unstaged changes
```

### Version(s)

v3.6.2

```YAML
See description
```

```text
```

```text
```

DavidGamba:
A git artifact is a new clean checkout so forcing the checkout to avoid the error shouldn't affect any existing workflows people have.

Fixes #14141 

### Motivation

Want to be able to checkout a particular commit in my workflows rather than the HEAD of the branch.

### Modifications

Enable force checkout.

### Verification

Issue #14141 describes a separate test script that I used to reproduce my problem and to validate that this PR will fix it.

https://github.com/argoproj/argo-workflows/issues/14141#issuecomment-2631619270

### DIFF
--- a/workflow/artifacts/git/git.go
+++ b/workflow/artifacts/git/git.go
@@ -164,7 +164,7 @@ func (g *ArtifactDriver) Load(inputArtifact *wfv1.Artifact, path string) error {
 		if err != nil {
 			return fmt.Errorf("failed to get resolve revision: %w", err)
 		}
-		if err := w.Checkout(&git.CheckoutOptions{Hash: plumbing.NewHash(h.String())}); err != nil {
+		if err := w.Checkout(&git.CheckoutOptions{Hash: plumbing.NewHash(h.String()), Force: true}); err != nil {
 			return fmt.Errorf("failed to checkout %q: %w", h, err)
 		}
 	}


MasonM (commented):
Thanks for your contribution! I left a comment at https://github.com/argoproj/argo-workflows/issues/14141#issuecomment-2646018553, but I think we should have a way of consistently reproducing this first. Defaulting to force checkout seems dangerous to me because it introduces the possibility of data loss. If you could add a test case here, that'd be great: https://github.com/argoproj/argo-workflows/blob/abdd9dab71d66f835322a1fdf8fc17d132b5d50a/test/e2e/artifacts_test.go#L772

Joibel:
Given that this is a change in behaviour that others may rely on (a workflow failing when their worktree is dirty) I would want this to be configurable, defaulting to the old behaviour.

DavidGamba:
When I wrote this PR I wasn't thinking of the interaction this could have with volumes since I was just using it for workflows that didn't mount a volume and where always fresh.

I expanded my use cases to leverage volumes and I now see how the force flag would be dangerous whereas before I always though it would be a fresh environment.

I think this is a bug in the upstream git library rather than necessarily and issue with Argo Workflows itself.
I have moved on to run git clone from my container scripts and fully control that behaviour rather than relying on the artifact due to this issue so I won't be pursuing this solution anymore.

Thanks for the attention given to the review.

PR #14151, opened 2025-02-05
feat: enable cherry-pick bot

Joibel:
Part of #12592

### Motivation

Maintaining released versions of argo-workflows is not automated. This will help by allowing the backporting of commits as they are merged rather than by the release manager at the time of release.

Each PR can be CI tested independently then, and only those that pass get merged.

### Modifications

Enabled CI on release branch PRs. 

Added a new /cherry-pick action, using the [same cherry-pick-bot](https://github.com/googleapis/repo-automation-bots/tree/main/packages/cherry-pick-bot) as argo-cd already use. I have already enabled that bot on this project, but it needs a configuration file as provided here.

### Verification

This mostly needs testing after merging.

### Documentation

Added documentation for how to use `/cherry-pick`, and for the other comment based command which already existed: `/retest`

PR #14164, opened 2025-02-10
fix: remove `/test` command comment

Joibel:
Partial revert of #14151

Fixes [/test command not showing in github checks](https://github.com/argoproj/argo-workflows/pull/14151#issuecomment-2643256922) and [extra CI jobs running](https://github.com/argoproj/argo-workflows/pull/14151#issuecomment-2644729397)

### Motivation

The `/test` command isn't as important as the rest of #14151, so just remove it for now as it doesn't work as well as hoped and has issues

### Modifications

Remove `issue_comment` from the actions and associated `if`

### Verification

Untested as github action code, and changes are a revert.

### Documentation

Reverted documentation for releasing process to use the old manual push, and removed docs on `/test`

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -9,8 +9,6 @@ on:
     branches:
       - "main"
       - "release-*"
-  issue_comment:
-    types: [created]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


MasonM (approved):

isubasinghe (approved):
This is unfortunate but fair enough.

Issue #14165, opened 2025-02-10
v3.6.3 regression. Can't reliably call workflowTemplates using workflowTemplateRef


tico24:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

1. Upgrade to 3.6.3
2. Have a simple workflowTemplate that has some parameters
3. Call that using `templateRef` and `withItems`
4. Error: `failed to resolve {{item.path}}`

![Image](https://github.com/user-attachments/assets/a220d6b5-c901-4374-af5b-6f01a442a6fa)

### Version(s)

v3.6.3

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: 363-test-tmp
  namespace: ci
spec:
  templates:
    - name: 363-test-tmp
      nodeSelector:
        nodegroup: arm-spot
      inputs:
        parameters:
          - name: path
          - name: service
          - name: build_arg
          - name: run_on
          - name: arch
          - name: docker_org
      container:
        image: alpine
        command:
          - sh
          - -c
          - |
            echo "path: {{inputs.parameters.path}}"
            echo "service: {{inputs.parameters.service}}"
            echo "build_arg: {{inputs.parameters.build_arg}}"
            echo "run_on: {{inputs.parameters.run_on}}"
            echo "arch: {{inputs.parameters.arch}}"
            echo "docker_org: {{inputs.parameters.docker_org}}"
---
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: 363-test-
  namespace: ci
spec:
  activeDeadlineSeconds: 10800
  entrypoint: main
  serviceAccountName: ci

  templates:
    - name: main
      dag:
        tasks:
          - name: withitems
            templateRef:
              name: 363-test-tmp
              template: 363-test-tmp
            arguments:
              parameters:
                - name: path
[...]

PR #14053, opened 2025-01-06
fix(controller): validation failed when dynamic templateRef is used in nested template

chengjoey:
Fixes #14042 #13773

### Motivation
if nested template has reference, input tmpl args 

### Verification
unit test

PR #14166, opened 2025-02-10
Revert "fix(controller): validation failed when dynamic templateRef is used in nested template"

Joibel:
Reverts argoproj/argo-workflows#14053

This PR causes #14165


Joibel:
The revert button in Github doesn't meet DCO, so I'll do this manually.

PR #14179, opened 2025-02-11
chore: build argoexec on windows 2022 & 2025

mweibel:
### Motivation

New major windows version is out and we'll be testing the version soon.

### Modifications

Adjusted release build to build on windows 2022 and windows 2025 while also adjusting the generated manifest.

### Verification

Couldn't test it since it's a GitHub runner adjustment.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -95,7 +95,10 @@ jobs:
   build-windows:
     name: Build & push windows
     if: github.repository == 'argoproj/argo-workflows'
-    runs-on: windows-2022
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


tico24:
It would be more common to adjust on your fork to test so that you can demonstrate it works.

mweibel:
@tico24 I adjusted the workflow so it can run - the build was successful but due to missing secrets I of course couldn't push the image. 

https://github.com/helio/argo-workflows/actions/runs/13303462419/job/37149113756

does that suffice?
Otherwise I could probably rename the branch to start with `dev-*` and it would run the workflow here, too. Would you prefer that?

mweibel:
oh, I'm sorry. I just saw something upon doing a deeper inspection. Need to verify it actually works.

PR #14181, opened 2025-02-11
fix: manual retries exit handler cleanup. Fixes #14180

### LINKED ISSUE #14180: Manual retry does not work with exit handlers.
### Pre-requisites

Manual retries are incorrectly typed and result in a panic of the argo server.

This can be reproduced by calling retry on any node of an exit-handler. 

### Version(s)

v3.6.3

```YAML
# An exit handler is a template reference that executes at the end of the workflow
# irrespective of the success, failure, or error of the primary workflow. To specify
# an exit handler, reference the name of a template in 'spec.onExit'.
# Some common use cases of exit handlers are:
# - sending notifications of workflow status (e.g. e-mail/slack)
# - posting the pass/fail status to a webhook result (e.g. github build result)
# - cleaning up workflow artifacts
# - resubmitting or submitting another workflow
kind: Workflow
metadata:
  generateName: exit-handlers-
spec:
  entrypoint: intentional-fail
  onExit: exit-handler
  templates:
  # primary workflow template
  - name: intentional-fail
    container:
      image: alpine:latest
      command: [sh, -c]

  # exit handler related templates
  # After the completion of the entrypoint template, the status of the
  # workflow is made available in the global variable {{workflow.status}}.
  # {{workflow.status}} will be one of: Succeeded, Failed, Error
  - name: exit-handler
    steps:
    - - name: notify
        template: send-email
      - name: celebrate
        template: celebrate
        when: "{{workflow.status}} == Succeeded"
      - name: cry
        template: cry
  - name: send-email
    container:
      image: alpine:latest
      command: [sh, -c]
      # Tip: {{workflow.failures}} is a JSON list. If you're using bash to read it, we recommend using jq to manipulate
      # it. For example:
      #
      # echo "{{workflow.failures}}" | jq -r '.[] | "Failed Step: \(.displayName)\tMessage: \(.message)"'
      #
      # Will print a list of all the failed steps and their messages. For more info look up the jq docs.
      # Note: jq is not installed by default on the "alpine:latest" image, however it can be installed with "apk add jq"
      args: ["echo send e-mail: {{workflow.name}} {{workflow.status}} {{workflow.duration}}. Failed steps {{workflow.failures}}"]
  - name: celebrate
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo hooray!"]
  - name: cry
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo boohoo!"]
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #14180

### Motivation
N/A

### Modifications

This change now correctly types the items entered into the list. 

### Verification

Verified with a regression test  and additionally manually tested. 
Manual resets on an exit node resets the entire.

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -1322,14 +1322,14 @@ func FormulateRetryWorkflow(ctx context.Context, wf *wfv1.Workflow, restartSucce
 		}
 		if n.Name == onExitNodeName {
 			queue := list.New()
-			queue.PushBack(n)
+			queue.PushBack(&n)
 			for {
 				currNode := queue.Front()
 				if currNode == nil {
 					break
 				}
-				curr := currNode.Value.(wfv1.NodeStatus)
-				deletedPods, podsToDelete = deletePodNodeDuringRetryWorkflow(wf, curr, deletedPods, podsToDelete)
+				curr := currNode.Value.(*wfv1.NodeStatus)
+				deletedPods, podsToDelete = deletePodNodeDuringRetryWorkflow(wf, *curr, deletedPods, podsToDelete)
 				for i := range curr.Children {
 					child, err := wf.Status.Nodes.Get(curr.Children[i])
 					if err != nil {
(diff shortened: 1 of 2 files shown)


Joibel (approved):
Makes sense.

Would have preferred a better PR description here.

PR #14183, opened 2025-02-11
fix(ci): python sdk release process (release-3.6)

### LINKED ISSUE #14159: SDK releasing is failing
### Pre-requisites

https://github.com/argoproj/argo-workflows/actions/workflows/sdks.yaml the 3.5 and 3.6 releases both failed SDK builds

### Version(s)

release process

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

Joibel:
Fixes #14159

### Motivation

SDK releases are failing since 3.6.3 and 3.5.14

### Modifications

See https://github.com/pypa/twine/issues/1216 - the suggestions there are either to upgrade packaging or fiddle with setuptools.

Python SDK is deprecated so this code should end up being removed soon.

### Verification

I cannot reproduce the problem locally in devcontainer, so this is a speculative fix.

### Documentation

None needed.

### DIFF
--- a/sdks/python/Makefile
+++ b/sdks/python/Makefile
@@ -11,7 +11,7 @@ DOCKER = docker run --rm --user $(shell id -u):$(shell id -g) -v $(WD):/wd --wor
 CHOWN = chown -R $(shell id -u):$(shell id -g)
 
 publish: generate
-	pip install setuptools twine build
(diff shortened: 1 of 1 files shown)


MasonM (approved):
Don't know if this will work, but it's worth a shot

PR #14184, opened 2025-02-11
fix: correct semaphore configmap keys for multiple semaphores

Joibel:
No known bug reports around this so far

### Motivation

Whilst looking at deprecating single semaphores completely I discovered `getSemaphoreConfigMapRef()` was not multiple semaphore aware.

### Modifications

Rename `getSemaphoreConfigMapRef` to `getSemaphoreConfigMapRefs` and return a list of pointers instead of a pointer.

Make the caller of this iterate the list.

### Verification

Updated the test case that tests `GetSemaphoreKeys` which is the only user of `getSemaphoreConfigMapRefs` to test this and added an extra entry to be sure it can successfully manage a list.
 
### Documentation

Not relevant, internal code

### DIFF
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -501,7 +501,7 @@ func (wf *Workflow) GetSemaphoreKeys() []string {
 	if wf.Spec.WorkflowTemplateRef == nil {
 		templates = wf.Spec.Templates
 		if wf.Spec.Synchronization != nil {
-			if configMapRef := wf.Spec.Synchronization.getSemaphoreConfigMapRef(); configMapRef != nil {
+			for _, configMapRef := range wf.Spec.Synchronization.getSemaphoreConfigMapRefs() {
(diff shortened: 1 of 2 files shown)


rohankmr414 (approved):
LGTM

rohankmr414:
@Joibel I think we should also update `TestWorkflowSemaphoreKeysIndexFunc` here: https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/indexes/workflow_index_test.go#L73, the assertions only check for 1 entry which is `Synchronization.Semaphore` and not `Synchronization.Semaphores`.

rohankmr414:
Also, I'm curious if multiple keys are expected here: https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/controller.go#L271

Joibel:
> Also, I'm curious if multiple keys are expected here: https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/controller.go#L271

This index is already multi-key aware, and this fix makes sure all the semaphores get into the index.

isubasinghe (approved):

PR #14189, opened 2025-02-13
chore (deprecation): remove 3.6 deprecated features for 3.7

Joibel:
### Motivation

As part of 3.6 we deprecated 4 features:
* `podPriority` doesn't work any modern kubernetes and has been deprecated for a while
* `schedule` is replaced by `schedules` in cron workflows - this is a trivial update to make to the yaml
* `mutex` and `semaphore` are replaced by the pluralised version - this is a trivial update to make to the yaml

Removing legacy code reduces maintainer burden and the chance for errors.

### Modifications

Removed the features mentioned.
Removed podPriority tests.
Updated a message about use of schedule and schedules at the same time, and fixed the test. Code needed moving slightly to make a test case pass as the validator only returns one error.
Otherwise Updated all tests which used these features to use the new version.

### Verification

Relying on updated tests.

### Documentation

Documentation for this is mostly auto-generated, but has been updated where necessary.

# Draft

This will remain in draft until #14184 on which it depends is merged.


MasonM:
Mentioned in the meeting, but the UI changes for the `schedule` deprecation overlaps with https://github.com/argoproj/argo-workflows/pull/14197, which handles transparently upgrading `schedule` to `schedules`: https://github.com/argoproj/argo-workflows/blob/e9e7c4398c00621f697acf0002b7992028958daa/ui/src/shared/services/cron-workflow-service.ts#L5-L17

That's now been merged, hence the conflicts.

isubasinghe (approved):

PR #14196, opened 2025-02-15
ci: drop support for k8s 1.29 and centralize config

MasonM:
### Motivation

This drops support for Kubernetes 1.29, which will be end-of-life in 2 weeks (https://endoflife.date/kubernetes). This will allow us to take advantage of new features in 1.30, like `ValidatingAdmissionPolicies` (https://github.com/argoproj/argo-workflows/pull/14045). See https://github.com/argoproj/argo-workflows/pull/14142#issuecomment-2629570868 for previous discussions.

Also, I centralized the config for defining the tested Kubernetes versions so that we don't duplicate that in both `ci-build.yaml` and `pre-build.sh`.

### Modifications
Copied the config changes from https://github.com/argoproj/argo-workflows/pull/14176 and bumped the `min` version in `hack/k8s-versions.sh`.

An alternative to `hack/k8s-version.sh` would be to move all the logic for installing `k3d`/`k3s`/`kubectl` into the `Makefile` and then defining versions there, which would arguably be cleaner, but that'd require a lot more changes.

### Verification

Rebuilt DevContainer and verified everything came up successfully:
```
$ kubectl version
Client Version: v1.30.9
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.9+k3s1
```

### Documentation

Docs changes will be handled in https://github.com/argoproj/argo-workflows/pull/14176


Joibel (approved):
This needs a fixup, but looks good to me.

Joibel (approved):

PR #14205, opened 2025-02-18
Cache semaphore limit lookup

drabinowitz:
Fixes #TODO

### Motivation

This PR came out of the slack discussion [here](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1739325430319219):

Each `semaphore` must acquire a universal lock on the `SyncManager` and reload the corresponding `ConfigMap` from the kube api on every operation call. This can result in a performance degradation when a large number of workflows with semaphores are queued as all workflows using synchronization will have to wait in this same queue for each get call.

### Modifications

Added a `--semaphore-size-cache-ttl` argument. When specified, `SyncManager` will wait the provided number of seconds before refetching the semaphore size from the `ConfigMap`.

### Verification

Added associated unit tests to verify the caching behavior.

### Documentation

Added a line in the scaling docs to explain the flag and its purpose.

### DIFF
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -59,7 +59,7 @@ func NewRootCommand() *cobra.Command {
 		podCleanupWorkers       int    // --pod-cleanup-workers
 		cronWorkflowWorkers     int    // --cron-workflow-workers
 		workflowArchiveWorkers  int    // --workflow-archive-workers
-		semaphoreCacheTTL       int    // --semaphore-cache-ttl
+		SemaphoreSizeCacheTTL   int    // --semaphore-size-cache-ttl
 		burst                   int
 		qps                     float32
 		namespaced              bool   // --namespaced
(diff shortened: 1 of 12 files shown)


drabinowitz:
One thing this PR does not have is an integrated stress test i couldn't find a good way to add that with the existing e2e test setup but maybe i was missing something. Let me know if there's a way I can add stress tests with some assumed configmap get request latency to confirm that this change improves the system.

cmd/workflow-controller/main.go
```diff
@@ -59,6 +59,7 @@ func NewRootCommand() *cobra.Command {
 		podCleanupWorkers       int    // --pod-cleanup-workers
 		cronWorkflowWorkers     int    // --cron-workflow-workers
 		workflowArchiveWorkers  int    // --workflow-archive-workers
+		semaphoreLimitCacheTTL  int    // --semaphore-limit-cache-ttl
```
Joibel, line 62: I think this should be in the workflow-controller-configmap instead of as a command line parameter. I'd like this to be changed please.

workflow/sync/common.go
```diff
@@ -13,5 +13,10 @@ type semaphore interface {
 	getCurrentPending() []string
 	getName() string
 	getLimit() int
+	getLimitTimestamp() time.Time
+	resetLimitTimestamp()
 	resize(n int) bool
 }
+
+// expose for overriding in tests
+var now = time.Now
```
Joibel, line 22: This is clever, but I need to think about whether it's OK.

At the very least I'd prefer the variable had a longer name than just `now` so as to be more obvious as to what it does and to avoid anyone doing other bad things with it.

Joibel (changes requested):

workflow/sync/common.go
drabinowitz: yeah i didn't love the approach either i just couldn't come up with another way to mock time and didn't want to add a dependency

drabinowitz: made other fixes left this one unchanged for now since there's still a decision to be made if we even want to go this route

Joibel: I don't have a better way of doing this, but would still prefer the word `now` to be something more explicit. Perhaps just `nowFn` or `nowFunc` to illustrate it's an odd thing. Happy for other suggestions though.

drabinowitz: updated to `nowFn` i'm good with that approach

Joibel:
Thanks for working through this!

PR #14207, opened 2025-02-19
docs: Correct sub-optimal burst value for resourceRateLimit

tico24:
### Motivation

In both the example configmap and the docs, the `resourceRateLimit` `burst` was set to a value lower than the `limit`.  Even worse, it was set to `1`. This means that the `limit` would be perpetually in place even when not needed.

### Modifications

- Update the docs to explain burst a little clearer. Update the example
- Update the example controller configmap to match the docs.
- Added a missing line break

### Verification

`make docs`


terrytangyuan (approved):

terrytangyuan:
Thanks!

PR #14209, opened 2025-02-20
chore(deps)!: bump k8s dependencies to 1.32

blkperl:
### Motivation

adds support for K8s 1.32, potentially drops support for K8s 1.29 which is EOL

### Modifications

* updated go.mod and makefile with new k8s and controller-runtime version

### Verification

e2e will verify

### Documentation

Docs are not needed


MasonM (approved):

Makefile
```diff
@@ -664,7 +664,7 @@ dist/kubernetes.swagger.json: Makefile
 	@mkdir -p dist
 	# recurl will only fetch if the file doesn't exist, so delete it
 	rm -f $@
-	./hack/recurl.sh $@ https://raw.githubusercontent.com/kubernetes/kubernetes/v1.31.3/api/openapi-spec/swagger.json
+	./hack/recurl.sh $@ https://raw.githubusercontent.com/kubernetes/kubernetes/v1.32.2/api/openapi-spec/swagger.json
```
Joibel, line 667: Would it makes sense if this be derived from https://github.com/argoproj/argo-workflows/blob/main/hack/k8s-versions.sh?

Not necessary for this PR, but something to consider.

Joibel (approved):

Makefile
blkperl: I think not because thats for the CI version and we will probably update the CI and k8s libs separately

PR #14104, opened 2025-01-20
feat: label actor action when making change to workflow/template. Fixes #14102

tczhao:
Fixes #14102 

### Motivation

when sso is enabled, server will add actor labels so that we know who was making the changes to resource
instead of relying on matching timeline with the audit log

### Modifications

added labels
```
workflows.argoproj.io/actor:
workflows.argoproj.io/actor-email:
workflows.argoproj.io/actor-preferred-username:
workflows.argoproj.io/action:
```

relevant label will be added when user perform the following action
- workflow
  - terminate
  - stop
  - suspend
  - resume
- workflowTemplate
  - update
- clusterWorkflowTemplate
  - update
- cronWorkflow
  - update

The code follows the same pattern as the existing creator code but instead, we are recording/labelling other workflow actions.

### Verification

Added unit tests

PR #14216, opened 2025-02-23
docs: label user action

tczhao:
Fixes #14104

### DIFF
--- a/docs/workflow-actor.md
+++ b/docs/workflow-actor.md
@@ -2,7 +2,7 @@
 
 > v3.6.5 and after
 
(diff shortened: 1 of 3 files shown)


docs/workflow-actor.md
```diff
@@ -0,0 +1,28 @@
+# Workflow Actor
+
+> v3.6.5 and after
```
tczhao, line 3: Let's align on this,
I feel like it's a feature, should be on 3.7 but it's small enough and no harm to include in the next patch

MasonM, line 3: I think you'll need to add this here to make the spellchecker happy: https://github.com/argoproj/argo-workflows/blob/f1722947cb177b711c62eaa6cd292c812818480e/.spelling#L259

tczhao: Thanks

Joibel:
@tczhao As we missed getting this in for 3.6, fancy updating it and we'll have it marked as for 3.7? Or if you don't have time I'm happy to fix it.

PR #14217, opened 2025-02-23
feat: support fetch items from configmap in withItems. Part of #5978

### LINKED ISSUE #5978: Support for configmaps in `withItems` and `withParam` of loops
# Summary

What change needs making?

Add a way to reference values from configmaps in `withItems` / `withParam` in loops. Something like

```yaml
kind: Workflow
metadata:
  generateName: loops-
spec:
  entrypoint: loop-example
  templates:
  - name: loop-example
    steps:
    - - name: print-message
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "{{item}}"
        withItems:              # invoke whalesay once for each item in parallel
          configMapKeyRef:
            name: my-config
            key: tables

  - name: whalesay
    inputs:
      parameters:
      - name: message
    container:
      image: docker/whalesay:latest
      command: [cowsay]
```

```
apiVersion: v1
kind: ConfigMap
metadata:
 name: my-config
data:
  tables:
  - table1
  - table2
  - table3
```

# Use Cases

When would you use this?

My usecase is to reuse an existing configmap used by other deployments and I can easily change the configuration across my workflows and deployments.

---

shuangkun:
support fetch items from configmap in withItems. 

Fixes #5978

### Verification

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
data:
  tables: |
    [
      "table1",
      "table2",
      "ubuntu 16.10"
    ]
```
```
kind: Workflow
metadata:
  name: expand-with-items-from
  namespace: default
spec:
  entrypoint: loop-example
  templates:
  - name: loop-example
    steps:
    - - name: print-message
        template: whalesay
        arguments:
          parameters:
          - name: message
            value: "{{item}}"
        withItemsFrom: # invoke whalesay once for each item in parallel
          configMapKeyRef:
            name: my-config
            key: tables

  - name: whalesay
    inputs:
      parameters:
      - name: message
    container:
      image: alpine:latest
      command: [sh, -c]
      args: ["echo {{inputs.parameters.message}}"]
```


shuangkun:
> Could you add some documentation to https://argo-workflows.readthedocs.io/en/latest/walk-through/loops/ please?
> 
> Could you also add an example to /examples?
> 
Thanks， I will add it in these fields.

> Is there a reason you chose to do this to `withItems` rather than the more powerful and useful `withParam` style. In general `withParam` can do anything `withItems` can do, so if we're only going to support one `withParam` would be the preference.

My proposal is to add ConfigMap support for both approaches. Since we adopted withItems early in our Argo Workflows journey (likely influenced by initial examples/documentation), but now need a centralized configuration mechanism.
We periodically synchronize resources (e.g., container images, Helm charts, automation scripts) to multiple regions. The goal is to maintain a single source of truth (e.g., a ConfigMap) listing all target regions, which can be dynamically referenced by various WorkflowTemplates. This would allow region additions/removals to propagate automatically across all workflows without modifying individual YAML definitions.

PR #14218, opened 2025-02-23
test: fix intermittent cron failures due to parallelism limit

MasonM:
### Motivation

Occasionally, the cron E2E suite fails and you'll see the following in the output ([example](
https://github.com/argoproj/argo-workflows/actions/runs/13480065787/job/37664515754?pr=13747)):
>  ◷ test-cron-wf-fail-1-1740291060 Workflow 0s      Workflow processing has been postponed because too many workflows are already running

I think this is happening because the `E2ESuite.DeleteResources()` method that's supposed to clean up `CronWorkflow`s from previous tests isn't being run for subtests, and nearly all the tests in `test/e2e/cron_test.go` are subtests. 

### Modifications
This adds a `TearDownSubTest` method to cleanup all resources after every subtest. Docs:
https://pkg.go.dev/github.com/stretchr/testify/suite#TearDownSubTest

I also deleted the redundant `BeforeTest()` and `TearDownSuite()` methods.

### Verification
Ran `make test-cron` locally

### DIFF
--- a/test/e2e/cron_test.go
+++ b/test/e2e/cron_test.go
@@ -21,19 +21,16 @@ type CronSuite struct {
 	fixtures.E2ESuite
 }
 
-func (s *CronSuite) SetupSuite() {
-	s.E2ESuite.SetupSuite()
-	// Since tests run in parallel, delete all cron resources before the test suite is run
+func (s *CronSuite) TearDownSubTest() {
+	// Delete all CronWorkflows after every subtest (as opposed to after each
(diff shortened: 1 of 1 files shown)


blkperl (approved):

Joibel:
The cron tests are supposed to run in parallel, but haven't done for years. Hopefully one day we can go back to doing that - but testify.suite gets in the way. This removes some more use of suite though so might be overall beneficial to that.

Joibel (approved):

PR #14233, opened 2025-02-27
fix: incorrect description of value field in Workflow.Parameter

luispflamminger:
### Motivation

The current description of the `value` field on `Workflow.Parameter` is confusing and suggests that parameters passed to the workflow will not be respected if a value has been set in the workflow definition.
The actual behavior is opposite: The specified value acts as a default and is overridden by any passed values.

### Modifications

- Updated the description to match the actual behavior
- Added *Deutsche Telekom AG* to the list of users

### Verification

Ran `make pre-commit -B`.


tczhao (approved):
Thanks for the fix.
The original description was indeed not correct

PR #14242, opened 2025-03-03
chore: remove `onExit` in example

elliotgunton:
### Motivation

One liner for a line missed in https://github.com/argoproj/argo-workflows/pull/14127


Joibel (approved):
Good spot, thanks

Issue #13168, opened 2024-06-12
UI: Share URLs with search criteria set


panicboat:
# Summary

What change needs making?

Search criteria can be set in CronWorkflow and WorkflowTemplate lists.
However, URLs with search criteria cannot be shared with others.
Since we use Argo Workflows as a team, there are many situations where we want to share URLs.
Specifically, we would like to add the search criteria set in the Web UI to the URL query parameters as well.

## Use Cases

We have multiple teams using Argo Workflows, so we have a large number of workflows deployed.
Therefore, we set owner information for each Workflow and often search by owner information (we also search by microservice name, project name, etc.).
Therefore, when communicating in some way, the Workflow list is used to share with multiple people.

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
You can do this for most of the filters in the Workflows List.
It might not be supported in the CronWorkflow and WorkflowTemplate Lists though -- but you can filter by those in the Workflow List.

panicboat:
Thank you for your prompt confirmation.
The Workflow list does not seem to be suitable for our purposes.
We sometimes communicate primarily about what CronWorkflow / WorkflowTemplate is being deployed.
Therefore, we would like to share the list of CronWorkflows and WorkflowTemplates.

MasonM:
This should probably share some of the code from https://github.com/argoproj/argo-workflows/pull/13922

pradeept:
Hey, can I work on this ?

Aditya200247:
Hi team, 
I would like to work on this issue. I've been looking into the UI codebase and I believe I can implement this to improve the shareability of workflow views.

My action plan is :- 
1) Locate state management - I will identify the state variables controlling the filters in the Workflow List component
2) Sync to URL - I will update the filter change handlers to modify the browser's URL query parameters by using replace or history.push whenever a user applies a search criterion.
3) Hydrate from URL I will add an initialization check to parse the URL query parameters on page load and pre-fill the filter state.
4) Verification: I will ensure that copying the URL and opening it in a new window correctly preserves the active filters.

If this approach can fix the issue or aligns with the project, could you please assign this issue to me

kunalworldwide:
[...]

PR #14245, opened 2025-03-04
feat: Specify conditions in query parameters

panicboat:
Fixes https://github.com/argoproj/argo-workflows/issues/13168

### Motivation

Currently query parameters cannot be specified on the CronWorkflow or WorkflowTemplate list screens.
The experience would be improved if it could be displayed with the search criteria entered so that the browser can be bookmarked.


panicboat:
Closing due to duplication with https://github.com/argoproj/argo-workflows/pull/14816

PR #14248, opened 2025-03-04
fix: wait for workflow informer to sync before pod informer

rohankmr414:
### Motivation

There's a bug in this PR: https://github.com/argoproj/argo-workflows/pull/14129

The bug is that the podInformer is running and active (https://github.com/argoproj/argo-workflows/blob/b005487837dd065960932bba32bf75e85222e548/workflow/controller/pod/controller.go#L74) before the workflowInformer has completed cache synchronisation.
This means a pod can be orphaned (https://github.com/argoproj/argo-workflows/blob/b005487837dd065960932bba32bf75e85222e548/workflow/controller/pod/controller.go#L203) when it isn't really, because the workflowInformer hasn't yet seen the workflow that the orphan pod check should check.

We must delay starting the informer until after the workflow informer has started

### Modifications

-  Changed the order to check for cache sync status of wfInformer before podInformer while running the pod controller


workflow/controller/pod/controller.go
```diff
@@ -113,7 +113,7 @@ func (c *Controller) HasSynced() func() bool {
 func (c *Controller) Run(ctx context.Context, workers int) {
 	defer c.workqueue.ShutDown()
 	go c.podInformer.Run(ctx.Done())
```
Joibel, line 115: The podInformer.Run() call will start the pod informer, which can still make incorrect decisions based on the unsynced wfInformer. I think this Run() needs to move after the waitforcachesync on wfInfomer.

Joibel (changes requested):

workflow/controller/pod/controller.go
rohankmr414: If we move `podInformer.Run()` below the `waitforcachesync`, will `podInformer.HasSynced` return `true`?

Joibel: I think we'll have to call WaitForCacheSync twice, once for `wfInformer`, then `podInformer.Run`, then wait on the podInformer in a second call.

rohankmr414: resolved in 6b0aaf39d465037da697ab9a1cf35a8067a28fba

Joibel (approved):

workflow/controller/pod/controller.go
Joibel: Thanks

Issue #14120, opened 2025-01-23
GitHub CI Action test-executor E2E Test Often Fails - TestStopBehavior


wesleyscholl:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

Over the last week when updating my PR #13895, the GitHub CI Action often fails on the `test-executor` E2E test - TestSignalsSuite/TestStopBehavior.

```shell
=== FAIL: SignalsSuite/TestStopBehavior
FAIL	github.com/argoproj/argo-workflows/v3/test/e2e	623.621s
FAIL
make: *** [Makefile:609: test-executor] Error 1
Error: Process completed with exit code 2.
```

Failed CI runs for my PR:

- https://github.com/argoproj/argo-workflows/actions/runs/12936289255/job/36081693785?pr=13895
- https://github.com/argoproj/argo-workflows/actions/runs/12934602535
- https://github.com/argoproj/argo-workflows/actions/runs/12933983700
- https://github.com/argoproj/argo-workflows/actions/runs/12914162834
- https://github.com/argoproj/argo-workflows/actions/runs/12910303232
- https://github.com/argoproj/argo-workflows/actions/runs/12909711015
- https://github.com/argoproj/argo-workflows/actions/runs/12890135207
- https://github.com/argoproj/argo-workflows/actions/runs/12891494408

Other community members affected:

- https://github.com/argoproj/argo-workflows/actions/runs/12925025600
- https://github.com/argoproj/argo-workflows/actions/runs/12925812461
- https://github.com/argoproj/argo-workflows/actions/runs/12885673508
- https://github.com/argoproj/argo-workflows/actions/runs/12875943776

I found these recurring issues and PRs regarding this failing TestStopBehavior E2E test.

- #8532
- #11881
- #11880
- #11753
- #9027
- #8096

Let me know if you need more information and how I can help resolve this issue, thanks.

### Version(s)

GitHub Action - latest

[...]

PR #13831, opened 2024-10-29
fix: don't print help for non-validation errors. Fixes #13826

MasonM:
Fixes #13826

### Motivation
https://github.com/argoproj/argo-workflows/pull/13656 changed all CLI commands to use `RunE` instead of `Run` and to use [Cobra validators](https://cobra.dev/#positional-and-custom-arguments) for validating arguments. The default behavior with Cobra is to print the help text for all errors, which can be tedious to scroll through.
### Modifications

This changes the CLI to only print the help text for argument validation errors, which should match the previous behavior.

### Verification
```
$ ./dist/argo submit 'ANYexampleworkflow.yaml'
Error: open ANYexampleworkflow.yaml: no such file or directory

 $ ./dist/argo submit --from workflow/basic 'ANYexampleworkflow.yaml'
Error: cannot combine --from with file arguments
Usage:
  argo submit [FILE... | --from `kind/name] [flags]
<SNIP>
```

PR #14249, opened 2025-03-04
fix: don't print help for non-validation errors. Fixes argoproj#14234

### LINKED ISSUE #14234: 3.6.2 CLI: Do not print help output on submission error
### Pre-requisites

In previous versions of cmd/argoexec (< 3.6), when a workflow submission failed, only the error message was output.

Currently, I am using v3.6.2, and in case of a failure, the entire help message is printed along with the error message. This results in a lot of unnecessary log output, which adds noise and makes it inconvenient to check the actual error.

A similar issue has already been reported in the following issue.
However, that issue only addressed the cmd/argo side. It seems that cmd/argoexec also needs to be addressed in the same way.

### Version(s)

v3.6.2

```YAML
argo submit 'ANYexampleworkflow.yaml'
```

```text
Not relevant to this issue.
```

```text
Not relevant to this issue.
```

jumpe1:
Fixes #14234

### Motivation

The default behavior with Cobra is to print the help text for all errors, which can be tedious to scroll through.

### Modifications

This changes the CLI to only print the help text for argument validation errors.
Here are the related PRs.
https://github.com/argoproj/argo-workflows/pull/13831

### Verification
I reproduced the issue by applying the following manifest and intentionally causing a failure when accessing `/apis/argoproj.io/v1alpha1/namespaces/{namespace}/workflowtaskresults`.
I confirmed that the help log does not appear.

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: deny-to-api-server
spec:
  podSelector: {}
  policyTypes:
    - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 172.22.0.1/32
      ports:
        - protocol: TCP
          port: 6443
```

#### Note

Some CI tests are failing, but this is likely caused by known flaky tests, as documented in [#14120](https://github.com/argoproj/argo-workflows/issues/14120).

I have confirmed that the relevant tests for this change pass locally.
These CI failures are likely unrelated to the changes in this PR.

### DIFF
--- a/test/e2e/expectedfailures/serviceaccount-insufficient-permissions.yaml
+++ b/test/e2e/expectedfailures/serviceaccount-insufficient-permissions.yaml
@@ -1,0 +1,13 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: serviceaccount-insufficient-permissions
+spec:
+  # This ServiceAccount is only intended for use in webhooks, and doesn't have
+  # the permissions needed by the executor.
+  serviceAccountName: github.com
+  entrypoint: main
+  templates:
+  - name: main
+    container:
+      image: argoproj/argosay:v2
--- a/test/e2e/fixtures/then.go
+++ b/test/e2e/fixtures/then.go
@@ -3,6 +3,7 @@ package fixtures
 import (
 	"context"
 	"fmt"
+	"io"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


MasonM:
Thanks for the fix! I'd like to understand the behavior you're seeing, so if you could provide some more details, that'd be very helpful: https://github.com/argoproj/argo-workflows/issues/14234#issuecomment-2699801439

MasonM:
Thanks for the details in https://github.com/argoproj/argo-workflows/issues/14234#issuecomment-2700042258! I understand this now, and your fix looks right to me.

I think it's worth having a test for this, though it's slightly tricky. The best way I found to trigger this is to submit a workflow that has a `serviceAccountName` lacking sufficient permissions. That's really a scenario we should be testing anyway, so I went ahead and wrote a test for that (and refactored an existing test): https://github.com/MasonM/argo-workflows/commit/b1a24421678875dcbe10f77f3281a9def31e0597

Here's commands to merge it into your branch:
```
git fetch git@github.com:MasonM/argo-workflows.git b1a24421678875dcbe10f77f3281a9def31e0597 
git merge b1a24421678875dcbe10f77f3281a9def31e0597   
```

though if you can think of a better way to test this, then feel free to do that instead.

jumpe1:
@MasonM 
Thanks for adding the test.
Since this is my first contribution, I took some time to think about it, and I think your testing approach is the best.
I have merged your commit. Please review it.

MasonM (approved):
Good job!

Joibel (approved):
I appreciate the code reuse change in the test framework too, thank you.

PR #14254, opened 2025-03-05
chore: use expr built-in functions instead of pkg (Fixes #14255)

### LINKED ISSUE #14255: Use new expr-lang builtins instead of pkg functions
# Summary

expr-lang has included `int`, `float`, and `string` functions since 1.14.0. Argo Workflows currently uses expr 1.16.9. Instead of using those functions as implemented in argoproj/pkg, we should use the built-ins.

We should also inline the implementation of `jsonpath`, since it is trivial and doesn't make sense to externalize to the pkg repo.

## Use Cases

Simplifies maintenance, because we don't have to worry about the pkg repo lifecycle as much.

---

crenshaw-dev:
Fixes https://github.com/argoproj/argo-workflows/issues/14255

### Motivation

expr has had built-in `int`, `float`, and `string` functions since [1.14.0](https://github.com/expr-lang/expr/releases/tag/v1.14.0) (Workflows is currently using 1.16.9). 

The `jsonpath` function is trivial and doesn't make sense to maintain in a separate library.

More info here: https://github.com/argoproj/argo-cd/issues/22175

### Verification

Compared the [expr-lang builtins](https://github.com/expr-lang/expr/blob/77e876bb7cda24c5aeedbfd1a0efa1465fb503bb/builtin/lib.go) with the [pkg functions](https://github.com/argoproj/pkg/blob/cbc37dc61de58f5431fddc2fc377b56a69e94de2/expr/function.go) to confirm they're functionally identical.


terrytangyuan (commented):
Thanks! This definitely simplifies things!

Joibel (approved):

PR #14257, opened 2025-03-06
fix(api/jsonschema): use working `$id`

rogpeppe:
As suggested in #14092, it's fairly conventional for JSON Schema ID URIs to be working URLs too. Change the schema to use a URL that works and returns the latest tip version.


Joibel:
Thanks @rogpeppe

PR #14274, opened 2025-03-07
docs: fix kubernetes versions

Joibel:
Fixes kubernetes versions in documentation - this doesn't work as hoped from #14176 

### Motivation

Docs get rebuilt on every push to a branch, so `RELEASE_TAG` isn't usable

### Modifications

Use the `GIT_BRANCH` variable and check we're on an appropriate branch

### Verification

`make docs`

### Documentation

Not needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -762,7 +762,7 @@ docs: /usr/local/bin/mkdocs \
 	# check environment-variables.md contains all variables mentioned in the code
 	./hack/docs/check-env-doc.sh
 	# build the docs
-ifeq ($(RELEASE_TAG),true)
+ifeq ($(shell echo $(GIT_BRANCH) | head -c 8),release-)
 	./hack/docs/tested-versions.sh > docs/tested-kubernetes-versions.md
 endif
 	TZ=UTC mkdocs build --strict


Makefile
```diff
@@ -762,7 +762,7 @@ docs: /usr/local/bin/mkdocs \
 	# check environment-variables.md contains all variables mentioned in the code
 	./hack/docs/check-env-doc.sh
 	# build the docs
-ifeq ($(RELEASE_TAG),true)
+ifeq ($(shell echo $(GIT_BRANCH) | head -c 8),release-)
```
MasonM, line 765: I actually started off doing this, but went with the release tag approach because you mentioned concerns in https://github.com/argoproj/argo-workflows/pull/14176#discussion_r1955763642 that an update to the k8s version may have occurred on the branch without a release. I'm assuming that's rare enough this isn't a big problem.

I did start on a branch to fully automate this using the `peter-evans/create-pull-request` action, though I'm not sure it's worth it, and determining the branch name from the release tag turned out to be a bit more complicated than I expected: 
https://github.com/argoproj/argo-workflows/compare/main...MasonM:argo-workflows:automate-docs-update

MasonM (approved):
Thanks!

PR #14275, opened 2025-03-07
fix(controller): step with items failed to resolve inputs; fixes #14267

### LINKED ISSUE #14267: Optional artifact yielding failed to resolve when not passed alongside withParams or withItems
### Pre-requisites

- [ ] I have searched existing issues and could not find a match for this bug

After upgrading argo from 3.5.11 to 3.5.13 we started to observe _failed to resolve_ errors in the workflows when artifact marked as _optional_ is not passed. Now, in 3.5.14, the problem still exists. 

The issue only arises when dag or task includes withItems or withParams parameter (similarly to [this issue](https://github.com/argoproj/argo-workflows/issues/6805)). To highlight, in 3.5.11 version such problem does not occur.

An example of failing workflow:

```yaml
kind: Workflow
metadata:
  name: fail-optional-artifact-with-items
spec:
  serviceAccountName: argo
  entrypoint: artifact-example
  templates:
    - name: artifact-example
      steps:
        - - name: generate-artifact
            template: whalesay
        - - name: consume-artifact
            template: print-message
            arguments:
              artifacts:
                - name: message
                  optional: true
                  from: "{{steps.generate-artifact.outputs.artifacts.message}}"
            withItems:
              - hello world

    - name: whalesay
      container:
        image: argoproj/argosay:v1
        command: [sh, -c]
        args: ["sleep 1;"]
      outputs:
        artifacts:
          - name: message
            optional: true
            path: /tmp/hello_world
          

    - name: print-message
      inputs:
        artifacts:
          - name: message
            optional: true
            path: /tmp/message
      container:
        image: alpine:latest
        command: [sh, -c]
        args: ["test -e /tmp/message && ls /tmp/message || echo 0"]
```

And similar workflow, without withItems in task (removed lines 20, 21), that succeeds:

```yaml
kind: Workflow
metadata:
  name: success-optional-artifact-no-with-items
spec:
  serviceAccountName: argo
  entrypoint: artifact-example
  templates:
    - name: artifact-example
      steps:
        - - name: generate-artifact
            template: whalesay
        - - name: consume-artifact
            template: print-message
            arguments:
              artifacts:
                - name: message
                  optional: true
                  from: "{{steps.generate-artifact.outputs.artifacts.message}}"

    - name: whalesay
      container:
        image: argoproj/argosay:v1
        command: [sh, -c]
        args: ["sleep 1;"]
      outputs:
        artifacts:
          - name: message
            optional: true
            path: /tmp/hello_world
          

    - name: print-message
      inputs:
        artifacts:
          - name: message
            optional: true
            path: /tmp/message
      container:
        image: alpine:latest
        command: [sh, -c]
        args: ["test -e /tmp/message && ls /tmp/message || echo 0"]
```

Find also the error message from when those workflows were run on argo 3.5.14.

Image

### Version(s)

v3.5.11, v3.5.13, v3.5.14

chengjoey:
Fixes #14267

### Motivation

with-items ignore optional artifact repo input

### DIFF
--- a/test/e2e/smoke/with-items-with-artifact-repo-ref.yaml
+++ b/test/e2e/smoke/with-items-with-artifact-repo-ref.yaml
@@ -1,0 +1,43 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: success-optional-artifact-with-items-
+spec:
+  entrypoint: artifact-example
+  templates:
+    - name: artifact-example
+      steps:
+        - - name: generate-artifact
+            template: whalesay
+        - - name: consume-artifact
+            template: print-message
+            arguments:
+              artifacts:
+                - name: message
+                  optional: true
+                  from: "{{steps.generate-artifact.outputs.artifacts.message}}"
+            withItems:
+              - hello world
+
+    - name: whalesay
+      container:
+        image: argoproj/argosay:v1
+        command: [ sh, -c ]
+        args: [ "echo 'hello world' > /tmp/hello_world;" ]
+      outputs:
+        artifacts:
+          - name: message
+            optional: true
+            path: /tmp/hello_world
+
+
+    - name: print-message
+      inputs:
+        artifacts:
+          - name: message
+            optional: true
+            path: /tmp/message
+      container:
+        image: argoproj/argosay:v1
+        command: [ sh, -c ]
+        args: [ "test -e /tmp/message && ls /tmp/message || echo 0" ]
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -1356,6 +1356,14 @@ func (s *FunctionalSuite) TestWithItemsWithHooks() {
 		WaitForWorkflow(fixtures.ToBeSucceeded)
 }
 
+func (s *FunctionalSuite) TestWithItemsWithArtifactRepositoryRef() {
+	s.Given().
+		Workflow("@smoke/with-items-with-artifact-repo-ref.yaml").
+		When().
+		SubmitWorkflow().
+		WaitForWorkflow(fixtures.ToBeSucceeded)
+}
+
 // when you terminate a workflow with onexit handler,
 // then the onexit handler should fail along with steps and stepsGroup
 func (s *FunctionalSuite) TestTerminateWorkflowWhileOnExitHandlerRunning() {
--- a/workflow/controller/dag.go
+++ b/workflow/controller/dag.go
@@ -810,6 +810,13 @@ func expandTask(task wfv1.DAGTask) ([]wfv1.DAGTask, error) {
 	task.WithItems = nil
 	task.WithParam = ""
 	task.WithSequence = nil
+	var inputArts wfv1.Artifacts
+	for _, art := range task.Arguments.Artifacts {
+		if !art.Optional {
+			inputArts = append(inputArts, art)
+		}
+	}
+	task.Arguments.Artifacts = inputArts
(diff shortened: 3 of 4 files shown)


workflow/controller/operator.go
```diff
@@ -3648,7 +3648,7 @@ func processItem(tmpl template.Template, name string, index int, item wfv1.Item,
 	if err == nil && !proceed {
 		newStepStr, err = tmpl.Replace(replaceMap, true)
 	} else {
-		newStepStr, err = tmpl.Replace(replaceMap, false)
+		newStepStr, err = tmpl.Replace(replaceMap, true)
```
Joibel, line 3651: This is now the same as 2 lines above, so the whole if clause would be unnecessary.

Joibel (changes requested):
I am not at all convinced this is the correct fix as you haven't explained anything in the PR message about why you think this is correct. Gut feel says that this is too broad a change, and that we should only allow failure for optional inputs at this step.

It definitely needs a regression test whatever.

chengjoey:
> I am not at all convinced this is the correct fix as you haven't explained anything in the PR message about why you think this is correct. Gut feel says that this is too broad a change, and that we should only allow failure for optional inputs at this step.
> 
> It definitely needs a regression test whatever.

Agree, please re-review this PR. I think we should ignore the optional artifact ref, WDYT?

test/e2e/smoke/with-items-with-artifact-repo-ref.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - name: print-message
+      inputs:
+        artifacts:
+          - name: message
+            optional: true
+            path: /tmp/message
+      container:
+        image: argoproj/argosay:v1
+        command: [ sh, -c ]
+        args: [ "test -e /tmp/message && ls /tmp/message || echo 0" ]
```
anarsen, line 43: I know we came up with this repro workflow, so this is an attempt of me fixing our past mistakes 😅 Shouldn't it be like this?

```suggestion
        args: [ "test -e /tmp/message && ls /tmp/message || echo 0" ]
```

Joibel:
> > I am not at all convinced this is the correct fix as you haven't explained anything in the PR message about why you think this is correct. Gut feel says that this is too broad a change, and that we should only allow failure for optional inputs at this step.
> > It definitely needs a regression test whatever.
> 
> Agree, please re-review this PR. I think we should ignore the optional artifact ref, WDYT?
> 
> Reference：
> 
> https://github.com/argoproj/argo-workflows/blob/f687a824348ab37d2622d9f0ebfa544c28e0df42/workflow/controller/steps.go#L473-L479

Won't this break all optional input artifacts. I haven't tested it, but it feels like we should attempt to resolve all optional inputs and just not complain when they don't work.

caelan-io:
Simply circling back on this to say that we appreciated you taking this on @chengjoey ! Per @Joibel's comments, we had to go in a different direction with the fix based on some deeper context the maintainers have in how the Argo controller works, so @isubasinghe created the fix in #14503. Please do let us know if any questions, and we're always happy to help you with contributing other fixes! 👍  

I'll close the PR since the fix is merged. More issues here we would appreciate contributions on:
- [Good First Issues label](https://github.com/argoproj/argo-workflows/issues/?q=is:issue%20state:open%20sort:comments-desc%20label:%22good%20first%20issue%22)
- [Top bugs](https://github.com/argoproj/argo-workflows/issues/?q=is:issue%20state:open%20label:type/bug%20sort:comments-desc)

PR #14280, opened 2025-03-10
fix: remove UID from securitycontext, redundant to UID in Dockerfile Fixes #14279

### LINKED ISSUE #14279: Artifact Garbage Collect UID in artgc pod should not be defined by default
### Pre-requisites

In OpenShift environment, the artgc pod is in error because by default the UID 8737 is set at https://github.com/argoproj/argo-workflows/blob/v3.6.4/workflow/common/security_context.go#L13

I expect the UID to not be set by default, so that:
* in OpenShift environment, the UID is in the authorized range
* in other environment, the UID is the one define the container image

Workaround:
It is possible to override that.
According to doc https://github.com/argoproj/argo-workflows/issues/11485 it is possible to override the pod definition of artgc, to be set in values.yaml of helm chart:
```
  workflowDefaults:
    spec:
      artifactGC:
        podSpecPatch: |+
          containers:
          - name: main
            securityContext:
              # OpenShift: must be in UID range.
              runAsUser: 123456789
```

I also tried many other things like `runAsUser: null`, `runAsUser: "null"`, `runAsUser: "null"`, `runAsUser: ""` but none of these deletes the 8737 UID.
Also tried strategic merge patch:
```
  workflowDefaults:
    spec:
      artifactGC:
        podSpecPatch: |+
          containers:
          - name: main
            securityContext:
              $patch: delete
              runAsUser: 123456789
```

This will result in UID=8737.

### Version(s)

v3.6.4

```YAML
Any helloworld that has artifact S3 as output.
```

```text
```

```text
```

antoinetran:
Fixes #14279

### Motivation

Argo should work easily in OpenShift and non-OpenShift environment. Currently, the UID 8737 is hard-coded in code, both in Dockerfile and in YAML resource to apply. This is redundant, and makes it fail in OpenShift, where a pod policy such as "UID must be in range 10000000-10010000" . This pod policy can also be dynamic, per OpenShift project, so another one might says "UID must be in range 20000000-20010000". So it is a nightmare to deploy ARGO and specifies each UID, while NOT settting the runAsUser UID in YAML resource, allows Kubernetes to use the default one in Dockerfile in non-OpenShift environment, and use the default high UID in the range in OpenShift environment.

### Modifications

Just removed the UID in runAsUser by setting it to `nil`.

### Documentation

This changes nothing in use, and override of this UID for artgc pod has not yet been documented I believe. But from the other issue I found, the override with PodSpecPatch (described in my issue #14279 workaround) works before and after this pullrequest.

### DIFF
--- a/workflow/common/security_context.go
+++ b/workflow/common/security_context.go
@@ -10,7 +10,7 @@ func MinimalCtrSC() *corev1.SecurityContext {
 		Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
 		Privileged:               ptr.To(false),
 		RunAsNonRoot:             ptr.To(true),
-		RunAsUser:                ptr.To(int64(8737)),
+		RunAsUser:                nil,
 		ReadOnlyRootFilesystem:   ptr.To(true),
 		AllowPrivilegeEscalation: ptr.To(false),
 		SeccompProfile:           &corev1.SeccompProfile{Type: "RuntimeDefault"},
@@ -20,7 +20,7 @@ func MinimalCtrSC() *corev1.SecurityContext {
 func MinimalPodSC() *corev1.PodSecurityContext {
 	return &corev1.PodSecurityContext{
 		RunAsNonRoot:   ptr.To(true),
-		RunAsUser:      ptr.To(int64(8737)),
+		RunAsUser:      nil,
 		SeccompProfile: &corev1.SeccompProfile{Type: "RuntimeDefault"},
 	}
 }


antoinetran:
FYI, this is where Dockerfile hard-code 8737 https://github.com/argoproj/argo-workflows/blob/v3.6.4/Dockerfile#L108, and this is ok, as long as this is not specified again in securityContext.

antoinetran:
Hi @Garett-MacGowan , I have an issue with one test with my change. I can see you wrote that test. I have no idea how to debug, I mean there is an issue with TestGlobalArtifactPassing and I can see exit code 2 of one container, but I cannot see the logs. Do you have a hint of how to progress?

Garett-MacGowan:
> Hi @Garett-MacGowan , I have an issue with one test with my change. I can see you wrote that test. I have no idea how to debug, I mean there is an issue with TestGlobalArtifactPassing and I can see exit code 2 of one container, but I cannot see the logs. Do you have a hint of how to progress?

Could it be that the workflow container itself no longer has the privileges it needs to write to /tmp/artifact.txt? e.g. `argoproj/argosay:v2` doesn't specify `USER 8737` and therefore won't have privilege anymore.

antoinetran:
> Could it be that the workflow container itself no longer has the privileges it needs to write to `/tmp/artifact.txt`? e.g. `argoproj/argosay:v2` doesn't specify `USER 8737` and therefore won't have privilege anymore.

`argoproj/argosay:v2` is running as root by default. In the workflow in failure (https://github.com/argoproj/argo-workflows/blob/main/test/e2e/testdata/global-artifact-passing.yaml#L64), I don't see any user override in the workflow, so unless the CI/CD runs with a default non root user, this container should be run as root. So it shouldn't has issue with writing to  `/tmp/artifact.txt`.

Normally, my modification only impacts these file who uses the `MinimalCtrSC` function:
* https://github.com/argoproj/argo-workflows/blob/release-3.6/workflow/controller/agent.go#L173
* https://github.com/argoproj/argo-workflows/blob/release-3.6/workflow/controller/artifact_gc.go#L448
* https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/workflowpod.go#L693
which all revolves around the executor image (argoexec), which is precisely why I added UID 8737 by default in its `Dockerfile`.

How can I be sure that the image used in CI/CD is the one built from this pull request? Can I access the CI/CD environment, do exec inside the container somehow?

antoinetran:
> > How can I be sure that the image used in CI/CD is the one built from this pull request? Can I access the CI/CD environment, do exec inside the container somehow?
> 
> The CICD builds the image first then imports to k3d cluster for e2e test (all from current branch)

Thanks for the doc of how to run e2e test locally. I need time to setup the test environment. Probably there won't be movement here until some time, hope I get time!

antoinetran:
Ok I could test with github codespace, what an amazing tool!! Could spawn a k3s cluster and reproduce the issue of test `TestGlobalArtifactPassing`:
```
/bin/sh: 1: cannot create /tmp/artifact.txt: Permission denied
```
The error is because, in the container main, the UID is 8737 (as before the PR), however the difference is that now that the securityContext has been removed, the volumeMount below:
```
    volumeMounts:
    - mountPath: /tmp/artifact.txt
      name: input-artifacts
      subPath: input
  volumes:
  - emptyDir: {}
    name: tmp-dir-argo
```
as mounted as root.

antoinetran:
I tried doing chmod in code for group, so that we have:
```
-rw-rw-r-- 1 8737 root    0 Apr  7 17:34 artifact.txt
```
I still get permission denied. After losing a few hairs, I found out this is because of sysctl protected_regular (see [here](https://docs.kernel.org/admin-guide/sysctl/fs.html#protected-regular)). In `/tmp` in particular, noone can modify any file, even root, if the owner is not the user. Even if we have 777.

I changed the test workflow so that we get the input in `/data/artifact.txt` instead of `/tmp/artifact.txt` and now it works.

The fundamental question for me is: is-it ok if we change the default behavior from
```
-rw------- 1 root root    0 Apr  7 17:34 artifact.txt
```
to
```
-rw-rw-r-- 1 8737 root    0 Apr  7 17:34 artifact.txt
```
?
In most workflows, this will work unless they are using /tmp.

antoinetran:
I keep having failure or pending phase for the step onExit at TestStopBehavior test. In github codespace, I ran multiple time the test. Sometime it works, sometime it fails with "failed" phase, sometime with "pending". When I look at the pod logs, nothing is wrong. Exit code 0. I tried changing timeout of the test, no change. I start to believe this test failure has nothing to do with my modification.

@tczhao  any idea?

test/e2e/smoke/runasnonroot-output-params-pipeline.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   entrypoint: main
   securityContext:
     runAsNonRoot: true
-    runAsUser: 8737
+    runAsUser: 12345
```
tczhao, line 9: What's the purpose of this change?

test/e2e/functional/stop-terminate.yaml
```diff
@@ -16,16 +16,38 @@ spec:
...
       container:
         image: argoproj/argosay:v1
-        command: [ sleep ]
-        args: [ "5" ]
+        command:
+          - /bin/bash
+          - "-c"
+          - "-x"
+          - |
+            date -Is --uc
```
tczhao, line 49: I couldn't find the reference for `--uc` Do you mean `--utc`

workflow/common/security_context.go
```diff
@@ -10,7 +10,7 @@ func MinimalCtrSC() *corev1.SecurityContext {
 		Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
 		Privileged:               ptr.To(false),
 		RunAsNonRoot:             ptr.To(true),
-		RunAsUser:                ptr.To(int64(8737)),
+		RunAsUser:                nil,
```
tczhao, line 13: I think we better make this configurable so it doesn't become a breaking change. e.g. set from env var but defaults to 8737.

tczhao (changes requested):

tczhao:
>I start to believe this test failure has nothing to do with my modification.
Yea, it was due to a flaky test.

test/e2e/functional/stop-terminate.yaml
antoinetran: mistyped (but it somehow still works), fixed

test/e2e/smoke/runasnonroot-output-params-pipeline.yaml
antoinetran: The runasnonroot test purpose is to test any non root user and see if argo can pass the UID. There is no reason to expect the UID choice to be exactly the Argo default UID, it can be anything. The fix in the pull request is supposed to removed any hard-coded 8737 except where is is truly needed, so I replaced this.

workflow/common/security_context.go
antoinetran: But it is already configurable, not with env var but with:
* for argo server: security context in argo helm ([here](https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/values.yaml#L494)), or directly in the deployment yaml
* for argo controller: security context in argo helm ([here](https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/values.yaml#L167)), or directly in the deployment yaml
* for argo artgc: with podSpecPatch as described in workaround of the issue
* for argo init/wait: security context in argo helm ([here](https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/values.yaml#L465)), or directly in the deployment yaml

My change will be seamless, except that now the Argo init container writes artifacts with UID 8737 instead of root. The break can happen in the case of the e2e test (write in `/tmp`), and can be fixed easily with a cp (as done in the test), or a change to another path. This is the only breaking case that I can see.

antoinetran:
Any news of this?

PR #14291, opened 2025-03-12
fix: Prevent dropdown from disappearing when shrinking the viewport

### LINKED ISSUE #14290: UI issue: Dropdown disappears when shrinking the viewport
### Pre-requisites

When the UI is resized, argo-dropdown moves out of the screen.

Image

### Version(s)

:latest

```YAML
kind: Workflow                  # new type of k8s spec
metadata:
  generateName: hello-world-    # name of the workflow spec
spec:
  entrypoint: hello-world       # invoke the hello-world template
  templates:
    - name: hello-world         # name of the template
      container:
        image: busybox
        command: [ echo ]
        args: [ "hello world" ]
        resources: # limit the resources
          limits:
            memory: 32Mi
            cpu: 100m
```

```text
none
```

```text
none
```

garireo2549:
…14290

Fixes #14290

### Motivation

When multiple browsers are displayed on a single screen, argo-dropdown moves off-screen, making it unusable.

### Modifications

- Fixed CSS issues and set argo-dropdown to a fixed position.

スクリーンショット 2025-03-12 17 26 13

### Verification

スクリーンショット 2025-03-12 17 42 27

### DIFF
--- a/ui/src/shared/components/dropdown/dropdown.scss
+++ b/ui/src/shared/components/dropdown/dropdown.scss
@@ -8,7 +8,8 @@
     }
 
     &__content {
-        position: fixed;
+        position: absolute;
+        left: 80px !important;
         z-index: $notifications-z-index;
         padding: 1rem;
         background-color: $white-color;


tczhao (approved):
Thanks for the fix.

The `important` overwrite make sense since the left menu bar is using a fixed width

PR #14302, opened 2025-03-16
build: add devcontainer-lock.json and pin base image

MasonM:
### Motivation

The Dev Container currently isn't reproducible, which means contributors can run into weird bugs that we can't reproduce locally.

### Modifications

* Add a `devcontainer-lock.json` file, which was auto-generated by VS Code, so that the features we use are pinned. Spec: https://github.com/devcontainers/spec/blob/482e5f9e022c82f0f12e776f196df81a440eae03/docs/specs/devcontainer-lockfile.md
* Update Dependabot to keep the lockfile up-to-date. Docs: https://containers.dev/guide/dependabot
* Pin the base image to the Ubuntu version used in CI: https://github.com/argoproj/argo-workflows/blob/8691b9b186ef0d8157933d35d1d658b0e1729b74/.github/workflows/ci-build.yaml#L30

### Verification

Rebuilt the dev container in VS Code
### Documentation
N/A

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,24 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {
+      "version": "2.12.0",
+      "resolved": "ghcr.io/devcontainers/features/docker-in-docker@sha256:5f3e2005aad161ce3ff7700b2603f11935348c039f9166960efd050d69cd3014",
+      "integrity": "sha256:5f3e2005aad161ce3ff7700b2603f11935348c039f9166960efd050d69cd3014"
+    },
+    "ghcr.io/devcontainers/features/go:1": {
+      "version": "1.3.1",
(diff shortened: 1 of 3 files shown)


Joibel (approved):
Looks like a good change - easiest just to push this and see what breaks for anyone.

MasonM:
@Joibel Thanks for the details. That's an incompatibility with Ubuntu 24.04. The [official docs for `mcr.microsoft.com/vscode/devcontainers/base:ubuntu`](https://github.com/devcontainers/images/tree/main/src/base-ubuntu)  say that the `ubuntu` tag tracks the latest Ubuntu LTS release, which is 24.04, but it's currently pointing to 22.04, the previous LTS release. This is a good example of why pining the image is important, because as soon as Microsoft gets around to updating that tag to point to 24.04, everyone provisioning the dev container would've had that error, and you wouldn't be able to reproduce it unless you manually pulled the image. 

Anyway, I entered https://github.com/argoproj/argo-workflows/pull/14323 to downgrade back to 22.04.

PR #14306, opened 2025-03-17
fix: properly filter map fields in nodes for large workflows. Fixes #14305.

### LINKED ISSUE #14305: Support fields filtering for map values in Workflow API responses
# Summary

The `Fields` parameter in WorkflowGetRequest [allows filtering](https://github.com/argoproj/argo-workflows/blob/main/server/workflow/workflow_server.go#L160) response fields like `status.phase` or `status.startedAt`. However, this filtering doesn't work properly for map structures like `status.nodes`.

When I specify fields like `status.nodes.id` or `status.nodes.phase`, I still get all fields for all nodes. Using the exclusion pattern with `-status.nodes.templateName` also doesn't work.

The current [Cleaner implementation](https://github.com/argoproj/argo-workflows/blob/main/util/fields/fields.go) doesn't respect field filtering within map values. This results in much larger responses than necessary when retrieving workflows with many nodes.

## Use Cases
- Reducing response size for workflows with many nodes (10 000+) to avoid GRPC message size limits
- Improving performance when only specific node fields are needed
- Reducing network bandwidth usage for frequent workflow status checks

Current workaround is to request only `status.phase,status.startedAt,status.finishedAt` and skip `status.nodes` entirely, but this loses important node status information.

---

sosiska:
Fixes #14305

### Motivation

Large workflows with thousands of nodes currently return overly large responses. Filtering subfields (e.g., `status.nodes.id`) does not work as expected, causing all node fields to appear in the response. This PR addresses that limitation by properly handling map subfield filters for `status.nodes` and reducing response size.

### Modifications

- Enhanced the `Cleaner` logic to support filtering nested fields in maps (`status.nodes.*`).
- Added new unit tests (`TestCleanNodeMapFields` and `TestCleanNodeMapFieldsExclude`) to confirm that excluded node fields are stripped out and included node fields remain.
- Verified no breaking changes to existing filtering behavior outside of maps.

### Verification

- **Note**: I encountered issues running `make pre-commit -B` on the main branch due to local path/module constraints (the codegen step conflicts with v3 module paths). However, basic linting and tests pass successfully.
- Confirmed all relevant Go tests (including newly added ones) are successful.
- Commits are signed off (DCO).

### Documentation

- No additional user-facing documentation is needed since this fixes existing behavior rather than adding a new feature.
  - (If needed, we can add a small note in the existing docs about how map field filtering works.)


tczhao (commented):
The `Reducing response size for workflows with many nodes (10 000+) to avoid GRPC message size limits` seems to be a necessary fix.

A documentation is needed since this introduced a new behaviour.
I'm not an expert in design/define API behaviour, would prefer to have another reviewer to have a look

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+					break
 				}
-			} else {
+			}
+
+			if shouldExclude || (f.exclude && pathIn) || (!f.exclude && !parentPathIn) {
 				delete(mapItem, k)
+			} else {
+				f.cleanItem(append(path, k), v)
 			}
```
Joibel, line 152: **Include-mode exact matches no longer keep their subtree — breaks `argo list`, the UI, and the documented `fields` contract.**

The old code only recursed when `!pathIn`, so an exactly-requested field kept its entire subtree. This branch now recurses unconditionally when keeping a key, and the recursion then deletes every child, because no requested field has the prefix `child.path.`.

Consequences:

- `fields=metadata` returns `metadata: {}`. The CLI's own defaults hit this — `argo list` would render empty names/namespaces:
  https://github.com/argoproj/argo-workflows/blob/32cbee030e5bfb4f3c0e819312c7ae65a501860e/cmd/argo/commands/list.go#L42-L45
- The UI requests `items.metadata.labels`, `items.metadata.annotations`, `items.spec.arguments` (`ui/src/shared/services/workflows-service.ts`) — all become empty objects.
- `fields=status.nodes` alone now returns an **empty nodes map**: the special case above finds no `status.nodes.*` entries, falls through to this loop, and every node key is deleted (`status.nodes.<nodeID>` matches nothing).

No existing or new test covers a bare subtree include (they all use leaf selectors), so the suite passes despite this.

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+
+				if strings.HasPrefix(fieldName, nodePath+".") {
+					parts := strings.SplitN(strings.TrimPrefix(fieldName, nodePath+"."), ".", 2)
+					if len(parts) == 1 {
+						if exclude {
+							excludeFields[parts[0]] = true
+						} else {
+							includeFields[parts[0]] = true
+						}
+					}
```
Joibel, line 97: **Node subfield selectors deeper than one segment are silently dropped — and delete explicitly requested data.**

`len(parts) == 1` discards e.g. `status.nodes.outputs.parameters`:

- `fields=status.nodes.id,status.nodes.outputs.parameters` → `includeFields={id}` only, so `outputs` is deleted from every node even though it was explicitly requested — and the early `return` below prevents deeper recursion from honouring it.
- `fields=status.nodes.outputs.parameters` alone → both sets stay empty → falls through to the standard loop, which deletes every node key → empty nodes map.

Either support deeper paths or reject them; silently returning less than requested is the worst of both.

Joibel (commented):
### Code review

Found 9 issues. Eight are posted as inline comments on the diff; the remaining one anchors to files this PR doesn't touch:

1. **New user-facing `fields` syntax is undocumented — and documentation was explicitly requested in review.** The per-entry `-` form (e.g. `status.phase,-status.nodes.message`) exists nowhere in the proto comments, the generated swagger, or `docs/`. The only description of the parameter is the proto comment, which documents a different grammar (a single leading `-` on the whole string):

   https://github.com/argoproj/argo-workflows/blob/32cbee030e5bfb4f3c0e819312c7ae65a501860e/pkg/apiclient/workflow/workflow.proto#L26-L28

   https://github.com/argoproj/argo-workflows/blob/32cbee030e5bfb4f3c0e819312c7ae65a501860e/pkg/apiclient/workflow/workflow.proto#L33-L35

   @tczhao already asked for this ("A documentation is needed since this introduced a new behaviour"). The PR description's "no additional user-facing documentation is needed since this fixes existing behavior rather than adding a new feature" doesn't hold: #14305 is a feature request, and the mixed per-field `-` syntax did not exist before this PR.

Notes, not counted as findings:

- The PR is currently in merge conflict with `main` — `util/fields/fields.go` has since been modernized (`interface{}` → `any`, `strings.SplitSeq`, shadow lint, `/v4` module path), so a rebase is needed before CI signal means much.
- One piece is worth keeping standalone regardless of the rest: the dot-qualified prefix fix `strings.HasPrefix(checkField, fieldPath+".")` corrects a real pre-existing false positive (a request for `status.phaseX` previously kept `status.phase`).

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+			excludeFields := make(map[string]bool)
+
+			for field := range f.fields {
+				fieldName := field
+				var exclude bool
+
+				if strings.HasPrefix(field, "-") {
+					fieldName = field[1:]
+					exclude = true
+				}
```
Joibel, line 87: **The new per-field `-` syntax isn't understood by `matches`/`WillExclude` — the server skips node hydration.**

`matches` does plain `HasPrefix` in both directions against the raw stored string, so `-status.nodes.message` matches neither direction against `status.nodes`, and `WillExclude("status.nodes")` returns `!f.exclude` = `true`:

https://github.com/argoproj/argo-workflows/blob/32cbee030e5bfb4f3c0e819312c7ae65a501860e/util/fields/fields.go#L50-L68

The server gates offloaded-node hydration on `WillExclude("status.nodes")` (`server/workflow/workflow_server.go`). So `fields=status.phase,-status.nodes.message` — the exact shape this PR's own tests use — makes the server skip `hydrator.Hydrate`/offload loading entirely: excluding **one** node field returns **no nodes at all** for workflows with offloaded node status.

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+
+				if strings.HasPrefix(fieldName, nodePath+".") {
```
Joibel, line 89: **The new syntax is silently inert on `WatchWorkflows`.**

The watch endpoint builds its cleaner with `NewCleaner(req.Fields).WithoutPrefix("result.object.")` (`server/workflow/workflow_server.go`). `WithoutPrefix` uses `strings.TrimPrefix(k, prefix)`:

https://github.com/argoproj/argo-workflows/blob/32cbee030e5bfb4f3c0e819312c7ae65a501860e/util/fields/fields.go#L161-L167

`TrimPrefix` cannot strip `result.object.` from `-result.object.status.nodes.message` — the `-` is at position 0 — so the field is stored untrimmed and never matches this `nodePath+"."` prefix check. The same query filters correctly on `GetWorkflow` but does nothing on the watch endpoint.

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+			nodePath := strings.Join(path, ".")
+
+			// Collect fields for node filtering
+			includeFields := make(map[string]bool)
+			excludeFields := make(map[string]bool)
+
+			for field := range f.fields {
+				fieldName := field
+				var exclude bool
+
```
Joibel, line 83: **Per-field `-` collides with `NewCleaner`'s global `-` parsing — the grammar is position-dependent.**

`NewCleaner` strips a leading `-` from the *whole* string and sets a global exclude flag before splitting on commas:

https://github.com/argoproj/argo-workflows/blob/32cbee030e5bfb4f3c0e819312c7ae65a501860e/util/fields/fields.go#L8-L20

So `fields=-status.nodes.message,-status.nodes.inputs` produces `f.exclude=true` with fields `{status.nodes.message, -status.nodes.inputs}` — the first entry silently becomes a global-exclude entry (which is itself a no-op here, see the comment below) while the second is a per-field exclude. The meaning of an entry depends on its position in the list.

Relatedly, in include mode a negative entry marks its ancestors as included (the parent-path check strips `-` before matching), so `fields=metadata.name,-spec.templates` keeps `spec` as an empty object in the response.

The syntax should be parsed once in `NewCleaner` into explicit include/exclude sets shared by `cleanItem`, `matches`/`WillExclude`, and `WithoutPrefix`.

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+				for _, nodeMap := range mapItem {
+					if nodeObj, ok := nodeMap.(map[string]interface{}); ok {
+						for key := range nodeObj {
+							if excludeFields[key] || !f.exclude && len(includeFields) > 0 && !includeFields[key] {
+								delete(nodeObj, key)
+							}
+						}
+					}
+				}
+				return
```
Joibel, line 112: **The documented exclusion form is a silent no-op for node subfields — the exact case from #14305.**

With `fields=-status.nodes.message` (the documented syntax — the proto comment's `"-status.nodes"` form), `NewCleaner` strips the `-` and sets `f.exclude=true`, so the entry lands in `includeFields` here. This delete condition is then always false — `excludeFields[key]` is false and the second arm is gated by `!f.exclude` — so nothing is deleted, and the `return` skips all further cleaning of the subtree. Issue #14305 explicitly cites `-status.nodes.templateName` not working; after this PR it still doesn't. Only the new, undocumented mixed form (`status.phase,-status.nodes.message`) works.

util/fields/fields.go
```diff
@@ -69,24 +69,86 @@ func (f Cleaner) matches(x string) bool {
...
+
+			// Check for fields with "-" prefix
+			shouldExclude := false
+			for field := range f.fields {
+				if strings.HasPrefix(field, "-") && fieldPath == field[1:] {
+					shouldExclude = true
+					break
 				}
-			} else {
+			}
```
Joibel, line 146: **Two O(len(fields)) scans per map key, in the hot path of a performance-motivated change.**

This new `shouldExclude` loop is equivalent to the O(1) lookup `f.fields["-"+fieldPath]`, and it runs for every key of every map at every recursion level, on top of the parent-path scan above (which also re-strips `-` on every iteration). The `-`-prefixed set — like the include/exclude sets rebuilt each time a `nodes` map is encountered — is constant for the Cleaner's lifetime and could be computed once in `NewCleaner`. For the 10,000+-node workflows this PR targets, that's O(nodes × keys × fields) avoidable work.

util/fields/fields_test.go
```diff
@@ -76,3 +128,68 @@ func TestCleanFieldsExclude(t *testing.T) {
...
+func TestCleanNodeMapFields(t *testing.T) {
+	var wf, cleanWf wfv1.Workflow
+	wfv1.MustUnmarshal([]byte(complexWorkflow), &wf)
+	fields := "status.phase,status.startedAt,status.finishedAt,status.nodes," +
+		"status.nodes.id,status.nodes.name,status.nodes.displayName," +
+		"status.nodes.type,status.nodes.phase"
+
+	ok, err := NewCleaner(fields).Clean(wf, &cleanWf)
+	require.NoError(t, err)
+	assert.True(t, ok)
```
Joibel, line 141: **These tests are shaped so they can't catch the regressions this PR introduces.**

Both new tests pair `status.nodes` with single-segment subfield filters, which always takes the special-case early-return path — so the bare-subtree breakage (`fields=status.nodes` alone, or `fields=metadata`, now return empty objects; see comments on `fields.go`) is never exercised. Also untested: the documented global-exclude form (`-status.nodes.message`, currently a no-op), subfields deeper than one segment, and `WillExclude`/`WithoutPrefix` with `-`-prefixed entries.

Smaller points: `node-2` is never asserted on (only `assert.Len(..., 2)`); `status.nodes.type` is requested but `node1.Type` is never checked; and there are no positive-control assertions that the fixture contains these values *before* cleaning — a typo in the fixture YAML would make the `assert.Empty` checks pass vacuously.

Issue #13114, opened 2024-05-29
UI: Open custom `links` in new tab or with router navigation depending on address


agilgur5:
# Summary

The custom [Links](https://argo-workflows.readthedocs.io/en/latest/links/) feature allows you to specify links in various places in the UI, but does not currently specify _how to open_ the link.

It does allow `Ctrl+Click` to open in new tab as is standard in the browser, but some users would prefer that be the default, per https://github.com/argoproj/argo-workflows/discussions/12605#discussioncomment-9469587, https://github.com/argoproj/argo-workflows/issues/11112#issuecomment-1746988447, and this [Slack thread](https://cloud-native.slack.com/archives/C01QW9QSSSK/p1707989479129399).

Similarly, per https://github.com/argoproj/argo-workflows/issues/11112#issuecomment-1558714545, some users would want internal link to load via router navigation instead of reloading the whole page.

In the Slack thread, I suggested:
> We might want to auto-detect internal vs external links and make external ones open in a new tab by default.

## Use Cases

1. If you link to a Logs provider from the Logs Viewer, you might want to open that in a new tab by default
1. If you link to a specific filter in the Workflows List, you might want that to use router navigation to change the filter without reloading the page
1. ... and similar ...

## Implementation Details

1. Check if the link URL matches the UI's URL (consider hostname, subdomain, and base href), if so, use router navigation
1. If not, open in a new tab by default

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

agilgur5:
Possibly something for @Adrien-D to tackle?

PR #14314, opened 2025-03-20
feat: support open custom links in new tab.Part of #13114

shuangkun:
Fixes #TODO


ui/src/shared/models/info.ts
```diff
@@ -1,6 +1,7 @@
 export interface Link {
     name: string;
     scope: string;
+    type?: string;
```
tczhao, line 4: open for discussion,
I'm thinking if using `target` makes more sense, user can use value as if its HTML target attribute. `_blank`, `_self`
`internal` is not very precise since its opening the link in the same tab

Joibel: I agree with @tczhao

pkg/apis/workflow/v1alpha1/info.go
```diff
@@ -8,6 +8,8 @@ type Link struct {
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
 	// "workflow", "pod", "pod-logs", "event-source-logs", "sensor-logs", "workflow-list" or "chat"
 	Scope string `json:"scope" protobuf:"bytes,2,opt,name=scope"`
+	// Target attribute specifies where a linked document will be opened when a user clicks on a link. E.g. "_blank", "_self". If the target is _blank, it will open in a new tab.
+	Target string `json:"target" protobuf:"bytes,4,opt,name=target"`
 	// The URL. Can contain "${metadata.namespace}", "${metadata.name}", "${status.startedAt}", "${status.finishedAt}" or any other element in workflow yaml, e.g. "${workflow.metadata.annotations.userDefinedKey}"
 	URL string `json:"url" protobuf:"bytes,3,opt,name=url"`
```
tczhao: The order is incorrect
```suggestion
	// The URL. Can contain "${metadata.namespace}", "${metadata.name}", "${status.startedAt}", "${status.finishedAt}" or any other element in workflow yaml, e.g. "${workflow.metadata.annotations.userDefinedKey}"
	URL string `json:"url" protobuf:"bytes,3,opt,name=url"`
	// Target attribute specifies where a linked document will be opened when a user clicks on a link. E.g. "_blank", "_self". If the target is _blank, it will open in a new tab.
	Target string `json:"target" protobuf:"bytes,4,opt,name=target"`
```

tczhao (changes requested):

pkg/apis/workflow/v1alpha1/info.go
shuangkun: OK, Modified.

PR #14316, opened 2025-03-20
feat: add jobs parameter to support multiple files concurrently for OSS.

shuangkun:
Supports parallel upload when there are multiple files. 

reference: https://github.com/aliyun/ossutil/blob/778640b2c45c644f174b0ea94e4816bf19142727/lib/cp.go#L960


tczhao (commented):
Could you describe how you verify the change

PR #14318, opened 2025-03-20
refactor: tidy tools in makefile

Joibel:
Fixes the problem found in https://github.com/argoproj/argo-workflows/pull/14302#issuecomment-2742883052

### Motivation

Whilst adding another tool not mentioned in this commit I thought this looked like a better pattern for tool management.

### Modifications

Create a makefile variable for all the tool binaries and use them everywhere.

This will force use some tools which were previously discovered by path to be the actual versions we installed, which might help in some cases.

Install `mkdocs` into a venv to fix the regression introduced in #14302 

Added some dependencies which were missed when doing #13623

`clang-format` is still used from the path, because the rule to install it allows this and I didn't feel like investigating this behaviour right now.

The paths for the node tools were not correct, so corrected them.

I don't really understand why the `requirements.txt` change is necessary, but without it the superfences pymdown module is missing when you run make docs.

### Verification

`make -B codegen` works in a devcontainer
`make docs`
CI will test this

### Documentation

Not relevant

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -343,7 +343,7 @@ jobs:
           name: cli
           path: dist/
       - name: Prepare binaries
-        run: chmod +x dist/* && make --touch dist/*
+        run: chmod +x dist/* && touch dist/*
       - name: Set-up /etc/hosts
         run: |
           echo '127.0.0.1 dex'      | sudo tee -a /etc/hosts
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -343,7 +343,7 @@ jobs:
           name: cli
           path: dist/
       - name: Prepare binaries
-        run: chmod +x dist/* && touch dist/*
+        run: chmod +x dist/* && make --touch dist/*
       - name: Set-up /etc/hosts
         run: |
           echo '127.0.0.1 dex'      | sudo tee -a /etc/hosts
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,28 @@ ifeq ($(PROFILE),sso)
 AUTH_MODE                     := sso
 endif
 
+# Makefile managed tools
+TOOL_MOCKERY                := $(GOPATH)/bin/mockery
+TOOL_CONTROLLER_GEN         := $(GOPATH)/bin/controller-gen
+TOOL_GO_TO_PROTOBUF         := $(GOPATH)/bin/go-to-protobuf
+TOOL_PROTOC_GEN_GOGO        := $(GOPATH)/bin/protoc-gen-gogo
+TOOL_PROTOC_GEN_GOGOFAST    := $(GOPATH)/bin/protoc-gen-gogofast
+TOOL_PROTOC_GEN_GRPC_GATEWAY:= $(GOPATH)/bin/protoc-gen-grpc-gateway
+TOOL_PROTOC_GEN_SWAGGER     := $(GOPATH)/bin/protoc-gen-swagger
+TOOL_OPENAPI_GEN            := $(GOPATH)/bin/openapi-gen
+TOOL_SWAGGER                := $(GOPATH)/bin/swagger
+TOOL_GOIMPORTS              := $(GOPATH)/bin/goimports
+TOOL_GOLANGCI_LINT          := $(GOPATH)/bin/golangci-lint
+
+# npm bin -g will do this on later npms than we have
+TOOL_NODE_BINDIR            := $(shell npm config get prefix)/bin
+TOOL_CLANG_FORMAT           := /usr/local/bin/clang-format
+TOOL_MDSPELL                := $(TOOL_NODE_BINDIR)/mdspell
+TOOL_MARKDOWN_LINK_CHECK    := $(TOOL_NODE_BINDIR)/markdown-link-check
+TOOL_MARKDOWNLINT           := $(TOOL_NODE_BINDIR)/markdownlint
+TOOL_MKDOCS_DIR             := $(HOME)/.venv/mkdocs
+TOOL_MKDOCS                 := $(TOOL_MKDOCS_DIR)/bin/mkdocs
+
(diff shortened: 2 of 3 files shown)


tico24 (changes requested):
When I run `make` in a devcontainer, I no longer get a list of possible commands. I get this:

```bash
$ make
GIT_COMMIT=b0fdb2ac814a74d64899a5a383ab195b6d087845 GIT_BRANCH=makefile-tool-tidy GIT_TAG=untagged GIT_TREE_STATE=clean RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=k3d-k3s-default K3D=true DOCKER_PUSH=false TARGET_PLATFORM=linux/arm64
make: Nothing to be done for 'cli'.
```

edit: I was thinking of a different project. I'll go away now.

tico24 (approved):

.github/workflows/ci-build.yaml
```diff
@@ -343,7 +343,7 @@ jobs:
           name: cli
           path: dist/
       - name: Prepare binaries
-        run: chmod +x dist/* && make --touch dist/*
+        run: chmod +x dist/* && touch dist/*
```
MasonM, line 346: This is going to slow down the tests. The reason I added this in https://github.com/argoproj/argo-workflows/pull/14028 is because otherwise every E2E test job is going to rebuild the binaries instead of re-using the pre-built binaries from the `Download controller`/`Download CLI` steps. You can tell that's happening in this PR because the `Start controller/API` step for every E2E test job is taking ~3m, e.g. https://github.com/argoproj/argo-workflows/actions/runs/13989885038/job/39173751538?pr=14318

Normally, that step takes ~1m, e.g. https://github.com/argoproj/argo-workflows/actions/runs/13966659800/job/39098757154?pr=14316

The reason `touch dist/*` isn't sufficient is because of the pre-requisites for the Make targets. Here's a command to see what it's doing:
```shell
$ make --dry-run --trace dist/workflow-controller  | grep ^Make
Makefile:362: update target '/home/vscode/go/bin/openapi-gen' due to: Makefile
Makefile:657: update target 'pkg/apis/workflow/v1alpha1/openapi_generated.go' due to: /home/vscode/go/bin/openapi-gen
Makefile:330: update target '/home/vscode/go/bin/go-to-protobuf' due to: Makefile
Makefile:673: update target 'pkg/apis/workflow/v1alpha1/zz_generated.deepcopy.go' due to: /home/vscode/go/bin/go-to-protobuf
Makefile:252: update target 'dist/workflow-controller' due to: pkg/apis/workflow/v1alpha1/openapi_generated.go pkg/apis/workflow/v1alpha1/zz_generated.deepcopy.go
```

We probably should add `--trace` here so it's easier to identify/debug this sort of thing: 
https://github.com/argoproj/argo-workflows/blob/50219cee29ed4e9fac54703192338d2f8277a2f3/Makefile#L9

Makefile
```diff
@@ -92,6 +92,28 @@ ifeq ($(PROFILE),sso)
...
+TOOL_SWAGGER                := $(GOPATH)/bin/swagger
+TOOL_GOIMPORTS              := $(GOPATH)/bin/goimports
+TOOL_GOLANGCI_LINT          := $(GOPATH)/bin/golangci-lint
+
+# npm bin -g will do this on later npms than we have
+TOOL_NODE_BINDIR            := $(shell npm config get prefix)/bin
+TOOL_CLANG_FORMAT           := /usr/local/bin/clang-format
+TOOL_MDSPELL                := $(TOOL_NODE_BINDIR)/mdspell
+TOOL_MARKDOWN_LINK_CHECK    := $(TOOL_NODE_BINDIR)/markdown-link-check
+TOOL_MARKDOWNLINT           := $(TOOL_NODE_BINDIR)/markdownlint
```
MasonM, line 113: Since this is a simply expanded variable, it's going to execute this every time you run `make`, even if it's not actually used. This takes ~100ms for me, which isn't huge, but that can add up:
```
$ time npm config get prefix
/usr/local/share/nvm/versions/node/v20.19.0

real    0m0.114s
user    0m0.104s
sys     0m0.024s
```

The devcontainer uses `nvm` for handling Node versions, which makes this available in the `NVM_BIN` variable: https://github.com/nvm-sh/nvm?tab=readme-ov-file#environment-variables

So you could use optional assignment to make use of that and eliminate the unnecessary overhead, while still preserving compatibility for people without `nvm`:
```suggestion
NVM_BIN                     ?= $(shell npm config get prefix)/bin
TOOL_CLANG_FORMAT           := /usr/local/bin/clang-format
TOOL_MDSPELL                := $(NVM_BIN)/mdspell
TOOL_MARKDOWN_LINK_CHECK    := $(NVM_BIN)/markdown-link-check
TOOL_MARKDOWNLINT           := $(NVM_BIN)/markdownlint
```

docs/requirements.txt
```diff
@@ -1,2 +1,3 @@
 mkdocs-material==9.*
 mkdocs-redirects==1.*
+pymdown-extensions
```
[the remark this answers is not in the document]
Joibel: Maybe, I don't really understand why I had to add it to make this work. Done.

.github/workflows/ci-build.yaml
Joibel: I've reverted to demonstrate why I'd changed it:
https://github.com/argoproj/argo-workflows/actions/runs/14036203761/job/39295011929?pr=14318

```
Run chmod +x dist/* && make --touch dist/*
  chmod +x dist/* && make --touch dist/*
  shell: /usr/bin/bash -e {0}
  env:
    KUBECONFIG: /home/runner/.kubeconfig
    E2E_ENV_FACTOR: 2
GIT_COMMIT=44124a468207cd41e92059399af665667f33c465 GIT_BRANCH=HEAD GIT_TAG=untagged GIT_TREE_STATE=clean RELEASE_TAG=false DEV_BRANCH=true VERSION=latest
KUBECTX=default K3D=false DOCKER_PUSH=false TARGET_PLATFORM=linux/amd64
make: touch: open: /bin/openapi-gen: Permission denied
touch /bin/openapi-gen
```

I hadn't realised what you'd done, and why, so it's a good point.

I have moved some stuff around to hopefully fix my self-induced issue.

MasonM (approved):
Thanks, this is a lot cleaner. I confirmed `make docs` works on the Ubuntu 22.04 image with these changes.

Joibel:
> Thanks, this is a lot cleaner. I confirmed `make docs` works on the Ubuntu 22.04 image with these changes.

Thanks for verifying. This does also work on 24.04 because of the venv change, so one of us can put in a PR for that once this is in.

isubasinghe (approved):

PR #14341, opened 2025-03-31
build: add CI to pre-build and validate dev container

MasonM:
### Motivation
This updates the [dev container](https://containers.dev/) to use a pre-built image generated using GitHub actions, which greatly reduces the time to get a dev environment up and running. It also serves to validate changes to the dev container definitions in PRs.

### Modifications

This is loosely based on the guide at [Speed Up Your Workflow with Prebuilds](https://containers.dev/guide/prebuild). Initially, I was hoping to use [Codespaces prebuilds](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds), but we can't because it's only available for organizations in a team or enterprise plan, which `argoproj` isn't. From what I gathered, the way prebuilds work in Codespaces is very similar to what's depicted in the guide. There's a few notable differences between this PR and the guide:
1. Instead of using the [devcontainers/ci](https://github.com/devcontainers/ci/) action to build the image, I directly called [devcontainers/cli](https://github.com/devcontainers/cli), since that's what the former uses under the hood, and it gives us more flexibility.
2. Pushes the image to quay.io instead of ghcr.io, since all our other images are on Quay. I'm not 100% sure this is going to work, since I don't have access to push images to https://quay.io/organization/argoproj. In my fork, I tested this using ghcr.io: https://github.com/MasonM/argo-workflows/pkgs/container/argo-workflows-devcontainer
3. Instead of using the [inline cache](https://docs.docker.com/build/cache/backends/inline/) for Docker caching, I had it use the [registry cache](https://docs.docker.com/build/cache/backends/registry/), as that's more flexible and produces much smaller images.

Additionally, I made a couple minor optimizations to improve the start up time:
1. Added a `Dockerfile` containing k3d and some other tools that rarely change.
2. Removed `make kit` and `kit build` from `.devcontainer/pre-build.sh` because installing Kit and building deps is handled automatically when running a command that needs to do so, e.g. `make start`. If you're just editing documentation, then you won't need to do that.

For Codespaces, I verified in my fork that this reduces the time to have a fully ready container from ~13m to ~1m30s. Most of that ~1m30s is spent pulling the pre-built image, which is ~843MB compressed, and ~2.5GB uncompressed:
```
$ docker manifest inspect ghcr.io/masonm/argo-workflows-devcontainer:latest --verbose | jq '.[0] | [.OCIManifest | .layers[] | .size] | add / 1024 / 1024'
842.8535175323486

$ docker image ls --filter=reference=ghcr.io/masonm/argo-workflows-devcontainer
REPOSITORY                                   TAG       IMAGE ID       CREATED      SIZE
ghcr.io/masonm/argo-workflows-devcontainer   latest    b81831b01361   3 days ago   2.5GB
```

### Verification
Created a CodeSpace at https://github.com/MasonM/argo-workflows/:
image

I also verified I can load the devcontainer in VS Code on both my amd64 Linux desktop and arm64 Macbook:
image
### Documentation
I restructured the "Running Locally" page to make it clearer. Also, I made it clear that using the dev container is the fastest way to get started, since many people (myself included) wasted a lot of time trying to manually set things up before giving up and using the dev container.

Preview: https://masonm-argo-workflows.readthedocs.io/en/latest/running-locally/


Makefile
```diff
@@ -818,3 +833,22 @@ release-notes: /dev/null
...
+
+$(TOOL_DEVCONTAINER): Makefile
+	npm list -g @devcontainers/cli@0.75.0 > /dev/null || npm i -g @devcontainers/cli@0.75.0
+
+.PHONY: devcontainer-build
+devcontainer-build: $(TOOL_DEVCONTAINER)
+	devcontainer build \
+		--workspace-folder . \
+		--config .devcontainer/builder/devcontainer.json \
+		--platform $(TARGET_PLATFORM) \
```
tczhao, line 847: I think we need set buildx context then build it to allow it to be multiarch

.devcontainer/devcontainer.json
```diff
@@ -1,17 +1,10 @@
...
-      "version": "20"
-    },
-    "ghcr.io/devcontainers/features/docker-in-docker:2": {},
-    "ghcr.io/devcontainers/features/python:1": {}
-  },
+  // Dev container definition. Reference: https://containers.dev/implementors/json_reference/
+  "name": "Argo Workflows (pre-built)",
+
+  // This image is built and pushed by .github/workflows/devcontainer.yaml using .devcontainer/builder/devcontainer.json
+  "image": "quay.io/argoproj/argo-workflows-devcontainer",
```
tczhao, line 6: shall we version this? Although we don't change devcontianer that much but we want to make sure it works when we use a different argo version instead of figuring it how to make it work

docs/running-locally.md
```diff
@@ -1,53 +1,43 @@
...
+Next, add the following to your `/etc/hosts`:
 
-Note:
+```text
+127.0.0.1 dex
+127.0.0.1 minio
+127.0.0.1 postgres
+127.0.0.1 mysql
+127.0.0.1 azurite
+```
```
tczhao, line 22: I don't think this is needed for devcontainer unless something changed in this PR
I use wsl2 without any of this setup, it works fine

tczhao (commented):
Great idea!

Makefile
MasonM: I think you mean we need to use the [docker container build driver](https://docs.docker.com/build/builders/drivers/docker-container/). That's handled by the `docker/setup-buildx-action` action, and you can see that running in https://github.com/argoproj/argo-workflows/actions/runs/14185152964/job/39739033877?pr=14341

Specifically, the action executes this command to create the driver and use it as the default:
```
  /usr/bin/docker buildx create --name builder-fe8bbb6f-d828-4446-a45a-4e7df05ff27d --driver docker-container --buildkitd-flags --allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host --use
``

docs/running-locally.md
MasonM: You're right, this is only need for manual installation. I pushed a commit to move this to that section.

.devcontainer/devcontainer.json
MasonM: I thought about that, but unfortunately Dependabot doesn't support updating the `"image"` property in `devcontainer.json` yet (https://github.com/dependabot/dependabot-core/issues/7000), so there's no easy way of automating that yet AFAIK.

This image should work fine for older Argo releases, though I haven't tested that extensively. The only thing you might need to do manually is download an older `kubectl` version if you need to test old Kubernetes version. Note that the `pre-build.sh` script is not run as part of the image build process, so the cluster created by `k3d cluster create` is not part of the image.

MasonM:
Build is failing when pulling packages from the NPM registry, which is clearly unrelated, and it seems this is a NPM-wide issue: https://status.npmjs.org/

tczhao (approved):
Nice, works with wsl2

blkperl (approved):
Great job MasonM!  Excited to see contributors using this! 🌸

PR #14352, opened 2025-04-04
fix: argoexec kill command in Windows to use osspecific.Kill

### LINKED ISSUE #14297: Argoexec kill command does not work in case of Windows container failures
### Pre-requisites

This is a follow-up of https://github.com/argoproj/argo-workflows/pull/13693.

When a Windows container fails e.g. due to a crash it goes into Error state. The `wait` container will trigger `argoexec kill` command to signal to argo workflow controller that the step should be retried by creating another Pod. In the Windows container case, the error failed with code 64 (error in argoexec), since `os.Kill` was used. Upon replacing the latter with `osspecific.Kill` the issue was still not solved because as opposed to Linux, Windows containers don't have PID 1 so `osspecific.Kill` had to be adapted to also find the PID of the `argoexec` process. In our tests the fix works as expected, now Windows containers in Error state don't get stuck anymore, Argo will create another Pod and when the workflow completes the erroring Pod will be cleaned up.  Please see our fix here: https://github.com/argoproj/argo-workflows/compare/main...helio:argo-workflows:fix-argoexec-kill-windows.

We have a question before submitting the PR, in `osspecific.Kill` we don't really know if we want to kill PID 1 or not, so maybe we'd need to extract the "find PID logic" outside of `osspecific.Kill` and supply its result to the kill function. However we would need a os check condition, or even better add a method to the `signal_<os>.go` files to retrieve the PID (in Linux it would be just a "return 1"). Any thoughts on this and the PR? Thanks! 

/cc @mweibel

### Version(s)

v3.6.5

```YAML
n/a
```

```text
n/a
```

```text
n/a
```

criscola:
Fixes #14297 (see discussion also).

### Motivation

When a Windows container fails e.g. due to a crash it goes into Error state. The wait container will trigger argoexec kill command to signal to argo workflow controller that the step should be retried by creating another Pod. In the Windows container case, the error failed with code 64 (error in argoexec), since os.Kill was used. Upon replacing the latter with osspecific.Kill the issue was still not solved because as opposed to Linux, Windows containers don't have PID 1 so osspecific.Kill had to be adapted to also find the PID of the argoexec process. In our tests the fix works as expected, now Windows containers in Error state don't get stuck anymore, Argo will create another Pod and when the workflow completes the erroring Pod will be cleaned up.

### Modifications

Modifies argoexec kill command to use `osspecific.Kill` and implement the relative Windows code for it.

### Verification

Triggered a bluescreen by killing `svchost.exe` in the container, then seen argoexec not fail anymore when trying to kill the erroring Pod.


criscola:
Hello, can someone please help me understand what's wrong with the tests? I bet most of them are flakes, however, I re-run the testing suite about 3 times and I'm still getting failures, in particular, could [this](https://github.com/argoproj/argo-workflows/actions/runs/15563856915/job/43823248917?pr=14352) be a flake or does changing from `p.Kill` to `osspecific.Kill` might have had some side-effects I'm not aware of? cc those who I see contributing to `area/windows` most recently: @juliev0 @agilgur5 @Joibel

PR #14373, opened 2025-04-11
fix: added pagination to the workflow and pod listings. Fixes #14374

### LINKED ISSUE #14374: Listing of workflows/pods should paginate to support the case where there is a large number  of objects.
### Pre-requisites

The current informer logic's ListFunc directly calls the clients List without any sort of pagination. 
While this is good enough when there is a small number of workflows, it becomes unresponsive for a large number. These list functions should change to support pagination. 

### Version(s)

all

```YAML
NA
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #14374

### Motivation

This change ensures that pagination happens in our list calls to both pod and workflows. This is needed because when k8s has a large number of resources, it is generally better to perform some amount of pagination. 

### Modifications

Simple change that modified the list calls to use 'continue on' tokens.

### Verification

Tested to ensure the k8s api calls function as is.

### DIFF
--- a/util/unstructured/unstructured.go
+++ b/util/unstructured/unstructured.go
@@ -14,6 +14,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+const workflowPaginationLimit = 500
+
 // NewUnstructuredInformer constructs a new informer for Unstructured type.
 // Always prefer using an informer factory to get a shared informer instead of getting an independent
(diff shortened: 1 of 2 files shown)


isubasinghe:
Should we maybe have a flag that allows it to switch between the two methods? There is a valid argument to be made about this introducing some latency. Giving the user some choice might make sense too.

Joibel (approved):

PR #14389, opened 2025-04-15
fix(server/auth/webhook): update GitHub webhook events list and library version

### LINKED ISSUE #14388: GitHub webhook authentication fails for `workflow_run` event type but works for release event type
### Pre-requisites

When sending GitHub webhook events to Argo Workflows, I've discovered that some event types are properly authenticated while others are rejected with a 401 error. Specifically:

- GitHub `release` events with action `published` are properly processed
- GitHub `workflow_run` events with action `completed` are rejected with error: `{"code":16,"message":"token not valid. see https://argo-workflows.readthedocs.io/en/release-3.5/faq/"}`

The issue appears to be that the GitHub webhook handler in Argo Workflows doesn't recognize or properly authenticate newer GitHub event types like `workflow_run`, `workflow_dispatch`, and `workflow_job`. These events are documented in GitHub's webhook documentation but aren't included in Argo's event handling code.

**Expected behavior:** All valid GitHub webhook events should be authenticated and processed similarly, regardless of event type.

### Version(s)

v3.5.8

```YAML
The issue can be reproduced with this WorkflowEventBinding:
kind: WorkflowEventBinding
metadata:
  name: workflow-run-completed-binding
  namespace: github-webhook-processor
spec:
  event:
    selector: payload.action == "completed" && metadata["x-github-event"] == ["workflow_run"]
  submit:
    workflowTemplateRef:
      name: workflow-run-completed
    arguments:
      parameters:
        - name: payload
          valueFrom:
            event: payload
        - name: github-event
          valueFrom:
            event: metadata["x-github-event"][0]
        - name: github-hook-id
          valueFrom:
            event: metadata["x-github-hook-id"][0]
        - name: github-delivery
          valueFrom:
            event: metadata["x-github-delivery"][0]

And webhook clients secret:
kind: Secret
apiVersion: v1
metadata:
  name: argo-workflows-webhook-clients
  namespace: github-webhook-processor
stringData:
  github.com: |
    type: github
    secret: "your-webhook-secret"
```

```text
argo-workflows-server-d8675cc7c-hw6gs time="2025-04-15T12:54:41.542Z" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = token not valid. see https://argo-workflows.readthedocs.io/en/release-3.5/faq/" grpc.code=Unauthenticated grpc.method=ReceiveEvent grpc.service=event.EventService grpc.start_time="2025-04-15T12:54:41Z" grpc.time_ms=0.04 span.kind=server system=grpc
```

```text
nothing here
```

malpou:
Fixes #14388

### Motivation
GitHub has expanded their webhook events system to include newer event types such as `workflow_run`, `workflow_dispatch`, and `workflow_job`. Currently, Argo Workflows doesn't recognize these newer event types in its GitHub webhook authentication mechanism, causing `401` authentication failures when these events are sent. This PR updates the list of supported GitHub webhook event types to align with GitHub's current list of events (within the newest version of the used library).

### Modifications
Updated the `githubMatch` function in `server/auth/webhook/github.go` to include the missing GitHub webhook event types:
- CodeScanningAlertEvent
- DependabotAlertEvent
- DeployKeyEvent
- GitHubAppAuthorizationEvent
- WorkflowDispatchEvent
- WorkflowJobEvent
- WorkflowRunEvent

Additionally, the updated webhook library (`go-playground/webhooks/v6`) now uses SHA-256 instead of SHA-1 for GitHub webhook signature validation. This required updating the test cases to use the `X-Hub-Signature-256` header with the `sha256=` prefix format instead of the previous SHA-1 format.

These changes ensure that Argo Workflows can properly authenticate and process both existing and newer event types from GitHub webhooks, while maintaining compatibility with GitHub's current security standards.

### Verification
I verified the changes by:
- Deploying the updated code to our Argo Workflows instance
- Configuring a GitHub webhook to send both release and `workflow_run` events
- Setting up WorkflowEventBindings for both event types
- Triggering both events from GitHub
- Confirming that both event types were properly authenticated and processed

Before the fix, the `workflow_run` events were being rejected with a `401` error, while `release` events were processed correctly. After the fix, both event types are properly authenticated and processed.

### Documentation
No documentation updates are required as this is a bug fix that aligns the code with expected behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/go-git/go-git/v5 v5.16.0
 	github.com/go-jose/go-jose/v3 v3.0.4
 	github.com/go-openapi/jsonreference v0.21.0
+	github.com/go-playground/webhooks/v6 v6.4.0
 	github.com/go-sql-driver/mysql v1.9.2
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/protobuf v1.5.4
--- a/go.sum
+++ b/go.sum
@@ -326,6 +326,8 @@ github.com/go-openapi/jsonreference v0.21.0 h1:Rs+Y7hSXT83Jacb7kFyjn4ijOuVGSvOdF
 github.com/go-openapi/jsonreference v0.21.0/go.mod h1:LmZmgsrTkVg9LG4EaHeY8cBDslNPMo06cago5JNLkm4=
 github.com/go-openapi/swag v0.23.1 h1:lpsStH0n2ittzTnbaSloVZLuB5+fvSY/+hnagBjSNZU=
 github.com/go-openapi/swag v0.23.1/go.mod h1:STZs8TbRvEQQKUA+JZNAm3EWlgaOBGpyFDqQnDHMef0=
+github.com/go-playground/webhooks/v6 v6.4.0 h1:KLa6y7bD19N48rxJDHM0DpE3T4grV7GxMy1b/aHMWPY=
+github.com/go-playground/webhooks/v6 v6.4.0/go.mod h1:5lBxopx+cAJiBI4+kyRbuHrEi+hYRDdRHuRR4Ya5Ums=
 github.com/go-sql-driver/mysql v1.9.2 h1:4cNKDYQ1I84SXslGddlsrMhc8k4LeDVj6Ad6WRjiHuU=
 github.com/go-sql-driver/mysql v1.9.2/go.mod h1:qn46aNg1333BRMNU69Lq93t8du/dwxI64Gl8i5p1WMU=
 github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
--- a/server/auth/webhook/github.go
+++ b/server/auth/webhook/github.go
@@ -3,7 +3,7 @@ package webhook
 import (
 	"net/http"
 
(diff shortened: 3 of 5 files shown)


Joibel:
The updated webhook library authentication change appears to be a breaking change, is this correct?

malpou:
The updated library does change from using SHA-1 signatures to SHA-256, but this shouldn't be a breaking change for actual GitHub webhook calls because GitHub sends both signature headers in their webhook requests:
```
X-Hub-Signature: sha1=6d98d96238fa69e943a39e12d832153861bxxx
X-Hub-Signature-256: sha256=fed7cd188037a3592e98464395e51577a04776bc879cd71d778c0a98d4xxx
```

The [library now looks for the X-Hub-Signature-256 header instead of X-Hub-Signature](https://github.com/go-playground/webhooks/commit/53694f85d27d5446a0790b0408b6799ac9709b98), but since GitHub provides both, real-world webhook calls should continue to function correctly.

I had to update the webhook library because it was the only way to add support for the newer GitHub event types (`workflow_run`, `workflow_dispatch`, etc.) that we need. 

The test cases needed updating to use SHA-256 signatures with the "sha256=" prefix format instead of the old SHA-1 format. 

I'm still working through the E2E test case related to the signature calculation using the "shh!" secret, but once resolved, the tests will validate that everything works as expected in real scenarios.

Joibel (approved):
Thanks for your contribution. Could you resolve the conflicts and then I can merge this.

Issue #14422, opened 2025-04-23
Service Unavailable: connection error: desc = "transport: authentication handshake failed: credentials: cannot check peer: missing selected ALPN property


yukikea:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

I started seeing `Service Unavailable: connection error: desc = "transport: authentication handshake failed: credentials: cannot check peer: missing selected ALPN property. If you upgraded from a grpc-go version earlier than 1.67, your TLS connections may have stopped working due to ALPN enforcement. For more details, see: https://github.com/grpc/grpc-go/issues/434" on the UI.
` from this Tuesday after `argo-server` restarted not sure if that's the cause though.
We have an entire stack deployed via Metaflow so basically https://github.com/outerbounds/metaflow-tools/tree/master/gcp/terraform on our GCP project.

### Version(s)

fa87072583a04f5cc278379cd16778cec3607c822cf8efd6fed2f0f6b56a5bec

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

```YAML
We can't proceed with running any workflow at the moment. The above error is on UI https://localhost:2746/
```

### Logs from the workflow controller

```text
time="2025-04-23T23:40:21.379Z" level=info msg="Processing workflow" Phase=Running ResourceVersion=474752098 namespace=argo workflow=ffpfulfilmentforecast.prod.lightgbmpredict-1745451000
time="2025-04-23T23:40:21.379Z" level=info msg="Task-result reconciliation" namespace=argo numObjs=29 workflow=ffpfulfilmentforecast.prod.lightgbmpredict-1745451000
time="2025-04-23T23:40:21.382Z" level=warning msg="was unable to obtain the node for ffpfulfilmentforecast.prod.lightgbmpredict-1745451000-26505037, taskName end"
[...]

PR #14433, opened 2025-04-29
fix: disable ALPN in argo-server as a workaround

Joibel:
### Motivation

Upgrading from 3.6.5 to 3.6.6 will probably cause you to see ALPN errors as per https://github.com/grpc/grpc-go/issues/434 and #14422.
### Modifications

This bakes the disabling flag into the docker image for this warning as a workaround for 3.6.

### Verification

Built image and substituted it into a running 3.6.6 and found the error went away

### Documentation

Added some notes to the argo server page to document this. I will also highlight this in slack.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,6 +109,8 @@ USER 8737
 
 WORKDIR /home/argo
 
+# Temporary workaround for https://github.com/grpc/grpc-go/issues/434
+ENV GRPC_ENFORCE_ALPN_ENABLED=false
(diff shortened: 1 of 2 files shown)


tico24 (approved):

terrytangyuan (approved):

terrytangyuan:
LGTM as long as we document this and provide a long-term plan

Joibel:
Longer term fix to be decided on and fixed in #14422

PR #14442, opened 2025-05-01
fix: Remove nil ConfigMap reference in cache gc

thevilledev:
### Motivation

Prevents `nil` pointer dereference when attempting to log invalid objects during ConfigMap cache garbage collection.

### Modifications

Changed the error log from `log.WithField().Error()` to `log.Error()` as there's no additional fields to be added.

### Verification

`make test` passes locally.

I've also prepared a lint rule change in #14443.

### Documentation

No documentation changes needed.


thevilledev:
Tests seem to fail due to unrelated reasons. I think this is ready for review.

MasonM (approved):
Thanks for your contribution!

Joibel (approved):
Thanks, looks good

PR #14446, opened 2025-05-02
feat!: Introduce curated Sprig function allowlist

thevilledev:
### Motivation

This PR aims to curate the available Sprig functions, exposing only those that provide unique value beyond the [expr-lang](https://github.com/expr-lang/expr) built-ins and are considered safe for use within workflow expressions. This reduces complexity, relying on expr-lang's built-in functions and memory budgeting where possible.

The [sprig library](https://github.com/Masterminds/sprig) has not received a new release since August 2024 and has many open issues, regarding Go vulnerabilities and bug fixes. This PR reduces dependencies on it, in favor of built-in expr functions.

NOTE: This is a breaking change.

### Modifications

Refactored the initialization of the Sprig function map in `util/expr/env/env.go`. Instead of exposing all functions from `sprig.GenericFuncMap()`, it now filters the full map against an allowlist during package initialization.

Defined the allowlist to include only Sprig functions that do not have direct equivalents in expr-lang and are considered safe. Potentially unsafe functions and those redundant with expr-lang built-ins are excluded.

### Verification

Added comprehensive unit tests in `util/expr/env/env_test.go` which didn't exist before, covering the new allowlist logic and existing helper functions. Test coverage of the package was 0% previously, now at 96%.

Tests explicitly verify the presence and absence of specific Sprig functions based on the defined allowlist.

Core expression evaluation relies on the expr-lang library, which has its own testing, including memory budgeting features.

### Documentation

Updated docs in `docs/variables.md`.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -152,6 +152,7 @@ codebase
 config
 cpu
 cron
+crypto
 daemoned
 dependabot
 dev
@@ -219,6 +220,7 @@ runtime
 runtimes
 s3
 sandboxed
+semver
 shortcodes
 stateful
 stderr
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -3,6 +3,12 @@
 Breaking changes  typically (sometimes we don't realise they are breaking) have "!" in the commit message, as per
 the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
 
+## Upgrading to v3.7
+
+### Deprecations
+
+Several Sprig functions have been deprecated in favor of Expr standard library alternatives. While these functions continue to work, they will be removed in a future version. See [available Sprig functions](variables.md#sprig-functions) for the complete list and [Expression language](variables.md#expression) for alternatives.
+
 ## Upgrading to v3.6
 
 See also the list of [new features in 3.6](new-features.md).
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -113,9 +113,9 @@ Extract data from JSON:
 jsonpath(inputs.parameters.json, '$.some.path')
 ```
 
-You can also use [Sprig functions](http://masterminds.github.io/sprig/):
+#### Sprig Functions
 
-Trim a string:
+You can also use a curated set of [Sprig functions](http://masterminds.github.io/sprig/):
 
 ```text
 sprig.trim(inputs.parameters['my-string-param'])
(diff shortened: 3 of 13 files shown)


blkperl:
Seems like a great first step at deprecating Sprig! Thanks for opening the PR 🌸 

Please change the whitelist references to allowlist as it doesn't align with the CNCF's inclusive language 

https://www.cncf.io/announcements/2021/10/13/inclusive-naming-initiative-announces-new-community-resources-for-a-more-inclusive-future/

thevilledev:
> Please change the whitelist references to allowlist

Thanks for the reminder! Done now.

> Are we able to have the argo linter and/or controller emit a deprecation warning for this change?

Interesting idea. This would be a totally new feature I guess? I couldn't at least find anything similar from the existing codebase, although the `util/deprecated` package could be extended to this.

Covering both `argo lint` and `argo submit` makes sense, but also anyone interfacing through `kubectl`. And warnings generated by the controller would be nice too. Sounds like a validating webhook and then some shared logic for lint/submit commands. This would be a generic solution and could be used for any workflow syntax deprecation in the future.

.spelling
```diff
@@ -212,13 +213,21 @@ pprof
 pre-commit
 pytorch
 qps
+randAlpha
```
Joibel, line 216: This block shouldn't be needed as you've backticked the words in the docs.

.spelling
```diff
@@ -227,6 +236,7 @@ temporality
 tolerations
 triaged
 un-reconciled
+uuidv4
```
Joibel, line 239: ditto: should be unnecessary

.spelling
```diff
@@ -212,13 +213,21 @@ pprof
...
+randNumeric
 rc2
 repo
 roadmap
 runtime
 runtimes
 s3
 sandboxed
+semver
+semverCompare
```
Joibel, line 230: ditto for semverCompare

Joibel (changes requested):
Please could this also be added to docs/upgrading.md under a new 3.7 heading - it can link into docs/variables.md for the list, but people need to be made aware, and that's our mechanism.

Otherwise LGTM.

.spelling
thevilledev: Ah, right. Fixed!

Joibel (approved):
LGTM, thank you

Joibel (approved):

PR #14449, opened 2025-05-05
feat: Allow overriding the argoexec binary location

mprahl:
### Motivation

The specific usecase is to allow a distribution of Argo Workflows to build one container image with a CGO_ENABLED=1 binary of argoexec and one with CGO_ENABLED=0. When the cluster must be FIPS compliant, we use the CGO_ENABLED=1 binary to use FIPS approved C libraries by setting the ARGO_EXEC_PATH environment variable. When the cluster does not require FIPS, we will just use the CGO_ENABLED=0 binary which allows much more flexibility for the container image used by the main container as the C libraries don't need to match from the build environment.

### Modifications

This introduces the ARGO_EXEC_PATH environment variable and when set, the value is used as the path to copy to /var/run/argo/argoexec.

### Verification

I deployed the workflow controller with the the `ARGO_EXEC_PATH` environment variable and looked for the new logs to ensure the right code path was being executed.

### Documentation

Documented in docs/environment-variables.md.


terrytangyuan (commented):
@Joibel Do you have any concerns about this change?

Joibel:
I don't think it's the right approach. Building a container with both binaries wouldn't satisfy some users, so they'd just build a single binary container and switch to using that for all workflows by setting `executor.image` in the workflow-controller-configmap.

If you really need to switch on a per workflow or per template basis this is already doable using a podSpecPatch which you can attempt with this (not-working because `alpine` isn't argoexec) example:

`podSpecPatch: '{"containers":[{"name":"wait", "image":"alpine"}],"initContainers":[{"name":"init", "image":"alpine","command":["echo","hello"]}]}'`

Instead of this approach in the PR using an env variable I'd support doing this through a workflow and/or template level entry which worked just like the workflow-controller-configmap `executor.image`.

I'd actually like it if this was implemented as a full `PodSpec` override just like it is there. This would then allow per template resource requests for the executor, which would be useful where specific templates use a lot of artifacts for example.

mprahl:
@Joibel did you get a chance to see my previous comment? Thanks for your help.

terrytangyuan:
Alan is on vacation and will be back next week

mprahl:
@Joibel did you get a chance to see my previous comment? Thanks for your help.

HumairAK:
bump on this, would love this feature thanks!

PR #14450, opened 2025-05-06
feat: allow last retry variables in expressions. Fixes #10364

### LINKED ISSUE #10364: Allow use of lastRetry in podSpecPatch resources request

isubasinghe:
Fixes #10364

### Motivation

This PR resurrects the older [PR](https://github.com/argoproj/argo-workflows/pull/12722) by @eduardodbr to support last retry variable in expressions. 

### Modifications 
This adds a generic function to replace expressions. My change differs somewhat to @eduardodbr's. I didn't agree with his implementation so I changed it. The actual injection is pretty much  exactly how retries get injected. 

### Verification
Verified with the test.

### Documentation
Modified the docs to add documentation on the `lastRetry` variable.


util/template/expression_template.go
```diff
@@ -40,6 +40,17 @@ func expressionReplace(w io.Writer, expression string, env map[string]interface{
 		return fmt.Fprintf(w, "{{%s%s}}", kindExpression, expression)
 	}
 
+	if (hasVariableInExpression(unmarshalledExpression, "lastRetry.exitCode") && !hasVarInEnv(env, "lastRetry.exitCode") ||
```
Joibel, line 43: Oh, this hack is horrid. But it mirrors what's happening with retries, so I guess we go with it for now.

Can we make this less evil code wise - perhaps list of all the variables we're going to do this for and iterate? Including retry?

Joibel (changes requested):

util/template/expression_template.go
isubasinghe: Sure I think we can  do this for lastRetry but I do not want to do this for retries. I think they need to be distinct.
I could maybe be refactored to share some common logic though.

eduardodbr:
Thanks for resurrecting it! 

Looks good to me.

Did you have the time to check @juliev0 suggestion? https://github.com/argoproj/argo-workflows/pull/12722#discussion_r1597789358

Joibel:
Oops, this is a feature so shouldn't be back ported. I was trying to sort out the dependencies for backporting #15033 to release-3.6

PR #14454, opened 2025-05-06
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

There is some fixes in the code as well, but eyeballing those changes they seems to be kosher... well -- tests should show, right?

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,go.sum,*.lock,.codespellrc,CHANGELOG*.md
+check-hidden = true
+ignore-regex = \S{50,}|\bList[A-Z][a-z]+\b
+ignore-words-list = notin,implementors,bui,te,alpha-numeric,freee,ans
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within .codespellrc
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
--- a/Makefile
+++ b/Makefile
@@ -818,3 +818,7 @@ release-notes: /dev/null
 .PHONY: checksums
 checksums:
 	sha256sum ./dist/argo-*.gz | awk -F './dist/' '{print $$1 $$2}' > ./dist/argo-workflows-cli-checksums.txt
+
+.PHONY: codespell
+codespell:
+	codespell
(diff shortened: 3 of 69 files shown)


Joibel:
@yarikoptic - thanks for this, it looks interesting and is something we could possibly adopt.

We'd obviously need it to pass CI. My brief take on getting that to happen:
* Sign DCO
* We'd need codespell action pinning
* We'd need codegen to pass, which probably means that the Makefile needs to learn to codespell. `make -B codegen` inside a devcontainer is the easiest way to see what this will do. It might be necessary to change which files get checked. Ideally codespell would learn not to check generated files - all ours should be detectable.
* Even if you don't need to make the Makefile know how to codespell for codegen, I'd still want a Makefile target like that so I can know my code passes before pushing.

yarikoptic:
> * Sign DCO

done

> * We'd need codespell action pinning

do you use any helper for that, e.g. smth like https://www.npmjs.com/package/pin-github-action

> * We'd need codegen to pass, 

it seems that test fails are due to interaction with other components of argo [which still carry those typos, like `conditons`](https://github.com/search?q=org%3Aargoproj%20conditons&type=code)... yet to see what the "sequence" of fixes across repos to be, please advise

> which probably means that the Makefile needs to learn to codespell. `make -B codegen` inside a devcontainer is the easiest way to see what this will do. It might be necessary to change which files get checked. Ideally codespell would learn not to check generated files - all ours should be detectable.

ATM there is no (easy/built-in) way to skip files based on content (e.g. matching "Code generated by"), could only easily skip based on matching filenames (e.g. `*_generated*`)... but also raises the question: why generated files should have typos?

> * Even if you don't need to make the Makefile know how to codespell for codegen, I'd still want a Makefile target like that so I can know my code passes before pushing.

done, there is `codespell` target now.  Should it be added to `lint` or `pre-commit` dependencies?

Joibel:
> > * Sign DCO
> 
> done

Thanks

> > * We'd need codespell action pinning
> 
> do you use any helper for that, e.g. smth like https://www.npmjs.com/package/pin-github-action

No, we just pin manually.

> > * We'd need codegen to pass,
> 
> it seems that test fails are due to interaction with other components of argo [which still carry those typos, like `conditons`](https://github.com/search?q=org%3Aargoproj%20conditons&type=code)... yet to see what the "sequence" of fixes across repos to be, please advise

This would require argo-events to be fixed, then a release made, and then us to import that new release in go.mod. (Technically the release could be skipped and we import as a SHA, but for spelling only that feels like it does more harm than good).

> > which probably means that the Makefile needs to learn to codespell. `make -B codegen` inside a devcontainer is the easiest way to see what this will do. It might be necessary to change which files get checked. Ideally codespell would learn not to check generated files - all ours should be detectable.
> 
> ATM there is no (easy/built-in) way to skip files based on content (e.g. matching "Code generated by"), could only easily skip based on matching filenames (e.g. `*_generated*`)... but also raises the question: why generated files should have typos?

If generated files have typos but they're imported typos (e.g. we're embedding a documented struct from another project as we do) we can't fix them.

The non-generated file typos should be detected and then the generated files regenerated to be fixed also.

> > * Even if you don't need to make the Makefile know how to codespell for codegen, I'd still want a Makefile target like that so I can know my code passes before pushing.
> 
> done, there is `codespell` target now. Should it be added to `lint` or `pre-commit` dependencies?

Yes, this must be in the dependencies for pre-commit.

In order that this is satisfactory it should get installed, and we do this (somewhat badly) via the Makefile, so that you can spin up a devcontainer and run `make pre-commit` and it just works - installing the tools it needs along the way.

yarikoptic:
ok, let's then be "sequential" to avoid messing with exclusions etc, and first get 
- https://github.com/argoproj/argo-events/pull/3584 

done

PR #14457, opened 2025-05-07
feat!: don't push to docker hub

### LINKED ISSUE #13968: proposal: stop publishing container images to docker hub
# Summary

None of the rest of argoproj publish containers to docker hub, so lets stop doing that for workflows.

The most recent non-workflows images are from [Argo CD which stopped publishing there 2 and a bit years ago](https://github.com/argoproj/argo-cd/pull/8238)

This reduces the number of keys we need to maintain and the amount of clean-up necessary if I mess up releasing again.

Joibel:
Fixes #13968

### Motivation

None of the rest of argoproj pushes to docker hub any more. With their rate limiting it makes it hard to use.

We have pushed to quay.io for years, so make that the only place we push to

### Modifications

Remove all the actions which push to docker hub.

Fix up a few docs which would attempt to pull from docker hub not quay.io

This is not a 100% purge of images in docker hub, there are still some references to things that don't mention a registry.

There is no intention to delete any old images, just to stop new ones going there.

I'm intending to backport this to 3.6, so comments around whether that is acceptable to reviewers would be helpful. This will allow me to test the release changes.

I have retained docker login within the codebase as our Dockerfile depends on images coming from dockerhub, and logging in should remove rate limiting.

I haven't renamed `DOCKERIO_ORG` (which == `argoproj`), just to reduce likelyhood of things going wrong. I'm unsure why this comes from a secret at all.

### Verification

There really has to be a release for the bulk of it to be tested.

### Documentation

None here, should be mentioned in the release notes. The image sources aren't mentioned in the current docs.

The official helm chart already pulls from quay, as do our manifests.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -89,7 +89,6 @@ jobs:
             --platform="${PLATFORM}" \
             --target $TARGET \
             --provenance=false \
-            --tag $image_name \
             --tag quay.io/$image_name .
 
   build-windows:
@@ -133,8 +132,6 @@ jobs:
               -t $image_name \
               -f Dockerfile.windows \
               .
-
-            docker push $image_name
 
             docker tag $image_name quay.io/$image_name
             docker push quay.io/$image_name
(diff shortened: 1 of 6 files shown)


tico24 (approved):
Docs LGTM

PR #14459, opened 2025-05-08
fix: namespace permissions for per-namespace parallelism. Fixes #14460

### LINKED ISSUE #14460: per-namespace parallelism doesn't have the correct perms
### Pre-requisites

N/A

### Version(s)

latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #14460

### Motivation

This PR fixes an issue with the missing permissions that was introduced in https://github.com/argoproj/argo-workflows/pull/14188. 

### Modifications

This change edits the permissions such that we add the get/watch/list verbs for namespace resources

### Verification

Ran locally, but no tests were added. 

### Documentation

No documentation needed


manifests/namespace-install/workflow-controller-rbac/workflow-controller-role.yaml
```diff
@@ -28,6 +28,7 @@ rules:
       - ""
     resources:
       - configmaps
+      - namespaces
```
Joibel, line 31: I don't think we should be doing this for namespaced installs, it feels like they wouldn't use this feature.

workflow/controller/ns_watcher.go
```diff
@@ -31,6 +32,10 @@ type resetFunc = func(string)
 
 func (wfc *WorkflowController) newNamespaceInformer(ctx context.Context, kubeclientset kubernetes.Interface) (cache.SharedIndexInformer, error) {
 
+	can, _ := authutil.CanI(ctx, wfc.kubeclientset, []string{"get", "watch", "list"}, "", metav1.NamespaceAll, "namespaces")
+	if !can {
+		logrus.Error("was unable to get permissions for get/watch/list verbs on the namespace resource, per-namespace parallelism will not work")
```
Joibel, line 37: Make this a warning, it's not strictly an error

Joibel (changes requested):
I think some documentation in around this RBAC being optional would also be useful - https://argo-workflows.readthedocs.io/en/latest/parallelism/

Make it clear that you must grant this RBAC for the feature to work, but it's fine not to grant RBAC if you don't want the feature.

manifests/namespace-install/workflow-controller-rbac/workflow-controller-role.yaml
isubasinghe: Yes they won't use this feature, but they might still expect the dynamic limits to work? I think that is a fair expectation.

isubasinghe: Addressed

Joibel (approved):

dyatlov:
If the feature required only `get/list/watch` then why did you give it also `create/update/delete/patch` permissions?
This is absolutely not needed, way too wide, and requires us to now manually patch the ClusterRole to limit potential blast radius.

PR #14473, opened 2025-05-17
fix: Respect NameFilter in Workflow Archive Fixes #14069

### LINKED ISSUE #14069: Argo UI cannot filter archived workflow with name filter
### Pre-requisites

In version v3.6.2, argo merged live workflow and archived workflow into to same page
However, we noticed unexpected behavior in the name filter functionality.

e.g.
we have 5 completed workflows in our env
allwf

but we can only filter the live workflows using the name filter
截圖 2025-01-09 中午12 32 17

it appears that the archived workflows can only be queried with an exact name match
截圖 2025-01-09 中午12 30 14

### Version(s)

v3.6.2

```YAML
N/A
```

```text
```

```text
```

arminfriedl:
Fixes #14069

Related: 
- PR #13160: Implementation of the NameFilter for non-archived workflows
- Issue #12161: Contains some background about how this bug came to be, especially https://github.com/argoproj/argo-workflows/issues/12161#issuecomment-2155272711
- PR #11121: Where the unified UI was implemented, ultimately causing the NameFilter behavior for Archived and Non-Archived Workflows to be inconsistent

### Motivation

Archived Workflows did not respect the NameFilter (Exact/Prefix/Contains) predicate. Now they do :tada: 

### Modifications

Generate query for Archived Workflows with a Name condition depending on the NameFilter predicate Exact/Prefix/Contains). If no NameFilter is present falls back to the previous behavior of using an exact match.

### Verification

Added e2e tests and manually via UI

### Documentation

I don't think this is specifically mentioned anywhere. Users will discover it by using the Name filter in the UI and it working like one would expect, also for Archived Workflows.


test/e2e/argo_server_test.go
```diff
@@ -942,6 +942,216 @@ func (s *ArgoServerSuite) TestWorkflowService() {
...
+			String().IsEqual(uidAliceWf)
+
+		j.Path("$.items[1].metadata.uid").
+			String().IsEqual(uidBobWf)
+	})
+
+	s.Run("Cleanup", func() {
+		s.e().DELETE("/api/v1/archived-workflows/{uid}", uidBobWf)
+		s.e().DELETE("/api/v1/archived-workflows/{uid}", uidAliceWf)
+	})
```
MasonM: This isn't doing anything, since `httpexpect` doesn't send a request unless you chain an `.Expect()` to this. But you don't need to do explicit cleanup, because that's handled automatically here: https://github.com/argoproj/argo-workflows/blob/a5d43eb2025fffe86817d23b8cf488596d7ca0d8/test/e2e/fixtures/e2e_suite.go#L177
```suggestion
```

test/e2e/argo_server_test.go
```diff
@@ -942,6 +942,216 @@ func (s *ArgoServerSuite) TestWorkflowService() {
...
+
+		j.Path("$.items").
+			Array().Length().IsEqual(2)
+
+		j.Path("$.items[0].metadata.uid").
+			String().IsEqual(uidAliceWf)
+
+		j.Path("$.items[1].metadata.uid").
+			String().IsEqual(uidBobWf)
+	})
```
MasonM: This seems possibly flaky, because it's possible that the Alice workflow starts before Bob. You can fix that using `IsEqualUnordered()` here, which is also simpler:
```suggestion
	s.Run("ListNameContainsTest", func() {
		s.e().GET("/api/v1/workflows/argo").
			WithQuery("listOptions.fieldSelector", "metadata.name=%test").
			WithQuery("nameFilter", "Contains").
			Expect().
			Status(200).
			JSON().
			Path("$.items[*].metadata.uid").
			Array().
			IsEqualUnordered([]interface{}{uidAliceWf, uidBobWf})
	})
```

test/e2e/argo_server_test.go
```diff
@@ -942,6 +942,216 @@ func (s *ArgoServerSuite) TestWorkflowService() {
...
+			Expect().
+			Status(200).
+			JSON()
+
+		j.Path("$.items").
+			Array().Length().IsEqual(1)
+
+		j.Path("$.items[0].metadata.uid").
+			String().IsEqual(uidBobWf)
+
```
MasonM: Minor: this (and every other test in this suite) could be simplified a bit using `IsEqualUnordered()`, like I suggested above:
```suggestion
		s.e().GET("/api/v1/workflows/argo").
			WithQuery("listOptions.fieldSelector", "metadata.name="+nameBobWf).
			Expect().
			Status(200).
			JSON().
			Path("$.items[*].metadata.uid").
			Array().
			IsEqualUnordered([]interface{}{uidBobWf})
```

test/e2e/argo_server_test.go
arminfriedl: Thanks, removed it

test/e2e/argo_server_test.go
arminfriedl: Thanks, changed it to `IsEqualUnordered` - this indeed also reflects the intention better, this is not supposed to test any specific order.

test/e2e/argo_server_test.go
arminfriedl: That's right, I like this simplification - changed

arminfriedl:
> Thanks for this! Could you update `pkg/apiclient/workflowarchive/workflow-archive.proto` spec to include the `nameFilter` parameter so the API clients are updated accordingly? Should be nearly the same change as was done in https://github.com/argoproj/argo-workflows/pull/13160/files#diff-1f6bb75f9e833cb6f66ada3c22cd4218d7776fbfad7ef40583d9a3c45534059aR36

Added the `nameFilter` parameter also to `/archived-workflows` now. I hope I did the necessary codegen correctly.

I think there is some potential for confusion with the already existing `namePrefix` parameter there. I didn't remove the `namePrefix` as this would mean a breaking API change, but it is in some sense superfluous now.

With this new parameter I think this is not just a fix anymore - changed the title accordingly and added a description in .feature too.

MasonM (approved):
Good job @arminfriedl!

tczhao (approved):

PR #14477, opened 2025-05-19
feat: non-root argoexec

Joibel:
As discussed at Feb 17th 2025 contributor meeting

### Motivation

Security policies for some companies do not like having images where the user is `0` or root.
`argoexec` defaults to running as root, but changing this is likely to break a lot of users workflows.

### Modifications

Supplement this with an `argoexec-nonroot` image which runs as the user 8737, the same user as the rest of the images.

Also update the documentation around this, and remove references executor choices.

### Verification

Built the image using the Makefile, and ran it in production for a week at pipekit.

### Documentation

Update the executor documentation to match the style guide better, and remove references to choices around the executor. I've retained the `Features` list mostly even though it feels a bit redundant.

Update the workflow pod security context documentation to prefer using this as best practice and encourage good pod security practice overall

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -264,6 +264,7 @@ else
 endif
 
 argoexec-image:
+argoexec-nonroot-image:
 
 %-image:
 	[ ! -e dist/$* ] || mv dist/$* .
--- a/docs/workflow-executors.md
+++ b/docs/workflow-executors.md
@@ -1,8 +1,8 @@
 # Workflow Executors
 
-A workflow executor runs in the pods that execute your workloads.
+A Workflow executor runs in the Pods that execute your workloads.
 It runs as both an init container and as a sidecar to the container you specify.
-It allows Argo to perform certain actions like monitoring pod logs, providing and collecting artifacts, and managing container life-cycles.
+It allows Argo to perform certain actions like monitoring Pod logs, providing and collecting artifacts, and managing container life-cycles.
 
 Historically there were multiple available executor types, but as of 3.4 the only available executor is called `emissary`.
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -264,6 +264,7 @@ else
 endif
 
 argoexec-image:
+argoexec-nonroot-image:
```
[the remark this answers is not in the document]
Joibel: The other similar targets don't do this, so not changing it here.

docs/workflow-executors.md
```diff
@@ -1,32 +1,29 @@
...
-    * `command` should be specified for containers.
+    * Output artifacts can be located on the base layer (e.g. `/tmp`)
+
+### Container Command
 
-You can determine values as follows:
+You can determine the default command for a container image using:
 
 ```bash
 docker image inspect -f '{{.Config.Entrypoint}} {{.Config.Cmd}}' argoproj/argosay:v2
```
tico24, line 29: This image doesn't exist any more does it? I can't run the command and get a useful answer from docker anyway.

tico24 (changes requested):
docs review only: I couldn't repeat the 'how to find the command' script.

If I'm being picky, you're missing some caps on proper nouns and k8s things like Pods.

tico24 (approved):

isubasinghe (approved):

docs/workflow-executors.md
```diff
@@ -1,46 +1,51 @@
 # Workflow Executors
 
-A workflow executor is a process that conforms to a specific interface that allows Argo to perform certain actions like monitoring pod logs, collecting artifacts, managing container life-cycles, etc.
+A Workflow executor runs in the Pods that execute your workloads.
+It runs as both an init container and as a sidecar to the container you specify.
+It allows Argo to perform certain actions like monitoring Pod logs, providing and collecting artifacts, and managing container life-cycles.
 
-## Emissary (emissary)
+Historically there were multiple available executor types, but as of 3.4 the only available executor is called `emissary`.
```
terrytangyuan, line 7: I wonder if we should link to an old commit for the relevant doc in case users are interested

PR #14494, opened 2025-05-23
fix: populate optional artifacts with empty data

isubasinghe:
Fixes #TODO


isubasinghe:
I was wondering if there should also be a `exists` variable as well such that the expression logic can deal with missing output artifacts with more precision.

workflow/controller/dag.go
```diff
@@ -657,7 +657,7 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
 func (woc *wfOperationCtx) buildLocalScopeFromTask(dagCtx *dagContext, task *wfv1.DAGTask) (*wfScope, error) {
 	// build up the scope
 	scope := createScope(dagCtx.tmpl)
-	woc.addOutputsToLocalScope("workflow", woc.wf.Status.Outputs, scope)
+	woc.addOutputsToLocalScope("workflow", woc.wf.Status.Outputs, scope, nil)
```
Joibel, line 660: nit (edit): wot copilot said

workflow/controller/operator.go
```diff
@@ -3222,12 +3223,22 @@ func (woc *wfOperationCtx) requeueIfTransientErr(err error, nodeName string) (*w
 // buildLocalScope adds all of a nodes outputs to the local scope with the given prefix, as well
 // as the global scope, if specified with a globalName
 func (woc *wfOperationCtx) buildLocalScope(scope *wfScope, prefix string, node *wfv1.NodeStatus) {
+
```
Joibel, line 3226: nit: Remove leading new line

Joibel (changes requested):
This whole thing still feels like it's the wrong fix. It's fixing outputs, which are fine, to mask a problem in inputs, which are broken.

We have a problem with input artifacts, marked as optional, causing the processing of those to cause an error. They're optional, so they shouldn't cause an error, and this doesn't fix that.

This feels like a hack, undoing some of #13678 which made sure that optional output artifacts which didn't exist really didn't exist.

You've suggested a flag `exists` on the artifact rather than an empty value as you've implemented here - but that's pretty much already determinable with the code prior to this - it just doesn't exist.

If you'd implemented `exists` where would you check and use that information - that's where I think this fix should be applied?

We're missing the test case as a regression test too.

Joibel:
Replaced by #14521 now? Close?

PR #14507, opened 2025-05-27
ci: disable flaky test

isubasinghe:
We are wasting money/time/cpu on a single flaky test, we should disable it until the flakiness has been resolved.


Joibel (approved):

isubasinghe:
> Not super keen on just ignoring it, are you going to investigate a proper fix?

I agree we should investigate a proper fix. I would like to.

PR #14515, opened 2025-05-27
docs: auto-generate wf-controller-configmap docs

Joibel:
### Motivation

The workflow controller configmap is documented by an example .yaml file only.

Instead, following fields.md, generate a file documenting it from the comments

### Modifications

Added a new generator to hack/docs to generate this file. It can't use the swagger process that `fields.md` uses so instead parse the .go and use the resulting AST to extract the details. We could potentially do this for fields.md also, but I didn't try.

Added one more thing to fields.md (`InfoResponse`) as we're linking into that for fields that are in our public interface, and these aren't documented there.

Added comments throughout config for fields and structs which are missing them.

I'm not sure if this should have been a separate (new) .md file, but I've bunged it on to the end of the existing one. To edit the beginning you'd need to change the hack/doc/workflow-controller-configmap.md file, which isn't super intuitive. Comments welcome about what's best here.

### Verification

Looked at the resulting documentation locally.

### Documentation

This is a docs commit, so probably doesn't need docs!


docs/workflow-controller-configmap.md
```diff
@@ -60,3 +60,288 @@ data:                      # "config: |" key is optional in 2.7+!
...
+## NodeEvents
+
+NodeEvents configures how node events are emitted
+
+### Fields
+
+| Field Name | Field Type | Description |
+|:----------:|:----------:|:------------|
+| `Enabled` | `bool` | Enabled controls whether node events are emitted |
+| `SendAsPod` | `bool` | SendAsPod TODO: describe what this field does |
```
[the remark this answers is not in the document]
Joibel: Done, I just hadn't regenerated

hack/docs/workflow-controller-configmap.md
```diff
@@ -0,0 +1,62 @@
+# Workflow Controller Config Map
```
tico24: for consistency, can any instance of `Config Map` be `ConfigMap`?

Joibel: This is just a moved file, but yes.

hack/docs/configdoc.go
```diff
@@ -0,0 +1,392 @@
...
+	"strings"
+)
+
+// Set this to the root of the repo
+const configDir = "config"
+
+//go:embed workflow-controller-configmap.md
+var header string
+
+const outputFile = "docs/workflow-controller-configmap.md"
```
isubasinghe, line 24: hahah love this

hack/docs/configdoc.go
```diff
@@ -0,0 +1,392 @@
...
+	buf.WriteString(fmt.Sprintf("\n## %s\n\n", name))
+	// Check for comment from GenDecl first, then TypeSpec
+	if comment, ok := typeComments[name]; ok && comment != nil {
+		buf.WriteString(normalizeComment(comment.Text()) + "\n\n")
+	} else if ts.Doc != nil {
+		buf.WriteString(normalizeComment(ts.Doc.Text()) + "\n\n")
+	}
+	buf.WriteString("### Fields\n\n")
+	buf.WriteString("| Field Name | Field Type | Description |\n")
+	buf.WriteString("|:----------:|:----------:|:------------|\n")
```
isubasinghe: Fine I suppose, but I thought we maybe had some library to better manipulate markdown?

isubasinghe (approved):

hack/docs/configdoc.go
Joibel: Good point. I remember checking that library and rejecting it, but looking again now it feels right. I'll update

Joibel: Done

PR #14519, opened 2025-05-27
chore(deps)!: bump k8s dependencies to 1.33

blkperl:
### Motivation

adds support for K8s 1.33, potentially drops support for K8s 1.30 which is EOL

### Modifications

updated go.mod and makefile with new k8s, controller-runtime version and golangcilint version

### Verification

e2e

### Documentation

No docs

PR #14525, opened 2025-05-28
ci: add support for k8s 1.33 and drop 1.30

blkperl:
Fixes #TODO

### Motivation

* Add testing for k8s 1.33 and drop support for EOL 1.30

### Modifications

* updated the min and max k8s versions

### Verification

e2e will verify

### Documentation

docs will be updated automatically


Joibel (approved):
1.30 doesn't go EOL [until 28th June](https://kubernetes.io/releases/#release-v1-30) so we technically have a month to go, but it feels unlikely we get 3.7 out before then.

Will wait to merge until after #14519

PR #14527, opened 2025-05-30
feat: logging refactor to `slog`. Fixes #11120

### LINKED ISSUE #11120: Prune unmaintained dependencies

isubasinghe:
Fixes #11120  (partially do not close this one out)

### Motivation
Logrus is in maintenance mode, but more importantly this is about making logging better and more obvious about where the origin of log message is from. This mostly achieves that.

### Modifications
This consists of a few changes:
* Introduction of a logger interface so that we may add other logging backends far more  easily.
* Migrated to using `slog` from the golang standard library

### Verification

This  is all logging changes so no verification needed really.

### DIFF
--- a/.features/pending/feat-logging.md
+++ b/.features/pending/feat-logging.md
@@ -1,0 +1,9 @@
+Component: General
+Issues: 11120
+Description: This migrates most of the logging off logrus and onto a custom logger.
+Author: [Isitha Subasinghe](https://github.com/isubasinghe)
+
+Currently it is quite hard to identify log lines with it's corresponding workflow.
+This change propagates a context object down the call hierarchy containing an 
+annotated logging object. This allows context aware logging from deep within the
+codebase.
--- a/cmd/argo/commands/archive/list_label_values.go
+++ b/cmd/argo/commands/archive/list_label_values.go
@@ -1,7 +1,10 @@
 package archive
 
 import (
+	"context"
 	"fmt"
+
+	"github.com/argoproj/argo-workflows/v3/util/logging"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -39,11 +42,17 @@ func NewListLabelValueCommand() *cobra.Command {
 			for _, str := range labels.Items {
 				fmt.Printf("%s\n", str)
 			}
+
 			return nil
 		},
 	}
+	ctx := command.Context()
+	if ctx != nil {
+		ctx = logging.WithLogger(context.Background(), logging.NewSlogLogger(logging.GetGlobalLevel(), logging.GetGlobalFormat()))
+		command.SetContext(ctx)
+	}
 	command.Flags().StringVarP(&selector, "selector", "l", "", "Selector (label query) to query on, allows 1 value (e.g. -l key1)")
 	err := command.MarkFlagRequired("selector")
-	errors.CheckError(err)
+	errors.CheckError(ctx, err)
 	return command
 }
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -73,6 +73,11 @@ func NewRootCommand() *cobra.Command {
 			defer runtimeutil.HandleCrashWithContext(c.Context(), runtimeutil.PanicHandlers...)
 
 			log := logging.NewSlogLogger(logging.GetGlobalLevel(), logging.GetGlobalFormat())
+			if c.Context() == nil {
+				ctx := context.Background()
+				ctx = logging.WithLogger(ctx, log)
+				c.SetContext(ctx)
+			}
 
 			cmdutil.SetLogLevel(logLevel)
 			cmdutil.SetGLogLevel(glogLevel)
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -215,6 +215,13 @@ func NewRootCommand() *cobra.Command {
 	command.Flags().BoolVar(&namespaced, "namespaced", false, "run workflow-controller as namespaced mode")
 	command.Flags().StringVar(&managedNamespace, "managed-namespace", "", "namespace that workflow-controller watches, default to the installation namespace")
 	command.Flags().BoolVar(&executorPlugins, "executor-plugins", false, "enable executor plugins")
+	ctx := command.Context()
+	if ctx == nil {
+		ctx = context.Background()
+		log := logging.NewSlogLogger(logging.GetGlobalLevel(), logging.GetGlobalFormat())
+		ctx = logging.WithLogger(ctx, log)
+		command.SetContext(ctx)
+	}
 
 	// set-up env vars for the CLI such that ARGO_* env vars can be used instead of flags
 	viper.AutomaticEnv()
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 	"strings"
@@ -10,7 +11,6 @@ import (
 	"time"
 
 	"github.com/argoproj/pkg/stats"
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
@@ -30,6 +30,7 @@ import (
 	cmdutil "github.com/argoproj/argo-workflows/v3/util/cmd"
 	"github.com/argoproj/argo-workflows/v3/util/env"
 	kubecli "github.com/argoproj/argo-workflows/v3/util/kube/cli"
+	"github.com/argoproj/argo-workflows/v3/util/logging"
 	"github.com/argoproj/argo-workflows/v3/util/logs"
 	pprofutil "github.com/argoproj/argo-workflows/v3/util/pprof"
 	"github.com/argoproj/argo-workflows/v3/workflow/common"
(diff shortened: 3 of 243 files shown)


isubasinghe:
@Joibel I may need you to do a `make codegen` since that command doesn't work for me. 

EDIT: Lets see how the devcontainer's `make codegen` works.

cmd/argo/commands/archive/list_label_values.go
```diff
@@ -39,11 +40,12 @@ func NewListLabelValueCommand() *cobra.Command {
...
 				fmt.Printf("%s\n", str)
 			}
+
 			return nil
 		},
 	}
 	command.Flags().StringVarP(&selector, "selector", "l", "", "Selector (label query) to query on, allows 1 value (e.g. -l key1)")
 	err := command.MarkFlagRequired("selector")
-	errors.CheckError(err)
+	errors.CheckError(context.Background(), err)
```
Joibel: command.Context()?

cmd/workflow-controller/main.go
```diff
@@ -233,10 +237,10 @@ func init() {
 }
 
 func initConfig() {
```
Joibel: Is this relevant any more. Otherwise remove?

util/logging/slog.go
```diff
@@ -0,0 +1,317 @@
...
+		return slog.LevelInfo
+	case Panic:
+		return slog.LevelError
+	default:
+		return slog.LevelInfo
+	}
+}
+
+// NewSlogLogger returns a slog based logger
+func NewSlogLogger(logLevel Level, format LogType) Logger {
```
Joibel: nit: Was expecting to find this at the top of the file - it's convention.

util/logging/slog.go
```diff
@@ -0,0 +1,317 @@
...
+	panic(msg)
+}
+
+func (s *slogLogger) Panicf(ctx context.Context, format string, args ...interface{}) {
+	msg := fmt.Sprintf(format, args...)
+	s.Panic(ctx, msg)
+}
+
+func (s *slogLogger) AddHook(hook Hook) {
+	if hook == nil {
```
Joibel: The cost of this being dynamic is very large - all the read locks. I suggest you MUST add hooks at logger creation time, which covers the current use cases to mitigate this. Avoids all the hook copies too.

util/logging/logging.go
```diff
@@ -0,0 +1,165 @@
...
+// Hook is used to tap into the log
+type Hook interface {
+	Levels() []Level
+	Fire(level Level, msg string)
+}
+
+// Logger exports a logging interface
+type Logger interface {
+	WithFields(ctx context.Context, fields Fields) Logger
+	WithField(ctx context.Context, name string, value interface{}) Logger
```
Joibel: nit: replace `interface{}` with `any` everywhere

util/logging/logging.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Hook is used to tap into the log
+type Hook interface {
+	Levels() []Level
+	Fire(level Level, msg string)
+}
+
+// Logger exports a logging interface
+type Logger interface {
```
Joibel, line 128: I suggest a future pass on this stuff reduces the number of levels here

util/logging/slog.go
```diff
@@ -0,0 +1,317 @@
...
+
+type slogLogger struct {
+	fields Fields
+	logger *slog.Logger
+	hooks  map[Level][]Hook
+	mu     sync.RWMutex
+}
+
+var (
+	defaultLogger = NewSlogLogger(Info, Text)
```
Joibel: Suggest `defaultLogger = NewSlogLogger(globalLevel, globalFormat)` to reduce duplication of these fields.

util/logging/logging.go
```diff
@@ -0,0 +1,165 @@
...
+	LoggerKey CtxKey = "logger"
+)
+
+var (
+	globalLevel  = Info
+	globalFormat = Text
+)
+
+// SetGlobalLevel sets the global log level
+func SetGlobalLevel(level Level) {
```
Joibel, line 36: Yuck. I probably would have failed in the urge to get rid of this and replaced it with a simpler system. The logging module shouldn't be tracking this stuff, and we never use the facility to change it on the fly.

util/logging/test_hook.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Reset clears all captured entries
+func (h *TestHook) Reset() {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.entries = make([]TestEntry, 0)
+}
+
+// NewTestLogger creates a logger that doesn't output to stdout for testing
```
Joibel, line 68: Possibly make some common func for this and the NewSlogLogger, as it's mostly the same

util/logging/logging.go
```diff
@@ -0,0 +1,165 @@
...
+
+	Panic(ctx context.Context, msg string)
+	Panicf(ctx context.Context, format string, args ...interface{})
+
+	AddHook(hook Hook)
+}
+
+// GetLoggerFromContext returns a logger from context or the default logger if none is found
+func GetLoggerFromContext(ctx context.Context) Logger {
+	if ctx == nil {
```
Joibel: This feels like the results might be surprising if things go wrong, and we won't know until too late. Why do we allow `nil` contexts and for the context not to have a value in it?

util/logging/slog.go
```diff
@@ -0,0 +1,317 @@
...
+)
+
+// DefaultSlogLogger returns the default logger
+func DefaultSlogLogger() Logger {
+	lock.Lock()
+	defer lock.Unlock()
+	return defaultLogger
+}
+
+func (s *slogLogger) WithFields(_ context.Context, fields Fields) Logger {
```
Joibel: Why require an unused context here?

util/logging/slog.go
```diff
@@ -0,0 +1,317 @@
...
+
+	return &slogLogger{
+		fields: newFields,
+		logger: logger,
+		hooks:  newHooks,
+		mu:     sync.RWMutex{},
+	}
+}
+
+func (s *slogLogger) WithField(_ context.Context, name string, value interface{}) Logger {
```
Joibel: Why require an unused context here?

Joibel (changes requested):

cmd/workflow-controller/main.go
isubasinghe: Not really, but left it around so the config options for logrus were still around, but removed.

cmd/argo/commands/archive/list_label_values.go
isubasinghe: Well it will be nil here. But changed it to be populated first.

util/logging/slog.go
isubasinghe: Addressed

util/logging/slog.go
isubasinghe: This is mostly for the interface, I added a context just in case another logging library needed it for whatever reason.

util/logging/slog.go
isubasinghe: same as above ^

util/logging/slog.go
isubasinghe: Sure that works for me.

util/logging/logging.go
isubasinghe: Done

util/logging/logging.go
isubasinghe: Yeah I agree

util/logging/logging.go
isubasinghe: This isn't trivial to get rid of, it requires a large refactor, even bigger than this.

util/logging/slog.go
isubasinghe: Sure

util/logging/test_hook.go
isubasinghe: Done

util/logging/logging.go
isubasinghe: Addressed

util/logging/slog.go
isubasinghe: Addressed

util/logging/logging.go
Joibel: Happy with that

Joibel (approved):
Happy with this, thanks for the epic effort

isubasinghe:
@Joibel I made a small change after your approval: https://github.com/argoproj/argo-workflows/pull/14527/commits/532e7cc352030ad0fff51358d54d4b5062361efa

Logrus isn't completely out, so we set the level and format for it still. 
We keep a single source of truth in the field information instead of having it duplicated in the slog logger and in the abstraction. 

I shall merge this Monday morning unless you aren't okay with this.

Joibel:
> I shall merge this Monday morning unless you aren't okay with this. 

Can it wait until 3.7.0 is out, so it's not in RC4 if that needs to happen?

isubasinghe:
@Joibel some features have already been merged onto main, I will need to cherry-pick anyway for a fix only RC4. 
But more than that, I really don't want to rebase this again. It's just a lot of work.

.features/pending/feat-logging.md
```diff
@@ -0,0 +1,9 @@
+Component: General
```
terrytangyuan, line 1: What is this file for?

terrytangyuan: and it should be under approved instead of pending?

Joibel: This is for #14491  - it provides us with a documented list of features ready for feature releases - so 3.8/4.0
will be easier.

It's under `pending` as it's pending release - if that's confusing it could be renamed. It would get moved into a different directory in the release process - see #14491's documentation.

terrytangyuan: Got my answer offline.

PR #14536, opened 2025-06-04
fix: ensure task results sync when calling fullfilled.

isubasinghe:
Fixes #TODO


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2370,8 +2401,12 @@ func (n *NodeStatus) IsWorkflowStep() bool {
 }
 
 // Fulfilled returns whether a phase is fulfilled, i.e. it completed execution or was skipped or omitted
-func (phase NodePhase) Fulfilled() bool {
-	return phase.Completed() || phase == NodeSkipped || phase == NodeOmitted
+func (phase NodePhase) Fulfilled(synced *bool) bool {
```
isubasinghe, line 2424: Note that Fulfilled is a method on both NodePhase and NodeStatus, they are actually semantically different, this is almost certainly a subtle source of bugs, we should refactor this into one function in a different PR.

workflow/controller/operator.go
```diff
@@ -2720,8 +2750,14 @@ func (woc *wfOperationCtx) markNodePhase(nodeName string, phase wfv1.NodePhase,
 		woc.log.Warningf("workflow '%s' node '%s' uninitialized when marking as %v: %s", woc.wf.Name, nodeName, phase, message)
 		node = &wfv1.NodeStatus{}
 	}
+	// if we not in a running state (not expecting task results)
+	// and transition into a state that ensures we will never run mark the task results synced
+	if node.Phase != wfv1.NodeRunning && phase.FailedOrError() && node.TaskResultSynced != nil {
+		tmp := true
+		node.TaskResultSynced = &tmp
+	}
```
isubasinghe, line 2755: Maybe this is bad and should be handled by taskResultReconciliation instead? But arguably it does make sense here as well.

isubasinghe:
There exists a spec outline of the previous logic here: https://github.com/isubasinghe/argo-tlaplus in PlusCal
This hasn't been updated with the current fix (mostly because I am unsure how to represent a timeout at first glance).

I suppose if in the evicted ghost state we could transition to  a "detected evicted" state? Hmm.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2355,6 +2374,18 @@ type NodeStatus struct {
 
 	// SynchronizationStatus is the synchronization status of the node
 	SynchronizationStatus *NodeSynchronizationStatus `json:"synchronizationStatus,omitempty" protobuf:"bytes,25,opt,name=synchronizationStatus"`
+
+	// TaskResultSynced is used to determine if the node's output has been received
+	TaskResultSynced *bool `json:"taskResultSynced,omitempty" protobuf:"bytes,28,opt,name=taskResultSynced"`
```
JPZ13, line 2399: Why make this a pointer to a boolean rather than just a boolean? Reading the PR, I don't see where having the extra state of the `nil` condition gets you that much to where it justifies the extra complexity of doing the nil checks and setting the references

workflow/controller/operator.go
```diff
@@ -101,7 +101,7 @@ type wfOperationCtx struct {
 	eventRecorder    record.EventRecorder
 	// preExecutionNodePhases contains the phases of all the nodes before the current operation. Necessary to infer
 	// changes in phase for metric emission
-	preExecutionNodePhases map[string]wfv1.NodePhase
+	preExecutionNodePhases map[string]wfv1.NodeStatus
```
JPZ13: Is the plan to consolidate the `NodeStatus` and `NodePhase` types into `NodeStatus`? Also, it's a nit, but it would be good to update `preExecutionNodePhases` to `preExecutionNodeStatuses` along with the associated comment

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: This is for backwards compatibility reasons, this field isn't really a boolean, its trilean (?) I guess. 
Existence of pointer shows that this node was created after PR and uses it. 

So yes, there are three states encoded in this boolean.

workflow/controller/operator.go
isubasinghe: Well the nodestatus does contain the node phase inside them, but I agree I should rename this. its mostly used for phases so thats why I didn't bother.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2370,8 +2401,12 @@ func (n *NodeStatus) IsWorkflowStep() bool {
...
 
 // Fulfilled returns whether a phase is fulfilled, i.e. it completed execution or was skipped or omitted
-func (phase NodePhase) Fulfilled() bool {
-	return phase.Completed() || phase == NodeSkipped || phase == NodeOmitted
+func (phase NodePhase) Fulfilled(synced *bool) bool {
+	if synced == nil {
+		tmp := true
+		synced = &tmp
+	}
+	return phase.Completed() && *synced || phase == NodeSkipped || phase == NodeOmitted
```
Joibel: I found this hard to parse. Consider
```golang
	isSynced := synced == nil || *synced
	return (phase.Completed() && isSynced) || phase == NodeSkipped || phase == NodeOmitted
```

docs/environment-variables.md
```diff
@@ -45,7 +45,8 @@ This document outlines environment variables that can be used to customize behav
...
+| `RECENTLY_DELETED_POD_DURATION`          | `time.Duration`     | `2m`                                                                                       | The duration of a pod before the pod is considered to be recently deleted.                                                                                                                                                                                                |
+| `TASK_RESULT_TIMEOUT_DURATION`           | `time.Duration`     | `10m`                                                                                       | The duration of time before a node is marked completed but the task result has not arrived yet. Prefer to use this over `RECENTLY_DELETED_POD_DURATION` since this accounts for pods that are not removed but haven't produced `taskresults`                             |
```
Joibel: As a user how do I "use this" over the alternative. They both appear to be here and have defaults so both are "used".

Do we need to keep both of these mechanisms, or both of these timeouts. Couldn't they just be the same timeout? Can we remove RECENTLY_DELETED_POD_DURATION as a mechanism completely?

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2008,13 +2009,31 @@ func (in *WorkflowStatus) MarkTaskResultIncomplete(name string) {
 		in.TaskResultsCompletionStatus = make(map[string]bool)
 	}
 	in.TaskResultsCompletionStatus[name] = false
+	node, err := in.Nodes.Get(name)
+	if err != nil {
+		return
+	}
+	if node.TaskResultSynced != nil {
+		tmp := false
+		node.TaskResultSynced = &tmp
```
Joibel: Use `k8s.io/utils/ptr` instead: ptr.To(bool(false))

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2008,13 +2009,31 @@ func (in *WorkflowStatus) MarkTaskResultIncomplete(name string) {
...
 		in.TaskResultsCompletionStatus = make(map[string]bool)
 	}
 	in.TaskResultsCompletionStatus[name] = true
+	node, err := in.Nodes.Get(name)
+	if err != nil {
+		return
+	}
+	if node.TaskResultSynced != nil {
+		tmp := true
+		node.TaskResultSynced = &tmp
```
Joibel: Use `ptr` again here

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2408,12 +2443,11 @@ func (in WorkflowStatus) FinishTime() *metav1.Time {
...
-}
-
-// Completed returns whether a node completed. Notably, a skipped node is not considered as having completed
-func (n NodeStatus) Completed() bool {
-	return n.Phase.Completed()
+	synced := true
+	if n.TaskResultSynced != nil {
+		synced = *n.TaskResultSynced
+	}
+	return n.Phase.Fulfilled(n.TaskResultSynced) && synced || n.IsDaemoned() && n.Phase != NodePending
```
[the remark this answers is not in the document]
isubasinghe: sorta right, but skipped node won't have a task result, so will inherit true.

But yeah I suppose it makes it more obvious though to make the change anyway.

pkg/apis/workflow/v1alpha1/workflow_types.go
Joibel: Could you pop in a Task Issue for this please

workflow/controller/operator.go
Joibel: I think it should be renamed for sanity

Joibel (changes requested):
I think we should decide if this is going to get backported. It feels like a fundamentally useful change and ideally we don't change the CRDs in backports.

We can get away without the new TaskResultsSynced field as demonstrated here https://github.com/pipekit/argo-workflows/pull/35. I personally don't like where that data is held, and would rather it was moved to where you have it, but that's a breaking change.

We certainly shouldn't track the same information in two places, that feels very wrong.

docs/environment-variables.md
isubasinghe: I meant more, put more emphasis on this value when configuring argo. I think it's good to keep both, because you can be more aggressive with `RECENTLY_DELETED_POD_DURATION` since you know the pod is gone. 
Hopefully that makes sense?

workflow/controller/plugin_template.go
```diff
@@ -10,7 +10,7 @@ func (woc *wfOperationCtx) executePluginTemplate(nodeName string, templateScope
 		if opts.boundaryID == "" {
 			woc.log.Warnf("[DEBUG] boundaryID was nil")
 		}
-		node = woc.initializeExecutableNode(nodeName, wfv1.NodeTypePlugin, templateScope, tmpl, orgTmpl, opts.boundaryID, wfv1.NodePending, opts.nodeFlag)
+		node = woc.initializeExecutableNode(nodeName, wfv1.NodeTypePlugin, templateScope, tmpl, orgTmpl, opts.boundaryID, wfv1.NodePending, opts.nodeFlag, true)
```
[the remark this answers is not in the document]
isubasinghe: No

isubasinghe:
@Joibel yes lets backport this but without the new field, but keep this one in for 3.8?

I get your point that this is duplicated, but it still is a single source of truth though (on the truth path at least), it's only really set true via MarkTaskResults(In)Complete, so it is more a derived property. 

I do not plan to ship a version of argo with both these fields (although, arguably we could ship this in 3.8 and then remove `taskResultStatuses` in 3.9 and that would be less of a hassle to upgrade into as a result.

I'm just stating ideas, I remember convincing you of doing it this way on a call so I am hoping I can re-convince you, mostly because this should be fixed at some point. Maybe if we are going to do a v4 release then this can be delayed until then though.

test/e2e/daemon_pod_test.go
```diff
@@ -200,7 +200,15 @@ spec:
     container:
       image: argoproj/argosay:v2
       command: ["bash"]
-      args: ["-c", "sleep 10 && exit 1"]
+      args:
```
Joibel, line 203: Is this intentionally left in?

workflow/controller/pod_cleanup.go
```diff
@@ -35,11 +35,12 @@ func (woc *wfOperationCtx) queuePodsForCleanup() {
 		}
 		nodeID := woc.nodeID(pod)
 		nodePhase, err := woc.wf.Status.Nodes.GetPhase(nodeID)
+		node, _ := woc.wf.Status.Nodes.Get(nodeID)
```
Joibel: Should check for err before and after this line I believe, as in Nodes.Get should check for err. Or at least reorder because the GetPhase and it's error check should be adjacent.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -2004,20 +2006,41 @@ type WorkflowStatus struct {
 }
 
 func (in *WorkflowStatus) MarkTaskResultIncomplete(name string) {
```
Joibel, line 2010: Perhaps explain this with a comment?

Joibel (changes requested):
Minor nits really

test/e2e/daemon_pod_test.go
isubasinghe: Yup, I needed to make this test less flakey.

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: Done

workflow/controller/pod_cleanup.go
isubasinghe: Ah thanks

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: Done

PR #14552, opened 2025-06-05
feat: handle interruptions as a transient error

rwong2888:
### Motivation

I have had this in my `TRANSIENT_ERROR_PATTERN` to handle automatically retrying pods that have been removed for one reason or another. This is to add it to the upstream argo-workflows. I feel it would be useful for users, especially those running workflows on preemptive or spot instances.

### Modifications

Added below as a transient error.
`(OOMKilled|imminent node shutdown|node was low on resource|pod deleted|node is shutting down|Deadline exceeded)`

### Verification

Used the TRANSIENT_ERROR_PATTERN at Forbes.

### Documentation

This file is already referenced in the [documentation](https://argo-workflows.readthedocs.io/en/latest/retries/#retry-policies).


util/errors/errors.go
```diff
@@ -93,6 +94,24 @@ func isTransientEtcdErr(err error) bool {
 	return false
 }
 
+func isTransientInterruptionErr(err error) bool {
+	// Handle transient node/resource related errors
+	if strings.Contains(err.Error(), "OOMKilled") {
```
Joibel: OOMKilled isn't always transient.

There is no way of removing items from this list for the setups or scenarios where it's not transient, so I'd like this removing.

Joibel (changes requested):
I'm not totally sure I like the other items in this either, but certainly OOMKilled needs removing.

I have a feeling we're mixing up a bunch of different uses for transient error detection and would like to think about it a bit. We're considering all error types from different scenarios (here pod errors, elsewhere API call errors) - maybe this is fine. Not caused by this PR, just a problem in general.

util/errors/errors.go
rwong2888: that's fair. that was a recent addition. removed oomkilled.

util/errors/errors.go
```diff
@@ -93,6 +94,22 @@ func isTransientEtcdErr(err error) bool {
 	return false
 }
 
+func isTransientInterruptionErr(err error) bool {
+	// Handle transient node/resource related errors
+	if strings.Contains(err.Error(), "imminent node shutdown") {
+		return true
+	} else if strings.Contains(err.Error(), "node was low on resource") {
```
MasonM:
```suggestion
	} else if strings.Contains(err.Error(), "The node was low on resource: ") {
		// Eviction due to resource pressure.
		// Source: https://github.com/kubernetes/kubernetes/blob/ec80dcaf3892d0cf89e4fd7b1ea60300fb417d8a/pkg/kubelet/eviction/helpers.go#L43
```

util/errors/errors.go
```diff
@@ -93,6 +94,22 @@ func isTransientEtcdErr(err error) bool {
 	return false
 }
 
+func isTransientInterruptionErr(err error) bool {
+	// Handle transient node/resource related errors
+	if strings.Contains(err.Error(), "imminent node shutdown") {
```
MasonM:
```suggestion
	if strings.Contains(err.Error(), "Pod was terminated in response to imminent node shutdown") {
		// Pod killed due to node shutdown.
		// Source: https://github.com/kubernetes/kubernetes/blob/ec80dcaf3892d0cf89e4fd7b1ea60300fb417d8a/pkg/kubelet/nodeshutdown/nodeshutdown_manager.go#L89
```

util/errors/errors.go
```diff
@@ -93,6 +94,22 @@ func isTransientEtcdErr(err error) bool {
...
+		return true
+	} else if strings.Contains(err.Error(), "node was low on resource") {
+		return true
+	} else if strings.Contains(err.Error(), "pod deleted") {
```
MasonM: This is a reference to this error, right? https://github.com/argoproj/argo-workflows/blob/e0a0f2856ba6e8a49b1ea28b15d50d39f3106125/workflow/controller/operator.go#L1262

From looking at that method, it seems this error should only happen during reconciliation when a pod is deleted after the `RECENTLY_STARTED_POD_DURATION` period without the controller being informed, which is unusual. Have you been encountering that under normal operation?

util/errors/errors.go
```diff
@@ -93,6 +94,22 @@ func isTransientEtcdErr(err error) bool {
...
+		return true
+	} else if strings.Contains(err.Error(), "node is shutting down") {
```
MasonM:
```suggestion
	} else if strings.Contains(err.Error(), "Pod was rejected as the node is shutting down.") {
		// Source: https://github.com/kubernetes/kubernetes/blob/ec80dcaf3892d0cf89e4fd7b1ea60300fb417d8a/pkg/kubelet/nodeshutdown/nodeshutdown_manager.go#L85
```

util/errors/errors.go
```diff
@@ -93,6 +94,22 @@ func isTransientEtcdErr(err error) bool {
...
+		return true
+	} else if strings.Contains(err.Error(), "Deadline exceeded") {
```
MasonM: This is a reference to this error, right?
https://github.com/argoproj/argo-workflows/blob/e0a0f2856ba6e8a49b1ea28b15d50d39f3106125/workflow/controller/operator.go#L120
https://github.com/argoproj/argo-workflows/blob/e0a0f2856ba6e8a49b1ea28b15d50d39f3106125/workflow/controller/operator.go#L2007-L2012

This isn't always a transient error, and I noticed we have code that has special handling for this, e.g. https://github.com/argoproj/argo-workflows/blob/e0a0f2856ba6e8a49b1ea28b15d50d39f3106125/workflow/controller/operator.go#L374-L378

so I'm not sure this is safe.

MasonM (commented):
Thanks @rwong2888! I did some sleuthing to try to identify the source of these errors and match the message as exactly as possible; let me know if I got anything wrong. As @Joibel mentioned, this is a bit tricky because there's a lot of different code paths that end up calling `IsTransientErr()`, and what counts as "transient" isn't the same in every one.

PR #14555, opened 2025-06-10
fix: retry forevever on task result creation related transient errors. Fixes #14560

### LINKED ISSUE #14560: task result creation should retry forever on transient errors.
### Pre-requisites

Currently transient errors do not retry, it has a cap of a max of 5 retries then gives up and reports container failure. 
This is not ideal, nor does it make sense. The task results are a must and we will spin the controller forever waiting for them if they do not arrive. This happens when this issue is  combined with say a podGC policy of OnSuccess. 

### Version(s)

:latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #14560

### Motivation

Currently the task result creation is only attempted 5 times when transient errors occur, this doensn't make any sense since we are relient on them to progress the state of the workflow. This change ensure that we retry effectively always when transient errors happen.

### Modifications
Modify the number of steps from 5 to the maximum int32 value.

### Verification

None performed since this only modifies the parameters passed.

### Documentation

Not documented .

### DIFF
--- a/workflow/executor/executor.go
+++ b/workflow/executor/executor.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"math"
 	"os"
 	"path"
 	"path/filepath"
@@ -858,18 +859,33 @@ func (we *WorkflowExecutor) ReportOutputsLogs(ctx context.Context) error {
 	return we.reportResult(ctx, wfv1.NodeResult{Outputs: &outputs})
 }
 
+func shouldLogRetry(attempt uint64) bool {
+	switch {
+	case attempt < 10:
+		return true // log first 10 attempts
+	case attempt < 100:
+		return attempt%10 == 0 // log every 10 attempts (10–99)
+	case attempt < 1000:
+		return attempt%50 == 0 // log every 50 attempts (100–999)
+	default:
+		return attempt%100 == 0 // log every 100 attempts (1000+)
+	}
+}
+
 func (we *WorkflowExecutor) reportResult(ctx context.Context, result wfv1.NodeResult) error {
+	var count uint64 // used to avoid spamming with these messages
 	return retryutil.OnError(wait.Backoff{
 		Duration: time.Second,
-		Factor:   2,
-		Jitter:   0.1,
-		Steps:    5,
+		Factor:   2.0,
+		Jitter:   0.2,
+		Steps:    math.MaxInt32, // effectively infinite retries
 		Cap:      30 * time.Second,
 	}, errorsutil.IsTransientErr, func() error {
 		err := we.upsertTaskResult(ctx, result)
-		if apierr.IsForbidden(err) {
+		if apierr.IsForbidden(err) && shouldLogRetry(count) {
 			log.WithError(err).Warn("failed to patch task result, see https://argo-workflows.readthedocs.io/en/latest/workflow-rbac/")
 		}
+		count++
 		return err
 	})
 }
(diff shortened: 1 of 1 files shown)


workflow/executor/executor.go
```diff
@@ -858,18 +859,33 @@ func (we *WorkflowExecutor) ReportOutputsLogs(ctx context.Context) error {
...
-		Steps:    5,
+		Factor:   2.0,
+		Jitter:   0.2,
+		Steps:    math.MaxInt32, // effectively infinite retries
 		Cap:      30 * time.Second,
 	}, errorsutil.IsTransientErr, func() error {
 		err := we.upsertTaskResult(ctx, result)
-		if apierr.IsForbidden(err) {
+		if apierr.IsForbidden(err) && shouldLogRetry(count) {
 			log.WithError(err).Warn("failed to patch task result, see https://argo-workflows.readthedocs.io/en/latest/workflow-rbac/")
```
Joibel: Log a count here?

jswxstw:
Some conflict errors are being ignored in `errorsutil.IsTransientErr`.

I think `isResourceQuotaConflictErr(err)` should be replaced with `apierr.IsConflict(err)`.

https://github.com/argoproj/argo-workflows/blob/2104eda0351cd67ed01d1e02544fea4b08d519f5/util/errors/errors.go#L52

workflow/executor/executor.go
```diff
@@ -858,18 +859,33 @@ func (we *WorkflowExecutor) ReportOutputsLogs(ctx context.Context) error {
...
-		Jitter:   0.1,
-		Steps:    5,
+		Factor:   2.0,
+		Jitter:   0.2,
+		Steps:    math.MaxInt32, // effectively infinite retries
 		Cap:      30 * time.Second,
 	}, errorsutil.IsTransientErr, func() error {
 		err := we.upsertTaskResult(ctx, result)
-		if apierr.IsForbidden(err) {
+		if apierr.IsForbidden(err) && shouldLogRetry(count) {
```
jswxstw: `shouldLogRetry` here will not take any effect because `apierr.IsForbidden(err)` is not transient error.

isubasinghe:
@jswxstw I think lets address that in a different PR to keep the PR atomic, and also just to get this one merged in fast (this is mostly why)

workflow/executor/executor.go
isubasinghe: Sure

jswxstw:
Sure, I will submit a new PR to fix my issue.

workflow/executor/executor.go
isubasinghe: You are correct, thanks changing it now. I am surprised there is no logging of failed transient errors in that case.

workflow/executor/executor.go
```diff
@@ -858,18 +859,35 @@ func (we *WorkflowExecutor) ReportOutputsLogs(ctx context.Context) error {
 	return we.reportResult(ctx, wfv1.NodeResult{Outputs: &outputs})
 }
 
+func shouldLogRetry(attempt uint64) bool {
```
jswxstw: I think `shouldLogRetry` is unnecessary,  because backoff is already configured during retries.

isubasinghe: I disagree, the delay between logs at its worse case will be 30 seconds. As time goes on, `shouldLogRetry` basically makes sure we place less emphasis on this container in the logs (since it is likely the user probably noticed it and likely doesn't care, this is my hypothesis anyway).

Joibel: I'm inclined to agree with @jswxstw here. This container is dying, and nothing else should be happening, this is just unnecessary complexity.

Joibel (approved):
Approving anyway, but think we can lose the logging complexity.

PR #14574, opened 2025-06-17
docs: release docs for 3.7

isubasinghe:
### Documentation

This adds documentation for the release of 3.7, mostly the new-features documentation along with a note regarding using non-root argoexec.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,7 +1,5 @@
 # Metrics
 
-> v2.7 and after
-
 ## Introduction
 
 Argo emits a certain number of controller metrics that inform on the state of the controller at any given time.
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -7,6 +7,8 @@ the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summar
 
 See also the list of [new features in 3.7](new-features.md).
 
+For upgrades to older versions of Argo Workflows, please change to the documentation for the version of interest.
+
 ### Deprecations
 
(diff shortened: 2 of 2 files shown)


.spelling
```diff
@@ -79,6 +79,7 @@ Kaniko
 Katacoda
 Katib
 Kerberos
+KeyValueEditor
```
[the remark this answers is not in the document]
Joibel: You're not very good at this are you Copilot?

docs/upgrading.md
```diff
@@ -3,9 +3,9 @@
 Breaking changes  typically (sometimes we don't realise they are breaking) have "!" in the commit message, as per
 the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).
 
-## Upgrading to v3.6
+## Upgrading to v3.7
```
Joibel, line 6: I generally like this change as not being an ever growing document. To save the questions in slack, please can we also pop a note to say please change to the older version's documentation to read the notes for upgrading to that version.

docs/metrics.md
```diff
@@ -2,9 +2,6 @@
 
 > v2.7 and after
```
tico24: we can drop this now can't we?

isubasinghe: Yes, we could indeed.

docs/upgrading.md
isubasinghe: Sure

Joibel (approved):

Joibel:
Will need cherry-picking to release/3.7 when that is created.

PR #14579, opened 2025-06-18
fix: set creator when use X509 client certificates. Fixes: #14578

### LINKED ISSUE #14578: Support obtaining user information from X509 client certificates kubeconfig
# Summary

Obtaining user information from X509 client certificates kubeconfig

## Use Cases

Set Creator, Administrators always want to know who created the workflow.
https://argo-workflows.readthedocs.io/en/latest/workflow-creator/

X509 client certificates are a very popular authentication method. In this case, we are currently unable to resolve user information and associate it with Workflows. Only Basic Auth（Username/Password）、Bearer Token（ServiceAccount Token） work.
https://github.com/argoproj/argo-workflows/blob/main/server/auth/serviceaccount/claims.go

---

shuangkun:
Set creator when use X509 client certificates.

Fixes #14578 

### Verification

Loccal Test:
When use an alibaba cloud  ACK cluster kubeconfig.
image
You can see that after the update, there is creator. So do a local k3d cluster.

### DIFF
--- a/server/auth/serviceaccount/claims.go
+++ b/server/auth/serviceaccount/claims.go
@@ -25,7 +25,7 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
 		return ClaimSetWithBearerToken(restConfig)
 	}
 
-        if restConfig.CertFile != "" || len(restConfig.CertData) > 0 {
+	if restConfig.CertFile != "" || len(restConfig.CertData) > 0 {
 		return ClaimSetWithX509(restConfig)
 	}
 	return nil, nil
--- a/server/auth/serviceaccount/claims.go
+++ b/server/auth/serviceaccount/claims.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 
@@ -81,7 +80,7 @@ func ClaimSetWithX509(restConfig *rest.Config) (*types.Claims, error) {
 		cert, _ = x509.ParseCertificate(block.Bytes)
 	} else if restConfig.TLSClientConfig.CertFile != "" {
 		// Load certificate from file
-		data, err := ioutil.ReadFile(restConfig.TLSClientConfig.CertFile)
+		data, err := os.ReadFile(restConfig.TLSClientConfig.CertFile)
 		if err != nil {
 			return nil, fmt.Errorf("failed to read certificate file: %w", err)
 		}
--- a/server/auth/serviceaccount/claims.go
+++ b/server/auth/serviceaccount/claims.go
@@ -1,9 +1,12 @@
 package serviceaccount
 
 import (
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


server/auth/serviceaccount/claims.go
```diff
@@ -60,3 +67,55 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
 
 	return claims, nil
 }
+
+func ClaimSetWithX509(restConfig *rest.Config) (*types.Claims, error) {
+	var cert *x509.Certificate
+
+	if len(restConfig.CertData) > 0 {
```
tczhao, line 77: reading on [TLSClientConfig](https://github.com/kubernetes/client-go/blob/v0.33.2/rest/config.go#L234), wouldn't CertData be nested under Config.TLSClientConfig

server/auth/serviceaccount/claims.go
```diff
@@ -60,3 +67,55 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
...
+		// Decode certificate from memory data
+		block, _ := pem.Decode(restConfig.CertData)
+		if block == nil || block.Type != "CERTIFICATE" {
+			return nil, fmt.Errorf("failed to parse certificate PEM")
+		}
+		cert, _ = x509.ParseCertificate(block.Bytes)
```
tczhao: we need to handle error here

tczhao (changes requested):

server/auth/serviceaccount/claims.go
shuangkun: The rest.Config structure embeds the TLSClientConfig type, so the result should be the same.

lint will make this change automatically. So I change restConfig.TLSClientConfig.CertData to restConfig.CertData
```
-		block, _ := pem.Decode(restConfig.TLSClientConfig.CertData)
+		block, _ := pem.Decode(restConfig.CertData)
```

server/auth/serviceaccount/claims.go
```diff
@@ -18,10 +20,15 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
...
 		return &types.Claims{Claims: jwt.Claims{Subject: username}}, nil
 	}
-	if restConfig.BearerToken == "" && restConfig.BearerTokenFile == "" {
-		return nil, nil
+
+	if restConfig.BearerToken != "" || restConfig.BearerTokenFile != "" {
+		return ClaimSetWithBearerToken(restConfig)
 	}
 
+	return ClaimSetWithX509(restConfig)
```
tczhao: Would this look cleaner? instead of having default nil,nil inside the X509
```suggestion
        if restConfig.CertFile != "" || len(restConfig.CertData) > 0 {
		return ClaimSetWithX509(restConfig)
	}
	return nil, nil
```

server/auth/serviceaccount/claims.go
```diff
@@ -60,3 +70,61 @@ func ClaimSetFor(restConfig *rest.Config) (*types.Claims, error) {
...
+		if block == nil || block.Type != "CERTIFICATE" {
+			return nil, fmt.Errorf("failed to parse certificate PEM")
+		}
+		cert, err = x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate: %w", err)
+		}
+	} else {
+		return nil, nil // No certificate info available
+	}
```
tczhao:
```suggestion
	} else {
		// Load certificate from file
		data, err := os.ReadFile(restConfig.CertFile)
		if err != nil {
			return nil, fmt.Errorf("failed to read certificate file: %w", err)
		}
		block, _ := pem.Decode(data)
		if block == nil || block.Type != "CERTIFICATE" {
			return nil, fmt.Errorf("failed to parse certificate PEM")
		}
		cert, err = x509.ParseCertificate(block.Bytes)
		if err != nil {
			return nil, fmt.Errorf("failed to parse certificate: %w", err)
		}
	}
```

tczhao (approved):

Issue #14334, opened 2025-03-27
UI: Display when Conditions in CronWorkflow UI


garireo2549:
# Summary
The when condition feature was introduced in Argo Workflows v3.6 to allow more flexible scheduling for CronWorkflow. However, currently, this condition is not visible in the UI, requiring users to check the YAML definition manually. Displaying it in the UI would improve the user experience and make it easier to manage scheduled workflows.

## Use Cases

Users often need to verify these conditions before making modifications or debugging failed executions. Displaying the when condition directly in the UI will enhance usability and reduce the need to check YAML configurations manually.

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

garireo2549:
PR: https://github.com/argoproj/argo-workflows/pull/14585

PR #14585, opened 2025-06-19
feat: Display when Conditions in CronWorkflow UI : Fixes: #14334

garireo2549:
Fixes https://github.com/argoproj/argo-workflows/issues/14334

### Motivation

There was a need to display and edit the when conditions of CronWorkflows in the UI. The current UI had no way to display or edit the when property even when it was included in the API response. This change allows users to easily view and manage the when conditions of their CronWorkflows.

### Modifications

Added when?: string[] property to the CronWorkflowSpec type

Modified ui/src/shared/models/cron-workflows.ts
Added a "WhenConditions" section to the CronWorkflowSpecEditor component

Modified ui/src/cron-workflows/cron-workflow-spec-editor.tsx
Added a new section right after the Schedules section
Added updateWhenAtIndex function for array operations
Modified the normalizeSchedules function to properly handle the when property when it's included in the API response

Modified ui/src/shared/services/cron-workflow-service.ts
Set an empty array if the when property doesn't exist
Convert when to an array if it's a string

### Verification

Started the UI in a local environment and created a CronWorkflow with the when property
Opened the CronWorkflow details page and confirmed that the newly added "WhenConditions" section appears in the "Cron" tab
Verified that editing existing when conditions and adding new ones works correctly
Confirmed that the when property from the API response is displayed correctly

スクリーンショット 2025-06-19 17 48 32

### Documentation

This change is a UI enhancement that extends the existing CronWorkflow concept, so no additional documentation is required. The UI change is intuitive, and users can easily find the new "WhenConditions" section within the "Cron" tab.

### Beyond this PR

Thank you for submitting this! Have you ever thought of becoming a Reviewer or Approver on the project?

Argo Workflows is seeking more community involvement and ultimately more [Reviewers and Approvers](https://github.com/argoproj/argoproj/blob/main/community/membership.md) to help keep it viable.
See [Sustainability Effort](https://github.com/argoproj/argo-workflows/blob/main/community/sustainability_effort.md) for more information.

-->

### DIFF
--- a/ui/src/cron-workflows/cron-workflow-spec-editor.tsx
+++ b/ui/src/cron-workflows/cron-workflow-spec-editor.tsx
@@ -28,7 +28,7 @@ export function CronWorkflowSpecEditor({onChange, spec}: {spec: CronWorkflowSpec
                         {spec.when ? (
                             <TextInput value={spec.when} onChange={newCondition => onChange({...spec, when: newCondition})} />
                         ) : (
-                            <TextInput value='' placeholder='Enter condition' onChange={newCondition => onChange({...spec, when: newCondition})} />
+                            <TextInput value='' onChange={newCondition => onChange({...spec, when: newCondition})} />
                         )}
                     </div>
                 </div>


ui/src/shared/services/cron-workflow-service.ts
```diff
@@ -13,6 +13,14 @@ function normalizeSchedules(cronWorkflow: any): CronWorkflow {
...
         delete cronWorkflow.spec.schedule;
     }
+
+    // Ensure when property is properly handled
+    cronWorkflow.spec.when ??= [];
+    // If when is a string in the API response, convert it to an array
+    if (typeof cronWorkflow.spec.when === 'string') {
+        cronWorkflow.spec.when = [cronWorkflow.spec.when];
+    }
+
```
tczhao: Should we leave out the array part?
accordingly to the [spec doc](https://github.com/argoproj/argo-workflows/pull/13474/files#diff-ce7b64e72add3d44567b11ab1a4b1e6bfc9345a86b4a352283140e4974776c4dR8850), it only returns string

garireo2549: Fixed it: 15e37a92c1593091e82e97aec8d877cf89dc3225

ui/src/cron-workflows/cron-workflow-spec-editor.tsx
```diff
@@ -22,6 +22,16 @@ export function CronWorkflowSpecEditor({onChange, spec}: {spec: CronWorkflowSpec
                         ))}
                     </div>
                 </div>
+                <div className='row white-box__details-row'>
+                    <div className='columns small-3'>WhenConditions</div>
```
tczhao: I see the other fields in the UI follows the same name as the api doc,
perhaps let keep it consistent and only use when here
```suggestion
                    <div className='columns small-3'>When</div>
```

garireo2549: Fixed it : 55b810d5923f5337515bd8b30c09e563c79159f4

garireo2549:
@tczhao 
Thank you for your review. I have completed the revisions based on the CI and the points raised during the review, so please check them at your convenience.

tczhao (approved):

tczhao:
Thank you for the fix!

PR #14589, opened 2025-06-20
docs: Added doc for GC directory

### LINKED ISSUE #14565: [S3] artgc Artifact Garbage Collect cannot delete directories
### Pre-requisites

When uploading a directory to a S3 bucket, it is successful. When the workflow is deleted, the artgc pod tries to remove the directory but can't. It can only deletes list of files.

### Version(s)

v3.6.7

```YAML
metadata:
  name: testtodel-template
  generateName: testtodel-
  namespace: argo-workflows
spec:
  entrypoint: init-pipeline
  templates:
    - name: init-pipeline
      steps:
        - - name: upload-file-test
            template: upload-file
        - - name: upload-directory-test
            template: upload-directory

    - name: upload-file
      outputs:
        artifacts:
          - name: file_to_localstorage
            path: "/tmp/not_a_directory"
            archive:
              # Avoid having tgz file.
              none: {}
            s3:
              key: "{{workflow.name}}/not_a_directory"
      script:
        image: "bash:3.1.23-alpine3.21"
        command: ["bash"]
        source: |
          set -x
          id
          date -Is --utc
          echo 123 >/tmp/not_a_directory
        resources:
          requests:
            memory: "50Mi"
            cpu: "50m"
          limits:
            memory: "50Mi"
            cpu: "50m"

    - name: upload-directory
      outputs:
        artifacts:
          - name: file_to_localstorage
            path: "/tmp/directory"
            archive:
              # Avoid having tgz file.
              none: {}
            s3:
              key: "{{workflow.name}}/directory"
      script:
        image: "bash:3.1.23-alpine3.21"
        command: ["bash"]
        source: |
          set -x
          id
          date -Is --utc
          mkdir /tmp/directory
          echo 123 >/tmp/directory/onefile
        resources:
          requests:
            memory: "50Mi"
            cpu: "50m"
          limits:
            memory: "50Mi"
            cpu: "50m"
```

```text
not relevant
```

```text
not relevant
```

jmabuin:
Fixes [14565](https://github.com/argoproj/argo-workflows/issues/14565)

### Motivation

This PR makes clearer the way of using the artifact GC when the S3 key is a whole directory. Comes from a problem stated at [14565](https://github.com/argoproj/argo-workflows/issues/14565)

### Modifications

Added documentation so this case is clear.

### Verification
Verified in the specified issue [14565](https://github.com/argoproj/argo-workflows/issues/14565)

### Documentation

Added

### DIFF
--- a/docs/walk-through/artifacts.md
+++ b/docs/walk-through/artifacts.md
@@ -155,11 +155,11 @@ spec:
 
 Consider parameterizing your S3 keys by {{workflow.uid}}, etc (as shown in the example above) if there's a possibility that you could have concurrent Workflows of the same spec. This would be to avoid a scenario in which the artifact from one Workflow is being deleted while the same S3 key is being generated for a different Workflow.
 
-In the case of having a whole directory as S3 key, please pay attention of the key value:
-- (A) and changing the default archive option to none (see example below), it is important that it ends with a "/". Otherwise, the directory will be created in S3 but the GC pod won't be able to remove it.
-- (B) and keeping the default archive option to tgz (see example below), the it is import that it does NOT end with "/". Otherwise Argo will fail to create the archive file.
+In the case of having a whole directory as S3 key, please pay attention to the key value. Here are two examples:
+- (A) When changing the default archive option to none, it is important that it ends with a "/". Otherwise, the directory will be created in S3 but the GC pod won't be able to remove it.
+- (B) When keeping the default archive option to `.tgz`, in this case, it is important that it does NOT end with "/". Otherwise, Argo will fail to create the archive file.
 
-Example (A) without packaging as tgz
+Example (A) without packaging as `.tgz`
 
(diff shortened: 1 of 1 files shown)


antoinetran:
Hi @jmabuin , thanks for the contribution!

I do not know the process of suggesting change in your branch, so I created another branch in my fork here: https://github.com/antoinetran/argo-workflows/tree/14565-artifact-gc-docs

Feel free to pull them in your branch if you feel this is OK. I tested your code, the example did not work because:
1. the mkdir was missing
2. the archive was default (tgz file). In this particular case of directory + archive not none, the argo will fail, thus I created another example with both case.

antoinetran:
@jmabuin I just saw the DCO failure, I pushed force in my repository with my commit with "signed off" field. I think you should do the same on your side?

tczhao (approved):

PR #14596, opened 2025-06-24
chore(deps): bump golang version

Joibel:
Just a version bump for golang

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ ARG GIT_COMMIT=unknown
 ARG GIT_TAG=unknown
 ARG GIT_TREE_STATE=unknown
 
-FROM golang:1.24-alpine3.21 as builder
+FROM golang:1.24.4-alpine3.22 as builder
 
 # libc-dev to build openapi-gen
 RUN apk update && apk add --no-cache \
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/argoproj/argo-workflows/v3
 
-go 1.24.2
+go 1.24.4
 
 require (
 	cloud.google.com/go/storage v1.55.0


Dockerfile
```diff
@@ -3,7 +3,7 @@ ARG GIT_COMMIT=unknown
 ARG GIT_TAG=unknown
 ARG GIT_TREE_STATE=unknown
 
-FROM golang:1.24-alpine3.21 as builder
+FROM golang:1.24.4-alpine3.22 as builder
```
tczhao, line 6: Shall we keep this as 1.24?
this way we can build any 1.24.x version specified by go.mod
(go installation version needs to be >= go mod version)

Otherwise, we may need to update 1.24 to 1.24.4 for windows Dockerfile, devcontainer and github action for consistency and this is more to maintain

Joibel: `golang:1.24-alpine3.22` still contains a specific version of go each time you run it. Right now that's the same as this specific image:
```bash
$  docker run --rm golang:1.24-alpine3.22 go version
go version go1.24.4 linux/amd64
```

I was just pinning the image to an unmoving target as it's usually better practice to do that.

The `go.mod` value drives the github action which is the main one that matters.

Right now devcontainer contains `1.24.3` which isn't what we wanted (e.g. 1.24.2)
```
vscode ➜ ~/go/src/github.com/argoproj/argo-workflows (main) $ go version
go version go1.24.3 linux/amd64
```

I'd rather we control more of this rather than less - happy to update the devcontainer and windows Dockerfile if that would be acceptable. I know it's somewhat more effort and golang is good about compatibility.

tczhao: Make sense,
I checked both the Dockerfile.windows and devcontainer installs go on the flight rather than having it build-in to the base image. It will work with the current change even without explicitly specifying the patch version.
I'm fine either way, will approve for now

tczhao (approved):

PR #14598, opened 2025-06-24
fix: executor workflowtaskresult retry should use the default retry and configurable

### LINKED ISSUE #14381: executor workflowtaskresult retry should use the default retry and configurable
# Summary

What change needs making?

taskresult related operation retry timings are currently hardcoded
https://github.com/argoproj/argo-workflows/blob/8793d8893b68ca81800ff9c3512780636ed733cd/workflow/executor/executor.go#L804-L810

It should use the default retry like all other operation that interacts with k8s api
https://github.com/argoproj/argo-workflows/blob/8793d8893b68ca81800ff9c3512780636ed733cd/workflow/executor/executor.go#L706-L712

## Use Cases

When would you use this?

So that we can configure retry timing, incase k8s api server is unstable, we can still have workflow succeeding

---

akashjkhamkar:
Fixes #14381 

### Motivation
- Some of the executor functions are using hard coded backoff params, preventing us from configuring backoff values to avoid failures due to k8s service downtimes

### Modifications
[workflow/executor/retry/retry.go](https://github.com/argoproj/argo-workflows/pull/14598/files#diff-6351f1acf764a41c79df22af22100dc14026ba3974877bfb464c20cdd143b300R45)
- Added a new func `GetExecutorRetry` to get a configurable backoff object, which gives env's precedence over the arguments passed

- New env variable `EXECUTOR_RETRY_BACKOFF_CAP` to make the cap configurable

[argo-workflows/workflow/executor/executor.go](https://github.com/argoproj/argo-workflows/blob/8793d8893b68ca81800ff9c3512780636ed733cd/workflow/executor/executor.go)

Changes all the hardcoded `wait.Backoff{` to use `executorretry.GetExecutorRetry`
Keeping all the currently passed values as it is,
If the user has set envs , then they will take precedence over the values that are passed


workflow/executor/executor.go
```diff
@@ -859,13 +845,7 @@ func (we *WorkflowExecutor) ReportOutputsLogs(ctx context.Context) error {
...
 func (we *WorkflowExecutor) reportResult(ctx context.Context, result wfv1.NodeResult) error {
 	var count uint64 // used to avoid spamming with these messages
-	return retryutil.OnError(wait.Backoff{
-		Duration: time.Second,
-		Factor:   2.0,
-		Jitter:   0.2,
-		Steps:    math.MaxInt32, // effectively infinite retries
-		Cap:      30 * time.Second,
-	}, errorsutil.IsTransientErr, func() error {
+	return retryutil.OnError(retryutil.DefaultRetry, errorsutil.IsTransientErr, func() error {
```
tczhao: can we keep the default and only overwrite if the config is present

akashjkhamkar: can you elaborate @tczhao 

didn't get what you mean by "if the config is present"

akashjkhamkar: made some changes

`GetExecutorRetry` is a new function , that can take params and return an backoff object
If envs are set, then envs take precedence over valeus passed by the user

workflow/executor/retry/retry.go
```diff
@@ -22,3 +22,32 @@ var (
...
+//
+// Returns:
+//   - wait.Backoff: configured backoff parameters.
+func GetExecutorRetry(duration time.Duration, factor, jitter float64, steps int, cap time.Duration) wait.Backoff {
+	return wait.Backoff{
+		Steps:    env.LookupEnvIntOr("EXECUTOR_RETRY_BACKOFF_STEPS", steps),
+		Duration: env.LookupEnvDurationOr("EXECUTOR_RETRY_BACKOFF_DURATION", duration),
+		Factor:   env.LookupEnvFloatOr("EXECUTOR_RETRY_BACKOFF_FACTOR", factor),
+		Jitter:   env.LookupEnvFloatOr("EXECUTOR_RETRY_BACKOFF_JITTER", jitter),
+		Cap: cap,
```
tczhao: Perhaps we can use `EXECUTOR_RETRY_BACKOFF_CAP`,
since we are providing a default `cap` we want the ability to overwrite this

tczhao (changes requested):
also need to fix the lint CI failure

akashjkhamkar:
fixed, ptal @tczhao

docs/environment-variables.md
```diff
@@ -102,6 +102,7 @@ spec:
 | `EXECUTOR_RETRY_BACKOFF_FACTOR`        | `float`         | `1.6`   | The retry back-off factor when the workflow executor performs retries.                                 |
 | `EXECUTOR_RETRY_BACKOFF_JITTER`        | `float`         | `0.5`   | The retry back-off jitter when the workflow executor performs retries.                                 |
 | `EXECUTOR_RETRY_BACKOFF_STEPS`         | `int`           | `5`     | The retry back-off steps when the workflow executor performs retries.                                  |
+| `EXECUTOR_RETRY_BACKOFF_CAP`         | `time.Duration`           | `30s`     | The retry back-off cap duration when the workflow executor performs retries.                                  |
```
tczhao: Could you format this, have the `|` aligned

akashjkhamkar: done

tczhao: The trailing `|` is not aligned

akashjkhamkar: neither of the other entries have it aligned

akashjkhamkar: can format the whole table instead, since there are inconsistencies in the existing rows

workflow/executor/retry/retry.go
```diff
@@ -20,5 +20,35 @@ var (
 	Duration      = env.LookupEnvDurationOr("EXECUTOR_RETRY_BACKOFF_DURATION", 1*time.Second)
 	Factor        = env.LookupEnvFloatOr("EXECUTOR_RETRY_BACKOFF_FACTOR", 1.6)
 	Jitter        = env.LookupEnvFloatOr("EXECUTOR_RETRY_BACKOFF_JITTER", 0.5)
-	ExecutorRetry = wait.Backoff{Steps: Steps, Duration: Duration, Factor: Factor, Jitter: Jitter}
+	Cap           = env.LookupEnvDurationOr("EXECUTOR_RETRY_BACKOFF_CAP", 30*time.Second)
+	ExecutorRetry = wait.Backoff{Steps: Steps, Duration: Duration, Factor: Factor, Jitter: Jitter, Cap: Cap}
```
tczhao: I feel we better leave this as is, don't change
If we want to change, set default to something that behave the same as not setting it

akashjkhamkar: keeping it 0, as by default its 0

akashjkhamkar: pushed

docs/environment-variables.md
akashjkhamkar: let me know what you think

tczhao (approved):
👍

tczhao:
Could you take a look as well
@isubasinghe

isubasinghe:
@tczhao I think my latest change kinda covers this right? I think we should retry forever on transient fails.

isubasinghe (changes requested):
I put in a new PR to retry forever on task result failures, but I only changed the reportResult function. Maybe change this PR so that we retry forever on failed patches as welll? Or just open another PR

akashkhamkarr:
@isubasinghe Would infinite retries be ideal ? (for extreme cases)
I think it's best to have retry strategies configurable

isubasinghe:
@akashjkhamkar yes, workflowtaskresults are sorta critical.

akashjkhamkar:
@isubasinghe reverted previous changes, and added infinite retries for the concerned functions.

Please take a look, and let me know if there are any other places where we might need infinite retries

isubasinghe:
@akashjkhamkar could you copy my logging logic as well please?
https://github.com/argoproj/argo-workflows/commit/b81e8c9fc04667eb973e05a8f4aace7ee81e0f1a

PR #14599, opened 2025-06-24
fix: restore the CRDs to install.yaml

Joibel:
3.7-rc [`install.yaml`](https://github.com/argoproj/argo-workflows/releases/download/v3.7.0-rc2/install.yaml) doesn't contain the CRDs.

I believe this an unintentional change from #14048 - flagging @MasonM to verify.

Verifiable as a much smaller size for the asset at https://github.com/argoproj/argo-workflows/releases/tag/v3.7.0-rc2, which none of the other manifests have.

In order that the test manifests can use the full CRDs I've split the cluster-install into one without the CRDs (`cluster-install-no-crds`) and then made the `cluster-install` just take that and inject the minimized CRDS.


Joibel:
```
ls -lh dist/manifests/
total 652K
-rw-r--r-- 1 alan 1000 118K Jun 24 11:52 install.yaml
-rw-r--r-- 1 alan 1000 116K Jun 24 11:52 namespace-install.yaml
-rw-r--r-- 1 alan 1000 135K Jun 24 11:52 quick-start-minimal.yaml
-rw-r--r-- 1 alan 1000 137K Jun 24 11:52 quick-start-mysql.yaml
-rw-r--r-- 1 alan 1000 137K Jun 24 11:52 quick-start-postgres.yaml
```

isubasinghe (approved):
Thanks

Joibel:
> I'm trying to think of how to prevent this kind of bug, and introducing something like https://www.conftest.dev/ would work, but I'm not sure that's worth it.

Yeah, it doesn't feel that it's the best use of effort compared to other things we could improve. We're aware we can make this mistake so perhaps we won't make it again.

PR #14613, opened 2025-06-28
fix: avoid healthz check restart controller. Fixes: #14526

### LINKED ISSUE #14526: large-scale scenarios the controller restarted
### Pre-requisites

When more than 8,000 workflows are submitted at the same time, even if the pod worker is opened as high as 128, but the controller still always restarts，like this：
```
workflow-controller-6f55c5f474-gt5vd   1/1     Running   41 (11m ago)   6h41m
```
the root cause is some workflows cannot be reconceiled within 5 minutes. At this time, the workflow-controller will restart because of healthz check, increasing the system burden.

```
		for _, wf := range unreconciledWorkflows {
			if time.Since(wf.GetCreationTimestamp().Time) > age {
				return fmt.Errorf("workflow never reconciled: %s", wf.Name)
			}
		}
```

### Version(s)

latest

```YAML
many workflows scenarios
```

```text
```

```text
```

shuangkun:
avoid healthz check restart controller in large-scale scenarios.

Fixes #14526 

### Motivation
When submitting large-scale workflows at one time (8000+), avoid frequent restarts of the workflow-controller due to healthz problems

### Modifications

Add a cache of workflows that were not reconceiled last time.
If there are workflows that have not been reconceiled for more than 5 minutes, check whether the workflows in the cache have decreased.
If there are decreases, it means that the workflow-controller is still processing, but it is slow, and there is no need to restart.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -150,7 +150,8 @@ type WorkflowController struct {
 	executorPlugins          map[string]map[string]*spec.Plugin // namespace -> name -> plugin
 
 	recentCompletions recentCompletions
-	lastUnreconciled  map[string]*wfv1.Workflow
+	// lastUnreconciledWorkflows is a map of workflows that have been recently unreconciled
+	lastUnreconciledWorkflows map[string]*wfv1.Workflow
 }
 
 const (
--- a/workflow/controller/healthz.go
+++ b/workflow/controller/healthz.go
@@ -76,7 +76,7 @@ func (wfc *WorkflowController) Healthz(w http.ResponseWriter, r *http.Request) {
 			log.Info("healthz: workflows exceed max age")
 			// Check if there is progress by comparing with the last check:
 			// If all workflows from last time are still present, it means no progress
-			for key := range wfc.lastUnreconciled {
+			for key := range wfc.lastUnreconciledWorkflows {
 				if _, exists := unreconciledWorkflows[key]; !exists {
 					// At least one workflow has been reconciled, so there is progress
 					noProgress = false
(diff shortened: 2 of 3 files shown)


tczhao:
Any thoughts on whether this fix could mask a bigger problem?

The new and original implementation seems to serve a different purpose

- The originally implementation is checking _if a workflow e.g. workflow A has reconciled within 5min_
- The new implementation is checking _if there's progress in any unreconciled workflows._

shuangkun:
> Any thoughts on whether this fix could mask a bigger problem?
> 
> The new and original implementation seems to serve a different purpose
> 
> * The originally implementation is checking _if a workflow e.g. workflow A has reconciled within 5min_
> * The new implementation is checking _if there's progress in any unreconciled workflows._

@Joibel  Do you have any comments? Thanks

Joibel:
So I do agree that this is a problem, and the original code isn't great - we've seen heavy WF-of-WF patterns causing controller restarts because of this, and this then makes the problem worse because the controller is obviously much less able to keep up as it's spending a lot of time getting ready so it's hard for it to service all the workflows.

The original implementation only requires that one workflow is exceeding the 5 minute reconciliation window to cause a problem. If others have reconciled but one hasn't this counts against it - and does seem an unreasonable expectation to meet.

As for this change - it would alleviate that problem. It doesn't alleviate the problem of one workflow being fully stuck, which the old mechanism would catch but this wouldn't. This change feels like the correct solution to this problem, so I agree with doing it. @tczhao - did you have a specific scenario (or multiple scenarios) that this change would mask in mind?

tczhao:
Thanks @Joibel , no issue from my side

tczhao (approved):

Joibel:
Should we cherry-pick this back to 3.6?

PR #14618, opened 2025-07-01
fix: create task results only once. Fixes: #14617

### LINKED ISSUE #14617: Avoid frequent calls to createTaskResult in large-scale scenarios
### Pre-requisites

In large-scale scenarios (more than 10,000 pods), we.monitorProgress causes frequent calls to create the workflowtaskresult interface, resulting in a low success rate for the entire apiserver write and a significant decrease in overall performance.

In Alibaba Cloud's log system, we can see a single workflow pod called 57 times, post workflowtaskresult requests, and the apiserver returned 409. This call is invalid, which increases the pressure on the APIserver and causes poor performance, especially when there are many pods in the cluster.

Image

### Version(s)

v3.5.13

```YAML
any workflow
```

```text
```

```text
```

shuangkun:
Create taskresult only once. Avoid excessive pressure on the API server and performance degradation in large-scale scenarios.

Fixes #14617

### DIFF
--- a/workflow/executor/taskresult.go
+++ b/workflow/executor/taskresult.go
@@ -18,18 +18,15 @@ func (we *WorkflowExecutor) upsertTaskResult(ctx context.Context, result wfv1.No
 	if !we.taskResultCreated {
 		err := we.createTaskResult(ctx, result)
 		if apierr.IsAlreadyExists(err) {
-			we.taskResultCreated = true // Mark as already existed
 			return we.patchTaskResult(ctx, result)
 		}
 		if err != nil {
 			return err
 		}
-		we.taskResultCreated = true // Mark as created after successful creation
 	} else {
 		err := we.patchTaskResult(ctx, result)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


workflow/executor/taskresult.go
```diff
@@ -15,11 +15,27 @@ import (
...
+		if err != nil {
+			return err
+		}
+		we.taskResultCreated = true // Mark as created after successful creation
+	} else {
+		err := we.patchTaskResult(ctx, result)
+		if err != nil {
+			if apierr.IsNotFound(err) {
+				we.taskResultCreated = false // Reset flag since it's no longer available
+				return we.createTaskResult(ctx, result)
```
Joibel, line 30: At this point we'll create a task result and not record it as created. Not a disaster, but perhaps setting the flag true inside `createTaskResult` instead of in this function could fix that.

shuangkun: Thanks, I modified it

PR #14625, opened 2025-07-02
fix: Make the phase of the node unchange when the pod is completed and outputs are not set in the status.node

### LINKED ISSUE #14497: workflow status node output parameters is missing occasionally when use withParam
### Pre-requisites

workflow status node output parameters is missing occasionally,  and **it's hard to reproduce it.**
I checked the wait container exit code is 0, and the main container logs are successfully uploaded to s3, so I think wait container has successfully created the workflowtaskresult. 
My pod gc strategy is OnPodCompletion, when pod was deleted, the workflowtaskresult will be deleted too. I think the problem is in the workflow-controller, I have checked the related code, but can't find the root cause.

The correct node status output is:

Image

The abnormal node status output is like this:

Image

### Version(s)

v3.4.8

```YAML
kind: Workflow
metadata:
  generateName: parallel-tasks-
spec:
  entrypoint: flow
  templates:
    - name: flow
      dag:
        tasks:
          - name: start
            template: argosay
            arguments: {}
          - name: start-test-output
            template: argosay
            dependencies:
              - start
          - name: start-test-output-foreach-input-list
            template: start-test-output-foreach-input-list
            arguments:
              parameters:
                - name: split-index
                  value: "{{item}}"
            dependencies:
              - start-test-output
            withParam: "[0,1,2,3,4,5,6,7,8,9]"
          - name: end-test-output
            template: argosay
            dependencies:
              - start-test-output-foreach-input-list

    - name: start-test-output-foreach-input-list
      inputs:
        parameters:
          - name: split-index
      outputs:
        parameters:
          - name: task-id
            valueFrom:
              parameter: '{{tasks.test-output.outputs.parameters.task-id}}'
      dag:
        tasks:
          - name: test-output
            template: test-output
            arguments:
              parameters:
                - name: task-index
                  value: "{{inputs.parameters.split-index}}"
        failFast: true

    - name: argosay
      script:
        image: python:3.6-alpine
        command: [python]
        source: |
          import random
          i = random.randint(1, 100)
          print(i)

    - name: test-output
      inputs:
        parameters:
          - name: task-index
      outputs:
        parameters:
          - name: task-id
            valueFrom:
              path: /mnt/out/task_id
      activeDeadlineSeconds: 3600
      retryStrategy:
        limit: 1
        retryPolicy: Always
        backoff:
          duration: "10s"
        failFast: true
      container:
        image: python:3.6-alpine
        command: [sh, -c]
        args: ["echo -n hello-{{inputs.parameters.task-index}} > /mnt/out/task_id"]
        volumeMounts:
          - name: out
            mountPath: /mnt/out
      volumes:
        - name: out
          emptyDir: {}
```

```text
nil
```

```text
nil
```

ChenRussell:
Fixes #14497 

### Motivation

This issue is difficult to reproduce. The scenario occurs when:

1. During taskResultReconciliation() execution, the Pod has either not yet finished running, or the Pod has finished but the task result cache has not been synced in time. As a result, the outputs parameters in taskResult are not set in status.nodes.

2. However, when podReconciliation() executes, the Pod has already completed, and the node is marked as Succeeded. This causes the workflow to proceed further. When the outputs of the node is required by other node, that would lead to an error.

### Modifications

Make the phase of the node unchange when the pod is completed and outputs are not set in the status.node

### Verification

local test

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1515,6 +1515,15 @@ func (woc *wfOperationCtx) assessNodeStatus(ctx context.Context, pod *apiv1.Pod,
 		woc.wf.Status.MarkTaskResultComplete(nodeID)
 	}
 
+	// If the node template has outputs Parameters/Artifacts/Result, we should not change the phase to Succeeded until the outputs are set.
+	if tmpl != nil && tmpl.Outputs.HasOutputs() && new.Outputs != nil && new.Phase == wfv1.NodeSucceeded &&
+		((tmpl.Outputs.Parameters != nil && new.Outputs.Parameters == nil) ||
+			(tmpl.Outputs.Artifacts != nil && new.Outputs.Artifacts == nil) ||
+			(tmpl.Outputs.Result != nil && new.Outputs.Result == nil)) {
+		woc.log.WithField("new.phase", new.Phase).Info("leaving phase un-changed: outputs are not yet set")
+		new.Phase = old.Phase
+	}
+
 	// if we are transitioning from Pending to a different state (except Fail or Error), clear out unchanged message
 	if old.Phase == wfv1.NodePending && new.Phase != wfv1.NodePending && new.Phase != wfv1.NodeFailed && new.Phase != wfv1.NodeError && old.Message == new.Message {
 		new.Message = ""


ChenRussell:
> as far as I'm aware,
> 
> * [the output artifact can be optional through optional flag.](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/functional/output-artifact-optional.yaml)
> * [the output parameter can be optional through default flag.](https://github.com/argoproj/argo-workflows/blob/main/examples/output-parameter.yaml#L40)
> 
> need to consider the above scenarios

Both scenarios have been covered. The wait container handles both output parameters and output artifacts when them have default flag or optional flag, so they will not be nil.

ChenRussell:
@tczhao Could you merge this PR if it looks good?

jswxstw:
What if the wait container exits abnormally? Will the workflow get stuck in `Running`?

ChenRussell:
> What if the wait container exits abnormally? Will the workflow get stuck in `Running`?

When the wait container exits abnormally, new.Phase will not be Succeeded, so it will not enter this if statement.

ChenRussell:
I noticed that the original PR [#12537](https://github.com/argoproj/argo-workflows/pull/12537) could have fixed this issue. However, since it caused problems during Argo version upgrades—specifically, when older workflows lacked the status.taskResultsCompletionStatus field, nodes would get stuck in the Running state—the follow-up PR [#13332](https://github.com/argoproj/argo-workflows/pull/13332) modified part of the logic to fix it. Unfortunately, this change rendered the original fix ineffective, so the issue still persists.

ChenRussell:
@Joibel Could you review and merge this PR if it looks good?

PR #14631, opened 2025-07-04
fix: artgc delete dir with fallback as directory. Fixes #14565 (#14631)

### LINKED ISSUE #14565: [S3] artgc Artifact Garbage Collect cannot delete directories

antoinetran:
Fixes #14565  artgc Artifact Garbage Collect cannot delete directories - fallback as directory

### Modifications

In case the delete fail because of NoSuchKey, it tries to delete a directory. Same code already existed in loadS3Artifact() https://github.com/argoproj/argo-workflows/blob/v3.6.7/workflow/artifacts/s3/s3.go#L105

### Verification

Nothing verified!!
TODO: how to?

### Documentation

Another PR for doc, link later

### DIFF
--- a/workflow/artifacts/s3/s3.go
+++ b/workflow/artifacts/s3/s3.go
@@ -277,7 +277,12 @@ func (s3Driver *ArtifactDriver) Delete(artifact *wfv1.Artifact) error {
 
 		// check suffix instead of s3cli.IsDirectory as it requires another request for file delete (most scenarios)
 		if !strings.HasSuffix(artifact.S3.Key, "/") {
-			return s3cli.Delete(artifact.S3.Bucket, artifact.S3.Key)
+			err := s3cli.Delete(artifact.S3.Bucket, artifact.S3.Key)
+			if !argos3.IsS3ErrCode(err, "NoSuchKey") {
+				return err
+			}
+			// If we get here, the error was a NoSuchKey. The key might be an s3 "directory"
+			log.Infof("S3 Delete artifact attempt as file failed, trying as directory (consider adding \"/\" to the key to avoid this failed attempt): key: %s", artifact.S3.Key)
 		}
 
 		keys, err := s3cli.ListDirectory(artifact.S3.Bucket, artifact.S3.Key)
--- a/workflow/artifacts/s3/s3.go
+++ b/workflow/artifacts/s3/s3.go
@@ -278,7 +278,7 @@ func (s3Driver *ArtifactDriver) Delete(artifact *wfv1.Artifact) error {
 		// check suffix instead of s3cli.IsDirectory as it requires another request for file delete (most scenarios)
 		if !strings.HasSuffix(artifact.S3.Key, "/") {
 			err := s3cli.Delete(artifact.S3.Bucket, artifact.S3.Key)
-			if !argos3.IsS3ErrCode(err, "NoSuchKey") {
+			if !IsS3ErrCode(err, "NoSuchKey") {
 				return err
 			}
 			// If we get here, the error was a NoSuchKey. The key might be an s3 "directory"


workflow/artifacts/s3/s3.go
```diff
@@ -277,7 +277,12 @@ func (s3Driver *ArtifactDriver) Delete(artifact *wfv1.Artifact) error {
 
 		// check suffix instead of s3cli.IsDirectory as it requires another request for file delete (most scenarios)
 		if !strings.HasSuffix(artifact.S3.Key, "/") {
-			return s3cli.Delete(artifact.S3.Bucket, artifact.S3.Key)
+			err := s3cli.Delete(artifact.S3.Bucket, artifact.S3.Key)
+			if !argos3.IsS3ErrCode(err, "NoSuchKey") {
+				return err
+			}
+			// If we get here, the error was a NoSuchKey. The key might be an s3 "directory"
+			log.Infof("S3 Delete artifact attempt as file failed, trying as directory (consider adding \"/\" to the key to avoid this failed attempt): key: %s", artifact.S3.Key)
```
tczhao, line 285: I've approved the doc, it make sense https://github.com/argoproj/argo-workflows/pull/14589#pullrequestreview-2989408523.

However, for this code change, user may already code their yamls in a way that adapts to the existing behaviour, changing here will cause some issue.
Also like you mentioned, its best to use suffix "/" if its a dir, that's the proper way rather than wasintg 1 Delete request

antoinetran: >However, for this code change, user may already code their yamls in a way that adapts to the existing behaviour, changing here will cause some issue.

I don't think so! This code only add a fallback behavior, that is consistent with the already existing loadS3Artifact (https://github.com/argoproj/argo-workflows/blob/v3.6.7/workflow/artifacts/s3/s3.go#L105).
Old code works exactly the same, except that in the case of no archive tgz + directory + key without "/", now the artgc can delete without issue. So there is no compatibility issue with old yamls (with or without "/" in keys).

Besides, we could also implement a similar code to the loadS3Artifact in fact (the "/" as a convention to avoid a double s3 request).

However, not merging is ok for me too, as long as the doc is clear and now it is merged (thanks) :)

antoinetran: The only downside of not merging is, if people haven't seen the doc in the particular section of artgc for a directory, the user will not see any error in GUI nor in Argo workflows controller. There will be a directory in s3 that is not deleted and the only relevant logs are in a artgc pod, that is deleted very quick. It is too bad we cannot at least write a warning in Argo GUI or in Argo logs.

PR #14640, opened 2025-07-09
fix: fix for feature note changed files

Joibel:
This PR fixes the feature note detector by using `tj-actions/changed-files`

Note, the output `any-added` of `tj-actions/changed-files` doesn't actually exist in a released version of the action, so I spent several confused attempts to get this to work.

Tested on my fork:
Added a feature note: https://github.com/Joibel/argo-workflows/actions/runs/16148097687
Didn't add a feature note: https://github.com/Joibel/argo-workflows/actions/runs/16148153499

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -39,20 +39,25 @@ jobs:
         with:
           fetch-depth: 50
 
-      - name: Ensure ./.features/*.md addition(s)
+      - name: Ensure ./.features/pending/*.md addition(s)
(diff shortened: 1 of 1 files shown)


Joibel:
> Hmm this action is maintained by just one person so I am a bit ehh about this but I guess its locked so meh I suppose.

We already use the action in our ci workflow which is more important, so figured it was better than our hand-rolled version.

shuangkun:
Great！Thanks

PR #14642, opened 2025-07-09
chore: remove tracing logging level

Joibel:
To reduce the number of log levels down to the standard set (debug, info, warn, error), this commit removes trace level and maps it to debug. This was an undocumented level, and used once.

Transition the one log line changed to be structured log.


isubasinghe (approved):

Joibel:
> I almost forgot, this change should probably force logrus to use Debug when 'trace' is passed. This is the `SetLogLevel` function I think

This is why `ParseLevel` still understands `trace`. `SetGlobalLevel` and `convertLevel` take a `Level` which can't come from code because Trace is removed as a Level - so I think it could only come from `ParseLevel`

isubasinghe:
This isn't complete unfortunately, logrus has its own ParseLevel function, so `SetLogLevel` needs to be updated to cooerce `trace` to `debug`.

Joibel:
> This isn't complete unfortunately, logrus has its own ParseLevel function, so `SetLogLevel` needs to be updated to cooerce `trace` to `debug`.

Like this change?

isubasinghe:
Yes, thanks

PR #14646, opened 2025-07-11
fix: make loop index parsing more robust

isubasinghe:
Fixes #TODO


isubasinghe:
Yeah this code  sucks

PR #14655, opened 2025-07-14
docs(cli): add usage examples for argo auth token command Fixes argoproj#11898

miinsun:
Signed-off-by: 민선 (minnie) <61786235+miinsun@users.noreply.github.com>

Fixes #TODO


Jack-R-lantern:
It seems more appropriate to use a related keyword than to use a fix keyword because there is something unresolved within the issue.

also remove the fixes keyword of the commit message

miinsun:
Thanks, @Jack-R-lantern. I replaced it with Refs #11898 so that the issue remains open for further discussion.

Joibel:
Thank you for your contribution!

PR #14665, opened 2025-07-21
docs(cli): add example field to argo cluster-template lint. Fixes #11898

SinnoLn:
Related to #11898

### Motivation:

The help output for the `argo cluster-template lint` command lacks practical examples, making it difficult for new users to understand its usage.

### Modifications:

- Added an `Example` field to `cmd/argo/commands/clustertemplate/lint.go`.
- Ran `make codegen` to update the corresponding CLI documentation.

### Verification:
run argo cluster-template lint --help


Jack-R-lantern:
It seems more appropriate to use a related keyword than to use a fix keyword because there is something unresolved within the issue.

SinnoLn:
Thanks for the suggestion, @Jack-R-lantern! 
I didn't realize `Fixes` would auto-close the issue. I've updated the description as you recommended.

Joibel (approved):

PR #14676, opened 2025-07-22
fix: pagination during list workflows. Fixes: #14672

### LINKED ISSUE #14672: Workflow controller startup fails in large-scale scenarios
### Pre-requisites

In large-scale scenarios (8000+ workflows), the workflow controller failed to restart, causing the system to crash.
```
time="2025-07-21T07:32:26.636Z" level=info msg=Plugins executorPlugins=true
time="2025-07-21T07:32:28.914Z" level=fatal msg="rpc error: code = ResourceExhausted desc = grpc: trying to send message larger than max (2409346693 vs. 2147483647)"
```

### Version(s)

3.5.13

```YAML
8000+ running workflows
```

```text
```

```text
```

shuangkun:
pagination during list workflows, avoid rpc error.

Fixes #14672

### Verification

Before: rpc error
```
time="2025-07-21T07:32:26.636Z" level=info msg=Plugins executorPlugins=true
time="2025-07-21T07:32:28.914Z" level=fatal msg="rpc error: code = ResourceExhausted desc = grpc: trying to send message larger than max (2409346693 vs. 2147483647)"
```

After fix:
```
time="2025-07-23T07:33:30.567Z" level=info msg="Watching task results" labelSelector="!workflows.argoproj.io/controller-instanceid,workflows.argoproj.io/workflow"
time="2025-07-23T07:33:30.567Z" level=info msg=Plugins executorPlugins=true
time="2025-07-23T07:36:57.501Z" level=info msg="Successfully listed 61052 workflows"
time="2025-07-23T07:36:57.505Z" level=info msg="Manager initialized successfully"
```

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -445,8 +445,8 @@ func (wfc *WorkflowController) initManagers(ctx context.Context) error {
 	if req != nil {
 		labelSelector = labelSelector.Add(*req)
 	}
-	listOpts := metav1.ListOptions{LabelSelector: labelSelector.String()}
-	wfList, err := wfc.wfclientset.ArgoprojV1alpha1().Workflows(wfc.GetManagedNamespace()).List(ctx, listOpts)
+
+	wfList, err := wfc.listWorkflowsPaginated(ctx, labelSelector)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


Joibel:
This looks similar to #14373, @isubasinghe?

isubasinghe:
@Joibel the code  is similar but mine works on unstructured, this is a requirement since we use an unstructured informer. I think this code is fine as a result. I do think the overall logic could ideally be refactored though, its almost the exact same.

shuangkun:
> @Joibel the code is similar but mine works on unstructured, this is a requirement since we use an unstructured informer. I think this code is fine as a result. I do think the overall logic could ideally be refactored though, its almost the exact same.

I tried refactoring, but the conversion between the two types of unstructured.UnstructuredList and wfv1.WorkflowList increased the complexity.   

One is at the NS level, and the other is global.

PR #14677, opened 2025-07-23
chore: update devcontainer golang version

djanjic:
### Motivation
There's a mismatch between the Golang version defined by `gomod` and the version used in the `devcontainer`. This change fixes that. 

### Modifications
- Updated `.devcontainer/builder/devcontainer.json` to use Golang `1.24.4`
- used `devcontainer` CLI to update "lock" file

### Verification
- tested manually that the devcontainer is building using `make devcontainer-build`


Joibel (approved):
LGTM thank you

PR #14696, opened 2025-07-25
fix: GC realtime metrics after workflow completed. Fixes #14694

### LINKED ISSUE #14694: Real-time metrics disappear while workflow is running when metricsTTL is < workflow run time
### Pre-requisites

#13265 refactored the metric module, but it did't include the fix from #12830  and consequently #12790 was reintroduced.

### Version(s)

v3.6.7

```YAML
no workflow
```

```text
```

```text
```

jswxstw:
Fixes #14694

### Motivation

Realtime metrics may be cleaned up, even if the workflow is still running.

### Modifications

- Adding boolean value: `complete` in metric to indicate whether the metric is completed.
- Adding two functions: `CompleteRealtimeMetricsForWfUID` and `DeleteRealtimeMetricsForWfUID` to clearly distinguish whether to directly clean realtime metrics or only mark them as completed.
- Enable realtime metric gc only after its workflow is completed.

### Verification

local test and UT

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -1030,7 +1030,7 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context) 
 		DeleteFunc: func(obj interface{}) {
 			wf, ok := obj.(*unstructured.Unstructured)
 			if ok { // maybe cache.DeletedFinalStateUnknown
-				wfc.metrics.StopRealtimeMetricsForWfUID(string(wf.GetUID()))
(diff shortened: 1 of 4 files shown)


Joibel (commented):
To avoid a regression again, do you think you could add a test similar to TestRealtimeMetricGC but to prove this is also working?

Otherwise LGTM.

jswxstw:
@Joibel `TestRealtimeMetricGC` is the one I added earlier to verify this issue, but it also had bugs which I fixed in this PR.

Joibel (approved):
Thanks for fixing this

PR #14703, opened 2025-07-28
fix: prevent thundering herd on cache save/load. Fixes #14701

### LINKED ISSUE #14701: Thundering herd problem when updating cache
### Pre-requisites

If a workflow uses the same key a large number of times, it is possible for us to arrive at a thundering herd problem where we all try to update the configmap at the same time. 

The code for Load() and Save() have locks in place which in theory should prevent this issue but doesn't in practice (I have observed this in a workflow). I suspect that this is because it is possible for `NewConfigMapCache` to be racy. 

The fix for this should a) add a lock to `NewConfigMapCache` and b) retry on conflict.

### Version(s)

:latest

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #14701

### Motivation
Empirical data shows a thundering herd problem on configmaps when workflows are highly parallel. 

### Modifications

We do not gain exclusive access to configmaps on load/save operations despite the current code suggesting that we do. I believe this is a result of many `MemoizationCache` objects being created. This change ensures that we obtain a lock on the cacheFactory before a call to `GetCache`, this should ensure only one cache. 

I've also added a retry block on save/load operations, treating conflict errors as transient errors. 
The decision to not add to the TransientError function was deliberate. I think conflict errors being treated as transient
errors are somewhat context dependent, hence the omission. 

This was a small fix, I think the ideal fix for these sort of issues is much more work. We should have a priortised queue along with batching for these type of updates to k8s objects. It doesn't really make sense to perform `Load` 10000 times because 10000 parallel nodes used the cache, we can also batch updates to a configmap since the data is partitioned by key. This fix was what I initially wanted to attempt when I found this problem, but I found it to be:
a) too time consuming, realistically take a week or more
b) hard to get right, the merge strategy becomes complicated when duplicate values are present for the same key. 

So I gave up on this idea.

### Verification

None needed, no semantic change to how configmaps work.

### DIFF
--- a/workflow/controller/cache/configmap_cache.go
+++ b/workflow/controller/cache/configmap_cache.go
@@ -74,7 +74,7 @@ func (c *configMapCache) Load(ctx context.Context, key string) (*Entry, error) {
 		Steps:    5,
 		Cap:      30 * time.Second,
 	}, func(err error) bool {
-		return argoerr.IsTransientErr(ctx, err) || (apierr.IsConflict(err) && strings.Contains(err.Error(), ""))
+		return argoerr.IsTransientErr(ctx, err) || (apierr.IsConflict(err) && strings.Contains(err.Error(), "Operation cannot be fulfilled on configmaps"))
 	}, func() error {
 		var innerErr error
 		entry, innerErr = c.load(ctx, key)
--- a/workflow/controller/cache/configmap_cache.go
+++ b/workflow/controller/cache/configmap_cache.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
@@ -74,7 +73,7 @@ func (c *configMapCache) Load(ctx context.Context, key string) (*Entry, error) {
 		Steps:    5,
 		Cap:      30 * time.Second,
 	}, func(err error) bool {
-		return argoerr.IsTransientErr(ctx, err) || (apierr.IsConflict(err) && strings.Contains(err.Error(), "Operation cannot be fulfilled on configmaps"))
+		return argoerr.IsTransientErr(ctx, err) || apierr.IsConflict(err)
 	}, func() error {
 		var innerErr error
 		entry, innerErr = c.load(ctx, key)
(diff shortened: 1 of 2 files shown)


workflow/controller/cache/configmap_cache.go
```diff
@@ -61,6 +66,24 @@ func (c *configMapCache) validateConfigmap(ctx context.Context, cm *apiv1.Config
...
 func (c *configMapCache) Load(ctx context.Context, key string) (*Entry, error) {
+	var entry *Entry
+	err := retry.OnError(kwait.Backoff{
+		Duration: time.Second,
+		Factor:   2,
+		Jitter:   0.1,
+		Steps:    5,
+		Cap:      30 * time.Second,
+	}, func(err error) bool {
+		return argoerr.IsTransientErr(ctx, err) || (apierr.IsConflict(err) && strings.Contains(err.Error(), ""))
```
Joibel: What's this strings.Contains() trying to do?

isubasinghe: Sigh I hate stringly typed stuff,this was meant to do the exact same as the strings.Contain before

workflow/controller/cache/configmap_cache.go
```diff
@@ -107,14 +128,28 @@ func (c *configMapCache) Load(ctx context.Context, key string) (*Entry, error) {
...
+	err := retry.OnError(kwait.Backoff{
+		Duration: time.Second,
+		Factor:   2,
+		Jitter:   0.1,
+		Steps:    5,
+		Cap:      30 * time.Second,
+	}, func(err error) bool {
+		return argoerr.IsTransientErr(ctx, err) || (apierr.IsConflict(err) && strings.Contains(err.Error(), "Operation cannot be fulfilled on configmaps"))
+	}, func() error {
+		var innerErr = c.save(ctx, key, nodeID, value)
```
Joibel: nit: drop the `var` here and use `:=`

workflow/controller/cache/configmap_cache.go
```diff
@@ -61,6 +66,24 @@ func (c *configMapCache) validateConfigmap(ctx context.Context, cm *apiv1.Config
...
 func (c *configMapCache) Load(ctx context.Context, key string) (*Entry, error) {
+	var entry *Entry
+	err := retry.OnError(kwait.Backoff{
+		Duration: time.Second,
+		Factor:   2,
+		Jitter:   0.1,
+		Steps:    5,
+		Cap:      30 * time.Second,
+	}, func(err error) bool {
+		return argoerr.IsTransientErr(ctx, err) || (apierr.IsConflict(err) && strings.Contains(err.Error(), "Operation cannot be fulfilled on configmaps"))
```
Joibel: Properly reviewing this now, why would other IsConflicts not want a retry - I'd prefer we drop the fallible and possibly breaking in the future string match here and in the other place and just check `IsConflict`

Joibel (commented):
I'm mostly happy with this

workflow/controller/cache/configmap_cache.go
isubasinghe: Well it's more so that I don't expect other IsConflict errors here. That being said the string matching is how all of our k8s error matching works, but happy to drop the string matching.

PR #14706, opened 2025-07-28
chore: fully structured logging

### LINKED ISSUE #2308: Migrate toward structured logging
Some parts of the code already have structured logging -- for example:

https://github.com/argoproj/argo/blob/214c451535ebeb6e68f1599c2c0a4a4d174ade25/persist/sqldb/workflow_archive.go#L63-L64

But most don't. Let's migrate the rest.

#2307

Joibel:
Follow up to #14680
Fixes #2308 

### Motivation

Make all logging structured for better telemetry

### Modifications

Remove all the printf style methods in the logger, and fix the fallout of doing so by structuring all remaining logs.

There is still a small shim in event_recorder_manager which produces unstructured logs from an external component, I didn't look in to fixing that.

### Verification

CI mostly

### Documentation

This doesn't use the common telemetry attributes yet, that would be a further thing to move to, so these became more well documentable.

### DIFF
--- a/workflow/cron/operator.go
+++ b/workflow/cron/operator.go
@@ -41,7 +41,6 @@ const (
 
 type cronWfOperationCtx struct {
 	// CronWorkflow is the CronWorkflow to be run
-	name            string
 	cronWf          *v1alpha1.CronWorkflow
 	wfClientset     versioned.Interface
 	wfClient        typed.WorkflowInterface
@@ -63,7 +62,6 @@ func newCronWfOperationCtx(ctx context.Context, cronWorkflow *v1alpha1.CronWorkf
 ) *cronWfOperationCtx {
 	log := logging.RequireLoggerFromContext(ctx)
 	return &cronWfOperationCtx{
-		name:            cronWorkflow.Name,
 		cronWf:          cronWorkflow,
 		wfClientset:     wfClientset,
 		wfClient:        wfClientset.ArgoprojV1alpha1().Workflows(cronWorkflow.Namespace),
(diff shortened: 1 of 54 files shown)


workflow/cron/operator.go
```diff
@@ -41,7 +41,6 @@ const (
 
 type cronWfOperationCtx struct {
 	// CronWorkflow is the CronWorkflow to be run
-	name            string
```
isubasinghe, line 44: Is this on purpose?

workflow/cron/operator.go
```diff
@@ -95,7 +93,7 @@ func (woc *cronWfOperationCtx) Run() {
 func (woc *cronWfOperationCtx) run(ctx context.Context, scheduledRuntime time.Time) {
 	defer woc.persistUpdate(ctx)
 
-	woc.log.Infof(ctx, "Running %s", woc.name)
+	woc.log.Info(ctx, "Running")
```
isubasinghe, line 96: We do not log the name anymore

workflow/cron/operator.go
Joibel: Yes.
Name is set to `cronWorkflow.Name`, but we're also storing `cronWorkflow`. 
I was inspecting the logging to work out when and why we were using `woc.name` vs `woc.cronWf.Name`, and then noticed this and thought it best just to remove the duplication.

workflow/cron/operator.go
Joibel: The logger at this point has a permanent field of `workflow` set to this value (line 73), so no.

PR #8028, opened 2022-02-28
feat: Remove plugin Kube API access by default.

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>

Fixes #TODO

PR #14711, opened 2025-07-30
chore: Add support for mounting default workflow ServiceAccount in sidecar as a configurable option

ntny:
In https://github.com/argoproj/argo-workflows/pull/8028  mounting the default ServiceAccount token for Kubernetes API access was disabled. This means that to allow sidecars (like the executor plugin) to access the Kubernetes API, one must create and configure a separate ServiceAccount and explicitly mount its token.

This PR introduces a configuration option to mount the workflow's default ServiceAccount token into the executor plugin sidecar.
In my case, the agent sidecar needs to interact with the Kubernetes API, and I’d prefer to run it under the workflow’s ServiceAccount, avoiding the need to create and manage an additional one.

@alexec , do you have any concerns about this change?


Joibel:
This is not a `chore` it is a feature/`feat` type PR.

Joibel:
I have quite a lot of concern over adding this as a feature. I don't know how many people actually need it, it can be done other ways, and it encourages worse security practice.

ntny:
Hi, @Joibel! Sorry for the late reply.
I generally agree with your concerns. When I was looking into how the ServiceAccount is mounted in the executor plugin, it seemed like this feature could be useful. But after weighing the pros and cons, I realized that the risks outweigh the potential convenience.

PR #14713, opened 2025-07-30
fix: cron patch use defaultRetry. Fixes #14712

### LINKED ISSUE #14712: exec.resource/cron k8s ops should use defaultRetry instead of defaultBackoff
### Pre-requisites

all other k8s ops in argo are using defaultRetry, we should make all the behaviour consistent

[defaultRetry ](https://github.com/argoproj/argo-workflows/blob/680ee6c2223e3ba0403fbc86bd3644fa98177a4f/util/retry/retry.go#L20)is a configurable retry parameter
where [defaultBackoff ](https://pkg.go.dev/k8s.io/client-go/util/retry#pkg-variables)is a static parameter

cron
https://github.com/argoproj/argo-workflows/blob/680ee6c2223e3ba0403fbc86bd3644fa98177a4f/workflow/cron/operator.go#L183

resource
https://github.com/argoproj/argo-workflows/blob/680ee6c2223e3ba0403fbc86bd3644fa98177a4f/workflow/executor/resource.go#L44

### Version(s)

latest, 680ee6c2223e3ba0403fbc86bd3644fa98177a4f

```YAML
N/A
```

```text
```

```text
```

tczhao:
Fixes #14712

### Motivation

allow cron patch to use `TRANSIENT_ERROR_PATTERN` env var
just like how all other codes that interact with the k8s api

### DIFF
--- a/workflow/cron/operator.go
+++ b/workflow/cron/operator.go
@@ -180,7 +180,7 @@ func (woc *cronWfOperationCtx) patch(ctx context.Context, patch map[string]inter
 		woc.log.WithError(err).Error(ctx, "failed to marshall cron workflow status.active data")
 		return
 	}
-	err = waitutil.Backoff(retry.DefaultRetry, func() (bool, error) {
+	err = waitutil.Backoff(retry.DefaultRetry(ctx), func() (bool, error) {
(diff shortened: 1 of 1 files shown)


Joibel (approved):

Joibel:
LGTM thanks

PR #14721, opened 2025-08-01
fix: added createdAfter query support

isubasinghe:
Fixes #TODO

### DIFF
--- a/persist/sqldb/migrate.go
+++ b/persist/sqldb/migrate.go
@@ -23,6 +23,7 @@ func Migrate(ctx context.Context, session db.Session, clusterName, tableName str
     workflow text,
     startedat timestamp default CURRENT_TIMESTAMP,
     finishedat timestamp default CURRENT_TIMESTAMP,
+		creationtimestamp timestamp default CURRENT_TIMESTAMP,
 
     primary key (id, namespace)
 )`),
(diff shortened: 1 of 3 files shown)


persist/sqldb/migrate.go
```diff
@@ -217,5 +218,14 @@ func Migrate(ctx context.Context, session db.Session, clusterName, tableName str
 			sqldb.Postgres: sqldb.AnsiSQLChange(`drop index argo_archived_workflows_i4`),
 		}),
 		sqldb.AnsiSQLChange(`create index argo_archived_workflows_i4 on argo_archived_workflows (clustername, startedat)`),
+		// add creationtimestamp column to argo_archived_workflows table
+		sqldb.AnsiSQLChange(`alter table argo_archived_workflows add column creationtimestamp timestamp`),
+		sqldb.AnsiSQLChange(`update argo_archived_workflows set creationtimestamp = startedat where creationtimestamp is null`),
+		sqldb.ByType(dbType, sqldb.TypedChanges{
+			sqldb.MySQL:    sqldb.AnsiSQLChange(`alter table argo_archived_workflows modify column creationtimestamp timestamp not null default CURRENT_TIMESTAMP`),
```
Joibel, line 224: CURRENT_TIMESTAMP is not what I expected here, but I don't have a better value, so fine.

persist/sqldb/migrate.go
```diff
@@ -23,6 +23,7 @@ func Migrate(ctx context.Context, session db.Session, clusterName, tableName str
     workflow text,
     startedat timestamp default CURRENT_TIMESTAMP,
     finishedat timestamp default CURRENT_TIMESTAMP,
+		creationtimestamp timestamp default CURRENT_TIMESTAMP,
```
Joibel: nit: Github shows this as indented badly

Joibel (commented):
My main issue here is why did we implement created after at all, rather than started at instead. It feels like it's too late to fix that, but it makes me sad.

persist/sqldb/migrate.go
JPZ13: Fixed

persist/sqldb/migrate.go
JPZ13: C'est la vie, cher!

JPZ13:
@terrytangyuan - can you give this a review/approve? I updated for @Joibel's comments while he's on vacation. It looks like there's a flaky test that needs re-running

persist/sqldb/workflow_archive.go
```diff
@@ -313,6 +309,20 @@ func startedAtFromClause(from time.Time) db.Cond {
...
+	}
+	return db.Cond{}
+}
+
+func finishedBeforeClause(finishedBefore time.Time) db.Cond {
+	if !finishedBefore.IsZero() {
+		return db.Cond{"finishedat <=": finishedBefore}
+	}
+	return db.Cond{}
+}
```
terrytangyuan, line 324: non-blocking: Are "="s intentional?

terrytangyuan:
> add support for live workflows (this is impossible without CRD changes or runtime checks).

Maybe a follow-up but we might want to handle the unsupported case with a warning or documentation

persist/sqldb/workflow_archive.go
isubasinghe: Yes

PR #14724, opened 2025-08-01
feat: Add support for creating a configmap semaphore config using CLI. Fixes #14671

### LINKED ISSUE #14671: Add support for creating a configmap semaphore config using CLI
# Summary
Currently, the user must manually create a ConfigMap with the semaphore configuration in order to use it. This process can be difficult for users who are not familiar with Kubernetes, and it also requires the user to have the necessary permissions. Additionally, the user must be careful about the format of the ConfigMap.

By adding support for creating the configuration via the CLI, users can more easily add new semaphore configurations. This would also simplify future support for creating database-backed semaphore configurations, and eventually, for adding UI support as well.

## Use Cases
- creating a new configmap semaphore configuration using CLI
- make semaphore configuration creation less error-prone

---

djanjic:
Fixes #14671

### Motivation
This adds support for creating configmap sync configuration using CLI.

### Modifications
- added new endpoints to accomodate this change
- added CLI support for creating configmap sync config.
- added tests

### Verification
- manual testing
- e2e tests
- unit tests

### Documentation
- Documentation is automatically generated for new CLI commands


server/sync/sync_server.go
```diff
@@ -0,0 +1,155 @@
...
+	return &syncpkg.SyncLimitResponse{
+		Name:      cm.Name,
+		Namespace: cm.Namespace,
+		Key:       req.Key,
+		SizeLimit: req.SizeLimit,
+	}, nil
+}
+
+func (s *syncServer) DeleteSyncLimit(ctx context.Context, req *syncpkg.DeleteSyncLimitRequest) (*syncpkg.DeleteSyncLimitResponse, error) {
+	fmt.Printf("Deleting sync limit for ConfigMap %s in namespace %s with key %s\n", req.Name, req.Namespace, req.Key)
```
eduardodbr: You can remove this print or replace it proper logging

cmd/argo/commands/sync/configmap/get.go
```diff
@@ -0,0 +1,62 @@
...
+	key string // --key
+}
+
+func NewGetCommand() *cobra.Command {
+	var cliGetOpts = cliGetOpts{}
+	command := &cobra.Command{
+		Use:     "get",
+		Short:   "Get a configmap sync limit",
+		Args:    cobra.ExactArgs(1),
+		Example: `argo sync get configmap my-cm --key my-key`,
```
eduardodbr:
```suggestion
		Example: `argo sync configmap get my-cm --key my-key`,
```
the command and the method are switched

server/sync/sync_server.go
```diff
@@ -0,0 +1,155 @@
...
+	cm, err := configmapGetter.Create(ctx, cm, metav1.CreateOptions{})
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return s.updateSyncLimit(ctx, &syncpkg.UpdateSyncLimitRequest{
+				Name:      req.Name,
+				Namespace: req.Namespace,
+				Key:       req.Key,
+				SizeLimit: req.SizeLimit,
+			}, false)
+		}
```
eduardodbr: A Create doing an Update violates some API principles, is this common in the current codebase?

server/sync/sync_server.go
```diff
@@ -0,0 +1,155 @@
...
+	kubeClient := auth.GetKubeClient(ctx)
+
+	configmapGetter := kubeClient.CoreV1().ConfigMaps(req.Namespace)
+
+	cm, err := configmapGetter.Get(ctx, req.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, sutils.ToStatusError(err, codes.Internal)
+	}
+
+	delete(cm.Data, req.Key)
```
eduardodbr, line 149: if the deleted key is the only key in the Configmap, should the CM also be deleted? otherwise the CM will be created with no data

docs/cli/argo_sync_configmap_get.md
```diff
@@ -0,0 +1,61 @@
...
+Get a configmap sync limit
+
+```
+argo sync configmap get [flags]
+```
+
+### Examples
+
+```
+argo sync get configmap my-cm --key my-key
```
eduardodbr:
```suggestion
argo sync configmap get my-cm --key my-key
```

server/sync/sync_server.go
isubasinghe: Yeah I agree with this, we should fail on any error really.

server/sync/sync_server.go
isubasinghe: I think its fine to leave it around

isubasinghe (changes requested):
We shouldn't attempt an update, apart from this issue, it looks good.

server/sync/sync_server.go
djanjic: This change should create a cm with a defined key/value for the semaphore. If cm already exists, we are just trying to add another field to that configmap, which could have multiple semaphores for different workflows. At least, that's the idea.

Joibel: I mostly agree with @djanjic, it's not completely pure:

I think this should:
* If the CM does not exist, create it (we probably all agree on this)
* If the CM does exist, but the key does not, create the key within in (by performing a CM update call).
* If the CM does exist and the key also does, fail.

PR #14726, opened 2025-08-01
feat(workflowtemplate): evaluate expressions in workflow template parameters

k3rnL:
### Motivation
Currently, the default values for parameters in a `WorkflowTemplate` are static. This limits the ability to create dynamic and reusable templates where default configurations might depend on other parameters or dates. This change introduces the capability to use expressions for default parameter values, allowing for more powerful and flexible template definitions. For instance, a template-level parameter could have a default value that is derived from a global parameter, enhancing modularity and configurability.

A use case in my office is to provide date parameters to backfill cron jobs, the start_date and end_date parameters are filled with the date of the day and now() - 24h. Aloowing the company's operators to start a backfill job in one click or to change the date with a prefilled format. Avoiding human errors.

### Modifications
- Updated the `GetWorkflowTemplate` method in the `WorkflowTemplateServer`.
- Added logic to detect and evaluate expressions within the `default` field of parameters. Expressions are identified by the `{{= ... }}` syntax.
- The evaluation happens in two distinct phases:
    1. Global parameters (`spec.arguments.parameters`) are evaluated first. Their values are made available for subsequent steps.
    2. Template-level parameters (`spec.templates[].inputs.parameters`) are evaluated next. They can reference the evaluated global parameters through the `workflow.parameters` map within their expressions.

- The result of a successfully evaluated expression is populated into the parameter's `value` field, making the computed default visible in the API response.

### Verification
1. Create a `WorkflowTemplate` containing global and template-level parameters with default values using `{{= ... }}` expressions. Ensure some template-level parameters reference global ones.
2. Use `argo template get` or call the API endpoint for the created template. `GetWorkflowTemplate`
3. Verify that the output shows the `value` field correctly populated with the evaluated result of the expressions for both global and template-level parameters.

Or, from the UI you can see in the submit form the evaluated values:
image

You can test using a template like so:
```yaml
metadata:
  name: omniscient-whale
  namespace: argo
spec:
  templates:
    - name: whalesay-template
      inputs:
        parameters:
          - name: message
            default: "{{workflow.parameters.test}}"
      outputs: {}
      metadata: {}
      container:
        name: ""
        image: docker/whalesay
        command:
          - cowsay
        args:
          - "{{inputs.parameters.message}}"
        resources: {}
  entrypoint: whalesay-template
  arguments:
    parameters:
      - name: test
        default: "{{=now().Format('2006-01-02 15:04:05')}}"
```

### Documentation
The documentation for this will need to be updated to include this new feature. This includes: `WorkflowTemplates`
- Explaining the `{{= ... }}` syntax for expression-based default values.
- Providing examples of how to define and use global and template-level parameters with expressions.
- Documenting how template-level parameters can reference global parameters within their expressions using the `workflow.parameters` context.

### AI
Used for self-review

## Summary by CodeRabbit

* **New Features**
  * Workflow and cluster workflow template retrieval now evaluates expression-based parameter defaults.
  * Dependent defaults resolve in order, while explicit parameter values take precedence.
  * Unresolved expressions remain unchanged, and evaluated results populate the parameter value.

* **Bug Fixes**
  * Invalid expressions now return a clear internal error instead of producing invalid results.

* **Documentation**
  * Added guidance and examples for expression-based parameter defaults.


k3rnL:
Just a reminder, can I have a reviewer ?

k3rnL:
@isubasinghe Can you review this feature ? Thanks :)

.features/pending/expr-eval-in-template.md
```diff
@@ -0,0 +1,37 @@
+Description: Evaluate expressions in workflow template parameters.
+Author: [Erwan Daniel](https://github.com/k3rnL)
+Component: UI
+Issues: 14726
+
+This feature is useful when you want to make your `WorkflowTemplates` more dynamic.
```
tico24: Nothing here says `{{= }}` is what makes a default an expression rather than ordinary substitution. This note gets compiled into `docs/new-features.md` and read on its own, so the distinction has to be in the prose. At the moment a reader infers it by diffing the two YAML defaults against each other.

Worth stating: a default wrapped in `{{= }}` is evaluated as an expression, the result is written to the parameter's `value` while `default` keeps the expression, and it happens when the API server serves a Get rather than when the workflow runs.

.features/pending/expr-eval-in-template.md
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  name: omniscient-whale
+  namespace: argo
+spec:
+  templates:
+    - name: whalesay-template
+      inputs:
+        parameters:
+          - name: message
+            default: "{{workflow.parameters.test}}"
```
tico24: This line is ordinary substitution and resolves at runtime, so it's right as it stands.

It does mean the note never demonstrates the second half of the change: a template input default can itself be `{{= }}`, with the evaluated globals in scope as `workflow.parameters.*`. That's the part nobody will guess from the example. Either show it here, or say it in the prose above.

.features/pending/expr-eval-in-template.md
```diff
@@ -0,0 +1,37 @@
+Description: Evaluate expressions in workflow template parameters.
+Author: [Erwan Daniel](https://github.com/k3rnL)
+Component: UI
+Issues: 14726
+
+This feature is useful when you want to make your `WorkflowTemplates` more dynamic.
+For example, you can prefill your submit parameters with the date of today.
+Making it easier for operators to start a job and avoid mistakes.
```
tico24: Two edges worth a line each. ClusterWorkflowTemplate gets the same treatment, since the PR changes both servers, but the note only mentions WorkflowTemplates. And an expression that fails to evaluate is dropped silently: `if err == nil`, so the parameter keeps its raw default and nothing is logged. If that's deliberate, say so, because a typo will otherwise look like the feature not working at all. If it isn't, that's one for the approvers.

.features/pending/expr-eval-in-template.md
```diff
@@ -0,0 +1,37 @@
...
+
+ex.
```
tico24: `ex.` reads oddly here. `For example:` works, or drop the line since the code block follows on from the sentence anyway.

While you're in there: line 9 is a fragment hanging off line 8. Something like `For example, you can prefill submit parameters with today's date, which makes it easier for operators to start a job without typos.` folds it back into one sentence.

.features/pending/expr-eval-in-template.md
```diff
@@ -0,0 +1,37 @@
...
+    - name: whalesay-template
+      inputs:
+        parameters:
+          - name: message
+            default: "{{workflow.parameters.test}}"
+      outputs: {}
+      metadata: {}
+      container:
+        name: ""
+        image: docker/whalesay
```
tico24: Docs examples have mostly moved to `argoproj/argosay:v2`: 8 files under `docs/` use it against 2 still on `docker/whalesay`. Not a blocker, just more consistent with the rest of the docs. Plenty of test fixtures still use whalesay, so treat this as a docs point rather than a repo-wide one.

The `outputs: {}`, `metadata: {}`, `resources: {}` and `name: ""` lines look like round-trip noise from an API response. Dropping them makes the example shorter and it still applies cleanly.

tico24 (commented):
Docs only: `.features/pending/expr-eval-in-template.md`. The other four files are for @argoproj/argo-workflows-approvers.

The example works, but it exercises one of the two things this PR does, and the note never says what the syntax is.

`GetWorkflowTemplate` evaluates in two passes: globals under `spec.arguments.parameters`, then template inputs, with the evaluated globals exposed to the second pass as `workflow.parameters.*`. The example uses the first pass on `test`, then reaches that value through ordinary substitution on `message`, which resolves at runtime. What it never shows is a template input default that is itself an expression, which is the case the new `workflow.parameters` scope exists for.

.features/pending/expr-eval-in-template.md
k3rnL: Updated in 5be3b2f. The note now explicitly states that {{= ... }} marks an expression default, that evaluation happens when the API server handles the Get request, and that the result is returned in value while default retains the original expression.

.features/pending/expr-eval-in-template.md
k3rnL: Updated in 5be3b2f. The example now makes the template input default an expression itself and references the first-pass result through workflow.parameters['run-date'], so both evaluation passes are demonstrated.

.features/pending/expr-eval-in-template.md
k3rnL: Updated in 5be3b2f. The note now covers both WorkflowTemplate and ClusterWorkflowTemplate. I also removed the direct expr.Eval / if err == nil handling: both servers now use a shared helper backed by the Argo template evaluator. Allowed unresolved expressions leave value unset, preserve the raw default, and use the existing debug logging; invalid expression syntax is returned as an error. Tests cover resolved typed values, unresolved expressions, dependent unresolved defaults, and invalid syntax.

.features/pending/expr-eval-in-template.md
k3rnL: Updated in 5be3b2f. I removed ex. and rewrote the fragmented text as one sentence introducing the example.

.features/pending/expr-eval-in-template.md
k3rnL: Updated in 5be3b2f. The example now uses argoproj/argosay:v2, and the round-trip-only empty fields were removed.

server/utils/parameter_defaults.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	for i := range spec.Templates {
+		for j := range spec.Templates[i].Inputs.Parameters {
+			parameter := &spec.Templates[i].Inputs.Parameters[j]
+			value, resolved, _, err := evaluateParameterDefault(ctx, parameter.Default, replaceMap)
+			if err != nil {
+				return fmt.Errorf("failed to evaluate default for template %q input parameter %q: %w", spec.Templates[i].Name, parameter.Name, err)
+			}
+			if resolved {
+				parameter.Value = v1alpha1.AnyStringPtr(value)
```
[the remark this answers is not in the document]
k3rnL: Fixed in 3cf8d26. Explicit workflow parameter values now take precedence over defaults and are seeded into workflow.parameters; explicit template input values also skip default evaluation. Added regression coverage for both cases, including a template default that reads the explicit workflow value.

Issue #14679, opened 2025-07-23
Keep orphans option when deleting CronWorkflows in UI


bencoates08:
# Summary

When deleting a CronSchedule in the Workflows UI, any Workflows spawned from that CronSchedule will also be deleted (garbage collected by Kubernetes). This behaviour isn't obvious when a user goes to delete a Cron workflow and may lead to them losing data.

I propose that an option should be added to the delete confirmation modal that allows users to keep any orphaned Workflows from the CronSchedule. Alternatively, improved messaging in the modal that states any Workflows will be lost.

## Use Cases

When a user wishes to delete a CronSchedule but keep any of its associated Workflows

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

miinsun:
Hi @bencoates08 . I’ve submitted a small UI PR that adds a warning to the delete confirmation modal indicating that deleting a CronWorkflow will also delete any Workflows it created.

As a follow-up, I’m planning a separate PR to add an option in the modal—e.g., a checkbox labeled “Keep orphaned Workflows”—so users can delete the CronWorkflow while retaining its spawned Workflows. By the way, feedback on the wording is welcome.

bencoates08:
Thanks @miinsun, I've left a small comment on your MR about this only affecting non-archived workflows.

Joibel:
The requested opt-out part of this is still open, and it is a nicely scoped UI-only task for a new contributor, because the server side already supports it: `DeleteCronWorkflow` forwards `deleteOptions` to Kubernetes (server/cronworkflow/cron_workflow_server.go), and `deleteOptions.propagationPolicy` is already a documented query parameter on `DELETE /api/v1/cron-workflows/{namespace}/{name}`.

Suggested approach:
1. In `ui/src/cron-workflows/cron-workflow-details.tsx`, pass a render function to `popup.confirm` instead of a plain string and include a "Keep Workflows created by this CronWorkflow" checkbox. See `renderSuspendNodeOptions` in `workflow-details.tsx` for the pattern.
2. In `ui/src/shared/services/cron-workflow-service.ts`, let `delete()` take an optional flag and append `?deleteOptions.propagationPolicy=Orphan` when it is set.

This won't affect archived workflows.

JerryNee:
[...]

PR #14727, opened 2025-08-01
feat(ui): add warning message when deleting CronWorkflow about orphaned Workflows

miinsun:
Refs #14679 

### Motivation

When a user deletes a CronWorkflow (CronSchedule) from the Workflows UI, any associated Workflow resources are also garbage collected by Kubernetes. However, this behavior isn’t clearly communicated in the UI, which could result in accidental data loss.

To improve clarity, I’ve added an informational message to the delete confirmation modal to let users know that Workflows created by the CronSchedule will also be deleted. This is the first step toward eventually adding an option to retain orphaned Workflows in a future release.

### Modifications
- Added a warning message to the delete confirmation modal for CronWorkflows.
- The message informs users that all Workflows created by the CronWorkflow will also be deleted.
- UI/UX change only — no functional logic added yet.

image

### Verification
- Manually tested the delete modal in the Workflows UI.

### Documentation
- Not required at this point.

### DIFF
--- a/ui/src/cron-workflows/cron-workflow-details.tsx
+++ b/ui/src/cron-workflows/cron-workflow-details.tsx
@@ -152,7 +152,7 @@ export function CronWorkflowDetails({match, location, history}: RouteComponentPr
                 iconClassName: 'fa fa-trash',
                 disabled: edited,
                 action: () => {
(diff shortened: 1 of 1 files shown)


bencoates08:
It might be worth prefacing that only non-archived workflows will be deleted

MasonM (approved):

PR #14754, opened 2025-08-08
fix: allow query parameters for DELETE method and disable body marshaling. Fixes #14753

### LINKED ISSUE #14753: CLI force parameter is not passed when trying to delete a workflow using an HTTP connection
### Pre-requisites

`force` parameter while using `argo CLI` (`argo delete <wf> --force`) is not passed to the argo server when using an HTTP connection. When you check the `facade` code used for HTTP connection, you can see that only the `GET` method can have query parameters, and the `grpc-gateway` code expects that.

`facade.go`:
```
func (h Facade) url(method, path string, in interface{}) (*url.URL, error) {
	query := url.Values{}
	for s, v := range flatten.Flatten(in) {
		x := "{" + s + "}"
		if strings.Contains(path, x) {
			path = strings.Replace(path, x, v, 1)
		} else if method == "GET" {
			query.Set(s, v)
		}
	}
	// remove any that were not provided
	path = regexp.MustCompile("{[^}]*}").ReplaceAllString(path, "")
	return url.Parse(h.baseURL + path + "?" + query.Encode())
}
```

`workflow.pb.gw.go` `delete` handler:
```
func request_WorkflowService_DeleteWorkflow_0(ctx context.Context, marshaler runtime.Marshaler, client WorkflowServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
	var protoReq WorkflowDeleteRequest
	var metadata runtime.ServerMetadata

	var (
		val string
		ok  bool
		err error
		_   = err
	)

	val, ok = pathParams["namespace"]
	if !ok {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "missing parameter %s", "namespace")
	}

	protoReq.Namespace, err = runtime.String(val)

	if err != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", "namespace", err)
	}

	val, ok = pathParams["name"]
	if !ok {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "missing parameter %s", "name")
	}

	protoReq.Name, err = runtime.String(val)

	if err != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", "name", err)
	}

	if err := req.ParseForm(); err != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
	}
	if err := runtime.PopulateQueryParameters(&protoReq, req.Form, filter_WorkflowService_DeleteWorkflow_0); err != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
	}

	msg, err := client.DeleteWorkflow(ctx, &protoReq, grpc.Header(&metadata.HeaderMD), grpc.Trailer(&metadata.TrailerMD))
	return msg, metadata, err

}
```

# How to reproduce:
- use HTTP CLI connection
- add log line in Argo Server `DeleteWorkflow` method to see `force` value
- run `argo delete <wf> --force`
- check argo server logs

### Version(s)

:latest, 8fb4bbaf783d7b912db3a9c69629962172893f73

```YAML
-
```

```text
-
```

```text
```

djanjic:
Fixes #14753

### Motivation
As described in issue #14753, query parameters are not passed when using a DELETE request. Code generated by `grpc-gateway` only allows path and query parameters for a `DELETE` request, so the `facade` code is updated to accommodate that. In general, the body can be used, but the command for generating the code needs to be updated.

https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/main.go#L22

### Modifications
Allowed

### Verification
- manual testing
- added unit test

### DIFF
--- a/pkg/apiclient/http1/facade.go
+++ b/pkg/apiclient/http1/facade.go
@@ -93,7 +93,7 @@ func (h Facade) EventStreamReader(ctx context.Context, in interface{}, path stri
 func (h Facade) do(ctx context.Context, in interface{}, out interface{}, method string, path string) error {
 	log := logging.RequireLoggerFromContext(ctx)
 	var data []byte
-	if method != "GET" {
+	if method != "GET" && method != "DELETE" {
 		var err error
 		data, err = json.Marshal(in)
 		if err != nil {
@@ -148,7 +148,7 @@ func (h Facade) url(method, path string, in interface{}) (*url.URL, error) {
 		x := "{" + s + "}"
 		if strings.Contains(path, x) {
 			path = strings.Replace(path, x, v, 1)
-		} else if method == "GET" {
+		} else if method == "GET" || method == "DELETE" {
 			query.Set(s, v)
 		}
 	}
--- a/pkg/apiclient/http1/facade_test.go
+++ b/pkg/apiclient/http1/facade_test.go
@@ -13,4 +13,8 @@ func TestFacade_do(t *testing.T) {
 	u, err := f.url("GET", "/{namespace}/{name}", &metav1.ObjectMeta{Namespace: "my-ns", Labels: map[string]string{"foo": "1"}})
 	require.NoError(t, err)
 	assert.Equal(t, "http://my-url/my-ns/?labels.foo=1", u.String())
+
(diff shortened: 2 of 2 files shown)


Joibel (approved):
Makes sense to me.

PR #14768, opened 2025-08-18
feat: support tracking workflow of workflows

### LINKED ISSUE #6922: UI: Workflow of Workflows representation
# Summary

Currently a workflow of workflows (using the resource template) creates a seperate workflow representation in the UI with no form of reference or link between the created workflow and the "parent" workflow. Having some way to see the entire workflow of workflows as a single UI visualization would be great.

Terms:
Parent referrers to any Workflow that creates a workflow
Child referrers to any Workflow created by a (parent) workflow

Ideas for implementation:
- Add a parent setting to the workflow spec, the parent setting can be filled in with a workflow name and a step name for what workflow created it. This is simpler than a children setting given that the parent when creating the child knows it's own name, however a child's name isn't always known by the parent due to generated names.
- When a workflow step has children display them in the UI Workflow view visualisation, timeline, etc.

# Use Cases

This can be used for visualising and calculating complicated Workflow of Workflows scenarios where workflow templates aren't feasible, such as dynamically created workflows.

---

cw-Guo:
This is a workaround solution to fix #6922

### Motivation
If we use the [workflow of workflows](https://argo-workflows.readthedocs.io/en/latest/workflow-of-workflows/) pattern, it's painful to check the details of the spawned child workflows. We need to check the logs of the parent workflow's to see the name for the child workflow, and then use the name to go to the workflows page to search for workflow details. We need some quick links to avoid such operations.

### Modifications
- add two labels to the manifest defined in the resource template when `SetOwnerReference ` is `true`

### Verification
local tests

### Documentation

see feature description

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -3562,30 +3562,32 @@ func (woc *wfOperationCtx) executeResource(ctx context.Context, nodeName string,
 
 	tmpl = tmpl.DeepCopy()
 
-	obj := unstructured.Unstructured{}
-	err = yaml.Unmarshal([]byte(tmpl.Resource.Manifest), &obj)
-	if err != nil {
-		return node, err
-	}
 	if tmpl.Resource.SetOwnerReference {
+		obj := unstructured.Unstructured{}
+		err := yaml.Unmarshal([]byte(tmpl.Resource.Manifest), &obj)
+		if err != nil {
+			return node, err
+		}
+
 		ownerReferences := obj.GetOwnerReferences()
 		obj.SetOwnerReferences(append(ownerReferences, *metav1.NewControllerRef(woc.wf, wfv1.SchemeGroupVersion.WithKind(workflow.WorkflowKind))))
-	}
 
-	// attach the labels for better tracking
-	labels := obj.GetLabels()
-	if labels == nil {
-		labels = make(map[string]string)
-	}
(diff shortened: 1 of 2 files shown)


workflow/controller/operator.go
```diff
@@ -3571,11 +3571,22 @@ func (woc *wfOperationCtx) executeResource(ctx context.Context, nodeName string,
...
 		obj.SetOwnerReferences(append(ownerReferences, *metav1.NewControllerRef(woc.wf, wfv1.SchemeGroupVersion.WithKind(workflow.WorkflowKind))))
+
+		// attach the labels for better tracking
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
+		}
+		labels[common.LabelKeyResourceParentPodName] = wfutil.GeneratePodName(woc.wf.Name, nodeName, tmpl.Name, node.ID, wfutil.GetWorkflowPodNameVersion(woc.wf))
+		labels[common.LabelKeyResourceParentWorkflowName] = woc.wf.Name
+		obj.SetLabels(labels)
```
eduardodbr, line 3582: I believe this is only useful for child Workflows, so maybe we could only add the labels if `obj.GetKind() == workflow.WorkflowKind`

cw-Guo:
closed in favor of #14872

PR #14771, opened 2025-08-19
fix(ui): add tooltips to WORKFLOW TEMPLATE and CRON WORKFLOW dropdowns

### LINKED ISSUE #14770: UI Improvement: Issue with long text being cut off in the dropdowns for WORKFLOW TEMPLATE and CRON WORKFLOW
# Summary

The WORKFLOW TEMPLATE and CRON WORKFLOW dropdown lists in the Workflows tab need improvement to handle long text. Currently, when template or workflow names are long, the text gets truncated without any way to view the full content. This makes it difficult for users to identify and select the correct template or workflow.

We need to implement tooltips that show the full text when hovering over truncated names in these dropdowns. This would significantly improve usability when working with long template or workflow names.

## Use Cases

This feature would be useful in the following scenarios:

1. When working with templates that follow naming conventions that include detailed information (e.g., environment, purpose, version)
2. When multiple similar templates exist and need to be distinguished by their full names
3. When templates are generated programmatically with long descriptive names

---

garireo2549:
Fixes #14770 

### Motivation
The current implementation of WORKFLOW TEMPLATE and CRON WORKFLOW dropdowns in the Workflows tab has a usability issue. When template or workflow names are long, the text gets truncated without any way to view the full content. This makes it difficult for users to identify and select the correct template or workflow, especially when there are multiple items with similar names.

Users need to be able to see the full name of templates and workflows in the dropdown lists, even when those names are long. Currently, hovering over truncated text does not display any tooltip or additional information.

### Modifications
This PR implements the following changes to improve the user experience:

1. Implement logic in the DataLoaderDropdown component to add a title attribute to each option. Set the title attribute for static option lists and use MutationObserver to set the title attribute for dynamically added option elements as well. 

2. Improve how long text is displayed in the workflow filter stylesheet by showing overflowed text with an ellipsis (text-overflow: ellipsis) and displaying the full text as a tooltip on hover.

Screenshots after the changes:

スクリーンショット 2025-08-19 13 04 14
スクリーンショット 2025-08-19 13 04 09

### Verification
The changes have been tested with the following scenarios:

1. Created workflow templates with very long names
2. Verified that long names are properly truncated with ellipsis in the dropdown
3. Confirmed that hovering over truncated text displays a tooltip with the full name
4. Verified that the functionality works for both WORKFLOW TEMPLATE and CRON WORKFLOW dropdowns

### Documentation
No documentation updates are required for this change as it's a UI improvement that doesn't introduce new features or change existing functionality. The behavior (showing tooltips on hover) follows standard web UI patterns that users are already familiar with.

### DIFF
--- a/ui/src/shared/components/data-loader-dropdown.tsx
+++ b/ui/src/shared/components/data-loader-dropdown.tsx
@@ -1,24 +1,86 @@
 import {DataLoader} from 'argo-ui/src/components/data-loader';
 import {Select, SelectOption} from 'argo-ui/src/components/select/select';
 import * as React from 'react';
-import {useState} from 'react';
+import {useEffect, useRef, useState} from 'react';
 
 export function DataLoaderDropdown(props: {load: () => Promise<(string | SelectOption)[]>; onChange: (value: string) => void; placeholder?: string}) {
     const [selected, setSelected] = useState('');
+    const selectRef = useRef<HTMLDivElement>(null);
+    const [options, setOptions] = useState<(string | SelectOption)[]>([]);
+
+    const enhanceOptionsWithTitles = (options: (string | SelectOption)[]) => {
+        setOptions(options);
+        return options.map(option => {
+            if (typeof option === 'string') {
+                return {
+                    value: option,
+                    title: option
+                };
+            } else {
+                return option;
+            }
+        });
+    };
+
+    useEffect(() => {
+        const addTitleAttributes = () => {
+            setTimeout(() => {
+                const optionElements = document.querySelectorAll('.select__option');
+                optionElements.forEach(element => {
+                    const text = element.textContent || '';
+                    if (text) {
+                        element.setAttribute('title', text);
+                    }
+                });
+            }, 100);
+        };
+
+        if (options.length > 0) {
+            addTitleAttributes();
+        }
+    }, [options]);
+
+    useEffect(() => {
+        const handleDropdownOpen = () => {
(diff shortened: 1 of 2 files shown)


ui/src/shared/components/data-loader-dropdown.tsx
```diff
@@ -1,24 +1,86 @@
...
+    useEffect(() => {
+        const handleDropdownOpen = () => {
+            const observer = new MutationObserver(mutations => {
+                mutations.forEach(mutation => {
+                    if (mutation.type === 'childList' && mutation.addedNodes.length > 0) {
+                        const optionElements = document.querySelectorAll('.select__option');
+                        optionElements.forEach(element => {
+                            const text = element.textContent || '';
+                            if (text) {
+                                element.setAttribute('title', text);
```
MasonM, line 52: Modifying DOM elements outside React is liable to cause bugs ([source](https://react.dev/learn/manipulating-the-dom-with-refs#best-practices-for-dom-manipulation-with-refs)):
>  Avoid changing DOM nodes managed by React. Modifying, adding children to, or removing children from elements that are managed by React can lead to inconsistent visual results or crashes like above.

Also, I don't see understand this is necessary. Why can't the `title` attribute be set at the time the `<Select>` component is instantiated?

garireo2549: @MasonM 
Thank you for your review.
I was not aware of that rule 😢 
I will make the necessary corrections, so please wait a moment.

PR #14782, opened 2025-08-24
fix: retry strategy being ignored by daemoned nodes. Fix #14715

### LINKED ISSUE #14715: failed daemon leaves workflow in bad state
### Pre-requisites

possibly related to #14544 

Image

i have transient retries on things such as pod deletion

that staging app server errored out with status 143
the other app server deleted
not sure why
then it looks like it spun back up due to transient errors and retries we have with pod deletions
then it looks like it tried to do 2 more retries with the git clone steps again
and workflows are just stuck in a weird state now
the workflow looks failed, but still has pods running

### Version(s)

3.7.0

```YAML
n/a
```

```text
running into character limit. linking to logs in slack snippet
https://cloud-native.slack.com/archives/C01QW9QSSSK/p1753901691851319?thread_ts=1753901385.723079&cid=C01QW9QSSSK
```

```text
garbage collection ran i think


kubectl logs -n frontend -c wait -l workflows.argoproj.io/workflow=playwright-simple-site-dg7wr,workflow.argoproj.io/phase!=Succeeded
No resources found in frontend namespace.
```

eduardodbr:
Fixes #14715

### Motivation

Whenever a daemon pod Fails/errors the retryStategy is being ignored. The current code logic will skip the retry, and The message “Node not set to be retried” shows on the logs. However:
- workflow controller is still creating a new daemon node, ignoring the retryStrategy rules
- that node is not copying the children (step 2 in picture below)
- the killed daemoned node is marked as fulfilled with success ( I think this is by design)
- the workflow is set as failed event though the new daemon node is running

This leaves a zombie pod (monitor(1)) and If the global retry strategy is set, the workflow is retried.

image

### Modifications

No longer creates new children when a daemoned retry node was set as failed or errored, which only happens when woc.processNodeRetries() decides that is not to be retried.

### Verification

```
kind: Workflow
metadata:
  generateName: daemon-example-
spec:
  entrypoint: main
  templates:
    - name: main
      steps:
        - - name: monitor
            template: monitor

        - - name: fail
            template: fail 

    - name: monitor
      daemon: true
      retryStrategy:
        limit: 2
      script:
        image: python:alpine3.6
        command: [ python ]
        source: |
          import time
          print("Daemon running successfully", flush=True)
          time.sleep(10)
          sys.exit(1) # exit with non-zero code to failed node

    - name: fail
      script:
        image: python:alpine3.6
        command: [ python ]
        source: |
          import time
          import sys
          time.sleep(20)
          sys.exit(1) # remove this to test successfull workflow
```

The daemon pod is always failing, which triggers a Failed node. Kill the pod (`kubectl kill pod <name>`) to test a Errored node.

With `retryPolicy: Always`

Daemon pods are retried twice. After the last failure, the workflow is marked as failed and daemon nodes marked as successful (this is by design, all daemoned containers are set as Succeeded.

image

With `retryPolicy: OnFailure`

If pod is killed by outside it is marked as `Errored`, so no retry:

image

If pod exits with `sys.exit(1)` it is marked as `Failed` so it retries:

image

With a global retry strategy, workflow will retry if daemoned pod fails, even if all other steps succeed 

image

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -2178,7 +2178,7 @@ func (woc *wfOperationCtx) executeTemplate(ctx context.Context, nodeName string,
 		childNodeIDs, lastChildNode := getChildNodeIdsAndLastRetriedNode(retryParentNode, woc.wf.Status.Nodes)
 
 		// The retry node might have completed by now.
-		if retryParentNode.Fulfilled() && woc.childrenFulfilled(retryParentNode) { // if retry node is daemoned we want to check those explicitly
+		if retryParentNode.Fulfilled() && (woc.childrenFulfilled(retryParentNode) || (retryParentNode.IsDaemoned() && retryParentNode.FailedOrError())) { // if retry node is daemoned we want to check those explicitly
 			// If retry node has completed, set the output of the last child node to its output.
 			// Runtime parameters (e.g., `status`, `resourceDuration`) in the output will be used to emit metrics.
 			if lastChildNode != nil {


isubasinghe (approved):

eduardodbr:
@MenD32 I believe this was introduced in #13738, can you please take a look?

MenD32 (approved):

MenD32:
Looks good! must have missed that in my testing...

PR #14784, opened 2025-08-25
feat: Add support for creating a database semaphore config using CLI

### LINKED ISSUE #14783: Add support for creating a database semaphore config using CLI
# Summary
Currently, the user must manually add a new entry into database with the semaphore configuration in order to use it. This process can be difficult for some users, and it also requires the user to have the necessary permissions.

By adding support for creating the configuration via the CLI, users can more easily add new semaphore configurations. This work will rely on configmap work described in #14671 and will help for eventually adding the UI support.

## Use Cases
- creating a new database semaphore configuration using CLI
 - make semaphore configuration creation less error-prone

---

djanjic:
Note: This PR is based on #14671

Fixes #14783

### Motivation
This adds support for creating database semaphore configuration using CLI.

### Modifications
- based on configmap PR (check #14671). Did a small refactor there
- moved some common db handling code from wf controller package to utility in order to be able to reuse it.
- added CLI support for creating database sync config.
- added tests

### Verification
- manual testing
- e2e tests
- unit tests

### Documentation
- Documentation is automatically generated for new CLI commands


cmd/argo/commands/sync/db/create.go
```diff
@@ -0,0 +1,66 @@
...
+
+func NewCreateCommand() *cobra.Command {
+
+	var cliCreateOpts = cliCreateOpts{}
+
+	command := &cobra.Command{
+		Use:     "create",
+		Short:   "create a db sync limit",
+		Args:    cobra.ExactArgs(1),
+		Example: `argo sync db create my-key --size-limit 10`,
```
isubasinghe: While I understand why there is no --key flag here, I kind of don't like the inconsistency.

isubasinghe: Makes me wonder if the cli should look more like one of these:

`argo sync create my-key --type db`

`argo sync create my-key  --configmap cmapname`

`argo sync create my-key --type configmap --name cmapname`

Joibel: I agree with this, consistency between the two commands would be preferable.

docs/cli/argo_sync_db_create.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+argo sync db create my-key --size-limit 10
+```
+
+### Options
+
+```
+  -h, --help               help for create
+      --size-limit int32   Size limit of the sync limit
```
Joibel: I think this could be simplified to either --size or --limit here and in the other places

cmd/argo/commands/sync/db/create.go
eduardodbr: my preferred solution would also be:

argo sync create <key> --type [configmap|db] --limit <n> [--configmap <name>]

server/sync/sync_db.go
```diff
@@ -0,0 +1,108 @@
...
+
+type dbSyncProvider struct {
+	db syncdb.SyncQueries
+}
+
+var _ SyncConfigProvider = &dbSyncProvider{}
+
+func (s *dbSyncProvider) createSyncLimit(ctx context.Context, req *syncpkg.CreateSyncLimitRequest) (*syncpkg.SyncLimitResponse, error) {
+	allowed, err := auth.CanI(ctx, "create", workflow.WorkflowPlural, req.Namespace, "")
+	if err != nil {
```
eduardodbr, line 28: just wondering, why do you think it is necessary to check permissions for DB sync limits but not for ConfigMap sync?

server/sync/sync_cm.go
```diff
@@ -0,0 +1,154 @@
...
+
+	kubeClient := auth.GetKubeClient(ctx)
+
+	configmapGetter := kubeClient.CoreV1().ConfigMaps(req.Namespace)
+
+	cm, err := configmapGetter.Get(ctx, req.Name, metav1.GetOptions{})
+	if err == nil {
+		_, has := cm.Data[req.Key]
+		if has {
+			return nil, fmt.Errorf("sync limit cannot be created as it already exists")
```
eduardodbr: this should return a status.Error(codes.AlreadyExists) error

server/sync/sync_db.go
Joibel: This was my suggestion. ConfigMap access is controlled by Kubernetes standard RBAC. There is no such equivalent for the database access, so the idea was to make it use an in cluster object to see if the request was "reasonable".

@djanjic - this behaviour should be documented, and this whole class of db access should be controllable (e.g. the admin can disable it if they don't want it).

server/sync/sync_cm.go
djanjic: updated

docs/cli/argo_sync_db_create.md
djanjic: updated

cmd/argo/commands/sync/db/create.go
djanjic: This is now updated. Both cm and db part.

server/sync/sync_db.go
djanjic: updated

cmd/argo/commands/sync/get.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	req := &syncpkg.GetSyncLimitRequest{
+		CmName:    cliGetOpts.cmName,
+		Namespace: client.Namespace(ctx),
+		Key:       key,
+		Type:      syncpkg.SyncConfigType(syncpkg.SyncConfigType_value[cliGetOpts.syncType]),
+	}
+
+	fmt.Printf("Request: %+v\n", req)
```
Joibel: Remove this line please

docs/synchronization-config.md
Joibel, line 1: Suggestions for this sent in slack.

docs/synchronization-config.md
```diff
@@ -1,22 +1,89 @@
-# Synchronization
+# Managing Synchronization Limits via API
 
-It is possible to create a lock configuration using the Argo Server API (and CLI). A ConfigMap API is always available and the database one is disabled by default.
-For the API reference, please check the [Swagger documentation](swagger.md), and for CLI usage, refer to the [CLI documentation](cli/argo_sync.md).
+This page explains how to manage synchronization limits (mutexes and semaphores) using the Argo Server API and CLI.
```
Joibel: Humm, my fault, but we should remove the word mutexes from this line.

djanjic: updated

PR #14801, opened 2025-08-30
fix(ui): remove duplicate params when create workflow using by workfl…

### LINKED ISSUE #14772: Argo UI displays duplicate parameters after selecting entrypoint while submitting workflow
### Pre-requisites

Please see images attached. 
Image 1 shows input parameters configuration where entry point is `default` while submitting the WF.
Image 2 shows input parameters listed twice where entry point is switched to `main` while submitting the WF.

Image
Image

### Version(s)

v3.6.7

```YAML
metadata:
  name: sample-workflow
  namespace: argo-workflows
spec:
  templates:
    - name: main
      inputs:
        parameters:
          - name: parameter1
          - name: parameter2
      outputs: {}
      metadata: {}
      container:
        name: ""
        image: bitnami/kubectl:latest
        command:
          - /bin/bash
          - -c
        args:
          - >
            set -e

            sleep 3600
        resources: {}
  entrypoint: main
  arguments:
    parameters:
      - name: parameter1
        description: parameter1
      - name: parameter2
        default: "a"
        enum:
          - "a"
          - "b"
        description: parameter2
```

mg5566:
…ow template

Fixes #14772 

I removed 'parameters' because 'workflowParameters' exists.

I used the manifest below, and removed the duplicate param as shown in the image.

```yaml
metadata:
  name: sample-workflow
  namespace: argo-workflows
spec:
  templates:
    - name: main
      inputs:
        parameters:
          - name: parameter1
          - name: parameter2
      outputs: {}
      metadata: {}
      container:
        name: ""
        image: bitnami/kubectl:latest
        command:
          - /bin/bash
          - -c
        args:
          - >
            set -e

            sleep 3600
        resources: {}
  entrypoint: main
  arguments:
    parameters:
      - name: parameter1
        description: parameter1
      - name: parameter2
        default: "a"
        enum:
          - "a"
          - "b"
        description: parameter2
```
스크린샷 2025-08-30 오후 3 20 08

### DIFF
--- a/ui/src/workflows/components/submit-workflow-panel.tsx
+++ b/ui/src/workflows/components/submit-workflow-panel.tsx
@@ -64,10 +64,7 @@ export function SubmitWorkflowPanel(props: Props) {
         try {
             const submitted = await services.workflows.submit(props.kind, props.name, props.namespace, {
                 entryPoint: entrypoint === workflowEntrypoint ? null : entrypoint,
-                parameters: [
-                    ...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p)),
-                    ...parameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))
-                ],
+                parameters: [...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))],
                 labels: labels.join(',')
             });
             navigation.goto(uiUrl(`workflows/${submitted.metadata.namespace}/${submitted.metadata.name}`));


ui/src/workflows/components/submit-workflow-panel.tsx
```diff
@@ -65,8 +65,8 @@ export function SubmitWorkflowPanel(props: Props) {
             const submitted = await services.workflows.submit(props.kind, props.name, props.namespace, {
                 entryPoint: entrypoint === workflowEntrypoint ? null : entrypoint,
                 parameters: [
-                    ...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p)),
-                    ...parameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))
+                    ...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))
+                    // ...parameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))
```
eduardodbr: this should be removed instead of commented

mg5566: 🫡 Remove Complete!

eduardodbr:
I tested the code and the issue is still hapening if you submit workflow and change entrypoint to main->default>main 

image

ui/src/workflows/components/submit-workflow-panel.tsx
```diff
@@ -64,10 +64,7 @@ export function SubmitWorkflowPanel(props: Props) {
         try {
             const submitted = await services.workflows.submit(props.kind, props.name, props.namespace, {
                 entryPoint: entrypoint === workflowEntrypoint ? null : entrypoint,
-                parameters: [
-                    ...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p)),
-                    ...parameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))
-                ],
+                parameters: [...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))],
```
MasonM, line 67: Unfortunately, this is going to break submission of workflows that define input parameters that aren't workflow arguments. Try creating the following as a workflow template and then submitting it:
```yaml
metadata:
  name: sample-workflow2
spec:
  templates:
    - name: main
      inputs:
        parameters:
          - name: parameter1
          - name: parameter2
      container:
        image: busybox
        command: [echo]
        args: ["{{inputs.parameters.parameter1}} {{inputs.parameters.parameter2}} {{workflow.parameters.parameter1}}"]
  entrypoint: main
  arguments:
    parameters:
      - name: parameter1
        description: parameter1
```

You'll see an error similar to `Bad Request: templates.main inputs.parameters.parameter2 was not supplied`.

I think the correct fix here is to merge the `workflowParameters` and `parameters` arrays, and if there's any parameters with the same name, the one in `workflowParameters` takes precedence. This matches the corresponding logic in the backend: https://github.com/argoproj/argo-workflows/blob/1e2a87f2afdebbcd0e55069df5a945f5faca9d45/workflow/common/util.go#L166-L181

MasonM (changes requested):
Thanks for your contribution!

mg5566:
> I tested the code and the issue is still hapening if you submit workflow and change entrypoint to main->default>main
> 
Hmm... I don't have that problem.
Can you check my video


https://github.com/user-attachments/assets/ef2dd721-23e4-416f-adfd-affb74fc68ea

ui/src/workflows/components/submit-workflow-panel.tsx
mg5566: Thnaks @MasonM 

I'll try recreate it by referring to the backend logic you told me.

srgva:
Possibly related issue https://github.com/argoproj/argo-workflows/issues/15068

PR #14813, opened 2025-09-04
fix: Document the /api/v1/workflow-templates API in swagger

### LINKED ISSUE #14799: API /api/v1/workflow-templates seems to be undocumented
### Pre-requisites

The public documentation lists only `/api/v1/workflow-templates/{namespace}`, and I don't see `/api/v1/workflow-templates` documented here. 

https://argo-workflows.readthedocs.io/en/latest/swagger/

Is this just a mistake? Happy to update docs/swagger to clarify that `namespace` is not a required parameter as the docs says. 

### Version(s)

ALL

```YAML
NA
```

```text
NA
```

```text
NA
```

jaseemabid:
I copied the /api/v1/workflow-templates/{namespace} block as it is without doing much else.

The JSON looks pretty verbose and duplicated, but I'm not sure what else I could do unless this file is generated from something else.

Fixes #14799

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -2592,6 +2592,97 @@
         }
       }
     },
+    "/api/v1/workflow-templates": {
+      "get": {
+        "tags": [
+          "WorkflowTemplateService"
+        ],
+        "operationId": "WorkflowTemplateService_ListWorkflowTemplates",
+        "parameters": [
+          {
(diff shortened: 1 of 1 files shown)


jaseemabid:
# Before

image

# After

image

Joibel:
You can't manually add things to swagger.json like this I'm afraid, it's autogenerated.

PR #14816, opened 2025-09-07
feat(ui): add label filter sync with URL query params

### LINKED ISSUE #14807: WorkflowTemplates page does not support label filtering via query parameters in the UI
# Summary

Currently, the Argo Workflows UI allows filtering of Workflows using label selectors via query parameters in the URL (e.g.,
https://\<host\>/workflows?namespace=xyz&label=key%3Dvalue), which is very helpful for narrowing down views.

However, this feature does not work for WorkflowTemplates. When attempting to filter the WorkflowTemplates list using similar query parameters (e.g.,
https://\<host\>/workflow-templates?namespace=xyz&label=key%3Dvalue), the label filter is silently ignored, and the UI lists all available templates.

This is despite the fact that the API supports filtering WorkflowTemplates via label selectors, as demonstrated by a direct API call to:
/api/v1/workflow-templates/<namespace>?listOptions.labelSelector=key%3Dvalue

This inconsistency:

- Limits the ability to create UI links that show only a subset of user-facing templates (e.g., hide internal or system templates).
- Breaks the symmetry with workflow filtering behavior.

**Expected Behavior:**

When visiting the WorkflowTemplates page with a _label=_ query parameter, only the templates matching the _label_ selector should be shown, just like it works for workflows.

**Actual Behavior:**

The _label_ query parameter is ignored in the WorkflowTemplates UI, and all templates are displayed.

**Proposed Fix:**

Support _label=\<labelSelector\>_ query parameters in the /workflow-templates UI route, and apply them consistently when rendering the list view.

## Use Cases

Argo Workflows allows adding custom UI links (via workflow-controller-configmap). These links could point directly to filtered views of WorkflowTemplates, making it easier for users to launch workflows relevant to their context. However, since label filtering currently doesn't work for the /workflow-templates UI route, this use case is blocked.

---

puretension:
Fixes #14807 

### Motivation

WorkflowTemplates UI silently ignores label query parameters (e.g., `?label=key%3Dvalue`) unlike Workflows list, breaking URL filtering and preventing custom UI links for filtered template views.

### Modifications

- Added label query parameter sync with URL in WorkflowTemplates UI
- Combined URL updates and localStorage persistence in single useEffect
- Now matches Workflows list behavior for consistent filtering

https://github.com/user-attachments/assets/e61c8f78-f715-4e64-99a8-31b762ea1b1f

### Verification

- `?label=key%3Dvalue` now properly filters WorkflowTemplates
- URL updates when changing filters in UI
- Filters persist on page refresh

### DIFF
--- a/.features/pending/14807-workflow-templates-label-filter-sync.md
+++ b/.features/pending/14807-workflow-templates-label-filter-sync.md
@@ -1,0 +1,9 @@
+Component: UI
+Issues: 14807
+Description: Add label query parameter sync with URL in WorkflowTemplates UI to match Workflows list behavior for consistent filtering.
+Author: [puretension](https://github.com/puretension)
+
+- WorkflowTemplates UI now properly handles label query parameters (e.g., ?label=key%3Dvalue)
+- Combined URL updates and localStorage persistence in single useEffect
+- Enables custom UI links for filtered template views
+- Verified that URL updates when changing filters and filters persist on page refresh


puretension:
Hi @eduardodbr,

Thanks for approving! After implementing this fix, 
I noticed we have duplicate label synchronization logic across multiple components:

## Duplicate patterns found:
• **workflows-list.tsx** + workflow-template-list.tsx: Same useState + queryParams + localStorage logic
• **cron-workflows/cron-workflow-list.tsx**: Has labels state but missing URL sync  
• **reports/reports.tsx**: Different queryParams pattern

## Duplicated code pattern:
```
typescript
// Repeated in multiple files
const [labels, setLabels] = useState(() => {
    const queryParams = new URLSearchParams(location.search);
    const labelQueryParam = queryParams.getAll('label');
    return labelQueryParam.length > 0 ? labelQueryParam : savedLabels;
});
```


## Proposal:
Would you be interested in a follow-up PR after this merges that extracts this 
into a shared custom hook? This would:

• Eliminate code duplication
• Improve maintainability 
• Standardize label filtering across all components

Let me know what you think!

eiriklid:
Hey, I would really like to have this feature. 

Is there any way I can contribute so that it can be released?

PR #14832, opened 2025-09-10
fix: reset the traskgroup

isubasinghe:
Fixes #TODO


Joibel:
Any reason not to add the test file as a regression test?

isubasinghe:
> Any reason not to add the test file as a regression test?

Added it now

Issue #14042, opened 2025-01-01
dynamic template naming on workflowTemplates fails unexpectedly


MenD32:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [X] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

templating a template name fails when using a workflowTemplate's input, this bug can be mitigated by using an intermediate template to output the input of the workflowTemplate 

### Version(s)

v3.6.2

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

```YAML
---
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: example-workflow-template-a
spec:
  templates:
  - name: whalesay-template
    inputs:
      parameters:
        - name: message
    steps:
      - - name: whalesay
          templateRef:
            name: example-workflow-template-b
            template: "{{ inputs.parameters.message }}"
---
apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: example-workflow-template-b
spec:
  templates:
  - name: whalesay-template
    container:
      image: docker/whalesay:latest
      command: [cowsay]
      args: ["hello from template"]
---
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: example-workflow-
spec:
  entrypoint: whalesay
  templates:
  - name: whalesay
    steps:
      - - name: whalesay
          templateRef:
            name: example-workflow-template-a
            template: whalesay-template
          arguments:
            parameters:
              - name: message
                value: "whalesay-template"
```

### Logs from the workflow controller

```text
not relevant
```

### Logs from in your workflow's wait container

```text
not relevant
```

tczhao:
It is the same issue as https://github.com/argoproj/argo-workflows/issues/13773
but this example is much simpler

shuangkun:
Will you submit a PR to fix this issue? @MenD32

MenD32:
[...]

Issue #13773, opened 2024-10-17
template validation/submit fail if dynamic templateRef is used in nested template


tczhao:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

When using dynamic template in nested workflow template calls, the validation failed and couldn't submit workflow.
calling `wftmpl1 -> wftmpl2:dynamic template -> wftmpl3` has issue
calling `wftmpl2:dynamic template -> wftmpl3` is fine
The issue applies to both clusterworkflowtemplate and workflowtemplate

To reproduce the error, run
```
argo submit --from clusterworkflowtemplate/cwt1  --dry-run -o yaml
```
This would produce error
```
FATA[2024-10-17T12:09:45.670Z] Failed to submit workflow: rpc error: code = InvalidArgument desc = templates.main.tasks.cwt1-s1 templates.main2.steps[0].cwt2-s1 template reference cwt3.placeholder-11 not found
```

### Version(s)

latest, e18570c 

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflows that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: ClusterWorkflowTemplate
metadata:
  name: cwt1
spec:
  entrypoint: main
  templates:
  - name: main
    dag:
      tasks:
      - name: cwt1-s1
        templateRef:
          name: cwt2
          template: main2
          clusterScope: true
---
apiVersion: argoproj.io/v1alpha1
kind: ClusterWorkflowTemplate
metadata:
  name: cwt2
spec:
  entrypoint: main2
  templates:
  - name: main2
    inputs:
      parameters:
      - name: api-template
        value: "abc-abc"
    outputs:
      parameters:
      - name: output-foo
        valueFrom:
          parameter: "{{steps.cwt2-s1.outputs.parameters.foo}}"
    steps:
    - - name: cwt2-s1
        templateRef:
          name: cwt3
          template: "{{inputs.parameters.api-template}}"
          clusterScope: true
---
[...]

PR #14850, opened 2025-09-12
fix: skip validate dynamic templateref

tczhao:
Fixes https://github.com/argoproj/argo-workflows/issues/14042 https://github.com/argoproj/argo-workflows/issues/13773

Before this fix, the validation will failed if workflow contains dynamic templateRef.
Decide to skip validate dynamic templateRef instead of find a way to validate them due to the complexity and cases where its impossible to validate
- e.g. the templateRef depends on the output from a previous step

### Verification

Added test for dynamic templateRef validation. copy from https://github.com/argoproj/argo-workflows/pull/14053
Added test to validate the regression introduced by previous templateRef validation implementation attempt

### DIFF
--- a/workflow/validate/validate.go
+++ b/workflow/validate/validate.go
@@ -574,6 +574,10 @@ func (tctx *templateValidationCtx) validateTemplateHolder(ctx context.Context, t
 	tmplCtx, resolvedTmpl, _, err := tmplCtx.ResolveTemplate(ctx, tmplHolder)
 	if err != nil {
 		if argoerr, ok := err.(errors.ArgoError); ok && argoerr.Code() == errors.CodeNotFound {
+			if tmplRef != nil && strings.Contains(tmplRef.Template, "placeholder") {
+				// placeholder indicate this is a dynamic template, skip validation
+				return nil, nil
+			}
 			if tmplRef != nil {
 				return nil, errors.Errorf(errors.CodeBadRequest, "template reference %s.%s not found", tmplRef.Name, tmplRef.Template)
 			}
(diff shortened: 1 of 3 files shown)


workflow/validate/validate.go
```diff
@@ -574,6 +574,10 @@ func (tctx *templateValidationCtx) validateTemplateHolder(ctx context.Context, t
 	tmplCtx, resolvedTmpl, _, err := tmplCtx.ResolveTemplate(ctx, tmplHolder)
 	if err != nil {
 		if argoerr, ok := err.(errors.ArgoError); ok && argoerr.Code() == errors.CodeNotFound {
+			if tmplRef != nil && strings.Contains(tmplRef.Template, "placeholder") {
```
Joibel, line 577: Can we make it less likely to mismatch on non-placeholder templates with
```suggestion
			if tmplRef != nil && strings.HasPrefix(tmplRef.Template, "placeholder") {
```

Joibel (changes requested):
Minor suggestion

workflow/validate/validate.go
tczhao: The dynamic template parameter can be partial, using prefix matching won't cover the following use case.
e.g.
`"prod-{{ inputs.parameters.message }}"` this will translate to `prod-placeholder-x`.

If a user uses string:placeholder in templateRef, the worst-case scenario is that they experience an issue during runtime instead of catch it at the validation phase.

Joibel: Thanks for that, understand

PR #14852, opened 2025-09-13
fix: http template read response.Body after cancel(), sometimes it return a context canceled error

AnonyScorpio:
### Motivation

 fix: http template read response.Body after cancel(), sometimes it return a context canceled error

### Modifications

Return response.Body in string directly for sometimes read after call cancel() return a context canceled error
For more detail  https://groups.google.com/g/golang-nuts/c/2FKwG6oEvos


AnonyScorpio:
https://github.com/argoproj/argo-workflows/pull/14853  has merged

PR #14853, opened 2025-09-13
fix: http template read response.Body after cancel(), sometimes it return a context canceled error

AnonyScorpio:
### Motivation

http template read response.Body after cancel(), sometimes it return a context canceled error

### Modifications

Read response.Body after cancel(), sometimes it return a context canceled error
For more detail  https://groups.google.com/g/golang-nuts/c/2FKwG6oEvos

### DIFF
--- a/workflow/executor/agent.go
+++ b/workflow/executor/agent.go
@@ -249,7 +249,13 @@ func (ae *AgentExecutor) executeHTTPTemplate(ctx context.Context, tmpl wfv1.Temp
 	if tmpl.HTTP == nil {
 		return 0, nil
 	}
-
+	// Read response.Body after cancel(), sometimes it return a context canceled error
+	// For more detail  https://groups.google.com/g/golang-nuts/c/2FKwG6oEvos
+	var cancel context.CancelFunc
+	if tmpl.HTTP.TimeoutSeconds != nil {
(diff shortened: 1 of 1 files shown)


AnonyScorpio:
There is no impact because it will be retried several times.

AnonyScorpio:
> There is no impact because it will be retried several times.

Ok, problems can arise if the HTTP request is non-idempotent or time-consuming.

Joibel:
> > There is no impact because it will be retried several times.
> 
> Ok, problems can arise if the HTTP request is non-idempotent or time-consuming.

Why does this PR cause this risk to change? (the risk of having retries call a non-idempotent URL multiple times)

AnonyScorpio:
> > > There is no impact because it will be retried several times.
> > 
> > Ok, problems can arise if the HTTP request is non-idempotent or time-consuming.
> 
> Why does this PR cause this risk to change? (the risk of having retries call a non-idempotent URL multiple times)
Read `response.Body` before `response.Body.Close` (`defer response.Body.Close()`) by handle `ctx(tmpl.HTTP.TimeoutSeconds)` and `defer cancel()` out of `executeHTTPTemplateRequest`

AnonyScorpio:
@Joibel Is there any problem with merging this PR?

Joibel:
> @Joibel Is there any problem with merging this PR?

Just me getting free time to do things.

Thanks for this contribution.

PR #14892, opened 2025-09-28
feat: optimize pagination when counting workflows in archive. Fixes:#13948

### LINKED ISSUE #13948: /api/v1/workflows/ has slow response time when, there are many archived workflows in 3.6.0
### Pre-requisites

After upgrade 3.6.0 (and latest) from 3.5.8,
`http://argo-server/workflows/?&limit=50` takes response time too long.

I've found there is log for `argo-server`

```
Query:          SELECT count(*) as total FROM "argo_archived_workflows" WHERE (("clustername" = $1 AND "instanceid" = $2) AND not exists (select 1 from argo_archived_workflows_labels where clustername = argo_archived_workflows.clustername and uid = argo_archived_workflows.uid and name = 'workflows.argoproj.io/controller-instanceid'))
Arguments:      []interface {}{"default", ""}
Error:          upper: slow query
Time taken:     363.58917s
<...>

time="2024-11-27T10:52:09.899Z" level=info duration=6m3.617456665s method=GET path=/api/v1/workflows/ size=33993 status=0
```

There seems to be an issue with the count query fetching archived_workflow after the upgrade.

You can see memory of database has significant increase.
image

There are 2 million archived workflows in the database:
```
# SELECT count(*) as total FROM "argo_archived_workflows" WHERE (("clustername" = 'default' AND "instanceid" = '') AND not exists (select 1 from argo_archived_workflows_labels where clustername = argo_archived_workflows.clustername and uid = argo_archived_workflows.uid and name = 'workflows.argoproj.io/controller-instanceid'))
postgres-# ;
  total
---------
 2040568
(1 row)
```

In helm values.yaml, we need to set workflows ttl 400days.
```
controller:
  persistence:
    archive: true
    archiveTTL: 400d
```

### Version(s)

a1f67794be68c2bcdfb8900e20ce18b0ab9115ebfe0adf1f1f5100eb6bd7604c

```YAML
.
```

```text
.
```

```text
.
```

shuangkun:
Optimize pagination when counting workflows in archive

Fixes #13948

### Motivation

When the database has more than 1 million workflows, this query statement becomes a bottleneck, and the UI becomes inaccessible, impacting workflow query and observation.

After optimization, testing in a production environment with 5 million records, the UI can be opened normally.

### DIFF
--- a/.features/pending/pagination-when-counting-workflows.md
+++ b/.features/pending/pagination-when-counting-workflows.md
@@ -1,0 +1,16 @@
+<!-- Required: All of these fields are required, including at least one issue -->
+Description: <!-- A brief one line description of the feature -->
+Author: <!-- Author name and GitHub link in markdown format e.g. [Alan Clucas](https://github.com/Joibel) -->
+Component: <!-- component name here, see hack/featuregen/components.go for the list -->
+Issues: <!-- Space separated list of issues 1234 5678 -->
+
+<!--
+Optional
+Additional details about the feature written in markdown, aimed at users who want to learn about it
+* Explain when you would want to use the feature
+* Include code examples if applicable
+  * Provide working examples
+  * Format code using back-ticks
+* Use Kubernetes style
+* One sentence per line of markdown
+-->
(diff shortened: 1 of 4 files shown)


Joibel:
Thanks for this, looks great.

Issue #14668, opened 2025-07-21
Preserve original URL after SSO login redirect


DamienMatias:
# Summary

The Argo Workflows UI should preserve the original URL and redirect users back to it after successful SSO authentication. Currently, when users access a direct link to a specific workflow without being logged in, they are redirected to the SSO login page, but after authentication, they are not returned to the originally requested workflow page. Users must click login again to reach their intended destination.

## Use Cases

This improvement would be useful when:

- Team members share direct links to specific workflows in Slack, email, or documentation
- Monitoring/alerting systems include direct links to problematic workflows

---

**Message from the maintainers**:

Love this feature request? Give it a 👍. We prioritise the proposals with the most 👍.

tczhao:
duplicate of https://github.com/argoproj/argo-workflows/issues/13031
The fix is merged but yet to be released

svenakela:
I have a use case where this PR seems to not solve the problem, but it is related. 
Our services are behind an ingress, with the `BASE_HREF` configured to `/[env]/[service]` for each service to match the path in the ingress.
I.e. browsing the Workflows UI would be `https://<server-name>/test/argo-workflows`.
If a user clicks the _single sign-on_ button the auth flow ends up with a redirect to `/workflows`, which doesn't exist.
If I on the other hand login with a bearer token, I am redirected to `/test/argo-workflows/workflows` 

It seems the redirect is not taking the base href into consideration anymore. A quick look at the #13031 files I get a feeling that the base href is removed?

Using the latest release, 3.7.1 

<img width="1306" height="276" alt="Image" src="https://github.com/user-attachments/assets/9e9bae2a-c6f2-420d-9613-f6558226895e" />

svenakela:
> It seems the redirect is not taking the base href into consideration anymore. A quick look at the [#13031](https://github.com/argoproj/argo-workflows/issues/13031) files I get a feeling that the base href is removed?
> 
> Using the latest release, 3.7.1
> 

Adding here that an older platform, in this case 3.6.7 works as expected. This should probably be its own bug?

MasonM:
@svenakela It's quite possible that https://github.com/argoproj/argo-workflows/pull/13747 broke something here, but I can't reproduce the behavior you're seeing. I entered a PR (https://github.com/argoproj/argo-workflows/pull/14894) to update the local development environment to allow testing `BASE_HREF`, then I followed these steps to try and reproduce it:
[...]

PR #14894, opened 2025-09-29
fix(ui): support base HREF in dev environment

MasonM:
### Motivation
This updates the dev environment to allow testing the [base HREF option](https://argo-workflows.readthedocs.io/en/stable/argo-server/#base-href), since some users have reported issues with that (https://github.com/argoproj/argo-workflows/issues/14668#issuecomment-3207368612).

### Modifications

* Previously, we used `hack/update-sso-redirect-url.sh` to update the SSO redirect URL when `UI_SECURE` is set, but that didn't account for `BASE_HREF`. Updating it to do so would be possible, but it gets messy because we now need to update two different `ConfigMaps`, not just `workflow-controller-configmap`: 
  * https://github.com/argoproj/argo-workflows/blob/cb7ebd9393f3322abf455d906e39a3a976421b30/manifests/components/sso/overlays/workflow-controller-configmap.yaml#L15
  * https://github.com/argoproj/argo-workflows/blob/cb7ebd9393f3322abf455d906e39a3a976421b30/manifests/components/sso/dex/dex-cm.yaml#L20-L23

  Instead, this removes that script and moves the logic to  `Makefile`, where it does a global `sed` on the output of `kustomize` to replace all instances of the redirect URL. Obviously, this is rather hacky, but I couldn't think of a better way of doing this.
* Update `ui/webpack.config.js` to respect `ARGO_BASE_HREF` and use it for both populating the `<base href="...">` tag and proxying connections to the API server. This required removing the tag from `ui/src/index.html`, since Webpack will populate it now. 
* Update `tasks.yaml` to restart Webpack if `webpack.config.js` changes, to simplify development

### Verification
Tested by running the following commands to start up the environment, then verified I can access the workflows page and login:
1. `make start UI=true`
image

2. `make start UI=true PROFILE=sso BASE_HREF=/test`
image

3. `make start UI_SECURE=true PROFILE=sso BASE_HREF=/test/argo-workflows`
image

4. `make start UI_SECURE=true BASE_HREF=test`
image

### Documentation
I added a "Proxying" section to the "Running Locally" page to document this

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -214,6 +214,7 @@ parameterizing
 params
 pprof
 pre-commit
+proxying
 pytorch
 qps
 ray
@@ -279,6 +280,7 @@ versioning
 webHDFS
 webhook
 webhooks
+webpack-dev-server
 workflow-controller-configmap
 workqueue
 yaml
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,23 @@ endif
 ifeq ($(UI),true)
 TASKS                 := controller server ui
 endif
+
+# -- SSO options
+# Need to rewrite the SSO redirect URL referenced in ConfigMaps when UI_SECURE and/or BASE_HREF is set.
+# Can't use "kustomize" or "kubectl patch" because the SSO config is a YAML string in those ConfigMaps.
+SSO_REDIRECT_URL   := http
+ifeq ($(UI_SECURE),true)
+SSO_REDIRECT_URL   := https
+endif
+ifeq ($(BASE_HREF),)
+BASE_HREF          := /
+else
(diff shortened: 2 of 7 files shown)


Joibel:
Neat, thank you. Tested locally.

Joibel:
~~@MasonM would you mind creating a cherry-pick PR for this targetting 3.7 - I believe this is needed there? I'm unsure of how to resolve login.tsx correctly and rather than guessing it felt better for you to do it.~~

Joibel:
> @MasonM would you mind creating a cherry-pick PR for this targetting 3.7 - I believe this is needed there? I'm unsure of how to resolve login.tsx correctly and rather than guessing it felt better for you to do it.

Wrong PR to comment on, sorry.

PR #14909, opened 2025-10-05
fix(ui): login/logout bug when base HREF set. Fixes #14897

### LINKED ISSUE #14897: SSO Login from Login URL ends up in 404 Not Found
### Pre-requisites

When using SSO Login and BASE_HREF, if a user starts with the Login path a successful login ends up in a 404 where the BASE_HREF is not taken into consideration.

Our services are behind an ingress, with the BASE_HREF configured to /[env]/[service] for each service to match the path in the ingress.
I.e. browsing the Workflows UI would be https://<server-name>/test/argo-workflows.
If a user clicks the single sign-on button the auth flow ends up with a redirect to /workflows, which doesn't exist.
If I on the other hand login with a bearer token, I am redirected to /test/argo-workflows/workflows

Image

Also, see the comments from https://github.com/argoproj/argo-workflows/issues/14668#issuecomment-3344735779 and below.
ping @MasonM.

### Version(s)

v3.7.1

```YAML
not needed
```

```text
```

```text
```

MasonM:
Fixes #14897

### Motivation
This fixes two bugs when the [base HREF option](https://argo-workflows.readthedocs.io/en/stable/argo-server/#base-href) is set:
1. The "Logout" button doesn't delete the `Authorization` cookie properly.
2. If you visit the login page directly and click the SSO button, you'll be redirected to the `/workflows` page without the base HREF, which will likely result in a 404. I think this was a regression from https://github.com/argoproj/argo-workflows/pull/13747

### Modifications
The first issue is because logging out calls `document.cookie = 'authorization=;Max-Age=0';` without the `path` attribute, whereas the cookie is set using `document.cookie = 'authorization=TOKEN;SameSite=Strict;path=' = uiUrl('')`, which does have the `path` attribute. Per [RFC 6265](https://httpwg.org/specs/rfc6265.html#sane-set-cookie-semantics), the `path` attribute must match when deleting or overriding a cookie. If you don't have base HREF set, that's not a problem, since the default value of `path` will be `/`.

The second issue is because the default value of the `redirect` query parameter was hardcoded to `'/workflows'`, when it should be `uiUrl('/workflows')`. The `uiUrl()` function handles prepending the base HREF.

I had to do some refactoring to make this easily testable using [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/).

### Verification
Tested by following these steps:
1. Merge these changes with https://github.com/argoproj/argo-workflows/pull/14894
2. Run `make start UI_SECURE=true PROFILE=sso BASE_HREF=/test/argo-workflows`
3. Visit https://localhost:8080/test/argo-workflows/login
4. Click the first "Login" button
5. Click "Log in with Example" at Dex
6. Click "Grant Access"
7. Verify I'm redirected to https://localhost:8080/test/argo-workflows/workflows
8. Go back to https://localhost:8080/test/argo-workflows/login and click "Logout"
9. Verify the `Authorization` cookie is deleted

https://github.com/user-attachments/assets/a60ddc46-8f66-492a-b2f7-fb30739477e8

### Documentation

N/A

### DIFF
--- a/ui/src/login/login.test.tsx
+++ b/ui/src/login/login.test.tsx
@@ -1,0 +1,80 @@
+import {fireEvent, render} from '@testing-library/react';
+import {createMemoryHistory, History} from 'history';
+import React from 'react';
+
+import {deleteCookie, setCookie} from '../shared/cookie';
+import {Login} from './login';
+
+jest.mock('../shared/cookie');
+
+describe('Login', () => {
+    const LoginWithHistory = (history: History) => <Login history={history} match={null} location={history.location} />;
+
+    beforeEach(() => {
+        const base = document.createElement('base');
+        base.setAttribute('href', '/');
+        document.head.appendChild(base);
+    });
+
+    afterEach(() => {
+        document.querySelector('base').remove();
+    });
+
+    describe('SSO login', () => {
+        it('button has right href', () => {
+            const {getAllByText} = render(LoginWithHistory(createMemoryHistory()));
+            const button = getAllByText('Login')[0];
+            expect(button.getAttribute('href')).toBe('/oauth2/redirect?redirect=%2Fworkflows');
+        });
+
+        it('button has right href with custom <base>', () => {
+            document.querySelector('base').setAttribute('href', '/test/');
+            const {getAllByText} = render(LoginWithHistory(createMemoryHistory()));
+
+            const button = getAllByText('Login')[0];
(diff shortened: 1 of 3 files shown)


Joibel:
@MasonM would you mind creating a cherry-pick PR for this targetting 3.7 - I believe this is needed there? I'm unsure of how to resolve login.tsx correctly and rather than guessing it felt better for you to do it.

MasonM:
@Joibel Sure: https://github.com/argoproj/argo-workflows/pull/14964

PR #14915, opened 2025-10-07
feat: artifact plugins

### LINKED ISSUE #5862: Custom artifact repository plugins

Joibel:
Fixes #5862 

### Motivation

Allow users to create their own artifact drivers.
This avoids new artifact drivers being gatekeepered by the main project, and reduces the burden of maintenance on the main project. 

### Modifications

TODO

### Verification

Unit and e2e tests added.
Manual testing of the workflows and the argo-server also performed.

### Documentation

New section in the artifacts code.
New artifact plugin documentation, which was verified and updated by @elliotgunton who wrote a plugin against the documentation.


docs/configure-artifact-repository.md
```diff
@@ -762,11 +763,171 @@ configuring the default artifact repository described previously.
...
+1. **Workflow Usage**: Reference plugins in your workflow artifact definitions
+
+#### System Configuration
+
+Plugins must be registered in the workflow controller ConfigMap under the [`artifactDrivers`](workflow-controller-configmap.md#artifactdriver) section.
+This allows the system administrator to control which plugins are available and their versions.
+
+See the [Workflow Controller ConfigMap documentation](workflow-controller-configmap.md#artifactdriver) for detailed configuration of the `artifactDrivers` field.
+
+Example workflow controller ConfigMap configuration:
```
tico24, line 794: Is it worth considering putting something to this effect into the main example workflow controller configmap file?

Joibel: Good idea, done

tico24 (approved):
Docs approved. Thanks

cmd/argoexec/commands/wait.go
```diff
@@ -79,11 +84,37 @@ func waitContainer(ctx context.Context) error {
 	logArtifacts := wfExecutor.SaveLogs(bgCtx)
 	artifacts = append(artifacts, logArtifacts...)
 
-	// Try to upsert TaskResult. If it fails, we will try to update the Pod's Annotations
 	err = wfExecutor.ReportOutputs(bgCtx, artifacts)
```
Joibel, line 89: Comment deliberately removed as it is incorrect since somewhere between 3.5-3.6

cmd/argoexec/commands/artifact_plugin_init.go
```diff
@@ -0,0 +1,80 @@
...
+							logger.WithField("signal", s).Debug(ctx, "ignore signal")
+							continue
+						}
+
+						logger.WithField("signal", s).Debug(ctx, "forwarding signal")
+						_ = osspecific.Kill(command.Process.Pid, s.(syscall.Signal))
+					}
+				}()
+			}()
+			err := loadArtifactPlugin(cmd.Context(), wfv1.ArtifactPluginName(artifactPlugin))
```
eduardodbr: should you reuse `ctx` here ?

cmd/argoexec/commands/artifact/delete.go
```diff
@@ -46,7 +51,25 @@ func NewArtifactDeleteCommand() *cobra.Command {
...
+				return err
+			}
+			pluginNamesEnv := os.Getenv(common.EnvVarArtifactPluginNames)
+			if pluginNamesEnv == "" {
+				logger.Info(ctx, "no artifact sidecars to kill")
+				return nil
+			}
+			artifactSidecars := strings.Split(pluginNamesEnv, ",")
+			logger.WithFields(logging.Fields{"artifactSidecars": artifactSidecars}).Info(ctx, "killing artifact sidecars")
+			err = em.Kill(ctx, artifactSidecars, executor.GetTerminationGracePeriodDuration())
```
eduardodbr: except for the log message, this is the same code as killArtifactSidecars() in `wait.go`, does it make sense to make it DRY ?

cmd/argoexec/commands/artifact_plugin_sidecar.go
```diff
@@ -0,0 +1,109 @@
...
+		Short: "Run an artifact plugin as a sidecar",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			exitCode := 64
+			ctx := cmd.Context()
+			logger := logging.RequireLoggerFromContext(ctx)
+
+			osspecific.AllowGrantingAccessToEveryone()
+
+			// Dir permission set to rwxrwxrwx, so that non-root containers can write exitcode and signal files.
+			if err := os.MkdirAll(varRunArgo+"/ctr/"+containerName, 0o777); err != nil {
```
eduardodbr: we could use filepath.Join() to be safer and align with the other code changes in `emissary.go`

cmd/argoexec/commands/artifact_plugin_sidecar.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			defer closer()
+			// setup signal handlers
+			signals := make(chan os.Signal, 1)
+			defer close(signals)
+			signal.Notify(signals)
+			defer signal.Reset()
+
+			defer func() {
+				err := os.WriteFile(varRunArgo+"/ctr/"+containerName+"/exitcode", []byte(strconv.Itoa(exitCode)), 0o644)
```
eduardodbr: same here, we could use use filepath.Join()

cmd/argoexec/commands/artifact_plugin_sidecar.go
```diff
@@ -0,0 +1,109 @@
...
+						s, _ := strconv.Atoi(string(data))
+						logger.WithFields(logging.Fields{
+							"signal":     s,
+							"signalPath": signalPath,
+						}).Info(ctx, "received signal")
+						if s > 0 {
+							_ = osspecific.Kill(command.Process.Pid, syscall.Signal(s))
+						}
+						time.Sleep(2 * time.Second)
+					}
```
eduardodbr: this code is the same as the one on `emissary.go`, do you think it would be worth to create a function? I understand that sometimes it is worth having them repeated but wanted to make sure

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -860,6 +860,75 @@ func (a Artifacts) GetArtifactByName(name string) *Artifact {
...
+			log.WithFields(logging.Fields{"plugin": defaultPluginName}).Debug(ctx, "adding default repo plugin")
+			plugins[defaultPluginName] = true
+		default:
+			log.WithFields(logging.Fields{"artifact": art.Name, "defaultRepo": defaultRepo}).Debug(ctx, "no plugin")
+		}
+	}
+	// TODO separate this out into a separate function
+	if includeLogs == IncludeLogs &&
+		defaultRepo != nil &&
+		defaultRepo.ArchiveLogs != nil &&
```
eduardodbr: was this forgotten or is it for another PR?

workflow/artifacts/plugin/plugin.go
```diff
@@ -0,0 +1,283 @@
...
+	client     artifact.ArtifactServiceClient
+}
+
+// NewDriver creates a new plugin artifact driver
+func NewDriver(ctx context.Context, pluginName wfv1.ArtifactPluginName, socketPath string, connectionTimeoutSeconds int32) (*Driver, error) {
+	// Check for the unix socket, retrying for up to a minute if it doesn't exist immediately
+	logger := logging.RequireLoggerFromContext(ctx)
+
+	// Try for up to 120 seconds, checking once per second
+	const maxRetries = 120
```
eduardodbr, line 33: log says it is retrying for a minute but code is setting up 120 retries (2 minutes)

workflow/artifacts/plugin/plugin.go
```diff
@@ -0,0 +1,283 @@
...
+	var info os.FileInfo
+	var statErr error
+	var socketExists bool
+
+	for retry := 0; retry < maxRetries; retry++ {
+		info, statErr = os.Stat(socketPath)
+		if statErr == nil {
```
eduardodbr, line 40: nit: normally I would suggest using a retry lib instead of implementing it. However the only retry lib in the dependencies is k8s util/wait which I understand could feel weird to use here

workflow/artifacts/plugin/plugin.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(connectionTimeoutSeconds)*time.Second)
+	defer cancel()
+
+	conn.Connect()
+
+	// Wait for the connection to be ready within the timeout.
+	if !conn.WaitForStateChange(ctx, connectivity.Idle) {
+		_ = conn.Close()
+		return nil, fmt.Errorf("timeout waiting for plugin %s connection to become active (socket=%q)", pluginName, socketPath)
```
eduardodbr: comment says it is waiting for the connection to be ready, but function is only checking if state changed from `Idle` to any other. Are we sure we want to continue if state changes from `Idle` to any other then `Ready`? maybe thats not a possible scenario?

workflow/executor/executor.go
```diff
@@ -1255,22 +1279,27 @@ func (we *WorkflowExecutor) monitorDeadline(ctx context.Context, containerNames
...
-	logger.Info(ctx, "Starting deadline monitor")
+	logger.WithField("containers", containerNames).Info(ctx, "Starting deadline monitor")
 	select {
 	case <-ctx.Done():
 		logger.Info(ctx, "Deadline monitor stopped")
 		return
 	case <-deadlineExceeded:
 		message = "Step exceeded its deadline"
 	}
+	// TODO: Handle not killing the artifact sidecars
```
eduardodbr: it this a forgotten comment? this is not killing plugin sidecars and they are being killed later by the wait container in killArtifactSidecars()

cmd/argoexec/commands/wait.go
```diff
@@ -79,11 +84,37 @@ func waitContainer(ctx context.Context) error {
...
 	if err != nil {
 		wfExecutor.AddError(ctx, err)
 	}
 
+	err = killArtifactSidecars(bgCtx)
+	if err != nil {
+		wfExecutor.AddError(ctx, err)
+	}
```
eduardodbr: this is just a question, probably due to lack of knowledge on my end. Why cant this be a method of wfExecutor, like `wfExecutor.KillArtifactSidecars(bgCtx)` ?

cmd/argoexec/commands/artifact_plugin_init.go
Joibel: Yes, done.

workflow/artifacts/plugin/plugin.go
Joibel: Fixed (changed comment)

workflow/executor/executor.go
Joibel: Yep, forgotten comment, I thought I'd removed that one!

cmd/argoexec/commands/artifact_plugin_sidecar.go
Joibel: :+1: I've moved it to be common.

cmd/argoexec/commands/artifact_plugin_sidecar.go
Joibel: Done

cmd/argoexec/commands/artifact_plugin_sidecar.go
Joibel: Done

cmd/argoexec/commands/artifact/delete.go
Joibel: Will do, good plan

cmd/argoexec/commands/wait.go
Joibel: I've done this as part of your first comment.

workflow/artifacts/plugin/plugin.go
Joibel: Our `util/wait` only has backoff in it, and we really don't want a backoff strategy here, we just want this simple stupid polling as this is holding off all work being done.

workflow/artifacts/plugin/plugin.go
Joibel: We're waiting for the connection to go out of Idle here, which from my testing means there is something on the other end. It doesn't verify that it's a useful something, but that's all we're really concerned about at this stage - we defer failure in the protocol to other points in the process. Updated the comment.

pkg/apis/workflow/v1alpha1/workflow_types.go
Joibel: It was forgotten. Thanks, done now

docs/artifact-plugin.md
```diff
@@ -0,0 +1,317 @@
...
+### 3. Implement Required Methods
+
+Your GRPC server must implement these six methods from the `ArtifactService` interface:
+
+#### Required Methods
+
+1. **`Load(LoadArtifactRequest)` → `LoadArtifactResponse`**
+   - Downloads an artifact from your storage system to the specified local path
+   - Called during workflow execution to retrieve input artifacts
+
```
eduardodbr, line 139: this is not being rendered as expected:

image

I tried rebuilding with a tab and it works.

docs/configure-artifact-repository.md
```diff
@@ -762,11 +763,171 @@ configuring the default artifact repository described previously.
...
+- Each plugin requires its own dedicated volume and mount path at `/tmp/artifact-plugins/{plugin-name}`.
+- The socket path must be `/tmp/artifact-plugins/{plugin-name}/socket` where `{plugin-name}` exactly matches the `name` field in the workflow controller's `artifactDrivers` configuration.
+- Both the argo-server container and the plugin sidecar must mount the same volume at the same path.
+- The socket file name must be `socket` (not arbitrary) to match what the code expects.
+- Use `emptyDir` volumes as they provide efficient shared storage for Unix sockets.
+
+#### Workflow Usage
+
+Once registered, you can use plugins in your workflows by specifying the `plugin` field in artifact definitions:
+
```
eduardodbr, line 868: It might be helpful to include a reference to [key-only artifacts](https://argo-workflows.readthedocs.io/en/latest/key-only-artifacts/) to let users know they can configure the plugin in the controller config map and then use only the key in the workflow.

docs/artifact-plugin.md
Joibel: Ok, I think I've fixed this now. Good spot :eagle: 

image

docs/configure-artifact-repository.md
Joibel: Added

.features/pending/artifact-plugins.md
```diff
@@ -0,0 +1,8 @@
+Description: Artifact Drivers as plugins
+Author: [Alan Clucas](https://github.com/Joibel), [JP Zivalich](https://github.com/JPZ13), [Elliot Gunton](https://github.com/elliotgunton)
```
terrytangyuan, line 2:
```suggestion
Authors: [Alan Clucas](https://github.com/Joibel), [JP Zivalich](https://github.com/JPZ13), [Elliot Gunton](https://github.com/elliotgunton)
```

.features/pending/artifact-plugins.md
```diff
@@ -0,0 +1,8 @@
...
+Component: General
+Issues: 5862
+
+Artifact Drivers can now be added via a plugin mechanism.
+You can write a GRPC server which acts as an artifact driver to upload and download artifacts to a repository, and supply that as a docker image.
```
terrytangyuan:
```suggestion
You can write a GRPC server which acts as an artifact driver to upload and download artifacts to a repository, and supply that as a container image.
```

.features/pending/artifact-plugins.md
Joibel: Currently the feature tool requires this to be singular. This should probably be acceptable syntax, but isn't right now, so won't fix here.

cmd/argoexec/commands/signal.go
```diff
@@ -0,0 +1,48 @@
...
+				if err != nil || s <= 0 {
+					continue
+				}
+				_ = os.Remove(signalPath)
+				logger.WithFields(logging.Fields{
+					"signal":     s,
+					"signalPath": signalPath,
+				}).Info(ctx, "received signal")
+				_ = osspecific.Kill(pid, syscall.Signal(s))
+			}
```
isubasinghe, line 45: Why do we not break after the kill is sent? The os.Readfile is guaranteed to fail due to the `os.Remove`.

server/apiserver/argoserver.go
```diff
@@ -449,6 +470,111 @@ func (as *argoServer) newHTTPServer(ctx context.Context, port int, artifactServe
...
+	errorChannel := make(chan error, len(cfg.ArtifactDrivers))
+
+	// Validate each driver connection in parallel
+	for _, driver := range cfg.ArtifactDrivers {
+		wg.Add(1)
+		go func(driver config.ArtifactDriver) {
+			defer wg.Done()
+
+			// Create a new driver connection
+			pluginDriver, err := plugin.NewDriver(ctx, driver.Name, driver.Name.SocketPath(), 5) // replace with driver.ConnectionTimeoutSeconds once we have it
```
isubasinghe: Could we move this to a const for now ?
Should we also bump this up ? We have no idea how long this could take for certain artifact plugins.

workflow/artifacts/plugin/plugin.go
```diff
@@ -0,0 +1,283 @@
...
+	if connectionTimeoutSeconds == 0 {
+		connectionTimeoutSeconds = defaultConnectionTimeoutSeconds
+	}
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(connectionTimeoutSeconds)*time.Second)
+	defer cancel()
+
+	conn.Connect()
+
+	// Wait for the connection to be possibly ready within the timeout.
+	if !conn.WaitForStateChange(ctx, connectivity.Idle) {
```
isubasinghe: This isn't quite what we need though, this will pass if `TransientFailure` or `Shutdown`  is the next state.
We should be monitoring state to check if we hit `Ready`. 

Having looked at the code for the grpc client, we can hit `TransientFailure` if the addresses are unreachable.

Not being in `Idle` doesn't give us assurance that there is something on the other end.

Why not call Invoke on a dummy method and then wait for an error response? If the error is a network error, then we didn't connect, if the error is an `Unimplemented` error then we know that the endpoint exists.

isubasinghe (changes requested):
Minor issues really.

cmd/argoexec/commands/signal.go
Joibel: This is basically copied from https://github.com/argoproj/argo-workflows/blob/0f74e76011d5758aa15544a07a39cb221fef0cd5/cmd/argoexec/commands/emissary.go#L173.
It is now used in multiple places, hence making this common function.

It is slightly tidied to use a `ticker` rather than a `sleep`, but is trying to achieve the same thing, and should be functionally the same.

`os.ReadFile` will repeatedly fail every 2 seconds under normal conditions, but could get a SIGTERM and then say 10 seconds later a SIGKILL from the same file, hence not `break`ing.

server/apiserver/argoserver.go
Joibel: I thought we'd already fixed this. I've made a separate commit in the chain to show how this has changed the codebase to save too much re-reviewing. Basically we need drivers to have this field, but also allow it for artifacts - at this stage there are no artifacts, just drivers.

We also need transport for the field from the driver configuration in the controller to argoexec in the pods, so the field in artifacts remains.

Joibel: We don't know how long it will take, except for the timeout. But holding off startup whilst we test is by design. They are tested in parallel, not series.

workflow/artifacts/plugin/plugin.go
Joibel: I've made sure it reaches Ready now. I don't see that validating we get Unimplemented is worth the extra code complexity - this is theoretically only a sanity check for sysadmins, it's not supposed to be ensuring plugin authors have written a useful plugin.

.features/pending/artifact-plugins.md
Joibel: See #14958

PR #14943, opened 2025-10-20
chore: replace golang.org/x/exp/maps with standard library maps package

shota3506:
Replace the experimental `golang.org/x/exp/maps` package with the standard library `maps` package

### Motivation

  - Use stable standard library APIs
  - Reduce dependency on experimental packages
 
### Modifications

- use standard `maps` and `slices` package instead of experimental package
- update go.mod

### DIFF
--- a/workflow/controller/artifact_gc.go
+++ b/workflow/controller/artifact_gc.go
@@ -6,7 +6,6 @@ import (
 	"hash/fnv"
 	"maps"
 	"slices"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Joibel:
Thanks @shota3506 - is your aim to run through all of [`modernize`](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize)? If so, once you're done I'd also appreciate a PR to run `modernize` as part of linting.

shota3506:
hi @Joibel ✋ 
As far as I know, migrating from the exp package to the std package is not within the scope of gopls modernize for now.

But yes, I agree it would be a great idea to run modernize in linting. Would it be okay if I do that as a separate task?

shota3506:
Let me add modernize analyzer after next golangci-lint minor version release
https://github.com/golangci/golangci-lint/pull/6126

Joibel:
> Let me add modernize analyzer after next golangci-lint minor version release [golangci/golangci-lint#6126](https://github.com/golangci/golangci-lint/pull/6126)

Oh, that's nice that it's being added there, hadn't seen that.

Joibel:
> hi @Joibel ✋ As far as I know, migrating from the exp package to the std package is not within the scope of gopls modernize for now.
> 
> But yes, I agree it would be a great idea to run modernize in linting. Would it be okay if I do that as a separate task?

Separate task would be best, yes please. Possibly split up some of the fixes it provides from actually enabling it in linting too, just so the PRs are more reviewable.

shota3506:
@Joibel Okay, I’ll open some PRs to address modernize alerts in the existing codebase before enabling it in lint.

This PR is independent of that, so please review it when you have a moment. (I assume the CI errors are unrelated to these changes?)

PR #14949, opened 2025-10-20
feat: cache updated workflow to avoid reconciling outdated objects. Fixes: #14833 #14780

### LINKED ISSUE #12352: Add some doc about environment variable "INFORMER_WRITE_BACK"
# Summary

I'd like to add some comments at environment variable "INFORMER_WRITE_BACK".
documentation description now is INFORMER_WRITE_BACK: "Whether to write back to informer instead of catching up." until 2023-12-12.

here is my scene
I have a controller that turn the Workflow Spec.Suspend from true to false. In some race cases, I found my changes were not caught by Workflow Controller, because in Workflow Controller, I see "workflow suspended" log after my controller reported "change suspend from true to false".
so I checked workflow code,  https://github.com/argoproj/argo-workflows/blob/23292c429430e71fdc7c59c89878053ee821b80d/workflow/controller/operator.go#L737. I believe my controller wrote  Spec and Workflow Controller wrote status, so they were both accepted by kubernetes Apiserver. So this https://github.com/argoproj/argo-workflows/blob/23292c429430e71fdc7c59c89878053ee821b80d/workflow/controller/operator.go#L739C2-L739C2 Warning didn't caught by Workflow Controller log.
Finally, my Spec event first reported to Informer and synced with Store, and then Workflow Controller write workflow back into Informer which cover my Spec changes https://github.com/argoproj/argo-workflows/blob/23292c429430e71fdc7c59c89878053ee821b80d/workflow/controller/operator.go#L776C19-L776C19

Since I turn "INFORMER_WRITE_BACK" to false, I see no missing change so far for 4 days, so I think my judging is right. This environment variable could result in the lost of consistency between Workflow Controller Store Cache and Kubernetes Apiserver in this race condition.  Any suggestion from Community?
If any improve will be made, I'd like to do some contribute and be a naive contributor

## Use Cases

When would you use this?

---

**Beyond this issue**:

### LINKED ISSUE #14780: WriteBack Informer not Take effect
### Pre-requisites

I find the same workflow reconceil logs:
```
time="2025-08-21T23:15:21.713Z" level=info msg="Workflow update successful" namespace=test phase=Running resourceVersion=607575187
time="2025-08-21T23:16:48.142Z" level=info msg="Processing workflow" Phase=Pending ResourceVersion=592494448 namespace=test
```

### Version(s)

v3.5.15

```YAML
large number (10w+) workflows
```

```text
```

```text
```

### LINKED ISSUE #14833: Pod Created twice when cluster under pressure
### Pre-requisites

The log like this, the same pod created twice:
```
time="2025-09-09T16:35:08.922Z" level=info msg="Created pod: workflow-template-dag-wsklb.B.D (workflow-template-dag-wsklb-echo-19727502)" namespace=default workflow=workflow-template-dag-wsklb
time="2025-09-09T16:29:34.050Z" level=info msg="cleaning up pod" action=deletePod key=default/workflow-template-dag-wsklb-echo-19727502/deletePod
time="2025-09-09T16:20:22.287Z" level=info msg="Created pod: workflow-template-dag-wsklb.B.D (workflow-template-dag-wsklb-echo-19727502)" namespace=default workflow=workflow-template-dag-wsklb
```

### Version(s)

v3.5.13

```YAML
1w+ workflows in cluster.
```

```text
```

```text
```

shuangkun:
Create new cache instead of writeInformerBack.

Fixes #14833
Fixes #14780
Fixes #12352


eduardodbr:
This PR is very similar to #14532 , both are adding a new cache in front of the informar cache, which  might feel weird. Please review and join the discussion if you have a different opinion

shuangkun:
> This PR is very similar to #14532 , both are adding a new cache in front of the informar cache, which might feel weird. Please review and join the discussion if you have a different opinion

Oh, the problems we're trying to solve may be somewhat different, but there's also some overlap, such as double pod creation. I'm trying to address controller informer anomalies at very large scales (300,000+ workflows), such as event overwriting and workflow controllers missing workflows. Therefore, I'd like to remove the "writeInformerBack" method, which is the official recommended approach: instead of writing to the same InformerCache, create new expectations.

Joibel:
Thanks for this, and sorry it sat so long.

Both sides of this PR have since been handled on main in other ways:

- `INFORMER_WRITE_BACK` and `writeBackToInformer()` were removed outright
in #15627 (after being disabled by default in #15079), so the deletions here no longer apply to any existing code.
- Reconciling stale workflow objects is now handled by the
`lastWrittenVersions` / `recordWorkflowWrite` resourceVersion comparison added in #16357, rather than by a second cache in front of the informer.

Accordingly #14833 and #14780 were both closed as fixed. I'm closing this as obsolete - if you're still seeing stale-object reconciliation at very large scale (300k+ workflows) on current main, please open a fresh issue with details and we'll look at it against the new mechanism.

PR #14958, opened 2025-10-21
feat: update featuregen to use plural "Authors:" field

Joibel:
Vibe coded by claude web UI as an experiment

### Motivation

https://github.com/argoproj/argo-workflows/pull/14915#discussion_r2440256419

### Modifications

- Changed TEMPLATE.md to use "Authors:" instead of "Author:"
- Updated metadataFields to include both "Authors" and "Author" for backwards compatibility
- Updated validation to accept either "Authors:" or "Author:" field
- Updated all tests to use "Authors:" field
- Added test case for backwards compatibility with old "Author:" field

### Verification

Built in tests

### Documentation

Self documenting


Joibel:
> Cool :-) Is this working as expected?

Yes, seems to be. It's got a reasonable test suite, and the PR itself is a feature with "Authors:" in the feature note.

PR #14662, opened 2025-07-18
refactor(ui): Similar login page to argo cd. Fixes #10816

jiwlee97:
Fixes #10816 

### Motivation
The login page was structured like a subpage.

### Modifications
Transformed the login page to feel more like a standalone “gate” rather than a subpage.
Also updated the UI to resemble Argo CD’s login page for visual consistency.

### Verification
I tested locally.

after changes:
Screenshot 2025-07-19 at 1 23 12 AM

before changes:
image

PR #14964, opened 2025-10-22
fix(ui): login/logout bug when base HREF set. Fixes #14897 (cherry-pick #14909 for 3.7)

### LINKED ISSUE #14897: SSO Login from Login URL ends up in 404 Not Found

MasonM:
### Motivation
Cherry-picks #14909 for `release-3.7`.

### Modifications
Manually resolved conflicts with `ui/src/login/login.tsx` due to https://github.com/argoproj/argo-workflows/pull/14662 not being in `release-3.7`. Also had to tweak `ui/src/login/login.test.tsx` so the tests pass.

### Verification

Followed the same verification steps as in https://github.com/argoproj/argo-workflows/pull/14909

### Documentation
N/A

### DIFF
--- a/ui/src/login/login.test.tsx
+++ b/ui/src/login/login.test.tsx
@@ -1,0 +1,80 @@
+import {fireEvent, render} from '@testing-library/react';
+import {createMemoryHistory, History} from 'history';
+import React from 'react';
+
+import {deleteCookie, setCookie} from '../shared/cookie';
+import {Login} from './login';
+
+jest.mock('../shared/cookie');
+
+describe('Login', () => {
+    const LoginWithHistory = (history: History) => <Login history={history} match={null} location={history.location} />;
+
+    beforeEach(() => {
+        const base = document.createElement('base');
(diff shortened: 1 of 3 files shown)


Joibel:
Thank you.

PR #14979, opened 2025-10-23
fix: enforce metrics attributes, also fixing one doc

Joibel:
### Motivation

Follow up to #13943 where I mention we could enforce attributes as documented in values.yaml in the golang side to ensure they lined up.

The discovery of #14976 lead me to believe this was worth doing now.
It has discovered one missing attribute - see the docs change.

### Modifications

Vibed with my mate claude from anthropic to get a generator for helper functions for all builtin metrics types.

Then use those throughout the metrics codebase.

### Verification

CI will check the e2e tests, but unit tests pass without modification.

### Documentation

Fixed docs for one metric. Otherwise no user facing changes


util/telemetry/builder/metrics_helpers.go
```diff
@@ -0,0 +1,192 @@
...
+		return
+	}
+	defer f.Close()
+
+	fmt.Fprintf(f, "%s\n", generatedBanner)
+	fmt.Fprintf(f, "package telemetry\n\n")
+	fmt.Fprintf(f, "import (\n")
+	fmt.Fprintf(f, "\t\"context\"\n\n")
+	fmt.Fprintf(f, "\t\"go.opentelemetry.io/otel/metric\"\n")
+	fmt.Fprintf(f, ")\n\n")
```
isubasinghe: Do we have to individually do this for every line? Maybe move this into a single text bit, it will be cleaner to read

util/telemetry/builder/metrics_helpers.go
```diff
@@ -0,0 +1,192 @@
...
+
+			fmt.Fprintf(f, "// %s sets the %s attribute\n", funcName, attribDef.displayName())
+			fmt.Fprintf(f, "func %s(%s %s) %s {\n", funcName, paramName, paramType, optionTypeName)
+			fmt.Fprintf(f, "\treturn func(a *InstAttribs) {\n")
+			fmt.Fprintf(f, "\t\t*a = append(*a, InstAttrib{\n")
+			fmt.Fprintf(f, "\t\t\tName:  Attrib%s,\n", attr.Name)
+			fmt.Fprintf(f, "\t\t\tValue: %s,\n", paramName)
+			fmt.Fprintf(f, "\t\t})\n")
+			fmt.Fprintf(f, "\t}\n")
+			fmt.Fprintf(f, "}\n\n")
```
isubasinghe: Is there a cleaner way of manipulating the AST ? 
Templates could be a good middle ground between this and using something like: https://github.com/dave/jennifer

isubasinghe (approved):
No massive issues

util/telemetry/builder/metrics_helpers.go
Joibel: I'll redo these with `text/template`

util/telemetry/builder/metrics_helpers.go
Joibel: `jennifer` looks interesting but a bit abandoned (albeit something that doesn't really need much maintainence). I'll use `text/template`

Issue #14977, opened 2025-10-23
Remove deprecated singular CRD entries


Joibel:
As part of 3.6 we deprecated 4 features:

* `podPriority` doesn't work any modern kubernetes and has been deprecated for a while
* `schedule` is replaced by `schedules` in cron workflows - this is a trivial update to make to the yaml
* `mutex` and `semaphore` are replaced by the pluralised version - this is a trivial update to make to the yaml

Removing legacy code reduces maintainer burden and the chance for errors.

Joibel:
Completed

PR #14996, opened 2025-10-29
feat: CLI convert command

Joibel:
Part of #14977 

### Motivation

An assistant to help upgrade workflows/cron/templates so that everyone doesn't have to write this this separately.

### Modifications

Added a new CLI command which is based around the lint command in structure and the convert command from kubectl.

### Verification

New e2e test cases

### Documentation

Documentation auto added and feature note added.

Deprecations.md updated - (spreading this change over multiple PRs would make it into a very messy merge)

### DIFF
--- a/test/e2e/cli_test.go
+++ b/test/e2e/cli_test.go
@@ -2077,17 +2077,7 @@ func (s *CLISuite) TestWorkflowConvert() {
 		})
 	})
 	s.Run("ConvertStdin", func() {
-		tmp, err := os.CreateTemp("", "convert-test-*.yaml")
-		s.CheckError(err)
-		defer os.Remove(tmp.Name())
-
-		data, err := os.ReadFile("testdata/convert/convert-legacy-cron.yaml")
-		s.CheckError(err)
-		_, err = tmp.Write(data)
-		s.CheckError(err)
-		tmp.Close()
-
-		s.Given().RunCli([]string{"convert", tmp.Name()}, func(t *testing.T, output string, err error) {
+		s.Given().RunCliStdin([]string{"convert", "-"}, "testdata/convert/convert-legacy-cron.yaml", func(t *testing.T, output string, err error) {
 			require.NoError(t, err)
 			assert.Contains(t, output, "schedules:")
 			assert.Contains(t, output, "- 0 0 * * *")
--- a/test/e2e/fixtures/given.go
+++ b/test/e2e/fixtures/given.go
@@ -221,10 +221,10 @@ var OutputRegexp = func(rx string) func(t *testing.T, output string, err error) 
 	}
 }
 
-func (g *Given) Exec(name string, args []string, block func(t *testing.T, output string, err error)) *Given {
+func (g *Given) Exec(name string, args []string, stdin string, block func(t *testing.T, output string, err error)) *Given {
 	g.t.Helper()
 	ctx := logging.TestContext(g.t.Context())
-	output, err := Exec(ctx, name, args...)
+	output, err := Exec(ctx, name, stdin, args...)
 	block(g.t, output, err)
 	return g
 }
@@ -232,11 +232,15 @@ func (g *Given) Exec(name string, args []string, block func(t *testing.T, output
 // Use Kubectl to server-side apply the given file
 func (g *Given) KubectlApply(file string, block func(t *testing.T, output string, err error)) *Given {
 	g.t.Helper()
-	return g.Exec("kubectl", append([]string{"-n", Namespace, "apply", "--server-side", "-f"}, file), block)
+	return g.Exec("kubectl", append([]string{"-n", Namespace, "apply", "--server-side", "-f"}, file), "", block)
 }
 
 func (g *Given) RunCli(args []string, block func(t *testing.T, output string, err error)) *Given {
-	return g.Exec("../../dist/argo", append([]string{"-n", Namespace}, args...), block)
+	return g.RunCliStdin(args, "", block)
+}
+
+func (g *Given) RunCliStdin(args []string, stdin string, block func(t *testing.T, output string, err error)) *Given {
+	return g.Exec("../../dist/argo", append([]string{"-n", Namespace}, args...), stdin, block)
 }
 
 func (g *Given) ClusterWorkflowTemplate(text string) *Given {
--- a/test/e2e/fixtures/then.go
+++ b/test/e2e/fixtures/then.go
@@ -321,7 +321,7 @@ func (t *Then) ExpectWorkflowTaskSet(block func(t *testing.T, wfts *wfv1.Workflo
 func (t *Then) RunCli(args []string, block func(t *testing.T, output string, err error)) *Then {
 	t.t.Helper()
 	ctx := logging.TestContext(t.t.Context())
-	output, err := Exec(ctx, "../../dist/argo", append([]string{"-n", Namespace}, args...)...)
+	output, err := Exec(ctx, "../../dist/argo", "", append([]string{"-n", Namespace}, args...)...)
 	block(t.t, output, err)
 	return t
 }
--- a/test/e2e/fixtures/util.go
+++ b/test/e2e/fixtures/util.go
@@ -18,8 +18,15 @@ func errorln(args ...interface{}) {
 	_, _ = fmt.Fprint(os.Stderr, args...)
 }
 
-func Exec(ctx context.Context, name string, args ...string) (string, error) {
+func Exec(ctx context.Context, name string, stdin string, args ...string) (string, error) {
 	cmd := exec.CommandContext(ctx, name, args...)
+	if stdin != "" {
+		data, err := os.ReadFile(stdin)
+		if err != nil {
+			return "", err
+		}
+		cmd.Stdin = bytes.NewReader(data)
+	}
 	cmd.Env = os.Environ()
 	_, _ = fmt.Println(cmd.String())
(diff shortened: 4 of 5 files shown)


cmd/argo/commands/convert.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return outputObject(converted, outputFormat, isJSON)
+}
+
+func outputObject(obj interface{}, format string, preferJSON bool) error {
+	var outBytes []byte
+	var err error
+
+	// If input was JSON and format is yaml, convert to JSON
```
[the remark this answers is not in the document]
Joibel: Updated

workflow/convert/legacy_types.go
```diff
@@ -0,0 +1,195 @@
...
+	if ls.Mutex != nil && len(sync.Mutexes) == 0 {
+		sync.Mutexes = []*wfv1.Mutex{ls.Mutex}
+	}
+
+	return sync
+}
+
+// LegacyTemplate wraps Template with legacy synchronization support
+type LegacyTemplate struct {
+	wfv1.Template
```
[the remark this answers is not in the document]
Joibel: Kept, but comments updated for reasoning.

workflow/convert/legacy_types.go
```diff
@@ -0,0 +1,195 @@
...
+	tmpl := lt.Template
+	if lt.Synchronization != nil {
+		tmpl.Synchronization = lt.Synchronization.ToCurrent()
+	}
+	return tmpl
+}
+
+// LegacyWorkflowSpec wraps WorkflowSpec with legacy synchronization support
+type LegacyWorkflowSpec struct {
+	wfv1.WorkflowSpec
```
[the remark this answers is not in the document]
Joibel: As previous

.features/pending/convert.md
```diff
@@ -0,0 +1,9 @@
+Description: `convert` CLI command to convert to new workflow format
+Authors: [Alan Clucas](https://github.com/Joibel)
+Component: CLI
+Issues: 14977
+
+A new CLI command `convert` which will convert Workflows, CronWorkflows, and (Cluster)WorkflowTemplates to the new format.
```
elliotgunton, line 6: I would suggest `migrate` as the CLI command name. `convert` sounds like it might convert between `Workflows`/`CronWorkflows`/etc. (Might even be a good feature for later.)

Joibel: In the PR description I mention the name comes from `kubectl convert` which 
> Convert config files between different API versions. Both YAML and JSON formats are accepted. Note - requires kubectl-convert plugin to be installed.

cmd/argo/commands/convert.go
```diff
@@ -0,0 +1,172 @@
...
+var yamlSeparator = regexp.MustCompile(`\n---`)
+
+func runConvert(ctx context.Context, args []string, output string) error {
+	for _, file := range args {
+		err := fileutil.WalkManifests(ctx, file, func(path string, data []byte) error {
+			isJSON := jsonpkg.IsJSON(data)
+
+			if isJSON {
+				// Parse single JSON document
+				if err := convertDocument(data, output, isJSON); err != nil {
```
elliotgunton: nit: Might as well skip the isJSON variable

```suggestion
			if jsonpkg.IsJSON(data) {
				// Parse single JSON document
				if err := convertDocument(data, output, true); err != nil {
```

docs/cli/argo.md
```diff
@@ -109,6 +109,7 @@ argo [flags]
 * [argo auth](argo_auth.md)	 - manage authentication settings
 * [argo cluster-template](argo_cluster-template.md)	 - manipulate cluster workflow templates
 * [argo completion](argo_completion.md)	 - output shell completion code for the specified shell (bash, zsh or fish)
+* [argo convert](argo_convert.md)	 - convert workflow manifests from legacy format to current format
```
elliotgunton, line 112: Very much a nit: inconsequential in a `.md` file but there's a stray tab here. The rest are double-spaced
```suggestion
* [argo convert](argo_convert.md)  - convert workflow manifests from legacy format to current format
```

workflow/convert/legacy_types.go
```diff
@@ -0,0 +1,202 @@
...
+	// Deprecated v3.5 and before: singular mutex
+	Mutex *wfv1.Mutex `json:"mutex,omitempty"`
+	// v3.6 and after: plural semaphores
+	Semaphores []*wfv1.SemaphoreRef `json:"semaphores,omitempty"`
+	// v3.6 and after: plural mutexes
+	Mutexes []*wfv1.Mutex `json:"mutexes,omitempty"`
+}
+
+// ToCurrent converts a LegacySynchronization to the current Synchronization type
+func (ls *LegacySynchronization) ToCurrent() *wfv1.Synchronization {
```
elliotgunton, line 26: I'm a bit concerned about the general approach being time/context dependent, i.e. "ToCurrent" rather than "from version X to Y", and also the "Legacy" wording could mean anything in a couple of years (unless we think "Legacy"=="not current").

I don't think these fields will change again but do we plan to build on the `convert` command for later migrations/deprecations? Will we update `ToCurrent` functions directly or will we use a series of functions that do `3.5 -> 3.7+ -> 4.something -> 5.something-else`?

elliotgunton (commented):
Code LGTM, couple of nits and a general question 🚀

docs/cli/argo.md
Joibel: This is autogenerated, so shouldn't be touched.

workflow/convert/legacy_types.go
Joibel: This tool, the argocli, is released as a versioned component of the system.

My idea was that convert would always bring it up to date with the version of the tool that you're using. If 5.0 has more breaking changes in it, if you run argo v5.0 you'll get it updated to 5.0 version of the CRD. I wouldn't provide a command or parameter to allow argo 5 to only bring the CRD up to version 4's CRD.

The implementation of `ToCurrent()` can be modified in the future - if the best way to do that is to have separate functions for 3->4 and 4->5 that's whoever implements it's choice.

In theory we should be versioning the CRDs, but I'm not considering doing that at the moment.

cmd/argo/commands/convert.go
```diff
@@ -0,0 +1,170 @@
...
+		output = common.EnumFlagValue{
+			AllowedValues: []string{"yaml", "json"},
+			Value:         "yaml",
+		}
+	)
+
+	command := &cobra.Command{
+		Use:   "convert FILE...",
+		Short: "convert workflow manifests from legacy format to current format",
+		Long:  "Converts singular schedule, mutex and semaphore to the new plural version",
```
elliotgunton: Nitting
```suggestion
		Long:  "Converts singular schedule, mutex and semaphore fields to the new plural versions",
```

IMO it's also worth saying "Only supports outputting to stdout" -- otherwise the behaviour when converting a directory is not necessarily obvious.

cmd/argo/commands/convert.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+	}
+
+	command.Flags().VarP(&output, "output", "o", "Output format. "+output.Usage())
+	command.Flags().BoolVar(&common.NoColor, "no-color", false, "Disable colorized output")
+
+	return command
+}
+
+var yamlSeparator = regexp.MustCompile(`\n---`)
```
elliotgunton, line 58: Could this be `const`?
```suggestion
const yamlSeparator = regexp.MustCompile(`\n---`)
```

It's only used once below so I would even suggest moving it into the function. Plus there's the "Split YAML documents" comment to describe it

Joibel: MustCompile is a function which takes time, by making it an evil global we save running it every time, and breaking it will break all of argocli which makes it easier to detect you've got it wrong - the panic will happen at program startup. It's a common pattern to have it this way, even if it involves an unnecessary (package scoped) global.

We can't make this `const` as the result of `MustCompile` isn't const in the eyes of the compiler. Essentially the compiler can't run it at compile time, and must wait until runtime.

cmd/argo/commands/convert.go
Joibel: Done

test/e2e/cli_test.go
```diff
@@ -2037,6 +2037,72 @@ func (s *CLISuite) TestPluginStruct() {
...
+			lines := strings.Split(strings.TrimSpace(output), "\n")
+			require.NotEmpty(t, lines)
+			err = json.Unmarshal([]byte(lines[0]), &cronWf)
+			require.NoError(t, err)
+			// Verify conversion happened - legacy schedule field should be converted to schedules array
+			assert.Len(t, cronWf.Spec.Schedules, 1)
+			assert.Equal(t, "0 0 * * *", cronWf.Spec.Schedules[0])
+		})
+	})
+	s.Run("ConvertStdin", func() {
```
MasonM, line 2079: minor: This isn't really testing stdin, since it's creating a temp file and reading for that. Here's a commit that changes this test to actually test from stdin: https://github.com/MasonM/argo-workflows/commit/fa9f1831bf54bc91015499b10c80e127d94fa1f2

MasonM (approved):

test/e2e/cli_test.go
Joibel: Thanks for doing that, I've cherry-picked it into this PR.

PR #15003, opened 2025-11-03
chore(manifests): switch to full CRDs by default

MasonM:
### Motivation

After the work in https://github.com/argoproj/argo-workflows/pull/14044, the full CRDs are now usable, and no issues have been reported. This switches the install manifests to use them by default.

### Modifications

Update the the `kustomization.yaml` files to reference the full CRDs
### Verification

Ran `make manifests-validate` (which is also run in CI builds)
### Documentation
* Moved the "Full CRDs" section to a more appropriate part of the page and gave more details: https://argo-workflows--15003.org.readthedocs.build/en/15003/installation/#full-crds
* Updated the upgrading guide: https://argo-workflows--15003.org.readthedocs.build/en/15003/upgrading/#full-crds

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,6 +21,13 @@ You can use Kustomize to patch your preferred [configurations](managed-namespace
     `latest` is the tip of the `main` branch and may not be stable.
     In production, you should use a specific release version.
 
+#### Full CRDs
+
+As of version 4.0, the official release manifests use CRDs with full validation information.
+They must be applied using [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) to workaround [Kubernetes size limits](https://github.com/kubernetes/kubernetes/issues/82292).
+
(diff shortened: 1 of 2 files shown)


docs/installation.md
```diff
@@ -25,16 +25,6 @@ You can use Kustomize to patch your preferred [configurations](managed-namespace
...
-### Full CRDs
-
-The official release manifests come with stripped-down CRDs that omit validation information.
-This is a workaround for [Kubernetes size limitations](https://github.com/kubernetes/kubernetes/issues/82292) when using client-side apply.
-As of version 3.7, the full CRDs can be installed using [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) via the following command:
-
-```bash
-kubectl apply --server-side --kustomize https://github.com/argoproj/argo-workflows/manifests/base/crds/full?ref=v3.7.0
-```
-
```
Joibel, line 37: I think we need to leave in instructions for using SSA to install (perhaps on the releases page), otherwise people will struggle. Perhaps we should also have the inverted instructions for the minified CRDs here, at least until we discover if anyone needs them.

PR #15010, opened 2025-11-07
docs: Update Instance ID documentation

scottmelhop:
### Motivation

The use of Instance ID is not super clear from the current docs, hopefully this helps :)

### Modifications

Adjusted and added to the Instance ID section, and linked to it from the workflow controller configmap docs.


Joibel:
Not cherry-picking for complexity.

PR #15013, opened 2025-11-10
chore(ci): fix up generated docs in CI

Joibel:
PR #15010 made me realise we don't have all the necessary entries in place to do codegen - the PR should have failed as that section of workflow-controller-configmap.md is generated.

I've fixed the ones which seemed obvious to me.


gianarb (approved):
:+1:

isubasinghe (approved):

PR #15030, opened 2025-11-11
fix!: return latest workflow for retried-persisted workflows

### LINKED ISSUE #15015: Retried "persisted" workflows are not fetchable from the API
### Pre-requisites

- Have archiving to postgres setup in Workflow Controller (use the `quick-start-postgres.yaml` from [the release](https://github.com/argoproj/argo-workflows/releases/tag/v3.7.3))
- Run any basic workflow, and you will see `curl -k https://localhost:2746/api/v1/workflows/argo/<workflow-name>` works
- Wait for it to complete and "archive" -- the `workflows.argoproj.io/workflow-archiving-status` label is "archived"
- Delete it from the cluster (manually) (curl still works), note that the label in the Argo UI and in the database will become "persisted"
- **Retry** (_not resubmit_) the workflow -- curl works and fetches the newer workflow (check uid)
- Wait for it to archive (curl _still_ works)
- Delete it from the cluster manually
  - `curl -k https://localhost:2746/api/v1/workflows/argo/<workflow-name>` **no longer works**, and gives you the error message `{"code":5,"message":"workflows.argoproj.io \"workflow-name\" not found"}`
  - Note that the archived-workflows endpoint still works (curl -k https://localhost:2746/api/v1/archived-workflows/<workflow-uid>)

We would expect the non-archived API to still work (and fetch the newer workflow) as the UI has unified the live/archived workflows.

Server logs for the non-working curl (it can see the archived workflows):

```
argo-server-7d6d865b46-knk48 time="2025-11-10T10:45:27.845Z" level=error msg="failed to get live workflow: workflows.argoproj.io \"test-workflow\" not found; failed to get archived workflow: found 2 archived workflows with namespace
argo-server-7d6d865b46-knk48 time="2025-11-10T10:45:27.845Z" level=info msg="finished unary call with code NotFound" error="rpc error: code = NotFound desc = workflows.argoproj.io \"test-workflow\" not found" grpc.code=NotFound grpc
argo-server-7d6d865b46-knk48 time="2025-11-10T10:45:27.847Z" level=info duration=25.737541ms method=GET path=/api/v1/workflows/argo/test-workflow size=72 status=404
```

### Version(s)

v3.7.3

```YAML
N/A
```

```text
N/A
```

```text
N/A
```

elliotgunton:
Fixes #15015 

### Modifications

* Remove the error return from the workflowArchive `GetWorkflow` function, just log a debug message instead
* Add `OrderBy(-startedat)` to the database query so we can retrieve the latest workflow (via the subsequent `One` function)

### Verification

Tested with `make start PROFILE=postgres UI=true` and then `curl -k http://localhost:2746/api/v1/workflows/argo/<workflow-name>` at each stage described in the issue. If there are multiple persisted workflows with the same name (but different uids) then the `GetWorkflow` endpoint now returns the most recently finished workflow, matching non-archived behaviour.

### DIFF
--- a/persist/sqldb/workflow_archive.go
+++ b/persist/sqldb/workflow_archive.go
@@ -509,6 +509,7 @@ func phaseEqual(phase string) db.Cond {
 }
 
 func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace string, name string) (*wfv1.Workflow, error) {
+	logger := logging.RequireLoggerFromContext(ctx)
 	var err error
 	archivedWf := &archivedWorkflowRecord{}
 	if uid != "" {
@@ -533,7 +534,11 @@ func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace
 			}
 			num := int64(total.Total)
 			if num > 1 {
-				return nil, fmt.Errorf("found %d archived workflows with namespace/name: %s/%s", num, namespace, name)
+				logger.WithFields(logging.Fields{
+					"namespace": namespace,
+					"name":      name,
+					"num":       num,
+				}).Debug(ctx, "returning latest of archived workflows")
 			}
 			err = r.session.SQL().
 				Select("workflow").
@@ -541,6 +546,7 @@ func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace
 				Where(r.clusterManagedNamespaceAndInstanceID()).
 				And(namespaceEqual(namespace)).
 				And(nameEqual(name)).
+				OrderBy("-startedat").
 				One(archivedWf)
 		} else {
 			return nil, sutils.ToStatusError(fmt.Errorf("both name and namespace are required if uid is not specified"), codes.InvalidArgument)


persist/sqldb/workflow_archive.go
```diff
@@ -533,14 +534,15 @@ func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace
...
-				return nil, fmt.Errorf("found %d archived workflows with namespace/name: %s/%s", num, namespace, name)
+				logger.Debug(ctx, fmt.Sprintf("returning latest of %d found archived workflows with namespace/name: %s/%s", num, namespace, name))
 			}
 			err = r.session.SQL().
 				Select("workflow").
 				From(archiveTableName).
 				Where(r.clusterManagedNamespaceAndInstanceID()).
 				And(namespaceEqual(namespace)).
 				And(nameEqual(name)).
+				OrderBy("-finishedat").
```
Joibel: I was expecting this to be "startedat", that's how I think of most recent? Is there precedent for chosing based on finishedat?

persist/sqldb/workflow_archive.go
```diff
@@ -533,14 +534,15 @@ func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace
 			}
 			num := int64(total.Total)
 			if num > 1 {
-				return nil, fmt.Errorf("found %d archived workflows with namespace/name: %s/%s", num, namespace, name)
+				logger.Debug(ctx, fmt.Sprintf("returning latest of %d found archived workflows with namespace/name: %s/%s", num, namespace, name))
```
Joibel: Use WithFields() for logging so that the fields are machine parseable (structured logging), don't printf please.

persist/sqldb/workflow_archive.go
elliotgunton: Nope, I think a brain fart happened. Will use `startedat`

elliotgunton:
> What happens if you have an archived and a live (in cluster) workflow?

It will _not_ fall into the archived GetWorkflow path here https://github.com/argoproj/argo-workflows/blob/436a8d56fb1307b7c894a42331993dc970af235a/server/workflow/workflow_server.go#L750-L752

persist/sqldb/workflow_archive.go
Joibel: /demand :brain_fart: emoji

PR #15059, opened 2025-11-19
fix: set current platform as default platform when retrieving image entrypoint. Fixes #15058

### LINKED ISSUE #15058: Default platform linux/amd64 used when retrieving image entrypoint from container registry
### Pre-requisites

### Issue Description

**Summary**
When a workflow pod spec does not have a commandset, Argo Workflows queries the container registry to get the image's entrypoint. During this lookup, the code does not specify the platform (OS/Architecture), causing the underlying go-containerregistry library to default to linux/amd64. This fails when the image does not support the linux/amd64 platform (e.g., ARM-only images), resulting in a "MANIFEST_UNKNOWN" or "manifest unknown" error.

**Steps to Reproduce**
Run an Argo Workflows server on an ARM64 (linux/arm64) node.
Submit a Workflow where the pod spec:
- Uses an image that is only available for linux/arm64​ (e.g., a private registry, single-arch ARM image).
- Does not​ have a command field defined.

The workflow pod will fail to create. The logs will show an error similar to:
"MANIFEST_UNKNOWN: manifest unknown; map[architecture:amd64 os:linux]"

**Expected Behavior**
Argo Workflows should retrieve the image configuration for the platform architecture of the node it is scheduled on (linux/arm64 in this case), not the default linux/amd64.

**Root Cause**
The issue is in the Lookup method of the containerRegistryIndex(e.g., workflow/controller/entrypoint/container_registery_index.go). When calling remote image or similar methods from the go-containerregistry library, no platform option is provided. This causes the library to use its default hard-coded value of linux/amd64.

### Version(s)

v3.5.0 and later

```YAML
kind: Workflow
metadata:
  generateName: simple-no-command-
spec:
  entrypoint: main
  templates:
    - name: main
      container:
        image: alpine:latest-arm64-only  # assume only arm64 image in registry
```

```text
step group deemed errored due to child xxxxxxxx error: failed to look-up entrypoint/cmd for image "some-arm64-only-image", you must either explicitly specify the command, or list the image's command in the index: https://argoproj.github.io/argo-workflows/workflow-executors/#emissary-emissary: no child with platform linux/amd64 in index some-arm64-only-image
```

```text
# no container created
```

guamian-delicious:
…trypoint from container registry

Fixes #15058 Default platform linux/amd64 used when retrieving image entrypoint from container registry

### Motivation

to fix an issue as above

### Modifications

set default platform as current node platform when retrieving image entrypoint config from registry

### Verification

i build the new binary and run it in my own dev environment

### Documentation

**Summary**
This PR fixes a platform compatibility issue where Argo Workflows would fail to retrieve image configurations for non-linux/amd64 architectures when no commandis specified in the workflow template.

**Changes**
1. Modified image configuration lookup to explicitly specify the node's platform architecture
2. Added platform detection using runtime.GOOS and runtime.GOARCH
3. Ensured compatibility with multi-arch images and single-arch images on matching platforms

**Impact**
Before this fix:
Workflows would fail with "MANIFEST_UNKNOWN" error when:
- Using single-architecture images (ARM, etc.)
- Running on non-AMD64 nodes
- No commandspecified in template

After this fix:
- Workflows automatically use the correct platform architecture for image config lookup
- Supports ARM64, AMD64, and other architectures transparently
- Maintains backward compatibility

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -320,6 +320,7 @@ jobs:
         with:
           name: ui-dist
           path: ui/dist/
+          if-no-files-found: error
 
   build-clis:
     name: Build CLI ${{ matrix.goos }}-${{ matrix.goarch }}
@@ -375,7 +376,8 @@ jobs:
         uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
         with:
           name: argo-${{ matrix.goos }}-${{ matrix.goarch }}
-          path: dist/argo-${{ matrix.goos }}-${{ matrix.goarch }}.gz
+          path: dist/argo-${{ matrix.goos }}-${{ matrix.goarch }}${{ matrix.goos == 'windows' && '.exe' || '' }}.gz
+          if-no-files-found: error
 
   publish-release:
     permissions:
@@ -389,17 +391,9 @@ jobs:
       COSIGN_PASSWORD: ${{secrets.COSIGN_PASSWORD}}
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4.0.1
-        with:
-          node-version: "20" # change in all GH Workflows
       - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version: "1.24"
-      - name: Restore node packages cache
-        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
-        with:
-          path: ui/node_modules
-          key: ${{ runner.os }}-node-dep-v1-${{ hashFiles('**/yarn.lock') }}
       - name: Install cosign
         uses: sigstore/cosign-installer@e1523de7571e31dbe865fd2e80c5c7c23ae71eb4 # v3.4.0
         with:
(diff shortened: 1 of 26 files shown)


workflow/controller/entrypoint/container_registry_index.go
```diff
@@ -2,6 +2,9 @@ package entrypoint
 
 import (
 	"context"
+	"runtime"
+	
+	v12 "github.com/google/go-containerregistry/pkg/v1"
 
 	"github.com/google/go-containerregistry/pkg/authn/k8schain"
 	"github.com/google/go-containerregistry/pkg/name"
```
jswxstw, line 8: Lint checks failed.
```suggestion
	"github.com/google/go-containerregistry/pkg/name"
	pkgv1 "github.com/google/go-containerregistry/pkg/v1"
```

jswxstw (changes requested):
"v12" -> "pkgv1" seems better to me.

workflow/controller/entrypoint/container_registry_index.go
```diff
@@ -27,7 +29,11 @@ func (i *containerRegistryIndex) Lookup(ctx context.Context, image string, optio
 	if err != nil {
 		return nil, err
 	}
-	img, err := remote.Image(ref, remote.WithAuthFromKeychain(kc))
+	var defaultPlatform = pkgv1.Platform{
+		Architecture: runtime.GOARCH,
```
MasonM, line 33: Note to other reviewers: I wasn't sure if there was a one-to-one correspondence between `GOARCH` and the OCI `architecture` field, so I checked [the spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md), and it says:
> Image indexes SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [GOARCH](https://golang.org/doc/install/source#environment).

MasonM (approved):

PR #15014, opened 2025-11-10
chore: flip filtering label to mark examples with tests #15002

gianarb:
Reference: #15002

### Motivation

We want to increase the test coverage for our examples since they are important
to guide users and for documentations we want them to run correctly.

### Modifications

Thi PR flips how we mark example with validation tests. Previously we were
marking only the examples with a test, it makes this process easy to kind of
chill out and avoid writing a validation test. We decided to run example tests
by default for all examples that do not explicitly exclude. This should help us
to writing those tests.

### Verification

As part of exeuction of the tests I wrote the path of the examples that ran as
test in a file prev and after the change to validate that the same number of
examples got executed

### Documentation

I could not find documentation mentioning the old label used so I didn't add
it. If you can point me to where I should add it I am happy to write something
about it.

Issue #15002, opened 2025-10-31
testing: improve testing of examples


Joibel:
# Summary

We have a number of examples which are not tested, and would provide more test coverage if we tested them.
It would be good if we also knew we had not regressed and that all examples worked.

## Use Cases

The code to run examples is in [`test/e2e/examples_test.go`](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/examples_test.go)

Multiple suggested improvements to the test framework for examples. These can be tackled in separate PRs.
* Get the tests to run in parallel. Currently they run under `testify/suite`, which [breaks](https://github.com/stretchr/testify/issues/187) t.Parallel. So divorce it from suite as the setup for this is trivial, and then run each test in a [subtest](https://pkg.go.dev/testing#T.Run)?
* Currently we filter the tests based on seeing `workflows.argoproj.io/test` in the example. In the interests of more and better coverage I suggest flipping this to `workflows.argoproj.io/no-test`, probably with two options:
  * `workflows.argoproj.io/no-test-broken` for genuinely broken examples which can be fixed
  * `workflows.argoproj.io/no-test-environment` for examples which need to run in a different environment, e.g. need access to artifactory.
* Run `argo lint` on every example. Failing examples could be flagged with `workflows.argoproj.io/no-lint`, but really we should fix these!

gianarb:
Morning @Joibel I would like to help out with this one if you don't mind. it will help me to get argo-workflow running locally and it sounds like a good place to start.

Since there are a few things into this issue let me double-check if I got what you are saying at least for a subset of it.
I see there are examples with the label `workflows.argoproj.io/test` those are the one having a test. We want to remove such label and adding `workflows.argoproj.io/no-test` to all the tests that don't have the `test` label and change the executor to exclude only the example with `no-test`, `no-test-environment` or `no-test-broken`. does it sound correct? It makes sense to me so by default CI will fail if there is not a test for a new example, and this will help contributors to write one.

Speaking about executor I see we are hitting a limitation with `testify/suite` but looking at your message it looks like there is something left in your keyboard

> So divorce it from suite as the setup for this is trivial, and then run each test in a

Probably in a [subtest](https://pkg.go.dev/testing#T.Run)?

If it makes sense I will start from those two points

Joibel:
> Morning [@Joibel](https://github.com/Joibel) I would like to help out with this one if you don't mind. it will help me to get argo-workflow running locally and it sounds like a good place to start.
> 
[...]

PR #15063, opened 2025-11-20
chore: move example suite from testify to go subtest #15002

gianarb:
Reference #15002 on top of https://github.com/argoproj/argo-workflows/pull/15014

### Motivation

We want to move example suite from testify suite to go subtest because there is a long standing issue open for testify suite that can't run in parallel.

### Modifications

Remove reference of testify suite from example_test.go and make the `Given` fixture easy to instantiate on its own.

### Verification

Example test suite should pass like before and tests should run in parallel


examples/cluster-workflow-template/workflow-template-ref.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: cluster-workflow-template-hello-world-
+  labels:
+    workflows.argoproj.io/no-test-environment: "true"
```
isubasinghe: If you set this to false, the current code will not correctly handle it

test/e2e/examples_test.go
```diff
@@ -5,45 +5,85 @@ package e2e
...
 		if err != nil {
-			s.T().Fatalf("Error parsing %s: %v", path, err)
+			t.Fatalf("Error parsing %s: %v", path, err)
 		}
 		for _, wf := range wfs {
-			if _, ok := wf.GetLabels()["workflows.argoproj.io/test"]; ok {
-				s.T().Logf("Found example workflow at %s with test label\n", path)
-				s.Given().
+			_, noTestBrokenLabelExists := wf.GetLabels()["workflows.argoproj.io/no-test-broken"]
+			_, noTestBrokenEnvironmentLabelExists := wf.GetLabels()["workflows.argoproj.io/no-test-environment"]
```
isubasinghe: I think we should just use the previous test label, or in the very least just invert the logic by using "no-test". I don't see a point in introducing two new labels to avoid testing a workflow.

Joibel: This was my request in #15002 and also implemented #15014.

The ones which are marked broken should be fixable by someone  - which could be a next step. Examples which won't work are not good - if they won't work because they require artifactory or to run in azure then they get marked as 'no-test-environment' meaning they should work if we had that as part of their test environment. Better names welcome.

examples/timeouts-step.yaml
```diff
@@ -6,6 +6,17 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: timeouts-step-
+  labels:
+    workflows.argoproj.io/no-test: "broken"
```
Joibel: "expected failure" - it's designed to timeout after 10 seconds, yet tries to sleep for 1d.

examples/http-success-condition.yaml
```diff
@@ -13,7 +13,7 @@ spec:
         - - name: http-status-is-201-succeeds
             template: http-status-is-201
             arguments:
-              parameters: [{name: url, value: "http://httpstat.us/201"}]  # Returns status code 201
+              parameters: [{name: url, value: "http://httpbin:9100/status/201"}]  # Returns status code 201
```
Joibel: This breaks it as an example. It would only work in our test environment, so not happy to make this change.

test/e2e/examples_test.go
```diff
@@ -3,47 +3,87 @@
...
+					versioned.NewForConfigOrDie(restConfig).ArgoprojV1alpha1().ClusterWorkflowTemplates(),
+					versioned.NewForConfigOrDie(restConfig).ArgoprojV1alpha1().CronWorkflows(fixtures.Namespace),
+					hydrator.New(persistence.OffloadNodeStatusRepo),
+					kubeClient,
+					string(sec.Data["token"]),
+					restConfig,
+					config,
+				)
+
+				given.KubectlApply("../../examples/configmaps/simple-parameters-configmap.yaml", fixtures.NoError)
```
Joibel: This doesn't need applying every loop around.

I was hoping we could look through all the files twice:
* First time apply all the ConfigMaps, WorkflowTemplates and ClusterWorkflowTemplates. (Then a few more files will be testable from the `workflow-template` and `cluster-workflow-template` directories
* Second time, do this loop.

test/e2e/examples_test.go
```diff
@@ -3,47 +3,87 @@
...
+				given.
```
Joibel: nit: just chain all the givens rather than declaring it as a variable.

test/e2e/examples_test.go
```diff
@@ -3,47 +3,87 @@
...
-			if _, ok := wf.GetLabels()["workflows.argoproj.io/test"]; ok {
-				s.T().Logf("Found example workflow at %s with test label\n", path)
-				s.Given().
+			t.Run(path, func(t *testing.T) {
+				t.Parallel()
+				noTestKeyword, noTextLabelExists := wf.GetLabels()["workflows.argoproj.io/no-test"]
+				if noTextLabelExists {
+					t.Skip(fmt.Sprintf("Impossible to run this example: %s", noTestKeyword))
+				}
+				given := fixtures.NewGiven(
```
Joibel: My preference here is that we have a new package that looks a lot like `fixtures/e2e_suite.go` but doesn't inherit from `suite.Suite`. Let's refactor this later though.

examples/cluster-workflow-template/workflow-template-ref.yaml
gianarb: We fixed that by reverting to a single label called `no-test` and using the value of such label to express why this example can't be tested. In this way there is no really a point to support "true" because "no-test" is there only if you want to no-test it

examples/http-success-condition.yaml
gianarb: Sounds good to me, I can use https://httpbin.org/status/201 so it will go to the internet but when I fixed this example such API was down. It looks up for now, but I am not sure how reliable it is.

test/e2e/examples_test.go
gianarb: Ok, do you think such struct that won't inherit from suite.Suite should turn to be a dependecy of the e2e suite or not?

test/e2e/examples_test.go
gianarb: I created the first loop that applies all the resources that are not workflows from the example directory and I enabled a couple more tests than now are working

test/e2e/examples_test.go
Joibel: Don't do this now, do it in a separate PR.

Ideally we DRY, and common up some code between an e2e_suite that uses Testify, and one that doesn't. I've not tried to do this, so can't really work out what that would look like exactly. If the two can look the same for tests that would be awesome, but the main aim is to pull all of the setup code out of here and make it somewhere that can be reused. The cron tests could drop testify quite easily and be worth doing for example

test/e2e/examples_test.go
```diff
@@ -3,47 +3,172 @@
...
+	}
+	sec, err := clientset.CoreV1().Secrets(fixtures.Namespace).Get(ctx, secrets.TokenName("argo-server"), metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	kindToApply := map[string]bool{
+		"ConfigMap":               true,
+		"PersistentVolumeClaim":   true,
+		"ResourceQuota":           true,
```
Joibel: I wouldn't include ResourceQuota here - installing it won't enable any tests to work.

test/e2e/examples_test.go
```diff
@@ -3,47 +3,172 @@
...
+		"WorkflowEventBinding":    true,
```
Joibel: Same for WorkflowEventBinding, don't apply as we're not testing them

test/e2e/examples_test.go
```diff
@@ -3,47 +3,172 @@
...
+		t.Fatal(err)
+	}
+	sec, err := clientset.CoreV1().Secrets(fixtures.Namespace).Get(ctx, secrets.TokenName("argo-server"), metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	kindToApply := map[string]bool{
+		"ConfigMap":               true,
+		"PersistentVolumeClaim":   true,
```
Joibel, line 60: Does installing a PVC actually work, and does it enable a test?

Joibel (changes requested):
Minor nits

test/e2e/examples_test.go
gianarb: removed

test/e2e/examples_test.go
gianarb: removed

test/e2e/examples_test.go
gianarb: I enabled this test that now works because of the persistent volume claim applied: `examples/volumes-existing.yaml`

Joibel (approved):
Yay, thanks

PR #15076, opened 2025-11-25
fix: http template read response.Body after cancel(), sometimes it return a context canceled error (#14853)

AnonyScorpio:
Fixes #15075


Joibel:
Thanks for doing this

PR #15078, opened 2025-11-25
fix: http template read response.Body after cancel(), sometimes it return a context canceled error (cherry-pick #14853 for 3.6) #15074

AnonyScorpio:
Fixes #15074


Joibel:
Thanks again!

PR #15086, opened 2025-11-27
feat(controller): auto-restart pods that failed before starting

### LINKED ISSUE #12572: Pod was rejected: The node had condition: [DiskPressure].

Joibel:
Fixes #12572 

### Motivation

A retryStrategy can restart pods, but sometimes this is not desired or hard to implement everywhere.
Add a new restart strategy which will recreate pods when they fail prior to starting `main`

### Modifications

Adds automatic restart for pods that fail due to infrastructure issues before the main container starts. This handles transient failures like node evictions, disk pressure, or unexpected admission errors without requiring a retryStrategy.
                                                           
When a pod fails before its main container enters Running state, the controller checks if the failure reason indicates an infrastructure issue. If so, the pod is deleted and the node is marked Pending to recreate it. 
                                                           
Restartable failure reasons:
  - Evicted (node pressure eviction)                     
  - NodeShutdown (graceful node shutdown)
  - NodeAffinity (node affinity/selector no longer matches)            
  - UnexpectedAdmissionError                                                                                           
                                                           
You must enable this in workflow-controller-configmap:
                                       
```                    
      failedPodRestart:
        enabled: true                        
        maxRestarts: 3                                                                                  
```
Added a metric to track this firing
  - Node status includes `FailedPodRestarts` counter
  - New `pod_restarts_total` metric with reason, condition, and namespace labels

### Verification

New unit tests which are mostly pretty lightweight.
E2e tests for the metrics and the actual feature, which fakes pod eviction.

### Documentation

Added a new doc page

### DIFF
--- a/hack/manifests/crdgen.sh
+++ b/hack/manifests/crdgen.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eu -o pipefail
 
 cd "$(dirname "$0")/../.." # up to repo root


util/telemetry/metrics_list.go
```diff
@@ -416,5 +435,8 @@ var InstrumentWorkflowtemplateTriggeredTotal = BuiltinInstrument{
 		{
 			name: AttribTemplateCluster,
 		},
+		{
+			name: AttribWorkflowPhase,
+		},
```
eduardodbr, line 440: is this related to this PR?

util/telemetry/metrics_helpers.go
```diff
@@ -60,13 +73,13 @@ func (m *Metrics) AddErrorCount(ctx context.Context, val int64, errorCause strin
...
-func (m *Metrics) ObserveGauge(ctx context.Context, o metric.Observer, val int64, workflowStatus string) {
+func (m *Metrics) ObserveGauge(ctx context.Context, o metric.Observer, val int64, workflowPhase string) {
 	inst := m.GetInstrument(InstrumentGauge.Name())
 	if inst == nil {
 		return
 	}
 	attribs := InstAttribs{
-		{Name: AttribWorkflowStatus, Value: workflowStatus},
+		{Name: AttribWorkflowPhase, Value: workflowPhase},
 	}
```
eduardodbr, line 83: this also doesnt seem related?

util/telemetry/metrics_list.go
Joibel: These came from codegen, for some reason they're not getting picked up properly.

util/telemetry/metrics_helpers.go
Joibel: Same

isubasinghe (approved):

PR #15092, opened 2025-12-01
fix: always `convert` singular to plural

Joibel:
In #14996 the singular version of each of `semaphore`, `mutex` would not get moved over to the plural in the presence of both fields.

This is not really correct as if you have both sets in 3.7 you will get the behaviour of the appended list+singular. So this PR makes convert append instead of only replace.

For `schedule`, if present it would mask `schedules`. I have kept this behaviour.

Added tests for this.

### DIFF
--- a/workflow/convert/legacy_types.go
+++ b/workflow/convert/legacy_types.go
@@ -31,6 +31,7 @@ func (ls *LegacySynchronization) ToCurrent() *wfv1.Synchronization {
 	sync := &wfv1.Synchronization{}
 
 	// Copy semaphores to avoid aliasing, then append singular if present
+	// Unlike schedules, we honor both singular and plurals here
 	sync.Semaphores = make([]*wfv1.SemaphoreRef, len(ls.Semaphores))
 	copy(sync.Semaphores, ls.Semaphores)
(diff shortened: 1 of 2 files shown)


workflow/convert/legacy_types.go
```diff
@@ -111,9 +114,16 @@ type LegacyCronWorkflowSpec struct {
 
 // ToCurrent converts a LegacyCronWorkflowSpec to the current CronWorkflowSpec type
 func (lcs *LegacyCronWorkflowSpec) ToCurrent() wfv1.CronWorkflowSpec {
-	spec := wfv1.CronWorkflowSpec{
+	// Copy schedules to avoid aliasing, then append singular if present
```
MasonM: I think the original behavior was correct, because `CronWorkflowSpec.getSchedules()` in v3.7.4 ignores `Schedules` if `Schedule` is defined:
https://github.com/argoproj/argo-workflows/blob/v3.7.4/pkg/apis/workflow/v1alpha1/cron_workflow_types.go#L192-L199

Joibel: Argh, you're right. Reverting this bit.

PR #15095, opened 2025-12-02
docs: Add Python side by side examples to walkthrough

elliotgunton:
### Motivation

Argo Workflows focuses on using YAML to write Workflows. It is not immediately obvious from the Walkthrough that Python is usable through Hera. Using the tabbed approach for the code examples means users can view the entire walkthrough in their preferred language, which should be even more useful if/when Argo Workflows supports SDKs in other languages (notably Go could be used). The tab is "linked" throughout the docs, so once a reader clicks "Python", all tabbed examples will switch to Python.

### Modifications

* Updated the `mkdocs.yml` configuration to allow features like the tabbed examples
* Added Python examples to the walkthrough, in particular improving the `Scripts and Results` page wording and introducing Hera for Python.
* I mostly converted the YAML from the walkthrough using the Hera CLI to convert it to Python code, then tidied up the boilerplate to be a bit more idiomatic (though most of the examples use `container` which isn't the main strength of Hera).
* The first few examples also use the "annotate" feature of the material mkdocs theme, to tell the user to install the `hera` package and link to the Hera walkthrough for a Python-centric walkthrough.

The finished result looks like:
YAML:
image

Python:
image

Annotation:
image

### DIFF
--- a/docs/walk-through/artifacts.md
+++ b/docs/walk-through/artifacts.md
@@ -11,6 +11,8 @@ Often, the output artifacts of one step may be used as input artifacts to a subs
 The below workflow spec consists of two steps that run in sequence.
 The first step named `generate-artifact` will generate an artifact using the `hello-world-to-file` template that will be consumed by the second step named `print-message-from-file` that then consumes the generated artifact.
 
+/// tab | YAML
+
 ```yaml
 apiVersion: argoproj.io/v1alpha1
 kind: Workflow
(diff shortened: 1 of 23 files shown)


gianarb (approved):
In general, I really like documentations that are not tight to a specific language if they support more than one :+1:

aaronmell:
Once this is merged, I plan to add the same docs for typescript (juno)

PR #15115, opened 2025-12-06
perf: set ResourceVersion=0 in deleteTaskResults to reduce etcd pressure.

shuangkun:
Set ResourceVersion=0 in ListOptions when deleting task-results collection to avoid hitting etcd directly. This reduces API server pressure in high-volume short-task scenarios by using the watch cache instead.

In large-scale scenarios (100k+ workflows) with short tasks (5min per workflow), this optimization reduces request latency by 10x and decreases API server CPU utilization by 40%.

### DIFF
--- a/examples/artifact-disable-archive.yaml
+++ b/examples/artifact-disable-archive.yaml
@@ -7,6 +7,8 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: artifact-disable-archive-
+  labels:
+    workflows.argoproj.io/no-test: "flaky"
 spec:
   entrypoint: artifact-disable-archive
   templates:
--- a/examples/http-success-condition.yaml
+++ b/examples/http-success-condition.yaml
@@ -2,6 +2,8 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: http-template-condition-
+  labels:
+    workflows.argoproj.io/no-test: "flaky"
   annotations:
     workflows.argoproj.io/description: |
       Exemplifies usage of successCondition in HTTP template (available since v3.3)


workflow/controller/operator.go
```diff
@@ -839,7 +839,10 @@ func (woc *wfOperationCtx) deleteTaskResults(ctx context.Context) error {
 		DeleteCollection(
 			ctx,
 			metav1.DeleteOptions{PropagationPolicy: &deletePropagationBackground},
-			metav1.ListOptions{LabelSelector: common.LabelKeyWorkflow + "=" + woc.wf.Name},
+			metav1.ListOptions{
+				LabelSelector:   common.LabelKeyWorkflow + "=" + woc.wf.Name,
+				ResourceVersion: "0",
```
MasonM, line 847: I'm trying to understand the semantics of using `resourceVersion` with deletes, but I can't seem to find the relevant docs. Here's the docs for `get` and `list`, but there's no mention of `delete`: https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list

Do you know?

Joibel: I am also curious about this.

shuangkun:
@Joibel 
The DeleteCollection operation depends on an underlying LIST call. Therefore, when deleting a collection, the client will first perform a LIST. If we don't explicitly set resourceVersion: "0" in the ListOptions, the request will bypass the apiserver cache and hit etcd directly.

This is a real business scenario， each workflow 5 minutes:
image

This shows the comparison of the results before and after the change.
image

Joibel:
Is there a risk that we miss some of the WTFResults by only inspecting the cache. In that case those would get deleted when the workflow gets deleted via ownerReference?

It seems unlikely that we'd consider a workflow succeeded without all of the results, but are their failures?

I'm still probably OK with missing some as long as they don't leak long term.

shuangkun:
> Is there a risk that we miss some of the WTFResults by only inspecting the cache. In that case those would get deleted when the workflow gets deleted via ownerReference?
> 
> It seems unlikely that we'd consider a workflow succeeded without all of the results, but are their failures?
> 
> I'm still probably OK with missing some as long as they don't leak long term.

You're right, it's possible that some resources could be missed, though the probability is very low. However, when the parent Workflow is deleted, the kube-controller-manager will garbage-collect these orphaned resources via their ownerReference. Therefore, compared to the significant performance impact of an uncached LIST operation at scale, this risk is considered acceptable.

workflow/controller/operator.go
MasonM, line 847: Minor: I think it's worth adding a comment to clarify what this is doing

```suggestion
				// DeleteCollection does a "list" operation to get the resources to delete, which by default does a strongly consistent read of the most recent version.
				// This can be slow for Kubernetes versions before 1.34, so we set resourceVersion=0 to relax consistency and tell the k8s API to return any resource version.
				// It's possible for this to miss some resources, but those should be GC'd when the parent workflow is deleted.
				ResourceVersion: "0",
```

MasonM (approved):
Thanks for the explanation @shuangkun.  I'm a bit concerned about the risk of introducing a race condition, but I agree the risk is low.

Also, I noticed in [the docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list) that Kubernetes 1.34 now uses the watch cache for "most recent" reads, so that should theoretically remove the need for this, but it's going to be a very long time before we can drop support for older versions.

PR #15116, opened 2025-12-06
fix: add timeout for database query in workflow estimation to prevent blocking.

shuangkun:
I found that when the database was locked, it blocked the execution of a large number of our workflows.

Add timeout control for GetWorkflowForEstimator database query to prevent workflow execution from being blocked when database is slow or locked.

- Add default 5 second timeout for database queries
- Configurable via WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT environment variable
- Return default estimator on timeout/error to ensure workflow continues
- Add warning logs for timeout and error cases
- Add documentation for new environment variable

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -58,7 +58,7 @@ This document outlines environment variables that can be used to customize behav
 | `SEMAPHORE_NOTIFY_DELAY`                 | `time.Duration`     | `1s`                                                                                        | Tuning Delay when notifying semaphore waiters about availability in the semaphore                                                                                                                                                                                        |
 | `WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS` | `bool` | `true` | Whether to watch the Controller's ConfigMap and semaphore ConfigMaps for run-time changes. When disabled, the Controller will only read these ConfigMaps once and will have to be manually restarted to pick up new changes. |
 | `SKIP_WORKFLOW_DURATION_ESTIMATION` | `bool` | `false` | Whether to lookup resource usage from prior workflows to estimate usage for new workflows. |
-| `WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT` | `time.Duration` | `5s` | Timeout for database queries when estimating workflow duration. Prevents workflow execution from being blocked when database is slow or locked. |
+| `WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT_SECONDS` | `int` | `5` | Timeout in seconds for database queries when estimating workflow duration. Prevents workflow execution from being blocked when database is slow or locked. |
 
 CLI parameters of the Controller can be specified as environment variables with the `ARGO_` prefix.
 For example:
--- a/persist/sqldb/workflow_archive.go
+++ b/persist/sqldb/workflow_archive.go
@@ -28,9 +28,9 @@ const (
 	archiveTableName        = "argo_archived_workflows"
 	archiveLabelsTableName  = archiveTableName + "_labels"
 	postgresNullReplacement = "ARGO_POSTGRES_NULL_REPLACEMENT"
-	// Default timeout for database queries in GetWorkflowForEstimator to prevent blocking workflow execution
-	// Can be overridden by WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT environment variable
-	defaultEstimationDBQueryTimeout = 5 * time.Second
+	// Default timeout in seconds for database queries in GetWorkflowForEstimator to prevent blocking workflow execution
+	// Can be overridden by WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT_SECONDS environment variable
+	defaultEstimationDBQueryTimeoutSeconds = 5
 )
 
 type archivedWorkflowMetadata struct {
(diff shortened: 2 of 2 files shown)


workflow/controller/estimation/estimator_factory.go
```diff
@@ -84,9 +89,37 @@ func (f *estimatorFactory) NewEstimator(ctx context.Context, wf *wfv1.Workflow)
...
+			var queryCtx context.Context //nolint:contextcheck // queryCtx is assigned from context.WithTimeout in the if-else below
+			var cancel context.CancelFunc
+			if ctx != nil {
+				queryCtx, cancel = context.WithTimeout(ctx, queryTimeout)
+			} else {
+				queryCtx, cancel = context.WithTimeout(context.Background(), queryTimeout)
+			}
+			defer cancel()
+
+			baselineWF, err := f.wfArchive.GetWorkflowForEstimator(queryCtx, wf.Namespace, requirements)
```
MasonM: It looks like this function isn't using the `ctx` argument at all: https://github.com/argoproj/argo-workflows/blob/1d00f495f5c1574ad9a76ddda3dd2c00f2ab7fb9/persist/sqldb/workflow_archive.go#L590-L628

so I don't think this is going to work. I think you'll need to change that to something like:
```go
	selector := r.session.
		WithContext(ctx).
		SQL().
```

IMO it's also worth moving all this logic to the `persist/sqldb/workflow_archive.go` file, since that's meant to contain the DB logic.

shuangkun: Thanks, I hadn't noticed this problem before.

shuangkun: Moved

persist/sqldb/workflow_archive.go
```diff
@@ -588,7 +592,29 @@ func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace
...
+			}
+			queryCtx, cancel = context.WithTimeout(ctx, queryTimeout)
+			defer cancel()
+		}
+	} else {
+		// If ctx is nil, create a new context with timeout
+		queryTimeout := defaultEstimationDBQueryTimeout
+		queryCtx, cancel = context.WithTimeout(context.Background(), queryTimeout)
+		defer cancel()
+	}
```
MasonM: I don't think the `if _, hasDeadline := ctx.Deadline()` check is necessary because there's already logic that's used when `context.WithTimeout()` is called that does basically the same thing:  
https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/context/context.go;l=636

Also, I don't understand why the `WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT` environment variable is only read if `ctx` is non-nil. If `ctx` is nil, why wouldn't we want to read `WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT`?

Here's a simplified version:
```suggestion
	queryTimeout := env.LookupEnvDurationOr(ctx, "WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT", defaultEstimationDBQueryTimeout)
	if ctx == nil {
		ctx = context.Background()
	}
	queryCtx, cancel := context.WithTimeout(ctx, queryTimeout)
	defer cancel()
```

docs/environment-variables.md
```diff
@@ -58,6 +58,7 @@ This document outlines environment variables that can be used to customize behav
 | `SEMAPHORE_NOTIFY_DELAY`                 | `time.Duration`     | `1s`                                                                                        | Tuning Delay when notifying semaphore waiters about availability in the semaphore                                                                                                                                                                                        |
 | `WATCH_CONTROLLER_SEMAPHORE_CONFIGMAPS` | `bool` | `true` | Whether to watch the Controller's ConfigMap and semaphore ConfigMaps for run-time changes. When disabled, the Controller will only read these ConfigMaps once and will have to be manually restarted to pick up new changes. |
 | `SKIP_WORKFLOW_DURATION_ESTIMATION` | `bool` | `false` | Whether to lookup resource usage from prior workflows to estimate usage for new workflows. |
+| `WORKFLOW_ESTIMATION_DB_QUERY_TIMEOUT` | `time.Duration` | `5s` | Timeout for database queries when estimating workflow duration. Prevents workflow execution from being blocked when database is slow or locked. |
```
Joibel: Don't use `time.Duration` as go's duration syntax isn't a standard for other languages. We shouldn't use it for new things, instead specify TIMEOUT_SECONDS and take a number please.

persist/sqldb/workflow_archive.go
```diff
@@ -587,8 +591,22 @@ func (r *workflowArchive) GetWorkflow(ctx context.Context, uid string, namespace
 	return wf, nil
 }
 
+//nolint:contextcheck // ctx may be nil, in which case we create a new context with logger
 func (r *workflowArchive) GetWorkflowForEstimator(ctx context.Context, namespace string, requirements []labels.Requirement) (*wfv1.Workflow, error) {
-	selector := r.session.SQL().
+	// Add timeout to database query to prevent blocking workflow execution
+	// if database is slow or locked. Check if context already has a deadline,
+	// if not, set a default timeout from environment variable or use default.
+	if ctx == nil {
```
Joibel: I don't believe we should be able to call this without a valid `ctx` from within our codebase. I think you should be able to just take the ctx and create queryCtx using ctx.WithTimeout

I've gone through and removed nearly all context creation elsewhere.

Joibel (changes requested):

docs/environment-variables.md
shuangkun: modified. thanks

PR #15117, opened 2025-12-07
fix: add urllib3 compatibility for HTTPResponse getheader methods

shuangkun:
fix test-python-sdk error

Add compatibility layer for urllib3 HTTPResponse getheader and getheaders methods to support both old and new versions of urllib3.

- Check for getheader/getheaders methods (urllib3 < 2.0)
- Fall back to headers attribute (urllib3 >= 2.0)
- Fixes AttributeError when using urllib3 2.x in Python SDK tests

### DIFF
--- a/sdks/python/Makefile
+++ b/sdks/python/Makefile
@@ -48,6 +48,9 @@ else
 		--generate-alias-as-model
 	# https://vsupalov.com/docker-shared-permissions/#set-the-docker-user-when-running-your-container
 	$(CHOWN) $(WD) || sudo $(CHOWN) $(WD)
+	# Apply urllib3 compatibility fix for rest.py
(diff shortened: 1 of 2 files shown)


MasonM:
Thanks @shuangkun! Since we already decided to remove the Python SDK, it's probably best to move forward with https://github.com/argoproj/argo-workflows/pull/14995 instead of adding hacks to patch it.

Joibel:
I've merged #14995 so we don't need this on `main`. I've backported it to 3.6 and 3.7 though as we're not removing the SDK there. Thank you for the work on it.

shuangkun:
> I've merged #14995 so we don't need this on `main`. I've backported it to 3.6 and 3.7 though as we're not removing the SDK there. Thank you for the work on it.

Thanks！

PR #15146, opened 2025-12-11
chore: rename IsWorkflowDeleted to WorkflowExists for clarity

shuangkun:
The condition was inverted - locks should be released when workflows are deleted, not when they exist. Changed from `!sm.isWFDeleted(wfKey)` to `sm.isWFDeleted(wfKey)`.

This fixes a bug where locks for deleted workflows were not being released, potentially causing resource leaks.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -427,7 +427,7 @@ func (wfc *WorkflowController) createSynchronizationManager(ctx context.Context)
 		wfc.wfQueue.AddAfter(key, semaphoreNotifyDelay)
 	}
 
-	isWFDeleted := func(key string) bool {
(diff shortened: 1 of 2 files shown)


Joibel (changes requested):
[isWFDeleted](https://github.com/argoproj/argo-workflows/blob/18d0e18fbbd253729618cb6de4719b1c026914d0/workflow/controller/controller.go#L430) is really badly named.

In the non error case it returns `exists` which is true if the workflow exists, and false if it doesn't. So it returns the opposite of what you'd expect, so this logic is correct.

I'd appreciate a PR which renames that function, but I think the logic is correct, as the unit test shows.

Joibel (approved):

PR #15150, opened 2025-12-11
fix: if pod fails without container termination, don't mark node succeeded always

Joibel:
### Motivation

`inferFailedReason` doesn't attempt to cope well with a `Failed` Pod where the containers are not marked `Terminated`. We've seen this situation in the field - likely down to pod eviction.

The current behaviour would eventually be to drop out the bottom of the function and return `NodeSucceeded` which feels quite optimistic for a failed pod.

### Modifications

Instead monitor whether both `wait` and `main` exited 0, if they did, then lets call this a Success, but otherwise return Failure telling the user that the pod has failed.

### Verification

Unit tests, but I don't know how to produce this usefully in e2e.

### Documentation

None required.

### DIFF
--- a/workflow/controller/operator.go
+++ b/workflow/controller/operator.go
@@ -1681,6 +1681,12 @@ func (woc *wfOperationCtx) inferFailedReason(ctx context.Context, pod *apiv1.Pod
 		return wfv1.NodeFailed, fmt.Sprintf("can't find failed message for pod %s namespace %s", pod.Name, pod.Namespace)
 	}
 
+	// Track whether critical containers completed successfully (terminated with exit code 0).
+	// We must confirm both to return success—otherwise a pod-level failure (eviction, node death)
+	// could be incorrectly reported as success.
+	mainContainerSucceeded := false
+	waitContainerSucceeded := false
+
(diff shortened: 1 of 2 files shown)


workflow/controller/operator.go
```diff
@@ -1681,6 +1681,12 @@ func (woc *wfOperationCtx) inferFailedReason(ctx context.Context, pod *apiv1.Pod
 		return wfv1.NodeFailed, fmt.Sprintf("can't find failed message for pod %s namespace %s", pod.Name, pod.Namespace)
 	}
 
+	// Track whether critical containers completed successfully (terminated with exit code 0).
+	// We must confirm both to return success—otherwise a pod-level failure (eviction, node death)
+	// could be incorrectly reported as success.
+	mainContainerSucceeded := false
+	waitContainerSucceeded := false
```
terrytangyuan, line 1688: Default should be false anyways?

Joibel: Yeah, I preferred this explicit initialising over `var mainContainerSucceeded bool` way of writing the same thing - it's less typing for one thing.

PR #15160, opened 2025-12-12
fix: use alpine 3.23 version in tests, examples, workflows

forzagreen:
### Motivation

All alpine versions prior to 3.20 are outdated and no longer receive security updates.

See: https://endoflife.date/alpine-linux

### Modifications

Updated all Alpine versions throughout the project.

UPDATE: using alpine version `:3.23`. As of 2025-12-12, this is the [current](https://www.alpinelinux.org/) stable version.

~~Since the commands used inside the Alpine images are basic, I believe it’s safe to rely on `latest`.
Alpine also has a very short release cycle, which means keeping explicit versions updated in the repository would require continuous and unnecessary maintenance.~~

### DIFF
--- a/cmd/argo/commands/common/get_test.go
+++ b/cmd/argo/commands/common/get_test.go
@@ -203,7 +203,7 @@ spec:
       - sh
       - -c
       - sleep 10
-      image: alpine:latest
+      image: alpine:3.23
     name: sleep
 status:
   conditions:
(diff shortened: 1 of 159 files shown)


Joibel:
Please don't use latest. Pin them to something more recent, but not `:latest`

Joibel:
Thanks, this is nice work

Issue #15149, opened 2025-12-11
new-features.md should be lintable and spell checked


Joibel:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

When generating the [new features document](https://argo-workflows.readthedocs.io/en/latest/releasing/#documentation) for release it fails markdown lint and spelling.

It would be good if it didn't fail, and we knew this up front.

I've not dug into what this would involve exactly, but basically the feature.md documents being produced should be ready for incorporation into new-features.md and we shouldn't ever get a failure at this stage of the process.

I think this involves spell checking the individual feature documents, and post processing the generated features-new.md so it complies with the linter after generation.

These could be two separate PRs.

### Version(s)

v4.0-rc1

rohithnarasimha:
Hi @Joibel 
 I’d like to take this issue if it’s available. Could you please assign this issue to me?

ogulcanaydogan:
Hi @Joibel! I'd like to work on this. Looking at the Makefile, `docs-spellcheck` explicitly excludes `new-features.md`:

```
-not -name new-features.md
```

Running `mdspell` on `new-features.md` gives ~53 warnings, but most are GitHub usernames from author links (e.g., `Shuangkun`, `Joibel`, `bradfordwagner`) and URL fragments — not actual misspellings. There's one real typo (`singluar` → `singular` on line 97).

I see two approaches:

1. **Add all author names/tech terms to `.spelling`** — straightforward but adds ~30 entries just for usernames
2. **Pre-process `new-features.md` to strip author links before spell check** — cleaner long-term, since each release adds new contributor names

Which would you prefer? Happy to implement either one. The `docs-lint` (markdownlint) already includes `new-features.md`, so that side is covered.

Green00101:
Hi, I’m interested in this issue and would like to work on it. Please let me know if it’s available.

myzk-a:
I think the substance of this is now handled on `main`, mostly via #16272:

[...]

PR #15167, opened 2025-12-15
chore: from mdspell to codespell #15149

gianarb:
Related #15149

### Motivation

mdspell is a bit too strict let's explore an alternative. I see
codespell has a good amount of starts and it looks maintained.

### Modifications

Replace mdsell with codespell in Makefile

### Verification

CI should be happy


Makefile
```diff
@@ -830,16 +830,27 @@ docs/cli/argo.md: $(CLI_PKG_FILES) go.sum ui/dist/app/index.html hack/docs/cli.g
 docs/go-sdk-guide.md: $(TOOL_SNIPDOC)
 	$(TOOL_SNIPDOC) run
 
-$(TOOL_MDSPELL): Makefile
 # update this in Nix when upgrading it here
```
Joibel, line 833: Should also remove these lines.

Makefile
```diff
@@ -830,16 +830,27 @@ docs/cli/argo.md: $(CLI_PKG_FILES) go.sum ui/dist/app/index.html hack/docs/cli.g
...
 ifneq ($(USE_NIX), true)
 	npm list -g markdown-spellcheck@1.3.1 > /dev/null || npm i -g markdown-spellcheck@1.3.1
 endif
 
+$(TOOL_CODESPELL): Makefile
```
Joibel, line 838: No need for duplicate lines here, just put the dependencies on the same line as the target.

Joibel (changes requested):
Whilst I wouldn't mind using codespell for the code, it's not really suitable for checking documentation, it's not a spell checker in the traditional sense.

I quite liked the idea of using vale, but it might be difficult or too draconian on style until we'd updated all the docs to match kubernetes style. I don't have a correct answer here.

We'd also need to delete the .spelling file if we merged this PR.

gianarb:
> Whilst I wouldn't mind using codespell for the code, it's not really suitable for checking documentation, it's not a spell checker in the traditional sense.

Makes sense, I didn't expect this to be the right approach, but it is a way to start the conversation.

The issue targets `new-features.md` and not the entire doc, so what I can do is to keep new-features.md out from mdspell check and figure out a different one for it? 

Or I can add the words that mdspell complains about into the .spelling files so it will start to work again for it. Then I will add mdspell check for the various features make targets similar to what I did for the #15165 )

I think the last solution is the easiest and does not require new tooling. What do you prefer?

Joibel:
> > Whilst I wouldn't mind using codespell for the code, it's not really suitable for checking documentation, it's not a spell checker in the traditional sense.
> 
> Makes sense, I didn't expect this to be the right approach, but it is a way to start the conversation.
> 
> The issue targets `new-features.md` and not the entire doc, so what I can do is to keep new-features.md out from mdspell check and figure out a different one for it?
> 
> Or I can add the words that mdspell complains about into the .spelling files so it will start to work again for it. Then I will add mdspell check for the various features make targets similar to what I did for the #15165 )
> 
> I think the last solution is the easiest and does not require new tooling. What do you prefer?

Ok, lets spell check the individual features and we can switch to a better spelling tool another time then. I'd prefer fixes to the feature files over lots and lots of new words in .spelling

Joibel:
Thanks for kicking off this conversation @gianarb - it did lead somewhere. We ended up switching spelling providers in #16272: `mdspell` and `.spelling` are gone, replaced by `typos` (common misspellings, broad file set) and `cspell` with technology dictionaries for prose docs. The Makefile this PR patches no longer looks like this, and both typo fixes here ("attch" -> "attach", "singluar" -> "singular") are already on main.

Closing as superseded by #16272.

gianarb:
Thanks for sharing!! Ciao!

Il 11 agosto 2026 18:17:22 CEST, Alan Clucas ***@***.***> ha scritto:

PR #14777, opened 2025-08-21
fix(manifests): add descriptions to CRDs. Fixes #8190

MasonM:
Fixes #8190

### Motivation

This adds field descriptions to the CRDs, allowing the use of `kubectl explain` for nearly every field. Example:
```console
$ kubectl explain workflow.spec.templates.memoize
GROUP:      argoproj.io
KIND:       Workflow
VERSION:    v1alpha1

FIELD: memoize <Object>

DESCRIPTION:
    Memoize allows templates to use outputs generated from already executed
    templates

FIELDS:
  cache <Object> -required-
    Cache sets and configures the kind of cache

  key   <string> -required-
    Key is the key to use as the caching key

  maxAge        <string> -required-
    MaxAge is the maximum age (e.g. "180s", "24h") of an entry that is still
    considered valid. If an entry is older
    than the MaxAge, it will be ignored.
```

### Modifications

Remove the `crd:maxDescLen=0` option when calling `controller-gen` so that field descriptions aren't dropped. I believe the main reason this was added is that descriptions massively increase the CRD sizes, which was a major concern when using client-side apply, but that's mostly a non-issue with server-side apply (https://github.com/argoproj/argo-workflows/pull/14048). The one exception is `manifests/base/crds/full/argoproj.io_workflows.yaml`: even with server-side apply, trying to apply that with every field description causes errors, e.g.:
```console
I0818 05:10:53.286386  111645 round_trippers.go:553] PATCH https://0.0.0.0:45333/apis/apiextensions.k8s.io/v1/customresourcedefinitions/workflows.argoproj.io?fieldManager=kubectl&fieldValidation=Strict&force=false 413 Request Entity Too Large in 3 milliseconds
I0818 05:10:53.286418  111645 round_trippers.go:570] HTTP Statistics: GetConnection 0 ms ServerProcessing 9223372036854 ms Duration 3 ms
I0818 05:10:53.286424  111645 round_trippers.go:577] Response Headers:
I0818 05:10:53.286431  111645 round_trippers.go:580]     Cache-Control: no-cache, private
I0818 05:10:53.286437  111645 round_trippers.go:580]     Content-Type: application/json
I0818 05:10:53.286440  111645 round_trippers.go:580]     X-Kubernetes-Pf-Flowschema-Uid: 087c75fa-394c-4cc0-85ac-603a543c72a6
[...]

PR #15175, opened 2025-12-16
fix: make CRDs smaller (in general) and fix cron schedule regex

Joibel:
### Motivation

Testing 4.0.0-rc1 we noticed
* The CRDs were too big to install on EKS even with SSA, and even with there being no CRD at all. I don't understand how they work on k3d and k3s (I've tried them on my personal k3s cluster, and k3d is tested in CI).
* CronWorkflows with `-` in the schedule were rejected

#14777 is the reason the CRDs have grown a lot since 3.7.

### Modifications

1. Fix the regex for `schedules` to allow `-` for `1-5` values in the cron. I thought this would work with the previous regex, but I was wrong.
2. Shrink the CRDs until they install on a stock AWS cluster. I did experiment with removing lots of whitespace (making them more like JSON with no content changes) but that makes them more difficult to read and doesn't actually seem to help at all.
  * Remove descriptions from the status block always
  * Strip status.storedTemplates and status.storedWorkflowTemplateSpec from workflow objects - these are huge and controller managed
  * Remove descriptions from the spec of the controller managed objects (TaskResults, GCTasks, TaskSets)
  * Remove descriptions from templateDefaults in all 4 CRDs which have them, rather than just workflows.

I've also made the minimization kick in at 1Mb rather than 512k as intended in #14777. 

As a result of these changes in the minimized set:
* GCTasks have lost their descriptions (so are smaller).
* TaskSets are nolonger minimized in the minimized set.

### Verification

Modified an example so it was tested through the fixed regex.
Installed onto our production cluster running 4.0.0-rc1 on EKS.

### Documentation

None

### Note

Yes, the cron change would be nice separated, but then they'd collide on merge and in the interests of getting an installable rc2 out I've combined these changes


Joibel:
@MasonM - as you've had a hand in this I'd appreciate you taking a look before merge.

PR #15185, opened 2025-12-17
fix(ui): fix column alignment in workflow list

gcampax:
### Motivation

The current workflow list has a bug where the started and finished columns are not sized properly. This causes the title row to become misaligned with the actual list, which confused me to no end.

### Modifications

Fixed a stray apostrophe in the CSS classes of the workflow row component.

image


Joibel (approved):

Joibel:
Thank you for your contribution and finding this

PR #15186, opened 2025-12-17
chore: lint markdown new-features.md #15149

gianarb:
Reference #15149

### Motivation

Currently we do not lint the new-features.md file because it fails
straight during a release.

### Modifications

This pR makes the new-features.md compliant with the markdownlint
tool and it adds linting check at the right time and not during a CI
release.

I had to fix a few `.features/pending` files that could not be combined into a
new-features.md without making the linter to go crazy.

And then I added markdownlint in a few more places in the Makefile.

### Verification

`make features-update` should work with no issue and lint the generated file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -863,7 +863,7 @@ endif
 .PHONY: docs-lint
 docs-lint: $(TOOL_MARKDOWNLINT) docs/metrics.md
 	# lint docs
-	$(TOOL_MARKDOWNLINT) docs --fix --ignore docs/fields.md --ignore docs/executor_swagger.md --ignore docs/cli --ignore docs/walk-through/the-structure-of-workflow-specs.md --ignore docs/tested-kubernetes-versions.md --ignore docs/new-features.md --ignore docs/go-sdk-guide.md
+	$(TOOL_MARKDOWNLINT) docs --fix --ignore docs/fields.md --ignore docs/executor_swagger.md --ignore docs/cli --ignore docs/walk-through/the-structure-of-workflow-specs.md --ignore docs/tested-kubernetes-versions.md --ignore docs/go-sdk-guide.md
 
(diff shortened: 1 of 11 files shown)


Makefile
Joibel, line 929: I think people who're developing a new feature would expect to be able to run `make features-validate` and have it tell them if it would fail linting - but currently this won't.

gianarb: I thought about the same, but I see a bit of a race condition here because in order to run the mdcheck we need to actually have a file and not print to stdout. This makes the `features-validate` very similar to `features-update` one. But if we merge those it means that the user running `features-update` in order to run correctly needs to know the `VERSION` to set, and I think it is not what we want.

Another option I thought about is to change the `features-validate` target to temporarily write into a file somewhere, mdcheck and remove the file. In this way it is "okey" if the generated output does not have the correct version but the default one (latest).

Joibel: I don't really mind how we achieve this, the validate target's only job is to validate the markdown - and fixing it so VERSION doesn't matter is fine. Be as brutal as needed. If it effectively runs update, that's also fine - perhaps we can avoid a temp file by piping the output to markdown lint?

Makefile
```diff
@@ -924,9 +924,10 @@ feature-new: hack/featuregen/featuregen
 	$< new --filename $(FEATURE_FILENAME)
 
 .PHONY: features-validate
-features-validate: hack/featuregen/featuregen
+features-validate: hack/featuregen/featuregen $(TOOL_MARKDOWNLINT)
 	# Validate all pending feature documentation files
 	$< validate
+	$< update --dry |  tail +4 | markdownlint -s | $(TOOL_MARKDOWNLINT) -s
```
Joibel: We seem to pipe through markdownlint twice here!

gianarb: my fingers sometime go to fast sorry

PR #15198, opened 2025-12-23
feat!: allow for usage of workflow name in `archive` subcommand. Fixes #15199

### LINKED ISSUE #15199: Ensure `archive` command echos it's non-archived command in the CLI.
# Summary
* `argo archive` commands should accept workflow **name** (not just UID) and auto-detect which lookup method to use.
* Currently users must use UIDs: `argo archive get a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11`
Want: `argo archive get my-workflow` to just work.

## Use Cases
 - Get/delete/resubmit/retry archived workflows by human-readable name instead of copying UUIDs
 - Consistent UX with non-archive commands (`argo get my-wf` already works by name)
---

isubasinghe:
Fixes #15199

### Motivation

This PR allows for access to the workflow archive via name instead of via uid. 
This mirrors how the cli is used for non-archived workflows and is intended to simplify the CLI experience. 

### Modifications

* Added flags to force arguments to be interpreted as either uid or as a workflow name. 
* Added logic to determine if argument is a uid or name.
* Allow for names to be used as an access method to the archive.

### Verification

Existing tests pass and additionally manual testing of the CLI was performed.

### Documentation
Updated the documentation (via codegen).

### DIFF
--- a/cmd/argo/commands/archive/util.go
+++ b/cmd/argo/commands/archive/util.go
@@ -25,7 +25,7 @@ func isUID(s string, forceUID bool, forceName bool) bool {
 	if forceName {
 		return false
 	}
-	return uuidRegex.MatchString(strings.ToLower(s))
+	return uuidRegex.MatchString(s)
 }
 
 func resolveUID(ctx context.Context, serviceClient workflowarchivepkg.ArchivedWorkflowServiceClient, identifier string, namespace string, forceUID bool, forceName bool) (string, error) {


isubasinghe:
@coderabbitai review

.features/pending/archive-cli.md
```diff
@@ -0,0 +1,8 @@
+Description: Allow archive cli commands to use workflow name instead of uid.
+Author: [Isitha Subasinghe](https://github.com/isubasinghe)
+Component: CLI
+Issues: 15199 
+
+This change allows for `archive` related cli commands to use the workflow name
+instead of relying upon the uid. This is explcitly a user experience related improvement.
```
eduardodbr: typo `explicitly`

cmd/argo/commands/archive/delete.go
```diff
@@ -25,14 +41,33 @@ func NewDeleteCommand() *cobra.Command {
...
-				if _, err = serviceClient.DeleteArchivedWorkflow(ctx, &workflowarchivepkg.DeleteArchivedWorkflowRequest{Uid: uid}); err != nil {
+			namespace := client.Namespace(ctx)
+			for _, identifier := range args {
+				var req *workflowarchivepkg.DeleteArchivedWorkflowRequest
+				isUID := isUID(identifier)
+				if forceUID {
+					isUID = true
+				} else if forceName {
+					isUID = false
+				}
```
eduardodbr: this logic is repeated in all commands, maybe we can have a function in `util.go`:
```
func isUID(identifier string, forceName, forceUID bool) (isUID bool) {
    if forceUID {
        return true
    }
    if forceName {
        return false
    }
    return uuidRegex.MatchString(strings.ToLower(identifier))
}
```

server/workflowarchive/archived_workflow_server.go
```diff
@@ -136,7 +144,7 @@ func (w *archivedWorkflowServer) DeleteArchivedWorkflow(ctx context.Context, req
 		// no need for ToStatusError since it is already the same time
 		return nil, status.Error(codes.PermissionDenied, "permission denied")
 	}
-	err = w.wfArchive.DeleteWorkflow(ctx, req.Uid)
+	err = w.wfArchive.DeleteWorkflow(ctx, string(wf.UID))
```
Joibel: Is this deliberately inconsistent. We seem to have added a field to pass name (see workflow-archive.proto change) but then not use it here. The other 3 calls on the server side seems to honor the name+namespace version.

cmd/argo/commands/archive/util.go
```diff
@@ -0,0 +1,72 @@
...
+
+// isUID returns true if the input string matches the UUID format used by Kubernetes UIDs
+func isUID(s string, forceUID bool, forceName bool) bool {
+	if forceUID {
+		return true
+	}
+	if forceName {
+		return false
+	}
+	return uuidRegex.MatchString(strings.ToLower(s))
```
Joibel: isUID will return true on an upper case UID, which we then continue to treat as the UID, without re-lowering it when we hit the DB.

Easiest path seems to be to remove the ToLower here to preserve original behaviour? Not sure though

Joibel:
### Code review

Found 1 issue:

1. `resolveUID` uses `NamePrefix` with `NameFilter: "Exact"`, but `NameFilter` has no effect on `NamePrefix` — the query performs a prefix match instead of exact match

   The request sets `NamePrefix: identifier` alongside `NameFilter: "Exact"`, but `NameFilter` only applies when `options.Name != ""` ([`persist/sqldb/selector.go` L26](https://github.com/argoproj/argo-workflows/blob/a5241e4824ea2519609c872dc0608bb34ee279d6/persist/sqldb/selector.go#L26)). Since `resolveUID` sets `NamePrefix` (not `Name`), `NameFilter` is ignored entirely. `NamePrefix` always generates `LIKE 'identifier%'` via [`namePrefixClause`](https://github.com/argoproj/argo-workflows/blob/a5241e4824ea2519609c872dc0608bb34ee279d6/persist/sqldb/workflow_archive.go#L539-L544). This means searching for `my-workflow` will also match `my-workflow-v2`, `my-workflow-retry-xyz`, etc., causing false "Multiple archived workflows found" errors. The fix is to use the `Name` field (via field selector) instead of `NamePrefix`, which would respect `NameFilter: "Exact"` and generate `name = 'identifier'` SQL.

   https://github.com/argoproj/argo-workflows/blob/a5241e4824ea2519609c872dc0608bb34ee279d6/cmd/argo/commands/archive/util.go#L37-L42

cmd/argo/commands/archive/util.go
isubasinghe: Yeah I agree this is best.

PR #15214, opened 2026-01-02
chore: remove testify from cron e2e test suite #15138

### LINKED ISSUE #15138: Build abstraction for the code copied from E2ESuite during examples_test migration out from Testify
# Summary

Now the example test suite does not rely on `testify.Suite` and it does not use the [E2ESuite struct](https://github.com/argoproj/argo-workflows/blob/9242c6f1c2d72b9535dba8ecbfd1313b54a15547/test/e2e/fixtures/e2e_suite.go#L54) anymore.

In order to achieve such migration as part of PR #15063 we copied a few things (Given initialization, kube client initialization and things like that) out from the E2ESuite straight up into the `examples_test.go` file.

## Use Cases

It works but @Joibel pointed out that will be nice to abstract this part of the code in its own structure in this way we will be able to move other suite out from testify in order to support parallelization. A good candidate for this migration is the Cron test suite.

---

gianarb:
Fixes #15138

### Motivation

This commit explores how to drop testify from another (first was example
suite) e2e suite. In this case the cronjob one with particular focus on
how to abstract the code used for both migrations in a reusable fashion

### Modifications

I created a new structure to pre-configure and dispatch the Given fixture.
I refactored both Example and Cron suite to use it

### Verification

CI/CD is green

### Documentation

N/A

### DIFF
--- a/test/e2e/cron_test.go
+++ b/test/e2e/cron_test.go
@@ -17,8 +17,7 @@ import (
 )
 
 func TestBasic(t *testing.T) {
-	runner := fixtures.NewRunner(t)
-	defer runner.DeleteResources(t)
+	runner := fixtures.NewRunner(t, fixtures.WithTestResourceCleanupEnabled(true))
 	runner.Given().
 		CronWorkflow(`apiVersion: argoproj.io/v1alpha1
 kind: CronWorkflow
@@ -51,6 +50,7 @@ spec:
 			assert.Greater(t, cronWf.Status.LastScheduledTime.Time, time.Now().Add(-1*time.Minute))
 		})
 }
+
 func TestBasicTimezone(t *testing.T) {
 	// This test works by scheduling a CronWorkflow for the next minute, but using the local time of another timezone
 	// then seeing if the Workflow was ran within the next minute. Since this test would be trivial if the selected
@@ -65,8 +65,7 @@ func TestBasicTimezone(t *testing.T) {
 		hour = (hour + 1) % 24
 	}
 	scheduleInTestTimezone := strconv.Itoa(minute) + " " + strconv.Itoa(hour) + " * * *"
-	runner := fixtures.NewRunner(t)
-	defer runner.DeleteResources(t)
+	runner := fixtures.NewRunner(t, fixtures.WithTestResourceCleanupEnabled(true))
 	runner.Given().
 		CronWorkflow(fmt.Sprintf(`
 apiVersion: argoproj.io/v1alpha1
@@ -95,9 +94,9 @@ spec:
 			assert.Greater(t, cronWf.Status.LastScheduledTime.Time, time.Now().Add(-1*time.Minute))
 		})
 }
+
 func TestSuspend(t *testing.T) {
-	runner := fixtures.NewRunner(t)
-	defer runner.DeleteResources(t)
+	runner := fixtures.NewRunner(t, fixtures.WithTestResourceCleanupEnabled(true))
 	runner.Given().
 		CronWorkflow(`apiVersion: argoproj.io/v1alpha1
 kind: CronWorkflow
@@ -131,8 +130,7 @@ spec:
 }
 
 func TestResume(t *testing.T) {
-	runner := fixtures.NewRunner(t)
-	defer runner.DeleteResources(t)
+	runner := fixtures.NewRunner(t, fixtures.WithTestResourceCleanupEnabled(true))
 	runner.Given().
 		CronWorkflow(`apiVersion: argoproj.io/v1alpha1
 kind: CronWorkflow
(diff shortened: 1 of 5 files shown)


gianarb:
I do not like the name `Runner`, it is too generic. it is at this point a placeholder because I did not know how this struct would look like at the end. I still don't know where I am so far with this but happy to get feedback

test/e2e/fixtures/e2e_suite.go
```diff
@@ -179,43 +178,32 @@ func (s *E2ESuite) DeleteResources() {
...
-	}
-	return resourceInterface.Namespace(Namespace)
-}
-
 func (s *E2ESuite) CheckError(err error) {
-	s.T().Helper()
-	if err != nil {
-		s.T().Fatal(err)
-	}
+	CheckError(s.T(), err)
```
Joibel, line 206: Why both this and the previous changes of `s.CheckError`->`CheckError`.

This still should be an `s.T().Helper()`

test/e2e/fixtures/e2e_suite.go
```diff
@@ -71,14 +70,14 @@ type E2ESuite struct {
 func (s *E2ESuite) SetupSuite() {
 	var err error
 	s.RestConfig, err = kubeconfig.DefaultRestConfig()
-	s.CheckError(err)
+	CheckError(s.T(), err)
```
Joibel: Can't we stick with `s.CheckError(err)` here and all the way down?

test/e2e/examples_test.go
```diff
@@ -10,50 +10,45 @@ import (
...
+	ctx := logging.TestContext(t.Context())
 	restConfig, err := kubeconfig.DefaultRestConfig()
 	if err != nil {
 		t.Fatal(err)
 	}
-	kubeClient, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		t.Fatal(err)
-	}
+	//kubeClient, err := kubernetes.NewForConfig(restConfig)
```
Joibel: Remove dead code rather than comment it out please

test/e2e/cron_test.go
```diff
@@ -58,31 +42,33 @@ spec:
...
+		When().
+		CreateCronWorkflow().
+		WaitForWorkflow(fixtures.ToBeSucceeded).
+		Then().
+		ExpectCron(func(t *testing.T, cronWf *wfv1.CronWorkflow) {
+			assert.Equal(t, cronWf.Spec.GetScheduleWithTimezoneString(), cronWf.GetLatestSchedule())
+			assert.Greater(t, cronWf.Status.LastScheduledTime.Time, time.Now().Add(-1*time.Minute))
+		})
+}
+func TestBasicTimezone(t *testing.T) {
```
Joibel, line 54: Blank new line between functions please

test/e2e/cron_test.go
```diff
@@ -10,33 +10,17 @@ import (
...
-	s.DeleteResources()
-}
-
-func (s *CronSuite) TestBasic() {
-	s.Run("TestBasic", func() {
-		s.Given().
-			CronWorkflow(`apiVersion: argoproj.io/v1alpha1
+func TestBasic(t *testing.T) {
+	runner := fixtures.NewRunner(t)
+	defer runner.DeleteResources(t)
```
Joibel: Requiring this everywhere is a bit messy and prone to mistakes (missing the call to DeleteResources). Also, it makes it impossible to t.Parallel() these as they all nuke all resources after they complete.

Can we run this in some wrapper like the former suite?

Joibel (changes requested):

test/e2e/cron_test.go
gianarb: The current cronjob suite does not support parallelization. `defer` does not look like the problem here. Parallelization will not work because the problem is in how tbe DeleteResources function works. It is a copy of the old one, and it does not support parallelization because there is no segmentation by test name. We should do something like labeling all the resources created in a test with their test name and the DeleteResources should delete only the one with such label. Right now the DeleteResources deletes across tests if I understand it correctly, and it works only because parallelization is disabled.

This is based on what we had before where the cleanup happened after each cron subtest. If we want to move it at the end of the all suite we can use the `func TestMain(m *testing.M)` structure but ready the comment related to why we do cleanup after each subtest I think we should keep it like it is so it means that in order to parallelize w eneed to do some work to improve the DeleteResources function

Joibel: Yeah, the current version doesn't support parallelisation, but it did used to have this enabled (but not working, due to testify). The linter `testifylint` complained about them having it enabled but disfunctional, so the code was removed.

Ok, let's go with this for now and we can parallelise in a separate PR if we ever get around to it.

I'd still like us not to have to remember the defer though, so wrapping it.

gianarb: Not sure if you like the approach, but I added options to the Runner and the cleanup runs if enabled as `t.Cleanup` function. It sounds idiomatic and I like the API. Let me know

test/e2e/fixtures/e2e_suite.go
Joibel: I'd still like this to be an s.T().Helper()

PR #15216, opened 2026-01-05
fix: dedupe realtime metrics and handle delete tombstone

jswxstw:
### Motivation

- Fix realtime custom metrics attachment on a workflow: allow multiple realtime metrics per workflow.
- Ensure realtime metrics state is cleaned up when a workflow is deleted, including when the delete event is delivered as a tombstone (cache.DeletedFinalStateUnknown).

### Modifications

- Deduplicate realtime custom-metric trackers per workflow by metricSpec.GetKey() in attachCustomMetricToWorkflow, so the same realtime metric isn’t attached multiple times for the same workflow.
- Update the workflow informer delete handler to handle cache.DeletedFinalStateUnknown by extracting the underlying workflow object and calling DeleteRealtimeMetricsForWfUID, avoiding stale realtime metric values and tracker entries after deletion.

## Bug Fixes

* Improved stability during workflow deletion operations with enhanced type safety and null checks to prevent potential crashes.
* Streamlined realtime metrics tracking to improve efficiency and reduce unnecessary processing.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -1039,8 +1039,14 @@ func (wfc *WorkflowController) addWorkflowInformerHandlers(ctx context.Context) 
 	}
 	_, err = wfc.wfInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		DeleteFunc: func(obj interface{}) {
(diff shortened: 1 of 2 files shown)


Joibel:
@coderabbitai review

PR #15235, opened 2026-01-13
chore(deps): bump golang to 1.25

blkperl:
### Motivation

* go1.24 is EOL in Feb

### Modifications

* bumped golang version to 1.25
* bumped go-swagger version

### Verification

* e2e tests

### Documentation

* no docs required

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -1110,10 +1110,12 @@ func dagSortedNodes(nodes []*dagNode, rootNodeName string) []*dagNode {
 	}
 
 	queue := make([]*dagNode, 0)
+	visited := make(map[string]bool)
 
 	for _, n := range nodes {
 		if n.n.Name == rootNodeName {
 			queue = append(queue, n)
+			visited[n.n.ID] = true
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


Joibel:
Note: Please don't merge until after 4.0.

If we need a big 4.0-rc-4 we might merge this into 4.0, but I'd like to keep `main` clear for an emergency fix of 4.0.

blkperl:
My understanding is go1.24 will be EOL shortly after 1.26 release in February so we will need this in 4.0 eventually and the other branches.

docs/executor_swagger.md
```diff
@@ -113,7 +113,7 @@ ownership management and SELinux relabeling.
 
 
 
-`interface{}`
+[any](#any)
```
blkperl: @MasonM any idea why swagger changed this?

MasonM: Probably a side-effect of https://github.com/go-swagger/go-swagger/pull/3200

Joibel: `any` is the modern way of writing that - so semi-glad that's updated. At some point I'd like to enable the `modernize` linter on our codebase to do the same.

Anyway - the old version got mangled [here](https://github.com/argoproj/argo-workflows/blob/4540e18aac7cbd28f39c138e32bc8f612f7.87883/Makefile#L358) so that needs updating.

blkperl:
@Joibel all set to merge. I think we will need to backport this to the other active branches as well since 1.24 will likely EOL in February. https://endoflife.date/go

Joibel:
Thanks for doing this. We've tended not to backport these changes to older branches unless needed for other reasons - even though the compiler is EOL due to adding risk.

PR #15237, opened 2026-01-13
feat: Upload input artifacts when submitting workflows from the UI

### LINKED ISSUE #12656: Accept CSV and other files as input values for workflow
# Summary

The input value for WorkflowTemplate is a string, but we would also like the CSV file to be an input value.
This may be just our team, but we can work around it by opening the contents of the CSV file and copy/pasting those strings, but mistakes are often made.
Therefore, entering the file format is a safe way to execute the workflow.

## Use Cases

It is often used to execute WorkflowTemplate by non-developers.

---

panicboat:
Fixes https://github.com/argoproj/argo-workflows/issues/12656

> **Depends on #16407** (`feat: add SaveStream to artifact drivers with plugin gRPC streaming`) — must merge first. That PR adds `SaveStream` to the `ArtifactDriver` interface and implements it across all drivers plus a plugin gRPC streaming path. This PR builds the upload endpoint and UI on top of it, and does not touch driver code itself.

### Motivation

Currently, when submitting a workflow from the UI, users cannot provide input artifacts directly.
The only way to use input artifacts is to pre-upload files to the artifact repository and manually specify the key in the WorkflowTemplate.
This creates a poor user experience, especially for ad-hoc workflow executions that require different input files each time.
This PR enables users to upload files directly through the UI when submitting a workflow, making input artifacts truly usable from the web interface.

### Modifications

**Backend:**
- New API endpoint `POST /upload-artifacts/{namespace}/{workflowTemplateName}/{artifactName}` streams a multipart file upload into the workflow's default artifact repository via `ArtifactDriver.SaveStream` (provided by #16407)
- Upload endpoint enforces: `auth.CanI` (`get workflowtemplates` + `create workflows`) → `http.MaxBytesReader` (`ARGO_SERVER_MAX_ARTIFACT_UPLOAD_BYTES`, returns 413) → `ParseMultipartForm` → `defer MultipartForm.RemoveAll()`, in that order
- Server-generated key is `uploads/{namespace}/{uuid}/{filename}` using `github.com/google/uuid` (RFC-4122)
- New `SubmitOpts.Artifacts []string` (format `name=key` only) lets a submitter point a WorkflowTemplate's input artifact at a previously uploaded key
- Submit-time key validation: prefix, `path.Clean` idempotence, four-segment shape, UUID form, `path.Base`; rejects anything else with `InvalidArgument`
- Shared `ResolveArtifactLocation` helper resolves the namespace's default artifact repository for both the upload and submit paths
- Shared `ParseArtifactOverrides` + `ApplyOverridesToTemplateArtifacts` helpers unify the override parsing and application (fail-fast on malformed input, `last-wins` on duplicate names)
- Artifact merge fix in `JoinWorkflowSpec`: a `key`-only override in the Workflow spec now preserves `bucket`/`endpoint`/`credentialSecret` inherited from defaults

**Frontend:**
- `ArtifactsInput` component: XHR upload with progress, drag-and-drop, per-artifact error display
- `SubmitWorkflowPanel` integration: blocks submission while any upload is in progress, then submits with `Artifacts: ["name=key", ...]`
- `upload-artifacts` added to webpack proxy configuration for development

#### Architecture

```mermaid
sequenceDiagram
    participant User
    participant UI as Argo UI
    participant Server as Argo Server
    participant S3 as S3/GCS/Azure
    participant K8s as Kubernetes

    User->>UI: Select file
    UI->>Server: POST /upload-artifacts/{ns}/{tmpl}/{artifact}
    Server->>K8s: Get WorkflowTemplate
    K8s-->>Server: Template config
    Server->>S3: SaveStream (uploads/{ns}/{uuid}/{file})
    S3-->>Server: Success
    Server-->>UI: {name, key, location}

    User->>UI: Click Submit
    UI->>Server: POST /api/v1/workflows/{ns}/submit
    Note right of UI: artifacts: ["name=uploads/{ns}/{uuid}/{file}"]
    Server->>Server: Validate key against uploads/{ns}/{uuid}/{file} shape
    Server->>K8s: Get WorkflowTemplate
    Server->>Server: Override artifact key
    Server->>K8s: Create Workflow
    K8s-->>Server: Workflow
    Server-->>UI: Success
```

<details>
<summary>Data flow</summary>

```mermaid
flowchart TD
    subgraph Frontend
        A[Select File] --> B[XHR Upload]
        B --> C{Success?}
        C -->|Yes| D[Store in uploadedArtifacts]
        C -->|No| E[Show Error]
        D --> F[Submit Button]
        F --> G[Build artifactOverrides]
        G --> H[POST /submit]
    end

    subgraph Backend Upload
        B --> I[UploadInputArtifact]
        I --> J[Get WorkflowTemplate]
        J --> K[Get artifact config]
        K --> L[Generate new key]
        L --> M[SaveStream]
        M --> N[Return response]
    end

    subgraph Backend Submit
        H --> O[SubmitWorkflow]
        O --> P[Get WorkflowTemplate]
        P --> Q[Validate + Override key]
        Q --> R[Create Workflow]
    end
```
</details>

### Verification

1. Create a WorkflowTemplate with an input artifact defined in `arguments.artifacts`
2. Open the WorkflowTemplate in the UI and click "Submit"
3. Upload a file using the new file input field
4. Submit the workflow
5. Verify the workflow runs successfully using the uploaded file

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: s3-creds
  namespace: argo
type: Opaque
stringData:
  accessKey: AKI**********
  secretKey: s9K**********
--
kind: WorkflowTemplate
metadata:
  name: my-template
  namespace: argo
spec:
  entrypoint: argosay
  arguments:
    artifacts:
      - name: my-uploaded-file
        s3:
          endpoint: s3.amazonaws.com
          region: us-east-1
          bucket: my-bucket
          key: input-file/placeholder.zip  # overridden per-submit via the upload endpoint
          accessKeySecret:
            name: s3-creds
            key: accessKey
          secretKeySecret:
            name: s3-creds
            key: secretKey
  templates:
    - name: argosay
      inputs:
        artifacts:
          - name: my-uploaded-file
            path: /tmp/input-file
      container:
        image: alpine:latest
        command: [cat]
        args: ["/tmp/input-file"]
```

https://github.com/user-attachments/assets/096f8621-07e8-4b13-9eaa-335af928a515

### Operational note — cleanup of abandoned uploads

Objects written under `uploads/{namespace}/{uuid}/` become orphaned if the user never submits, re-uploads, or the workflow is deleted. **This PR relies on operator-configured bucket lifecycle rules under the `uploads/` prefix** (7-day TTL suggested). See `.features/pending/input-artifact-upload.md` for S3/GCS/OSS lifecycle rule examples and the corresponding CLI commands.

### Scope

**In scope**: WorkflowTemplate upload via UI, key validation, size limit, RBAC on the upload endpoint, operator docs for bucket lifecycle.

**Out of scope (deferred to follow-up issues)**:

- **ClusterWorkflowTemplate and CLI support** (e.g., `argo submit --from wftmpl/name --input-artifact name=./file.zip`) — can reuse the `/upload-artifacts/` endpoint introduced here
- **Cryptographic ownership of upload keys** (HMAC-signed upload token bound to namespace/subject/wftmpl/artifact/key/expiry/kid) — the current key validation restricts to the server-controlled prefix but does not prove ownership within the same namespace; the UUID's high entropy is the mitigation. Design sketch kept locally for when audit needs arise.
- **GC linkage for submitted uploads** (marking uploaded-input artifacts and hooking them into the existing artifact GC) — this PR meets the "at least a bucket lifecycle/TTL **or** GC linkage" bar via the bucket lifecycle side of the `or`. Design sketch kept locally for a follow-up that touches controller/GC semantics.
- **Active GC of abandoned uploads** (uploaded but never submitted) — requires an upload-session registry; bucket lifecycle handles this operationally.

### Documentation

Added feature documentation in `.features/pending/input-artifact-upload.md`, including the operator guidance for bucket lifecycle rules.
Users can discover this feature when they:
1. View a WorkflowTemplate with input artifacts defined
2. Click "Submit" button
3. See the file upload fields for each defined artifact

## Summary

* **New Features**
  * Added submit UI to select/upload workflow input artifacts (drag/drop, progress, per-artifact success) and send them as `name=key` overrides.
  * Added server-side input artifact uploads with default repository location resolution during submission.
* **Documentation**
  * Added guidance on abandoned upload cleanup and the upload-size limit, including supported storage providers.
* **Bug Fixes**
  * Improved artifact override/merge behavior with precedence-by-name and safer uploaded artifact key validation.
* **Chores**
  * Updated API schema and UI submit options to support artifact overrides.


Joibel:
@coderabbitai review

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,181 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+	if !artifactCopy.HasLocation() {
+		http.Error(w, fmt.Sprintf("Artifact '%s' does not have a storage location configured (s3, gcs, azure, oss). Please configure a storage location in the WorkflowTemplate or set up a default artifact repository.", artifactName), http.StatusBadRequest)
+		return
+	}
+
+	// Generate unique key for the artifact
+	uuid := generateUUID()
+	originalKey, _ := artifactCopy.GetKey()
+	// Replace the key with uploaded file path under uploads/
+	newKey := fmt.Sprintf("uploads/%s/%s/%s", namespace, uuid, header.Filename)
```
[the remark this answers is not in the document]
panicboat: Fixed in 46d97435a. Sanitized with `path.Base()` and added rejection for empty/dot-only filenames. Test added.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,181 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+		a.logger.WithError(err).Error(ctx, "Failed to encode response")
+	}
+}
+
+// generateUUID generates a simple UUID for unique artifact keys
+func generateUUID() string {
+	b := make([]byte, 16)
+	_, _ = rand.Read(b)
+	return hex.EncodeToString(b)
+}
```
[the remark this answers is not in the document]
panicboat: Fixed in 7cceac929. `generateUUID()` now returns `(string, error)` and the caller handles the error.

ui/src/shared/components/artifacts-input/artifacts-input.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    } else {
+                        reject(new Error(`Upload failed: ${xhr.statusText}`));
+                    }
+                };
+
+                xhr.onerror = () => {
+                    reject(new Error('Network error'));
+                };
+
+                xhr.open('POST', `/upload-artifacts/${namespace}/${workflowTemplateName}/${artifactName}`);
```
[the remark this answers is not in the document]
panicboat: Fixed in 63f85400c. All path parameters are now wrapped with `encodeURIComponent()`.

ui/src/shared/components/artifacts-input/artifacts-input.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    return (
+        <div className='artifacts-input'>
+            <input ref={fileInputRef} type='file' onChange={handleFileChange} style={{display: 'none'}} />
+
+            {status === 'idle' && (
+                <div className='artifacts-input__dropzone' onClick={handleFileSelect}>
+                    <i className='fa fa-upload' />
+                    <span>Click to select a file or drag and drop</span>
+                </div>
```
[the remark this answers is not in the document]
panicboat: Fixed in fa81eb65a. Replaced `location: any` with `Record<string, unknown>`.

ui/src/workflows/components/submit-workflow-panel.tsx
```diff
@@ -62,13 +69,20 @@ export function SubmitWorkflowPanel(props: Props) {
...
             const submitted = await services.workflows.submit(props.kind, props.name, props.namespace, {
                 entryPoint: entrypoint === workflowEntrypoint ? null : entrypoint,
                 parameters: [
                     ...workflowParameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p)),
                     ...parameters.filter(p => getValueFromParameter(p) !== undefined).map(p => p.name + '=' + getValueFromParameter(p))
                 ],
-                labels: labels.join(',')
+                labels: labels.join(','),
+                artifacts: artifactOverrides.length > 0 ? artifactOverrides : undefined
             });
```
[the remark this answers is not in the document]
panicboat: Fixed in 19b7841d4. Added `uploadingArtifacts` state to track in-progress uploads. Submit button is disabled and shows "Uploading..." while any upload is in progress.

workflow/artifacts/gcs/gcs.go
```diff
@@ -241,6 +241,31 @@ func (h *ArtifactDriver) Save(ctx context.Context, path string, outputArtifact *
...
+				wc.Close()
+				return !isTransientGCSErr(ctx, err), err
+			}
+			if err = wc.Close(); err != nil {
+				return !isTransientGCSErr(ctx, err), err
+			}
+			return true, nil
+		})
+	return err
+}
```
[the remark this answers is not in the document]
panicboat: Fixed in 2022b9211. GCS SaveStream now buffers the stream to a temp file before entering the retry loop, consistent with the HDFS driver pattern.

workflow/artifacts/oss/oss.go
```diff
@@ -275,6 +275,34 @@ func (ossDriver *ArtifactDriver) Save(ctx context.Context, path string, outputAr
...
+			bucket, err := osscli.Bucket(bucketName)
+			if err != nil {
+				return !isTransientOSSErr(ctx, err), err
+			}
+			err = bucket.PutObject(outputArtifact.OSS.Key, reader)
+			if err != nil {
+				return !isTransientOSSErr(ctx, err), err
+			}
+			return true, nil
+		})
```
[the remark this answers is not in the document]
panicboat: Fixed in 2022b9211. OSS SaveStream now buffers the stream to a temp file before entering the retry loop.

workflow/artifacts/oss/oss.go
```diff
@@ -275,6 +275,34 @@ func (ossDriver *ArtifactDriver) Save(ctx context.Context, path string, outputAr
...
+				return !isTransientOSSErr(ctx, err), err
+			}
+			bucket, err := osscli.Bucket(bucketName)
+			if err != nil {
+				return !isTransientOSSErr(ctx, err), err
+			}
+			err = bucket.PutObject(outputArtifact.OSS.Key, reader)
+			if err != nil {
+				return !isTransientOSSErr(ctx, err), err
+			}
```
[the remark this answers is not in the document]
panicboat: Fixed in 2022b9211. OSS SaveStream now buffers the stream to a temp file before entering the retry loop, so the reader is fresh on each retry.

workflow/artifacts/s3/s3.go
```diff
@@ -267,6 +270,27 @@ func (s3Driver *ArtifactDriver) Save(ctx context.Context, path string, outputArt
...
+	defer cancel()
+	log := logging.RequireLoggerFromContext(ctx)
+	err := waitutil.Backoff(executorretry.ExecutorRetry(ctx),
+		func() (bool, error) {
+			log.WithField("key", outputArtifact.S3.Key).Info(ctx, "S3 SaveStream")
+			s3cli, err := s3Driver.newClient(ctx)
+			if err != nil {
+				return !isTransientS3Err(ctx, err), fmt.Errorf("failed to create new S3 client: %w", err)
+			}
+			err = s3cli.PutStream(outputArtifact.S3.Bucket, outputArtifact.S3.Key, reader, -1)
```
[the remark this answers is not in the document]
panicboat: Fixed in 2022b9211. S3 SaveStream now buffers the stream to a temp file before the retry loop. Also passes actual file size to PutStream instead of -1.

panicboat:
Addressed the nitpick comments from the review:

- **workflow_server.go:834-847 (verbose logging):** Fixed in 1bf2106d8. Removed anonymous function and unnecessary fields. Logging now only fires when `SubmitOptions` is non-nil.
- **workflow_server.go:882-887 (malformed override):** Fixed in 244cbd1b8. Malformed artifact override strings now produce a warning log.
- **artifacts-input.tsx:14-18 (location: any):** Fixed in fa81eb65a. Replaced with `Record<string, unknown>`.
- **artifacts-input.tsx:107-110 (drag-and-drop):** Fixed in fa81eb65a. Added `onDragOver`, `onDragLeave`, `onDrop` handlers to the dropzone element.
- **azure.go:307-323 (error wrapping):** Fixed in 2022b9211. `UploadStream` error is now wrapped with `fmt.Errorf` for consistent reporting.

ui/src/shared/components/artifacts-input/artifacts-input.tsx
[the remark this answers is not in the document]
panicboat: Sorry for the confusion in my previous reply. This is fixed in fa81eb65a — added `onDragOver`, `onDragLeave`, and `onDrop` handlers to the dropzone element, along with an `isDragging` state for visual feedback.

workflow/artifacts/gcs/gcs_test.go
```diff
@@ -47,3 +51,63 @@ func TestIsTransientGCSErr(t *testing.T) {
...
+	files, err := listFileRelPaths(tempDir+string(os.PathSeparator), "")
+	require.NoError(t, err)
+	assert.Empty(t, files)
+}
+
+// TestListFileRelPathsNonExistent tests listing a non-existent directory
+func TestListFileRelPathsNonExistent(t *testing.T) {
+	_, err := listFileRelPaths("/non/existent/path/", "")
+	require.Error(t, err)
+}
```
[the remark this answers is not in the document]
panicboat: This test runs in CI on Linux only. Will revisit if cross-platform support becomes a requirement.

workflow/artifacts/http/http.go
```diff
@@ -140,6 +140,45 @@ func (h *ArtifactDriver) Save(ctx context.Context, path string, outputArtifact *
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = res.Body.Close()
+	}()
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
+		return errors.InternalErrorf("saving stream to %s failed with reason: %s", url, res.Status)
+	}
+	return nil
```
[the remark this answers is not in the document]
panicboat: The HTTP artifact driver is read-only (download). `SaveStream` is not expected to be called for HTTP artifacts in practice, so redirect handling for uploads is not applicable here.

workflow/artifacts/s3/s3_test.go
```diff
@@ -713,6 +748,127 @@ func TestNewClientWithDiff(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			reader := strings.NewReader(tc.content)
+			err := tc.s3client.PutStream(tc.bucket, tc.key, reader, int64(len(tc.content)))
+			if tc.errMsg == "" {
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
+				// For this test, we're testing the mock client directly
+				// The actual error wrapping happens in the ArtifactDriver.SaveStream
+			}
```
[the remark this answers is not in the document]
panicboat: Noted. The existing S3 tests cover the underlying client behavior. Adding SaveStream-specific integration tests is a good follow-up but out of scope for this PR.

workflow/util/merge.go
```diff
@@ -131,6 +131,20 @@ func JoinWorkflowSpec(wfSpec, wftSpec, wfDefaultSpec *wfv1.WorkflowSpec) (*wfv1.
...
+	// Apply the same priority logic for artifacts.
+	// Artifacts from the workflow spec should take precedence over those from the template.
+	wfArtifactsMap := artifactsToMapByName(wfSpec)
+	wftArtifactsMap := artifactsToMapByName(wftSpec)
+	for index, artifact := range targetWf.Spec.Arguments.Artifacts {
+		if art, ok := wfArtifactsMap[artifact.Name]; ok {
+			targetWf.Spec.Arguments.Artifacts[index] = *art.DeepCopy()
+		} else if art, ok := wftArtifactsMap[artifact.Name]; ok {
+			targetWf.Spec.Arguments.Artifacts[index] = *art.DeepCopy()
+		}
```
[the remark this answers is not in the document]
panicboat: The current behavior is intentional. When an artifact override is provided, `DeepCopy()` replaces the entire artifact, which is the correct semantics for overriding uploaded artifact keys. Partial field overlay would risk mixing stale location fields with the new key.

workflow/artifacts/oss/oss.go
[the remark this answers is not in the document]
panicboat: Good catch. Fixed in 803344135 — added `CreateBucketIfNotPresent` and `LifecycleRule` setup to `SaveStream` to match `Save` behavior.

workflow/artifacts/s3/s3_test.go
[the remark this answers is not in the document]
panicboat: Yes, please open an issue for that. Happy to address it in a follow-up PR.

panicboat:
@Joibel @terrytangyuan
Thank you for reviewing this PR.
I’ve addressed the points raised by CodeRabbitAI, so I’d appreciate it if you could take another look when you have a moment.
If you notice anything else that needs attention, please feel free to let me know.

Joibel (commented):
## Code review — automated multi-agent pass

I ran 11 independent review passes over this PR and confidence-scored each finding (0–100). **20 issues** are flagged as inline comments below. The repo's `CLAUDE.md`/`AGENTS.md` contain only GitNexus tooling instructions (no human PR conventions), so these are scored purely on whether they're real, not on style-guide compliance.

**Suggested merge blockers:**
- Missing authorization on the upload endpoint (inline, confidence 85)
- No upload-size limit + multipart temp-file leak (confidence 90 / 92)
- Client-supplied artifact key applied without validation (confidence 85)

**Worth cleaning up before merge:**
- Triplicated + partly-dead `name=key` override logic (confidence 85)
- Scope creep: ~39 unrelated `.spelling` entries, unrelated `controller_test`/azure/hdfs changes, author's sandbox bucket + AWS account id in fixtures

The rest (debug logging, hand-rolled UUID, misleading format docs, missing frontend/cluster-branch tests, mock-only s3 test) are polish. Full detail is in the inline comments.

<sub>🤖 Generated with multi-agent code review. Each finding was independently verified and confidence-scored.</sub>

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,193 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+	if len(requestPath) < 5 {
+		http.Error(w, "Invalid path. Expected: /upload-artifacts/{namespace}/{workflowTemplateName}/{artifactName}", http.StatusBadRequest)
+		return
+	}
+	namespace := requestPath[2]
+	workflowTemplateName := requestPath[3]
+	artifactName := requestPath[4]
+
+	// Authenticate and authorize
+	ctx, err := a.gateKeeping(r, types.NamespaceHolder(namespace))
```
Joibel, line 98: **Authorization missing on upload endpoint (confidence 85).** This handler calls only `gateKeeping` (authentication). Every other sensitive handler in this file pairs it with `validateAccess`/`auth.CanI` (e.g. L353, L549). In Server/SSO-without-RBAC modes the resolved clients are the server's privileged service account, so any authenticated user can read any WorkflowTemplate and write into any namespace's artifact store. Recommend adding an explicit `CanI`/instanceID check here.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,193 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+	}
+
+	a.logger.WithFields(logging.Fields{
+		"namespace":            namespace,
+		"workflowTemplateName": workflowTemplateName,
+		"artifactName":         artifactName,
+	}).Info(ctx, "Upload artifact")
+
+	// Parse multipart form (max 32MB in memory, rest on disk)
+	if parseErr := r.ParseMultipartForm(32 << 20); parseErr != nil {
```
Joibel, line 162: **No total upload-size limit — disk/DoS vector (confidence 90).** `ParseMultipartForm(32<<20)` only bounds *in-memory* buffering; the body is uncapped and spills to disk (then buffered again by the driver's `SaveStream`). There's no `http.MaxBytesReader`. The "32MB" framing is misleading — consider wrapping `r.Body` with `http.MaxBytesReader`.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,193 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+		http.Error(w, "Failed to parse multipart form: "+parseErr.Error(), http.StatusBadRequest)
+		return
```
Joibel, line 169: **Multipart temp files are never cleaned up (confidence 92).** After `ParseMultipartForm`, parts >32MB are written to disk and the caller must `defer r.MultipartForm.RemoveAll()`. This handler never does, across all return paths, so large uploads leak temp files. Add `defer r.MultipartForm.RemoveAll()` right after a successful parse.

server/workflow/workflow_server.go
```diff
@@ -827,11 +831,111 @@ func (s *workflowServer) SubmitWorkflow(ctx context.Context, req *workflowpkg.Wo
...
+								logger.WithFields(logging.Fields{
+									"artifactName": tmplArt.Name,
+									"repoRef":      repoRef,
+								}).Debug(ctx, "Resolved artifact location from default repository for override")
+							}
+						}
+					}
+				}
+
+				if setErr := artCopy.SetKey(newKey); setErr != nil {
```
Joibel: **Client-supplied artifact key applied without validation (confidence 85).** Upload generates a sanitized server key (`uploads/{ns}/{uuid}/{file}`), but here the `name=key` value from the request is passed straight to `SetKey`. There's no check that the key belongs to an upload by this user / lives under `uploads/{ns}/`. A submitter can redirect an input artifact to any object key in the template's bucket and read it inside their container. Consider constraining the key to the server-controlled prefix.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,193 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+	}
+	originalKey, _ := artifactCopy.GetKey()
+	// Sanitize filename to prevent path traversal attacks
+	sanitizedFilename := path.Base(header.Filename)
+	if sanitizedFilename == "." || sanitizedFilename == "/" || sanitizedFilename == "" {
+		http.Error(w, "Invalid filename", http.StatusBadRequest)
+		return
+	}
+	// Replace the key with uploaded file path under uploads/
+	newKey := fmt.Sprintf("uploads/%s/%s/%s", namespace, uuid, sanitizedFilename)
```
Joibel: **Uploaded artifacts are never garbage-collected (confidence 85).** Objects written under `uploads/{ns}/{uuid}/…` are orphaned if the user never submits, re-uploads, or deletes the workflow — nothing ties them to a workflow lifecycle for cleanup. Worth at least a bucket lifecycle/TTL or GC linkage before this sees production use.

workflow/artifacts/http/http.go
```diff
@@ -140,6 +140,45 @@ func (h *ArtifactDriver) Save(ctx context.Context, path string, outputArtifact *
...
+	var err error
+	if outputArtifact.Artifactory != nil && outputArtifact.HTTP == nil {
+		url = outputArtifact.Artifactory.URL
+		req, err = http.NewRequestWithContext(ctx, http.MethodPut, url, reader)
+		if err != nil {
+			return err
+		}
+		req.SetBasicAuth(h.Username, h.Password)
+	} else {
+		url = outputArtifact.HTTP.URL
```
Joibel: **Potential nil-pointer deref (confidence 75).** This `else` branch dereferences `outputArtifact.HTTP.URL` with no guard; `retrieveContent` (L49) guards the same shape with an explicit error. Not reachable under normal driver selection, but the new method drops the safety the rest of the file has.

server/workflow/workflow_server.go
```diff
@@ -827,11 +831,111 @@ func (s *workflowServer) SubmitWorkflow(ctx context.Context, req *workflowpkg.Wo
...
+		}
+
+		// Parse artifact overrides: name=key format
+		overrides := make(map[string]string)
+		for _, artifactStr := range req.SubmitOptions.Artifacts {
+			parts := strings.SplitN(artifactStr, "=", 2)
+			if len(parts) == 2 {
+				overrides[parts[0]] = parts[1]
+			} else {
+				logger.WithField("artifact", artifactStr).Warn(ctx, "Ignoring malformed artifact override (expected name=key format)")
```
Joibel: **Divergent error handling for the same input (confidence 85).** This inline override block only *warns and skips* malformed `name=key` entries, while `overrideArtifacts` in `workflow/util/util.go` (L346) *returns an error* for the identical input. Same user input, two behaviors — and the parsing logic is duplicated. Recommend consolidating into one shared helper with consistent handling.

workflow/util/util.go
```diff
@@ -329,6 +333,46 @@ func overrideParameters(wf *wfv1.Workflow, parameters []string) error {
...
+	for _, artifactStr := range artifacts {
+		parts := strings.SplitN(artifactStr, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("expected artifact of the form: NAME=KEY. Received: %s", artifactStr)
+		}
+		overrides[parts[0]] = parts[1]
+	}
+
+	// Override artifact keys in workflow arguments
+	// Note: For workflowTemplateRef workflows, artifacts are handled in workflow_server.go
```
Joibel, line 408: **`overrideArtifacts` is effectively dead for the UI path (confidence 85).** `NewWorkflowFromWorkflowTemplate` produces an empty `Spec.Arguments.Artifacts`, so for `workflowTemplateRef` workflows this loop iterates nothing — which is why the real logic was duplicated into `workflow_server.go` (as this comment admits). The shared util addition no-ops for its primary use case.

workflow/artifacts/s3/s3.go
```diff
@@ -267,6 +270,52 @@ func (s3Driver *ArtifactDriver) Save(ctx context.Context, path string, outputArt
 	return err
 }
 
+// SaveStream saves an artifact from an io.Reader to S3 compliant storage
+func (s3Driver *ArtifactDriver) SaveStream(ctx context.Context, reader io.Reader, outputArtifact *wfv1.Artifact) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Buffer to a temp file so the reader can be re-read on retry
+	tmpFile, err := os.CreateTemp("", "s3-upload-*")
```
Joibel: **Temp-file buffering boilerplate duplicated across 5 drivers (confidence 85).** This `CreateTemp → io.Copy → close → defer remove` block (same error strings) is copy-pasted into s3/gcs/oss/hdfs/plugin `SaveStream`. A shared helper (e.g. `common.BufferReaderToTempFile`) would remove ~5 copies.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,193 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+	artifactCopy := templateArtifact.DeepCopy()
+
+	// If the artifact doesn't have a full location, try to resolve from default artifact repository
+	if !artifactCopy.HasLocation() {
+		// Try to resolve default artifact repository for the namespace
+		// This handles cases where:
+		// 1. WorkflowTemplate specifies artifactRepositoryRef explicitly
+		// 2. Namespace has artifact-repositories ConfigMap
+		// 3. workflow-controller-configmap has default artifactRepository
+		repoRef, resolveErr := a.artifactRepositories.Resolve(ctx, wfTemplate.Spec.ArtifactRepositoryRef, namespace)
```
Joibel: **Default-repo resolution block duplicated (confidence 85).** This `!HasLocation() → Resolve → Get → ToArtifactLocation → DeepCopy` sequence (including the "We don't use Relocate()" comment) is repeated in `server/workflow/workflow_server.go` (~L899). Extract into one helper called from both.

server/workflow/workflow_server.go
```diff
@@ -827,11 +831,111 @@ func (s *workflowServer) SubmitWorkflow(ctx context.Context, req *workflowpkg.Wo
 
 	s.instanceIDService.Label(wf)
 	creator.LabelCreator(ctx, wf)
+
+	logger := logging.RequireLoggerFromContext(ctx)
+	if req.SubmitOptions != nil {
+		logger.WithFields(logging.Fields{
+			"artifactsCount":         len(req.SubmitOptions.Artifacts),
+			"hasWorkflowTemplateRef": wf.Spec.WorkflowTemplateRef != nil,
+		}).Debug(ctx, "SubmitWorkflow: checking conditions")
```
Joibel: **Leftover Debug scaffolding logging (confidence 75).** `SubmitWorkflow` adds many Debug lines ("checking conditions", "Parsed artifact overrides", per-artifact "Applied artifact key override" with an inline closure built only to log the old key, and a "Final workflow artifacts before creation" block that builds a `[]map[string]string` purely to log). Reads as dev instrumentation on a core RPC handler — recommend trimming.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,193 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(response); err != nil {
+		a.logger.WithError(err).Error(ctx, "Failed to encode response")
+	}
+}
+
+// generateUUID generates a simple UUID for unique artifact keys
+func generateUUID() (string, error) {
```
Joibel: **Reinvents an available UUID library, and isn't a UUID (confidence 75).** This produces a 32-char hex string (no RFC-4122 format) despite `github.com/google/uuid` already being in `go.mod` and `k8s.io/apimachinery/.../uuid` used elsewhere. Prefer `uuid.NewString()`.

pkg/apis/workflow/v1alpha1/common.go
```diff
@@ -72,4 +72,6 @@ type SubmitOpts struct {
 	// Priority is used if controller is configured to process limited number of workflows in parallel, higher priority workflows
 	// are processed first.
 	Priority *int32 `json:"priority,omitempty" protobuf:"bytes,14,opt,name=priority"`
+	// Artifacts overrides input artifact locations. Format: name=s3://bucket/key or name=key
```
Joibel: **Misleading format documentation (confidence 80).** This says `name=s3://bucket/key or name=key` (and the UI model says `name=gcs://bucket/key`), but parsing does `SplitN(…,"=",2)` → `SetKey(value)`, so only a bare key works — a full URL is stored literally and fails at runtime. Only `submit-workflow-panel.tsx`'s comment is correct. Recommend aligning all to `name=key`.

workflow/artifacts/common/common.go
```diff
@@ -24,6 +24,10 @@ type ArtifactDriver interface {
 	// Save uploads the path to artifact destination
 	Save(ctx context.Context, path string, outputArtifact *v1alpha1.Artifact) error
 
+	// SaveStream saves an artifact from an io.Reader.
+	// This allows for memory-efficient streaming uploads.
```
Joibel: **Comment inaccurate for most implementations (confidence 75).** "memory-efficient streaming uploads" is true only for Azure/HTTP; s3/gcs/oss/hdfs/plugin buffer the entire reader to a temp file first. Consider rewording to reflect the temp-file buffering.

.spelling
```diff
@@ -32,6 +32,7 @@ Ang
 Anthos
 Approvers
 ArgoLabs
+ArgoServerOpts
```
Joibel: **Unrelated dictionary additions — scope creep (confidence 92).** ~39 of the added tokens (`WatchWorkflows`, `ArgoServerOpts`, `WorkflowNodePodDetails`, `workflow_server.go`, `sqldb`, …) don't appear in any file this PR touches; they look like leftovers from another branch. Recommend dropping the ones unrelated to this feature.

workflow/util/merge_test.go
```diff
@@ -727,3 +727,77 @@ func TestAllWorkflowSpecFieldsAccountedFor(t *testing.T) {
...
+  generateName: workflow-with-artifact-override-
+spec:
+  workflowTemplateRef:
+    name: template-with-artifact
+  arguments:
+    artifacts:
+      - name: my-uploaded-file
+        s3:
+          endpoint: s3.amazonaws.com
+          bucket: panicboat-sandbox-723535945756
```
Joibel: **Author-specific values in test fixtures (confidence 75).** `panicboat-sandbox-723535945756` (a 12-digit AWS account id) and `debug-s3-creds` look copy-pasted from manual debugging. Recommend neutral placeholders (e.g. `my-bucket`, `s3-creds`).

workflow/artifacts/hdfs/hdfs.go
```diff
@@ -57,7 +58,9 @@ func ValidateArtifact(errPrefix string, art *wfv1.HDFSArtifact) error {
 	if art.Path == "" {
 		return errors.Errorf(errors.CodeBadRequest, "%s.path is required", errPrefix)
 	}
-	if !filepath.IsAbs(art.Path) {
+	// HDFS paths are always Unix-style, so check for leading "/" instead of filepath.IsAbs
```
Joibel: **Unrelated change bundled in (confidence 75).** This `filepath.IsAbs → strings.HasPrefix` validation fix (and the azure Windows path-separator fix, and the `controller_test.go` task-result fixture change) aren't used by the new `SaveStream` paths and have no stated connection to this feature. Consider splitting into a separate PR.

ui/src/shared/components/artifacts-input/artifacts-input.tsx
```diff
@@ -0,0 +1,181 @@
+import * as React from 'react';
```
Joibel, line 1: **No tests for this component (confidence 75).** This 181-line component has non-trivial XHR/state-machine/drag-drop logic but no `*.test.tsx`, despite the repo's established UI component test pattern (button, object-editor, etc.). Worth adding coverage for the success/error/parse-failure paths.

workflow/artifacts/s3/s3_test.go
```diff
@@ -713,6 +748,127 @@ func TestNewClientWithDiff(t *testing.T) {
 	})
 }
 
+func TestSaveStreamS3Artifact(t *testing.T) {
```
Joibel: **Test exercises only the mock, not the driver (confidence 85).** `TestSaveStreamS3Artifact`/`TestArtifactDriverSaveStream` call `mockS3Client.PutStream` directly and assert the mock's own return — the real `ArtifactDriver.SaveStream` (temp-file buffering, retry/backoff, error wrapping) is never run. The new `S3Client` interface + second mock add no real coverage.

server/workflow/workflow_server.go
```diff
@@ -827,11 +831,111 @@ func (s *workflowServer) SubmitWorkflow(ctx context.Context, req *workflowpkg.Wo
...
+	// When using workflowTemplateRef, wf.Spec.Arguments.Artifacts is empty,
+	// so we need to copy artifacts from the template and override the keys
+	if req.SubmitOptions != nil && len(req.SubmitOptions.Artifacts) > 0 && wf.Spec.WorkflowTemplateRef != nil {
+		logger.WithFields(logging.Fields{
+			"artifacts":           req.SubmitOptions.Artifacts,
+			"workflowTemplateRef": wf.Spec.WorkflowTemplateRef.Name,
+		}).Debug(ctx, "Processing artifact overrides for workflowTemplateRef")
+		var tmplArtifacts []wfv1.Artifact
+		var artifactRepositoryRef *wfv1.ArtifactRepositoryRef
+		if wf.Spec.WorkflowTemplateRef.ClusterScope {
```
Joibel, line 860: **Untested branches (confidence 75).** `TestSubmitWorkflowWithArtifactOverride` only covers namespaced WorkflowTemplate; this `ClusterScope` path (distinct API call + error) and the malformed-override warn-and-skip branch are uncovered.

workflow/artifacts/plugin/plugin.go
```diff
@@ -214,6 +214,26 @@ func (d *Driver) Save(ctx context.Context, path string, outputArtifact *wfv1.Art
 	return nil
 }
 
+// SaveStream implements ArtifactDriver.SaveStream by using a temporary file
+// Note: gRPC streaming upload is complex, so we use a temp file fallback
+func (d *Driver) SaveStream(ctx context.Context, reader io.Reader, outputArtifact *wfv1.Artifact) error {
```
Joibel, line 234: I would have expected this to allow the plugin to stream the artifact via a new driver call if the driver implements it, only falling back to calling `Save` if it is unimplemented.

workflow/artifacts/azure/azure.go
```diff
@@ -304,6 +304,26 @@ func (azblobDriver *ArtifactDriver) Save(ctx context.Context, path string, outpu
 	return nil
 }
 
+// SaveStream saves an artifact from an io.Reader to Azure Blob Storage
+func (azblobDriver *ArtifactDriver) SaveStream(ctx context.Context, reader io.Reader, outputArtifact *wfv1.Artifact) error {
```
Joibel, line 308: This and the other drivers that are bouncing the file off disk should probably have a common implementation, like `LoadToStream` is a common implementation for OpenStream in drivers that bounce off disk.

Joibel (changes requested):

workflow/artifacts/s3/s3.go
panicboat: Covered in the prerequisite PR #16407. A single `common.SaveStreamViaTempFile` helper is used by s3/gcs/oss/hdfs, and each driver delegates back to its existing `Save` so bucket creation, key normalization, and retry/backoff aren't reimplemented per driver. Azure is native `UploadStream` and HTTP feeds the request body directly, so neither goes through the helper.

workflow/artifacts/plugin/plugin.go
panicboat: Addressed in the prerequisite PR #16407. `artifact.proto` gains a client-streaming `SaveStream` RPC and a `GetCapabilities` RPC (no HTTP gateway — plugins talk direct gRPC over a unix socket). The driver calls `GetCapabilities` first; if streaming is advertised it sends chunks (2 MiB), otherwise it falls back to buffering to a temp file and calling the unary `Save`. Older plugins that don't implement the new RPCs surface `Unimplemented` and hit the fallback, so it's backward compatible. Mid-stream failures are hard errors rather than silent fallbacks (the reader can't be rewound). Tests in #16407 cover streaming success, capability=false fallback, and mid-stream failure.

workflow/artifacts/azure/azure.go
panicboat: Addressed in the prerequisite PR #16407. `common.SaveStreamViaTempFile` centralizes the temp-file → `Save` handoff for s3/gcs/oss/hdfs; azure keeps its native `UploadStream` path and http keeps direct request-body streaming, so neither one goes through the helper.

workflow/artifacts/http/http.go
panicboat: Covered in the prerequisite PR #16407, which replaces the `SaveStream` path in `workflow/artifacts/http/http.go` and shares construction with `Save` — the nil guard is in that shared setup, so both entry points are safe.

workflow/artifacts/common/common.go
panicboat: Superseded by #16407's revised interface comment, which is now the authoritative description of the reader contract ("the reader may be partially consumed on failure; callers must not retry with the same reader").

workflow/artifacts/s3/s3_test.go
panicboat: The retry/temp-file behavior this comment was pointing at has moved to the prerequisite PR #16407. The s3 driver's `SaveStream` now delegates to `Save` through `common.SaveStreamViaTempFile`, so the retry/backoff semantics are exercised by the existing `Save` tests plus a `common/streaming_test.go` covering the buffering helper (create-temp → copy → close → cleanup, and cleanup idempotency). The mock-only `TestSaveStreamS3Artifact` from this branch was dropped in fb333866 because the code path it was mocking no longer exists on this side.

.spelling
panicboat: Resolved by rebase. `.spelling` was removed from the repo in #16272, so the stale additions this PR carried have no target file anymore and disappeared. `make lint-docs` passes.

server/artifacts/artifact_server.go
panicboat: Addressed in 5e0152a2 (+ test in 30e6182e5).

Order was chosen deliberately per your point: `auth.CanI` runs **before** `MaxBytesReader`/`ParseMultipartForm` so an unauthorized caller is rejected without reading the body. Permissions required: `get workflowtemplates` (to fetch the template) **and** `create workflows` in the target namespace. `instanceIDService.Validate(wfTemplate)` runs after Get but before touching the artifact.

server/artifacts/artifact_server.go
panicboat: Addressed in f3000584e (+ test in 2c27c907c).

`http.MaxBytesReader` wraps the body immediately after `auth.CanI` passes. The limit is controlled by env `ARGO_SERVER_MAX_ARTIFACT_UPLOAD_BYTES`. `ParseMultipartForm` errors are unwrapped: `*http.MaxBytesError` returns **413**, other parse failures return 400. Documented in `.features/pending/input-artifact-upload.md`.

server/artifacts/artifact_server.go
panicboat: Addressed in 592401ecf (+ test in 35d7c4041).

`defer r.MultipartForm.RemoveAll()` is placed right after `ParseMultipartForm` succeeds, so every return path — including the error paths downstream (bucket resolution, driver save) — removes the spilled parts. The test uses `t.TempDir` and asserts via `os.Stat` that no leftover files remain on both success and error paths.

server/workflow/workflow_server.go
panicboat: Addressed in 59590b07a (+ test in 651cd950c), with a5ff37a8b (+ test in fe044bdff) landing first so the format is well-defined before we validate against it.

The submitted key must satisfy **all** of: exact `uploads/{namespace}/` prefix; `path.Clean(key) == key`; no `..` segments; not absolute; four `/`-separated segments; segment 3 parses as an RFC-4122 UUID; segment 4 equals `path.Base(...)`. Rejection returns `codes.InvalidArgument`. The upload endpoint self-validates the key it just generated so the two sides can't drift.

**Residual risk (documented in the feature doc):** this restricts the key to the server-controlled prefix but does not *prove ownership* — a submitter with `workflows create` in the same namespace could technically point at another user's upload key mid-upload. The UUID's high entropy makes this hard to guess in practice. A cryptographic ownership token (HMAC over `namespace/subject/wftmpl/artifact/key/expiry/kid`) would be worth doing if audit requirements come up, but it adds new signing-key infrastructure so I've kept it out of this PR and will file a follow-up.

server/artifacts/artifact_server.go
panicboat: Addressed at the "at least a bucket lifecycle/TTL" end of your ask in 37044818c.

`.features/pending/input-artifact-upload.md` now has an **Abandoned Upload Cleanup** section with S3/GCS/OSS lifecycle rule examples targeting the `uploads/` prefix, plus the CLI equivalents. The PR description also flags this operational dependency.

GC linkage (marking submitted uploads and hooking them into the existing artifact GC) is out of scope for this PR — it touches controller/GC semantics and I'd rather land it separately once the operator docs bake in. Will file a follow-up.

server/workflow/workflow_server.go
panicboat: Addressed in 221d9953e (+ test in 2f0508367), with a follow-up c7fcce6c0 so the parse error surfaces as `codes.InvalidArgument` on both the templateRef branch and the `util.ApplySubmitOpts` path.

Both code paths now go through `ParseArtifactOverrides`, which is `error`-returning (fail-fast). Semantics fixed by tests: duplicate name → **last wins**; empty name/key or missing `=` → **rejected**.

workflow/util/util.go
panicboat: Addressed together with the parse unification in 221d9953e. `workflow/util` now owns the pure functions (`ParseArtifactOverrides`, `ApplyOverridesToTemplateArtifacts`), and `server/workflow/workflow_server.go` composes them with the shared repo-resolution helper so the two entry points share the same override semantics. Deliberately kept artifact-repository resolution on the server side (needs k8s client access) rather than pulling it into `util`.

server/artifacts/artifact_server.go
panicboat: Addressed in 539e77666 (+ test in 03f88e99e). Both `artifact_server.go` (upload) and `workflow_server.go` (submit) call the shared `ResolveArtifactLocation` helper, and the key validator lives alongside it so the server-side concerns stay together.

server/workflow/workflow_server.go
panicboat: Addressed in ba6948a8c. The per-artifact debug closure and the "Final workflow artifacts before creation" block are gone; the surviving logs are meaningful ones (override count summary).

server/artifacts/artifact_server.go
panicboat: Addressed in a5ff37a8b (+ test in fe044bdff).

`generateUUID` is removed; the upload path uses `github.com/google/uuid` (`uuid.NewString`). This landed before the key validator (59590b07a) so the validator can assert RFC-4122 format on segment 3.

pkg/apis/workflow/v1alpha1/common.go
panicboat: Addressed in 831f24331 (with codegen).

The doc string, model comments, and generated schemas all say `name=key` now. Kept the parser at `name=key` deliberately — accepting `name=s3://bucket/key` would put storage-location switching into `SetKey`, which pulls credential/endpoint concerns into an API surface that's supposed to just point at an object; combined with the key-validation constraints the tradeoffs get ugly fast.

workflow/util/merge_test.go
panicboat: Addressed in dfddb29ea.

Replaced the author-specific values only (`panicboat-sandbox-723535945756` → `my-bucket`, `debug-s3-creds` → `s3-creds`). Kept the artifact keys (`uploads/argo/{uuid}/file.zip`, `input-file/placeholder`) since they carry the override semantics the test is asserting.

workflow/artifacts/hdfs/hdfs.go
panicboat: Addressed as 3a7bf997f and 257b20f82.

Small caveat on the path-separator revert: it turned out the new `hdfs_test.go`/`azure_test.go` cases in this PR depended on the `strings.HasPrefix` behavior on Windows (`filepath.IsAbs` doesn't treat drive-less `/x` as absolute on Windows), so the fix was reapplied in e0547f918 to keep the Windows CI green. The controller_test fixture revert stays.

server/workflow/workflow_server.go
panicboat: Addressed in d5fea6f1d. Adds `TestSubmitWorkflowWithArtifactOverride_ClusterScope` (exercises the `ClusterWorkflowTemplate` path so the client fetches from `ClusterWorkflowTemplates()` rather than `WorkflowTemplates(ns)`, and the key override + non-key location preservation still hold) and `TestSubmitWorkflowWithArtifactOverride_MalformedOverride` (missing `=` / empty name / empty key → `codes.InvalidArgument`).

ui/src/shared/components/artifacts-input/artifacts-input.tsx
panicboat: Addressed in 98f89430e. `artifacts-input.test.tsx` covers: file selection → POST /upload-artifacts/... with URL-encoded path segments, `onUploadStart` fires, `onUploadComplete` receives the parsed server response; upload progress events update the progress indicator; non-2xx response surfaces the status text as an error; 2xx with malformed JSON body raises "Failed to parse response"; XHR `onerror` surfaces "Network error"; drag-and-drop file onto the dropzone triggers the same upload path.

panicboat:
**Reviewer note — please review #16407 first.**

This PR depends on **#16407** (`feat: add SaveStream to artifact drivers with plugin gRPC streaming`). #16407 adds `SaveStream` to the `ArtifactDriver` interface, implements it across all drivers (s3/gcs/oss/hdfs via a shared `SaveStreamViaTempFile` helper; azure/http natively), and adds a plugin gRPC client-streaming path with capability negotiation and a temp-file fallback. It is a self-contained prerequisite.

Because #16407's base is also `argoproj/argo-workflows:main`, GitHub currently shows #16407's commits in this PR's diff too — that noise disappears once #16407 merges. If you review this PR before #16407 lands, please skip the driver-side changes and focus on:

- `server/artifacts/artifact_server.go` — upload endpoint, authz ordering, size limit, multipart cleanup, key generation
- `server/workflow/workflow_server.go` — submit-side key validation, override apply
- `workflow/util/{util,merge}.go` — pure override helpers and the field-preserving merge fix
- `ui/src/shared/components/artifacts-input/**` and `ui/src/workflows/components/submit-workflow-panel.tsx` — upload UI
- `.features/pending/input-artifact-upload.md` — operator guidance for the `uploads/` bucket lifecycle

workflow/artifacts/gcs/gcs_test.go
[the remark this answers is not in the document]
panicboat: The gcs test cases this comment was pointing at (`TestListFileRelPaths*`) were removed as tangential in the prerequisite PR #16407 (`test(artifacts): fix azure test comment and drop tangential gcs test`), so the Windows-portability concern is resolved by their removal rather than a `t.TempDir`-based fix. Resolving.

workflow/artifacts/http/http.go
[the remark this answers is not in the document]
panicboat: Covered in the prerequisite PR #16407, which replaces the `SaveStream` path in `workflow/artifacts/http/http.go` and shares construction with `Save` — the nil guard is in that shared setup, so both entry points are safe. Resolving.

workflow/artifacts/oss/oss.go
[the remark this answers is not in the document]
panicboat: Confirmed as addressed. The `SaveStream` setup was aligned with `Save` in cc74c24e8. In the prerequisite PR #16407 this is unified further — `SaveStream` delegates directly to `Save` through `common.SaveStreamViaTempFile`, so there is a single setup path end-to-end. Resolving.

workflow/artifacts/s3/s3_test.go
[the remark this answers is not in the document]
panicboat: Covered in the prerequisite PR #16407 rather than here. The s3 driver's `SaveStream` now delegates to `Save` through `common.SaveStreamViaTempFile`, so the retry/backoff semantics are exercised by the existing `Save` tests plus a `common/streaming_test.go` covering the buffering helper. The mock-only `TestSaveStreamS3Artifact` from this branch was dropped in fb333866 because the code path it was mocking no longer exists on this side. Resolving.

workflow/util/merge.go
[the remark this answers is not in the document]
panicboat: Addressed in 30f49cc8d (+ test in 297ca1b29). The merge now walks the artifact fields explicitly and only overrides those that are non-zero on the higher-precedence side, so a `key`-only override in the Workflow spec keeps the `bucket`/`endpoint`/`credentialSecret` inherited from defaults. Resolving.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,217 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+		}
+		http.Error(w, "Failed to parse multipart form: "+parseErr.Error(), http.StatusBadRequest)
+		return
+	}
+	defer func() {
+		if r.MultipartForm != nil {
+			_ = r.MultipartForm.RemoveAll()
+		}
+	}()
+
```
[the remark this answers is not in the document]
panicboat: Applied in d7acbf0f4: hoisted `defer r.MultipartForm.RemoveAll()` above `ParseMultipartForm`.

I want to be transparent about the scope though: on current Go (1.20+), `mime/multipart.ReadForm` already has an internal `defer` that calls `form.RemoveAll()` on parse error, including the `MaxBytesError` path. I verified this with a spill-to-disk MaxBytes test against the pre-fix code — no temp files leaked. So this is not a real leak today; the change is a defence-in-depth positioning so the handler's cleanup contract does not depend on that stdlib internal.

server/artifacts/artifact_server.go
```diff
@@ -68,6 +71,217 @@ func (a *ArtifactServer) GetInputArtifact(w http.ResponseWriter, r *http.Request
...
+	if !artifactCopy.HasLocation() {
+		http.Error(w, fmt.Sprintf("Artifact '%s' does not have a storage location configured (s3, gcs, azure, oss). Please configure a storage location in the WorkflowTemplate or set up a default artifact repository.", artifactName), http.StatusBadRequest)
+		return
+	}
+
+	// Generate unique key for the artifact
+	uploadUUID := uuid.NewString()
+	originalKey, _ := artifactCopy.GetKey()
+	// Sanitize filename to prevent path traversal attacks
+	sanitizedFilename := path.Base(header.Filename)
```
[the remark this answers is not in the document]
panicboat: Applied in d7acbf0f4 with `path.Base(strings.ReplaceAll(header.Filename, "\\", "/"))` and a new test `Success - Windows-style backslash traversal in filename is sanitized` that asserts `..\..\etc\passwd` reduces to `passwd`.

For context on the severity: any filename containing `..` was already rejected downstream by `sutils.ValidateUploadedArtifactKey`'s `strings.Contains(key, "..")` check, so the traversal payload in the finding was blocked even without this fix. The gap this closes is filenames with backslashes but no `..` (e.g. `foo\bar`), which passed through and could take on separator meaning in filesystem-oriented drivers — so the change is applied as defence-in-depth.

.features/pending/input-artifact-upload.md
```diff
@@ -0,0 +1,21 @@
...
+
+This feature works with all supported artifact repositories (S3, GCS, Azure Blob Storage, OSS, HDFS).
+
+Uploaded files are written under the `uploads/{namespace}/{uuid}/{filename}` key in the artifact
+repository before the workflow is submitted. The maximum accepted upload size is controlled by the
+`ARGO_SERVER_MAX_ARTIFACT_UPLOAD_BYTES` environment variable on the Argo Server (default `1073741824`,
+i.e. 1GiB); requests over this size receive `413 Request Entity Too Large`.
+
+Abandoned uploads (never submitted) rely on operator-configured bucket lifecycle under
+`uploads/{namespace}/`; see [Configuring Your Artifact Repository](configure-artifact-repository.md#abandoned-upload-cleanup).
```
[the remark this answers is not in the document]
panicboat: Not applied — the current bare link is intentional.

`.features/pending/*.md` fragments are concatenated by `hack/featuregen/main.go` (`docsOutput = "docs/new-features.md"`) into `docs/new-features.md`, so links resolve from `docs/`, not from `.features/pending/`. The bare `configure-artifact-repository.md#abandoned-upload-cleanup` therefore points at the sibling doc in the rendered output. Switching to `../../docs/configure-artifact-repository.md#abandoned-upload-cleanup` would break the link in the published docs. The same bare convention is already used elsewhere (see `.features/pending/configurable-node-status-compression.md` referencing `offloading-large-workflows.md`).

The trade-off is that the link is not clickable when viewing the raw fragment on GitHub, but that matches the project-wide convention for feature notes.

panicboat:
@Joibel Thank you very much for your thorough review.
I believe I have addressed all of the points you raised, so I would appreciate it if you could review it again when you have a moment.

Joibel (approved):
Thanks for all the hard work on this one.

Joibel:
@panicboat - have you considered becoming a [member](https://github.com/argoproj/argoproj/blob/main/community/membership.md) of argoproj. I'd sponsor your application.

panicboat:
@Joibel Thank you very much.
I’m really happy about that offer.
I’d like to keep contributing going forward, so I’d like to consider applying.
However, I don’t have any potential sponsors in mind, so would it be appropriate to ask for sponsors in [#argo-workflows-contributors](https://cloud-native.slack.com/archives/C0510EUH90V)?

Joibel:
Asking in #argo-workflows-contributors should work, if it doesn't I will reach out on your behalf - you've easily done enough so it will just be a case of ensuring someone reads the message. Feel free to open a membership request with just me as a sponsor for now, and we can add a second sponsor later.

panicboat:
Understood, thank you.
I will contact you via Slack 🙏

panicboat:
I've created this as a WIP for now.
https://github.com/argoproj/argoproj/issues/459

PR #15239, opened 2026-01-13
fix: remove holder from waiting list when semaphore lock is acquired.

shuangkun:
When a workflow acquires a semaphore lock, it should be removed from the waiting list to prevent the workflow YAML from becoming too large and to ensure accurate status reporting.

This fix ensures that:
- When LockAcquired is called, the holder is removed from Waiting list
- If the waiting list becomes empty, the entire waiting entry is removed
- The fix prevents stale entries in the waiting status

Fixes the issue where workflows that have acquired locks still appear in the waiting list, causing workflow YAML to grow unnecessarily.

Fixes #TODO

### DIFF
--- a/pkg/apis/workflow/v1alpha1/workflow_types.go
+++ b/pkg/apis/workflow/v1alpha1/workflow_types.go
@@ -4125,29 +4125,31 @@ func (ss *SemaphoreStatus) LockWaiting(holderKey, lockKey string, currentHolders
 func (ss *SemaphoreStatus) LockAcquired(holderKey, lockKey string, currentHolders []string) bool {
 	i, semaphoreHolding := ss.GetHolding(lockKey)
 	holdingName := holderKey
-
-	// Remove from waiting list if present
-	waitingIdx, semaphoreWaiting := ss.GetWaiting(lockKey)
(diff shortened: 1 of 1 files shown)


pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -4125,6 +4125,20 @@ func (ss *SemaphoreStatus) LockWaiting(holderKey, lockKey string, currentHolders
...
 	holdingName := holderKey
+
+	// Remove from waiting list if present
+	waitingIdx, semaphoreWaiting := ss.GetWaiting(lockKey)
+	if waitingIdx >= 0 {
+		semaphoreWaiting.Holders = slices.DeleteFunc(semaphoreWaiting.Holders,
+			func(x string) bool { return x == holdingName })
+		if len(semaphoreWaiting.Holders) == 0 {
+			// Remove the entire waiting entry if no holders left
+			ss.Waiting = append(ss.Waiting[:waitingIdx], ss.Waiting[waitingIdx+1:]...)
```
Joibel: perhaps use `slices.Delete()` here?

PR #15241, opened 2026-01-14
fix: Optimize DAG sort for retry

spruett345:
### Motivation
The existing retry functionality has a combinatorial explosion during the DAG sort stage. Consider a workflow
composed of sequential stages, each stage having a set of items in parallel:
```
                START
                  │
          ┌───────┼───────┐
          ▼       ▼       ▼
        [1-1]   [1-2]   [1-3]
          └───────┼───────┘
                  │
          ┌───────┼───────┐
          ▼       ▼       ▼
        [2-1]   [2-2]   [2-3]
          └───────┼───────┘
                  │
          ┌───────┼───────┐
          ▼       ▼       ▼
        [3-1]   [3-2]   [3-3]
          └───────┼───────┘
                  │
                 END
```

The current DAG sort code can reach each next stage as a `child` of all steps from the current stage. Each of them is processed, adding their children, ... which explodes the number of times the child nodes are processed. This leads to either very slow retries (referring to the manual retry button, `argo retry` equivalent) for complex workflows, or OOMs of the Argo server.

### Modifications

This optimizes the topsort to avoid re-visiting the same nodes more than once. This ensures that
the DAG sort is bounded (O(N)) in time and memory.

### Verification

Submit this workflow: https://gist.github.com/spruett345/a73c0c9dd49fcd030c95ad015b03f158
This has 10 sequential stages, each with 20 parallel steps, each with 3 sub steps (mimicking the structure of a real Workflow which I ran to hit this pathological case). Once it has failed, attempt to retry it with `argo retry` -- it OOMs on my machine (10+GB of RAM). 

With this patch, the retry completes in a reasonable time frame without noticeable memory usage.

### Documentation

N/A: no expected behavior change

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved duplicate node processing in workflow traversal operations, preventing nodes from being processed multiple times and improving system performance and correctness.

### DIFF
--- a/workflow/util/util.go
+++ b/workflow/util/util.go
@@ -1110,10 +1110,12 @@ func dagSortedNodes(nodes []*dagNode, rootNodeName string) []*dagNode {
 	}
 
 	queue := make([]*dagNode, 0)
+	visited := make(map[string]bool)
 
 	for _, n := range nodes {
 		if n.n.Name == rootNodeName {
 			queue = append(queue, n)
+			visited[n.n.ID] = true
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


Joibel:
@coderabbitai review

jsloyer (approved):
LGTM

Issue #15246, opened 2026-01-16
`make pre-commit` fails on aarch64 Linux: snipdoc binary not available


panicboat:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [x] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

Running `make pre-commit` on an aarch64 (ARM64) Linux machine fails with a 404 error when downloading snipdoc:

```
Downloading snipdoc from https://github.com/kaplanelad/snipdoc/releases/download/v0.1.12/snipdoc-aarch64-linux.tar.xz
curl: (22) The requested URL returned error: 404
make: *** [Makefile:455: /home/user/.local/bin/snipdoc] Error 22
```

**What I expected:**

The `make pre-commit` command should complete successfully, or gracefully handle the missing binary.

**Root cause:**

The `hack/install-snipdoc.sh` script generates a download URL based on the detected architecture and OS. However, the upstream snipdoc project (kaplanelad/snipdoc) does not provide pre-built binaries for `aarch64-linux`.

Available assets for v0.1.12:
- snipdoc-aarch64-macos.tar.xz
- snipdoc-x86_64-linux.tar.xz
- snipdoc-x86_64-macos.tar.xz
- snipdoc-x86_64-windows.zip

**Proposed fix:**

Add a fallback to `cargo install snipdoc` when the pre-built binary is not available.

### Version(s)

227878a76c673c2a1745d1f362d4a310c6a305fe

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

```YAML
None
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

Joibel:
Closed by #15274

PR #15247, opened 2026-01-16
fix: support snipdoc installation on aarch64-linux via cargo fallback

panicboat:
Fixes https://github.com/argoproj/argo-workflows/issues/15246

### Motivation

snipdoc pre-built binaries are not available for aarch64-linux platforms, causing installation failures in devcontainer environments on ARM-based systems (e.g., Apple Silicon Macs, ARM cloud instances). This prevents
developers from building documentation on these platforms.

### Modifications

- Modified `hack/install-snipdoc.sh` to implement a fallback mechanism:
  - First attempts to download pre-built binary
  - Falls back to `cargo install` if download fails (e.g., on aarch64-linux)
  - Adds proper error handling and user feedback
  - Copies cargo-installed binary to the expected location
- Added Rust feature to `.devcontainer/devcontainer.json` to ensure Rust toolchain is available for building snipdoc from source when needed

### Verification

Tested on aarch64-linux devcontainer environment:
- Verified that pre-built binary download fails gracefully
- Confirmed cargo fallback successfully builds and installs snipdoc
- Verified the installed binary is placed in the correct location and is executable
- Confirmed documentation builds work correctly with cargo-installed snipdoc

### Documentation

No documentation update needed.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,6 +5,11 @@
   // This image is built and pushed by .github/workflows/devcontainer.yaml using .devcontainer/builder/devcontainer.json
   "image": "quay.io/argoproj/argo-workflows-devcontainer",
 
+  "features": {
+    // Required for building snipdoc on aarch64-linux (no pre-built binary available)
+    "ghcr.io/devcontainers/features/rust:1": {}
+  },
+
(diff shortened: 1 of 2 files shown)


Joibel:
Thanks for doing this, but I didn't like adding rust as a dependency, so I've done #15274 as a different approach as we don't need most of the power of snipdoc.

PR #15250, opened 2026-01-18
fix: optimize index to prevent 'out of sort memory' error Fixes #14240

### LINKED ISSUE #14240: UI shows out of memory in sort
### Pre-requisites

I have a namespace with more than 50 workflows
When I use latest version:

Image

When I use v3.6.4, it works well:

Image

When the workflows num is only five, all work well.

### Version(s)

latest

```YAML
over 50 simple workflows in one namespace.
```

```text
```

```text
```

downfa11:
Fixes #14240 

### Motivation

The existing index(`argo_archived_workflows_i1`) lacked a sort order, causing the database to extract all data first and then perform sorting on startedat within the sort buffer. This resulted in 'Out of sort memory' errors and system failure.
  
This issue specifically occurs under the following conditions:
- The optimizer selects `argo_archived_workflows_i1` instead of `argo_archived_workflows_i4` during namespace filtering.
- The workflow JSON data is large enough to exceed the MySQL `sort_buffer_size`.

### Modifications

Since this modification includes all existing index columns, it resolves the failure without causing any side effects to existing queries.
- Added `startedat DESC` to the `argo_archived_workflows_i1(clustername, instanceid, namespace)` index.

### Verification

Detailed reproduction steps and query execution plans can be found in this https://github.com/argoproj/argo-workflows/issues/14240#issuecomment-3753236025

**Actual UI failure state**:
스크린샷 2026-01-18 133344

**Resolved state under the same conditions**:
스크린샷 2026-01-18 133302

### DIFF
--- a/persist/sqldb/migrate.go
+++ b/persist/sqldb/migrate.go
@@ -188,12 +188,6 @@ func Migrate(ctx context.Context, session db.Session, clusterName, tableName str
 		}),
 		// add argo_archived_workflows index
 		sqldb.AnsiSQLChange(`create index argo_archived_workflows_i1 on argo_archived_workflows (clustername,instanceid,namespace)`),
-		// change argo_archived_workflows_i1 index to add startedat DESC to resolve MySQL out of sort memory issues. #14240
-		sqldb.ByType(dbType, sqldb.TypedChanges{
-			sqldb.MySQL:    sqldb.AnsiSQLChange(`drop index argo_archived_workflows_i1 on argo_archived_workflows`),
-			sqldb.Postgres: sqldb.AnsiSQLChange(`drop index argo_archived_workflows_i1`),
-		}),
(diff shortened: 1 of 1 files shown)


shuangkun:
Nice. Could you add Fixes #14240 at the end of the title.

persist/sqldb/migrate.go
```diff
@@ -188,6 +188,12 @@ func Migrate(ctx context.Context, session db.Session, clusterName, tableName str
 		}),
 		// add argo_archived_workflows index
 		sqldb.AnsiSQLChange(`create index argo_archived_workflows_i1 on argo_archived_workflows (clustername,instanceid,namespace)`),
+		// change argo_archived_workflows_i1 index to add startedat DESC to resolve MySQL out of sort memory issues. #14240
```
MasonM: These lines should go at the end of the `changes []Change` array, since the `schema_version` is based on the index of each entry in the array.

downfa11: Thank you for the review! I missed that part, but I've corrected it now. 

could you take another look when you have a moment?

PR #15251, opened 2026-01-18
fix: Add optional UID query parameter to GetWorkflow

### LINKED ISSUE #15232: UI fails to load retried archived workflow
### Pre-requisites

When an archived workflow is retried, the workflow gets a new UID.  The UI lists both workflows:

Image

However, when clicking in any of the workflows, they all link to the same workflow (the most recent one).

This happens because the workflow server `GetWorkflow` is ignoring the UID when requesting from the archived storage:

https://github.com/argoproj/argo-workflows/blob/227878a76c673c2a1745d1f362d4a310c6a305fe/server/workflow/workflow_server.go#L752C2-L768C3

```go
    wf, origErr := wfClient.ArgoprojV1alpha1().Workflows(namespace).Get(ctx, name, options)
    // fallback to retrieve from archived workflows
    if wf == nil || origErr != nil {
        allowed, err := auth.CanI(ctx, "get", workflow.WorkflowPlural, namespace, name)
        if err != nil {
            return nil, getWorkflowOrigErr(ctx, origErr, err)
        }
        if !allowed {
            err = status.Error(codes.PermissionDenied, "permission denied")
            return nil, getWorkflowOrigErr(ctx, origErr, err)
        }


        wf, err = s.wfArchive.GetWorkflow(ctx, "", namespace, name)
        if wf == nil || err != nil {
            return nil, getWorkflowOrigErr(ctx, origErr, err)
        }
    }
```

by default the archived workflow returns the most recent one when the UID isn't provided:

https://github.com/argoproj/argo-workflows/blob/227878a76c673c2a1745d1f362d4a310c6a305fe/persist/sqldb/workflow_archive.go#L556C4-L562C5

```go
if num > 1 {
				logger.WithFields(logging.Fields{
					"namespace": namespace,
					"name":      name,
					"num":       num,
				}).Debug(ctx, "returning latest of archived workflows")
			}
```

### Version(s)

v3.7.7

```YAML
# This example demonstrates the use of retries for a single container.
kind: Workflow
metadata:
  generateName: retry-on-error-
  labels:
    workflows.argoproj.io/no-test: "environment"
spec:
  entrypoint: error-container
  ttlStrategy:
    secondsAfterCompletion: 10 # Time to live after the workflow is completed, regardless of its status (overrides secondsAfterSuccess and secondsAfterFailure).
    secondsAfterSuccess: 5     # Time to live after workflow is successful
    secondsAfterFailure: 5
  templates:
  - name: error-container
    retryStrategy:
      limit: "1"
      retryPolicy: "Always"   # Retry on errors AND failures. Also available: "OnFailure" (default), "OnError", and "OnTransientError" (retry only on transient errors such as i/o or TLS handshake timeout. Available after v3.0.0-rc2)
    container:
      image: python:alpine3.23
      command: ["python", "-c"]
      # fail with a 80% probability
      args: ["import random; import sys; exit_code = random.choice(range(0, 5)); sys.exit(exit_code)"]
```

```text
```

```text
```

eduardodbr:
Fixes #15232

This PR adds support for an optional `uid` query parameter to the existing `GetWorkflow` API endpoint (`/api/v1/workflows/{namespace}/{name}`). This enables more precise workflow identification, particularly useful when accessing archived workflows that may have been recreated with the same name. This implementation uses a query parameter approach (?uid=...), this ensures full backward compatibility with all existing endpoints.

### Motivation

When a workflow is archived and a new workflow with the same name is created, the existing API cannot distinguish between them. This makes it difficult to access the correct workflow, especially in the UI when viewing workflow history or when workflows have been archived. More info #15232

### Modifications

API/Server:

- Added  a new field `uid` in the protobuf definition
- Search for `namespace/name` in the API server, if not found or if `uid` is provided and doesn't match the live workflow's UID, the server falls back to the workflow archive to retrieve the correct workflow by UID

UI:

- Modified UI to pass the UID from URL query params when fetching workflows

### Verification

Unit tests and E2E tests


isubasinghe (approved):

isubasinghe:
@eduardodbr can you make this a fix instead? Amend the commit message as well.

Thanks

eduardodbr:
@isubasinghe  I did amend the commit message but I can't change the PR title. I also remove the feature file. CI is still running `feature-pr-handling`.

PR #15260, opened 2026-01-20
feat: #12442 improve s3 upload speed

### LINKED ISSUE #12442: Implement parallelization to speed up S3 artifacts upload and download
# Summary

Implement a client that supports parallelism such as [s5cmd](https://github.com/peak/s5cmd) to enable much faster S3 artifacts I/O speeds, which is crucial when working with many and\or large files.

## Use Cases

When working with many and\or large artifacts via S3 artifact repository on a Workflow, the part of uploading and downloading artifacts in the beginning of every step can take more than the processing time of the step itself.

---

### LINKED ISSUE #15636: feat: improve s3 upload speed
# Summary

What change needs making?

The S3 upload code used in artifact upload.

## Use Cases

When would you use this?

In environment where workflows with large artifacts (>10GB) where stage out is a bit slow due to non optimal upload thread number/part size.

In such environment, one can configure with env variables inside the Argo Workflows ConfigMap the number of threads / part size that works well with a given S3 provider.

Benchmark and more detail in https://github.com/argoproj/argo-workflows/pull/15260
---

antoinetran:
Fixes #12442 improve s3 upload speed (issue ticket)
Fixes #15636 (feature ticket)

### Motivation

The need:
Upload in argo wait container is "slow" (23min for 35GiB) while a file download is fast (2min for 35GiB) to write on disk. More benchmark will follow to confirm.

The how:
The argo code uses minio but does not allow the number of threads for S3 multipart upload, nor does it allow setting the part size. Reading the minio code, it seems the default threads is 4 and the default part size is calculated accordingly to the file size, but generally it is 16MiB (for file size <= 156GiB). This PR allows setting both of them with argo env var, as it is the case for other argo params for argoexec image.

### Modifications

The change is to add env var in doc and read them, if they are defined, to set the part size and threads number. The part size is also checked so that it is a valid value for Minio: 5MiB < part size < 5GiB and file size > part size.

### Verification

Workflow to test:
```
kind: Workflow
metadata:
  generateName: test-artifact-s3-upload-
spec:
  entrypoint: main
  templates:
    - name: main
      container:
        image: busybox
        command: [ sh, -c ]
        args: [ "echo hello world | tee /tmp/hello_world.txt" ]
      outputs:
        artifacts:
          - name: hello_world
            path: /tmp/hello_world.txt
      podSpecPatch: |
        containers:
        - name: wait
          env:
            - name: ARGO_AGENT_S3_UPLOAD_PART_SIZE_MIB
              value: "500"
            - name: ARGO_AGENT_S3_UPLOAD_THREADS
              value: "8"
```
Expected result in wait logs:
```
Number of threads for S3 multipart upload detected from env ARGO_AGENT_S3_UPLOAD_THREADS: 8
Size of S3 multipart upload part size detected from env ARGO_AGENT_S3_UPLOAD_PART_SIZE_MIB: 500 MiB
```

### Documentation

This page https://argo-workflows.readthedocs.io/en/latest/environment-variables/ will be update thanks to the related commit that add doc.

## Summary by CodeRabbit

* **New Features**
  * Added environment variable configuration for S3 multipart uploads, allowing customization of concurrent upload threads and part size to optimize artifact transfer performance.

* **Documentation**
  * Documented new S3 upload configuration environment variables for Controller settings.


antoinetran:
I tested with this branch https://github.com/antoinetran/argo-workflows/tree/v3.7.1-hardcode-s3-multithread that is created from v3.7.1 in my test cluster with argo v3.7.1. But this pull request is in a branch created from main today.

antoinetran:
We did thorough benchmarks and here are the results.

Test env: private kubernetes cluster with private s3 server
Argo v3.7.1 with patched argoexec (https://github.com/antoinetran/argo-workflows/tree/v3.7.1-hardcode-s3-multithread which shares almost the same code as this pull request, except this is for old 3.7.1)
Benchmark of 10GiB uploads from tmpfs (RAM, so no disk IO) with various threads and part sizes. 10 measure of uploads per couple (threads/partsize) and the average speed is retrieved. 3 cases:
1/ original argoexec
2/ patched argoexec with configurable threads/part size + `ConcurrentStreamParts = false`
3/ patched argoexec with configurable threads/part size + `ConcurrentStreamParts = true`

1/ shows upload speeds at ~30MiB/s
2/ shows upload speeds that increases from ~50MiB/s to ~130MiB/s, with this observation: the speed increases with the with the part size but is mostly independent from the number of threads
3/ shows uploads speeds that is ~200MiB/s, with little effect from part size, and increase when the threads increase. But at threads = 4, it is already almost the maximum.

My understanding of `ConcurrentStreamParts` after reading minio code is this enables intermediate buffers for parallel uploads, while disabling it will still do parallel uploads, but with seekable input (no buffer). The only drawback would be it requires a bit more memory: number of threads * part size. But with the default value of 16MiB part size and 4 threads, it means the wait container must be >64MiB of RAM, which seems ok for general usage. So I do not see any good reason to add a configuration env var for this parameter, unless someone says the contrary here.

antoinetran:
This is a graph of the case 3 benchmarks
image

antoinetran:
This is a graph of the case 2 benchmarks
newplot

Joibel:
@coderabbitai review

workflow/artifacts/s3/s3.go
```diff
@@ -504,6 +505,99 @@ func NewS3Client(ctx context.Context, opts S3ClientOpts) (S3Client, error) {
...
+		nbThreads, err = strconv.Atoi(nbThreadsStr)
+		if err != nil {
+			logging.RequireLoggerFromContext(s.ctx).WithFields(logging.Fields{"envvar": common.EnvAgentS3UploadThreads, "nbThreads": nbThreadsStr}).Error(s.ctx, "Could not convert to int the env var. Ignoring and using default value of 4")
+			nbThreads = defaultThreads
+		}
+	} else {
+		nbThreads = defaultThreads
+	}
+	return nbThreads
+}
```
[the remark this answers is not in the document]
antoinetran: fixed as requested

antoinetran:
Can someone rerun the test? One test failed but I guess it is a random fail that has nothing to do with my change.

antoinetran:
Hi! I don't know if there is anything more I can do on this? ping @Joibel ? Thanks!

workflow/artifacts/s3/s3.go
```diff
@@ -513,8 +607,26 @@ func (s *s3client) PutFile(bucket, key, path string) error {
...
+
+	partSizeMiB := s.getFromEnvS3UploadPartSizeMiB()
+	if partSizeMiB > 0 {
+		checkedPartSize, _ := s.getCheckedPartSize(path, partSizeMiB)
+		if checkedPartSize > 0 {
+			logging.RequireLoggerFromContext(s.ctx).WithFields(logging.Fields{"checkedPartSize": checkedPartSize}).Info(s.ctx, "Part size MiB for S3 multipart upload")
+			opts.PartSize = uint64(checkedPartSize)
+		}
+	} else {
+		logging.RequireLoggerFromContext(s.ctx).Info(s.ctx, "Part size not configured from env var. Letting Minio calculating it.")
```
Joibel: nit
```suggestion
		logging.RequireLoggerFromContext(s.ctx).Info(s.ctx, "Part size not configured from env var. Letting Minio calculate it.")
```

workflow/common/common.go
```diff
@@ -185,6 +185,10 @@ const (
 	EnvVarDefaultRequeueTime = "DEFAULT_REQUEUE_TIME"
 	// EnvVarPodStatusCaptureFinalizer is used to prevent pod garbage collected before argo captures its exit status
 	EnvVarPodStatusCaptureFinalizer = "ARGO_POD_STATUS_CAPTURE_FINALIZER"
+	// Number of threads for artifact upload through S3. Default: 4.
```
Joibel: Please follow style for the rest of this:
```suggestion
	// EnvAgentS3UploadThreads is the number of threads for artifact upload through S3. Default: 4.
```

workflow/common/common.go
```diff
@@ -185,6 +185,10 @@ const (
...
+	EnvAgentS3UploadThreads = "ARGO_AGENT_S3_UPLOAD_THREADS"
+	// Size in MiB of the part of a MultiPartUpload. Default: let Minio calculate automatically (16MiB for file <= 156GiB).
```
Joibel:
```suggestion
	// EnvAgentS3UploadPartSizeMiB is the size in MiB of the part of a MultiPartUpload. Default: let Minio calculate automatically (16MiB for file <= 156GiB).
```

Joibel (changes requested):
This is really a feature. Please could you retitle it as such, and add the necessary feature description file. If you'd like it backported I think we can make an exception for this and put it in 4.0.

antoinetran:
> This is really a feature. Please could you retitle it as such, and add the necessary feature description file. 

Thanks for the feedback! I renamed to "feat" this PullRequest, and I created a feature issue https://github.com/argoproj/argo-workflows/issues/15636 and linked it to this ticket. Added the feature description file (MD file).

> If you'd like it backported I think we can make an exception for this and put it in 4.0.

I think you meant if I want to backport it to argo workflows 3.X instead of 4.X? Because the latest tags are 4.X. I am ok with not backporting it to 3.X and migrating to 4.X when this is merged and tagged as a 4.X release.

docs/environment-variables.md
```diff
@@ -22,6 +22,8 @@ This document outlines environment variables that can be used to customize behav
...
 | `ARGO_AGENT_MEMORY_LIMIT`                | `resource.Quantity` | `256m`                                                                                      | Memory resource limit for the agent.                                                                                                                                                                                                                                     |
+| `ARGO_AGENT_S3_UPLOAD_THREADS`           | `int`               | `4`                                                                                         | Number of threads for artifact upload through S3. Default: 4.                                                                                                                                                                                                            |
```
Joibel: These are not agent related controls, so they should be renamed.

They should be in the Executor section of this document. This makes me wonder if you've actually seen them work?

workflow/artifacts/s3/s3.go
```diff
@@ -513,8 +606,26 @@ func (s *s3client) PutFile(bucket, key, path string) error {
 	if err != nil {
 		return err
 	}
+	opts := minio.PutObjectOptions{SendContentMd5: s.SendContentMd5, ServerSideEncryption: encOpts}
+
+	// Enables better performance for parallel uploads (3 times faster in a benchmark of 4 threads 16MiB part size).
+	opts.ConcurrentStreamParts = true
```
Joibel, line 612: This specific option has no effect on FPutObject, so best removed as it's misleading. I'm wondering how  the previous line comment was established

https://github.com/minio/minio-go/blob/v7.0.92/api-put-object-file-context.go#L54
https://github.com/minio/minio-go/blob/v7.0.92/api-put-object.go#L349-L353

workflow/artifacts/s3/s3.go
```diff
@@ -513,8 +606,26 @@ func (s *s3client) PutFile(bucket, key, path string) error {
...
 
-	_, err = s.minioClient.FPutObject(s.ctx, bucket, key, path, minio.PutObjectOptions{SendContentMd5: s.SendContentMd5, ServerSideEncryption: encOpts})
+	nbThreads := s.getFromEnvS3UploadNbThreads()
+	nbThreadsU := uint(nbThreads)
+	opts.NumThreads = nbThreadsU
+
+	partSizeMiB := s.getFromEnvS3UploadPartSizeMiB()
+	if partSizeMiB > 0 {
+		checkedPartSize, _ := s.getCheckedPartSize(path, partSizeMiB)
```
Joibel: This is the only call to getCheckedPartSize(), which returns an error which we're discarding here. We should handle that. I suggest propagating.

Joibel (changes requested):
As this PR will change memory usage of the wait container, I think it needs notes in docs/upgrading.md, and I won't backport it.

docs/environment-variables.md
antoinetran: Oh this is a vocabulary issue, because I thought Argo wait containers are also called agents. Didn't notice that it was under a controller section. Will move it to executor section.

> This makes me wonder if you've actually seen them work?

Yes, you can see the benchmark.

workflow/artifacts/s3/s3.go
antoinetran: I can see why you are saying this. But if you look a bit further, you would see that the option is called later.
```
FPutObject() -> putObjectCommon() -> putObjectMultipartStream()
```
Then in this last function the option is called https://github.com/minio/minio-go/blob/v7.0.92/api-put-object-streaming.go#L47

This theory has also been confirmed by benchmarks which showed, with only this parameter changed, significant performance boost.

antoinetran: > I'm wondering how the previous line comment was established

The comment statement has been established thanks to the benchmarks here https://github.com/argoproj/argo-workflows/pull/15260/#issuecomment-3823320690

Case 2 is with the option as false, case 3 as true.

docs/environment-variables.md
antoinetran: Are `WAIT_CONTAINER_S3_UPLOAD_THREADS` and `WAIT_CONTAINER_S3_UPLOAD_PART_SIZE_MIB` ok?

workflow/artifacts/s3/s3.go
antoinetran: done: err propagated

docs/environment-variables.md
antoinetran: done: doc renamed

antoinetran: I don't know if I am supposed to click on "resolve comment" or let you do that.

docs/environment-variables.md
```diff
@@ -107,6 +108,8 @@ spec:
 | `REMOVE_LOCAL_ART_PATH`                | `bool`          | `false` | Whether to remove local artifacts.                                                                     |
 | `RESOURCE_STATE_CHECK_INTERVAL`        | `time.Duration` | `5s`    | The time interval between resource status checks against the specified success and failure conditions. |
 | `WAIT_CONTAINER_STATUS_CHECK_INTERVAL` | `time.Duration` | `5s`    | The time interval for wait container to check whether the containers have completed.                   |
+| `WAIT_CONTAINER_S3_UPLOAD_THREADS`           | `int`               | `4`                                                                                         | Number of threads for artifact upload through S3. Default: 4.                                                                                                                                                                                                            |
```
Joibel: Sorry, didn't respond before you did. I'd rather they were `ARTIFACT_S3_UPLOAD_THREADS`. See #15237 for why - this will then work in argo-server too.

workflow/artifacts/s3/s3.go
Joibel: Happy with that, thanks.

docs/environment-variables.md
Joibel: > I don't know if I am supposed to click on "resolve comment" or let you do that.

I prefer "resolve" if you're sure the reviewer is happy - e.g. they've said what they want and you've done it - for something like this I prefer not. I wish Github made this a more obvious thing though, different people I've worked with want different things!

Joibel (changes requested):
As this PR will change memory usage of the wait container, I think it needs notes in docs/upgrading.md, and I won't backport it.

docs/environment-variables.md
antoinetran: Good idea! Done!

antoinetran:
Hi @Joibel , it seems to me that any change you asked were addressed. Is-there anything else I can help with?

antoinetran:
> As this PR will change memory usage of the wait container, I think it needs notes in docs/upgrading.md, and I won't backport it.

After this feature, with default settings and input file < GiB, it requires 4 * 16MiB so the minimum with margin would be ~128MiB . Before this feature, after reading the code it seems to me that the number of threads is still 4 and part size still 16MiB, so I don't think it changes the memory consumption in practice.

antoinetran:
Hi, the CI/CD is green again :) Anything else I can do @Joibel ? This feature is a really much needed feature for us for a scientific community project :)

Joibel (approved):
Thanks for your patience and my double posting silliness. LGTM

antoinetran:
Hi @Joibel , may I ask if this will be taken in the 4.0 release (https://github.com/argoproj/argo-workflows/blob/release-4.0/workflow/artifacts/s3/s3.go)? I can see it is in main but not in latest tag 4.0.7. Or will-it be in 4.1.X?

Joibel:
It will be in 4.1, but not in 4.0 (I said above due to memory usage changes I wouldn't back port it). Features usually only go into new releases. 4.1 release candidate is very close, this month for sure.

PR #15345, opened 2026-01-27
chore: enable errorlint linter

Joibel:
### Motivation

Enabling more linters for higher quality code and better standards compliance

### Modifications

Mostly this is a case of switching out %s or %v for %w when printfing errors.

Where `argo-workflows/v3/errors` is imported alongside `errors` the former is imported as `argoerrors` now.

Use errors.Is or errors.As instead of cast or comparisons.

Two areas of note:
* gcs.go: `isTransientGCSErr` now will return true if the string check passes even if the Temporary check returns false. This feels more correct.
* git.go: the fallthrough comment wouldn't have fallen through, but now it does because this is correct

### Verification

CI tests prove or disprove

### Documentation

None needed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling throughout the system to properly wrap and propagate error chains, enabling more accurate error diagnosis and debugging when operations fail.

* **Chores**
  * Enhanced error inspection and wrapping patterns across the codebase for consistency and compatibility with Go's standard error handling practices.

### DIFF
--- a/workflow/artifacts/gcs/gcs.go
+++ b/workflow/artifacts/gcs/gcs.go
@@ -39,6 +39,9 @@ var (
 
 // from https://github.com/googleapis/google-cloud-go/blob/master/storage/go110.go
 func isTransientGCSErr(ctx context.Context, err error) bool {
+	if err == nil {
+		return false
+	}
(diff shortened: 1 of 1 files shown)


Joibel:
@coderabbitai review

isubasinghe (approved):

PR #15463, opened 2026-02-02
refactor: use wg.Go() instead of manual Add+Defer

Joibel:
### Motivation

Simplify goroutine management by using `wg.Go()` method instead of manually calling `wg.Add(1)`, and `wg.Done()`, and passing closure parameters. This reduces boilerplate code and makes the codebase more idiomatic.

### Modifications

Replaced manual `sync.WaitGroup` patterns with `wg.Go()` across multiple files.

### Verification

These are refactoring changes that maintain the same functionality while improving code clarity. The behavior of all goroutine management remains identical - only the implementation pattern has changed to use the more idiomatic `wg.Go()` method.

### Documentation

No documentation changes needed - this is an internal refactoring with no user-facing changes.

### AI

This was mostly written by Claude.

### DIFF
--- a/cmd/argo/commands/common/wait.go
+++ b/cmd/argo/commands/common/wait.go
@@ -24,14 +24,11 @@ func WaitWorkflows(ctx context.Context, serviceClient workflowpkg.WorkflowServic
 	wfSuccessStatus := true
 
 	for _, name := range workflowNames {
-		wg.Add(1)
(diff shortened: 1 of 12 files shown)


Joibel:
Merged into #15464 as the linter will do this for some cases and therefore fail CI without it.

PR #15466, opened 2026-02-02
docs: be explicit about artifact security

Joibel:
### Documentation

This documents that workflows is not responsible for artifact repository security - this is done by the credentials provided only.

AI must be able to see that we allow path traversal as we've had a number of nearly identical security issues around this.
I've discussed it at argo #sig-security to ensure we're aligned on this not being a problem we're going to solve.

### DIFF
--- a/docs/configure-artifact-repository.md
+++ b/docs/configure-artifact-repository.md
@@ -24,6 +24,12 @@ The actual repository used by a workflow is chosen by the following rules:
 2. From a config map named `artifact-repositories` if it has the `workflows.argoproj.io/default-artifact-repository` annotation in the workflow's namespace.
 3. From a workflow controller config-map.
 
+## Security
+
+You should ensure that the credentials provided to your workflow limits access to the artifacts you wish users to be able to access.
+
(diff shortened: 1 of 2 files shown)


1seal:
@Joibel thanks for clarifying the threat model (“argo is not responsible for artifact repository security”) and documenting that ../ traversal is allowed.

i think the docs still need one explicit operator-facing warning to prevent common multi-tenant misconfiguration: if multiple namespaces/tenants share the same artifact

repository and credentials (same bucket/prefix/iam principal), then any user who can read artifacts can potentially enumerate/read artifacts outside their expected namespace/workflow grouping. kubernetes namespace boundaries do not provide isolation at the artifact layer.

suggested doc additions (high signal, no behavior change):
 - “do not share artifact repository credentials across tenants/namespaces if you expect isolation; use separate buckets/prefixes and separate credentials per tenant.”
 - “if you must share a backend, enforce isolation at the repository policy layer (prefix-scoped creds), not via argo-server request paths.”

optional defense-in-depth: consider a config flag to reject .. (or enforce “final key must stay under the computed workflow prefix”) for operators who prefer stricter behavior, even if default remains unchanged.

Issue #15266, opened 2026-01-21
apiclient can't be used in project that uses k8s.io/api 0.35.0


TvdW:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

From the [k8s 1.35.0 changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.35.md):

> Kubernetes API Go types removed runtime use of the github.com/gogo/protobuf library, and are no longer registered into the global gogo type registry. Kubernetes API Go types were not suitable for use with the google.golang.org/protobuf library, and no longer implement `ProtoMessage()` by default to avoid accidental incompatible use. If removal of these marker methods impacts your use, it can be re-enabled for one more release with a `kubernetes_protomessage_one_more_release` build tag, but will be removed in 1.36. (https://github.com/kubernetes/kubernetes/pull/134256, `@liggitt`) [SIG API Machinery, Apps, Architecture, Auth, CLI, Cluster Lifecycle, Instrumentation, Network, Node, Scheduling and Storage]

This potentially breaks using the argo-workflows apiclient in a project that also uses the K8s API client.

### Version(s)

v3.7.8, 832a44dac839665446ab5e8a6dc4fe9a54d23a0e

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

`go get k8s.io/api@v0.35.0`

```
package main

import (
	_ "github.com/argoproj/argo-workflows/v3/pkg/apiclient"
)

func main() {

}
```

blkperl:
This was fixed in the linked PR.

PR #15573, opened 2026-02-16
chore(deps): update k8s.io libraries to v0.35.1

Joibel:
### Motivation

Keeping up with k8s libs.

This PR does not attempt to switch away from gogo-protobuf nor upgrade to grpc-gateway v2.

### Modifications

Bump k8s.io/api, k8s.io/apimachinery, and k8s.io/client-go from v0.33.1 to v0.35.1 in the main module and all SDK go.mod files.

Compatibility fixes:
- Add eventProtoAdapter to wrap corev1.Event for proto.Message (k8s v0.35 dropped ProtoMessage() from core types, which breaks grpc-gateway v1)
- Pass kubernetes_protomessage_one_more_release as build tag... but also
- Strip kubernetes_protomessage_one_more_release build tag from generated.protomessage.pb.go so v1alpha1 types satisfy proto.Message unconditionally for tools that need it.
- Post-process both of these in the Makefile during codegen
- Add Lookup() to gjsonLabels for the new labels.Labels interface
- Add NewClientset() to the argo fake clientset, migrate callers off the deprecated NewSimpleClientset()
- Wrap custom informers with ToListWatcherWithWatchListSemantics (k8s v0.35 enables WatchListClient by default). See https://github.com/kubernetes/kubernetes/issues/135895
- Fix packer/hydrator test thresholds for smaller JSON output (omitzero on zero-value structs shrinks serialized NodeStatus)
- Drop creationTimestamp: null from expected YAML in convert tests due to switch to omitzero
- Make proto-export always re-clone repos instead of caching stale checkouts, so argo-proto.yaml ref bumps take effect immediately (this had me stumped for a bit, so I'm calling this a better choice)
- Codegen regenerated the rest

Note: node status is now smaller, but compatible. New code will read old node status, but we can now fit more in before compressing and presumably before requiring offload.

This should be followed up by a further change to try to remove the protomessage flag and upgrade to grpc-gateway v2, but keeping them separate as this is already big enough

### Verification

Locally ran unit tests, will let CI run 

### Documentation

Internal only, no docs


hyperlingg:
@Joibel could we please merge this? should also resolve https://github.com/argoproj/argo-workflows/issues/15266 and similar issues.

Joibel:
> @Joibel could we please merge this? should also resolve #15266 and similar issues.

Bugging me won't help, it needs a review by another maintainer.

PR #15578, opened 2026-02-17
fix: Disable dualStackEndpoints in minioClient for isolated AWS regions.

AsKc2000:
### Description

- Disable dualStackEndpoints in minioClient for isolated AWS regions.
  -  minioClient by default enables DualStack endpoints but dualStack endpoints are not supported in ISO regions.
  - If the region is an ISO region, set dualstack to false. 

### Verification

- Added Unit tests for the functionality.

### DIFF
--- a/workflow/artifacts/s3/s3.go
+++ b/workflow/artifacts/s3/s3.go
@@ -38,6 +38,13 @@ import (
 
 const nullIAMEndpoint = ""
 
+// isoRegions defines AWS ISO regions (C2S/SC2S) where dual-stack is not supported
+var isoRegions = map[string]bool{
+	"us-iso-east-1":  true,
+	"us-iso-west-1":  true,
+	"us-isob-east-1": true,
+}
+
 type Client interface {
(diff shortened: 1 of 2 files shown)


Joibel:
Why does the workflows code need to know this information - why can't the S3 bucket in a region which doesn't support dual stack not advertise dual stack?

I don't really like hardcoding information into the controller like this either - I'd be happier if the S3 artifact driver learnt a new config parameter to allow everyone (not just AWS users) to disable dualstack.

vnzongzna:
Hi @Joibel thank for feedback, we did tried to address the issue in minio api itself [here](https://github.com/minio/minio-go/pull/2205) but the project has been [archived](https://github.com/minio/minio/blob/master/README.md) & maintainers are not addressing features/fixes. Hence as a fallback we wanted to add this change here to continue consuming OSS argo. We should ideally also plan on looking at alternative for minio API in argo as well.

Joibel:
minio-go isn't archived, just minio.

PR #15582, opened 2026-02-18
chore: add OpenTelemetry tracing infrastructure

Joibel:
# Based on PR #15576

## Chain of upstream PRs as of 2026-02-18

* PR #15576:
  `main` ← `tracing-dbtype`

  * **PR #15582 (THIS ONE)**:
    `tracing-dbtype` ← `tracing-infrastructure`

## Tracing

Part of #12077

This PR is heavily interlinked with #15585 - which actually implements most of the spans which are defined in this PR

These two PRs should be merged at the same time really, split just for reviewability.

### Motivation

Adds foundational tracing support using OpenTelemetry.

The aim is to support workflow tracing via manually added trace points, along with autogenerated http, db and aws libraries.

### Modifications

Infrastructure:
- Tracing config in workflow-controller-configmap (enabled, url, insecure, etc.)
- TracerProvider lifecycle management in controller and executor
- Span creation/propagation helpers in workflow/tracing and workflow/executor/tracing
- HTTP, K8s API, and S3 transport wrappers for context propagation
- Workflow-level trace context management (trace ID in labels/annotations)

Refactoring:
- Rename Config -> MetricsConfig, Metrics -> MetricsProvider for clarity
- Split telemetry builder into metrics + tracing components
- Add telemetry/config.go for unified telemetry configuration

Test infrastructure:
- Test tracing exporter for unit tests (util/telemetry/test_tracing_exporter.go)
- Package-level test helpers (workflow/tracing/test_helpers.go, etc.)
- Builder validation tests

### Verification

Most of the tests are in #15585 

### Documentation

Autogenerated docs are in this PR, but most of them are in #15583

### AI

Claude helped with splitting this from #15585

## Release Notes

* **New Features**
  * OpenTelemetry tracing integration for workflow execution, including startup, node lifecycle, artifact operations, and container runs.
  * Database query tracing support via otelsql integration.
  * Rate limiter tracing and instrumentation for client requests.
  * Configurable OTLP trace exporters (gRPC and HTTP).

* **Documentation**
  * New comprehensive tracing documentation with trace diagrams and span references.
  * Updated metrics documentation clarifying optional fields.

* **Examples**
  * Enhanced artifact-passing example demonstrating multi-artifact workflows.

PR #15576, opened 2026-02-17
refactor: thread DBType through DB session creation instead of detecting from driver

Joibel:
Part of #12077 

### Motivation

In order to inject tracing into the database drivers in a future commit, we lose the ability to perform type assertion to determine dbtype - so I've removed DBTypeFor and replaced it with explicit returns.

### Modifications

Rather than going for more complexity here, just do the simple thing and return DBType and thread it around

### Verification

This commit doesn't make functional changes. CI will test it.

### Documentation

Not required.

## Release Notes

* **Documentation**
  * Corrected grammar in configuration documentation for clarity.

* **Refactor**
  * Improved internal database handling implementation throughout the codebase.
  * Adjusted dependency structure for database drivers.

PR #15584, opened 2026-02-18
fix: deterministic span and trace ids

Joibel:
# Based on PR #15585

## Chain of upstream PRs as of 2026-02-18

* PR #15576:
  `main` ← `tracing-dbtype`

  * PR #15582:
    `tracing-dbtype` ← `tracing-infrastructure`

    * PR #15585:
      `tracing-infrastructure` ← `tracing-spans`

      * **PR #15584 (THIS ONE)**:
        `tracing-spans` ← `tracing-deterministic`

Part of tracing, but separated to reduce complexity of reviews

### Motivation

Some spans are stored in memory - but this will get wiped if the controller restarts. The spans are lost, and due to https://github.com/open-telemetry/semantic-conventions/issues/1648 there isn't really a way to do this without adding more complexity. Once otel supports long running spans we'll adopt their mechanism.

### Modifications

Renamed context package to wfcontext so as not to conflict with stdlib. Added CreationTimestamp to the context as this+name+namespace is unique. (Workflows can be recreated and then won't be unique).

This PR does not attempt to address retries in workflows which would be problematic for tracing after this change.

Added deterministic flag and implemented it, so that we have fixed span and trace IDs for a given workflow and node, even over a restart.

### Verification

Some new unit tests

CI checks

### Documentation

Implementation detail, not required.

## Summary by CodeRabbit

- **New Features**
  - Added deterministic trace and span identifiers derived from workflow, node, and phase metadata.
  - Telemetry now consistently reports deterministic identifiers and producer spans for node-related activity.
  - Expanded telemetry schema support for file key references and optional configuration.

- **Bug Fixes**
  - Improved trace continuity across workflow and node lifecycle events.
  - Added handling for missing or invalid metadata while preserving valid telemetry generation.


util/telemetry/builder/tracing.go
```diff
@@ -160,25 +142,61 @@ func (s *span) typeName() string {
...
 func generateStart(f io.Writer, s *span, attributes *attributesList) error {
 	methodName := fmt.Sprintf("Start%s", s.Name)
-
-	params := buildParameterList([]string{"ctx context.Context"}, &s.common, spanType, attributes)
-	methodCall := fmt.Sprintf("t.tracer.Start(ctx, \"%s\"", toLowerCamelCase(s.Name)) // TODO best practice
-
-	// Convert parents to their runtime (lowerCamelCase) names
-	var runtimeParents []string
-	for _, p := range s.Parents {
-		runtimeParents = append(runtimeParents, toLowerCamelCase(p))
```
isubasinghe, line 170: Claude reckons this can be better templated. 

```go
type startFuncData struct {
	MethodName        string
	DisplayName       string
	Params            string
	RuntimeParents    []string
	JoinedParents     string
	ParentCondition   string
	DeterministicRoot bool
	Deterministic     bool
	WithAttributes    string
	ReturnCall        string
}

var startFuncTmpl = template.Must(template.New("startFunc").Parse(`// {{.MethodName}} starts a {{.DisplayName}} span
func (t *Tracing) {{.MethodName}}({{.Params}}) (context.Context, trace.Span) {
parent := trace.SpanFromContext(ctx)
if roParent, ok := parent.(sdktrace.ReadOnlySpan); ok {
parentName := roParent.Name()
{{if .RuntimeParents -}}
if {{.ParentCondition}} {
// TODO .WithFatal()
logging.RequireLoggerFromContext(ctx).WithFields(logging.Fields{"startMethod": {{printf "%q" .MethodName}}, "expectedParents": {{printf "%q" .JoinedParents}}, "actualParent": parentName}).Error(ctx, "incorrect trace parentage")
}
{{- else -}}
logging.RequireLoggerFromContext(ctx).WithFields(logging.Fields{"startMethod": {{printf "%q" .MethodName}}, "actualParent": parentName}).Info(ctx, "trace parent") // TODO remove
{{- end}}
}
{{if .DeterministicRoot}}ctx = WithTraceID(ctx, traceID)
{{end -}}
{{if .Deterministic}}ctx = WithSpanID(ctx, spanID)
{{end -}}
{{.WithAttributes}}return {{.ReturnCall}}
}

`))

func generateStart(f io.Writer, s *span, attributes *attributesList) error {
	methodName := fmt.Sprintf("Start%s", s.Name)

	baseParams := []string{"ctx context.Context"}
	if s.Deterministic {
		if s.Root {
			baseParams = append(baseParams, "traceID trace.TraceID")
		}
		baseParams = append(baseParams, "spanID trace.SpanID")
	}
	params := buildParameterList(baseParams, &s.common, spanType, attributes)

	runtimeParents := make([]string, len(s.Parents))
	for i, p := range s.Parents {
		runtimeParents[i] = toLowerCamelCase(p)
	}

	conds := make([]string, len(runtimeParents))
	for i, p := range runtimeParents {
		conds[i] = fmt.Sprintf("parentName != %q", p)
	}

	methodCall := fmt.Sprintf("t.tracer.Start(ctx, %q", toLowerCamelCase(s.Name))

	return startFuncTmpl.Execute(f, startFuncData{
		MethodName:        methodName,
		DisplayName:       s.displayName(),
		Params:            strings.Join(params, ", "),
		RuntimeParents:    runtimeParents,
		JoinedParents:     strings.Join(runtimeParents, ", "),
		ParentCondition:   strings.Join(conds, " && "),
		DeterministicRoot: s.Deterministic && s.Root,
		Deterministic:     s.Deterministic,
		WithAttributes:    buildTracingWithAttributes(&s.Attributes, attributes),
		ReturnCall:        buildTracingCall(methodCall, &s.common, spanKind(s.Kind)),
	})
}
```

util/telemetry/idgen.go
```diff
@@ -0,0 +1,105 @@
...
+	if override, ok := ctx.Value(traceIDOverrideKey{}).(*idOverride[trace.TraceID]); ok {
+		if id, taken := override.take(); taken {
+			tid = id
+		}
+	}
+	if !tid.IsValid() {
+		for {
+			binary.NativeEndian.PutUint64(tid[:8], rand.Uint64())
+			binary.NativeEndian.PutUint64(tid[8:], rand.Uint64())
+			if tid.IsValid() {
```
isubasinghe, line 71: I kinda don't like this?

TraceID has a pre-determined size doesn't it? 
It doesn't need the loop? It just needs insurance that rand.Uint64() > 0 and thats  easy to do (+ 1).

util/telemetry/idgen.go
```diff
@@ -0,0 +1,105 @@
...
+		if id, taken := override.take(); taken {
+			return id
+		}
+	}
+	sid := trace.SpanID{}
+	for {
+		binary.NativeEndian.PutUint64(sid[:], rand.Uint64())
+		if sid.IsValid() {
+			break
+		}
```
isubasinghe, line 102: Same here, looping is unnecessarily complex.

isubasinghe (changes requested):

util/telemetry/idgen.go
Joibel: This is a complete and deliberate copy of https://github.com/open-telemetry/opentelemetry-go/blob/911ed35fa0ed380b61c888a15ecf161f14b907a4/sdk/trace/id_generator.go

util/telemetry/idgen.go
Joibel: Same here, I did this the same as the default generator which is sadly not exported so that we didn't re-invent the wheel

util/telemetry/idgen.go
isubasinghe: I am not a huge fan of that, is there any reason we need to copy them? 
```go
package main

import (
    "crypto/rand"
    "fmt"
)

func main() {
    b := make([]byte, 16)
    _, err := rand.Read(b)
    if err != nil {
        panic(err)
    }
    fmt.Printf("%x\n", b) // hex string
}
```

This will be a more secure/ less collision likely way of generating random arrays of data.

util/telemetry/builder/tracing.go
Joibel: Done

Joibel: I thought the template got too complex to be readable, but it's fine.

util/telemetry/idgen.go
isubasinghe: Hmm youd still need to account for the case where zero array is generated though. I guess or the last byte with 1.

isubasinghe: But maybe its fine as it is.

util/telemetry/builder/tracing.go
isubasinghe: Feel free to revert if you feel the template is too difficult to parse.

isubasinghe (approved):

PR #15615, opened 2026-02-24
fix(lint): enable shadow linter and fix issues

Joibel:
### Motivation

#15609 would have been caught if we'd had a shadow linter on. (This linter does catch that issue)

### Modifications

Enabled all of `govet` linter except for `fieldalignment`. Fieldalignment isn't bad per-se, but the --fix rewrites any structs without comments, and this is bad. So do that another time.

The only issue these found (as many of them are duplicated in other linters) is one `unusedwrite` which I've flagged in the comments.

There are 3 commits in this PR. The first is all about the variable called `err` being shadowed, and is mostly AI. It isn't all the `err` variables.

The last commit has my hand done shadow fixes, where I've made a decision about whether shadowing was actually unintended.

### Verification

CI only

### Documentation

Not required, code only

### DIFF
--- a/cmd/workflow-controller/main.go
+++ b/cmd/workflow-controller/main.go
@@ -125,6 +125,9 @@ func NewRootCommand() *cobra.Command {
 				if err := wfController.ShutdownTracing(context.WithoutCancel(ctx)); err != nil {
 					log.WithError(err).Error(ctx, "Failed to shutdown tracing")
 				}
+				if err := wfController.ShutdownMetrics(context.WithoutCancel(ctx)); err != nil {
+					log.WithError(err).Error(ctx, "Failed to shutdown metrics")
+				}
 			}()
 
 			leaderElectionOff := os.Getenv("LEADER_ELECTION_DISABLE")
--- a/util/telemetry/metrics.go
+++ b/util/telemetry/metrics.go
@@ -20,6 +20,7 @@ import (
 type Metrics struct {
 	otelMeter *metric.Meter
 	config    *MetricsConfig
+	provider  *metricsdk.MeterProvider
 
 	// Ensures mutual exclusion in instruments
 	mutex       sync.RWMutex
@@ -49,6 +50,14 @@ func (m *Metrics) IterateROInstruments(fn func(i *Instrument)) {
 	for _, i := range m.instruments {
 		fn(i)
 	}
+}
+
+// Shutdown flushes any remaining metrics and shuts down the meter provider.
+func (m *Metrics) Shutdown(ctx context.Context) error {
+	if m.provider != nil {
+		return m.provider.Shutdown(ctx)
+	}
+	return nil
 }
 
 func NewMetrics(ctx context.Context, serviceName, prometheusName string, config *MetricsConfig, extraOpts ...metricsdk.Option) (*Metrics, error) {
@@ -116,6 +125,7 @@ func NewMetrics(ctx context.Context, serviceName, prometheusName string, config 
 	metrics := &Metrics{
 		otelMeter:   &meter,
 		config:      config,
+		provider:    provider,
 		instruments: make(map[string]*Instrument),
 	}
 
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -301,6 +301,11 @@ func (wfc *WorkflowController) ShutdownTracing(ctx context.Context) error {
 	return wfc.tracing.Shutdown(ctx)
 }
 
+// ShutdownMetrics flushes any remaining metrics and shuts down the meter provider.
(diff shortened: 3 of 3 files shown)


cmd/argo/commands/clustertemplate/create.go
```diff
@@ -17,27 +17,27 @@ type cliCreateOpts struct {
 }
 
 func NewCreateCommand() *cobra.Command {
-	var cliCreateOpts = cliCreateOpts{output: common.NewPrintWorkflowOutputValue("")}
+	var opts = cliCreateOpts{output: common.NewPrintWorkflowOutputValue("")}
```
Joibel, line 20: This and all similar are the struct and variable name matching. I'm sure there is another linter that's more aggressive about this which we haven't enabled yet, but that's why all these are renamed.

cmd/argoexec/commands/emissary.go
```diff
@@ -132,7 +132,7 @@ func NewEmissaryCommand() *cobra.Command {
 								}
 							default:
 								data, _ := os.ReadFile(filepath.Clean(varRunArgo + "/ctr/" + y + "/exitcode"))
-								exitCode, err := strconv.Atoi(string(data))
+								exitCode, err = strconv.Atoi(string(data))
```
Joibel, line 161: This feels like it's possibly an actual bug, but for non-zero exit codes we return anyway.

util/template/simple_template.go
```diff
@@ -18,12 +18,12 @@ func simpleReplace(ctx context.Context, w io.Writer, tag string, replaceMap map[
 		if index := strings.LastIndex(tag, "{{"); index > 0 {
 			nestedTagPrefix := tag[:index]
 			nestedTag := tag[index+2:]
-			if replacement, ok := replaceMap[nestedTag]; ok {
-				replacement, isStr := replacement.(string)
+			if replNested, replOk := replaceMap[nestedTag]; replOk {
```
Joibel, line 21: This one is weird, but I don't think there's an actual bug happening here.

workflow/artifacts/git/git.go
```diff
@@ -127,7 +127,8 @@ func (g *ArtifactDriver) Load(ctx context.Context, inputArtifact *wfv1.Artifact,
 	r, err := git.PlainClone(path, false, cloneOptions)
 	if errors.Is(err, transport.ErrEmptyRemoteRepository) {
 		logging.RequireLoggerFromContext(ctx).Info(ctx, "Cloned an empty repository")
-		r, initErr := git.PlainInit(path, false)
+		var initErr error
```
Joibel, line 130: This doesn't change behaviour, as we never leave this if block as there's an unconditional return at the end of it.

workflow/controller/artifact_gc.go
```diff
@@ -158,17 +158,17 @@ func (woc *wfOperationCtx) processArtifactGCStrategy(ctx context.Context, strate
 	podNames := make(map[string]podInfo)
 
 	var podName string
-	var podInfo podInfo
+	var info podInfo
```
Joibel, line 161: Another case of struct name shadowing

workflow/controller/operator.go
```diff
@@ -3783,10 +3781,6 @@ func (woc *wfOperationCtx) executeSuspend(ctx context.Context, nodeName string,
 	var requeueTime *time.Time
 
 	if tmpl.Suspend.Duration != "" {
-		node, err := woc.wf.GetNodeByName(nodeName)
```
Joibel, line 3853: This appears unnecessary - `node` should already contain this node. I'm hoping CI will confirm or deny this.

workflow/controller/steps.go
```diff
@@ -177,10 +177,6 @@ func (woc *wfOperationCtx) executeSteps(ctx context.Context, nodeName string, tm
 	}
 
 	if outputs != nil {
-		node, err := woc.wf.GetNodeByName(nodeName)
```
Joibel, line 178: Another unnecessary reget of node I think

workflow/sync/sync_manager.go
```diff
@@ -219,15 +219,15 @@ func (sm *Manager) Initialize(ctx context.Context, wfs []wfv1.Workflow) {
 
 		if wf.Status.Synchronization.Semaphore != nil {
 			for _, holding := range wf.Status.Synchronization.Semaphore.Holding {
-				semaphore := sm.syncLockMap[holding.Semaphore]
-				if semaphore == nil {
+				sem := sm.syncLockMap[holding.Semaphore]
```
Joibel, line 222: semaphore is a struct name

workflow/controller/steps.go
Joibel: I was wrong for this one, it can get mutated so this reget is needed

PR #15622, opened 2026-02-24
fix: go module versioning

isubasinghe:
### Motivation

Go versioning is broken on 4.x

### Modifications

Update all references to v4

### Verification

Tests work as usual

### Documentation

None

### DIFF
--- a/sdks/go/alternate_auth/go.mod
+++ b/sdks/go/alternate_auth/go.mod
@@ -5,7 +5,7 @@ go 1.25.6
 replace github.com/argoproj/argo-workflows/v4 => ../../..
 
 require (
-	github.com/argoproj/argo-workflows/v4 v3.0.0-00010101000000-000000000000
+	github.com/argoproj/argo-workflows/v4 v4.0.0-00010101000000-000000000000
 	k8s.io/apimachinery v0.33.1
 	k8s.io/client-go v0.33.1
 )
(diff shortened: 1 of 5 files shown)


server/auth/sso/clients_test.go
```diff
@@ -105,6 +105,7 @@ func TestCreateHTTPClient_DefaultConfig(t *testing.T) {
 
 	if client == nil {
 		t.Fatal("createHTTPClient() returned nil client")
+		return
```
Joibel, line 108: Why these?

pkg/apiclient/workflowtemplate/workflow-template.proto
```diff
@@ -1,9 +1,9 @@
 syntax = "proto3";
-option go_package = "github.com/argoproj/argo-workflows/pkg/apiclient/workflowtemplate";
+option go_package = "github.com/argoproj/argo-workflows/v4/pkg/apiclient/workflowtemplate";
```
Joibel, line 2: Ok, interesting.

server/auth/sso/clients_test.go
isubasinghe: The linter started complaining for some reason

pkg/apiclient/workflowtemplate/workflow-template.proto
isubasinghe: It's also not really needed.

server/auth/sso/clients_test.go
isubasinghe: staticheck to be precise.

PR #15442, opened 2026-01-29
fix: requeue workflow if expected variables are missing. Fixes #15513

isubasinghe:
Fixes #15513

### Motivation

This PR forces workflows to be requeue-ed when they are missing the variables referenced in their expression. 
This ensures we do not make invalid progress on expressions. 

### Modifications

* Modified `hasVarInEnv` to handle more cases (previously it couldn't handle `time.Time` due to being unable to flatten the struct thanks to its unexported fields which resulted in a panic). 
* Lex and check for variables, if there are missing we requeue
* added some code to allow for relaxations of strictness when placeholders are present. 

### Verification

Added some tests along with fuzzing. 
We fuzz using the old definition of expressionReplace to ensure semantic equivalence. 

### Documentation

N/A

## Summary by CodeRabbit

* **New Features**
  * Added strict template replacement mode (ReplaceStrict) and a requeue signal for unresolved references.

* **Improvements**
  * Stricter variable validation for templates with configurable strict prefixes.
  * Workflow: dependency and step reference resolution now defers or requeues on missing variables; nested multi-pass variable resolution improved.
  * Placeholder prefix standardized for internal placeholders.
  
* **Tests**
  * Added extensive unit, comparison and fuzz tests covering template replacement, expressions, and placeholder behavior.

PR #15630, opened 2026-02-26
feat!: Short-circuit evaluation in enhanced depends logic. Fixes #15244

### LINKED ISSUE #15244: Short-circuit evaluation in enhanced depends logic
# Summary

Short-circuit evaluation should be implemented in the enhanced depends logic. The benefit is faster workflow execution.

## Use Cases

Here's a simple example where I'd like task `C` to start as soon as `A` completes, instead of waiting for `B` unnecessarily:
```
kind: Workflow
metadata:
  generateName: racing-
spec:
  serviceAccountName: argo-workflows-sa
  entrypoint: race

  templates:
    - name: race
      dag:
        tasks:
          - name: A
            template: foo

          - name: B
            template: foo
            arguments:
              parameters:
                - name: sleep
                  value: 20

          - name: C
            template: foo
            depends: "A || B"

    - name: foo
      inputs:
        parameters:
          - name: sleep
            value: 0
      container:
        image: alpine:3
        command: ["sh", "-c"]
        args:
          - |
            sleep "{{inputs.parameters.sleep}}"
            exit 0
```

---

prashanthjos:
Fixes #15244

### Motivation

The enhanced depends logic in DAG workflows currently requires all dependencies to be fulfilled before evaluating the depends expression for a task. This means that even when the final outcome of a boolean expression is already determined by a subset of completed dependencies, the workflow controller still waits for every dependency to finish. This leads to unnecessary delays in workflow execution.

For example, consider a task with `depends: "A.Succeeded || B.Succeeded"`. If task A has already succeeded, the expression will evaluate to true regardless of what task B does. However, the current implementation forces the workflow to wait for B to complete before it will evaluate the expression and allow the dependent task to proceed.

This becomes especially impactful in workflows with many parallel branches and complex dependency expressions, where entire chains of downstream tasks are blocked waiting for irrelevant upstream tasks to finish.

### Modifications

**Short-circuit evaluation in the depends logic evaluator:**

The depends logic evaluator has been reworked to support short-circuit evaluation. Previously, the function would immediately return "not ready" if any dependency was not yet fulfilled. Now, the function separates dependencies into two groups: resolved (completed and fulfilled) and pending (still running, not yet started, or waiting on hooks).

For resolved dependencies, their actual task results (Succeeded, Failed, Errored, Skipped, Omitted, Daemoned, AnySucceeded, AllFailed) are collected as before. When there are no pending dependencies, the expression is evaluated normally with no change in behavior.

When pending dependencies exist, a two-evaluation boundary check is performed to determine whether the expression outcome is already fixed:

1. The depends expression is evaluated with all pending dependencies' result fields set to false (simulating the case where no pending dependency achieves any result).
2. The depends expression is evaluated again with all pending dependencies' result fields set to true (simulating the case where every pending dependency achieves every possible result).
3. If both evaluations produce the same boolean result, the outcome is determined regardless of what the pending dependencies eventually resolve to, and the function short-circuits immediately.
4. If the two evaluations disagree, the outcome genuinely depends on the pending dependencies, and the function correctly reports that it cannot yet determine the result.

This approach is sound because the boundary values (all-false and all-true) represent the extreme ends of all possible outcomes for pending dependencies. If both extremes agree, then every possible intermediate state must also agree.

**Graceful handling of missing dependency nodes:**

When short-circuit evaluation determines that a task should be executed or omitted early, some of its dependency tasks may not have created workflow nodes yet. Two downstream functions have been updated to handle this gracefully:

- The dependency connector, which wires up parent-child relationships in the DAG node graph, 
- now skips dependencies that do not have corresponding nodes instead of causing a nil pointer
  panic.
- The local scope builder, which collects outputs from ancestor tasks for template parameter resolution, now skips missing ancestor nodes instead of returning an internal error. This is safe because a task that was short-circuited should not reference outputs from a dependency that never ran.

**New helper function:**

A small utility function was added to deep-copy the evaluation scope map, which is needed to create the two boundary evaluation scopes without mutating the original resolved results.

### Verification

A new test suite was added with 7 subtests that thoroughly exercise the short-circuit evaluation logic:

- **OR short-circuit to execute**: Verifies that `A.Succeeded || B.Succeeded` short-circuits to execute when A has succeeded, without waiting for B (which is still running).
- **AND short-circuit to omit**: Verifies that `A.Succeeded && B.Succeeded` short-circuits to omit when A has failed, without waiting for B (which is still running).
- **Indeterminate result**: Verifies that `A.Succeeded && B.Succeeded` correctly waits when A has succeeded but B is still running, since B's outcome affects the result.
- **Dependency not yet started**: Verifies that `A.Succeeded || B.Succeeded` short-circuits when A has succeeded and B has no node at all (not yet created).
- **All dependencies pending**: Verifies that `A.Succeeded || B.Succeeded` correctly waits when neither A nor B have started.
- **Negation prevents false short-circuit**: Verifies that `!A.Succeeded` with A still running correctly waits instead of prematurely short-circuiting, since A could either succeed or not.
- **Mixed result types**: Verifies that `A.Succeeded || B.Failed` short-circuits when A has succeeded, regardless of B's status.

All existing unit tests continue to pass with no regressions, confirming that the change is fully backward-compatible and only affects the timing of evaluation, not the final outcome.

### Documentation
No documentation changes are required. This is a transparent performance optimization to the existing enhanced depends logic. The syntax and semantics of depends expressions remain unchanged, tasks will reach the same final state, but may do so faster when short-circuit evaluation applies.


Joibel:
If C has inputs from both a A and B then a workflow which currently works will break with this change.

prashanthjos:
> If C has inputs from both a A and B then a workflow which currently works will break with this change.

@Joibel This is a great point to raise but I see the existing code already handles this case safely.

When the short-circuit determines that a task should execute, the next step in `executeDAGTask` is `resolveDependencyReferences`, which calls buildLocalScopeFromTask to collect outputs from ancestor tasks and then uses `ReplaceStrict` to resolve all `{{tasks.*}}` references. If a referenced task's outputs aren't available (because that task hasn't run yet), `ReplaceStrict` returns a `MissingVariableErr`, which triggers a requeue, the task simply returns without executing and gets retried on the next reconciliation loop:

```
newTaskStr, err := template.ReplaceStrict(ctx, string(taskBytes), woc.globalParams.Merge(scope.getParameters()), []string{"tasks", "steps"})
if err != nil {
    if template.IsMissingVariableErr(err) {
        woc.requeue()
        return nil, ErrRequeue
    }
    return nil, err
}

```
So in the scenario where C has depends: "A.Succeeded || B.Succeeded" but references outputs from both A and B:

- A succeeds, B is still running --> short-circuit says "execute C"
- C tries to resolve {{tasks.B.outputs.parameters.foo}} --> B hasn't completed, so the variable is missing
- ReplaceStrict detects the missing variable --> requeues

On next reconciliation, if B has completed, C resolves successfully and executes. The workflow won't break, it will just requeue until the referenced outputs are available, which is the same mechanism already used for race conditions in the existing code (as noted in the comment above ReplaceStrict).

_That said, a workflow like this is arguably misconfigured: if C needs outputs from both A and B, the depends clause should be A.Succeeded && B.Succeeded, not A.Succeeded || B.Succeeded. The depends field is meant to declare the task's actual data dependencies._

prashanthjos:
@Joibel if the concern is breaking the backward compatibility we can also keep this functionality behind a flag with three options that I see 

**Option 1**

Add a field like shortCircuitDepends: true on the DAG template spec:

```
dag:
  shortCircuitDepends: true
  tasks:
    - name: A
      ...
    - name: C
      depends: "A.Succeeded || B.Succeeded"
```

**Option 2**

Add a flag in the workflow controller configmap (workflow-controller-configmap):

`shortCircuitDependsLogic: true`

**Option 3**

Use an annotation to enable it:

```
metadata:
  annotations:
    workflows.argoproj.io/short-circuit-depends: "true"
```

I would prefer option 1 or option 2, do let me know your thoughts.

workflow/controller/dag.go
```diff
@@ -939,9 +946,57 @@ func (d *dagContext) evaluateDependsLogic(ctx context.Context, taskName string)
 	}
 
 	evalLogic := strings.ReplaceAll(d.GetTaskDependsLogic(ctx, taskName), "-", "_")
-	execute, err := argoexpr.EvalBool(evalLogic, evalScope)
+
+	// If there are no pending dependencies, evaluate normally
+	if len(pendingDeps) == 0 {
```
Joibel, line 951: This block feels over complicated. We calculate the result of this first EvalBool below as `allFalseResult`, and it feels harder than necessary to follow and read:
Something like this (untested) is easier for my head to get around:
```
        // Evaluate with resolved deps only. Pending deps are absent from the                                                                                                                                                                  
        // scope, so expr returns zero-value TaskResults (all fields false) for                                                                                                                                                                
        // them — equivalent to the "all-false" boundary.                                                                                                                                                                                      
        result, err := argoexpr.EvalBool(evalLogic, evalScope)                                                                                                                                                                                 
        if err != nil {                                                                                                                                                                                                                        
                return false, false, fmt.Errorf("unable to evaluate expression '%s': %w", evalLogic, err)                                                                                                                                      
        }                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                               
        if len(pendingDeps) == 0 {                                                                                                                                                                                                             
                // All deps resolved, result is definitive.                                                                                                                                                                                    
                return result, true, nil                                                                                                                                                                                                       
        }                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                               
        // Pending deps exist. Check the "all-true" boundary: if it agrees with                                                                                                                                                                
        // the all-false result, the outcome is determined regardless of what the                                                                                                                                                              
        // pending deps resolve to.                                                                                                                                                                                                            
        allTrueScope := maps.Clone(evalScope)                                                                                                                                                                                                  
        for _, dep := range pendingDeps {                                                                                                                                                                                                      
                allTrueScope[dep] = TaskResults{                                                                                                                                                                                               
                        Succeeded: true, Failed: true, Errored: true,                                                                                                                                                                          
                        Skipped: true, Omitted: true, Daemoned: true,                                                                                                                                                                          
                        AnySucceeded: true, AllFailed: true,                                                                                                                                                                                   
                }                                                                                                                                                                                                                              
        }                                                                                                                                                                                                                                      
        allTrueResult, err := argoexpr.EvalBool(evalLogic, allTrueScope)                                                                                                                                                                       
        if err != nil {                                                                                                                                                                                                                        
                return false, false, fmt.Errorf("unable to evaluate expression '%s': %w", evalLogic, err)                                                                                                                                      
        }                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                               
        if result == allTrueResult {                                                                                                                                                                                                           
                return result, true, nil                                                                                                                                                                                                       
        }                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                
        return false, false, nil                
```

Joibel (changes requested):

Joibel:
> > If C has inputs from both a A and B then a workflow which currently works will break with this change.
> 
> When the short-circuit determines that a task should execute, the next step in `executeDAGTask` is `resolveDependencyReferences`, which calls buildLocalScopeFromTask to collect outputs from ancestor tasks and then uses `ReplaceStrict` to resolve all `{{tasks.*}}` references. If a referenced task's outputs aren't available (because that task hasn't run yet), `ReplaceStrict` returns a `MissingVariableErr`, which triggers a requeue, the task simply returns without executing and gets retried on the next reconciliation loop:

This behaviour was added after you raised this PR, so it wasn't considered at design time. This PR would still cause the controller to busy wait on missing results, rather than going back to sleep.

[Do not continue to reply to me with AI generated text](https://github.com/argoproj/argoproj/blob/main/community/genai.md).

I am happy to have this as the default behaviour as it does make sense mostly - I'm just worried about breaking things. So for that I'd probably like something like this at the task level, with the flag being absent == true:
```
dag:
  tasks:
    - name: A
      ...
    - name: C
      shortCircuitDepends: false
      depends: "A.Succeeded || B.Succeeded" 
```

Summoning @isubasinghe for his view on this as he wrote #15442

prashanthjos:
> This behaviour was added after you raised this PR, so it wasn't considered at design time. This PR would still cause the controller to busy wait on missing results, rather than going back to sleep.

I agree, I would wait for @isubasinghe to add more comments

prashanthjos:
@Joibel @caelan-io not sure if @isubasinghe is around to review, any thoughts on how we should move forward?

caelan-io:
Hey @prashanthjos - We have many other items to attend to, so Isitha will be reviewing as soon as there's capacity. We have already responded with an initial review, comments and Zoom call within a few days of you opening this PR, which is much better than the average CNCF open source project. 

If your review is urgent for your company, we suggest seeking out SLAs that match your business needs.

isubasinghe:
@prashanthjos thanks for the PR, my biggest issue here is how we handle variables in expressions that are optional. Currently, due to my PR, this aspect is a bit poorly handled in this PR.

prashanthjos:
@isubasinghe thank you for the comment, I am thinking the short-circuit evaluation is safe with respect to your `ReplaceStrict` (atleast that is what my eyes see :) )  changes because in the depends evaluation path, we always provide all variables (including pending deps) to EvalBool, pending deps get boundary values (all-false and all-true), so EvalBool never encounters a missing variable and, in the template resolution path, if a short-circuited task references outputs from a pending dep in its arguments, `ReplaceStrict` will correctly detect the missing variable and requeue, so a task can never execute with unresolved references.

I am still trying to  get my mind around your concern about "optional" variables correctly, are you referring to a case where a task in the depends expression might never create a node at all (e.g., behind a when clause or empty withParams), meaning `GetTaskDependencies` returns it but it's permanently absent from the eval scope? If so, could you share a specific scenario so I can add a test case and handle it properly?

prashanthjos:
@Joibel @caelan-io @isubasinghe any thoughts here?

PR #15649, opened 2026-03-03
feat: pause the running pod based on the existence of a file

guanguxiansheng:
### Motivation

If the environment variable `ARGO_DEBUG_PAUSE_AFTER` is not configured as `true` before the Pod is created, it is impossible to pause a running Pod. However, in actual production tasks, users often need to temporarily pause running Pods for on-site analysis. In these scenarios, the Pods do not have the `ARGO_DEBUG_PAUSE_AFTER` environment variable configured in advance, so an additional method for Pod pause needs to be provided.

### Modifications

Add a way to pause a running Pod: Before emissary exits, check if the file `/var/run/argo/ctr/{containerName}/after-pause` exists; if it exists, then pause the Pod.

### DIFF
--- a/.features/pending/pause-running-pod.md
+++ b/.features/pending/pause-running-pod.md
@@ -1,0 +1,11 @@
+Description: Pause the running pod after container exit based on file existence or environment variable
+Author: [guanguxiansheng](https://github.com/guanguxiansheng)
+Component: General
+Issues: 0
+
+When the main process in the container exits, the emissary can optionally pause instead of exiting immediately.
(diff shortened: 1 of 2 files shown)


shuangkun:
after-puase -> after-pause ?

shuangkun:
I think this feature is great; it helps with debugging.

Please fix the DCO and PR feature check.

guanguxiansheng:
Thank you, I have completed the revisions.

isubasinghe (changes requested):
This needs more documentation.
Can you please write an example workflow that demonstrates this in use? 
You may use container sets here to manage the creation and removal of the file.

PR #15665, opened 2026-03-06
feat: Add template name as attribute to node spans

### LINKED ISSUE #15768: Add `node` name to generated spans
# Summary

The new tracing functionality for workflows is really nice 🚀 it would be even more useful if one could tell from a node span which template it executed (and, prospectively, also with which parameters etc.).

## Use Cases

- Investigating traces for workflows

---

borchero:
Fixes https://github.com/argoproj/argo-workflows/issues/15768

### Motivation

The new tracing functionality for workflows is really nice 🚀 it would be even more useful if one could tell from a `node` span which template it executed (and, prospectively, also with which parameters etc.).

### Modifications

- Add `template_name` to the span attributes of the `node` spans

### Verification

n/a, seems pretty straightforward


chansuke:
https://github.com/argoproj/argo-workflows/actions/runs/22782853074/job/66092725512?pr=15665
Could you check the message?

borchero:
@chansuke I updated the existing pending feature related to tracing. I don't think this change warrants a dedicated feature addition.

Issue #15653, opened 2026-03-04
Argo Cron Workflow - CET timezone not triggering


data-han:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

Previously when my pipeline is set using CET timezone, it was able to trigger pipeline based on the time schedule. However, as of yesterday and tested today, this timzone seems to have issue triggering the pipelines, causing no pipeline to be run if it was set using this timezone

<img width="1412" height="74" alt="Image" src="https://github.com/user-attachments/assets/bab83b34-9cf0-4c34-961b-e2df16324803" />

### Version(s)

v3.7.10

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

```YAML
spec:
  concurrencyPolicy: "Forbid"
  schedules:
    - "00 18 * * *" # US-run - 18:00 CET = 12AM SGT
  timezone: "CET"
```

### Logs from the workflow controller

```text
kubectl logs -n argo deploy/workflow-controller | grep ${workflow}
```

### Logs from in your workflow's wait container

```text
kubectl logs -n argo -c wait -l workflows.argoproj.io/workflow=${workflow},workflow.argoproj.io/phase!=Succeeded
```

thevilledev:
It looks like the workflow-controller `v3.7.10` image is lacking "CET" from the timezone database. But it's not the only image, as can be seen here:

```bash
for v in v3.6.0 v3.7.10 v3.7.11 v4.0.1 v4.0.2; 
do
  name="tz-$$-$v";
  docker create --name "$name" "quay.io/argoproj/workflow-controller:$v" 2>/dev/null;
  has=$(docker export "$name" 2>/dev/null | tar -tf - 2>/dev/null | grep -c "^usr/share/zoneinfo/CET$");
  echo "$v: CET $([ "$has" -gt 0 ] && echo "PRESENT" || echo "MISSING")";
  docker rm "$name" >/dev/null 2>&1;
done
```

Output:

```
[...]

PR #15741, opened 2026-03-13
chore(deps): pin distroless base to debian13

### LINKED ISSUE #15733: Pin distroless base images to Debian release track
### Pre-requisites

Relates to #15653 and #15732.

The Dockerfile uses `gcr.io/distroless/static` without a Debian version suffix or digest pin. This tag is a moving pointer that silently switches between Debian releases when distroless rebuilds.

This already caused an issue: when distroless rebuilt on Debian 13 (Trixie), the `tzdata-legacy` package was not included, stripping backward-compatible timezone entries like CET. The Argo Workflows v3.7.10 and v4.0.1 releases were built against this snapshot, breaking CronWorkflows that used timezone abbreviations. The v3.7.11 and v4.0.2 releases happened to be built after the upstream fix.

The distroless maintainers recommend against using unversioned tags:
> "We generally don't recommend using the non-debian version suffixed image references for this exact reason." (https://github.com/GoogleContainerTools/distroless/issues/1996#issuecomment-3854372767)

### Proposal

Pin `Dockerfile` to use Debian 13 release track for all three containers (argoexec-base, workflow-controller, argocli):

```
FROM gcr.io/distroless/static-debian13@sha256:abc123... AS workflow-controller
```

With a Renovate rule to auto-bump.

### Version(s)

v3.7.10, v4.0.1 

```YAML
cat <<'EOF' | kubectl apply -n argo -f - 2>&1
kind: CronWorkflow
metadata:
  name: test-cet-timezone
spec:
  schedules:
    - "*/1 * * * *"
  timezone: "CET"
  workflowSpec:
    entrypoint: whalesay
    templates:
      - name: whalesay
        container:
          image: busybox
          command: [echo, "hello CET"]
EOF
```

```text
error="provided bad location CET: unknown time zone CET"
```

```text
```

thevilledev:
Fixes #15733.

### Motivation

The Dockerfile uses `gcr.io/distroless/static` without a Debian version suffix or digest pin. This tag is a moving pointer that silently switches between Debian releases when distroless rebuilds. This already caused a production issue: the v3.7.10 and v4.0.1 images were built against a distroless snapshot where the Debian `tzdata-legacy` package was missing, stripping backward-compatible timezone entries like CET (#15653).

The distroless maintainers recommend against using unversioned tags:
> "We generally don't recommend using the non-debian version suffixed image references for this exact reason." — distroless#1996 (https://github.com/GoogleContainerTools/distroless/issues/1996#issuecomment-3854372767)

### Modifications

Pin all three `FROM gcr.io/distroless/static` references (argoexec-base, workflow-controller, argocli) to `gcr.io/distroless/static-debian13:latest@sha256:...`. The `:latest` tag documents intent, the `@sha256:` digest locks the exact image. Debian 13 is used because the unpinned `static:latest` already resolves to it — this pins what we already use, not a change in base OS. Previously the transition from Debian 12 to Debian 13 happened without the project knowing about it.

### Verification

Image digest resolved through:

```bash
docker buildx imagetools inspect gcr.io/distroless/static-debian13:latest
```

Built all three `Dockerfile` targets locally with the pinned base image and confirmed they produce working images. Verified `static:latest` and `static-debian13:latest` resolve to the same image (6.36 MB, identical digest).

### Documentation

Not needed as this is not a user-facing change.


Joibel:
I like this, but by itself it wouldn't have prevented the TZ problem as we wouldn't have noticed it. Still very much worth doing.

Joibel (approved):

PR #15743, opened 2026-03-15
fix(validate): match internal placeholder prefix

thevilledev:
### Motivation

PR #14850 added a `strings.Contains(tmplRef.Template, "placeholder")` check to skip validation for dynamic template references (e.g. `template: "{{ inputs.parameters.message }}`"). This match is too broad and might clash with user-supplied template names like `run-placeholder-task`, silently skipping validation errors.

### Modifications

Match the full internal prefix `__argo__internal__placeholder-` instead of the bare word, correctly handling both full and partial dynamic template names. The placeholder prefix is exported from the `template` package now.

Added regression test.

### Verification

In addition to the unit test, verified buggy behaviour on a live kind cluster. Created a WorkflowTemplate `test-wft` with one template `real-template`, then submitted two workflows referencing non-existent templates:

- if the template name is `nonexistent-template`: invalid spec, not found
- if the template name is `run-placeholder-task`: step group deemed errored

With the fix, both fail the same way:

```
Failed: invalid spec: templates.main.steps[0].step1 template reference test-wft.run-placeholder-task not found
```
### Documentation

No documentation changes needed. Internal validation logic only.


Joibel:
Thanks for this

PR #15745, opened 2026-03-16
fix: fix pagination when archive is disabled. Fixes #15639

### LINKED ISSUE #15639: In UI the pagination of Workflows is broken when archive feature is not enabled
### Pre-requisites

Testing the new v4.0.1 we immediately noticed something wrong in the UI.
In the Workflows page the pagination didn't work anymore. For example selecting max 5 entries, the UI listed 5 entries (out of more than 20) but I was not able to click on the button `Next page`.

In this screenshot the UI shows the button in grey, but also the annoying warning "Workflows cannot be globally sorted when paginated"
Image

But the output of the command `kubectl get workflows` was showing more than 20 results
Image

I was expecting to see 5 Workflows, but also to be able to click on "Next page" to see the other ones.

This bug disappears when I enabled the `archive` feature.
I suppose there is an issue with [179571ee1](https://github.com/argoproj/argo-workflows/commit/179571ee154c55c831b2d74e685420072cdea1ea) feat(server): optimize pagination when counting workflows in archive. Fixes:#13948 (#14892)

Here the workflow-controller logs
[workflows-controller_v4_test-B.txt](https://github.com/user-attachments/files/25614462/workflows-controller_v4_test-B.txt)

Thanks a lot in advance for any help!

### Version(s)

v4.0.1

```YAML
kind: Workflow
metadata:
  generateName: cowsay-
  namespace: argo
  annotations:
    workflows.argoproj.io/title: Cowsay
    workflows.argoproj.io/description: This is a dummy pipeline.
spec:
  serviceAccountName: cicd-pipeline
  entrypoint: main
  templates:
    - name: main
      dag:
        tasks:
          - name: A
            template: cowsay
          - name: B
            template: cowsay
            depends: "A"
          - name: C
            template: cowsay
            depends: "A"
          - name: D
            template: cowsay
            depends: "B && C"
    - name: cowsay
      container:
        image: grycap/cowsay:latest
        command: [/usr/games/cowsay]
        args: ["hello world"]
```

```text
$ kubectl logs -n argo deploy/workflow-controller

See file in the description
```

```text
not an issue of a single specific workflow
```

himeshp:
Fixes #15639

### Motivation

When the archive feature is disabled, `ListWorkflows` determined whether a next page existed by calling `s.wfArchive.HasMoreWorkflows()`. However `NullWorkflowArchive.HasMoreWorkflows()` always returns `false` — it has no knowledge of live workflows. This meant `meta.Continue` was never set, disabling the "Next page" button in the UI even when there were more live workflows than the page size.

### Modifications

- `server/workflow/workflow_server.go`: In the `!ShowRemainingItemCount` branch of `ListWorkflows`, replaced the single `HasMoreWorkflows` call with a two-step check:
  1. First checks if there are more live workflows arithmetically (`offset+limit < liveWfCount`) — no DB query needed
  2. Only queries the archive if live workflows are exhausted, using the same `archivedOffset`/`archivedLimit` arithmetic as the actual archive fetch (lines 244–250), so the probe lands at the correct position in the archive
  3. Skips the archive check entirely when `limit=0` (fetch-all, no pagination)

- `server/workflow/workflow_server_test.go`: Added two regression tests:
  - `TestListWorkflowPaginationWithoutArchive` — reproduces the reported bug: archive disabled, 3 live workflows, limit=2 → Next page must be available
  - `TestListWorkflowPaginationMixedLiveAndArchive` — covers the live/archive boundary: 3 live + 4 archived, limit=5 → page 1 returns 5 items with Continue set

### Verification

- `go test ./server/workflow/...` — all tests pass including the two new regression tests
- Manually verified with archive disabled and 21 live workflows: `GET /api/v1/workflows/argo?listOptions.limit=5` now returns `meta.continue="5"` and the "Next page" button is clickable in the UI

Screenshot 2026-03-16 at 12 08 12 PM

### Documentation

No documentation changes needed — this is a bug fix restoring expected pagination behaviour.

## Release Notes

* **Bug Fixes**
  * Improved pagination accuracy for workflow lists by fixing calculation of remaining items when displaying both active and archived workflows in a single list view.

* **Tests**
  * Added pagination tests for workflow listings to ensure correct behavior when pagination spans boundaries between active and archived workflow collections.


isubasinghe:
@coderabbitai review

PR #15748, opened 2026-03-16
chore: modify the tooltip title for the nav-bar

nakatani-yo:
### Motivation

The tooltip for the user menu item in the navigation bar simply said "User", which was vague. Changing it to "User Info" better describes the destination and improves clarity for users.

### Modifications

Changed the tooltip title from User to User Info in ui/src/app-router.tsx

### Documentation

No documentation changes required


Joibel:
Simple enough improvement, thank you

PR #15754, opened 2026-03-18
fix: fixed invalid JSON in template substitution, fixes #15512

### LINKED ISSUE #15512: Produce invalid json when template substitution due to special character
### Pre-requisites

When I was creating a workflow, I encountered an invalid json error.
```
fail to create workflow with namespace(canal-flow) and name(null), reason: {"code":3,"message":"templates.main cannot finish template replacement because the result was invalid JSON"}
```

After investigating the codebase, I found that the problem might be occurring during parameter value replacement in `executeTemplate`.
Currently, the parameter value is directly written into the buffer.
In our case, the parameter value contains special characters. (`🏴󠁧󠁢󠁳󠁣󠁴󠁿`)

```
// SubstituteParams returns a new copy of the template with global, pod, and input parameters substituted
func SubstituteParams(tmpl *wfv1.Template, globalParams, localParams Parameters) (*wfv1.Template, error) {
	tmplBytes, err := json.Marshal(tmpl)
	if err != nil {
		return nil, errors.InternalWrapError(err)
	}
	// First replace globals & locals, then replace inputs because globals could be referenced in the inputs
	replaceMap := globalParams.Merge(localParams)
	globalReplacedTmplStr, err := template.Replace(string(tmplBytes), replaceMap, true)
	if err != nil {
		return nil, err
	}
	var globalReplacedTmpl wfv1.Template
	err = json.Unmarshal([]byte(globalReplacedTmplStr), &globalReplacedTmpl)
	if err != nil {
		return nil, errors.InternalWrapError(err)
	}
```
[Reference](https://github.com/argoproj/argo-workflows/blob/741ab0ef7b6432925e49882cb4294adccf5912ec/workflow/common/util.go#L240-L255)

```
func simpleReplace(w io.Writer, tag string, replaceMap map[string]interface{}, allowUnresolved bool) (int, error) {
					...
					replacement = strconv.Quote(replacement)
					replacement = replacement[1 : len(replacement)-1]
					return w.Write([]byte("{{" + nestedTagPrefix + replacement))
					...
}
```
[Reference](https://github.com/argoproj/argo-workflows/blob/741ab0ef7b6432925e49882cb4294adccf5912ec/util/template/simple_template.go#L14-L26)

However, when the value contains special characters, golang writes a string of special characters in Unicode, making the replaced template json invalid.
e.g. `🏴󠁧󠁢󠁳󠁣󠁴󠁿`
```
{ "name": "test-config", "value": "\U0001f1e7"}
```
[Reproduce in golang playground](https://go.dev/play/p/hh-Q41BESbd?v=goprev)

---
I think a safer approach would be to use json marshal before writing it to ensure that the value to be replaced is a valid json string.
This approach successfully solves the issue in our own customized argo-workflows.
```
func simpleReplace(w io.Writer, tag string, replaceMap map[string]interface{}, allowUnresolved bool) (int, error) {
	replacement, ok := replaceMap[strings.TrimSpace(tag)]
	if !ok {
		// Attempt to resolve nested tags, if possible
		if index := strings.LastIndex(tag, "{{"); index > 0 {
			nestedTagPrefix := tag[:index]
			nestedTag := tag[index+2:]
			if replacement, ok := replaceMap[nestedTag]; ok {
				replacement, isStr := replacement.(string)
				if isStr {
					escaped, err :=  json.Marshal(replacement)   <-- here
					if err != nil {
						return 0, err
					}
					replacement = escaped[1 : len(escaped)-1]
					return w.Write([]byte("{{" + nestedTagPrefix + replacement))
				}
			}
```

### Version(s)

v3.6.2

```YAML
kind: Workflow
metadata:
  generateName: special-char-
spec:
  entrypoint: main
  arguments:
    parameters:
    - name: flag
      value: "🏴󠁧󠁢󠁳󠁣󠁴󠁿"
  templates:
  - name: main
    inputs:
      parameters:
      - name: flag
    container:
      image: alpine:3.19
      command: [sh, -c]
      args:
      - echo "{{inputs.parameters.flag}}"
```

```text
N/A
```

```text
N/A
```

nebojsaj1726:
Fixes #15512

### Motivation

Workflow template substitution fails with "invalid JSON" error when parameter values contain special Unicode characters (emoji, flags, etc.).

### Modifications

Implemented the fix suggested in the issue:
- Replaced `strconv.Quote` with `json.Marshal` in `util/template/simple_template.go`

Added unit tests to cover emoji and special Unicode characters.

### Verification
- New unit tests pass
- All existing template tests pass
- Manually tested with workflow YAML from the issue

### Documentation
No documentation changes needed - this is a bug fix.

### DIFF
--- a/util/template/simple_template_test.go
+++ b/util/template/simple_template_test.go
@@ -1,0 +1,73 @@
+package template
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/argoproj/argo-workflows/v4/util/logging"
+)
+
+func TestReplaceWithEmoji(t *testing.T) {
+	ctx := logging.TestContext(t.Context())
+
+	replaceMap := map[string]string{
+		"inputs.parameters.flag": "🏴󠁧󠁢󠁳󠁣󠁴󠁿",
+	}
+
+	test := toJSONString(`{{inputs.parameters.flag}}`)
+	replacement, err := Replace(ctx, test, replaceMap, false)
+
+	require.NoError(t, err, "Should not error on emoji substitution")
+	assert.Equal(t, toJSONString("🏴󠁧󠁢󠁳󠁣󠁴󠁿"), replacement, "Should preserve emoji character")
+}
+
+func TestReplaceWithSpecialCharacters(t *testing.T) {
+	ctx := logging.TestContext(t.Context())
+
+	testCases := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "Emoji flag",
+			input:    "🏴󠁧󠁢󠁳󠁣󠁴󠁿",
+			expected: "🏴󠁧󠁢󠁳󠁣󠁴󠁿",
+		},
+		{
+			name:     "Unicode emoji",
+			input:    "Hello 👋 World",
+			expected: "Hello 👋 World",
+		},
+		{
+			name:     "Chinese characters",
+			input:    "你好世界",
+			expected: "你好世界",
+		},
+		{
+			name:     "Arabic characters",
+			input:    "مرحبا بالعالم",
+			expected: "مرحبا بالعالم",
+		},
+		{
+			name:     "Mixed content",
+			input:    "Test 测试 🚀",
+			expected: "Test 测试 🚀",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			replaceMap := map[string]string{
+				"inputs.parameters.value": tc.input,
+			}
+			test := toJSONString(`{{inputs.parameters.value}}`)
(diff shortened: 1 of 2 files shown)


util/template/simple_template_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/argoproj/argo-workflows/v4/util/logging"
+)
+
+func TestReplaceWithEmoji(t *testing.T) {
```
Joibel, line 12: This appears to be a one off test of something that's also tested in the next test?

nebojsaj1726: The idea was for TestReplaceWithEmoji to be a focused test for this particular issue, and TestReplaceWithSpecialCharacters to be a broader test that covers various characters.

But happy to remove it and keep just the other one, if you prefer.

PR #15773, opened 2026-03-19
chore: tell gotestsum to re-run root test on failure

MasonM:
### Motivation

[gotestsum](https://github.com/gotestyourself/gotestsum/tree/06f60b3249917ddd1a04c8e2586116d1a87bc67c) is configured to re-run failed tests, but it only re-runs the specific subtest that fails. This is a problem for tests that are part of a suite, where subtests depend on earlier subtests.

For example, `ArgoServerSuite/TestWorkflowServiceListArchived/ListAll` test is currently flaky, but the re-runs don't help, because the `CreateArchivedBobWf` and `CreateAlice` subtests need to run beforehand, otherwise the `nameBobWf` and `nameAliceWf` variables will be empty and you'll see a bunch of errors like `Waiting up to 1m30s for workflow with field selector 'metadata.name='`
https://github.com/argoproj/argo-workflows/blob/7824aa613546b8e7cb7d406d79768947805ff802/test/e2e/argo_server_test.go#L1031-L1211

### Modifications
Add `--rerun-fails-run-root-test `. Docs: https://github.com/gotestyourself/gotestsum/blob/06f60b3249917ddd1a04c8e2586116d1a87bc67c/cmd/testdata/gotestsum-help-text#L30

### Verification

E2E tests
### Documentation

N/A

### DIFF
--- a/test/e2e/argo_server_test.go
+++ b/test/e2e/argo_server_test.go
@@ -430,8 +430,11 @@ func (s *ArgoServerSuite) createServiceAccount(name string) {
 	secret, err := s.KubeClient.CoreV1().Secrets(fixtures.Namespace).Create(ctx, secrets.NewTokenSecret(name), metav1.CreateOptions{})
 	s.Require().NoError(err)
 	s.T().Cleanup(func() {
-		_ = s.KubeClient.CoreV1().Secrets(fixtures.Namespace).Delete(ctx, secret.Name, metav1.DeleteOptions{})
-		_ = s.KubeClient.CoreV1().ServiceAccounts(fixtures.Namespace).Delete(ctx, name, metav1.DeleteOptions{})
+		ctx := logging.TestContext(s.T().Context())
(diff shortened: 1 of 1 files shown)


test/e2e/argo_server_test.go
```diff
@@ -430,8 +430,11 @@ func (s *ArgoServerSuite) createServiceAccount(name string) {
 	secret, err := s.KubeClient.CoreV1().Secrets(fixtures.Namespace).Create(ctx, secrets.NewTokenSecret(name), metav1.CreateOptions{})
 	s.Require().NoError(err)
 	s.T().Cleanup(func() {
-		_ = s.KubeClient.CoreV1().Secrets(fixtures.Namespace).Delete(ctx, secret.Name, metav1.DeleteOptions{})
-		_ = s.KubeClient.CoreV1().ServiceAccounts(fixtures.Namespace).Delete(ctx, name, metav1.DeleteOptions{})
+		ctx := logging.TestContext(s.T().Context())
```
MasonM, line 433: Added this because these service accounts weren't being cleaned up, and when I debugged, it seems the context was cancelled

PR #15796, opened 2026-03-20
fix(ui): populate URL filter parameters on first load. Fixes #15794

### LINKED ISSUE #15794: UI: filter parameters are not populated in the URL on first load
### Pre-requisites

While reviewing #15217 I saw that the URL is not populated on first load to the workflows list and workflow template pages. This is due to the early return via `isFirstRender.current`. We can combine this approach to use `history.replace` with all the parameters on the first render, and push otherwise, which avoids the "multiple backs" issue originally described in #13940, while updating the URL to show the actual search parameters that are active.

This is a small quality-of-life improvement so that users can share a filter URL directly from the page, instead of having to click (on some filter) twice to update the URL to what they actually want to share.

### Version(s)

latest main (5456d80f25c2437c4d717e6e5f266e26ab496449)

```YAML
* Click onto the workflows list and note the URL is `<host>/workflows`. Take note of the search filters that are pre-set (from a prior session) in the UI, e.g. namespace. These are not reflected in the URL.
* Update a filter, e.g. set "name contains" to "test"
* Observe the URL updating to `<host>/workflows?namespace=argo&limit=10&Contains=test`

Expected behaviour: we see `<host>/workflows?namespace=argo&limit=10` on first load.
```

```text
N/A
```

```text
N/A
```

elliotgunton:
Fixes #15794

### Motivation

Small quality of life improvement for query filter parameter behaviour

### Modifications

* For the Workflow, WorkflowTemplate and CronWorkflow pages

### Verification

https://github.com/user-attachments/assets/c1cd2e00-6af4-4142-adea-3bdbeea4a927

### DIFF
--- a/ui/src/cron-workflows/cron-workflow-list.tsx
+++ b/ui/src/cron-workflows/cron-workflow-list.tsx
@@ -51,16 +51,13 @@ export function CronWorkflowList({match, location, history}: RouteComponentProps
 
     // save history
     useEffect(() => {
-        if (isFirstRender.current) {
-            isFirstRender.current = false;
-            return;
-        }
-        history.push(
+        (isFirstRender.current ? history.replace : history.push)(
             historyUrl('cron-workflows' + (nsUtils.getManagedNamespace() ? '' : '/{namespace}'), {
                 namespace,
                 sidePanel
             })
         );
+        isFirstRender.current = false;
     }, [namespace, sidePanel]);
 
     // internal state
--- a/ui/src/workflow-templates/workflow-template-list.tsx
+++ b/ui/src/workflow-templates/workflow-template-list.tsx
@@ -64,22 +64,19 @@ export function WorkflowTemplateList({match, location, history}: RouteComponentP
     );
 
     useEffect(() => {
-        if (isFirstRender.current) {
-            isFirstRender.current = false;
-            return;
-        }
         storage.setItem('options', {labels}, {});
         const params = new URLSearchParams();
         labels?.forEach(label => params.append('label', label));
         if (sidePanel) {
             params.append('sidePanel', 'true');
         }
-        history.push(
+        (isFirstRender.current ? history.replace : history.push)(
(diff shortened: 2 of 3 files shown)


themavik (commented):
history.replace on first render syncs workflows, cron-workflows, and workflow-templates URLs without stacking a duplicate push. nit: workflow-template-list still encodes sidePanel as the string true in query params—confirm readers still treat that the same after the reorder.

ui/src/cron-workflows/cron-workflow-list.tsx
MasonM: Minor: this could be de-duplicated a bit:
```suggestion
(isFirstRender.current ? history.replace : history.push)(
    historyUrl('cron-workflows' + (nsUtils.getManagedNamespace() ? '' : '/{namespace}'), {
        namespace,
        sidePanel
    })
);
isFirstRender.current = false;
```

ui/src/workflow-templates/workflow-template-list.tsx
MasonM: Same as above:
```suggestion
        (isFirstRender.current ? history.replace : history.push)(
            historyUrl('workflow-templates' + (nsUtils.getManagedNamespace() ? '' : '/{namespace}'), {
                namespace,
                extraSearchParams: params
            })
        );
        isFirstRender.current = false;
```

MasonM (commented):
Good job! Just some minor comments about de-duplication

ui/src/cron-workflows/cron-workflow-list.tsx
MasonM: For some reason, GitHub isn't showing this suggestion correctly. This is meant to replace lines 54 to 69

elliotgunton: Nice, I didn't know that was possible!

MasonM (approved):
Thanks!

PR #15788, opened 2026-03-20
fix(controller): preserve NodeError phase in steps

thevilledev:
Fixes #14337

### Motivation

The steps controller unconditionally converts `NodeError` to `NodeFailed` in two locations when propagating phases through the step hierarchy. When a step child errors (e.g. pod error, expression evaluation error), the step group and parent `Steps` node are both hardcoded to `NodeFailed` regardless of the child's actual phase. This loses the `Error` vs `Failed` distinction, which matters for [`continueOn.error`](https://argo-workflows.readthedocs.io/en/latest/fields/#continueon) and for users who rely on the workflow phase to distinguish infrastructure errors from logical failures.

### Modifications

Two changes:

- when marking the parent `Steps` node from a failed/errored step group, use `sgNode.Phase` instead of hardcoded `wfv1.NodeFailed`
- when marking the step group from a failed/errored child, use `childNode.Phase` instead of hardcoded `wfv1.NodeFailed`

### Verification

I added a regression test `TestStepsPreserveNodeErrorPhase`, with a Steps template with a child in `NodeError` phase. Asserts both the step group and Steps node preserve `NodeError`. Fails on main (gets Failed), passes with fix.

I also had to update a few existing tests whose expectations changed from Failed to Error. The old code masked the true phase by converting all errors to failures. These tests use pre-populated node state without real pods in the kubeclient, so when the controller assesses pod status and finds no pod, it marks the node as Error (infrastructure problem) rather than Failed (application exit).

Test manifest for local Kind cluster:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: verify-steps-error-
spec:
  entrypoint: main
  serviceAccountName: argo
  templates:
  - name: main
    steps:
    - - name: run-eval
        template: erroring-eval
        arguments:
          parameters:
          - name: INPUT
            value: "no-colon-here"
  - name: erroring-eval
    inputs:
      parameters:
      - name: INPUT
    steps:
    - - name: placeholder
        template: noop
        when: "false"
    outputs:
      parameters:
      - name: RESULT
        valueFrom:
          expression: "inputs.parameters.INPUT.splitList(':')[1]"
  - name: noop
    suspend:
      duration: "0"
```

Currently `main` branch sets the workflow phase to `Failed` where as this fix sets it to `Error`.

### Documentation

This is a bug fix that corrects phase propagation. No documentation changes needed.

PR #15798, opened 2026-03-20
fix: webhook dispatch applies template metadata

thevilledev:
### Motivation

When the webhook event handler creates a workflow from a `WorkflowTemplateRef`, it does not copy `spec.workflowMetadata.labels` or `spec.workflowMetadata.annotations` from the resolved template onto the workflow. These are only applied later asynchronously by the controller in `updateWorkflowMetadata()`.

This causes two problems:

- Label-selector watches cannot find the workflow between creation and controller reconciliation, leading to timeouts in `WaitForWorkflow()`.
- If the workflow is archived before cleanup runs (or cleanup uses label selectors), stale archived workflows without the expected labels are never cleaned up, polluting subsequent test runs. See [this](https://github.com/argoproj/argo-workflows/actions/runs/23351490877/job/67931896440?pr=15788) CI run from #15788. Overall seems to manifest into `CI / E2E Tests (test-api, mysql, true)`.

CronWorkflows already apply `workflowMetadata` at creation time.

### Modifications

After fetching the WorkflowTemplate and calling `NewWorkflowFromWorkflowTemplate()`, copy `spec.workflowMetadata.labels` and `spec.workflowMetadata.annotations` from the template onto the workflow. This runs before `populateWorkflowMetadata()` so event binding metadata can still override template values. `labelsFrom` is intentionally left to the controller since it requires runtime parameter evaluation.

### Verification

Added two unit tests that verify:

- A workflow created via webhook dispatch carries `workflowMetadata` labels and annotations from the template alongside standard dispatch labels.
- Event binding metadata takes precedence over template metadata for the same label key.
 
Both tests fail against the unfixed code and pass with the fix.

### Documentation

No documentation changes needed. This aligns webhook dispatch behavior with the existing CronWorkflow behavior.

### DIFF
--- a/server/event/dispatch/operation.go
+++ b/server/event/dispatch/operation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"strings"
 
 	"github.com/expr-lang/expr"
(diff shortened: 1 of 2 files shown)


thevilledev:
Thanks @isubasinghe! Tests look green, except the code coverage upload (which seems unrelated). Hopefully the tests pass more often now ❤️

PR #15818, opened 2026-03-29
fix(test): poll in TestWorkflowArchiveServiceList

thevilledev:
### Motivation

Address test failure from another test fix PR: #15816 and https://github.com/argoproj/argo-workflows/actions/runs/23716864596/job/69085590152?pr=15816

### Modifications

`WaitForWorkflow(ToBeArchived)` watches K8s labels, but the `/api/v1/archived-workflows` endpoint queries the archive DB which may lag behind. The bare assertion fails when one workflow has not been persisted yet. Wrap in `s.Eventually` to poll until both workflows appear, matching the sibling test.

### Verification

Tests pass.

### Documentation

Tests-only change, no need for a documentation update.


Joibel:
I'd also considered doing this change, so thank you

thevilledev:
Nice, thanks! I forgot to take this out of the drafts but ready now.

PR #15819, opened 2026-03-30
POC: K8s API Aggregation

droctothorpe:
This PR is just a POC. I wanted to see if it was possible to refactor AWF to use K8s API Aggregation to store CRs in a sql database instead of etcd. It turns out: it is. Unsurprisingly, this is almost entirely LLM generated, though it required a crap ton of redirection. Don't bother reading it line-by-line. The [readme](https://github.com/droctothorpe/argo-workflows/blob/b35cab1b63b1a8197fc682e7259d81925d0eaff4/docs/aggregated-apiserver.md) provides a good overview. [This](https://k8s.info/docs/expert/api-aggregation) is a useful reference as well.

There are undoubtedly tons of drawbacks, but this approach theoretically solves for persistence, eliminates the need for workflow archives, eliminates the need for node status offloading, and overcomes the etcd `--max-request-bytes` limit. 

If there's a real appetite for this feature, I'm happy to move it beyond just a POC.

### DIFF
--- a/cmd/argo/commands/server.go
+++ b/cmd/argo/commands/server.go
@@ -57,8 +57,12 @@ func NewServerCommand() *cobra.Command {
 		kubeAPIQPS               float32
 		kubeAPIBurst             int
 		allowedLinkProtocol      []string
-		logFormat                string // --log-format
-		logLevel                 string // --loglevel
+		logFormat                  string // --log-format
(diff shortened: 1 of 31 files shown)


droctothorpe:
image

Screenshot of a workflow that completed using the API Aggregation.

PR #15844, opened 2026-04-02
feat: added config option to disable agent pod creation. Fixes #7891

### LINKED ISSUE #7891: Global agent pod instead of one per workflow
# Summary
Hi all, I find argo workflows support plugins since 3.3. An agent pod will load plugins to serve no pods tasks.
In #5544 , agent pod is designed to be one per workflow.
I'm wondering if agent pod could be a global one instead of one per workflow. The agent pod could watch all workflowset  and send request to the plugins. The global agent pod will save k8s resources and also improve performance.

# Use Cases
In our prod, we create hundred thousands of workflows everyday. Most of workflows are short time tasks. We would use plugins to do some non-pods task(eg. email notifications or light calculation).
One global agent pod can meet our requirements and also shorten workflow execution time.

---

gaurang9991:
Fixes #7891

### Motivation

Motivation for the changes are to utlize executor plugin concept to run high-frequency batch tasks without creating multiplepods. The workload for agnet pod can be off loaded to external agents or controller independent of argo workflow(long running). This reduces the time taken for agent pog creation. Worklfow controller only needs to manage the task sets (create, update, delete) and wait for them get updated with outputs. 

### Modifications

Added a conifgurable key in workflow controler configmap to enable and disable this behaviour with changes in agent pod creation method.

### Verification

Added key to workflow controller confgimap with `disableAgentPodCreation: "true"`
Submitted a workflow with plugin. The task set was created and no agent pod was created.

### Documentation

Added key in the controller-config-map example in docs


Hariharasuthan99 (approved):
LGTM

adityaraj178:
LGTM

rajarshipal-lab (approved):
LGTM 👍

Joibel (approved):

Joibel:
We can extend this in future for more use cases, like per template or per workflow behaviour. This is quite a brutal hammer of a config option!

PR #15875, opened 2026-04-07
feat: Add pod-free wait-for-delete resource action. Fixes #15874

AndreKurait:
Fixes #15874

### Motivation

There is currently no clean way to block a workflow until a Kubernetes resource is fully deleted without creating a pod. Users resort to script templates running `kubectl wait --for=delete` in a loop, which incurs pod scheduling, image pull, and memory overhead for what is fundamentally a single API call.

This adds a new pod-free `action: wait` with `waitFor: delete` for resource templates, following the same controller-side pattern used by Suspend nodes.

### Modifications

- **API** (`workflow_types.go`): Added `wait` to the `Action` enum and a new `WaitFor` field (enum: `delete`) on `ResourceTemplate`.
- **Validation** (`validate.go`): `waitFor` is required when `action: wait`, forbidden otherwise. `successCondition`, `failureCondition`, and outputs are disallowed with `wait`.
- **Controller** (`operator.go`): `executeResource` intercepts `action: wait` before pod creation. `checkResourceWaitForDelete` uses the dynamic client to check resource existence — 404 means succeeded, exists means requeue. `podReconciliation` skips wait-action nodes so they are not flagged as pod-missing.
- **Controller** (`controller.go`): Added a `restMapper` to resolve GVK → GVR for the dynamic client GET.
- **CRDs/OpenAPI**: Regenerated via `controller-gen` and `openapi-gen`.

### Verification

- **Unit tests** (`operator_test.go`): Two test cases — resource already deleted (→ Succeeded), resource still exists (→ Running, no pod created).
- **Unit tests** (`validate_test.go`): Five test cases covering valid usage and all invalid combinations (missing waitFor, waitFor on non-wait action, with successCondition, with outputs).
- **Existing test regression**: Updated `TestInvalidResourceWorkflow` expected error message to include `wait` in the valid actions list. Full `workflow/controller` (82s) and `workflow/validate` test suites pass.
- **E2E test** (`resource_template_test.go`): `TestWaitForDeleteAlreadyGone` — submits a workflow that waits for a non-existent ConfigMap, expects immediate success.
- **Lint**: `golangci-lint` v2.11.3 passes with 0 issues. `go vet` clean. Feature doc validated.

### Documentation

- `docs/resource-template.md`: Added "Wait for Resource Deletion" section with usage example.
- `.features/pending/wait-for-delete-resource-action.md`: Feature description file for release notes.
- The `waitFor` enum is designed for future extension (e.g. `waitFor: condition`) without breaking changes.

### DIFF
--- a/.features/pending/wait-for-delete-resource-action.md
+++ b/.features/pending/wait-for-delete-resource-action.md
@@ -1,0 +1,26 @@
+Description: Add pod-free `action: wait` with `waitFor: delete` for resource templates
+Authors: [Andre Kurait](https://github.com/AndreKurait)
+Component: General
+Issues: 15874
+
+The new `action: wait` with `waitFor: delete` allows workflows to block until a Kubernetes resource is fully deleted, without creating a pod.
+The controller checks resource existence directly via the K8s API on each reconciliation cycle.
+
+This eliminates pod overhead, OOM risk, and image pull costs for deletion-watching patterns.
+
+```yaml
+- name: wait-for-job-deleted
+  resource:
+    action: wait
+    waitFor: delete
+    manifest: |
+      apiVersion: batch/v1
+      kind: Job
+      metadata:
+        name: my-job
+```
+
+If the resource doesn't exist when the step starts, it succeeds immediately.
+Use `activeDeadlineSeconds` to set a timeout.
+
+`successCondition`, `failureCondition`, and outputs cannot be used with the `wait` action.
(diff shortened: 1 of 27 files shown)


Joibel:
I won't accept this as is. This requires the controller to have global permissions to do the action you requested, and that's not OK. In general performing any node action in the controller is not OK. This also puts the controller into a busy loop waiting for deletion, which is not OK.

I'd accept modifications to the resource templates:
* `delete` with wait would be the most obvious choice here
* `get` with more control over behaviour when object is gone (currently delete will fail the node).

Joibel (changes requested):
I don't like the design - this comment is my "request changes" comment to block for me to review later.

PR #15884, opened 2026-04-08
fix: unresolved reference for omitted/skipped children when using steps

KowalczykBartek:
Fixes https://github.com/argoproj/argo-workflows/issues/15873

### Motivation

Try to resolve child's output when steps is actually skipped.


isubasinghe (changes requested):
@KowalczykBartek can you please signoff your commit so that DCO is happy? 

Also please add some tests like this: github.com/argoproj/argo-workflows/pull/15841

KowalczykBartek:
@isubasinghe sure I will add test - but one question regarding commit - I already signed it, commit is verified 
```
GPG Key ID: 8EB42525E8A929A9
Verified on Apr 8, 2026, 09:23 PM
```
can we re-run this check ?

KowalczykBartek:
@isubasinghe  test case added

isubasinghe:
@KowalczykBartek I also realised that this PR is to `release-3.7`, it should be towards `main`. 

The easiest way I think to resolve these issues is to git squash your commits into one. 
Then dump out a .patch file and then cherry-pick that patch to a branch off main.

KowalczykBartek:
@isubasinghe I create another PR https://github.com/argoproj/argo-workflows/pull/15926/changes to main branch (signed :) )

PR #15897, opened 2026-04-11
feat(persistence): Allow using custom database schema for persistence. Fixes #2452

### LINKED ISSUE #2452: Custom db schema for persistence tables
# Summary

In the latest release of argo (v2.6.2 at the time of writting) it is not possible to define a schema when enabling persistence. I would like to have persistence stored in a schema different that `public`. 

# Motivation

Many applications or organizations use a database for multiple purposes. Having the ability to choose the schema where to store the information about argo would be useful. 

# Proposal

An extra parameter in the **workflow-controller-configmap** when persistence is enabled: 
```yaml
postgresql:
        host: localhost
        port: 5432
        database: postgres
        tableName: argo_workflows
        schema: argo # new attribute
        userNameSecret:
          name: argo-postgres-config
          key: username
        passwordSecret:
          name: argo-postgres-config
          key: password
```

---

jnt0r:
Fixes #2452

### Motivation
Currently, Argo Workflows uses the default database schema for persistence. This PR introduces the ability to configure a custom schema. As MySQL does not have the concept of schemas and used the database name as kind of a schema, it is only introduced to postgres.

Key benefits:

- Namespace Isolation: Clearly separate Argo Workflows metadata from other tables in the same database.
- Permissions Management: Allows DBAs to restrict the Argo service user to a specific schema.
- Flexibility: Supports complex infrastructure setups where the default schema (e.g., public in PostgreSQL) is restricted or reserved for other purposes.

### Modifications

- It is now possible to set the optional "schema" Property in the  persistence config for the workflow controller (configmap).
- The schema is created for postgresql if the schema is set and does not already exist (migration)

### Verification

I adjusted the Unit Tests to verify the schema is set to the database connections only when schema is explicitly set. When the schema is not overridden the default schema is used to provide downward compatibility.

Locally tested with hack/db (setting custom dsn like it would be generated in code: -d postgres://postgres:password@host.docker.internal:5432/postgres?search_path=argo)

Also, I created a new integration test with postgres and mysql testcontainers to test the connection is correctly using the new schema.

Should I also create a new E2E test?

### Documentation

Added the schema property in examples and added additional chapter to "workflow archive" chapter.

## Summary by CodeRabbit

* **New Features**
  * Added support for a custom PostgreSQL logical schema for persistence, synchronization, and workflow archive tables.
  * PostgreSQL sessions now default to the configured schema via connection settings.
  * Migrations can run within (and create, if needed) the selected schema.

* **Documentation**
  * Documented the new `persistence.postgresql.schema` / config guidance with examples.
  * Updated required PostgreSQL permissions, including options for existing vs auto-created schemas.
  * Clarified schema behavior is PostgreSQL-specific (MySQL uses `database` as an alias).

* **Tests**
  * Added integration coverage to verify schema usage for PostgreSQL and MySQL.


docs/workflow-archive.md
```diff
@@ -97,3 +97,12 @@ Example:
 
     persistence:
       archive: false
+
+## Customizing PostgreSQL Database Schema
+
+To change the schema for the tables in PostgreSQL, set `schema:` to the desired schema in the `persistence` section of [your configuration](workflow-controller-configmap.yaml). Only available on PostgreSQL.
```
tico24: Each sentence on a new line please. It make future reviews easier. Thanks!

jnt0r:
@tico24 can you take another look at this change?

tico24 (commented):
Only reviewed the docs files (`.features` and `docs/`), since those are what I own per CODEOWNERS. Haven't looked at the Go side.

docs/workflow-archive.md
```diff
@@ -99,3 +99,13 @@ Example:
 
     persistence:
       archive: false
+
+## Customizing PostgreSQL Database Schema
+
+To change the schema for the tables in PostgreSQL, set `schema:` to the desired schema in the `persistence` section of [your configuration](workflow-controller-configmap.yaml).
```
tico24, line 119: Have you thought about the permissions side here? The "Required database permissions" section above says you need `CREATE`/`USAGE` on `public` — with a custom schema set, that's no longer quite right. You'd need either `CREATE` on the database (since the migration runs `CREATE SCHEMA IF NOT EXISTS`), or `CREATE`/`USAGE` on a pre-existing custom schema. Worth a line in this new section?

.features/pending/persistence-schema.md
```diff
@@ -0,0 +1,6 @@
+Description: You can now configure a specific postgres database schema for Argo Persistence in the controller config map.
+Authors: [Jonathan Pollert](https://github.com/jnt0r)
+Component: General
+Issues: 2452
+
+Added support for custom database schemas to improve data isolation and security in shared environments. This allows Argo to operate within a designated logical schema rather than the default. Note: This feature is not applicable to MySQL, as it does not support logical schemas; for MySQL users, database names should continue to be used for application separation.
```
tico24: Could you break this onto one sentence per line? See `aws-rds-iam-auth.md` for the format other entries use here.

docs/workflow-archive.md
jnt0r: Yes, completely right. I adjusted that.

.features/pending/persistence-schema.md
jnt0r: did that

PR #15903, opened 2026-04-13
docs: add notes on pod security and format security.md

Joibel:
### Motivation

Explain that we don't gateway pod creation

### Modifications

Docs only.

### Verification

make docs

### Documentation

One sentence per line.

Explain that Argo is a pod controller, and we don't gateway pod creation.


Joibel:
@thevilledev, tagging you for comments on this - does it make sense to you?

isubasinghe (approved):

thevilledev (approved):
@Joibel LGTM! Thanks for clarifying this

PR #15926, opened 2026-04-13
fix: populate scope with empty values for outputs of skipped/omitted …

KowalczykBartek:
Fixes https://github.com/argoproj/argo-workflows/issues/15873

### Motivation

Try to resolve child's output when steps is actually skipped.


isubasinghe:
@KowalczykBartek your commit isn't signed off. I've addressed the linter issues and squashed my  commit onto yours. Hopefully you are okay with this?

KowalczykBartek:
@isubasinghe all fine, thank you very much for help, but I cannot understand why my commit is not visible as signed :D I signed it, it also visible here on github as verified by my pgp key 

EDIT: ok I know - i used **-S** instead of **-s** - details matters

isubasinghe:
@KowalczykBartek can you please open another PR and sign it off? I cannot approve this according the CNCF rules.

KowalczykBartek:
@isubasinghe I believe this one is the last one https://github.com/argoproj/argo-workflows/pull/15932 (its signed and verified) :)

KowalczykBartek:
@isubasinghe when we can expect that to be merged? and **released** ?

PR #15933, opened 2026-04-15
docs: generate database migration reference

Joibel:
### Motivation

With database migrate set to false operators are forced to read the code to understand the database setup and migrations.

This can be (mostly) generated into documentation, so I've done that here.

### Modifications

Add a generator at hack/docs/migrate_docs.go that produces docs/database-migrations.md from the archive and sync migration change lists. The generator extracts SQL for MySQL and PostgreSQL separately by type-asserting against AnsiSQLChange and TypedChange.

To support the generator, export MigrateChanges from persist/sqldb and util/sync/db, and export TypedChange from util/sqldb so ByType returns an inspectable value.

Cross-link the new page from workflow-archive.md and synchronization.md, add it to the Configuration nav in mkdocs.yml, and wire it into the Makefile.

### Verification

Visual inspection of the new documentation.
`make docs`

### Documentation

This is really only a change aimed at documentation - a small amount of real code changed to make that pretty.


isubasinghe (approved):
This is really cool

PR #15936, opened 2026-04-15
fix: MariaDB compatibility for JSON queries and native password auth. Fixes #15413

### LINKED ISSUE #15413: UI errors when filtering workflows with MariaDB
### Pre-requisites

I have just upgraded to v3.7.8 (from v3.6.4) and immediately started experiencing the following error in the UI:

Image

Things I noticed, and could help pinpoint the problem:
1. As you can see in the screenshot, the UI code seems to crash and workflows are not loaded.
2. This seems to happen when a filter is applied in the left sidebar: right now I have entered a `namespace` and selected `Phases` as `Error` (it happens with other combinations, too). When the filter is cleared, and the page reloaded, everything works again.
3. The error message in the screenshot shows a MariaDB error, seemingly complaining about the `-->` operator in combination with a JSON query on labels and annotations.
4. This does not seem to affect running workflows, nor queries made via CLI: only UI.

Given the above, I would guess that this is a regression of https://github.com/argoproj/argo-workflows/issues/13202. However, I have no experience with the Argo codebase to trace the problem, so I might be wrong.

### Version(s)

v3.7.8 (Helm chart 0.47.1)

```YAML
n/a
```

```text
n/a
```

```text
n/a
```

mariadb-MarioEApostolov:
Fixes #15413

### Motivation

The MySQL branch of [ListWorkflows](cci:1://file:///Users/marioapostolov/Documents/work/tmp/argo-workflows/persist/sqldb/workflow_archive.go:77:1-78:87) uses `->` and `->>` JSON operators which are not supported by MariaDB, causing SQL syntax errors when listing archived workflows.

Additionally, `go-sql-driver/mysql` v1.9.x changed the default of `AllowNativePasswords` to `false`, which prevents connections to MariaDB since it uses `mysql_native_password` by default.

### Modifications

- Use `JSON_EXTRACT()`/`JSON_UNQUOTE()` instead of `->` / `->>` in the MySQL branch of [ListWorkflows](cci:1://file:///Users/marioapostolov/Documents/work/tmp/argo-workflows/persist/sqldb/workflow_archive.go:77:1-78:87) (compatible with MySQL 5.7+ and MariaDB 10.2+)
- Set `AllowNativePasswords: true` in the MySQL driver config
- Add integration tests for both [ListWorkflows](cci:1://file:///Users/marioapostolov/Documents/work/tmp/argo-workflows/persist/sqldb/workflow_archive.go:77:1-78:87) and session creation, running against MySQL 8.4 and MariaDB 11.4 via testcontainers
- Add MariaDB to the supported databases list in the workflow archive documentation

### Verification

Ran `goimports`, `go vet`, and the full test suites for the affected packages. All existing and new tests pass. No regressions.

### Documentation

Updated [docs/workflow-archive.md](cci:7://file:///Users/marioapostolov/Documents/work/tmp/argo-workflows/docs/workflow-archive.md:0:0-0:0) to include MariaDB (>= 10.2) as a supported database.

## Release Notes

* **New Features**
  * Added MariaDB (version >= 10.2) as a newly supported database backend alongside Postgres and MySQL, expanding database flexibility for deployments.

* **Documentation**
  * Updated workflow archive documentation to explicitly list MariaDB as a supported database option alongside Postgres and MySQL.

* **Improvements**
  * Enhanced compatibility for archived workflow metadata queries and listing operations across MySQL and MariaDB databases.


Joibel:
@coderabbitai review

Joibel (commented):
### Claude Code review

Found 9 issues:

1. **Container-leak risk in test setup helpers** — both `setupMySQLArchiveTest` and `setupMySQLContainer` start a testcontainers container, then call `require.NoError` on several subsequent steps. If any of those fail, the container is never terminated because the cleanup closure is only invoked once the caller defers it. Should register `t.Cleanup(func() { testcontainers.TerminateContainer(c) })` immediately after container start. (bug — resource leak)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/persist/sqldb/workflow_archive_mysql_test.go#L36-L81

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/util/sqldb/sqldb_test.go#L29-L61

2. **Test scaffolding duplicated verbatim across two new files** — `mysqlVariant` struct, `mysqlVariants` map, and most of the container-bootstrap code are copy-pasted between the two new test files. Image tags, wait-log strings, and setup behaviour will drift independently. Extract to a shared test helper. (DRY / maintainability)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/persist/sqldb/workflow_archive_mysql_test.go#L24-L50

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/util/sqldb/sqldb_test.go#L17-L42

3. **`//nolint:errcheck` silently swallows `TerminateContainer` errors** — the existing pattern in `workflow/sync/database_helper_test.go` logs the termination error via `t.Logf`. The new code swallows it entirely, which will mask container leaks in CI. (inconsistent with repo convention)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/util/sqldb/sqldb_test.go#L58-L62

4. **Test assertions cover only 3 of 8 changed JSON expressions** — the PR's primary fix is 8 `JSON_EXTRACT` / `JSON_UNQUOTE` expressions. `TestMySQLListWorkflows` asserts only `Progress`, `Message`, and one label. `Annotations`, `Arguments`, `Suspend`, `EstimatedDuration`, and `ResourcesDuration` are never asserted. `Arguments.Parameters[msg=hello]` is set in the test input but never validated on the way back. A bug in any of those extractions would not be caught. (test coverage)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/persist/sqldb/workflow_archive_mysql_test.go#L127-L137

5. **`require.Equal(t, MySQL, dbType)` is tautological** — `CreateDBSessionWithCreds` returns `MySQL` whenever `cfg.MySQL != nil`; the assertion checks a Go branch, not a property of the MariaDB connection. It does not differentiate MySQL from MariaDB in any way. (ineffective assertion)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/util/sqldb/sqldb_test.go#L73-L84

6. **Misleading `archive-strategy` label in test input** — the test sets `workflows.argoproj.io/archive-strategy: Persisted` on the input workflow, but `ArchiveWorkflow` unconditionally overwrites the archiving-status label before persisting. The input value has no bearing on what gets stored or retrieved. (misleading test fixture)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/persist/sqldb/workflow_archive_mysql_test.go#L100-L110

7. **`AllowNativePasswords: true` has no inline comment** — the rationale (MariaDB defaults to `mysql_native_password`, go-sql-driver v1.9 changed the zero-value default to `false` when using the struct literal) is non-obvious and lives only in the PR description. A future refactor is likely to "clean it up". Add a brief comment. (maintainability)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/util/sqldb/sqldb.go#L196-L210

8. **Windows skip uses runtime check instead of `//go:build !windows` build tag** — commit 7674c7f5f just standardised testcontainers tests on the build-tag pattern. These new files revert to the older runtime check, inconsistent with the freshly-established convention (and still compiles the testcontainers import on Windows). (consistency)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/persist/sqldb/workflow_archive_mysql_test.go#L84-L92

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/util/sqldb/sqldb_test.go#L64-L72

9. **Other docs still advertise only MySQL** — the `AllowNativePasswords` fix in `util/sqldb/sqldb.go` applies to all MySQL connections, so offloading/synchronization features also now work with MariaDB. `docs/memoization.md`, `docs/offloading-large-workflows.md`, `docs/synchronization.md`, and `docs/synchronization-config.md` still say "PostgreSQL or MySQL" without mentioning MariaDB. (docs scope)

https://github.com/argoproj/argo-workflows/blob/7007a62c3f78a76fd1c25ed38d3e830a8207f73b/docs/workflow-archive.md#L1-L8

mariadb-MarioEApostolov:
Thanks for the review. All feedback addressed in the latest push:

1. Container leak → `t.Cleanup` right after container start
2. DRY → shared `MySQLVariant` in `util/sqldb/mysql_test_helper.go`
3. TerminateContainer → error logged via `t.Logf`
4. Test coverage → assertions for annotations, arguments, suspend, estimatedDuration
5. Tautological assertion → removed
6. Misleading label → removed from fixture
7. AllowNativePasswords → inline comment added
8. Build tag → `//go:build !windows`
9. Docs → updated synchronization, synchronization-config, offloading docs
10. Missing `AS suspend` alias → fixed with round-trip assertion
11. Inconsistent `JSON_UNQUOTE` → removed from arguments

All tests pass locally against MySQL 8.4 and MariaDB 11.4.

Joibel (approved):
Thanks for fixing this up. Great to have team Maria on board!

PR #14030, opened 2024-12-24
fix: search and replace null bytes for postgres only.Fixes #13711

isubasinghe:
# Motivation
Unfortunately postgres does not handle NULL bytes in the json, this is as a result of a non-compliant implementation of JSON that exists in postgres. See  more details about this limitation [here](https://github.com/argoproj/argo-workflows/issues/13711#issuecomment-2507193576). 

# Modifications
This attempts a search and replace in postgres, it does this by reserving a special phrase for replacement of null bytes. 
On the `GetWorkflows` path, this special phrase is replaced with a null byte.

# Verification
This has tests and additionally been verified by checking postgres.

PR #15938, opened 2026-04-15
feat: SQL database-backed memoization cache

droctothorpe:
## Summary

Add an opt-in SQL-backed memoization cache for workflows. When `workflow-controller-configmap` includes `memoization:`, memoization entries are stored in PostgreSQL or MySQL instead of ConfigMaps. ConfigMap-backed memoization remains the default when `memoization:` is omitted.

## What changed

- Added `config.MemoizationConfig` plus `util/memo/db` for SQL memoization configuration, table migration, and query logic, including configurable `tableName` and `skipMigration`.
- Added a SQL memoization cache implementation and updated the cache factory/controller config flow so memoization can switch safely between ConfigMap-backed, SQL-backed, and disabled/no-op modes during config changes.
- SQL cache entries are stored by `namespace`, `cache_name`, and `cache_key` in the configured table (`cache_entries` by default), with `created_at` / `expires_at` timestamps and pruning by `expires_at`.
- `MemoizationCache.Save()` now accepts template `maxAge`; SQL-backed entries derive TTL from `memoize.maxAge`, `DEFAULT_MAX_AGE`, or the 30-day default, and the controller prunes expired rows on `MEMO_CACHE_GC_PERIOD`.
- Updated memoization load/save paths in `operator.go`, `dag.go`, `steps.go`, and `taskset.go` so backend errors still surface, while SQL backend unavailability is treated as a cache miss / skipped save instead of silently creating a ConfigMap.
- Updated controller config docs, memoization docs, env-var docs, and MySQL/PostgreSQL example manifests for the new `memoization:` block.

## Testing

- Added unit coverage for memoization DB config/migration/query logic, SQL cache behavior, controller config switching, cache-factory backend changes, and memoization GC.
- Added `test/e2e/sqldb_memoize_test.go` and CI coverage for SQL-backed memoization on both MySQL and PostgreSQL.
- Included the Windows `workflow/sync` test stabilization that makes `TestParallelismUpdate` deterministic across platforms since it was leading to flakiness in this PR.

## Notes

- On the SQL backend, `memoize.cache.configMap.name` is still required, but it is used as the logical cache name in the database; no ConfigMap is created.
- If `memoization:` is configured but the SQL backend cannot be initialized, memoization is disabled rather than falling back to ConfigMap-backed storage. ConfigMap-backed memoization is only used when `memoization:` is not configured.

## Backend selection / routing

```mermaid
flowchart TD
  A[Workflow node<br>uses memoize] --> B[get memoization cache]
  B --> C{memoization config set?}
  C -->|no| D[use ConfigMap cache]
  C -->|yes| E{SQL backend enabled?}
  E -->|yes| F[use SQL cache]
  E -->|no| G[use disabled or<br>no-op SQL cache]
  D --> H[load or save via ConfigMap]
  F --> I[load or save via SQL]
  G --> J[treat lookup as cache miss]
  G --> K[skip cache save]
```

## ConfigMap path

```mermaid
sequenceDiagram
  participant N as Workflow node
  participant C as WorkflowController
  participant F as cacheFactory
  participant M as ConfigMapCache
  participant K as Kubernetes API

  N->>C: getMemoizationCache(ns, cacheName)
  C->>F: GetCache(ConfigMapCache, ns, cacheName)
  F-->>C: ConfigMapCache

  alt load
    C->>M: Load(key)
    M->>K: GET ConfigMap
    K-->>M: ConfigMap / NotFound
    alt entry exists
      M->>K: UPDATE lastHitTimestamp
      M-->>C: Entry
    else miss
      M-->>C: nil
    end
  else save
    Note over M: maxAge is ignored for ConfigMap storage
    C->>M: Save(key, nodeID, outputs, maxAge)
    M->>K: GET or CREATE ConfigMap
    M->>K: UPDATE entry JSON
    M-->>C: ok / error
  end
```

## SQL path

```mermaid
sequenceDiagram
  participant N as Workflow node
  participant C as WorkflowController
  participant F as cacheFactory
  participant S as sqlDBCache
  participant D as SQL database
  participant G as GC ticker

  N->>C: getMemoizationCache(ns, cacheName)
  C->>F: GetCache(ConfigMapCache, ns, cacheName)
  F-->>C: sqlDBCache

  alt SQL backend enabled
    alt load
      C->>S: Load(key)
      S->>D: SELECT active entry
      D-->>S: row or no row
      S-->>C: entry or nil
    else save
      C->>S: Save(key, nodeID, outputs, maxAge)
      S->>D: UPSERT row with created_at and expires_at
      D-->>S: ok
      S-->>C: ok / error
    end

    G->>D: DELETE expired rows on each GC tick
  else SQL configured but unavailable
    C->>S: Load(key) / Save(...)
    S-->>C: nil on load / no-op on save
    C-->>N: treat as cache miss or skip save
  end
```

## AI

LLMs were used in the creation of this PR.


droctothorpe:
## Manual validation evidence

I ran two identical workflows. The first one results in a cache miss, the second one results does not:
image

I compared the status of the miss (left) and hit (right). Note how the hit has no resource duration:
image

The miss has an executor pod, the hit does not:
image

Here's the cache entry in postgres:
image

Here are the runs in the UI:
image

Here's the hit with a duration of 0s:
image

Joibel:
@coderabbitai review

Joibel:
### Claude Code review

Found 18 issues:

1. `tableName` in every doc example and overlay manifest is nested inside `postgresql:`/`mysql:` blocks, but the code reads only the top-level `MemoizationConfig.TableName`. The nested value is silently ignored; it only *appears* to work because the value matches the default.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/docs/memoization.md#L41-L55

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/config.go#L23-L36

2. `MEMO_CACHE_GC_PERIOD=0s` causes `time.NewTicker(0)` to panic. The sibling ConfigMap GC explicitly guards `if cacheGCPeriod != 0`; this new goroutine does not.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/workflow/controller/controller.go#L748-L756

3. `Prune` calls `sp.Session().SQL().DeleteFrom(...).Exec()` directly, bypassing `sp.With()`; `SessionProxy.Session()` godoc explicitly warns this path loses reconnect-on-error. Running on a 24h ticker, the connection is very likely to have gone idle — failures then persist until controller restart. `Load`/`Save` correctly use `TxWith`.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/queries.go#L75-L88

4. When the `memoization` block is removed from the ConfigMap at runtime, the `else` branch only logs — it never calls `cacheFactory.SetSessionProxy(nil, \"\")`, so SQL routing continues until controller restart. Also, a failed initial DB connection leaves `memoSessionProxy=nil` and the `if wfc.memoSessionProxy == nil` guard never retries.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/workflow/controller/config.go#L80-L96

5. `memodb.Migrate` logs migration failure as a warning but returns no error, and the caller still proceeds to `SetSessionProxy`. Compare `util/sync/db/config.go` which clears `SessionProxy` on failure, and persistence `initDB` which is fatal. Subsequent Load/Save will fail silently on every memoized node.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/config.go#L58-L75

6. `outputs` stored as raw `text` via `json.Marshal`. Postgres rejects `\u0000` null bytes in text columns; PR #14030 fixed exactly this class of bug in `persist/sqldb/workflow_archive.go` — the same null-byte handling is not applied here. Memoized step outputs can contain null-byte escapes.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/queries.go#L88-L119

7. `outputs text` column is capped at 65,535 bytes on MySQL, contradicting the headline claim of replacing the 1 MB ConfigMap limit. `persist/sqldb/migrate.go` uses `ByType` with MySQL-specific JSON/LONGTEXT conversion for exactly this reason; this migration does not.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/migrate.go#L11-L26

8. Every `Load` (cache hit) opens a transaction and issues SELECT + UPDATE to refresh `last_hit_at`. With the stated scale motivation (\"thousands of memoized nodes\"), every cache read becomes a write with transaction overhead — consider coalescing/debouncing `last_hit_at` updates.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/queries.go#L36-L73

9. `cf.namespace` is the controller's install namespace (from `NewCacheFactory(kubeclient, ns)`), not the workflow's namespace. Every row's `namespace` column has the same value, so the column is non-partitioning and the `(namespace, cache_name, key)` PK misleads about multi-tenant isolation.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/workflow/controller/cache/cache.go#L52-L74

10. Stale comment `// Only config maps are currently supported for caching` directly above `ConfigMapCache`. After this PR, `GetCache(ConfigMapCache, name)` routes to SQL when `sessionProxy != nil`; the `Type` enum is functionally dead and the comment is false.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/workflow/controller/cache/cache.go#L76-L124

11. `.features/pending/sql-memoization-cache.md` has `Issues: 15938` — the PR number itself. `hack/featuregen` generates a link to `/issues/15938` which GitHub resolves to the PR, not a feature-request issue. Convention in `.features/released/**` is to reference the originating issue.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/.features/pending/sql-memoization-cache.md#L1-L10

12. `_ = db.ErrNoMoreRows // keep import used` is dead code; `db` from `upper/db/v4` is otherwise unused in the e2e test and the import can simply be removed.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/test/e2e/sqldb_memoize_test.go#L140-L144

13. `MEMO_CACHE_GC_PERIOD` is out of alphabetical order in the Controller env-var table (placed between `CACHE_GC_AFTER_NOT_HIT_DURATION` and `CRON_SYNC_PERIOD` instead of after `MAX_OPERATION_TIME`).

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/docs/environment-variables.md#L27-L31

14. Docs say `cacheTTL: \"0\"` disables GC, but code treats `ttl == 0` as \"use default 90 days\" — only `ttl < 0` disables. Users following the documented path get 90-day pruning instead of disabled GC.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/workflow/controller/controller.go#L738-L748

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/docs/memoization.md#L56-L60

15. `Save` does DELETE-then-INSERT inside a transaction instead of UPSERT (`ON CONFLICT DO UPDATE` / `ON DUPLICATE KEY UPDATE`). Two concurrent Saves to the same (ns, cache_name, key) can both DELETE (seeing empty snapshot) and both INSERT, producing a PK-violation on the second.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/queries.go#L87-L119

16. `Prune` accepts `ctx context.Context` but never propagates it — `.Exec()` is called without context binding, so controller shutdown cannot cancel a long DELETE. Load/Save plumb ctx via `TxWith`.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/queries.go#L75-L88

17. `SessionProxyFromConfig` omits `sqldb.ConfigureDBSession`, so `MemoizationConfig.DBConfig.ConnectionPool` is silently ignored — connection pool settings have no effect. Persistence calls `ConfigureDBSession` in `workflow/controller/config.go` for exactly this.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/config.go#L38-L56

18. Column named `key` (a MySQL reserved word) and raw-SQL test queries using `WHERE key = ?` without backticks will fail on MySQL.

https://github.com/argoproj/argo-workflows/blob/0d9402b2d0662ab6d5f28fbf6bcfdbc4f468153d/util/memo/db/migrate.go#L13-L25

droctothorpe:
All 18 issues from the review have been addressed in 93b5b1925:

| Issue | Solution |
|-------|----------|
| 1. `tableName` nested inside `postgresql:`/`mysql:` blocks is silently ignored; code reads top-level `MemoizationConfig.TableName` | Moved `tableName` to top-level `memoization:` block in docs, overlay manifest, and quick-start manifest |
| 2. `MEMO_CACHE_GC_PERIOD=0s` causes `time.NewTicker(0)` panic | Added `periodicity <= 0` guard before `NewTicker` |
| 3. `Prune` calls `sp.Session()` directly, bypassing reconnect-on-error | Changed `Prune` to use `sp.With(ctx, ...)` |
| 4. Removing `memoization` config block at runtime doesn't clear SQL routing | Added `else` branch to call `SetSessionProxy(nil, "")` and nil out `memoSessionProxy` |
| 5. `Migrate` swallows errors; caller proceeds to `SetSessionProxy` | Changed `Migrate` to return `error`; caller falls back to ConfigMap on failure |
| 6. `outputs` stored as raw text without null-byte handling breaks Postgres | Added `postgresNullReplacement` handling in `Save`/`Load` (same pattern as `workflow_archive.go`) |
| 7. MySQL `text` column is capped at 64KB | Used `sqldb.ByType` to split migration: `text` for Postgres, `longtext` for MySQL |
| 8. Every `Load` opens a transaction with SELECT + UPDATE for `last_hit_at` | Debounced: `last_hit_at` is only updated if older than 1 hour; reads are now transaction-free |
| 9. `namespace` column is always the controller install namespace, not the workflow namespace | Added clarifying comment on the `namespace` field documenting this behavior |
| 10. Stale comment "Only config maps are currently supported for caching" | Updated comment to reflect SQL-backed caching support |
| 11. `.features/pending/` file has `Issues: 15938` (the PR number, not a feature-request issue) | Changed to `PRs: 15938` |
| 12. `_ = db.ErrNoMoreRows` dead code keeping unused import alive | Removed import and workaround line |
| 13. `MEMO_CACHE_GC_PERIOD` out of alphabetical order in env-var docs table | Moved to after `MAX_OPERATION_TIME` |
| 14. Docs say `cacheTTL: "0"` disables GC, but code treats 0 as default 90 days | Fixed all docs: `"0"` uses the default; negative value disables GC |
| 15. `Save` DELETE+INSERT race: concurrent Saves can both DELETE then both INSERT, causing PK violation | Replaced with `ON CONFLICT DO UPDATE` (Postgres) / `ON DUPLICATE KEY UPDATE` (MySQL) |
| 16. `Prune` doesn't propagate context; shutdown can't cancel long DELETE | Now uses `sp.With(ctx, ...)` which propagates context |
| 17. `SessionProxyFromConfig` omits `ConfigureDBSession`; connection pool settings silently ignored | Added `sqldb.ConfigureDBSession` call after creating the session proxy |
| 18. Column `key` is a MySQL reserved word; raw SQL queries without quoting fail on MySQL | Used backtick-quoting for MySQL, double-quote for Postgres in migration DDL and upsert SQL |

config/config.go
```diff
@@ -353,6 +357,20 @@ type SyncConfig struct {
...
+	DBConfig
+	// TableName is the name of the table to use for memoization cache entries.
+	// Defaults to "memoization_cache" if not set.
+	TableName string `json:"tableName,omitempty"`
+	// SkipMigration skips automatic database migration on startup.
+	SkipMigration bool `json:"skipMigration,omitempty"`
+	// CacheTTL is the duration after which cache entries that have not been hit are deleted by the
+	// garbage collector. Defaults to "2160h" (90 days). Set to "0" to disable GC.
+	CacheTTL TTL `json:"cacheTTL,omitempty"`
+}
```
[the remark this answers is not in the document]
droctothorpe: Fixed — updated the doc comment to say `Set to a negative value (e.g., "-1s") to disable GC`, matching the implementation where `ttl == 0` means "use default 90 days" and only `ttl < 0` disables GC. Also updated `docs/memoization.md` and `docs/workflow-controller-configmap.md` consistently.

docs/memoization.md
```diff
@@ -14,16 +14,73 @@ If you are using workflows prior to version 3.5 you should look at the [work avo
...
+        name: argo-postgres-config
+        key: username
+      passwordSecret:
+        name: argo-postgres-config
+        key: password
+    cacheTTL: "2160h"
+```
+
+The `cacheTTL` field controls how long cache entries that have not been accessed are retained.
+It defaults to `"2160h"` (90 days). Set to `"0"` to disable automatic pruning.
```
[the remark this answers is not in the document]
droctothorpe: Fixed — the docs now read: `Setting it to "0" uses the default TTL. Set to a negative value (e.g., "-1s") to disable automatic pruning.`

util/memo/db/config.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return cfg.TableName
+}
+
+func ConfigFromConfig(cfg *config.MemoizationConfig) Config {
+	return Config{
+		TableName:     TableName(cfg),
+		SkipMigration: cfg.SkipMigration,
+	}
+}
```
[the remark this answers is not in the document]
droctothorpe: Fixed — added a nil guard at the top of `ConfigFromConfig`:
```go
if cfg == nil {
    return Config{TableName: defaultTableName}
}
```

util/memo/db/migrate.go
```diff
@@ -0,0 +1,26 @@
...
+    key        varchar(256) not null,
+    node_id    text         not null,
+    outputs    text         not null,
+    created_at timestamp    not null,
+    last_hit_at timestamp   not null,
+    primary key (namespace, cache_name, key)
+)`),
+		sqldb.AnsiSQLChange(`create index if not exists imemo_namespace_cache on ` + tableName + ` (namespace, cache_name)`),
+		sqldb.AnsiSQLChange(`create index if not exists imemo_last_hit_at on ` + tableName + ` (last_hit_at)`),
+	})
```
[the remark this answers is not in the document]
droctothorpe: Both issues fixed:

1. **SQL injection** — added `validTableName` regex (`^[A-Za-z0-9_]+$`) validation before any SQL concatenation.
2. **Redundant index** — removed the `imemo_namespace_cache` index on `(namespace, cache_name)` since the primary key already covers it.

Additionally, the migration now uses `sqldb.ByType` to split DDL per database: Postgres uses `"key"` (double-quoted) and `text` for outputs, MySQL uses `` `key` `` (backtick-quoted, since `key` is a reserved word) and `longtext` (to avoid the 64KB `text` limit).

workflow/controller/controller.go
```diff
@@ -721,6 +724,52 @@ func (wfc *WorkflowController) archivedWorkflowGarbageCollector(ctx context.Cont
...
+			logger.Info(ctx, "Performing memoization cache GC")
+			n, err := queries.Prune(ctx, wfc.memoSessionProxy, ttl)
+			if err != nil {
+				logger.WithError(err).Error(ctx, "Failed to prune memoization cache")
+			} else {
+				logger.WithField("deleted", n).Info(ctx, "Memoization cache GC complete")
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
droctothorpe: Both issues fixed:

1. **TTL=0 behavior** — the implementation is correct (`0` → default 90 days, negative → disable). Fixed all documentation to match. 
2. **`cfg`/`queries` outside the loop** — these were already outside the loop (lines 755–756 before the `for`), so no change needed there.

Additionally:
- Added a `periodicity <= 0` guard before `time.NewTicker` to prevent a panic when `MEMO_CACHE_GC_PERIOD=0s`.
- `Prune` now uses `sp.With(ctx, ...)` instead of `sp.Session()` directly, gaining reconnect-on-error.
- `Save` replaced DELETE+INSERT with `ON CONFLICT DO UPDATE` / `ON DUPLICATE KEY UPDATE` to eliminate the concurrent-write race condition.

droctothorpe:
## Manual Validation

Deployed the latest changes to a local Kubernetes cluster (OrbStack, `linux/arm64`) and ran the e2e test suite.

### 1. Controller startup — memoization migration

Controller logs confirm the memoization DB was detected, migrated, and the GC goroutine started:

```
level=INFO msg="Memoization database configuration enabled"
level=INFO msg="Running memoization db migration"
level=INFO msg="apply change" change="create table if not exists memoization_cache (...)"
level=INFO msg=done rowsaffected=0
level=INFO msg="apply change" change="create index if not exists imemo_last_hit_at on memoization_cache (last_hit_at)"
level=INFO msg=done rowsaffected=0
level=INFO msg="Memoization db migration complete"
level=INFO msg="Starting memoization cache GC" ttl=2160h0m0s periodicity=24h0m0s
```

### 2. E2E test — `TestSQLDBMemoize`

```
$ go test -v -tags corefunctional -run TestSQLDBMemoizeSuite ./test/e2e/ -count=1 -timeout 5m

=== RUN   TestSQLDBMemoizeSuite/TestSQLDBMemoize

 ✔ sqldb-memoize-mqkxg   Workflow  10s
 └ ✔ [0]                 StepGroup 10s
 └ ✔ run1                Pod       5s      ← executed container, saved to SQL cache
 └ ✔ [1]                 StepGroup 0s
 └ ✔ run2                Pod       0s      ← cache hit, no container execution

--- PASS: TestSQLDBMemoizeSuite/TestSQLDBMemoize (10.58s)
PASS
ok  	github.com/argoproj/argo-workflows/v4/test/e2e	11.314s
```

The test verifies:
- At least one node **saved** to the cache (run1)
- At least one node **hit** the cache (run2)
- The cache entry exists in the `memoization_cache` Postgres table
- **No ConfigMap** was created (i.e., the controller did not fall back)

### 3. ConfigMap used

```yaml
memoization:
  tableName: memoization_cache
  postgresql:
    host: postgres
    port: 5432
    database: postgres
    userNameSecret:
      name: argo-postgres-config
      key: username
    passwordSecret:
      name: argo-postgres-config
      key: password
```

droctothorpe:
Addressed all feedback. Failures are CI flakiness I think. CC @Joibel.

droctothorpe:
Disregard for now. Still working on this. Need to make sure other step fields are also part of the cache key.

droctothorpe:
# Updated manual validation evidence

# E2E Test Evidence: Memoization Cache Miss → Checksum → Cache Hit

**Date**: 2026-04-19  
**Workflow name**: `memo-e2e-v2-v2mgq`  
**Cluster**: OrbStack (local kubernetes, arm64)  
**Controller image**: `workflow-controller:local` (build from HEAD, `buildDate=2026-04-19T04:24:04Z`)  
**Executor image**: `argoexec:local` (build from HEAD)  
**Cache backend**: PostgreSQL (`memoization_cache` table, host `postgres:5432`)  

---

## Key change validated

This run uses the new **minimal `memoize: {}`** syntax — no `key`, no `cache`, no `maxAge` required.
The controller automatically:
- Derives a cache key from the template name + input parameters
- Routes entries to the `"default"` bucket in the SQL database

---

## Workflow Manifest

```yaml
kind: Workflow
metadata:
  generateName: memo-e2e-v2-
  namespace: argo
  labels:
    test: memo-e2e-v2
spec:
  entrypoint: test-memoization
  arguments:
    parameters:
      - name: message
        value: "hello-memoize"

  templates:
    # Orchestrator: runs the same memoized template twice in sequence.
    # The first call is a cache miss (template executes, checksum generated).
    # The second call is a cache hit (template is skipped, outputs replayed).
    - name: test-memoization
      steps:
        - - name: first-run
            template: compute
            arguments:
              parameters:
                - name: message
                  value: "{{workflow.parameters.message}}"
        - - name: second-run
            template: compute
            arguments:
              parameters:
                - name: message
                  value: "{{workflow.parameters.message}}"

    # Memoized worker template.
    # memoize: {} — no key, cache, or maxAge needed.
    # The controller auto-derives the key and uses the SQL "default" bucket.
    - name: compute
      inputs:
        parameters:
          - name: message
      memoize: {}
      container:
        image: alpine:3.18
        command: [sh, -c]
        args:
          - |
            echo "=== compute running ==="
            echo "input message: {{inputs.parameters.message}}"
            echo "{{inputs.parameters.message}}" > /tmp/result.txt
            echo "timestamp: $(date -u +%Y-%m-%dT%H:%M:%SZ)" >> /tmp/result.txt
            echo "=== done ==="
      outputs:
        parameters:
          - name: result
            valueFrom:
              path: /tmp/result.txt
        artifacts:
          - name: result-file
            path: /tmp/result.txt
            s3:
              bucket: my-bucket
              endpoint: minio:9000
              insecure: true
              accessKeySecret:
                name: my-minio-cred
                key: accesskey
              secretKeySecret:
                name: my-minio-cred
                key: secretkey
```

---

## Workflow Execution Summary

```
Workflow:   memo-e2e-v2-v2mgq
Phase:      Succeeded
Started:    2026-04-19T04:26:24Z
Finished:   2026-04-19T04:26:34Z
Duration:   10s
Progress:   2/2
```

---

## Step 1: first-run — Cache MISS + Checksum Generated

**Node**: `memo-e2e-v2-v2mgq[0].first-run`  
**Node ID**: `memo-e2e-v2-v2mgq-235193569`  
**Pod**: `memo-e2e-v2-v2mgq-compute-235193569` (ran on `orbstack`)  
**Phase**: Succeeded  
**Started**: 2026-04-19T04:26:24Z  
**Finished**: 2026-04-19T04:26:28Z  

### Memoization Status (from workflow node)
```json
{
  "cacheName": "default",
  "hit": false,
  "key": "66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd267d77b3d912857cf"
}
```

- **`hit: false`** — cache was empty; the container ran
- **`cacheName: "default"`** — automatically assigned SQL bucket (no `cache:` field in manifest)
- **`key`** — SHA-256 auto-derived from `template:compute` + `param:message=hello-memoize`

### Outputs (from workflow node)
```json
{
  "exitCode": "0",
  "parameters": [
    {
      "name": "result",
      "value": "hello-memoize\ntimestamp: 2026-04-19T04:26:26Z"
    }
  ],
  "artifacts": [
    {
      "name": "result-file",
      "path": "/tmp/result.txt",
      "checksum": "sha256:ec53fe5b05aefceb93e662796ddd32704c4405b3a45d8b7810147d72719f2e32",
      "s3": {
        "key": "memo-e2e-v2-v2mgq/memo-e2e-v2-v2mgq-compute-235193569/result-file.tgz"
      }
    }
  ]
}
```

- **`checksum: "sha256:ec53fe5b..."`** — SHA-256 of `/tmp/result.txt` computed by `argoexec` at upload time

### Executor (argoexec wait container) Logs

```
time=2026-04-19T04:26:25.884Z level=INFO msg="Starting Workflow Executor" version=latest+unknown.dirty buildDate=2026-04-19T04:24:04Z
time=2026-04-19T04:26:27.897Z level=INFO msg="Main container completed" error=<nil>
time=2026-04-19T04:26:27.897Z level=INFO msg="Saving output parameters"
time=2026-04-19T04:26:27.897Z level=INFO msg="Saving path output parameter" name=result
time=2026-04-19T04:26:27.898Z level=INFO msg="Successfully saved output parameter" name=result
time=2026-04-19T04:26:27.898Z level=INFO msg="Saving output artifacts"
time=2026-04-19T04:26:27.898Z level=INFO msg="Staging artifact" name=result-file
time=2026-04-19T04:26:27.898Z level=INFO msg="Copying from container base image layer" path=/tmp/result.txt localArtPath=/tmp/argo/outputs/artifacts/result-file.tgz
time=2026-04-19T04:26:27.898Z level=INFO msg="S3 Save" key=memo-e2e-v2-v2mgq/memo-e2e-v2-v2mgq-compute-235193569/result-file.tgz path=/tmp/argo/outputs/artifacts/result-file.tgz
time=2026-04-19T04:26:27.999Z level=INFO msg="Save artifact" artifactName=result-file duration=101.222871ms error=<nil>
```

---

## Step 2: second-run — Cache HIT (no pod executed)

**Node**: `memo-e2e-v2-v2mgq[1].second-run`  
**Node ID**: `memo-e2e-v2-v2mgq-3723553210`  
**Pod**: none — node was resolved instantly from cache  
**Phase**: Succeeded  
**Started**: 2026-04-19T04:26:34Z  
**Finished**: 2026-04-19T04:26:34Z (0s — no pod ran)  

### Memoization Status (from workflow node)
```json
{
  "cacheName": "default",
  "hit": true,
  "key": "66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd267d77b3d912857cf"
}
```

- **`hit: true`** — controller found the key in the SQL cache; no pod was scheduled
- **`key`** — identical to first-run (same template name + same inputs → same deterministic hash)

### Outputs replayed from cache (from workflow node)
```json
{
  "exitCode": "0",
  "parameters": [
    {
      "name": "result",
      "value": "hello-memoize\ntimestamp: 2026-04-19T04:26:26Z"
    }
  ],
  "artifacts": [
    {
      "name": "result-file",
      "path": "/tmp/result.txt",
      "checksum": "sha256:ec53fe5b05aefceb93e662796ddd32704c4405b3a45d8b7810147d72719f2e32",
      "s3": {
        "key": "memo-e2e-v2-v2mgq/memo-e2e-v2-v2mgq-compute-235193569/result-file.tgz"
      }
    }
  ]
}
```

- **timestamp unchanged** (`2026-04-19T04:26:26Z`) — proves the second node did NOT re-execute; it returned the cached output
- **`checksum`** identical — persisted through the cache round-trip
- **`s3.key`** points to first-run's artifact — the cache replays the exact same artifact location

---

## PostgreSQL Cache Table

Query run immediately after workflow completion:

```sql
SELECT namespace, cache_name, key, node_id,
       outputs::jsonb->'artifacts'->0->>'checksum' AS checksum,
       created_at, last_hit_at
FROM memoization_cache;
```

```
 namespace | cache_name |                key (SHA-256)                     |           node_id           |              checksum                                                   |         created_at         |        last_hit_at
-----------+------------+--------------------------------------------------+-----------------------------+-------------------------------------------------------------------------+----------------------------+---------------------------
 argo      | default    | 66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd | memo-e2e-v2-v2mgq-235193569 | sha256:ec53fe5b05aefceb93e662796ddd32704c4405b3a45d8b7810147d72719f2e32 | 2026-04-19 04:26:34.142141 | 2026-04-19 04:26:34.142141
(1 row)
```

- **`cache_name: "default"`** — no explicit `cache:` in manifest; controller used the SQL default bucket
- `node_id` = first-run's pod ID (the node whose outputs are replayed)
- `checksum` stored inside the `outputs` JSON blob, round-tripped correctly

---

## Controller Startup Logs (confirming SQL memoization active)

```
time=2026-04-19T04:24:04Z level=INFO msg="Memoization database configuration enabled"
time=2026-04-19T04:24:04Z level=INFO msg="Running memoization db migration"
time=2026-04-19T04:24:04Z level=INFO msg="Memoization db migration complete"
```

---

## Cache Key Derivation

The key `66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd267d77b3d912857cf` is the SHA-256 of:

```
v1
template:compute
param:message=13:hello-memoize
```

- `v1` — format version prefix
- `template:compute` — scopes key to this template
- `param:message=13:hello-memoize` — length-prefixed parameter (13 = len("hello-memoize"))

No `memoize.key` was specified in the workflow spec. The key was auto-derived by the controller.

---

## Summary

| | first-run | second-run |
|---|---|---|
| Cache hit | `false` | `true` |
| Pod executed | ✅ yes (`orbstack`) | ❌ no (skipped) |
| Cache name | `default` (auto) | `default` (auto) |
| Auto-derived key | `66bde39e…` | `66bde39e…` (identical) |
| Output checksum | `sha256:ec53fe5b…` | `sha256:ec53fe5b…` (replayed) |
| Output timestamp | `04:26:26Z` | `04:26:26Z` (replayed, not re-run) |
| Duration | 4s | 0s (instant) |
  

---

## Workflow Manifest

```yaml
kind: Workflow
metadata:
  generateName: memo-e2e-
  namespace: argo
  labels:
    test: memo-e2e
spec:
  entrypoint: test-memoization
  arguments:
    parameters:
      - name: message
        value: "hello-memoize"

  templates:
    # Orchestrator: runs the same memoized template twice in sequence.
    # The first call is a cache miss (template executes, checksum generated).
    # The second call is a cache hit (template is skipped, outputs replayed).
    - name: test-memoization
      steps:
        - - name: first-run
            template: compute
            arguments:
              parameters:
                - name: message
                  value: "{{workflow.parameters.message}}"
        - - name: second-run
            template: compute
            arguments:
              parameters:
                - name: message
                  value: "{{workflow.parameters.message}}"

    # Memoized worker template.
    # memoize.key is intentionally omitted so the controller auto-derives it
    # from: template name ("compute") + input parameter ("message").
    - name: compute
      inputs:
        parameters:
          - name: message
      memoize:
        cache:
          configMap:
            name: memo-e2e-cache
        maxAge: "1h"
      container:
        image: alpine:3.18
        command: [sh, -c]
        args:
          - |
            echo "=== compute running ==="
            echo "input message: {{inputs.parameters.message}}"
            echo "{{inputs.parameters.message}}" > /tmp/result.txt
            echo "timestamp: $(date -u +%Y-%m-%dT%H:%M:%SZ)" >> /tmp/result.txt
            echo "=== done ==="
      outputs:
        parameters:
          - name: result
            valueFrom:
              path: /tmp/result.txt
        artifacts:
          - name: result-file
            path: /tmp/result.txt
            s3:
              bucket: my-bucket
              endpoint: minio:9000
              insecure: true
              accessKeySecret:
                name: my-minio-cred
                key: accesskey
              secretKeySecret:
                name: my-minio-cred
                key: secretkey
```

---

## Workflow Execution Summary

```
Workflow:   memo-e2e-mxt7t
Phase:      Succeeded
Started:    2026-04-19T03:52:02Z
Finished:   2026-04-19T03:52:12Z
Duration:   10s
Progress:   2/2
```

---

## Step 1: first-run — Cache MISS + Checksum Generated

**Node**: `memo-e2e-mxt7t[0].first-run`  
**Node ID**: `memo-e2e-mxt7t-1152369653`  
**Pod**: `memo-e2e-mxt7t-compute-1152369653` (ran on `orbstack`)  
**Phase**: Succeeded  
**Started**: 2026-04-19T03:52:02Z  
**Finished**: 2026-04-19T03:52:05Z  

### Memoization Status (from workflow node)
```json
{
  "cacheName": "memo-e2e-cache",
  "hit": false,
  "key": "66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd267d77b3d912857cf"
}
```

- **`hit: false`** — cache was empty; the container ran
- **`key`** — SHA-256 auto-derived from `template:compute` + `param:message=hello-memoize` (no explicit `memoize.key` set)

### Outputs (from workflow node)
```json
{
  "exitCode": "0",
  "parameters": [
    {
      "name": "result",
      "value": "hello-memoize\ntimestamp: 2026-04-19T03:52:03Z"
    }
  ],
  "artifacts": [
    {
      "name": "result-file",
      "path": "/tmp/result.txt",
      "checksum": "sha256:c30a6cea8867ce43c2cdb23907f31fbd6257e17b3811d88226f86c6cabe972c0",
      "s3": {
        "key": "memo-e2e-mxt7t/memo-e2e-mxt7t-compute-1152369653/result-file.tgz"
      }
    }
  ]
}
```

- **`checksum: "sha256:c30a6cea..."`** — SHA-256 of `/tmp/result.txt` computed by the patched `argoexec` at upload time and stored on the `Artifact` object

### Executor (argoexec wait container) Logs

```
time=2026-04-19T03:52:03.498Z level=INFO msg="Starting Workflow Executor" version=v0.0.0+unknown
time=2026-04-19T03:52:05.509Z level=INFO msg="Main container completed" error=<nil>
time=2026-04-19T03:52:05.509Z level=INFO msg="Saving output parameters"
time=2026-04-19T03:52:05.509Z level=INFO msg="Saving path output parameter" name=result
time=2026-04-19T03:52:05.509Z level=INFO msg="Successfully saved output parameter" name=result
time=2026-04-19T03:52:05.509Z level=INFO msg="Saving output artifacts"
time=2026-04-19T03:52:05.509Z level=INFO msg="Staging artifact" name=result-file
time=2026-04-19T03:52:05.509Z level=INFO msg="Copying from container base image layer" path=/tmp/result.txt localArtPath=/tmp/argo/outputs/artifacts/result-file.tgz
time=2026-04-19T03:52:05.509Z level=INFO msg="S3 Save" path=/tmp/argo/outputs/artifacts/result-file.tgz key=memo-e2e-mxt7t/memo-e2e-mxt7t-compute-1152369653/result-file.tgz
time=2026-04-19T03:52:05.527Z level=INFO msg="Save artifact" artifactName=result-file duration=17.620254ms error=<nil>
time=2026-04-19T03:52:05.527Z level=INFO msg="Successfully saved file" path=/tmp/argo/outputs/artifacts/result-file.tgz
```

The patched argoexec computed `ChecksumPath("/tmp/argo/outputs/artifacts/result-file.tgz")` (SHA-256)
and stored it on `art.Checksum` before writing the TaskResult back to the API server.

---

## Step 2: second-run — Cache HIT (no pod executed)

**Node**: `memo-e2e-mxt7t[1].second-run`  
**Node ID**: `memo-e2e-mxt7t-4142332998`  
**Pod**: none — node was resolved instantly from cache  
**Phase**: Succeeded  
**Started**: 2026-04-19T03:52:12Z  
**Finished**: 2026-04-19T03:52:12Z (0s — no pod ran)  

### Memoization Status (from workflow node)
```json
{
  "cacheName": "memo-e2e-cache",
  "hit": true,
  "key": "66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd267d77b3d912857cf"
}
```

- **`hit: true`** — controller found the key in the SQL cache; no pod was scheduled
- **`key`** — identical to first-run (same template name + same inputs → same deterministic hash)

### Outputs replayed from cache (from workflow node)
```json
{
  "exitCode": "0",
  "parameters": [
    {
      "name": "result",
      "value": "hello-memoize\ntimestamp: 2026-04-19T03:52:03Z"
    }
  ],
  "artifacts": [
    {
      "name": "result-file",
      "path": "/tmp/result.txt",
      "checksum": "sha256:c30a6cea8867ce43c2cdb23907f31fbd6257e17b3811d88226f86c6cabe972c0",
      "s3": {
        "key": "memo-e2e-mxt7t/memo-e2e-mxt7t-compute-1152369653/result-file.tgz"
      }
    }
  ]
}
```

- **timestamp unchanged** (`2026-04-19T03:52:03Z`) — proves the second node did NOT re-execute; it returned the cached output
- **`checksum`** identical — persisted through the cache round-trip
- **`s3.key`** points to first-run's artifact — the cache replays the exact same artifact location

---

## PostgreSQL Cache Table

Query run immediately after workflow completion:

```sql
SELECT namespace, cache_name, key, node_id,
       outputs::jsonb->'artifacts'->0->>'checksum' AS checksum,
       created_at, last_hit_at
FROM memoization_cache;
```

```
 namespace |   cache_name   |                key (SHA-256)                     |         node_id           |              checksum               |         created_at         |        last_hit_at
-----------+----------------+--------------------------------------------------+---------------------------+-------------------------------------+----------------------------+---------------------------
 argo      | memo-e2e-cache | 66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd | memo-e2e-mxt7t-1152369653 | sha256:c30a6cea8867ce43c2cdb2390...  | 2026-04-19 03:52:12.074337 | 2026-04-19 03:52:12.074337
(1 row)
```

- One row — one unique (namespace, cache_name, key) tuple
- `node_id` = first-run's pod ID (the node whose outputs are replayed)
- `checksum` stored inside the `outputs` JSON blob, round-tripped correctly

---

## Controller Startup Logs (confirming SQL memoization active)

```
time=2026-04-19T03:51:26.161Z level=INFO msg="Memoization caches will be garbage-collected if they have not been hit after" gcAfterNotHitDuration=30s
time=2026-04-19T03:51:26.204Z level=INFO msg="Memoization database configuration enabled"
time=2026-04-19T03:51:26.240Z level=INFO msg="Running memoization db migration"
time=2026-04-19T03:51:26.316Z level=INFO msg="apply change" dbType=postgres change="create table if not exists memoization_cache (...)"
time=2026-04-19T03:51:26.363Z level=INFO msg="Memoization db migration complete"
```

---

## Cache Key Derivation

The key `66bde39ee73e41af649c29cb401af496c08ea3f883ec5dd267d77b3d912857cf` is the SHA-256 of:

```
v1
template:compute
param:message=13:hello-memoize
```

- `v1` — format version prefix
- `template:compute` — scopes key to this template
- `param:message=13:hello-memoize` — length-prefixed parameter (13 = len("hello-memoize"))

No `memoize.key` was specified in the workflow spec. The key was auto-derived by the controller from the template name and resolved input parameters.

---

## Summary

| | first-run | second-run |
|---|---|---|
| Cache hit | `false` | `true` |
| Pod executed | ✅ yes (`orbstack`) | ❌ no (skipped) |
| Auto-derived key | `66bde39e…` | `66bde39e…` (identical) |
| Output checksum | `sha256:c30a6cea…` | `sha256:c30a6cea…` (replayed) |
| Output timestamp | `03:52:03Z` | `03:52:03Z` (replayed, not re-run) |
| Duration | 3s | 0s (instant) |

droctothorpe:
Some important design questions that we should probably discuss:

* We see significant pressure on the pod used for workflow archives as workflow quantity and complexity increases. Does it make sense to provide a standalone database for the cache to decouple scaling considerations and reduce the blast radius?
* How do you feel about the blank `memoize: {}` interface? The goal is to make it as turnkey as possible for end users ofc, but maybe there's a better way to design the UX.
* How do you feel about 3 months as the default TTL for cache entries?

docs/workflow-controller-configmap.md
```diff
@@ -369,6 +370,22 @@ SyncConfig contains synchronization configuration for database locks (semaphores
...
+| `SkipMigration`     | `bool`                                                                                                                                                                                                                                                                                   | SkipMigration skips automatic database migration on startup.                                                                                                                                        |
+| `CacheTTL`          | `TTL` (TTL is a time.Duration wrapper that supports human-readable unmarshalling, since time.Duration forces you to specify in millis and does not support days. See https://stackoverflow.com/questions/48050945/how-to-unmarshal-json-into-durations (underlying type: time.Duration)) | CacheTTL is the duration after which cache entries that have not been hit are deleted by the garbage collector. Defaults to "2160h" (90 days). Set to a negative value (e.g., "-1s") to disable GC. |
```
Joibel, line 387: Please don't use time.Duration for new fields as it's not a well known format. Make this CacheTTLSeconds instead.

Joibel (changes requested):

Joibel:
>     * We see significant pressure on the pod used for workflow archives as workflow quantity and complexity increases. Does it make sense to provide a standalone database for the cache to decouple scaling considerations and reduce the blast radius?

We've already got that here, haven't we? Well, you can use the same one, or a different one, depending upon how you configure it.

>     * How do you feel about the blank `memoize: {}` interface? The goal is to make it as turnkey as possible for end users ofc, but maybe there's a better way to design the UX.

You appear to have implemented multiple features here. It's big enough and AI enough as is without automatic key generation - please don't do that in this PR.

>     * How do you feel about 3 months as the default TTL for cache entries?

It's way bigger than I'd expect. We already have maxAge, which isn't perfect. Again, split this into a separate feature if you'd like TTL. It should apply to any backend equally and can be reasoned about separately.

manifests/components/postgres/overlays/workflow-controller-configmap.yaml
```diff
@@ -43,6 +43,18 @@ data:
       passwordSecret:
         name: argo-postgres-config
         key: password
+  memoization: |
```
droctothorpe, line 46: Added memoization to `quick-start-postgres` in order to run the e2e `sqldb_memoize_test.go` test. Let me know if you prefer for that to happen in a dedicated overlay, e.g. `quick-start-postgress-with-caching` or something to that effect.

Joibel: I'd like to e2e test this on both postgres and mysql. Happy to leave it configured here though, but needs duplicating for mysql.

test/e2e/sqldb_memoize_test.go
```diff
@@ -0,0 +1,150 @@
...
+            name: sqldb-memo-cache
+      container:
+        image: argoproj/argosay:v2
+        command: [echo]
+        args: ["{{inputs.parameters.message}}"]
+`, cacheKey, cacheKey)
+}
+
+func (s *SQLDBMemoizeSuite) TestSQLDBMemoize() {
+	if s.Config.Memoization == nil {
```
Joibel: Can we drop this - always test the feature. Dangerous to have the only test for a feature accidentally stop running.

droctothorpe:
# Manual validation evidence

## Validate ConfigMap 

I deployed AWF on my local orbstack-backed K8s cluster with the changes in this PR.

Setup port-forwarding:
```bash
kubectl -n argo port-forward svc/argo-server 2746:2746
```

I added the `create` and `update` verbs to `configmap` resource in the `argo-cluster-role` configmap.

I ran the following e2e test:

```bash
GOFLAGS='-mod=mod' go test ./test/e2e -tags corefunctional -run 'TestFunctionalSuite/TestStepLevelMemoize$' -count=1 -v
```

It succeeded:

```
...
=== RUN   TestFunctionalSuite
=== RUN   TestFunctionalSuite/TestStepLevelMemoize
Submitting workflow  steps-memoize-
Waiting up to 1m0s for workflow with field selector 'metadata.name=steps-memoize-cxvds' and label selector 'workflows.argoproj.io/test'
 ? steps-memoize-cxvds Workflow 0s      

 ● steps-memoize-cxvds   Workflow  0s      
 └ ● steps-memoize-cxvds Steps     0s      
 └ ◷ hello1              Steps     0s      
 └ ● [0]                 StepGroup 0s      
 └ ◷ cache               Pod       0s      
 └ ● [0]                 StepGroup 0s      

 ✔ steps-memoize-cxvds   Workflow  12s     
 └ ✔ steps-memoize-cxvds Steps     12s     
 └ ✔ hello1              Steps     12s     
 └ ✔ [0]                 StepGroup 12s     
 └ ✔ cache               Pod       7s      
 └ ✔ [0]                 StepGroup 12s     
 └ ✔ hello2a             Steps     0s      
 └ ✔ [1]                 StepGroup 0s      

Condition "to be Succeeded" met after 11s
Checking expectation steps-memoize-cxvds
steps-memoize-cxvds : Succeeded 
=== PASS: FunctionalSuite/TestStepLevelMemoize
--- PASS: TestFunctionalSuite (12.24s)
    --- PASS: TestFunctionalSuite/TestStepLevelMemoize (11.95s)
PASS
ok      github.com/argoproj/argo-workflows/v4/test/e2e  14.556s
```

I verified that a CM was created:

```
Name:         my-config-memo-step
Namespace:    argo
Labels:       workflows.argoproj.io/configmap-type=Cache
Annotations:  <none>

Data
====
hello1:
----
{"nodeID":"steps-memoize-cxvds-1782520801","outputs":{"parameters":[{"name":"output","value":"hello1"}]},"creationTimestamp":"2026-05-15T02:07:24Z","lastHitTimestamp":"2026-05-15T02:07:24Z"}


BinaryData
====

Events:  <none>
```

## Validate SQL-backed approach

Now that I've verified the default, CM-based approach works as expected, I moved on to the SQL-based approach.

I enabled the SQL-backed approach:

```bash
kubectl apply --server-side --force-conflicts \
-f manifests/cluster-install-no-crds/workflow-controller-rbac/workflow-controller-clusterrole.yaml

make install PROFILE=postgres

kubectl -n argo scale deploy/workflow-controller deploy/argo-server --replicas=1
kubectl -n argo set env deploy/argo-server CI_ONLY_DISABLE_ARTIFACT_SERVER_CHECKS=true

kubectl -n argo rollout status deploy/postgres --timeout=300s
kubectl -n argo rollout status deploy/workflow-controller --timeout=300s
kubectl -n argo rollout status deploy/argo-server --timeout=300s
```

I validated that the sql memoization was enabled in the config by running:

```bash
kubectl -n argo get cm workflow-controller-configmap -o jsonpath='{.data.memoization}'; echo
kubectl -n argo logs deploy/workflow-controller --tail=100 | egrep 'Memoization database configuration enabled|Running memoization db migration'
```

The output was:

```
❯ kubectl -n argo get cm workflow-controller-configmap -o jsonpath='{.data.memoization}'; echo

tableName: cache_entries
postgresql:
  host: postgres
  port: 5432
  database: postgres
  userNameSecret:
    name: argo-postgres-config
    key: username
  passwordSecret:
    name: argo-postgres-config
    key: password


❯ kubectl -n argo logs deploy/workflow-controller --tail=100 | egrep 'Memoization database configuration enabled|Running memoization db migration'
time=2026-05-15T02:19:06.924Z level=INFO msg="Memoization database configuration enabled" component=configmap_watcher
time=2026-05-15T02:19:06.924Z level=INFO msg="Running memoization db migration" component=configmap_watcher
```

I port-forwarded the postgres service to local:

```bash
kubectl -n argo port-forward svc/postgres 5432:5432 --address localhost
```

I then kicked off the same e2e workflow as before:

```bash
GOFLAGS='-mod=mod' go test ./test/e2e -tags sqldbmemoize -run '^TestSQLDBMemoizeSuite$' -count=1 -v
```

It succeeded:

```
❯ argo-workflows main* ⇡ GOFLAGS='-mod=mod' go test ./test/e2e -tags sqldbmemoize -run '^TestSQLDBMemoizeSuite$' -count=1 -v
=== RUN   TestSQLDBMemoizeSuite
=== RUN   TestSQLDBMemoizeSuite/TestSQLDBMemoize
Submitting workflow  sqldb-memoize-
Waiting up to 1m0s for workflow with field selector 'metadata.name=sqldb-memoize-cjrfk' and label selector 'workflows.argoproj.io/test'
 ? sqldb-memoize-cjrfk Workflow 0s      

 ● sqldb-memoize-cjrfk   Workflow  0s      
 └ ● sqldb-memoize-cjrfk Steps     0s      
 └ ◷ run1                Pod       0s      
 └ ● [0]                 StepGroup 0s      

 ✔ sqldb-memoize-cjrfk   Workflow  10s     
 └ ✔ sqldb-memoize-cjrfk Steps     10s     
 └ ✔ run1                Pod       6s      
 └ ✔ [0]                 StepGroup 10s     
 └ ✔ [1]                 StepGroup 0s      
 └ ✔ run2                Pod       0s      

 ✔ sqldb-memoize-cjrfk   Workflow  10s     
 └ ✔ sqldb-memoize-cjrfk Steps     10s     
 └ ✔ run1                Pod       6s      
 └ ✔ [0]                 StepGroup 10s     
 └ ✔ [1]                 StepGroup 0s      
 └ ✔ run2                Pod       0s      

Condition "to be Succeeded" met after 10s
Checking expectation sqldb-memoize-cjrfk
sqldb-memoize-cjrfk : Succeeded 
=== PASS: SQLDBMemoizeSuite/TestSQLDBMemoize
--- PASS: TestSQLDBMemoizeSuite (12.18s)
    --- PASS: TestSQLDBMemoizeSuite/TestSQLDBMemoize (12.02s)
PASS
ok      github.com/argoproj/argo-workflows/v4/test/e2e  12.817s
```

I ran the following command to verify that the cache key is stored in the postgres database:

```bash
kubectl -n argo exec svc/postgres -- \
   psql -U postgres -d postgres \
   -c "SELECT namespace, cache_name, cache_key, created_at, expires_at FROM cache_entries ORDER BY created_at DESC LIMIT 20;"
```

Here's the output:

```
 namespace |    cache_name    |            cache_key            |        created_at         |        expires_at         
-----------+------------------+---------------------------------+---------------------------+---------------------------
 argo      | sqldb-memo-cache | hello-sqldb-1778812065286276000 | 2026-05-15 02:27:55.93877 | 2026-05-15 02:37:55.93877
(1 row)
```

droctothorpe:
Test failures were just flakes.

workflow/controller/cache/cache.go
```diff
@@ -51,35 +101,56 @@ func (e *Entry) GetOutputsWithMaxAge(maxAge time.Duration) (*wfv1.Outputs, bool)
...
+func (cf *cacheFactory) SetQueries(q memodb.MemoizationDB) {
+	cf.lock.Lock()
+	defer cf.lock.Unlock()
+	cf.queries = q
+	cf.caches = make(map[string]MemoizationCache)
+}
+
+// GetCache returns a cache scoped to the given workflow namespace if it exists and creates it
+// otherwise.
+func (cf *cacheFactory) GetCache(ctx context.Context, ct Type, namespace, name string) MemoizationCache {
```
Joibel, line 144: Caches have moved from controller namespace to workflow namespace, which breaks a decent number of things and is an undocumented and undesirable change.

workflow/controller/config.go
```diff
@@ -78,6 +104,42 @@ func (wfc *WorkflowController) updateConfig(ctx context.Context) error {
 		logger.Info(ctx, "Persistence configuration disabled")
 	}
 
+	memoCfg := wfc.Config.Memoization
+	if memoCfg != nil {
+		logger.Info(ctx, "Memoization database configuration enabled")
+		sessionProxy := wfc.memoSessionProxy
+		if sessionProxy == nil {
```
Joibel, line 111: I think if we modify the session config we don't fully reinitialise the connection here. Persistence does the right thing (L81).

workflow/controller/cache/sqldb_cache.go
```diff
@@ -0,0 +1,94 @@
...
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	queries := memodb.NullMemoizationDB
+	if c.getQueries != nil {
+		if q := c.getQueries(); q != nil {
+			queries = q
+		}
+	}
+	return fn(queries)
```
Joibel, line 46: We still hold the mutex for the actual query here, we probably shouldn't

workflow/controller/cache_test.go
```diff
@@ -104,3 +104,15 @@ func TestConfigMapCacheSave(t *testing.T) {
 	wfv1.MustUnmarshal([]byte(cm.Data["hi-there-world"]), &entry)
 	assert.Equal(t, entry.LastHitTimestamp.Time, entry.CreationTimestamp.Time)
 }
+
+func TestConfigMapCacheSaveIgnoresInvalidDefaultMaxAge(t *testing.T) {
```
Joibel, line 108: I don't think this tests anything useful. We're proving that a badly configured DEFAULT_MAX_AGE allows `Save`? Why do we want this behaviour?

workflow/controller/operator_test.go
```diff
@@ -6389,11 +6390,33 @@ func TestConfigMapCacheLoadOperateMaxAge(t *testing.T) {
 	}
 }
 
+func TestUnavailableSQLMemoizationBackendTreatsLookupAsCacheMiss(t *testing.T) {
+	wf := wfv1.MustUnmarshalWorkflow(workflowCachedMaxAge)
+	ctx := logging.TestContext(t.Context())
+	cancel, controller := newController(ctx)
```
Joibel, line 6396: We've built a controller without SQL memoization backend. This would pass even without the rest of the code.

workflow/controller/operator.go
```diff
@@ -2808,14 +2829,19 @@ func (woc *wfOperationCtx) initializeExecutableNode(ctx context.Context, nodeNam
 		node.Inputs = executeTmpl.Inputs.DeepCopy()
 	}
 
-	// Set the MemoizationStatus
-	if node.MemoizationStatus == nil && executeTmpl.Memoize != nil {
-		memoizationStatus := &wfv1.MemoizationStatus{
+	// Set the MemoizationStatus only when the user explicitly supplied a key.
+	// When key is auto-derived, executeTemplate sets MemoizationStatus with the
```
Joibel, line 2833: I think this is a legacy comment and should be removed - auto-deriving keys has been removed

workflow/controller/config.go
```diff
@@ -78,6 +104,42 @@ func (wfc *WorkflowController) updateConfig(ctx context.Context) error {
...
+			sessionProxy = memoSessionProxyFromConfig(ctx, wfc.kubeclientset, wfc.namespace, memoCfg)
+			if sessionProxy == nil {
+				logger.Warn(ctx, "Memoization database unavailable; memoization disabled")
+				wfc.resetMemoizationBackend(ctx, nil, memodb.NullMemoizationDB)
+				goto memoizationConfigured
```
Joibel, line 116: I don't like gotos much.

workflow/controller/config.go
```diff
@@ -78,6 +104,42 @@ func (wfc *WorkflowController) updateConfig(ctx context.Context) error {
...
 	}
 
+	memoCfg := wfc.Config.Memoization
+	if memoCfg != nil {
+		logger.Info(ctx, "Memoization database configuration enabled")
+		sessionProxy := wfc.memoSessionProxy
+		if sessionProxy == nil {
+			sessionProxy = memoSessionProxyFromConfig(ctx, wfc.kubeclientset, wfc.namespace, memoCfg)
+			if sessionProxy == nil {
+				logger.Warn(ctx, "Memoization database unavailable; memoization disabled")
```
Joibel, line 114: In a startup connection failure we semi-silently carry on not memoizing to a database until controller restart. I think we should hard fail when we can't setup memoization having been asked to use it - same as persistence

docs/environment-variables.md
```diff
@@ -40,6 +41,7 @@ This document outlines environment variables that can be used to customize behav
...
 | `MAX_OPERATION_TIME`                     | `time.Duration`     | `30s`                                                                                       | The maximum time a workflow operation is allowed to run for before re-queuing the workflow onto the work queue.                                                                                                                                                          |
+| `MEMO_CACHE_GC_PERIOD`                   | `time.Duration`     | `24h`                                                                                       | How often the SQL-backed memoization cache garbage collector runs to prune entries that have exceeded their TTL.                                                                                                                                                         |
```
Joibel, line 44: Should document `0` special behaviour

util/memo/db/queries.go
```diff
@@ -0,0 +1,181 @@
...
+	now := time.Now().UTC()
+	expiresAt := now.Add(time.Duration(maxAgeSeconds) * time.Second)
+	record := &CacheRecord{
+		Namespace: namespace,
+		CacheName: cacheName,
+		CacheKey:  cacheKey,
+		NodeID:    nodeID,
+		Outputs:   string(outputsJSON),
+		CreatedAt: now,
+		ExpiresAt: expiresAt,
```
Joibel, line 176: TTL is determined at `Save` for DB, at `Load` for Configmap. I think that's fine, but can we document the difference.

workflow/controller/dag.go
```diff
@@ -379,11 +378,17 @@ func (woc *wfOperationCtx) executeDAG(ctx context.Context, nodeName string, tmpl
 		woc.wf.Status.Nodes.Set(ctx, node.ID, *node)
 	}
 	if node.MemoizationStatus != nil {
-		c := woc.controller.cacheFactory.GetCache(controllercache.ConfigMapCache, node.MemoizationStatus.CacheName)
-		saveErr := c.Save(ctx, node.MemoizationStatus.Key, node.ID, node.Outputs)
-		if saveErr != nil {
-			woc.log.WithField("nodeID", node.ID).WithError(saveErr).Error(ctx, "Failed to save node outputs to cache")
-			node.Phase = wfv1.NodeError
+		c := woc.controller.getMemoizationCache(ctx, woc.wf.Namespace, node.MemoizationStatus.CacheName)
```
Joibel, line 381: DRY. Can we make this sequence a common op on wfOperationCtx, shared with steps.go, taskset.go, and operator.go

workflow/controller/cache/cache.go
```diff
@@ -51,35 +101,56 @@ func (e *Entry) GetOutputsWithMaxAge(maxAge time.Duration) (*wfv1.Outputs, bool)
...
 }
 
 type Type string
 
 const (
-	// Only config maps are currently supported for caching
+	// ConfigMapCache is a cache type identifier used as a key prefix in the cache map.
+	// When a MemoizationDB is configured, SQL-backed memoization semantics are used instead.
 	ConfigMapCache Type = "ConfigMapCache"
 )
```
Joibel, line 129: I suspect no human looked at this, but it's very odd to not have a SQL entry in this list and use it.

If we're really going without it, we should drop it as it's confusing and wasteful to retain.

workflow/controller/operator.go
```diff
@@ -2808,14 +2829,19 @@ func (woc *wfOperationCtx) initializeExecutableNode(ctx context.Context, nodeNam
...
+	// computed effectiveKey before reaching this point.
+	if node.MemoizationStatus == nil && executeTmpl.Memoize != nil && executeTmpl.Memoize.Key != "" {
+		cacheName := ""
+		if executeTmpl.Memoize.Cache != nil {
```
Joibel, line 2837: Guard `executeTmpl.Memoize.Cache.ConfigMap != nil` too?

Joibel (changes requested):

PR #15953, opened 2026-04-17
chore(ui): replace `<pre>` with `<p><code>` in plugin list zero state

nakatani-yo:
### Motivation

The plugin list zero-state used a `<pre>` tag for the `kubectl` command, which renders as a preformatted block. It is displayed as shown in Fig. 1. As shown in Fig. 2, the `curl` command is rendered as inline code when "Nothing to show" is displayed on the Workflow Event Bindings screen. I would like the `kubectl` command to be rendered similarly.

image

- Fig. 1. Plugins

image

- Fig. 2. Workflow Event Bindings

### Modifications

Replaced `<pre>` with `<p><code>` in `ui/src/plugins/plugin-list.tsx` so the command renders as inline code within a paragraph, matching the style of the rest of the zero-state content.

### Verification

Visually inspected the plugin list page with no plugins configured and confirmed the command renders correctly as inline code.

image

### Documentation

No documentation changes needed — this is a minor UI markup fix.


Joibel:
What do you see before your change - I see 
image
which looks fine.

nakatani-yo:
@Joibel 
Thank you very much for your review.
I've corrected the "Motivation" section as it was unclear.

Joibel (approved):
Thanks

PR #16021, opened 2026-04-23
fix(ui): respect target field in workflow-list scope links

### LINKED ISSUE #16020: `workflow-list` scope links ignore `target: _blank`
### Pre-requisites

When configuring a `links` entry in `workflow-controller-configmap` with `scope: workflow-list` and `target: _blank`, the link opens in the **same tab** instead of a new tab. The `target` field is completely ignored.

**Expected:** Clicking the link opens a new browser tab (or behaves according to the configured `target` value).

**Actual:** The link always navigates in the current tab, regardless of the `target` field.

### Version(s)

`:latest`

```YAML
apiVersion: v1
kind: ConfigMap
metadata:
  name: workflow-controller-configmap
  namespace: argo
data:
  links: |
    - name: Examples
      scope: workflow-list
      url: https://github.com/argoproj/argo-workflows/tree/main/examples
      target: _blank
```

**Steps to reproduce:**
1. Apply the configmap above
2. Open the Argo Workflows UI and navigate to the Workflows list page
3. Click the "Examples" link in the action menu (top right)
4. Observe: navigates in the same tab instead of opening a new tab

```text
```

```text
```

panicboat:
Fixes #16020

### Motivation

When configuring a `links` entry in `workflow-controller-configmap` with `scope: workflow-list` and `target: _blank`, the link opens in the same tab instead of a new tab. The `target` field is silently ignored.

`workflow-list` scope links were implemented using `window.location.href = link.url`, which always navigates in the current tab regardless of the `target` field. The `target` field was introduced in v4.0.0 for other scopes, but `workflow-list` was not updated at that time.

### Modifications

Updated `workflow-list` scope links to use `openLinkWithKey()`, consistent with how other scopes handle the `target` field.

```diff
- action: () => (window.location.href = link.url)
+ action: () => openLinkWithKey(link.url, link.target)
```

### Verification

Applied the following configmap and confirmed that clicking the link in the action menu on the Workflows list page opens in a new tab:

```yaml
  links: |
    - name: Examples
      scope: workflow-list
      url: https://github.com/argoproj/argo-workflows/tree/main/examples
      target: _blank
```

### Documentation

No documentation changes needed. This is a bug fix that makes the `target` field behave as already documented.

### AI

Used Claude (claude-sonnet-4-6) to investigate.

### DIFF
--- a/ui/src/workflows/components/workflows-list/workflows-list.tsx
+++ b/ui/src/workflows/components/workflows-list/workflows-list.tsx
@@ -8,6 +8,7 @@ import {uiUrl} from '../../../shared/base';
 import {CostOptimisationNudge} from '../../../shared/components/cost-optimisation-nudge';
 import {ErrorNotice} from '../../../shared/components/error-notice';
 import {ExampleManifests} from '../../../shared/components/example-manifests';
+import {openLinkWithKey} from '../../../shared/components/links';
 import {Loading} from '../../../shared/components/loading';
 import {PaginationPanel} from '../../../shared/components/pagination-panel';
 import {TimestampSwitch} from '../../../shared/components/timestamp';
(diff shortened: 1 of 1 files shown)


Joibel (approved):
Thanks for fixing

Issue #10, opened 2017-08-23
Minion-manager should revert to correct spot-instance configuration after upgrades


shrinandj:
It's been reported that after upgrades, the minions are which are switched to on-demand nodes are not getting switched back to spot instances as per the configuration.

PR #16031, opened 2026-04-23
fix(emissary): prevent output path traversal in saveParameter and saveArtifact

jayantkamble10000:
## Summary

Both `saveParameter` and `saveArtifact` built destination paths without
enforcing that the result stays inside `/var/run/argo/outputs/`.

`saveParameter` used raw string concatenation:
```go
// vulnerable
dstPath := varRunArgo + "/outputs/parameters/" + srcPath
```

A workflow spec with valueFrom.path: "../../ctr/containerB/exitcode"
resolves to /var/run/argo/ctr/containerB/exitcode — a world-writable
directory (created 0o777 by design for non-root containers) — letting
one container in a ContainerSet falsify another container's exit code
and bypass DAG dependency checks.

saveArtifact had the same issue: a traversal path resolves outside outputs/artifacts/ regardless of the .tgz suffix appended to the filename

## Fix

Add a filepath.Join + strings.HasPrefix bounds check in both
functions. The check is placed before os.Open/os.Stat so traversal
paths are rejected even when the source file does not yet exist.

## Tests

Added TestSaveParameterPathTraversal and TestSaveArtifactPathTraversal
covering:

- Legitimate absolute paths → allowed
- ../../argoexec traversal → blocked
- ../../ctr/sidecar/exitcode traversal → blocked
- ../template traversal → blocked

Fixes https://github.com/argoproj/argo-workflows/security/advisories/GHSA-r9w2-rp4m-gwr9

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how artifacts and parameters are persisted to prevent path traversal and symlink escape from writing outside the expected output directories.
  * Optional outputs that are missing are now safely skipped.
* **Tests**
  * Added coverage for path traversal and symlink-escape scenarios for both artifacts and parameters, verifying allowed in-tree mappings and rejecting unsafe inputs.

### DIFF
--- a/cmd/argoexec/commands/emissary.go
+++ b/cmd/argoexec/commands/emissary.go
@@ -401,20 +401,33 @@ func saveArtifact(ctx context.Context, srcPath string) error {
 		logger.WithField("srcPath", srcPath).Info(ctx, "no need to save artifact - on overlapping volume")
 		return nil
 	}
+	// Normalize absolute paths (e.g. /tmp/artifact → tmp/artifact) so that
+	// filepath.IsLocal can validate the result, and os.Root.Create can safely
+	// write it within the outputs directory.
+	relSrc := strings.TrimPrefix(strings.TrimSuffix(srcPath, "/"), "/")
+	dstRel := relSrc + ".tgz"
+	if !filepath.IsLocal(dstRel) {
+		return fmt.Errorf("path traversal in output artifact path %q", srcPath)
+	}
+	artifactsDir := filepath.Join(varRunArgo, "outputs/artifacts")
+	dstPath := filepath.Join(artifactsDir, dstRel)
 	if _, err := os.Stat(srcPath); os.IsNotExist(err) { // might be optional, so we ignore
 		logger.WithField("srcPath", srcPath).WithError(err).Warn(ctx, "cannot save artifact")
 		return nil
 	}
-	dstPath := filepath.Join(varRunArgo, "/outputs/artifacts/", strings.TrimSuffix(srcPath, "/")+".tgz")
 	logger.WithFields(logging.Fields{
 		"src": srcPath,
 		"dst": dstPath,
 	}).Info(ctx, "saving artifact")
-	z := filepath.Dir(dstPath)
-	if err := os.MkdirAll(z, 0o755); err != nil { // chmod rwxr-xr-x
-		return fmt.Errorf("failed to create directory %s: %w", z, err)
+	if err := os.MkdirAll(filepath.Dir(dstPath), 0o755); err != nil { // chmod rwxr-xr-x
+		return fmt.Errorf("failed to create directory %s: %w", filepath.Dir(dstPath), err)
 	}
-	dst, err := os.Create(dstPath)
+	root, err := os.OpenRoot(artifactsDir)
+	if err != nil {
+		return fmt.Errorf("failed to open artifacts output root: %w", err)
+	}
+	defer root.Close()
+	dst, err := root.Create(dstRel)
 	if err != nil {
 		return fmt.Errorf("failed to create destination %s: %w", dstPath, err)
 	}
@@ -435,6 +448,15 @@ func saveParameter(ctx context.Context, srcPath string) error {
 		logger.WithField("src", srcPath).Info(ctx, "no need to save parameter - on overlapping volume")
 		return nil
 	}
+	// Normalize absolute paths (e.g. /tmp/parameter → tmp/parameter) so that
+	// filepath.IsLocal can validate the result, and os.Root.Create can safely
+	// write it within the outputs directory.
+	relPath := strings.TrimPrefix(srcPath, "/")
+	if !filepath.IsLocal(relPath) {
+		return fmt.Errorf("path traversal in output parameter path %q", srcPath)
+	}
+	parametersDir := filepath.Join(varRunArgo, "outputs/parameters")
+	dstPath := filepath.Join(parametersDir, relPath)
 	src, err := os.Open(filepath.Clean(srcPath))
 	if os.IsNotExist(err) { // might be optional, so we ignore
 		logger.WithField("src", srcPath).WithError(err).Error(ctx, "cannot save parameter, does not exist")
@@ -444,16 +466,19 @@ func saveParameter(ctx context.Context, srcPath string) error {
 		return fmt.Errorf("failed to open %s: %w", srcPath, err)
 	}
 	defer func() { _ = src.Close() }()
-	dstPath := varRunArgo + "/outputs/parameters/" + srcPath
 	logger.WithFields(logging.Fields{
 		"src": srcPath,
 		"dst": dstPath,
 	}).Info(ctx, "saving parameter")
-	z := filepath.Dir(dstPath)
-	if mkdirErr := os.MkdirAll(z, 0o755); mkdirErr != nil { // chmod rwxr-xr-x
-		return fmt.Errorf("failed to create directory %s: %w", z, mkdirErr)
+	if mkdirErr := os.MkdirAll(filepath.Dir(dstPath), 0o755); mkdirErr != nil { // chmod rwxr-xr-x
+		return fmt.Errorf("failed to create directory %s: %w", filepath.Dir(dstPath), mkdirErr)
 	}
-	dst, err := os.Create(dstPath)
+	root, err := os.OpenRoot(parametersDir)
+	if err != nil {
+		return fmt.Errorf("failed to open parameters output root: %w", err)
+	}
+	defer root.Close()
+	dst, err := root.Create(relPath)
 	if err != nil {
 		return fmt.Errorf("failed to create %s: %w", srcPath, err)
 	}
--- a/cmd/argoexec/commands/emissary.go
+++ b/cmd/argoexec/commands/emissary.go
@@ -401,32 +401,35 @@ func saveArtifact(ctx context.Context, srcPath string) error {
 		logger.WithField("srcPath", srcPath).Info(ctx, "no need to save artifact - on overlapping volume")
 		return nil
 	}
-	// Normalize absolute paths (e.g. /tmp/artifact → tmp/artifact) so that
-	// filepath.IsLocal can validate the result, and os.Root.Create can safely
-	// write it within the outputs directory.
-	relSrc := strings.TrimPrefix(strings.TrimSuffix(srcPath, "/"), "/")
-	dstRel := relSrc + ".tgz"
-	if !filepath.IsLocal(dstRel) {
+	// Map the (possibly absolute) source path to a path relative to the outputs
+	// directory, e.g. /tmp/artifact -> tmp/artifact. TrimLeft drops any leading
+	// slashes (including "//"); Clean collapses any in-bounds "..". IsLocal then
+	// rejects anything that would still escape (leading "..", absolute paths).
+	relSrc := filepath.Clean(strings.TrimLeft(strings.TrimSuffix(srcPath, "/"), "/"))
+	if !filepath.IsLocal(relSrc) {
 		return fmt.Errorf("path traversal in output artifact path %q", srcPath)
 	}
-	artifactsDir := filepath.Join(varRunArgo, "outputs/artifacts")
-	dstPath := filepath.Join(artifactsDir, dstRel)
 	if _, err := os.Stat(srcPath); os.IsNotExist(err) { // might be optional, so we ignore
 		logger.WithField("srcPath", srcPath).WithError(err).Warn(ctx, "cannot save artifact")
 		return nil
 	}
+	// dstRel is relative to varRunArgo; both the directory and the file are
+	// created through a single os.Root opened at that trusted base, so a
+	// symlinked path component cannot redirect the write outside the tree.
+	dstRel := filepath.Join("outputs/artifacts", relSrc+".tgz")
+	dstPath := filepath.Join(varRunArgo, dstRel)
 	logger.WithFields(logging.Fields{
 		"src": srcPath,
 		"dst": dstPath,
 	}).Info(ctx, "saving artifact")
-	if err := os.MkdirAll(filepath.Dir(dstPath), 0o755); err != nil { // chmod rwxr-xr-x
+	root, err := os.OpenRoot(varRunArgo)
+	if err != nil {
+		return fmt.Errorf("failed to open output root %s: %w", varRunArgo, err)
+	}
+	defer func() { _ = root.Close() }()
+	if err = root.MkdirAll(filepath.Dir(dstRel), 0o755); err != nil { // chmod rwxr-xr-x
 		return fmt.Errorf("failed to create directory %s: %w", filepath.Dir(dstPath), err)
 	}
-	root, err := os.OpenRoot(artifactsDir)
-	if err != nil {
-		return fmt.Errorf("failed to open artifacts output root: %w", err)
-	}
-	defer root.Close()
 	dst, err := root.Create(dstRel)
 	if err != nil {
 		return fmt.Errorf("failed to create destination %s: %w", dstPath, err)
@@ -448,15 +451,19 @@ func saveParameter(ctx context.Context, srcPath string) error {
 		logger.WithField("src", srcPath).Info(ctx, "no need to save parameter - on overlapping volume")
 		return nil
 	}
-	// Normalize absolute paths (e.g. /tmp/parameter → tmp/parameter) so that
-	// filepath.IsLocal can validate the result, and os.Root.Create can safely
-	// write it within the outputs directory.
-	relPath := strings.TrimPrefix(srcPath, "/")
-	if !filepath.IsLocal(relPath) {
+	// Map the (possibly absolute) source path to a path relative to the outputs
+	// directory, e.g. /tmp/parameter -> tmp/parameter. TrimLeft drops any leading
+	// slashes (including "//"); Clean collapses any in-bounds "..". IsLocal then
+	// rejects anything that would still escape (leading "..", absolute paths).
+	relSrc := filepath.Clean(strings.TrimLeft(srcPath, "/"))
+	if !filepath.IsLocal(relSrc) {
 		return fmt.Errorf("path traversal in output parameter path %q", srcPath)
 	}
-	parametersDir := filepath.Join(varRunArgo, "outputs/parameters")
-	dstPath := filepath.Join(parametersDir, relPath)
+	// dstRel is relative to varRunArgo; both the directory and the file are
+	// created through a single os.Root opened at that trusted base, so a
+	// symlinked path component cannot redirect the write outside the tree.
+	dstRel := filepath.Join("outputs/parameters", relSrc)
+	dstPath := filepath.Join(varRunArgo, dstRel)
 	src, err := os.Open(filepath.Clean(srcPath))
 	if os.IsNotExist(err) { // might be optional, so we ignore
 		logger.WithField("src", srcPath).WithError(err).Error(ctx, "cannot save parameter, does not exist")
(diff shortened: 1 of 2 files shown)


isubasinghe:
Hey @jayantkamble10000 could you please sign off your commit?

jayantkamble10000:
Please check its done

cmd/argoexec/commands/emissary.go
```diff
@@ -401,11 +401,15 @@ func saveArtifact(ctx context.Context, srcPath string) error {
 		logger.WithField("srcPath", srcPath).Info(ctx, "no need to save artifact - on overlapping volume")
 		return nil
 	}
+	dstPath := filepath.Join(varRunArgo, "outputs/artifacts", strings.TrimSuffix(srcPath, "/")+".tgz")
+	safeArtifactBase := filepath.Clean(filepath.Join(varRunArgo, "outputs/artifacts")) + string(os.PathSeparator)
+	if !strings.HasPrefix(filepath.Clean(dstPath)+string(os.PathSeparator), safeArtifactBase) {
+		return fmt.Errorf("path traversal in output artifact path %q", srcPath)
+	}
```
isubasinghe: if !filepath.IsLocal(srcPath)  is much cleaner

isubasinghe (changes requested):
https://go.dev/blog/osroot please take a read here

jayantkamble10000:
Done — replaced strings.
HasPrefix bounds check with filepath.IsLocal(srcPath) in both functions as suggested. Also updated the test from LegitimateAbsolutePath to LegitimateRelativePath since filepath.IsLocal correctly rejects absolute paths (which is the right behaviour — output paths should be relative to the container's workingDir). 
All tests pass.

cmd/argoexec/commands/emissary.go
jayantkamble10000: Done — replaced strings.HasPrefix bounds check with filepath.IsLocal(srcPath) in both functions as suggested. Also updated the test from LegitimateAbsolutePath to LegitimateRelativePath since filepath.IsLocal correctly rejects absolute paths (which is the right behaviour — output paths should be relative to the container's workingDir). All tests pass.

jayantkamble10000:
Updated as suggested — switched to \`os.OpenRoot\` for creating the destination file so the OS enforces the directory boundary (also protects against symlink escapes, which a pure lexical check can't catch). Combined with \`filepath.IsLocal\` for an early, clear error before the file-create attempt.

Regarding the two failing E2E checks: both failures are pre-existing flaky tests unrelated to this change:
- Previous run: \`RetryTestSuite/TestWorkflowTemplateWithRetryStrategyInContainerSet\` (the \`min\` variant of the same suite passed)
- This run: \`ArgoServerSuite/TestWorkflowArchiveServiceList\` and \`TestWorkflowServiceListArchived\` (MySQL-specific; the identical \`postgres\` variant passed)

Also rebased onto the latest \`main\` — branch is now a single clean commit.

jayantkamble10000:
@isubasinghe All 40 checks are now passing (including DCO). I believe I've addressed all the feedback — switched to `os.OpenRoot` as suggested and rebased to a single clean commit. Would you be able to take another look when you get a chance? Happy to make any further changes.

jayantkamble10000:
Done @isubasinghe

jayantkamble10000 (commented):
@isubasinghe May i please get an update?

jayantkamble10000:
Gentle nudge — this PR has been ready to merge for ~7 weeks and is blocked only by a stale review.

**Summary for reviewers:**
- This is a path-traversal (CWE-22) security fix in `saveParameter`/`saveArtifact`.
- @isubasinghe requested using `os.OpenRoot` (per https://go.dev/blog/osroot) — that change was implemented on May 7, so the requested change is complete.
- All 40 CI checks pass (including DCO), it's a single clean commit, and it's mergeable apart from the standing "changes requested" review.

@isubasinghe — would you be able to re-review or dismiss the now-addressed review when you have a moment? And if you're tied up, could one of the owners/approvers help unblock this — @joibel @terrytangyuan @alexmt @jessesuen @juliev0 @masonm @tczhao? Happy to rebase or make any further changes. Thanks all! 🙏

isubasinghe:
Claude output:
```console
Review complete. The workflow ran 47 agents (10 finders → 34 candidates → independent verify pass → 13 refuted, 12 reported). I've grounded the top findings against the actual code. Here they are, most-severe first.

What the branch does

saveArtifact/saveParameter in cmd/argoexec/commands/emissary.go now normalize the path, gate it through filepath.IsLocal, and write via os.OpenRoot+root.Create instead of os.Create. Good intent. But the sandbox is applied to only one of the two filesystem operations, and the normalization regresses several legitimate paths.

---
1. 🔴 Internal .. breaks legitimate output collection (regression, confirmed + reproduced)

emissary.go:413/430 (artifact), :459/481 (parameter)

filepath.Join (→ dstPath, used by MkdirAll) cleans the path; root.Create gets the uncleaned dstRel/relPath. filepath.IsLocal("a/../b") is true, so it passes the guard — but MkdirAll creates the cleaned dir (b's parent) while root.Create("a/../b") resolves component-by-component, hits the never-created a, and fails no such file or directory. Any output path with an in-bounds .. (e.g. /out/../result) now fails the whole step. Worked on main.

Fix: feed root.Create the same cleaned relative path MkdirAll uses — filepath.Clean(dstRel) / filepath.Clean(relPath).

2. 🟠 Sandbox is bypassed for directory creation and for the base dir itself (security, confirmed)

emissary.go:422/473 (os.MkdirAll), :425/476 (os.OpenRoot)

This is a path-traversal hardening PR, but two operations escape the sandbox:
- os.MkdirAll(filepath.Dir(dstPath)) runs on the raw joined path before and outside the root — it follows symlinks. A main container that plants a symlink as a path component under outputs/artifacts gets attacker-named dirs created outside the tree (then root.Create errors, so the output is also lost).
- os.OpenRoot(artifactsDir) follows a symlink in the base name itself — if outputs/artifacts is a symlink, the whole root rebases onto the target.

(Both assume the main container can write into the outputs tree before the post-exec save — the exact threat os.Root was added to stop.)

Fix (collapses #1 + #2): Go is 1.26.1, so os.Root.MkdirAll exists. Open the root once at the trusted fixed base (varRunArgo), then do root.MkdirAll(dir) + root.Create(rel) with a cleaned relative path under outputs/artifacts. Everything stays sandboxed and cleaned.

3. 🟠 Leading // now rejected (regression, confirmed)

emissary.go:454 / :407

strings.TrimPrefix(srcPath, "/") strips only one slash. A POSIX-valid //tmp/result → /tmp/result → IsLocal false → hard "path traversal" error. main handled it fine. Use a normalization that collapses leading slashes (or filepath.Clean then strip).

4. 🟡 Optional out-of-bounds artifact now hard-errors instead of skipping (behavior change, plausible)

emissary.go:409

The IsLocal check now runs before the os.Stat existence short-circuit. A non-existent optional artifact with an out-of-bounds path previously returned nil (skipped); now it fails the step. Arguably acceptable (it's out-of-bounds), but it's a behavior change — decide intentionally, or move the IsLocal check after the existence skip.

5. 🟡 The one positive test never exercises the write path (confirmed)

emissary_test.go:215–219

LegitimateRelativePath writes to tmp/result.txt but saveParameter(ctx, "result.txt") opens relative to CWD → not found → swallowed by the os.IsNotExist branch → test passes green without ever calling root.Create. So regression #1 ships with no test catching it. There's also zero happy-path test for saveArtifact (:238). Fix the cwd mismatch and add an artifact write test — ideally with an internal-.. case to lock down #1.

6. 🟢 Trailing-slash + empty// edge regressions (lower)

- Parameter path with trailing / → root.Create is a directory (artifact trims it, parameter doesn't) — :444.
- Empty or / path → IsLocal("") false → hard error; artifact / → destination literally named .tgz — :407/455.

7. 🟢 Test isolation (plausible)

emissary_test.go:209 — both new tests rely on the package-level template global left by TestEmissary; the traversal asserts only pass because that template happens to declare no volumeMounts (otherwise FindOverlappingVolume short-circuits before the guard). Reset template in the new tests.

---
Bottom line: #1 and #2 are the ones to fix before merge — #1 silently breaks valid workflows, #2 leaves the hardening half-done. Both, plus most of the edge regressions, fold into one rewrite: open the root at varRunArgo, clean the relative path, and route both MkdirAll and Create through the root.
```

isubasinghe:
I will verify claude's output on Monday if I have time

jayantkamble10000:
Thanks for running that review — genuinely useful findings. I've pushed a commit addressing them (76dbe641b). Mapping each one:

**#1 🔴 internal `..` regression** — Fixed. The relative path is now `filepath.Clean`-ed *before* use, and the same cleaned path feeds both `MkdirAll` and `Create`. So `/out/../result` → `result` and writes correctly. Added a `LegitimateInternalDotDot` test for both saveParameter and saveArtifact to lock this down.

**#2 🟠 sandbox half-applied** — Fixed via the consolidation you suggested. I now open a single `os.Root` at the **trusted fixed base** (`varRunArgo`) and route *both* the directory creation and the file creation through it using `os.Root.MkdirAll` + `root.Create`. No `os.MkdirAll`/`os.OpenRoot` on attacker-influenceable paths anymore, so a symlinked component under `outputs/` (or the base dir itself) can't redirect the write.

**#3 🟠 leading `//`** — Fixed. Switched `strings.TrimPrefix(…, "/")` → `strings.TrimLeft(…, "/")` so all leading slashes are dropped; `//tmp/result` now maps to `tmp/result`.

**#5 🟡 tests didn't exercise the write path** — Fixed. The legitimate cases now `t.Chdir` into the source dir so `os.Open` resolves the source and the write actually runs, and they assert the output file exists under `outputs/…`. Added saveArtifact happy-path coverage (previously none).

**#7 🟢 test isolation** — Fixed. New tests reset the package-level `template`.

**#4 🟡 (out-of-bounds optional artifact now errors instead of skipping)** — Left intentional: I kept the `IsLocal` guard before the existence check so an out-of-bounds path fails loudly regardless of whether the file exists (an out-of-bounds path is never legitimate). Happy to flip the order if you'd prefer the skip-on-missing behavior.

**#6 🟢 trailing-slash/empty edges** — The `filepath.Clean` normalization now collapses trailing slashes for parameters too; empty/`/` inputs stay in-bounds (and missing sources are skipped by the existence check).

Important correctness note on the guard placement: `IsLocal` is applied to the user-controlled relative portion (`relSrc`) *before* joining the `outputs/...` prefix — if I checked the joined path instead, enough `..` would `Clean` away and a path could escape `outputs/artifacts` into `varRunArgo` while still reading as "local". Verified all cases (legit abs/rel, `//`, internal `..`, `../../argoexec`, `a/../../b`).

All `TestEmissary` + traversal tests pass, gofmt/vet clean. No rush given you mentioned Monday — thanks again for the thorough pass.

jayantkamble10000:
Quick CI follow-up on that last push:

- **Lint** was failing with a real (mine) issue — `shadow: declaration of "err" shadows declaration at line 425` in `saveArtifact`, introduced because I moved `MkdirAll` below the `root, err :=` line. Fixed by reusing the outer `err` (`=` instead of `:=`). Verified locally with the `shadow` analyzer (clean) plus `gofmt`/`go vet`.
- The **`test-functional` E2E** failures are the known `RetryTestSuite` flakes (`TestRetryLimit`, `TestRetryBackoff`, `TestWorkflowTemplateWithRetryStrategyInContainerSet`) — the logs show `LONG SET-UP took 27s/36s (maybe previous test was slow)` and no artifact/path errors. The unit tests that directly exercise retry + artifact saving (`TestEmissary/RetryContainerSet*`) pass locally.

Force-pushed the amended commit so it stays a single, clean "address review" change on top of the original fix. Everything else green; re-run should clear the flake.

jayantkamble10000:
Gentle nudge — please review the change if possible.

isubasinghe:
```console
  filepath.IsLocal(".") bypasses the guard (both functions) · emissary.go:409 and :459

Go's isLocal only rejects "", absolute paths, and paths that clean to ../../…. It returns true for .. Any srcPath that cleans to . slips through:

- srcPath = "/" → TrimLeft("/") = "" → Clean("") = "." → IsLocal(".") true.
- srcPath = "/tmp/.." → TrimLeft = "tmp/.." → Clean = "." → passes.

In saveArtifact, os.Stat("/") then succeeds and archive.TarGzToWriter(ctx, "/", …) tarballs the entire container filesystem into an output artifact — unbounded archive, disk/emptyDir exhaustion, and it defeats the guard's purpose. (In saveParameter the same path passes the gate but io.Copy from a directory then errors with EISDIR, so it's a hard-fail rather than a clean rejection — still wrong.)

One-line fix, both call sites:
if relSrc == "." || !filepath.IsLocal(relSrc) {
    return fmt.Errorf("path traversal in output artifact path %q", srcPath)
}
relSrc == "." only happens when srcPath is /, empty, or x/.. — never a valid single output path — so rejecting it is safe.
```

jayantkamble10000:
Great catch — you're right, and I should have caught this one myself (I'd actually seen `/` clean to `.` in local testing and wrongly waved it off as harmless, missing that `saveArtifact` would then tar the whole source root). 

Pushed `d2b29c57b` applying your exact fix in both functions:

```go
if relSrc == "." || !filepath.IsLocal(relSrc) {
    return fmt.Errorf("path traversal in output artifact path %q", srcPath)
}
```

Added a `RootPathCleansToDot` regression test to both `TestSaveParameterPathTraversal` and `TestSaveArtifactPathTraversal` covering `"/"`, `"/tmp/.."`, and `""` — all now rejected with a `path traversal` error rather than tarring root / hitting EISDIR. Full `TestEmissary` + traversal suites pass, and `gofmt`/`go vet`/`shadow` are clean. Thanks again for the thorough review.

jayantkamble10000:
While hardening this I realized the traversal tests all get rejected by the lexical guard (`relSrc`/`IsLocal`) **before** `os.Root` is ever reached — so the `os.Root` sandbox (the actual core of the fix) had **no** coverage. Reverting it back to `os.MkdirAll`/`os.Create` — the exact bug #2 you rejected earlier — kept the entire suite green. That's a gap worth closing given os.Root routing is the whole point here.

Pushed `d0be09cba` adding `SymlinkedDestComponentBlocked` to both `TestSaveParameterPathTraversal` and `TestSaveArtifactPathTraversal`:
- plants `outputs/<kind>/escape` as a symlink pointing outside `varRunArgo`
- calls `save*` with an `IsLocal` source (`escape/a`) that passes the string guard and reaches the filesystem layer
- asserts the write is refused **and** `require.NoFileExists` in the external dir

I verified it's load-bearing: it **fails** on an `os.Create`/`os.MkdirAll` regression (tarball escapes to the external dir) and **passes** on the current `os.Root` code — so it actually pins the sandbox guarantee rather than passing vacuously. Also added the missing `TraversalToTemplate` subtest to `saveArtifact` for parity.

Full `cmd/argoexec/commands` suite passes; `gofmt`/`go vet`/`shadow` clean.

isubasinghe:
I ran another claude review:

1. Optional outputs with traversal-shaped paths now hard-fail — cmd/argoexec/commands/emissary.go:465 (and :412 for artifacts). The old code's first action was the os.IsNotExist skip ("cannot save parameter, does not exist" → return nil), so a nonexistent optional output never failed the emissary regardless of path shape. The new traversal guard runs before the existence check, so a template with an optional output whose path is traversal-shaped and never created now fails the step. Previously-green workflows break. Fix: run the existence check first, or only enforce the guard once you know the file exists.

2. Exit code 0 is recorded while argoexec exits non-zero — cmd/argoexec/commands/emissary.go:263. The new error returns fire after exitCode is finalized (0 on command success), and the deferred write at emissary.go:59 still writes 0 to ctr/main/exitcode. In a containerSet, a dependent container gating on that file via waitForDependencyExitCode sees success and runs against never-staged outputs, and the wait container substitutes ValueFrom.Default for the missing parameter. The guard failure needs to be reflected in the exitcode file (or outputs saved before finalizing it).
3. Windows paths are rejected as traversal — emissary.go:461 / :408. strings.TrimLeft(srcPath, "/") only strips forward slashes, and on Windows filepath.IsLocal rejects rooted backslash paths (\out\result.txt) and reserved names (aux.log). Windows argoexec is supported (emissary_windows_test.go, docs/windows.md), and these paths worked before. Verifier confirmed against Go's Windows isLocal implementation.

4. path: / (or .) whole-filesystem exports now fail — emissary.go:412. Previously / produced outputs/artifacts/.tgz, which is exactly what the wait container's CopyFile looks up after Artifact.CleanPath, so whole-fs/cwd artifact exports worked end-to-end. The new relSrc == "." rejection breaks them. This is the one place where "reject" may still be the right call — but it's a behavior change to previously-working workflows, so it should be deliberate and documented, not incidental.

5. The os.Root sandbox still follows in-root symlinks; the tests overstate the guarantee — emissary.go:436, emissary_test.go:262. Empirically reproduced on Go 1.26.1: root.Create("outputs/parameters/out") follows a planted symlink outputs/parameters/out -> ../../ctr/main/exitcode because the target stays inside varRunArgo, truncating the exitcode file. Your tests only cover symlinks pointing outside the root. Caveat: the main container can already write the shared volume directly, so this is defense-in-depth shortfall rather than a new escalation — but the test names claim protection the sandbox doesn't provide. Worth an os.Lstat/root.Stat check or at least honest test coverage/comments.

isubasinghe:
Cleanup

6. emissary.go:496 — root.Create(dstRel) failure reports srcPath ("failed to create /tmp/result.txt") instead of the destination; the saveArtifact twin gets it right.
7. emissary.go:461/:408 — saveParameter/saveArtifact duplicate ~30 lines of security-critical guard/sandbox logic, and the copies have already drifted (TrimSuffix only in the artifact copy). One shared helper.
8. emissary_test.go:280 — the four rejection subtests are copy-pasted verbatim between the parameter and artifact tests; a shared table over func(ctx, path) error covers both.
9. emissary.go:428 — os.OpenRoot(varRunArgo) is opened/closed per output inside the caller's loop; open once around the output-saving block.
10. emissary.go:412 — consider also rejecting traversal-shaped output paths in workflow/validate so argo lint/submission catches it, instead of failing after the container has run to completion (keep the runtime check as defense-in-depth).

The 6 refuted candidates were mostly variations of "Clean-before-.tgz diverges writer from reader" — refuted because Artifact.CleanPath() runs before CopyFile, so reader and writer agree — plus a claim that os.OpenRoot needs MkdirAll (varRunArgo is created earlier in RunE).

jayantkamble10000:
Thanks — really thorough pass. Pushed `efe393153` addressing the code-level items, and grouping the rest below.

**Fixed in this PR:**
- **#1 (optional hard-fail):** moved the existence check *before* the guard in both functions, so a nonexistent optional output is skipped regardless of path shape — an escaping path whose source *exists* is still rejected. Added `NonexistentTraversalSkipped` tests. (The traversal tests now nest `varRunArgo` and create existing out-of-bounds targets, since the guard is now only reached for existing sources.)
- **#3 (Windows separators):** now `TrimLeft`/`TrimRight` over ``/\`` so rooted `\out\x` gets the same normalization and isn't wrongly rejected by `IsLocal`. Agree the `aux.log` case isn't real (Win11 `IsLocal` accepts names-with-extension); only bare device names are rejected, which is acceptable. A dedicated `//go:build windows` test is a good follow-up since the current suite is `!windows`.
- **#5 (symlink honesty):** renamed to `SymlinkedEscapingDestBlocked` and corrected the comments — `os.Root` blocks destination symlinks that *escape* the tree; in-root symlinks aren't covered. No longer overstates.
- **#6:** `saveParameter` create error now reports the destination.

**Design points I'd value your call on before coding:**
- **#4 ("/" export):** I actually don't think the two review rounds conflict — both point to rejecting `/` (the earlier one flagged it as an unbounded-archive DoS, this one notes it's a behavior change). I've kept the rejection and documented it in-code. Happy to keep as-is unless you'd prefer to preserve whole-fs export.
- **#2 (exitcode 0 on save failure):** confirmed real, but it's **pre-existing** — the `exitCode=0` / save-loop / `return err` block is byte-identical to the parent commit (`80dc102f4`); our guard only adds one more way to trigger a save failure that this latent bug mishandles. I'd prefer to fix it in a focused follow-up (set `exitCode` before the save-loop returns) rather than fold exitcode semantics into this traversal PR — but I'll do it here if you'd rather.

**Follow-ups I'll open once this lands (happy to do sooner if you prefer):**
- **#7/#8:** extract the shared guard/sandbox into one helper + a shared test table (removes the drift you flagged).
- **#9:** open `os.Root` once around the output-saving loop instead of per-output.
- **#10:** reject traversal-shaped output paths at `workflow/validate` so lint/submission catches them, keeping the runtime check as defense-in-depth.

Full `cmd/argoexec/commands` suite green; gofmt/vet/shadow clean.

PR #16036, opened 2026-04-24
fix(crds): escape template variables in CRD descriptions to prevent Helm rendering errors

ferhatguneri:
Fix CRD template variable rendering and regenerate generated artifacts

Escape Argo variable references in CRD descriptions (e.g. workflow.outputs.artifacts.XXXX)
so that Helm does not interpret them as template expressions during chart rendering.

Previously, raw `{{...}}` expressions in CRD descriptions could be evaluated by Helm,
causing rendering failures when installing manifests.

This change ensures these expressions are rendered as literal text.

Also regenerates code-generated artifacts to keep CRDs, OpenAPI spec, and client SDKs
in sync with updated API definitions.

### DIFF
--- a/api/jsonschema/schema.json
+++ b/api/jsonschema/schema.json
@@ -4119,7 +4119,7 @@
           "description": "Git contains git artifact location details"
         },
         "globalName": {
-          "description": "GlobalName exports an output artifact to the global scope, making it available as '{{io.argoproj.workflow.v1alpha1.outputs.artifacts.XXXX}} and in workflow.status.outputs.artifacts",
+          "description": "GlobalName exports an output artifact to the global scope, making it available as io.argoproj.workflow.v1alpha1.outputs.artifacts.XXXX and in workflow.status.outputs.artifacts",
           "type": "string"
         },
         "hdfs": {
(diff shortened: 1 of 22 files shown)


Joibel:
Hi @ferhatguneri - I appreciate what you're trying to fix here, but you've tackled this from the wrong place. These are all generated files, so to fix them you'll need to amend the comments in the files in [pkg/apis/workflow/v1alpha1](https://github.com/argoproj/argo-workflows/tree/main/pkg/apis/workflow/v1alpha1) and then run codegen.

I think it might be preferable just to drop all of the {{ and }} from those comments instead of double escaping them - they'd be confusing in other contexts then.

ferhatguneri:
Hi @Joibel Thank you so much for pointing to the right direction. I have done another commit, but to be honest due to some restrictions I couldn't build the crds to see if it removes the {{ }}.

Joibel:
> Hi @Joibel Thank you so much for pointing to the right direction. I have done another commit, but to be honest due to some restrictions I couldn't build the crds to see if it removes the {{ }}.

Have you tried using a github codespace - a free github account should have enough minutes to use `make codegen` to get it to fix this for you.

ferhatguneri:
Thank you so much again for help @Joibel . I could make codegen in Github codespace. The PR ir ready now. Could you please review ?

Joibel (approved):

PR #16038, opened 2026-04-25
fix(errors): treat kube-apiserver HTTP 500 as transient error

### LINKED ISSUE #14220: Argo fails to retry workflows when kube-apiserver returns a transient Internal Server Error
### Pre-requisites

The [`isTransientErr`](https://github.com/argoproj/argo-workflows/blob/5d9bcd974b5034e6ecb3132849f86269ad25d8b6/util/errors/errors.go#L30) function currently considers only the following kubernetes-native errors as transient. 

`apierr.IsTooManyRequests(err)`
`apierr.IsServerTimeout(err)`
`apierr.IsServiceUnavailable(err)`

in [link](https://github.com/argoproj/argo-workflows/blob/5d9bcd974b5034e6ecb3132849f86269ad25d8b6/util/errors/errors.go#L50-L60). 

This can lead to cases where the retry mechanism will not be triggered if the kube-apiserver returns transient Internal Server Errors. 

### Version(s)

3.6.4

```YAML
Description includes enough information.
```

```text
```

```text
```

HsiuChuanHsu:
Fixes #14220

## Motivation
The Workflows fails immediately when the kube-apiserver returns an HTTP 500 error. 
This happens because the system does not recognize `general 500 errors` as transient, turning temporary server issues into permanent workflow failures.

https://github.com/argoproj/argo-workflows/blob/7a7e60839796b0cb3a1343f2ec0968bd2ac2d694/util/errors/errors.go#L39-L56

> There is already a function `isResourceQuotaTimeoutErr` that calls `apierr.IsInternalError(err)`, but it narrows it with a specific message check ("resource quota evaluation timed out"), so every other HTTP 500 falls through.

https://github.com/argoproj/argo-workflows/blob/7a7e60839796b0cb3a1343f2ec0968bd2ac2d694/util/errors/errors.go#L73-L75

### 5XX Http Error that are  transient for now
|Function|Http Status|
|--|--|
|IsServerTimeout| HTTP 504(specific server-timeout reasons)|
|IsServiceUnavailable| HTTP 503 |
|isResourceQuotaTimeoutErr|HTTP 500(resource quota evaluation timed out)|

## Modifications
Updates the error handling logic to classify all HTTP 500 (Internal Server Error) responses as transient, enabling automatic retries to improve workflow resilience.
**Logic Update**: Added `apierr.IsInternalError(err)` to the `isTransientErr` function to catch 500 errors. 

## Verification
Updated unit tests to ensure HTTP 500 now triggers a retry and verified the fix with all existing sub-tests.

## AI
Claude did the testing part.

### DIFF
--- a/util/errors/errors.go
+++ b/util/errors/errors.go
@@ -47,6 +47,7 @@ func isTransientErr(err error) bool {
 		isResourceQuotaTimeoutErr(err) ||
 		isTransientNetworkErr(err) ||
 		apierr.IsServerTimeout(err) ||
+		apierr.IsInternalError(err) ||
 		apierr.IsTimeout(err) ||
 		apierr.IsServiceUnavailable(err) ||
 		isTransientEtcdErr(err) ||
--- a/util/errors/errors_test.go
+++ b/util/errors/errors_test.go
@@ -69,8 +69,9 @@ func TestIsTransientErr(t *testing.T) {
 		assert.Contains(t, hook.LastEntry().Msg, "Transient error")
 	})
 	t.Run("ResourceQuotaTimeoutErr", func(t *testing.T) {
-		assert.False(t, IsTransientErr(ctx, apierr.NewInternalError(errors.New(""))))
+		assert.True(t, IsTransientErr(ctx, apierr.NewInternalError(errors.New(""))))
 		assert.True(t, IsTransientErr(ctx, apierr.NewInternalError(errors.New("resource quota evaluation timed out"))))
+		assert.True(t, IsTransientErr(ctx, apierr.NewInternalError(errors.New("some other transient 500"))))
 	})
 	t.Run("ExceededQuotaErr", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


isubasinghe (changes requested):
Yeah I don't feel comfortable about a blanket acceptance of all 500 errors as transient. 
Maybe we could instead extend this to support custom transient errors.

isubasinghe:
I've closed this @HsiuChuanHsu but happy to reopen to discuss :)

HsiuChuanHsu:
Thanks for the feedback and sorry for the long delay in getting back to this.
I agree with you that a blanket 500 is too broad.

> Maybe we could instead extend this to support custom transient errors.

Here's an alternative I'm thinking of: introduce a `TRANSIENT_HTTP_STATUS_CODES` env var (comma-separated, e.g. `"500,503"`) that lets operators explicitly opt-in, following the same opt-in pattern as the existing `TRANSIENT_ERROR_PATTERN`.

Default behavior is unchanged — no operator gets unexpected retries. Operators who know their 500s are transient can opt in explicitly.

Does this direction work for you? Happy to update the PR.

PR #16053, opened 2026-04-29
feat: Add configurable batching for nodeResults in TaskSet patch operations to prevent etcd request size limits

### LINKED ISSUE #16054: Add configurable batching for nodeResults in TaskSet patch operations to prevent etcd request size limits
Summary

Add batching support for nodeResults in TaskSet patch operations to improve scalability and reduce patch size.

This change addresses an issue observed in highly parallel workflows (AWF), where within a single requeue period the accumulated nodeResults produced a patch request exceeding the etcd size limit.
In such cases, the TaskSet patch would fail with an error similar to:
`
etcdserver: request is too large`

The batch size is configurable via an environment variable and defaults to a sufficiently large value (e.g., 1000), allowing most workflows to operate without additional tuning while still providing flexibility for large-scale scenarios.

Use Cases
Large workflows with many parallel nodes
Avoiding oversized patch requests that may exceed etcd limits
Fine-tuning patch behavior via configurable batch size
Improving reliability of TaskSet updates under high load

---

ntny:
Motivation

In highly parallel Argo Workflows executed via the executor plugin (tested with ~200 parallel nodes), nodeResults can accumulate within a single requeue period and produce TaskSet patch payloads that exceed etcd size limits.

This leads to patch failures with errors such as:

(or slightly different variants depending on Argo version, e.g. v4.0.4).

To address this, this change introduces batching of nodeResults before patching TaskSets.

What this PR does
Introduces batching of nodeResults in TaskSet patch operations (executor plugin path)
Splits large result sets into smaller chunks before sending patch requests
Ensures each patch stays within etcd size limits
Applies patches sequentially per batch
Adds configurable batch size via environment variable (default: 1000)

-->

Fixes https://github.com/argoproj/argo-workflows/issues/16054

Future possible improvement: split TaskSet into metadata and payload, keep metadata in etcd with a reference to the payload, and store payloads in external storage to further reduce etcd pressure.

### DIFF
--- a/.features/pending/feat-support-batching.md
+++ b/.features/pending/feat-support-batching.md
@@ -1,0 +1,12 @@
+Issues: 16054
+Authors: [Anton Pechenin](ntny1986@gmail.com)
+Description: Add batching support for nodeResults in TaskSet patch operations
+Component: General
+
+You can now batch nodeResults when patching Workflow TaskSets.
+This improves reliability for workflows with a large number of parallel nodes by preventing patch payloads from exceeding etcd size limits.
+
+Previously, workflows with many parallel nodes (tested with ~200), particularly when using the agent executor plugin, could generate patch requests that exceeded etcd size limits when multiple TaskSet updates occurred within a single reconciliation period.
+With batching enabled, nodeResults are split into smaller chunks and patched sequentially, keeping each request within safe size limits.
+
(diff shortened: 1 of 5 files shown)


ntny:
Closing this PR because of a different root cause. Will fix and follow up.

Issue #13239, opened 2024-06-24
Template with retry and exit hook does not behave as expected on error


McKMarcBruchner:
### Pre-requisites

- [X] I have double-checked my configuration
- [X] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [X] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened/what did you expect to happen?

Hi argo team,

I have two problems with the exact same workflow template, depending on the version I'm using.

In my workflow, I have template DAG task which has both a retry option (limit and policy doesn't seem to matter, it doesn't work with any setting) and a simple exit lifecycle hook. After that step, I have an additional step which should not be executed if the retry fails (either until the limit is reached or because the error does not match the retryPolicy).

Here are my issues:
1. Versions up to `v3.4.17` (latest `v.3.4.X` version), or up to `v3.5.4`: 
The workflow runs the task with the retry, failing because of some reason (in my example workflow by intention), then it either retries or not depending on the retryPolicy and the error type.
Afterwards the hook runs as expected, but then the workflow seems to want to continue, the next step taking the **exit hook step** as its predecessor, thus skipping. The whole workflow is then marked as succeeded, while the retry step clearly failed. This is very unexpected behaviour to me!
![Screenshot 2024-06-24 at 08 45 21](https://github.com/argoproj/argo-workflows/assets/87852749/36cbaf4b-4e04-404a-84b1-cea56cdeec85)
![Screenshot 2024-06-24 at 08 46 21](https://github.com/argoproj/argo-workflows/assets/87852749/cc465a7b-c5e5-4fc5-a07f-baefe3e3dd7c)
I would expect the workflow to fail directly after the exit hook of the failing step has been executed, as one of its steps has failed.

2. Versions `v3.5.5` up to `v3.5.8` (currently `latest` version):
[...]

Issue #14392, opened 2025-04-15
"Max parallelism reached" error in DAG with task level exit handlers


igorcalabria:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

DAG Workflows with `parallelism: 1` and dag-level exit handlers always "fail" due to "Max parallelism reached".  Regardless of what happens with the workflow, it always be marked with "Error". Setting the parallelism to 2 avoids the issue but it's really counter intuitive since exit handlers will always run after the task finishes. I think the error makes sense when using lifecycle hooks for "Running" tasks, but not for exit handlers. 

### Version(s)

v3.4.10 up to at least latest (18e1aa6da3459d8890f0ba8954b7602762b912fb11bb1b2e32ba4047214f3ee6)

### Paste a minimal workflow that reproduces the issue. We must be able to run the workflow; don't enter a workflow that uses private images.

```YAML
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: exit-hanlder-dag-level-
spec:
  entrypoint: main
  parallelism: 1
  templates:
    - name: main
      dag:
        tasks:
          - name: A
            onExit: exit
            template: echo
            arguments:
              parameters: [{name: message, value: A}]
    - name: echo
      inputs:
        parameters:
          - name: message
      container:
        image: busybox
        command: [echo]
        args: ["{{inputs.parameters.message}}"]

    - name: exit
      container:
        image: busybox
        command: [echo]
        args: ["task cleanup"]
```

### Logs from the workflow controller

```text
time="2025-04-15T16:30:57.988Z" level=info msg="Performing periodic GC" periodicity=5m0s
time="2025-04-15T16:30:57.988Z" level=info msg="Persistence disabled - so archived workflow GC disabled - you must restart the controller if you enable this"
time="2025-04-15T16:30:57.989Z" level=info msg="Starting CronWorkflow controller"
[...]

PR #12281, opened 2023-11-28
fix: allow withItems when hooks are involved

Joibel:
processAggregateNodeOutputs expects all children to be withItems children, but when there are hooks involved, this is not true. It uses `DisplayName` in an evil way to order the children, but the hooks don't fit it's idea of how displayName should look, so it will panic. I don't attempt to address this evil.

Instead I've opted filter the children in place to make it work.

Fixes #12279 

### Verification

Functional test included.

PR #12192, opened 2023-11-13
fix: ensure wftmplLifecycleHook wait for each dag task

toyamagu-2021:
Signed-off-by: toyamagu-2021 <toyamagu2021@gmail.com>

Fixes #12120

### Motivation

* Ensure WorkflowTemplateLevelLifeCycleHook wait for each dag task
* Almost same as `onExit` handler #3435

### Modifications

* Add `CheckAllHooksFullfilled` to check child hooked nodes.

### Verification

This PR resolve #12120

<details>
<summary>yaml</summary>

```yaml
spec:
  templates:
    - name: main
      inputs: {}
      outputs: {}
      metadata: {}
      dag:
        tasks:
          - name: some-task
            template: some-task
            arguments: {}
            hooks:
              failure:
                template: exit-handler
                arguments: {}
                expression: tasks["some-task"].status == "Failed"
          - name: finish
            template: finish
            arguments: {}
            dependencies:
              - some-task
    - name: some-task
      inputs: {}
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: alpine:latest
        command:
          - sh
          - '-c'
        args:
          - |
            echo "Doing great things";
            echo "Failing...";
            exit 1;
        resources: {}
    - name: exit-handler
      inputs: {}
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: alpine:latest
        command:
          - sh
          - '-c'
        args:
          - |
            echo "Sending updates to github";
            sleep 5;
            echo "Done!";
        resources: {}
      retryStrategy:
        limit: 1
        retryPolicy: Always
    - name: finish
      inputs: {}
      outputs: {}
      metadata: {}
      container:
        name: ''
        image: alpine:latest
        command:
          - sh
          - '-c'
        args:
          - |
            echo "Finished!";
        resources: {}
  entrypoint: main
```

</details>

* Success hook case

image

* Failed hook case

image

PR #16058, opened 2026-04-30
fix: schedule lifecycle hooks for non-leaf and retrying DAG tasks. Fixes #16057

### LINKED ISSUE #16057: DAG task lifecycle hooks silently skipped when task failure triggers Omit cascade on downstream tasks
### Pre-requisites

When a DAG task fails and has downstream dependent tasks, those downstream tasks are marked `Omitted`. Any lifecycle hook declared on the failing task is never scheduled.
This affects all DAG topologies where the failing task has at least one downstream dependent.

_What happens when you execute the sample workflow attached below:_

Image

_What I expected to happen:_

Image

The notify hook on `step-a` should fire when `step-a` fails, regardless of whether `step-a` has downstream dependents. Downstream tasks should be marked `Omitted` only after all hooks on the failing task have completed.

### Version(s)

v4.0.2, v4.0.5, 0eb6267ac

```YAML
kind: Workflow
metadata:
  generateName: hook-bug-
spec:
  entrypoint: main
  templates:
    - name: main
      dag:
        tasks:
          - name: step-a
            template: fail-inner
            hooks:
              notify:
                template: echo
                expression: tasks["step-a"].status == "Failed"
          - name: step-b
            template: succeed-inner
            depends: step-a

    - name: fail-inner
      dag:
        tasks:
          - name: inner
            template: fail-pod

    - name: succeed-inner
      dag:
        tasks:
          - name: inner
            template: succeed-pod

    - name: fail-pod
      container:
        image: alpine:3.23
        command: [sh, -c]
        args: ["exit 1"]

    - name: succeed-pod
      container:
        image: alpine:3.23
        command: [sh, -c]
        args: ["echo done"]

    - name: echo
      container:
        image: alpine:3.23
        command: [echo]
        args: ["hook fired"]
```

```text
{"time":"2026-04-30T08:38:02.541709186Z","level":"INFO","msg":"Processing workflow","lastSeenVersion":"","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","phase":"","resourceVersion":"701913157"}
{"time":"2026-04-30T08:38:02.553802829Z","level":"INFO","msg":"resolved artifact repository","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","artifactRepositoryRef":{"default":true}}
{"time":"2026-04-30T08:38:02.555229614Z","level":"INFO","msg":"Task-result reconciliation","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","component":"workflow_worker","numObjs":0}
{"time":"2026-04-30T08:38:02.555291112Z","level":"INFO","msg":"updated phase","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","fromPhase":"","toPhase":"Running","component":"workflow_worker"}
{"time":"2026-04-30T08:38:02.555369943Z","level":"WARN","msg":"Node was nil, will be initialized as type Skipped","namespace":"hxdr-processing","component":"workflow_worker","workflow":"hook-bug-rxph8"}
{"time":"2026-04-30T08:38:02.556007549Z","level":"INFO","msg":"was unable to obtain node, letting display name to be nodeName","component":"workflow_worker","workflow":"hook-bug-rxph8","boundaryID":"","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.556038162Z","level":"INFO","msg":"node initialized","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","node":"hook-bug-rxph8","phase":"Running","message":""}
{"time":"2026-04-30T08:38:02.557104092Z","level":"INFO","msg":"Event occurred","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"},"fieldPath":"","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","component":"workflow_worker","type":"Normal","reason":"WorkflowRunning","message":"Workflow Running"}
{"time":"2026-04-30T08:38:02.557373467Z","level":"WARN","msg":"was unable to obtain the node","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeID":"hook-bug-rxph8-231077340","taskName":"step-a"}
{"time":"2026-04-30T08:38:02.557419344Z","level":"WARN","msg":"was unable to obtain the node","taskName":"step-b","nodeID":"hook-bug-rxph8-281410197","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.557430755Z","level":"WARN","msg":"was unable to obtain the node","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeID":"hook-bug-rxph8-281410197","taskName":"step-b","component":"workflow_worker"}
{"time":"2026-04-30T08:38:02.557507729Z","level":"WARN","msg":"was unable to obtain the node","namespace":"hxdr-processing","component":"workflow_worker","workflow":"hook-bug-rxph8","nodeID":"hook-bug-rxph8-231077340","taskName":"step-a"}
{"time":"2026-04-30T08:38:02.558240925Z","level":"WARN","msg":"was unable to obtain the node","nodeID":"hook-bug-rxph8-231077340","taskName":"step-a","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.558258175Z","level":"INFO","msg":"All of node dependencies completed","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeName":"hook-bug-rxph8.step-a","dependencies":null}
{"time":"2026-04-30T08:38:02.558337572Z","level":"WARN","msg":"Node was nil, will be initialized as type Skipped","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.558654851Z","level":"INFO","msg":"node initialized","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","node":"hook-bug-rxph8-231077340","phase":"Running","message":""}
{"time":"2026-04-30T08:38:02.559189186Z","level":"WARN","msg":"was unable to obtain the node","nodeID":"hook-bug-rxph8-158441414","taskName":"inner","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.559204461Z","level":"WARN","msg":"was unable to obtain the node","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeID":"hook-bug-rxph8-158441414","taskName":"inner"}
{"time":"2026-04-30T08:38:02.559420877Z","level":"WARN","msg":"was unable to obtain the node","nodeID":"hook-bug-rxph8-158441414","taskName":"inner","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.559439528Z","level":"INFO","msg":"All of node dependencies completed","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeName":"hook-bug-rxph8.step-a.inner","dependencies":null}
{"time":"2026-04-30T08:38:02.559466871Z","level":"WARN","msg":"Node was nil, will be initialized as type Skipped","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","component":"workflow_worker"}
{"time":"2026-04-30T08:38:02.561275107Z","level":"INFO","msg":"node initialized","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","message":"","node":"hook-bug-rxph8-158441414","phase":"Pending"}
{"time":"2026-04-30T08:38:02.561642363Z","level":"INFO","msg":"getExecutorLogOpts","workflow":"hook-bug-rxph8","loglevel":"info","namespace":"hxdr-processing","nodeName":"hook-bug-rxph8.step-a.inner","nodeID":"hook-bug-rxph8-158441414","component":"workflow_worker"}
{"time":"2026-04-30T08:38:02.561909522Z","level":"INFO","msg":"getExecutorLogOpts","nodeID":"hook-bug-rxph8-158441414","loglevel":"info","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeName":"hook-bug-rxph8.step-a.inner"}
{"time":"2026-04-30T08:38:02.563554166Z","level":"INFO","msg":"getExecutorLogOpts","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeName":"hook-bug-rxph8.step-a.inner","nodeID":"hook-bug-rxph8-158441414","loglevel":"info"}
{"time":"2026-04-30T08:38:02.619104625Z","level":"INFO","msg":"add pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:02.620025865Z","level":"INFO","msg":"Created pod","nodeName":"hook-bug-rxph8.step-a.inner","nodeID":"hook-bug-rxph8-158441414","podName":"hook-bug-rxph8-fail-pod-158441414","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.620106533Z","level":"WARN","msg":"was unable to obtain the node","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeID":"hook-bug-rxph8-281410197","taskName":"step-b","component":"workflow_worker"}
{"time":"2026-04-30T08:38:02.620291132Z","level":"WARN","msg":"was unable to obtain the node","taskName":"step-b","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeID":"hook-bug-rxph8-281410197"}
{"time":"2026-04-30T08:38:02.620517468Z","level":"INFO","msg":"TaskSet Reconciliation","namespace":"hxdr-processing","component":"workflow_worker","workflow":"hook-bug-rxph8"}
{"time":"2026-04-30T08:38:02.621013565Z","level":"INFO","msg":"reconcileAgentPod","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:02.627086138Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:02.627939571Z","level":"INFO","msg":"Workflow to be dehydrated","namespace":"hxdr-processing","component":"workflow_worker","workflow":"hook-bug-rxph8","Workflow Size":1796}
{"time":"2026-04-30T08:38:02.644870561Z","level":"INFO","msg":"Workflow update successful","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","phase":"Running","lastSeenVersion":"701913157","resourceVersion":"701913165"}
{"time":"2026-04-30T08:38:02.644962837Z","level":"INFO","msg":"Event occurred","reason":"WorkflowNodeRunning","message":"Running node hook-bug-rxph8","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"},"fieldPath":"","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","type":"Normal","component":"workflow_worker"}
{"time":"2026-04-30T08:38:02.64498548Z","level":"INFO","msg":"Event occurred","message":"Running node hook-bug-rxph8.step-a","component":"workflow_worker","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"},"fieldPath":"","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","type":"Normal","reason":"WorkflowNodeRunning"}
{"time":"2026-04-30T08:38:02.692696293Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:03.638424389Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:04.680255958Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:06.664619838Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:07.670408178Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:08.678119865Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:09.777745233Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:12.620124913Z","level":"INFO","msg":"Processing workflow","resourceVersion":"701913165","lastSeenVersion":"701913157","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","phase":"Running"}
{"time":"2026-04-30T08:38:12.621090875Z","level":"INFO","msg":"Task-result reconciliation","numObjs":1,"namespace":"hxdr-processing","component":"workflow_worker","workflow":"hook-bug-rxph8"}
{"time":"2026-04-30T08:38:12.62133436Z","level":"INFO","msg":"Pod failed","displayName":"inner","templateName":"fail-pod","pod":"hook-bug-rxph8-fail-pod-158441414","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","message":"main: Error (exit code 1)"}
{"time":"2026-04-30T08:38:12.621360031Z","level":"INFO","msg":"Pod restart check: analyzing pod status","podName":"hook-bug-rxph8-fail-pod-158441414","workflow":"hook-bug-rxph8","containerStatuses":2,"podPhase":"Failed","podReason":"","podMessage":"","namespace":"hxdr-processing","component":"workflow_worker","initContainerStatuses":1}
{"time":"2026-04-30T08:38:12.621404615Z","level":"INFO","msg":"Pod restart check: pod does not qualify for restart","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","podName":"hook-bug-rxph8-fail-pod-158441414","podReason":""}
{"time":"2026-04-30T08:38:12.621745553Z","level":"WARN","msg":"was unable to obtain the node","taskName":"step-b","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","component":"workflow_worker","nodeID":"hook-bug-rxph8-281410197"}
{"time":"2026-04-30T08:38:12.621764419Z","level":"WARN","msg":"was unable to obtain the node","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","nodeID":"hook-bug-rxph8-281410197","taskName":"step-b"}
{"time":"2026-04-30T08:38:12.622193414Z","level":"INFO","msg":"Outbound nodes set","nodeID":"hook-bug-rxph8-231077340","outbound":["hook-bug-rxph8-158441414"],"component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:12.622216891Z","level":"INFO","msg":"node phase changed","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","node":"hook-bug-rxph8-231077340","fromPhase":"Running","toPhase":"Failed"}
{"time":"2026-04-30T08:38:12.622250044Z","level":"INFO","msg":"node finished","namespace":"hxdr-processing","node":"hook-bug-rxph8-231077340","finishedAt":"2026-04-30T08:38:12Z","component":"workflow_worker","workflow":"hook-bug-rxph8"}
{"time":"2026-04-30T08:38:12.622287129Z","level":"WARN","msg":"was unable to obtain the node","nodeID":"hook-bug-rxph8-281410197","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","taskName":"step-b"}
{"time":"2026-04-30T08:38:12.622451296Z","level":"INFO","msg":"node initialized","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","node":"hook-bug-rxph8-281410197","phase":"Omitted","message":" (message: omitted: depends condition not met)"}
{"time":"2026-04-30T08:38:12.622554063Z","level":"INFO","msg":"Outbound nodes set","nodeID":"hook-bug-rxph8","outbound":["hook-bug-rxph8-281410197"],"workflow":"hook-bug-rxph8","namespace":"hxdr-processing","component":"workflow_worker"}
{"time":"2026-04-30T08:38:12.62277553Z","level":"INFO","msg":"node phase changed","fromPhase":"Running","toPhase":"Failed","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","node":"hook-bug-rxph8"}
{"time":"2026-04-30T08:38:12.622797356Z","level":"INFO","msg":"node finished","node":"hook-bug-rxph8","finishedAt":"2026-04-30T08:38:12Z","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","component":"workflow_worker"}
{"time":"2026-04-30T08:38:12.622854104Z","level":"INFO","msg":"TaskSet Reconciliation","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:12.622862647Z","level":"INFO","msg":"reconcileAgentPod","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:12.622876202Z","level":"INFO","msg":"updated phase","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","fromPhase":"Running","toPhase":"Failed"}
{"time":"2026-04-30T08:38:12.622915108Z","level":"INFO","msg":"Marking workflow completed","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:12.622931875Z","level":"INFO","msg":"Marking workflow as pending archiving","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:12.623422585Z","level":"INFO","msg":"Event occurred","fieldPath":"","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","type":"Warning","reason":"WorkflowFailed","message":"","component":"workflow_worker","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"}}
{"time":"2026-04-30T08:38:12.623717208Z","level":"INFO","msg":"Workflow to be dehydrated","component":"workflow_worker","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","Workflow Size":4004}
{"time":"2026-04-30T08:38:12.638297554Z","level":"INFO","msg":"Workflow update successful","workflow":"hook-bug-rxph8","namespace":"hxdr-processing","component":"workflow_worker","resourceVersion":"701913333","phase":"Failed","lastSeenVersion":"701913165"}
{"time":"2026-04-30T08:38:12.638476075Z","level":"INFO","msg":"Event occurred","fieldPath":"","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","type":"Normal","reason":"WorkflowNodeRunning","message":"Running node hook-bug-rxph8.step-a.inner: main: Error (exit code 1)","component":"workflow_worker","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"}}
{"time":"2026-04-30T08:38:12.63893103Z","level":"INFO","msg":"Event occurred","component":"workflow_worker","apiVersion":"argoproj.io/v1alpha1","type":"Warning","reason":"WorkflowNodeFailed","message":"Failed node hook-bug-rxph8.step-a.inner: main: Error (exit code 1)","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"},"fieldPath":"","kind":"Workflow"}
{"time":"2026-04-30T08:38:12.638997684Z","level":"INFO","msg":"Event occurred","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"},"component":"workflow_worker","fieldPath":"","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","type":"Warning","reason":"WorkflowNodeFailed","message":"Failed node hook-bug-rxph8.step-a"}
{"time":"2026-04-30T08:38:12.639036565Z","level":"INFO","msg":"Event occurred","fieldPath":"","component":"workflow_worker","kind":"Workflow","apiVersion":"argoproj.io/v1alpha1","type":"Warning","reason":"WorkflowNodeFailed","message":"Failed node hook-bug-rxph8","object":{"name":"hook-bug-rxph8","namespace":"hxdr-processing"}}
{"time":"2026-04-30T08:38:13.647792376Z","level":"INFO","msg":"queueing pod for cleanup after","podName":"hook-bug-rxph8-fail-pod-158441414","action":"deletePod","after":5000000000,"component":"pod_controller","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:13.648444786Z","level":"INFO","msg":"archiving workflow","component":"archive_worker","uid":"50cad03e-ed67-40f5-b4ca-6a173f3db159","namespace":"hxdr-processing","workflow":"hook-bug-rxph8"}
{"time":"2026-04-30T08:38:14.739053229Z","level":"INFO","msg":"Queueing workflow for delete due to TTL","phase":"Failed","workflow":"hxdr-processing/hook-bug-rxph8","addAfter":118000000000,"component":"gc_controller"}
{"time":"2026-04-30T08:38:18.648218723Z","level":"INFO","msg":"cleaning up pod","component":"pod_controller","podName":"hook-bug-rxph8-fail-pod-158441414","key":"hxdr-processing/hook-bug-rxph8-fail-pod-158441414/deletePod","action":"deletePod","namespace":"hxdr-processing"}
{"time":"2026-04-30T08:38:18.658154979Z","level":"INFO","msg":"update pod event","component":"pod_controller","pod":"hook-bug-rxph8-fail-pod-158441414"}
{"time":"2026-04-30T08:38:18.665262697Z","level":"INFO","msg":"delete pod event","pod":"hook-bug-rxph8-fail-pod-158441414","component":"pod_controller"}
{"time":"2026-04-30T08:40:13.002419201Z","level":"INFO","msg":"Deleting garbage collected workflow","workflow":"hxdr-processing/hook-bug-rxph8","component":"gc_controller"}
{"time":"2026-04-30T08:40:13.016836493Z","level":"INFO","msg":"Successful request to be deleted","component":"gc_controller","workflow":"hxdr-processing/hook-bug-rxph8"}
```

```text
```

fcolombo7:
Fixes #16057

### Motivation

When a non-leaf DAG task (one whose template is itself a DAG or steps) fails, its lifecycle hooks are never scheduled if the failure causes downstream tasks to be marked `Omitted`.

The root cause: hook evaluation in `executeDAGTask` runs at the top of the function while the node is still `Running`. For non-leaf tasks, the phase transition `Running → Failed` happens later inside `executeTemplate`. By then hook evaluation has already passed and no further sweep is triggered for that task, so the hook is permanently lost.

The same applies to tasks with a `retryStrategy`, leaf or not, since a retry node's phase is also settled inside `executeTemplate`, and to expanded tasks (`withItems`, `withParam`, `withSequence`), whose `TaskGroup` node reaches its phase in the same sweep its dependents are `Omitted`. Plain leaf tasks (bare container pods) are unaffected because their phase is updated by `podReconciliation` before `executeDAGTask` runs.

### Modifications

`dag.go`: re-evaluate hooks in the existing `if node.Completed()` block at the bottom of `executeDAGTask`'s inner loop, where `runOnExitNode` already handles the same late phase transition. Ordering against the `Omit` cascade comes from the existing `CheckAllHooksFullfilled` gate in `evaluateDependsLogic`, since `executeTmplLifeCycleHook` links the hook node as a child before returning, so no early return is needed. Expanded tasks are skipped there and evaluated once at task level against the `TaskGroup` node, so the hook fires once per task rather than once per item.

`operator.go`: `getOutboundNodes` now skips hooked children when picking a retry node's outbound node, matching `possiblyGetRetryChildNode` and `getRetryNodeChildrenIds`. Without it, a succeeded hook becomes the outbound node of a failed retrying task and therefore the parent of that task's dependents, so `assessDAGPhase` propagates the hook's phase and the workflow reports `Succeeded`.

One change is independent of the hook fix and worth backporting on its own: `executeDAGTask` now logs and returns when `buildLocalScopeFromTask` fails, instead of falling through and dereferencing the nil scope it returns, which panicked the controller.

### Verification

Four regression tests in `dag_test.go`, each failing without its corresponding change. The non-leaf cases use nested inner DAG templates (required to reproduce; bare container tasks are reconciled before `executeDAGTask` runs):

- `TestDAGHookNonLeafSerialHeadFails`: non-leaf task fails with one downstream dependent
- `TestDAGHookNonLeafFanInBothFail`: two parallel non-leaf tasks both fail with a shared downstream
- `TestDAGHookWithItemsFiresOnce`: expanded task raises exactly one hook node, on the task rather than per item
- `TestDAGHookRetryTaskOutbound`: a succeeded hook does not mask a retrying task's failure

`TestDagWftmplHookWithRetry` and `TestDAGOrphanedTaskGroupCompletes` still pass, so the #13239 and #16450 fixes hold. Full `go test ./workflow/controller/` passes with no regressions.

Also verified on a cluster with a controller built from this branch: the workflow from this issue, retry non-leaf, retry leaf, `withItems`, fan-in with two failing dependencies, retry with both a non-exit and an exit hook, and an all-succeeding run.

### Documentation

No documentation change needed.

### AI

Claude (Anthropic) was used to navigate and understand the codebase, identify the root cause, assist in writing the fix and regression tests, and to reproduce and work through the review feedback. All code was reviewed and validated by the author per the [Argo project Generative AI policy](https://github.com/argoproj/argoproj/blob/main/community/genai.md).

## Summary by CodeRabbit

* **Bug Fixes**
  * Lifecycle hooks now run reliably when DAG tasks fail, including serial, parallel, and retry scenarios.
  * Dependent tasks are omitted only after applicable failure hooks complete.
  * Failure hooks for expanded item tasks execute once at the overall task level.
  * Successful hooks no longer mask the failure of retrying tasks.
* **Tests**
  * Added regression coverage for lifecycle hook execution and workflow failure behavior.


Joibel (commented):
### Code review

Thanks for this — the write-up is genuinely good, and the diagnosis is correct. The root cause you identified is real, the fix is load-bearing (delete the hook block and `TestDAGHookNonLeafFanInBothFail` fails outright), and you were right that the existing `executeDAG` hook loop can't cover it, since that loop only runs for target tasks and `step-a` isn't one. `executeDAGTask` is the right function. The problem is only *where inside it* the hook is evaluated.

The headline concern is the first comment below: for a task with `retryStrategy`, this change makes a failing workflow report `Succeeded`. I reproduced it, so please treat that as blocking.

Most of the rest share one root cause. The new call passes the node `executeTemplate` returned for the *expanded* task, while the pre-existing evaluation at the top of `executeDAGTask` passes the task-level node — so the two sites disagree about which node a hook hangs off. That single choice drives comments 1 and 2, and probably has a one-line fix: pass `dagCtx.getTaskNode(ctx, taskName)`, or skip the new call when `task.ShouldExpand()`.

Some history that may help, since the current layout looks arbitrary until you see it. #10307 ("Fix not working Running state lifecycle hooks in dag task", #9897) fixed the near-identical bug by *relocating* the hook evaluation up to the top of the function and keying it on the task-level node — that relocated block is what this PR works around. And #16088 ("do not re-run `onExitNode`", #14392) *removed* a duplicated hook-launch call from `executeDAG`, leaving a standing comment in `main` warning that calling `executeTemplate` a second time on the same node "would re-run checkParallelism against the count this very pass just bumped". Worth reading both before the next revision.

Two other things:

- **`steps.go` does not need the same fix.** I checked, building the steps analogue of your repro: the hook node is created Pending on the sweep the step goes Failed, and the workflow waits for it. `steps.go` is immune precisely because its hook evaluation is already a post-`executeTemplate` pass over the children — which is the shape this change probably wants to adopt.
- **This needs a rebase.** The branch is ~252 commits behind. #16043 changed `executeTmplLifeCycleHook` and `runOnExitNode` to take `varkeys.NodeRefKeys` plus a name instead of the `"tasks."+taskName` string, and replaced `scope.addParamToScope(fmt.Sprintf("tasks.%s.status", ...))` with `varkeys.TasksNodeRef.Status.Set(...)`. Both new call sites use the old forms, so it won't compile as-is.

Worth running `HooksSuite` on this branch too — `TestTemplateLevelHooksDagHasDependencyVersion` asserts `Progress("1/2")` under the comment "Make sure unnecessary hooks are not triggered", which is the assertion most exposed to comment 2.

workflow/controller/dag.go
```diff
@@ -643,8 +643,19 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
 			scope, err := woc.buildLocalScopeFromTask(ctx, dagCtx, task)
 			if err != nil {
 				woc.markNodeError(ctx, node.Name, err)
+				return
 			}
 			scope.addParamToScope(fmt.Sprintf("tasks.%s.status", task.Name), string(node.Phase))
+			// Re-evaluate hooks: executeTemplate may have transitioned this node from Running to
+			// Fulfilled within this sweep, after the earlier hook evaluation ran.
+			hookCompleted, err := woc.executeTmplLifeCycleHook(ctx, scope, task.Hooks, node, dagCtx.boundaryID, dagCtx.tmplCtx, "tasks."+taskName)
```
Joibel: **Blocking: a task with `retryStrategy` plus hooks now reports the workflow as `Succeeded` when it should be `Failed`.**

`executeTmplLifeCycleHook` adds the hook node as a child of `parentNode` (`hooks.go`, via `addChildNode`). For a retry task that parent is the **Retry** node, and `getOutboundNodes` returns the *last* child for `NodeTypeRetry`:

```go
case wfv1.NodeTypeRetry:
    numChildren := len(node.Children)
    if numChildren > 0 {
        return []string{node.Children[numChildren-1]}
    }
```

So the succeeded hook becomes the task's outbound node, hence the parent of the omitted downstream task, and `assessDAGPhase` propagates `Succeeded` upward.

Reproduced by adding `retryStrategy: {limit: "1", retryPolicy: Always}` to `fail-inner` in your own `dag_hook_non_leaf_serial_head_fails.yaml`, with the hook pod succeeding and the task pods failing:

```
step-a              type=Retry  phase=Failed
  step-a(0)         type=DAG    phase=Failed
  step-a(1)         type=DAG    phase=Failed
  step-a.hooks.notify           phase=Succeeded  hooked=true
step-b                          phase=Omitted
>>> FINAL WORKFLOW PHASE = Succeeded      <-- should be Failed
```

Reverting only the `dag.go` hunk gives `Failed`. This also affects **leaf** tasks with `retryStrategy` + hooks, so it isn't limited to the non-leaf case this PR targets.

In fairness, before this change the hook never fired at all for retry tasks, so you are genuinely extending the fix — but turning a correctly-failing workflow green is worse than the bug being fixed. Likely needs hooked children excluded when computing a retry node's outbound nodes, plus a test. Note #13239 ("Template with retry and exit hook stuck in `Pending`") is a closed bug on this same combination, so it's worth confirming that fix still holds.

Joibel: **`withItems`/`withParam` tasks now fire their hook once per expanded item *plus* once at task-group level, instead of exactly once.**

Two things go wrong here, both from the arguments passed:

**1. Wrong node → duplicate hooks.** `node` here is what `executeTemplate` returned for the *expanded* task `t`, and `generateLifeHookNodeName` builds the hook node name from `parentNode.Name`. The pre-existing evaluation at the top of `executeDAGTask` passes the task-level (TaskGroup) node instead, so the two produce different names and **both** fire. On a `withItems: [foo, bar, baz]` task with one `notify` hook:

```
step-a(0:foo).hooks.notify
step-a(1:bar).hooks.notify
step-a(2:baz).hooks.notify
step-a.hooks.notify            <-- from the top-of-function evaluation
```

Four hook executions for three items. Reverting the hunk gives exactly one. A 100-item fan-out with a notification hook would fire 101 hooks. `docs/lifecyclehook.md` says a LifecycleHook "executes during execution time and executes once".

**2. Wrong hooks spec → `{{item}}` never substituted.** `expandTask` marshals the task, substitutes the item and unmarshals, so `t.Hooks` carries item-substituted expressions and arguments while `task.Hooks` does not. With a hook argument `value: "item-is-{{item}}"`, every per-item hook node is created with the literal placeholder:

```
step-a(0:foo).hooks.notify  inputs=[msg=item-is-{{item}}]
step-a(1:bar).hooks.notify  inputs=[msg=item-is-{{item}}]
step-a(2:baz).hooks.notify  inputs=[msg=item-is-{{item}}]
```

If per-item hook nodes are kept they must be fed `t.Hooks`; this becomes moot if the node fix above means they're never created.

Both existing guards for this are in the repo already but won't catch it: `test/e2e/smoke/with-items-with-hooks.yaml` (from #12281, "allow withItems when hooks are involved") is exactly this shape — `withItems: [1,2,3]` on a non-leaf `steps` template with a non-exit `running` hook — but only asserts `ToBeSucceeded`.

workflow/controller/dag.go
```diff
@@ -643,8 +643,19 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
...
+			if err != nil {
+				woc.markNodeError(ctx, node.Name, err)
+				return
+			}
+			if !hookCompleted {
+				return
+			}
```
Joibel: **This early return looks redundant, and it's what causes the serialisation.**

I deleted just these three lines and both of your new tests still pass. The Omit cascade is already held back by the pre-existing `CheckAllHooksFullfilled` gate in `evaluateDependsLogic` (added by #12192, "ensure wftmplLifecycleHook wait for each dag task"), which walks the dependency's children and returns false for any unfulfilled hooked child. Since `executeTmplLifeCycleHook` calls `addChildNode` before returning, that child link already exists in the same sweep the hook node is created — which is the whole reason your fix works.

What the return costs, meanwhile:

- **Serialises expanded items.** It returns from `executeDAGTask` outright, abandoning the rest of `for _, t := range expandedTasks`. On the 3-item probe, one item's hook fired per reconcile sweep and the workflow took 4 sweeps instead of 2; the delay is linear in item count. Removing the return restored baseline timing, with all hooks starting in the same sweep.
- **Blocks real work under `parallelism`.** The next item's pod isn't created either, and the pending hook holds the parallelism slot.
- **Skips the trailing `taskGroupNode` phase-marking block**, leaving the TaskGroup `Running` with every child Succeeded. That's the same failure mode #16454 exists to heal (#16450, TaskGroups stuck `Running`).
- **Defers `runOnExitNode`** on the very next line — the call #8992 added specifically to stop retry-type DAG tasks with exit nodes hanging (#6557) — by at least one reconcile for any task with both a non-exit hook and an `onExit`.

If a gate is wanted here after all, tracking a pending-hooks flag and returning after the loop (the way `steps.go` does) would preserve the task-group and onExit behaviour. But the simplest change is to drop these three lines.

workflow/controller/dag.go
```diff
@@ -643,8 +643,19 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
 			scope, err := woc.buildLocalScopeFromTask(ctx, dagCtx, task)
 			if err != nil {
 				woc.markNodeError(ctx, node.Name, err)
+				return
```
Joibel, line 693: This `return` is a real fix rather than tidying — `buildLocalScopeFromTask` returns `(nil, err)`, and the next statement calls `scope.addParamToScope`, which is `s.scope[key] = val` and would nil-deref and panic the controller. Good catch.

Two small requests:

1. **Add the log line.** Both sibling handlers for this same call in this function log before returning, e.g. `log.WithError(err).Error(ctx, "Failed to build local scope from task")`. That pattern exists because it was asked for during review of #10307 ("just add a log and return"), so it's worth matching here.
2. **Mention it in the PR body.** It isn't required by the hook fix — I removed it and both new tests still pass — so it's an independent latent-panic fix riding along. A line in the description (or a separate commit) keeps it from getting lost, since it's worth backporting on its own merits.

workflow/controller/dag.go
```diff
@@ -643,8 +643,19 @@ func (woc *wfOperationCtx) executeDAGTask(ctx context.Context, dagCtx *dagContex
...
 			}
 			scope.addParamToScope(fmt.Sprintf("tasks.%s.status", task.Name), string(node.Phase))
+			// Re-evaluate hooks: executeTemplate may have transitioned this node from Running to
+			// Fulfilled within this sweep, after the earlier hook evaluation ran.
```
Joibel: Small wording point: this says the node transitioned "from Running to **Fulfilled**", but the block is gated on `node.Completed()`, which is true only for Succeeded/Failed/Error. `Fulfilled()` is a distinct and wider predicate — it also covers Skipped and Omitted, and additionally requires the task result to be synced (`workflow/controller/AGENTS.md`: "A node is only truly done when its phase is fulfilled **and** its task result is synced — `NodeStatus.Fulfilled()` checks both").

The distinction is deliberate in this codebase; commit `132b947ad6`, "fix: Differentiate between Fulfilled and Completed" (#3083), exists specifically to separate the two across `dag.go`, `steps.go` and `operator.go`. And `Fulfilled()` is called just a few lines below on `onExitNode`, so a future reader has both words in view at once.

Something like "from Running to a Completed phase (e.g. Failed)" would match what the guard actually checks.

workflow/controller/dag_test.go
```diff
@@ -4188,3 +4188,110 @@ func TestDAGWhenExprSkipEval(t *testing.T) {
...
+		panic(err)
+	}
+	for _, pod := range pods.Items {
+		if pod.Status.Phase == v1.PodFailed || pod.Status.Phase == v1.PodSucceeded {
+			continue
+		}
+		if pod.Status.Phase == phase {
+			continue
+		}
+		pod.Status.Phase = phase
```
Joibel: This forks `makePodsPhase` rather than extending it — about 25 of the 29 lines are verbatim, with two extra `continue`s as the only real difference.

`makePodsPhase` already has a variadic `with ...with` extension point (used by `withOutputs` and `withExitCode`), so a terminal-phase skip could go there, or be a thin wrapper, instead of a second copy that will drift.

A few smaller things while you're in here:

- The `pod.Status.Phase == phase` check is unreachable at both call sites: both pass `PodSucceeded`, which the preceding terminal-phase check already skips.
- It drops the `with ...with` variadic that `makePodsPhase` has.
- It lives in `dag_test.go`, while every sibling pod helper (`makePodsPhase`, `createRunningPods`, `deletePods`, `syncPodsInformer`) is in `controller_test.go`.
- The name misleads — it doesn't act on Pending pods, it skips *terminal* ones, so it also transitions Running and Unknown. Something like `makeNonTerminalPodsPhase`, next to `makePodsPhase`, would say what it does and keep the helpers together.

workflow/controller/dag.go
fcolombo7: Confirmed and fixed in `c053e3dfe`. `getOutboundNodes` skips hooked children when picking a
retry node's outbound node, matching `possiblyGetRetryChildNode` and
`getRetryNodeChildrenIds`.

Added `TestDAGHookRetryTaskOutbound`, which fails without the change.
`TestDagWftmplHookWithRetry`, covering #13239, still passes.

Worth flagging: on `main` a retry task fires no hooks at all, leaf or non-leaf, because
`processNodeRetries` settles the retry node's phase inside `executeTemplate`. So #16057
covers retry tasks too, and this only surfaced once they started firing.

workflow/controller/dag.go
fcolombo7: Reworded in `5ea4fba27` to "from Running to a Completed phase (e.g. Failed)", with a note
that the block is gated on `Completed()`, not `Fulfilled()`.

workflow/controller/dag_test.go
fcolombo7: Done in `71a634a5d`. `makeNonTerminalPodsPhase` sits beside `makePodsPhase` in
`controller_test.go`, both delegating to a shared `setPodsPhase` with a `skipTerminal` flag,
so the `with ...with` extension point survives. Dropped the unreachable equal-phase check
and renamed as you suggested.

The variadic could not carry the skip itself: a `with` func mutates the pod rather than
filtering it, so it is a wrapper.

workflow/controller/dag.go
fcolombo7: Added in `c22922dfc`.

I merged main in rather than rebasing, so I could not split the guard out of the original
commit without rewriting it. It is called out in that commit message and in the PR
description instead. Say if you want it isolated.

workflow/controller/dag.go
fcolombo7: Confirmed and dropped in `b5ad9050d`. Both tests still pass. On a 3-item `withItems` fan-out the
sweep count goes from 6 to 3.

#16454 has landed since you wrote this, which likely softens the `taskGroupNode` point.
`TestDAGOrphanedTaskGroupCompletes` passes with the return removed.

workflow/controller/dag.go
fcolombo7: Both confirmed. Fixed in `c9b3de9dc`: the post-`executeTemplate` evaluation is gated on
`!task.ShouldExpand()`, so there are no per-item hook nodes and the `t.Hooks` versus
`task.Hooks` problem goes with them.

Gating alone left an expanded non-leaf task with no hook at all, so the commit also
evaluates once at task level against the TaskGroup node.
`TestDAGHookWithItemsFiresOnce` asserts exactly one hook node, raised on the task.

workflow/controller/dag_test.go
```diff
@@ -4734,3 +4809,68 @@ func TestOnExitDAGNotFailedOnShutdownStop(t *testing.T) {
...
+func TestDAGHookWithItemsFiresOnce(t *testing.T) {
+	wf := wfv1.MustUnmarshalWorkflow("@testdata/dag_hook_withitems_fires_once.yaml")
+	ctx := logging.TestContext(t.Context())
+	woc := newWoc(ctx, *wf)
+
+	// Drive the workflow to completion, failing task pods and succeeding hook pods.
+	for i := 0; i < 8 && !woc.wf.Status.Phase.Completed(); i++ {
+		woc.operate(ctx)
+		makeNonTerminalPodsPhase(ctx, woc, v1.PodFailed)
+	}
```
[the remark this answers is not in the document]
fcolombo7: Comment addressed: the test now completes the hook and asserts the terminal phases

PR #16059, opened 2026-05-02
fix: Compress large WorkflowTaskSetSpec to avoid etcd size limit errors

ntny:
### Motivation

When using executor plugins, Argo stores task definitions and execution results in WorkflowTaskSet resources.

Unlike Workflow, both _WorkflowTaskSet.spec.Tasks_ and _WorkflowTaskSet.status.Nodes_ are not compressed, which can lead to exceeding etcd object size limits (~1MB) in large workflows.

This becomes relevant in high fan-out workflows, where WorkflowTaskSet may grow to multiple megabytes and cause create/patch failures due to etcd request limits.
```
TaskSet Patch Failed
```

### Modifications

This change introduces per-entry compression for:

WorkflowTaskSet.spec.Tasks
WorkflowTaskSet.status.Nodes

### Key changes:

each task and node is compressed individually
compression is applied unconditionally (when beneficial)
no full-map serialization is used
Why per-entry compression

Both Tasks and Nodes are map-based collections where each entry is fully replaced per key via Kubernetes merge semantics.

Because of this, compressing the full map would:

require full decompression for any update
break per-key lifecycle handling
increase patch amplification

Per-entry compression preserves update semantics while reducing object size.

### Tradeoff

Unconditional compression introduces additional CPU overhead in controller/agent paths, but ensures deterministic avoidance of etcd size limit failures.

### Verification
large TaskSets successfully created and patched within etcd limits
round-trip validated: compress → store → decompress → execute
backward compatibility confirmed for existing objects without compressed fields

### DIFF
--- a/api/jsonschema/schema.json
+++ b/api/jsonschema/schema.json
@@ -7305,6 +7305,10 @@
           "description": "AutomountServiceAccountToken indicates whether a service account token should be automatically mounted in pods. ServiceAccountName of ExecutorConfig must be specified if this value is false.",
           "type": "boolean"
         },
+        "compressedTemplate": {
+          "description": "Compressed and base64 decoded Template",
+          "type": "string"
+        },
         "container": {
           "$ref": "#/definitions/io.k8s.api.core.v1.Container",
           "description": "Container is the main container image to run in the pod"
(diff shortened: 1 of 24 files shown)


ntny:
The actual root cause of the error has been identified and fixed in this [PR](https://github.com/argoproj/argo-workflows/pull/16075)

PR #16063, opened 2026-05-07
fix: remove duplicate archiveWorkflowAux call in archiveWorkflow

SebTardif:
Fixes #15780 (regression)

### Motivation

`archiveWorkflow` calls `archiveWorkflowAux` twice: once on line 1121 with error handling, and again unconditionally on line 1126 as the return value. When the first call succeeds, the second call re-archives the same workflow, causing:

- Double database writes to the workflow archive
- Double Kubernetes API Patch calls to update the `archivingStatus` label
- Redundant hydration and JSON marshaling cycles

This was introduced in #15780, which refactored `archiveWorkflow` from returning `void` to returning `error`. The original call at the bottom (`return wfc.archiveWorkflowAux(ctx, obj)`) was left in place instead of being replaced with `return nil`.

```go
// Before (buggy):
err = wfc.archiveWorkflowAux(ctx, obj)  // first call
if err != nil {
    logger.WithField("key", key).WithError(err).Error(ctx, "failed to archive workflow")
    return nil // non-retryable
}
return wfc.archiveWorkflowAux(ctx, obj)  // second call - duplicate

// After (fixed):
err = wfc.archiveWorkflowAux(ctx, obj)
if err != nil {
    logger.WithField("key", key).WithError(err).Error(ctx, "failed to archive workflow")
    return nil // non-retryable
}
return nil
```

### Modifications

- Replace the duplicate `archiveWorkflowAux` call with `return nil` in `workflow/controller/controller.go`.

### Verification

- Code compiles successfully (`go build ./workflow/controller/`)
- Existing archive-related tests pass
- Traced git history to confirm the duplicate was unintentional (introduced in #15780)

### Documentation

No documentation changes needed. This is a bugfix with no user-facing API changes.

### AI

Generative AI (Grok) was used to help identify this bug during a codebase audit and to draft this PR description.

### DIFF
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -1113,7 +1113,7 @@ func (wfc *WorkflowController) archiveWorkflow(ctx context.Context, obj any) err
 	}
 	wfc.workflowKeyLock.Lock(key)
 	defer wfc.workflowKeyLock.Unlock(key)
-	key, err = cache.MetaNamespaceKeyFunc(obj)
+	_, err = cache.MetaNamespaceKeyFunc(obj)
 	if err != nil {
 		logger.Error(ctx, "failed to get key for object after locking")
 		return nil // non-retryable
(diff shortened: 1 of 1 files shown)


Joibel:
Nice find. This makes that function always return nil, which also feels wrong. I think we should return err instead (losing line 1124 to make that happen)?

SebTardif:
Good call. Updated to return the error directly (`return wfc.archiveWorkflowAux(ctx, obj)`), which also removes the redundant Error log since the caller already logs at Warn level with the requeue context.

PR #16064, opened 2026-05-07
fix(ui): fix mixed bold/notbold markdown in title annotations

johnke:
### Motivation

While standardising our workflow titles, I noticed a weird behaviour when mixing bold and non-bold text, where the first instance of bold markdown would be marked italics with literal asterisks. Any bold text after the beginning (e.g. after un-formatted text) behaved as expected.

image

### Modifications

The `escapeInvalidMarkdown` function in `ui/src/workflows/utils.ts` has a regex intended to escape markdown list items (`- item` or `* item`), but the pattern `/([-*])\s/g` matches any `*` followed by a space anywhere in the string, not just at the start of a line.

This causes `**bold** notbold` to be transformed into `**bold*\* notbold`, which the markdown renderer interprets as italics with a literal asterisk instead of bold. 

The fix anchors the list-item and ordered-list regexes to the start of a line using `^` with the `m` (multiline) flag, so they only match actual list items.

### Verification

- `yarn eslint --ext .ts,.tsx src/workflows/utils.ts src/shared/components/workflows-utils.test.ts` - clean
- Added a test for the `**bold** notbold` scenario
- `./node_modules/.bin/jest --config jest.config.js src/shared/components/workflows-utils.test.ts` - passes


Joibel (approved):
LGTM, thanks

PR #16075, opened 2026-05-10
fix: WorkflowTaskSets size bloat for large workflows

ntny:
### Motivation

removeCompletedTaskSetStatus was executed after workflow dehydration, causing completed task set nodes to be skipped during cleanup and leading to WorkflowTaskSets size bloat. This is because getDeleteTaskAndNodePatch [relies on in-memory status.nodes](https://github.com/ntny/argo-workflows/blob/8f75471639320b37a510b67fd8dbbd42fd4faa3e/workflow/controller/taskset.go#L33), which are cleared during workflow compression.

As a result, large compressed workflows could continuously accumulate completed task set data and eventually fail with errors such as:
```
TaskSet Patch Failed
```

This change:

-  moves cleanup before dehydration
- checks that the workflow is not compressed before tracking task set nodes

Fixes #TODO

### Verification

Verified the changes on:

- regular workflows
- compressed workflows (AWF) that previously reproduced the issue

Confirmed that completed task sets are cleaned up correctly and workflows no longer fail with oversized patch requests.

### DIFF
--- a/test/e2e/fixtures/then.go
+++ b/test/e2e/fixtures/then.go
@@ -83,6 +83,18 @@ func (t *Then) ExpectWorkflowDeleted() *Then {
 	return t
 }
 
+func (t *Then) ExpectWorkflowCompressed() *Then {
+	ctx := logging.TestContext(t.t.Context())
+	wf, err := t.client.Get(ctx, t.wf.Name, metav1.GetOptions{})
+	if err != nil {
+		t.t.Fatal(err)
+	}
+	if wf.Status.CompressedNodes == "" {
+		t.t.Errorf("expected workflow to be compressed")
+	}
+	return t
+}
+
 // ExpectWorkflowNode checks on a specific node in the workflow.
 // If no node matches the selector, then the NodeStatus and Pod will be nil.
 // If the pod does not exist (e.g. because it was deleted) then the Pod will be nil too.
--- a/test/e2e/manifests/plugins/massive-executor-plugin-configmap.yaml
+++ b/test/e2e/manifests/plugins/massive-executor-plugin-configmap.yaml
@@ -1,15 +1,28 @@
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: massive-executor-plugin
+  labels:
+    workflows.argoproj.io/configmap-type: ExecutorPlugin
+  annotations:
+    workflows.argoproj.io/description: |
+      This plugin returns a massive string (~1000 chars).
 data:
-  sidecar.automountServiceAccountToken: "true"
+  sidecar.automountServiceAccountToken: "false"
   sidecar.container: |
     args:
     - |
       import json
+      import random
+      import string
       from http.server import BaseHTTPRequestHandler, HTTPServer
 
       with open("/var/run/argo/token") as f:
           token = f.read().strip()
 
+      def gen_text():
+        base = "HELLO_WORLD_123 "
+        return base * (100_000 // len(base))
 
       class Plugin(BaseHTTPRequestHandler):
 
@@ -21,10 +34,6 @@ data:
               self.end_headers()
               self.wfile.write(json.dumps(reply).encode("UTF-8"))
 
-          def forbidden(self):
-              self.send_response(403)
-              self.end_headers()
-
           def unsupported(self):
               self.send_response(404)
               self.end_headers()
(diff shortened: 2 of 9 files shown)


terrytangyuan (commented):
Could you add some tests to exercise this change?

ntny:
> Could you add some tests to exercise this change?

Sure, I added a test to verify that compressed workflows now fail explicitly with a clear error instead of silently proceeding with invalid taskset cleanup logic.

I also was considering adding a large stress/e2e test for the executor plugin around the 2MB limit, but I’m not sure it’s a good idea.

ntny:
@terrytangyuan 
I’ve added unit and e2e tests
Could you please review?
- unit `RemoveCompletedTaskSetStatusFailsForCompressedWorkflow` just check that compressed workflows now fail explicitly with a clear error instead of silently proceeding with invalid taskset cleanup logic (the operator should decompress the workflow before cleanup calls)
- E2E `TestCompressedTemplateExecutor_WorkflowTaskSetIsProperlyCleaned` verifies that for an AWF that has reached the compressed size limit and has been compressed, tasksets were properly cleaned up

terrytangyuan (commented):
@Joibel Can you take another look here?

test/e2e/fixtures/then.go
```diff
@@ -83,6 +83,18 @@ func (t *Then) ExpectWorkflowDeleted() *Then {
 	return t
 }
 
+func (t *Then) ExpectWorkflowCompressed() *Then {
+	ctx := logging.TestContext(t.t.Context())
```
Joibel, line 87: Follow `ExpectWorkflow`, `ExpectWorkflowName` and call  expectWorkflow(name, func) instead of hand rolling this.

test/e2e/manifests/plugins/massive-executor-plugin-configmap.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: massive-executor-plugin
+  labels:
+    workflows.argoproj.io/configmap-type: ExecutorPlugin
+  annotations:
+    workflows.argoproj.io/description: |
+      This plugin returns a massive string (~1000 chars).
```
Joibel:
```suggestion
      This plugin returns a massive string (~100000 chars).
```

util/file/watch.go
```diff
@@ -113,7 +113,7 @@ func watchFilePoll(ctx context.Context, path string, onChange func()) error {
 				}
 				return err
 			}
-			if last == nil || fi.ModTime() != last.ModTime() || fi.Size() != last.Size() {
+			if last == nil || !fi.ModTime().Equal(last.ModTime()) || fi.Size() != last.Size() {
```
Joibel, line 116: Is this relevant? It looks like a good change, but perhaps should be separate - is there something here that tickled this problem?

Having said that, this is better so happy to retain the change.

workflow/controller/taskset.go
```diff
@@ -224,3 +228,14 @@ func (woc *wfOperationCtx) createTaskSet(ctx context.Context) error {
 	}
 	return nil
 }
+
+// ensureNotCompressed verifies that the workflow is not in a compressed state
+// The workflow operator is responsible for controlling when a workflow is
+// compressed or decompressed.
+// Tasksets operate on in-memory nodes and must ensure the workflow is decompressed before proceeding
+func (woc *wfOperationCtx) ensureNotCompressed() error {
+	if woc.wf != nil && woc.wf.Status.CompressedNodes != "" {
```
Joibel: We shouldn't check woc.wf here - if it really was nil we'd explode somewhere anyway, and we've just declared that this non-existent workflow is not compressed.

```suggestion
	if woc.wf.Status.CompressedNodes != "" {
```

Actually, this whole ensureNotCompressed guard isn't doing what it looks like — the call site only logs at Warn, so a future reordering that puts cleanup back below Dehydrate would silently re-introduce the bloat with no enforcement, while legitimate paths into persistUpdates (e.g. controller.go's Hydrate-failure branch) will hit the guard normally and emit a misleading "workflow must be decompressed" warning on top of the real error. Both problems share a root cause: the ordering requirement is expressed as a runtime precondition on shared state, rather than as a data dependency in the function signature. Could we drop the guard and instead have removeCompletedTaskSetStatus take the hydrated nodes snapshot as an explicit parameter, so callers physically can't invoke it without the data it needs?

Joibel (changes requested):

test/e2e/fixtures/then.go
ntny: Good point. I considered using ExpectWorkflow as well, but it dehydrates the workflow [here](https://github.com/argoproj/argo-workflows/blob/main/test/e2e/fixtures/then.go#L68) before the function check, so it doesn't work for this particular check.

util/file/watch.go
ntny: This is not directly related to the fix. I noticed it while working on this area and changed it as a small cleanup.

I thought it might not be worth opening a separate PR and running the full e2e pipeline just for this small change. That said, if you prefer to keep the PR strictly focused, I can remove it.

ntny:
@Joibel could you please take another look? I’ve addressed your review comments.

Joibel (approved):
Looks, good, thank you

ntny:
Hi @Joibel, thanks for merging this. Would it be possible to backport this fix to the 3.7 release branch as well? If anything is needed for that, I’d be happy to help.

Joibel:
> Hi @Joibel, thanks for merging this. Would it be possible to backport this fix to the 3.7 release branch as well? If anything is needed for that, I’d be happy to help.

Can you create a PR against release-3.7 for it - if you follow the PR title convention for cherry-picks by looking at others in the repo that helps.

Issue #9140, opened 2022-07-12
UI/API: show all namespaces possibilities 


bengoldenberg:
we would like to have the option to see all the namespaces we have in our cluster, in the Namespace section inside the ui as a dropdown list or something like this.
today we have many namespaces and once we have added more namespaces, we cant manage them manually.
![Screen Shot 2022-07-12 at 14 06 17](https://user-images.githubusercontent.com/71871067/178476348-33f9a875-cf00-4938-8900-06a0bed9dc1d.png)

amitde69:
This is a critical feature in order to provide argo workflows as a tool to the organization.

OneCricketeer:
Or workaround has been to create an external tool to provision namespaces and app git repos (e.g. Backstage software templates) with markdown links in the README that link direct to the workflow UI with `/<namespace>` where that repo would be built

angapov:
namespace dropdown would be awesome

fingeromer:
Hi, is this feature in your roadmap?

agilgur5:
The main issue with this is that this requires a namespace query, which the Server and users do not normally have permissions to do. The alternative would be to pull out the namespace from every Workflow a user has permission to see, but there's performance ramifications with that as well

agilgur5:
Turns out this is effectively a duplicate of #7405, which mentions needing a new `ListNamespaces` query and RBAC as well

PR #16087, opened 2026-05-13
feat(ui): autocomplete namespace filter from displayed resources (refs #7405)

### LINKED ISSUE #7405: Add autocomplete for namespace filter
# Summary

Add the ability to autocomplete the namespace filter in "Workflow Template" and "Workflow" screens using the already available namespaces in the cluster.

# Use Cases

In our specific use case, we build and destroy namespaces dynamically per development or production branch. and use these namespaces a the isolation level for our different workflows. selecting these namespaces in the UI for the first time is a bit cumbersome, as we need to type them in manually.

---

callmemorgan:
Fixes #7405

### Motivation

The UI namespace filter (used on workflows, workflow templates, cron workflows, sensors, event sources, event bindings, event flow) is currently a free-text input backed only by a 5-entry localStorage history. Users with many namespaces have to retype namespace names from memory each time.

Issue #7405 has tracked this for 4+ years. In [a comment on the dup #9140](https://github.com/argoproj/argo-workflows/issues/9140#issuecomment-2432936375), @agilgur5 noted that one safe and proper way to provide namespace autocomplete is to derive suggestions from the namespaces of workflows the user has already loaded — since the user already has `list` permission on those namespaces, this leaks no information beyond k8s RBAC's existing answer.

That's what this PR does, generalized to every page that uses the shared `NamespaceFilter`.

### Modifications

- `shared/components/input-filter.tsx` gains an optional `extraSuggestions?: string[]` prop. Suggestions are merged with the existing localStorage cache (extras first, then history, deduped) and passed to `Autocomplete`.
- `shared/components/namespace-filter.tsx` gains a matching `extraNamespaces?: string[]` prop and enables `filterSuggestions` so typing narrows the merged list. The managed-namespace short-circuit is unchanged.
- `shared/namespaces.ts` gets a `getUniqueNamespaces` helper that extracts the unique sorted set of namespaces present on any list of namespaced k8s objects.
- 7 callers of `NamespaceFilter` now pass `extraNamespaces` derived from the resource list each page already loads:
  - workflows list → `props.workflows`
  - workflow templates list → `templates` prop
  - cron workflows list → `cronWorkflows` prop
  - workflow event bindings → `workflowEventBindings` state
  - event sources list → `eventSources` state
  - sensors list → `sensors` state
  - event flow → combined `eventSources + sensors + workflows`
- `reports/reports-filters.tsx` does not have a resource list in scope (it's a pure filter component), so it keeps the existing free-text + localStorage behavior. Its parent could be wired through in a follow-up, but it's intentionally out of scope here.
- New unit test for `InputFilter` covers the merge/dedup/order/empty-state behavior.

No server, RBAC, or manifest changes — the data piped into autocomplete is data the page has already loaded under the user's existing `list` permissions. Cluster-wide list responses already include `metadata.namespace` per item (e.g., `workflow-template-list.tsx:90`), which is what populates the suggestions.

This is purposefully a small fix. Some honest limitations:
- It does not surface empty namespaces (no resources visible).
- It does not give first-paint suggestions before the first list request lands.
- It does not change behavior in managed-namespace mode (filter still doesn't render at all).

Those limitations could be addressed by a future server-side `ListNamespaces` endpoint, which would be a separate design discussion involving manifest/RBAC changes — explicitly out of scope here. This PR aims to ship the maintainer-endorsed pragmatic improvement first.

Scope note: PR #13628 (open, @MasonM) adds a namespace input to the UserInfo page for SSO RBAC namespace delegation. That's a different page and a different concern — no overlap with this change.

### Verification

- `yarn --cwd ui lint` passes.
- `yarn --cwd ui test` passes; new test cases in `input-filter.test.tsx` cover:
  1. `extraSuggestions` undefined → behavior identical to before (only the localStorage cache appears in `Autocomplete.items`).
  2. `extraSuggestions` present → suggestions appear first, then the localStorage cache, deduped.
  3. Empty `extraSuggestions` → empty items list.
- Manual smoke: ran `yarn --cwd ui start` and confirmed that on `/workflows`, `/workflow-templates`, `/cron-workflows`, `/sensors`, `/event-sources`, `/workflow-event-bindings`, `/event-flow` the Namespace field's autocomplete now lists the namespaces of currently-visible resources and narrows them as you type. Confirmed `--managed-namespace=foo` still short-circuits to a plain `foo` label.

### Documentation

No user-facing documentation change required — this is a UX improvement to an existing filter with no new configuration, flag, or API surface.

### AI

Claude (Anthropic) was used to:
- Survey existing `NamespaceFilter` callers and confirm what data each had in scope.
- Verify existing helpers (`util/auth.CanIArgo`, `server/auth.CanI`) and their callers.
- Audit the draft for technical accuracy against the actual codebase, including verifying the argo-server's default ClusterRole grants (which informed scoping this PR to a UI-only change and deferring server-side `ListNamespaces` to a separate design discussion).
- Draft this PR description and the unit test.

All code in this PR was reviewed by the author before commit.

### DIFF
--- a/.features/pending/namespace-filter-autocomplete.md
+++ b/.features/pending/namespace-filter-autocomplete.md
@@ -1,0 +1,10 @@
+Description: Autocomplete the Namespace filter from the namespaces of resources currently visible on the page
+Author: [Morgan Allen](https://github.com/callmemorgan)
+Component: UI
+Issues: 7405
+
+The Namespace filter on the workflows, workflow templates, cron workflows, sensors, event sources, event bindings, and event flow pages now autocompletes from the namespaces of resources already loaded for that page, in addition to the existing localStorage history.
+
+When the user is viewing resources across multiple namespaces (cluster-wide mode), the namespace filter dropdown now lists the namespaces present on the current page and narrows as you type. No new server endpoint or RBAC is involved — suggestions are derived from data the user already has permission to see.
+
+The managed-namespace short-circuit (where the filter renders as plain text) is unchanged.
--- a/ui/src/cron-workflows/cron-workflow-filters.tsx
+++ b/ui/src/cron-workflows/cron-workflow-filters.tsx
@@ -5,6 +5,7 @@ import {CheckboxFilter} from '../shared/components/checkbox-filter/checkbox-filt
 import {NamespaceFilter} from '../shared/components/namespace-filter';
 import {TagsInput} from '../shared/components/tags-input/tags-input';
 import * as models from '../shared/models';
+import {getUniqueNamespaces} from '../shared/namespaces';
 
 import './cron-workflow-filters.scss';
 
@@ -45,6 +46,7 @@ export function CronWorkflowFilters({cronWorkflows, namespace, labels, states, o
                         onChange={ns => {
                             onChange(ns, labels, states);
                         }}
+                        extraNamespaces={getUniqueNamespaces(cronWorkflows)}
                     />
                 </div>
                 <div className='columns small-2 xlarge-12'>
--- a/ui/src/event-flow/event-flow-page.tsx
+++ b/ui/src/event-flow/event-flow-page.tsx
@@ -256,7 +256,14 @@ export function EventFlowPage({history, location, match}: RouteComponentProps<an
                         }
                     ]
                 },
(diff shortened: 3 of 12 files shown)


MasonM (approved):
Thanks, this is definitely an improvement!

For future reference: the way to test this in the local development environment is `make start UI=true NAMESPACED=false`, since the default mode is namespaced.

PR #16093, opened 2026-05-14
feat(server): add PKCE support to OAuth2 SSO flow

trollkarlen:
Enables PKCE (RFC 7636) on the authorization code flow by default using S256, per the OAuth 2.0 Security BCP. The code verifier is round-tripped through the existing state cookie. Adds insecureSkipPKCE opt-out for identity providers that reject the code_challenge parameter.

### Fixes
missing PKCE support

### Motivation
Security

### Verification
Added tests

### Documentation
Added docs

### AI
yes, opencode


trollkarlen:
Will add helm chart changes when this comes in, or should i say if :)

trollkarlen:
tumbleweed_gif

isubasinghe:
Hey @trollkarlen you are missing a features document in .features/pending. This is why CI is failing for you

trollkarlen:
What, where and how?
Please give me som information on where this "process" is explained, or in the best of worlds help out.

PS.
I thought this was open source, not cooperate bureaucracy 😉

config/sso.go
```diff
@@ -35,6 +35,12 @@ type SSOConfig struct {
 	FilterGroupsRegex []string `json:"filterGroupsRegex,omitempty"`
 	// custom PEM encoded CA certificate file contents
 	RootCA string `json:"rootCA,omitempty"`
+	// InsecureSkipPKCE disables PKCE (RFC 7636) on the OAuth2 authorization
```
MasonM: If PKCE is going to be made the default (which makes this a backwards-incompatible change), I think there needs to be a justification beyond defense-in-depth or the fact it's recommended by RFC 9700. How exactly could an attacker pull of an authorization code injection without PKCE? The only ways I can think of are when TLS is disabled or the attacker manages to install a malicious extension in the victim's browser, and in those cases the victim has much bigger problems. Defense-in-depth for CSRF attacks is nice, but not worth breaking compatibility for.

MasonM:
@trollkarlen Here's instructions to add a feature document: https://argo-workflows.readthedocs.io/en/latest/running-locally/#creating-feature-descriptions

config/sso.go
trollkarlen: Most oauth vendors(and libs for implementing oauth) have this enabled by default and you must enable InsecureSkipPKCE on their end to use this code as it was before this change.

The attack f.ex:
They inject the victim's authorization code into the attacker's own login flow on the legitimate client. The result: the attacker is logged into the legitimate app as the victim. 

The threat isn't just CSRF defense-in-depth — it's mix-up attacks, referrer leakage to ad networks, history exposure, and log capture, all of which work over fully-functional TLS against confidential clients (client_secret does not stop code injection)

MasonM: There's a high bar for breaking backwards compatibility. If there's a concrete security vulnerability here, then that could meet the bar, but I'm not seeing that.

To be clear, I support adding support for PKCE, but I'm not convinced it should be the default. 

> Most oauth vendors(and libs for implementing oauth) have this enabled by default 

Argo Workflows is frequently run in enterprise environments that aren't using the most up-to-date versions of these libraries.

> They inject the victim's authorization code into the attacker's own login flow on the legitimate client. 

How? What precise steps would an attacker follow to do this?

trollkarlen:
> @trollkarlen Here's instructions to add a feature document: https://argo-workflows.readthedocs.io/en/latest/running-locally/#creating-feature-descriptions

Thx

config/sso.go
trollkarlen: - Referrer leakage: victim's callback page app.com/callback?code=ABC&state=XYZ loads any third-party resource (Google Analytics, an ad pixel, a CDN font, a Facebook button). With a default/permissive Referrer-Policy, the browser sends Referer: https://app.com/callback?code=ABC&state=XYZ to those third parties. Attacker who controls or observes any of those endpoints (ad network bid stream, compromised analytics vendor, log access) reads the code.
- Open redirector at the client or AS: attacker crafts an auth URL with redirect_uri=https://app.com/legit-callback but chains it through an open redirect or a redirect_uri validation bug (path confusion, parameter pollution — Innocenti et al. 2023 found this in 10/16 major IdPs) so the response lands at attacker.com instead.
- Mix-up attack: attacker tricks the client into starting a flow against a malicious AS that the client also trusts. The malicious AS proxies the user to the honest AS. The honest AS issues a code and sends it back through the redirect — but because of the mix-up, the code (issued for app.com at honest-AS) ends up visible to the attacker.
- Logs / proxy / error tracker: full URL with code captured in WAF logs, reverse-proxy access logs, Sentry breadcrumbs, browser history on a shared machine, or browser-profile sync.

So with [Supply Chain Attacks](https://auth0.com/blog/secure-nodejs-applications-from-supply-chain-attacks/) being hot now this is even more of an issue.

Sites that support it, the vast majority — Keycloak, Auth0, Okta, Google, GitHub, Dex, Azure AD, etc. all accept it on confidential clients.

So its a real issue and if you have the chance as an Enterprice(or as an open-source software) to keep you code as safe as possible, and from this project push the awareness of issues like this I think you should.

MasonM: > Referrer leakage: victim's callback page app.com/callback?code=ABC&state=XYZ loads any third-party resource (Google Analytics, an ad pixel, a CDN font, a Facebook button). 

Not an issue in Argo Workflows. We don't load any third-party resources, nor do we allow users to inject any. 

> Open redirector at the client

Not an issue in Argo Workflows. We do strict validation on the redirect URI.

I'm going to stop there. Everything you listed are valid vulnerabilities in other clients, and I'm sure that's the reason RFC 9700 recommends everyone use PKCE, but that doesn't mean they're valid in Argo Workflows. 

If you can't think of an attack scenario that applies specifically to Argo Workflows, then please change the default to be PKCE disabled. Also, I think the option should be renamed to `enablePKCEAuthentication` for consistency with Argo CD: https://github.com/argoproj/argo-cd/blob/998aac60b060a09e62ffeccb25da7a1ff79bc1b1/util/settings/settings.go#L219

trollkarlen: You still have the redirect and/or proxy and logs leaks.

Supply chain attacks from go packages can also steal the secrets.

But I can agree that the less secure option can be default, but then I suggest a log about this during startup.

MasonM: I don't agree that leaving PCKE disabled is "less secure" given our threat model (which is different than the threat model for many other clients), and I don't think that's how it should be communicated in the documentation or the logs. I think it's okay to say "enabling PCKE is recommended and required with certain authorization servers", but anything that implies leaving PKCE disabled is a vulnerability should be backed up with an explanation of how that pertains specifically to Argo Workflows

trollkarlen:
I think all comments is fixed now

MasonM:
Thanks @trollkarlen. I'll review this by the end of the week.

server/auth/sso/sso.go
```diff
@@ -227,10 +228,40 @@ func newSso(
...
+}
+
+// decodeStateCookieValue is the inverse of encodeStateCookieValue. It returns
+// the PKCE code verifier (empty when not present) and the final redirect URL.
+func decodeStateCookieValue(value string) (verifier, finalRedirectURL string) {
+	if before, after, ok := strings.Cut(value, stateCookieSeparator); ok {
+		return before, after
+	}
+	return "", value
+}
```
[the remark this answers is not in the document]
MasonM: This is correct, and I can reproduce this by visiting the following URL in the dev environment: https://localhost:8080/login?redirect=%2Fworkflows%3Fnamespace%3Dargo%26sidePanel%3Dsubmit-new-workflow%26template%3Ddelightful-python%26parameters%255Bmessage%255D%3Dtest:hello2%26phase%3DSucceeded%26limit%3D50

Image

I'm not a fan of the proposed fix though. Choosing a different delimiter seems better. ArgoCD uses a newline: https://github.com/argoproj/argo-cd/blob/19d99f5d799e5ec3ae9399487f32459af9a4e5f8/util/oidc/oidc.go#L301

but raw newlines aren't allowed in cookies. ArgoCD gets around that by encrypting the cookie, which it seems we should be doing anyway (see below)

server/auth/sso/sso.go
```diff
@@ -242,17 +273,30 @@ func (s *sso) HandleRedirect(w http.ResponseWriter, r *http.Request) {
...
+	var verifier string
+	if s.pkceEnabled {
+		verifier = oauth2.GenerateVerifier()
+		authCodeOptions = append(authCodeOptions, oauth2.S256ChallengeOption(verifier))
+	}
+
 	http.SetCookie(w, &http.Cookie{
 		Name:     state,
-		Value:    finalRedirectURL,
+		Value:    encodeStateCookieValue(verifier, finalRedirectURL),
```
MasonM, line 292: This cookie isn't encrypted, unlike the `authorization` cookie above. This basically means PKCE isn't providing any defense-in-depth, because the point of PKCE is to have a challenge that's only known to the client. Per [RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636):
> Typically, the "code_challenge" and "code_challenge_method" values
   are stored in encrypted form in the "code" itself but could
   alternatively be stored on the server associated with the code.  The
   server MUST NOT include the "code_challenge" value in client requests
   in a form that other entities can extract.

You should be able to use the `s.encrypter` above to encrypt the cookie, like we're doing with `authorization`.

MasonM (commented):
Thanks! Besides the issues above, one thing I think we need is the ability to quickly test this in the dev environment. We get a lot of bug reports related to the SSO integration, and the first step is typically trying to reproduce it to see if it's an actual bug or a misconfiguration (most of the time it's a misconfiguration). I went ahead and created a commit that implements that and improves the docs a little, which you're free to pull into your branch: https://github.com/MasonM/argo-workflows/commit/4ec6f4ad8a0ba8ab3bbc2807bf119022dbf79506

PR #16095, opened 2026-05-14
fix(ui): show full tag value on hover in TagsInput. Fixes #16096

### LINKED ISSUE #16096: Show full tag value on hover in TagsInput
# Summary

The `TagsInput` component truncates long tag values via CSS, but does not provide any way to view the full value. Adding a `title` attribute to each tag element would show the full value as a native browser tooltip on hover.

## Use Cases

- When filtering workflows by labels, users work with long key=value pairs (e.g., `workflows.argoproj.io/workflow-template=workflow-template-with-a-long-name`) that get truncated, making it difficult to verify the full content.
- When reviewing workflow metadata or annotations, users need to confirm the exact tag value without clicking into edit mode or inspecting the DOM.

---

nakatani-yo:
Fixes #16096

### Motivation

The `TagsInput` component truncates long tag values via CSS, but hovering over a truncated tag does not reveal the full value. Users must click into edit mode or inspect the DOM to verify the exact content, which interrupts their workflow when filtering by labels or reviewing metadata.

### Modifications

- Added `title={tag}` attribute to the `<span className='tags-input__tag'>` element in `ui/src/shared/components/tags-input/tags-input.tsx` so the browser displays the full tag value as a native tooltip on hover.
- Added a unit test in `ui/src/shared/components/tags-input/tags-input.test.tsx` to verify the `title` attribute is set correctly.

### Verification

- Ran the new unit test with `jest` and confirmed it passes.
- Manually verified in the browser that hovering over a truncated tag in the workflow list filter shows the full value as a tooltip.

image

### Documentation

Not needed. This is a small UX improvement using a native HTML attribute with no new configuration or user-facing API changes.

### AI

Codex, Gemini


Joibel (approved):

nakatani-yo:
@Joibel 
Thank you for your review.

PR #16097, opened 2026-05-14
feat(server): allow SSO ServiceAccount to override UI default namespace

trollkarlen:
Adds a 'workflows.argoproj.io/default-namespace' annotation, read from the matched SSO-mapping ServiceAccount in gatekeeper. When set, its value replaces ServiceAccountNamespace in the claims returned via /api/v1/userinfo, so the UI lands the user in their tenant namespace instead of the install namespace. Opt-in; absent annotation preserves prior behavior.

### Fixes
So the default namespace can be a user scoped NS and not install NS 

### Motivation
For multi tenancy with one server and oauth and dynamic provisioning 

### Verification
Added tests

### Documentation
Added docs

### AI
yes, opencode

### DIFF
--- a/docs/argo-server-sso.md
+++ b/docs/argo-server-sso.md
@@ -112,6 +112,40 @@ As of Kubernetes v1.24, secrets for a service account token are no longer automa
 Therefore, service account secrets for SSO RBAC must be created manually.
 See [Service Account Secrets](service-account-secrets.md) for detailed instructions.
 
+### Per-user default UI namespace
+
+By default, the Argo UI's "current namespace" is initialized from the matched
+SSO ServiceAccount's own `metadata.namespace`. SSO-mapping ServiceAccounts
+must live in the install namespace (Argo only walks that namespace when
+matching `rbac-rule` annotations against claims), so in a multi-tenant install
+where users have permissions only in their **tenant** namespace, the UI lands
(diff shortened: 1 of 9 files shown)


trollkarlen:
Please review this

trollkarlen:
I think the failing test is a test env failure(s).

docs/argo-server-sso.md
```diff
@@ -112,6 +112,40 @@ As of Kubernetes v1.24, secrets for a service account token are no longer automa
 Therefore, service account secrets for SSO RBAC must be created manually.
 See [Service Account Secrets](service-account-secrets.md) for detailed instructions.
 
+### Per-user default UI namespace
+
+By default, the Argo UI's "current namespace" is initialized from the matched
```
tico24, line 117: The docs guidelines ask for one sentence per line of markdown, and `MD013` (line length) is off in this repo, so there is no need to hard-wrap. The rest of this file already does it, e.g. the `As of Kubernetes v1.24...` lines just above. Mind reflowing this new section to one sentence per line? Keeps the diff clean when someone later edits a single sentence.

tico24 (commented):
Only reviewed the docs here, since that is the part I am codeowner of (`docs/`, `.features/`). I have not looked at the Go server changes, the UI changes, or `.spelling`.

PR #16100, opened 2026-05-15
fix: allow cron aliases in schedule validation

### LINKED ISSUE #16052: Months in cron expression are rejected by validation
### Pre-requisites

Hello,

I've just updated to the latest version, coming from v3, and I see a regression in the expression allowed in the cron scheduling. 
Before I was using this expression : 00 8 3 Jan,Apr,Jul,Oct *
Which is valid and were working on the v3, this is a screenshot from the v3:

Image

The validation of the workflow is rejected with this error:
  Error: UPGRADE FAILED: cannot patch "density-map-quarterly" with kind CronWorkflow: CronWorkflow.argoproj.io "density-map-quarterly" is invalid: spec.schedules[0]: Invalid value: "00 8 3 Jan,Apr,Jul,Oct *": spec.schedules[0] in body should match '^(@(yearly|annually|monthly|weekly|daily|midnight|hourly)|@every\s+([0-9]+(ns|us|µs|ms|s|m|h))+|([0-9*,/?-]+\s+){4}[0-9*,/?-]+)$'

### Version(s)

v4.0.5

```YAML
kind: CronWorkflow
metadata:
  name: test
spec:
  schedules:
   - 00 8 3 Jan,Apr,Jul,Oct *
  timezone: UTC
  concurrencyPolicy: Forbid
  successfulJobsHistoryLimit: 7
  failedJobsHistoryLimit: 7
  startingDeadlineSeconds: 0
  workflowSpec:
    entrypoint: date
    templates:
    - name: date
      container:
        image: alpine:3.23
        command: [sh, -c]
        args: ["date; sleep 90"]
```

```text
N/A
```

```text
N/A
```

PradeepSD476:
Fixes #16052

### Motivation

CronWorkflow validation rejected valid cron aliases such as Jan and Mon-Fri even though they are supported by the runtime cron parser.

### Modifications

Updated the validation regex to allow alphabetic cron aliases in CronWorkflow schedules.

This keeps semantic validation delegated to the runtime cron parser while preventing valid aliases from being rejected during CRD validation.

### Verification

Verified the following schedules successfully:

*/1 * * May Fri-Sat
0 0 * * MON-FRI
0 8 3 JAN *

Also verified invalid aliases such as May,Hello are rejected by the runtime cron parser with a message:  cron schedule */1 * * May,Hello Fri,Sat is malformed: failed to parse int from Hello: strconv.Atoi: parsing "Hello": invalid syntax.

### Documentation

No documentation changes were required.

### AI

Used ChatGPT for debugging discussion and PR preparation.


MasonM (approved):
Thanks, this looks right to me. Here's the list of the aliases the parser allows: https://github.com/robfig/cron/blob/bc59245fe10efaed9d51b56900192527ed733435/spec.go#L26-L48

This could be made stricter, but that doesn't seem necessary to me.

@Joibel This is related to https://github.com/argoproj/argo-workflows/pull/15028

PradeepSD476:
@MasonM  Thanks for the review and approval.

I noticed this PR was referenced from #15028. Please let me know if there are any additional changes you'd like me to do.

MasonM (approved):

isubasinghe (approved):

Joibel (approved):

PR #16102, opened 2026-05-15
fix: retry for database transaction errors. Fixes #16101

### LINKED ISSUE #16101: Database sync: transient SQLSTATE 40001 misclassified as "bug"
### Pre-requisites

When a workflow uses a database-backed mutex or semaphore and the SERIALIZABLE transaction in TryAcquire hits a 40001 serialization conflict, the workflow node is permanently marked Error instead of being retried.

  acquire logs the SQL error but returns only false, dropping the cause. tryAcquireImpl then wraps the bare failure as "bug: failed to acquire something that should have been checked:  ". The retry loop in TryAcquire substring-matches for serialization/dependencies/deadlock/rollback — none of which appear
  in the  string so the loop falls through to the non-retryable branch on attempt 1.

  Net effect: a transient PostgreSQL serializable conflict becomes a permanent workflow-node
  error.

### Version(s)

:latest

```YAML
impossible to reproduce on a single controller setup.
```

```text
N/A
```

```text
N/A
```

isubasinghe:
Fixes #16101
 
 ### Motivation

  Workflows using database-backed mutexes or semaphores fail permanently on PostgreSQL SQLSTATE 40001 during lock acquisition. 40001 is a normal serializable-isolation
  conflict; the retry loop in `TryAcquire` is built for it but never fires.

  `databaseSemaphore.acquire` drops the SQL error and returns plain `false`. `tryAcquireImpl` wraps the bare failure as `bug: failed to acquire something that should have
  been checked: ` (empty payload). The retry detector substring-matches `serialization`/`dependencies`/`deadlock`/`rollback` against that string, finds nothing, and gives
  up after one attempt.

  ### Modifications

  - `acquire` returns `(bool, error)`; `tryAcquire` returns `(bool, string, error)`. SQL error propagates.
  - `tryAcquireImpl` returns the underlying error directly. The `bug:` wrap stays for its original case: `(false, nil)` after `checkAcquire` approved.
  - Retry loop rewritten with `retry.OnError` and `wait.Backoff{Duration: 10ms, Factor: 2, Steps: 5, Jitter: 0.5, Cap: 600ms}`. Old loop had zero delay; jitter avoids
  replica lockstep.
  - Substring check extracted into `isRetryableSyncError`.
  - `prioritySemaphore` and the `semaphore` interface updated to match. In-memory `acquire` returns nil error. `Initialize` callers still drop the error — pre-existing,
  separate change.

  ### Verification
Test pass

  ### Documentation

  None needed.

  ### AI

  Claude Code (Claude Opus 4.7) used for investigation, tests, fix, and this description. Reviewed by hand; `Cap` raised from 200ms to 600ms during review.

### DIFF
--- a/workflow/sync/database_mutex_test.go
+++ b/workflow/sync/database_mutex_test.go
@@ -41,18 +41,18 @@ func TestDatabaseMutexAcquireRelease(t *testing.T) {
 			require.NoError(t, mutex.addToQueue(ctx, "default/workflow2", 0, now.Add(time.Second)))
 
 			// First acquisition should succeed
-			acquired, _ := mutex.tryAcquire(ctx, "default/workflow1", tx)
+			acquired, _, _ := mutex.tryAcquire(ctx, "default/workflow1", tx)
 			assert.True(t, acquired, "First acquisition should succeed")
 
 			// Second acquisition should fail
-			acquired, _ = mutex.tryAcquire(ctx, "default/workflow2", tx)
+			acquired, _, _ = mutex.tryAcquire(ctx, "default/workflow2", tx)
 			assert.False(t, acquired, "Second acquisition should fail")
 
 			// Release the mutex
 			mutex.release(ctx, "default/workflow1")
 
 			// Now acquisition should succeed again
-			acquired, _ = mutex.tryAcquire(ctx, "default/workflow2", tx)
+			acquired, _, _ = mutex.tryAcquire(ctx, "default/workflow2", tx)
 			assert.True(t, acquired, "Acquisition after release should succeed")
 		})
 	}
@@ -76,11 +76,11 @@ func TestDatabaseMutexQueueOrder(t *testing.T) {
 			require.NoError(t, mutex.addToQueue(ctx, "default/workflow1", 0, now))
 			require.NoError(t, mutex.addToQueue(ctx, "default/workflow2", 0, now.Add(time.Second)))
 
-			acquired, _ := mutex.tryAcquire(ctx, "default/workflow2", tx)
+			acquired, _, _ := mutex.tryAcquire(ctx, "default/workflow2", tx)
 			assert.False(t, acquired, "Second workflow should not acquire the mutex")
 
 			// Acquire the first one
-			acquired, _ = mutex.tryAcquire(ctx, "default/workflow1", tx)
+			acquired, _, _ = mutex.tryAcquire(ctx, "default/workflow1", tx)
 			assert.True(t, acquired, "First workflow should acquire the mutex")
 
 			// Release it - this should notify the next one
--- a/workflow/sync/database_mutex_test.go
+++ b/workflow/sync/database_mutex_test.go
@@ -41,18 +41,21 @@ func TestDatabaseMutexAcquireRelease(t *testing.T) {
 			require.NoError(t, mutex.addToQueue(ctx, "default/workflow2", 0, now.Add(time.Second)))
 
 			// First acquisition should succeed
-			acquired, _, _ := mutex.tryAcquire(ctx, "default/workflow1", tx)
+			acquired, _, err := mutex.tryAcquire(ctx, "default/workflow1", tx)
+			require.NoError(t, err)
 			assert.True(t, acquired, "First acquisition should succeed")
 
 			// Second acquisition should fail
-			acquired, _, _ = mutex.tryAcquire(ctx, "default/workflow2", tx)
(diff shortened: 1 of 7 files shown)


workflow/sync/semaphore.go
```diff
@@ -250,10 +251,10 @@ func (s *prioritySemaphore) tryAcquire(ctx context.Context, holderKey string, tx
 			"limit":     limit,
 		}).Info(ctx, "acquired")
 		s.notifyWaiters(ctx)
-		return true, ""
+		return true, "", nil
 	}
 	logger.WithField("lockHolder", s.lockHolder).Debug(ctx, "Current semaphore Holders")
-	return false, msg
+	return false, msg, nil
```
isubasinghe, line 257: Should the errror from the s.acquire call be passed here?

isubasinghe: no

workflow/sync/sync_manager.go
```diff
@@ -320,50 +323,61 @@ func (sm *Manager) TryAcquire(ctx context.Context, wf *wfv1.Workflow, nodeName s
...
+// text.
+func isRetryableSyncError(err error) bool {
+	if err == nil {
+		return false
+	}
+	s := strings.ToLower(err.Error())
+	return strings.Contains(s, "serialization") ||
+		strings.Contains(s, "dependencies") ||
+		strings.Contains(s, "deadlock") ||
+		strings.Contains(s, "rollback")
```
eduardodbr, line 382: I know you just moved the code to a func but I am wondering if using sql error codes would be possible? I

eduardodbr (approved):
LGTM, just minor question

workflow/sync/sync_manager.go
```diff
@@ -244,8 +245,10 @@ func (sm *Manager) Initialize(ctx context.Context, wfs []wfv1.Workflow) {
 						continue
 					}
 					key := getUpgradedKey(&wf, holders, level)
-					if sem != nil && sem.acquire(ctx, key, sm.dbInfo.SessionProxy) {
-						sm.log.WithFields(logging.Fields{"key": key, "semaphore": holding.Semaphore}).Info(ctx, "Lock acquired")
+					if sem != nil {
+						if acquired, _ := sem.acquire(ctx, key, sm.dbInfo.SessionProxy); acquired {
+							sm.log.WithFields(logging.Fields{"key": key, "semaphore": holding.Semaphore}).Info(ctx, "Lock acquired")
+						}
```
Joibel, line 255: Feels like logging the dropped err here would be helpful.

workflow/sync/sync_manager.go
```diff
@@ -320,50 +323,61 @@ func (sm *Manager) TryAcquire(ctx context.Context, wf *wfv1.Workflow, nodeName s
...
-				already, updated, msg, failedLockName, txErr = sm.tryAcquireImpl(ctx, wf, sp, holderKey, failedLockName, syncItems, lockKeys)
-				sm.log.WithFields(logging.Fields{
-					"holderKey": holderKey,
-					"attempt":   retryCounter + 1,
-				}).Info(ctx, "TryAcquire - transaction completed")
-				return txErr
+		// Backoff bounds: sm.lock is held for the whole loop, so cap each sleep
+		// modestly. Jitter prevents a fleet of replicas from retrying in lockstep
+		// after a shared conflict burst.
+		backoff := wait.Backoff{
```
Joibel, line 333: 👍

workflow/sync/database_mutex_test.go
```diff
@@ -41,18 +41,18 @@ func TestDatabaseMutexAcquireRelease(t *testing.T) {
 			require.NoError(t, mutex.addToQueue(ctx, "default/workflow2", 0, now.Add(time.Second)))
 
 			// First acquisition should succeed
-			acquired, _ := mutex.tryAcquire(ctx, "default/workflow1", tx)
+			acquired, _, _ := mutex.tryAcquire(ctx, "default/workflow1", tx)
 			assert.True(t, acquired, "First acquisition should succeed")
```
Joibel, line 46: I feel like a few require.NoError(t, err) here would be a bit defensive. Every one, probably overkill

workflow/sync/sync_manager.go
isubasinghe: Done

workflow/sync/sync_manager.go
isubasinghe: I agree this is nicer, but its  also harder to do because we rely on upper db as a medium to abstract between the underlying databases we use.

Joibel (approved):

PR #16104, opened 2026-05-17
feat(executor): report artifact metadata before upload completes

### LINKED ISSUE #16091: Async artifact uploading
# Summary

Add an opt-in async output artifact upload mode.

Today, a workflow task does not become available to unrelated downstream work until its output artifact uploads finish. For large artifacts, this can unnecessarily block DAG or steps branches that do not consume those artifacts.

The proposal is to let Argo report output metadata, including final artifact locations, as soon as those locations are known, then continue uploading artifact bytes in the wait container. Tasks that consume those artifacts should still wait until the referenced upload has completed successfully. Workflow completion and pod cleanup should also wait for pending async uploads by default.

  ## Use Cases

Our main use case is around build cache; allowing workflow to continue and upload the cache in the background can save precious minutes in case of large builds (Rust, big UI node projects).

  ---
  
SimbaKingjoe:
## Summary

Report artifact metadata (S3 keys, types) to the controller **before** the actual upload completes. The wait container generates artifact keys from the deterministic archive location, reports outputs immediately, then uploads files concurrently via goroutines. This allows the controller to access artifact metadata earlier, and makes uploads non-blocking within the wait container.

## Fixes

#16091

## Motivation

Today, `SaveArtifacts` uploads all output artifacts synchronously before `ReportOutputs`. For large artifacts (build caches, ML models), this blocks the wait container and prevents the controller from seeing any output metadata until every byte is uploaded. For downstream DAG/Steps tasks that do NOT consume these artifacts, this is unnecessary delay.

This PR lays the foundation for full async artifact uploads by:
1. Decoupling key generation from file upload
2. Reporting metadata early
3. Making uploads concurrent within the container

## Modifications

**`workflow/executor/executor.go`** (+111/-13):
- `generateArtifactKey()` — new method that computes the artifact's S3/GCS key from the archive location, without uploading
- `StagedArtifact` — new struct pairing an artifact with its staged local file path
- `GenerateArtifactOutputs()` — stages artifact files and populates keys/types, returns `[]StagedArtifact`
- `SaveArtifactsAsync()` — uploads staged artifacts concurrently via goroutines with `sync.WaitGroup` + error aggregation
- `saveArtifactFromFile()` — refactored to call `generateArtifactKey` instead of inlining key generation

**`cmd/argoexec/commands/wait.go`** (+21/-1):
- Reordered wait container flow: `GenerateArtifactOutputs` → `ReportOutputs` → `SaveArtifactsAsync`
- Metadata reported immediately, uploads run concurrently after

**`workflow/executor/executor_test.go`** (+47):
- `TestGenerateArtifactKey` — verifies key generation and HasKey() idempotency
- `TestGenerateArtifactOutputs_EmptyOutputs` — empty outputs return nil
- `TestSaveArtifactsAsync_EmptyList` — nil/empty input returns nil

## Verification

```bash
go test ./workflow/executor/ -short    # all existing + new tests pass
go build ./workflow/executor/           # compiles
go build ./cmd/argoexec/commands/       # compiles
```

## Behavioral Impact

- **Backward compatible**: `FinalizeOutput` (which sets `report-outputs-completed=true`) is still deferred and runs after all uploads complete. Downstream DAG dependencies still wait correctly.
- **Metadata available earlier**: Controller's `taskResultReconciliation` copies outputs to NodeStatus regardless of completion flag (taskresult.go:124-143), so artifact keys are visible in UI/CLI as soon as `ReportOutputs` completes.
- **Concurrent uploads**: Multiple artifacts now upload in parallel goroutines instead of serial.

### DIFF
--- a/.features/pending/async-artifact-upload.md
+++ b/.features/pending/async-artifact-upload.md
@@ -1,0 +1,11 @@
+Description: Report artifact metadata before upload completes, with concurrent uploads.
+Author: [daixin1204](https://github.com/SimbaKingjoe)
+Component: General
+Issues: 16091
+
+This change separates artifact key generation from file uploads, allowing the
+wait container to report output metadata (S3 keys, types) to the controller
+immediately. Actual artifact uploads then run concurrently via goroutines with
+sync.WaitGroup synchronization. This lays the foundation for full async artifact
+uploading where downstream tasks that do not consume the artifacts can proceed
+without waiting for upload completion.
--- a/cmd/argoexec/commands/wait.go
+++ b/cmd/argoexec/commands/wait.go
@@ -10,6 +10,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/argoproj/argo-workflows/v4/cmd/argoexec/executor"
+	wfv1 "github.com/argoproj/argo-workflows/v4/pkg/apis/workflow/v1alpha1"
 	"github.com/argoproj/argo-workflows/v4/util/logging"
 	"github.com/argoproj/argo-workflows/v4/workflow/executor/tracing"
 )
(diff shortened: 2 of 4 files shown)


Joibel:
This breaks tests and doesn't really implement something usable. It's supposed to be an optional feature and requires a large amount of controller side design too.

Issue #15275, opened 2026-01-22
Argo Workflows UI v3.7.7: the GRAPH in Cron Workflows view shows "Cannot read properties of undefined (reading 'forEach')"


ncs-coder-ops:
### Pre-requisites

- [x] I have double-checked my configuration
- [x] I have tested with the `:latest` image tag (i.e. `quay.io/argoproj/workflow-controller:latest`) and can confirm the issue still exists on `:latest`. If not, I have explained why, **in detail**, in my description below.
- [x] I have searched existing issues and could not find a match for this bug
- [ ] I'd like to contribute the fix myself (see [contributing guide](https://github.com/argoproj/argo-workflows/blob/main/docs/CONTRIBUTING.md))

### What happened? What did you expect to happen?

I upgraded from v3.6.4 to v3.7.7 and noticed the addition of the GRAPH menu in the CronWorkflow view in the Argo Workflows UI. When I click on it I get the following output:

```
Cannot read properties of undefined (reading 'forEach')
 [Reload this page](javascript:document.location.reload();) to try again.

Stack Trace
TypeError: Cannot read properties of undefined (reading 'forEach')
    at https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:1671900
    at nl (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:1674085)
    at gi (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:563288)
    at Cs (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:577326)
    at kl (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:623096)
    at yc (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:611481)
    at vc (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:611409)
    at gc (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:611272)
    at ac (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:608107)
    at oc (https://argo-workflows.example-domain.com/main.5025d1c0e176f5e8a2e0.js:2:606663)
Component Stack

[...]

PR #16071, opened 2026-05-09
fix(ui): fix error on "Graph" tab when viewing CronWorkflow and add tests. Fixes #15275

MasonM:
Fixes #15275

### Motivation

When viewing a `CronWorkflow` that uses `workflowTemplateRef` to reference a `WorkflowTemplate`, clicking the "Graph" tab will generate the following error:
```
Cannot read properties of undefined (reading 'forEach')
```

Here's a minimal WorkflowTemplate and CronWorkflow to reproduce this:

```yaml

apiVersion: argoproj.io/v1alpha1
kind: WorkflowTemplate
metadata:
  name: hello-world
spec:
  entrypoint: print-message
  templates:
    - name: print-message
      container:
        image: busybox
        command: [echo]
        args: ["hello world"]
---
apiVersion: argoproj.io/v1alpha1
kind: CronWorkflow
metadata:
  name: daily-job
spec:
  schedules: ["0 2 * * *"]
  workflowSpec:
    workflowTemplateRef:
      name: hello-world
```

### Modifications

This has been present since https://github.com/argoproj/argo-workflows/pull/14034 and is happening in the `templates.forEach((template: Template) => {` line. The `templates` variable is set to `workflow.spec.templates`, and a `CronWorkflow` with `workflowTemplateRef` won't have `workflow.spec.templates` defined. The `setWorkflowFromRefrence()` function is supposed to prevent this by resolving the `workflowTemplateRef`, but because of the way the `useEffect()` block is structured, it wasn't being run.

To minimize breakage, I had Claude write a bunch of unit tests in commit 7139d8cc646d5921555c05e571812b87c393d87c. That required refactoring
`graph-viewer.tsx` a bit. This is easiest to review by ignoring whitespace (I manually edited Claude's output to structure it so the diff looks nice): https://github.com/argoproj/argo-workflows/pull/16071/changes?w=1

The error can be fixed by simply appending `|| []` to `const templates = workflow.spec.templates`, but that means the example `CronWorkflow` will show "Nothing to show". To really fix it so it shows the graph properly
required changing the `useEffect()` block. I also did some cleanup:
1. Removed the ad-hoc loading indicator and error displayer and used the standard `<Loading>` and `<ErrorNotice>` components.
2. Removed the `generateNamePostfix()` logic and simply show the workflow names without a randomly-generated suffix, since the non-determinism introduced by that complicates testing and doesn't seem useful to me.

### Verification
[...]

PR #16119, opened 2026-05-19
fix(ui): guard missing DAG dependency nodes in GraphViewer

Zakharden:
## Summary\n\nThis PR fixes a Graph Viewer crash when rendering DAG tasks with unresolved dependencies by guarding null lookup paths in dependency resolution.\n\n## Changes\n- In `GraphViewer`, guard dependency node lookups before reading `genre` in DAG tasks.\n- Skip dependency wiring when template metadata can’t be resolved instead of throwing\n  runtime errors.\n- Add regression test: `GraphViewer` rendering with a missing DAG dependency no longer throws and still renders the graph view.\n\n## Related issue\n\nCloses #15275

### DIFF
--- a/ui/src/shared/components/editors/graph-viewer.test.tsx
+++ b/ui/src/shared/components/editors/graph-viewer.test.tsx
@@ -1,0 +1,52 @@
+import {render} from '@testing-library/react';
+import React from 'react';
+
+import {GraphViewer} from './graph-viewer';
+
+jest.mock('../graph/graph-panel', () => ({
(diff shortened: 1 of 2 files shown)


MasonM (commented):
Thanks for your contribution. It looks like the linter is failing. I suggest waiting until https://github.com/argoproj/argo-workflows/pull/16071 is merged, which also has a `ui/src/shared/components/editors/graph-viewer.test.tsx` file with much more extensive tests.

Zakharden:
Fixed the DCO issue and the UI lint failure from eslint import ordering in `graph-viewer.test.tsx`. `corepack yarn --cwd ui lint` passes locally.

PR #16124, opened 2026-05-20
fix: change log level because behavior is expected

spaced:
Fixes log level because behavior is expected
- skipped step using `when` 
- optional artifacts

### Motivation

do not flood with `WARN` if behavior is expected

### Modifications

change log level to `INFO`

### Verification

scan log output


Joibel (approved):
Thanks for taking the time to do this

PR #16125, opened 2026-05-21
feat: workflowtasksets for resource monitors. fixes #16126

### LINKED ISSUE #16126: Run resource monitoring pods as agents.
# Summary

Re-use the existing support for http/plugins that run a per workflow pod to monitor resources created by the workflow we run. 

## Use Cases

* For massive fan out workflow of workflow patterns, there is a giant waste of resources if monitoring of these workflows are wanted/required.

---

isubasinghe:
Fixes #16126

### Motivation

  Resource templates currently spawn a wait pod per invocation just to poll the created object. For workflow-of-workflows that's a pod per child, which adds up fast.

  This moves the polling into the workflow's agent pod, same way HTTP and Plugin templates already work.

  ### Modifications

  - New node type `NodeTypeResourceMonitor`. The operator no longer creates a pod for resource templates; it puts the template on the WorkflowTaskSet and the agent picks
  it up.
  - Two new label keys: `workflows.argoproj.io/monitored-resource` (workflow name, used as the informer's selector) and `workflows.argoproj.io/monitored-resource-node-id`
  (used to route events back to the originating task).
  - Agent gets a `MonitoredResourceInformer`: one dynamic informer per GVR, label-filtered, with a single handler that dispatches by node-ID label.
  - Agent shells out to `kubectl` for the action, infers the GVR from the response, then watches. Terminal phase is posted when a success/failure condition matches.
  - Output parameter extraction supports `jsonPath`, `jqFilter`, and `default`.
  - `manifestFrom.artifact` is resolved inside the agent via the artifact driver. Single-file archives only; multi-file archives are rejected with a pointer back to the
  legacy path.
  - Agent pod gets a 64Mi tmpfs `/tmp` so kubectl scratch files work under `ReadOnlyRootFilesystem=true`.
  - UI: `ResourceMonitor` added to the DAG genre map.
  
  ### Verification
  - `make test`, including new informer unit tests and an agent integration test (fake dynamic client plus stubbed kubectl) that covers success match, failure match,
  JSONPath/JQ outputs, and the delete short-circuit.
  - Ran the new examples on a kind cluster against the quick-start manifests. Workflow-of-workflows runs without per-step wait pods. Stripped `list`/`watch` from the agent
   role and confirmed the failure mode matches what the docs say.

  ### Documentation

  `docs/workflow-rbac.md` has the new RBAC section with a worked example role. The walk-through points at it. The quick-start manifest matches the docs.

  ### AI

  The majority of the code was written by me. Claude Code (Opus) was used for boilerplate around the informer and for help drafting docs and this PR description.


isubasinghe:
These defaults are wayy too permisive to let the tests pass. 

Opening this PR not because this is the state I want it merged in but more to open a discussion about the direction to go in order to reduce the permissions here. 

We could also consider drop support for logs in this mode of execution perhaps.

workflow/executor/agent_resource_informer.go
```diff
@@ -0,0 +1,179 @@
...
+func (m *MonitoredResourceInformer) Watch(ctx context.Context, gvr schema.GroupVersionResource) error {
+	logger := logging.RequireLoggerFromContext(ctx)
+
+	m.mu.Lock()
+	inf, exists := m.informers[gvr]
+	if !exists {
+		factory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(
+			m.dynClient,
+			m.resyncPeriod,
+			m.namespace,
```
Joibel, line 87: If `m.namespace != createdObject.namespace` we're never going to observe it.

workflow/executor/agent_resource_informer.go
```diff
@@ -0,0 +1,179 @@
...
+			func(opts *metav1.ListOptions) {
+				opts.LabelSelector = m.labelSelector()
+			},
+		)
+		shared := factory.ForResource(gvr).Informer()
+		if _, err := shared.AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    func(obj any) { m.dispatch(ctx, obj, false) },
```
Joibel, line 94: This `ctx` is wrong for second and subsequent calls to Watch, it has captured the first caller for any given GVR.

workflow/controller/agent.go
```diff
@@ -147,20 +167,91 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
...
 	serviceAccountName := woc.execWf.Spec.ServiceAccountName
+	if woc.execWf.Spec.Executor != nil && woc.execWf.Spec.Executor.ServiceAccountName != "" {
+		serviceAccountName = woc.execWf.Spec.Executor.ServiceAccountName
+	}
 	tokenVolume, tokenVolumeMount, err := woc.getServiceAccountTokenVolume(ctx, serviceAccountName)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get token volumes: %w", err)
 	}
 
+	// Artifact-plugin sidecars run with the pod's AutomountServiceAccountToken
```
Joibel, line 186: This all fixes up the agent to "work with resources" but we've lost all the per template pod spec capabilities - principally the SA. This isn't acceptable.

workflow/executor/agent.go
```diff
@@ -385,6 +426,734 @@ func (ae *AgentExecutor) executePluginTemplate(ctx context.Context, tmpl wfv1.Te
...
+		return manifest, nil
+	}
+	if action == "patch" && mergeStrategy == "json" {
+		var ops []map[string]any
+		if err := yaml.Unmarshal([]byte(manifest), &ops); err != nil {
+			return "", fmt.Errorf("parse json patch: %w", err)
+		}
+		escapedMonitored := strings.ReplaceAll(common.LabelKeyMonitoredResource, "/", "~1")
+		escapedNodeID := strings.ReplaceAll(common.LabelKeyMonitoredResourceNodeID, "/", "~1")
+		// "add" on /metadata/labels is a no-op overwrite when labels already
```
Joibel, line 448: add isn't a "no-op", I'm not sure what that part is trying to say. It replaces if exists IIRC

examples/resource-pod-spec-patch.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: argoproj.io/v1alpha1
```
Joibel, line 1: Does this even work, it feels unlikely.

Joibel (changes requested):
I think, as is, this can't fly.

The old model of SA per pod is very useful for some environments, and making this the only path to use resource templates is problematic. The SA on the resource template is just silently ignored, and that's no good.

PR #15217, opened 2026-01-05
fix(ui): workflow list sidebar filter correctly update URL query parameters for multi-value #15213 

fruitriin:
Fixes #15213 

### Motivation

The Workflows list page does not correctly update URL query parameters when using multi-value filters (Phase, Labels):

1. **Multiple selections not preserved**: Selecting multiple phases (e.g., Running + Failed) only shows the last one in the URL
2. **Unchecking not reflected**: Unchecking a filter doesn't remove it from the URL

### ui/src/workflows/components/workflows-list/workflows-list.tsx
- Start with empty `URLSearchParams` instead of inheriting existing params
- Explicitly preserve only non-managed parameters (e.g., `sidePanel`)
- Build all managed parameters from scratch

### ui/src/shared/history.ts
- Change `queryParams.set()` to `queryParams.append()` for `extraSearchParams` processing
- This allows multiple values with the same key (e.g., `?phase=Running&phase=Failed`)

### Verification

- [x] `yarn lint` passes
- [x] `yarn build` succeeds
  - [x] Select single phase → URL reflects correctly
  - [x] Select multiple phases → All phases in URL (`?phase=Running&phase=Failed`)
  - [x] Uncheck a phase → Phase removed from URL
  - [x] Labels filter works correctly with multiple selections
  - [x] Name filter works correctly
  - [x] Page refresh with params → Filters restored correctly
  - [x] Other pages (Workflow Templates, Cron Workflows, etc.) work correctly

PR #16139, opened 2026-05-23
fix(ui): workflow list sidebar filter correctly update URL query parameters for multi-value. Fixes #15213

### LINKED ISSUE #15213: UI: Workflows list Phase/Labels filters not correctly reflected in URL
### Pre-requisites

When using multi-value filters (Phase, Labels) on the Workflows list page, the URL query parameters are not updated correctly:

1. Selecting multiple phases (e.g., Running + Failed) only shows the last one in the URL
2. Unchecking a filter doesn't remove it from the URL
3. Labels and other multi-value parameters have the same issue

I expected the URL to accurately reflect all selected filters, for example `?phase=Running&phase=Failed` when both are checked.

### Version(s)

:latest cf749355bd5eead70552611734ddb9135da63836

1. Navigate to the Workflows list page
2. Check "Running" in the Phase filter → URL shows `?phase=Running`
3. Also check "Failed" → Expected: `?phase=Running&phase=Failed`, Actual: `?phase=Failed` only
4. Uncheck "Running" → `phase=Running` remains in the URL

### Root Cause

Two issues were identified:

In `workflows-list.tsx`, the useEffect inherits all existing query parameters and appends new values without clearing old ones.

In `history.ts`, the shared `historyUrl` function uses `set()` instead of `append()` when processing `extraSearchParams`, causing multiple values with the same key to be overwritten.

### Proposed Fix

1. In `workflows-list.tsx`: Start with empty `URLSearchParams` and explicitly preserve only non-managed parameters
2. In `history.ts`: Change `set()` to `append()` for `extraSearchParams` processing
```

myzk-a:
Fixes #15213

### Motivation

The Workflows list page does not correctly update URL query parameters when using multi-value filters (Phase, Labels):

1. **Multiple selections not preserved**: Selecting multiple phases (e.g., Running + Failed) only shows the last one in the URL
1. **Unchecking not reflected**: Unchecking a filter doesn't remove it from the URL

### Modifications

 This PR continues the work from #15217. Thanks to the original author for the investigation and implementation in that PR.

- Preserve repeated query parameters passed through `historyUrl` by appending `extraSearchParams` instead of replacing existing values with the same key; add a unit test for this behavior.
- Keep single-value Workflows list query parameters (`offset`, `limit`, name filter, `createdAfter`, and `finishedBefore`) as single values when rebuilding the URL.
- Prevent `?namespace=` from appearing twice in the URL when the namespace is already set as a named parameter and also present in `extraSearchParams`; add a unit test for this behavior.
- Fix duplicate `parameters[*]` query parameters accumulating in the URL when the sidePanel is open: use `params.set` instead of `params.append` so that a parameter key appearing multiple times in the current URL is collapsed to one value; add a regression test for this behavior.
- Add Workflows list tests that verify:
  - selecting multiple phase and label filters adds repeated query parameters to the URL;
  - removing one selected phase and one selected label removes only the corresponding query parameters;
  - opening a URL with repeated phase and label query parameters restores the selected filters in the UI.

### Verification

- Selecting multiple phases should be reflected in the URL.
- Deselecting a phase should be reflected in the URL.

https://github.com/user-attachments/assets/e2729dd6-c7d8-4654-8d21-bdbe521a7e69

- Selecting multiple labels should be reflected in the URL.
- Deselecting a label should be reflected in the URL.

https://github.com/user-attachments/assets/45fc99fd-fc54-4990-9f17-55e05d02efec

### AI

I used Codex and Claude Code to help understand the related issue and pull request context, including #15213 and #15217, and to help prepare the test code for this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow list filtering so selected phases, labels, and other filters update the URL consistently.
  * Fixed duplicated query values when opening the side panel or changing filters, including cases with repeated parameters already in the URL.
  * Ensured filter selections from the URL are restored correctly and remain in sync with the visible filter state.


myzk-a:
The PR title check is failing with `Bad credentials` from the GitHub REST
  API, not with a semantic title validation error. Could a maintainer re-run
  the job or check the workflow token configuration?

PR #16140, opened 2026-05-25
fix: added error tolerant list/watch

### LINKED ISSUE #11278: One invalid WorkflowTemplate make all WorkflowTemlate invisible
### Pre-requisites

- If we have one invalid `WorkflowTemplates` the `argo-server` returns error response `InternalServeError`, then entire `WorkflowTemplates` are invisible.

- image
- cli
```console
$ argo template -n argo-workflows list
2023/06/29 21:08:17 rpc error: code = Internal desc = json: cannot unmarshal string into Go struct field Container.items.spec.templates.container.command of type []string
```

### Version

latest(v3.4.8)

```YAML
kind: WorkflowTemplate
metadata:
  name: good
  namespace: argo-workflows
spec:
  entrypoint: whalesay-template
  templates:
    - name: whalesay-template
      container:
        image: docker/whalesay
        command: [cowsay]
---
kind: WorkflowTemplate
metadata:
  name: bad
  namespace: argo-workflows
spec:
  entrypoint: whalesay-template
  templates:
    - name: whalesay-template
      container:
        image: docker/whalesay
        command: this-is-invalid


-
```

```text
Logs from workflow server


argo-workflows-server-69ffff554c-7qskg argo-server time="2023-06-29T12:08:03.355Z" level=info duration=6.17825ms method=GET path=/api/v1/workflows/argo size=52 status=0
argo-workflows-server-69ffff554c-7qskg argo-server time="2023-06-29T12:08:17.239Z" level=error msg="finished unary call with code Internal" error="rpc error: code = Internal desc = json: cannot unmarshal string into Go struct field Container.items.spec.templates.container.command of type []string" grpc.code=Internal grpc.method=ListWorkflowTemplates grpc.service=workflowtemplate.WorkflowTemplateService grpc.start_time="2023-06-29T12:08:17Z" grpc.time_ms=2.287 span.kind=server system=grpc
argo-workflows-server-69ffff554c-7qskg argo-server time="2023-06-29T12:08:17.239Z" level=info duration=2.655417ms method=GET path=/api/v1/workflow-templates/argo-workflows size=140 status=500
```

```text
N/A
```

isubasinghe:
Fixes #11278

  ### Motivation

  On clusters running the minimal CRDs (no admission-time schema validation), a single malformed resource can be written to etcd. The typed client decodes the whole `List`
  response with one `json.Unmarshal`, so one bad object 500s every list API call (and breaks watches) for that resource type — one invalid WorkflowTemplate makes all
  WorkflowTemplates invisible in the UI/CLI.

  ### Modifications

  - Added `TolerantList` and `TolerantWatch` in `server/utils`: list/watch via the dynamic client and decode each item individually, logging and skipping items that fail
  to decode instead of failing the whole request.
  - Per-item decoding goes through a JSON marshal/unmarshal roundtrip rather than `runtime.DefaultUnstructuredConverter.FromUnstructured`, since the latter uses reflection
  and bypasses custom `json.Unmarshaler` implementations (`ParallelSteps`, `Item`, `AnyString`, etc.) — reflection-based conversion silently dropped every workflow with
  `steps`.
  - Bookmark watch events are decoded through the same path into typed objects; the previous pass-through of `*unstructured.Unstructured` made the reflector reject
  bookmarks, stalling watch-list sync (list APIs returned `{"items":null}` for ~60s after argo-server start).
  - Switched the workflow, workflow template, cluster workflow template, cron workflow, and event binding servers (plus the kube lister) to the tolerant list/watch,
  threading a `dynamic.Interface` through the server constructors and auth context.

  ### Verification

  New unit tests in `server/utils` covering: malformed items skipped while valid items are returned, list errors propagated, custom unmarshalers preserved (workflow with
  `ParallelSteps`), and bookmark event conversion. Updated existing server tests to provide a fake dynamic client. Verified E2E that list APIs return workflows immediately
  after server start.

  ### Documentation

  No user-facing behavior change beyond malformed resources being skipped (and logged) instead of failing the request — no docs needed.

  ### AI

  Claude Code was used to help write code, tests, and commit messages in this PR.


Joibel:
### Code review

Found 6 issues (none are merge blockers — the change correctly fixes #11278 and is well-tested):

1. **`ReceiveEvent` still uses the intolerant typed client.** The PR makes `ListWorkflowEventBindings` tolerant, but the hotter inbound-event path still lists via `auth.GetWfClient(...).List()`. A single malformed `WorkflowEventBinding` here will still 500 every inbound event — the exact failure class this PR exists to prevent.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/event/event_server.go#L73-L78

2. **`CountWorkflows` now fully decodes every workflow just to return `len()`.** It delegates to `ListWorkflows`, which runs the per-item JSON marshal/unmarshal decode and then discards the objects. Because the list handler calls count then list, large namespaces pay the expensive per-item decode twice per request.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/workflow/store/lister.go#L38-L44

3. **`TolerantWatch` doc comment is inaccurate.** It says "Error and Bookmark events are forwarded as-is," but Bookmark events are not forwarded as-is — they are converted from `*unstructured.Unstructured` to a typed `*T` via `decodeUnstructured`. Only Error events are truly forwarded as-is.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/utils/tolerant_watch.go#L29-L33

4. **The bookmark test doesn't exercise the branch it guards, and `ListMeta` propagation is untested.** `TestTolerantWatch_ConvertsBookmarkEvent` sends a metadata-only bookmark that decodes via the success path, so the decode-failure fallback below (issue 5) is never hit — the test passes for the wrong reason.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/utils/tolerant_watch_test.go#L79-L100

   Separately, every `TolerantList` test discards the returned `meta` with `_`, so the `Continue` / `ResourceVersion` / `RemainingItemCount` pagination machinery has no assertions — a dropped continue token would pass all tests.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/utils/tolerant_list_test.go#L49-L51

5. **Bookmark decode-failure fallback drops `resourceVersion`, and the comment overstates it.** On decode failure for a bookmark, the code emits a zero-value object; `item` was never populated, so `resourceVersion` is empty. The comment claims the fallback carries "the bookmark's metadata," which it does not. Latent only (well-formed bookmarks decode via the success path), but the comment is misleading and the fallback should copy `un.GetResourceVersion()` onto the emitted object.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/utils/tolerant_watch.go#L105-L112

6. **Repeated `gvr`-build + `*List` re-wrap boilerplate across ~6 call sites.** Because `TolerantList` returns `(items, meta)` rather than a typed `*List`, every caller repeats `gvr := ...WithResource(...Plural)` → `TolerantList[...]` → `&XxxList{ListMeta: meta, Items: items}`. Non-blocking; Go generics can't fully collapse this, but it's worth noting.

https://github.com/argoproj/argo-workflows/blob/0a3fd04eea24ef04e14c3282e86785226b30088d/server/utils/tolerant_list.go#L45-L86

Joibel:
@isubasinghe - the API test failures are real I think

server/workflow/store/lister.go
```diff
@@ -15,25 +17,26 @@ type WorkflowLister interface {
...
 	if err != nil {
 		return nil, err
 	}
-	return wfList, nil
+	return &wfv1.WorkflowList{ListMeta: meta, Items: items}, nil
 }
 
 func (k *kubeLister) CountWorkflows(ctx context.Context, namespace, nameFilter, createdAfter, finishedBefore string, listOptions metav1.ListOptions) (int64, error) {
-	wfList, err := k.wfClient.ArgoprojV1alpha1().Workflows(namespace).List(ctx, listOptions)
+	wfList, err := k.ListWorkflows(ctx, namespace, nameFilter, createdAfter, finishedBefore, listOptions)
```
Joibel: **`CountWorkflows` now fully decodes every workflow just to count them, and silently undercounts.**

By delegating to `ListWorkflows`, the count path now runs `TolerantList`, which does the per-item marshal/unmarshal roundtrip (the 3-pass cost your own `ponytail:` note in `tolerant_list.go` flags) and allocates a typed `wfv1.Workflow` for every item — only to return `len(...)`. The previous typed-client version decoded the list once. On the hot list path (`GetWorkflows` calls `CountWorkflows` and then `ListWorkflows`) this roughly doubles the decode work.

Additionally, the count now excludes malformed workflows (they're dropped by `TolerantList`), so it can disagree with `RemainingItemCount`/pagination. A count could read `len(ul.Items)` off the raw `UnstructuredList` with no per-item decode.

_Flagged as: efficiency regression + semantic change (silent undercount)._

server/workflow/workflow_server.go
```diff
@@ -330,8 +336,12 @@ func (s *workflowServer) WatchWorkflows(req *workflowpkg.WatchWorkflowsRequest,
...
 	}
 	s.instanceIDService.With(opts)
-	wfIf := wfClient.ArgoprojV1alpha1().Workflows(req.Namespace)
-	watch, err := wfIf.Watch(ctx, *opts)
+	// Use the tolerant dynamic watch so a single malformed workflow in the namespace
+	// cannot tear down this user-facing stream with a decode error — the outage class
+	// this package exists to prevent. Malformed Add/Modify events are dropped; the
+	// stream stays alive for every other workflow.
+	gvr := wfv1.SchemeGroupVersion.WithResource(workflow.WorkflowPlural)
+	watch, err := sutils.TolerantWatch[wfv1.Workflow, *wfv1.Workflow](ctx, auth.GetDynamicClient(ctx), gvr, req.Namespace, *opts)
```
Joibel, line 344: **`WatchWorkflows` now silently drops malformed events on the interactive stream.**

This replaces a typed watch (where a decode failure surfaced as a watch *error* to the client) with `TolerantWatch`, which drops malformed Add/Modify events. For the cached reflector path this is exactly the desired resilience, but `WatchWorkflows` also backs the interactive `argo watch` / `argo wait` path: watching a workflow that happens to be malformed now hangs with no events and no error, instead of failing fast.

Likely an acceptable tradeoff, but worth a conscious sign-off since it changes fail-fast semantics for that narrow interactive case.

_Flagged as: behavior change / UX regression for the interactive watch path._

server/event/event_server.go
```diff
@@ -73,12 +74,15 @@ func (s *Controller) ReceiveEvent(ctx context.Context, req *eventpkg.EventReques
 	options := metav1.ListOptions{}
 	s.instanceIDService.With(&options)
 
-	list, err := auth.GetWfClient(ctx).ArgoprojV1alpha1().WorkflowEventBindings(req.Namespace).List(ctx, options)
+	// Use the tolerant dynamic list so a single malformed WorkflowEventBinding
+	// cannot 500 every inbound event (the failure class this package guards).
+	gvr := wfv1.SchemeGroupVersion.WithResource(workflow.WorkflowEventBindingPlural)
+	items, _, err := sutils.TolerantList[wfv1.WorkflowEventBinding](ctx, auth.GetDynamicClient(ctx), gvr, req.Namespace, options)
```
Joibel, line 80: **Malformed `WorkflowEventBinding`s are now silently skipped on the event-trigger path.**

Unlike a read/list, `ReceiveEvent` is a *dispatch* path: a binding that fails per-item decode is dropped (warn log only), so an inbound event that should have triggered a workflow via that binding will silently fail to fire — versus the old behavior of erroring loudly. The comment frames this as intended (avoiding a 500 on every inbound event), which is reasonable, but converting a hard failure into a silently-missed trigger is a product decision worth an explicit confirmation.

_Flagged as: silent failure on a trigger path (intended, but worth sign-off)._

server/utils/tolerant_list.go
```diff
@@ -0,0 +1,106 @@
...
+// JSON marshal/unmarshal roundtrip — the same decode the typed client performs on
+// the wire bytes this path replaces.
+//
+// Going through json.Unmarshal is what invokes the custom json.Unmarshaler
+// implementations several workflow types rely on (ParallelSteps is an anonymous
+// list; Item/AnyString/Plugin/Amount/Object accept multiple shapes); a naive
+// reflective field copy would silently zero or drop them. (Current apimachinery's
+// runtime.DefaultUnstructuredConverter.FromUnstructured does honor custom
+// unmarshalers too, so it would also work — the JSON roundtrip is just the most
+// direct equivalent of the typed decode it stands in for.)
```
Joibel: **The justification for the JSON roundtrip contradicts the PR's own stated reason.**

This comment correctly notes that `runtime.DefaultUnstructuredConverter.FromUnstructured` "does honor custom unmarshalers too, so it would also work." But the PR description and the test comment in `tolerant_list_test.go` (`TestTolerantList_PreservesCustomUnmarshalers`) state that reflection-based conversion "silently dropped every Workflow with `steps`" — which can't be true for the pinned apimachinery if `FromUnstructured` honors `ParallelSteps`' custom unmarshaler.

A future maintainer reading the two will be misled about *why* the more expensive (3-pass, per your `ponytail:` note just below) path was chosen. Either reconcile the comments, or switch to the cheaper single-pass `FromUnstructured` (the existing test guards either implementation).

_Flagged as: inaccurate / self-contradictory comment (comment rot risk)._

Joibel (commented):
### Code review

Automated review of the error-tolerant list/watch change. Solid, well-tested PR with no correctness-breaking bugs — the items below are mostly intended tradeoffs worth an explicit sign-off, plus one self-contradictory comment. Left as inline comments on the relevant lines.

server/utils/tolerant_list.go
```diff
@@ -0,0 +1,124 @@
...
+// several workflow types rely on (ParallelSteps is an anonymous list;
+// Item/AnyString/Plugin/Amount/Object accept multiple shapes). It is the same
+// decode the typed client this path replaces performs on the wire bytes, so the
+// result is identical by construction. apimachinery's
+// runtime.DefaultUnstructuredConverter.FromUnstructured also honors these custom
+// unmarshalers (it decodes ParallelSteps correctly), so a single-pass
+// FromUnstructured would work too — the JSON roundtrip is chosen only as the most
+// direct equivalent of the typed decode it stands in for.
+//
+// ponytail: the roundtrip is 3 serialization passes per item (wire→unstructured by
```
Joibel, line 29: ponytail?

server/workflow/workflow_server.go
Joibel, line 217: Humm, possibly unresolvable and unnecessary to fix, but count!=number we'll actually list, so we'd see 50/50 with only 49 displayed.

Joibel (commented):
Any chance we could reduce the comment spam in this - claude has gone overly verbose here I think.

PR #16151, opened 2026-05-28
fix(server): strip baseHRef prefix from asset paths in static file server

irwinrex:
When baseHRef is set to a non-root value (e.g. /argo-workflows/), the browser requests assets relative to that path (e.g. /argo-workflows/main.js). The static file server was using the full request path directly to look up files in the embedded filesystem, causing a mismatch since the embedded assets are at the root level. This resulted in index.html being served instead of the actual JS/CSS files, causing syntax errors. Now the baseHRef prefix is stripped from the request path before checking asset existence and serving files, while still preserving the SPA fallback behavior for client-side routes.

Fixes #TODO

### Motivation
When Argo Workflows is deployed behind an AWS ALB (Application Load Balancer) on EKS, the ALB forwards requests at a subpath (e.g., `/argo-workflows/*`) to the Argo Server without stripping the path prefix. The `--base-href` flag is used to inform the browser that all assets are relative to `/argo-workflows/`. However, the static file server did not account for this prefix when resolving asset requests internally, causing all JS/CSS files to be served as `index.html` instead. This broke the UI entirely with `Uncaught SyntaxError: expected expression, got '<'`.
This is a common deployment pattern — ALB ingress rules typically map a path prefix to a service without path rewriting (unlike nginx ingress which supports `rewrite-target`). The fix ensures the server correctly handles requests with the baseHRef prefix regardless of whether the reverse proxy strips it.

### Modifications
- `server/static/static.go`: Modified `ServerFiles()` to strip the `baseHRef` prefix from `r.URL.Path` before checking whether a UI asset exists and before serving it via `http.FileServer`. The normalization handles both `/argo-workflows/` and `/argo-workflows` formats.

### Verification
- Existing unit tests in `server/static/static_test.go` pass
- Verified with integration test that:
  - `/main.js` with root baseHRef → returns JS (regression)
  - `/argo-workflows/main.js` with `/argo-workflows/` baseHRef → returns JS (fix)
  - `/argo-workflows/` → returns index.html with correct `<base href>`
  - `/argo-workflows/workflows/foo` (SPA route) → returns index.html (fallback preserved)
  - Works with and without trailing slash in baseHRef value
  
### Documentation
No documentation changes needed — this is a bug fix that makes the existing `--base-href` / `ARGO_BASE_HREF` feature work correctly.

### AI
None

For the ALB part specifically: AWS Load Balancer Controller on EKS uses pathType: Prefix with ALB. Unlike nginx ingress (which supports nginx.ingress.kubernetes.io/rewrite-target: /$2 to strip path prefixes), ALB forwards the full request path to the target service.I know we can use ALB Urlwrite to use prefix path still the issue exits.This means the Argo Server must handle requests like /argo-workflows/main.js directly, which is exactly what this fix enables.

Signed-off-by: Irwin Rex <irwinrex.a@gmail.com>


irwinrex:
PR is outdated

PR #16156, opened 2026-05-29
fix(ci): use client-id for GitHub Action "Create GitHub App Token"

nakatani-yo:
### Motivation

[The `Cherry Pick` workflow](https://github.com/argoproj/argo-workflows/actions/workflows/cherry-pick.yml) has started emitting warnings because it still uses the legacy `app-id` input for `actions/create-github-app-token`.

image

`actions/create-github-app-token` recommends using the `client-id` input instead.

Reference: https://github.com/actions/create-github-app-token#inputs

### Modifications

Updated the `Cherry Pick Single` workflow to pass `CHERRYPICK_APP_ID` via `client-id` instead of `app-id` when creating a GitHub App token.

### Verification

Not run locally; this is a workflow configuration-only change. It can be verified by running the `Cherry Pick Single` workflow and confirming that the GitHub App token is created successfully.

### Documentation

No documentation update is needed. This change only updates an internal GitHub Actions workflow.

### AI

Codex

### DIFF
--- a/.github/workflows/cherry-pick-single.yml
+++ b/.github/workflows/cherry-pick-single.yml
@@ -37,7 +37,7 @@ jobs:
         id: generate-token
         uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
         with:
-          app-id: ${{ secrets.CHERRYPICK_APP_ID }}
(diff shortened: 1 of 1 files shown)


MasonM (approved):
Thanks!

PR #13821, opened 2024-10-26
fix(deps): revert React 18 upgrade back to React 16 

agilgur5:
**Draft until #13818 is merged, as this builds on top of it**

Fixes #13410 per https://github.com/argoproj/argo-workflows/pull/13069#pullrequestreview-2079815858, effectively reverting the React upgrade part of #13069

### Motivation

- not all of our deps are compatible with React 18 and so this caused many warnings as well as some bugs
  - those need more updates, primarily breaking changes within `argo-ui` or decoupling the dependency on `argo-ui`

- the main purpose of the upgrade was to be able to upgrade `swagger-ui-react`, which had an [alternate React 16 compatible path](https://github.com/swagger-api/swagger-ui/wiki/Using-older-version-of-React) and has now been removed from the UI codebase entirely as of #13818 
  - so the upgrade is no longer necessary for that purpose

### Modifications

- downgrade back to React 16 and revert `render` changes to use React 16 syntax

PR #16159, opened 2026-05-31
chore(ui): modernize the UI dependency stack

tico24:
### Motivation

The web UI was several major versions behind on most of its dependencies. React, react-router, and the build and test tooling were old enough to be a security concern and a drag on future UI work. This brings the stack up to current versions while keeping the UI behavior the same.

### Modifications

- React 18 to 19, @types/react to 19
- react-router 4 to 7 (full routing rewrite), history 4 to 5
- argo-ui pinned to its latest commit and patched so its nav bar works under react-router 7
- TypeScript 4.9 to 5.9
- Jest 26 to 29 (ts-jest 29, jest-environment-jsdom)
- typescript-eslint 6 to 8 (ESLint stays on 8)
- Chart.js 2 to 4, react-chartjs-2 2 to 5
- react-markdown 8 to 10, remark-gfm 3 to 4
- react-datepicker 4 to 9 (adds date-fns)
- monaco-editor 0.45 to 0.55, react-monaco-editor 0.50 to 0.59
- cron-parser 4 to 5, cronstrue 2 to 3
- Build tooling: sass-loader 17, style-loader 4 with css-loader, copy-webpack-plugin 14, webpack-cli 7
- dagre replaced with @dagrejs/dagre
- Resolved the outstanding production-dependency CVEs (production audit is clean), removed stale @types packages, and added tests for the migrated routing, charts, datepicker, markdown, and nav-bar code

sass-loader 17 requires Node >= 22.11, so the Node baseline moves from 20 to 22. This is applied consistently across `ui/package.json` engines, the `argo-ui` Dockerfile stage (`node:22-alpine`), the CI/release/snyk GitHub workflows, the devcontainer, and the nix dev shell, so local, CI, and image builds stay in sync.

### Verification

- tsc, jest (108 tests), eslint, deduplicate, and the production webpack build all pass on Node 22, with `git diff --exit-code` clean
- The production image builds with the UI embedded and serves it: built the `argocli` image, ran the server from it, and confirmed it returns the embedded `index.html` (HTTP 200) and the compiled `main.<hash>.js` bundle (HTTP 200) over the static-files path
- Ran the full UI against a backend built from this branch (controller and argo-server from source on a local cluster) and clicked through the main flows: workflow list, DAG view, live log streaming, submitting a workflow from the editor, the reports charts, templates, and cron workflows. All worked against real data.

### Documentation

No documentation changes are needed. The build and run instructions are unchanged. The supported Node version moves to 22, which the dependency bumps require.

### AI

AI tooling was used to assist with the dependency upgrades, code migration, and testing.


jmeridth (approved):

ui/package.json
```diff
@@ -10,90 +10,96 @@
...
-        "react-datepicker": "^4.25.0",
-        "react-dom": "^18.3.1",
-        "react-markdown": "^8.0.7",
-        "react-monaco-editor": "^0.50.1",
-        "react-router": "^4.3.1",
-        "react-router-dom": "^4.2.2",
-        "remark-gfm": "^3.0.0",
-        "rxjs": "^7.8.1",
-        "superagent": "^10.2.3",
+        "react": "^19",
```
MasonM, line 34: React 19 is a large, high-risk upgrade. The last React upgrade caused tons of problems (see https://github.com/argoproj/argo-workflows/pull/13821), some of which still haven't been fixed.

I think we need to go through the fix the existing React 18 warnings before upgrading, because a lot of them are basically saying "this will break in React 19", e.g.  
> Warning: Route2 uses the legacy childContextTypes API which is no longer supported and will be removed in the next major release. Use React.createContext() instead

ui/patches/argo-ui+1.0.0.patch
MasonM, line 1: Patching https://github.com/argoproj/argo-ui is going to increase the maintenance burden and make it harder to update in the future. Why can't we update it at the source?

MasonM (changes requested):
I'm seeing quite a few regressions, e.g. some of the drop-downs on the log viewer page are misaligned:
Image

Modernizing the UI stack is great, but I think we need to take this slow and create focused PRs (especially for the React upgrade, since the last one caused a lot of breakage). We have very poor test coverage for the UI, so the risk of breakage is high. Improving the test coverage would really help.

PR #16161, opened 2026-06-02
feat: add opt-in init-less pod layout (beta)

### LINKED ISSUE #16154: Remove our init container (beta)
# Summary

The only thing that the initContainer called `init` that we inject into `container` templates and friends that couldn't be achieved by `wait` is providing the argoexec binary for `main` to use.

With Image Volumes in k8s now we can use those to provide argoexec

* k8s 1.36 has image volumes as GA
* 1.35 has them on by default
* 1.32 has them in beta but not on by default. 1.33 is our oldest supported version.

So not everyone can use them. Provide them via a configmap setting defaulting to zero.

This proposal changes nothing about user provided init containers.

This avoids the k8s API cost and time to run the init container completely.

`supervisor` would replace `init` and `wait` as a container doing artifact download, script prep etc, and then argoexec would read from the shared filesystem when it was allowed to kick off the main workload in `main`.

## Use Cases

When would you use this?

Always, no known downsides and many speedup benefits

## Future

I'd plan to remove init in the future (release n+2) once this is made default in whatever feature release n+1 is.

We could provide a hook to allow users to run code in `main` whilst artifacts are being made available to preheat the container and potentially replace some user init container uses.

---

Joibel:
Fixes #16154

:warning: This PR switches all e2e CI from using docker backend of k3d to containerd.

### Motivation

Every workflow pod currently runs an `argoexec init` container before any regular container starts, adding sequential startup latency to every pod. The init container's responsibilities also overlap conceptually with `wait` — both are Argo-infrastructure containers running `argoexec` — and a plugin used for both inputs and outputs ends up as two containers per pod (an init container for Load and a sidecar for Save).

This introduces an opt-in, controller-wide `initlessPod` mode (Beta, off by default) that removes the init container entirely, cutting pod startup latency and collapsing the input/output plugin split.

### Modifications

New controller-wide `initlessPod.enabled` setting (workflow controller ConfigMap). When enabled, every subsequently scheduled workflow pod uses the init-less layout; the legacy `wait` + init-container behaviour is unchanged when disabled (the default).

- The `argoexec` binary is delivered to `main` via a Kubernetes image volume (KEP-4639 — Beta in K8s 1.33 behind a feature gate, GA in 1.36) instead of being copied by an init container.
- A new `supervisor` container subsumes init + wait: template write, script staging, input artifact download, readiness signalling, then the post-main responsibilities (observe `main`, collect outputs/logs/artifacts). Pods run with **zero** init containers.
- Artifact plugins run as a single sidecar per plugin, invoked by `supervisor` for both Load and Save.
- Input artifacts are delivered via a whole-volume mount that the emissary symlinks into place once `supervisor` signals ready (per-artifact SubPath mounts race kubelet in the init-less layout, where `main` and `supervisor` start concurrently).
- Shared post-main logic is extracted into `WorkflowExecutor.PostMain`, used by both `wait` and `supervisor`.

Enable with `initlessPod.enabled: true`; roll back by setting it to `false` (in-flight pods keep their original layout).

### Verification

- New unit tests covering: pod-spec shape (zero init containers, `supervisor` present, image volume mounted, emissary entrypoint retargeted), plugin sidecar dedup across input/output references, template-env handling for templates without a supervisor, ready/failed marker waiting and writing, input-artifact symlink behaviour, supervisor pre-main orchestration, and output staging when an input artifact path is reused as an output path.
- New `initless: true` CI matrix dimension (restricted to `k8s_version: max`, bumped 1.35 → 1.36 where image volumes are GA) re-runs `test-corefunctional`, `test-functional`, `test-artifacts`, and `test-plugins` under the new layout.
- Legacy-mode behaviour is held invariant by existing tests plus explicit init-less-off assertions.

### Documentation

- New `docs/initless-pod.md` (architecture, pod lifecycle, failure modes, the input-artifact symlink-vs-bind-mount semantics table, and plugin author notes).
- Updates to `docs/architecture.md`, `docs/tracing.md`, `docs/workflow-controller-configmap.md`, and `mkdocs.yml`.
- Users discover the feature through the controller ConfigMap reference and the dedicated docs page.

### AI

Generative AI (Claude) was used to assist in preparing this PR — including code, tests, documentation, and commit messages — with human review throughout. See the Argo project's [Generative AI policy](https://github.com/argoproj/argoproj/blob/main/community/genai.md).

## Summary by CodeRabbit

* **New Features**
  * Introduced an opt-in beta “init-less pod” layout for workflow pods.
  * Added a supervisor container to manage pre-main staging, readiness signaling, and post-main responsibilities.
  * Updated artifact plugins to run as sidecars driven by the supervisor.

* **Configuration**
  * Added `InitlessPod.Enabled` to the workflow controller ConfigMap to enable/disable the mode.

* **Documentation**
  * Added operator and architecture documentation for init-less pods, plus updated tracing notes.

* **Bug Fixes**
  * Improved signal forwarding behavior and centralized subprocess exit handling.
  * Hardened artifact/input linking to be mount-aware; added additional edge-case validation.

* **Tests / CI**
  * Extended E2E coverage and adjusted CI execution and image import steps for init-less mode.


Joibel:
@coderabbitai review

.features/pending/initless-pod.md
```diff
@@ -0,0 +1,11 @@
+Description: Opt-in pod layout that removes the `argoexec init` container
+Authors: [Alan Clucas](https://github.com/Joibel)
```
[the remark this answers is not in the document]
Joibel: No, dunno why you think github handles match peoples names

cmd/argoexec/commands/emissary.go
```diff
@@ -281,6 +291,164 @@ func NewEmissaryCommand() *cobra.Command {
...
+		if _, err := os.Lstat(dst); err == nil {
+			if rmErr := os.RemoveAll(dst); rmErr != nil {
+				return fmt.Errorf("failed to clear existing path for artifact %q at %s: %w", art.Name, dst, rmErr)
+			}
+		} else if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to stat artifact path %q at %s: %w", art.Name, dst, err)
+		}
+		if err := os.Symlink(src, dst); err != nil {
+			return fmt.Errorf("failed to symlink input artifact %q (%s -> %s): %w", art.Name, dst, src, err)
+		}
```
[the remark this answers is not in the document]
Joibel: Good catch — switched to `os.Remove` and now return an error rather than recursively deleting a pre-existing directory at the artifact path (matches legacy SubPath shadowing). Test updated to assert the directory is preserved.

workflow/common/common.go
```diff
@@ -314,9 +353,54 @@ func IsArtifactPluginSidecar(containerName string) bool {
 }
 
 func IsArgoSidecar(containerName string) bool {
-	return containerName == WaitContainerName || IsArtifactPluginSidecar(containerName)
+	return containerName == WaitContainerName || containerName == SupervisorContainerName || IsArtifactPluginSidecar(containerName)
 }
```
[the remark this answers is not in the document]
Joibel: Added reserved-name validation in `workflow/validate` (it can't live in `v1alpha1` — import cycle with `workflow/common`). User containerSet containers and sidecars are now rejected when the name matches `common.IsArgoSidecar` (wait/supervisor/plugin-sidecar). Test added.

cmd/argoexec/commands/emissary.go
```diff
@@ -281,6 +291,164 @@ func NewEmissaryCommand() *cobra.Command {
...
+		logger.Info(ctx, "supervisor is ready")
+		return nil
+	case errors.Is(err, errFailedSeen):
+		return fmt.Errorf("supervisor reported pre-main failure: %s", string(failedBody))
+	case err != nil:
+		return fmt.Errorf("waiting for supervisor readiness: %w", err)
+	default:
+		// errgroup returned nil with neither sentinel — should not happen,
+		// but treat as success if it does (both watchers exited cleanly).
+		return nil
```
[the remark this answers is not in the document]
Joibel: Fixed — parent-context cancellation now propagates: the watchers only treat a cancel as a clean exit when `ctx.Err() == nil`, and the default case returns `ctx.Err()` instead of nil. Added a regression test that cancels the parent and asserts an error.

workflow/controller/workflowpod.go
```diff
@@ -1500,19 +1745,141 @@ func (woc *wfOperationCtx) addArtifactPlugins(ctx context.Context, pod *apiv1.Po
...
+	}
+	// Plugin sidecars run the plugin's own image, so they need argoexec
+	// delivered via the argoexec-bin image volume (their command points at
+	// argoBinExecutorPath — see artifactSidecarContainer).
+	argoBinMount := argoBinVolumeMount()
+	for i := range sidecars {
+		sidecars[i].VolumeMounts = append(sidecars[i].VolumeMounts, argoBinMount)
+		sidecars[i].VolumeMounts = append(sidecars[i].VolumeMounts, pluginSidecarMounts...)
+		pod.Spec.Containers = append(pod.Spec.Containers, sidecars[i])
+		sidecarNames = append(sidecarNames, sidecars[i].Name)
```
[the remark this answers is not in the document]
Joibel: Fixed — input-only plugin socket volumes are now appended to `pod.Spec.Volumes` in `addArtifactPluginsInitless`, deduped against the output-plugin volumes already added by `createArtifactVolumeMounts`.

workflow/controller/operator_test.go
```diff
@@ -7801,6 +7801,94 @@ func TestPodFailureWithContainerOOM(t *testing.T) {
...
+	var pod apiv1.Pod
+	wfv1.MustUnmarshal(podWithSupervisorPreMainFailure, &pod)
+	ctx := logging.TestContext(t.Context())
+	nodeStatus, msg := newWoc(ctx).inferFailedReason(ctx, &pod, nil)
+	// Supervisor's NodeError must win over main's exit-65 NodeFailed; the
+	// emitted message must name the supervisor and quote its real reason.
+	assert.Equal(t, wfv1.NodeError, nodeStatus)
+	assert.Contains(t, msg, "supervisor")
+	assert.Contains(t, msg, "connection refused")
+}
```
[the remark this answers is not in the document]
Joibel: Tightened — added `assert.NotContains` for the losing container's signal in both tests (main's exit-65 placeholder, and the supervisor in the success-supervisor/fail-main case).

workflow/executor/executor.go
```diff
@@ -1426,14 +1537,24 @@ func (we *WorkflowExecutor) Init() error {
 	return nil
 }
 
+// WriteTemplate writes the template JSON to the shared volume without
+// copying the argoexec binary. Used by the init-less supervisor, where
+// the binary is delivered via a Kubernetes image volume.
+func (we *WorkflowExecutor) WriteTemplate() error {
+	if w, ok := we.RuntimeExecutor.(TemplateWriter); ok {
```
isubasinghe, line 1482: We don't really need this indirection, its adding more cognitive overhead.

workflow/executor/executor.go
```diff
@@ -1426,14 +1537,24 @@ func (we *WorkflowExecutor) Init() error {
 	return nil
 }
```
isubasinghe, line 1477: No need for this cast as well, although this was not your code it seems.

cmd/argoexec/commands/supervisor.go
```diff
@@ -0,0 +1,186 @@
...
+	osspecific.AllowGrantingAccessToEveryone()
+
+	// Make restarts idempotent: a previous crashed attempt may have left a
+	// failed marker in the shared emptyDir, which would otherwise race the
+	// ready marker on the emissary's watcher and fail the pod even after
+	// pre-main succeeds. Ignore not-exist; surface other errors so a broken
+	// mount fails fast rather than silently producing the wrong outcome.
+	if err := os.Remove(common.FailedMarkerPath); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to clean stale failed marker: %w", err)
+	}
```
isubasinghe: I don't know if I like this, this can still exhibit the behaviour we are trying to avoid, but now its just racey instead. Some kind of epoch based marker is likely ideal here.

cmd/argoexec/commands/emissary.go
```diff
@@ -125,7 +125,21 @@ func NewEmissaryCommand() *cobra.Command {
...
-			data, err := os.ReadFile(varRunArgo + "/template")
+			// In init-less pod mode the supervisor, not an init container, writes
+			// /var/run/argo/template. Supervisor and main start concurrently, so
+			// block until supervisor signals readiness (or failure) before reading
+			// the template. Gated on an env var so legacy pods are unaffected.
+			waitForReady := os.Getenv(common.EnvVarWaitForReady) == "true"
+			if waitForReady {
+				if waitErr := waitForSupervisorReady(ctx); waitErr != nil {
+					// Distinct exit code so the controller attributes the failure
+					// to supervisor pre-main setup rather than the user command.
```
isubasinghe, line 136: What happens if the supervisor dies/errors out here and somehow fails to write the failure marker?
It to me feels like this protocol could be improved a bit to handle the restart case as well here. 
It feels like the state machine is:
start->working->ready
start->working->failure. 

I wonder if we should have a "started generation x" marker, then we can wait on generation x.
This makes restarts also non-racey.

workflow/signal/signal.go
```diff
@@ -17,22 +16,53 @@ import (
...
+		if c.Name != container {
+			continue
+		}
+		var hasVarRunArgo, hasArgoExecBin bool
+		for _, m := range c.VolumeMounts {
+			switch m.MountPath {
+			case common.VarRunArgoPath:
+				hasVarRunArgo = true
+			case common.ArgoExecBinMountPath:
+				hasArgoExecBin = true
```
isubasinghe: The supervisor container should just use the bare `argoexec` which should exists in its PATH.

isubasinghe (commented):
I found some issues I think, claude found more. I haven't verified claude's output but I think 1,7 and 8 are real. I think I came to the same conclusion regarding 1 actually:

 1. Supervisor crash hangs the workflow forever — emissary.go:134. If the supervisor dies without writing the ready or failed marker (OOM during artifact download, or a panic), main's emissary blocks in waitForSupervisorReady
  with no timeout. restartPolicy: Never means the container isn't restarted, there's no defer/recover to write a marker on crash, and the controller only calls MarkTaskResultComplete for a non-zero supervisor while main is
  Running — it never fails the node. So the pod sits Running until activeDeadlineSeconds (which has no default). Legacy init-container OOM failed the pod promptly. This is the same family as the bug you just reproduced, but worse
  — even pod deletion is the only escape, and nothing triggers it automatically.

  3. manifestFrom resource templates never get their manifest — workflowpod.go:289. The supervisor is only added (hasAuxCtr) for resource templates if archiveLogs is on. A resource.manifestFrom.artifact template without
  archiveLogs gets no supervisor and no init container, so nothing downloads the manifest. StageFiles assumes it was already staged and ExecResource then fails reading a file that was never delivered. No validation guard.

  4. readOnlyRootFilesystem breaks input artifacts — emissary.go:379. Legacy delivered input artifacts via kubelet SubPath bind mounts (work regardless of fs permissions). Init-less instead does MkdirAll/RemoveAll/Symlink inside
  main's own filesystem, which returns EROFS under readOnlyRootFilesystem: true (a common PSS/policy hardening). Same workflow succeeds legacy, fails init-less. Undocumented.

  5. os.RemoveAll can wipe a mounted volume — emissary.go:373. linkInputArtifactsAt runs os.RemoveAll(art.Path) on paths legacy never deleted, and FindOverlappingVolume only detects art.Path inside a mountPath, not a mountPath
  inside art.Path. An artifact path that's an ancestor of a mounted PVC (PVC at /data/shared, artifact path /data) recursively destroys the live volume. The controller-side CleanPath validation is also skipped on the init-less
  branch (workflowpod.go:1407).

  6. Pod-polling fallback is RBAC-dead and adds API load — executor.go:1372. waitMainContainersTerminated polls Pods().Get every 3s for every main container, in legacy pods too (which previously made zero API calls here). But the
  documented minimal executor Role lacks pods get, so it 403s forever (logged at Debug) and the OOM-detection fallback it implements never works on standard installs. It also duplicates executor/common.WaitForTermination and
  ignores the existing WAIT_CONTAINER_STATUS_CHECK_INTERVAL knob.

  7. Plugin-sidecar safety net is dead code — artifact_plugin_sidecar.go:92. startAuxExitWatcher waits for /var/run/argo/ctr/{wait,supervisor}/exitcode, but those containers run argoexec directly and never write exitcode files
  (only emissary-wrapped containers do). Its IsInitlessPod() check is also always false in the sidecar — because, as you saw in the YAML, ARGO_INITLESS_POD is set on the supervisor only, not delivered to sidecars. So on a
  supervisor crash the plugin sidecar is orphaned until pod terminationGracePeriod — exactly what the watcher claims to prevent.

  9. UI logs viewer can't show supervisor logs — workflow-logs-viewer.tsx:208. The container dropdown hardcodes ['init', 'wait'] and the diff touches no UI files. In init-less pods the supervisor — where pre-main download and
  post-main upload errors land — is unreachable from the UI. CLI is fine (argo logs -c supervisor).

  10. Config hot-reload can tear pod construction — workflowpod.go:249. isInitlessPodEnabled() is read at ~9 points during one createWorkflowPod, while wfc.Config is overwritten un-synchronized by the ConfigMap watcher. A flip
  mid-construction can emit a self-inconsistent pod (e.g. ARGO_WAIT_FOR_READY=true with no supervisor → main blocks forever). Low probability, but the first config field whose torn reads deadlock a pod. Fix: read once at the top,
  thread through.

workflow/executor/executor.go
Joibel:
```
	return we.RuntimeExecutor.TemplateWriter.WriteTemplate(we.Template)
```
gives
```
we.RuntimeExecutor.TemplateWriter undefined (type ContainerRuntimeExecutor has no field or method TemplateWriter)
```

The only executor left (emissary) implements both, but CRE doesn't, so we have to assert. We could drop the interface as emissary is the only one, but not proposing to do it here.

Not intending to change anything

workflow/executor/executor.go
Joibel: Same

workflow/signal/signal.go
Joibel: Ok, I've tried to implement that

cmd/argoexec/commands/emissary.go
Joibel: I've redone this as a single file rather than separate success/failure files, and that file "always" exists and is rewritten to heartbeat.

Joibel:
> I found some issues I think, claude found more. I haven't verified claude's output but I think 1,7 and 8 are real. I think I came to the same conclusion regarding 1 actually:
> 
>     1. Supervisor crash hangs the workflow forever — emissary.go:134. If the supervisor dies without writing the ready or failed marker (OOM during artifact download, or a panic), main's emissary blocks in waitForSupervisorReady
>        with no timeout. restartPolicy: Never means the container isn't restarted, there's no defer/recover to write a marker on crash, and the controller only calls MarkTaskResultComplete for a non-zero supervisor while main is
>        Running — it never fails the node. So the pod sits Running until activeDeadlineSeconds (which has no default). Legacy init-container OOM failed the pod promptly. This is the same family as the bug you just reproduced, but worse
>        — even pod deletion is the only escape, and nothing triggers it automatically.

Added a heartbeat in here, so main will die if supervisor doesn't heartbeat during startup phase.

>     2. manifestFrom resource templates never get their manifest — workflowpod.go:289. The supervisor is only added (hasAuxCtr) for resource templates if archiveLogs is on. A resource.manifestFrom.artifact template without
>        archiveLogs gets no supervisor and no init container, so nothing downloads the manifest. StageFiles assumes it was already staged and ExecResource then fails reading a file that was never delivered. No validation guard.

Nice find, a gap in the e2e capabilities - the test there has archiveLogs on always. Not added e2e due to cost though.

>     3. readOnlyRootFilesystem breaks input artifacts — emissary.go:379. Legacy delivered input artifacts via kubelet SubPath bind mounts (work regardless of fs permissions). Init-less instead does MkdirAll/RemoveAll/Symlink inside
>        main's own filesystem, which returns EROFS under readOnlyRootFilesystem: true (a common PSS/policy hardening). Same workflow succeeds legacy, fails init-less. Undocumented.

Needs documenting. Added - it's not really possible to fix any other way.

>     4. os.RemoveAll can wipe a mounted volume — emissary.go:373. linkInputArtifactsAt runs os.RemoveAll(art.Path) on paths legacy never deleted, and FindOverlappingVolume only detects art.Path inside a mountPath, not a mountPath
>        inside art.Path. An artifact path that's an ancestor of a mounted PVC (PVC at /data/shared, artifact path /data) recursively destroys the live volume. The controller-side CleanPath validation is also skipped on the init-less
>        branch (workflowpod.go:1407).

Another good find - I've double guarded this, both in the controller and again in the executor to avoid data wipes.

>     5. Pod-polling fallback is RBAC-dead and adds API load — executor.go:1372. waitMainContainersTerminated polls Pods().Get every 3s for every main container, in legacy pods too (which previously made zero API calls here). But the
>        documented minimal executor Role lacks pods get, so it 403s forever (logged at Debug) and the OOM-detection fallback it implements never works on standard installs. It also duplicates executor/common.WaitForTermination and
>        ignores the existing WAIT_CONTAINER_STATUS_CHECK_INTERVAL knob.

Yeah, good find. Removed.

>     6. Plugin-sidecar safety net is dead code — artifact_plugin_sidecar.go:92. startAuxExitWatcher waits for /var/run/argo/ctr/{wait,supervisor}/exitcode, but those containers run argoexec directly and never write exitcode files
>        (only emissary-wrapped containers do). Its IsInitlessPod() check is also always false in the sidecar — because, as you saw in the YAML, ARGO_INITLESS_POD is set on the supervisor only, not delivered to sidecars. So on a
>        supervisor crash the plugin sidecar is orphaned until pod terminationGracePeriod — exactly what the watcher claims to prevent.

Removed, it was belt and braces but fixing that is out of scope.

>     7. UI logs viewer can't show supervisor logs — workflow-logs-viewer.tsx:208. The container dropdown hardcodes ['init', 'wait'] and the diff touches no UI files. In init-less pods the supervisor — where pre-main download and
>        post-main upload errors land — is unreachable from the UI. CLI is fine (argo logs -c supervisor).

Added.

>     8. Config hot-reload can tear pod construction — workflowpod.go:249. isInitlessPodEnabled() is read at ~9 points during one createWorkflowPod, while wfc.Config is overwritten un-synchronized by the ConfigMap watcher. A flip
>        mid-construction can emit a self-inconsistent pod (e.g. ARGO_WAIT_FOR_READY=true with no supervisor → main blocks forever). Low probability, but the first config field whose torn reads deadlock a pod. Fix: read once at the top,
>        thread through.

I have a follow on PR which aims to fix this along with a lot of other sprawl - happy to wait for that? This PR doesn't refactor workflowpod.go to make the changes more obvious, and then is followed by one that rewrites it. Basically it creates a PodSpecBuilder which is pure.

Joibel:
@coderabbitai review

workflow/executor/executor.go
isubasinghe: Yeah would be nice to remove the interface, but yeah happy to leave it for a later cleanup.

cmd/argoexec/commands/emissary.go
```diff
@@ -125,7 +125,21 @@ func NewEmissaryCommand() *cobra.Command {
...
+					exitCode = common.ExitCodeSupervisorPreMainFailure
+					return waitErr
```
isubasinghe: `waitError`  might not implement errors.Extited which will result in this exitCode being ignored and exit with code 64 instead. Check argoexec/main.go:27

cmd/argoexec/commands/emissary.go
```diff
@@ -281,6 +291,265 @@ func NewEmissaryCommand() *cobra.Command {
...
+		// recurse into and destroy a live PVC/hostPath/emptyDir.
+		if _, err := os.Lstat(dst); err == nil {
+			mp, mErr := mountPointAtOrUnder(mountinfoPath, dst)
+			if mErr != nil {
+				return fmt.Errorf("failed to check for mounts under artifact path %q at %s: %w", art.Name, dst, mErr)
+			}
+			if mp != "" {
+				return fmt.Errorf("refusing to stage input artifact %q at %s: the path is or contains mount point %s, and clearing it would destroy the mounted volume; change the artifact path or volume mount so they do not overlap", art.Name, dst, mp)
+			}
+			if rmErr := os.RemoveAll(dst); rmErr != nil {
```
isubasinghe, line 405:
```yaml
  kind: Workflow
  metadata:
    generateName: initless-removeall-volume-wipe-
  spec:
    entrypoint: demo
    volumeClaimTemplates:
      - metadata: { name: vol }
        spec:
          accessModes: ["ReadWriteOnce"]
          resources: { requests: { storage: 64Mi } }
    templates:
      - name: demo
        steps:
          - - { name: seed,   template: seed }
          - - { name: victim, template: victim }
          - - { name: verify, template: verify }

      # 1) Put irreplaceable data INSIDE the volume, under a subdirectory.
      - name: seed
        container:
          image: alpine:3
          command: [sh, -c]
          args:
            - |
              mkdir -p /mnt/vol/sub
              echo "CRITICAL PRODUCTION DATA" > /mnt/vol/sub/important.txt
              echo "seeded:"; ls -la /mnt/vol/sub
          volumeMounts: [{ name: vol, mountPath: /mnt/vol }]

      # 2) The victim. Image has /data -> /mnt/vol. Input artifact path /data/sub is
      #    UNDER the symlink. Controller overlap checks compare the literal string
      #    "/data/sub" against the declared mount "/mnt/vol" -> no overlap -> admission
      #    passes, artifact staged to the shared emptyDir. main's emissary then, for
      #    /data/sub: MkdirAll("/data") [no-op symlink], Lstat("/data/sub") [follows
      #    /data->/mnt/vol, sees the real dir], mountPointAtOrUnder("/data/sub")
      #    [literal compare, /mnt/vol matches nothing -> ""], RemoveAll("/data/sub")
      #    [follows the symlink, recursively deletes /mnt/vol/sub], then symlinks.
      - name: victim
        inputs:
          artifacts:
            - name: payload
              path: /data/sub
              http:
                url: https://raw.githubusercontent.com/argoproj/argo-workflows/main/README.md
        container:
          image: YOUR_REGISTRY/symlink-demo:latest
          command: [sh, -c]
          args:
            - |
              echo "victim main running (destruction already happened in pre-main staging)"
              echo "/data/sub now points to:"; readlink /data/sub || true
              ls -la /mnt/vol
          volumeMounts: [{ name: vol, mountPath: /mnt/vol }]

      # 3) Independent witness: fresh pod, normal image, same PVC. Proves the data is
      #    gone from the volume itself, not just hidden behind the victim's symlink view.
      - name: verify
        container:
          image: alpine:3
          command: [sh, -c]
          args:
            - |
              echo "volume now:"; ls -la /mnt/vol
              echo "sub is:"; readlink /mnt/vol/sub || echo "(not a symlink)"
              if [ -f /mnt/vol/sub/important.txt ]; then
                echo "RESULT: important.txt SURVIVED (bug not triggered / already fixed)"
              else
                echo "RESULT: important.txt DESTROYED — RemoveAll recursed through the symlink into the PVC"
              fi
          volumeMounts: [{ name: vol, mountPath: /mnt/vol }]
```

This is kinda broken

Joibel: Ok, I'll work out what we can do that doesn't involve what I'm doing here - good find and totally unfixable in current form.

Joibel: Basically I've retained the removeall where I can prove it's safe, and fail at runtime if it's crosslinked.

cmd/argoexec/commands/emissary.go
Joibel: Not yet addressed

Joibel: Ok, wrapped it now. Another fine fault find!

isubasinghe (approved):
I think this is good to go now

PR #15932, opened 2026-04-15
fix: populate scope with empty values for outputs of skipped/omitted …

KowalczykBartek:
Fixes https://github.com/argoproj/argo-workflows/issues/15873
Motivation

Try to resolve child's output when steps is actually skipped.

PR #16223, opened 2026-06-04
fix!: drop values when skipped arguments are being substituted

isubasinghe:
### Motivation

  Skipped/omitted node outputs resolved to `""`, so: producer `valueFrom.default` was ignored, consumer input defaults were clobbered, and `??` never fired (value was
  present-empty, not absent).

  ### Modifications

  - Skipped/omitted outputs are `nil` in scope (absent optional); simple `{{...}}` refs still flatten to `""` (preserves #15932 no-requeue behavior).
  - Producer `valueFrom.default` applies; else consumer input default applies (`dropSkippedDefaultedArgs`); else `""`.
  - New `ReplaceStrictAny` preserves nil into expression envs, so `??` works in inline `{{=...}}` args and `when` clauses, not just
  `valueFrom.expression`/`fromExpression`.
  - Nil expression result with all identifiers present in env collapses to `""` instead of staying unresolved.
  - Refactor: unified identifier-vs-env checks on AST-based `missingVarsInEnv`; removed lexer-token `hasVariableInExpression`/`searchTokens`/`filterEOF`.
  - Behavior change: expressions doing e.g. `skippedRef + '-suffix'` now see nil (error) instead of `""`; use `??`.

  ### Verification

  New unit tests (`TestReplaceStrictAnyNilValues`, `Test_missingVarsInEnv`) and controller tests (`TestDAGSkippedOutput*`, `TestDAGSkippedInlineExpressionFallback`,
  `TestStepsSkippedOutputDefault`, `examples_skipped_defaults_test.go`). Full `workflow/controller`, `workflow/validate`, `workflow/common`, `util/template` suites pass.

  ### Documentation

  New examples in `examples/skipped-output-defaults/`.

  ### AI

  Claude Code was used for code, tests, and this description, under human review.

### DIFF
--- a/util/template/expression_template.go
+++ b/util/template/expression_template.go
@@ -8,14 +8,13 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
 	"github.com/expr-lang/expr"
 	"github.com/expr-lang/expr/ast"
-	"github.com/expr-lang/expr/file"
 	"github.com/expr-lang/expr/parser"
-	"github.com/expr-lang/expr/parser/lexer"
 
 	"github.com/argoproj/argo-workflows/v4/util/logging"
 	"github.com/argoproj/argo-workflows/v4/util/maps"
@@ -40,10 +39,33 @@ var (
 	}
 )
 
+// missingVarsInEnv returns the identifiers referenced by the expression that are absent from env
+// (a present-but-nil leaf counts as present). Errors if the expression cannot be parsed.
+func missingVarsInEnv(expression string, env map[string]any) ([]string, error) {
+	identifiers, err := getIdentifiers(expression)
+	if err != nil {
+		return nil, err
+	}
+	var missing []string
+	for _, id := range identifiers {
+		if !hasVarInEnv(env, id) {
+			missing = append(missing, id)
+		}
+	}
+	return missing, nil
+}
+
+// anyVarNotInEnv returns the first late-binding variable (variablesToCheck) that the expression
+// references but env lacks, or nil if there is none.
 func anyVarNotInEnv(expression string, env map[string]any) *string {
-	for _, variable := range variablesToCheck {
-		if hasVariableInExpression(expression, variable) && !hasVarInEnv(env, variable) {
-			return &variable
+	missing, err := missingVarsInEnv(expression, env)
+	if err != nil {
+		// Unparseable expressions can't be checked; compile/run will surface the error.
+		return nil
+	}
+	for _, id := range missing {
+		if slices.Contains(variablesToCheck, id) {
+			return &id
 		}
 	}
 	return nil
@@ -58,17 +80,10 @@ func expressionReplaceStrict(ctx context.Context, w io.Writer, expression string
 		return expressionReplaceCore(ctx, w, expression, env, false)
 	}
 
-	identifiers, err := getIdentifiers(unmarshalledExpression)
+	missingIdentifiers, err := missingVarsInEnv(unmarshalledExpression, env)
 	if err != nil {
 		// If we can't parse, we can't check variables. Fallback to expressionReplaceCore(false) to report syntax error.
 		return expressionReplaceCore(ctx, w, expression, env, false)
-	}
-
-	missingIdentifiers := []string{}
-	for _, id := range identifiers {
-		if !hasVarInEnv(env, id) {
-			missingIdentifiers = append(missingIdentifiers, id)
-		}
 	}
 
 	for _, id := range missingIdentifiers {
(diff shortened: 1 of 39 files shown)


workflow/controller/dag_test.go
```diff
@@ -4121,6 +4121,211 @@ func TestDAGSkippedOutputRef(t *testing.T) {
...
+      - name: in
+        default: "FALLBACK-FROM-INPUT"
+    container:
+      image: alpine:3.23
+      command: [echo, "{{inputs.parameters.in}}"]
+`
+
+// TestDAGSkippedInputDefaultUsed verifies that when a producer is skipped and its output declares NO
+// valueFrom.default, a consumer referencing that output in its input falls back to the consumer's OWN
+// input default rather than receiving the empty skipped-marker. The Parameter.Skipped marker carries
```
Joibel: Parameter.Skipped isn't implemented that way any more - probably just comment is wrong

workflow/controller/scope.go
```diff
@@ -61,14 +83,99 @@ func (s *wfScope) addArtifactToScope(key string, artifact wfv1.Artifact) {
...
+	tag, _, err := s.resolveVar(val)
+	return err == nil && s.skippedOutputs[tag]
+}
+
+// dropSkippedDefaultedArgs removes resolved arguments whose ORIGINAL value was a pure reference to a
+// skipped/omitted node's output with no producer default, WHEN the consumed template declares its own
+// input default for that parameter. Removing the argument lets that input default apply instead of the
+// empty skipped placeholder. Arguments whose input has no default are left intact (empty value) so the
+// reference still resolves and the node stays live (preserving the #15932 requeue-loop fix).
+func (woc *wfOperationCtx) dropSkippedDefaultedArgs(ctx context.Context, tmplCtx *templateresolution.TemplateContext, holder wfv1.TemplateReferenceHolder, origArgs wfv1.Arguments, resolved *wfv1.Arguments, scope *wfScope) {
```
Joibel: This really feels like it should return err and let callers deal with those failures we're swallowing here otherwise those failures will result in the old bad behaviour

workflow/controller/scope.go
```diff
@@ -61,14 +83,99 @@ func (s *wfScope) addArtifactToScope(key string, artifact wfv1.Artifact) {
...
+	}
+	if tmpl.Outputs.Result != nil {
+		scope.addSkippedParamToScope(fmt.Sprintf("%s.outputs.result", prefix))
+	}
+}
+
+// addSkippedOutputParamToScope registers an output parameter of a skipped/omitted node. If the
+// producing template declared a valueFrom.default for the parameter, that default is used as the
+// scope value so every consumer (task inputs and template-output aggregation alike) sees it.
+// Otherwise it falls back to addSkippedParamToScope's empty-value-plus-marker behavior.
```
Joibel: empty-value-plus-marker is no longer what happens here (just comment is wrong)

Joibel (commented):
docs/variables.md needs updating too for this behaviour (steps L#157 and dag L#176)

docs/enahanced-depends-logic.md should should mention new behaviour too

This is breaking so should get ! in PR title and upgrading.md note

workflow/controller/dag_test.go
```diff
@@ -4121,6 +4121,211 @@ func TestDAGSkippedOutputRef(t *testing.T) {
...
+      - name: in
+    container:
+      image: alpine:3.23
+      command: [echo, "{{inputs.parameters.in}}"]
+`
+
+// TestDAGSkippedOutputDefault verifies that when a DAG task is skipped and its template declares
+// an output parameter with a valueFrom.default, a downstream task referencing that output in its
+// INPUT receives the producer's declared default instead of an empty string.
+func TestDAGSkippedOutputDefault(t *testing.T) {
```
Joibel, line 4170: Please add Steps equivalent of this too, it seems interestingly different enough to warrant a test.

util/template/expression_template.go
```diff
@@ -187,6 +212,13 @@ func expressionReplaceCore(ctx context.Context, w io.Writer, expression string,
 		return 0, fmt.Errorf("failed to evaluate expression: %w", err)
 	}
 	result, err := expr.Run(program, env)
+	if err == nil && result == nil && !shouldAllowFailure && allVarsInEnv(unmarshalledExpression, env) {
```
Joibel: This collapses any expression evaluating to nil to "" when all referenced identifiers are present in env.  Expressions that legitimately evaluate to nil for reasons unrelated to skipped outputs (map-miss, nil-returning sprig/jsonpath) would have previously errored ("failed to evaluate expression", behavior from #15839) and now silently become "".

workflow/controller/dag_test.go
isubasinghe: yes, I initially implemented this by making a CRD change. But I experimented around to see if I can get away without using a CRD change. Nice catch, will address.

workflow/controller/dag_test.go
isubasinghe: This already exists

Joibel (approved):

PR #16226, opened 2026-06-04
fix: set GCS content type from file extension. Fixes #16051

### LINKED ISSUE #16051: GCS json artifact type is text/plain instead of application/json
### Pre-requisites

Uploading a json artifact to GCS with a .json extension results in an uploaded object that has type text/plain. The expected type of the uploaded object is application/json.

possibly related issues:
https://github.com/argoproj/argo-workflows/issues/6553
https://github.com/golang/go/issues/10630

### Version(s)

v.4.0.5

```YAML
kind: Workflow
metadata:
  generateName: upload-json-to-gcs-
spec:
  entrypoint: generate-json
  templates:
  - name: generate-json
    container:
      image: python:3.9-slim
      command: [python, -c]
      args: 
        - |
          import json
          data = {"status": "success", "message": "hello world"}
          with open('/tmp/data.json', 'w') as f:
              json.dump(data, f)
    outputs:
      artifacts:
        - name: json-output
          path: /tmp/data.json
          archive:
            none: {}
          gcs:
            bucket: my-test-bucket
            key: path/in/bucket/data.json
            serviceAccountKeySecret:
              name: my-gcs-credentials
              key: serviceAccountKey
```

```text
```

```text
```

PradeepSD476:
Fixes #16051

## Motivation

When uploading JSON artifacts to GCS, the resulting object was assigned the content type `text/plain` instead of `application/json`.

This occurs because the GCS writer does not automatically infer the desired content type from the artifact filename, causing JSON artifacts to be uploaded with incorrect metadata.

## Modifications

Set the GCS object's content type using the artifact file extension before upload.

The implementation derives the MIME type using `mime.TypeByExtension(filepath.Ext(localPath))` and assigns it to the GCS writer's `ContentType` field prior to copying the object contents.

A fallback of `application/octet-stream` is used when the file extension cannot be mapped to a known MIME type.

## Verification

Reproduced the issue using a workflow that uploads a JSON artifact to GCS.

Before the change:

* Uploaded object content type: `text/plain`

After the change:

* Uploaded object content type: `application/json`

Also verified that MIME type detection correctly resolves common extensions such as:

* `.json` → `application/json`
* `.csv` → `text/csv`
* `.html` → `text/html`
* `.xlsx` → `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`

## Documentation

No documentation changes were required.

## AI

Used ChatGPT for debugging discussion and PR preparation.

### DIFF
--- a/workflow/artifacts/gcs/gcs.go
+++ b/workflow/artifacts/gcs/gcs.go
@@ -318,8 +318,6 @@ func uploadObject(ctx context.Context, client *storage.Client, bucket, key, loca
 	ext := filepath.Ext(localPath)
 	mimeType := mime.TypeByExtension(ext)
 
-	fmt.Println("mimeType =", mimeType)
-
 	if mimeType == "" {
 		mimeType = "application/octet-stream"
 	}
--- a/workflow/artifacts/gcs/gcs.go
+++ b/workflow/artifacts/gcs/gcs.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


mailnike (commented):
Thanks for adding content-type detection for GCS uploads. One blocker before this can merge: the diff includes a stray `fmt.Println("mimeType =", mimeType)` in `uploadObject()`, which writes to stdout on every artifact upload and bypasses the logging framework. That debug line should be removed (or converted to a logger call at debug level).

One thing worth considering as well: Argo archives artifacts as `.tgz` by default, so for archived artifacts an extension-based mime type may not reflect the actual stored content. Otherwise the approach looks reasonable.

PradeepSD476:
@mailnike Thanks for catching that! I've removed the stray fmt.Println

Regarding the archived artifacts point. My current implementation derives the MIME type from the extension of the file being uploaded, so if the artifact is archived as .tgz, the uploaded object's content type would reflect the archive rather than the original artifact type. I wasn't sure whether the intended behavior should instead preserve the original artifact's MIME type, so I left the implementation as-is. Happy to adjust it if that's the preferred behavior.

PR #16238, opened 2026-06-09
fix: resolve race condition in custom metric initialization

### LINKED ISSUE #16106: OCTOU Race Condition in Custom Metric Initialization Causes Workflow to Fail with Panic
### Pre-requisites

When a custom metric with `realtime: true` is defined (e.g. via `workflowDefault`) and two or more workflows are processed by the controller simultaneously for the first time after a controller restart, a TOCTOU race condition in `ensureBaseMetric` causes a panic.

```
internal error: unexpected userdata on custom metric <metric-name>
```

The workflow immediately transitions to Error phase without executing any steps.

### Root cause

[ensureBaseMetric](https://github.com/argoproj/argo-workflows/blob/351c306070548bc1403ff17d5829cc3bd65e1eef/workflow/metrics/metrics_custom.go#L158-L178) has a window between [createCustomMetric](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/workflow/metrics/metrics_custom.go#L247-L271) and [inst.SetUserdata](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/workflow/metrics/metrics_custom.go#L176) where another goroutine can find the newly created Instrument via [matchExistingMetric](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/workflow/metrics/metrics_custom.go#L130-L156),
call [getOrCreateValue](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/workflow/metrics/metrics_custom.go#L79-L92), and then [customUserData(inst, true)](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/workflow/metrics/metrics_custom.go#L67-L77) — which panics because [inst.GetUserdata()](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/workflow/metrics/metrics_custom.go#L68) is still nil at that point.

```
Goroutine 1: matchExistingMetric → nil
Goroutine 1: createCustomMetric   (userdata = nil)
Goroutine 2: matchExistingMetric → inst found (userdata = nil)
Goroutine 2: getOrCreateValue → customUserData(inst, true) → PANIC
Goroutine 1: inst.SetUserdata(newUserData())
```

This only occurs when the metric is being initialized for the first time (i.e., after a controller restart), so it is rare in practice.

### Version(s)

v4.0.5

```YAML
kind: Workflow
metadata:
  generateName: metrics-race-
spec:
  entrypoint: main
  metrics:
    prometheus:
      - name: exec_duration_gauge
        help: "Duration of execution"
        labels:
          - key: workflow_uid
            value: "{{ workflow.uid }}"
          - key: status
            value: "{{ workflow.status }}"
        gauge:
          realtime: true
          value: "{{ workflow.duration }}"
  templates:
    - name: main
      container:
        image: argoproj/argosay:v2
        command: [sh, -c]
        args: ["sleep 5"]
```

```text
time=2026-05-16T15:05:00.019Z level=ERROR msg="Recovered from panic" r="internal error: unexpected userdata on custom metric exec_duration_gauge" component=workflow_worker workflow=<workflow-name> namespace=<namespace> stack="goroutine 358 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:26 +0x64
github.com/argoproj/argo-workflows/v4/workflow/controller.(*wfOperationCtx).operate.func2()
	/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:201 +0x50
panic({0x229e2c0?, 0x400103e680?})
	/usr/local/go/src/runtime/panic.go:783 +0x120
github.com/argoproj/argo-workflows/v4/workflow/metrics.customUserData(0x4000fc8480?, 0x24?)
	/go/src/github.com/argoproj/argo-workflows/workflow/metrics/metrics_custom.go:73 +0xa8
github.com/argoproj/argo-workflows/v4/workflow/metrics.getOrCreateValue(0x4000e36660?, {0x40009428c0, 0x134}, {0x40008fa380, 0x9, 0x10})
	/go/src/github.com/argoproj/argo-workflows/workflow/metrics/metrics_custom.go:80 +0x44
github.com/argoproj/argo-workflows/v4/workflow/metrics.(*Metrics).UpsertCustomMetric(0x4000700c80, {0x2ced3b0, 0x4001581080}, 0x4000e36660, {0x4000fc9b00, 0x24}, 0x400103e350)
	/go/src/github.com/argoproj/argo-workflows/workflow/metrics/metrics_custom.go:188 +0x84
github.com/argoproj/argo-workflows/v4/workflow/controller.(*wfOperationCtx).computeMetrics(0x400141c840, {0x2ced3b0, 0x4001581080}, {0x4000dca120, 0x2, 0x1a?}, 0x4000f82e10, 0x4000f82e70, 0x1)
	/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:4122 +0x5e0
github.com/argoproj/argo-workflows/v4/workflow/controller.(*wfOperationCtx).operate(0x400141c840, {0x2ced3b0, 0x4001581080})
	/go/src/github.com/argoproj/argo-workflows/workflow/controller/operator.go:286 +0xa34
..."
time=2026-05-16T15:05:00.019Z level=INFO msg="updated phase" workflow=<workflow-name> namespace=<namespace> component=workflow_worker fromPhase="" toPhase=Error
```

```text
N/A — the workflow fails before any pod is scheduled (`fromPhase=""`)
```

nebojsaj1726:
Fixes #16106

### Motivation

Concurrent workflow submissions could trigger a panic in `ensureBaseMetric` and data races in `UpsertCustomMetric`. The root cause is a TOCTOU race: one goroutine creates a metric instrument and a second goroutine finds it via `matchExistingMetric` before `SetUserdata` is called, resulting in `customUserData` receiving a nil userdata and panicking. The race detector also revealed additional write-write and read-write races on `customMetricValue` fields accessed concurrently by the controller and the Prometheus scraper.

### Modifications

- `util/telemetry/instrument.go`: added `sync.RWMutex` to `Instrument` to make `SetUserdata`/`GetUserdata` thread-safe.
- `workflow/metrics/metrics.go`: added `customMetricsMutex sync.Mutex` to `Metrics`.
- `workflow/metrics/metrics_custom.go`:
  - Added `sync.RWMutex` to `customMetricValue`.
  - Wrapped the create+initialize sequence in `ensureBaseMetric` with `customMetricsMutex` using a double-checked locking pattern.
  - Protected `prometheusValue`, `rtValueFunc`, and `lastUpdated` writes in `UpsertCustomMetric` with `metricValue.mutex.Lock()`.
  - Protected reads of the same fields in `customCallback` and `runCustomGC` with `RLock()`.

### Verification

Added `TestUpsertCustomMetric_Concurrency` in `workflow/metrics/metrics_custom_test.go` which spawns 200 goroutines all calling `UpsertCustomMetric` simultaneously on the same metric. Before this fix, `go test -race` reported multiple data races and caught panics. After the fix, the test passes cleanly with no races detected.

Note: make lint-go revealed a pre-existing unrelated lint issue in util/file/watch.go (revive time-equal), not introduced by this change.

### AI

Claude, all code reviewed and verified by the author.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal synchronization mechanisms across metrics and telemetry systems to ensure reliable and consistent behavior during concurrent access and operations.

* **Tests**
  * Added comprehensive concurrency testing for custom metrics operations to validate system stability under simultaneous metric updates.

### DIFF
--- a/workflow/metrics/metrics_custom.go
+++ b/workflow/metrics/metrics_custom.go
@@ -17,6 +17,7 @@ import (
 type RealTimeValueFunc func() float64
 
 type customMetricValue struct {
+	mutex sync.RWMutex
 	// We are faking a settable and incrementable up/down gauge/coutner
 	// for compatibility with prometheus. callback() reads this.
 	// This is used for counters, gauges and realtime custom metrics,
@@ -104,11 +105,15 @@ func (i *customInstrument) customCallback(ctx context.Context, o metric.Observer
 	ud.mutex.RLock()
 	defer ud.mutex.RUnlock()
 	for _, value := range ud.values {
+		// Manual RLock/RUnlock here instead of defer because we are in a loop.
+		// There are no return paths inside this loop, so this is safe.
+		value.mutex.RLock()
 		if value.rtValueFunc != nil {
 			i.ObserveFloat(ctx, o, value.rtValueFunc(), value.getLabels())
 		} else {
 			i.ObserveFloat(ctx, o, value.prometheusValue, value.getLabels())
 		}
+		value.mutex.RUnlock()
 	}
 	return nil
 }
@@ -156,11 +161,30 @@ func (m *Metrics) matchExistingMetric(metricSpec *wfv1.Prometheus) (*telemetry.I
 }
 
 func (m *Metrics) ensureBaseMetric(metricSpec *wfv1.Prometheus, ownerKey string) (*telemetry.Instrument, error) {
+	// Fast path: check if metric already exists and is fully initialized (double-checked locking).
 	metric, err := m.matchExistingMetric(metricSpec)
 	if err != nil {
 		return nil, err
 	}
 	if metric != nil {
+		if customUserData(metric, false) != nil {
+			m.attachCustomMetricToWorkflow(metricSpec, ownerKey)
+			return metric, nil
+		}
+	}
+
+	m.customMetricsMutex.Lock()
+	defer m.customMetricsMutex.Unlock()
+
+	// Slow path: re-check under lock before creating.
+	metric, err = m.matchExistingMetric(metricSpec)
+	if err != nil {
+		return nil, err
+	}
+	if metric != nil {
+		if customUserData(metric, false) == nil {
+			metric.SetUserdata(newUserData())
+		}
 		m.attachCustomMetricToWorkflow(metricSpec, ownerKey)
 		return metric, nil
 	}
@@ -186,6 +210,8 @@ func (m *Metrics) UpsertCustomMetric(ctx context.Context, metricSpec *wfv1.Prome
 		return err
 	}
 	metricValue := getOrCreateValue(baseMetric, metricSpec.GetKey(), metricSpec.Labels)
+	metricValue.mutex.Lock()
+	defer metricValue.mutex.Unlock()
 	metricValue.lastUpdated = time.Now()
 
 	metricType := metricSpec.GetMetricType()
@@ -213,6 +239,7 @@ func (m *Metrics) UpsertCustomMetric(ctx context.Context, metricSpec *wfv1.Prome
 		if err != nil {
 			return err
 		}
+		// Record does its own locking if needed by the exporter
 		baseMetric.Record(ctx, val, metricValue.getLabels())
 	case metricType == wfv1.MetricTypeCounter:
 		val, err := strconv.ParseFloat(metricSpec.Counter.Value, 64)
@@ -288,11 +315,17 @@ func (m *Metrics) runCustomGC(ttl time.Duration) {
 		}
 		ud.mutex.Lock()
 		for key, value := range ud.values {
-			if time.Since(value.lastUpdated) > ttl {
+			value.mutex.RLock()
+			lastUpdated := value.lastUpdated
+			isRealtime := value.rtValueFunc != nil
+			completed := value.completed
+			value.mutex.RUnlock()
+
+			if time.Since(lastUpdated) > ttl {
 				switch {
-				case value.rtValueFunc != nil && value.completed:
+				case isRealtime && completed:
 					delete(ud.values, key)
-				case value.rtValueFunc == nil:
+				case !isRealtime:
 					delete(ud.values, key)
 				}
 			}
--- a/workflow/metrics/metrics_custom.go
+++ b/workflow/metrics/metrics_custom.go
@@ -108,15 +108,17 @@ func (i *customInstrument) customCallback(ctx context.Context, o metric.Observer
 	ud.mutex.RLock()
 	defer ud.mutex.RUnlock()
 	for _, value := range ud.values {
-		// Manual RLock/RUnlock here instead of defer because we are in a loop.
-		// There are no return paths inside this loop, so this is safe.
-		value.mutex.RLock()
-		if value.rtValueFunc != nil {
-			i.ObserveFloat(ctx, o, value.rtValueFunc(), value.getLabels())
-		} else {
-			i.ObserveFloat(ctx, o, value.prometheusValue, value.getLabels())
-		}
-		value.mutex.RUnlock()
+		// Wrapped in a closure so defer runs per iteration, protecting against
+		// panics in rtValueFunc leaving the lock permanently held.
+		func() {
+			value.mutex.RLock()
+			defer value.mutex.RUnlock()
+			if value.rtValueFunc != nil {
+				i.ObserveFloat(ctx, o, value.rtValueFunc(), value.getLabels())
+			} else {
+				i.ObserveFloat(ctx, o, value.prometheusValue, value.getLabels())
+			}
+		}()
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


Joibel:
@coderabbitai review

Joibel:
### Code review

Found 3 issues:

1. Residual nil-userdata panic window via the Prometheus scraper: in `ensureBaseMetric`'s slow path, `createCustomMetric` registers `customCallback` with the OTel SDK (`RegisterCallback`) *before* `inst.SetUserdata(newUserData())` runs. The scraper goroutine never acquires `customMetricsMutex`, so a scrape landing between callback registration and `SetUserdata` still hits `customUserData(i.Instrument, true)` with nil userdata and panics — the same panic class this PR fixes for concurrent controller goroutines. Consider setting userdata before registering the callback (or making `customCallback` tolerate nil userdata).

https://github.com/argoproj/argo-workflows/blob/741fc1e31074a874a15cd38291797f39809b98d8/workflow/metrics/metrics_custom.go#L190-L201

2. The new concurrency test only races `UpsertCustomMetric` callers against each other — it never triggers a metric collection/scrape (`customCallback`) or `runCustomGC`, so the new `RLock` guards in those paths (the "controller vs Prometheus scraper" races in the PR description) are unverified, and the window in issue 1 can't be caught. It also uses a non-realtime gauge while #16106 is about `realtime: true`, and its only assertion is a panic count, which is meaningful only under `go test -race` (the field races are silent without it).

https://github.com/argoproj/argo-workflows/blob/741fc1e31074a874a15cd38291797f39809b98d8/workflow/metrics/metrics_custom_test.go#L11-L53

3. Inconsistent lock discipline on `completed`: `runCustomGC` now reads `value.completed` under `value.mutex.RLock()`, but `handleRealtimeMetricsForWfUID` writes it holding only `ud.mutex` (it wasn't updated to take `value.mutex`). Not a data race today since both paths also hold `ud.mutex`, but the per-value-mutex invariant isn't uniform — a future reader of `completed` that doesn't hold `ud.mutex` would silently race. Worth taking `value.mutex` in the writer or documenting which lock guards the field.

https://github.com/argoproj/argo-workflows/blob/741fc1e31074a874a15cd38291797f39809b98d8/workflow/metrics/metrics_custom.go#L316-L323

nebojsaj1726:
@Joibel I've addressed the feedback by implementing the following changes:
   - Closed the Panic Window: Reordered initialization to set userdata before registering callbacks and updated customCallback to gracefully handle nil userdata.
   - Fixed Lock Discipline: Ensured value.completed is updated under value.mutex in handleRealtimeMetricsForWfUID for consistency.
   - Hardened Initialization: Added nil-checks to all SetUserdata paths to prevent accidental data loss.
   - Expanded Coverage: Upgraded TestUpsertCustomMetric_Concurrency to a comprehensive stress test (including scrapes and GC) and added TestUpsertCustomMetric_PanicWindow to verify the fix.

Note on Test Races: While verifying with -race, I identified pre-existing data races in TestMetricGC and TestRealtimeMetricGC. I've kept this PR focused on the reported issues but can provide a separate commit to fix those test races (by adding locks to assertions) if desired.

workflow/metrics/metrics_custom.go
```diff
@@ -100,15 +101,22 @@ type customInstrument struct {
...
+		return nil
+	}
 	ud.mutex.RLock()
 	defer ud.mutex.RUnlock()
 	for _, value := range ud.values {
+		// Manual RLock/RUnlock here instead of defer because we are in a loop.
+		// There are no return paths inside this loop, so this is safe.
+		value.mutex.RLock()
 		if value.rtValueFunc != nil {
 			i.ObserveFloat(ctx, o, value.rtValueFunc(), value.getLabels())
```
isubasinghe: what if the rtValueFunc() panics? 
It feels like a defer is needed for that case? 

```go
    func() {
        value.mutex.RLock()
        defer value.mutex.RUnlock()
        if value.rtValueFunc != nil {
            i.ObserveFloat(ctx, o, value.rtValueFunc(), value.getLabels())
        } else {
            i.ObserveFloat(ctx, o, value.prometheusValue, value.getLabels())
        }
    }()
```

isubasinghe (approved):
From my look through, the rtValueFuncs, seem to not panic though so I think this is fine. 

Will wait for another review. 

But if you address it, I will merge it in right away.

workflow/metrics/metrics_custom.go
nebojsaj1726: Addressed it with an IIFE to ensure defer runs per iteration even if rtValueFunc panics.

Joibel:
LGTM, thank you

PR #16261, opened 2026-06-12
feat: configurable compression algorithms for node compression. Fixes #16262

### LINKED ISSUE #16262: Consider allowing for choosing the compression algorithm and level
# Summary

Allow users to choose between compressing with gzip as done previously or to use zstd. 

## Use Cases

In my synthetic benchmarks I was able to witness a 18% reduction in size after switching to zstd.
This will allow users to stretch the standalone version of the controller more before needing node offloading.

It makes sense to make zstd the default, but this should be done in later release as it will be a breaking change.

---

isubasinghe:
Fixes #16262

### Motivation

Node statuses are gzip-compressed into `/status/compressedNodes`. In my benchmarks zstd is ~12–18% smaller at the same speed, brotli ~28–41% smaller for more CPU, which stretches how far you get before needing offloading. Switching the default would break downgrades, so this just makes it configurable; gzip stays the default.

### Modifications

- New `WORKFLOW_COMPRESSION_ALGORITHM` (`gzip`/`zstd`/`brotli`) and `WORKFLOW_COMPRESSION_LEVEL` env vars in `util/file`. Invalid values warn and fall back.
- Decompression auto-detects the algorithm (magic bytes; brotli by elimination), independent of the env vars.
- No new deps: `klauspost/compress` and `andybalholm/brotli` promoted from indirect to direct.
- `hack/compression-bench`: the harness behind the numbers. It also ruled out zstd dictionaries and proto encoding.

### Verification

Unit tests in `util/file` (round-trips, detection, fallbacks) and end-to-end zstd/brotli tests in `workflow/packer`. The harness round-trip-verifies every codec.

### Documentation

New section in `docs/offloading-large-workflows.md` with the benchmark table and a downgrade warning. Env vars listed in `docs/environment-variables.md`. Feature file added for release notes.

### AI

Claude Code helped with the code, tests, docs, and this description. Reviewed and benchmarked by me.

## Summary by CodeRabbit

* **New Features**
  * Workflow node status compression can now use `gzip`, `zstd`, or `brotli`, with configurable compression level settings.
  * Decompression now automatically detects the compression format, improving compatibility when reading stored workflows.

* **Bug Fixes**
  * Improved handling of large workflow statuses to support more compression options without breaking restoration.

* **Documentation**
  * Updated environment-variable and workflow documentation with guidance, defaults, and compatibility warnings for compression settings.


util/file/fileutil_test.go
```diff
@@ -37,6 +37,103 @@ func TestCompressContentString(t *testing.T) {
 	}
 }
 
+// TestZstdCompression ensures the writer switches to zstd via the environment
+// variable while the reader detects the algorithm from the content alone.
+func TestZstdCompression(t *testing.T) {
+	ctx := logging.TestContext(t.Context())
+	content := "{\"pod-limits-rrdm8-591645159\":{\"id\":\"pod-limits-rrdm8-591645159\",\"phase\":\"Succeeded\"}}"
+
+	var compressed string
```
Joibel: `TestZstdCompression` and `TestBrotliCompression` share this `compressed` variable across subtests: the first subtest writes it, later subtests read it. Running a later subtest in isolation (e.g. `go test -run 'TestZstdCompression/DecompressDetectsZstdWithoutEnvVar'`) leaves it empty and the test fails — verified locally. The subtests depend on source-order execution and aren't independently runnable/debuggable. Consider compressing fresh inside each subtest (or a helper) so each stands alone.

util/file/fileutil_test.go
```diff
@@ -37,6 +37,103 @@ func TestCompressContentString(t *testing.T) {
...
+		comp := file.CompressEncodeString(ctx, content)
+		resultString, err := file.DecodeDecompressString(ctx, comp)
+		require.NoError(t, err)
+		assert.Equal(t, content, resultString)
+	})
+}
+
+// TestCompressionLevel ensures WORKFLOW_COMPRESSION_LEVEL applies to each
+// algorithm with algorithm-specific ranges.
+func TestCompressionLevel(t *testing.T) {
```
Joibel, line 115: These level tests (and `TestBrotliCompression`'s `InvalidLevelFallsBackToDefault`/`ExplicitLevel`) only assert that compress→decompress round-trips — which passes regardless of which level was actually used. A regression that ignored `WORKFLOW_COMPRESSION_LEVEL` entirely would still go green, so the `compressionLevel()` parsing/validation path is effectively unverified. Observing an effect of the level would fix this, e.g. assert that max-level output is smaller than (or differs from) min-level output for the same input, or that invalid-level output is byte-identical to unset-level output.

util/file/fileutil_test.go
```diff
@@ -37,6 +37,103 @@ func TestCompressContentString(t *testing.T) {
 	}
 }
 
+// TestZstdCompression ensures the writer switches to zstd via the environment
+// variable while the reader detects the algorithm from the content alone.
+func TestZstdCompression(t *testing.T) {
+	ctx := logging.TestContext(t.Context())
+	content := "{\"pod-limits-rrdm8-591645159\":{\"id\":\"pod-limits-rrdm8-591645159\",\"phase\":\"Succeeded\"}}"
```
Joibel: Nit: this 90-character JSON fixture (and the `ctx := logging.TestContext(t.Context())` setup) is duplicated verbatim in all three new test functions. A package-level `const` and/or a small round-trip helper would remove the repetition.

util/file/fileutil.go
```diff
@@ -101,15 +153,32 @@ func DecodeDecompressString(ctx context.Context, content string) (string, error)
 	return string(dBuf), nil
 }
 
-// CompressContent will compress the byte array using zip writer
+// CompressContent will compress the byte array using the algorithm selected
+// by `CompressionAlgorithmEnvVarKey` (gzip by default). DecompressContent
+// detects the algorithm from the content, so the variable only affects writes.
 func CompressContent(ctx context.Context, content []byte) []byte {
+	switch env.GetString(CompressionAlgorithmEnvVarKey, GZipAlgorithm) {
```
Joibel: An unrecognized `WORKFLOW_COMPRESSION_ALGORITHM` value (typo, or wrong case like `ZSTD` — the comparison is case-sensitive) silently falls through to gzip with no log line, whereas an invalid `WORKFLOW_COMPRESSION_LEVEL` warns via `compressionLevel()`. An operator who sets the algorithm and fat-fingers it gets no feedback that their setting was ignored. Suggest logging a warning in a `default:` case for parity.

hack/compression-bench/main.go
```diff
@@ -0,0 +1,239 @@
+// compression-bench measures candidate replacements for the node-status
+// compression in workflow/packer (JSON+gzip+base64). It synthesizes node
+// statuses from the examples/ specs, trains zstd dictionaries on half the
+// corpus, and reports sizes/ratios/timings for each codec on the held-out
+// half. See docs/superpowers/specs/2026-06-12-node-compression-benchmark-design.md.
```
Joibel: This references `docs/superpowers/specs/2026-06-12-node-compression-benchmark-design.md`, which doesn't exist in the repo — the design doc actually ships as `hack/compression-bench/DESIGN.md` in this PR. Looks like a leftover pointer to local working notes; suggest pointing at `DESIGN.md` or dropping the sentence.

hack/compression-bench/DESIGN.md
```diff
@@ -0,0 +1,89 @@
+# Node-Status Compression Benchmark
+
+## Goal
+
+Measure whether replacing the current node-status compression (JSON → gzip →
+base64, `workflow/packer`) with proto serialization and/or zstd with a trained
+dictionary is worth a format migration. This is a measurement prototype only:
```
Joibel: Should this harness ship in the PR at all? Its own design doc describes it as "a measurement prototype only" whose numbers "decide whether an integration phase happens later" — and that decision is now made (the results are baked into `docs/offloading-large-workflows.md`). As committed it: benchmarks codecs the feature doesn't ship (proto encoding, zstd dictionaries), is the sole user of `klauspost/compress/dict`, isn't wired into the Makefile or CI (unlike the other `hack/` tools, per `hack/README.md`), yet must compile and pass lint forever via `go build ./...`. Consider dropping it (keeping the measured results in the docs) or moving it out of the mainline tree.

.features/pending/configurable-node-status-compression.md
```diff
@@ -0,0 +1,8 @@
+Description: Configurable node status compression algorithm (gzip, zstd, or brotli) and level
+Authors: [Isitha Subasinghe](https://github.com/isubasinghe)
+Component: General
+Issues: 16262
+
+Large workflow node statuses can now be compressed with `zstd` or `brotli` instead of gzip via the `WORKFLOW_COMPRESSION_ALGORITHM` environment variable, with `WORKFLOW_COMPRESSION_LEVEL` tuning the level.
+Decompression auto-detects the algorithm.
+See [node status compression](../docs/offloading-large-workflows.md#node-status-compression).
```
Joibel: This link will break when rendered: `hack/featuregen` copies the feature body verbatim into `docs/new-features.md`, and from there `../docs/offloading-large-workflows.md` escapes the mkdocs docs root (mkdocs runs with `strict: true`). Should be the bare form used everywhere else in `docs/` — `offloading-large-workflows.md#node-status-compression` — matching this PR's own `environment-variables.md`.

workflow/packer/packer_test.go
```diff
@@ -55,6 +56,40 @@ func TestDecompressWorkflow(t *testing.T) {
 		assert.NotEmpty(t, wf.Status.Nodes)
 		assert.Empty(t, wf.Status.CompressedNodes)
 	})
+	t.Run("LargeWorkflowZstd", func(t *testing.T) {
```
Joibel: `LargeWorkflowZstd` and `LargeWorkflowBrotli` are line-for-line copies of the existing `LargeWorkflow` subtest, differing only by the `t.Setenv` line — three copies of the same ~17-line body. A table-driven subtest over `{"", file.ZStdAlgorithm, file.BrotliAlgorithm}` would cover all three with one body.

Joibel (approved):

PR #16267, opened 2026-06-13
feat: archive init and wait container logs. Fixes #12640

### LINKED ISSUE #12640: Support archive init and wait container logs
# Summary

Currently argo `ArchiveLogs` only supports logs from `main` container.
From time to time user is asking if it supports `init` and `wait` container log.

I'm aware the archivelog docs recommend a proper logging facility for logging,
Maybe this is still something we could consider if there's enough 👍

## Use Cases

When would you use this?
Archive `init` and `wait` container log for debugging purpose

---

myzk-a:
Fixes #12640

> **Heads-up on the diff size (for reviewers):** GitHub shows `+2,567 / 50 files`, but **~1,880 lines (~73%) are generated by `make codegen`** — the `quick-start-*.yaml` manifests, CRD manifests under `manifests/base/crds/`, `generated.pb.go`, `openapi_generated.go`, `swagger.json` / `schema.json`, `zz_generated.deepcopy.go`, `generated.proto`, `docs/fields.md` / `executor_swagger.md`, and the Java SDK docs.
>
> The **hand-written change is ~690 lines**, of which ~370 is tests and ~75 is docs. The **core logic is only ~240 lines**, concentrated in:
> - `cmd/argoexec/commands/logs.go` / `init.go` / `auxiliary.go`
> - `workflow/executor/executor.go`
> - `workflow/controller/workflowpod.go`
> - `util/logging/slog.go`
> - the type definitions in `pkg/apis/workflow/v1alpha1/`

### Motivation

Today Argo only archives the `main` container's logs. When a Pod is garbage collected, the logs of Argo's own system containers — the `init` container (which loads input artifacts) and the `wait` container (which saves outputs and logs), or the single `supervisor` container that plays both roles under the init-less pod layout (#16161) — are lost forever.

This makes it hard to debug problems that happen *outside* the user's `main` container after the Pod is gone, for example:

- artifact upload failures or other errors in `wait`,
- reviewing which input artifacts the `init` container loaded.

This PR lets users opt in to archiving the system container logs as artifacts, controlled independently from `archiveLogs` so you only pay to store them when you need them.

### Modifications

Added a new `archiveSystemContainerLogs` flag, settable at the controller config-map, workflow spec, or template level, following the same priority rules as the existing `archiveLogs`.

Since #16161 introduced the opt-in init-less pod layout, the set of system containers depends on the layout, and this PR supports both:

| | Legacy layout | Init-less layout |
|---|---|---|
| Archived artifacts | `init-logs`, `wait-logs` | `supervisor-logs` |
| When input artifact loading fails | Nothing is archived — `init` fails and `wait` never starts (documented limitation) | `supervisor-logs` **is** archived, because the supervisor still runs its post-main phase |

#### Architecture (after this change)

The system containers tee their own logs into per-container `combined` files on the shared `/var/run/argo` emptyDir volume — the same convention the emissary already uses for `main`. The auxiliary container (`wait`, or `supervisor` in init-less mode) then reads those files back in `SaveLogs` and uploads each as an artifact, alongside the existing `main-logs`. The tee is unconditional (the flag is not yet readable at that point); the archiving decision is made once, in `SaveLogs`.

```mermaid
flowchart TD
    subgraph Pod["Workflow Pod (legacy layout)"]
        direction TB
        Init["init container<br/>(argoexec init)"]
        Main["main container<br/>(user cmd via emissary)"]
        Wait["wait container<br/>(argoexec wait)"]
        Vol[("/var/run/argo/ctr<br/>shared emptyDir<br/>init/combined · main/combined · wait/combined")]
        Init -- tee --> Vol
        Main -- emissary tee --> Vol
        Wait -- tee --> Vol
        Vol -- "wait reads back (SaveLogs)" --> Wait
    end
    Wait -- upload --> Store["Artifact storage<br/>main-logs · init-logs (new) · wait-logs (new)"]
```

```mermaid
flowchart TD
    subgraph Pod2["Workflow Pod (init-less layout)"]
        direction TB
        Sup["supervisor container<br/>(argoexec supervisor)"]
        Main2["main container<br/>(user cmd via emissary)"]
        Vol2[("/var/run/argo/ctr<br/>shared emptyDir<br/>supervisor/combined · main/combined")]
        Sup -- tee --> Vol2
        Main2 -- emissary tee --> Vol2
        Vol2 -- "supervisor reads back (SaveLogs)" --> Sup
    end
    Sup -- upload --> Store2["Artifact storage<br/>main-logs · supervisor-logs (new)"]
```

- **Type / API** (`pkg/apis/workflow/v1alpha1/`): added the `ArchiveSystemContainerLogs` field to `ArtifactRepository` and the workflow/template types, plus a `SaveSystemContainerLogsAsArtifact()` helper mirroring the existing `SaveLogsAsArtifact()`.
- **Controller** (`workflow/controller/workflowpod.go`): added `IsArchiveSystemContainerLogs(tmpl)`, a faithful copy of `IsArchiveLogs` with the same `controller(on) > template > workflow > controller(off)` priority.
- **Executor**:
  - `cmd/argoexec/commands/auxiliary.go`: the tee lives in `runAuxiliaryContainer`, the lifecycle scaffolding shared by the legacy `wait` command and the init-less `supervisor` command, so both layouts are covered in one place. The legacy-only `init` command (`init.go`) sets up its own tee, since it does not go through `runAuxiliaryContainer`. A tee setup failure is treated as a log-archive failure and does not block the container's primary work.
  - `util/logging/slog.go`: added `TeeLogger` and a `multiHandler` so a logger can fan out to its original destination **and** the combined file, preserving fields, format, and hooks.
  - `workflow/executor/executor.go`: `SaveLogs` derives the system container names from the pod layout (`common.IsInitlessPod()`) and uploads their combined files as `<container>-logs` artifacts when the flag is set. The auxiliary container's own log is saved last so it includes the preceding save operations; a missing combined file is skipped via `fs.ErrNotExist` as a safety net for tee setup failure.
- **UI** (`ui/src/shared/services/workflows-service.ts`): dropped the `container === 'main'` restriction so archived logs of non-main containers can be viewed after the Pod is gone.

### Verification

- **Unit tests**:
  - `workflow/executor/executor_test.go` — `TestSaveLogs`: covers each flag combination, the `ErrNotExist` skip, that the auxiliary container is saved last, and the init-less cases (`t.Setenv(common.EnvVarInitlessPod, "true")` switches the targets to `supervisor`).
  - `cmd/argoexec/commands/logs_test.go` — `Test_teeContainerLogs`: combined file creation and write tee-through.
- **E2E tests** (layout-aware via `fixtures.AuxContainerName()` / `E2E_INITLESS`, exercised by both the regular and the `initless: true` CI dimensions):
  - `TestArchiveSystemContainerLogs` — verifies `main-logs` and the per-layout system log artifacts exist on a succeeded Pod.
  - `TestArchiveSystemContainerLogsWhenArtifactLoadFails` — legacy: a failing `init` container still ends cleanly in the `WorkflowError` phase (regression guard for the tee added in `init.go`); init-less: `supervisor-logs` is archived even though pre-main failed.
- `make pre-commit -B` passes (codegen + lint + docs + features-validate), and `go build ./...` is clean.

### Documentation

- Added an "Archiving System Container Logs" section (with a "Limitations" subsection) to `docs/configure-archive-logs.md`, including an example manifest and the init-less behavior.
- Added a feature note at `.features/pending/archive-system-container-logs.md`.
- Documented limitations: in the legacy layout a *failing* `init` container is **not** captured, because the `wait` container never starts to archive it (the init-less layout does capture this case).

### AI

Claude Code was used to help review the implementation, write tests, and draft the docs, commit messages, and this PR description. All changes were reviewed by the author.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in archiving for system container logs as `init-logs`, `wait-logs`, or `supervisor-logs`.
  * Supports independent configuration from main-container log archiving with defined setting priorities.
  * Archived logs for non-main containers can be viewed after workflow completion.
* **Documentation**
  * Added configuration guidance, field references, examples, and limitations.
* **Tests**
  * Added coverage for system-log archiving, artifact-loading failures, and init-less workflows.


myzk-a:
Addressed nitpicks #2–#4 in [284defa](https://github.com/argoproj/argo-workflows/pull/16267/commits/284defa0f1c18738824eb047567d857b91e45cd6). Two notes:

- **3**: adopted the diagnosis, but `require.Empty` would fail — `main` legitimately records one unconfigured-storage error here. Used `require.Len(t, we.errors, 1)` + `assert.EqualError` instead.
- **1**: deliberately not adopted. Asserting `init-logs` absence would pass even if archiving were broken, and would freeze a documented limitation into a test contract. The limitation is documented in `docs/configure-archive-logs.md`.

myzk-a:
Good catch on the resource templates — fixed in 35fc008: `needsAuxCtr` now also
considers `SaveSystemContainerLogsAsArtifact()`, with unit tests for both the legacy
(wait) and init-less (supervisor) layouts.

Data templates are intentionally unchanged: they never get an aux container even with
`archiveLogs` enabled, because argoexec runs as the main container there (see the
comment above `needsAuxCtr`). Changing that would alter pre-existing behavior beyond
the scope of this PR.

docs/configure-archive-logs.md
```diff
@@ -62,6 +62,61 @@ spec:
       archiveLogs: true
 ```
 
+## Archiving Init and Wait Container Logs
+
+By default, only the `main` container's logs are archived. Every Workflow Pod also runs Argo's own
```
tico24: Can we get one sentence per line please as per the docs guidelines? Just makes it much easier to diff in future. Thanks

myzk-a: Thanks for the review, and for the pointer to the docs guidelines! Fixed in e5af2dfde110399db240e35012e9e3aa45c50df8 — reformatted the whole section to one sentence per line.

myzk-a:
Since #16161 (init-less pod layout) was merged, I've merged `main` and extended this PR so `archiveSystemContainerLogs` works under both pod layouts: the legacy layout archives `init-logs` / `wait-logs`, and the init-less layout archives a single `supervisor-logs` — notably still capturing it when input artifact loading fails, which the legacy layout cannot. The init-less delta is small: a container-name switch in `SaveLogs` plus moving the tee into the shared `runAuxiliaryContainer`. If you'd prefer the init-less support split into a follow-up PR to keep review scope down, happy to do that. The PR description has been updated accordingly.

myzk-a:
@coderabbitai resume

myzk-a:
> **Add `archiveSystemContainerLogs` to the stub and test the new feature.**

Done in 7c47be27e — added the toggle to `stubPodBuilderDeps` and positive tests for both layouts (`TestBuild_InitlessResourceSystemLogsOnly_SchedulesSupervisor` / `TestBuild_LegacyResourceSystemLogsOnly_SchedulesWaitContainer`).

myzk-a:
@coderabbitai resume

myzk-a:
Merged latest `main` and regenerated — back to mergeable.

**On the one red check:** `Windows Unit Tests` fails on `workflow/sync.TestParallelismUpdate` (`:225` / `:226`), the flake #16224 fixes — it also reproduces on `main` (run 30632470836), and this PR touches nothing under `workflow/sync`. Everything else is green, including all 31 E2E jobs.

This round also picks up the composition-root convention from #16554 — `SaveLogs` now reads `we.initlessPod` instead of calling `common.IsInitlessPod()` from inside the executor package.

@Joibel this implements the separate-flag design you asked for in #12640, and extends it to the init-less layout from #16161 (`supervisor-logs`). It's been open a while and has taken two rounds of protobuf renumbering from unrelated merges, so a review when you have a moment would be much appreciated.

Thanks @thc1006 for checking that field 46 stayed free for this PR in #16587.

docs/configure-archive-logs.md
```diff
@@ -62,6 +62,54 @@ spec:
...
+      command: [echo]
+      args: ["hello world"]
+    archiveLocation:
+      archiveLogs: true
+      archiveSystemContainerLogs: true
+```
+
+### Limitations
+
+* **`init` failures are not archived.**
```
tico24: These all name `init`/`wait`, so I read the section as legacy-only, but the feature note says init-less does archive `supervisor-logs` when artifact loading fails. Worth saying which layout this applies to?

docs/workflow-controller-configmap.yaml
```diff
@@ -115,6 +115,9 @@ data:
   artifactRepository: |
     # archiveLogs will archive the main container logs as an artifact
     archiveLogs: true
+    # archiveSystemContainerLogs will also archive the logs of Argo's system containers
+    # (init and wait, or supervisor in the init-less pod layout)
+    archiveSystemContainerLogs: true
```
tico24: configure-archive-logs.md says you only pay for these when you need them, but setting it here turns it on for every workflow and there's no way to switch it back off. Should the example be `false`?

tico24 (commented):
One sentence per line is sorted across the new section. Two questions below. Docs review only, the code is for @argoproj/argo-workflows-approvers.

thc1006:
Thanks for the mention. I checked the latest head again from the #16587 side. `archiveSystemContainerLogs` uses WorkflowSpec field 46, and `resourceClaims` remains at 47, so there is no protobuf conflict.

I also checked the two docs questions against the current behavior. I agree with both:

- The `init` failure limitation is specific to the legacy layout. The other `wait` limitations should also mention `supervisor` for init-less Pods, where `supervisor-logs` can still be archived after an artifact-loading failure.
- `archiveSystemContainerLogs: false` is safer in the controller config example. With the controller-level value set to `true`, workflow/template `false` cannot disable it.

One small docs mismatch I noticed: the priority should be `controller(on) > template > workflow > controller(off)`, matching `IsArchiveSystemContainerLogs`.

This is only a cross-PR compatibility check from my side; the author and approvers can decide the final wording.

docs/configure-archive-logs.md
myzk-a: Good catch — that section was written for the legacy layout and never revisited after init-less landed. Fixed: each limitation now names the layout it applies to, and the `init` failure bullet says explicitly that init-less is not affected, because the `supervisor` container still archives `supervisor-logs` after an artifact-loading failure. I also qualified the earlier mention of the `init` limitation in the body, which had the same problem, and retitled the section from "Archiving Init and Wait Container Logs" to "Archiving System Container Logs" — the heading itself was steering you toward the legacy-only reading, while the field name is deliberately layout-neutral.

docs/workflow-controller-configmap.yaml
myzk-a: Changed to `false`, with a comment spelling out that enabling it here cannot be undone per workflow. I had mirrored the adjacent `archiveLogs: true`, but you're right that the two aren't equivalent: `archiveLogs` covers the logs users actually want, whereas these are debug-only, so shipping the example with it on imposes storage cost on everyone who copies the block.

myzk-a:
Pushed docs-only fixes for all three points — no behavior change.

While fixing the precedence wording I found that the existing line for `archiveLogs` at the top of the page contradicts the truth table directly below it. The line reads `workflow-controller config (on) > workflow spec (on/off) > template (on/off)`, but the table's `false | true | false → false` row has the template value winning over the workflow spec, which is what `IsArchiveLogs` does and what `TestConditionalAddArchiveLocationTemplateArchiveLogs` pins. The line has been there since 2021 and my `archiveSystemContainerLogs` paragraph had copied its wording, so mine was wrong in the same way but without a table to correct it. I've fixed the line to match the table and dropped my duplicate of it, leaving one statement of the ordering for both fields. Happy to split that one-line change into its own PR if you'd rather keep this one narrowly scoped.

Thanks @thc1006 for re-checking field 46 from the #16587 side and for catching the precedence mismatch.

tico24 (commented):
Docs only again, rest is for @argoproj/argo-workflows-approvers. Both my earlier points are sorted, and the `archiveLogs` priority line you fixed now matches `IsArchiveLogs` and the table under it, so I wouldn't bother splitting that into its own PR. Two things left on the feature note.

.features/pending/archive-system-container-logs.md
```diff
@@ -0,0 +1,17 @@
+Description: Archive the `init` and `wait` container logs, not just the `main` container
```
tico24: This one's still init/wait only, and `Description` is what ends up as the release-note headline. Same reason you retitled the docs section, I think?

.features/pending/archive-system-container-logs.md
```diff
@@ -0,0 +1,17 @@
...
+Under the opt-in init-less pod layout, the single `supervisor` container performs both roles, so its log is archived as one `supervisor-logs` artifact instead.
+This is useful for inspecting what the system containers did after the Pod is gone, such as debugging output or artifact upload failures in `wait`, or reviewing which input artifacts the `init` container loaded.
+
+`archiveSystemContainerLogs` is controlled separately from `archiveLogs`, so you only pay to store these extra logs when you need them.
+It can be set at the workflow-controller config-map, workflow spec, or template level, following the same priorities as `archiveLogs`.
+
+In the legacy pod layout a *failing* `init` container is not captured, because the `wait` container never starts to archive it; the init-less layout does capture the `supervisor` log even when loading input artifacts fails.
+See [Configuring Archive Logs](https://argo-workflows.readthedocs.io/en/latest/configure-archive-logs/) for the example manifest and the full list of limitations.
```
tico24: These notes get pasted into `docs/new-features.md` as-is, so `/en/latest/` will point at whatever's current rather than the release this belongs to. `configurable-node-status-compression.md` uses a relative link if you want the pattern.

.features/pending/archive-system-container-logs.md
myzk-a: Yes, exactly the same reason. `Description` is what `hack/featuregen/contents.go` emits as the `- ...` bullet in `docs/new-features.md`, so the legacy-only framing would have become the release-note headline. Changed it to "Archive the logs of Argo's system containers, not just the `main` container", which matches the field name and the retitled docs section. I left the `init`/`wait` enumeration in the body, where there is room for the following line to name `supervisor` for the init-less layout.

.features/pending/archive-system-container-logs.md
myzk-a: Good catch — switched to `configure-archive-logs.md#archiving-system-container-logs`, following the `configurable-node-status-compression.md` pattern. Anchored to the section itself, since that's where the example manifest and the limitations list are.

myzk-a:
Both feature-note points are fixed, and thanks for the call on the precedence line — leaving that in this PR then. I also checked the rest of the docs surface for the same legacy-only framing; the generated files all inherit the `pkg/apis` field comments, which already name `supervisor`.

One change you didn't ask for, in the same file: the note repeated the "you only pay to store these extra logs when you need them" line without the caveat you raised on the config-map example, so I added the same sentence I used in `configure-archive-logs.md`. Better than letting it reach the release notes.

Merging `main` picked up #16224, so `Windows Unit Tests` is green again. The one red check left is `test-examples` on `artifact-passing-explicit-plugin.yaml`, where the `generate-artifact` Pod sat in `PodInitializing` until the 90s wait timeout rather than failing an assertion. Looks like runner contention: the `initless` cell of the same suite passed on this commit, and both cells passed at `f5c074db2`, which has byte-identical Go code.

PR #16274, opened 2026-06-15
fix: honor ?? and ?. guards in strict missing-variable check

### LINKED ISSUE #16276: Strict "is missing" check rejects `??` / `?.`-guarded member paths
### Pre-requisites

In `withParam` / `withItems` (and other strict) contexts, the static pre-flight identifier check rejects an expression that references a map key absent from the item — **even when the reference is explicitly guarded by the nil-coalescing operator `??` or optional chaining `?.`**, whose entire purpose is to tolerate the missing value.

A loop over heterogeneous items, where some omit an optional key, fails before evaluation with:

```
failed to evaluate expression: item.optionalKey is missing
```

**Expected:** `item.optionalKey ?? 'fallback'` should resolve the missing key to `fallback`, matching expr-lang's own runtime semantics (verified against the vendored `expr-lang` v1.17.8: the guarded expression evaluates with no error and returns the fallback).

**Root cause:** the check (`getIdentifiers` → `identifierVisitor` → `getMemberPath`, introduced in `d84169f` / #15442) is purely structural. It walks the AST and extracts every `item.<key>` member path with no awareness that the path is guarded by `??` / `?.`, so a legitimately-optional key is rejected exactly like a genuinely-undefined variable. This is distinct from #15839, which only fixed the non-strict (`allowUnresolved=true`) path.

### Version(s)

v3.7.11 up to :latest

Sample workflow:
```yaml
kind: Workflow
metadata:
  generateName: nil-coalesce-repro-
  namespace: argo
spec:
  entrypoint: main
  templates:
    - name: main
      steps:
        - - name: echo
            template: echo
            withItems:
              - { name: "a", optionalKey: "value" }
              - { name: "b" }
            arguments:
              parameters:
                - name: msg
                  value: "{{= item.optionalKey ?? 'fallback' }}"
    - name: echo
      inputs:
        parameters:
          - name: msg
      container:
        image: alpine:3.20
        command: [echo]
        args: ["{{inputs.parameters.msg}}"]
```

## fail

setup any version, from `v3.5.11 up to latest`. Run the wf, you will get:
`Message: failed to evaluate expression: item.optionalKey is missing`

Image

Image

Image

## fix

build from https://github.com/argoproj/argo-workflows/pull/16274, re-execute the same workflow:

Image

Image

Image

## interim workaround

Reference the key via the `get()` builtin, so no `item.optionalKey` member path is produced:

```
{{= get(item, 'optionalKey') ?? 'fallback' }}
```

```text

level=ERROR msg="marking node as error" workflow=nil-coalesce-repro-drv96 namespace=argo error="failed to evaluate expression: item.optionalKey is missing" nodeName=nil-coalesce-repro-drv96[0]
level=INFO  msg="step group ... was unsuccessful: failed to evaluate expression: item.optionalKey is missing" workflow=nil-coalesce-repro-drv96 namespace=argo
level=INFO  msg="updated phase" workflow=nil-coalesce-repro-drv96 namespace=argo fromPhase=Running toPhase=Failed
```

```text

# N/A — the workflow fails during withItems/withParam expansion in the controller,
# before any pod (and therefore any wait container) is created. See the empty Logs view above.
```

zvdy:
Fixes https://github.com/argoproj/argo-workflows/issues/16276

### Motivation

Commit 9f32c38 introduced a strict missing-variable check in `ReplaceStrict`. Before evaluating an expression template, it walks the AST, collects every referenced identifier, and requeues the workflow if any of them are not in scope yet. That is correct behavior for plain member paths like `tasks.a.outputs.result` — if the variable is not there yet, we should wait.

The problem is that the check did not account for `??` (nil-coalescing) and `?.` (optional chaining). An expression like:

```
{{= tasks.a.outputs.result ?? 'default' }}
```

is designed to handle a missing value gracefully, but the pre-check still treated `tasks.a.outputs.result` as strictly required and triggered a requeue, even though the `??` fallback would have handled it just fine at evaluation time.

### Reproduction

take any DAG where a downstream task optionally consumes the output of a sibling using `??`:

```yaml
kind: Workflow
metadata:
  generateName: nil-coalescing-bug-
spec:
  entrypoint: main
  templates:
    - name: main
      dag:
        tasks:
          - name: produce
            template: echo
            arguments:
              parameters:
                - name: message
                  value: "hello"
          - name: consume
            template: echo
            arguments:
              parameters:
                - name: message
                  value: "{{= tasks['produce'].outputs.result ?? 'fallback' }}"
```

Even though `tasks['produce'].outputs.result` is guarded by `??`, the workflow gets stuck requeuing and never reaches the `consume` task. The error you see in the controller logs is:

```
failed to evaluate expression: tasks.produce.outputs.result is missing
```

Note the identifier is reported in dot notation (`tasks.produce.outputs.result`) because `getMemberPath` normalizes bracket notation when building the required-identifier set.

### expr-lang 

image

What happens under the hood: [expr-lang](https://github.com/expr-lang/expr) parses the expression into an AST where `tasks['produce'].outputs.result` becomes a chain of `MemberNode`s and `??` becomes a `BinaryNode` with operator `??`. The strict check in `ReplaceStrict` walks that AST with `identifierVisitor`, which collected every `MemberNode` path as a required identifier without looking at what operator surrounded it. So `tasks.produce.outputs.result` was added to the required set, the check found it missing, and triggered a requeue — completely ignoring the `??` that was meant to handle exactly that case.

Worth noting: this is a different issue from #15839, which fixed runtime eval errors in the `allowUnresolved=true` path (`SubstituteParams`). This PR fixes the strict identifier pre-check path used by `resolveDependencyReferences`.

### Modifications

Added a `guardVisitor` / `memberMarker` pre-pass that walks the AST before identifier collection and marks any member node sitting under a `??` or `?.` operator as guarded. The existing `identifierVisitor` now skips those nodes so they are not counted as strictly required. `getMemberPath` also bails out early on optional-chaining members for the same reason.

The base variable of a guarded path (e.g. `tasks` in `tasks.a.outputs.result ?? 'default'`) is intentionally kept in the required set. If the whole variable does not exist yet, the requeue behavior is still correct.

### Verification

Added two new test functions:

`Test_getIdentifiers` covers plain required members, `??`-guarded paths, bracket notation (`item['key'] ?? ''`), optional chaining (`item?.key`, `item.a?.b`), and mixed expressions where some paths are required and others are guarded.

`Test_ReplaceStrict_NilCoalescing` runs `ReplaceStrict` end-to-end with a present and a missing optional key for each operator variant, plus a case confirming that a missing base variable still correctly triggers the strict check.

All existing `util/template` tests continue to pass.

### Documentation

not needed

### AI

Claude Opus helped with the investigation, implementation and tests.

### DIFF
--- a/util/template/expression_template.go
+++ b/util/template/expression_template.go
@@ -103,9 +103,13 @@ func expressionReplaceStrict(ctx context.Context, w io.Writer, expression string
 type identifierVisitor struct {
 	identifiers []string
 	seen        map[string]bool
+	guarded     map[ast.Node]bool
 }
 
 func (v *identifierVisitor) Visit(node *ast.Node) {
+	if v.guarded[*node] {
+		return
+	}
 	if n, ok := (*node).(*ast.IdentifierNode); ok {
 		if !v.seen[n.Value] {
 			v.identifiers = append(v.identifiers, n.Value)
@@ -127,6 +131,9 @@ func getMemberPath(node *ast.MemberNode) (string, bool) {
 	var parts []string
 	curr := node
 	for {
+		if curr.Optional {
+			return "", false
+		}
 		prop, ok := curr.Property.(*ast.StringNode)
 		if !ok {
 			return "", false
@@ -146,13 +153,49 @@ func getMemberPath(node *ast.MemberNode) (string, bool) {
 	}
 }
 
+// guardVisitor collects the member nodes whose access is guarded by the
+// nil-coalescing (??) or optional-chaining (?.) operators, so they are not
+// reported as strictly-required identifiers. Base identifiers are left
+// untouched so a genuinely-unavailable variable still triggers a requeue.
+type guardVisitor struct {
+	guarded map[ast.Node]bool
+}
+
+func (v *guardVisitor) Visit(node *ast.Node) {
+	switch n := (*node).(type) {
+	case *ast.BinaryNode:
+		if n.Operator == "??" {
+			ast.Walk(&n.Left, &memberMarker{guarded: v.guarded})
+		}
+	case *ast.MemberNode:
+		if n.Optional {
+			if _, ok := n.Node.(*ast.MemberNode); ok {
+				v.guarded[n.Node] = true
+			}
+		}
+	}
+}
+
+type memberMarker struct {
+	guarded map[ast.Node]bool
+}
+
+func (m *memberMarker) Visit(node *ast.Node) {
+	if _, ok := (*node).(*ast.MemberNode); ok {
+		m.guarded[*node] = true
+	}
+}
+
 func getIdentifiers(expression string) ([]string, error) {
 	tree, err := parser.Parse(expression)
 	if err != nil {
 		return nil, err
 	}
+	guarded := make(map[ast.Node]bool)
+	ast.Walk(&tree.Node, &guardVisitor{guarded: guarded})
 	visitor := &identifierVisitor{
-		seen: make(map[string]bool),
+		seen:    make(map[string]bool),
+		guarded: guarded,
 	}
 	ast.Walk(&tree.Node, visitor)
 	return visitor.identifiers, nil
--- a/util/template/expression_template_test.go
+++ b/util/template/expression_template_test.go
@@ -99,6 +99,57 @@ func Test_hasVarInEnv(t *testing.T) {
 	})
 }
 
+func Test_getIdentifiers(t *testing.T) {
+	tests := []struct {
+		name       string
+		expression string
+		want       []string
+	}{
+		{
+			name:       "plain member path is required",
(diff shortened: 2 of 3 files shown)


isubasinghe:
@zvdy can you please rebase

zvdy:
@isubasinghe done, let me know if anything else is missing, I would like this to be backported if possible from `v3.7.11` upwards

zvdy:
@isubasinghe ci failure is unrelated, the rebase introduced it.

```
└ ✖ workflow-template-containerset Pod       3s      c2: Error (exit code 64): failed to find name in PATH: exec: "invalid": executable file not found in $PATH

```

zvdy:
@isubasinghe take a look

isubasinghe (approved):

PR #16275, opened 2026-06-15
fix: prevent NaN display in cron workflow number input fields

nakatani-yo:
### Motivation

When creating or editing a CronWorkflow in the UI, the number input fields (startingDeadlineSeconds, successfulJobsHistoryLimit, failedJobsHistoryLimit) display "NaN" when their values are undefined (specifically, when a non-numeric key is pressed in the input field). This occurs because `parseInt(undefined, 10)` returns `NaN`, which is then rendered as a string in the input field.

Additionally, once "NaN" is displayed in the input field, it cannot be removed using the Backspace key (Windows) or Delete key (Mac), which is inconvenient for users.

image

### Modifications

- Updated `NumberInput` component (`ui/src/shared/components/number-input.tsx`) to guard against `NaN` values:
  - Added an `isNaN()` check when converting the value to a string for display, so undefined/NaN values render as an empty string instead of "NaN".
  - Changed the `onChange` callback to return `undefined` instead of `NaN` when the input is cleared or contains an invalid number.
- Added unit tests for `CronWorkflowSpecEditor` (`ui/src/cron-workflows/cron-workflow-spec-editor.test.tsx`) to verify:
  - No NaN is displayed when numeric fields are undefined.
  - Numeric values are displayed correctly when set.
  - Clearing a numeric field calls onChange with `undefined`.
  - Entering a valid number calls onChange with the parsed number.

### Verification

- Added unit tests covering the NaN display bug and correct numeric input behavior.
- Manually verified in the UI that CronWorkflow number input fields no longer show "NaN" when values are not set.

### Documentation

No documentation changes needed. This is a bug fix for existing UI behavior.

### AI

Codex

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric input parsing and display so invalid or cleared values no longer show `NaN`.
  * The editor now treats cleared/invalid numeric entries as `undefined`, sending `undefined` to change handlers and rendering an empty field.
  * Valid numeric values still display and edit correctly.
* **Tests**
  * Added React Testing Library coverage for the cron workflow spec editor to verify `NaN` never renders when numeric fields (including `startingDeadlineSeconds`) are omitted or cleared.
  * Verified that entering, clearing, and `undefined` numeric states behave as expected.

### DIFF
--- a/ui/src/cron-workflows/cron-workflow-spec-editor.test.tsx
+++ b/ui/src/cron-workflows/cron-workflow-spec-editor.test.tsx
@@ -1,0 +1,78 @@
+import {fireEvent, render, screen} from '@testing-library/react';
+import React from 'react';
+
+import {CronWorkflowSpec} from '../shared/models';
+import {CronWorkflowSpecEditor} from './cron-workflow-spec-editor';
+
+const baseSpec: CronWorkflowSpec = {
+    schedules: ['* * * * *'],
+    workflowSpec: {
+        entrypoint: 'main',
+        templates: [{name: 'main', container: {name: 'main', image: 'argoproj/argosay:v2', command: ['cowsay']}}]
+    }
+};
+
+describe('CronWorkflowSpecEditor', () => {
+    it('does not display NaN when startingDeadlineSeconds is undefined', () => {
+        render(<CronWorkflowSpecEditor spec={baseSpec} onChange={() => undefined} />);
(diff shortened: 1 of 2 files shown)


Joibel:
@coderabbitai review

nakatani-yo:
@Joibel 

Thank you for your running `coderabbitai review`.
I plan to make revisions based on the review results.

isubasinghe (approved):

PR #16277, opened 2026-06-16
feat: add cronWorkflowDefaults to controller config. Fixes #12627

### LINKED ISSUE #12627: Add `cronWorkflowDefaults`
# Summary

What change needs making?

There is a [workflow default](https://argo-workflows.readthedocs.io/en/latest/default-workflow-specs/), but not one for crons.

## Use Cases

When would you use this?

Setting `CronWorkflow history limits` mentioned in the [Cost Optimization doc](https://argo-workflows.readthedocs.io/en/latest/cost-optimisation/#limit-the-total-number-of-workflows-and-pods) cannot be set globally. It would be nice to have this option similar to the `Workflow Defaults`.

---

alien2003:
Fixes #12627

### Motivation

`workflowDefaults` lets an operator set controller-wide defaults for `Workflows`, but there is no equivalent for `CronWorkflows`. Fields like `successfulJobsHistoryLimit`, `failedJobsHistoryLimit`, `concurrencyPolicy` and `timezone` have to be repeated on every `CronWorkflow`, or enforced with an external policy engine. This adds a `cronWorkflowDefaults` to set them once.

### Modifications

Add a `cronWorkflowDefaults` key to the `workflow-controller-configmap` (`config.Config.CronWorkflowDefaults *wfv1.CronWorkflowSpec`).

The defaults are merged into each `CronWorkflow` during reconciliation by a new `util.MergeToCronWorkflowSpec`, which mirrors `util.MergeTo`: a strategic merge that lets the `CronWorkflow`'s own values win over the defaults. The merge runs on the controller's in-memory copy in `newCronWfOperationCtx`, before validation, scheduling and runtime-policy checks, so every read sees the defaulted spec.

The defaults are not written back to the resource. The cron reconcile loop only persists `status` and `metadata`, never `spec`, so this keeps defaults out of the stored object and avoids fighting GitOps controllers that own the `CronWorkflow` manifest (the concern @anarsen raised on the issue). The shared defaults are deep-copied per reconcile because the merge temporarily mutates its patch argument and workers run concurrently.

### Verification

`go test ./config/... ./workflow/util/... ./workflow/cron/...`

- `TestMergeToCronWorkflowSpec`: default fills unset field, target overrides default, nil patch is a no-op, the shared patch is not mutated.
- `TestApplyCronWorkflowDefaults`: defaults fill unset fields on the in-memory `CronWorkflow`, an explicit `CronWorkflow` value is preserved, and the `CronWorkflow`'s own `schedules` survive the merge.

### Documentation

`docs/default-workflow-specs.md` gets a "Setting Default CronWorkflow Values" section; the example `docs/workflow-controller-configmap.yaml` gets the new key; `docs/workflow-controller-configmap.md` is regenerated. Feature file added under `.features/pending/`.

### AI

AI-assisted (Claude Code). I scoped the design (in-memory application vs writing back to the resource, and the config type), reviewed and tested the implementation, and ran the builds, tests and linter in a pinned container. The decision to apply defaults in-memory rather than mutate the resource follows the GitOps concern raised by @anarsen on the issue.


Joibel:
I'm not sure we should include workflowSpec in the merged result, I think it should be banned from the cronWorkflowDefaults. What do you think @isubasinghe?

My reasoning is that we've then got two ways to set workflow defaults for crons, and that's confusing and error prone. I don't want to support fixing that when it's surprising.

workflow/cron/operator.go
```diff
@@ -46,6 +46,7 @@ type cronWfOperationCtx struct {
 	wfClientset     versioned.Interface
 	wfClient        typed.WorkflowInterface
 	wfDefaults      *v1alpha1.Workflow
+	cronWfDefaults  *v1alpha1.CronWorkflowSpec
```
Joibel, line 49: **Dead struct field (confidence 80).** `cronWfDefaults` is stored here and assigned in `newCronWfOperationCtx` (line 87), but never read again — defaults are applied eagerly in place by `applyCronWorkflowDefaults`. Contrast `wfDefaults`, which *is* read later (at submit/validate). The constructor only needs the parameter; this field can be dropped.

workflow/util/merge.go
```diff
@@ -191,6 +191,70 @@ func MergeTo(patch, target *wfv1.Workflow) error {
 	return nil
 }
 
+// MergeToCronWorkflowSpec merges the "patch" CronWorkflowSpec into the "target" CronWorkflowSpec.
+// If the target defines a field, that value takes precedence over the patch. This mirrors MergeTo,
+// including the manual handling of the embedded WorkflowSpec's hooks and labelsFrom, which strategic
+// merge does not handle.
+func MergeToCronWorkflowSpec(patch, target *wfv1.CronWorkflowSpec) error {
```
Joibel, line 198: **A non-zero default can't be overridden back to its zero value (confidence 75).** The CronWorkflow is applied as the strategic-merge patch over the defaults, so `omitempty` value-typed fields can't be reset. Sharpest case: `CronWorkflowSpec.Suspend` is a plain `bool`, so `cronWorkflowDefaults: {suspend: true}` would force-suspend every CronWorkflow with no opt-out. The Workflow path explicitly compensates in `JoinWorkflowSpec` (and uses `*bool`); there's no equivalent here. Also affects `when`/`timezone`.

workflow/cron/operator.go
```diff
@@ -57,16 +58,33 @@ type cronWfOperationCtx struct {
...
+// working copy: the reconcile loop never persists the spec, so defaults don't cause GitOps drift.
+// The defaults are deep-copied because MergeToCronWorkflowSpec temporarily mutates its patch argument
+// and cronWfDefaults is shared across concurrent workers.
+func applyCronWorkflowDefaults(ctx context.Context, cronWf *v1alpha1.CronWorkflow, cronWfDefaults *v1alpha1.CronWorkflowSpec) {
+	if cronWfDefaults == nil {
+		return
+	}
+	if err := util.MergeToCronWorkflowSpec(cronWfDefaults.DeepCopy(), &cronWf.Spec); err != nil {
+		logging.RequireLoggerFromContext(ctx).WithError(err).Error(ctx, "Failed to apply CronWorkflow defaults")
+	}
```
Joibel, line 72: **Merge error is swallowed, and the merge can wipe the spec (confidence 75).** On error this only logs and continues. Worse, `MergeToCronWorkflowSpec` does `*target = wfv1.CronWorkflowSpec{}` *before* its final unmarshal (merge.go:231), so an unmarshal failure leaves `cronWf.Spec` zeroed and reconciliation proceeds on an empty spec. The analogous `setWorkflowDefaults`/`MergeTo` returns the error to the caller instead. Rare in practice, but weaker than the pattern it copies.

docs/default-workflow-specs.md
```diff
@@ -55,3 +55,29 @@ data:
...
+## Setting Default CronWorkflow Values
+
+> v4.1 and after
+
+Default `CronWorkflow` spec values can be specified under the `cronWorkflowDefaults` key in the `workflow-controller-configmap`.
+These apply to all `CronWorkflows` executed from the controller, such as `concurrencyPolicy`, `startingDeadlineSeconds`, `successfulJobsHistoryLimit`, `failedJobsHistoryLimit` and `timezone`.
+If a `CronWorkflow` sets a value that also has a default, the `CronWorkflow`'s value takes precedence.
+See the [Field Reference](./fields.md#cronworkflowspec) for the full list of `CronWorkflowSpec` fields.
+
+Defaults are applied to the controller's in-memory copy during reconciliation and are never written back to the `CronWorkflow`, so they do not conflict with GitOps tools that manage the resource.
```
Joibel, line 68: **This "never written back" claim isn't strictly true for `timezone` (confidence 72).** A defaulted `Spec.Timezone` flows into `SetSchedule(...GetScheduleWithTimezoneString())` and the resulting `annotationKeyLatestSchedule` annotation, which `persistUpdate` patches into `metadata.annotations`. So a defaulted value does reach the stored object (a controller-managed annotation, not the spec). Worth qualifying the claim.

workflow/controller/controller.go
```diff
@@ -278,7 +278,7 @@ func (wfc *WorkflowController) runPodController(ctx context.Context, podGCWorker
 func (wfc *WorkflowController) runCronController(ctx context.Context, cronWorkflowWorkers int) {
 	defer runtimeutil.HandleCrashWithContext(ctx, runtimeutil.PanicHandlers...)
 
-	cronController := cron.NewCronController(ctx, wfc.wfclientset, wfc.dynamicInterface, wfc.namespace, wfc.GetManagedNamespace(), wfc.Config.InstanceID, wfc.metrics, wfc.eventRecorderManager, cronWorkflowWorkers, wfc.wftmplInformer, wfc.cwftmplInformer, wfc.Config.WorkflowDefaults)
+	cronController := cron.NewCronController(ctx, wfc.wfclientset, wfc.dynamicInterface, wfc.namespace, wfc.GetManagedNamespace(), wfc.Config.InstanceID, wfc.metrics, wfc.eventRecorderManager, cronWorkflowWorkers, wfc.wftmplInformer, wfc.cwftmplInformer, wfc.Config.WorkflowDefaults, wfc.Config.CronWorkflowDefaults)
```
Joibel, line 281: **Defaults are controller-only; not applied on the API-server / `argo cron lint` path (confidence 72).** `workflowDefaults` is also wired into validate/submit on the server, but `cronWorkflowDefaults` is reconcile-only — so a CronWorkflow that only validates *because* of a default passes at the controller yet can't be linted with defaults via the API. Likely deliberate (GitOps-drift avoidance), but an undocumented behavioral asymmetry worth calling out.

workflow/util/merge_test.go
```diff
@@ -813,3 +813,55 @@ func TestAllWorkflowSpecFieldsAccountedFor(t *testing.T) {
 		}
 	}
 }
+
+func TestMergeToCronWorkflowSpec(t *testing.T) {
```
Joibel, line 817: **The bug-prone hooks/labelsFrom path has no coverage (confidence 75).** The hand-rolled Hooks save/restore and `WorkflowMetadata.LabelsFrom` re-merge are the whole reason a plain strategic merge won't work, yet this test asserts nothing about either. `MergeTo` has dedicated `TestMergeHooks`/`TestMergeLabelsFrom`; please add cron equivalents.

workflow/util/merge_test.go
```diff
@@ -813,3 +813,55 @@ func TestAllWorkflowSpecFieldsAccountedFor(t *testing.T) {
...
+		assert.Equal(t, wfv1.ForbidConcurrent, target.ConcurrencyPolicy)
+	})
+
+	t.Run("PatchIsNotMutated", func(t *testing.T) {
+		defaults := &wfv1.CronWorkflowSpec{Timezone: "America/New_York"}
+		target := &wfv1.CronWorkflowSpec{Timezone: "Asia/Tokyo"}
+		require.NoError(t, MergeToCronWorkflowSpec(defaults, target))
+		assert.Equal(t, "America/New_York", defaults.Timezone)
+		assert.Equal(t, "Asia/Tokyo", target.Timezone)
+	})
```
Joibel, line 866: **`PatchIsNotMutated` doesn't test what its name claims (confidence 75).** It only sets `Timezone`, which the save/restore dance never touches — this subtest would pass even if the entire restore block were deleted. To actually verify it, the patch must contain `Hooks` and `WorkflowMetadata.LabelsFrom` and assert they survive the call.

workflow/util/merge.go
```diff
@@ -191,6 +191,70 @@ func MergeTo(patch, target *wfv1.Workflow) error {
...
+
+	if len(patchHooks) != 0 && target.WorkflowSpec.Hooks == nil {
+		target.WorkflowSpec.Hooks = make(wfv1.LifecycleHooks)
+	}
+	for name, hook := range patchHooks {
+		// If the patch hook doesn't exist in target
+		if _, ok := target.WorkflowSpec.Hooks[name]; !ok {
+			target.WorkflowSpec.Hooks[name] = hook
+		}
+	}
```
Joibel, line 244: **No field-accounting guard protects this manual handling (confidence 70).** `MergeTo` is backed by `TestAllWorkflowSpecFieldsAccountedFor`, which forces a conscious decision whenever `WorkflowSpec` gains a field. This duplicated manual hooks/labelsFrom handling has no equivalent guard, so a future `WorkflowSpec`/`CronWorkflowSpec` field needing special handling could silently break on the cron path.

workflow/util/merge_test.go
```diff
@@ -813,3 +813,55 @@ func TestAllWorkflowSpecFieldsAccountedFor(t *testing.T) {
...
+		}
+		target := &wfv1.CronWorkflowSpec{
+			ConcurrencyPolicy:       wfv1.ForbidConcurrent,
+			StartingDeadlineSeconds: new(int64(10)),
+			Schedules:               []string{"* * * * *"},
+		}
+		require.NoError(t, MergeToCronWorkflowSpec(defaults, target))
+		assert.Equal(t, wfv1.ForbidConcurrent, target.ConcurrencyPolicy)
+		assert.Equal(t, int64(10), *target.StartingDeadlineSeconds)
+	})
```
Joibel, line 852: **Missing the `*int32` 0-vs-unset boundary (confidence 70).** `successfulJobsHistoryLimit`/`failedJobsHistoryLimit` are `*int32` where an explicit `0` is meaningful and distinct from unset. Please add a case where the default is `4` and the CronWorkflow explicitly sets `0`, asserting the explicit `0` wins — it's the headline use case in the docs.

workflow/cron/operator_test.go
```diff
@@ -815,3 +815,46 @@ func TestEvaluateWhenUnresolvedOutside(t *testing.T) {
...
 }
+
+func TestApplyCronWorkflowDefaults(t *testing.T) {
+	ctx := logging.TestContext(t.Context())
+
+	newCronWf := func() *v1alpha1.CronWorkflow {
+		cronWf := &v1alpha1.CronWorkflow{}
+		v1alpha1.MustUnmarshal([]byte(scheduledWf), cronWf)
+		return cronWf
+	}
```
Joibel, line 826: **The DeepCopy safety property is untested at this layer (confidence 60).** `applyCronWorkflowDefaults` deep-copies precisely because the shared `cronWfDefaults` is mutated transiently during the merge. No case here asserts that the shared `defaults` object is left unmutated (especially when it contains Hooks/LabelsFrom). A future refactor dropping the `DeepCopy()` wouldn't be caught.

Joibel (commented):
### Code review

Re-posting as inline comments. Nice, well-scoped feature mirroring `workflowDefaults`. I verified several things are deliberately correct and **not** issues: the strategic-merge direction (CronWorkflow wins), the `DeepCopy()` for concurrency safety, no informer-cache corruption (both call sites build fresh objects from unstructured), the `LabelsFrom` deref is safely guarded just like `MergeTo`, and `new(int64(60))` in tests is valid Go 1.26. No CLAUDE.md compliance issues.

14 findings inline below (confidence in parentheses; ~70+ are worth acting on, the rest are nits).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

docs/default-workflow-specs.md
```diff
@@ -55,3 +55,29 @@ data:
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: workflow-controller-configmap
+data:
+  cronWorkflowDefaults: |
+    concurrencyPolicy: "Replace"
+    successfulJobsHistoryLimit: 4
+    failedJobsHistoryLimit: 4
+    timezone: "America/New_York"
```
Joibel, line 82: **The example sets more than the prose explains (confidence 70).** The lead-in says only "to keep at most four successful and four failed child Workflows", but the YAML also sets `concurrencyPolicy: "Replace"` and `timezone: "America/New_York"`. Either trim the example to match, or broaden the lead-in.

config/config.go
```diff
@@ -86,6 +86,9 @@ type Config struct {
 	// WorkflowDefaults are values that will apply to all Workflows from this controller, unless overridden on the Workflow-level
 	WorkflowDefaults *wfv1.Workflow `json:"workflowDefaults,omitempty"`
 
+	// CronWorkflowDefaults are values that will apply to all CronWorkflows from this controller, unless overridden on the CronWorkflow-level
+	CronWorkflowDefaults *wfv1.CronWorkflowSpec `json:"cronWorkflowDefaults,omitempty"`
```
Joibel, line 90: **Undocumented two-layer defaulting (confidence 55).** Because the type is `CronWorkflowSpec` (which embeds `WorkflowSpec`), `cronWorkflowDefaults` can also set arbitrary workflow-spec fields (entrypoint, templates, etc.), which then re-merge with `workflowDefaults` on the spawned Workflow. The docs only advertise cron-level fields and don't explain this overlap or its precedence.

workflow/util/merge.go
```diff
@@ -191,6 +191,70 @@ func MergeTo(patch, target *wfv1.Workflow) error {
...
+	if patch.WorkflowSpec.WorkflowMetadata != nil {
+		patchLabelsFrom = patch.WorkflowSpec.WorkflowMetadata.LabelsFrom
+		patch.WorkflowSpec.WorkflowMetadata.LabelsFrom = nil
+	}
+
+	patchBytes, err := json.Marshal(patch)
+	patch.WorkflowSpec.Hooks = patchHooks
+	if len(patchLabelsFrom) != 0 {
+		patch.WorkflowSpec.WorkflowMetadata.LabelsFrom = patchLabelsFrom
+	}
```
Joibel, line 216: **DRY: this is a near-verbatim copy of `MergeTo` (confidence 50).** The subtle hooks/labelsFrom "remove-before-marshal, restore-after, re-add-to-target" dance is duplicated; a fix to one copy won't propagate to the other, and nothing enforces parity. Consider a single generic helper parameterized over the embedded `WorkflowSpec`. Maintainability only.

docs/default-workflow-specs.md
```diff
@@ -55,3 +55,29 @@ data:
       parallelism: 3
 
 ```
+
+## Setting Default CronWorkflow Values
+
+> v4.1 and after
+
+Default `CronWorkflow` spec values can be specified under the `cronWorkflowDefaults` key in the `workflow-controller-configmap`.
+These apply to all `CronWorkflows` executed from the controller, such as `concurrencyPolicy`, `startingDeadlineSeconds`, `successfulJobsHistoryLimit`, `failedJobsHistoryLimit` and `timezone`.
```
Joibel, line 64: **Dangling "such as" clause (confidence 50).** "These apply to all `CronWorkflows` executed from the controller, such as `concurrencyPolicy`, ..." reads as if the CronWorkflows are "such as concurrencyPolicy". It should attach to the default *fields*, e.g. "Default values can be set for fields such as ...".

config/config.go
Joibel: I've made a separate comment - but my feeling is we must disallow this.

docs/default-workflow-specs.md
```diff
@@ -55,3 +55,29 @@ data:
       parallelism: 3
 
 ```
+
+## Setting Default CronWorkflow Values
+
+> v4.1 and after
+
+Default `CronWorkflow` spec values can be specified under the `cronWorkflowDefaults` key in the `workflow-controller-configmap`.
```
tico24, line 63: When would I use this instead of `workflowDefaults`? That's the thing I'd want the section to answer, especially as `workflowSpec` is itself a `CronWorkflowSpec` field, so both keys can set the same thing and it's not obvious which one wins. [docs guide](https://argo-workflows.readthedocs.io/en/latest/doc-changes/)

docs/default-workflow-specs.md
tico24, line 64: The "such as" list hangs off CronWorkflows rather than off the values, so it reads like `concurrencyPolicy` is a kind of CronWorkflow. Two sentences would fix it.

docs/default-workflow-specs.md
```diff
@@ -55,3 +55,29 @@ data:
...
+These apply to all `CronWorkflows` executed from the controller, such as `concurrencyPolicy`, `startingDeadlineSeconds`, `successfulJobsHistoryLimit`, `failedJobsHistoryLimit` and `timezone`.
+If a `CronWorkflow` sets a value that also has a default, the `CronWorkflow`'s value takes precedence.
```
tico24, line 65: Do these need back-ticks? CronWorkflow and CronWorkflows are both in `.spelling` already, and the Workflow section above writes them plain, which is what the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects) wants for API objects.

tico24 (commented):
Docs only from me, the rest is for @argoproj/argo-workflows-approvers. A few questions on the new section.

docs/default-workflow-specs.md
tico24, line 68: "the controller's in-memory copy during reconciliation" is a bit controller-internals for someone editing a configmap. Isn't the useful bit that `kubectl get cronworkflow` still shows exactly what they applied?

PR #16290, opened 2026-06-21
refactor(controller): unify DAG and Steps execution in a shared Engine

isubasinghe:
See the [pull request guide](https://argo-workflows.readthedocs.io/en/latest/pull-requests/) for details on each item.

- [x] Ran `make pre-commit -B`
- [x] Signed-off commits with [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) messages
- [x] PR title is a conventional commit message (it becomes the release notes entry)
- [x] Unit or e2e tests cover the change
- [ ] For features: an associated issue and a feature description file (`make feature-new`)
- [x] Opened as draft; will mark "Ready for review" once builds are green

Fixes #12192, #16450, #16454

### Motivation

`dag.go` and `steps.go` have been two separate implementations of the same job for a long time. Each one works out which tasks can run, expands `withItems`/`withParam`/`withSequence`, handles retries and hooks, decides what gets omitted, and rolls phases up to the parent. They were written at different times and have drifted, so a fix on one side often never made it to the other (the hook gating from #12192 is one example). Readiness was also worked out by recursing through the dependencies of every task on every reconcile, which made cascading omission slow and awkward to change.

This PR replaces both with one Engine that drives DAG and Steps templates the same way.

### Modifications

The evaluation logic now lives in a new package, `workflow/common/dag`. `DAGEvaluator` takes the workflow and a task list and reports which tasks are ready, waiting, omitted or retrying. It never writes to the workflow. Expansion and the dependency topology are in the same package. Steps templates get the same treatment through an adapter that gives each step a synthetic dependency on the previous group, so the evaluator does not know or care which template type it is looking at. We effectively rewrite steps as a dag on the fly, and then only have a dag engine.

`workflow/controller/engine.go` consumes those results. It loops evaluate, dispatch, assess until nothing new happens in a cycle, creates omitted nodes, assesses TaskGroup, StepGroup and DAG phases, and runs lifecycle and exit hooks. `dag.go` and `steps.go` are now thin wrappers that build an Engine.

Along the way:

- hook orchestration moved into `hook_handler.go`
- the template execution half of `operator.go` moved into `operator_template_execution.go` (`reconcileTemplate` and `executeProcessedTemplate`)
- node phase transitions are an explicit table in `node_phase_sm.go`
- `desired_state.go` and `reconciler_k8s.go` separate "what should exist" from the k8s calls
- the `depends` grammar is one parser shared by validation and the evaluator (`common.ParseDepends`), the backoff formula is one function (`common.RetryBackoffWait`), and the retry policy decision is made in one place (`dag.RetryDecider`)

Two things change for users. A `depends` expression is now re-evaluated whenever a task it references changes state, and takes effect as soon as it is settled, instead of only after every referenced task has finished. And a StepGroup whose steps were all omitted is now Omitted rather than Succeeded, which memoized resubmit (#16938) needs.

The branch was rebased twice while open. Fixes that landed on main in the meantime were ported into the Engine: #12762, #16745, #16408, #16377, #16488, #16883/#16376, #16839, #16955.

### Verification

Unit tests: `KUBECONFIG=/dev/null go test ./workflow/controller/ ./workflow/common/... ./workflow/validate/ ./workflow/util/ ./pkg/...` is green.

The evaluator has its own suite in `workflow/common/dag` (`argo_test.go`, `expansion_test.go`, `bug_regression_test.go`, plus benchmarks). The Engine is covered by `engine_test.go`, `engine_hooks_test.go`, `engine_per_child_test.go`, `steps_engine_test.go`, `operator_concurrency_test.go` and the existing `dag_test.go`/`steps_test.go`, which were ported across.

Regressions found in review each got a dedicated test that was checked to fail at the base commit before the fix went in: `exit_hook_stuck_regression_test.go`, `dag_target_phase_regression_test.go`, `lastretry_defaults_regression_test.go`, `stepgroup_regression_test.go`, `retry_backoff_regression_test.go`, `skipped_defaults_regression_test.go`.

The full e2e matrix in CI is green.

### Documentation

- `docs/enhanced-depends-logic.md` has a new "When a task runs" section for the settled-evaluation rule.
- `docs/upgrading.md` has a note on the `depends` timing change with a migration example.
- `workflow/common/dag/README.md` and `doc.go` describe the evaluator package.

### AI

Claude Code was used to review this branch, to draft the regression tests, and for some of the follow-up fix commits. Everything it produced was read and checked by the authors.


Joibel:
### Review notes — design-level points for the author

These came out of a multi-agent review of this PR plus a follow-up pass on the `dag-refactor-engine-v4` branch. The regressions and the mechanical fixes from that pass have landed as separate commits on the branch (see the commit log from `a6fb7d02d` onward); the items below are judgment calls I'd rather leave to you than fix unilaterally.

1. **`K8sTaskReconciler.Reconcile` aborts the batch on the first error** (`workflow/controller/reconciler_k8s.go:88`). One failing task skips every remaining `DesiredTask` in the same pass — notably the sibling expanded children dispatched together from `reconcileExpanded`. Is per-task isolation (continue, collect errors) intended here, as it is in `converge`?
2. **`topologicalSort` tolerates cycles silently** (`workflow/common/dag/topology.go`, Kahn's algorithm with an input-order fallback). Validation rejects cycles so this is defensive, but `util/sorting.TopologicalSorting` — already applied to the same graph in `validate.go` — returns an error instead. Consider reusing it or at least logging when the fallback triggers.
3. **Three `shouldExecute` functions.** `workflow/common/dag/expansion.go` (argoexpr, on an *unsubstituted* `when` during expansion), `workflow/controller/engine.go` (govaluate, post-substitution — moved from `steps.go`), and `workflow/cron/operator.go` (govaluate, pre-existing, with a now-stale `TODO: refactor shouldExecute in steps.go`). The first two share a name but not semantics; behaviour on a malformed `withParam` differs from the pre-Engine path. Worth a rename and a decision on which evaluator owns which case.
4. **The evaluator's API is order-dependent.** `evaluateAllStates` mutates `previouslyOmitted`/phase state and must run before any phase-reading method, but `GetTargetTasks`, `FindLeafTaskNames` and `GetDependencies` don't call it and the Engine invokes them outside an evaluation. It works today because `EvaluateAll` always runs first; a pure `Evaluate(ctx) Snapshot` with readers on the snapshot would remove the hazard.
5. **`workflowStore.getPhase` reports a running daemon as `Succeeded`** (`store.go`). The same intent is already expressed by `EvaluationResult.FulfilledForDeps`; having both means one of them is a lie to some reader (e.g. `CurrentPhase` flowing into `assessDAGPhase`).
6. **`workflowStore`'s `RWMutex` guards only `phases`** while `nodes` is read unguarded in the same methods. Either the store is single-goroutine (then the mutex advertises a guarantee it doesn't provide) or it isn't (then the unguarded map is the bigger exposure).
7. **`exit_handler_test.go` calls `MarkTaskResultComplete` on a stale `woc`** (the `leafA` loop, copying the pre-existing `leafB` block); `newWorkflowOperationCtx` deep-copies, so the call is a no-op and the test passes for a different reason than it says.

Joibel:
### Rebased onto `main` (`8f0d28008`) — what needed porting by hand

The branch was 174 commits behind `main`. Conflicts were confined to the first three commits of the PR (the refactor, the test port, the dead-code removal); everything after applied cleanly. The interesting part is the set of upstream fixes that landed in code this PR moved or deleted, so `git` could not carry them across — each was re-applied in its new home and verified by the test that accompanied it upstream:

1. **#12762 `pendingTimeout`** — the old `checkTemplateTimeout` no longer exists. `checkConstraints` (in `operator_template_execution.go`) now calls `checkTemplateTimeouts`, deletes a timed-out pod by UID, and requeues at the earliest deadline, mirroring upstream's `executeTemplate`. `TestCheckTemplateTimeouts` passes.
2. **#16745 retryable sync-lock errors** — `handleSynchronization` now treats `IsRetryableSyncError` / transient errors as "leave the node Pending and requeue" instead of erroring it. The sync tests from #16745, #16777 and #16776 pass.
3. **#16408 `typedConfigMapInformer`** — the rename applied at the PR's `ProcessArgs` call sites in `engine.go` and `operator.go`.
4. **#16377 `podBuilder`** — upstream now starts the `create_workflow_pod` span after the shutdown guard; the PR's tracing re-root (`ContextWithNode`) is kept and its own, now duplicate, span start dropped.
5. **#16488 onExit DAG handler under `Stop` shutdown** — the only port the suite caught after the rebase (`TestOnExitDAGNotFailedOnShutdownStop`): upstream added `&& !dagCtx.onExitTemplate` to the shutdown condition in the old `dag.go`; the Engine has the same flag, so the equivalent one-liner is in `Engine.finalize`, as its own commit (`04ae9d75e`).
6. **#16454 orphaned TaskGroup nodes stuck Running** — needed **no** port: the Engine's TaskGroup assessment already covers it (`TestDAGOrphanedTaskGroupCompletes` passes) and the old `taskGroupsToComplete` plumbing went with `dag.go`.

Verification on the rebased tree: `go build`/`go vet` clean, `golangci-lint` clean on every touched package, and the full `workflow/controller`, `workflow/common`, `workflow/common/dag` and `workflow/validate` suites green. The rewritten commits are unsigned (DCO trailers intact).

isubasinghe:
One thing to note here is that step withItems and dag withItems now both have a task group. This is better, it fixes some issues in expression/variables

isubasinghe:
This looks good, it just needs a rebase

Joibel:
Rebased onto main at 4389bbf96 (114 upstream commits since the previous base 8f0d28008). The previous head is kept at `pipekit/dag-refactor-engine-v4` reflog / locally as 04ae9d75e.

Conflicts were again confined to the first refactor commit and the test-port commit. As before, I took the Engine's side of those conflicts and then ported what main had changed in the code the Engine replaced. The ports are two commits on top of the rebased history so they can be reviewed on their own:

**fefcb1e8b — port the upstream fixes dropped by the rebase**

- #16883 / #16376 (node ID hash collisions): `Engine.taskNodeID`, `Engine.parentNodeNames`, the `workflow/common/dag` store and the synchronization lock in `operator_template_execution.go` now resolve node IDs through `wf.ResolveNodeID`, matching what main did to `dagContext.taskNodeID` and `executeTemplate`. The retry attempt's `addChildNode` moved to just before dispatch in `handleRetries`, as main did in `executeTemplate`, so an edge is never persisted for a node that is not created. `initializeNode` keeps the Engine's allowance for overwriting a task-result placeholder, checked against the 32-bit slot because `ResolveNode` only claims nodes whose `Name` matches. Main's three new `TestNodeIDCollision*` tests pass against the Engine.
- Main's `expandingTaskGroups` guard: a TaskGroup whose expanded items have not all materialized this reconcile (parallelism, deadline, pending exit hook) must not be completed from the children created so far. The Engine tracks the same thing in `Engine.expandingTaskGroups` via `noteExpansionProgress`, consulted by `assessTaskGroups`. New regression test `TestDAGTaskGroupSkippedThenDeferredItems` (a `when`-skipped first item plus two parallelism-deferred items) fails without the guard.
- #16839 (optional artifact from a skipped step): the check moved from `resolveReferences`, which the Engine deleted, to `wfScope.resolveArguments`. `TestResolveReferencesSkipsOptionalArtifactFromSkippedStep` is replaced by `TestResolveArgumentsSkipsOptionalArtifactFromSkippedStep` against that function.
- #16955 log quieting: the "Node was nil" warning in `prepareNode` is gone, as in main's `executeTemplate`.
- #16000 (containerSet siblings) supersedes the Engine's `containerSetRetry` special case in `podHasContainerNeedingTermination`; main's version is kept and both tests remain.
- Test fixtures for the evaluator name their nodes so `ResolveNodeID` finds them; `TestSubstitute` builds a controller since `maxOperationTime` now lives on it (#16702).

**129729e6c — mark a fully omitted StepGroup Omitted, not Succeeded**

Main's #16938 makes memoized resubmit reuse the retry reset planner, which refuses to reset a failed node that has a Succeeded descendant. When an early step group fails, the Engine creates the later groups' steps as Omitted and previously assessed those groups as Succeeded; they hang off the failed group's outbound nodes, so `TestMemoizedResubmitStepsNestedFanout` kept the failure instead of re-running it. A group whose every step was omitted never ran, so it is now Omitted, and `Running → Omitted` is added to the node phase state machine since StepGroups are initialized Running.

Everything else from main merged cleanly. Locally: build, vet and golangci-lint clean; `workflow/controller`, `workflow/common`, `workflow/util`, `pkg` and `util` unit suites green.

PR #16291, opened 2026-06-22
fix: add configurable database connection timeout

### LINKED ISSUE #14112: PostgreSQL cannot report an error when it cannot connect.
### Pre-requisites

When users sometimes misconfigure PostgreSQL information (perhaps due to network disconnection), errors cannot be reported and workflow-controller will freeze.

```
time="2025-01-21T07:09:49.701Z" level=info msg="Persistence configuration enabled"
time="2025-01-21T07:09:49.706Z" level=warning msg="Non-transient error: <nil>"
time="2025-01-21T07:09:49.709Z" level=warning msg="Non-transient error: <nil>"

```

### Version(s)

latest

```YAML
Unreachable PostgreSQL address
```

```text
```

```text
```

HsiuChuanHsu:
Fixes #14112

### Motivation

The DB session is built with no application-level connect timeout, so it relies on the OS default.

### Modifications

- Add `connectionTimeoutSeconds` to the DB config (default 5s), mirroring `ArtifactDriver.ConnectionTimeoutSeconds`. Applies to both persistence and synchronization (shared `DBConfig`).
- **PostgreSQL:** set `connect_timeout` (covers dial + handshake read, then resets — later queries unaffected), on the creds and Azure/AWS RDS paths.
- **MySQL:** set the dial `Timeout` only; `ReadTimeout` is intentionally left unset since it would cap every query read.

### Documentation

Documented in `docs/workflow-controller-configmap.yaml`; feature file added under `.features/pending/`.

### AI

Claude Sonnet 4.8
- Debug core logic
- Refine comment wording
- Unit tests

## Summary by CodeRabbit

* **New Features**
  * Added `connectionTimeoutSeconds` to persistence and synchronization config, defaulting to 5 seconds.
  * Applied the setting to PostgreSQL and MySQL connection establishment (with the timeout applied specifically to connecting).

* **Bug Fixes**
  * Prevented stalled or half-open database connections from hanging indefinitely; sessions now fail within the configured timeout.

* **Tests**
  * Added unit and regression tests validating default/custom timeout behavior and half-open connection handling.

* **Documentation**
  * Updated workflow controller config documentation to include the new timeout fields.


.features/pending/14112-db-connection-timeout.md
```diff
@@ -0,0 +1,8 @@
+Component: General
```
isubasinghe: It could be argued that this is a feature or a fix. 
We should probably make the decision to classify this as one or the  other though. 

This is arguably a fix more than a feature, so I would prefer to remove this file.

isubasinghe (commented):
I am fine with this but remove the file from the .features folder please

.features/pending/14112-db-connection-timeout.md
HsiuChuanHsu: Thanks for the feedback.
Agree - fix is the right classification. I've removed the feature file.

docs/workflow-controller-configmap.yaml
```diff
@@ -267,6 +267,10 @@ data:
 
   # enable persistence using postgres
   persistence: |
+    # Timeout in seconds for establishing a database connection. Default 5s.
```
Joibel, line 270: This should probably be duplicated into synchronization too for discoverability.

docs/workflow-controller-configmap.yaml
```diff
@@ -267,6 +267,10 @@ data:
...
+    # Prevents the controller from hanging indefinitely when the DB is unreachable
+    # or half-open; on timeout it fails fast so Kubernetes can restart it.
```
Joibel, line 272: half-open on MySQL would need read-timeout setting too.

Joibel (changes requested):

util/sqldb/connection_timeout_test.go
```diff
@@ -0,0 +1,122 @@
...
+func (sl *silentListener) hostPort(t *testing.T) (string, int) {
+	t.Helper()
+	host, portStr, err := net.SplitHostPort(sl.ln.Addr().String())
+	require.NoError(t, err)
+	port, err := strconv.Atoi(portStr)
+	require.NoError(t, err)
+	return host, port
+}
+
+// TestPostgresConnectionTimeoutHalfOpen is the regression test for issue #14112:
```
Joibel, line 63: It would be nice to see an identical test for MySQL

docs/workflow-controller-configmap.yaml
HsiuChuanHsu: Thanks for the review.

I had concern in the beginning about the read-timeout setting would cap every query read, so i intentionally left it. However, rather than setting `readTimeout`, I wrapped the `driver's Connector.Connect` in a `context.WithTimeout` (see `timeoutConnector` in `sqldb.go`).

Now, `readTimeout` applies to every read for the connection's entire lifetime, so setting it short enough to fail-fast on a hung handshake would also cut off long-running archive/list queries.

util/sqldb/connection_timeout_test.go
HsiuChuanHsu: Added.

docs/workflow-controller-configmap.yaml
HsiuChuanHsu: Updated, thanks!

Joibel (approved):
👍

PR #16292, opened 2026-06-22
fix(sso): fix "asymmetric encryption algorithms not supported for JWT". Fixes #16232

### LINKED ISSUE #16232: SSO login failure: invalid key encryption algorithm: RSA-OAEP-256
### Pre-requisites

While deploying a new instance, I encountered an error trying to login via SSO. `argo-server` completes the authorization code flow successfully and responds with a `set-cookie` header containing an `authorization` cookie as expected. The callback response redirects to `/`, but subsequent requests to `/api/v1/userinfo` and `/api/v1/info` return 401 with this response:

```json
{"code":16,"message":"failed to parse encrypted token: asymmetric encryption algorithms not supported for JWT: invalid key encryption algorithm: RSA-OAEP-256"}
```

The application then navigates to `/login` where you can start the process over again.

I first tried deleting the `sso` secret and restarting `argo-server`, but I investigated further after this didn't change the behavior. I noticed that the error message is [emitted by `go-jose/v4`](https://github.com/go-jose/go-jose/blob/d50302948f33c6365a8aa2d690dae02ebdd548a5/jwt/jwt.go#L117) , and that PR #16213 was merged yesterday. A quick check at quay.io confirmed that `latest` was rebuilt yesterday, which introduced the bug.

I pinned images to `v4.0.5`, deleted the `sso` secret again, restarted `argo-server`, cleared cookies, and login succeeded.

### Version(s)

f7b79b64650002f9982677670fde2e87788741d642354cd9ce850dcb5f1cb975

```YAML
New installation with SSO (OAuth 2.0 - not Dex)
```

```text
No longer available
```

```text
No longer available
```

MasonM:
Fixes #16232

### Motivation

https://github.com/argoproj/argo-workflows/pull/16213 introduced a bug when SSO is enabled: after logging in, the JWT stored in the `authorization` cookie is unusable. Any attempt to access an API with it results in the following error:
```
{"code":16,"message":"failed to parse encrypted token: asymmetric encryption algorithms not supported for JWT: invalid key encryption algorithm: RSA-OAEP-256"}
```
image

### Modifications
This is happening because go-jose v4 dropped support for encrypting JWTs with a public key, specifically in https://github.com/go-jose/go-jose/pull/78. 

The fix is to use a nested JWT inside a JWE, since go-jose still supports decrypting a JWE with a public key. Note that this does mean user sessions will be effectively invalidated and they'll be forced to login again, but that seems minor enough it's not worth calling out in the release notes.
 
### Verification

Tested locally using `make start PROFILE=sso UI=true`
Screenshot_20260621_211236

### Documentation

N/A

### AI

Used GPT 5.5 for initial code and tests.

### DIFF
--- a/server/auth/sso/sso.go
+++ b/server/auth/sso/sso.go
@@ -59,6 +59,7 @@ type sso struct {
 	baseHRef          string
 	secure            bool
 	privateKey        crypto.PrivateKey
+	signer            jose.Signer
 	encrypter         jose.Encrypter
 	rbacConfig        *config.RBACConfig
 	expiry            time.Duration
@@ -190,7 +191,12 @@ func newSso(
 		Scopes:       append(c.Scopes, oidc.ScopeOpenID),
 	}
 	idTokenVerifier := provider.Verifier(&oidc.Config{ClientID: config.ClientID})
-	encrypter, err := jose.NewEncrypter(jose.A256GCM, jose.Recipient{Algorithm: jose.RSA_OAEP_256, Key: privateKey.Public()}, &jose.EncrypterOptions{Compression: jose.DEFLATE})
+	signer, err := jose.NewSigner(jose.SigningKey{Algorithm: jose.RS256, Key: privateKey}, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create JWT signer: %w", err)
+	}
+	encrypterOptions := (&jose.EncrypterOptions{Compression: jose.DEFLATE}).WithContentType("JWT")
+	encrypter, err := jose.NewEncrypter(jose.A256GCM, jose.Recipient{Algorithm: jose.RSA_OAEP_256, Key: privateKey.Public()}, encrypterOptions)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create JWT encrpytor: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


isubasinghe (approved):
I don't understand the point of this, we encrypt using the public key and decrypt using the private key?

It feels like we should just  use AES-GCM. 

The public key luckily is not serialised anywhere, but if it were, we would be open to token forgery.

MasonM:
@isubasinghe Thanks for the review!

> I don't understand the point of this, we encrypt using the public key and decrypt using the private key?

Yes, this is basically identical to the old behavior. With go-jose v3, the `jwt.ParseEncrypted()` function delegated to `jose.ParseEncrypted()`: https://github.com/go-jose/go-jose/blob/v3.0.5/jwt/jwt.go#L105-L115

and the call to `enc.Decrypt(s.privateKey)` happens indirectly in the `if err := tok.Claims(s.privateKey, c); err != nil {` line.

> It feels like we should just use AES-GCM.

If we were doing this from scratch, I'd definitely agree, but the reason I didn't do that here is backwards-compatibility. The RSA private key is stored in the secret `sso` as `cookieEncryptionPrivateKey`, and if we switched to AES, we'd need to store it somewhere else (or overwrite the RSA key, but then we need code to determine if `cookieEncryptionPrivateKey` needs to be migrated).

isubasinghe:
@MasonM thanks, I just wanted confirmation that we should fix this up in the future really. I have approved this for now. But seems like 5.0 should just use AES-GCM

isubasinghe:
Should only be ported depending on what https://github.com/argoproj/argo-workflows/pull/16213 gets ported to

PR #16344, opened 2026-07-02
feat: configurable allowlist

### LINKED ISSUE #16345: Make the workflowTemplateRef override allow-list configurable
# Summary

What change needs making?
Add a `WORKFLOW_USER_OVERRIDE_ALLOWLIST` environment variable that lets an
operator add `WorkflowSpec` fields to the allow-list, e.g.
`WORKFLOW_USER_OVERRIDE_ALLOWLIST=podSpecPatch,volumes`. Unknown field names
fail the controller at startup so typos surface. The nested `artifactGC`
security fields (`podSpecPatch`, `serviceAccountName`, `podMetadata`) stay
blocked regardless.

## Use Cases

When would you use this?

- An operator runs `templateReferencing: Secure` for the security guarantees,
  but wants to let submitters set one or two normally-blocked fields (e.g.
  `volumes` or `podSpecPatch`) that they've decided are safe in their cluster —
  without forking the controller.
- Adapting the restriction policy per environment (dev vs prod) via config/env
  rather than a code change.

isubasinghe:
Fixes #16345

## Motivation

Under `templateReferencing: Strict`/`Secure`, a `workflowTemplateRef` submitter may only set a hard-coded allow-list of `WorkflowSpec` fields; everything else is rejected so the template's security settings can't be overridden. That list was baked into the binary, so operators who wanted to permit an extra field they'd deemed safe had no choice but to fork and rebuild the controller.

## Modifications

- Added `WORKFLOW_USER_OVERRIDE_ALLOWLIST`: a comma-separated list of `WorkflowSpec` field names (the YAML/JSON names, e.g. `podSpecPatch,volumes`) that get added to the allow-list.
- Parsed and applied once at controller startup via `ConfigureUserOverrideAllowlistFromEnv`. Names are validated against the real `WorkflowSpec` fields, so a typo fails the controller at boot instead of silently leaving a field blocked.
- Nested `artifactGC` security fields (`podSpecPatch`, `serviceAccountName`, `podMetadata`) stay blocked regardless — this var doesn't touch them.

## Verification

`TestParseUserOverrideAllowlist` covers the empty/whitespace input, YAML→Go name mapping, and the two error paths (Go identifiers and unknown names both rejected). Existing `ValidateUserOverrides` tests still pass.

## Documentation

- `docs/environment-variables.md` — new Controller env-var row.
- `docs/workflow-restrictions.md` — "Extending the Allow-List" section.

## AI

Used Claude Code to help with the docs and tests.

### DIFF
--- a/.features/pending/configurable-allowlist.md
+++ b/.features/pending/configurable-allowlist.md
@@ -1,0 +1,11 @@
+Description: Allow for configuring the allow list when using workflow template refs.
+Authors: [Isitha Subasinghe](https://github.com/isubasinghe)
+Component: General
+Issues: 23
+
+When the controller runs with `templateReferencing: Strict` or `Secure`, workflows using a `workflowTemplateRef` may only set an allow-listed set of `WorkflowSpec` fields (`arguments`, `entrypoint`, `suspend`, and other benign knobs); every other field is blocked so a submitter cannot override the template's security settings.
+The `WORKFLOW_USER_OVERRIDE_ALLOWLIST` environment variable lets operators add fields to this allow-list.
+Set it on the workflow-controller to a comma-separated list of `WorkflowSpec` field names, using the YAML/JSON names as written in a workflow, for example `WORKFLOW_USER_OVERRIDE_ALLOWLIST=podSpecPatch,volumes`.
+Use this when your environment has decided a normally-blocked field is safe for submitters to override.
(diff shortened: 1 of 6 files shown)


terrytangyuan (approved):
Great addition

Joibel (approved):

Joibel:
Backporting as the fix that made this important was applied as a fix to 3.7 and 4.0

PR #16357, opened 2026-07-02
fix: reject stale copies of completed workflows using resourceVersion comparison

### LINKED ISSUE #16305: Recently completed workflow cache stores namespace/name but checks name
### Pre-requisites

There is an internal key mismatch in the controller's recently completed workflow cache.

On current `main` (`cec639645bedde59909a785fceb6bb2fa2dc4fa3`), `processNextItem` checks the cache using only the workflow name:

```go
if wf.Status.Phase != "" && wfc.checkRecentlyCompleted(wf.Name) {
```

But completed workflows are recorded using the informer key from `cache.MetaNamespaceKeyFunc`, which is `namespace/name`:

```go
key, _ := cache.MetaNamespaceKeyFunc(un)
wfc.recordCompletedWorkflow(key)
```

The same pattern is used on delete via `cache.DeletionHandlingMetaNamespaceKeyFunc(obj)`.

This means the cache stores values like `default/my-wf`, but later queries `my-wf`. In a namespaced controller or normal multi-namespace controller setup, this makes the recently completed workflow check mostly ineffective.

Expected behavior: the cache should use one consistent identity format for both recording and lookup. Either record and check `namespace/name`, or change the cache to explicitly store the same value that `processNextItem` checks. It should also account for workflow identity correctly if workflow names are reused.

This appears to have been introduced by #12198 (`62732b30a3724dad702352160cdf2a9e55a922c6`), where completed workflow tracking was added.

### Version(s)

Current `main`: `cec639645bedde59909a785fceb6bb2fa2dc4fa3`

The mismatch appears to have been introduced by #12198 / `62732b30a3724dad702352160cdf2a9e55a922c6`.

### Paste a minimal workflow that reproduces the issue

This is a controller bookkeeping bug visible from the code path rather than a workflow-spec-specific issue. Any completed workflow recorded through the informer completed/deleted path is stored under `namespace/name`, while `processNextItem` checks only `name`.

A behavioral reproduction would require a completed workflow to be evicted from the informer cache and then an older copy of that workflow to re-enter the controller queue within the recent-completion retention window.

No controller logs are required to observe the mismatch. The relevant code paths are:

- `workflow/controller/controller.go`: `processNextItem` calls `checkRecentlyCompleted(wf.Name)`
- `workflow/controller/controller.go`: completed workflow filter records `cache.MetaNamespaceKeyFunc(un)`
- `workflow/controller/controller.go`: delete handler records `cache.DeletionHandlingMetaNamespaceKeyFunc(obj)`

Not applicable; this is controller cache bookkeeping and does not depend on wait-container behavior.

Joibel:
Fixes #16305

### Motivation

  The recently-completed workflow cache added in #12198 records informer keys
  (`namespace/name`) but `processNextItem` checked it with `name` only. A
  Kubernetes name can never contain `/`, so the check never matched and the
  "Cache: Rejecting recently deleted" guard was dead code: a stale, still-running
  copy of a recently completed or deleted workflow re-entering the informer
  (e.g. via a relist after watch interruption) could be reprocessed,
  potentially recreating pods.

  Fixing just the key would introduce a new problem: a name-keyed membership
  check would wrongly reject a retried workflow, or a recreated workflow
  reusing a name, for up to 10 minutes after completion.

  Since #15090 the controller also tracks staleness for in-flight workflows via
  the `workflows.argoproj.io/last-seen-version` annotation plus an in-memory
  UID→resourceVersion map, using equality comparison because resourceVersions
  could not officially be ordered. `k8s.io/apimachinery/pkg/util/resourceversion`
  now provides sanctioned ordered comparison of resourceVersions of the same
  object, which lets both mechanisms be replaced by one simpler, UID-keyed one.

### Modifications

  - Replace `recentCompletions` and `lastSeenVersions` with a single
    `lastWrittenVersions` map, keyed by workflow UID, recording the last
    resourceVersion this controller wrote (or saw complete/be deleted) and
    when the workflow completed.
  - `isOutdated` now compares resourceVersions with
    `resourceversion.CompareResourceVersion` and reports whether the recorded
    version was the workflow's completed/deleted state: stale copies of
    in-flight workflows are requeued (as before), stale copies of completed
    workflows are dropped. Incomparable resourceVersions fail open (the
    workflow is processed) with a warning.
  - Retry is unaffected because this is an ordering check, not a membership
    check: the recorded completion resourceVersion acts as a floor, so only
    copies strictly older than it are rejected. A retry mutates the same
    object (same UID) and therefore always carries a newer resourceVersion;
    the controller's first write of the retried workflow then clears the
    completed state, restoring requeue semantics for in-flight staleness.
    A recreated workflow reusing a name has a new UID and is untracked.
  - `persistUpdates`/`persistWorkflowSizeLimitErr` record the Update response
    resourceVersion; the controller no longer writes the
    `workflows.argoproj.io/last-seen-version` annotation, and the constant is
    removed. Existing annotations on workflows are ignored and harmless.
  - Records of completed workflows expire after 10 minutes (matching the old
    cache's retention); expiry scans are throttled to once per minute.
    A stale completion event cannot regress the recorded version.

### Verification

  - New unit tests (`TestIsOutdated`, `TestExpireCompletedVersions`) cover the
    #16305 regression scenario, the full retry cycle (newer copy passes while
    the completed record stands; the first controller write after retry clears
    the completed state so stale copies requeue rather than drop), name reuse,
    fail-open on incomparable resourceVersions, no-regress on stale completion
    events, and record expiry. Neither previous mechanism had unit coverage.
  - `go test ./workflow/controller/... ./workflow/common/...` passes.
  - `golangci-lint run` on the changed packages reports only pre-existing
    issues on untouched lines.

  ### Documentation

  Not needed: this is internal controller bookkeeping. The removed
  `last-seen-version` annotation was never documented.

  ### AI

The idea to swtich to resourceVersion was mine.
This PR was developed with Claude Code (Claude Fable 5): analysis of the issue, the code changes, tests, and this PR text, reviewed and directed by me.

### DIFF
--- a/workflow/common/common.go
+++ b/workflow/common/common.go
@@ -57,9 +57,6 @@ const (
 	// the strategy whose artifacts are being deleted
 	AnnotationKeyArtifactGCStrategy = workflow.WorkflowFullName + "/artifact-gc-strategy"
 
-	// AnnotationKeyLastSeenVersion stores the last seen version of the workflow when it was last successfully processed by the controller
-	AnnotationKeyLastSeenVersion = workflow.WorkflowFullName + "/last-seen-version"
-
 	// LabelParallelismLimit is a label applied on namespace objects to control the per namespace parallelism.
 	LabelParallelismLimit = workflow.WorkflowFullName + "/parallelism-limit"
 
--- a/workflow/controller/controller.go
+++ b/workflow/controller/controller.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/resourceversion"
 	runtimeutil "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
@@ -72,19 +73,21 @@ import (
 
 const maxAllowedStackDepth = 100
 
-type recentlyCompletedWorkflow struct {
-	key  string
-	when time.Time
+// lastWrittenVersion records the most recent resourceVersion of a workflow
+// that this controller has written, or has seen complete or be deleted.
+type lastWrittenVersion struct {
+	resourceVersion string
+	// completedAt is set when the workflow completed or was deleted. The
+	// record is retained for completedVersionRetention afterwards so that
+	// stale copies of the workflow re-entering the informer can still be
+	// rejected, and is then expired.
+	completedAt time.Time
 }
 
-type recentCompletions struct {
-	completions []recentlyCompletedWorkflow
+type lastWrittenVersions struct {
+	versions    map[types.UID]lastWrittenVersion
+	nextCleanup time.Time
 	mutex       gosync.RWMutex
-}
-
-type lastSeenVersions struct {
-	versions map[string]string
-	mutex    gosync.RWMutex
 }
 
 // WorkflowController is the controller for workflow resources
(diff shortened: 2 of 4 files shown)


terrytangyuan (approved):

isubasinghe (approved):

Joibel:
Not cherry-picking to 3.7 because the k8s libs don't support comparing resource versions

Joibel:
> Not cherry-picking to 3.7 because the k8s libs don't support comparing resource versions

Actually, going to try the same tactic for 3.7 as 4.0.

PR #15440, opened 2026-01-28
feat: Add workflow-level executor plugin settings definition

ntny:
Fixes [#15234](https://github.com/argoproj/argo-workflows/issues/15234)

### Motivation

Enable workflow-level configuration of the Argo Workflow Executor Plugin. When settings are specified in the workflow spec under executorPlugins.spec.sidecar, the plugin uses that configuration directly. If no sidecar is defined in the workflow spec, it falls back to the ConfigMap, preserving the previous behavior.

In other words, if `executorPlugins.spec` is provided in the workflow spec, the settings from the ConfigMap are ignored, and the agent pod will launch only the sidecars defined in the workflow spec.

**Example Argo Workflow Spec**

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: http-template
  namespace: default
spec:
  podSpecPatch: |
    nodeName: virtual-node
  entrypoint: main
  executorPlugins: #executor plugin settings
    -  metadata:
        name: test-sidecar
    spec:
        sidecar:
          container:
            name: test-sidecar
            image: busybox:1.35
            ports:
              - containerPort: 8080
            resources:
              requests:
                cpu: "100m"
                memory: "128Mi"
              limits:
                cpu: "200m"
                memory: "256Mi"
            securityContext:
              runAsUser: 1000
              runAsGroup: 1000
              runAsNonRoot: true
  templates:
    - name: main
      steps:
        ...
```

This feature enable via `ARGO_WORKFLOW_LEVEL_EXECUTOR_PLUGINS=true` controller env variable

### How I tested
Sample of the Argo Workflow spec:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: wf-level-plugin
  namespace: argo
spec:
  entrypoint: hello-hello-hello
  executorPlugins:  # configuration for executor plugins
    - metadata:
       name:  print-message-plugin
    spec:
        sidecar:
          container:
            name: print-message-plugin
            image: print-message-plugin:latest
            imagePullPolicy: IfNotPresent
            ports:
              - containerPort: 8080
            resources:
              limits:
                cpu: 500m
                memory: 128Mi
              requests:
                cpu: 250m
                memory: 64Mi
            securityContext:
              runAsNonRoot: true
              runAsUser: 65534

  templates:
    - name: hello-hello-hello
      steps:
        - - name: hello1
            template: print-message
            arguments:
              parameters:
                - name: message
                  value: "hello1"
[...]

PR #16372, opened 2026-07-03
feat: added a resource agent to monitor resources. Fixes #16126

### LINKED ISSUE #16126: Run resource monitoring pods as agents.

isubasinghe:
Fixes #16126

### Motivation

This adds the ability for resources to be monitored by an agent style singular pods

### Modifications

* CRD modification to indicate the usage of resource agents.
* Added reconciler for agents pods as well as normal resource pods
* new service account added.

### Verification

* Tested on examples in the examples folder

### Documentation

Added documentation

### AI

Initial design was hand written, but was then wired up with claude. 
commit 4005f85 onwards was AI driven.

## Summary by CodeRabbit

* **New Features**
  * Added agent-based execution for resource templates using a shared per-workflow resource-agent pod.
  * Added support for monitoring created resources, evaluating success/failure conditions, and extracting output parameters.
  * Added `manifestFrom` support for loading resource manifests from artifacts.
  * Added examples and quick-start configuration, including required service accounts and RBAC.

* **Documentation**
  * Added guidance for configuration, RBAC, limitations, troubleshooting, and agent resynchronization settings.

* **Bug Fixes**
  * Improved resource-agent status updates, retries, restart recovery, and progress reporting.
  * Prevented invalid multi-document manifests and unsupported service-account settings in agent mode.

### DIFF
--- a/.features/pending/resource-agent.md
+++ b/.features/pending/resource-agent.md
@@ -3,7 +3,7 @@ Authors: [Isitha Subasinghe](https://github.com/isubasinghe)
 Component: General
 Issues: 16126
 
-Resource templates can now opt into agent-based execution by setting `resource.agent: true`.
+Resource templates can now opt into agent-based execution by setting `resource.mode: agent`.
 Instead of creating a separate workflow pod for every resource-template node, a shared resource agent pod executes and monitors all opted-in resource templates for the workflow.
 This reduces pod creation overhead for workflows that manage many Kubernetes resources.
 
--- a/.features/pending/resource-agent.md
+++ b/.features/pending/resource-agent.md
@@ -4,6 +4,8 @@ Component: General
 Issues: 16126
 
 Resource templates can now opt into agent-based execution by setting `resource.agent: true`.
-Instead of creating a separate workflow pod for every resource-template node, a shared resource agent pod executes and monitors all opted-in resource templates for the workflow. This reduces pod creation overhead for workflows that manage many Kubernetes resources.
+Instead of creating a separate workflow pod for every resource-template node, a shared resource agent pod executes and monitors all opted-in resource templates for the workflow.
+This reduces pod creation overhead for workflows that manage many Kubernetes resources.
 
-The resource agent runs under a dedicated `<workflow service account>-resource-agent` service account and requires permissions to create and watch the resource kinds used by the workflow. See `docs/resource-template.md` for configuration details, required RBAC, and examples.
+The resource agent runs under a dedicated `<workflow service account>-resource-agent` service account and requires permissions to create and watch the resource kinds used by the workflow.
+See `docs/resource-template.md` for configuration details, required RBAC, and examples.
--- a/.features/pending/resource-agent.md
+++ b/.features/pending/resource-agent.md
@@ -1,7 +1,7 @@
 Description: Execute resource templates with a shared resource agent
 Authors: [Isitha Subasinghe](https://github.com/isubasinghe)
 Component: General
-Issues: 99999
+Issues: 16126
 
 Resource templates can now opt into agent-based execution by setting `resource.agent: true`.
 Instead of creating a separate workflow pod for every resource-template node, a shared resource agent pod executes and monitors all opted-in resource templates for the workflow. This reduces pod creation overhead for workflows that manage many Kubernetes resources.
--- a/.features/pending/resource-agent.md
+++ b/.features/pending/resource-agent.md
@@ -1,0 +1,9 @@
+Description: Execute resource templates with a shared resource agent
+Authors: [Isitha Subasinghe](https://github.com/isubasinghe)
+Component: General
+Issues: 99999
+
+Resource templates can now opt into agent-based execution by setting `resource.agent: true`.
+Instead of creating a separate workflow pod for every resource-template node, a shared resource agent pod executes and monitors all opted-in resource templates for the workflow. This reduces pod creation overhead for workflows that manage many Kubernetes resources.
+
+The resource agent runs under a dedicated `<workflow service account>-resource-agent` service account and requires permissions to create and watch the resource kinds used by the workflow. See `docs/resource-template.md` for configuration details, required RBAC, and examples.
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@ dist/
 # delve debug binaries
 cmd/**/debug
 hack/**/debug
+# binaries built in-place (e.g. `go build ./cmd/argoexec` from its own dir)
+cmd/argo/argo
+cmd/argoexec/argoexec
+cmd/workflow-controller/workflow-controller
 hack/embeddoc/embeddoc
 hack/featuregen/featuregen
 /argo
--- a/.gitignore
+++ b/.gitignore
@@ -9,10 +9,6 @@ dist/
 # delve debug binaries
 cmd/**/debug
 hack/**/debug
-# binaries built in-place (e.g. `go build ./cmd/argoexec` from its own dir)
-cmd/argo/argo
-cmd/argoexec/argoexec
-cmd/workflow-controller/workflow-controller
 hack/embeddoc/embeddoc
 hack/featuregen/featuregen
 /argo
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -101,7 +101,7 @@ spec:
 
 | Name                                   | Type            | Default | Description                                                                                            |
 |----------------------------------------|-----------------|---------|--------------------------------------------------------------------------------------------------------|
-| `ARGO_AGENT_RESOURCE_INFORMER_RESYNC`  | `time.Duration` | `5m`    | How often the resource agent's informers resync as a safety net against missed watch events.          |
+| `ARGO_AGENT_RESOURCE_INFORMER_RESYNC`  | `time.Duration` | `5m`    | How often the resource agent's informers resync as a safety net against missed watch events. Read by the executor/agent process; see the other `ARGO_AGENT_*` variables in the Controller table above. |
 | `ARGO_DEBUG_PAUSE_AFTER`               | `bool`          | `false` | Enable [Debug Pause](debug-pause.md) after step execution
 | `ARGO_DEBUG_PAUSE_BEFORE`              | `bool`          | `false` | Enable [Debug Pause](debug-pause.md) before step execution
 | `ARTIFACT_S3_UPLOAD_THREADS`           | `int`               | `4`                                                                                         | Number of threads for artifact upload through S3. Default: 4.                                                                                                                                                                                                            |
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -101,6 +101,7 @@ spec:
 
 | Name                                   | Type            | Default | Description                                                                                            |
 |----------------------------------------|-----------------|---------|--------------------------------------------------------------------------------------------------------|
+| `ARGO_AGENT_RESOURCE_INFORMER_RESYNC`  | `time.Duration` | `5m`    | How often the resource agent's informers resync as a safety net against missed watch events.          |
 | `ARGO_DEBUG_PAUSE_AFTER`               | `bool`          | `false` | Enable [Debug Pause](debug-pause.md) after step execution
 | `ARGO_DEBUG_PAUSE_BEFORE`              | `bool`          | `false` | Enable [Debug Pause](debug-pause.md) before step execution
 | `ARTIFACT_S3_UPLOAD_THREADS`           | `int`               | `4`                                                                                         | Number of threads for artifact upload through S3. Default: 4.                                                                                                                                                                                                            |
--- a/docs/resource-template.md
+++ b/docs/resource-template.md
@@ -3,3 +3,21 @@
 > v2.0 and after
 
 See [Kubernetes Resources](walk-through/kubernetes-resources.md).
+
+## Agent-based execution
+
+When resource templates are executed by the agent instead of a per-node pod, be aware of the following:
+
+* Created resources are labeled with `workflows.argoproj.io/agent-resource: <workflow UID>` and
+  annotated with `workflows.argoproj.io/node-id` plus the template's success/failure conditions.
+  The agent's informers select on the label and evaluate the annotations, so these must not be
+  stripped by admission controllers or other mutating controllers.
+* The agent's service account needs `list` and `watch` on every resource kind your templates
+  create (to watch conditions), and `get` on `secrets` and `configmaps` if you use
+  `manifestFrom` artifacts, since the agent downloads them via the Kubernetes API rather than an
+  init container.
+* Node results are reported at most once per node. If the agent pod restarts, an
+  already-completed node's result may be patched into the `WorkflowTaskSet` a second time; this
+  is harmless.
+* Output parameters (`jsonPath`/`jqFilter`) are resolved from the watched object's state at the
+  moment its success conditions are met, not via `kubectl get`.
--- a/docs/resource-template.md
+++ b/docs/resource-template.md
@@ -36,6 +36,11 @@ When resource templates are executed by the agent instead of a per-node pod, be 
   is harmless.
 * Output parameters (`jsonPath`/`jqFilter`) are resolved from the watched object's state at the
   moment its success conditions are met, not via `kubectl get`.
+* Only a single manifest document is supported per template. A multi-document manifest
+  (`---`-separated) is rejected; split it into separate resource templates.
+* `manifestFrom` loads the artifact in-process, so it works only with the native artifact drivers
+  (`s3`, `gcs`, `oss`, `http`, `git`, `raw`). A `plugin:`-backed artifact cannot be used, because the
+  agent pod runs no artifact-plugin sidecars.
 
 ### Service account and RBAC
 
--- a/docs/resource-template.md
+++ b/docs/resource-template.md
@@ -6,14 +6,14 @@ See [Kubernetes Resources](walk-through/kubernetes-resources.md).
 
 ## Agent-based execution
 
-Set `resource.agent: true` on a resource template to run it on the shared per-workflow resource
+Set `resource.mode: agent` on a resource template to run it on the shared per-workflow resource
 agent pod instead of a dedicated per-node pod:
 
 ```yaml
 - name: create-thing
   resource:
     action: create
-    agent: true
+    mode: agent
     successCondition: status.phase == Running
     manifest: |
       ...
@@ -38,9 +38,13 @@ When resource templates are executed by the agent instead of a per-node pod, be 
   moment its success conditions are met, not via `kubectl get`.
 * Only a single manifest document is supported per template. A multi-document manifest
   (`---`-separated) is rejected; split it into separate resource templates.
-* `manifestFrom` loads the artifact in-process, so it works only with the native artifact drivers
-  (`s3`, `gcs`, `oss`, `http`, `git`, `raw`). A `plugin:`-backed artifact cannot be used, because the
-  agent pod runs no artifact-plugin sidecars.
+* `manifestFrom` artifacts are loaded by the agent pod itself. The native artifact drivers
+  (`s3`, `gcs`, `oss`, `http`, `git`, `raw`) run in-process; `plugin:`-backed artifacts work too,
+  because every [artifact driver plugin](configure-artifact-repository.md) registered in the
+  controller's `artifactDrivers` configuration is installed into the agent pod as a sidecar.
+* `{{pod.name}}` still resolves in agent-based resource templates — to the name the per-node pod
+  *would* have had — so manifests keep working unchanged when switching modes. No pod with that
+  name actually exists, so don't pass it to anything that expects a real pod.
 
 ### Service account and RBAC
 
--- a/docs/resource-template.md
+++ b/docs/resource-template.md
@@ -8,16 +8,62 @@ See [Kubernetes Resources](walk-through/kubernetes-resources.md).
 
 When resource templates are executed by the agent instead of a per-node pod, be aware of the following:
 
+* Agent-based resource templates run in their own pod, named `<workflow>-<id>-resource-agent` and
+  labeled `workflows.argoproj.io/component: resource-agent`, separate from the HTTP/plugin agent
+  pod. Like the HTTP agent, one such pod serves all resource-template nodes of the workflow.
 * Created resources are labeled with `workflows.argoproj.io/agent-resource: <workflow UID>` and
   annotated with `workflows.argoproj.io/node-id` plus the template's success/failure conditions.
   The agent's informers select on the label and evaluate the annotations, so these must not be
   stripped by admission controllers or other mutating controllers.
-* The agent's service account needs `list` and `watch` on every resource kind your templates
-  create (to watch conditions), and `get` on `secrets` and `configmaps` if you use
-  `manifestFrom` artifacts, since the agent downloads them via the Kubernetes API rather than an
-  init container.
 * Node results are reported at most once per node. If the agent pod restarts, an
   already-completed node's result may be patched into the `WorkflowTaskSet` a second time; this
   is harmless.
 * Output parameters (`jsonPath`/`jqFilter`) are resolved from the watched object's state at the
   moment its success conditions are met, not via `kubectl get`.
+
+### Service account and RBAC
+
+The resource agent watches every resource kind your templates create, which requires `list` and
+`watch` on the whole kind — a broader grant than workflow pods should carry. It therefore runs
+under a dedicated service account named `<workflow service account>-resource-agent` (for example
+`default-resource-agent` for workflows using the `default` service account). The workflow errors
+if this service account does not exist.
+
+The service account needs:
+
+* `list` and `watch` on `workflowtasksets`, and `patch` on `workflowtasksets/status` (to receive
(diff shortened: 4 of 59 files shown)


workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+	if err != nil {
+		logger.WithError(err).Error(ctx, "failed to create resource")
+		rae.report(nodeID, wfv1.NodeResult{Phase: wfv1.NodeFailed, Message: fmt.Sprintf("failed to create resource: %v", err)})
+		return
+	}
+	logger.WithFields(logging.Fields{"namespace": obj.GetNamespace(), "name": obj.GetName(), "kind": obj.GetKind()}).Info(ctx, "Created resource")
+
+	// A delete leaves no live object to watch, and zero requirements (empty or whitespace-only
+	// conditions) mean there is nothing to wait for: succeed immediately, as WaitResource does.
+	if tmpl.Resource.Action == "delete" || (len(rc.successReqs) == 0 && len(rc.failReqs) == 0) {
```
Joibel: **[bug]** A whitespace-only success/failure condition (e.g. produced by parameter substitution) parses to zero label requirements, so this branch reports immediate `Succeeded` without ever reading the resource back.

The per-pod path was deliberately hardened against exactly this case by this same PR — `WaitResource` in `workflow/executor/resource.go` (lines 219‑223) only short-circuits when both conditions are the **empty string**, with a comment explaining that a whitespace-only condition “must still poll at least once … instead of being reported a spurious success”. The agent path diverges from that guard: consider checking the raw strings here (as `WaitResource` does) rather than the parsed requirement counts.

workflow/controller/resource_template.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		_, node = woc.initializeExecutableNode(ctx, nodeName, wfv1.NodeTypeResourceMonitor, templateScope, tmpl, orgTmpl, opts.boundaryID, wfv1.NodePending, opts.nodeFlag, true)
+	}
+	if !node.Fulfilled() {
+		// The agent applies exactly one object per node; a multi-document inline manifest would be
+		// silently partly applied (only doc 1 created, reported Succeeded). Reject it loudly here,
+		// matching the agent's own single-doc guard. setOwnerReference is injected by the agent's
+		// create path (withAgentMetadata) so it also covers manifestFrom, whose content we never see.
+		if common.ManifestDocCount([]byte(tmpl.Resource.Manifest)) > 1 {
+			return woc.markNodeError(ctx, nodeName, fmt.Errorf("agent-based resource templates support only a single manifest document"))
```
Joibel, line 24: **[validation gap]** The single-manifest-document constraint on `resource.agent: true` is only enforced here at runtime, so a workflow with a static multi-document manifest passes lint/submit and fails mid-run. For static manifests this is checkable at submit time in `workflow/validate` (manifests coming from parameters/artifacts can still be left to this runtime check).

Same feedback pattern as the previous agent-routed field: on #15440 a maintainer asked for “some code in `workflow/validate` to validate this field”.

Joibel (commented):
### Code review

Found 19 issues (7 high-confidence, 12 minor). Each is also posted as an inline comment on the relevant file/line. Overall this is a carefully written change — it preserves the prior taskset fixes (#16075, #15220), handles the tricky races (restart safety, create-to-watch gap, terminal-node resurrection), and the client-go usage is idiomatic. The one real bug found is #1.

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+	if action == "" {
+		action = "create"
+	}
+	args, err := getKubectlArguments(action, manifestPath, resource.Flags, resource.MergeStrategy)
+	if err != nil {
+		return nil, gvr, namespaced, err
+	}
+
+	// runKubectl mutates global os.Args and kubectl's fatal handler; serialize across workers.
+	rae.kubeCTLMutex.Lock()
```
Joibel: **[scalability]** Because in-process kubectl mutates global state (`os.Args`, the fatal handler), this global mutex serializes resource *creation* across all task workers — partly defeating the worker-pool concurrency and the PR's goal of reducing overhead for workflows that create many resources. The per-pod executor got away with global-state kubectl because each pod ran exactly one invocation; the shared agent amplifies it.

The executor already holds a `DynamicClient` (used for get/list/watch/informers); create could go through it, avoiding both the global-state hazard and the serialization. Reusing kubectl buys behavior parity (`resource.Flags` passthrough, merge strategies, server-side apply), so if the mutex stays it's worth a comment acknowledging the throughput ceiling as a conscious trade-off.

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+	}
+
+	go rae.patchWorker(ctx, rae.WorkflowInterface.ArgoprojV1alpha1().WorkflowTaskSets(rae.Namespace), requeueTime)
+	go rae.runEventWorker(ctx)
+	go func() {
+		<-ctx.Done()
+		rae.eventQueue.ShutDown()
+	}()
+
+	<-ctx.Done()
```
Joibel, line 940: **[reliability]** The resource agent only exits on `ctx.Done()`. The sibling `AgentExecutor` (workflow/executor/agent.go, ~lines 101‑113) additionally self-terminates when the taskset is deleted (`watch.Deleted`) and when the workflow-completed label is set (`IsWorkflowCompleted`). Here the taskset informer's `DeleteFunc` only logs, and `LabelKeyCompleted` is never checked — so any path where the controller fails to delete the pod leaves it running indefinitely. Consider mirroring `AgentExecutor`'s belt-and-suspenders self-exit.

workflow/executor/agent.go
```diff
@@ -233,6 +233,11 @@ func (ae *AgentExecutor) processTask(ctx context.Context, tmpl wfv1.Template) (*
 		executeTemplate = ae.executeHTTPTemplate
 	case tmpl.Plugin != nil:
 		executeTemplate = ae.executePluginTemplate
+	case tmpl.Resource != nil:
```
Joibel, line 254: **[coupling]** This skip and `ResourceAgentExecutor`'s inverse skip (non-resource tasks) are the only things keeping the two agents from stepping on each other's nodes in the shared `WorkflowTaskSet`, and the relationship is enforced only by comments. Adding a third agent-served template type would require updating every agent's skip logic or nodes error spuriously. Worth a cross-referencing comment at both sites at minimum (or a shared predicate both agents consume).

docs/resource-template.md
```diff
@@ -3,3 +3,87 @@
...
+    agent: true
+    successCondition: status.phase == Running
+    manifest: |
+      ...
+```
+
+See [`resource-agent.yaml`](https://github.com/argoproj/argo-workflows/blob/main/examples/resource-agent.yaml)
+for a complete workflow example.
+
+When resource templates are executed by the agent instead of a per-node pod, be aware of the following:
```
Joibel, line 25: **[docs]** This “be aware” list omits a hard user-facing limitation: agent-based resource templates reject multi-document manifests (`ManifestDocCount > 1` → “agent-based resource templates support only a single manifest document”). A user splitting resources with `---` — a very common k8s pattern — hits a runtime error the docs never warn about. Worth a bullet here.

docs/resource-template.md
```diff
@@ -3,3 +3,87 @@
...
+
+The service account needs:
+
+* `list` and `watch` on `workflowtasksets`, and `patch` on `workflowtasksets/status` (to receive
+  tasks and report results, the same as the [HTTP agent](workflow-rbac.md)),
+* on every resource kind your templates touch, the verbs matching your actions plus the ones the
+  agent always needs: `create`, `get`, `list` and `watch` at minimum (the agent creates the
+  resource, reads it back for restart-safety, and watches/polls it for its success and failure
+  conditions), plus `update` for `apply`/`replace`, `patch` for `patch`, and `delete` for `delete`,
+* `get` on `secrets` and `configmaps` if you use `manifestFrom` artifacts, since the agent
```
Joibel, line 65: **[docs]** The example file header (`examples/resource-agent-manifest-from.yaml`) documents that agent-based `manifestFrom` only works with in-process artifact drivers (native s3/gcs/oss/http/git/raw) and cannot use plugin-backed artifacts, but that constraint never made it into this doc, where users will actually look. Worth stating alongside the `manifestFrom` RBAC note.

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+	return nil
+}
+
+// withAgentMetadata parses a single-object manifest and re-marshals it with the UID-valued
+// agent-resource label (informer selection) plus the success/failure conditions and owning
+// node ID as annotations, making every event self-describing — handlers can evaluate and
+// attribute results with no agent-side state, and a restarted agent's re-list resumes cleanly.
+// It returns both the re-marshaled manifest and the parsed object.
+// ponytail: single-document manifests only; multi-doc is rejected rather than silently partly
```
Joibel: **[nit]** Stray token: “ponytail” isn't a recognized comment convention (TODO/NOTE/…) and doesn't relate to the sentence. Looks like a mis-paste/authoring artifact — drop it or replace with `Note:`.

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+}
+
+// ResourceAgentExecutor holds state for all resource agent work
+type ResourceAgentExecutor struct {
+	WorkflowName      string
+	WorkflowUID       string
+	ClientSet         kubernetes.Interface
+	DynamicClient     dynamic.Interface
+	WorkflowInterface workflow.Interface
+	RESTClient        rest.Interface
```
Joibel: **[dead code]** `RESTClient` is assigned in `NewResourceAgentExecutor` but never read — the agent resolves GVRs via the RESTMapper and does all reads/writes through `DynamicClient`/`ClientSet`/`WorkflowInterface`. The dead plumbing extends to the constructor's `restClient` parameter and `clientSet.RESTClient()` in `cmd/argoexec/commands/resource_agent.go`.

workflow/controller/taskset.go
```diff
@@ -175,7 +198,9 @@ func (woc *wfOperationCtx) createTaskSet(ctx context.Context) error {
 	}
 
 	woc.log.Info(ctx, "Creating TaskSet")
-	labels := map[string]string{}
+	labels := map[string]string{
+		common.LabelKeyWorkflowUID: string(woc.wf.UID),
```
Joibel: **[dead code]** `LabelKeyWorkflowUID` is written here (and in `createTaskSet` below) but never read anywhere in the codebase. Its doc comment in `workflow/common/common.go` says it exists “so per-workflow agents can select exactly their own taskset”, but both agents select their taskset by `metadata.name` field selector — and the resource agent's own comment argues the name selector is the spoof-proof choice. Looks like a leftover from an earlier label-based design; either remove it or wire selection to it (removal seems right).

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case nodeID := <-taskQueue:
+			rae.createAndWatchResource(ctx, rae.templateForNode(nodeID), nodeID)
+		}
+	}
+}
+
+func (rae *ResourceAgentExecutor) patchWorker(ctx context.Context, taskSetInterface v1alpha1.WorkflowTaskSetInterface, requeueTime time.Duration) {
```
Joibel, line 766: **[dry]** This mirrors `AgentExecutor.patchWorker` (workflow/executor/agent.go:169): same `json.Marshal({"status": WorkflowTaskSetStatus{Nodes}})` payload, `retry.OnError` with backoff, `Patch(…, MergePatchType, …, "status")`, and mark-pending-nodes-`NodeError` on non-transient failure — the comment below even says “matching AgentExecutor.patchWorker”. The two patch paths will drift over time; a shared taskset-status-patch helper embedded by both would prevent that. (Backoff params also differ: 5 steps/30s cap here vs 10/60s there — if intentional, worth a note.)

docs/environment-variables.md
```diff
@@ -101,6 +101,7 @@ spec:
 
 | Name                                   | Type            | Default | Description                                                                                            |
 |----------------------------------------|-----------------|---------|--------------------------------------------------------------------------------------------------------|
+| `ARGO_AGENT_RESOURCE_INFORMER_RESYNC`  | `time.Duration` | `5m`    | How often the resource agent's informers resync as a safety net against missed watch events.          |
```
Joibel: **[docs nit]** Every sibling `ARGO_AGENT_*` variable (`ARGO_AGENT_TASK_WORKERS`, `ARGO_AGENT_PATCH_RATE`, `ARGO_AGENT_CPU_LIMIT`, `ARGO_AGENT_MEMORY_LIMIT`) lives in the “Controller” table above. This one is genuinely read by the executor process, so it's not wrong here, but readers scanning agent config will look for it next to the others — consider co-locating or cross-referencing.

.features/pending/resource-agent.md
```diff
@@ -0,0 +1,9 @@
+Description: Execute resource templates with a shared resource agent
+Authors: [Isitha Subasinghe](https://github.com/isubasinghe)
+Component: General
+Issues: 16126
+
+Resource templates can now opt into agent-based execution by setting `resource.agent: true`.
+Instead of creating a separate workflow pod for every resource-template node, a shared resource agent pod executes and monitors all opted-in resource templates for the workflow. This reduces pod creation overhead for workflows that manage many Kubernetes resources.
```
Joibel: **[style]** `.features/TEMPLATE.md` asks for “One sentence per line of markdown”; this line contains two sentences.

workflow/executor/agent_test.go
```diff
@@ -40,6 +41,19 @@ func TestUnsupportedTemplateTaskWorker(t *testing.T) {
 	assert.Contains(t, response.Result.Message, "agent cannot execute: unknown task type")
 }
 
+func TestAgentSkipsResourceTasks(t *testing.T) {
```
Joibel, line 49: **[test nit]** This test over-specifies: it sets `Resource.Agent: true` and initializes `consideredTasks`, but the code under test (`case tmpl.Resource != nil` in `processTask`) checks neither — it skips *all* resource templates regardless of the `Agent` flag and returns before touching `consideredTasks`. The test would pass with `Agent: false`, so the name/setup imply agent-flag-specific gating that doesn't exist. Consider renaming or dropping the misleading setup.

workflow/executor/resource-agent_test.go
```diff
@@ -0,0 +1,270 @@
...
+	k8stesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/yaml"
+
+	wfv1 "github.com/argoproj/argo-workflows/v4/pkg/apis/workflow/v1alpha1"
+	"github.com/argoproj/argo-workflows/v4/util/logging"
+	"github.com/argoproj/argo-workflows/v4/workflow/common"
+)
+
+func TestManifestDocCount(t *testing.T) {
```
Joibel: **[test nit]** `TestManifestDocCount` tests `common.ManifestDocCount` (defined in `workflow/common/common.go` and also used by the controller) — it belongs in `workflow/common/common_test.go` rather than the executor package's test file.

workflow/executor/resource.go
```diff
@@ -86,14 +73,42 @@ func (we *WorkflowExecutor) ExecResource(ctx context.Context, action string, man
...
+		}
+		err = argoerrors.Wrap(err, argoerrors.CodeBadRequest, "no more retries "+err.Error())
+	}
+	return out, err
+}
+
+// inferPluralName guesses the lowercase plural resource name for a kind.
+// This is the best guess we can do here and is what `kubectl` uses under the hood. Hopefully future versions of the
+// REST client would remove the need to infer the plural name.
+func inferPluralName(kind string) string {
```
Joibel: **[nit]** `inferPluralName` is extracted but has exactly one caller (`inferObjectSelfLink`, same as before) — the resource agent uses the RESTMapper for pluralization instead. The extraction adds churn without reuse; fine to keep, but it isn't needed for this PR.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -102,17 +102,18 @@ type NodeType string
...
+	NodeTypeContainer       NodeType = "Container"
+	NodeTypeSteps           NodeType = "Steps"
+	NodeTypeStepGroup       NodeType = "StepGroup"
+	NodeTypeDAG             NodeType = "DAG"
+	NodeTypeTaskGroup       NodeType = "TaskGroup"
+	NodeTypeRetry           NodeType = "Retry"
+	NodeTypeSkipped         NodeType = "Skipped"
+	NodeTypeSuspend         NodeType = "Suspend"
+	NodeTypeHTTP            NodeType = "HTTP"
+	NodeTypeResourceMonitor NodeType = "ResourceMonitor"
```
Joibel: **[naming nit]** The node both *creates* and monitors the resource; `ResourceMonitor` undersells the create half. Public API enum value, so worth getting right before it ships.

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3480,12 +3491,23 @@ func (tmpl *Template) GetNodeType() NodeType {
...
 		return true
 	}
 	return false
 }
 
+// ProvidesPodNameVar reports whether the {{pod.name}} variable should be injected into this
+// template's scope. Pod templates always do; resource templates do too (including the agent
+// variant, which is not IsPodType) so that a manifest referencing {{pod.name}} keeps resolving
+// the same way regardless of resource.agent.
+func (tmpl *Template) ProvidesPodNameVar() bool {
```
Joibel, line 3538: **[nit]** For agent-based resource templates this keeps `{{pod.name}}` resolving even though no per-node pod exists, so the variable yields a pod name that never exists. The comment says this is intentional for template-compat with the non-agent path — fair, but consider documenting the semantics (users may pass `{{pod.name}}` to things that expect a real pod).

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,977 @@
...
+	// sit on the queue forever. Clamp to the default, as requeueTime does above.
+	if taskWorkers <= 0 {
+		logger.WithField("taskWorkers", taskWorkers).Warn(ctx, "invalid ARGO_AGENT_TASK_WORKERS; falling back to 16")
+		taskWorkers = 16
+	}
+	logger.WithField("taskWorkers", taskWorkers).
+		WithField("requeueTime", requeueTime).
+		Info(ctx, "Starting Agent")
+
+	taskQueue := make(chan string, 32)
```
Joibel: **[scalability nit]** `taskQueue <- nodeID` inside the informer event handler blocks once >32 tasks are in flight beyond the workers, stalling the WorkflowTaskSet informer delivery goroutine until workers drain (each `createAndWatchResource` can block ~30s on kubectl + cache sync). Backpressure rather than deadlock, and the scale example uses 30 children, but large fan-outs will feel it — a workqueue (as used elsewhere in this file) wouldn't block the handler.

workflow/controller/agent.go
```diff
@@ -151,8 +209,26 @@ func (woc *wfOperationCtx) createAgentPod(ctx context.Context) (*apiv1.Pod, erro
...
+				Value: resync,
+			})
+		}
+	}
 
 	serviceAccountName := woc.execWf.Spec.ServiceAccountName
+	if resourceAgent {
+		// The resource agent's informers need list+watch on whole GVRs — broader than
+		// workflow pods should carry — so it runs under a dedicated service account,
+		// following the `<name>-executor-plugin` convention. Fails if the SA is missing.
```
Joibel: Comment says -executor-plugin. Code says "-resource-agent". Nothing anywhere seems to tell the user that their actual SA they requested is being ignored, validation doesn't validate they haven't specified it, and we'd probably want a runtime "nope" as well.

docs/resource-template.md
Joibel: Still not addressed.

workflow/executor/resource.go
```diff
@@ -390,22 +390,43 @@ func jqFilter(ctx context.Context, input []byte, filter string) (string, error)
...
+// kubectlMutex serializes in-process kubectl: NewKubectlCommand binds flags to kubectl
+// package-level variables (e.g. profiling.go), so concurrent construction/parsing races even
+// with per-call args and IO streams. Callers such as the resource agent's task workers may call
+// runKubectl concurrently; the lock lives here so the invariant cannot drift to call sites.
+var kubectlMutex sync.Mutex
+
+func runKubectl(ctx context.Context, args ...string) (out []byte, err error) {
+	logging.RequireLoggerFromContext(ctx).Info(ctx, strings.Join(args, " "))
+	installKubectlFatalHandler()
+	kubectlMutex.Lock()
```
Joibel, line 416: Doesn't this lock pretty much prevent multiple workers from being useful?

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -102,17 +102,18 @@ type NodeType string
...
+	NodeTypeContainer     NodeType = "Container"
+	NodeTypeSteps         NodeType = "Steps"
+	NodeTypeStepGroup     NodeType = "StepGroup"
+	NodeTypeDAG           NodeType = "DAG"
+	NodeTypeTaskGroup     NodeType = "TaskGroup"
+	NodeTypeRetry         NodeType = "Retry"
+	NodeTypeSkipped       NodeType = "Skipped"
+	NodeTypeSuspend       NodeType = "Suspend"
+	NodeTypeHTTP          NodeType = "HTTP"
+	NodeTypeResourceAgent NodeType = "ResourceAgent"
```
Joibel, line 115: Why do we actually need a new node type - it feels like it should be hidden?

pkg/apis/workflow/v1alpha1/workflow_types.go
```diff
@@ -3414,6 +3415,11 @@ type ResourceTemplate struct {
 	// 	"--validate=false"  # disable resource validation
 	// ]
 	Flags []string `json:"flags,omitempty" protobuf:"varint,7,opt,name=flags"`
+
+	// Agent, when true, runs this resource template on the shared per-workflow resource agent pod
+	// (which creates the resource and watches it for its success/failure conditions) instead of a
+	// dedicated per-node pod. See docs/resource-template.md.
+	Agent bool `json:"agent,omitempty" protobuf:"varint,9,opt,name=agent"`
```
Joibel: I don't like the name "agent" here, nor the boolean rather than an extendable enum.

workflow/executor/resource.go
```diff
@@ -414,11 +435,11 @@ func runKubectl(ctx context.Context, args ...string) ([]byte, error) {
 			WithDiscoveryBurst(300).
 			WithDiscoveryQPS(50.0),
 		IOStreams: genericclioptions.IOStreams{Out: &buf, ErrOut: os.Stderr},
-	}).Execute(); err != nil {
+	})
+	// Execute parses os.Args unless told otherwise (kubectl's own tests set args the same way).
+	cmd.SetArgs(args[1:])
+	if err := cmd.Execute(); err != nil {
```
Joibel: I'm surprised the linter didn't complain. Prefer ExecuteContext(ctx) here.

workflow/executor/resource-agent.go
```diff
@@ -0,0 +1,985 @@
...
+		})
+	}
+}
+
+func (rae *ResourceAgentExecutor) resolveManifest(ctx context.Context, tmpl *wfv1.Template) ([]byte, error) {
+	if tmpl.Resource.ManifestFrom == nil || tmpl.Resource.ManifestFrom.Artifact == nil {
+		return []byte(tmpl.Resource.Manifest), nil
+	}
+	name := tmpl.Resource.ManifestFrom.Artifact.Name
+	art := tmpl.Inputs.GetArtifactByName(name)
```
Joibel, line 543: Can we DRY and reuse the standard executor artifact code here - this looks like a duplicate of that code.

.gitignore
```diff
@@ -9,6 +9,10 @@ dist/
 # delve debug binaries
 cmd/**/debug
 hack/**/debug
+# binaries built in-place (e.g. `go build ./cmd/argoexec` from its own dir)
```
Joibel: Drop this please

Joibel (changes requested):
This needs (probably more than one) e2e testing

It feels like the resource agent should have all artifact plugins installed in it, then they'd all just work.

pkg/apis/workflow/v1alpha1/workflow_types.go
isubasinghe: We definitely need one as we make decisions based upon if its a resource agent or not. 
Like technically we can derive it from the node struct, but none of the other node types do this so it would be kind of weird to do so.

workflow/executor/agent.go
isubasinghe: addressed

docs/resource-template.md
isubasinghe: addressed, but also this is behaviour for the non agent path as well

isubasinghe:
https://github.com/argoproj/argo-workflows/pull/16602/changes probably conflicts with this

PR #16375, opened 2026-07-04
fix(controller): retry workflows awaiting a missing WorkflowTemplate

### LINKED ISSUE #14615: workflowtemplates.argoproj.io not found should be retried for workflows
### Pre-requisites

When deploying both workflow and workflowtemplate in one go, the workflow fails to start because of 

```
xyz-init-workflow   Error    2m3s   workflowtemplates.argoproj.io "xyz-workflow-template" not found
```

This is because the workflow is deployed before the workflowtemplate. But even if we deploy it in the correct order, ArgoWF actually takes some time to process the workflowtemplate and know about it. So it still failed.

Naturally, the retryStrategy should make the workflow retry in such scenarios. But it doesn't do retries.

### Version(s)

v3.4.17

```YAML
kind: Workflow
metadata:
  name: xyz-init-workflow
spec:
  workflowTemplateRef:
    name: xyz-workflow-template
  retryStrategy:
    limit: 3
    retryPolicy: "Always"
```

```text
n/a
```

```text
Workflow does not start.
```

HsiuChuanHsu:
Fixes #14615

### Motivation

A `Workflow` that references a `WorkflowTemplate` via `workflowTemplateRef` is put straight into the terminal `Error` phase with `workflowtemplates.argoproj.io "..." not found` when the template is not yet in the controller's informer cache.

A user would expect `retryStrategy` to cover this, but it does not: template resolution happens in `setExecWorkflow` at the very start of `operate()`, before any node executes, so the workflow reaches the terminal `Error` phase before the retry machinery (`processNodeRetries`) is ever involved.

### Modifications
- **Transient Grace Period**: A missing template (NotFound) is now treated as a `temporary` issue for a set grace period. The Workflow remains in a `Pending` state and automatically `requeues` until the template is ready.
- **Deferred Fast-Fail**: If the template is still missing after the grace period expires (e.g., due to a typo), the workflow will fail permanently, preserving the original fast-fail contract.
- **Configurable Timeout**: Introduced a new controller setting, `workflowTemplateNotFoundGracePeriod` (default: 30s; 0 restores immediate-fail behavior), measured from the Workflow's creation time.
- 

### AI
Claude Sonnet 4.8
- Debug core logic
- Refine comment wording
- Unit tests

## Summary by CodeRabbit

* **New Features**
  * Added a configurable grace period for workflows waiting on a referenced WorkflowTemplate to appear.
  * Workflows can now stay Pending and retry for a short time before failing if the template is temporarily unavailable.

* **Bug Fixes**
  * Reduced noisy error handling when a referenced WorkflowTemplate is not yet ready.
  * Added clearer behavior for immediate failure when the grace period is disabled or expires.

* **Documentation**
  * Updated config documentation to include the new setting and its default behavior.


Joibel:
Please see https://github.com/argoproj/argo-workflows/issues/14615#issuecomment-4894277872

HsiuChuanHsu:
Sorry for missing the message, and thanks for the review.

